PR #32226, opened 2024-02-20
Move PendingResponseTracker into interaction-model

andy31415:
Without this, there is a circular dependency loop between app and interaction-model throgh CommandSender header:

image


andy31415:
On a second view, CommandSender by itself is circular - CommandSender is in app, however the header references im. I think we have to break circularity in commandsender to begin with before being able to fix things. Reverting this to fix command sender header.

PR #32227, opened 2024-02-20
Move CommandSender and PendingResponseTracker to IM.

andy31415:
Previous location of these files was `app` however this created apparent circular dependencies for includes.

### Changes:

- replace include for commandsender to only include the timeouts header instead of the whole IM
- pull commandsender (both header and cpp) into interactionmodel as well as endingResponseTracker.

Before:

image

After:

image


yunhanw-google (approved):

tcarmelveilleux (approved):

src/app/BUILD.gn
```diff
@@ -140,6 +140,8 @@ static_library("interaction-model") {
     "CASEClientPool.h",
     "CASESessionManager.cpp",
     "CASESessionManager.h",
+    "CommandSender.cpp",
+    "CommandSender.h",
```
bzbarsky-apple, line 144: We should probably move the Read/Write clients in here too.  It's odd to have these conceptually similar objects in different places.

andy31415: Based on header linkage I believe this is correct: clients belong here:

image

will make the change.

andy31415: Done and also pulled in `InteractionModelHelper.h` as that was very coupled to IM:

image

bzbarsky-apple (approved):

PR #32233, opened 2024-02-21
Improve DeviceCommissioner::StopPairing behavior

ksperling-apple:
Improve StopPairing behavior by cancelling the current read or invoke interaction to avoid stray callbacks messing up a future commissioning attempt.
    
Also
- Rename SendCommand to SendCommissioningCommand.
- Consistently check for presence of mDefaultCommissioner first
- Move OnBasic* into DeviceCommissioner as private helpers
- Prefix DeviceCommissioner members with 'm'

Darwin-only changes:
- Tidy up MTRError
- Add MTRErrorCodeCancelled
- Add tests for pairing after previous cancellation


src/controller/CHIPDeviceController.cpp
```diff
@@ -1746,17 +1769,12 @@ void DeviceCommissioner::CommissioningStageComplete(CHIP_ERROR err, Commissionin
...
     // Once this stage is complete, reset mDeviceBeingCommissioned - this will be reset when the delegate calls the next step.
     MATTER_TRACE_SCOPE("CommissioningStageComplete", "DeviceCommissioner");
-    if (mDeviceBeingCommissioned == nullptr)
-    {
-        // We are getting a stray callback (e.g. due to un-cancellable
-        // operations) when we are not in fact commissioning anything.  Just
-        // ignore it.
-        return;
-    }
+    VerifyOrDie(mDeviceBeingCommissioned);
```
bzbarsky-apple, line 1772: Are there no commissioning stages where we are waiting for our API consumer to call back into us?  This seems pretty dangerous to me...

src/controller/InvokeInteraction.h
```diff
@@ -81,6 +83,15 @@ InvokeCommandRequest(Messaging::ExchangeManager * aExchangeMgr, const SessionHan
     ReturnErrorOnFailure(commandSender->AddRequestData(commandPath, requestCommandData, timedInvokeTimeoutMs));
     ReturnErrorOnFailure(commandSender->SendCommandRequest(sessionHandle, responseTimeout));
 
+    // If requested by the caller, provide a way to cancel the invoke interaction.
+    if (outCancelFn != nullptr)
+    {
+        *outCancelFn = [rawDecoderPtr = decoder.get(), rawCommandSender = commandSender.get()]() {
+            chip::Platform::Delete(rawCommandSender);
+            chip::Platform::Delete(rawDecoderPtr);
```
bzbarsky-apple, line 91: So what happens if someone calls this cancel function after OnDone?

src/controller/CHIPDeviceController.cpp
ksperling-apple: `NetworkCredentialsReady` and `ICDRegistrationInfoReady` check `mCommissioningStage` and return an error  if called in the wrong state. `ContinueCommissioningAfterDeviceAttestation` checks both the stage and the device explicitly. Are there any others?

I'm not a huge fan of the kind of defensive programming where internal bugs are expected and ignored, but if we had something like `DebugVerifyOrDieOrReturn()` that could be useful here to only enable these checks in debug builds?

src/controller/InvokeInteraction.h
ksperling-apple: Bad things! This is intended to be just for internal use by the device commissioner at this point, but I don't know if there is a good way to enforce that without moving a bunch of code around.

I think medium term `TypedCommandCallback` should maybe go away and we should instead have something more like `ReadClient` (or `CommandSender`) where the client can explicitly manage the object lifetime, and maybe a `Auto...` variant / wrapper that heap-allocates and frees automatically?

ksperling-apple: I've documented `InvokeCancelFn` as Internal, so we don't have to consider it API and retain support for this exact mechanism going forward.

src/darwin/Framework/CHIPTests/MTRPairingTests.m
```diff
@@ -163,6 +163,12 @@ - (void)setUp
     [self setContinueAfterFailure:NO];
 }
 
+- (void)tearDown
+{
+    [sController setDeviceControllerDelegate:(id _Nonnull) nil queue:dispatch_get_main_queue()]; // TODO: do we need a clearDeviceControllerDelegate API?
```
bzbarsky-apple, line 168: Why do we need to clear it?  This isn't something things will need to do normally...  followup to sort this out is ok.

bzbarsky-apple (approved):

src/darwin/Framework/CHIPTests/MTRPairingTests.m
ksperling-apple: It's not strictly necessary here, but leaving stuff like delegates hanging around between tests can be annoying sometimes.

cecille (approved):
The commissioner stuff looks reasonable to me. I know nothing about the darwin stack, so I haven't reviewed that. but I'm assuming that boris did.

src/controller/CHIPDeviceController.cpp
cecille: Unresolveing this not becuase there's something specific to be done, but just so github doesn't collapse it. Feel free to close if you think there's sufficient debug info exposed.

ksperling-apple: With `CHIP_CONFIG_VERBOSE_VERIFY_OR_DIE` defined (which is true for Darwin and Linux among a few others) it translates to
```
    nlABORT_ACTION(aCondition, ChipLogDetail(Support, "VerifyOrDie failure at %s:%d: %s", __FILE__, __LINE__, #aCondition))
```

It's surprising that this is currently at Detail level, I'll change this to Error.

src/controller/CHIPDeviceController.cpp
```diff
@@ -938,6 +939,21 @@ CHIP_ERROR DeviceCommissioner::StopPairing(NodeId remoteDeviceId)
     return CHIP_NO_ERROR;
 }
 
+void DeviceCommissioner::CancelCommissioningInteractions()
+{
+    if (mReadClient)
+    {
+        ChipLogDetail(Controller, "Cancelling read request for step '%s'", StageToString(mCommissioningStage));
+        mReadClient = nullptr; // destructor cancels
```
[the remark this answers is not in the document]
ksperling-apple: Yes, I can change this to `reset()`. Given that smart pointers aren't super common in the code base I agree it's more readable to use methods rather than assignments.

woody-apple (approved):

PR #32236, opened 2024-02-21
[Android] Implement validate Device attestation API

joonhaengHeo:
In order to perform Device Attestation in an app, an API that can be verified is required. 
Additionally, the app must know vendorID and productID, which are the values required for Device Attestation.
To achieve this, we modified some code and added two new APIs.


src/controller/java/AndroidOperationalCredentialsIssuer.cpp
```diff
@@ -47,7 +47,8 @@ static CHIP_ERROR N2J_CSRInfo(JNIEnv * env, jbyteArray nonce, jbyteArray element
 
 static CHIP_ERROR N2J_AttestationInfo(JNIEnv * env, jbyteArray challenge, jbyteArray nonce, jbyteArray elements,
                                       jbyteArray elementsSignature, jbyteArray dac, jbyteArray pai, jbyteArray cd,
-                                      jbyteArray firmwareInfo, jobject & outAttestationInfo);
+                                      jbyteArray firmwareInfo, chip::VendorId vendorId, uint16_t productId,
```
tcarmelveilleux, line 50: Use uint16_t for VendorID, not `chip::VendorId` which is a very partial enum, and is getting phased out.

src/controller/java/src/chip/devicecontroller/ChipDeviceController.java
```diff
@@ -1314,6 +1314,16 @@ public static native byte[] createOperationalCertificate(
    */
   public native byte[] extractSkidFromPaaCert(byte[] paaCert);
 
+  public static native byte[] extractAkidFromPaiCert(byte[] paiCert);
+
+  public static native void validateAttestationInfo(
```
tcarmelveilleux, line 1319: - If this is static, it should be in a new interface + implementation class, not in ChipDeviceController.java directly.
- extractSkidFromPaaCert/extractAkidFromPaiCert should be in a new certificate utilities interface/class, and don't belong here.
- Attestation checks require the attestation challenge, which is not passed here.
- Suggest removing PAA cert from here, and let the implementation of device attestation checks use a method of its choosing to access the PAA set. This will allow better control over revocation, and avoid the caller needing to do a lookup, which, if the attestation check was done over cloud, could be based on a more dynamic PAA list than PAA trust store.
- This method doesn't take a callback, and doesn't return a value. How is the management done of the operation?

src/controller/java/AndroidOperationalCredentialsIssuer.cpp
joonhaengHeo: Modified. Thanks.

src/controller/java/src/chip/devicecontroller/ChipDeviceController.java
joonhaengHeo: - If this is static, it should be in a new interface + implementation class, not in ChipDeviceController.java directly.
- extractSkidFromPaaCert/extractAkidFromPaiCert should be in a new certificate utilities interface/class, and don't belong here.
  -> Yes, I can create a separate class. Would it be appropriate to also move the existing “extractSkidFromPaaCert” method?

- Attestation checks require the attestation challenge, which is not passed here.
- Suggest removing PAA cert from here, and let the implementation of device attestation checks use a method of its choosing to access the PAA set. This will allow better control over revocation, and avoid the caller needing to do a lookup, which, if the attestation check was done over cloud, could be based on a more dynamic PAA list than PAA trust store.
  -> I agree with you. I think it would be better to use AttestationTrustStore, but the problem is occurring because it returns a paa bytearray rather than a callback. Do you know a good solution?

- This method doesn't take a callback, and doesn't return a value. How is the management done of the operation?
 -> Currently, it is implemented to generate an exception if verification fails.

tcarmelveilleux: >-> Yes, I can create a separate class. Would it be appropriate to also move the existing “extractSkidFromPaaCert” method?

Yes

> -> I agree with you. I think it would be better to use AttestationTrustStore, but the problem is occurring because it returns a paa bytearray rather than a callback. Do you know a good solution?

The AttestationTrustStore is an implementation detail that tries to get the associated PAA by given SKI. The issue is that the AttestationTrustStore is called "too early" here --> The validateAttestationInfo should only take inputs from the commissionee (currently untrusted elements) and then do all the necessary internal calls to actually determine if PAA is available. This will also help roll-in revocation support later, which will require even more complex checks against  databases that are not directly  a replica of the DCL.

> - This method doesn't take a callback, and doesn't return a value. How is the management done of the operation?
-> Currently, it is implemented to generate an exception if verification fails.

What about on success?

joonhaengHeo: > The AttestationTrustStore is an implementation detail that tries to get the associated PAA by given SKI. The issue is that the AttestationTrustStore is called "too early" here --> The validateAttestationInfo should only take inputs from the commissionee (currently untrusted elements) and then do all the necessary internal calls to actually determine if PAA is available. This will also help roll-in revocation support later, which will require even more complex checks against databases that are not directly a replica of the DCL.

-> Yes, all processes verified by DefaultDACVerifier have been added to the currently implemented method.

> This method doesn't take a callback, and doesn't return a value. How is the management done of the operation?
> -> Currently, it is implemented to generate an exception if verification fails.
> What about success?

-> Moves to the next line without any action.

joonhaengHeo: @tcarmelveilleux 
I modified in seperate class. please review this.

src/controller/java/AndroidOperationalCredentialsIssuer.cpp
```diff
@@ -492,11 +498,11 @@ CHIP_ERROR N2J_AttestationInfo(JNIEnv * env, jbyteArray challenge, jbyteArray no
     SuccessOrExit(err);
 
     env->ExceptionClear();
-    constructor = env->GetMethodID(infoClass, "<init>", "([B[B[B[B[B[B[B[B)V");
+    constructor = env->GetMethodID(infoClass, "<init>", "([B[B[B[B[B[B[B[BII)V");
     VerifyOrExit(constructor != nullptr, err = CHIP_JNI_ERROR_METHOD_NOT_FOUND);
 
-    outAttestationInfo =
-        (jobject) env->NewObject(infoClass, constructor, challenge, nonce, elements, elementsSignature, dac, pai, cd, firmwareInfo);
+    outAttestationInfo = (jobject) env->NewObject(infoClass, constructor, challenge, nonce, elements, elementsSignature, dac, pai,
```
yunhanw-google, line 504: please use static_cast<jobject>

src/controller/java/src/chip/devicecontroller/AttestationInfo.java
```diff
@@ -11,13 +11,24 @@ public final class AttestationInfo {
   private byte[] pai;
   private byte[] certificationDeclaration;
   private byte[] firmwareInfo;
+  private int vendorId;
```
yunhanw-google, line 14: should these be final?

src/controller/java/src/chip/devicecontroller/DeviceAttestation.java
```diff
@@ -0,0 +1,57 @@
...
+package chip.devicecontroller;
+
+public class DeviceAttestation {
+  /**
+   * Extract skid from paa cert.
+   *
+   * @param paaCert The product attestation authority (PAA) cert
+   * @return The subject key identifier (SKID)
+   */
+  public static native byte[] extractSkidFromPaaCert(byte[] paaCert);
```
yunhanw-google, line 28: private?

yunhanw-google (approved):

src/controller/java/AndroidOperationalCredentialsIssuer.cpp
joonhaengHeo: Changed. Thanks.

PR #32245, opened 2024-02-21
Make MTRServerCluster threadsafe.

bzbarsky-apple:
If two API clients are both touching the same instance of MTRServerCluster on different threads, we should handle that correctly.

### DIFF
--- a/src/darwin/Framework/CHIP/ServerEndpoint/MTRServerCluster.mm
+++ b/src/darwin/Framework/CHIP/ServerEndpoint/MTRServerCluster.mm
@@ -88,10 +88,8 @@ MTR_DIRECT_MEMBERS
     NSArray<NSNumber *> * _generatedCommands;
 
     /**
-     * _lock always protects access to: _accessGrants, _attributes,
-     * _deviceController, _attributeAccessInterface, _matterAttributeMetadata,
-     * _matterAccessGrants, _parentEndpoint, _acceptedCommands,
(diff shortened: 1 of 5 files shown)


yunhanw-google (approved):

src/darwin/Framework/CHIP/ServerEndpoint/MTRServerCluster.mm
```diff
@@ -71,11 +72,28 @@ @implementation MTRServerCluster {
...
+
+    chip::EndpointId _parentEndpoint;
+
+    // _acceptedCommands and _generatedCommands are touched directly by our API
+    // consumer.
+    NSArray<NSNumber *> * _acceptedCommands;
+    NSArray<NSNumber *> * _generatedCommands;
+
+    /**
+     * _lock always protects access to: _accessGrants, _attributes,
```
ksperling-apple, line 91: Is this all the mutable state? Might be easier to document it that way (and any exceptions)

ksperling-apple (approved):
It would be useful to document MTRServerCluster as thread-safe in the header, and remove the "nonatomic" from any properties that are in fact thread-safe.

src/darwin/Framework/CHIP/ServerEndpoint/MTRServerCluster.mm
bzbarsky-apple: This is all the state that gets mutated after init, yes.  I will just say that.

PR #32246, opened 2024-02-21
OnOff Cluster - OFFONLY feature

OmAmbalkar:
- Added Cooktop device type on endpoint 3
- OnOff cluster is enabled with feature OFFONLY
- Only OFF command is supported, other commands will result in unsupported_command


OmAmbalkar:
This PR is opened - https://github.com/project-chip/connectedhomeip/pull/32265

PR #32247, opened 2024-02-21
[ICD] StayActiveRequest

### LINKED ISSUE #24259: [ICD] Implement ICDM cluster StayActiveRequest command handling
### Feature description

Once #24258 is done, the logic behind managing the stay active request can be done.  The manager will need to make sure the device stays in Active mode during the duration of one `ActiveThreshold`  but cannot be forced to stay in active mode indefinitely.

lpbeliveau-silabs:
Implemented the StayActiveRequest command as described in the spec.

Fixes: https://github.com/project-chip/connectedhomeip/issues/24259

### DIFF
--- a/src/app/clusters/icd-management-server/icd-management-server.cpp
+++ b/src/app/clusters/icd-management-server/icd-management-server.cpp
@@ -425,9 +425,14 @@ bool emberAfIcdManagementClusterUnregisterClientCallback(CommandHandler * comman
 bool emberAfIcdManagementClusterStayActiveRequestCallback(CommandHandler * commandObj, const ConcreteCommandPath & commandPath,
                                                           const Commands::StayActiveRequest::DecodableType & commandData)
 {
+// Note: We only need this #if statement for platform examples that enable the ICD management server without building the sample
+// as an ICD. Since this is not spec compliant, we should remove this #if statement once we stop compiling the ICD management
+// server in those examples.
+#if CHIP_CONFIG_ENABLE_ICD_SERVER
     IcdManagement::Commands::StayActiveResponse::Type response;
     response.promisedActiveDuration = Server::GetInstance().GetICDManager().StayActiveRequest(commandData.stayActiveDuration);
     commandObj->AddResponse(commandPath, response);
+#endif // CHIP_CONFIG_ENABLE_ICD_SERVER
     return true;
 }
 
--- a/src/app/clusters/icd-management-server/icd-management-server.cpp
+++ b/src/app/clusters/icd-management-server/icd-management-server.cpp
@@ -364,14 +364,6 @@ void ICDManagementServer::TriggerICDMTableUpdatedEvent()
 
 #endif // CHIP_CONFIG_ENABLE_ICD_CIP
 
-Status ICDManagementServer::StayActiveRequest(FabricIndex fabricIndex)
-{
-    // TODO: Implementent stay awake logic for end device
-    // https://github.com/project-chip/connectedhomeip/issues/24259
-    ICDNotifier::GetInstance().NotifyICDManagementEvent(ICDListener::ICDManagementEvents::kStayActiveRequestReceived);
-    return InteractionModel::Status::UnsupportedCommand;
-}
-
 void ICDManagementServer::Init(PersistentStorageDelegate & storage, Crypto::SymmetricKeystore * symmetricKeystore,
                                ICDConfigurationData & icdConfigurationData)
 {
@@ -433,10 +425,15 @@ bool emberAfIcdManagementClusterUnregisterClientCallback(CommandHandler * comman
 bool emberAfIcdManagementClusterStayActiveRequestCallback(CommandHandler * commandObj, const ConcreteCommandPath & commandPath,
                                                           const Commands::StayActiveRequest::DecodableType & commandData)
 {
-    ICDManagementServer server;
-    InteractionModel::Status status = server.StayActiveRequest(commandObj->GetAccessingFabricIndex());
-
-    commandObj->AddStatus(commandPath, status);
+#if CHIP_CONFIG_ENABLE_ICD_SERVER
+    IcdManagement::Commands::StayActiveResponse::Type response;
+    response.promisedActiveDuration = Server::GetInstance().GetICDManager().StayActiveRequest(commandData.stayActiveDuration);
+    commandObj->AddResponse(commandPath, response);
+#else
+    // Note: We only keep this line around for platforms examples that enable the ICD management server without building the sample
+    // as an ICD.
+    commandObj->AddStatus(commandPath, InteractionModel::Status::UnsupportedCommand);
+#endif // CHIP_CONFIG_ENABLE_ICD_SERVER
     return true;
 }
 
--- a/src/app/clusters/icd-management-server/icd-management-server.cpp
+++ b/src/app/clusters/icd-management-server/icd-management-server.cpp
@@ -425,15 +425,9 @@ bool emberAfIcdManagementClusterUnregisterClientCallback(CommandHandler * comman
 bool emberAfIcdManagementClusterStayActiveRequestCallback(CommandHandler * commandObj, const ConcreteCommandPath & commandPath,
                                                           const Commands::StayActiveRequest::DecodableType & commandData)
 {
-#if CHIP_CONFIG_ENABLE_ICD_SERVER
     IcdManagement::Commands::StayActiveResponse::Type response;
     response.promisedActiveDuration = Server::GetInstance().GetICDManager().StayActiveRequest(commandData.stayActiveDuration);
     commandObj->AddResponse(commandPath, response);
-#else
-    // Note: We only keep this line around for platforms examples that enable the ICD management server without building the sample
-    // as an ICD.
-    commandObj->AddStatus(commandPath, InteractionModel::Status::UnsupportedCommand);
-#endif // CHIP_CONFIG_ENABLE_ICD_SERVER
     return true;
 }
 
--- a/src/app/icd/server/ICDConfigurationData.h
+++ b/src/app/icd/server/ICDConfigurationData.h
@@ -61,7 +61,7 @@ public:
 
     System::Clock::Milliseconds16 GetActiveModeThreshold() { return mActiveThreshold; }
 
-    System::Clock::Milliseconds32 GetMaxStayActiveDuration() { return kMaxStayActiveDuration; }
+    System::Clock::Milliseconds32 GetMaxStayActiveDuration() { return kMaxGuaranteedStayActiveDuration; }
 
     Protocols::SecureChannel::CheckInCounter & GetICDCounter() { return mICDCounter; }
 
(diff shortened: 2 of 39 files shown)


lpbeliveau-silabs:
Added skip backwards compatibility as the command was previously unsupported.

yunhanw-google (approved):

src/app/clusters/icd-management-server/icd-management-server.cpp
```diff
@@ -433,10 +425,15 @@ bool emberAfIcdManagementClusterUnregisterClientCallback(CommandHandler * comman
...
-    ICDManagementServer server;
-    InteractionModel::Status status = server.StayActiveRequest(commandObj->GetAccessingFabricIndex());
-
-    commandObj->AddStatus(commandPath, status);
+#if CHIP_CONFIG_ENABLE_ICD_SERVER
+    IcdManagement::Commands::StayActiveResponse::Type response;
+    response.promisedActiveDuration = Server::GetInstance().GetICDManager().StayActiveRequest(commandData.stayActiveDuration);
+    commandObj->AddResponse(commandPath, response);
+#else
+    // Note: We only keep this line around for platforms examples that enable the ICD management server without building the sample
```
bzbarsky-apple, line 433:
```suggestion
    // Note: We only keep this line around for platform examples that enable the ICD management server without building the sample
```

bzbarsky-apple, line 433: But aren't those non-spec-compliant in that they claim the command in AcceptedCommandList?  I don't understand the setup that causes this code to be needed.

src/app/icd/server/ICDConfigurationData.h
```diff
@@ -123,6 +125,8 @@ class ICDConfigurationData
 
     static constexpr System::Clock::Seconds32 kMaxIdleModeDuration = System::Clock::Seconds32(18 * kSecondsPerHour);
     static constexpr System::Clock::Seconds32 kMinIdleModeDuration = System::Clock::Seconds32(1);
+    // As defined in the spec, the maximum duration for the StayActiveDuration is 30s (spec 7.5.1)
```
bzbarsky-apple, line 128: Which spec?  7.5.1 in the main Matter spec is "7.5.1. Accessing Fabric"...

If you are looking at spec bits directly on github, the section numbers will be variously wrong, generally.

src/app/icd/server/ICDConfigurationData.h
```diff
@@ -123,6 +125,8 @@ class ICDConfigurationData
...
+    static constexpr System::Clock::Milliseconds32 kMaxStayActiveDuration = System::Clock::Milliseconds32(30000);
```
bzbarsky-apple, line 129: So what I see in "9.17.7.5.1. PromisedActiveDuration Field" is that 30s is a _floor_, not a ceiling?

That is, a server can always respond with a PromisedActiveDuration that's as large as it wants.  This thing needs better naming.  Maybe kMaxGuaranteedStayActiveDuration?

src/app/icd/server/ICDManager.h
```diff
@@ -109,6 +109,14 @@ class ICDManager : public ICDListener
     void postObserverEvent(ObserverEventType event);
     OperationalState GetOperationalState() { return mOperationalState; }
 
+    /**
+     * @brief Extends the Active Mode duration for whichever is smallest between 30000 milliseconds and stayActiveDuration, taking
+     * in account the remaining active time.
```
bzbarsky-apple, line 114:
```suggestion
     * @brief Ensures that the remaining Active Mode duration is at least the smaller of 30000 milliseconds and stayActiveDuration.
     *
```

src/app/clusters/icd-management-server/icd-management-server.cpp
lpbeliveau-silabs: It's for the CI. We need to put an `#if` around `response.promisedActiveDuration = Server::GetInstance().GetICDManager()` otherwise they will cause compilation error.

That being said we could just remove the `#else` part and have this comment to explain the `#if` part instead.

lpbeliveau-silabs: From the results observed, the following CI will fail without an if `#i` `GetICDManager`:
Build example - ASR / ASR 
Build example - ESP32 / ESP32
Build example - Infineon / Infineon examples building
Build example - Linux ARM / Linux ARM Cross compile 
Build example - Mbed OS / Mbed OS examples building 
Build example - QPG / QPG 
Build example - TI CC26X2X7 / cc26x2x7  
Build example - TI CC32XX / cc32xx 
Tests / Test Suites - Darwin (no-ble-asan-clang) (pull_request)
Tests / REPL Tests - Linux (pull_request) 
Build example - ESP32 / ESP32_1 (pull_request)
...

From what I see, it seems that the lock app includes the icd management server cluster, without the actual command, but that the compiler still tries to build the emberAfcallbacks.

jmartinez-silabs (approved):

src/app/clusters/icd-management-server/icd-management-server.cpp
```diff
@@ -433,10 +425,14 @@ bool emberAfIcdManagementClusterUnregisterClientCallback(CommandHandler * comman
...
-    ICDManagementServer server;
-    InteractionModel::Status status = server.StayActiveRequest(commandObj->GetAccessingFabricIndex());
-
-    commandObj->AddStatus(commandPath, status);
+// Note: We only need this #if statement for platform examples that enable the ICD management server without building the sample
+// as an ICD. Since this is not spec compliant, we should remove this #if statement once we stop compiling the ICD management
+// server in those examples.
+#if CHIP_CONFIG_ENABLE_ICD_SERVER
+    IcdManagement::Commands::StayActiveResponse::Type response;
+    response.promisedActiveDuration = Server::GetInstance().GetICDManager().StayActiveRequest(commandData.stayActiveDuration);
```
mkardous-silabs, line 433: Can we avoid using the `Server::GetInstance()` here and match what as in the init of the cluster?
pass a pointer to the ICDManager to the ICDManagementServer to centralize the access to server.

src/app/clusters/icd-management-server/icd-management-server.cpp
```diff
@@ -433,10 +425,14 @@ bool emberAfIcdManagementClusterUnregisterClientCallback(CommandHandler * comman
...
                                                           const Commands::StayActiveRequest::DecodableType & commandData)
 {
-    ICDManagementServer server;
-    InteractionModel::Status status = server.StayActiveRequest(commandObj->GetAccessingFabricIndex());
-
-    commandObj->AddStatus(commandPath, status);
+// Note: We only need this #if statement for platform examples that enable the ICD management server without building the sample
+// as an ICD. Since this is not spec compliant, we should remove this #if statement once we stop compiling the ICD management
+// server in those examples.
+#if CHIP_CONFIG_ENABLE_ICD_SERVER
```
mkardous-silabs, line 431: Should we add another build flag for the StayActiveRequest to mimick the CIP and UAT defines?

src/app/icd/server/ICDConfigurationData.h
```diff
@@ -61,6 +61,8 @@ class ICDConfigurationData
 
     System::Clock::Milliseconds16 GetActiveModeThreshold() { return mActiveThreshold; }
 
+    System::Clock::Milliseconds32 GetMaxStayActiveDuration() { return kMaxGuaranteedStayActiveDuration; }
```
mkardous-silabs, line 64: For clarity, i would the `max` bits since it is just the GaranteedStayActiveDuration.

src/app/icd/server/ICDManager.cpp
```diff
@@ -120,6 +120,22 @@ bool ICDManager::SupportsFeature(Feature feature)
 #endif // !CONFIG_BUILD_FOR_HOST_UNIT_TEST
 }
 
+uint32_t ICDManager::StayActiveRequest(uint32_t stayActiveDuration)
+{
+    // This should only be called when the device is in ActiveMode
+    VerifyOrReturnValue(mOperationalState == OperationalState::ActiveMode, 0);
+
+    uint32_t promisedActiveDuration =
+        std::min(static_cast<uint32_t>(ICDConfigurationData::GetInstance().GetMaxStayActiveDuration().count()), stayActiveDuration);
```
mkardous-silabs, line 129: Do you need the static cast? count() should already return a uint32_t no?

src/app/icd/server/ICDManager.h
```diff
@@ -131,6 +139,12 @@ class ICDManager : public ICDListener
     void OnSubscriptionReport() override;
 
 protected:
+    /**
+     * @brief Hepler function that extends the Active Mode duration by the extendDuration parameter
+     *        as well as the Active Mode Jitter timer for the transition to iddle mode.
```
mkardous-silabs, line 144: This comment doesn't reflect what the function done.

This says :
currentTimer = 15s,
extend by 30s
new timer is 45s

when it does
current Timer 15s
extend to 30s
new timer 30s

src/app/tests/TestICDManager.cpp
```diff
@@ -540,6 +540,83 @@ class TestICDManager
...
+    static void TestICDMStayActive(nlTestSuite * aSuite, void * aContext)
+    {
+        TestContext * ctx                    = static_cast<TestContext *>(aContext);
+        ICDNotifier notifier                 = ICDNotifier::GetInstance();
+        ICDConfigurationData & icdConfigData = ICDConfigurationData::GetInstance();
+
+        // Verify That ICDManager starts in Idle
+        NL_TEST_ASSERT(aSuite, ctx->mICDManager.mOperationalState == ICDManager::OperationalState::IdleMode);
+
+        // Trigger a subscription report Put the ICD manager into active mode
```
mkardous-silabs, line 554:
```suggestion
        // Trigger a subscription report. Put the ICD manager into active mode.
```

src/app/tests/TestICDManager.cpp
```diff
@@ -540,6 +540,83 @@ class TestICDManager
...
+        notifier.NotifySubscriptionReport();
+        NL_TEST_ASSERT(aSuite, ctx->mICDManager.mOperationalState == ICDManager::OperationalState::ActiveMode);
+
+        // Advance time by the duration of the stay active request - 1
```
mkardous-silabs, line 558:
```suggestion
        // Advance time by the ActiveModeDuration - 1
```

src/app/tests/TestICDManager.cpp
```diff
@@ -540,6 +540,83 @@ class TestICDManager
...
+        NL_TEST_ASSERT(aSuite, ctx->mICDManager.mOperationalState == ICDManager::OperationalState::ActiveMode);
+
+        uint32_t stayActiveRequestedMs = 20000;
+        // Send a stay active request for 20 seconds
+        uint32_t stayActivePromisedMs = ctx->mICDManager.StayActiveRequest(stayActiveRequestedMs);
+        // confirm the promised time is the same as the requested time
+        NL_TEST_ASSERT(aSuite, stayActivePromisedMs == stayActiveRequestedMs);
+
+        // Advance time by the duration of the stay stayActiveRequestedMs - 1 ms
+        AdvanceClockAndRunEventLoop(ctx, System::Clock::Milliseconds32(stayActiveRequestedMs) - 1_ms);
```
mkardous-silabs, line 570: `_ms` works but it should be `_m32`

src/app/clusters/icd-management-server/icd-management-server.cpp
lpbeliveau-silabs: I don't think so. It makes the flow from enabling this command more complex because you need to enable it in zap, then to ensure the define is here with whatever build flag we add.
If we do not enable it in zap`emberAfIcdManagementClusterStayActiveRequestCallback` will be optimized out as unused and won't take space on the device.
The only reason to ifdef it at all is due to compilation errors when non-icd try to build this cluster.

src/app/clusters/icd-management-server/icd-management-server.cpp
lpbeliveau-silabs: We can, but then we will have more code, another ifdef and be using more RAM for the extra pointer. 

I get these are all of little impact, but what would be the benefit of centralizing the access to the server?

src/app/icd/server/ICDManager.h
lpbeliveau-silabs: Agreed but that behavior is dictated by the SystemLayer Implementation, so I think we cannot indicate the precise behavior here as it is the SystemLayer that decides in the end. I'll just avoid being specific about the exact extension.

mkardous-silabs (approved):

src/app/clusters/icd-management-server/icd-management-server.cpp
andy31415: This sounds like a build/layering/dependency issue. We should try to fix that instead.

lpbeliveau-silabs: @bzbarsky-apple Understood, UnsupportedCOmmand was removed and everything in the callback is ifdefed.
@andy31415 An issue about the need for ifdefing was opened [here](https://github.com/project-chip/connectedhomeip/issues/32321)

PR #32248, opened 2024-02-21
[Messages] Change message duration to uint64

hasty:
This changes the duration on Messages to be uint64 to match changes in the spec.

### DIFF
--- a/data_model/clusters/Messages.xml
+++ b/data_model/clusters/Messages.xml
@@ -216,7 +216,7 @@ Davis, CA 95616, USA
         <quality nullable="true"/>
         <mandatoryConform/>
       </field>
-      <field id="4" name="Duration" type="uint16" default="0">
+      <field id="4" name="Duration" type="uint64" default="0">
         <access fabricSensitive="true"/>
         <quality nullable="true"/>
         <mandatoryConform/>
--- a/examples/tv-app/android/java/MessagesManager.cpp
+++ b/examples/tv-app/android/java/MessagesManager.cpp
@@ -183,10 +183,10 @@ CHIP_ERROR MessagesManager::HandleGetMessages(AttributeValueEncoder & aEncoder)
             }
 
             jfieldID durationField = env->GetFieldID(messageClass, "duration", "I");
-            jint jduration         = env->GetIntField(messageObject, durationField);
+            jlong jduration        = env->GetLongField(messageObject, durationField);
             if (jduration >= 0)
             {
-                message.duration = DataModel::Nullable<uint16_t>(static_cast<uint64_t>(jduration));
+                message.duration = DataModel::Nullable<uint64_t>(static_cast<uint64_t>(jduration));
             }
 
             jfieldID getResponseOptionsField =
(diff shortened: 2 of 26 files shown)


examples/tv-app/tv-common/clusters/messages/MessagesManager.h
```diff
@@ -110,7 +110,7 @@ struct CachedMessage
     const chip::app::Clusters::Messages::MessagePriorityEnum mPriority;
     const chip::BitMask<chip::app::Clusters::Messages::MessageControlBitmap> mMessageControl;
     const chip::app::DataModel::Nullable<uint32_t> mStartTime;
-    const chip::app::DataModel::Nullable<uint16_t> mDuration;
+    const chip::app::DataModel::Nullable<uint64_t> mDuration;
```
chrisdecenzo, line 113: Line 129:
        const chip::app::DataModel::Nullable<uint64_t> & startTime, const chip::app::DataModel::Nullable<uint16_t> & duration,

examples/tv-app/android/java/MessagesManager.cpp
```diff
@@ -186,7 +186,7 @@ CHIP_ERROR MessagesManager::HandleGetMessages(AttributeValueEncoder & aEncoder)
             jint jduration         = env->GetIntField(messageObject, durationField);
             if (jduration >= 0)
             {
-                message.duration = DataModel::Nullable<uint16_t>(static_cast<uint16_t>(jduration));
+                message.duration = DataModel::Nullable<uint16_t>(static_cast<uint64_t>(jduration));
```
chrisdecenzo, line 189: other lines to change in this file due to data type change:

Line 77:
env->GetMethodID(managerClass, "presentMessages", "(Ljava/lang/String;IIJJLjava/lang/String;Ljava/util/HashMap;)Z");

Line 185:
jfieldID durationField = env->GetFieldID(messageClass, "duration", "J");
jlong jduration = env->GetLongField(messageObject, durationField);
if (jduration >= 0)
{
message.duration = DataModel::Nullable<uint64_t>(static_cast<uint64_t>(jduration));
}

Line 340:
jlong jduration = -1;
if (!duration.IsNull())
{
jduration = static_cast(duration.Value());
}

chrisdecenzo (changes requested):

examples/tv-app/tv-common/clusters/messages/MessagesManager.h
hasty: Done.

examples/tv-app/android/java/MessagesManager.cpp
hasty: Done.

chrisdecenzo (approved):

bzbarsky-apple (approved):

data_model/clusters/Messages.xml
```diff
@@ -216,7 +216,7 @@ Davis, CA 95616, USA
         <quality nullable="true"/>
         <mandatoryConform/>
       </field>
-      <field id="4" name="Duration" type="uint16" default="0">
+      <field id="4" name="Duration" type="uint64" default="0">
```
andy31415, line 219: This is based on a scrape of the specification (which was done last week).

It is not meant for manual change. Also looking at my copy of the spec:

image

andy31415 (dismissed):
@hasty  I do not see this change in the spec and this PR tries to manually change an automatically generated file.

Duration says is "in minutes" ... why do we want 64-bit there? Is about 45 days insufficient? even then probably 32-bit would be sufficient...

andy31415:
> @hasty I do not see this change in the spec and this PR tries to manually change an automatically generated file.
> 
> Duration says is "in minutes" ... why do we want 64-bit there? Is about 45 days insufficient? even then probably 32-bit would be sufficient...

I see https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/8909 exists ... however that has changes-requested.

I feel uncomfortable allowing a SDK change before spec changes (very late). @robszewczyk what is the plan for the spec here? I would be ok if we synchronize them at the same time however right before 1.3 it seems dangerous to allow one to diverge before the other.

robszewczyk (approved):

PR #32251, opened 2024-02-22
[Darwin] MTRDevice delegate should be notified when cache is primed with basic info

jtung-apple:
This PR introduces a new MTRDevice delegate callback `-deviceCachePrimed:` so that the delegate knows when the stack has gotten the subscription priming report and has cached basic configuration information about the device, e.g. Descriptor clusters.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice.h
+++ b/src/darwin/Framework/CHIP/MTRDevice.h
@@ -405,6 +405,10 @@ MTR_EXTERN NSString * const MTRDataVersionKey MTR_NEWLY_AVAILABLE;
 
 /**
  * Notifies delegate when the device attribute cache has been primed with initial configuration data of the device
+ *
+ * This is called when the MTRDevice object goes from not knowing the device to having cached the first attribute reports that include basic mandatory information, e.g. Descriptor clusters.
+ *
+ * The intention is that after this is called, the client should be able to call read for mandatory attributes and likely expect non-nil values.
  */
 - (void)deviceCachePrimed:(MTRDevice *)device MTR_NEWLY_AVAILABLE;
 
--- a/src/darwin/Framework/CHIP/MTRDevice.h
+++ b/src/darwin/Framework/CHIP/MTRDevice.h
@@ -403,6 +403,11 @@ MTR_EXTERN NSString * const MTRDataVersionKey MTR_NEWLY_AVAILABLE;
  */
 - (void)deviceBecameActive:(MTRDevice *)device MTR_AVAILABLE(ios(16.4), macos(13.3), watchos(9.4), tvos(16.4));
 
+/**
+ * Notifies delegate when the device attribute cache has been primed with initial configuration data of the device
+ */
+- (void)deviceCachePrimed:(MTRDevice *)device MTR_NEWLY_AVAILABLE;
+
 @end
 
 @interface MTRDevice (Deprecated)
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -503,12 +503,14 @@ typedef NS_ENUM(NSUInteger, MTRDeviceWorkItemDuplicateTypeID) {
     _weakDelegate = [MTRWeakReference weakReferenceWithObject:delegate];
     _delegateQueue = queue;
 
+    // If Check if cache is already primed and client hasn't been informed yet, call the -deviceCachePrimed: callback
+    if (!_delegateDeviceCachePrimedCalled && [self _isCachePrimedWithInitialConfigurationData]) {
+        [self _callDelegateDeviceCachePrimed];
+    }
+
     if (setUpSubscription) {
         [self _setupSubscription];
     }
-
-    // Check if cache is already primed from storage
-    [self _checkIfCacheIsPrimed];
 
     os_unfair_lock_unlock(&self->_lock);
 }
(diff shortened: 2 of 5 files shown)


src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -741,6 +768,12 @@ - (void)_handleReportEnd
     _receivingReport = NO;
     _receivingPrimingReport = NO;
     _estimatedStartTimeFromGeneralDiagnosticsUpTime = nil;
+
+    // First subscription report is priming report
```
bzbarsky-apple, line 772: Shouldn't this go in the "subscription established" callback, instead of guessing things based on "report end"?

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -2147,6 +2180,48 @@ - (void)_removeExpectedValueForAttributePath:(MTRAttributePath *)attributePath e
...
+{
+    os_unfair_lock_assert_owner(&self->_lock);
+
+    // Only send the callback once per lifetime of MTRDevice
+    if (_delegateDeviceCachePrimedCalled) {
+        return;
+    }
+
+    // Check if root node descriptor exists
+    NSDictionary * rootDescriptorPartsListDataValue = _readCache[[MTRAttributePath attributePathWithEndpointID:@(0) clusterID:@(MTRClusterIDTypeDescriptorID) attributeID:@(MTRAttributeIDTypeClusterDescriptorAttributePartsListID)]];
```
bzbarsky-apple, line 2195:
```suggestion
    NSDictionary * rootDescriptorPartsListDataValue = _readCache[[MTRAttributePath attributePathWithEndpointID:@(kRootEndpointId) clusterID:@(MTRClusterIDTypeDescriptorID) attributeID:@(MTRAttributeIDTypeClusterDescriptorAttributePartsListID)]];
```

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -2147,6 +2180,48 @@ - (void)_removeExpectedValueForAttributePath:(MTRAttributePath *)attributePath e
...
+    if (!rootDescriptorPartsListDataValue || ![MTRArrayValueType isEqualToString:rootDescriptorPartsListDataValue[MTRTypeKey]]) {
+        return;
+    }
+    NSArray * partsList = rootDescriptorPartsListDataValue[MTRValueKey];
+    if (![partsList isKindOfClass:[NSArray class]] || !partsList.count) {
+        MTR_LOG_ERROR("%@ unexpected type %@ for parts list %@", self, [partsList class], partsList);
```
bzbarsky-apple, line 2201: So how would we get here?  Only if we mess up our own data structures, right?  Do we need these various guards?

src/darwin/Framework/CHIP/MTRDevice.h
```diff
@@ -403,6 +403,11 @@ MTR_EXTERN NSString * const MTRDataVersionKey MTR_NEWLY_AVAILABLE;
  */
 - (void)deviceBecameActive:(MTRDevice *)device MTR_AVAILABLE(ios(16.4), macos(13.3), watchos(9.4), tvos(16.4));
 
+/**
+ * Notifies delegate when the device attribute cache has been primed with initial configuration data of the device
```
bzbarsky-apple, line 407: I think this needs to more clearly explain what "configuration data" is.  I was expecting this to happen at the end of the subscription priming read, but this can happen any time we have all the descriptor clusters, even if we don't have the other bits yet....

And the fact that _usually_ this will happen at the end of the priming read (if the delegate is added early) but then at a different time if the delegate is added later is a bit odd.

woody-apple: Note, this will only be primed once, ever, after the first subscription post pairing, if the client implements storage. @jtung-apple is working on making that change. This isn't expected to be called more than that.

src/darwin/Framework/CHIP/MTRDevice.mm
woody-apple: Guards are good :)

src/darwin/Framework/CHIP/MTRDevice.mm
jtung-apple: Good point. I've moved the logic there.

src/darwin/Framework/CHIP/MTRDevice.mm
jtung-apple: Ever since the persistence PR, I've felt that we should add more guards around things. I am planning to add more as I touch code around MTRDevice, and also think about failure modes and if something extra needs to be done when issues are detected.

src/darwin/Framework/CHIP/MTRDevice.h
jtung-apple: The change to make this call back only once is done in the latest commit.

woody-apple (approved):

src/darwin/Framework/CHIP/MTRDevice.h
```diff
@@ -403,6 +403,15 @@ MTR_EXTERN NSString * const MTRDataVersionKey MTR_NEWLY_AVAILABLE;
...
+ *
+ * This is called when the MTRDevice object goes from not knowing the device to having cached the first attribute reports that include basic mandatory information, e.g. Descriptor clusters.
+ *
+ * The intention is that after this is called, the client should be able to call read for mandatory attributes and likely expect non-nil values.
```
bzbarsky-apple, line 411: "mandatory" as determined by what?  Will it be able to read the "OnOff" attribute on the "OnOff" cluster?

PR #32253, opened 2024-02-22
Fix crash in Wi-Fi Connect Network

joonhaengHeo:
Fix crash in Wi-Fi ConnectNetwork stage during commissioning.


yunhanw-google (approved):

yunhanw-google (approved):

src/controller/CHIPDeviceController.cpp
```diff
@@ -3041,7 +3041,7 @@ void DeviceCommissioner::PerformCommissioningStep(DeviceProxy * proxy, Commissio
 
         CHIP_ERROR err = CHIP_NO_ERROR;
         ChipLogProgress(Controller, "SendCommand kWiFiNetworkEnable, supportsConcurrentConnection=%d",
-                        params.GetSupportsConcurrentConnection().Value());
+                        params.GetSupportsConcurrentConnection().ValueOr(false));
```
andy31415, line 3044: This makes the log not clear about if `false` or `optional::Missing` is there. We should probably use `%s` and the logic of `params.GetSupportsConcurrentConnection().HasValue() ? (params.GetSupportsConcurrentConnection().Value() ? "true" : "false") : "missing"` (or better place this in some getter function) .... very verbose but does not hide away things.

bzbarsky-apple, line 3044: At this point, shouldn't `params.GetSupportsConcurrentConnection()` have a value based on the earlier read?  Or is that only done by autocommissioner and other commissioners are not ensuring the right (new) invariants are followed?

joonhaengHeo: @bzbarsky-apple 
Yes, you are correct. Of course, crashes do not occur in general sequences. (This occurred during testing using the previous code.) However, if the code has changed and this value can be empty, I uploaded a PR because I thought it would be necessary to prevent a crash due to this code that simply shows the log.

joonhaengHeo: I understand. I modified.

bzbarsky-apple: In that case, it would be better to log separately for the "missing value" case, which should probably be logged as error.

bzbarsky-apple (approved):

chulspro (approved):

src/controller/CHIPDeviceController.cpp
joonhaengHeo: It was already added by andy advise.

PR #32257, opened 2024-02-22
[Python] fixed subscription crash

### LINKED ISSUE #32310: [BUG] python subscription crashed when error happend
TC-IDM-4.3 step 15:

### Reproduction steps

```
***[Initialization]***
sub1 = await dev_ctrl.ReadAttribute(
	nodeid=1,
	attributes=[(0, BasicInformation.Attributes.NodeLabel)],
	reportInterval=(3, 10),
	autoResubscribe=False,		# needs to set False
	keepSubscriptions=True		# First subscription, needs to set True
)
assert isinstance(sub1, SubscriptionTransaction), "Subscribe1 failed."
sub2 = await dev_ctrl.ReadAttribute(
	nodeid=1,
	attributes=[(0, GeneralCommissioning.Attributes.Breadcrumb)],
	reportInterval=(3, 10),
	autoResubscribe=False,  	# needs to set False
	keepSubscriptions=False		# Second subscription, needs to set False
)
assert isinstance(sub2, SubscriptionTransaction), "Subscribe2 failed."
# After the second subscription is set up, the first subscription will be terminated
await asyncio.sleep(15)
sub2.Shutdown()
# sub1 should be terminated
sub1.Shutdown()		# Crash
***[deInitialization]***
```

### Bug prevalence

every time

8fbe0a71deb96ade73282558a583386b80dbcd5a

### Platform

python

### Platform Version(s)

_No response_

### Anything else?

_No response_

tianfeng-yang:
Fixed #32310
1. The end of the `ReadClient` life cycle is `OnDone`. If `OnError` happend, `OnDone` will eventually be executed. In the existing Python code, `delete it` is executed in `OnDone`. At this time, `pCallback` and `pReadClient` (moved from `pychip_ReadClient_Read`) have both been deleted. , After that, if the python application executes `SubscriptionTransaction.Shutdown()` or `delete`, `pychip_ReadClient_Abort` will be executed, which will cause `pCallback` to be deleted again, causing the application to crash. Therefore, when calling `Shuddown()` or `delete` on the python side, you need to determine whether `OnError` happend. If `OnError` happend, you do not need to call `pychip_ReadClient_Abort` again. If not, it is a normal delete behavior on the application side.
2. The destructor of `ReadClient` also describes that after deleting `ReadClient`, `OnDone` will not be executed, so we can directly delete `ReadClient` in the callback of `OnDone`(the life cycle of `ReadClient` has ended), so that there is no need to move the memory of `pReadClient` to `ReadClientCallback` in `pychip_ReadClient_Read`.    
@Notes: There is no problem moving `ReadClient` to `pCallback` here, but `WriteClientCallback` is also not moved, just for consistency reasons.

[Follow the advice](https://github.com/project-chip/connectedhomeip/pull/32257#issuecomment-2061738561)

The following two points are split into PR #32308

1. Subscription support host test mode, align to `chip-tool`
2. Export `GetSubscriptionTimeoutMs` interface to check if the subscription is alive.

### DIFF
--- a/src/controller/python/chip/clusters/Attribute.py
+++ b/src/controller/python/chip/clusters/Attribute.py
@@ -110,9 +110,11 @@ class TypedAttributePath:
         '''Only one of either ClusterType and AttributeType OR Path may be provided.'''
 
         if (self.ClusterType is not None and self.AttributeType is not None) and self.Path is not None:
-            raise ValueError("Only one of either ClusterType and AttributeType OR Path may be provided.")
+            raise ValueError(
+                "Only one of either ClusterType and AttributeType OR Path may be provided.")
         if (self.ClusterType is None or self.AttributeType is None) and self.Path is None:
-            raise ValueError("Either ClusterType and AttributeType OR Path must be provided.")
+            raise ValueError(
+                "Either ClusterType and AttributeType OR Path must be provided.")
 
         # if ClusterType and AttributeType were provided we can continue onwards to deriving the label.
         # Otherwise, we'll need to walk the attribute index to find the right type information.
@@ -373,7 +375,8 @@ class AttributeCache:
             try:
                 decodedData = clusterType.FromDict(
                     data=clusterType.descriptor.TagDictToLabelDict([], self.attributeTLVCache[endpointId][clusterId]))
-                decodedData.SetDataVersion(self.versionList.get(endpointId, {}).get(clusterId))
+                decodedData.SetDataVersion(
+                    self.versionList.get(endpointId, {}).get(clusterId))
                 return decodedData
             except Exception as ex:
                 return ValueDecodeFailure(self.attributeTLVCache[endpointId][clusterId], ex)
@@ -404,12 +407,14 @@ class AttributeCache:
             clusterType = _ClusterIndex[clusterId]
 
             if self.returnClusterObject:
-                endpointCache[clusterType] = handle_cluster_view(endpointId, clusterId, clusterType)
+                endpointCache[clusterType] = handle_cluster_view(
+                    endpointId, clusterId, clusterType)
             else:
                 if clusterType not in endpointCache:
                     endpointCache[clusterType] = {}
                 clusterCache = endpointCache[clusterType]
-                clusterCache[DataVersion] = self.versionList.get(endpointId, {}).get(clusterId)
+                clusterCache[DataVersion] = self.versionList.get(
+                    endpointId, {}).get(clusterId)
 
                 if (clusterId, attributeId) not in _AttributeIndex:
                     #
@@ -419,7 +424,8 @@ class AttributeCache:
                     continue
 
                 attributeType = _AttributeIndex[(clusterId, attributeId)][0]
-                clusterCache[attributeType] = handle_attribute_view(endpointId, clusterId, attributeId, attributeType)
+                clusterCache[attributeType] = handle_attribute_view(
+                    endpointId, clusterId, attributeId, attributeType)
         self._attributeCacheUpdateNeeded.clear()
         return self._attributeCache
 
@@ -428,14 +434,18 @@ class SubscriptionTransaction:
     def __init__(self, transaction: AsyncReadTransaction, subscriptionId, devCtrl):
         self._onResubscriptionAttemptedCb: Callable[[SubscriptionTransaction,
                                                      int, int], None] = DefaultResubscriptionAttemptedCallback
-        self._onAttributeChangeCb: Callable[[TypedAttributePath, SubscriptionTransaction], None] = DefaultAttributeChangeCallback
-        self._onEventChangeCb: Callable[[EventReadResult, SubscriptionTransaction], None] = DefaultEventChangeCallback
-        self._onErrorCb: Callable[[int, SubscriptionTransaction], None] = DefaultErrorCallback
+        self._onAttributeChangeCb: Callable[[
+            TypedAttributePath, SubscriptionTransaction], None] = DefaultAttributeChangeCallback
+        self._onEventChangeCb: Callable[[
+            EventReadResult, SubscriptionTransaction], None] = DefaultEventChangeCallback
+        self._onErrorCb: Callable[[
+            int, SubscriptionTransaction], None] = DefaultErrorCallback
         self._readTransaction = transaction
         self._subscriptionId = subscriptionId
         self._devCtrl = devCtrl
         self._isDone = False
-        self._onResubscriptionSucceededCb: Optional[Callable[[SubscriptionTransaction], None]] = None
+        self._onResubscriptionSucceededCb: Optional[Callable[[
+            SubscriptionTransaction], None]] = None
         self._onResubscriptionSucceededCb_isAsync = False
         self._onResubscriptionAttemptedCb_isAsync = False
 
@@ -460,7 +470,8 @@ class SubscriptionTransaction:
     def OverrideLivenessTimeoutMs(self, timeoutMs: int):
         handle = chip.native.GetLibraryHandle()
         builtins.chipStack.Call(
-            lambda: handle.pychip_ReadClient_OverrideLivenessTimeout(self._readTransaction._pReadClient, timeoutMs)
+            lambda: handle.pychip_ReadClient_OverrideLivenessTimeout(
+                self._readTransaction._pReadClient, timeoutMs)
         )
 
     async def TriggerResubscribeIfScheduled(self, reason: str):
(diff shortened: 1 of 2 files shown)


src/controller/python/chip/clusters/Attribute.py
```diff
@@ -520,6 +520,22 @@ def GetReportingIntervalsSeconds(self) -> Tuple[int, int]:
 
         return minIntervalSec.value, maxIntervalSec.value
 
+    def GetSubscriptionTimeout(self) -> int:
```
andy31415, line 523: Please add unit of measurement. `GetSubscriptionTimeoutMs` and describe it in the documentation as well.

Otherwise typical timeouts in python for things like sleep is "float as seconds"

src/controller/python/chip/clusters/attribute.cpp
```diff
@@ -225,23 +225,25 @@ class ReadClientCallback : public ReadClient::Callback
...
-    void OnDone(ReadClient *) override
+    void OnDone(ReadClient * apReadClient) override
     {
         gOnReadDoneCallback(mAppContext);
 
-        delete this;
+        // SubscriptionType will delete with pychip_ReadClient_Abort
+        if (apReadClient->IsReadType())
+        {
+            delete apReadClient;
```
andy31415, line 235: the change of not having apReadClient as a member of the class anymore is not described in the summary of the PR.

Please explain changes and why in the summary of the PR.

src/controller/python/chip/clusters/attribute.cpp
```diff
@@ -225,23 +225,25 @@ class ReadClientCallback : public ReadClient::Callback
...
     void OnReportEnd() override { gOnReportEndCallback(mAppContext); }
 
-    void OnDone(ReadClient *) override
+    void OnDone(ReadClient * apReadClient) override
     {
         gOnReadDoneCallback(mAppContext);
 
-        delete this;
+        // SubscriptionType will delete with pychip_ReadClient_Abort
+        if (apReadClient->IsReadType())
```
yunhanw-google, line 233: onDone should take care of the readclient release. Abort is called from python application, and it is not called from IM's readClient, when subscription onError is called after backoff expired, onDone is further called, then memory leak for readClient would happen.

yunhanw-google (changes requested):

src/controller/python/chip/clusters/attribute.cpp
```diff
@@ -472,6 +475,22 @@ PyChipError pychip_ReadClient_GetReportingIntervals(ReadClient * pReadClient, ui
     return ToPyChipError(err);
 }
 
+void pychip_ReadClient_GetSubscriptionTimeout(ReadClient * pReadClient, uint32_t * milliSec)
```
yunhanw-google, line 478: when you find timeout happens, the existing abort should be enough for the callback/readclient destroy

src/controller/python/chip/clusters/attribute.cpp
tianfeng-yang: In the original design, when a subscription `OnError` occurs, the memory is released in `OnDone`. If `pychip_ReadClient_Abort` is called again at this time, it will cause a secondary release and cause a crash.

src/controller/python/chip/clusters/attribute.cpp
tianfeng-yang: On the contrary, calling `pychip_ReadClient_Abort` after OnError occurs will result in repeated release of memory

yunhanw-google: I think you should not call pychip_ReadClient_Abort when onError happens....

yunhanw-google: if onError happens, you call abort from application to release callback, then callback's onDone would be problematic...

tianfeng-yang: According to your idea, the python interface needs to be modified. Now `pychip_ReadClient_Abort` is called in `Shutdown`, which will cause a crash when it is released normally after use on the application side.

tianfeng-yang: > if onError happens, you call abort from application to release callback, then callback's onDone would be problematic...in readclient, ondone is called immediately after onerror..

Therefore, add ReadType judgment in `OnDone`. If it is SubscriptionType, hand it over to Abort to release it.

src/controller/python/chip/clusters/Attribute.py
tianfeng-yang: Ok

src/controller/python/chip/clusters/attribute.cpp
tianfeng-yang: Has been restored here

src/controller/python/chip/clusters/Attribute.py
```diff
@@ -585,10 +601,13 @@ def Shutdown(self):
             print("Subscription was already terminated previously!")
             return
 
-        handle = chip.native.GetLibraryHandle()
-        builtins.chipStack.Call(
-            lambda: handle.pychip_ReadClient_Abort(
-                self._readTransaction._pReadClient, self._readTransaction._pReadCallback))
+        # Abort only when no error occurs.
+        # If an error occurs, pReadClient has been deleted internally.
+        if self._readTransaction._resultError is None:
```
tcarmelveilleux, line 606: Please open an issue showing what is currently failing, so that it's clear what is being fixed.

tcarmelveilleux (changes requested):
- [ ] Please open issue related to what the problem is with pychip_ReadClient_Abort
- [ ] Better explain what is the current problem  with callbacks, given that large change to resource management are being done. A failing test case being documented in the issue, and added in this PR in a way that it now passes, with no change to the test, but just changes to the code, would be helpful.
- [ ] Consider splitting the other non-contentious bits to a different PR (e.g. access to GetSubscriptionTimeout).
- [ ] Indicate how this was tested to avoid regressions in certification test scripts

tianfeng-yang:
> * [ ]  Please open issue related to what the problem is with pychip_ReadClient_Abort
> * [ ]  Better explain what is the current problem  with callbacks, given that large change to resource management are being done. A failing test case being documented in the issue, and added in this PR in a way that it now passes, with no change to the test, but just changes to the code, would be helpful.
> * [ ]  Consider splitting the other non-contentious bits to a different PR (e.g. access to GetSubscriptionTimeout).
> * [ ]  Indicate how this was tested to avoid regressions in certification test scripts

Ok, I've split the PR as you suggested and created an Issue for the controversial parts

src/controller/python/chip/clusters/Attribute.py
tianfeng-yang: refer to the first point of comment

tianfeng-yang:
@tcarmelveilleux I have modified it according to your suggestion. Are there any problems now?

src/controller/python/chip/clusters/attribute.cpp
```diff
@@ -455,6 +453,7 @@ void pychip_ReadClient_Abort(ReadClient * apReadClient, ReadClientCallback * apC
     VerifyOrDie(apReadClient != nullptr);
     VerifyOrDie(apCallback != nullptr);
 
+    delete apReadClient;
```
tcarmelveilleux, line 456: This can lead to double-free if Abort is ever called after OnDone. Was this validated? This behavior existed before, however.

tcarmelveilleux (approved):

andy31415 (approved):

yunhanw-google:
@tianfeng-yang I would suggest to use ShutdownSubscriptions from https://github.com/project-chip/connectedhomeip/blob/master/src/app/InteractionModelEngine.cpp#L296, and remove abort API. and leave the below unchanged.
    *pReadClient = readClient.get();
    *pCallback   = callback.get();

    callback->AdoptReadClient(std::move(readClient));

    callback.release();

src/controller/python/chip/clusters/attribute.cpp
tianfeng-yang: It will not be double-free. The subscription will only execute `OnDone` when an error occurs. When calling `pychip_ReadClient_Abort`, it has been confirmed whether an error occurred.

tianfeng-yang:
> @tianfeng-yang I would suggest to use ShutdownSubscriptions from https://github.com/project-chip/connectedhomeip/blob/master/src/app/InteractionModelEngine.cpp#L296, and remove abort API. and leave the below unchanged.
> 
> ```
>     *pReadClient = readClient.get();
>     *pCallback   = callback.get();
> 
>     callback->AdoptReadClient(std::move(readClient));
> 
>     callback.release();
> ```

Changing the API like this is a bit extensive, and I can't confirm whether it will cause a chain reaction.

tianfeng-yang:
[Follow the advice](https://github.com/project-chip/connectedhomeip/pull/32257#issuecomment-2061738561)

andy31415:
NOTE on old PR staleness review: this PR seems to be under active development and could use some reviews.

tcarmelveilleux (approved):

yunhanw-google (approved):

arkq (approved):

PR #32264, opened 2024-02-22
Bump pigweed to latest version

andy31415:
Minimal delta:
- bump pigweed
- this brings in a newer clangtidy, which triggers several EnumCastOutOfRange. Will fix those as followups. Removed that
- clang-format update resulted in a delta from zap-regen. Checked in that file.
- clang-tidy still had 3 minor complains which I fixed: deleted an unused std::string and updated the for loop data types for two for loops.

### DIFF
--- a/scripts/tools/zap/tests/outputs/all-clusters-app/app-templates/endpoint_config.h
+++ b/scripts/tools/zap/tests/outputs/all-clusters-app/app-templates/endpoint_config.h
@@ -302,15 +302,15 @@
             { (uint16_t) 0x0, (uint16_t) 0x0, (uint16_t) 0x7 }, /* ControlMode */                                                  \
                                                                                                                                    \
             /* Endpoint: 1, Cluster: Thermostat (server) */                                                                        \
-            { (uint16_t) 0xA28, (uint16_t) -0x6AB3, (uint16_t) 0x7FFF }, /* OccupiedCoolingSetpoint */                             \
-            { (uint16_t) 0x7D0, (uint16_t) -0x6AB3, (uint16_t) 0x7FFF }, /* OccupiedHeatingSetpoint */                             \
-            { (uint16_t) 0x2BC, (uint16_t) -0x6AB3, (uint16_t) 0x7FFF }, /* MinHeatSetpointLimit */                                \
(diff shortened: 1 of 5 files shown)


arkq (approved):

woody-apple (approved):

src/lib/dnssd/minimal_mdns/core/RecordWriter.cpp
```diff
@@ -41,7 +41,7 @@ RecordWriter & RecordWriter::WriteQName(const FullQName & qname)
     size_t qNameWriteStart = mOutput->WritePos();
     bool isFullyCompressed = true;
 
-    for (uint16_t i = 0; i < qname.nameCount; i++)
+    for (uint16_t i = 0; i < static_cast<uint16_t>(qname.nameCount); i++)
```
bzbarsky-apple, line 44: So we do the following with `i`:

1. Add it to the `names` pointer.
2. Add it to `nameCount`, which is size_t.
3. Use it to index into `names` (which is again adding it to pointer).

Is there a reason `i` is not just `size_t`?

scripts/tools/zap/tests/outputs/all-clusters-app/app-templates/endpoint_config.h
```diff
@@ -302,15 +302,15 @@
             { (uint16_t) 0x0, (uint16_t) 0x0, (uint16_t) 0x7 }, /* ControlMode */                                                  \
                                                                                                                                    \
             /* Endpoint: 1, Cluster: Thermostat (server) */                                                                        \
-            { (uint16_t) 0xA28, (uint16_t) -0x6AB3, (uint16_t) 0x7FFF }, /* OccupiedCoolingSetpoint */                             \
```
bzbarsky-apple, line 305: I assume this is due to clang-format getting updated?

bzbarsky-apple (approved):

scripts/tools/zap/tests/outputs/all-clusters-app/app-templates/endpoint_config.h
andy31415: Yes, this is unfortunate :(

src/lib/dnssd/minimal_mdns/core/RecordWriter.cpp
andy31415: Updated.

p0fi (approved):

PR #32265, opened 2024-02-22
OnOff Cluster - OFFONLY feature example

OmAmbalkar:
Added Cooktop device type on endpoint 3
OnOff cluster is enabled with feature OFFONLY
Only OFF command is supported, other commands will result in unsupported_command


andy31415:
This is an old PR in draft form that has merge conflicts. Assuming stale and closing.

PR #32270, opened 2024-02-22
Move `declare_args` sections from `data_model` into AppConfig.h

andy31415:
Moved these arguments to a central place rather than placing them into the data_model.gni 

Placement in data_model gni has several drawbacks
  - declare_args become nested and harder to find
  - they were passed as compiler command-line arguments, so they are harder to integrate in alternate build systems
  - we have multiple data_model invocations, so having multiple declare_args seems potentially odd scoping (they are likely not scoped, however this is not obvious given the data_model invocation with a name.


src/app/BUILD.gn
```diff
@@ -43,6 +43,15 @@ declare_args() {
 
   enable_eventlist_attribute = false
 
+  # Allow building ota-requestor-app with a non-spec-compliant floor
+  # (i.e. smaller than 2 minutes) for action delays.
+  non_spec_compliant_ota_action_delay_floor = -1
+
+  # enable type sync client for use in `time-synchronization-server` (if linked)
```
bzbarsky-apple, line 50:
```suggestion
  # enable time sync client for use in `time-synchronization-server` (if linked)
```

src/app/clusters/ota-requestor/DefaultOTARequestorDriver.cpp
bzbarsky-apple, line 1: This file needs to include the relevant config header now that this is not using a command-line compiler arg, no?

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
```diff
@@ -37,10 +34,12 @@
 #include <platform/CHIPDeviceLayer.h>
 #include <platform/RuntimeOptionsProvider.h>
 
-#include <app-common/zap-generated/cluster-enums.h>
-
 #include <system/SystemClock.h>
 
+#if TIME_SYNC_ENABLE_TSC_FEATURE
```
bzbarsky-apple, line 39: Is this including the relevant config header?  I don't see it.

bzbarsky-apple (changes requested):

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
andy31415: the define is used in time-synchronization-server which includes the right include. You want to include it again?

src/app/clusters/ota-requestor/DefaultOTARequestorDriver.cpp
andy31415: I missed this one. Added it back.

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
bzbarsky-apple: If time-synchronization-server.h is including it, probably ok.

bzbarsky-apple (approved):

yunhanw-google (approved):

PR #32283, opened 2024-02-23
Use host toolchain for all mac and ios targets

ksperling-apple:
target_cpu is irrelevant for toolchain selection on these platforms.


bzbarsky-apple (approved):

ksperling-apple:

PR #32285, opened 2024-02-23
[Android] Support General value TLV parsing

### LINKED ISSUE #30564: [BUG] Android java lib not decoding custom attribute correctly
### Reproduction steps

Here it is in the report: @yunhanw-google 
image
Here it is in the Kotlin report variable:
image

JSON and TLV both have the 2, but valueObject is null, it should be 2.

Example of correctly decoded standard attribute:
image

### Bug prevalence

always

181b0cb14ff007ec912f2ba6627e05dfb066c008

### Platform

android

### Platform Version(s)

_No response_

### Anything else?

_No response_

joonhaengHeo:
Fix #30564 

Even if the result of DecodeAttributeValue is a null value, TLV parsing is possible for Integer, Float, Boolean, String, and byteArray.


src/controller/java/AndroidCallbacks.cpp
```diff
@@ -907,6 +915,77 @@ void InvokeCallback::ReportError(const char * message, ChipError::StorageType er
...
+    CHIP_ERROR err   = CHIP_NO_ERROR;
+    jobject retValue = nullptr;
+
+    TLV::TLVReader readerForJavaObject;
+    readerForJavaObject.Init(*apData);
+
+    switch (readerForJavaObject.GetType())
+    {
+    case TLV::kTLVType_SignedInteger: {
+        int64_t signedValue;
```
yunhanw-google, line 929: please set the initial value. thanks

src/controller/java/AndroidCallbacks.cpp
```diff
@@ -907,6 +915,77 @@ void InvokeCallback::ReportError(const char * message, ChipError::StorageType er
...
+        VerifyOrReturnValue(readerForJavaObject.Get(signedValue) == CHIP_NO_ERROR, nullptr,
+                            ChipLogProgress(Controller, "Get TLV Value fail!"));
+        err = JniReferences::GetInstance().CreateBoxedObject<jlong>("java/lang/Long", "(J)V", static_cast<jlong>(signedValue),
+                                                                    retValue);
+        VerifyOrReturnValue(err == CHIP_NO_ERROR, nullptr, ChipLogProgress(Controller, "Create Boxed Object fail!"));
+        return retValue;
+    }
+    case TLV::kTLVType_UnsignedInteger: {
+        uint64_t unsignedValue;
```
yunhanw-google, line 938: please set the initial value. thanks

yufengwangca (approved):

yunhanw-google (approved):

src/controller/java/AndroidCallbacks.cpp
```diff
@@ -907,6 +915,77 @@ void InvokeCallback::ReportError(const char * message, ChipError::StorageType er
...
+        std::unique_ptr<uint8_t[]> buffer = std::unique_ptr<uint8_t[]>(new uint8_t[bufferLen + 1]);
+        err                               = readerForJavaObject.GetBytes(buffer.get(), bufferLen + 1);
+        VerifyOrReturnValue(err == CHIP_NO_ERROR, nullptr, ChipLogProgress(Controller, "Get TLV Value fail!"));
+
+        jbyteArray valueByteArray = env->NewByteArray(static_cast<jsize>(bufferLen));
+        env->SetByteArrayRegion(valueByteArray, 0, static_cast<jsize>(bufferLen), reinterpret_cast<const jbyte *>(buffer.get()));
+
+        return static_cast<jobject>(valueByteArray);
+    }
+    default:
```
yunhanw-google, line 984: should we handle struct and array?, and dump it into json?

andy31415: It seems we return pure objects here ... it sounds like list/array should be a list as well. Unsure about structures though.

src/controller/java/AndroidCallbacks.cpp
joonhaengHeo: Modified. Thanks.

src/controller/java/AndroidCallbacks.cpp
joonhaengHeo: Modified. Thanks.

src/controller/java/AndroidCallbacks.cpp
joonhaengHeo: @yunhanw-google 
The purpose of this modification is to transmit at least some values of the Custom Attribute through TLV Parsing. As andy31415 commented, it may be possible to implement an Array/List with a simple value, but I think it will be difficult to implement anything else.

yunhanw-google: thanks!

PR #32289, opened 2024-02-23
Re-enable `clang-tidy` check for enum value range validation

andy31415:
Some minimal fixes (e.g. adding equality operator for BitFlags) and in many case I just flagged things as nolint so that we can iterate over fixes.

enum checks also somehow made `cast through void` detection much stricter, so I updated all the instances of that in linux platform (all DBUS casts seem to go through void, and I found no good way to mass-disable without disabling entire platform from all tidy checks)


src/app/icd/server/ICDManager.cpp
```diff
@@ -391,6 +391,8 @@ void ICDManager::SetKeepActiveModeRequirements(KeepActiveFlags flag, bool state)
 {
     assertChipStackLockedByCurrentThread();
 
+    // TODO: fix lint: https://github.com/project-chip/connectedhomeip/issues/32249
```
bzbarsky-apple, line 394: That dos not look like the right link.

src/app/icd/server/ICDManager.cpp
```diff
@@ -511,6 +513,9 @@ void ICDManager::OnActiveRequestWithdrawal(KeepActiveFlags request)
     {
         // Only 1 request per type (kCommissioningWindowOpen, kFailSafeArmed)
         // remove requirement directly
+
+        // TODO: fix lint: https://github.com/project-chip/connectedhomeip/issues/32249
```
bzbarsky-apple, line 517: Wrong link?

src/crypto/tests/TestChipCryptoPAL.cpp
```diff
@@ -2692,6 +2692,8 @@ static void TestVIDPID_StringExtraction(nlTestSuite * inSuite, void * inContext)
         { DNAttrType::kMatterPID, ByteSpan(reinterpret_cast<const uint8_t *>(sTestMatterAttribute14), strlen(sTestMatterAttribute14)), false, false, chip::VendorId::NotSpecified, 0, CHIP_ERROR_WRONG_CERT_DN },
         // Common Name (CN) VID/PID encoding examples:
         { DNAttrType::kCommonName, ByteSpan(reinterpret_cast<const uint8_t *>(sTestCNAttribute01), strlen(sTestCNAttribute01)), true, false, chip::VendorId::TestVendor1, 0, CHIP_NO_ERROR },
+        // TODO: fix lint: https://github.com/project-chip/connectedhomeip/issues/32249
```
bzbarsky-apple, line 2695: Wrong link?

src/lib/core/TLVReader.cpp
```diff
@@ -122,6 +122,9 @@ TLVType TLVReader::GetType() const
         return kTLVType_FloatingPointNumber;
     if (elemType == TLVElementType::NotSpecified || elemType >= TLVElementType::Null)
         return static_cast<TLVType>(elemType);
+
+    // TODO: fix lint: https://github.com/project-chip/connectedhomeip/issues/32249
```
bzbarsky-apple, line 126: All these issue links are wrong....

PR #32291, opened 2024-02-23
Replace hardcoded assumption with one dynamically calculated

tehampson:
Unit test were making a hardcoded assumption of the overhead required by AddRequestData. This was replaced with method that dynamically calculates the overhead

### DIFF
--- a/src/app/tests/TestCommandInteraction.cpp
+++ b/src/app/tests/TestCommandInteraction.cpp
@@ -379,7 +379,8 @@ private:
     static void AddInvokeResponseData(nlTestSuite * apSuite, void * apContext, CommandHandler * apCommandHandler,
                                       bool aNeedStatusCode, CommandId aResponseCommandId = kTestCommandIdWithData,
                                       CommandId aRequestCommandId = kTestCommandIdWithData);
-    static uint32_t GetAddResponseDataOverheadSizeForPath(nlTestSuite * apSuite, const ConcreteCommandPath & aRequestCommandPath, bool aIsForcedSizeBufferLargerThan255);
+    static uint32_t GetAddResponseDataOverheadSizeForPath(nlTestSuite * apSuite, const ConcreteCommandPath & aRequestCommandPath,
+                                                          bool aIsForcedSizeBufferLargerThan255);
     static void FillCurrentInvokeResponseBuffer(nlTestSuite * apSuite, CommandHandler * apCommandHandler,
                                                 const ConcreteCommandPath & aRequestCommandPath, uint32_t aSizeToLeaveInBuffer);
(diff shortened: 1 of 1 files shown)


yunhanw-google (approved):

src/app/tests/TestCommandInteraction.cpp
```diff
@@ -577,6 +579,36 @@ void TestCommandInteraction::AddInvokeResponseData(nlTestSuite * apSuite, void *
...
+                                                                       bool aIsForcedSizeBufferLargerThan255)
+{
+    BasicCommandPathRegistry<4> mBasicCommandPathRegistry;
+    CommandHandler commandHandler(kCommandHandlerTestOnlyMarker, &mockCommandHandlerDelegate, &mBasicCommandPathRegistry);
+    commandHandler.mReserveSpaceForMoreChunkMessages = true;
+    ConcreteCommandPath requestCommandPath1          = { kTestEndpointId, kTestClusterId, kTestCommandIdFillResponseMessage };
+    ConcreteCommandPath requestCommandPath2          = { kTestEndpointId, kTestClusterId, kTestCommandIdCommandSpecificResponse };
+    Optional<uint16_t> commandRef;
+    commandRef.SetValue(1);
+    mBasicCommandPathRegistry.Add(requestCommandPath1, commandRef);
```
andy31415, line 593: nit this seems to indicate a bit of an API deficiency ... I would have imagined we would want to support `Add(..., MakeOptional<uint16_t>(1))` Maybe an add with `&&` would work.

Could we maybe NL_TEST_ASSERT CHIP_NO_ERROR here since this returns an error code?

src/app/tests/TestCommandInteraction.cpp
```diff
@@ -577,6 +579,36 @@ void TestCommandInteraction::AddInvokeResponseData(nlTestSuite * apSuite, void *
...
+    commandRef.SetValue(2);
+    mBasicCommandPathRegistry.Add(requestCommandPath2, commandRef);
+
+    CHIP_ERROR err = commandHandler.AllocateBuffer();
+    NL_TEST_ASSERT(apSuite, err == CHIP_NO_ERROR);
+    uint32_t remainingSizeBefore = commandHandler.mInvokeResponseBuilder.GetWriter()->GetRemainingFreeLength();
+
+    // When ForcedSizeBuffer exceeds 255, an extra byte is needed for length, affecting the overhead size required by
+    // AddResponseData.
+    uint32_t sizeOfForcedSizeBuffer = aIsForcedSizeBufferLargerThan255 ? 256 : 0;
```
andy31415, line 603: this bool is not ideal in reading ... a `GetAddResponseDataOverheadSizeForPath(apSuite, aRequestCommandPath, true)` is not intuitive.

How about an enum class with kEmpty/kLarge or hard-coding the sizes and passing those around?

andy31415 (approved):

PR #32305, opened 2024-02-24
DM XMLs: Remove some in-progress that are not going to cert.

woody-apple:
Removing LIT-icd, aliro, per device credentials


cecille:
I'm going to land a clean scrape on here from the official 1.3 tag to match the 1.3 branch. Let's close this and not waste time trying to get it in, given that it's going to be overridden.

PR #32306, opened 2024-02-25
Updated OnOff Test script according most recent test plan.

beckerzito:
This PRs updates the OnOff test scripts according to the latest Test Plan version. It's necessary for SVE 1.3.
The Cluster Revision was bumped to 6. The example app was also updated.


ReneJosefsen (approved):

src/app/tests/suites/certification/Test_TC_OO_1_1.yaml
```diff
@@ -38,38 +38,75 @@ tests:
...
-      PICS: ( !OO.S.F00 && !OO.S.F01 )
+      PICS: ( OO.S.F00 && !OO.S.F02 )
       response:
-          value: 0
           constraints:
               type: bitmap32
-
+              hasMasksSet: [0x1]
     - label:
           "Step 3b: Given OO.S.F00(LT) ensure featuremap has the correct bit set"
```
tehampson, line 55: Is this label right? Seems like this should be different

src/app/tests/suites/certification/Test_TC_OO_1_1.yaml
```diff
@@ -38,38 +38,75 @@ tests:
...
+      PICS: ( OO.S.F01 && !OO.S.F02 )
       response:
           constraints:
               type: bitmap32
               hasMasksSet: [0x2]
+    - label:
+          "Step 3d: Given OO.S.F01(DF) ensure featuremap has the correct bit set"
+      command: "readAttribute"
+      attribute: "FeatureMap"
+      PICS: ( !( OO.S.F01 && !OO.S.F02 ) )
```
tehampson, line 77: Same, why not split into 2 for easier understanding

src/app/tests/suites/certification/Test_TC_OO_1_1.yaml
```diff
@@ -38,38 +38,75 @@ tests:
...
           constraints:
               type: bitmap32
-
+              hasMasksSet: [0x1]
     - label:
           "Step 3b: Given OO.S.F00(LT) ensure featuremap has the correct bit set"
       command: "readAttribute"
       attribute: "FeatureMap"
-      PICS: OO.S.F00
+      PICS: ( !( OO.S.F00 && !OO.S.F02 ) )
```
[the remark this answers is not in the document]
ReneJosefsen: As I understand this, it is to cover the "non-valid" case, so the conformances are somewhat linked. I didn't fully get this in my first pass on this, but I think I managed to make it make sense in my mind. This is becoming quite complex not though 😔

src/app/tests/suites/certification/Test_TC_OO_1_1.yaml
beckerzito: Correct Rene! The TP requires: "SHALL be 1 if and only if". The previous implementation was checking that if the feature is enabled, then the bit is 1, but wasn't checking the opposite, for example, the case where the feature is disabled on the device, but still, the feature map sets the bit as 1 (which is not allowed).

src/app/tests/suites/certification/Test_TC_OO_1_1.yaml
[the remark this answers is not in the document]
beckerzito: Correct Rene! The TP requires: "SHALL be 1 if and only if". The previous implementation was checking that if the feature is enabled, then the bit is 1, but wasn't checking the opposite, for example, the case where the feature is disabled on the device, but still, the feature map sets the bit as 1 (which is not allowed).

So the strategy here was to keep exactly the opposite from the previous checks to make clear the link with the test plan.

src/app/tests/suites/certification/Test_TC_OO_1_1.yaml
beckerzito: please check other responses above.

src/app/tests/suites/certification/Test_TC_OO_1_1.yaml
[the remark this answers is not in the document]
ReneJosefsen: > wasn't checking the opposite, for example, the case where the feature is disabled on the device, but still, the feature map sets the bit as 1 (which is not allowed)
I think this is a general issue with most of the 1.1 YAMLs to be honest.

src/app/tests/suites/certification/Test_TC_OO_1_1.yaml
ReneJosefsen: We should properly figure out what the plan is regarding the 1.1 tests. I know there have been talks about trying to remove these, and replace them by a single unified test instead. 

Also in case we really want to keep these and want to actually make them verify these cases, I'm wondering if it would not be worth simply rewriting these in Python, which I would assume would make them quite more readable.

beckerzito: I agree, python would be the best case in this scenario!

src/app/tests/suites/certification/Test_TC_OO_2_1.yaml
```diff
@@ -52,22 +52,70 @@ tests:
...
+          value: 30
+
+    - label:
+          "Step 7: TH writes the OnTime attribute as 'initial-On-time' to
+          restore the original value"
+      PICS: OO.S.A4001
+      command: "writeAttribute"
+      attribute: "OnTime"
+      arguments:
+          value: initialOnTime
```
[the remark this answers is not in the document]
beckerzito: my suggestion is to change on the TP once adding hyphen to variable names are not common.

src/app/tests/suites/certification/Test_TC_OO_2_2.yaml
```diff
@@ -57,6 +73,14 @@ tests:
       PICS: OO.S.C01.Rsp
       command: "On"
 
+    - label: "Wait 1000ms"
```
cecille, line 76: Do you know this is going to be sufficient for all the DUTs? Is there a limit to how long a device can take to process these commands?

src/app/tests/suites/certification/ci-pics-values
```diff
@@ -1211,6 +1211,7 @@ OO.S.C41.Rsp=1
 OO.S.C42.Rsp=1
 OO.S.F00=1
 OO.S.F01=0
+OO.S.F02=0
```
cecille, line 1214: Have you run these tests against any DUTs that have this feature enabled?

cecille (approved):

src/app/tests/suites/certification/Test_TC_OO_2_2.yaml
beckerzito: There's no a clear definition for that on the TP. my suggestion is that might be turned into a PIXIT variable in future. For now I just leverage the same 1 seconds that some other steps were already waiting and made it standard for all steps.

src/app/tests/suites/certification/ci-pics-values
beckerzito: There's no example app for now. We aligned that the test script would be urgent for SVE right now, but the example app would be available for v1.4.

There's a PR opened by Om https://github.com/project-chip/connectedhomeip/pull/32265, in which he's working on that... but we still need to do some updates there.

Originally we thought on implementing that on the all-cluster-app, but that will create some complexity to implement. So we might end up creating a new example application.

woody-apple (approved):

PR #32309, opened 2024-02-26
Add equality operator for ConcreteDataAttributePath

mwswartwout:
Add an equality operator for ConcreteDataAttributePath.

### DIFF
--- a/src/app/ConcreteAttributePath.h
+++ b/src/app/ConcreteAttributePath.h
@@ -121,13 +121,11 @@ struct ConcreteDataAttributePath : public ConcreteAttributePath
 
     ConcreteDataAttributePath(EndpointId aEndpointId, ClusterId aClusterId, AttributeId aAttributeId,
                               const Optional<DataVersion> & aDataVersion) :
-        ConcreteAttributePath(aEndpointId, aClusterId, aAttributeId),
-        mDataVersion(aDataVersion)
+        ConcreteAttributePath(aEndpointId, aClusterId, aAttributeId), mDataVersion(aDataVersion)
     {}
 
     ConcreteDataAttributePath(EndpointId aEndpointId, ClusterId aClusterId, AttributeId aAttributeId, ListOperation aListOp,
-                              uint16_t aListIndex) :
-        ConcreteAttributePath(aEndpointId, aClusterId, aAttributeId)
+                              uint16_t aListIndex) : ConcreteAttributePath(aEndpointId, aClusterId, aAttributeId)
     {
         mListOp    = aListOp;
         mListIndex = aListIndex;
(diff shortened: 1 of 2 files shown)


yunhanw-google (approved):

tehampson (approved):

src/app/tests/TestConcreteAttributePath.cpp
```diff
@@ -77,6 +115,14 @@ const nlTest sTests[] = {
                 TestConcreteDataAttributePathMatchesConcreteAttributePathEquality),
     NL_TEST_DEF("TestConcreteDataAttributePathMatchesConcreteAttributePathInequality",
                 TestConcreteDataAttributePathMatchesConcreteAttributePathInequality),
+    NL_TEST_DEF("TestConcreteDataAttributePathEqualityDefaultConstructor", TestConcreteDataAttributePathEqualityDefaultConstructor),
```
andy31415, line 118: could we get some inequality tests too? Right now this seems to pass if we implement ` bool operator==(...) { return true; }`

src/app/ConcreteAttributePath.h
```diff
@@ -144,9 +144,15 @@ struct ConcreteDataAttributePath : public ConcreteAttributePath
 
     bool MatchesConcreteAttributePath(const ConcreteAttributePath & aOther) { return ConcreteAttributePath::operator==(aOther); }
 
-    bool operator==(const ConcreteDataAttributePath & aOther) const = delete;
-    bool operator!=(const ConcreteDataAttributePath & aOther) const = delete;
-    bool operator<(const ConcreteDataAttributePath & aOther) const  = delete;
+    bool operator==(const ConcreteDataAttributePath & aOther) const
+    {
+        return ConcreteAttributePath::operator==(aOther) && (mListIndex == aOther.mListIndex) && (mListOp == aOther.mListOp) &&
```
andy31415, line 149: documentation says that mList Index is only valid if mListOp is a list operation, otherwise it is to be ignored.

This seems to imply this logic should apply to equality operators too (or do we guarantee it is 0 if list operation is NotList)?

bzbarsky-apple (approved):

src/app/tests/TestConcreteAttributePath.cpp
mwswartwout: Done.

src/app/ConcreteAttributePath.h
mwswartwout: This is guaranteed by our constructors (mListOp and mListIndex are only ever initialized together). But since they are public members there's nothing prevent some calling code from creating an invalid ConcreteDataAttributePath directly. If that's a concern I think it's worth refactoring this struct so that invalid states are unrepresentable.

PR #32311, opened 2024-02-26
DNSSD is registering, also can try dns resolve

tianfeng-yang:
1. [Linux] Avahi is registering, there is no need to reset
2. [Linux] Avahi collision, needn't restart dnssd, it depends on the HOST resolve service
3. DNSSD is registering, also can try dns resolve. There are differences between platforms.

### DIFF
--- a/src/lib/dnssd/Discovery_ImplPlatform.cpp
+++ b/src/lib/dnssd/Discovery_ImplPlatform.cpp
@@ -629,9 +629,10 @@ bool DiscoveryImplPlatform::IsInitialized()
 
 CHIP_ERROR DiscoveryImplPlatform::ResolveNodeId(const PeerId & peerId)
 {
-    // Resolve requests can only be issued once DNSSD is initialized and there is
-    // no caching currently
-    // When the DNSSD is kInitializing, can also try to resolve.
+    // It only determines that it is not in an uninitialized state, and cannot force initialization to be completed.
+    // There are differences between platforms.
+    // e.g. when a collision occurs on the Linux platform, initialization has not been completed at this time, but you can
+    // still try to parse.
     VerifyOrReturnError(mState != State::kUninitialized, CHIP_ERROR_INCORRECT_STATE);
 
     ChipLogProgress(Discovery, "Resolving " ChipLogFormatX64 ":" ChipLogFormatX64 " ...",
--- a/src/lib/dnssd/Discovery_ImplPlatform.cpp
+++ b/src/lib/dnssd/Discovery_ImplPlatform.cpp
@@ -631,7 +631,8 @@ CHIP_ERROR DiscoveryImplPlatform::ResolveNodeId(const PeerId & peerId)
 {
     // Resolve requests can only be issued once DNSSD is initialized and there is
     // no caching currently
-    VerifyOrReturnError(mState == State::kInitialized, CHIP_ERROR_INCORRECT_STATE);
+    // When the DNSSD is kInitializing, can also try to resolve.
+    VerifyOrReturnError(mState != State::kUninitialized, CHIP_ERROR_INCORRECT_STATE);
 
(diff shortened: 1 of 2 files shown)


src/lib/dnssd/Discovery_ImplPlatform.cpp
```diff
@@ -631,7 +631,8 @@ CHIP_ERROR DiscoveryImplPlatform::ResolveNodeId(const PeerId & peerId)
 {
     // Resolve requests can only be issued once DNSSD is initialized and there is
     // no caching currently
-    VerifyOrReturnError(mState == State::kInitialized, CHIP_ERROR_INCORRECT_STATE);
+    // When the DNSSD is kInitializing, can also try to resolve.
+    VerifyOrReturnError(mState != State::kUninitialized, CHIP_ERROR_INCORRECT_STATE);
```
andy31415, line 635: Could you point to some documentation stating this is ok? are all platforms ok to issue requests before async initialization completed?

src/platform/Linux/DnssdImpl.cpp
```diff
@@ -383,13 +383,19 @@ void MdnsAvahi::HandleClientState(AvahiClient * client, AvahiClientState state)
         break;
     case AVAHI_CLIENT_FAILURE:
         ChipLogError(DeviceLayer, "Avahi client failure");
-        mErrorCallback(mAsyncReturnContext, CHIP_ERROR_INTERNAL);
+        // error happend, we should to restart dnssd.
+        mErrorCallback(mAsyncReturnContext, CHIP_ERROR_FORCED_RESET);
```
cecille, line 387: If there's a real failure, won't this continuously restart avahi?

src/lib/dnssd/Discovery_ImplPlatform.cpp
tianfeng-yang: I am not sure that all platforms can try resolve. All platforms will still return errors when they cannot be resolved. Platform-related issues should not be handled here.

src/platform/Linux/DnssdImpl.cpp
tianfeng-yang: Yes, if an error occurs, it will continue to restart. This is the same as restarting the service when a collision error occurs. It will still restart continuously.

andy31415 (approved):

src/lib/dnssd/Discovery_ImplPlatform.cpp
```diff
@@ -631,7 +631,8 @@ CHIP_ERROR DiscoveryImplPlatform::ResolveNodeId(const PeerId & peerId)
 {
     // Resolve requests can only be issued once DNSSD is initialized and there is
     // no caching currently
-    VerifyOrReturnError(mState == State::kInitialized, CHIP_ERROR_INCORRECT_STATE);
+    // When the DNSSD is kInitializing, can also try to resolve.
```
bzbarsky-apple, line 634: Is this true on all platforms?   Does the API documentation clearly say that?

Because the comment right above the bit you are adding flatly contradicts what your new comment says.

What sort of audit was done here?  Why does the existing comment say what it does?

bzbarsky-apple (changes requested):

src/lib/dnssd/Discovery_ImplPlatform.cpp
tianfeng-yang: I have updated the comment

bzbarsky-apple: That answers literally none of the questions I asked.

bzbarsky-apple: So for example: is it true that _on all platforms_ when the state is kInitializing it is ok to try to resolve?  Is there anything that _guarantees_ that this will be ok?  I see nothing that would guarantee that.

PR #32320, opened 2024-02-26
remove the header `src/app/util/common.h`

andy31415:
This header is a header-aggregator and creates strong coupling and include loops between h and cpp file. The includes look like:

image

This creates odd things like:
  - attribute-table.cpp does not even include attribute-table.h
  - util.cpp does not include util.h
  - cpp includes the common.h header which in turn include the header files, making these files impossible to split in a dependency-layered way (it strongly couples util.cpp, attribute-storage.cpp and attribute-table.cpp). The name itself is 'common' (even worse, in a 'util' folder) so its actual functionality is unclear from the name.
  
After the changes, the includes look somewhat better:

image

at least this shows that attribute-storage is a base of everything and attribute-table & util are independent.


mkardous-silabs (approved):

tehampson (approved):

src/app/util/util.cpp
```diff
@@ -15,12 +15,14 @@
  *    limitations under the License.
  */
 
-#include "app/util/common.h"
+#include "util.h"
```
jepenven-silabs, line 18: since a lot of platform code might have some form of util.h somewhere in their SDK, should we use the complete path instead `/app/util/util.h` ? Also it preserve the original include form

src/app/util/attribute-storage.cpp
```diff
@@ -15,12 +15,12 @@
  *    limitations under the License.
  */
 
-#include "app/util/common.h"
+#include "attribute-storage.h"
```
jepenven-silabs, line 18: same comment as below, should use `app/util/attribute-storage.h`

src/app/util/attribute-table.cpp
```diff
@@ -14,13 +14,9 @@
  *    See the License for the specific language governing permissions and
  *    limitations under the License.
  */
-
-// this file contains all the common includes for clusters in the zcl-util
+#include "attribute-table.h"
```
jepenven-silabs, line 17: same comment

jepenven-silabs: After some research should not be a problem for same folder files.

src/app/util/attribute-storage.cpp
jepenven-silabs: After some research should not be a problem for same folder files.

woody-apple (approved):

PR #32324, opened 2024-02-26
Have PairingSession::EncodeSessionParameters require local MRP values

### LINKED ISSUE #30456: In `PairingSession::EncodeSessionParameters` figure out what MRP values we should sent when `providedMrpConfig.HasValue() == false`
Based on the spec `For backwards compatibility, if any tag after tag 2 (SESSION_ACTIVE_INTERVAL) is present, then the SESSION_ACTIVE_INTERVAL SHALL also be present.`

The spec really doesn't define what should happen if we don't have a local config. The other side would assume `GetDefaultMRPConfig()`. But it is not entirely clear if these are the same values on both side

tehampson:
Fixes: #30456

As per `GetLocalMRPConfig()` [documentation](https://github.com/project-chip/connectedhomeip/blob/44ca3957f15ed42f2e6d72fb1a4e1e827c8ed1da/src/messaging/ReliableMessageProtocolConfig.h#L216C1-L223C4):
```
/**
 *  @brief  The custom value of MRP config for the platform.
 *  @return Missing   If the value is same as default value defined by spec
 *          Value     The custom value for the platform
 *
 *  @note   This value is not used by our MRP manager. The value is advertised via mDNS or during PASE/CASE paring, and our peers
 *          use it when communicating with us.
 */
```

In addition to that comment I validated all location where we call `EncodeSessionParameters`. This happens in `src/protocols/secure_channel/CASESession.cpp` and `src/protocols/secure_channel/PASESession.cpp` with the member variable `mLocalMRPConfig`. The following are the four locations where that is set, and where they come from

* `CASESession::PrepareForSessionEstablishment`
  * This is provided by the afore mentioned `GetLocalMRPConfig()` [over here](https://github.com/project-chip/connectedhomeip/blob/44ca3957f15ed42f2e6d72fb1a4e1e827c8ed1da/src/protocols/secure_channel/CASEServer.cpp#L160)
* `CASESession::EstablishSession`
  * Called by `CASEClient::EstablishSession` [here](https://github.com/project-chip/connectedhomeip/blob/44ca3957f15ed42f2e6d72fb1a4e1e827c8ed1da/src/app/CASEClient.cpp#L53-L55)
    * Called by `OperationalSessionSetup::EstablishConnection` where MRP values are in `mInitParams.mrpLocalConfig` [here](https://github.com/project-chip/connectedhomeip/blob/44ca3957f15ed42f2e6d72fb1a4e1e827c8ed1da/src/app/OperationalSessionSetup.cpp#L294)
      * Struct's default is OptionalMissing (Doesn't seems like this is actually used) [code ref](https://github.com/project-chip/connectedhomeip/blob/44ca3957f15ed42f2e6d72fb1a4e1e827c8ed1da/src/app/CASEClient.h#L37)
      * Device Controller Factory where it is set by the afore mentioned `GetLocalMRPConfig()` https://github.com/project-chip/connectedhomeip/blob/44ca3957f15ed42f2e6d72fb1a4e1e827c8ed1da/src/controller/CHIPDeviceControllerFactory.cpp#L273
* `PASESession::WaitForPairing`
  * Called in two different places in `CommissioningWindowManager::AdvertiseAndListenForPASE` [here](https://github.com/project-chip/connectedhomeip/blob/44ca3957f15ed42f2e6d72fb1a4e1e827c8ed1da/src/app/server/CommissioningWindowManager.cpp#L259-L282)
    * In both cases set by the afore mentioned `GetLocalMRPConfig()`
* `PASESession::Pair`
  * Called by `DeviceCommissioner::EstablishPASEConnection` [here](https://github.com/project-chip/connectedhomeip/blob/44ca3957f15ed42f2e6d72fb1a4e1e827c8ed1da/src/controller/CHIPDeviceController.cpp#L751)
    * Set by the afore mentioned `GetLocalMRPConfig()`

### DIFF
--- a/src/protocols/secure_channel/CASESession.cpp
+++ b/src/protocols/secure_channel/CASESession.cpp
@@ -477,7 +477,7 @@ CASESession::PrepareForSessionEstablishment(SessionManager & sessionManager, Fab
     mFabricsTable             = fabricTable;
     mRole                     = CryptoContext::SessionRole::kResponder;
     mSessionResumptionStorage = sessionResumptionStorage;
-    mLocalMRPConfig           = mrpLocalConfig;
+    mLocalMRPConfig           = mrpLocalConfig.ValueOr(GetDefaultMRPConfig());
 
     ChipLogDetail(SecureChannel, "Allocated SecureSession (%p) - waiting for Sigma1 msg",
                   mSecureSessionHolder.Get().Value()->AsSecureSession());
@@ -525,7 +525,7 @@ CHIP_ERROR CASESession::EstablishSession(SessionManager & sessionManager, Fabric
     mFabricsTable             = fabricTable;
     mFabricIndex              = fabricInfo->GetFabricIndex();
     mSessionResumptionStorage = sessionResumptionStorage;
-    mLocalMRPConfig           = mrpLocalConfig;
+    mLocalMRPConfig           = mrpLocalConfig.ValueOr(GetDefaultMRPConfig());
 
     mExchangeCtxt->UseSuggestedResponseTimeout(kExpectedSigma1ProcessingTime);
     mPeerNodeId  = peerScopedNodeId.GetNodeId();
@@ -706,7 +706,7 @@ CHIP_ERROR CASESession::SendSigma1()
     ReturnErrorOnFailure(
         tlvWriter.PutBytes(TLV::ContextTag(4), mEphemeralKey->Pubkey(), static_cast<uint32_t>(mEphemeralKey->Pubkey().Length())));
 
-    ReturnErrorOnFailure(EncodeSessionParameters(TLV::ContextTag(5), mLocalMRPConfig.ValueOr(GetDefaultMRPConfig()), tlvWriter));
+    ReturnErrorOnFailure(EncodeSessionParameters(TLV::ContextTag(5), mLocalMRPConfig, tlvWriter));
 
     // Try to find persistent session, and resume it.
     bool resuming = false;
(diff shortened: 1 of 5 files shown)


src/protocols/secure_channel/CASESession.cpp
```diff
@@ -706,7 +706,7 @@ CHIP_ERROR CASESession::SendSigma1()
     ReturnErrorOnFailure(
         tlvWriter.PutBytes(TLV::ContextTag(4), mEphemeralKey->Pubkey(), static_cast<uint32_t>(mEphemeralKey->Pubkey().Length())));
 
-    ReturnErrorOnFailure(EncodeSessionParameters(TLV::ContextTag(5), mLocalMRPConfig, tlvWriter));
+    ReturnErrorOnFailure(EncodeSessionParameters(TLV::ContextTag(5), mLocalMRPConfig.ValueOr(GetDefaultMRPConfig()), tlvWriter));
```
Damian-Nordic, line 709: Would it make sense to make `mLocalMRPConfig` non-optional, initialized to `GetDefaultMRPConfig()`, instead?

Damian-Nordic (approved):

src/protocols/secure_channel/CASESession.cpp
Damian-Nordic: Also, `GetLocalMRPConfig()` might no longer need `Optional` if it's always needed.

bzbarsky-apple: For DNS-SD advertising, the Optional makes sense, because it allows us to not advertise default values and waste wire space, right?  Though that code could itself compare to default and skip if equal.  That feels like followup fodder.

Past that, I agree: we should just do the `ValueOr()` thing when setting `mLocalMRPConfig`; it's less code.

And I sort of question whether we should have the MRP config as an argument to this stuff anyway, if we are going to call the global getters after all.  But that again feels like followup fodder.

Damian-Nordic: Ah, right. For DNS-SD it makes sense.

tehampson: >Would it make sense to make `mLocalMRPConfig` non-optional, initialized to `GetDefaultMRPConfig()`, instead?

thanks, done

src/protocols/secure_channel/PairingSession.h
```diff
@@ -238,7 +238,7 @@ class DLL_EXPORT PairingSession : public SessionDelegate
 
     // mLocalMRPConfig is our config which is sent to the other end and used by the peer session.
     // mRemoteSessionParams is received from other end and set to our session.
-    Optional<ReliableMessageProtocolConfig> mLocalMRPConfig;
+    ReliableMessageProtocolConfig mLocalMRPConfig = GetDefaultMRPConfig();
```
bzbarsky-apple, line 241: Shouldn't this initialize to GetLocalMRPConfig?

bzbarsky-apple (approved):

pidarped (approved):

Issue #25649, opened 2023-03-13
[Java-matter-controller-test] upgrade the im test to use case connection if device has been paired.


yunhanw-google:

yufengwangca:
This is not only about test, the problem is in IM cluster commands implementation, currently, all IM cluster commands are registered under cluster 'im', but implemented as 'pairing' cluster command. The behavior is undefined, if we register a command in one cluster, but implemented it under another cluster. Lets follow-up this after this PR

PR #32326, opened 2024-02-26
[Android] Add initial extensible invoke command support

yunhanw-google:
ExtensibleInvoke can be used to send multiple invoke command at one time. In this PR.
-- Add extensibleInvoke and corresponding callback in Android Java/JNI
-- Add example test where we add two commands and send them out, and validate their response respectively.

### DIFF
--- a/src/controller/java/AndroidCallbacks.h
+++ b/src/controller/java/AndroidCallbacks.h
@@ -25,6 +25,7 @@
 #include <lib/core/CHIPError.h>
 #include <lib/support/JniTypeWrappers.h>
 #include <list>
+#include <unordered_map>
 #include <utility>
 
 namespace chip {
@@ -138,6 +139,20 @@ struct InvokeCallback : public app::CommandSender::Callback
     JniGlobalReference mWrapperCallbackRef;
 };
 
+struct ExtendableInvokeCallback : public app::CommandSender::ExtendableCallback
+{
+    ExtendableInvokeCallback(jobject wrapperCallback);
+    ~ExtendableInvokeCallback();
+
+    void OnResponse(app::CommandSender * commandSender, const app::CommandSender::ResponseData & aResponseData) override;
+    void OnNoResponse(app::CommandSender * commandSender, const app::CommandSender::NoResponseData & aNoResponseData) override;
+    void OnError(const app::CommandSender * apCommandSender, const app::CommandSender::ErrorData & aErrorData) override;
+    void OnDone(app::CommandSender * apCommandSender) override;
+
+    app::CommandSender * mCommandSender = nullptr;
+    JniGlobalReference mWrapperCallbackRef;
+};
+
 jlong newConnectedDeviceCallback(JNIEnv * env, jobject self, jobject callback);
 void deleteConnectedDeviceCallback(JNIEnv * env, jobject self, jlong callbackHandle);
 jlong newReportCallback(JNIEnv * env, jobject self, jobject subscriptionEstablishedCallbackJava,
@@ -147,6 +162,8 @@ jlong newWriteAttributesCallback(JNIEnv * env, jobject self);
 void deleteWriteAttributesCallback(JNIEnv * env, jobject self, jlong callbackHandle);
 jlong newInvokeCallback(JNIEnv * env, jobject self);
 void deleteInvokeCallback(JNIEnv * env, jobject self, jlong callbackHandle);
+jlong newExtendableInvokeCallback(JNIEnv * env, jobject self);
+void deleteExtendableInvokeCallback(JNIEnv * env, jobject self, jlong callbackHandle);
 
 } // namespace Controller
 } // namespace chip
--- a/src/controller/java/AndroidInteractionClient.cpp
+++ b/src/controller/java/AndroidInteractionClient.cpp
@@ -431,7 +431,7 @@ exit:
 
 CHIP_ERROR PutPreencodedInvokeRequest(app::CommandSender & commandSender, app::CommandPathParams & path, const ByteSpan & data)
 {
-    // PrepareCommand does nott create the struct container with kFields and copycontainer below sets the
+    // PrepareCommand does not create the struct container with kFields and copycontainer below sets the
     // kFields container already
     ReturnErrorOnFailure(commandSender.PrepareCommand(path, false /* aStartDataStruct */));
     TLV::TLVWriter * writer = commandSender.GetCommandDataIBTLVWriter();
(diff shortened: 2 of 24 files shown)


joonhaengHeo (approved):

examples/java-matter-controller/java/src/com/matter/controller/commands/pairing/PairOnNetworkLongImBatchInvokeCommand.kt
```diff
@@ -0,0 +1,182 @@
...
+        if (status != null) {
+          setFailure("invoke failure with incorrect status")
+        }
+      }
+      responseCount++
+    }
+
+    override fun onNoResponse(noInvokeResponseData: NoInvokeResponseData) {
+      logger.log(Level.INFO, "Batch Invoke receive onNoResponse on $noInvokeResponseData")
+      responseCount--
```
tehampson, line 98: I don't think you want to decrement here.

examples/java-matter-controller/java/src/com/matter/controller/commands/pairing/PairOnNetworkLongImBatchInvokeCommand.kt
```diff
@@ -0,0 +1,182 @@
...
+    val tlvWriter2 = TlvWriter()
+    tlvWriter2.startStructure(AnonymousTag)
+    tlvWriter2.put(ContextSpecificTag(0), number)
+    tlvWriter2.put(ContextSpecificTag(1), number)
+    tlvWriter2.endStructure()
+
+    val element2: InvokeElement =
+      InvokeElement.newInstance(
+        /* endpointId= */ 1,
+        CLUSTER_ID_TEST,
```
tehampson, line 146: Is this test cluster there for all different servers?

I am trying to understand why we are deciding to send this second command here. Is this a temporary things, or is this expected to be more permanent?

src/controller/java/AndroidInteractionClient.cpp
```diff
@@ -442,6 +442,195 @@ CHIP_ERROR PutPreencodedInvokeRequest(app::CommandSender & commandSender, app::C
...
+        isEndpointIdValid = (env->CallBooleanMethod(invokeElement, isEndpointIdValidMethod) == JNI_TRUE);
+        isGroupIdValid    = (env->CallBooleanMethod(invokeElement, isGroupIdValidMethod) == JNI_TRUE);
+
+        if (isEndpointIdValid)
+        {
+            endpointIdObj = env->CallLongMethod(invokeElement, getEndpointIdMethod, static_cast<jlong>(kInvalidEndpointId));
+            VerifyOrExit(!env->ExceptionCheck(), err = CHIP_JNI_ERROR_EXCEPTION_THROWN);
+        }
+
+        if (isGroupIdValid)
```
tehampson, line 534: This will always be false, right? Why is this here?

src/controller/java/AndroidCallbacks.h
```diff
@@ -138,6 +139,24 @@ struct InvokeCallback : public app::CommandSender::Callback
...
+
+    void OnResponse(app::CommandSender * commandSender, const app::CommandSender::ResponseData & aResponseData) override;
+    void OnNoResponse(app::CommandSender * commandSender, const app::CommandSender::NoResponseData & aNoResponseData) override;
+    void OnError(const app::CommandSender * apCommandSender, const app::CommandSender::ErrorData & aErrorData) override;
+    void OnDone(app::CommandSender * apCommandSender) override;
+
+    CHIP_ERROR GetIndexForCommandRef(uint16_t aCommandRef, size_t & aIndex);
+    CHIP_ERROR AddCommandRefToIndexLookup(uint16_t aCommandRef, size_t aIndex);
+
+    std::unordered_map<uint16_t, size_t> commandRefToIndex;
```
tehampson, line 155: If you don't want to, you don't have to use an unordered_map here. I would argue that is adds complexity without any benefit. This is something I am meaning to clean up in chip-repl.

In your case, because you are providing the `commandRef` to be the index already. You don't need this at all. If you want to be very safe, you can store the length of the array and validate that the CommandRef you get back in the OnResponse is less than the max. But you can rely on that `CommandRef` being the index for you.

src/controller/java/AndroidInteractionClient.cpp
```diff
@@ -442,6 +442,195 @@ CHIP_ERROR PutPreencodedInvokeRequest(app::CommandSender & commandSender, app::C
     return writer->CopyContainer(TLV::ContextTag(app::CommandDataIB::Tag::kFields), reader);
 }
 
+CHIP_ERROR PutPreencodedInvokeRequest(app::CommandSender & commandSender, app::CommandPathParams & path, const ByteSpan & data,
+                                      app::CommandSender::PrepareCommandParameters & prepareCommandParams)
+{
+    // PrepareCommand does nott create the struct container with kFields and copycontainer below sets the
```
tehampson, line 448: nit, 
```suggestion
    // PrepareCommand does not create the struct container with kFields and copycontainer below sets the
```

examples/java-matter-controller/java/src/com/matter/controller/commands/pairing/PairOnNetworkLongImBatchInvokeCommand.kt
yunhanw-google: This is the test, we pair firstly, then send two command in batch, thanks

src/controller/java/AndroidInteractionClient.cpp
```diff
@@ -442,6 +442,195 @@ CHIP_ERROR PutPreencodedInvokeRequest(app::CommandSender & commandSender, app::C
...
+    ChipLogDetail(Controller, "IM batchInvoke() called");
+
+    DeviceProxy * device = reinterpret_cast<DeviceProxy *>(devicePtr);
+    VerifyOrExit(device != nullptr, err = CHIP_ERROR_INCORRECT_STATE);
+    VerifyOrExit(device->GetSecureSession().HasValue(), err = CHIP_ERROR_MISSING_SECURE_SESSION);
+
+    VerifyOrExit(invokeElementList != nullptr, err = CHIP_ERROR_INVALID_ARGUMENT);
+    SuccessOrExit(err = JniReferences::GetInstance().GetListSize(invokeElementList, listSize));
+
+    if ((listSize > 1) && (device->GetSecureSession().Value()->IsGroupSession()))
```
tehampson, line 482: I don't think anything with group commands should ever call batch commands. They are two very different things

examples/java-matter-controller/java/src/com/matter/controller/commands/pairing/PairOnNetworkLongImBatchInvokeCommand.kt
yunhanw-google: Not permanent, this is temporary, we need the dedicated IM folder, and dedicated test suite that use IM for testing, that effort is still in radar , would happen in two or three months

src/controller/java/AndroidCallbacks.h
yunhanw-google: Agree, would remove it, thanks

src/controller/java/AndroidInteractionClient.cpp
```diff
@@ -442,6 +442,195 @@ CHIP_ERROR PutPreencodedInvokeRequest(app::CommandSender & commandSender, app::C
...
+        app::CommandSender::FinishCommandParameters finishCommandParams(convertedTimedRequestTimeoutMs != 0
+                                                                            ? Optional<uint16_t>(convertedTimedRequestTimeoutMs)
+                                                                            : Optional<uint16_t>::Missing());
+
+        finishCommandParams.commandRef = prepareCommandParams.commandRef;
+        SuccessOrExit(err = commandSender->FinishCommand(finishCommandParams));
+
+        // CommandSender provides us with the CommandReference for this associated command. In order to match responses
+        // we have to add CommandRef to index lookup.
+        SuccessOrExit(err = callback->AddCommandRefToIndexLookup(finishCommandParams.commandRef.Value(), i));
```
tehampson, line 598: If you remove the `unordered_map` then you no longer actually need to track here

chrisdecenzo (approved):

src/controller/java/AndroidInteractionClient.cpp
yunhanw-google: it would take care of all, updated with extenedableInvoke, thanks

src/controller/java/AndroidInteractionClient.cpp
yunhanw-google: removed, thanks

src/controller/java/AndroidInteractionClient.cpp
```diff
@@ -442,6 +442,195 @@ CHIP_ERROR PutPreencodedInvokeRequest(app::CommandSender & commandSender, app::C
...
+        }
+
+        app::CommandSender::FinishCommandParameters finishCommandParams(convertedTimedRequestTimeoutMs != 0
+                                                                            ? Optional<uint16_t>(convertedTimedRequestTimeoutMs)
+                                                                            : Optional<uint16_t>::Missing());
+
+        finishCommandParams.commandRef = prepareCommandParams.commandRef;
+        SuccessOrExit(err = commandSender->FinishCommand(finishCommandParams));
+
+        // CommandSender provides us with the CommandReference for this associated command. In order to match responses
```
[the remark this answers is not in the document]
yunhanw-google: removed, thanks

src/controller/java/AndroidCallbacks.h
yunhanw-google: removed, thanks

src/controller/java/AndroidInteractionClient.cpp
yunhanw-google: done, thanks

examples/java-matter-controller/java/src/com/matter/controller/commands/pairing/PairOnNetworkLongImBatchInvokeCommand.kt
yunhanw-google: thanks for catching. yes, removed, thanks

examples/java-matter-controller/java/src/com/matter/controller/commands/pairing/PairOnNetworkLongImBatchInvokeCommand.kt
yunhanw-google: it would be handled in https://github.com/project-chip/connectedhomeip/issues/25649, thanks

src/controller/java/AndroidInteractionClient.cpp
yunhanw-google: it is for batch, and group...Update with extensible. thanks

tehampson (approved):

src/controller/java/AndroidCallbacks.h
```diff
@@ -138,6 +139,20 @@ struct InvokeCallback : public app::CommandSender::Callback
     JniGlobalReference mWrapperCallbackRef;
 };
 
+struct ExtenedableInvokeCallback : public app::CommandSender::ExtendableCallback
```
andy31415, line 142: seems like an extra `e` compared with the other other one:

`ExtendableCallback`
`ExtenedableInvokeCallback `

src/controller/java/MatterInteractionClient-JNI.cpp
```diff
@@ -65,3 +65,15 @@ JNI_METHOD(void, invoke)
         ChipLogError(Controller, "JNI IM Invoke Error: %" CHIP_ERROR_FORMAT, err.Format());
     }
 }
+
+JNI_METHOD(void, extenedableInvoke)
```
andy31415, line 69: extra E here as well. I think we should grep for `extenedable` and replace with `extendable` everywhere

src/controller/java/src/chip/devicecontroller/ExtenedableInvokeCallback.java
```diff
@@ -0,0 +1,54 @@
...
+ *   limitations under the License.
+ *
+ */
+package chip.devicecontroller;
+
+import chip.devicecontroller.model.InvokeResponseData;
+import chip.devicecontroller.model.NoInvokeResponseData;
+
+/** An interface for receiving invoke response. */
+public interface ExtenedableInvokeCallback {
```
andy31415, line 24: move file and rename `s/Extenedable/Extendable/`

src/controller/java/MatterInteractionClient-JNI.cpp
yunhanw-google: updated, thanks

PR #32330, opened 2024-02-27
Fix MTRMetricsCollector.mm build failure using clang

anush-apple:
- Missing static_cast to int for enum class in a format string


ksperling-apple (approved):

ksperling-apple:

yunhanw-google (approved):

PR #32192, opened 2024-02-17
[ESP32] Read core dump from flash, for diagnostic crash logs

shubhamdp:
Send actual core dump as diagnostic crash logs.

Fixes #31727

#### Tests
Crashed device once and read and decoded the core dump

PR #32331, opened 2024-02-27
[ESP32] Provide core dump summary in response to crash diagnostic log requests

shubhamdp:
Only send the core dump summary instead of complete core dump.

This also addresses the after merge comment https://github.com/project-chip/connectedhomeip/pull/32192#pullrequestreview-1891091762

### DIFF
--- a/examples/temperature-measurement-app/esp32/main/diagnostic-logs-provider-delegate-impl.cpp
+++ b/examples/temperature-measurement-app/esp32/main/diagnostic-logs-provider-delegate-impl.cpp
@@ -92,9 +92,8 @@ size_t LogProvider::GetCrashSize()
 
 #if defined(CONFIG_ESP_COREDUMP_ENABLE_TO_FLASH) && defined(CONFIG_ESP_COREDUMP_DATA_FORMAT_ELF)
     // Verify that the crash is present and sane
-    size_t unusedOutAddr, unusedOutSize;
-    esp_err_t esp_err = esp_core_dump_image_get(&unusedOutAddr, &unusedOutSize);
-    VerifyOrReturnValue(esp_err == ESP_OK, 0, ChipLogError(DeviceLayer, "Failed to get core dump image, esp_err:%d", esp_err));
+    esp_err_t esp_err = esp_core_dump_image_check();
+    VerifyOrReturnValue(esp_err == ESP_OK, 0, ChipLogError(DeviceLayer, "Core dump image check failed, esp_err:%d", esp_err));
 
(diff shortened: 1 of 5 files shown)


shubhamdp:
@tcarmelveilleux Please take a look.

examples/temperature-measurement-app/esp32/main/diagnostic-logs-provider-delegate-impl.cpp
```diff
@@ -95,34 +91,17 @@ size_t LogProvider::GetCrashSize()
...
 
 #if defined(CONFIG_ESP_COREDUMP_ENABLE_TO_FLASH) && defined(CONFIG_ESP_COREDUMP_DATA_FORMAT_ELF)
-    size_t unusedOutAddr;
-    esp_err_t esp_err = esp_core_dump_image_get(&unusedOutAddr, &outSize);
+    // Verify that the crash is present and sane
+    size_t unusedOutAddr, unusedOutSize;
+    esp_err_t esp_err = esp_core_dump_image_get(&unusedOutAddr, &unusedOutSize);
     VerifyOrReturnValue(esp_err == ESP_OK, 0, ChipLogError(DeviceLayer, "Failed to get core dump image, esp_err:%d", esp_err));
+
+    outSize = sizeof(esp_core_dump_summary_t);
```
tcarmelveilleux, line 99: Need to check that the outsize from `esp_core_dump_image_get` is `>= sizeof(esp_core_dump_summary_t)` rather than ignoring the outsize.

tcarmelveilleux (approved):

examples/temperature-measurement-app/esp32/main/diagnostic-logs-provider-delegate-impl.cpp
shubhamdp: `esp_core_dump_image_get()` gets the complete core dump image and  outSize will always be greater than `sizeof(esp_core_dump_summary_t)`, and the check you suggested to add does make sense. But, as the core dump image size and summary size is not related, I used a different API, [`esp_core_dump_image_check()`](https://github.com/espressif/esp-idf/blob/master/components/espcoredump/include/esp_core_dump.h#L95-L106) here which also checks the integrity along with size checks as well.

dhrishi (approved):

examples/temperature-measurement-app/esp32/main/diagnostic-logs-provider-delegate-impl.cpp
shubhamdp: resolving the thread!

PR #32339, opened 2024-02-27
fix crash caused by lacking null pointer check and wrong cluster id usage

### LINKED ISSUE #32338: [1.3] TC-OO-2.3 fails on step 4a
### Reproduction steps

Device crashes when executing OnWithRecallGlobalScene.
Problem lies in handling recall scene.

### Bug prevalence

always

a2d98f869562ae1efc9a6c1c40f5049a12fa71e0

### Platform

core

### Platform Version(s)

1.3

### Type

Core SDK Crash

### Anything else?

_No response_

fessehaeve:
Crash detected while executing recall-scene for onoff cluster

- crash caused by lacking null pointer check in timer callback
- this was caused by using wrong cluster id in the transition time implementation
- fixes https://github.com/project-chip/connectedhomeip/issues/32338

### DIFF
--- a/src/app/clusters/on-off-server/on-off-server.cpp
+++ b/src/app/clusters/on-off-server/on-off-server.cpp
@@ -214,6 +214,7 @@ public:
               ScenesManagement::ScenesServer::Instance().IsHandlerRegistered(endpoint, LevelControlServer::GetSceneHandler())))
 #endif
         {
+            VerifyOrReturnError(mTransitionTimeInterface.sceneEventControl(endpoint) != nullptr, CHIP_ERROR_INVALID_ARGUMENT);
             OnOffServer::Instance().scheduleTimerCallbackMs(mTransitionTimeInterface.sceneEventControl(endpoint), timeMs);
         }
 
@@ -221,7 +222,7 @@ public:
     }
 
 private:
-    OnOffTransitionTimeInterface mTransitionTimeInterface = OnOffTransitionTimeInterface(Attributes::OnOff::Id, sceneOnOffCallback);
+    OnOffTransitionTimeInterface mTransitionTimeInterface = OnOffTransitionTimeInterface(OnOff::Id, sceneOnOffCallback);
 };
 static DefaultOnOffSceneHandler sOnOffSceneHandler;
 


src/app/clusters/on-off-server/on-off-server.cpp
```diff
@@ -214,14 +214,15 @@ class DefaultOnOffSceneHandler : public scenes::DefaultSceneHandlerImpl
...
+            VerifyOrReturnError(mTransitionTimeInterface.sceneEventControl(endpoint) != nullptr, CHIP_ERROR_INVALID_ARGUMENT);
             OnOffServer::Instance().scheduleTimerCallbackMs(mTransitionTimeInterface.sceneEventControl(endpoint), timeMs);
         }
 
         return CHIP_NO_ERROR;
     }
 
 private:
-    OnOffTransitionTimeInterface mTransitionTimeInterface = OnOffTransitionTimeInterface(Attributes::OnOff::Id, sceneOnOffCallback);
+    OnOffTransitionTimeInterface mTransitionTimeInterface = OnOffTransitionTimeInterface(OnOff::Id, sceneOnOffCallback);
```
andy31415, line 225: What will this Id be from? I believe we have both attributes and cluster as ID, it would be nice to be explicit.

bauerschwan (approved):

andy31415 (approved):

src/app/clusters/on-off-server/on-off-server.cpp
andy31415: I would prefer Clusters:: for clarity, however that also seems nitpicky. Will defer to any other reviewer to see how subjective that is.

This is not the first time I have see bugs due to different IDs being used. One would have hoped we would be more type safe than that :/

andy31415: I see Stefan LGTMed it as-is, so I will close my comment.

fessehaeve: You could as well see in this file that cluster ID is used without Clusters:: everywhere.

lpbeliveau (approved):

PR #32343, opened 2024-02-27
Define explicit `data_model` sub-targets for generated files

andy31415:
we currently code-generate across all layers, both config headers (that generally define endpoint composition) and cpp implementations like init callback stubs and data dispatch. As a result we cannot have a "data model target" as a stand-alone since this would imply a full monolith of src/app.

This PR contains two changes:
   - define separate targets for data_model that can be used in the future for GN dependencies
   - split out constants for privileges so that codegen can be independent from privilege-storage
   
This is what codegen we have for app (orange red bits):

image

Additional this is the constant include move for privilege-storage. Existing code would force me to combine generated code and non-generated:

image

Where as new code breaks that dependency into 2:

image

After the PR, this is what targets could exist (they do not as there is still no GN linkage ... I will need to include some form of `data_model_BACKEND` that differs across each app):

image

### DIFF
--- a/src/app/util/privilege-constants.h
+++ b/src/app/util/privilege-constants.h
@@ -1,0 +1,23 @@
+/**
+ *
+ *    Copyright (c) 2024 Project CHIP Authors
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+#pragma once
+
+inline constexpr int kMatterAccessPrivilegeView       = 0;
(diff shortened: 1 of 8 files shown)


tcarmelveilleux (approved):

yunhanw-google (approved):

tehampson (approved):

src/app/util/privilege-constants.h
```diff
@@ -0,0 +1,23 @@
...
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+#pragma once
+
+inline constexpr int kMatterAccessPrivilegeView       = 0;
+inline constexpr int kMatterAccessPrivilegeOperate    = 1;
+inline constexpr int kMatterAccessPrivilegeManage     = 2;
+inline constexpr int kMatterAccessPrivilegeAdminister = 3;
+inline constexpr int kMatterAccessPrivilegeMaxValue   = kMatterAccessPrivilegeAdminister;
```
bzbarsky-apple, line 23: So here is the thing.  These constants just duplicate constants from the cluster with exactly the same values.  Can we just use the cluster constants and stop pretending like there are these other constants involved?

Followup for that is probably OK, but this pretending like they are different constants but then breadking if they are not is just really really fragile.

bzbarsky-apple (approved):

src/app/util/privilege-constants.h
andy31415: Will look at this as a followup - have to make sure that we grab these from some generated code that is not a `chip_data_model` generated bit.

PR #32349, opened 2024-02-27
[SILABS] Multi-chip OTA work

mykrupp:
Silabs multi-chip ota feature:

Functionality to support arbitrary image downloads in a single OTA file. This can use pre-defined TLVs or custom TLVs. 
ota_image_tool.py script updated to create custom OTA file with TLVs embedded.


src/platform/silabs/efr32/args.gni
```diff
@@ -42,6 +41,7 @@ chip_system_config_use_open_thread_inet_endpoints = true
 chip_with_lwip = false
 
 chip_build_tests = false
+chip_enable_multi_ota_requestor = false
```
jmartinez-silabs, line 44: I don't think this should be here

src/platform/silabs/multi-ota/efr32/OTAHooks.cpp
```diff
@@ -0,0 +1,70 @@
...
+#define CONFIG_CHIP_SILABS_OTA_FACTORY_DATA_PROCESSOR 0
+#endif // CONFIG_CHIP_SILABS_OTA_FACTORY_DATA_PROCESSOR
+
+
+#if CONFIG_CHIP_SILABS_OTA_FACTORY_DATA_PROCESSOR
+#include <platform/silabs/multi-ota/efr32/FactoryDataProcessor.h>
+#endif // CONFIG_CHIP_SILABS_OTA_FACTORY_DATA_PROCESSOR
+
+
+extern "C" void ResetMCU(void);
```
jmartinez-silabs, line 43: Why this extern "C" functions? If you need an abstraction to reset platforms it should be a cpp function in Silabs Platform abstraction

src/platform/silabs/multi-ota/OTATlvProcessor.h
```diff
@@ -0,0 +1,162 @@
...
+#include <lib/core/Optional.h>
+#include <lib/support/ScopedBuffer.h>
+#include <lib/support/Span.h>
+
+namespace chip {
+
+#define CHIP_ERROR_TLV_PROCESSOR(e)                                                                                                \
+    ChipError(ChipError::Range::kLastRange, ((uint8_t) ChipError::Range::kLastRange << 3) | e, __FILE__, __LINE__)
+
+#define CHIP_OTA_TLV_CONTINUE_PROCESSING CHIP_ERROR_TLV_PROCESSOR(0x01)
```
jmartinez-silabs, line 30: all those define should be a enum class

src/platform/silabs/multi-ota/efr32/OTAHooks.cpp
```diff
@@ -0,0 +1,70 @@
...
+ * Enables default OTA TLV factory data processor.
+ * Disabled by default.
+ */
+#ifndef CONFIG_CHIP_SILABS_OTA_FACTORY_DATA_PROCESSOR
+#define CONFIG_CHIP_SILABS_OTA_FACTORY_DATA_PROCESSOR 0
+#endif // CONFIG_CHIP_SILABS_OTA_FACTORY_DATA_PROCESSOR
+
+
+#if CONFIG_CHIP_SILABS_OTA_FACTORY_DATA_PROCESSOR
+#include <platform/silabs/multi-ota/efr32/FactoryDataProcessor.h>
```
jmartinez-silabs, line 39: I don't see this header

src/platform/silabs/multi-ota/MultiOTAImageProcessorImpl.h
```diff
@@ -0,0 +1,117 @@
...
+/*
+ * OTA hooks that can be overwritten by application.
+ * Default behavior is implemented as WEAK symbols in platform OtaHooks.cpp.
+ */
+
+/*
+ * This hook is called at the end of OTAImageProcessorImpl::Init.
+ * It should generally register the OTATlvProcessor instances.
+ */
+extern "C" CHIP_ERROR OtaHookInit();
```
jmartinez-silabs, line 38: I don't quite get those 3 functions. why are they extern "C" ? Why are their prototype here?

Should that rather be virtual functions in the  OTAOmageProcessorInterface or something similar?

src/platform/silabs/efr32/args.gni
mykrupp: removed from here

src/platform/silabs/multi-ota/efr32/OTAHooks.cpp
mykrupp: They don't need to be extern, I removed the extern tag

src/platform/silabs/multi-ota/MultiOTAImageProcessorImpl.h
mykrupp: These don't need to be extern, removed the extern tag. They could be declared in a separate OTAHooks.h file

src/platform/silabs/efr32/BUILD.gn
```diff
@@ -77,10 +78,20 @@ static_library("efr32") {
...
+      "${silabs_platform_dir}/multi-ota/OTATlvProcessor.h",
+      "${silabs_platform_dir}/multi-ota/efr32/OTAFirmwareProcessor.cpp",
+      "${silabs_platform_dir}/multi-ota/efr32/OTAFirmwareProcessor.h",
+      "${silabs_platform_dir}/multi-ota/efr32/OTAHooks.cpp",
+    ]
+  } else if (chip_enable_ota_requestor) {
     sources += [
       "${silabs_platform_dir}/OTAImageProcessorImpl.h",
-      "OTAImageProcessorImpl.cpp",
+      "${silabs_platform_dir}/efr32/OTAImageProcessorImpl.cpp",
```
jmartinez-silabs, line 94: I guess it leads to the same result but `OTAImageProcessorImpl.cpp` is in the same folder as this build.gn which is `${silabs_platform_dir}/efr32/`. You can leave it with no additional path here.

src/platform/silabs/multi-ota/efr32/OTAHooks.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+#include <include/platform/CHIPDeviceLayer.h>
+#include <platform/silabs/multi-ota/OTAMultiImageProcessorImpl.h>
+
+#include <app/clusters/ota-requestor/OTARequestorInterface.h>
+
+#include <platform/silabs/multi-ota/efr32/OTACustomProcessor.h>
+#include <platform/silabs/multi-ota/efr32/OTAFirmwareProcessor.h>
+
+CHIP_ERROR ProcessDescriptor(void * descriptor)
```
jmartinez-silabs, line 27: Can you explain the implementation choice of OTAHook.cpp.
I am trying to understand why are those not in one of the OTA classes you created.
It is more of a C-esque implementation while you implement the rest in C++. Is there a reason for that?

Maybe it should be in part of the namespace.

mykrupp: I could add them to part of OTAMultiImageProcessorImpl. I had them in their own file since this function might be edited by a user who created their own TLV processors and wants to test them.

src/platform/silabs/multi-ota/OTATlvProcessor.h
mykrupp: Is there a way to do this with keeping them as a ChipError type?

src/platform/silabs/multi-ota/efr32/OTAHooks.cpp
mykrupp: added to OTAMultiImageProcessorImpl namespace

selissia (approved):

Damian-Nordic (approved):

PR #32223, opened 2024-02-20
Framework for emitting metrics data

anush-apple:
### Change Overview:

- Added mechanism to log simple metric data. Metric encapsulated by object to allow future additions.
- Metrics will be noop on platforms that don't enable the feature
- Switched the existing MATTER_TRACE_METRIC to the new metric macro MATTER_LOG_METRIC to match data logging patterns with other MATTER_LOG_XYZ formats
- Added sample metrics data in CHIPDeviceController::EstablishPASEConnection as a starter

Testing### 

- Verified metrics emitted using darwin-framework-tool
- Verified compilation for ESP32 and tv-casting-app compilation with the new changes

PR #32355, opened 2024-02-28
[ESP32] Tracing : Added the fix of data type mismatch while registering the metrics in esp32 tracing framework.

shripad621git:
**Problem**
- Testing the merged PR #32223 for esp32 showed failure to report metric data to esp32 tracing framework due to data type mismatch.

**Change Overview:**
- Added the fix for the metric reporting.
- Removed the heap diagnostic metrics in the traces due to framework built in support.

**Testing**
- Tested the esp32 lighting app with the fix and verified the metric report.

### DIFF
--- a/src/tracing/esp32_trace/esp32_tracing.cpp
+++ b/src/tracing/esp32_trace/esp32_tracing.cpp
@@ -134,10 +134,7 @@ void RemoveHashFromPermitlist(const char * str)
 #define LOG_HEAP_INFO(label, group, entry_exit)                                                                                    \
     do                                                                                                                             \
     {                                                                                                                              \
-        ESP_DIAG_EVENT("MTR_TRC", "%s - %s - %s Min Free heap - %u - LFB - %u Start free heap - %u", entry_exit, label, group,     \
-                       heap_caps_get_minimum_free_size(MALLOC_CAP_8BIT),                                                           \
-                       heap_caps_get_largest_free_block(MALLOC_CAP_INTERNAL | MALLOC_CAP_8BIT),                                    \
-                       heap_caps_get_free_size(MALLOC_CAP_8BIT));                                                                  \
+        ESP_DIAG_EVENT("MTR_TRC", "%s - %s - %s", entry_exit, label, group);                                                       \
     } while (0)
 
 void ESP32Backend::LogMessageReceived(MessageReceivedInfo & info) {}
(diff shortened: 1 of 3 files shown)


dhrishi (approved):

src/tracing/esp32_trace/esp32_tracing.cpp
```diff
@@ -157,32 +154,58 @@ void ESP32Backend::TraceCounter(const char * label)
 
 void ESP32Backend::LogMetricEvent(const MetricEvent & event)
 {
+    using ValueType = MetricEvent::Value::Type;
     if (!mRegistered)
```
andy31415, line 158: you have a single registered however several registration types. Is this ok?

i.e. if I log metrics like this:
   LogMetric(Int32)
   LogMetric(ErrorCode)

this will only ever register a "DATA_TYPE_INT" however then below it will send values of both int ant uint type.

"event.key" also varies throughout metrics ... we register only once for the first key we see, but then we try to upload data with every key. How does this work given that `mRegistered` is a single boolean value? Sounds like you would need some `set<tuple<string, type>>` to define what was registered and what not ... and that sounds potentially unbounded.

How is this supposed to work?

andy31415 (approved):
Giving checkmark because platform-specific code with platform approval ... however I have some questions that I placed in the PR ... the usage of a single mRegistered is suspect.

src/tracing/esp32_trace/esp32_tracing.cpp
shripad621git: Addressed in the latest change.

src/tracing/esp32_trace/esp32_tracing.cpp
```diff
@@ -155,34 +153,75 @@ void ESP32Backend::TraceCounter(const char * label)
...
+    if (mRegisteredMetrics.find(key) != mRegisteredMetrics.end())
+    {
+        if (mRegisteredMetrics[key] != type)
+        {
+            ESP_LOGE("SYS.MTR", "Type mismatch for metric key %s", key);
+            return;
+        }
+        else
+        {
+            return;
```
andy31415, line 168: nit: no return after else. This double return could be cleaned up.

shripad621git: Done.

PR #32356, opened 2024-02-28
MinMDNS skip interfaces without IPv4 addresses

agners:
IPv4 multicast require an IPv4 address to be present on a particular interface. Skip interfaces without IPv4 addresses in the default address policy. This avoids errors when trying to join the multicast group later on:

```
MDNS failed to join multicast group on veth3cdf62f for address type IPv4: src/inet/UDPEndPointImplSockets.cpp:777: Inet Error 0x00000110: Address not found
```

### DIFF
--- a/src/lib/dnssd/minimal_mdns/AddressPolicy_LibNlImpl.cpp
+++ b/src/lib/dnssd/minimal_mdns/AddressPolicy_LibNlImpl.cpp
@@ -48,6 +48,7 @@ private:
 
     nl_sock * mNlSocket            = nullptr;
     nl_cache * mNlCache            = nullptr;
+    nl_cache * mNlAddrCache        = nullptr;
     nl_object * mCurrentLink       = nullptr;
     IPAddressType mCurrentLinkType = IPAddressType::kUnknown;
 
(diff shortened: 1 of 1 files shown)


andy31415 (approved):

src/lib/dnssd/minimal_mdns/AddressPolicy_LibNlImpl.cpp
```diff
@@ -176,14 +182,42 @@ bool AllListenIterator::Next(InterfaceId * id, IPAddressType * type)
...
-        if (idx == 0)
+        int ifindex = rtnl_link_get_ifindex(CurrentLink());
+        if (ifindex == 0)
         {
             // Invalid index, move to the next interface
             continue;
         }
 
-        *id   = InterfaceId(idx);
+        // For IPv4, report only interfaces which have an IPv4 address
```
tcarmelveilleux, line 192: Will this make it so that an interface with only an IPv6 address will be skipped for v6 as well, just because IPv4 is enabled?

agners: No. The code always checks IPv6 then IPv4 before moving to the next link. This change only affects IPv4.

I could add ifdefs here, but it should not matter in practice.

woody-apple (approved):

PR #32357, opened 2024-02-28
Make use of build targets for chip_data_model generated code

andy31415:
This builds on #32346. 

## Changes

- make `privilege-storage` (which depends on access.h generated file) be a stand-alone target in CI
- make sure applications define a `chip_static_data_model_BACKEND` that specifies what data model gets compiled into src/app models.
- created a separate set of targets where the BACKEND points to by default which will always fail with a message. This both serves to quickly see what the problem is when compiling as well as to allow `bootstrap` to work as bootstrap resolves all dependencies and things in app depend on the backend.

The intent is to migrate away from implicit `data_model` including everything and instead to have bespoke targets. Started with privilege-storage since it seemed one of the simpler ones. main delta of this is to define chip_static_data_model_BACKEND everywhere.


andy31415:
After changing 100s of files for data model defaults, it looks to me that this has to be done incrementally - a one-change PR will not work. The dependencies are also very mixed:
   - chip_data_model includes `.cpp` files with ember-specific implementations
   - Expectation seems to be that `src/app` sources (e.g. interaction-model and app) use the forward-declared functions in headers (e.g. privilege-storage, attribute-storage) while the main app links the corresponding data model implementing things
   - as a result, having a single backend per build is currently not easily splittable:
      - some targets must have no backend set (e.g. qpg lib, integration tests provide their own)
      - gn_build.sh tries to build several targets at once with different models:
         - shell does not care, but needs something
         - chip-tool is unclear, however it probably needs something with OTA support.
         - im-initiator/responder/other require to provide their own
         
Closing this large PR for now and will look at incrementally adding support for dependencies. It looks like headers and sources must be split here as:
   - it seems impossible to have a "single backend per build" (even though that would be very much preferable...)
   - multi-backend support will need it's approach defined.

PR #32362, opened 2024-02-28
Fix incorrectly reported NoReponse when single batch command sent to …

### LINKED ISSUE #32360: OnNoResponse is always called using extendable invoke against the old device that don't support batch
when CommandRef is not in CommandResponseIB and total command in this transaction = 1 and onResponse is received, you may suppress onNoResponse

tehampson:
When new CommandSender APIs is used to send a single invoke request where `CommandRef` was provided in the request to a legacy device, CommandSender would incorrectly call the NoResposeCallback.

This occurs because legacy device is allowed to drop CommandRef (as it doesn't even know to look for CommandRef, nor does it know what it means).

Fix by assuming if getting a response to a single request, we assume the response matches the single pending tracked command.

fixes: https://github.com/project-chip/connectedhomeip/issues/32360

### DIFF
--- a/src/app/CommandSender.cpp
+++ b/src/app/CommandSender.cpp
@@ -441,7 +441,8 @@ CHIP_ERROR CommandSender::ProcessInvokeResponseIB(InvokeResponseIB::Parser & aIn
             }
         }
 
-        if (!commandRef.HasValue() && !commandRefRequired && mpPendingResponseTracker != nullptr && mpPendingResponseTracker->Count() == 1)
+        if (!commandRef.HasValue() && !commandRefRequired && mpPendingResponseTracker != nullptr &&
+            mpPendingResponseTracker->Count() == 1)
         {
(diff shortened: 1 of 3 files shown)


tcarmelveilleux (approved):

src/app/CommandSender.cpp
```diff
@@ -441,6 +441,14 @@ CHIP_ERROR CommandSender::ProcessInvokeResponseIB(InvokeResponseIB::Parser & aIn
             }
         }
 
+        if (!commandRef.HasValue() && !commandRefRequired && mpPendingResponseTracker != nullptr && mpPendingResponseTracker->Count() == 1)
```
tcarmelveilleux, line 444: Please add a test to cover detecting a regression of this PR's fix.

yunhanw-google (approved):

src/app/CommandSender.cpp
tehampson: Test has been added

woody-apple (approved):

PR #32367, opened 2024-02-29
Update the JSON<->TLV Converter doc to align with implementation

yufengwangca:
Update the JSON<->TLV Converter doc to align with implementation


yunhanw-google (approved):

andy31415:
fast tracking doc only change.

PR #32373, opened 2024-02-29
[Matter.framework] Add back downloadLogOfType onto MTRBaseDevice and MTRDevice

vivien-apple:
Add back what has been reverted in https://github.com/project-chip/connectedhomeip/pull/32319 and expose the API to `MTRDevice` as well.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice.h
+++ b/src/darwin/Framework/CHIP/MTRDevice.h
@@ -337,6 +337,27 @@ MTR_AVAILABLE(ios(16.1), macos(13.0), watchos(9.1), tvos(16.1))
  */
 - (void)removeClientDataForKey:(NSString *)key endpointID:(NSNumber *)endpointID MTR_UNSTABLE_API;
 
+/**
+ * Download log of the desired type from the device.
+ *
+ * Note: The consumer of this API should move the file that the url points to or open it for reading before the
+ * completion handler returns. Otherwise, the file will be deleted, and the data will be lost.
+ *
+ * @param type       The type of log being requested. This should correspond to a value in the enum MTRDiagnosticLogType.
+ * @param timeout    The timeout for getting the log. If the timeout expires, completion will be called with whatever
+ *                   has been retrieved by that point (which might be none or a partial log).
+ *                   If the timeout is set to 0, the request will not expire and completion will not be called until
+ *                   the log is fully retrieved or an error occurs.
+ * @param queue      The queue on which completion will be called.
+ * @param completion The completion handler that is called after attempting to retrieve the requested log.
+ *                     - In case of success, the completion handler is called with a non-nil URL and a nil error.
+ *                     - If there is an error, a non-nil error is used and the url can be non-nil too if some logs have already been downloaded.
+ */
+- (void)downloadLogOfType:(MTRDiagnosticLogType)type
+                  timeout:(NSTimeInterval)timeout
+                    queue:(dispatch_queue_t)queue
+               completion:(void (^)(NSURL * _Nullable url, NSError * _Nullable error))completion
+    MTR_NEWLY_AVAILABLE;
 @end
 
 MTR_EXTERN NSString * const MTRPreviousDataKey MTR_NEWLY_AVAILABLE;
--- a/src/darwin/Framework/CHIP/MTRDevice.h
+++ b/src/darwin/Framework/CHIP/MTRDevice.h
@@ -18,7 +18,6 @@
 #import <Foundation/Foundation.h>
 #import <Matter/MTRBaseDevice.h>
 #import <Matter/MTRDefines.h>
-#import <Matter/MTRDiagnosticLogsType.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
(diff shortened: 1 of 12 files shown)


src/darwin/Framework/CHIP/MTRDevice.h
```diff
@@ -329,9 +328,6 @@ MTR_AVAILABLE(ios(16.1), macos(13.0), watchos(9.1), tvos(16.1))
 /**
  * Download log of the desired type from the device.
  *
- * Note: The consumer of this API should move the file that the url points to or open it for reading before the
- * completion handler returns. Otherwise, the file will be deleted, and the data will be lost.
```
bzbarsky-apple, line 333: Why is this being removed?  Is this no longer true?  Why does the MTRBaseDevice version not have this?

src/darwin/Framework/CHIP/MTRDeviceController+XPC.h
```diff
@@ -183,6 +184,15 @@ typedef void (^MTRValuesHandler)(id _Nullable values, NSError * _Nullable error)
                              attributeId:(NSNumber * _Nullable)attributeId
                               completion:(MTRValuesHandler)completion;
 
+/**
+ * Requests downloading some logs
+ */
+- (void)downloadLogWithController:(id _Nullable)controller
+                           nodeId:(uint64_t)nodeId
```
bzbarsky-apple, line 191: Is this public API?  If so, it should use our normal public API naming conventions.  And yes, I know other things on here don't.

Also, if this is public API does it need availability annotations?

src/darwin/Framework/CHIP/MTRDiagnosticLogsDownloader.mm
```diff
@@ -165,7 +165,6 @@ - (instancetype)initWithType:(MTRDiagnosticLogType)type
                     // data in the logs that the caller may find useful. For this reason, fileURL is passed in even
                     // when there is an error but fileHandle is not nil.
                     completion(strongSelf->_fileHandle ? fileURL : nil, bdxError);
-                    [strongSelf deleteFile];
```
bzbarsky-apple, line 168: So what deletes the file and when?

woody-apple: Agree with @bzbarsky-apple why is this removed?

src/darwin/Framework/CHIP/MTRDeviceOverXPC.mm
```diff
@@ -427,6 +427,38 @@ - (void)openCommissioningWindowWithDiscriminator:(NSNumber *)discriminator
...
+                                         nodeId:self.nodeID.unsignedLongLongValue
+                                           type:type
+                                        timeout:timeout
+                                     completion:^(NSString * _Nullable url, NSError * _Nullable error) {
+                                         dispatch_async(queue, ^{
+                                             MTR_LOG_DEBUG("Download log");
+                                             completion([NSURL URLWithString:url], error);
+                                             // The following captures the proxy handle in the closure so that the
+                                             // handle won't be released prior to block call.
+                                             __auto_type handleRetainer = handle;
```
ksperling-apple, line 453: This is an odd construction.. why not just `(void)handle;` ?

src/darwin/Framework/CHIPTests/MTRXPCListenerSampleTests.m
```diff
@@ -1865,6 +1887,30 @@ - (void)test015_MTRDeviceInteraction
...
+                      timeout:10
+                        queue:queue
+                   completion:^(NSURL * _Nullable url, NSError * _Nullable error) {
+                       NSLog(@"downloadLogOfType: url: %@, error: %@", url, error);
+                       XCTAssertNil(error);
+
+                       NSError * readError;
+                       NSString * fileContent = [NSString stringWithContentsOfURL:url encoding:NSUTF8StringEncoding error:&readError];
+                       XCTAssertNil(readError);
+                       XCTAssertTrue([fileContent isEqualToString:@"This is a simple log\n"]);
```
ksperling-apple, line 1908: Please use `XCTAssertEqualObjects` for equality assertions (it provides a useful error message on failure)

src/darwin/Framework/CHIP/MTRDeviceOverXPC.mm
```diff
@@ -427,6 +427,38 @@ - (void)openCommissioningWindowWithDiscriminator:(NSNumber *)discriminator
...
+{
+    MTR_LOG_DEBUG("Downloading log ...");
+
+    __auto_type workBlock = ^(MTRDeviceControllerXPCProxyHandle * _Nullable handle, NSError * _Nullable error) {
+        if (error != nil) {
+            completion(nil, error);
+            return;
+        }
+
+        [handle.proxy downloadLogWithController:self.controllerID
```
ksperling-apple, line 443: How does the XCP client process have access to the file? Is there some magic that happens whenever an NSURL is passed through XCP? Or does it just work in our tests because we happen do be doing it within the same process / sandbox?

woody-apple: We’ll solve the sandbox stuff later when we officially support this

src/darwin/Framework/CHIP/MTRDeviceController+XPC.h
woody-apple: Agree it needs newly available, but base device appears to always use these number conventions

woody-apple: Let’s keep with what we have, as we want to slowly go away from base device

src/darwin/Framework/CHIP/MTRDeviceOverXPC.mm
vivien-apple: That is because this is how it is done elsewhere.  I would need to investigate but I suspect this is to avoid some compiler optimisation.

src/darwin/Framework/CHIP/MTRDeviceController+XPC.h
bzbarsky-apple: >  but base device appears to always use these number conventions

I'm not sure what you mean here.....

woody-apple (approved):

PR #32380, opened 2024-02-29
Fix incorrect snapshotting of MRP config by CASE client.

bzbarsky-apple:
Right now, we snapshot the MRP config as part of the CASEClientInitParams during stack startup.

After that, we will use that snapshotted config whenever we are the CASE initiator.

This config will not match the parameters we use as CASE responder or advertise over DNS-SD if the local MRP configuration ever changes.  Which for an ICD it can.

The fix is to stop the incorrect snapshotting and get the information we need from the right source of truth when we need it.

### DIFF
--- a/src/app/CASEClient.cpp
+++ b/src/app/CASEClient.cpp
@@ -16,6 +16,7 @@
  */
 
 #include <app/CASEClient.h>
+#include <messaging/ReliableMessageProtocolConfig.h>
 
 namespace chip {
 
@@ -49,10 +50,12 @@ CHIP_ERROR CASEClient::EstablishSession(const CASEClientInitParams & params, con
     Messaging::ExchangeContext * exchange = params.exchangeMgr->NewContext(session.Value(), &mCASESession);
     VerifyOrReturnError(exchange != nullptr, CHIP_ERROR_INTERNAL);
 
+    const Optional<ReliableMessageProtocolConfig> & mrpLocalConfig =
+        params.mrpLocalConfig.HasValue() ? params.mrpLocalConfig : GetLocalMRPConfig();
     mCASESession.SetGroupDataProvider(params.groupDataProvider);
     ReturnErrorOnFailure(mCASESession.EstablishSession(*params.sessionManager, params.fabricTable, peer, exchange,
                                                        params.sessionResumptionStorage, params.certificateValidityPolicy, delegate,
-                                                       params.mrpLocalConfig));
+                                                       mrpLocalConfig));
 
     return CHIP_NO_ERROR;
 }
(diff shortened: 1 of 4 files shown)


src/app/CASEClient.h
```diff
@@ -34,7 +34,6 @@ struct CASEClientInitParams
     Messaging::ExchangeManager * exchangeMgr                           = nullptr;
     FabricTable * fabricTable                                          = nullptr;
     Credentials::GroupDataProvider * groupDataProvider                 = nullptr;
-    Optional<ReliableMessageProtocolConfig> mrpLocalConfig             = Optional<ReliableMessageProtocolConfig>::Missing();
```
bzbarsky-apple, line 37: @tcarmelveilleux if you feel this is not compatible enough, we could leave the field and either ignore it or start failing if it's set to to a non-missing value....

tcarmelveilleux: If missing, it should use the local config.

bzbarsky-apple: Or I guess if it's set to something other than `GetLocalMRPConfig()` at the point when init happens.

bzbarsky-apple: > If missing, it should use the local config.

That is already the behavior, yes.

mkardous-silabs (approved):

src/app/CASEClient.h
```diff
@@ -34,7 +34,11 @@ struct CASEClientInitParams
...
+
+    // mrpLocalConfig should not generally be set to anything other than
+    // NullOptional.  Doing that can lead to different parts of the system
+    // claiming different MRP parameters for the same node.
```
tcarmelveilleux, line 40:
```suggestion
    // `mrpLocalConfig` should not generally be set to anything other than
    // NullOptional.  Doing that can lead to different parts of the system
    // claiming different MRP parameters for the same node. This remains
    // as backwards compatibility to previous public API.
```

tcarmelveilleux (approved):

src/app/CASEClient.cpp
```diff
@@ -49,10 +50,12 @@ CHIP_ERROR CASEClient::EstablishSession(const CASEClientInitParams & params, con
     Messaging::ExchangeContext * exchange = params.exchangeMgr->NewContext(session.Value(), &mCASESession);
     VerifyOrReturnError(exchange != nullptr, CHIP_ERROR_INTERNAL);
 
+    const Optional<ReliableMessageProtocolConfig> & mrpLocalConfig =
+        params.mrpLocalConfig.HasValue() ? params.mrpLocalConfig : GetLocalMRPConfig();
```
bzbarsky-apple, line 54: This part is needed because CASESession uses GetDefaultMRPConfig() if it gets nothing, since GetLocalMRPConfig() can still return nothing (and that means default).

tcarmelveilleux (approved):

tehampson (approved):

PR #32381, opened 2024-02-29
Clean up some includes in DataModelHandler

andy31415:
Clean up includes and useless file comment.

Overall the naming of this file is somewhat questionable as it contains a single init call. Still looking to determine if we should rename (we probably should).


plauric (approved):

bzbarsky-apple (approved):
There used to be more things in this file....

PR #32384, opened 2024-02-29
[Python] Add ruff as Python linter

agners:
This adds Ruff as Python linter. Ruff is an extremely fast Python linter and code formatter, written in Rust. Currently only it's linter capabilities are used. It replace flake8 currently in that capacity. Compared to flake8 it is much faster, and in the future additional checks can be enabled.

### DIFF
--- a/integrations/docker/images/base/chip-build/Dockerfile
+++ b/integrations/docker/images/base/chip-build/Dockerfile
@@ -144,7 +144,6 @@ RUN set -x \
     click \
     coloredlogs \
     cxxfilt \
-    flake8 \
     future \
     ghapi \
     mobly \
--- a/.flake8
+++ b/.flake8
@@ -1,6 +1,0 @@
-[flake8]
-max-line-length = 132
-exclude = third_party
-          .*
-          out/*
-          ./examples/common/QRCode/*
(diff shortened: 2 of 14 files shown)


agners:
NOTE: This requires a new chip-build release first. I am not sure what the process is to update the container. Should I create a separate PR to add `ruff` to the container first?

yunhanw-google:
@andreilitvin can help update the chip build container image.

integrations/docker/images/base/chip-build/Dockerfile
```diff
@@ -152,6 +152,7 @@ RUN set -x \
     portpicker \
     pygit \
     PyGithub \
+    ruff \
```
andy31415, line 155: @agners since you change an image, you have to update the version file as well.

I would split this into two PS:
- first to change DockerFile and version
- 2nd that makes use of it

agners: Step 1: #32388

andy31415: Can we remove flake8? I understand it includes flake8 checks: site says `Drop-in parity with [Flake8](https://docs.astral.sh/ruff/faq/#how-does-ruff-compare-to-flake8), isort, and Black`

agners: Yes, right, that make sense!

andy31415 (approved):

Damian-Nordic (commented):
Could you add a proper PR description why the change is being made etc.?

agners:
> Could you add a proper PR description why the change is being made etc.?

Good point, I've added a description to the PR.

Note that I am not sure yet if it (is/will be) a good idea to use its formatting capabilities. The repo is currently using restyled, for formatting, and Ruff is (not yet) supported.

Damian-Nordic (approved):

PR #32386, opened 2024-03-01
[ICD] Add missing include for the ICD specific defines

mkardous-silabs:
#### Description
Several sources files required the ICDServerConfig.h for the ICD defines.
The header was not explicitly included in the file but was included throught another include path/.

To avoid any inclusion problems in the feature, explicitly include the necessary header in the source file.

#### Tests
No necessary tests


jmartinez-silabs (approved):

yunhanw-google (approved):

yunhanw-google (commented):
You need modify the gn files where it needs depends on src/app/icd/server:icd-server-config

bzbarsky-apple (approved):

Issue #31405, opened 2024-01-13
[IM] Missing support for AttributeAccessInterface to report supported attributes


tcarmelveilleux:
CommandHandlerInterface supports listing the supported commands and using that knowledge within dispatch to properly handle the code-backed logic of a cluster for only available commands.

We are missing the equivalent for AttributeAccessInterface, which would allow dynamic reporting of supported attributes, such as for cases like NetworkCommissioningCluster that may have a set of attributes that is only known at runtime.

FYI @bzbarsky-apple

PR #32414, opened 2024-03-04
Introduce AttributeAccessInterface lookup cache

tcarmelveilleux:
- Running TC_DeviceBasicComposition test caused, for a single Wildcard read, 44622 iterations through the AttributeAccessInterface (AAI) lookup loop. This is significant runtime cost on embedded devices, when a wildcard subscription is established or when trying to iterate over changes of attributes.
- Instrumented analysis showed that significant locality exists within the lookup, which could be exploited by a cache.
- This PR introduces a cache, and its use, in AAI lookups. On same workload as the initial investigation, the total number of iterations of the costly loop went from 44622 --> 4864, a factor of almost 10, for very little additional RAM usage (single entry in cache).

Issue #31405

Testing done:
- Added unit tests
- Integration tests still pass
- TC_BasicDeviceComposition still succeeds

### DIFF
--- a/src/app/AttributeAccessInterfaceCache.h
+++ b/src/app/AttributeAccessInterfaceCache.h
@@ -1,0 +1,146 @@
+/*
+ *
+ *    Copyright (c) 2024 Project CHIP Authors
+ *    All rights reserved.
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+#pragma once
+
+#include <stddef.h>
+
+#include <app/AttributeAccessInterface.h>
+#include <lib/core/DataModelTypes.h>
+
+namespace chip {
+namespace app {
+
+/**
+ * @brief Cache to make look-up of AttributeAccessInterface (AAI) instances faster.
+ *
+ * @tparam N - The size of the cache (current suggested value == 1)
+ *
+ * This cache makes use of the fact that looking-up AttributeAccessInterface
+ * instances is usually done in loops, during read/subscription wildcard
+ * expansion, and there is a significant amount of locality.
+ *
+ * This cache records both "used" (i.e. uses AAI) and the single last
+ * "unused" (i.e. does NOT use AAI) entries. Combining positive/negative
+ * lookup led to factor of ~10 reduction of AAI lookups in total for wildcard
+ * reads on chip-all-clusters-app, with a cache size of 1. The size did not
+ * significantly improve the performance, but `N` is left to support better
+ * hashing/storage algorithms in the future if needed.
+ */
+template <size_t N>
+class AttributeAccessInterfaceCache
+{
+  public:
+    AttributeAccessInterfaceCache() { Invalidate(); }
+
+    /**
+     * @brief Invalidate the whole cache. Must be called every time list of AAI registrations changes.
+     */
+    void Invalidate()
+    {
+        for (auto & entry : mCacheSlots)
+        {
+          entry.Invalidate();
+        }
+        mLastUnusedEntry.Invalidate();
+    }
+
+    /**
+     * @brief Mark that we know a given <`endpointId`, `clusterId`> uses AAI, with instance `attrInterface`
+     */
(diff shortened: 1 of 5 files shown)


src/app/AttributeAccessInterfaceCache.h
```diff
@@ -0,0 +1,139 @@
...
+
+        return nullptr;
+    }
+
+    /**
+     * @brief returns true if <`endpointId`, `clusterId`> was last marked as NOT to using AAI.
+     *
+     * May return false even though it doesn't use AAI, on cache miss of unused slot.
+     */
+    bool IsUnused(EndpointId endpointId, ClusterId clusterId) const { return mLastUnusedEntry.Matches(endpointId, clusterId); }
```
andy31415, line 98: The semantics of guaranteed no false positive but not guaranteed false negatives in `usesAAI` seem not easy to follow. It looks like this class is a cache that knows some subset of `is using AAI` queries and may return true/false/unknown

I generally would imagine a `Get()` that returns the above 3 possibilies. Is there any chance we could use some variant/way Get() and return this type? like:

```cpp
enum class CacheState {
   kUnknown, // no info about it
   kUsingAccessInterface, // known using AAI
   kNotUsingAccessInterface, // known NOT to use AAI
};

/// gets the known status of endpoint/cluster
///
/// if AAI is used, return the pointer to it if the last arg is not nullptr.
CacheState Get(EndpointId, ClusterId, AttributeAccessInterface**);
```

src/app/AttributeAccessInterfaceCache.h
```diff
@@ -0,0 +1,139 @@
...
+
+    AttributeAccessCacheEntry * GetCacheSlot(EndpointId endpointId, ClusterId clusterId)
+    {
+        // A future implementation may use a better method.
+        size_t slotId = (clusterId ^ endpointId) & (N - 1);
+        return &mCacheSlots[slotId];
+    }
+
+    AttributeAccessCacheEntry mCacheSlots[N];
+    AttributeAccessCacheEntry mLastUnusedEntry;
```
andy31415, line 135: how about N/M here instead of N/1? Or does that incur extra cost?

src/app/util/attribute-storage.cpp
```diff
@@ -136,6 +137,7 @@ constexpr const EmberAfDeviceType fixedDeviceTypeList[]             = FIXED_DEVI
 DataVersion fixedEndpointDataVersions[ZAP_FIXED_ENDPOINT_DATA_VERSION_COUNT];
 
 AttributeAccessInterface * gAttributeAccessOverrides = nullptr;
+AttributeAccessInterfaceCache<1> gAttributeAccessInterfaceCache;
```
andy31415, line 140: nit: I would place a `kAttributeAccessInterfaceCacheSize` here to make it more visible as a set constant.

src/app/AttributeAccessInterfaceCache.h
tcarmelveilleux: It does incur cost, and does not significantly impact performance

src/app/util/attribute-storage.cpp
tcarmelveilleux: Removed the template.

src/app/AttributeAccessInterfaceCache.h
```diff
@@ -0,0 +1,145 @@
...
+ * @brief Cache to make look-up of AttributeAccessInterface (AAI) instances faster.
+ *
+ * This cache makes use of the fact that looking-up AttributeAccessInterface
+ * instances is usually done in loops, during read/subscription wildcard
+ * expansion, and there is a significant amount of locality.
+ *
+ * This cache records both "used" (i.e. uses AAI) and the single last
+ * "unused" (i.e. does NOT use AAI) entries. Combining positive/negative
+ * lookup led to factor of ~10 reduction of AAI lookups in total for wildcard
+ * reads on chip-all-clusters-app, with a cache size of 1. The size did not
```
bzbarsky-apple, line 38:
```suggestion
 * reads on chip-all-clusters-app, with a cache size of 1. Increasing the size did not
```
assuming that's what you mean?

src/app/AttributeAccessInterfaceCache.h
```diff
@@ -0,0 +1,145 @@
...
+
+    AttributeAccessCacheEntry * GetCacheSlot(EndpointId endpointId, ClusterId clusterId)
+    {
+        (void) endpointId;
+        (void) clusterId;
+        return &mCacheSlots[0];
+    }
+
+    AttributeAccessCacheEntry mCacheSlots[1];
+    AttributeAccessCacheEntry mLastUnusedEntry;
```
bzbarsky-apple, line 141: mLastUnusedEntry could probably be a ConcreteClusterPath and save one word of RAM... Either way, though.

bzbarsky-apple (approved):

src/app/AttributeAccessInterfaceCache.h
tcarmelveilleux: Then I would have to change the logic to invalidate, to be less regular. I will leave this one as-is.

andy31415 (approved):

PR #32422, opened 2024-03-04
Device composition tests: Add option to run tests from file

cecille:
This PR allows us to ingest data dumps from previous runs, and run tests against them. This works for device basic composition tests, device conformance tests, and pics tests.

#### Testing
Added a run to the device tests to run the test from file


src/controller/python/JsonTlvSupport.cpp
```diff
@@ -0,0 +1,48 @@
...
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+/**
+ *    @file
+ *      Support for JsonTlv.
```
andy31415, line 21: lets remove this doc-comment ... it does not seem helpful: says "support for jsontlv" in a file named "JsonTlvSupport.cpp".

andy31415: overall I wonder if the naming could be updated: support/helper/core and such are filler words that end up encouraging a mix of things to go into a single file.

maybe just `json_tlv.cpp` is sufficient.

src/controller/python/JsonTlvSupport.cpp
```diff
@@ -0,0 +1,48 @@
...
+//    size - output buffer size
+// Returns:
+//    Number of TLV bytes encoded. If this fails, 0 will be returned.
+size_t pychip_JsonToTlv(const char * json, uint8_t * buf, size_t size)
+{
+    MutableByteSpan tlvSpan(buf, size);
+    CHIP_ERROR err = JsonToTlv(json, tlvSpan);
+    if (err != CHIP_NO_ERROR)
+    {
+        return 0;
```
andy31415, line 44: should we at least log an error if it occurs? Otherwise this may be very hard to debug.

Even better if we can return a `PyChipError` that can be processed instead.

andy31415:
This is an old PR  that has merge conflicts. Assuming stale and closing.

src/controller/python/JsonTlvSupport.cpp
cecille: done.

src/controller/python/JsonTlvSupport.cpp
cecille: Added a log - I still like this passing back the size.

ReneJosefsen (approved):

src/controller/python/matter/ChipDeviceCtrl.py
```diff
@@ -3101,3 +3103,54 @@ def __init__(self, operationalKey: p256keypair.P256Keypair, noc: bytes,
         self._set_dev_ctrl(devCtrl, pairingDelegate)
 
         self._finish_init()
+
+
+class TLVJsonConverter():
+    ''' Converter class used to convert dumped Matter JsonTlv files into a cache. '''
```
andy31415: Description could be improved maybe. Reading it as a stand-alone, I am asking:
 - "what dumped matter jsontlv files? who generates those and how"
 - "what cache? 'a cache' seems vague"

I imagine an explanation of this receiving values in JsonTlv (explain why we have that as the format used ... what procedure dumps such a file?) and that we return an AttributeCache object would help.

src/controller/python/matter/ChipDeviceCtrl.py
```diff
@@ -3101,3 +3103,54 @@ def __init__(self, operationalKey: p256keypair.P256Keypair, noc: bytes,
...
+            Input:
+              json_tlv: json loaded from from the dump file.
+            Returns:
+              AttributeCache with the data from the json_string
+        '''
+        cache = AttributeCache()
+        for endpoint_id_str, endpoint in json_tlv.items():
+            endpoint_id = int(endpoint_id_str, 0)
+            for cluster_id_str, cluster in endpoint.items():
+                s = cluster_id_str.split(':')
```
andy31415: a comment explaining the data format here would help (or reference to jsontlv format), to explain why we split by `:`.

Maybe a link to https://github.com/project-chip/connectedhomeip/blob/master/src/lib/support/jsontlv/README.md is sufficient and some example here, so while we read the code we can follow it for the example.

andy31415: I wonder if something like:

```py
id_str, name = cluster_id_str.split(':', 2)
cluster_id =  int(id_str)
```

would be clearer to see what the parts split by : are.then for attributes we would split by ID and type I imagine.

src/controller/python/matter/ChipDeviceCtrl.py
andy31415: Is there a reason we have this in ChipDeviceCtrl.py instead of a separate stand-alone python file?

src/python_testing/matter_testing_infrastructure/matter/testing/basic_composition.py
```diff
@@ -170,6 +176,19 @@ def dump_wildcard(self, dump_device_composition_path: typing.Optional[str]) -> t
     async def setup_class_helper(self, allow_pase: bool = True):
         dev_ctrl = self.default_controller
         self.problems: list[ProblemNotice] = []
+        self.test_from_file = self.user_params.get("test_from_file", None)
```
andy31415, line 180: Some documentation somwhere of how one gets a `test_from_file` dump to use would be good.

andy31415 (approved):
Approving: mostly asking for documentation updates, however functionality seems reasonable. Updates can be here or in a followup.

src/controller/python/matter/ChipDeviceCtrl.py
cecille: nope, brain just defaulted. I've added more documentation and moved this into its own file now.

src/python_testing/matter_testing_infrastructure/matter/testing/basic_composition.py
cecille: Going to handle this in a separate PR so I can add in all the information for these tests. https://github.com/project-chip/connectedhomeip/pull/40973

src/controller/python/matter/ChipDeviceCtrl.py
cecille: changed (both clusters and attributes)

PR #32424, opened 2024-03-05
Allow multiple days to be specified in DaysMask parameter of SetWeekdaySchedule command

### LINKED ISSUE #32423: [DRLK] SetWeekdaySchedule only allows single day to be specified in DaysMask parameter
### Reproduction steps

The current SDK implementation of door-lock-server's SetWeekdaySchedule command only allows a single bit to be set in the DaysMask parameter. This appears to be a mis-interpretation of the spec. Specifically, the spec says that the SetWeekdaySchedule command shall return INVALID COMMAND if either of the following conditions is true:
- One or more fields violates constraints or is invalid.
- Door lock is unable to set Week Day schedule for more than one day in DaysMask map (e.g. need to use separate schedules for each day).

The second of the above conditions appears to be misconstrued to mean that DaysMask shall only allow a single day to be specified.

### Bug prevalence

Consistent

main

### Platform

core

### Platform Version(s)

_No response_

### Anything else?

_No response_

jrhees-cae:
Allow multiple days to be specified in DaysMask parameter of SetWeekdaySchedule command

Original SDK implementation misconstrues the spec and disallows anything but single-day schedules.

Fixes #32423

### DIFF
--- a/src/app/clusters/door-lock-server/door-lock-server.cpp
+++ b/src/app/clusters/door-lock-server/door-lock-server.cpp
@@ -1003,7 +1003,7 @@ void DoorLockServer::setWeekDayScheduleCommandHandler(chip::app::CommandHandler 
         return;
     }
 
-    uint8_t rawDaysMask   = daysMask.Raw();
+    uint8_t rawDaysMask = daysMask.Raw();
 
     // Check that bits are within range
     if ((0 == rawDaysMask) || (rawDaysMask & 0x80))
--- a/src/app/clusters/door-lock-server/door-lock-server.cpp
+++ b/src/app/clusters/door-lock-server/door-lock-server.cpp
@@ -1003,17 +1003,10 @@ void DoorLockServer::setWeekDayScheduleCommandHandler(chip::app::CommandHandler 
         return;
     }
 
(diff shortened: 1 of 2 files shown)


bzbarsky-apple (approved):

src/app/clusters/door-lock-server/door-lock-server.cpp
```diff
@@ -1003,17 +1003,10 @@ void DoorLockServer::setWeekDayScheduleCommandHandler(chip::app::CommandHandler
...
-    {
-        setBitsInDaysMask += rawDaysMask & 0x1;
-        rawDaysMask = static_cast<uint8_t>(rawDaysMask >> 1);
-    }
+    uint8_t rawDaysMask = daysMask.Raw();
 
-    // TODO: Check that bits are within range
-    if (setBitsInDaysMask == 0 || setBitsInDaysMask > 1)
+    // Check that bits are within range
+    if ((0 == rawDaysMask) || (rawDaysMask & 0x80))
```
bzbarsky-apple, line 1009: Ideally the call into emberAfPluginDoorLockSetSchedule would allow the app to communicate the "can't handle more than one bit, return InvalidCommand" situation if it exists.  But that can be a followup.

yufengwangca (approved):

PR #32425, opened 2024-03-05
[Darwin] Refactor MTRMetrics to have more well formed schema

anush-apple:
- MTRMetrics vends MTRMetricData which has well defined schema
- Cleaned up internals of MTRMetricData
- Fixed potential unbounded growth if metrics are enabled on Darwin
- Removed stale metric keys that are not used and were meant to be removed in initial framework for metrics

### DIFF
--- a/src/darwin/Framework/CHIP/MTRMetrics.h
+++ b/src/darwin/Framework/CHIP/MTRMetrics.h
@@ -21,6 +21,31 @@
 NS_ASSUME_NONNULL_BEGIN
 
 /**
+ * Representation of metric data corresponding to a metric event.
+ */
+MTR_NEWLY_AVAILABLE
+@interface MTRMetricData : NSObject
+
+/**
+ * Value for the metric data. The value may be nil depending on the event emitted.
+ */
+@property (nonatomic, nullable, readonly, copy) NSNumber * value;
+
+/**
+ * Error code for the metric data. This value when valid, holds the error code value
+ * of the operation associated with the event.
+ */
+@property (nonatomic, nullable, readonly, copy) NSNumber * errorCode;
+
+/**
+ * Duration of event associated with the metric. This value may be nil depending on
+ * the event emitted.
+ */
+@property (nonatomic, nullable, readonly, copy) NSNumber * durationMicroseconds;
+
+@end
+
+/**
  * A representation of a collection of metrics data for an operation.
  */
 MTR_NEWLY_AVAILABLE
@@ -35,13 +60,13 @@ MTR_NEWLY_AVAILABLE
 @property (nonatomic, readonly, copy) NSArray<NSString *> * allKeys;
 
 /**
- * @brief Returns metric object corresponding to the metric identified by its key
+ * @brief Returns metric data corresponding to the metric identified by its key.
  *
  * @param [in] key Name of the metric
  *
- * @return An object containing the metric data, nil if key is invalid
+ * @return An object containing the metric data, nil if key is invalid.
  */
-- (nullable id)valueForKey:(NSString *)key;
+- (nullable MTRMetricData *)metricDataForKey:(NSString *)key;
 
 @end
 
--- a/src/darwin/Framework/CHIP/MTRMetricsCollector.mm
+++ b/src/darwin/Framework/CHIP/MTRMetricsCollector.mm
@@ -18,6 +18,7 @@
 #import "MTRMetricsCollector.h"
 #import "MTRLogging_Internal.h"
 #include "MTRMetrics_Internal.h"
+#include <MTRMetrics.h>
 #import <MTRUnfairLock.h>
 #include <platform/Darwin/Tracing.h>
 #include <system/SystemClock.h>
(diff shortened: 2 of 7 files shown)


src/tracing/metric_keys.h
```diff
@@ -29,16 +29,7 @@ typedef const char * MetricKey;
 /**
  * List of supported metric keys
  */
-constexpr MetricKey kMetricDiscoveryOverBLE      = "disc-over-ble";
-constexpr MetricKey kMetricDiscoveryOnNetwork    = "disc-on-nw";
-constexpr MetricKey kMetricPASESession           = "pase-session";
-constexpr MetricKey kMetricPASESessionPair       = "pase-session-pair";
-constexpr MetricKey kMetricPASESessionBLE        = "pase-session-ble";
```
anush-apple, line 36: These keys should have been removed from the initial PR that put the metric framework up. These are not used and we should only add keys that are actually used commonly in the SDK here.

woody-apple (approved):

woody-apple (approved):

src/darwin/Framework/CHIP/MTRMetricsCollector.mm
```diff
@@ -41,60 +46,50 @@ - (instancetype)initWithMetricEvent:(const MetricEvent &)event
...
-        [dictRepresentation setValue:self.value forKey:kMTRMetricDataValueKey];
-    }
-    if (auto tmPt = self.timePointMicroseconds) {
-        [dictRepresentation setValue:tmPt forKey:kMTRMetricDataTimepointKey];
-    }
-    if (auto duration = self.durationMicroseconds) {
-        [dictRepresentation setValue:duration forKey:kMTRMetricDataDurationKey];
-    }
-    return dictRepresentation;
+    return [NSString stringWithFormat:@"<MTRMetricData>: Type %d, Value = %@, Error Code = %@, Duration = %@ us",
```
bzbarsky-apple, line 91: My understanding is that the closing `>` should be after the member details, no?  So `<MTRMetricData: Type %d, Value = %@, Error Code = %@, Duration = %@ us>`

src/darwin/Framework/CHIP/MTRMetrics.h
```diff
@@ -20,6 +20,31 @@
...
+@interface MTRMetricData : NSObject
+
+/**
+ * Value for the metric data. The value may be nil depending on the event emitted.
+ */
+@property (nonatomic, nullable, readonly, copy) NSNumber * value;
+
+/**
+ * Error code for the metric data. This value when valid, holds the error code value
+ * of the operation associated with the event.
```
bzbarsky-apple, line 36: It would be good to document what value space these error codes come from.  Are these CHIP_ERROR values?  Something else?

They look like CHIP_ERROR values to me, which is hard to document in our public API....  I wonder whether we should convert these to NSError instead of just reporting the raw codes, but that conversion can be lossy for error types we don't know about.

If this stays as-is, it needs at least _some_ documentation that would allow consumers to interpret these values.

src/darwin/Framework/CHIP/MTRMetrics.h
```diff
@@ -20,6 +20,31 @@
...
+MTR_NEWLY_AVAILABLE
+@interface MTRMetricData : NSObject
+
+/**
+ * Value for the metric data. The value may be nil depending on the event emitted.
+ */
+@property (nonatomic, nullable, readonly, copy) NSNumber * value;
+
+/**
+ * Error code for the metric data. This value when valid, holds the error code value
```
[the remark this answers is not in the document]
anush-apple: Will fix in a follow up PR

src/darwin/Framework/CHIP/MTRMetrics.h
anush-apple: This is subject to the metric and could be a non-CHIP error as well. Hence I left it open ended.

src/darwin/Framework/CHIP/MTRMetrics.h
```diff
@@ -20,6 +20,31 @@
...
+ * of the operation associated with the event.
+ */
+@property (nonatomic, nullable, readonly, copy) NSNumber * errorCode;
+
+/**
+ * Duration of event associated with the metric. This value may be nil depending on
+ * the event emitted.
+ */
+@property (nonatomic, nullable, readonly, copy) NSNumber * durationMicroseconds;
```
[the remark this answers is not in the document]
anush-apple: I can look to change this.

src/darwin/Framework/CHIP/MTRMetrics.h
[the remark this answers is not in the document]
anush-apple: Filed: https://github.com/project-chip/connectedhomeip/issues/32438

src/darwin/Framework/CHIP/MTRMetricsCollector.mm
anush-apple: Filed https://github.com/project-chip/connectedhomeip/issues/32439

src/darwin/Framework/CHIP/MTRMetrics.h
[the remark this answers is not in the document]
anush-apple: Filed https://github.com/project-chip/connectedhomeip/issues/32443

PR #32428, opened 2024-03-05
ESP: Enable Wi-Fi and Ethernet coexisting and add platform mdns for Ethernet

wqx6:
### Problems

1. Matter over Wi-Fi is disabled when Ethernet is enabled.
2. Ethernet device can only use minimal mDNS.

### Changes
1. Enable platform mDNS for Ethernet devices 
2. Make Wi-Fi and Ethernet be able to coexist

### Test
Tested on ESP32-Ethernet-Kit_A_V1.0, both commissioning over BLE-WiFi and commissioning over Ethernet works


examples/all-clusters-app/esp32/main/main.cpp
```diff
@@ -114,7 +114,7 @@ static void InitServer(intptr_t context)
 {
     Esp32AppServer::Init(&sCallbacks); // Init ZCL Data Model and CHIP App Server AND Initialize device attestation config
 
-#if !(CHIP_DEVICE_CONFIG_ENABLE_WIFI && CHIP_DEVICE_CONFIG_ENABLE_THREAD)
+#if !(CHIP_DEVICE_CONFIG_ENABLE_WIFI && (CHIP_DEVICE_CONFIG_ENABLE_THREAD || CHIP_DEVICE_CONFIG_ENABLE_ETHERNET))
```
chshu, line 117: Currently we assume Thread and Ethernet won't coexist. Please list the supported multi-network scenarios in the example readme, and introduce how to enable them.

chshu (approved):

src/platform/ESP32/ESP32DnssdImpl.cpp
```diff
@@ -418,8 +433,23 @@ static CHIP_ERROR ParseSrvResult(ResolveContext * ctx)
...
-            // If the InterfaceId in the context is Null, we will use the Station netif by default.
-            ctx->mService->mInterface = Inet::InterfaceId(DeviceLayer::Internal::ESP32Utils::GetStationNetif());
+            if (ctx->mSrvQueryResult->esp_netif)
+            {
+                ctx->mService->mInterface =
+                    Inet::InterfaceId(static_cast<struct netif *>(esp_netif_get_netif_impl(ctx->mSrvQueryResult->esp_netif)));
+            }
+            else
+            {
+                // If the InterfaceId in the context and esp_netif in current result is Null,
```
bzbarsky-apple, line 443: Seems like this code should be factored out into a helper function, since we have it in two places that now have to stay in sync and the code is a bit complicated.

bzbarsky-apple (approved):

yunhanw-google (approved):

src/platform/ESP32/ESP32DnssdImpl.cpp
wqx6: Done

Issue #32440, opened 2024-03-05
[Tooling] Re-enable MRP test, move to use mock clock


cecille:
### Reproduction steps / Feature

This test was flaking regularly on darwin, and it appears to be related to timing because the test is using the real clock. Please see discussion here: https://csamembers.slack.com/archives/G014G30SVV0/p1709651149059849

### Platform

core (please add to version below)

### Type

CI tested

PR #32441, opened 2024-03-05
Disable flaky MRP test

cecille:
See https://github.com/project-chip/connectedhomeip/issues/32440


chrisdecenzo (approved):

bzbarsky-apple (approved):
Approving, but this used to be non-flaky, so chances are someone broke something that causes it to now be flaky....

PR #32444, opened 2024-03-05
Zap updates to resolve zap warnings for lock, thermostat  and light-switch examples.

Sarthak-Shaha:
Problem:
Latest Zap shows error warning while building for the clusters / commands / attributes which are not up to the latest specifications. The zap files for lock app, thermostat and light-switch are not up to the spec.

Fix / solution:
Modifies the zap files for lock app, thermostat and light-switch to comply with latest spec
modifies the xml files to reflect the changes in mandatory commands
generated the .matter files for zap changes
Made following changes
For all zap files
a. enabled cluster: Diagnostic Logs server : RetrieveLogsResponse command

Light- switch :
a. Identify client : enabled identify outgoing commands
Thermostat
a. cluster: Identify, mandatory attribute: Identify Type.
Silabs Lighting app thread/wifi:
TODO : Need to enable scenes cluster in Matter_1.3

Closed previous PR as it was cluttered .


examples/lock-app/lock-common/lock-app.zap
```diff
@@ -2126,6 +2126,14 @@
               "source": "client",
               "isIncoming": 1,
               "isEnabled": 1
+            },
+            {
+              "name": "RetrieveLogsResponse",
```
andy31415, line 2131: Is this the correct thing to do? Do we have to enable every `response`?
It looks odd in .matter file to say `server handles the command FooResponse` overall however I see some existing examples doing this.

I do not quite understand this bit ... I expect servers to need to handle Requests but never Responses (they send these out only). what is the expected behaviour?

bzbarsky-apple (approved):

examples/lock-app/lock-common/lock-app.zap
Sarthak-Shaha: Diagnostic logs is optional cluster but if you enable the cluster `RetrieveLogsResponse` is a mandatory command according the spec.

jmartinez-silabs (approved):

yufengwangca (approved):

PR #32446, opened 2024-03-05
Make it possible to dynamically set MRP parameters.

bzbarsky-apple:
Right now MRP parameters have to be compile-time constants, except for ICDs. But in some cases the same binary might be used on different types of hardware that have different MRP requirements.  This change makes it possible to set the MRP parameters at run time, based on the actual hardware involved, not just compile time.  The new functionality is gated by a new default-off compiler flag, so only systems that really need it pay the cost for it.

### DIFF
--- a/src/messaging/ReliableMessageProtocolConfig.h
+++ b/src/messaging/ReliableMessageProtocolConfig.h
@@ -208,6 +208,25 @@ struct ReliableMessageProtocolConfig
         return mIdleRetransTimeout == that.mIdleRetransTimeout && mActiveRetransTimeout == that.mActiveRetransTimeout &&
             mActiveThresholdTime == that.mActiveThresholdTime;
     }
+
+#if CHIP_DEVICE_CONFIG_ENABLE_DYNAMIC_MRP_CONFIG
+    /**
+     * Set the local MRP configuration for the node.
+     *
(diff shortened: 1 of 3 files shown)


yunhanw-google (approved):

woody-apple (approved):

src/messaging/ReliableMessageProtocolConfig.h
```diff
@@ -208,6 +208,25 @@ struct ReliableMessageProtocolConfig
...
+     * CHIP_CONFIG_MRP_LOCAL_ACTIVE_RETRY_INTERVAL).
+     *
+     * Otherwise the value set via this function is used instead of the
+     * compiled-in settings, but can still be overridden by ICD configuration
+     * and other things that would override the compiled-in settings.
+     *
+     * @return whether the local MRP configuration actually changed as a result
+     *         of this call.  If it did, callers may need to reset DNS-SD
+     *         advertising to advertise the updated values.
+     */
```
pidarped, line 227: What happens to active Exchanges in flight? How would they be impacted? Need to add documentation on that to clarify.

bzbarsky-apple: Nothing happens with them, generally speaking.  This only affects new sessions that are established after this point.

pidarped (approved):

PR #32448, opened 2024-03-05
Stop explicitly building files that are part of libCHIP in Xcode.

bzbarsky-apple:
ember-strings and attribute-metadata used to be part of the data_model stuff that is part of the "app" build, but they've moved into libCHIP now, so we don't have to build them directly anymore.


andy31415 (approved):

andy31415:
Admin-merged because restyler was confused beacause only xcodeproj files were changed.

PR #32458, opened 2024-03-06
[Silabs][SiWx917] Getting the button pressed and button release in sync in sleep wakeup for btn0

chirag-silabs:
**Issue**
Once the device wakes up from sleep it handles a button release event due to the hardware(pull up resistor) 

**Fix**
Checking if the btn was not pressed just released no need of going forward. Making sure pressed and release are in sync

### DIFF
--- a/src/platform/silabs/platformAbstraction/WiseMcuSpam.cpp
+++ b/src/platform/silabs/platformAbstraction/WiseMcuSpam.cpp
@@ -30,10 +30,9 @@ extern "C" {
 #include "rsi_board.h"
 #include "sl_event_handler.h"
 #include "sl_si91x_button.h"
+#include "sl_si91x_button_pin_config.h"
 #include "sl_si91x_led.h"
 #include "sl_si91x_led_config.h"
-#include "sl_si91x_button.h"
-#include "sl_si91x_button_pin_config.h"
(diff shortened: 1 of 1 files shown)


src/platform/silabs/platformAbstraction/WiseMcuSpam.cpp
```diff
@@ -104,6 +108,13 @@ void SilabsPlatform::StartScheduler()
 extern "C" {
 void sl_button_on_change(uint8_t btn, uint8_t btnAction)
 {
+#if SL_ICD_ENABLED
+    // If device is sleeping making sure we get only one release per button press
+    if (btnAction == BUTTON_RELEASED && btn0_pressed == false)
```
jepenven-silabs, line 113: This seems to be completely wrong. You have an API that returns the button number (0 or 1 ) and the corresponding action. there shouldn't be any need for an external global variable...

jepenven-silabs (changes requested):
remove need for global variable

mkardous-silabs (changes requested):
I don't believe this is the right way to fix this.

mkardous-silabs (approved):

jepenven-silabs (approved):

PR #29682, opened 2023-10-10
Replaced nlunit-test with pw_unit_test in src/access/tests/

mbknust:
Used to part of https://github.com/project-chip/connectedhomeip/pull/29342, but was separated into smaller pieces.

PR #32460, opened 2024-03-06
Add pigweed support for nrfconnect

jlatusek:
Related to #29682

## Changes
Add a call of the wrapper for running all Pigweed tests in test_driver for nrfconnect and link the necessary library.

## Testing

CI will test it. In related PR, there are changes for all platforms. The CI there passes all checks.

### DIFF
--- a/src/test_driver/nrfconnect/main/runner.cpp
+++ b/src/test_driver/nrfconnect/main/runner.cpp
@@ -16,6 +16,7 @@
  */
 
 #include <lib/support/CodeUtils.h>
+#include <lib/support/UnitTest.h>
 #include <lib/support/UnitTestRegistration.h>
 #include <platform/CHIPDeviceLayer.h>
 
@@ -35,6 +36,7 @@ extern "C" int main(void)
 
     LOG_INF("Starting CHIP tests!");
     int status = RunRegisteredUnitTests();
+    status += chip::test::RunAllTests();
     LOG_INF("CHIP test status: %d", status);
 
     _exit(status);
--- a/config/nrfconnect/chip-gn/.gn
+++ b/config/nrfconnect/chip-gn/.gn
@@ -14,6 +14,7 @@
 
 import("//build_overrides/build.gni")
 import("//build_overrides/chip.gni")
+import("//build_overrides/pigweed.gni")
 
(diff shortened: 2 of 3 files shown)


src/test_driver/nrfconnect/main/runner.cpp
```diff
@@ -29,13 +30,24 @@ using namespace ::chip::DeviceLayer;
 
 LOG_MODULE_REGISTER(runner, CONFIG_MATTER_LOG_LEVEL);
 
-extern "C" int main(void)
+extern "C" {
+void console_init(void) {}
```
Damian-Nordic, line 34: This is a hack that shadows Zephyr function definitions. Could you at least add a TODO comment that the tests should rather configure and use Zephyr's console subsystem?

src/test_driver/nrfconnect/main/runner.cpp
```diff
@@ -29,13 +30,24 @@ using namespace ::chip::DeviceLayer;
...
+    return 0;
+}
+
+int main(void)
 {
     VerifyOrDie(settings_subsys_init() == 0);
 
     LOG_INF("Starting CHIP tests!");
     int status = RunRegisteredUnitTests();
+    status += chip::test::RunAllTests();
```
Damian-Nordic, line 48: Is the plan to completely replace the old method of registering UTs or we're going to support both in parallel?

jlatusek: The plan is to replace the old method completely successively test after test. The old method will be removed at the end of the process.

Damian-Nordic: Thanks for the info!

src/test_driver/nrfconnect/main/runner.cpp
jlatusek: I've checked, and this hack was necessary in the first version of my changes. This is no longer required, so I removed it.

Damian-Nordic: ok, thanks! :)

Damian-Nordic (approved):

arkq (approved):

PR #7538, opened 2021-06-10
IM command needs mandatory path

yunhanw-google:
#### Problem
IM command needs mandatory path

#### Change overview
-- Remove the test for IM command without path
-- Update CommandPathParams from pass by pointer to pass by reference for PrepareCommand and AddStatusCode so that path becomes mandatory in API

#### Testing
The existing test cover this.

PR #32467, opened 2024-03-06
Refactor CommandHandler to be more unittestable

tehampson:
### Ready for review

Summary of what has been refactored:
* Moved some method declaration in `CommandHandler.h` that were previously public to private.
* Inverting it such that `CommandResponder` contains `CommandHandler`
* `CommandResponder` is entirely responsible for responding for handling exchange. Remove `CommandHandler` from needing the exchange context. Any previous public APIs that relied on data from exchange now is tunneled through `CommandResponderInterface`
* `CommandHandler` gets object that implements`CommandResponderInterface`. This allows `CommandHandler` to be more unit testable.
* Allow `CommandHandler` to intentionally not provide instance of `CommandResponderInterface`. This allows application code to call an embr clsuter command dispatcher, such as `emberAfLevelControlClusterMoveWithOnOffCallback`, with a `CommandHandler` that won't send any responses.
* Removed declaration of `TestCommandHandlerWithSendEmptyResponse` from `src/app/tests/TestCommandInteraction.cpp`. Test itself was removed in #7538, this declaration stayed around for a long time.

Tests:
* CI Passes


src/app/tests/TestCommandInteraction.cpp
```diff
@@ -326,7 +343,7 @@ class TestCommandInteraction
     static void TestCommandHandlerEncodeSimpleStatusCode(nlTestSuite * apSuite, void * apContext);
     static void TestCommandHandlerWithSendEmptyResponse(nlTestSuite * apSuite, void * apContext);
 
-    static void TestCommandHandlerWithProcessReceivedEmptyDataMsg(nlTestSuite * apSuite, void * apContext);
+    static void TestCommandHandlerWithOnInvokeReceivedEmptyDataMsg(nlTestSuite * apSuite, void * apContext);
```
tehampson, line 346: TODO a few tests where `CommandHandler` does not have something that implements `CommandResponderInterface`

* TestCommandHandlerWithoutResponderCallingAddStatus
* TestCommandHandlerWithoutResponderCallingAddStatus
* TestCommandHandlerWithoutResponderCallingRawApis

src/app/CommandHandler.h
```diff
@@ -189,32 +192,37 @@ class CommandHandler
...
      *
      * isTimedInvoke is true if and only if this is part of a Timed Invoke
      * transaction (i.e. was preceded by a Timed Request).  If we reach here,
      * the timer verification has already been done.
+     *
+     * commandResponder handles sending InvokeResponses, added by clusters, to the client.
      */
-    void OnInvokeCommandRequest(Messaging::ExchangeContext * ec, const PayloadHeader & payloadHeader,
-                                System::PacketBufferHandle && payload, bool isTimedInvoke);
+    Protocols::InteractionModel::Status OnInvokeCommandRequest(CommandResponderInterface & commandResponder,
```
andy31415, line 217: This preserves a pointer to commandResponder insiude via SetCommandResponder. Could we have some details about the expected lifetime of commandResponder? how long does it have to live and when is it safe to drop?

src/app/CommandHandler.cpp
```diff
@@ -92,40 +99,20 @@ CHIP_ERROR CommandHandler::AllocateBuffer()
...
 {
-    System::PacketBufferHandle response;
-    Status status = Status::Failure;
-    VerifyOrDieWithMsg(ec != nullptr, DataManagement, "Incoming exchange context should not be null");
     VerifyOrDieWithMsg(mState == State::Idle, DataManagement, "state should be Idle");
 
-    // NOTE: we already know this is an InvokeCommand Request message because we explicitly registered with the
-    // Exchange Manager for unsolicited InvokeCommand Requests.
-    mResponseSender.SetExchangeContext(ec);
+    SetCommandResponder(&commandResponder);
```
andy31415, line 107: I was trying to look into the lifetiume of the commandResponder ... we set it oninvoke or test invoke, however it seems never reset. It seems potentially desirable to reset it back to invalid once the invoke processing is completed, so it is clear when it would be safe to be released by whoever uses the command handler.

src/app/CommandHandler.h
```diff
@@ -177,8 +177,11 @@ class CommandHandler
         bool mStartOrEndDataStruct = true;
     };
 
-    class TestOnlyMarker
+    class TestOnlyOverrides
```
andy31415, line 180: nit: maybe better represented as `struct` if just a container of public member

src/app/CommandHandler.h
```diff
@@ -473,6 +446,54 @@ class CommandHandler
...
+            return err;
+        }
+        ReturnErrorOnFailure(FinalizeInvokeResponseMessageAndPrepareNext());
+        err = addResponseFunction();
+        if (err != CHIP_NO_ERROR)
+        {
+            // The return value of RollbackResponse is ignored, as we prioritize
+            // conveying the error generated by addResponseFunction to the
+            // caller.
+            RollbackResponse();
```
andy31415, line 492: is rollbackreponse logging the failure? otherwise it seems valuable to log here for debug purposes, otherwise the error is forever lost.

src/app/CommandResponderInterface.h
```diff
@@ -0,0 +1,55 @@
...
+{
+public:
+    virtual ~CommandResponderInterface() = default;
+
+    /* ExchangeContext related getter methods */
+    virtual Messaging::ExchangeContext * GetExchangeContext() const = 0;
+    virtual Access::SubjectDescriptor GetSubjectDescriptor() const  = 0;
+    virtual FabricIndex GetAccessingFabricIndex() const             = 0;
+    virtual bool IsForGroup() const                                 = 0;
+    virtual GroupId GetGroupId() const                              = 0;
```
andy31415, line 44: does GroupId have any meaning if IsForGroup is false? If not, maybe this should be documented. Even wondering if we should combine these into a `Optional<GroupId>` value where missing means not group (although not sure if that is readable).

src/app/CommandResponderInterface.h
```diff
@@ -0,0 +1,55 @@
...
+
+    /* Exchange methods that CommandHandler's may need to perform while processing request */
+    virtual void FlushAcksRightNow() = 0;
+
+    /* CommandResponder core methods */
```
andy31415, line 49: I am unsure about the `core` description. Could we update the comments somewhat to make it clearer why these methods are special? Or even individually doc-comment them instead?

src/app/CommandResponderInterface.h
```diff
@@ -0,0 +1,55 @@
...
+    virtual ~CommandResponderInterface() = default;
+
+    /* ExchangeContext related getter methods */
+    virtual Messaging::ExchangeContext * GetExchangeContext() const = 0;
+    virtual Access::SubjectDescriptor GetSubjectDescriptor() const  = 0;
+    virtual FabricIndex GetAccessingFabricIndex() const             = 0;
+    virtual bool IsForGroup() const                                 = 0;
+    virtual GroupId GetGroupId() const                              = 0;
+
+    /* Exchange methods that CommandHandler's may need to perform while processing request */
```
andy31415, line 46: This seems to be a generic comment for a single method. How about just commenting on this method only?

andy31415: The comments also feel very coupled to commandhandler - instead of explaining what they are providing, they mention `CommandHandler` instead. Could we pull in expectations in these comments instead of referencing commandhandler?

src/app/CommandHandler.cpp
andy31415: I think setting to null on close is better - it would show that the lifetime was considered. It may also help in case we have bugs that access data past close.

andy31415 (approved):

.github/workflows/lint.yml
```diff
@@ -94,6 +94,7 @@ jobs:
                      --known-failure app/AttributeAccessToken.h \
                      --known-failure app/CommandHandler.h \
                      --known-failure app/CommandHandlerInterface.h \
+                     --known-failure app/CommandResponseSender.h \
```
bzbarsky-apple, line 97: I assume this is being added to known-failure because it's got the same issues as CommandHandler.h in terms of dependency loops?

src/app/CommandResponderInterface.h
```diff
@@ -0,0 +1,64 @@
...
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+#pragma once
+
+#include <lib/core/Optional.h>
+#include <messaging/ExchangeHolder.h>
```
bzbarsky-apple, line 21: Where is ExchangeHolder used here?  I don't think it is.  This should be including `ExchangeContext.h` and whatever file declares `SubjectDescriptor`, and `lib/core/DataModelTypes.h` for GroupId/FabricIndex.  Should be including `SystemPacketBuffer.h` since it uses `PacketBufferHandle`.

src/app/CommandResponderInterface.h
```diff
@@ -0,0 +1,64 @@
...
+namespace chip {
+namespace app {
+
+/**
+ * Interface for sending InvokeResponseMessage(s).
+ *
+ * Provides information about the associated exchange context.
+ *
+ * Design Rationale: This interface enhances unit testability and allows applications to
+ * customize CommandResponder behavior. For example, it can stub out sending a response.
```
bzbarsky-apple, line 33: This comment should probably explain why an "application" would want to do such a on-its-face-spec-violating thing.

src/app/CommandResponderInterface.h
```diff
@@ -0,0 +1,64 @@
...
+ */
+class CommandResponderInterface
+{
+public:
+    virtual ~CommandResponderInterface() = default;
+
+    /* ExchangeContext related getter methods */
```
bzbarsky-apple, line 40: This needs much better documentation.  The exchange involved is the one the InvokeRequestMessage was received on, presumably.  The subject descriptor is the one for the exchange presumably, right?  But do we in fact want to allow that to be configured separately from the exchange context (as opposed to having a method on this class that uses the overridable GetExchangeContext to implement GetSubjectDescriptor)?  Similar for GetAccessingFabricIndex().

GetGroupId() needs to document what its return value means.

src/app/CommandResponderInterface.h
```diff
@@ -0,0 +1,64 @@
...
+
+#include <lib/core/Optional.h>
+#include <messaging/ExchangeHolder.h>
+#include <protocols/interaction_model/StatusCode.h>
+
+namespace chip {
+namespace app {
+
+/**
+ * Interface for sending InvokeResponseMessage(s).
```
bzbarsky-apple, line 28: Seems like a combined interface for that and finding various information about the exchange the invoke is happening on...  Should these be separate interfaces?  Or should this have a name that makes it clearer what it really does (InvokeInteractionExchangeInterface or CommandHandlerExchangeInterface or something?).

src/app/CommandResponderInterface.h
```diff
@@ -0,0 +1,64 @@
...
+    virtual Access::SubjectDescriptor GetSubjectDescriptor() const  = 0;
+    virtual FabricIndex GetAccessingFabricIndex() const             = 0;
+    virtual Optional<GroupId> GetGroupId() const                    = 0;
+
+    /**
+     * @brief Flush acks right now.
+     *
+     * Typically called when processing a slow command.
+     */
+    virtual void FlushAcksRightNow() = 0;
```
bzbarsky-apple, line 51: I think we should name this based on the actual semantics involved.  As in, LongRunningCommandReceived() or some such.  The fact that this flushes acks is an implementation details (and not even true over TCP or BLE).

src/app/CommandResponderInterface.h
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * @brief Adds completed InvokeResponseMessage for sending to command requestor.
+     */
+    virtual void AddInvokeResponseToSend(System::PacketBufferHandle && aPacket) = 0;
+    /**
+     * @brief Called to indicate that response was dropped.
```
bzbarsky-apple, line 58: Dropped by who?  Why?  As an API consumer I have no idea when or whether to call this.

src/app/CommandResponseSender.cpp
```diff
@@ -38,12 +38,10 @@ CHIP_ERROR CommandResponseSender::OnMessageReceived(Messaging::ExchangeContext *
...
 
-        // If SendCommandResponse() fails, we are responsible for closing the exchange,
-        // as stipulated by the API contract. We fulfill this obligation by not sending
-        // a message expecting a response on the exchange. Since we are in the middle
-        // of processing an incoming message, the exchange will close itself once we are
-        // done processing it, if there is no response to wait for at that point.
         err = SendCommandResponse();
+        // If SendCommandResponse() fails, we must still close the exchange. Since sending an
+        // InvokeResponseMessage seems problematic, we'll send a StatusResponse with 'Failure'
+        // instead.
```
bzbarsky-apple, line 44: This lost the key part of the comment that was deleted, which is that if we don't send a message needing a response, that closes the exchange.

We could also send absolutely nothing and that would _also_ close the exchange.

So the comment no longer really explains why the code is doing what it's doing.  We need to fix this part.

1.  How are we ensuring the exchange is closed?  Just but not sending an "expect a reply" response.  This works because we are inside OnMessageReceived.
2. Why are we sending anything at all?

src/app/CommandResponseSender.cpp
```diff
@@ -78,16 +76,33 @@ void CommandResponseSender::OnResponseTimeout(Messaging::ExchangeContext * apExc
 {
     ChipLogDetail(DataManagement, "CommandResponseSender: Timed out waiting for response from requester mState=[%10.10s]",
                   GetStateStr());
+    // We don't need to consider remaining async CommandHandlers here. We become the exchange
+    // delegate only after CommandHandler::OnDone is called.
```
bzbarsky-apple, line 80: I don't see a CommandHandler::OnDone anywhere.  And there shouldn't be CommandHandlers (plural) involved anyway....

Trying to make sense of this comment and failing.

src/app/CommandResponseSender.cpp
```diff
@@ -78,16 +76,33 @@ void CommandResponseSender::OnResponseTimeout(Messaging::ExchangeContext * apExc
...
+        // tests to start failing.
+        //   ```
+        //   if (mExchangeCtx && mExchangeCtx->HasSessionHandle())
+        //   {
+        //       SendStatusResponse(Status::Failure);
+        //   }
+        //   Close();
+        //   return;
+        //   ```
+    }
```
bzbarsky-apple, line 105: but shouldn't we do the `Close(); return;` bit anyway?  It doesn't seem to make sense to run the HasMoreToSend() logic below if we failed to send... unless HasMoreToSend() will always be false in that case?

src/app/CommandResponseSender.cpp
```diff
@@ -157,12 +171,52 @@ void CommandResponseSender::MoveToState(const State aTargetState)
...
+{
+    VerifyOrDieWithMsg(ec != nullptr, DataManagement, "Incoming exchange context should not be null");
+    VerifyOrDieWithMsg(mState == State::ReadyForInvokeResponses, DataManagement, "state should be ReadyForInvokeResponses");
+
+    // NOTE: we already know this is an InvokeCommand Request message because we explicitly registered with the
+    // Exchange Manager for unsolicited InvokeCommand Requests.
+    mExchangeCtx.Grab(ec);
+    mExchangeCtx->WillSendMessage();
+
+    // Grabbing Handle to prevent OnDone from being called before OnInvokeCommandRequest returns.
```
bzbarsky-apple, line 188: I am not seeing a CommandResponseSender::OnDone.  Am I just missing it?

src/app/CommandResponseSender.cpp
```diff
@@ -157,12 +171,52 @@ void CommandResponseSender::MoveToState(const State aTargetState)
...
+    Status status = mCommandHandler.OnInvokeCommandRequest(*this, std::move(payload), isTimedInvoke);
+    if (status != Status::Success)
     {
-        mResponseSenderDoneCallback->mCall(mResponseSenderDoneCallback->mContext);
+        VerifyOrDie(mState == State::ReadyForInvokeResponses);
+        SendStatusResponse(status);
+        // The API contract of OnInvokeCommandRequest requires the CommandResponder instance to outlive
+        // the CommandHandler. Therefore, we cannot safely call Close() here, even though we have
+        // finished sending data. Closing must be deferred until the CommandHandler::OnDone callback.
+        mDelayCallingCloseUntilOnDone = true;
```
bzbarsky-apple, line 199: I don't quite get it. We have a state machine here, right?  Why are we using an out-of-band boolean instead of mState here to store our state?

src/app/CommandResponseSender.h
```diff
@@ -17,29 +17,71 @@
...
-typedef void (*OnResponseSenderDone)(void * context);
-class CommandHandler;
-
+// TODO(#30453): Rename CommandResponseSender to CommandResponder in follow up PR
 /**
- * Class manages the process of sending `InvokeResponseMessage`(s) back to the initial requester.
+ * Manages the process of sending InvokeResponseMessage(s) to the requester.
+ *
+ * Implements the CommandResponderInterface. Uses a CommandHandler member to process
+ * InvokeRequestCommand. The CommandHandler is provided a reference to this
```
bzbarsky-apple, line 34: What is "InvokeRequestCommand"?

src/app/CommandHandler.cpp
bzbarsky-apple, line 1: I have not reviewed the changes to this file or its header yet.  Trying to make sure I understand the new interface and CommandResponseSender bits first.

src/app/CommandResponseSender.h
```diff
@@ -51,91 +93,79 @@ class CommandResponseSender : public Messaging::ExchangeDelegate
...
-    GroupId GetGroupId()
+    Optional<GroupId> GetGroupId() const override
     {
         VerifyOrDie(mExchangeCtx);
-        return mExchangeCtx->GetSessionHandle()->AsIncomingGroupSession()->GetGroupId();
+        if (!mExchangeCtx->IsGroupExchangeContext())
+        {
+            return NullOptional;
+        }
+        return MakeOptional(mExchangeCtx->GetSessionHandle()->AsIncomingGroupSession()->GetGroupId());
```
bzbarsky-apple, line 123: Worth a comment that explains why we are guaranteed this is an _incoming_ group session.

Alternately, we should test for the session type, not the context type, if that's what we care about.

src/app/InteractionModelEngine.cpp
```diff
@@ -374,9 +374,14 @@ bool InteractionModelEngine::SubjectHasPersistedSubscription(FabricIndex aFabric
...
 
+void InteractionModelEngine::OnDone(CommandResponseSender & apResponderObj)
+{
+    mCommandResponderObjs.ReleaseObject(&apResponderObj);
+}
+
 void InteractionModelEngine::OnDone(CommandHandler & apCommandObj)
 {
-    mCommandHandlerObjs.ReleaseObject(&apCommandObj);
+    VerifyOrDie(false);
```
bzbarsky-apple, line 384: Needs documentation explaining why.

But also, feels like we should just fix our APIs so that this method doesn't need to exist here.  Followup for that, probably.

.github/workflows/lint.yml
tehampson: Yes. Same issue since app depending on im and im (InteractionModelEngine) including this header.

src/app/CommandResponseSender.cpp
tehampson: Removed the comment.

I got worried about a potential corner case that is non-existent.

src/app/CommandResponseSender.h
tehampson: I mean `mExchangeCtx->IsGroupExchangeContext()` is just doing `return mSession && mSession->IsGroupSession()`. For readability sake I will just check `sessionHandle->IsGroupSession()` here.

src/app/CommandResponseSender.cpp
tehampson: It is in the header. I moved out the implementation into the .cpp file now so it should be easier to find

src/app/CommandResponderInterface.h
tehampson: Right now, to give unit test some flexibility it is helpful if we provide this instead of having CommandHandler get it from exchange context.

I will add todo comment here to address this separately.

src/app/CommandHandler.h
```diff
@@ -189,32 +192,39 @@ class CommandHandler
...
+     * To prevent immediate OnDone invocation, callers can wrap their CommandHandler instance
+     * within a CommandHandler::Handle.
      *
      * isTimedInvoke is true if and only if this is part of a Timed Invoke
      * transaction (i.e. was preceded by a Timed Request).  If we reach here,
      * the timer verification has already been done.
+     *
+     * commandResponder handles sending InvokeResponses, added by clusters, to the client. The
+     * command responder object must outlive this CommandHandler object. It is only safe to
+     * release after client recieved OnDone callback.
```
bzbarsky-apple, line 217:
```suggestion
     * release after the caller of OnInvokeCommandRequest receives the OnDone callback.
```

src/app/CommandHandler.h
```diff
@@ -638,11 +654,15 @@ class CommandHandler
         return FinishCommand(/* aEndDataStruct = */ false);
     }
 
+    void SetCommandResponder(CommandHandlerExchangeInterface * commandResponder);
```
bzbarsky-apple, line 657: Maybe SetExchangeInterface?

bzbarsky-apple (approved):

src/app/CommandResponseSender.h
```diff
@@ -51,115 +77,116 @@ class CommandResponseSender : public Messaging::ExchangeDelegate
...
+    Optional<GroupId> GetGroupId() const override
     {
         VerifyOrDie(mExchangeCtx);
-        return mExchangeCtx->IsGroupExchangeContext();
+        auto sessionHandle = mExchangeCtx->GetSessionHandle();
+        if (!sessionHandle->IsGroupSession())
+        {
+            return NullOptional;
+        }
+        return MakeOptional(sessionHandle->AsIncomingGroupSession()->GetGroupId());
```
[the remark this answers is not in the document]
tehampson: I just checked if it was an incoming session.

src/app/CommandHandlerExchangeInterface.h
```diff
@@ -0,0 +1,105 @@
...
+public:
+    virtual ~CommandHandlerExchangeInterface() = default;
+
+    /**
+     * Gets the inner exchange context object associated with incoming
+     * InvokeRequestMessage, without ownership.
+     *
+     * @return The inner exchange context, might be nullptr if no
+     *         exchange context has been assigned or the context
+     *         has been released.
```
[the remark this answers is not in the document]
tehampson: I have added a little more context about what cases it might not be "assigned"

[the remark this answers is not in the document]
tehampson: Marking as resolved so I can merge and start working on follow-ups. If there is an issue with the updated comment I can address in a follow up

PR #32471, opened 2024-03-06
[ICD] Store ICD Idle/Active parameters

yunhanw-google:
--Store ActiveModeDuration, ActiveModeThreshold, IdleModeDuration in ICDClientStorage, and let user decide how to use them, for example, when user tries to interact with lit icd and have the queue, and UI can prompt user with the estimate waiting time window, if that waiting time is very long, user can decide whether to manually wakeup the device immediately.

### DIFF
--- a/src/app/icd/client/DefaultICDClientStorage.h
+++ b/src/app/icd/client/DefaultICDClientStorage.h
@@ -122,13 +122,16 @@ public:
 protected:
     enum class ClientInfoTag : uint8_t
     {
-        kPeerNodeId       = 1,
-        kFabricIndex      = 2,
-        kStartICDCounter  = 3,
-        kOffset           = 4,
-        kMonitoredSubject = 5,
-        kAesKeyHandle     = 6,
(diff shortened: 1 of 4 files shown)


src/app/icd/client/DefaultICDClientStorage.h
```diff
@@ -156,7 +159,8 @@ class DefaultICDClientStorage : public ICDClientStorage
     {
         // All the fields added together
         return TLV::EstimateStructOverhead(sizeof(NodeId), sizeof(FabricIndex), sizeof(uint32_t), sizeof(uint32_t),
```
mkardous-silabs, line 161: Might be worth while ading /**/ comments for what each field represents.

mkardous-silabs (approved):

src/app/icd/client/DefaultICDClientStorage.h
yunhanw-google: updated, thanks

yufengwangca (approved):

PR #32473, opened 2024-03-07
Get the commissioner reusable more quickly after StopPairing()

ksperling-apple:
Don't need to keep state and the CommissioneeDeviceProxy around; instead hold on to just the Session and evict it when the disarm request succeeds or fails.

### DIFF
--- a/src/controller/CHIPDeviceController.cpp
+++ b/src/controller/CHIPDeviceController.cpp
@@ -1676,7 +1676,7 @@ static void MarkForEviction(const Optional<SessionHandle> & session)
 
 void DeviceCommissioner::CleanupCommissioning(DeviceProxy * proxy, NodeId nodeId, const CompletionStatus & completionStatus)
 {
-    // proxy == mDeviceBeingCommissioned, nodeId == mDeviceBeingCommissioned->GetDeviceId()
+    // At this point, proxy == mDeviceBeingCommissioned, nodeId == mDeviceBeingCommissioned->GetDeviceId()
 
     mCommissioningCompletionStatus = completionStatus;
 
@@ -1696,7 +1696,7 @@ void DeviceCommissioner::CleanupCommissioning(DeviceProxy * proxy, NodeId nodeId
         // If we're cleaning up because cancellation has been requested via StopPairing(), expire the failsafe
         // in the background and reset our state synchronously, so a new commissioning attempt can be started.
         CommissioneeDeviceProxy * commissionee = FindCommissioneeDevice(nodeId);
-        SessionHolder session((commissionee == proxy) ? commissionee->ReleaseSecureSession().Value()
+        SessionHolder session((commissionee == proxy) ? commissionee->DetachSecureSession().Value()
                                                       : proxy->GetSecureSession().Value());
 
         auto request     = DisarmFailsafeRequest();
@@ -1718,7 +1718,7 @@ void DeviceCommissioner::CleanupCommissioning(DeviceProxy * proxy, NodeId nodeId
             ChipLogError(Controller, "Failed to send command to disarm fail-safe: %" CHIP_ERROR_FORMAT, err.Format());
         }
 
-        CleanupAfterErrorDone();
+        CleanupDoneAfterError();
     }
     else if (completionStatus.failedStage.HasValue() && completionStatus.failedStage.Value() >= kWiFiNetworkSetup)
     {
@@ -1740,7 +1740,7 @@ void DeviceCommissioner::CleanupCommissioning(DeviceProxy * proxy, NodeId nodeId
         {
             // We won't get any async callbacks here, so just pretend like the command errored out async.
             ChipLogError(Controller, "Failed to send command to disarm fail-safe: %" CHIP_ERROR_FORMAT, err.Format());
-            CleanupAfterErrorDone();
+            CleanupDoneAfterError();
         }
     }
 }
@@ -1750,17 +1750,17 @@ void DeviceCommissioner::OnDisarmFailsafe(void * context,
 {
     ChipLogProgress(Controller, "Failsafe disarmed");
     DeviceCommissioner * commissioner = static_cast<DeviceCommissioner *>(context);
-    commissioner->CleanupAfterErrorDone();
+    commissioner->CleanupDoneAfterError();
 }
 
 void DeviceCommissioner::OnDisarmFailsafeFailure(void * context, CHIP_ERROR error)
 {
     ChipLogProgress(Controller, "Ignoring failure to disarm failsafe: %" CHIP_ERROR_FORMAT, error.Format());
     DeviceCommissioner * commissioner = static_cast<DeviceCommissioner *>(context);
-    commissioner->CleanupAfterErrorDone();
+    commissioner->CleanupDoneAfterError();
 }
 
-void DeviceCommissioner::CleanupAfterErrorDone()
+void DeviceCommissioner::CleanupDoneAfterError()
 {
     // If someone nulled out our mDeviceBeingCommissioned, there's nothing else
     // to do here.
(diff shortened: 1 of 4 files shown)


yunhanw-google (approved):

src/controller/CHIPDeviceController.cpp
```diff
@@ -1674,13 +1691,40 @@ void DeviceCommissioner::CleanupCommissioning(DeviceProxy * proxy, NodeId nodeId
...
+        // If we're cleaning up because cancellation has been requested via StopPairing(), expire the failsafe
+        // in the background and reset our state synchronously, so a new commissioning attempt can be started.
+        CommissioneeDeviceProxy * commissionee = FindCommissioneeDevice(nodeId);
+        SessionHolder session((commissionee == proxy) ? commissionee->ReleaseSecureSession().Value()
+                                                      : proxy->GetSecureSession().Value());
+
+        auto request     = DisarmFailsafeRequest();
+        auto onSuccessCb = [session](const app::ConcreteCommandPath & aPath, const app::StatusIB & aStatus,
+                                     const decltype(request)::ResponseType & responseData) {
+            ChipLogProgress(Controller, "Failsafe disarmed");
```
cecille, line 1705: you do have a BIT of a race here because your new commissioner won't be able to start commissioning until the failsafe is disarmed on the device. Is this fix to basically allow the new proxy to start discovery earlier? If you have the address cached, you could end up with some failures if the resolution happens before the failsafe disarm goes though.

src/controller/CHIPDeviceController.cpp
```diff
@@ -1674,13 +1691,40 @@ void DeviceCommissioner::CleanupCommissioning(DeviceProxy * proxy, NodeId nodeId
...
+            MarkForEviction(session.Get());
+        };
+        auto onFailureCb = [session](CHIP_ERROR aError) {
+            ChipLogProgress(Controller, "Ignoring failure to disarm failsafe: %" CHIP_ERROR_FORMAT, aError.Format());
```
cecille, line 1709: I suppose there's not much you can do here, but it might make sense to retry because if you evict the session then nothing can disarm the failsafe other than waiting it out. Although I suppose if it failed once, it's likely to fail again. There might not be an elegant solution here, but possibly worth thinking through. Not a blocker though.

src/controller/CHIPDeviceController.cpp
```diff
@@ -1674,13 +1691,40 @@ void DeviceCommissioner::CleanupCommissioning(DeviceProxy * proxy, NodeId nodeId
...
+            MarkForEviction(session.Get());
+        };
+
+        ChipLogProgress(Controller, "Disarming failsafe on device %p in background", proxy);
```
bzbarsky-apple, line 1713: Is this pointer logged anywhere else?  If not, it doesn't seem useful to log it here.  Logging the nodeId might be more useful?

src/controller/CHIPDeviceController.cpp
```diff
@@ -1713,17 +1750,17 @@ void DeviceCommissioner::OnDisarmFailsafe(void * context,
...
 void DeviceCommissioner::OnDisarmFailsafeFailure(void * context, CHIP_ERROR error)
 {
     ChipLogProgress(Controller, "Ignoring failure to disarm failsafe: %" CHIP_ERROR_FORMAT, error.Format());
     DeviceCommissioner * commissioner = static_cast<DeviceCommissioner *>(context);
-    commissioner->DisarmDone();
+    commissioner->CleanupAfterErrorDone();
 }
 
-void DeviceCommissioner::DisarmDone()
+void DeviceCommissioner::CleanupAfterErrorDone()
```
bzbarsky-apple, line 1763: Maybe call this CleanupCompletedAfterError?  If the intent is to say that this is called when we hit an error, do some cleanup, then call this function.

src/controller/CHIPDeviceController.cpp
```diff
@@ -1735,13 +1772,14 @@ void DeviceCommissioner::DisarmDone()
...
     CommissioningStageComplete(CHIP_NO_ERROR);
-    SendCommissioningCompleteCallbacks(nodeId, mCommissioningCompletionStatus);
 
     // If we've disarmed the failsafe, it's because we're starting again, so kill the pase connection.
     if (commissionee != nullptr)
     {
         ReleaseCommissioneeDevice(commissionee);
     }
+
+    SendCommissioningCompleteCallbacks(nodeId, mCommissioningCompletionStatus);
```
bzbarsky-apple, line 1782: Is the ordering change here important?  If so, worth documenting why SendCommissioningCompleteCallbacks needs to come after ReleaseCommissioneeDevice.

src/controller/CommissioneeDeviceProxy.h
```diff
@@ -105,6 +105,11 @@ class CommissioneeDeviceProxy : public DeviceProxy, public SessionDelegate
      */
     void CloseSession();
 
+    /**
+     *  Releases the underlying session (if any) from this proxy, without expiring it.
+     */
+    chip::Optional<SessionHandle> ReleaseSecureSession();
```
bzbarsky-apple, line 111: I think this might be clearer as ForgetSecureSession, and document that it makes this proxy hand over its session to the caller and forget about it, without expiring it.

src/controller/CHIPDeviceController.cpp
ksperling-apple: Yeah there's about 6 existing places where we do this... I think the NodeID would probably be more useful. I can change them all to that in a follow-up.

src/controller/CommissioneeDeviceProxy.h
ksperling-apple: Hm, "release" is what this is called in unique_ptr. Not sure about "forget"... maybe "detach"?

src/controller/CHIPDeviceController.cpp
```diff
@@ -1658,13 +1658,30 @@ void DeviceCommissioner::OnBasicFailure(void * context, CHIP_ERROR error)
...
+    request.expiryLengthSeconds = 0; // Expire immediately.
+    request.breadcrumb          = 0;
+    return request;
+}
+
+static void MarkForEviction(const Optional<SessionHandle> & session)
+{
+    if (session.HasValue())
+    {
+        session.Value()->AsSecureSession()->MarkForEviction();
```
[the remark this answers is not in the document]
ksperling-apple: This pattern is used in a few places in this code and `AsSecureSession` has a `VerifyOrDie`, so I think this is fine?
I guess we could put a check into `PerformCommissioningStep` to check that the proxy has a secure session for all cases in one place?

src/controller/CHIPDeviceController.cpp
ksperling-apple: I'm not sure if additional retries beyond what MRP gives us are warranted or like to succeed here... I think I'll leave this as is for now.

src/controller/CHIPDeviceController.cpp
ksperling-apple: Added a comment

src/controller/CHIPDeviceController.cpp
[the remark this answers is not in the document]
bzbarsky-apple: My point is that it's easy to call this function with some other random session that has nothing to do with a DeviceProxy.  If the session being a SecureSession is a precondition, we should make that clearer somehow.  Ideally by having the argument type be a SecureSession, though that's a little annoying with the handle setup for sessions.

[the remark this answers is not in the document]
bzbarsky-apple: Anyway, followup is fine for sorting this out.

bzbarsky-apple (approved):

src/controller/CHIPDeviceController.cpp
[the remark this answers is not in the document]
ksperling-apple: https://github.com/project-chip/connectedhomeip/issues/32541

src/controller/CHIPDeviceController.cpp
ksperling-apple: https://github.com/project-chip/connectedhomeip/issues/32541

PR #32476, opened 2024-03-07
[Android] Support ICD Check in message

### LINKED ISSUE #31359: [ICD] Integrate ICD check-in in android chip-tool
Similar to what we integrate ICD check-in mechanism in chip-tool(https://github.com/project-chip/connectedhomeip/pull/30731/files), we also need integrate it with android chip-tool

joonhaengHeo:
Fix #31359

### DIFF
--- a/src/controller/java/AndroidDeviceControllerWrapper.h
+++ b/src/controller/java/AndroidDeviceControllerWrapper.h
@@ -24,6 +24,8 @@
 
 #include <jni.h>
 
+#include <app/icd/client/CheckInHandler.h>
+#include <app/icd/client/DefaultCheckInDelegate.h>
 #include <app/icd/client/DefaultICDClientStorage.h>
 #include <controller/CHIPDeviceController.h>
 #include <credentials/GroupDataProviderImpl.h>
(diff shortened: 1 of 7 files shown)


src/controller/java/AndroidDeviceControllerWrapper.h
```diff
@@ -24,6 +24,8 @@
 
 #include <jni.h>
 
+#include <app/icd/client/CheckInHandler.h>
+#include <app/icd/client/DefaultCheckInDelegate.h>
```
yunhanw-google, line 28: You should not use DefaultCheckInDelegate in android chip-tool, you should create a new CheckInDelegate, 
expose OnCheckInComplete to java/kotlin application.
Expose OnKeyRefreshNeeded to java/kotlin application so that symmetric token generation can happen in same place for both commissioning and keyRefreshStage, we also need make sure it is unique
Expose OnKeyRefreshDone to the java/kotlin application.

yunhanw-google (changes requested):
Please implement AndroidCheckInDelegate

yunhanw-google (commented):
what is enableServerInteractions for? Could you put more comment about this?

joonhaengHeo:
> what is enableServerInteractions for? Could you put more comment about this?

To receive a check in message, mDNS search of _matter._tcp must be possible. I thought a separate feature was needed to operate this function. (If the app requires only the commissioner, this function is not needed.)

yunhanw-google (approved):

src/controller/java/AndroidDeviceControllerWrapper.h
yunhanw-google: based upon offline chat, the android CheckInDelegate would be follow-up

mkardous-silabs (approved):

PR #32482, opened 2024-03-07
Add checks for mOTInst in GenericThreadStackManagerImpl_OpenThread

wqx6:
We didn't check mOTInst in the APIs in GenericThreadStackManagerImpl_OpenThread, which might cause some issues when they are called before Thread stack is initialized.

### DIFF
--- a/src/platform/OpenThread/GenericThreadStackManagerImpl_OpenThread.hpp
+++ b/src/platform/OpenThread/GenericThreadStackManagerImpl_OpenThread.hpp
@@ -149,6 +149,7 @@ void GenericThreadStackManagerImpl_OpenThread<ImplClass>::_ProcessThreadActivity
 template <class ImplClass>
 bool GenericThreadStackManagerImpl_OpenThread<ImplClass>::_HaveRouteToAddress(const Inet::IPAddress & destAddr)
 {
+    VerifyOrReturnValue(mOTInst, false);
     bool res = false;
 
     // Lock OpenThread
@@ -233,6 +234,7 @@ void GenericThreadStackManagerImpl_OpenThread<ImplClass>::_OnPlatformEvent(const
 template <class ImplClass>
 bool GenericThreadStackManagerImpl_OpenThread<ImplClass>::_IsThreadEnabled(void)
 {
+    VerifyOrReturnValue(mOTInst, false);
     otDeviceRole curRole;
 
     Impl()->LockThreadStack();
@@ -245,6 +247,7 @@ bool GenericThreadStackManagerImpl_OpenThread<ImplClass>::_IsThreadEnabled(void)
 template <class ImplClass>
 CHIP_ERROR GenericThreadStackManagerImpl_OpenThread<ImplClass>::_SetThreadEnabled(bool val)
 {
+    VerifyOrReturnError(mOTInst, CHIP_ERROR_INCORRECT_STATE);
     otError otErr = OT_ERROR_NONE;
 
     Impl()->LockThreadStack();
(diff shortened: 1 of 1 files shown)


Damian-Nordic:
Why is this necessary? `InitThreadStack()` should fail if the instance is not created, and an application should not use the thread stack if it failed to initialize.

wqx6:
> Why is this necessary? `InitThreadStack()` should fail if the instance is not created, and an application should not use the thread stack if it failed to initialize.

We are trying to build the example on our ESP32-C6 which can choose to work as a Wi-Fi or Thread device according the outside input(such as UART or flash storage). For this device, we cannot disable `CHIP_DEVICE_CONFIG_ENABLE_THREAD`. And when the device works as a Wi-Fi device, the Thread stack will not be initialized but some APIs in `GenericThreadStackManagerImpl_OpenThread` will be called somehow.

For instance, the `_GetPrimary802154MACAddress()` will be called from `Dnssd::AdvertiseOperational()`-> `GenericConfigurationManagerImpl::GetPrimaryMACAddress()`->`ThreadStackMgr::GetPrimary802154MACAddress()`.

That's why I add these checks.

src/platform/OpenThread/GenericThreadStackManagerImpl_OpenThread.hpp
```diff
@@ -1552,6 +1576,7 @@ CHIP_ERROR GenericThreadStackManagerImpl_OpenThread<ImplClass>::_AddSrpService(c
 template <class ImplClass>
 CHIP_ERROR GenericThreadStackManagerImpl_OpenThread<ImplClass>::_RemoveSrpService(const char * aInstanceName, const char * aName)
 {
+    VerifyOrReturnError(mOTInst, CHIP_ERROR_INCORRECT_STATE);
```
Damian-Nordic, line 1579: The SRP functions already check for `mSrpClient.mIsInitialized`, so no extra checks are needed, right?

Damian-Nordic (approved):

Damian-Nordic:
> > Why is this necessary? `InitThreadStack()` should fail if the instance is not created, and an application should not use the thread stack if it failed to initialize.
> 
> We are trying to build the example on our ESP32-C6 which can choose to work as a Wi-Fi or Thread device according the outside input(such as UART or flash storage). For this device, we cannot disable `CHIP_DEVICE_CONFIG_ENABLE_THREAD`. And when the device works as a Wi-Fi device, the Thread stack will not be initialized but some APIs in `GenericThreadStackManagerImpl_OpenThread` will be called somehow.
> 
> For instance, the `_GetPrimary802154MACAddress()` will be called from `Dnssd::AdvertiseOperational()`-> `GenericConfigurationManagerImpl::GetPrimaryMACAddress()`->`ThreadStackMgr::GetPrimary802154MACAddress()`.
> 
> That's why I add these checks.

Thanks for the info :). We also support dynamic switch between Thread and Wi-Fi on our platform but we still initialize the Thread stack even if we don't use it for Matter, so that's why we didn't hit on this issue.

yufengwangca (approved):

src/platform/OpenThread/GenericThreadStackManagerImpl_OpenThread.hpp
```diff
@@ -1107,20 +1124,23 @@ CHIP_ERROR GenericThreadStackManagerImpl_OpenThread<ImplClass>::DoInit(otInstanc
 template <class ImplClass>
 bool GenericThreadStackManagerImpl_OpenThread<ImplClass>::IsThreadAttachedNoLock(void)
 {
+    VerifyOrReturnValue(mOTInst, false);
```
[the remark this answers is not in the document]
wqx6: Removed.

src/platform/OpenThread/GenericThreadStackManagerImpl_OpenThread.hpp
wqx6: Keep the check in `ClearSrpHost()` and `ClearAllSrpHostAndServices()` which might be called before mIsInitialized is set to true and remove other.

PR #32483, opened 2024-03-07
Increase size to be more readable on 4k monitor

tehampson:
PR #32359 reduced the resolution of the image. This increase it back to roughly the same size as it was before

Pre #32359 resolution was 2830x1186
After this PR resoltuions 3135x1318

This is a screenshot. There is an export in Google Slide, but it exports to an even lower resolutions, by default. When I change the resolution of the slide, the image gets distorted in weird ways that makes it even more difficult to read as the arrow, shapes, and text do some funky scaling. Instead of wasting time trying to fix that, I simply took a screenshot on a 4k monitor.


bzbarsky-apple (approved):
Thank you!

cecille (approved):

yufengwangca (approved):

yunhanw-google (approved):

PR #32491, opened 2024-03-07
[Darwin] MTRDevice should support multiple delegates

jtung-apple:
This PR adds multiple delegate support for MTRDevice, and a unit test for it.

Also includes some comment fixes, logging additions, and MTRDevice os_unfair_lock usage cleanup for straight-forward lock/unlock situations.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice.h
+++ b/src/darwin/Framework/CHIP/MTRDevice.h
@@ -114,7 +114,26 @@ MTR_AVAILABLE(ios(16.1), macos(13.0), watchos(9.1), tvos(16.1))
  *
  * The delegate will be called on the provided queue, for attribute reports, event reports, and device state changes.
  */
-- (void)setDelegate:(id<MTRDeviceDelegate>)delegate queue:(dispatch_queue_t)queue;
+- (void)setDelegate:(id<MTRDeviceDelegate>)delegate queue:(dispatch_queue_t)queue MTR_NEWLY_DEPRECATED("Please use addDelegate:queue:interestedPaths:");
+
+/**
+ * Adds a delegate to receive asynchronous callbacks about the device.
+ *
+ * The delegate will be called on the provided queue, for attribute reports, event reports, and device state changes.
+ */
+- (void)addDelegate:(id<MTRDeviceDelegate>)delegate queue:(dispatch_queue_t)queue MTR_NEWLY_AVAILABLE;
+
+/**
+ * Adds a delegate to receive asynchronous callbacks about the device, and limit attribute reports to a specific set of paths.
+ *
+ * interestedAttributePaths may contain either MTRClusterPath or MTRAttributePath.
+ */
+- (void)addDelegate:(id<MTRDeviceDelegate>)delegate queue:(dispatch_queue_t)queue interestedAttributePaths:(NSArray *)interestedAttributePaths MTR_NEWLY_AVAILABLE;
+
+/**
+ * Removes the delegate from
+ */
+- (void)removeDelegate:(id<MTRDeviceDelegate>)delegate MTR_NEWLY_AVAILABLE;
 
 /**
  * Read attribute in a designated attribute path.  If there is no value available
@@ -389,7 +408,7 @@ MTR_EXTERN NSString * const MTRDataVersionKey MTR_AVAILABLE(ios(17.6), macos(14.
  *
  *                The data-value dictionary also contains this key:
  *
- *                MTRDataVersionKey : NSNumber-wrapped uin32_t. Monotonically increaseing data version for the cluster.
+ *                MTRDataVersionKey : NSNumber-wrapped uin32_t.
  */
 - (void)device:(MTRDevice *)device receivedAttributeReport:(NSArray<NSDictionary<NSString *, id> *> *)attributeReport;
 
--- a/src/darwin/Framework/CHIP/MTRDevice.h
+++ b/src/darwin/Framework/CHIP/MTRDevice.h
@@ -127,11 +127,13 @@ MTR_AVAILABLE(ios(16.1), macos(13.0), watchos(9.1), tvos(16.1))
  * Adds a delegate to receive asynchronous callbacks about the device, and limit attribute reports to a specific set of paths.
  *
  * interestedAttributePaths may contain either MTRClusterPath or MTRAttributePath.
+ *
+ * MTRDevice does not hold a strong reference to the delegate object.
  */
-- (void)addDelegate:(id<MTRDeviceDelegate>)delegate queue:(dispatch_queue_t)queue interestedAttributePaths:(NSArray *)interestedAttributePaths MTR_NEWLY_AVAILABLE;
+- (void)addDelegate:(id<MTRDeviceDelegate>)delegate queue:(dispatch_queue_t)queue interestedPathsForAttributes:(NSArray *)interestedPathsForAttributes MTR_NEWLY_AVAILABLE;
 
 /**
- * Removes the delegate from
+ * Removes the delegate from receiving callbacks abou the device.
  */
 - (void)removeDelegate:(id<MTRDeviceDelegate>)delegate MTR_NEWLY_AVAILABLE;
 
--- a/src/darwin/Framework/CHIP/MTRDevice.h
+++ b/src/darwin/Framework/CHIP/MTRDevice.h
@@ -133,7 +133,7 @@ MTR_AVAILABLE(ios(16.1), macos(13.0), watchos(9.1), tvos(16.1))
 - (void)addDelegate:(id<MTRDeviceDelegate>)delegate queue:(dispatch_queue_t)queue interestedPathsForAttributes:(NSArray *)interestedPathsForAttributes MTR_NEWLY_AVAILABLE;
 
 /**
- * Removes the delegate from receiving callbacks abou the device.
+ * Removes the delegate from receiving callbacks about the device.
  */
 - (void)removeDelegate:(id<MTRDeviceDelegate>)delegate MTR_NEWLY_AVAILABLE;
 
--- a/src/darwin/Framework/CHIP/MTRDevice.h
+++ b/src/darwin/Framework/CHIP/MTRDevice.h
@@ -120,17 +120,25 @@ MTR_AVAILABLE(ios(16.1), macos(13.0), watchos(9.1), tvos(16.1))
  * Adds a delegate to receive asynchronous callbacks about the device.
  *
  * The delegate will be called on the provided queue, for attribute reports, event reports, and device state changes.
+ *
+ * MTRDevice holds a weak reference to the delegate object.
  */
 - (void)addDelegate:(id<MTRDeviceDelegate>)delegate queue:(dispatch_queue_t)queue MTR_NEWLY_AVAILABLE;
 
 /**
- * Adds a delegate to receive asynchronous callbacks about the device, and limit attribute reports to a specific set of paths.
+ * Adds a delegate to receive asynchronous callbacks about the device, and limit attribute and/or event reports to a specific set of paths.
  *
- * interestedAttributePaths may contain either MTRClusterPath or MTRAttributePath.
+ * interestedPathsForAttributes may contain either MTRClusterPath or MTRAttributePath to specify interested clusters and attributes, or NSNumber for endpoints.
  *
- * MTRDevice does not hold a strong reference to the delegate object.
+ * interestedPathsForAttributes may contain either MTRClusterPath or MTREventPath to specify interested clusters and events, or NSNumber for endpoints.
+ *
+ * For both interested paths arguments, if nil is specified, then no filter will be applied.
+ *
+ * Calling addDelegate: again with the same delegate object will update the interested paths for attributes and events for this delegate.
+ *
+ * MTRDevice holds a weak reference to the delegate object.
  */
-- (void)addDelegate:(id<MTRDeviceDelegate>)delegate queue:(dispatch_queue_t)queue interestedPathsForAttributes:(NSArray *)interestedPathsForAttributes MTR_NEWLY_AVAILABLE;
+- (void)addDelegate:(id<MTRDeviceDelegate>)delegate queue:(dispatch_queue_t)queue interestedPathsForAttributes:(NSArray * _Nullable)interestedPathsForAttributes interestedPathsForEvents:(NSArray * _Nullable)interestedPathsForEvents MTR_NEWLY_AVAILABLE;
 
 /**
  * Removes the delegate from receiving callbacks about the device.
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -1052,7 +1052,7 @@ static NSString * const sLastInitialSubscribeLatencyKey = @"lastInitialSubscribe
     return NO;
 }
 
-- (void)_iterateDelegatesWithBlock:(void (^)(MTRDeviceDelegateInfo * delegateInfo))block
+- (void)_iterateDelegatesWithBlock:(void (NS_NOESCAPE ^)(MTRDeviceDelegateInfo * delegateInfo))block
 {
     os_unfair_lock_assert_owner(&self->_lock);
 
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -1834,7 +1834,7 @@ static NSString * const sLastInitialSubscribeLatencyKey = @"lastInitialSubscribe
 }
 
 // Returns filtered set of attributes using an interestedPaths array.
-// Returns nil if no attribute reports has a path that matches the paths in the interestedPaths array.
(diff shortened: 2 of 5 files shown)


woody-apple (changes requested):
Let's discuss the API here

jtung-apple:
@woody-apple 

I've changed the API to "addDelegate:..." and "removeDelegate:" for the new multiple delegate support. The old "setDelegate:..." will still work seamlessly.

woody-apple (approved):

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -922,23 +1044,76 @@ - (void)_readThroughSkipped
...
+    }
+
+    // Opportunistically remove defunct delegate references on every iteration
+    NSMutableSet * delegatesToRemove = [NSMutableSet set];
+    for (MTRDeviceDelegateInfo * delegateInfo in _delegates) {
+        if ([delegateInfo delegateIsNull]) {
+            [delegatesToRemove addObject:delegateInfo];
+        } else {
+            block(delegateInfo);
+        }
```
ksperling-apple, line 1073:
```suggestion
        id<MTRDeviceDelegate> strongDelegate = delegateInfo.delegate;
        if (strongDelegate) {
            block(delegateInfo);
            (void)strongDelegate; // ensure it stays alive
        } else {
            [delegatesToRemove addObject:delegateInfo];
        }
```

TOCTOU bug

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -922,23 +1044,76 @@ - (void)_readThroughSkipped
...
+- (void)_iterateDelegatesWithBlock:(void (^)(MTRDeviceDelegateInfo * delegateInfo))block
+{
+    os_unfair_lock_assert_owner(&self->_lock);
+
+    if (!_delegates.count) {
+        MTR_LOG("%@ no delegates to iterate", self);
+    }
+
+    // Opportunistically remove defunct delegate references on every iteration
+    NSMutableSet * delegatesToRemove = [NSMutableSet set];
```
ksperling-apple, line 1067: Given this method is likely called quite often it would be good to avoid allocating this temporary container until we actually know we need to remove a delegate.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -922,23 +1044,76 @@ - (void)_readThroughSkipped
...
-        return YES;
+    for (MTRDeviceDelegateInfo * delegateInfo in _delegates) {
+        if (!delegateInfo.delegateIsNull) {
+            return YES;
+        }
     }
     return NO;
 }
 
+- (void)_iterateDelegatesWithBlock:(void (^)(MTRDeviceDelegateInfo * delegateInfo))block
```
ksperling-apple, line 1058:
```suggestion
- (void)_iterateDelegatesWithBlock:(void (NS_NOESCAPE ^)(MTRDeviceDelegateInfo * delegateInfo))block
```

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -983,12 +1155,11 @@ - (void)_changeInternalState:(MTRInternalDeviceState)state
...
         /* BEGIN DRAGONS: This is a huge hack for a specific use case, do not rename, remove or modify behavior here */
         // TODO: This should only be called for thread devices
-        id<MTRDeviceDelegate> delegate = _weakDelegate.strongObject;
-        if ([delegate respondsToSelector:@selector(_deviceInternalStateChanged:)]) {
-            dispatch_async(_delegateQueue, ^{
-                [(id) delegate _deviceInternalStateChanged:self];
-            });
-        }
+        [self _callDelegatesWithBlock:^(id delegate) {
+            if ([delegate respondsToSelector:@selector(_deviceInternalStateChanged:)]) {
```
ksperling-apple, line 1159: I think it would be good to perform all the `respondsToSelector` calls when we add the delegate and cache this as a bit mask on the DelegateInfo, so we can avoid async-dispatching blocks that end up doing nothing.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -1647,28 +1809,75 @@ - (void)_handleReportEnd
...
     if (attributes.count) {
-        id<MTRDeviceDelegate> delegate = _weakDelegate.strongObject;
-        if (delegate) {
-            dispatch_async(_delegateQueue, ^{
-                [delegate device:self receivedAttributeReport:attributes];
-            });
-        }
+        [self _iterateDelegatesWithBlock:^(MTRDeviceDelegateInfo * delegateInfo) {
+            // make an autorelease pool so that temporary filtered attributes reports don't bloat memory
+            @autoreleasepool {
```
ksperling-apple, line 1872: It would probably be good to just have an `@autoreleasepool` in `_iterateDelegatesWithBlock` so it applies to all these iterations... auto release pool blocks are cheap.

src/darwin/Framework/CHIP/MTRDevice.h
```diff
@@ -114,7 +114,28 @@ MTR_AVAILABLE(ios(16.1), macos(13.0), watchos(9.1), tvos(16.1))
...
+ * Adds a delegate to receive asynchronous callbacks about the device, and limit attribute reports to a specific set of paths.
+ *
+ * interestedAttributePaths may contain either MTRClusterPath or MTRAttributePath.
+ *
+ * MTRDevice does not hold a strong reference to the delegate object.
+ */
+- (void)addDelegate:(id<MTRDeviceDelegate>)delegate queue:(dispatch_queue_t)queue interestedPathsForAttributes:(NSArray *)interestedPathsForAttributes MTR_NEWLY_AVAILABLE;
+
+/**
+ * Removes the delegate from receiving callbacks abou the device.
```
ksperling-apple, line 136:
```suggestion
 * Removes the delegate from receiving callbacks about the device.
```

src/darwin/Framework/CHIP/MTRDevice.h
```diff
@@ -114,7 +114,28 @@ MTR_AVAILABLE(ios(16.1), macos(13.0), watchos(9.1), tvos(16.1))
...
+- (void)addDelegate:(id<MTRDeviceDelegate>)delegate queue:(dispatch_queue_t)queue MTR_NEWLY_AVAILABLE;
+
+/**
+ * Adds a delegate to receive asynchronous callbacks about the device, and limit attribute reports to a specific set of paths.
+ *
+ * interestedAttributePaths may contain either MTRClusterPath or MTRAttributePath.
+ *
+ * MTRDevice does not hold a strong reference to the delegate object.
+ */
+- (void)addDelegate:(id<MTRDeviceDelegate>)delegate queue:(dispatch_queue_t)queue interestedPathsForAttributes:(NSArray *)interestedPathsForAttributes MTR_NEWLY_AVAILABLE;
```
ksperling-apple, line 133: Document that it's OK to re-add an already-added  delegate with this method, e.g. to update interestedPaths

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -922,23 +1044,76 @@ - (void)_readThroughSkipped
...
+- (BOOL)_delegateExists
 {
     os_unfair_lock_assert_owner(&self->_lock);
-    id<MTRDeviceDelegate> delegate = _weakDelegate.strongObject;
-    if (delegate) {
-        dispatch_async(_delegateQueue, ^{
-            block(delegate);
-        });
-        return YES;
+    for (MTRDeviceDelegateInfo * delegateInfo in _delegates) {
```
ksperling-apple, line 1050: You could just re-use `_iterateDelegatesWithBlock` here (maybe allow a `nil` block to only do a cleanup pass) and have that method return YES if there were any delegates alive.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -62,34 +62,90 @@
...
+@property (readonly) void * delegatePointerValue;
+
+- (instancetype)initWithDelegate:(id<MTRDeviceDelegate>)delegate queue:(dispatch_queue_t)queue interestedPathsForAttributes:(NSArray * _Nullable)interestedPathsForAttributes;
+
+// Returns YES if delegate and queue are both non-null, and the block is scheduled to run.
+- (BOOL)callDelegateWithBlock:(void (^)(id<MTRDeviceDelegate>))block;
+
+#ifdef DEBUG
+// Only used for unit test purposes - normal delegate should not expect or handle being called back synchronously.
+- (BOOL)callDelegateSynchronouslyWithBlock:(void (^)(id<MTRDeviceDelegate>))block;
```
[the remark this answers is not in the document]
jtung-apple: Filed https://github.com/project-chip/connectedhomeip/issues/33858

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -62,34 +62,90 @@
...
+#endif
+
+// Returns YES if the weak reference no longer points to a live object.
+- (BOOL)delegateIsNull;
```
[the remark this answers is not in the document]
jtung-apple: Hmm you're right about this. I'm looking at what I want to do.

src/darwin/Framework/CHIP/MTRDevice.h
```diff
@@ -114,7 +114,28 @@ MTR_AVAILABLE(ios(16.1), macos(13.0), watchos(9.1), tvos(16.1))
...
+ * Adds a delegate to receive asynchronous callbacks about the device.
+ *
+ * The delegate will be called on the provided queue, for attribute reports, event reports, and device state changes.
+ */
+- (void)addDelegate:(id<MTRDeviceDelegate>)delegate queue:(dispatch_queue_t)queue MTR_NEWLY_AVAILABLE;
+
+/**
+ * Adds a delegate to receive asynchronous callbacks about the device, and limit attribute reports to a specific set of paths.
+ *
+ * interestedAttributePaths may contain either MTRClusterPath or MTRAttributePath.
```
bzbarsky-apple, line 129: Can we also allow NSNumber for "whole endpoint"?

src/darwin/Framework/CHIP/MTRDevice.h
```diff
@@ -114,7 +114,28 @@ MTR_AVAILABLE(ios(16.1), macos(13.0), watchos(9.1), tvos(16.1))
...
+
+/**
+ * Adds a delegate to receive asynchronous callbacks about the device.
+ *
+ * The delegate will be called on the provided queue, for attribute reports, event reports, and device state changes.
+ */
+- (void)addDelegate:(id<MTRDeviceDelegate>)delegate queue:(dispatch_queue_t)queue MTR_NEWLY_AVAILABLE;
+
+/**
+ * Adds a delegate to receive asynchronous callbacks about the device, and limit attribute reports to a specific set of paths.
```
bzbarsky-apple, line 127: There's no control over event reporting?

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -62,34 +62,87 @@
...
-@interface MTRWeakReference () {
+// container of MTRDevice delegate weak reference, its queue, and its interested paths for attribute reports
+MTR_DIRECT_MEMBERS
+@interface MTRDeviceDelegateInfo : NSObject {
 @private
-    __weak id _object;
+    void * _delegatePointerValue;
+    __weak id _delegate;
+    dispatch_queue_t _queue;
+    NSArray * _interestedPathsForAttributes;
```
bzbarsky-apple, line 73: Should be marked nullable?

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -62,34 +62,87 @@
...
 }
+
+// Array of interested cluster paths or attribute paths, for attribute report filtering.
+@property (readonly) NSArray * interestedPathsForAttributes;
```
bzbarsky-apple, line 77: Nullable?

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -62,34 +62,87 @@
...
+    
+    block(strongDelegate);
+
+    return YES;
+}
+#endif
+
+- (BOOL)delegateIsNull
+{
+    return (!_delegate || !_queue);
```
bzbarsky-apple, line 140: Will this really do the right thing?  I thought `_delegate` (the weak pointer) will still be non-nil, but trying to assign to a strong pointer will cause it to be nil?  Or is that not how it works?

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -469,7 +522,7 @@ - (instancetype)initWithNodeID:(NSNumber *)nodeID controller:(MTRDeviceControlle
         _clusterDataToPersist = nil;
         _persistedClusters = [NSMutableSet set];
 
-        // If there is a data store, make sure we have an observer to
+        // If there is a data store, make sure we have an observer to monitor system clock changes, so NSDate-based write coalescing would
```
bzbarsky-apple, line 525: This sentence still has no end.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -1647,28 +1806,75 @@ - (void)_handleReportEnd
...
+    }];
+#endif
+}
+
+- (BOOL)_interestedPaths:(NSArray * _Nullable)interestedPaths includesPath:(MTRAttributePath *)attributePath
+{
+    for (id interestedPath in interestedPaths) {
+        if ([interestedPath isKindOfClass:[MTRClusterPath class]]) {
+            MTRClusterPath * interestedClusterPath = interestedPath;
+            if (interestedClusterPath.cluster.unsignedLongValue == attributePath.cluster.unsignedLongValue) {
```
bzbarsky-apple, line 1822: Is there a reason to not isEqualToNumber here and below?

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -1647,28 +1806,75 @@ - (void)_handleReportEnd
...
+            }
+        }
     }
-#endif
+
+    return NO;
+}
+
+// Returns filtered set of attributes using an interestedPaths array.
+// Returns nil if no attribute reports has a path that matches the paths in the interestedPaths array.
```
bzbarsky-apple, line 1837:
```suggestion
// Returns nil if no attribute report has a path that matches the paths in the interestedPaths array.
```

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -1647,28 +1806,75 @@ - (void)_handleReportEnd
...
+- (NSArray<NSDictionary<NSString *, id> *> *)_filteredAttributes:(NSArray<NSDictionary<NSString *, id> *> *)attributes forInterestedPaths:(NSArray * _Nullable)interestedPaths
+{
+    if (!interestedPaths.count) {
```
bzbarsky-apple, line 1840: So this treats empty array as "allow all".  This is not how the API documented it....

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -1647,28 +1806,75 @@ - (void)_handleReportEnd
...
+        return attributes;
+    }
+
+    NSMutableArray * filteredAttributes = nil;
+    for (NSDictionary<NSString *, id> * responseValue in attributes) {
+        MTRAttributePath * attributePath = responseValue[MTRAttributePathKey];
+        if ([self _interestedPaths:interestedPaths includesPath:attributePath]) {
```
bzbarsky-apple, line 1847: A question.  Should `_interestedPaths:includesPath:` be a class method?  It does not use any "self" state....  Not sure whether that's a convention ObjC uses.

Or even a static function or something.

src/darwin/Framework/CHIP/MTRDevice.mm
ksperling-apple: We don't seem to consistently do this in `.mm` files currently, but we should (including `NS_ASSUME_NONNULL...` etc)

src/darwin/Framework/CHIP/MTRDevice.mm
ksperling-apple: Static methods are a little tedious to use in ObjC because you have to explicitly call them with the class (or `self.class`), so I think it's fine for helpers to not use state and still be instance methods. static functions are OK can feel inconsistent due to the different style signatures. I think we should just put `MTR_DIRECT_MEMBERS` on all `@implementation` blocks, then these internal helpers effectively become static C functions but with ObjC syntax.

src/darwin/Framework/CHIP/MTRDevice.mm
ksperling-apple: This is treating either "no delegate" or "no queue" as "we have no delegate", however as mentioned above I think we should just hold the queue strongly and only the delegate weakly, and that will simplify this.

src/darwin/Framework/CHIP/MTRDevice.mm
jtung-apple: And added tests for it.

PR #32495, opened 2024-03-07
Add OperationalState to chef examples

tehampson:
`cluster-init-callback.cpp` is autogenerated when chef is compiled. It has a weak reference to `emberAfOperationalStateClusterInitCallback`. As a result any command before this PR to `OperationalState` would fail with `UNSUPPORTED_CLUSTER`

Files were copied from `examples/all-clusters-app/all-clusters-common/src/operational-state-delegate-impl.cpp` and `examples/all-clusters-app/all-clusters-common/include/operational-state-delegate-impl.h`

Test:
* `cd examples/chef`
* `python chef.py -d rootnode_laundrywasher_fb10d238c8 -t linux -b`
* Run `./rootnode_laundrywasher_fb10d238c8` in a terminal and commission with chip-tool
* `./chip-tool operationalstate stop 0x12344321 1`

### DIFF
--- a/examples/chef/common/chef-operational-state-delegate-impl.cpp
+++ b/examples/chef/common/chef-operational-state-delegate-impl.cpp
@@ -1,0 +1,219 @@
+/*
+ *    Copyright (c) 2023 Project CHIP Authors
+ *    All rights reserved.
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


tcarmelveilleux (approved):

tcarmelveilleux (approved):

marktrayer (approved):

younghak-hwang (approved):

examples/chef/common/chef-operational-state-delegate-impl.cpp
```diff
@@ -0,0 +1,219 @@
...
+void GenericOperationalStateDelegateImpl::HandleResumeStateCallback(GenericOperationalError & err)
+{
+    OperationalState::OperationalStateEnum state =
+        static_cast<OperationalState::OperationalStateEnum>(GetInstance()->GetCurrentOperationalState());
+
+    if (state == OperationalState::OperationalStateEnum::kStopped || state == OperationalState::OperationalStateEnum::kError)
+    {
+        err.Set(to_underlying(OperationalState::ErrorStateEnum::kCommandInvalidInState));
+        return;
+    }
```
bzbarsky-apple, line 87: This is duplicating cluster code.  Why is it doing that?  Those checks happen before this method is ever called.

Similar for various other bits in this file.

PR #32497, opened 2024-03-08
[Python] Make all subscription callbacks async capable

agners:
Support asyncio coroutines as callback arguments. This allows to write asyncio code easily. Still support regular callbacks.

This changes on what thread some of the callbacks are being made. So far, some of the callbacks have been made directly (from the Matter SDK thread), and some from the callee thread (from the event loop thread learned via `asyncio.get_running_loop()`).

- SetResubscriptionAttemptedCallback -> even loop thread
- SetResubscriptionSucceededCallback -> even loop thread
- SetAttributeUpdateCallback -> SDK thread
- SetEventUpdateCallback -> SDK thread
- SetErrorCallback -> even loop thread

~~With this PR, all callbacks are done in the even loop thread, no matter if coroutines are used or regular routines.~~

Calling all callbacks in the event loop thread causes issues for the `ZCLSubscribeAttribute`  (legacy?) API. This PR now simply makes all callbacks asyncio capable. Coroutines as callback can only be used when using the new `WriteAttribute()`/`ReadAttribute()` API.

This also addresses at least one bug where `create_task()` is used on the even loop from the SDK thread. `create_task()` is not thread safe. This PR uses the thread safe alternative `run_coroutine_threadsafe()`.

It also speeds up the tests slightly by avoiding fixed sleeps in `TestCaseEviction`.

### DIFF
--- a/src/controller/python/chip/clusters/Attribute.py
+++ b/src/controller/python/chip/clusters/Attribute.py
@@ -557,9 +557,9 @@ class SubscriptionTransaction:
         if callback is not None:
             self._onEventChangeCb = callback
 
-    def SetErrorCallback(self, callback: Callable[[int, SubscriptionTransaction], None]):
+    def SetErrorCallback(self, callback: Callable[[int, SubscriptionTransaction], Union[None, Awaitable[None]]]):
         '''
-        Sets the callback function in case a subscription error occured,
+        Sets the callback function in case a subscription error occurred,
         accepts a Callable accepts an error code and the cached data.
         '''
         if callback is not None:
@@ -574,7 +574,7 @@ class SubscriptionTransaction:
         return self._onEventChangeCb
 
     @property
-    def OnErrorCb(self) -> Callable[[int, SubscriptionTransaction], None]:
+    def OnErrorCb(self) -> Callable[[int, SubscriptionTransaction], Union[None, Awaitable[None]]]:
         return self._onErrorCb
 
     @property
@@ -787,7 +787,7 @@ class AsyncReadTransaction:
             # Clear it out once we've notified of all changes in this transaction.
         self._changedPathSet = set()
 
-    def _handleDone(self):
+    async def _handleDone(self):
         #
         # We only set the exception/result on the future in this _handleDone call (if it hasn't
         # already been set yet, which can be in the case of subscriptions) since doing so earlier
@@ -797,7 +797,10 @@ class AsyncReadTransaction:
         if not self._future.done():
             if self._resultError:
                 if self._subscription_handler:
-                    self._subscription_handler.OnErrorCb(self._resultError, self._subscription_handler)
+                    if inspect.iscoroutinefunction(self._subscription_handler.OnErrorCb):
+                        await self._subscription_handler.OnErrorCb(self._resultError, self._subscription_handler)
+                    else:
+                        self._subscription_handler.OnErrorCb(self._resultError, self._subscription_handler)
                 else:
                     self._future.set_exception(chip.exceptions.ChipStackError(self._resultError))
             else:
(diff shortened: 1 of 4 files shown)


agners:
I've added some additional commits which convert more of the tests to make use of the asyncio framework.

src/controller/python/chip/clusters/Attribute.py
```diff
@@ -723,9 +724,12 @@ def handleEventData(self, header: EventHeader, path: EventPath, data: bytes, sta
                 Header=header, Data=eventValue, Status=chip.interaction_model.Status(status))
             self._events.append(eventResult)
 
-            if (self._subscription_handler is not None):
-                self._subscription_handler.OnEventChangeCb(
-                    eventResult, self._subscription_handler)
+            if self._subscription_handler is not None:
+                callback = self._subscription_handler.OnEventChangeCb
+                if inspect.iscoroutinefunction(callback):
```
balloob, line 729: This call is quite expensive. Scheduling it as a task just to forward a piece of information can be quite some overhead. Would it be possible to allow decorating functions that they need to be run inside the event loop ? In Home Assistant we have a [`@callback`](https://github.com/home-assistant/core/blob/dev/homeassistant/core.py#L220-L228) decorator to indicate that a function is to be called in the event loop.

src/controller/python/chip/clusters/Attribute.py
```diff
@@ -723,9 +724,12 @@ def handleEventData(self, header: EventHeader, path: EventPath, data: bytes, sta
...
+                    asyncio.run_coroutine_threadsafe(callback(), self._event_loop)
```
balloob, line 730: shouldn't the arguments be the same regardless if it will be run in the event loop or not?

src/controller/python/chip/clusters/Attribute.py
agners: > This call is quite expensive. Scheduling it as a task just to forward a piece of information can be quite some overhead. Would it be possible to allow decorating functions that they need to be run inside the event loop ? In Home Assistant we have a [@callback](https://github.com/home-assistant/core/blob/dev/homeassistant/core.py#L220-L228) decorator to indicate that a function is to be called in the event loop.

Looking into the implementation of `iscoroutinefunction` it seems quite simple tough. 

Quick test:
```python
import timeit
setup="""
from inspect import iscoroutinefunction
async def test():
    pass
"""
timeit.timeit("iscoroutinefunction(test)",setup=setup)

setup="""\
def callback(func):
    setattr(func, "_hass_callback", True)
    return func
def is_callback(func) -> bool:
    return getattr(func, "_hass_callback", False) is True
@callback
async def test():
    pass
"""
timeit.timeit("is_callback(test)",setup=setup)
```

`iscoroutinefunction` is ~26s vs. `is_callback` ~5.4s, so quite a bit faster indeed.

There are actually two question here:
- Should we callback on the event loop or in the Thread the SDK calls us?
- Should we support coroutine functions (as in `async def...`)

In the end, both handled by the callback itself, essentially leave it up to the user of the read/write attribute APIs.

I think my main gripe here is that different callbacks have different semantics right now. And they are also not documented :cry: 

I probably gets a bit philosophical. What do you think how this API should behave? :thinking:

src/controller/python/chip/clusters/Attribute.py
agners: Right, good catch!

agners:
After further discussions internally we actually agreed that it is a more versatile API design to make all callbacks straight from the Thread the SDK makes the callback from. Some callbacks can be resorted already on that Thread, without the need to load the main event loop.

The main issue with the current implementation is that the callbacks are handled differently, and the semantics are not exactly documented. I intend to create a new PR which makes all callbacks to be called from the Thread directly, and document this accordingly. I expect that some tests will need adjustments as well.

agners:
Closing, reasoning see above.

PR #32501, opened 2024-03-08
Fix test app path in cirque test.

wqx6:
https://github.com/project-chip/connectedhomeip/pull/32496 accidentally disable lit-icd-app tests for subscription resumption, it may relate to the problematic test app path. This PR fixes this issue.


yunhanw-google (approved):

mkardous-silabs:
@wqx6 @yunhanw-google I'm not sure this is a good idea. LIT ICD app doesn't enable the timeout resumption feature since it conflicts with the Check-In Protocol.

wqx6:
> > @wqx6 @yunhanw-google I'm not sure this is a good idea. LIT ICD app doesn't enable the timeout resumption feature since it conflicts with the Check-In Protocol.
> 
> Agree. Thanks
> 
> Then we just merge the change without replacing all cluster app with lit example app, since the rest of change are still valuable for later lit test later. Thanks

Done

mkardous-silabs (approved):

PR #32506, opened 2024-03-08
Add Pigweed support for efr32

jlatusek:
Related to #29682

## Changes
* Add pigweed dependencies for gn 
* Add pigweed unit test service
* Create `pw_test_runner` script similar to `nl_test_runner`

## Testing
@jepenven-silabs 
I do not have the resources to test these changes. I would ask someone associated with the efr32 platform to test them and give feedback.

### DIFF
--- a/src/BUILD.gn
+++ b/src/BUILD.gn
@@ -145,7 +145,8 @@ if (chip_build_tests) {
       # TODO [PW_MIGRATION] There will be a list of already migrated platforms
       if (chip_device_platform == "esp32" ||
           chip_device_platform == "nrfconnect" ||
-          chip_device_platform == "openiotsdk") {
+          chip_device_platform == "openiotsdk" ||
+          chip_device_platform == "efr32") {
         deps += [ "${chip_root}/src/lib/support:pw_tests_wrapper" ]
       }
       build_monolithic_library = true
--- a/src/test_driver/efr32/.gn
+++ b/src/test_driver/efr32/.gn
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import("//build_overrides/build.gni")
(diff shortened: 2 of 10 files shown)


yufengwangca (approved):

jmartinez-silabs:
@jlatusek Does those pigweed change impact pw-rpc apps like the lighting app for example? I don't see any changes to that app gn files ? 

@tehampson Do you know?

tehampson:
> @jlatusek Does those pigweed change impact pw-rpc apps like the lighting app for example? I don't see any changes to that app gn files ?
> 
> @tehampson Do you know?

@jmartinez-silabs In what way do you think that this change impacts pw-rpc apps? My understanding of this change is needed to allow for pw unit testing. At a quick glance it is not clear to me how this would impact the pw-rpc of the example apps

jmartinez-silabs:
@tehampson I don't know. I see a bunch of new packages and dependencies mentioned. Yes they seem targeted to "test wrapper" but I don't know if this impact let's say RPC console. Therefore my questions. If there are completely separated (unit Test framework and RPC console) then my questions can be discarded

tehampson:
@jmartinez-silabs I don't think it does. There is a relationship where test-wrapper needs pw-rpc to stream the results of running the on device unit test. So that is why you see it added to `efr32_device_tests`. So unless examples are built using `efr32_device_tests`, I don't think there is an issue here. Unless I am missing something

jlatusek:
@jmartinez-silabs 
These changes are not intended to affect pw-rpc example apps like the lighting app. All I want to do is replace the nlunit tests with pigweed ones. Right now, no tests have been modified because I wanted to prepare each platform by properly modifying the build system and then modifying the tests.

PR #32508, opened 2024-03-08
Implement a Vector-based TLVBackingStore

yufengwangca:
When converting from stream-based input to TLV format, such as ArrayInputStream to TLV, accurately pre-estimating the required buffer size for the TLVWriter poses a significant challenge. 

To address this, it is necessary to implement a vector-based TLVBackingStore, which is capable of dynamically adjusting its size to accommodate the varying data sizes encountered during the conversion process. This approach ensures that the TLVWriter can efficiently handle inputs of any size without the risk of running out of buffer space, thereby enhancing the robustness and flexibility of the data serialization process.

### DIFF
--- a/src/lib/core/TLVVectorWriter.cpp
+++ b/src/lib/core/TLVVectorWriter.cpp
@@ -32,20 +32,20 @@ constexpr uint32_t kIpv6MtuSize = 1280;
 
 } // namespace
 
-TlvVectorWriter::TlvVectorWriter(std::vector<uint8_t> * buffer) : buffer_(buffer)
+TlvVectorWriter::TlvVectorWriter(std::vector<uint8_t> & buffer) : mVectorBuffer(buffer)
 {
-    Init(buffer_);
+    Init(mVectorBuffer);
 }
 
 TlvVectorWriter::~TlvVectorWriter() = default;
 
-TlvVectorWriter::TlvVectorBuffer::TlvVectorBuffer(std::vector<uint8_t> * buffer) : final_buffer_(*buffer) {}
+TlvVectorWriter::TlvVectorBuffer::TlvVectorBuffer(std::vector<uint8_t> & buffer) : mFinalBuffer(buffer) {}
 
 TlvVectorWriter::TlvVectorBuffer::~TlvVectorBuffer() {}
 
 CHIP_ERROR TlvVectorWriter::TlvVectorBuffer::OnInit(TLVWriter & /*writer*/, uint8_t *& bufStart, uint32_t & bufLen)
 {
-    VerifyOrReturnError(final_buffer_.empty(), CHIP_ERROR_INCORRECT_STATE);
+    VerifyOrReturnError(mFinalBuffer.empty(), CHIP_ERROR_INCORRECT_STATE);
 
     ResizeWriteBuffer(bufStart, bufLen);
 
@@ -54,8 +54,8 @@ CHIP_ERROR TlvVectorWriter::TlvVectorBuffer::OnInit(TLVWriter & /*writer*/, uint
 
 CHIP_ERROR TlvVectorWriter::TlvVectorBuffer::GetNewBuffer(TLVWriter & /*writer*/, uint8_t *& bufStart, uint32_t & bufLen)
 {
-    VerifyOrReturnError(!final_buffer_.empty(), CHIP_ERROR_INCORRECT_STATE);
-    VerifyOrReturnError(writing_buffer_.data() == bufStart, CHIP_ERROR_INCORRECT_STATE);
+    VerifyOrReturnError(!mFinalBuffer.empty(), CHIP_ERROR_INCORRECT_STATE);
+    VerifyOrReturnError(mWritingBuffer.data() == bufStart, CHIP_ERROR_INCORRECT_STATE);
 
     ResizeWriteBuffer(bufStart, bufLen);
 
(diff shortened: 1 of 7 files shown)


yunhanw-google (approved):

mwswartwout (approved):

src/lib/core/TLVVectorWriter.cpp
```diff
@@ -0,0 +1,93 @@
...
+namespace chip {
+namespace TLV {
+
+namespace {
+
+constexpr uint32_t kIpv6MtuSize = 1280;
+
+} // namespace
+
+TlvVectorWriter::TlvVectorWriter(std::vector<uint8_t> * buffer) : buffer_(buffer)
```
bzbarsky-apple, line 35: Is there a reason this takes pointer, not reference?  Is passing nullptr ok?

src/lib/core/TLVVectorWriter.h
```diff
@@ -0,0 +1,83 @@
...
+#include <lib/core/TLVBackingStore.h>
+#include <lib/core/TLVCommon.h>
+
+namespace chip {
+namespace TLV {
+
+// Implementation of TLVWriter that writes to a std::vector, automatically
+// resizing the vector as needed when writing.
+// Users of TlvVectorWriter may call any public API of TLVWriter, except for the
+// Init functions.
```
bzbarsky-apple, line 32: Why doesn't this class just have a private Init, or a deleted one?  That would prevent people calling Init by accident.

src/lib/core/TLVVectorWriter.h
```diff
@@ -0,0 +1,83 @@
...
+    private:
+        void ResizeWriteBuffer(uint8_t *& bufStart, uint32_t & bufLen);
+
+        // writing_buffer_ is the mutable buffer exposed via the TLVBackingStore
+        // interface. When FinalizeBuffer is called the contents of writing_buffer_
+        // are appended to final_buffer_ and writing_buffer_ is cleared. This allows
+        // for reading all written data from a single, contiguous buffer
+        // (final_buffer_).
+        std::vector<uint8_t> writing_buffer_;
+        std::vector<uint8_t> & final_buffer_;
```
bzbarsky-apple, line 76: Local style in this directory is to use `mSomething` for members if they have special naming at all, not trailing `_`, no?

src/lib/core/TLVVectorWriter.h
```diff
@@ -0,0 +1,83 @@
...
+    };
+
+    TlvVectorBuffer buffer_;
```
bzbarsky-apple, line 79: See above about naming.

src/lib/core/tests/TestTLVVectorWriter.cpp
```diff
@@ -0,0 +1,180 @@
...
+    NL_TEST_ASSERT(inSuite, reader.Next() == CHIP_NO_ERROR);
+    NL_TEST_ASSERT(inSuite, reader.GetTag() == AnonymousTag());
+
+    CharSpan span;
+    error = reader.Get(span);
+    NL_TEST_ASSERT(inSuite, error == CHIP_NO_ERROR);
+    NL_TEST_ASSERT(inSuite, std::string(span.data(), span.size()) == bytes);
+}
+
+void ManySmallDataRequiresNewBufferAllocation(nlTestSuite * inSuite, void * inContext)
```
pidarped, line 102: Is it possible to add a test to allocate a PacketBuffer(from heap) after a large TLV vector is populated and copy the contents of the vector into the allocated PacketBuffer?

src/lib/core/TLVVectorWriter.h
yufengwangca: Yes, passing by reference is better which guarantee to have a valid buffer

src/lib/core/tests/TestTLVVectorWriter.cpp
yufengwangca: We should not have this scenario, as the PacketBuffer is typically used for content transmission on the wire, while the TlvVectorBuffer is used for creating the TLV binary format to be stored within the PacketBuffer.

jmartinez-silabs (approved):

src/lib/core/TLVVectorWriter.cpp
```diff
@@ -0,0 +1,93 @@
...
+
+    mFinalBuffer.insert(mFinalBuffer.end(), mWritingBuffer.begin(), mWritingBuffer.end());
+    mWritingBuffer.resize(0);
+
+    mFinalBuffer.shrink_to_fit();
+
+    return CHIP_NO_ERROR;
+}
+
+void TlvVectorWriter::TlvVectorBuffer::ResizeWriteBuffer(uint8_t *& bufStart, uint32_t & bufLen)
```
tehampson, line 80: I am a little confused, what is actually resizing the `mWritingBuffer` vector in the test `ManySmallDataRequiresNewBufferAllocation`. How is there a guarantee that the address won't change?

yunhanw-google: ManySmallDataRequiresNewBufferAllocation is wring significant oversize buffer, which would trigger GetNewBuffer, then ResizeWriteBuffer is called,  mWritingBuffer is a vector in TLVVectorWrite, and its starting address is put in bufStart..and it is not changing.

Issue #32512, opened 2024-03-08
[BUG] dns-sd resolution APIs don't have a way to hand back results found on multiple interfaces


bzbarsky-apple:
### Reproduction steps

ResolvedNodeData (via CommonResolutionData) only allows specifying one resolve result (interface, hostname, port).  It does allow multiple IPs, but that does not help when the registration result found on some interface is a stale cache while the right result is available on a different interface.

Then the "platform" DNS-SD apis only allow one ResolvedNodeData to be handed back.

We really need to move to a situation where all the results are handed back, so OperationalSessionSetup can try them all as needed.

### Bug prevalence

Always

### GitHub hash of the SDK that was being used

28da08fbdcba82144be73ebffc9cf2f9a0be660f

### Platform

core

PR #32513, opened 2024-03-08
Prioritize DNS-SD results from the network on tvOS.

bzbarsky-apple:
Since we can only deliver results from one interface given the current state of Matter SDK APIs (see
https://github.com/project-chip/connectedhomeip/issues/32512), prioritize the interfaces that are more likely to have up-to-date results, not stale caches.

### DIFF
--- a/src/platform/Darwin/DnssdContexts.cpp
+++ b/src/platform/Darwin/DnssdContexts.cpp
@@ -21,6 +21,8 @@
 #include <lib/support/CHIPMemString.h>
 #include <platform/CHIPDeviceLayer.h>
 
+#include <net/if.h>
+
 using namespace chip::Dnssd;
 using namespace chip::Dnssd::Internal;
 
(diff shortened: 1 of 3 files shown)


src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -516,32 +518,35 @@ void ResolveContext::DispatchSuccess()
...
     bool needDelete = MdnsContexts::GetInstance().RemoveWithoutDeleting(this);
 
-    for (auto & interface : interfaces)
-    {
-        auto & ips = interface.second.addresses;
+#if TARGET_OS_TV
+    // On tvOS, priotize results from en0, en1, ir0 in that order, if those
+    // interfaces are present, since those will generally have more up-to-date
+    // information.
+    static const unsigned int priorityInterfaceIndices[] = {
```
Abhayakara, line 525: You could do this explicitly for the non-tvOS case as well, just in case lo0 isn't the lowest interface index for some reason.

src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -516,32 +518,35 @@ void ResolveContext::DispatchSuccess()
     // ChipDnssdResolveNoLongerNeeded don't find us and try to also remove us.
     bool needDelete = MdnsContexts::GetInstance().RemoveWithoutDeleting(this);
 
-    for (auto & interface : interfaces)
-    {
-        auto & ips = interface.second.addresses;
+#if TARGET_OS_TV
+    // On tvOS, priotize results from en0, en1, ir0 in that order, if those
```
Abhayakara, line 522: s/priotize/prioritize/

src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -516,32 +518,35 @@ void ResolveContext::DispatchSuccess()
...
-        auto addresses = Span<Inet::IPAddress>(ips.data(), ips.size());
-        if (nullptr == callback)
-        {
-            auto delegate = static_cast<CommissioningResolveDelegate *>(context);
-            DiscoveredNodeData nodeData;
-            service.ToDiscoveredNodeData(addresses, nodeData);
-            delegate->OnNodeDiscovered(nodeData);
-        }
-        else
+    for (auto & interface : interfaces)
```
Abhayakara, line 544: My only concern about this approach is that it will return the result for the same interface twice if it happens to be in the priority interface list.

Abhayakara (approved):
This looks good. Some caveats if you want to fix them.

src/platform/Darwin/DnssdContexts.cpp
bzbarsky-apple: Done.

src/platform/Darwin/DnssdContexts.cpp
bzbarsky-apple: Done.

woody-apple (approved):

src/platform/Darwin/DnssdContexts.cpp
bzbarsky-apple: No, because we never reach this code if we found the thing in the priority interface list and it actually had IP addresses.  Note the "return" in the case when TryReportingResultsForInterfaceIndex returned true.

woody-apple (approved):

src/platform/Darwin/DnssdContexts.cpp
Abhayakara: Right. Thanks. :)

arkq (approved):

PR #24228, opened 2023-01-02
SetupPayload: Python implementation to generate manual and QRCode

shubhamdp:
#### Problem
- No python script to generate Matter onboarding payloads.

#### Change Overview
- Functions to Base38 encode the payload
- Script to generate the manual pairing code and QRCode

#### Testing
- Validated the results by script which generates the random input values, then generate the QRCode and manual pairing code using `chip-tool` and python script and compare the results.
- Attaching [results.txt](https://github.com/project-chip/connectedhomeip/files/10331153/results.txt) when run against 100 different combinations
- Below is the core snippet, all the values are generated using `RANDOM` variable
```
ct_qr=$($CHIP_TOOL payload generate-qrcode --discriminator $disc --setup-pin-code $passcode --vendor-id $vid --product-id $pid --commissioning-mode $cf --rendezvous $dm | awk '{print $NF}' | awk -F ^[ print $1}')

ct_manual=$($CHIP_TOOL payload generate-manualcode --discriminator $disc --setup-pin-code $passcode --vendor-id $vid --product-id $pid --commissioning-mode $cf | awk '{print $NF}' | awk -F ^[ print $1}')

py_codes=$($PY_GEN -d $disc -p $passcode -cf $cf -vid $vid -pid $pid -dm $dm | xargs)
py_manual=$(echo $py_codes | awk '{print $3}')
py_qr=$(echo $py_codes | awk '{print $6}')

if [ "$ct_qr" = "$py_qr" ] && [ "$ct_manual" = "$py_manual" ]; then
    status="PASS"
else
    status="FAIL"
fi
```

PR #32516, opened 2024-03-09
setup_payload: Add support for parsing setup payloads in python impl

shubhamdp:
- Base38 decode impl in python
- use construct to generate/parse setup payload in python
- Add cli to parse and generate using click
- unit tests for parsing and verification using chip-tool
- removed the older script which only generated the codes
- replaced the usage of older utility with newer one


src/setup_payload/python/requirements.txt
```diff
@@ -1,2 +1,4 @@
-bitarray==2.6.0
+bitarray==2.8.1
+click==8.1.3
```
tcarmelveilleux, line 2: >=8.1.3

src/setup_payload/python/requirements.txt
```diff
@@ -1,2 +1,4 @@
...
+construct==2.10.68
```
tcarmelveilleux, line 3:
```suggestion
construct>=2.10.68
```

src/setup_payload/tests/run_python_setup_payload_test.py
```diff
@@ -62,6 +62,27 @@ def parse_setup_payload(chip_tool, payload):
     return parsed_params
 
 
+def chip_tool_generate_code(chip_tool, payload_data, cmd, key, is_qrcode):
```
tcarmelveilleux, line 65: This is not OK --> this should be logic since it's not portable to call chip-tool

shubhamdp: I have added some test data set and parsing the payload and checking against it.

Keeping the earlier test for validating generated payload and parsing it with chip-tool, as suggested by @bzbarsky-apple https://github.com/project-chip/connectedhomeip/pull/24228#pullrequestreview-1235034454

src/setup_payload/python/SetupPayload.py
```diff
@@ -0,0 +1,220 @@
...
+            'padding': 0,
+        }
+
+    def manualcode_dict(self):
+        return {
+            'version': 0,
+            'vid_pid_present': 0 if self.flow == CommissioningFlow.Standard else 1,
+            'discriminator': self.short_discriminator,
+            'pincode_lsb': self.pincode & 0x3FFF,    # 14 ls-bits
+            'pincode_msb': self.pincode >> 14,       # 13 ms-bits
```
tcarmelveilleux, line 100: Why as lsb/msb? Suggest having a member that is the "passcode"

Also suggest using @dataclass (Data classes) for these (as a follow-up) and use dataclasses.asdict(x) to dict-ify them for dicts versions.

src/setup_payload/python/SetupPayload.py
```diff
@@ -0,0 +1,220 @@
...
+
+        bytes = manualcode_format.build(self.manualcode_dict())
+        bits = bitarray()
+        bits.frombytes(bytes)
+
+        chunk1 = str(int(bits[CHUNK1_START:CHUNK1_START + CHUNK1_LEN].to01(), 2)).zfill(1)
+        chunk2 = str(int(bits[CHUNK2_START:CHUNK2_START + CHUNK2_LEN].to01(), 2)).zfill(5)
+        chunk3 = str(int(bits[CHUNK3_START:CHUNK3_START + CHUNK3_LEN].to01(), 2)).zfill(4)
+        chunk4 = str(self.vid).zfill(5) if self.flow != CommissioningFlow.Standard else ''
+        chunk5 = str(self.pid).zfill(5) if self.flow != CommissioningFlow.Standard else ''
```
tcarmelveilleux, line 127: Why `str` rather than `bytes` here?

src/setup_payload/tests/run_python_setup_payload_test.py
```diff
@@ -0,0 +1,213 @@
...
+    p['ProductID'] = pid
+    p['Custom flow'] = flow
+    p['Discovery Bitmask'] = discovery
+    p['Long discriminator'] = discriminator
+    p['Short discriminator'] = short_discriminator if short_discriminator is not None else (discriminator >> 8)
+    p['Passcode'] = passcode
+    return p
+
+
+def test_code_generation(chip_tool):
```
tcarmelveilleux, line 104: Suggest using plain old `unittest` module to convert the tests in here. Can be done as a follow-up.

src/setup_payload/python/requirements.txt
```diff
@@ -1,2 +1,4 @@
-bitarray==2.6.0
+bitarray==2.8.1
```
tcarmelveilleux, line 1: >=2.8.1

tcarmelveilleux (approved):

src/setup_payload/python/SetupPayload.py
shubhamdp: At the end we have to convert the numbers to string and to me this seems easier as we can pad the zeros using zfill().

src/setup_payload/python/SetupPayload.py
shubhamdp: Raised https://github.com/project-chip/connectedhomeip/issues/33549 for follow-up

src/setup_payload/tests/run_python_setup_payload_test.py
shubhamdp: https://github.com/project-chip/connectedhomeip/issues/33549 for follow up

dhrishi:
@andy31415 Can you please review?

PR #32520, opened 2024-03-10
Fix bug in level control move [bug fix]

volodymyr-zvarun-globallogic:
Command `chip-tool levelcontrol move 0 0 1 1 42 2` throws error `Stop reason: signal SIGFPE: integer divide by zero` because of division by zero in rate handling in LevelControl cluster. This PR checks if rate equal to zero, and if it is, consider it to be null.

### DIFF
--- a/src/app/clusters/level-control/level-control.cpp
+++ b/src/app/clusters/level-control/level-control.cpp
@@ -952,6 +952,12 @@ static void moveHandler(app::CommandHandler * commandObj, const app::ConcreteCom
         goto send_default_response;
     }
 
+    if (rate == 0) {
+      status = Status::Success;
+      goto send_default_response;
+    }
+
+
     if (!shouldExecuteIfOff(endpoint, commandId, optionsMask, optionsOverride))
     {
         status = Status::Success;


src/app/clusters/level-control/level-control.cpp
```diff
@@ -657,6 +657,10 @@ bool emberAfLevelControlClusterMoveCallback(app::CommandHandler * commandObj, co
     auto & optionsMask     = commandData.optionsMask;
     auto & optionsOverride = commandData.optionsOverride;
 
+    if (rate == 0) {
+      rate.SetNull();
+    }
```
tcarmelveilleux, line 662: The null value will cause the DefaultMoveRate to be used, which is writable and could also be zero. 

Recommend instead to set the rate to 1 instead of zero.

This is because:

>1.6.7.2.2. Rate Field
>
> This field SHALL indicate the rate of movement in units per second. The actual rate of movement **SHOULD be as close to this rate as the device is able**. If the Rate field is equal to null, then the value in DefaultMoveRate attribute SHALL be used. However, if the Rate field is equal to null and the DefaultMoveRate attribute is not supported, or if the Rate field is equal to null and the value of the DefaultMoveRate attribute is equal to null, then the device SHOULD move as fast as it is able. If the device is not able to move at a variable rate, this field MAY be disregarded.

tcarmelveilleux:
Closing as duplicate of https://github.com/project-chip/connectedhomeip/pull/32539

PR #32543, opened 2024-03-12
[nrfconnect] Switch to NCS 2.6.0 revision

ArekBalysNordic:
Regular update of nrfconnect SDK. 

This PR updates nrfconnect SDK to version 2.6.0.

Changelog:
- Enabled PSA Crypto API by default in nrfconnect examples.
- Updated Zephyr for nrfconnect SDK to 3.5.99.
- Aligned random include
- Aligned the nrf_power calls scheme
- Adapted to new zcbor 0.8.1 API
- Renamed obsolete hci_rpmsg name to hci_ipc
- Aligned the random generator definitions
- Aligned all examples to the PSA Crypto API

Work in progress, waiting for the NCS 2.6.0 TAG


Damian-Nordic (commented):
Looking good not approving until Docker image for 2.6.0 is ready.

src/platform/Zephyr/BLEManagerImpl.cpp
```diff
@@ -39,8 +39,14 @@
 
 #include <zephyr/bluetooth/addr.h>
 #include <zephyr/bluetooth/gatt.h>
+
 #include <zephyr/bluetooth/hci.h>
+#if CHIP_DEVICE_LAYER_TARGET_NRFCONNECT
```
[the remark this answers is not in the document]
axelnxp: my change was merged, you can update your branch to resolve the conflicts

Damian-Nordic (approved):

examples/chef/nrfconnect/main.cpp
```diff
@@ -110,6 +121,9 @@ int main()
 
     // Start IM server
     static chip::CommonCaseDeviceServerInitParams initParams;
+#ifdef CONFIG_CHIP_CRYPTO_PSA
+    initParams.operationalKeystore = &sPSAOperationalKeystore;
+#endif
     (void) initParams.InitializeStaticResourcesBeforeServerInit();
     err = chip::Server::GetInstance().Init(initParams);
     if (err != CHIP_NO_ERROR)
```
tehampson, line 129: Does this file not need the block inside `CONFIG_CHIP_MIGRATE_OPERATIONAL_KEYS_TO_ITS`?

ArekBalysNordic: The migration feature has been dedicated to in-field devices. We needed a mechanism to move all existing mbedTLS keys to PSA Crypto storage for customers who want to update their in-field devices. I don't think that the chef app is an example that is used in any in-field devices and it was created for testing purposes, so I guess migration is unnecessary here hence I didn't add it here.

tehampson: Sounds like it is not needed

tehampson (approved):

examples/chef/nrfconnect/main.cpp
tehampson: @erwinpan1 do you have an opinion?

It sounds to me like it should not be needed

ArekBalysNordic: @erwinpan1 config Crypto PSA is enabled by default for all samples, whereas the migration is optional and should be enableed only while updating the firmware for the device that runs mbedTLS. If both configs are enabled and the device does not have any mebdtls keys in memory then the migration method returns true. When the migration is enabled the method must iterate through saved fabrics and check whether any key needs to be migrated, so it should be disabled for new devices that originally runs PSA. Therefore I decided to not include it to the chef app, because I didn't see the case for migration there and leave it as simple as possible. But as I mentioned I can add it as follow-up if you need it there :)

PR #32546, opened 2024-03-12
[Silabs] Typing Event, state and wfx_rsi

lpbeliveau-silabs:
Added event enum for events and states, added a typedef for the wfx_rsi struct and changed name so the variable is not named like the type

### DIFF
--- a/examples/platform/silabs/SiWx917/SiWx917/wfx_rsi.h
+++ b/examples/platform/silabs/SiWx917/SiWx917/wfx_rsi.h
@@ -47,18 +47,18 @@ typedef enum
 
 typedef enum
 {
-    WFX_RSI_ST_DEV_READY       = 0x01,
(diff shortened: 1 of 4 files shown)


examples/platform/silabs/SiWx917/SiWx917/wfx_rsi.h
```diff
@@ -77,9 +84,9 @@ struct wfx_rsi
     sl_wfx_mac_address_t ap_bssid; /* To which our STA is connected */
     uint16_t join_retries;
     uint8_t ip4_addr[4]; /* Not sure if this is enough */
-};
+} wfx_rsi_t;
```
jepenven-silabs, line 87: Doesn't match with the previous CamelCase declaration... shouldn't it be WfxRsi_t ?

jepenven-silabs (approved):

tehampson (approved):

PR #32549, opened 2024-03-12
[Darwin] MTRDeviceController should clear attribute cache on pairing start

jtung-apple:
Changes in this PR:
* On pairing start, MTRDeviceController will clear information about the new node ID - both the in-memory MTRDevice and persisted attribute storage.
* Additional logging for attribute storage
* Bug fix for attribute storage `_pruneEmptyStoredAttributesBranches` method

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDeviceController.mm
+++ b/src/darwin/Framework/CHIP/MTRDeviceController.mm
@@ -44,6 +44,7 @@
 #import "MTRPersistentStorageDelegateBridge.h"
 #import "MTRServerEndpoint_Internal.h"
 #import "MTRSetupPayload.h"
+#import "MTRUnfairLock.h"
 #import "NSDataSpanConversion.h"
 #import "NSStringSpanConversion.h"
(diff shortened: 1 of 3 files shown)


src/darwin/Framework/CHIP/MTRDeviceController.mm
```diff
@@ -607,6 +608,13 @@ - (BOOL)setupCommissioningSessionWithPayload:(MTRSetupPayload *)payload
                                        error:(NSError * __autoreleasing *)error
 {
     auto block = ^BOOL {
+        // First reset previous information about this new nodeID
+        os_unfair_lock_lock(&_deviceMapLock);
+        [self _removeDeviceWithNodeID:newNodeID device:nil];
```
bzbarsky-apple, line 613: The problem is that the "node ID" in this API is not an actual operational node ID.  It does not have to match the operational node ID that the commissionee will get.  It's really a "commissioning process handle" or some such...

Some API consumers might have this match actual operational node IDs.  Others will not.  The latter will be broken by these changes, as far as I can tell.

PR #32550, opened 2024-03-12
Log node ID in exchanges

leonardmgh:
Previously, the SDK only recorded the exchange ID when logging information about exchanges, as illustrated by the following example:
`Failed to Send CHIP MessageCounter:68460922 on exchange 30872i sendCount: 4 max retries: 4` 
To enable more precise troubleshooting, this PR proposes to also log the peer node ID. The proposed log format change is as follows:
` Dropping unexpected message of type 0x1 with protocolId (0, 1) and MessageCounter:43669734 on exchange 11672r to: [1:0000000012344321]`
I have used the same format as here: 
https://github.com/project-chip/connectedhomeip/blob/ad14dbc9b5b1c8230bd29be4873f9367143bfc31/src/app/OperationalSessionSetup.cpp#L53

Further Context:
The HomeAssistant Matter server frequently logs the mentioned message when adding Nanoleaf Matter devices into the thread network. This situation has sparked debates regarding the source of the error – whether it is exclusive to Nanoleaf devices or  all devices relaying traffic through them, leading to timeout issues. This enhancement aims to provide clearer insights into such occurrences.

### DIFF
--- a/src/lib/support/logging/TextOnlyLogging.h
+++ b/src/lib/support/logging/TextOnlyLogging.h
@@ -268,9 +268,8 @@ using LogRedirectCallback_t = void (*)(const char * module, uint8_t category, co
 #define ChipLogValueMEI(aValue) static_cast<uint16_t>(aValue >> 16), static_cast<uint16_t>(aValue)
 
 /**
- * Logging helpers for exchanges.  For now just log the exchange id and whether
- * it's an initiator or responder, but eventually we may want to log the peer
- * node id as well (especially for the responder case).  Some callsites only
+ * Logging helpers for exchanges.  Log the exchange id, whether
+ * it's an initiator or responder and the peer node id.  Some callsites only
  * have the exchange id and initiator/responder boolean, not an actual exchange,
  * so we want to have a helper for that case too.
  */
--- a/src/lib/support/logging/TextOnlyLogging.h
+++ b/src/lib/support/logging/TextOnlyLogging.h
@@ -276,8 +276,14 @@ using LogRedirectCallback_t = void (*)(const char * module, uint8_t category, co
  */
 #define ChipLogFormatExchangeId "%u%c"
 #define ChipLogValueExchangeId(id, isInitiator) id, ((isInitiator) ? 'i' : 'r')
-#define ChipLogFormatExchange ChipLogFormatExchangeId
-#define ChipLogValueExchange(ec) ChipLogValueExchangeId((ec)->GetExchangeId(), (ec)->IsInitiator())
+#define ChipLogFormateExchangeNode "[%u:" ChipLogFormatX64 "]"
+#define ChipLogValueExchangeNode(fabric, id) fabric, ChipLogValueX64(id)
+#define ChipLogFormatExchange ChipLogFormatExchangeId " to: " ChipLogFormateExchangeNode
+#define ChipLogValueExchange(ec) ChipLogValueExchangeId((ec)->GetExchangeId(), (ec)->IsInitiator()),                               \
+    ChipLogValueExchangeNode(                                                                                                      \
+        ((ec)->HasSessionHandle() ? (ec)->GetSessionHandle()->GetSubjectDescriptor().fabricIndex : kUndefinedFabricIndex),         \
+        ((ec)->HasSessionHandle() ? (ec)->GetSessionHandle()->GetSubjectDescriptor().subject : kUndefinedNodeId))   
+
 #define ChipLogValueExchangeIdFromSentHeader(payloadHeader)                                                                        \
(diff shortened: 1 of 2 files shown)


src/lib/support/logging/TextOnlyLogging.h
```diff
@@ -268,16 +268,21 @@ using LogRedirectCallback_t = void (*)(const char * module, uint8_t category, co
...
+ * Logging helpers for exchanges.  Log the exchange id, whether
+ * it's an initiator or responder and the peer node id.  Some callsites only
  * have the exchange id and initiator/responder boolean, not an actual exchange,
  * so we want to have a helper for that case too.
  */
 #define ChipLogFormatExchangeId "%u%c"
 #define ChipLogValueExchangeId(id, isInitiator) id, ((isInitiator) ? 'i' : 'r')
-#define ChipLogFormatExchange ChipLogFormatExchangeId
-#define ChipLogValueExchange(ec) ChipLogValueExchangeId((ec)->GetExchangeId(), (ec)->IsInitiator())
+#define ChipLogFormateExchangeNode "[%u:" ChipLogFormatX64 "]"
```
bzbarsky-apple, line 278:
```suggestion
#define ChipLogFormatExchangeNode "[%u:" ChipLogFormatX64 "]"
```

src/lib/support/logging/TextOnlyLogging.h
```diff
@@ -268,16 +268,21 @@ using LogRedirectCallback_t = void (*)(const char * module, uint8_t category, co
...
+#define ChipLogValueExchangeNode(fabric, id) fabric, ChipLogValueX64(id)
+#define ChipLogFormatExchange ChipLogFormatExchangeId " to: " ChipLogFormateExchangeNode
+#define ChipLogValueExchange(ec) ChipLogValueExchangeId((ec)->GetExchangeId(), (ec)->IsInitiator()),                               \
+    ChipLogValueExchangeNode(                                                                                                      \
+        ((ec)->HasSessionHandle() ? (ec)->GetSessionHandle()->GetSubjectDescriptor().fabricIndex : kUndefinedFabricIndex),         \
+        ((ec)->HasSessionHandle() ? (ec)->GetSessionHandle()->GetSubjectDescriptor().subject : kUndefinedNodeId))
```
bzbarsky-apple, line 284: So this will have the following behavior:

1. For CASE sessions, what you would expect.
2. For incoming group messages, this will list the group id (as opposed to the source node id from the message).
3. For outgoing group messages, this will list undefined node id.
4. For PASE sessions as initiator, this will list undefined node id.
5. For PASE sessions as responder, this will list the made-up peer node id.
6. For unauthenticated sessions, this will list undefined node id.

The real question is this: is this better than doing `(ec)->GetSessionHandle()->GetPeer()` and logging it with `ChipLogFormatScopedNodeId`?  The latter is probably also smaller code.

Damian-Nordic:
This has significant impact on code size while usually the mapping between the exchange ID and the node ID can be found based on other messages. Could you make this feature optional so that we can disable it on platforms that prefer more concise logs for code size or performance reasons (printing logs to UART takes time and may affect the system performance)?

src/lib/support/logging/TextOnlyLogging.h
leonardmgh: It looks like the `ScopedNodeId` is used in most other places where the node id is logged, so i have changed my implementation.

leonardmgh:
> This has significant impact on code size while usually the mapping between the exchange ID and the node ID can be found based on other messages. Could you make this feature optional so that we can disable it on platforms that prefer more concise logs for code size or performance reasons (printing logs to UART takes time and may affect the system performance)?

I agree that the increase in code size is significant, which was unexpected. I could make this feature optional. However, perhaps I should entirely revise my approach to explicitly log the node ID every time `ChipLogError` utilizes `ChipLogValueExchange`. As you already mentioned, the node ID can be determined from other messages when the log level is below error.

andy31415:
@leonardmgh this requires restyle changes - please run `./scripts/helpers/restyle-diff.sh` or apply the patch that `restyled` proposes. Without that, reviewers are not added to the PR it seems.

leonardmgh:
> It's a little surprising that the codesize impact is that large, and concerning: How much codesize are we paying for the existing exchange logging?
> 
> The other thing that we could do is to only add the ScopedNodeId logging to specific logging locations, instead of making it part of the exchange logging in all cases....

For now i have made the changes optional, which is probably the easiest solution for now.

Otherwise adding the ScopedNodeId logging in specific locations could be a good alternative. Maybe the logging of the node id could replace the logging of the exchange context in some error messages, since i guess that it might be more useful to know node ID.

bzbarsky-apple (approved):

Damian-Nordic (approved):

PR #32556, opened 2024-03-13
Add a way for a client to know if the device cache is primed in Darwin

woody-apple:
Right now clients are called back when the cache is primed, but they have no way to know if it already was.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -28,6 +28,7 @@
 #import "MTRCommandTimedCheck.h"
 #import "MTRConversion.h"
 #import "MTRDefines_Internal.h"
+#import "MTRUnfairLock.h"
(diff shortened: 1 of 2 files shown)


src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -2017,6 +2017,19 @@ - (void)setAttributeValues:(NSArray<NSDictionary *> *)attributeValues reportChan
...
+{
+    BOOL isPrimed = NO;
+
+    os_unfair_lock_lock(&self->_lock);
+
+    isPrimed = [self _isCachePrimedWithInitialConfigurationData];
+
+    os_unfair_lock_unlock(&self->_lock);
+
+    return isPrimed;
```
anush-apple, line 2030:
```suggestion
    std::lock_guard lock(_lock);
    return [self _isCachePrimedWithInitialConfigurationData];
```

carol-apple (approved):

PR #32565, opened 2024-03-14
Darwin: Clean up includes/imports

ksperling-apple:
Import public headers via #import <Matter/Foo.h>
Import project headers via #import "Foo.h"
Include C++ SDK headers via #include <foo/bar.h>
Import counterpart header first in implementation file Import public header first in _Internal headers
Don't use #pragma once in ObjC(++)

Make MTRDeviceControllerLocalTestStorage.h a project header.

### DIFF
--- a/examples/darwin-framework-tool/commands/common/MTRError.mm
+++ b/examples/darwin-framework-tool/commands/common/MTRError.mm
@@ -19,9 +19,9 @@
 #import "MTRLogging.h"
 #import <Matter/Matter.h>
 
-#import <app/MessageDef/StatusIB.h>
-#import <inet/InetError.h>
-#import <lib/support/TypeTraits.h>
+#include <app/MessageDef/StatusIB.h>
+#include <inet/InetError.h>
+#include <lib/support/TypeTraits.h>
 
 #import <objc/runtime.h>
 
--- a/src/darwin/Framework/CHIP/MTRAttributeSpecifiedCheck.h
+++ b/src/darwin/Framework/CHIP/MTRAttributeSpecifiedCheck.h
@@ -16,10 +16,6 @@
  *    limitations under the License.
  */
 
-#pragma once
-
-#import <Foundation/Foundation.h>
-
 #include <lib/core/DataModelTypes.h>
 
 BOOL MTRAttributeIsSpecified(chip::ClusterId aClusterId, chip::AttributeId aAttributeId);
--- a/src/darwin/Framework/CHIP/MTRBaseDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRBaseDevice.mm
@@ -14,16 +14,18 @@
  *    See the License for the specific language governing permissions and
  *    limitations under the License.
  */
+
+#import "MTRBaseDevice_Internal.h"
+
+#import <Matter/MTRCluster.h>
 #import <Matter/MTRClusterConstants.h>
-#import <Matter/MTRDefines.h>
 
 #import "MTRAttributeTLVValueDecoder_Internal.h"
-#import "MTRBaseDevice_Internal.h"
 #import "MTRBaseSubscriptionCallback.h"
 #import "MTRCallbackBridgeBase.h"
-#import "MTRCluster.h"
 #import "MTRClusterStateCacheContainer_Internal.h"
 #import "MTRCluster_Internal.h"
+#import "MTRCommandPayloads_Internal.h"
 #import "MTRDevice_Internal.h"
 #import "MTRError_Internal.h"
 #import "MTREventTLVValueDecoder_Internal.h"
@@ -32,24 +34,22 @@
 #import "MTRSetupPayload_Internal.h"
 #import "NSDataSpanConversion.h"
 #import "NSStringSpanConversion.h"
-#import "zap-generated/MTRCommandPayloads_Internal.h"
-
-#include "app/ConcreteAttributePath.h"
-#include "app/ConcreteCommandPath.h"
-#include "app/ConcreteEventPath.h"
-#include "app/StatusResponse.h"
-#include "lib/core/CHIPError.h"
-#include "lib/core/DataModelTypes.h"
 
 #include <app/AttributePathParams.h>
 #include <app/BufferedReadCallback.h>
 #include <app/ClusterStateCache.h>
+#include <app/ConcreteAttributePath.h>
+#include <app/ConcreteCommandPath.h>
+#include <app/ConcreteEventPath.h>
 #include <app/InteractionModelEngine.h>
 #include <app/ReadClient.h>
+#include <app/StatusResponse.h>
 #include <controller/CommissioningWindowOpener.h>
 #include <controller/ReadInteraction.h>
 #include <controller/WriteInteraction.h>
 #include <crypto/CHIPCryptoPAL.h>
+#include <lib/core/CHIPError.h>
+#include <lib/core/DataModelTypes.h>
 #include <setup_payload/SetupPayload.h>
 #include <system/SystemClock.h>
 
--- a/src/darwin/Framework/CHIP/MTRCertificateInfo.mm
+++ b/src/darwin/Framework/CHIP/MTRCertificateInfo.mm
@@ -1,4 +1,5 @@
 /**
+ *
  *    Copyright (c) 2023 Project CHIP Authors
  *
  *    Licensed under the Apache License, Version 2.0 (the "License");
@@ -14,7 +15,7 @@
  *    limitations under the License.
  */
 
-#import "MTRCertificateInfo.h"
+#import <Matter/MTRCertificateInfo.h>
 
 #import "MTRConversion.h"
 #import "MTRDefines_Internal.h"
--- a/src/darwin/Framework/CHIP/MTRCertificates.mm
+++ b/src/darwin/Framework/CHIP/MTRCertificates.mm
@@ -14,7 +14,8 @@
  *    limitations under the License.
  */
 
-#import "MTRCertificates.h"
+#import <Matter/MTRCertificates.h>
+
 #import "MTRError_Internal.h"
 #import "MTRFramework.h"
 #import "MTRLogging_Internal.h"
--- a/src/darwin/Framework/CHIP/MTRCluster.mm
+++ b/src/darwin/Framework/CHIP/MTRCluster.mm
@@ -15,8 +15,10 @@
  *    limitations under the License.
  */
 
-#import "MTRBaseDevice.h"
 #import "MTRCluster_Internal.h"
+
+#import <Matter/MTRBaseDevice.h>
+
 #import "NSDataSpanConversion.h"
 #import "NSStringSpanConversion.h"
 
--- a/src/darwin/Framework/CHIP/MTRClusterStateCacheContainer+XPC.h
+++ b/src/darwin/Framework/CHIP/MTRClusterStateCacheContainer+XPC.h
@@ -16,8 +16,7 @@
  */
 
 #import <Foundation/Foundation.h>
-
-#import "MTRClusterStateCacheContainer.h"
+#import <Matter/MTRClusterStateCacheContainer.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
--- a/src/darwin/Framework/CHIP/MTRDeviceController.h
+++ b/src/darwin/Framework/CHIP/MTRDeviceController.h
@@ -15,8 +15,6 @@
  *    limitations under the License.
  */
 
-#import <Foundation/Foundation.h>
-
 #import <Matter/MTRCommissionableBrowserDelegate.h>
 #import <Matter/MTRDefines.h>
 #import <Matter/MTROperationalCertificateIssuer.h>
--- a/src/darwin/Framework/CHIP/MTRDeviceControllerLocalTestStorage.h
+++ b/src/darwin/Framework/CHIP/MTRDeviceControllerLocalTestStorage.h
@@ -1,4 +1,3 @@
-//
 /**
  *    Copyright (c) 2023 Project CHIP Authors
  *
@@ -15,14 +14,16 @@
  *    limitations under the License.
  */
 
-#import <Foundation/Foundation.h>
-#import <Matter/Matter.h>
+#import <Matter/MTRDeviceControllerStorageDelegate.h>
 
-#if MTR_PER_CONTROLLER_STORAGE_ENABLED
+#import "MTRDefines_Internal.h"
 
 NS_ASSUME_NONNULL_BEGIN
 
-MTR_EXTERN @interface MTRDeviceControllerLocalTestStorage : NSObject<MTRDeviceControllerStorageDelegate>
+#if MTR_PER_CONTROLLER_STORAGE_ENABLED
+
+MTR_TESTABLE
+@interface MTRDeviceControllerLocalTestStorage : NSObject <MTRDeviceControllerStorageDelegate>
 
 // Setting this variable only affects subsequent MTRDeviceController initializations
 @property (class, nonatomic, assign) BOOL localTestStorageEnabled;
@@ -32,6 +33,6 @@ MTR_EXTERN @interface MTRDeviceControllerLocalTestStorage : NSObject<MTRDeviceCo
 
 @end
 
-NS_ASSUME_NONNULL_END
-
 #endif // MTR_PER_CONTROLLER_STORAGE_ENABLED
+
+NS_ASSUME_NONNULL_END
--- a/src/darwin/Framework/CHIP/MTRDeviceControllerStartupParams.h
+++ b/src/darwin/Framework/CHIP/MTRDeviceControllerStartupParams.h
@@ -14,8 +14,6 @@
  *    limitations under the License.
  */
 
-#import <Foundation/Foundation.h>
-
 #import <Matter/MTRCertificates.h>
 #import <Matter/MTRDefines.h>
 #import <Matter/MTROperationalCertificateIssuer.h>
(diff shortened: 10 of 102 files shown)


examples/darwin-framework-tool/commands/common/MTRError.mm
```diff
@@ -19,9 +19,9 @@
 #import "MTRLogging.h"
 #import <Matter/Matter.h>
 
-#import <app/MessageDef/StatusIB.h>
-#import <inet/InetError.h>
-#import <lib/support/TypeTraits.h>
+#include <app/MessageDef/StatusIB.h>
```
woody-apple, line 22: Why are these inlcudes? Imports are more appropriate

src/darwin/Framework/CHIP/MTRBaseDevice.mm
```diff
@@ -14,16 +14,18 @@
  *    See the License for the specific language governing permissions and
  *    limitations under the License.
  */
+
+#import "MTRBaseDevice_Internal.h"
+
+#import <Matter/MTRCluster.h>
```
woody-apple, line 20: These should be " not <

src/darwin/Framework/CHIP/MTRBaseDevice.mm
```diff
@@ -32,24 +34,22 @@
...
-#include "app/ConcreteCommandPath.h"
-#include "app/ConcreteEventPath.h"
-#include "app/StatusResponse.h"
-#include "lib/core/CHIPError.h"
-#include "lib/core/DataModelTypes.h"
 
 #include <app/AttributePathParams.h>
 #include <app/BufferedReadCallback.h>
 #include <app/ClusterStateCache.h>
+#include <app/ConcreteAttributePath.h>
```
woody-apple, line 41: Why aren't these all imports?

src/darwin/Framework/CHIP/MTRCertificateInfo.mm
```diff
@@ -14,7 +15,7 @@
  *    limitations under the License.
  */
 
-#import "MTRCertificateInfo.h"
+#import <Matter/MTRCertificateInfo.h>
```
woody-apple, line 18: This is not correct, it should be "

src/darwin/Framework/CHIP/MTRCertificates.mm
```diff
@@ -14,7 +14,8 @@
  *    limitations under the License.
  */
 
-#import "MTRCertificates.h"
+#import <Matter/MTRCertificates.h>
```
woody-apple, line 17: This is not correct, it should be "This is not correct, it should be "

src/darwin/Framework/CHIP/MTRCluster.mm
```diff
@@ -15,8 +15,10 @@
  *    limitations under the License.
  */
 
-#import "MTRBaseDevice.h"
 #import "MTRCluster_Internal.h"
+
+#import <Matter/MTRBaseDevice.h>
```
woody-apple, line 20: This is not correct, it should be "

src/darwin/Framework/CHIP/MTRClusterStateCacheContainer+XPC.h
```diff
@@ -16,8 +16,7 @@
  */
 
 #import <Foundation/Foundation.h>
-
-#import "MTRClusterStateCacheContainer.h"
+#import <Matter/MTRClusterStateCacheContainer.h>
```
woody-apple, line 19: This is not correct, it should be "

woody-apple (changes requested):

woody-apple (approved):
Internal imports are ", not <

I'm not sure why we're moving import -> include, ObjC++ or ObjC files always use import.

woody-apple (changes requested):
Marking CR

examples/darwin-framework-tool/commands/common/MTRError.mm
ksperling-apple: Almost all our files follow the convention that C++ headers are included with `#include`, and ObjC++ headers with `#import`

src/darwin/Framework/CHIP/MTRAttributeSpecifiedCheck.h
```diff
@@ -16,10 +16,6 @@
  *    limitations under the License.
  */
 
-#pragma once
```
bzbarsky-apple, line 19: So these are not needed because the expectation is people use #import which handles this?

src/darwin/Framework/CHIP/MTRBaseDevice.mm
```diff
@@ -14,16 +14,18 @@
  *    See the License for the specific language governing permissions and
  *    limitations under the License.
  */
+
+#import "MTRBaseDevice_Internal.h"
+
+#import <Matter/MTRCluster.h>
 #import <Matter/MTRClusterConstants.h>
-#import <Matter/MTRDefines.h>
```
bzbarsky-apple, line 18: Is the assumption that all the SDK headers include this one?

src/darwin/Framework/CHIP/MTRDeviceController.h
```diff
@@ -15,8 +15,6 @@
  *    limitations under the License.
  */
 
-#import <Foundation/Foundation.h>
```
bzbarsky-apple, line 18: Seems to me that if this file declares interfaces inheriting from NSObject, it needs to import Foundation.

src/darwin/Framework/CHIP/MTRBaseDevice.mm
ksperling-apple: One simple convention that many people like to follow is to always refer to public headers via the qualified import, since this is a required when that import is itself within a public header. It sounds like you prefer to have implementation files use unqualified `""` imports no matter if the imported header is public or internal?

src/darwin/Framework/CHIP/MTRDeviceControllerStartupParams.h
```diff
@@ -14,8 +14,6 @@
  *    limitations under the License.
  */
 
-#import <Foundation/Foundation.h>
```
bzbarsky-apple, line 17: Again, we are using NSObject here.  We should be importing the right header for that, not randomly getting it from something else we import.

src/darwin/Framework/CHIP/MTRDeviceController_Internal.h
```diff
@@ -14,26 +14,10 @@
...
-#import "MTRDeviceController.h"
-
-#include <lib/core/CHIPError.h>
-#include <lib/core/DataModelTypes.h>
-
-#import "MTRBaseDevice.h"
-#import "MTRDeviceController.h"
-#import "MTRDeviceControllerDataStore.h"
-
-#import <Matter/MTRDefines.h>
```
bzbarsky-apple, line 36: Why is it ok to not import MTRDefines here?  Because we know MTRDeviceController does?

I'm trying to understand the motivation for removing some of these imports of stuff the files actually use.

src/darwin/Framework/CHIP/MTRError.h
```diff
@@ -15,8 +15,6 @@
  *    limitations under the License.
  */
 
-#import <Foundation/Foundation.h>
```
bzbarsky-apple, line 18: Where is definition of NSErrorDomain coming from?

src/darwin/Framework/CHIP/MTRLogging_Internal.h
```diff
@@ -1,7 +1,6 @@
-/*
+/**
  *
  *    Copyright (c) 2020 Project CHIP Authors
- *    All rights reserved.
```
bzbarsky-apple, line 4: Why this license change?

src/darwin/Framework/CHIP/ServerEndpoint/MTRServerCluster.h
```diff
@@ -14,7 +14,6 @@
  *    limitations under the License.
  */
 
-#import <Foundation/Foundation.h>
```
bzbarsky-apple, line 17: Where is this file getting NSObject?

src/darwin/Framework/CHIP/ServerEndpoint/MTRServerCluster_Internal.h
```diff
@@ -14,17 +14,15 @@
...
-#import <Matter/MTRDeviceController.h>
 #import <Matter/MTRServerCluster.h>
 
 #include <lib/core/DataModelTypes.h>
-
 // TODO: These attribute-*.h and Span bits are a hack that should eventually go away.
 #include <app/util/attribute-metadata.h>
 #include <lib/support/Span.h>
 
+@class MTRDeviceController;
```
bzbarsky-apple, line 24: Why? Including the public header seems strictly better than forward-declaring, and we know public headers can't depend on private, so dep loops are not an issue.

src/darwin/Framework/CHIP/ServerEndpoint/MTRServerEndpoint.h
```diff
@@ -14,7 +14,6 @@
  *    limitations under the License.
  */
 
-#import <Foundation/Foundation.h>
```
bzbarsky-apple, line 17: I'll stop asking about this, but I think these Foundation imports need to stay in a lot of these places.

src/darwin/Framework/CHIPTests/TestHelpers/MTRTestResetCommissioneeHelper.h
```diff
@@ -14,10 +14,7 @@
  *    limitations under the License.
  */
 
-#import <Foundation/Foundation.h>
```
bzbarsky-apple, line 17: Where is this getting dispatch_queue_t?

src/darwin/Framework/Matter.xcodeproj/project.pbxproj
```diff
@@ -244,7 +244,7 @@
 		5ACDDD7E27CD3F3A00EFD68A /* MTRClusterStateCacheContainer_Internal.h in Headers */ = {isa = PBXBuildFile; fileRef = 5ACDDD7B27CD14AF00EFD68A /* MTRClusterStateCacheContainer_Internal.h */; };
 		5AE6D4E427A99041001F2493 /* MTRDeviceTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 5AE6D4E327A99041001F2493 /* MTRDeviceTests.m */; };
 		75139A6F2B7FE5E900E3A919 /* MTRDeviceControllerLocalTestStorage.mm in Sources */ = {isa = PBXBuildFile; fileRef = 75139A6E2B7FE5E900E3A919 /* MTRDeviceControllerLocalTestStorage.mm */; };
-		75139A702B7FE68C00E3A919 /* MTRDeviceControllerLocalTestStorage.h in Headers */ = {isa = PBXBuildFile; fileRef = 75139A6D2B7FE5D600E3A919 /* MTRDeviceControllerLocalTestStorage.h */; settings = {ATTRIBUTES = (Private, ); }; };
```
bzbarsky-apple, line 247: This was Private for a reason, is my understanding.  Check with @jtung-apple .

src/darwin/Framework/CHIP/MTRDeviceControllerLocalTestStorage.h
```diff
@@ -15,14 +14,16 @@
...
 
-#if MTR_PER_CONTROLLER_STORAGE_ENABLED
+#import "MTRDefines_Internal.h"
 
 NS_ASSUME_NONNULL_BEGIN
 
-MTR_EXTERN @interface MTRDeviceControllerLocalTestStorage : NSObject<MTRDeviceControllerStorageDelegate>
+#if MTR_PER_CONTROLLER_STORAGE_ENABLED
+
+MTR_TESTABLE
```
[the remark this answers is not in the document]
ksperling-apple: I did check with him. It's available for testing and is enabled via a user default (not by anything importing this header). It was (incorrectly) marked `MTR_EXTERN` to get the test to see it.

src/darwin/Framework/CHIP/MTRAttributeSpecifiedCheck.h
ksperling-apple: Yes

src/darwin/Framework/CHIP/MTRBaseDevice.mm
ksperling-apple: No, simply because there's no need for `Foo.mm` to repeat imports from its counterparts (`Foo.h` and `Foo_Internal.h`). Similarly, `Foo_Internal.h` should always import `Foo.h` as it's first line, so no repeating of anything from `Foo.h` needed there either.

src/darwin/Framework/CHIP/MTRDeviceController.h
ksperling-apple: It seems OK to me to rely on MTRDefines importing Foundation?

src/darwin/Framework/CHIP/MTRDeviceController_Internal.h
ksperling-apple: I think it makes sense to view `Foo.h` and `Foo_Internal.h` as a sort of "stack", `Foo_Internal.h` always imports `Foo.h` at the start, and then adds more things to it, so I don't think there's any need to repeat imports from the public header in the internal one.

src/darwin/Framework/CHIP/MTRLogging_Internal.h
ksperling-apple: This license header didn't match the ones from other files

src/darwin/Framework/CHIPTests/TestHelpers/MTRTestResetCommissioneeHelper.h
ksperling-apple: Not from importing `<dispatch/dispatch.h>` apparently ;-)

src/darwin/Framework/CHIP/MTRDeviceControllerStartupParams.h
ksperling-apple: We have dozens of headers that don't explicitly import Foundation, generally they get it via `MTRDefines.h`. And any header that exports anything will be using `MTR_EXPORT` or an availability macro from MTRDefines anyway, so to me that seems like an acceptable way to reduce boilerplate.

You seem fine with relying on Foundation to import import `dispatch.h` for us, how is this different?

src/darwin/Framework/CHIP/ServerEndpoint/MTRServerCluster_Internal.h
ksperling-apple: We seem to forward-declare in quite a few places, and generally it's a useful way to reduce incremental build times, but if we have some general rule to always import instead of forward declaring that's fine by me.

src/darwin/Framework/CHIP/MTRDeviceControllerLocalTestStorage.h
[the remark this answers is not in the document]
bzbarsky-apple: OK, great.

src/darwin/Framework/CHIP/MTRBaseDevice.mm
bzbarsky-apple: Ah, so the key is that MTRBaseDevice.h imports MTRDefines?

src/darwin/Framework/CHIP/MTRDeviceController.h
bzbarsky-apple: OK.

src/darwin/Framework/CHIP/MTRDeviceControllerStartupParams.h
bzbarsky-apple: > You seem fine with relying on Foundation to import dispatch.h for us, how is this different?

I did not realize dispatch.h is a separate thing.  ;)

But more to the point, I assume that Foundation importing dispatch.h is part of its API contract, right?  We don't really have a clear API contract for what MTRDefines.h imports.

If we're OK saying it always imports Foundation and we depend on that, I'm OK with that.

src/darwin/Framework/CHIP/MTRDeviceController_Internal.h
bzbarsky-apple: Sounds good.

src/darwin/Framework/CHIP/MTRLogging_Internal.h
bzbarsky-apple: That really worries me, actually...  Probably best to make this change in a separate PR where we can figure out what the right text is.

src/darwin/Framework/CHIP/ServerEndpoint/MTRServerCluster_Internal.h
bzbarsky-apple: I would say we should import instead of forward-declaring except in cases where dependency loops would make that impossible.

src/darwin/Framework/CHIPTests/TestHelpers/MTRTestResetCommissioneeHelper.h
bzbarsky-apple: Like I said above, I did not realize that dispatch.h was a separate thing.  ;)

That said, sounds like the answer is "it's getting it via Matter.h, which includes MTRDefines.h, which includes Foundation.h".  And I guess having part of the public API contract for Matter.h be "includes Foundation.h" is reasonable, since it's defining ObjC types...

PR #32573, opened 2024-03-14
Remove redundant checks from the operational state impl in all-clusters-app.

bzbarsky-apple:
The cluster implementation does those checks.

### DIFF
--- a/examples/all-clusters-app/all-clusters-common/src/operational-state-delegate-impl.cpp
+++ b/examples/all-clusters-app/all-clusters-common/src/operational-state-delegate-impl.cpp
@@ -55,15 +55,6 @@ CHIP_ERROR GenericOperationalStateDelegateImpl::GetOperationalPhaseAtIndex(size_
 
 void GenericOperationalStateDelegateImpl::HandlePauseStateCallback(GenericOperationalError & err)
 {
-    OperationalState::OperationalStateEnum state =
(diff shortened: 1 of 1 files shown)


examples/all-clusters-app/all-clusters-common/src/operational-state-delegate-impl.cpp
```diff
@@ -78,15 +69,6 @@ void GenericOperationalStateDelegateImpl::HandlePauseStateCallback(GenericOperat
 
 void GenericOperationalStateDelegateImpl::HandleResumeStateCallback(GenericOperationalError & err)
 {
-    OperationalState::OperationalStateEnum state =
-        static_cast<OperationalState::OperationalStateEnum>(GetInstance()->GetCurrentOperationalState());
-
-    if (state == OperationalState::OperationalStateEnum::kStopped || state == OperationalState::OperationalStateEnum::kError)
-    {
-        err.Set(to_underlying(OperationalState::ErrorStateEnum::kCommandInvalidInState));
-        return;
```
tehampson, line 87: Isn't this early return still needed? That or we would need the cluster version to jump to exit [over here](https://github.com/project-chip/connectedhomeip/blob/master/src/app/clusters/operational-state-server/operational-state-server.cpp#L426-L430)

tehampson (approved):

arkq (approved):

woody-apple (approved):

PR #32582, opened 2024-03-14
[Darwin] Adding metrics for MTRDeviceController APIs

anush-apple:
- Adding initial set of metrics for MTRDeviceController APIs around device commissioning
- Added uniqueIdentifer to MTRMetrics
- Changed MetricsCollector to removed the end event for begin / instant event

### DIFF
--- a/src/darwin/Framework/CHIP/MTRBaseDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRBaseDevice.mm
@@ -62,6 +62,7 @@ using namespace chip::Protocols::InteractionModel;
 using chip::Optional;
 using chip::SessionHandle;
 using chip::Messaging::ExchangeManager;
+using namespace chip::Metrics::DarwinFramework;
 
 NSString * const MTRAttributePathKey = @"attributePath";
 NSString * const MTRCommandPathKey = @"commandPath";
@@ -1862,12 +1863,12 @@ MTREventPriority MTREventPriorityForValidPriorityLevel(chip::app::PriorityLevel 
                                             queue:(dispatch_queue_t)queue
                                        completion:(MTRDeviceOpenCommissioningWindowHandler)completion
 {
-    MATTER_LOG_METRIC_BEGIN(kMetricDarwinFrameworkOpenPairingWindow);
+    MATTER_LOG_METRIC_BEGIN(kMetricOpenPairingWindow);
 
     if (self.isPASEDevice) {
         MTR_LOG_ERROR("Can't open a commissioning window over PASE");
         dispatch_async(queue, ^{
-            MATTER_LOG_METRIC_END(kMetricDarwinFrameworkOpenPairingWindow, CHIP_ERROR_INCORRECT_STATE);
+            MATTER_LOG_METRIC_END(kMetricOpenPairingWindow, CHIP_ERROR_INCORRECT_STATE);
             completion(nil, [MTRError errorForCHIPErrorCode:CHIP_ERROR_INCORRECT_STATE]);
         });
         return;
@@ -1877,7 +1878,7 @@ MTREventPriority MTREventPriorityForValidPriorityLevel(chip::app::PriorityLevel 
     if (!CanCastTo<uint16_t>(durationVal)) {
         MTR_LOG_ERROR("Error: Duration %llu is too large.", durationVal);
         dispatch_async(queue, ^{
-            MATTER_LOG_METRIC_END(kMetricDarwinFrameworkOpenPairingWindow, CHIP_ERROR_INVALID_INTEGER_VALUE);
+            MATTER_LOG_METRIC_END(kMetricOpenPairingWindow, CHIP_ERROR_INVALID_INTEGER_VALUE);
             completion(nil, [MTRError errorForCHIPErrorCode:CHIP_ERROR_INVALID_INTEGER_VALUE]);
         });
         return;
@@ -1888,7 +1889,7 @@ MTREventPriority MTREventPriorityForValidPriorityLevel(chip::app::PriorityLevel 
     if (discriminatorVal > 0xFFF) {
         MTR_LOG_ERROR("Error: Discriminator %llu is too large. Max value %d", discriminatorVal, 0xFFF);
         dispatch_async(queue, ^{
-            MATTER_LOG_METRIC_END(kMetricDarwinFrameworkOpenPairingWindow, CHIP_ERROR_INVALID_INTEGER_VALUE);
+            MATTER_LOG_METRIC_END(kMetricOpenPairingWindow, CHIP_ERROR_INVALID_INTEGER_VALUE);
             completion(nil, [MTRError errorForCHIPErrorCode:CHIP_ERROR_INVALID_INTEGER_VALUE]);
         });
         return;
@@ -1900,7 +1901,7 @@ MTREventPriority MTREventPriorityForValidPriorityLevel(chip::app::PriorityLevel 
         if (!CanCastTo<uint32_t>(passcodeVal) || !SetupPayload::IsValidSetupPIN(static_cast<uint32_t>(passcodeVal))) {
             MTR_LOG_ERROR("Error: Setup passcode %llu is not valid", passcodeVal);
             dispatch_async(queue, ^{
-                MATTER_LOG_METRIC_END(kMetricDarwinFrameworkOpenPairingWindow, CHIP_ERROR_INVALID_INTEGER_VALUE);
+                MATTER_LOG_METRIC_END(kMetricOpenPairingWindow, CHIP_ERROR_INVALID_INTEGER_VALUE);
                 completion(nil, [MTRError errorForCHIPErrorCode:CHIP_ERROR_INVALID_INTEGER_VALUE]);
             });
             return;
@@ -1913,7 +1914,7 @@ MTREventPriority MTREventPriorityForValidPriorityLevel(chip::app::PriorityLevel 
             auto resultCallback = ^(CHIP_ERROR status, const SetupPayload & payload) {
                 if (status != CHIP_NO_ERROR) {
                     dispatch_async(queue, ^{
-                        MATTER_LOG_METRIC_END(kMetricDarwinFrameworkOpenPairingWindow, status);
+                        MATTER_LOG_METRIC_END(kMetricOpenPairingWindow, status);
                         completion(nil, [MTRError errorForCHIPErrorCode:status]);
                     });
                     return;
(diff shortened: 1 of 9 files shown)


src/darwin/Framework/CHIP/MTRDeviceController.mm
```diff
@@ -619,23 +637,38 @@ - (BOOL)setupCommissioningSessionWithPayload:(MTRSetupPayload *)payload
...
         self->_operationalCredentialsDelegate->SetDeviceID(nodeId);
-        auto errorCode = self->_cppCommissioner->EstablishPASEConnection(nodeId, [pairingCode UTF8String]);
+        errorCode = self->_cppCommissioner->EstablishPASEConnection(nodeId, [pairingCode UTF8String]);
         return ![MTRDeviceController checkForError:errorCode logMsg:kErrorPairDevice error:error];
     };
 
-    return [self syncRunOnWorkQueueWithBoolReturnValue:block error:error];
+    auto success = [self syncRunOnWorkQueueWithBoolReturnValue:block error:error];
+    if (!success) {
+        MATTER_LOG_METRIC_END(kMetricDarwinFrameworkDeviceCommissioning, errorCode);
```
ksperling-apple, line 646: Why don't we log the metric on success as well?

src/darwin/Framework/CHIP/MTRDeviceController.mm
```diff
@@ -602,11 +603,28 @@ - (NSNumber *)controllerNodeID
...
+
 - (BOOL)setupCommissioningSessionWithPayload:(MTRSetupPayload *)payload
                                    newNodeID:(NSNumber *)newNodeID
                                        error:(NSError * __autoreleasing *)error
 {
+    // Track overall commissioning
+    MATTER_LOG_METRIC_BEGIN(kMetricDarwinFrameworkDeviceCommissioning);
+    emitMetricForSetupPayload(payload);
+
+    // Capture in a block variable to avoid losing granularity for metrics,
```
ksperling-apple, line 620: This is a pretty clunky pattern... you could just submit the end metric from inside the block? There is not going to be a relevant difference in timing from submitting from the end of the block vs right after the dispatch_sync has returned.

src/darwin/Framework/CHIP/MTRDeviceController.mm
```diff
@@ -793,7 +834,11 @@ - (BOOL)continueCommissioningDevice:(void *)device
         auto deviceProxy = static_cast<chip::DeviceProxy *>(device);
         auto errorCode = self->_cppCommissioner->ContinueCommissioningAfterDeviceAttestation(deviceProxy,
             ignoreAttestationFailure ? chip::Credentials::AttestationVerificationResult::kSuccess : lastAttestationResult);
-        return ![MTRDeviceController checkForError:errorCode logMsg:kErrorPairDevice error:error];
+        auto failed = [MTRDeviceController checkForError:errorCode logMsg:kErrorPairDevice error:error];
```
ksperling-apple, line 837: I think the pattern you're using below somewhere of submitting the metric before the `checkForError:` is more readable

src/darwin/Framework/CHIP/MTRMetricKeys.h
```diff
@@ -0,0 +1,48 @@
...
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+#include <tracing/metric_event.h>
+
+constexpr chip::Tracing::MetricKey kMetricDarwinFrameworkDeviceCommissioning = "dwnfw_device_commissioning";
```
ksperling-apple, line 19: We could put these into a C++ namespace instead of using these super long names?
With code like

```
VerifyOrReturnValueWithMetric(kMetricDarwinFrameworkStartBrowseForCommissionables, ...)
```

it's starting to get hard to see the actual code.

src/darwin/Framework/CHIPTests/MTRMetricsTests.m
```diff
@@ -128,4 +128,13 @@ - (void)test005_TestValueRemoval
     XCTAssertTrue([keys count] == 0);
 }
 
+- (void)test006_TestUniqueIdentifier
+{
+    MTRMetrics * metrics1 = [[MTRMetrics alloc] initWithCapacity:1];
+    MTRMetrics * metrics2 = [[MTRMetrics alloc] initWithCapacity:1];
+    XCTAssertTrue(metrics1.uniqueIdentifier != nil);
+    XCTAssertTrue(metrics2.uniqueIdentifier != nil);
+    XCTAssertTrue([metrics1.uniqueIdentifier isEqual:metrics2.uniqueIdentifier] == NO);
```
ksperling-apple, line 137:
```suggestion
    XCTAssertNotNil(metrics1.uniqueIdentifier);
    XCTAssertNotNil(metrics2.uniqueIdentifier);
    XCTAssertNotEqualObjects(metrics1.uniqueIdentifier, metrics2.uniqueIdentifier);
```

src/darwin/Framework/CHIP/MTRDeviceController.mm
```diff
@@ -670,14 +703,22 @@ - (BOOL)setupCommissioningSessionWithDiscoveredDevice:(MTRCommissionableBrowserR
...
 }
 
 - (BOOL)commissionNodeWithID:(NSNumber *)nodeID
          commissioningParams:(MTRCommissioningParameters *)commissioningParams
                        error:(NSError * __autoreleasing *)error
 {
     auto block = ^BOOL {
+        // Capture metrics on overall status of this stage
+        CHIP_ERROR errorCode = CHIP_NO_ERROR;
+        MATTER_LOG_METRIC_SCOPE(kMetricDarwinFrameworkCommissionNode, errorCode);
```
ksperling-apple, line 720: Not sure it makes sense to capture this as begin / end since this only kicks off commissioning?

src/darwin/Framework/CHIP/MTRDeviceControllerDelegateBridge.mm
```diff
@@ -118,6 +119,7 @@
 void MTRDeviceControllerDelegateBridge::OnCommissioningComplete(chip::NodeId nodeId, CHIP_ERROR error)
 {
     MTR_LOG_DEFAULT("DeviceControllerDelegate Commissioning complete. NodeId %llu Status %s", nodeId, chip::ErrorStr(error));
+    MATTER_LOG_METRIC_END(kMetricDarwinFrameworkDeviceCommissioning, error);
```
anush-apple, line 122: Success or error is captured here

src/darwin/Framework/CHIP/MTRDeviceController.mm
anush-apple: I am trying to capture the error code form syncRunOnWorkQueueWithBoolReturnValue and end the metric since the actual end is in https://github.com/project-chip/connectedhomeip/pull/32582/files#r1525635022

anush-apple: Either way I will need to have the errorCode outside the block to capture the error here.

src/darwin/Framework/CHIP/MTRDeviceController.mm
anush-apple: It is here: https://github.com/project-chip/connectedhomeip/pull/32582/files#r1525635022

src/darwin/Framework/CHIP/MTRDeviceController.mm
anush-apple: Yup, will change

src/darwin/Framework/CHIP/MTRDeviceController.mm
anush-apple: Which place are you referring to?

src/darwin/Framework/CHIP/MTRMetricKeys.h
anush-apple: I could see what that looks like

src/darwin/Framework/CHIP/MTRDeviceController.mm
anush-apple: Found it

woody-apple (approved):

src/darwin/Framework/CHIP/MTRBaseDevice.mm
```diff
@@ -1861,9 +1863,12 @@ - (void)_openCommissioningWindowWithSetupPasscode:(nullable NSNumber *)setupPass
                                             queue:(dispatch_queue_t)queue
                                        completion:(MTRDeviceOpenCommissioningWindowHandler)completion
 {
+    MATTER_LOG_METRIC_BEGIN(kMetricOpenPairingWindow);
+
     if (self.isPASEDevice) {
         MTR_LOG_ERROR("Can't open a commissioning window over PASE");
         dispatch_async(queue, ^{
+            MATTER_LOG_METRIC_END(kMetricOpenPairingWindow, CHIP_ERROR_INCORRECT_STATE);
             completion(nil, [MTRError errorForCHIPErrorCode:CHIP_ERROR_INCORRECT_STATE]);
```
bzbarsky-apple, line 1872: Would have been less error, prone, I think, to define a block that takes a payload and a CHIP_ERROR and does the MATTER_LOG_METRIC_END and then the completion call.

Then call that block in all the places in this function where we call completion, instead of calling completion...

src/darwin/Framework/CHIP/MTRDeviceController.mm
```diff
@@ -1562,21 +1617,35 @@ - (BOOL)pairDevice:(uint64_t)deviceID
       setupPINCode:(uint32_t)setupPINCode
              error:(NSError * __autoreleasing *)error
 {
+    // Track overall commissioning
+    MATTER_LOG_METRIC_BEGIN(kMetricDeviceCommissioning);
```
bzbarsky-apple, line 1621: So we use the same metric for "start PASE handshake" and "start commissioning"?  That seems a bit confusing.

src/darwin/Framework/CHIP/MTRMetricKeys.h
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+#include <lib/core/CHIPError.h>
+#include <tracing/metric_event.h>
+
+namespace chip {
+namespace Tracing {
+namespace DarwinFramework {
+
+constexpr Tracing::MetricKey kMetricDeviceCommissioning = "dwnfw_device_commissioning";
```
bzbarsky-apple, line 24: These keys get exposed to API consumers, effectively, right?  Should there be something about "matter" in them?

src/darwin/Framework/CHIP/MTRDeviceController.mm
bzbarsky-apple: Also, I am not sure we are doing the corresponding METRIC_END on all the codepaths where things terminate with async failures (e.g. if commissioning discovery fails).

PR #32583, opened 2024-03-15
Parallelize Darwin CI build

ksperling-apple:
Split the monolithic `Darwin` build into separate jobs that can run in parallel.

Also fix a few issues surfaced by now running the watch and iOS builds with `-werror`.

### DIFF
--- a/.github/workflows/darwin.yaml
+++ b/.github/workflows/darwin.yaml
@@ -145,7 +145,7 @@ jobs:
               run: xcodebuild -target "MatterTvCastingBridge" -sdk iphoneos
 
     darwin:
-        name: Build Darwin
+        name: Build Darwin # Matches the previous monolithic build that's marked "required" for PRs
         needs: [ framework, tests ]
         runs-on: macos-latest
         steps:
--- a/.github/workflows/darwin.yaml
+++ b/.github/workflows/darwin.yaml
@@ -26,14 +26,25 @@ concurrency:
 
 env:
     CHIP_NO_LOG_TIMESTAMPS: true
-    
-jobs:
-    darwin:
-        name: Build Darwin
 
+jobs:
+    framework:
+        name: Build framework
         if: github.actor != 'restyled-io[bot]'
         runs-on: macos-latest
-
+        strategy:
+            matrix:
+                options: # We don't need a full matrix
+                    - flavor: macos-release-availability
+                      arguments: -sdk macosx -configuration Release   OTHER_CFLAGS='${inherited} -Werror -Wconversion -Wno-unguarded-availability-new'
+                    - flavor: ios-release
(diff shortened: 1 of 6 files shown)


cecille (approved):

jmartinez-silabs (approved):

woody-apple (approved):

.github/workflows/darwin.yaml
```diff
@@ -26,14 +26,25 @@ concurrency:
 
 env:
     CHIP_NO_LOG_TIMESTAMPS: true
-
```
woody-apple, line 29: Adding a comment to hold until I can update the build requirements

woody-apple (approved):

.github/workflows/darwin.yaml
```diff
@@ -103,40 +98,56 @@ jobs:
...
+              working-directory: examples/tv-casting-app/darwin/MatterTvCastingBridge
+              run: xcodebuild -target "MatterTvCastingBridge" -sdk iphoneos
+
+    darwin:
+        name: Build Darwin
+        needs: [ framework, tests ]
+        runs-on: macos-latest
+        steps:
+            - name: Done
+              run: 'true' # nothing to do
```
bzbarsky-apple, line 153: It would be good to have a comment here explaining why this is needed.  Just because we have an existing thing that checks that the "darwin" job succeeded, right?

src/darwin/Framework/CHIP/MTRDeviceControllerXPCConnection.mm
```diff
@@ -195,7 +195,7 @@ - (void)handleReportWithController:(id)controller
         if (!controllerDictionary) {
             return;
         }
-        NSNumber * nodeIdKey = [NSNumber numberWithUnsignedInteger:nodeId];
+        NSNumber * nodeIdKey = [NSNumber numberWithUnsignedLongLong:nodeId];
```
bzbarsky-apple, line 198: Is there any reason to not just do:
```suggestion
        NSNumber * nodeIdKey = @(nodeId);
```
and not mess with the `numberWithGuessWhatItShouldBe` bits?

bzbarsky-apple (approved):

.github/workflows/darwin.yaml
ksperling-apple: Yes, that was the idea, though I'm not sure that even works anyway

ksperling-apple: It does seem to work. I've put in a comment, but it's really just a temporary solution until the CI requirements for PRs have been switched over to directly point to the underlying framework build and test tasks. (It'll be nicer to see those correctly listed as "Required" in the checks list rather than indirectly via the aggregation job.)

src/darwin/Framework/CHIP/MTRDeviceControllerXPCConnection.mm
ksperling-apple: Done.

.github/workflows/darwin.yaml
ksperling-apple: The "Build Darwin" job with dependencies on the other jobs works for now, I'll take that out once you've changed the project build requirements.

PR #32610, opened 2024-03-18
CI: Don't run bootstrap if we restored from cache

ksperling-apple:
Handle caching and log upload directly within the bootstrap action, and don't perform the actual bootstrap if we have a cache hit. Include the platform name in the cache key.


andy31415:
@ksperling-apple do you happen to know how large our storage allocation for caching is? I believe our bootstrap includes cipd images for clang at least, so it may be quite large, so if we multiply it by the number of platforms we may increase the storage a lot. My current PW_ENVIRONMENT_ROOT on linux is 5.2GB. In setup we seem to have 15 separate requirements.txt.

andy31415:
> @ksperling-apple do you happen to know how large our storage allocation for caching is? I believe our bootstrap includes cipd images for clang at least, so it may be quite large, so if we multiply it by the number of platforms we may increase the storage a lot. My current PW_ENVIRONMENT_ROOT on linux is 5.2GB. In setup we seem to have 15 separate requirements.txt.

https://buildjet.com/for-github-actions/docs/about/pricing#build-jet-cache-limits says 20GB per week. this seems to work with 5GB, but will not work with 75.

ksperling-apple:
> > @ksperling-apple do you happen to know how large our storage allocation for caching is? I believe our bootstrap includes cipd images for clang at least, so it may be quite large, so if we multiply it by the number of platforms we may increase the storage a lot. My current PW_ENVIRONMENT_ROOT on linux is 5.2GB. In setup we seem to have 15 separate requirements.txt.
> 
> https://buildjet.com/for-github-actions/docs/about/pricing#build-jet-cache-limits says 20GB per week. this seems to work with 5GB, but will not work with 75.

At the moment I'm seeing cache sizes around 1.5-2 GB (tar ball size) in the logs, so I think we should be OK for the time being. If we're starting to hit the limit we can look at bootstrapping for all platforms instead of the targeted platform, so we have one large cache file shared by all jobs.

.github/workflows/cirque.yaml
```diff
@@ -57,6 +57,7 @@ jobs:
               with:
                 platform: linux
 
+            # TODO: Is what's being cached here actually compatible with a regular bootstrap?
```
ksperling-apple, line 60: @andy31415 Do you know any details here? It's not clear to me whether the `bootstrap.sh` run inside that container has exactly the same effect as running it directly, but I would expect there must be some difference, otherwise this workflow could just re-use the regular bootstrap action?

ksperling-apple: This should work as is for now since the `bootstrap-cache` action uses a different key from the cache step within the updated `bootstrap`, but it seems like a longer term solution would be to refactor this workflow to use a single cache step for both the bootstraps being performed here.

jmartinez-silabs (approved):

.github/workflows/cirque.yaml
andy31415: I think the compile bootstrap (via bootstrap.sh) should be compatible. Apart from that, cirque also has its own "boostrap" which is some docker image build and that is unrelated.

I believe for the purposes of bootstrap.sh and code compilation we should be fine.

andy31415: Actually looking at the script, it seems everything is funneled through chip-build-cirque/run.sh so I am not clear why a local cache would help (probably it does not).

@yunhanw-google do you know if we need this boostrap cache here at all? If we run all commands inside a docker image, bootstrapping outside docker does not seem useful.

we may want to test out just deleting this caching and see if cirque still passes.

andy31415 (approved):

.github/workflows/cirque.yaml
woody-apple: I don't believe the image has the bootstrap results inside of it. 


@ksperling-apple does this build faster?

ksperling-apple: > @ksperling-apple does this build faster?

Not doing the bootstrap if we have a cache hit seems to save around a minute per job, from looking at a small sample of runs.

ksperling-apple: I'm going to resolve this conversation to let the PR merge, leaving cirque to be addressed in a follow-up.

PR #32611, opened 2024-03-18
[Chef] Fix compilation error by removing printing ChipError number

erwinpan1:
Fix Chef  compilation error by removing printing ChipError number
which may cause compilation error due to its format could be `long` (formatted as "ld")
 or `int` (formatted as "d") depending on the platform

### DIFF
--- a/examples/chef/common/clusters/door-lock/chef-lock-endpoint.cpp
+++ b/examples/chef/common/clusters/door-lock/chef-lock-endpoint.cpp
@@ -590,8 +590,8 @@ bool LockEndpoint::weekDayScheduleForbidsAccess(uint16_t userIndex, bool * haveS
     auto chipError = chip::System::SystemClock().GetClock_RealTimeMS(cTMs);
     if (chipError != CHIP_NO_ERROR)
     {
-        ChipLogError(Zcl, "Lock App: unable to get current time to check user schedules [endpointId=%d,error=%d (%s)]", mEndpointId,
-                     chipError.AsInteger(), chipError.AsString());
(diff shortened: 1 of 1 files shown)


examples/chef/common/clusters/door-lock/chef-lock-endpoint.cpp
```diff
@@ -590,8 +590,8 @@ bool LockEndpoint::weekDayScheduleForbidsAccess(uint16_t userIndex, bool * haveS
     auto chipError = chip::System::SystemClock().GetClock_RealTimeMS(cTMs);
     if (chipError != CHIP_NO_ERROR)
     {
-        ChipLogError(Zcl, "Lock App: unable to get current time to check user schedules [endpointId=%d,error=%d (%s)]", mEndpointId,
-                     chipError.AsInteger(), chipError.AsString());
```
arkq, line 594: In other places in the codebase we have `static_cast<int>(err.AsInteger())`, so if the error number is valuable here maybe cast will better?

arkq (approved):

andy31415 (approved):

examples/chef/common/clusters/door-lock/chef-lock-endpoint.cpp
erwinpan1: @arkq On some platform (e.g. ARM 64 Linux) this could be `unsigned long`. So I don't think it still depends on the platform and we don't have a good way for all platforms. That's why I want to remove this at all.

bzbarsky-apple:
@erwinpan1 For future reference, you should just use CHIP_ERROR_FORMAT and error.Format().

PR #32623, opened 2024-03-18
Create `ember-compatibility-functions.h` to represent what `ember-compatibility-functions.cpp` contains

andy31415:
Without this, InteractionModelEngine.h actually declares what something in `util/ember-compatibility-functions.cpp` implements.

## Changes

- create `report/reporting.cpp` to contain what `report/reporting.h` header declares.
- Remove the re-declaration from InteractionModelEngine.h for what reporting.h declares
- Create `ember-compatibility-functions.h` to declare what `ember-compatibility-functions.cpp` implements and remove those declarations from InteractionModelEngine.h

## Updated dependency graph

The depenencies are still terrible and circular, however now at least we will have more of a chance to isolate them somehow.

image

My current goal is to have "if a function/class is implemented in foo.cpp, then foo.h should declare it".


src/app/util/ember-compatibility-functions.h
```diff
@@ -0,0 +1,115 @@
...
+} // namespace chip
+
+/** @brief Reporting Attribute Change
+ *
+ * This function is called by the framework when an attribute managed by the
+ * framework changes.  The application should call this function when an
+ * externally-managed attribute changes.  This function triggers attribute
+ * reports for subscriptions as needed.
+ */
+void MatterReportingAttributeChangeCallback(chip::EndpointId endpoint, chip::ClusterId clusterId, chip::AttributeId attributeId);
```
andy31415, line 105: these seem completely unrelated (and their definition was in reporting/reporting.h).

Adding this comment as I am considering splitting this to have a bit less circular dependency in things.

andy31415: Created 'reporting.cpp' instead now, to have things isolated.

The odd dependnecy is only because ember data model version is stored in generated code, so the header and cpp files cannot reside together now (global code needs to call 'notify' and actual implementation depends on generated code).

src/app/BUILD.gn
```diff
@@ -189,6 +189,9 @@ static_library("interaction-model") {
     "reporting/ReportSchedulerImpl.h",
     "reporting/SynchronizedReportSchedulerImpl.cpp",
     "reporting/SynchronizedReportSchedulerImpl.h",
+
+    # TODO: Reporting changes data model versions and is implemented in data_model.
```
bzbarsky-apple, line 193: This comment really confused me.  This is talking about cluster DataVersions, not the (compile-time-constant) data model revision, right?

Please rephrase to make it clearer.

src/app/dynamic_server/AccessControl.cpp
```diff
@@ -39,6 +44,11 @@ class DeviceTypeResolver : public Access::AccessControl::DeviceTypeResolver
 public:
     bool IsDeviceTypeOnEndpoint(DeviceTypeId deviceType, EndpointId endpoint) override
     {
+        // TODO: DynamicDispatcher.cpp actually hardcodes `false` here for every device/ep
+        //       we could remove the ember-compatibility-functions.h dependency if we hardcode
+        //       false here as well.
+        //
+        //       Should determine if hardcoding false here is the correct thing to do.
```
bzbarsky-apple, line 51: It's not in general, once you try to support a non-hardcoded policy of any sort.  Right now all the hardcoding is restricted to DynamicDispatcher.cpp..

src/darwin/Framework/Matter.xcodeproj/project.pbxproj
```diff
@@ -262,6 +262,8 @@
 		7596A8512878709F004DAE0E /* MTRAsyncCallbackQueueTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 7596A8502878709F004DAE0E /* MTRAsyncCallbackQueueTests.m */; };
 		7596A85528788557004DAE0E /* MTRClusters.mm in Sources */ = {isa = PBXBuildFile; fileRef = 7596A85228788557004DAE0E /* MTRClusters.mm */; };
 		7596A85728788557004DAE0E /* MTRClusters.h in Headers */ = {isa = PBXBuildFile; fileRef = 7596A85428788557004DAE0E /* MTRClusters.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		75A202E52BA8DBAC00A771DD /* reporting.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 75A202E42BA8DBAC00A771DD /* reporting.cpp */; };
```
bzbarsky-apple, line 265: So this can't be part of libCHIP because it uses emberAfDataVersionStorage which depends on generated stuff?

andy31415: I got a hit on `ember-compatibility-functions.cpp` inside this file, so I added it.

The reference seems to be at line 154 above here. Since I moved files from ember-compatibility-functions.cpp, I thought reporting.cpp has to reside here as well.

What is the correct setup here?

src/app/dynamic_server/AccessControl.cpp
andy31415: I removed the comment. However I believe long-term the fact that dynamic_server includes ember-compatibility-functions.h but provides its own implementation is somewhat odd. I believe we have to make the compatibility functions as some sort of interface that people can override (since dynamic-dispatch does just that, except it does so at link-time).

src/darwin/Framework/Matter.xcodeproj/project.pbxproj
bzbarsky-apple: The correct setup is that if reporting.cpp is built as part of libCHIP (which ember-compatibility-functions.cpp is not), then it should not be here, and it it's not built as part of libCHIP then it needs to be here, looks like.

The real question is whether reporting.cpp could be built as part of libCHIP.

bzbarsky-apple (approved):

src/darwin/Framework/Matter.xcodeproj/project.pbxproj
bzbarsky-apple: But the other comments explain why it can't be, so we are all good here.

harimau-qirex (approved):

andy31415:
Marking fast-track based on review from @harimau-qirex  ... still working to have Andy added to the org and google reviewers, that may take a bit due to MM timezone delays.

PR #32631, opened 2024-03-19
Dnssd changes to browse and resolve using open thread domain along wi…

### LINKED ISSUE #32630: Matter should browse both local domain and openthread.thread.home.arpa for Darwin
when trying to discover Thread accessories, the DNSServiceBrowse and the DNSServiceResolve must specify the "openthread.thread.home.arpa" domain. Currently we are passing "local" for the domain.

nivi-apple:
…th the local domain

Fixes: #32630

### DIFF
--- a/src/platform/Darwin/DnssdContexts.cpp
+++ b/src/platform/Darwin/DnssdContexts.cpp
@@ -463,7 +463,6 @@ ResolveContext::ResolveContext(void * cbContext, DnssdResolveCallback cb, chip::
     callback        = cb;
     protocol        = GetProtocol(cbAddressType);
     instanceName    = instanceNameToResolve;
-    domainName      = std::string(kLocalDot);
     consumerCounter = std::move(consumerCounterToUse);
 }
 
@@ -476,7 +475,6 @@ ResolveContext::ResolveContext(CommissioningResolveDelegate * delegate, chip::In
     callback        = nullptr;
     protocol        = GetProtocol(cbAddressType);
     instanceName    = instanceNameToResolve;
-    domainName      = std::string(kLocalDot);
     consumerCounter = std::move(consumerCounterToUse);
 }
 
@@ -528,7 +526,17 @@ void ResolveContext::DispatchSuccess()
 
     for (auto interfaceIndex : priorityInterfaceIndices)
     {
-        if (TryReportingResultsForInterfaceIndex(static_cast<uint32_t>(interfaceIndex)))
+        // Try finding interfaces for domains kLocalDot and kOpenThreadDot and delete them.
+        if (TryReportingResultsForInterfaceIndex(static_cast<uint32_t>(interfaceIndex), std::string(kLocalDot)))
+        {
+            if (needDelete)
+            {
+                MdnsContexts::GetInstance().Delete(this);
+            }
+            return;
+        }
+
+        if (TryReportingResultsForInterfaceIndex(static_cast<uint32_t>(interfaceIndex), std::string(kOpenThreadDot)))
         {
             if (needDelete)
             {
@@ -540,7 +548,7 @@ void ResolveContext::DispatchSuccess()
 
     for (auto & interface : interfaces)
     {
-        if (TryReportingResultsForInterfaceIndex(interface.first.first))
+        if (TryReportingResultsForInterfaceIndex(interface.first.first, interface.first.second))
         {
             break;
         }
@@ -552,7 +560,7 @@ void ResolveContext::DispatchSuccess()
     }
 }
 
-bool ResolveContext::TryReportingResultsForInterfaceIndex(uint32_t interfaceIndex)
+bool ResolveContext::TryReportingResultsForInterfaceIndex(uint32_t interfaceIndex, std::string domainName)
 {
     if (interfaceIndex == 0)
     {
@@ -560,7 +568,7 @@ bool ResolveContext::TryReportingResultsForInterfaceIndex(uint32_t interfaceInde
         return false;
     }
 
-    std::pair<uint32_t, std::string> interfaceKey = std::make_pair(interfaceIndex, this->domainName);
+    std::pair<uint32_t, std::string> interfaceKey = std::make_pair(interfaceIndex, domainName);
     auto & interface                              = interfaces[interfaceKey];
     auto & ips                                    = interface.addresses;
 
@@ -707,7 +715,7 @@ void ResolveContext::OnNewInterface(uint32_t interfaceId, const char * fullname,
     std::string domainFromHostname = GetDomainFromHostName(hostnameWithDomain);
     if (domainFromHostname.empty())
     {
-        ChipLogError(Discovery, "Mdns: Domain from hostname is empty");
+        ChipLogError(Discovery, "Mdns: No domain set in hostname %s", hostnameWithDomain);
         return;
     }
 
--- a/src/platform/Darwin/DnssdContexts.cpp
+++ b/src/platform/Darwin/DnssdContexts.cpp
@@ -436,7 +436,6 @@ void BrowseWithDelegateContext::OnBrowseAdd(const char * name, const char * type
     ChipLogProgress(Discovery, "Mdns: %s  name: %s, type: %s, domain: %s, interface: %" PRIu32, __func__, StringOrNullMarker(name),
                     StringOrNullMarker(type), StringOrNullMarker(domain), interfaceId);
 
-
     auto delegate = static_cast<DnssdBrowseDelegate *>(context);
     auto service  = GetService(name, type, protocol, interfaceId);
     delegate->OnBrowseAdd(service);
@@ -562,8 +561,8 @@ bool ResolveContext::TryReportingResultsForInterfaceIndex(uint32_t interfaceInde
     }
 
     std::pair<uint32_t, std::string> interfaceKey = std::make_pair(interfaceIndex, this->domainName);
-    auto & interface = interfaces[interfaceKey];
-    auto & ips       = interface.addresses;
+    auto & interface                              = interfaces[interfaceKey];
+    auto & ips                                    = interface.addresses;
 
     // Some interface may not have any ips, just ignore them.
     if (ips.size() == 0)
--- a/src/platform/Darwin/DnssdContexts.cpp
+++ b/src/platform/Darwin/DnssdContexts.cpp
@@ -705,7 +705,14 @@ void ResolveContext::OnNewInterface(uint32_t interfaceId, const char * fullname,
     // resolving.
     interface.fullyQualifiedDomainName = hostnameWithDomain;
 
-    std::pair<uint32_t, std::string> interfaceKey = std::make_pair(interfaceId, GetDomainNameFromHostName(hostnameWithDomain));
+    std::string domainFromHostname = GetDomainFromHostName(hostnameWithDomain);
+    if (domainFromHostname.empty())
+    {
+        ChipLogError(Discovery, "Mdns: Domain from hostname is empty");
+        return;
+    }
+
+    std::pair<uint32_t, std::string> interfaceKey = std::make_pair(interfaceId, domainFromHostname);
 
     interfaces.insert(std::make_pair(interfaceKey, std::move(interface)));
 }
--- a/src/platform/Darwin/DnssdContexts.cpp
+++ b/src/platform/Darwin/DnssdContexts.cpp
@@ -30,12 +30,6 @@ namespace {
 
 constexpr uint8_t kDnssdKeyMaxSize          = 32;
 constexpr uint8_t kDnssdTxtRecordMaxEntries = 20;
-constexpr char kLocalDot[]                  = "local.";
-
-bool IsLocalDomain(const char * domain)
-{
-    return strcmp(kLocalDot, domain) == 0;
-}
 
 std::string GetHostNameWithoutDomain(const char * hostnameWithDomain)
 {
@@ -252,6 +246,7 @@ void MdnsContexts::Delete(GenericContext * context)
     {
         DNSServiceRefDeallocate(context->serviceRef);
     }
+
     chip::Platform::Delete(context);
 }
 
@@ -388,7 +383,6 @@ void BrowseContext::OnBrowseAdd(const char * name, const char * type, const char
     ChipLogProgress(Discovery, "Mdns: %s  name: %s, type: %s, domain: %s, interface: %" PRIu32, __func__, StringOrNullMarker(name),
                     StringOrNullMarker(type), StringOrNullMarker(domain), interfaceId);
 
-    VerifyOrReturn(IsLocalDomain(domain));
     auto service = GetService(name, type, protocol, interfaceId);
     services.push_back(service);
 }
@@ -399,7 +393,6 @@ void BrowseContext::OnBrowseRemove(const char * name, const char * type, const c
                     StringOrNullMarker(type), StringOrNullMarker(domain), interfaceId);
 
     VerifyOrReturn(name != nullptr);
-    VerifyOrReturn(IsLocalDomain(domain));
 
     services.erase(std::remove_if(services.begin(), services.end(),
                                   [name, type, interfaceId](const DnssdService & service) {
@@ -443,7 +436,6 @@ void BrowseWithDelegateContext::OnBrowseAdd(const char * name, const char * type
     ChipLogProgress(Discovery, "Mdns: %s  name: %s, type: %s, domain: %s, interface: %" PRIu32, __func__, StringOrNullMarker(name),
                     StringOrNullMarker(type), StringOrNullMarker(domain), interfaceId);
 
-    VerifyOrReturn(IsLocalDomain(domain));
 
     auto delegate = static_cast<DnssdBrowseDelegate *>(context);
     auto service  = GetService(name, type, protocol, interfaceId);
@@ -456,7 +448,6 @@ void BrowseWithDelegateContext::OnBrowseRemove(const char * name, const char * t
                     StringOrNullMarker(type), StringOrNullMarker(domain), interfaceId);
 
     VerifyOrReturn(name != nullptr);
-    VerifyOrReturn(IsLocalDomain(domain));
 
     auto delegate = static_cast<DnssdBrowseDelegate *>(context);
     auto service  = GetService(name, type, protocol, interfaceId);
@@ -473,6 +464,7 @@ ResolveContext::ResolveContext(void * cbContext, DnssdResolveCallback cb, chip::
     callback        = cb;
     protocol        = GetProtocol(cbAddressType);
     instanceName    = instanceNameToResolve;
+    domainName      = std::string(kLocalDot);
     consumerCounter = std::move(consumerCounterToUse);
 }
 
@@ -485,6 +477,7 @@ ResolveContext::ResolveContext(CommissioningResolveDelegate * delegate, chip::In
     callback        = nullptr;
     protocol        = GetProtocol(cbAddressType);
     instanceName    = instanceNameToResolve;
+    domainName      = std::string(kLocalDot);
     consumerCounter = std::move(consumerCounterToUse);
 }
 
@@ -548,7 +541,7 @@ void ResolveContext::DispatchSuccess()
 
     for (auto & interface : interfaces)
     {
-        if (TryReportingResultsForInterfaceIndex(interface.first))
+        if (TryReportingResultsForInterfaceIndex(interface.first.first))
         {
             break;
         }
@@ -568,7 +561,8 @@ bool ResolveContext::TryReportingResultsForInterfaceIndex(uint32_t interfaceInde
         return false;
     }
 
-    auto & interface = interfaces[interfaceIndex];
+    std::pair<uint32_t, std::string> interfaceKey = std::make_pair(interfaceIndex, this->domainName);
+    auto & interface = interfaces[interfaceKey];
     auto & ips       = interface.addresses;
 
     // Some interface may not have any ips, just ignore them.
@@ -596,7 +590,7 @@ bool ResolveContext::TryReportingResultsForInterfaceIndex(uint32_t interfaceInde
     return true;
 }
 
-CHIP_ERROR ResolveContext::OnNewAddress(uint32_t interfaceId, const struct sockaddr * address)
+CHIP_ERROR ResolveContext::OnNewAddress(const std::pair<uint32_t, std::string> interfaceKey, const struct sockaddr * address)
 {
     // If we don't have any information about this interfaceId, just ignore the
     // address, since it won't be usable anyway without things like the port.
@@ -604,7 +598,9 @@ CHIP_ERROR ResolveContext::OnNewAddress(uint32_t interfaceId, const struct socka
     // on the system, because the hostnames we are looking up all end in
     // ".local".  In other words, we can get regular DNS results in here, not
     // just DNS-SD ones.
-    if (interfaces.find(interfaceId) == interfaces.end())
+    uint32_t interfaceId = interfaceKey.first;
+
+    if (interfaces.find(interfaceKey) == interfaces.end())
     {
         return CHIP_NO_ERROR;
     }
@@ -627,7 +623,7 @@ CHIP_ERROR ResolveContext::OnNewAddress(uint32_t interfaceId, const struct socka
         return CHIP_NO_ERROR;
     }
 
-    interfaces[interfaceId].addresses.push_back(ip);
+    interfaces[interfaceKey].addresses.push_back(ip);
 
     return CHIP_NO_ERROR;
 }
@@ -709,7 +705,9 @@ void ResolveContext::OnNewInterface(uint32_t interfaceId, const char * fullname,
     // resolving.
     interface.fullyQualifiedDomainName = hostnameWithDomain;
 
-    interfaces.insert(std::make_pair(interfaceId, std::move(interface)));
+    std::pair<uint32_t, std::string> interfaceKey = std::make_pair(interfaceId, GetDomainNameFromHostName(hostnameWithDomain));
+
+    interfaces.insert(std::make_pair(interfaceKey, std::move(interface)));
 }
 
 bool ResolveContext::HasInterface()
--- a/src/platform/Darwin/DnssdImpl.cpp
+++ b/src/platform/Darwin/DnssdImpl.cpp
@@ -177,11 +177,8 @@ namespace {
 void OpenThreadTimerExpiredCallback(System::Layer * systemLayer, void * callbackContext)
 {
     ChipLogProgress(Discovery, "Mdns: Resolve completed on the open thread domain.");
-    VerifyOrReturn(callbackContext != nullptr && systemLayer != nullptr,
-                   ChipLogError(Discovery, "Open thread timer callback context is null"));
-
-    auto sdCtx = reinterpret_cast<ResolveContext *>(callbackContext);
-    VerifyOrReturn(sdCtx != nullptr, ChipLogError(Discovery, "Resolve Context is null"));
+    auto sdCtx = static_cast<ResolveContext *>(callbackContext);
+    VerifyOrDie(sdCtx != nullptr);
     sdCtx->Finalize();
     hasOpenThreadTimerStarted = false;
 }
--- a/src/platform/Darwin/DnssdImpl.cpp
+++ b/src/platform/Darwin/DnssdImpl.cpp
@@ -152,7 +152,7 @@ namespace chip {
 namespace Dnssd {
 
 
-std::string GetDomainNameFromHostName(const char * hostname)
+std::string GetDomainFromHostName(const char * hostname)
 {
     if (HostNameHasDomain(hostname, kLocalDot))
     {
@@ -293,7 +293,12 @@ static void OnGetAddrInfo(DNSServiceRef sdRef, DNSServiceFlags flags, uint32_t i
     ReturnOnFailure(MdnsContexts::GetInstance().Has(sdCtx));
     LogOnFailure(__func__, err);
 
-    sdCtx->domainName = GetDomainNameFromHostName(hostname);
+    sdCtx->domainName = GetDomainFromHostName(hostname);
+    if (sdCtx->domainName.empty())
+    {
+        ChipLogError(Discovery, "Mdns: Domain name is not set");
+        return;
+    }
     if (kDNSServiceErr_NoError == err)
     {
         std::pair<uint32_t, std::string> key = std::make_pair(interfaceId, sdCtx->domainName);
--- a/src/platform/Darwin/DnssdImpl.cpp
+++ b/src/platform/Darwin/DnssdImpl.cpp
@@ -367,7 +367,7 @@ static void OnResolve(DNSServiceRef sdRef, DNSServiceFlags flags, uint32_t inter
         if (!sdCtx->isResolveRequested)
         {
             GetAddrInfo(sdCtx);
-            sdCtx->isResolveRequested = true;
+            sdCtx->isResolveRequested        = true;
             sdCtx->hasOpenThreadTimerStarted = false;
         }
     }
--- a/src/platform/Darwin/DnssdImpl.cpp
+++ b/src/platform/Darwin/DnssdImpl.cpp
@@ -26,15 +26,19 @@
 #include <lib/support/logging/CHIPLogging.h>
 #include <platform/CHIPDeviceLayer.h>
 
+using namespace chip;
 using namespace chip::Dnssd;
 using namespace chip::Dnssd::Internal;
 
 namespace {
 
-constexpr char kLocalDot[] = "local.";
+// The extra time in milliseconds that we will wait for the resolution on the open thread domain to complete.
+constexpr uint16_t kOpenThreadTimeoutInMsec = 250;
+
+static bool hasOpenThreadTimerStarted = false;
 
 constexpr DNSServiceFlags kRegisterFlags        = kDNSServiceFlagsNoAutoRename;
-constexpr DNSServiceFlags kBrowseFlags          = 0;
+constexpr DNSServiceFlags kBrowseFlags          = kDNSServiceFlagsShareConnection;
 constexpr DNSServiceFlags kGetAddrInfoFlags     = kDNSServiceFlagsTimeout | kDNSServiceFlagsShareConnection;
 constexpr DNSServiceFlags kResolveFlags         = kDNSServiceFlagsShareConnection;
 constexpr DNSServiceFlags kReconfirmRecordFlags = 0;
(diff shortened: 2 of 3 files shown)


src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -49,11 +53,27 @@ uint32_t GetInterfaceId(chip::Inet::InterfaceId interfaceId)
...
+    if (domainLength > hostnameLength)
+    {
+        return false;
+    }
+    const char * found = strstr(hostname, domain);
+    if (found == nullptr)
+    {
+        return false;
+    }
+    return (strncmp(found, domain, domainLength) == 0);
```
ksperling-apple, line 74: Isn't this meant to check for a suffix? `strstr` will find a match anywhere within the string, and the `strncmp` will always succeed because the comparison is restricted to `domainLength`. Also, shouldn't the check also include the `.` before `domain`?

src/platform/Darwin/DnssdImpl.h
```diff
@@ -27,9 +27,15 @@
 #include <string>
 #include <vector>
 
+constexpr char kLocalDot[] = "local.";
+
+constexpr char kOpenThreadDot[] = "openthread.thread.home.arpa";
```
ksperling-apple, line 32: How come there is no trailing`.` on this one?

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -131,10 +151,54 @@ std::shared_ptr<uint32_t> GetCounterHolder(const char * name)
...
+ * @param[in] callbackContext The context passed to the timer callback.
+ */
+void OpenThreadTimerExpiredCallback(System::Layer * systemLayer, void * callbackContext)
+{
+    ChipLogProgress(Discovery, "Mdns: Resolve completed on the open thread domain.");
+    VerifyOrReturn(callbackContext != nullptr && systemLayer != nullptr,
+                   ChipLogError(Discovery, "Open thread timer callback context is null"));
+
+    auto sdCtx = reinterpret_cast<ResolveContext *>(callbackContext);
+    VerifyOrReturn(sdCtx != nullptr, ChipLogError(Discovery, "Resolve Context is null"));
```
ksperling-apple, line 184:
```suggestion
    auto sdCtx = static_cast<ResolveContext *>(callbackContext);
    VerifyOrDie(sdCtx != nullptr);
```

If we want to check for a null context at all I think this should be a VerifyOrDie, or we can just remove the check and let things crash. The context is going to be whatever we passed in when we created the timer, so if it's null it's a programmer error and not something we should try to recover from at runtime.

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -26,15 +26,19 @@
...
 using namespace chip::Dnssd;
 using namespace chip::Dnssd::Internal;
 
 namespace {
 
-constexpr char kLocalDot[] = "local.";
+// The extra time in milliseconds that we will wait for the resolution on the open thread domain to complete.
+constexpr uint16_t kOpenThreadTimeoutInMsec = 250;
+
+static bool hasOpenThreadTimerStarted = false;
```
ksperling-apple, line 38: Shouldn't this be part of the `ResolveContext` rather than static?

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -219,25 +293,48 @@ static void OnGetAddrInfo(DNSServiceRef sdRef, DNSServiceFlags flags, uint32_t i
...
+        {
+            ChipLogProgress(Discovery, "Mdns: Resolve completed on the open thread domain.");
+            sdCtx->Finalize();
+        }
+        else if (sdCtx->domainName.compare(kLocalDot) == 0)
+        {
+            ChipLogProgress(
+                Discovery,
+                "Mdns: Resolve completed on the local domain. Starting a timer for the open thread resolve to come back");
+            if (!hasOpenThreadTimerStarted)
```
ksperling-apple, line 322: Why do we treat the local and openthread domains differently here, in the sense that we wait longer for the open thread resolve if necessary, but not the other way around if that one were to come back first? If that's the correct logic a comment explaining why we do that would be good.

src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -568,8 +560,9 @@ bool ResolveContext::TryReportingResultsForInterfaceIndex(uint32_t interfaceInde
         return false;
     }
 
-    auto & interface = interfaces[interfaceIndex];
-    auto & ips       = interface.addresses;
+    std::pair<uint32_t, std::string> interfaceKey = std::make_pair(interfaceIndex, this->domainName);
```
bzbarsky-apple, line 563: I don't get this part.  Why is this using `this->domainName`?  Are all the things in the map for the same domain?

src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -709,7 +704,16 @@ void ResolveContext::OnNewInterface(uint32_t interfaceId, const char * fullname,
     // resolving.
     interface.fullyQualifiedDomainName = hostnameWithDomain;
 
-    interfaces.insert(std::make_pair(interfaceId, std::move(interface)));
+    std::string domainFromHostname = GetDomainFromHostName(hostnameWithDomain);
+    if (domainFromHostname.empty())
+    {
+        ChipLogError(Discovery, "Mdns: Domain from hostname is empty");
```
bzbarsky-apple, line 710: Should this include the hostname in the log?

But also: are we guaranteed that the hostname will always have a domain in it?

src/platform/Darwin/DnssdImpl.cpp
bzbarsky-apple: It absolutely needs to be.

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -131,10 +151,54 @@ std::shared_ptr<uint32_t> GetCounterHolder(const char * name)
...
+{
+    if (HostNameHasDomain(hostname, kLocalDot))
+    {
+        return std::string(kLocalDot);
+    }
+    else if (HostNameHasDomain(hostname, kOpenThreadDot))
+    {
+        return std::string(kOpenThreadDot);
+    }
+    return std::string();
```
bzbarsky-apple, line 164: This looks wrong to me.  Why would we ignore all other domains?  Is this what Ted said to do?

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -219,25 +293,48 @@ static void OnGetAddrInfo(DNSServiceRef sdRef, DNSServiceFlags flags, uint32_t i
     ReturnOnFailure(MdnsContexts::GetInstance().Has(sdCtx));
     LogOnFailure(__func__, err);
 
+    sdCtx->domainName = GetDomainFromHostName(hostname);
```
bzbarsky-apple, line 296: So hold on.  Are there separate ResolveContext instances for the two domains?  Because if not, this is not making sense to me at all.

And looking above there are not.  So please remove the `domainName` member and fix all this code to work _without_ that.  I'll take another look once that's been done; stopping here.

bzbarsky-apple (dismissed):

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: i didn't have the domain name in the context before but recent changes to the code made me add that specifically this part - https://github.com/project-chip/connectedhomeip/blob/master/src/platform/Darwin/DnssdContexts.cpp#L539

nivi-apple: There is no hostname here to get the domain from.

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: Yeah i will add a comment. The reason for this is resolution on open thread take a little longer and in our discussions with Ted we decided to give it 250 millisecs more before we wrap up. I will add a meaningful comment

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: will fix.

src/platform/Darwin/DnssdContexts.cpp
nivi-apple: This is incorrect. Initially I was getting the domain from the host name  :) but when i sync'ed with the tip, I saw the changes from you - https://github.com/project-chip/connectedhomeip/commit/000edb450b7c975ff0ab0f60060f0605b75d8a44, and did a hasty fix. :(

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: so Ted said the domain can be null or the open thread one but he was talking from his perspective. But we could have other valid domains so you are right. I wanted to keep the scope of the change small since previous code was always hardcoding to local. I will fix the getdomainfromhostname to do the right string operation and get the last .<domainname>. . I asked Ted if the open thread domain name should have a trailing dot. if not, then it will be a bit tricky.

nivi-apple: also will remove the hardcoding for matching with the local domain name and the open thread one.

nivi-apple: @bzbarsky-apple So when we get a browse with null domain we should browse both on local and the open thread. if we get an explicit domain we should browse on that domain. so my code needs changes. as for the resolve I think ted said the hostname should have a domain. I am double checking with him. if he confirms then for the resolve part, I will fix my GetDomainFromHostName api and get the domain from there. Does this sound good?

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: done.

src/platform/Darwin/DnssdImpl.h
nivi-apple: I checked with Ted. he said trailing . is more correct? he doesn't care either way. I told him we will go with the "arpa."

src/platform/Darwin/DnssdContexts.cpp
nivi-apple: Confirmed with Ted the hostname should always have a domain in the resolve callback. Added hostname in the log

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: modified this code to get the domain by operating on the string with an assumption the hostname looks like "<hostnameWithoutDomain>.<domain>." will add a comment in code for the same.

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: done.

src/platform/Darwin/DnssdImpl.h
nivi-apple: Modified this to - "default.service.arpa." after discussion with Ted. he mentioned that he didn't think the domain name through :)

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: @bzbarsky-apple Would it make sense to look for interfaces with domains we care about [here]( https://github.com/project-chip/connectedhomeip/blob/master/src/platform/Darwin/DnssdContexts.cpp#L539) that is call get interfaces with interfacekey - interface index + kLocalDot and  interface index + kOpenThreadDot

woody-apple (approved):

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -131,10 +133,70 @@ std::shared_ptr<uint32_t> GetCounterHolder(const char * name)
...
+ *        The domainName returned from this API is "<domain>."
+ *
+ * @param[in] hostname The hostname with domain.
+ */
+std::string GetDomainFromHostName(const char * hostnameWithDomain)
+{
+    std::string hostname = std::string(hostnameWithDomain);
+
+    // Find the last occurence of '.'
+    size_t last_pos = hostname.find_last_of(".");
```
ksperling-apple, line 148: This should always be the last character of the string though? If it isn't then something else is going wrong.

src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -548,7 +548,7 @@ void ResolveContext::DispatchSuccess()
 
     for (auto & interface : interfaces)
     {
-        if (TryReportingResultsForInterfaceIndex(interface.first))
+        if (TryReportingResultsForInterfaceIndex(interface.first.first, interface.first.second))
```
ksperling-apple, line 551: I think this code would be more readable if we had structs with 2 meaningfully-named fields, rather than using std::pair. If you use aggregate initialization it doesn't involve writing more code either.

src/platform/Darwin/DnssdImpl.h
```diff
@@ -27,9 +27,15 @@
 #include <string>
 #include <vector>
 
+constexpr char kLocalDot[] = "local.";
+
+constexpr char kOpenThreadDot[] = "default.service.arpa.";
```
ksperling-apple, line 32: Maybe we should update the name to reference SRP instead of OpenThread? In the future we may well be using this on WiFi as well.

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -183,14 +245,24 @@ static void OnBrowse(DNSServiceRef sdRef, DNSServiceFlags flags, uint32_t interf
...
+    // We will browse on both the local domain and the open thread domain.
+    ChipLogProgress(Discovery, "Browsing for: %s on domain %s", StringOrNullMarker(type), kLocalDot);
+
+    auto sdRefLocal = sdCtx->serviceRef; // Mandatory copy because of kDNSServiceFlagsShareConnection
+    err             = DNSServiceBrowse(&sdRefLocal, kBrowseFlags, interfaceId, type, kLocalDot, OnBrowse, sdCtx);
+    VerifyOrReturnError(kDNSServiceErr_NoError == err, sdCtx->Finalize(err));
 
+    ChipLogProgress(Discovery, "Browsing for: %s on domain %s", StringOrNullMarker(type), kOpenThreadDot);
+
+    DNSServiceRef sdRefOpenThread = sdCtx->serviceRef; // Mandatory copy because of kDNSServiceFlagsShareConnection
```
ksperling-apple, line 260: Nit: Use `DNSServiceRef` on line 154 or `auto` here.

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -219,25 +291,52 @@ static void OnGetAddrInfo(DNSServiceRef sdRef, DNSServiceFlags flags, uint32_t i
...
+        return;
+    }
     if (kDNSServiceErr_NoError == err)
     {
-        sdCtx->OnNewAddress(interfaceId, address);
+        std::pair<uint32_t, std::string> key = std::make_pair(interfaceId, domainName);
+        sdCtx->OnNewAddress(key, address);
     }
 
     if (!(flags & kDNSServiceFlagsMoreComing))
```
ksperling-apple, line 306: Can `kDNSServiceFlagsMoreComing` be relied upon to work this way when `kDNSServiceFlagsShareConnection` is in use? It seems like we're assuming here that it means the `DNSServiceBrowse` and `DNSServiceGetAddrInfo` calls for the domain of this last result is done for now.

https://developer.apple.com/documentation/dnssd/1823436-anonymous/kdnsserviceflagsshareconnection

src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -536,7 +526,17 @@ void ResolveContext::DispatchSuccess()
 
     for (auto interfaceIndex : priorityInterfaceIndices)
     {
-        if (TryReportingResultsForInterfaceIndex(static_cast<uint32_t>(interfaceIndex)))
+        // Try finding interfaces for domains kLocalDot and kOpenThreadDot and delete them.
```
bzbarsky-apple, line 529: This comment should say something like:

> Prioritize results on the interfaces in priorityInterfaceIndices for kLocalDot and kOpenThreadDot domains.

But this raises the question: why are we not just prioritizing them for _all domains_?

src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -596,15 +597,17 @@ bool ResolveContext::TryReportingResultsForInterfaceIndex(uint32_t interfaceInde
     return true;
 }
 
-CHIP_ERROR ResolveContext::OnNewAddress(uint32_t interfaceId, const struct sockaddr * address)
+CHIP_ERROR ResolveContext::OnNewAddress(const std::pair<uint32_t, std::string> interfaceKey, const struct sockaddr * address)
```
bzbarsky-apple, line 600: Why is key passed by value here?  Should be const reference, I would think.

src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -560,16 +560,17 @@ void ResolveContext::DispatchSuccess()
     }
 }
 
-bool ResolveContext::TryReportingResultsForInterfaceIndex(uint32_t interfaceIndex)
+bool ResolveContext::TryReportingResultsForInterfaceIndex(uint32_t interfaceIndex, std::string domainName)
```
bzbarsky-apple, line 563: Why is name passed by value?  I'd think it should be const reference.

And really, the argument should be a const reference to the key type...

src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -709,7 +712,16 @@ void ResolveContext::OnNewInterface(uint32_t interfaceId, const char * fullname,
...
+    std::string domainFromHostname = GetDomainFromHostName(hostnameWithDomain);
+    if (domainFromHostname.empty())
+    {
+        ChipLogError(Discovery, "Mdns: No domain set in hostname %s", hostnameWithDomain);
+        return;
+    }
+
+    std::pair<uint32_t, std::string> interfaceKey = std::make_pair(interfaceId, domainFromHostname);
+
+    interfaces.insert(std::make_pair(interfaceKey, std::move(interface)));
```
bzbarsky-apple, line 724: Is there a reason this is not `std::move`-ing the key too?  It should.

src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -709,7 +712,16 @@ void ResolveContext::OnNewInterface(uint32_t interfaceId, const char * fullname,
...
 
-    interfaces.insert(std::make_pair(interfaceId, std::move(interface)));
+    std::string domainFromHostname = GetDomainFromHostName(hostnameWithDomain);
+    if (domainFromHostname.empty())
+    {
+        ChipLogError(Discovery, "Mdns: No domain set in hostname %s", hostnameWithDomain);
+        return;
+    }
+
+    std::pair<uint32_t, std::string> interfaceKey = std::make_pair(interfaceId, domainFromHostname);
```
bzbarsky-apple, line 722: This would have been a _great_ place to use `auto`.

src/platform/Darwin/DnssdImpl.cpp
bzbarsky-apple: Yes, exactly.  This is making some assumptions, but then acting as if those assumptions are not being made, which means it will fail in bad ways if those assumptions happen to be violated.

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -131,10 +133,70 @@ std::shared_ptr<uint32_t> GetCounterHolder(const char * name)
...
+    }
+}
+
+/**
+ * @brief Starts a timer to wait for the resolution on the kOpenThreadDot domain to happen.
+ *
+ * @param[in] timeoutSeconds The timeout in seconds.
+ * @param[in] ResolveContext The resolve context.
+ */
+void StartOpenThreadTimer(uint16_t timeoutInMSecs, ResolveContext * ctx)
```
bzbarsky-apple, line 193: Where do we stop the timer on destruction of the context?  This looks like potential use-after-free to me...

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -219,25 +291,52 @@ static void OnGetAddrInfo(DNSServiceRef sdRef, DNSServiceFlags flags, uint32_t i
...
 
     if (!(flags & kDNSServiceFlagsMoreComing))
     {
         VerifyOrReturn(sdCtx->HasAddress(), sdCtx->Finalize(kDNSServiceErr_BadState));
-        sdCtx->Finalize();
+
+        if (domainName.compare(kOpenThreadDot) == 0)
+        {
+            ChipLogProgress(Discovery, "Mdns: Resolve completed on the open thread domain.");
+            sdCtx->Finalize();
```
bzbarsky-apple, line 313: Where is the timer being cleared?

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -219,25 +291,52 @@ static void OnGetAddrInfo(DNSServiceRef sdRef, DNSServiceFlags flags, uint32_t i
...
+                "Mdns: Resolve completed on the local domain. Starting a timer for the open thread resolve to come back");
+
+            // Usually the resolution on the local domain is quicker than on the open thread domain. We would like to give the
+            // resolution on the open thread domain around 250 millisecs more to give it a chance to resolve before finalizing
+            // the resolution.
+            if (!sdCtx->hasOpenThreadTimerStarted)
+            {
+                // Schedule a timer to allow the resolve on OpenThread domain to complete.
+                StartOpenThreadTimer(kOpenThreadTimeoutInMsec, sdCtx);
+                sdCtx->hasOpenThreadTimerStarted = true;
```
bzbarsky-apple, line 328: Note that starting a timer can fail....

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -131,10 +133,70 @@ std::shared_ptr<uint32_t> GetCounterHolder(const char * name)
...
+ * @param[in] systemLayer The system layer.
+ * @param[in] callbackContext The context passed to the timer callback.
+ */
+void OpenThreadTimerExpiredCallback(System::Layer * systemLayer, void * callbackContext)
+{
+    ChipLogProgress(Discovery, "Mdns: Timer expired for resolve to complete on the open thread domain.");
+    auto sdCtx = static_cast<ResolveContext *>(callbackContext);
+    VerifyOrDie(sdCtx != nullptr);
+
+    if (sdCtx->hasOpenThreadTimerStarted)
```
bzbarsky-apple, line 181: What does this check even mean?  If we got here, the timer was clearly started.....

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -263,7 +362,14 @@ static void OnResolve(DNSServiceRef sdRef, DNSServiceFlags flags, uint32_t inter
     if (!(flags & kDNSServiceFlagsMoreComing))
     {
         VerifyOrReturn(sdCtx->HasInterface(), sdCtx->Finalize(kDNSServiceErr_BadState));
-        GetAddrInfo(sdCtx);
+
+        // If a resolve was not requested on this context, call GetAddrInfo and set the isResolveRequested flag to true.
+        if (!sdCtx->isResolveRequested)
```
bzbarsky-apple, line 367: I don't understand this logic.  We want to call GetAddrInfo any time we get OnResolve without kDNSServiceFlagsMoreComing, and _only do the GetAddrInfo on the things we didn't do it on yet_.

As in, this isResolveRequested flag needs to be per interface/domain pair, not per-context, no?

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -263,7 +362,14 @@ static void OnResolve(DNSServiceRef sdRef, DNSServiceFlags flags, uint32_t inter
...
+        {
+            GetAddrInfo(sdCtx);
+            sdCtx->isResolveRequested        = true;
+            sdCtx->hasOpenThreadTimerStarted = false;
```
bzbarsky-apple, line 371: Why are we setting hasOpenThreadTimerStarted false here?  Shouldn't it be set that way in the constructor?

src/platform/Darwin/DnssdImpl.h
bzbarsky-apple: Yes, please.

src/platform/Darwin/DnssdImpl.h
```diff
@@ -27,9 +27,15 @@
 #include <string>
 #include <vector>
 
+constexpr char kLocalDot[] = "local.";
```
bzbarsky-apple, line 30: Oh, and these non-namespaced things should not be in the header... Should be inside a namespace, and probably should be inline constexpr.

src/platform/Darwin/DnssdContexts.cpp
nivi-apple: I used std::pair from my conversation with Boris. I could add comments in code to say what first and second are if that helps.

src/platform/Darwin/DnssdContexts.cpp
nivi-apple: The problem here is I do not know what all domains to check for. For now we have the list of priority interfaces that have index ids but no domain but the interface key is a combination of interface ids + domain name. I am not sure if i can get all entries for an interface id from the map (can i provide a wild card domain name or something like that)?

nivi-apple: if you look in the latter part of the code [here](https://github.com/project-chip/connectedhomeip/pull/32631/files#diff-d3d1d99d691d0aef90e9de721cad391450d0e28aa0b727d812022acc3755f248R551) we are prioritizing whatever's in the map. for the deletion part since we have a harcoded list of interface indices, i am not sure what to use for the domain? if i could get all interfaces from the map that have the interface index I could iterate through them check if they need deletion and delete them

src/platform/Darwin/DnssdContexts.cpp
nivi-apple: done.

src/platform/Darwin/DnssdContexts.cpp
nivi-apple: done.

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: added a check that last_pos should be equal to the string length.

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: I had that code but got removed over iterations. My bad.

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: done.

src/platform/Darwin/DnssdContexts.cpp
nivi-apple: done.

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: added that.

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: When Boris, Ted and I discussed we didn't realize this can be an issue. I need to spend some time on this as I am not an expert in the domain. @bzbarsky-apple do you have an opinion here?

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: This is incorrect. I removed it before seeing your comment :D. I realized it was wrong

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: added check for error and finalizing the context if timer fails to start

src/platform/Darwin/DnssdContexts.cpp
Abhayakara: You should not be validating the domain. If you get a response, the response is valid. It's completely permissible and correct for a resolve, for example, to return a hostname in the SRV record that's looked up by DNS rather than mDNS, e.g. www.apple.com.

DNS responses always have an interface index of 0, and this includes lookups in the default.service.arpa domain. (We have a bug with this right now, but that's the intended behavior).

src/platform/Darwin/DnssdContexts.cpp
Abhayakara: I have to agree with Karsten here. Use a #define if you need to. :)

src/platform/Darwin/DnssdContexts.cpp
nivi-apple: I added the comment to indicate that I am deleting the interfaces with those 2 domains. AFAIK the code at line 551 is prioritizing all domains.

src/platform/Darwin/DnssdContexts.cpp
nivi-apple: for now I am going to add variables where i get the values from the pair with proper naming so there's no confusion. I need to get the addressed comments merged today. Please let me know if you still feel strongly about it.

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -131,10 +133,70 @@ std::shared_ptr<uint32_t> GetCounterHolder(const char * name)
...
+    {
+        // Get a substring without last '.'
+        std::string substring = hostname.substr(0, last_pos);
+
+        // Find the last occurence of '.' in the substring created above.
+        size_t pos = substring.find_last_of(".");
+        if (pos != std::string::npos)
+        {
+            // Return the domain name between the last 2 occurences of '.' including the trailing dot'.'.
+            return std::string(hostname.substr(pos + 1, last_pos));
```
Abhayakara, line 159: The domain is all the labels but the first, not the last label. So e.g. you could get an mDNS response like:

foo.link-0.local. In this case, the domain name is link-0.local, and the hostname is foo.

In the case of service instance names, we have <hostname>.<service-name>.<domain>, where hostname is one label, service-name is two labels, and domain is all the rest of the labels.

src/platform/Darwin/DnssdImpl.cpp
Abhayakara: kDNSServiceFlagsMoreComing says that we have any more data on the shared connection. If you are doing multiple queries on the same connection, the response that doesn't set the MoreComing flag for any particular series of results can happen on any of the queries.

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: removed.

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -276,8 +382,13 @@ static CHIP_ERROR Resolve(ResolveContext * sdCtx, uint32_t interfaceId, chip::In
     auto err = DNSServiceCreateConnection(&sdCtx->serviceRef);
     VerifyOrReturnError(kDNSServiceErr_NoError == err, sdCtx->Finalize(err));
 
-    auto sdRefCopy = sdCtx->serviceRef; // Mandatory copy because of kDNSServiceFlagsShareConnection
-    err            = DNSServiceResolve(&sdRefCopy, kResolveFlags, interfaceId, name, type, kLocalDot, OnResolve, sdCtx);
+    // Similar to browse, will try to resolve using both the local domain and the open thread domain.
+    auto sdRefLocal = sdCtx->serviceRef; // Mandatory copy because of kDNSServiceFlagsShareConnection
```
Abhayakara, line 386: This is incorrect. You should resolve the domain that you got in the browse callback. There is no promise that if you do a browse in local, all your responses will be in local. Same for default.service.arpa. There is no need to do two resolves here, and in fact it's incorrect because there might be totally unrelated data that happens to match the hostname you got for local in default.service.arpa, or vice versa.

src/platform/Darwin/DnssdImpl.h
Abhayakara: On WiFi you'll be using a default browsing domain (passing NULL for the domain in the browse) and there won't be a need for a second browse.

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: fixed. added a cancel timer in the destructor for ResolveContext

src/platform/Darwin/DnssdContexts.cpp
nivi-apple: fixed this by iterating through the map. If the interface index matches the key, we prioritize the results.

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: I see. I need to re-write this code again. Thanks for the response.

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: We discussed this with you in our meeting with Boris. So we get a browse before resolve only for the pairing use cases. for operational reads/writes, we do not get a browse before a resolve and hence we had to do the same thing as browse. I know you re-iterated that what we get in the browse result can be different than what we requested. so what i could do is if the domain in the browse result is not null i.e if a browse preceded a resolve, i will use the domain from the browse result otherwise we will resolve on local domain and srp domain. would that work?

src/platform/Darwin/DnssdImpl.h
nivi-apple: moved them back to the .cpp file under anonymous namespace as it was

src/platform/Darwin/DnssdImpl.h
nivi-apple: I used Srp instead of OpenThread

src/platform/Darwin/DnssdContexts.cpp
nivi-apple: so I tried to change this to a struct and running into issues where the map does some sorting based on keys and is looking for some operators on the struct. I spent some time adding < and + operator but seems like a non trivial fix here.

nivi-apple: Filed : #32678

src/platform/Darwin/DnssdImpl.cpp
Abhayakara: You should never get a NULL domain in a browse callback. Are you seeing that happen?

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: N/A. The API was changed to look for the first occurence of '.' after talking to Ted seems like the hostname with domain has hostname.<domainname> after the first '.'. I will follow up on this after checking the matter spec on how the hostname could look like in different variations and will make any fixes as required.

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: done

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: This needs more thought and has to be fixed correctly. I need to check the API documentation and see how best to handle the kDNSServiceFlagsMoreComing for a shared connection. will do as a follow up

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: I misunderstood this. This needs rework and I am going to do that as a follow up.

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: This needs a discussion with @bzbarsky-apple. The code right now should work for the 2 domains we care about the local and srp domain but definitely has a bug. Need to be re-written as follow up

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: filed - #32681

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: filed - #32681

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: Filed: #32681

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: filed: #32681

PR #32638, opened 2024-03-20
Include the work directory path in the bootstrap cache key (to account for containerized/non-containerized builds)

ksperling-apple:
The cache key needs to include the work directory path (which differs between containerized and non-containerized builds).

Also don't touch submodules during bootstrap, since we have a separate script / step for that.

Includes changes from #32610 that were reverted due to cache issues.


bzbarsky-apple (approved):

andy31415 (approved):

yunhanw-google (approved):

.github/actions/bootstrap/action.yaml
```diff
@@ -5,9 +5,56 @@ inputs:
...
+outputs:
+  cache-hit:
+    description: "Bootstrap environment was restored from cache"
+    value: ${{ fromJSON(steps.bootstrap-cache.outputs.outputs).cache-hit }} # retry returns all outputs in `outputs`
+
 runs:
   using: "composite"
   steps:
-    - name: Bootstrap
+    - name: Calculate bootstrap cache key
```
ksperling-apple, line 20: Block auto-merge so tests can be run a couple of extra times to ensure this is stable.

ksperling-apple: I did a few re-runs of various jobs and all attempts passed.

PR #32665, opened 2024-03-21
[Silabs] Wi-fi folders moved.

rcasallas-silabs:
Wi-fi example platform gathered into `examples/platform/silabs/wifi`.

Examples compiled:
- light-switch-app
- lighting-app
- lit-icd-app
- lock-app
- pump-app
- thermostat
- window-app

Flatforms compiled:
- BRD4166A
- BRD4187C + Si9116 NCP
- BRD4187C + Si917 NCP
- BRD4338A (Si917 SoC)


third_party/silabs/efr32_sdk.gni
```diff
@@ -249,11 +249,20 @@ template("efr32_sdk") {
...
+        # trng component
+        "${wifi_sdk_root}/components/device/silabs/si91x/wireless/crypto/inc",
+        "${wifi_sdk_root}/components/device/silabs/si91x/wireless/crypto/trng/inc",
+
         "${wifi_sdk_root}/components/protocol/wifi/si91x",
       ]
     }
     if (use_rs9116 || use_SiWx917) {
-      _include_dirs += [ "${chip_root}/src/platform/silabs/rs911x" ]
+      _include_dirs += [
```
jmartinez-silabs, line 260: when it was added to the template from the app build.gn we add this todo
`# TODO efr32_sdk should not need a header from this location`
Maybe add this todo back. I wonder why SDK files needs includes from example codes

examples/platform/silabs/wifi/sl_wifi_if.cpp
```diff
@@ -66,11 +66,9 @@ extern "C" {
 #include "sl_wifi.h"
 #include "sl_wifi_callback_framework.h"
 #include "wfx_host_events.h"
-#if SLI_SI91X_MCU_INTERFACE
 #include "sl_si91x_trng.h"
```
jmartinez-silabs, line 69: why is it ok to include this now with ncp options?

jmartinez-silabs (approved):

third_party/silabs/efr32_sdk.gni
rcasallas-silabs: The reason is that here, efr32_sdk is not a separate compilation unit, like a library. The efr32_sdk is a template, which results in a bundle of configuration and files that are merged with the application. So these files are actually part of the application, which depends on the platform code. Currently, there is now avoiding it. The SDK leave holes in that must be filled by the application, thus the template.
However, I have put the comment, if just to remember these dependencies are a mess.

examples/platform/silabs/wifi/sl_wifi_if.cpp
rcasallas-silabs: I am not sure ok is the right word, but the 917 NCP fails to compile without it.

andy31415:
PR #32665: Size comparison from 7f1413364e012fbf03cdba0aabd2fb73624e9311 to 796951d3eeef04c77338823650e22f808fa65c5c


<details>
<summary>Full report (3 builds for cc32xx, stm32)</summary>


| platform   | target       | config            | section      |   7f141336 |   796951d3 |   change |   % change |
|:-----------|:-------------|:------------------|:-------------|-----------:|-----------:|---------:|-----------:|
| cc32xx     | air-purifier | CC3235SF_LAUNCHXL | (read only)  |     583402 |     583402 |        0 |        0.0 |
|            |              |                   | (read/write) |     207992 |     207992 |        0 |        0.0 |
|            |              |                   | .bss         |     201228 |     201228 |        0 |        0.0 |
|            |              |                   | .data        |       1640 |       1640 |        0 |        0.0 |
|            |              |                   | .rodata      |      86610 |      86610 |        0 |        0.0 |
|            |              |                   | .text        |     494668 |     494668 |        0 |        0.0 |
|            | lock         | CC3235SF_LAUNCHXL | (read only)  |     628674 |     628674 |        0 |        0.0 |
|            |              |                   | (read/write) |     208336 |     208336 |        0 |        0.0 |
|            |              |                   | .bss         |     201720 |     201720 |        0 |        0.0 |
|            |              |                   | .data        |       1496 |       1496 |        0 |        0.0 |
|            |              |                   | .rodata      |     107066 |     107066 |        0 |        0.0 |
|            |              |                   | .text        |     519484 |     519484 |        0 |        0.0 |
| stm32      | light        | STM32WB5MM-DK     | (read/write) |     601241 |     601241 |        0 |        0.0 |
|            |              |                   | .bss         |     128344 |     128344 |        0 |        0.0 |
|            |              |                   | .data        |        668 |        668 |        0 |        0.0 |
|            |              |                   | .rodata      |      79700 |      79700 |        0 |        0.0 |
|            |              |                   | .text        |     382800 |     382800 |        0 |        0.0 |

</details>

PR #32670, opened 2024-03-21
Add GN arg to enable logging of node Id in exchange context

leonardmgh:
#32550 added the logging of the node ID to the logging of the exchange context. Since the code size increase caused by this was rather large, this has been made optional. This PR adds the option to enable the logging of the node ID when compiling using the arg `chip_exchange_node_id_logging=true`

### DIFF
--- a/src/lib/core/BUILD.gn
+++ b/src/lib/core/BUILD.gn
@@ -48,6 +48,7 @@ buildconfig_header("chip_buildconfig") {
     "CHIP_AUTOMATION_LOGGING=${chip_automation_logging}",
     "CHIP_PW_TOKENIZER_LOGGING=${chip_pw_tokenizer_logging}",
     "CHIP_USE_PW_LOGGING=${chip_use_pw_logging}",
+    "CHIP_EXCHANGE_NODE_ID_LOGGING=${chip_exchange_node_id_logging}",
     "CHIP_CONFIG_SHORT_ERROR_STR=${chip_config_short_error_str}",
     "CHIP_CONFIG_ENABLE_ARG_PARSER=${chip_config_enable_arg_parser}",
     "CHIP_TARGET_STYLE_UNIX=${chip_target_style_unix}",
--- a/src/lib/core/CHIPConfig.h
+++ b/src/lib/core/CHIPConfig.h
@@ -462,16 +462,6 @@
 #endif // CHIP_CONFIG_ENABLE_CONDITION_LOGGING
 
 /**
- *  @def CHIP_EXCHANGE_NODE_ID_LOGGING
- *
- *  @brief
- *    If asserted (1), enable logging of node IDs in exchange context.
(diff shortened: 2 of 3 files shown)


Damian-Nordic (approved):

src/lib/core/BUILD.gn
```diff
@@ -48,6 +48,7 @@ buildconfig_header("chip_buildconfig") {
     "CHIP_AUTOMATION_LOGGING=${chip_automation_logging}",
     "CHIP_PW_TOKENIZER_LOGGING=${chip_pw_tokenizer_logging}",
     "CHIP_USE_PW_LOGGING=${chip_use_pw_logging}",
+    "CHIP_EXCHANGE_NODE_ID_LOGGING=${chip_exchange_node_id_logging}",
```
bzbarsky-apple, line 51: If this is in buildconfig, its defaulting should be removed from CHIPConfig.h, no?

leonardmgh: I considered it, but for example `CHIP_AUTOMATION_LOGGING` also has a default value in `CHIPConfig.h`, so i tried to keep it consistent. Or is this by mistake?

bzbarsky-apple: I suspect that one is a mistake.  For example, CHIP_PW_TOKENIZER_LOGGING does not have any defaulting.

leonardmgh: I have removed the default from ChipConfig.h now. However i can see that all the following values from the BUILD.gn also have a default in ChipConfig.h :
`CHIP_CONFIG_TEST
CHIP_ERROR_LOGGING
CHIP_PROGRESS_LOGGING
CHIP_DETAIL_LOGGING
CHIP_CONFIG_LOG_MESSAGE_MAX_SIZE
CHIP_AUTOMATION_LOGGING
CHIP_CONFIG_SHORT_ERROR_STR
CHIP_CONFIG_ENABLE_ARG_PARSER
CHIP_CONFIG_MEMORY_MGMT_MALLOC
CHIP_CONFIG_MEMORY_DEBUG_CHECKS
CHIP_CONFIG_MEMORY_DEBUG_DMALLOC
CHIP_CONFIG_MINMDNS_DYNAMIC_OPERATIONAL_RESPONDER_LIST
CHIP_CONFIG_MINMDNS_MAX_PARALLEL_RESOLVES
CHIP_CONFIG_COMMAND_SENDER_BUILTIN_SUPPORT_FOR_BATCHED_COMMANDS ` 
Would it make sense to remove them in a separate commit to avoid future confusion?

bzbarsky-apple: > Would it make sense to remove them in a separate commit to avoid future confusion?

Yes, I think...

bzbarsky-apple (approved):

PR #32673, opened 2024-03-21
Remove weak-linkage usage for matter callbacks for pre/post attribute read/write and command invocation

andy31415:
These seem only to have ever been used for the placeholder apps. Migrated the placeholder apps to use the setter of the callbacks and a class.


Damian-Nordic (approved):

robszewczyk (approved):

src/app/util/MatterCallbacks.cpp
```diff
@@ -15,74 +15,11 @@
...
-                break;
-            }
-            break;
-        }
-    }
-
-    CHIP_ERROR PreCommandReceived(const chip::app::ConcreteCommandPath & commandPath,
-                                  const chip::Access::SubjectDescriptor & subjectDescriptor) override
-    {
-        return MatterPreCommandReceivedCallback(commandPath, subjectDescriptor);
```
bzbarsky-apple, line 73: So this is a public API change, right?

Can we do something to make it easy for SDK consumers to notice that their callbacks are not being called anymore?

PR #32675, opened 2024-03-21
Fix the dnssd code that browses on both the local and srp domains

### LINKED ISSUE #32668: Fix outstanding comments for https://github.com/project-chip/connectedhomeip/pull/32631

nivi-apple:
- Fixes the UAF issue with the timer

- Resolves critical comments from PR #32631

Fixes: #32668

### DIFF
--- a/src/platform/Darwin/DnssdContexts.cpp
+++ b/src/platform/Darwin/DnssdContexts.cpp
@@ -464,6 +464,7 @@ ResolveContext::ResolveContext(void * cbContext, DnssdResolveCallback cb, chip::
     protocol        = GetProtocol(cbAddressType);
     instanceName    = instanceNameToResolve;
     consumerCounter = std::move(consumerCounterToUse);
+    hasSrpTimerStarted = false;
 }
 
 ResolveContext::ResolveContext(CommissioningResolveDelegate * delegate, chip::Inet::IPAddressType cbAddressType,
@@ -476,9 +477,15 @@ ResolveContext::ResolveContext(CommissioningResolveDelegate * delegate, chip::In
     protocol        = GetProtocol(cbAddressType);
     instanceName    = instanceNameToResolve;
     consumerCounter = std::move(consumerCounterToUse);
+    hasSrpTimerStarted = false;
 }
 
-ResolveContext::~ResolveContext() {}
+ResolveContext::~ResolveContext() {
+    if (this->hasSrpTimerStarted)
+    {
+        CancelSrpTimer(this);
+    }
+}
 
 void ResolveContext::DispatchFailure(const char * errorStr, CHIP_ERROR err)
 {
@@ -526,8 +533,7 @@ void ResolveContext::DispatchSuccess()
 
     for (auto interfaceIndex : priorityInterfaceIndices)
     {
-        // Try finding interfaces for domains kLocalDot and kOpenThreadDot and delete them.
-        if (TryReportingResultsForInterfaceIndex(static_cast<uint32_t>(interfaceIndex), std::string(kLocalDot)))
+        if (TryReportingResultsForInterfaceIndex(interfaceIndex))
         {
             if (needDelete)
             {
@@ -536,7 +542,7 @@ void ResolveContext::DispatchSuccess()
             return;
         }
 
-        if (TryReportingResultsForInterfaceIndex(static_cast<uint32_t>(interfaceIndex), std::string(kOpenThreadDot)))
+        if (TryReportingResultsForInterfaceIndex(interfaceIndex))
         {
             if (needDelete)
             {
@@ -548,7 +554,8 @@ void ResolveContext::DispatchSuccess()
 
     for (auto & interface : interfaces)
     {
-        if (TryReportingResultsForInterfaceIndex(interface.first.first, interface.first.second))
+        auto interfaceId = interface.first.first;
+        if (TryReportingResultsForInterfaceIndex(interfaceId))
         {
             break;
         }
@@ -560,7 +567,7 @@ void ResolveContext::DispatchSuccess()
     }
 }
 
-bool ResolveContext::TryReportingResultsForInterfaceIndex(uint32_t interfaceIndex, std::string domainName)
+bool ResolveContext::TryReportingResultsForInterfaceIndex(uint32_t interfaceIndex)
 {
     if (interfaceIndex == 0)
     {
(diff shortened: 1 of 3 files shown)


src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -458,27 +458,35 @@ ResolveContext::ResolveContext(void * cbContext, DnssdResolveCallback cb, chip::
...
-    protocol        = GetProtocol(cbAddressType);
-    instanceName    = instanceNameToResolve;
-    consumerCounter = std::move(consumerCounterToUse);
+    type               = ContextType::Resolve;
+    context            = cbContext;
+    callback           = cb;
+    protocol           = GetProtocol(cbAddressType);
+    instanceName       = instanceNameToResolve;
+    consumerCounter    = std::move(consumerCounterToUse);
+    hasSrpTimerStarted = false;
```
ksperling-apple, line 467: Nit: This is not needed since you've already got it initialized to false in the header.

ksperling-apple (approved):

src/platform/Darwin/DnssdContexts.cpp
bzbarsky-apple, line 467: Why is this not default-initialized in the class decl, so you don't have to do it in both constructors?

I know other members were doing things this way; that's not a reason to copy the bad pattern.

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -144,59 +148,57 @@ std::string GetDomainFromHostName(const char * hostnameWithDomain)
...
 }
 
+/**
+ * @brief Cancels the timer that was started to wait for the resolution on the kSrpDot domain to happen.
+ *
+ * @param[in] ResolveContext The resolve context.
+ */
+void CancelSrpTimer(ResolveContext * ctx)
+{
+    DeviceLayer::SystemLayer().CancelTimer(SrpTimerExpiredCallback, reinterpret_cast<void *>(ctx));
```
bzbarsky-apple, line 195: Why do you need the cast here?  Just passing `ctx` as the second arg should work.  Same in StartTimer....

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -307,25 +309,37 @@ static void OnGetAddrInfo(DNSServiceRef sdRef, DNSServiceFlags flags, uint32_t i
     {
         VerifyOrReturn(sdCtx->HasAddress(), sdCtx->Finalize(kDNSServiceErr_BadState));
 
-        if (domainName.compare(kOpenThreadDot) == 0)
+        if (domainName.compare(kSrpDot) == 0)
```
bzbarsky-apple, line 312: I think this is broken.  You can get multiple SRP results, then a non-SRP result with !kDNSServiceFlagsMoreComing, and this code will end up sitting and waiting for 250ms when it shouldn't.  I think there's some confusion here about what kDNSServiceFlagsMoreComing means.  Which is understandable, because it can be confusing.  If set, all it means is "something on this same connection has more data to deliver right now".  In our case, it could be either of the two resolves, any of the GetAddrInfo those kicked off, etc.  In particular, it does not mean "no more data coming for the args that got passed here"; it means "no more stuff I know about right this second for anything that is using this connection".

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -307,25 +309,37 @@ static void OnGetAddrInfo(DNSServiceRef sdRef, DNSServiceFlags flags, uint32_t i
...
-            if (!sdCtx->hasOpenThreadTimerStarted)
+            if (!sdCtx->hasSrpTimerStarted)
             {
-                // Schedule a timer to allow the resolve on OpenThread domain to complete.
-                StartOpenThreadTimer(kOpenThreadTimeoutInMsec, sdCtx);
-                sdCtx->hasOpenThreadTimerStarted = true;
+                // Schedule a timer to allow the resolve on Srp domain to complete.
+                CHIP_ERROR error = StartSrpTimer(kSrpTimeoutInMsec, sdCtx);
+
+                // If the timer fails to start, finalize the context and return.
```
bzbarsky-apple, line 336: This comment should explain _why_, not _what_.  The "what" (finalize and return) is obvious from the code below.  The question is: why is that the right thing to do?  (It is, but the comment should explain why.)

src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -560,52 +568,60 @@ void ResolveContext::DispatchSuccess()
...
+        if (key.first == interfaceIndex)
+        {
+            auto & interface = interfaces[key];
+            auto & ips       = interface.addresses;
 
-    ChipLogProgress(Discovery, "Mdns: Resolve success on interface %" PRIu32, interfaceIndex);
+            // Some interface may not have any ips, just ignore them.
+            if (ips.size() == 0)
+            {
+                return false;
```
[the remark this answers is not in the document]
nivi-apple: I didn't modify this code. :) but yes I see we need to continue to the other interfaces

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: will do.

src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -548,7 +555,8 @@ void ResolveContext::DispatchSuccess()
 
     for (auto & interface : interfaces)
     {
-        if (TryReportingResultsForInterfaceIndex(interface.first.first, interface.first.second))
+        auto interfaceId = interface.first.first;
+        if (TryReportingResultsForInterfaceIndex(interfaceId))
```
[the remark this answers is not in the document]
nivi-apple: The reason I couldn't add the domain here is because there are hardcoded interfaces to prioritize that do not have a domain and I have no idea what domains to use here. we need to discuss this more for me to fix this properly.

src/platform/Darwin/DnssdContexts.cpp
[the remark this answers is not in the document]
bzbarsky-apple: > I didn't modify this code.

You changed the context around it, though.  Before, when there was only map entry the code was dealing with, return was correct. Now that it's trying to work with multiple map entries, returning is not correct.

So not modifying this code _is the bug_.

src/platform/Darwin/DnssdContexts.cpp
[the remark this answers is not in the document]
bzbarsky-apple: My point is that the "prioritize" bits indeed can only pass in the interface id (and then walk through all the matches and call the two-arg version for each one), but the bit here should pass in both pieces.

src/platform/Darwin/DnssdContexts.cpp
bzbarsky-apple: > i already have it defaulting to false in the class decl.

Then we shouldn't touch it in the constructor.

src/platform/Darwin/DnssdContexts.cpp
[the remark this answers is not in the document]
nivi-apple: yup. got that after i closley inspected the changes. it is a bug and i fixed it locally.

PR #32648, opened 2024-03-20
Start moving matter pre/post callbacks from `__weak__` handling to a injectable class instead.

andy31415:
I kept the weak linkage for now and only used injection in IM. 
If this passes tests, I will start updating our examples to use injection instead of overriding weak functions. Also removed the comment of `this file is generated by zap` since that is wrong.

## Changes

- created an interface equivalent to `MatterCallbacks` that existed before
- moved the `__weak__` empty declarations from various cpp files into `MatterCallbacks.cpp`
- Implemented a Get/Set instance for these callbacks

## Testing

Expect CI to validate this as part of e2e tests. I have kept the weak linkage support and I expect functionality to remain the same.

PR #32677, opened 2024-03-21
Testing Revert "Start moving matter pre/post callbacks from `__weak__` handling to a injectable class instead."

woody-apple:
Reverts project-chip/connectedhomeip#32648


woody-apple:
This may be breaking many builds, testing it

Issue #32587, opened 2024-03-15
[Workflow]  the QEMU tests are still failing in CI 


yufengwangca:
### Workflow issues

1.
2.
3.
...
https://github.com/project-chip/connectedhomeip/actions/workflows/qemu.yaml?query=branch%3Amaster

### Platform

esp32

andy31415:
Do we have the logs of what was happening here? the link seems to show all actions (which now are green) and I cannot find an example failure with logs.

andy31415:
I see some failures with `The hosted runner: GitHub Actions ... lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error.`

was this the error? Something else?

PR #32685, opened 2024-03-22
[ESP32]  Made changes to the ESP32 QEMU Dockerfile to address the QEMU CI failure.

shripad621git:
**Problem:**
- https://github.com/project-chip/connectedhomeip/pull/32587 - The tests before disabling the QEMU tests in CI were failing.
- When we run QEMU with address sanitizers enabled ,we get following warning at the start
`WARNING: ASan doesn't fully support makecontext/swapcontext functions and may produce false positives in some cases`!.
- For some case, this features works fine with Clang toolchain but has issues with GNU.
- I have attached the [reference Issue](https://github.com/espressif/qemu/issues/77.) here.
- Also it might consume more CPU or memory resources which might affect the CI .

**Change Overview**
- Modified the Dockerfile and built it locally and tested in the CI with the fix.

### DIFF
--- a/integrations/docker/images/stage-3/chip-build-esp32-qemu/Dockerfile
+++ b/integrations/docker/images/stage-3/chip-build-esp32-qemu/Dockerfile
@@ -14,7 +14,7 @@ WORKDIR /opt/espressif/qemu
 # Setup QEMU emulator for ESP32 platform
 RUN set -x \
     && git clone --depth 1 -b esp-develop-20210826 https://github.com/espressif/qemu.git ../qemu-src \
-    && ../qemu-src/configure --target-list=xtensa-softmmu --enable-debug --enable-sanitizers --disable-strip --disable-user --disable-capstone --disable-vnc --disable-sdl --disable-gtk \
+    && ../qemu-src/configure --target-list=xtensa-softmmu --enable-debug --disable-strip --disable-user --disable-capstone --disable-vnc --disable-sdl --disable-gtk \
     && make -j8 \
     && : # last line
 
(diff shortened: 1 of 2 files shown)


integrations/docker/images/stage-3/chip-build-esp32-qemu/Dockerfile
```diff
@@ -14,7 +14,7 @@ WORKDIR /opt/espressif/qemu
 # Setup QEMU emulator for ESP32 platform
 RUN set -x \
     && git clone --depth 1 -b esp-develop-20210826 https://github.com/espressif/qemu.git ../qemu-src \
-    && ../qemu-src/configure --target-list=xtensa-softmmu --enable-debug --enable-sanitizers --disable-strip --disable-user --disable-capstone --disable-vnc --disable-sdl --disable-gtk \
+    && ../qemu-src/configure --target-list=xtensa-softmmu --enable-debug --disable-strip --disable-user --disable-capstone --disable-vnc --disable-sdl --disable-gtk \
```
andy31415, line 17: please explain this change in the summary: *why* is this the correct change.

Since this modifies the image, you have to update chip-build/version as well.

shripad621git: When you run QEMU with address sanitizers enabled ,we get following warning at the start
`WARNING: ASan doesn't fully support makecontext/swapcontext functions and may produce false positives in some cases!`.
For some case, this features works fine with Clang toolchain but has issues with GNU.
[Reference issue](https://github.com/espressif/qemu/issues/77).
Also it might consume more CPU or memory resources which might affect the CI .

bzbarsky-apple: @shripad621git The point is, all that info should be in the commit message.

shripad621git: Updated the description.

andy31415 (approved):

PR #32694, opened 2024-03-24
Only genenerate the random fallback MAC address once

ksperling-apple:
This avoids changing our hostname every time we restart advertising, in scenarios where a real physical MAC is unavailable (e.g. on iOS).

Also set buf.size() correctly in GetPrimaryMACAddress(). When we fall back to returning the WiFi MAC on a device that supports Thread, the returned MAC is shorter than kPrimaryMACAddressLength.

### DIFF
--- a/src/include/platform/internal/GenericConfigurationManagerImpl.ipp
+++ b/src/include/platform/internal/GenericConfigurationManagerImpl.ipp
@@ -374,19 +374,21 @@ CHIP_ERROR GenericConfigurationManagerImpl<ConfigClass>::GetPrimaryMACAddress(Mu
     if (buf.size() != ConfigurationManager::kPrimaryMACAddressLength)
         return CHIP_ERROR_INVALID_ARGUMENT;
 
-    memset(buf.data(), 0, buf.size());
+    memset(buf.data(), 0, buf.size()); // zero the whole buffer, in case the caller ignores buf.size()
 
 #if CHIP_DEVICE_CONFIG_ENABLE_THREAD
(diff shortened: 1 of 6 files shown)


arkq (approved):

jmartinez-silabs (approved):

src/include/platform/internal/GenericConfigurationManagerImpl.ipp
```diff
@@ -374,19 +374,21 @@ CHIP_ERROR GenericConfigurationManagerImpl<ConfigClass>::GetPrimaryMACAddress(Mu
...
         return CHIP_ERROR_INVALID_ARGUMENT;
 
-    memset(buf.data(), 0, buf.size());
+    memset(buf.data(), 0, buf.size()); // zero the whole buffer, in case the caller ignores buf.size()
 
 #if CHIP_DEVICE_CONFIG_ENABLE_THREAD
     if (chip::DeviceLayer::ThreadStackMgr().GetPrimary802154MACAddress(buf.data()) == CHIP_NO_ERROR)
     {
         ChipLogDetail(DeviceLayer, "Using Thread extended MAC for hostname.");
+        buf.reduce_size(kThreadMACAddressLength);
```
Damian-Nordic, line 383: Sorry for a late review but just noticed that this does nothing as it only shrinks the local span, not the one passed by the caller.

bzbarsky-apple: @Damian-Nordic Good catch!  https://github.com/project-chip/connectedhomeip/pull/37609

PR #32695, opened 2024-03-25
Darwin: use macos-13 runners

ksperling-apple:
According to github docs these have 4 CPUs (vs 3 for macos-12 == latest)


ksperling-apple:
Running on macOS-13 doesn't seem like a major improvement at this point.

PR #32698, opened 2024-03-25
Configurable max network

Kshitijjain21:
As of now Matter only supports one network. The aim of the PR is to make the max network value configurable. This will unlock tests that need more that 1 networks testing removing networks, reordering networks and connecting to a different network.


src/platform/Linux/NetworkCommissioningDriver.h
```diff
@@ -147,11 +150,14 @@ class LinuxWiFiDriver final : public WiFiDriver
         Platform::SharedPtr<Crypto::P256Keypair> clientIdentityKeypair;
 #endif // CHIP_DEVICE_CONFIG_ENABLE_WIFI_PDC
     };
-
-    WiFiNetwork mSavedNetwork;
-    WiFiNetwork mStagingNetwork;
+    uint8_t currentNetId          = 0;
+    uint8_t connectedNetworkIndex = 0;
+    WiFiNetwork mSavedNetwork[kMaxNetworks];
```
bzbarsky-apple, line 155: Please update naming of these things to make it clear that there are multiple.

src/platform/Linux/NetworkCommissioningWiFiDriver.cpp
```diff
@@ -123,40 +128,56 @@ void LinuxWiFiDriver::Shutdown()
...
+    memcpy(mSavedNetwork, mStagingNetwork, sizeof(mSavedNetwork));
     return CHIP_NO_ERROR;
 }
 
 CHIP_ERROR LinuxWiFiDriver::RevertConfiguration()
 {
-    mStagingNetwork = mSavedNetwork;
+    currentNetId = 0;
+    memset(mStagingNetwork, 0, sizeof(mStagingNetwork));
+    memcpy(mStagingNetwork, mSavedNetwork, sizeof(mStagingNetwork));
```
bzbarsky-apple, line 172: Why do we need the memset, if we're going to memcpy anyway?

And presumably there should be a static_assert here that the two things have the same sizeof()?

src/platform/Linux/NetworkCommissioningDriver.h
```diff
@@ -147,11 +150,14 @@ class LinuxWiFiDriver final : public WiFiDriver
         Platform::SharedPtr<Crypto::P256Keypair> clientIdentityKeypair;
 #endif // CHIP_DEVICE_CONFIG_ENABLE_WIFI_PDC
     };
-
-    WiFiNetwork mSavedNetwork;
-    WiFiNetwork mStagingNetwork;
+    uint8_t currentNetId          = 0;
+    uint8_t connectedNetworkIndex = 0;
```
bzbarsky-apple, line 154: Please follow existing mWhatever naming convention.

And please document what these members mean.  I can't tell from the name what currentNetId is, and in code below it seems to be the number of free slots in mStagingNetwork?  But that makes no sense with the other uses.  It's really impossible to review this PR to tell which part is wrong without this being clearly documented.  But some part is wrong.

src/platform/Linux/NetworkCommissioningDriver.h
Kshitijjain21: Hi, Thank you for the feedback. Will update the PR with documentation and change the naming to the correct naming convention. There is still a bit of work remaining for this around storing the WiFi information in persistent storage. Will open a PR (not a draft) for review with all the changes in.

src/lib/core/CHIPConfig.h
```diff
@@ -1758,6 +1758,16 @@ extern const char CHIP_NON_PRODUCTION_MARKER[];
...
 #endif // CHIP_CONFIG_MAX_BDX_LOG_TRANSFERS
 
+/**
+ *  @def CHIP_CONFIG_MAX_WIFI_NETWORK
+ *
+ *  @brief
+ *    Maximum number of wifi network
+ */
+#ifndef CHIP_CONFIG_MAX_WIFI_NETWORK
+#define CHIP_CONFIG_MAX_WIFI_NETWORK 1
```
bzbarsky-apple, line 1768:
```suggestion
#define CHIP_CONFIG_MAX_WIFI_NETWORKS 1
```
and fix all the other places that mention it.

src/lib/core/CHIPConfig.h
```diff
@@ -1758,6 +1758,16 @@ extern const char CHIP_NON_PRODUCTION_MARKER[];
 #define CHIP_CONFIG_MAX_BDX_LOG_TRANSFERS 5
 #endif // CHIP_CONFIG_MAX_BDX_LOG_TRANSFERS
 
+/**
+ *  @def CHIP_CONFIG_MAX_WIFI_NETWORK
+ *
+ *  @brief
+ *    Maximum number of wifi network
```
bzbarsky-apple, line 1765:
```suggestion
 *    Maximum number of wifi networks
```

src/platform/Linux/NetworkCommissioningDriver.h
```diff
@@ -25,10 +25,14 @@
 
 #include <vector>
 
+#define CAN_HAVE_MAX_NETWORK 10
```
bzbarsky-apple, line 28: Shouldn't this be something like MAX_NETWORKS_MAXIMUM?

src/platform/Linux/NetworkCommissioningDriver.h
```diff
@@ -148,8 +153,16 @@ class LinuxWiFiDriver final : public WiFiDriver
 #endif // CHIP_DEVICE_CONFIG_ENABLE_WIFI_PDC
     };
 
-    WiFiNetwork mSavedNetwork;
-    WiFiNetwork mStagingNetwork;
+    // Number of entries in mStagedNetwork
```
bzbarsky-apple, line 156: There is no mStagedNetwork....

src/platform/Linux/NetworkCommissioningDriver.h
```diff
@@ -148,8 +153,16 @@ class LinuxWiFiDriver final : public WiFiDriver
...
+    uint8_t mCurrentNetId = 0;
+    WiFiNetwork mSavedNetwork[kMaxNetworks];
```
bzbarsky-apple, line 158: Again, please fix the naming.

bzbarsky-apple (changes requested):

src/platform/Linux/NetworkCommissioningWiFiDriver.cpp
```diff
@@ -34,16 +34,63 @@ namespace DeviceLayer {
...
+{
+    std::array<char[size], kMaxNetworks> arr{};
+
+    for (int i = 0; i < kMaxNetworks; i++)
+    {
+        for (size_t j = 0; j < size; j++)
+        {
+            arr[i][j] = keyName[j];
+        }
+        arr[i][size - 2] = intToChar(i);
```
bzbarsky-apple, line 70: So this is assuming that keyName[size - 1] is already 0?  Can we static_assert that, or is that not consteval (and if so, in what sense is any of this constexpr)?

src/platform/Linux/NetworkCommissioningWiFiDriver.cpp
```diff
@@ -52,58 +99,58 @@ inline CHIP_ERROR IgnoreNotFound(CHIP_ERROR err)
 }
 } // namespace
 
-// NOTE: For WiFiDriver, we uses two network configs, one is mSavedNetwork, and another is mStagingNetwork, during init, it will
-// load the network config from k-v storage, and loads it into both mSavedNetwork and mStagingNetwork. When updating the networks,
-// all changed are made on the staging network, and when the network is committed, it will update the mSavedNetwork to
-// mStagingNetwork and persist the changes.
+// NOTE: For WiFiDriver, we uses two network configs, one is mSavedNetworks, and another is mStagingNetworks, during init, it will
```
bzbarsky-apple, line 102:
```suggestion
// NOTE: For WiFiDriver, we use two sets of network configs, one is mSavedNetworks, and another is mStagingNetworks, during init, it will
```

src/platform/Linux/NetworkCommissioningWiFiDriver.cpp
```diff
@@ -52,58 +99,58 @@ inline CHIP_ERROR IgnoreNotFound(CHIP_ERROR err)
...
+// load the network config from k-v storage, and loads it into both mSavedNetworks and mStagingNetworks. When updating the networks,
+// all changed are made on the staging network, and when the network is committed, it will update the mSavedNetworks to
```
bzbarsky-apple, line 104:
```suggestion
// all changes are made on the staging network, and when the network is committed, it will update the mSavedNetworks to
```

src/platform/Linux/NetworkCommissioningWiFiDriver.cpp
```diff
@@ -123,96 +170,247 @@ void LinuxWiFiDriver::Shutdown()
 CHIP_ERROR LinuxWiFiDriver::CommitConfiguration()
 {
     auto & kvs = PersistedStorage::KeyValueStoreMgr();
-    ReturnErrorOnFailure(kvs.Put(kWiFiSSIDKeyName, mStagingNetwork.ssid, mStagingNetwork.ssidLen));
-#if CHIP_DEVICE_CONFIG_ENABLE_WIFI_PDC
-    if (mStagingNetwork.UsingPDC())
+
+    for (uint8_t networkIndex = 0; networkIndex < kMaxNetworks; networkIndex++)
```
bzbarsky-apple, line 174: Shouldn't this only go up to `mCurrentNetId` or whatever the actually initialized number is?  Otherwise it seems like we are persisting garbage data?

src/platform/Linux/NetworkCommissioningWiFiDriver.cpp
```diff
@@ -123,96 +170,247 @@ void LinuxWiFiDriver::Shutdown()
...
+    mStagingNetworks[mCurrentNetId].networkIdentityLen = 0;
+    mStagingNetworks[mCurrentNetId].clientIdentityLen  = 0;
+    mStagingNetworks[mCurrentNetId].clientIdentityKeypair.reset();
 #endif // CHIP_DEVICE_CONFIG_ENABLE_WIFI_PDC
+    mCurrentNetId++;
 
     return Status::kSuccess;
 }
 
+void LinuxWiFiDriver::ShiftNetworkAfterRemove()
```
bzbarsky-apple, line 273: Maybe this should be called something like "CompressStagingNetworksList"?  Not sure.

src/platform/Linux/NetworkCommissioningWiFiDriver.cpp
```diff
@@ -123,96 +170,247 @@ void LinuxWiFiDriver::Shutdown()
...
+
+    // Will need to reorder the list.
+    // Shift networks that are at a lower priority than the network to be removed if present in the list.
+    ShiftNetworkAfterRemove();
+    mCurrentNetId--;
+
     return Status::kSuccess;
 }
 
+bool LinuxWiFiDriver::StartReorderingEntries(uint8_t index, int8_t foundNetworkAtIndex)
```
bzbarsky-apple, line 326: I can't tell whether this function is doing what it means to, because I have no idea what it's trying to do.

src/platform/Linux/NetworkCommissioningWiFiDriver.cpp
```diff
@@ -123,96 +170,247 @@ void LinuxWiFiDriver::Shutdown()
...
+    {
+        if (mStagingNetworks[networkIndex].Matches(networkId))
+        {
+            foundNetworkAtIndex = networkIndex;
+            break;
+        }
+    }
+    VerifyOrReturnError(foundNetworkAtIndex != -1, Status::kNetworkIDNotFound);
+
+    // Network index larger to the current number of entries return out of range
```
bzbarsky-apple, line 385: What does "larger to" mean?

src/platform/Linux/NetworkCommissioningWiFiDriver.cpp
```diff
@@ -255,19 +453,34 @@ void LinuxWiFiDriver::ScanNetworks(ByteSpan ssid, WiFiDriver::ScanCallback * cal
...
 size_t LinuxWiFiDriver::WiFiNetworkIterator::Count()
 {
-    return driver->mStagingNetwork.Empty() ? 0 : 1;
+    size_t count = 0;
+
+    for (uint8_t networkIndex = 0; networkIndex < kMaxNetworks; networkIndex++)
+    {
+        if (!driver->mStagingNetworks[networkIndex].Empty())
+        {
+            count++;
```
bzbarsky-apple, line 462: Why are we bothering with `mCurrentNetId` then?  How does that differ from the count of non-empty networks?

src/platform/Linux/NetworkCommissioningWiFiDriver.cpp
```diff
@@ -333,24 +551,25 @@ CHIP_ERROR LinuxWiFiDriver::AddOrUpdateNetworkWithPDC(ByteSpan ssid, ByteSpan ne
 
 CHIP_ERROR LinuxWiFiDriver::GetNetworkIdentity(uint8_t networkIndex, MutableByteSpan & outNetworkIdentity)
 {
-    VerifyOrReturnError(!mStagingNetwork.Empty() && networkIndex == 0, CHIP_ERROR_INVALID_ARGUMENT);
-    VerifyOrReturnError(mStagingNetwork.UsingPDC(), CHIP_ERROR_INVALID_ARGUMENT);
-    return CopySpanToMutableSpan(ByteSpan(mStagingNetwork.networkIdentity, mStagingNetwork.networkIdentityLen), outNetworkIdentity);
+    VerifyOrReturnError(!mStagingNetworks[networkIndex].Empty() && networkIndex < kMaxNetworks, CHIP_ERROR_INVALID_ARGUMENT);
+    VerifyOrReturnError(mStagingNetworks[networkIndex].UsingPDC(), CHIP_ERROR_INVALID_ARGUMENT);
+    return CopySpanToMutableSpan(ByteSpan(mStagingNetworks.networkIdentity, mStagingNetworks.networkIdentityLen),
```
bzbarsky-apple, line 556: I don't understand how this compiles.  What is `mStagingNetworks.networkIdentity`?

src/platform/Linux/NetworkCommissioningWiFiDriver.cpp
```diff
@@ -333,24 +551,25 @@ CHIP_ERROR LinuxWiFiDriver::AddOrUpdateNetworkWithPDC(ByteSpan ssid, ByteSpan ne
...
 }
 
 CHIP_ERROR LinuxWiFiDriver::GetClientIdentity(uint8_t networkIndex, MutableByteSpan & outClientIdentity)
 {
-    VerifyOrReturnError(!mStagingNetwork.Empty() && networkIndex == 0, CHIP_ERROR_INVALID_ARGUMENT);
-    VerifyOrReturnError(mStagingNetwork.UsingPDC(), CHIP_ERROR_INVALID_ARGUMENT);
-    return CopySpanToMutableSpan(ByteSpan(mStagingNetwork.clientIdentity, mStagingNetwork.clientIdentityLen), outClientIdentity);
+    VerifyOrReturnError(!mStagingNetworks[networkIndex].Empty() && networkIndex < kMaxNetworks, CHIP_ERROR_INVALID_ARGUMENT);
+    VerifyOrReturnError(mStagingNetworks[networkIndex].UsingPDC(), CHIP_ERROR_INVALID_ARGUMENT);
+    return CopySpanToMutableSpan(ByteSpan(mStagingNetworks.clientIdentity, mStagingNetworks.clientIdentityLen), outClientIdentity);
```
bzbarsky-apple, line 564: Again, how does this compile?

src/platform/Linux/NetworkCommissioningWiFiDriver.cpp
```diff
@@ -333,24 +551,25 @@ CHIP_ERROR LinuxWiFiDriver::AddOrUpdateNetworkWithPDC(ByteSpan ssid, ByteSpan ne
...
 
 CHIP_ERROR LinuxWiFiDriver::SignWithClientIdentity(uint8_t networkIndex, const ByteSpan & message,
                                                    P256ECDSASignature & outSignature)
 {
-    VerifyOrReturnError(!mStagingNetwork.Empty() && networkIndex == 0, CHIP_ERROR_INVALID_ARGUMENT);
-    VerifyOrReturnError(mStagingNetwork.UsingPDC(), CHIP_ERROR_INVALID_ARGUMENT);
-    return mStagingNetwork.clientIdentityKeypair->ECDSA_sign_msg(message.data(), message.size(), outSignature);
+    VerifyOrReturnError(!mStagingNetworks[networkIndex].Empty() && networkIndex < kMaxNetworks, CHIP_ERROR_INVALID_ARGUMENT);
+    VerifyOrReturnError(mStagingNetworks[networkIndex].UsingPDC(), CHIP_ERROR_INVALID_ARGUMENT);
+    return mStagingNetworks.clientIdentityKeypair->ECDSA_sign_msg(message.data(), message.size(), outSignature);
```
bzbarsky-apple, line 572: And this.  How does this compile?

src/platform/Linux/NetworkCommissioningDriver.h
```diff
@@ -148,8 +153,16 @@ class LinuxWiFiDriver final : public WiFiDriver
...
-    WiFiNetwork mStagingNetwork;
+    // Number of entries in mStagingNetworks
+    uint8_t mCurrentNetId = 0;
+    WiFiNetwork mSavedNetworks[kMaxNetworks];
+    WiFiNetwork mStagingNetworks[kMaxNetworks];
+    ConnectCallback * mpConnectCallback;
+    // Index location of the network the device is connected to
+    int8_t mStagedConnectedNetworkIndex = -1;
+    int8_t mSavedConnectedNetworkIndex  = -1;
+    bool StartReorderingEntries(uint8_t index, int8_t foundNetworkAtIndex);
```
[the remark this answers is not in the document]
Kshitijjain21: Yes will update the documentation. Wanted to keep these function private as these would not be exposed outside of LinuxWiFiDriver class.

src/platform/Linux/NetworkCommissioningWiFiDriver.cpp
Kshitijjain21: Yeah the function can return mCurrentNetid instead of looping over mStagingNetworks to find non empty networks.

src/platform/Linux/NetworkCommissioningWiFiDriver.cpp
Kshitijjain21: Yes the assumption is that the keyName[size - 1] is 0 or null terminated. We can add a static assert but when adding the static assert in the function the function is no longer a consteval. However we could add a static assert for every individual keyname before calling `createNetworkStorageKeys` . Eg. 
`constexpr char ssidKeyName[SSID_KEY_NAME_LEN] = "wifi-ssid";
static_assert(ssidKeyName[SSID_KEY_NAME_LEN - 1] == 0, "ssidKeyName not null terminated");`
Thoughts on having it a static assert for every key name?

src/platform/Linux/NetworkCommissioningWiFiDriver.cpp
Kshitijjain21: Yes will make more sense to only go upto `mCurrentNetId`

src/platform/Linux/NetworkCommissioningWiFiDriver.cpp
Kshitijjain21: Will add an explanation of the function and whats it tries to achieve.

src/platform/Linux/NetworkCommissioningWiFiDriver.cpp
Kshitijjain21: These functions are included with the config option `CHIP_DEVICE_CONFIG_ENABLE_WIFI_PDC`. I think that the builds might not be using this config option so this bit of code never get compiled.

src/platform/Linux/NetworkCommissioningWiFiDriver.cpp
Kshitijjain21: Update the comment `Return out of range if the index value is greater than mStagingNetworkCount i.e the number of enties in mStagingNetworks`

src/platform/Linux/NetworkCommissioningDriver.h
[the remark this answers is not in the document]
bzbarsky-apple: They can be private without being under the members.

src/platform/Linux/NetworkCommissioningWiFiDriver.cpp
bzbarsky-apple: Adding asserts for every key name does not really make sense, in that it tells us nothing about what's seen inside the function that makes the assumption.

I guess we can just document the assumption, and make this function consteval once we switch to c++20....

src/platform/Linux/NetworkCommissioningWiFiDriver.cpp
bzbarsky-apple: Deleting stale entries would make sense, yes.

src/platform/Linux/NetworkCommissioningWiFiDriver.cpp
Kshitijjain21: Added the explanation to the NetworkCommissioningDriver.h file

andy31415:
This is an old PR with changes requested. Assuming stale and closing.

Kshitijjain21:
> This is an old PR with changes requested. Assuming stale and closing.

@andy31415 The change will help to unblock a couple of test that require 4 networks to be stored. Happy to open it again, feel like a good change to have to unblock a couple of tests and switch between networks.

andy31415:
@Kshitijjain21 this change has not been updated for many months and has a status of "changes requested" from Boris.
What is the current state for it? Is it reviewable/workable? Are you able to actively maintain it and see it go through? If so, we can re-open.

Kshitijjain21:
> @Kshitijjain21 this change has not been updated for many months and has a status of "changes requested" from Boris. What is the current state for it? Is it reviewable/workable? Are you able to actively maintain it and see it go through? If so, we can re-open.

Apologies about keeping this change in PR for long. I  will reopen this again and actively monitor it to see it go through. I checked the changes with the resent master and it's in workable state.

PR #32699, opened 2024-03-25
[NXP][k32w1] Remove the need for extra DAC private key conversion binary

andrei-menzopol:
The application will check at initialization whether the DAC private key has been
converted or not and convert it if needed. There is no need for an extra binary.


src/platform/nxp/k32w/k32w1/FactoryDataProviderImpl.cpp
```diff
@@ -118,18 +159,27 @@ CHIP_ERROR FactoryDataProviderImpl::SSS_Sign(uint8_t * digest, Crypto::P256ECDSA
...
-    ReturnErrorOnFailure(SSS_ExportBlob(blob, &blobSize, offset));
+    error = SSS_ExportBlob(blob, &blobSize, offset, convNeeded);
+    if (!convNeeded)
+    {
+        ChipLogError(DeviceLayer, "SSS: DAC private key already converted to blob");
+        chip::Platform::MemoryFree(data);
+        return CHIP_NO_ERROR;
+    }
+
+    ReturnErrorOnFailure(error);
```
arkq, line 182: I would suggest doing error checking before using values returned by a function. In your case it will work, but it looks like a potential bug.

andrei-menzopol: Thank you. I missed that

andy31415 (approved):

andy31415:
fast track platform specific changes

andrei-menzopol:
Thank you

PR #32701, opened 2024-03-25
[silabs] Add systemView for 917

jepenven-silabs:
System view was not implemented for the 917 GN build.

add it. Also add Matter specific configuration for SystemView since GSDK config isn't over-writable and SystemView buffer size are **WAY** too small for Matter.


examples/platform/silabs/sl_systemview_config.h
```diff
@@ -0,0 +1,80 @@
+/***************************************************************************/ /**
```
jmartinez-silabs, line 1: can you fix the comment banners 
the csa restyler doesn't like the silabs style at the end `*/ /**`

jmartinez-silabs (approved):

examples/platform/silabs/sl_systemview_config.h
jepenven-silabs: done

andy31415 (approved):

PR #32709, opened 2024-03-26
[Linux] Dispatch BlueZ signals in BluezObjectManager

arkq:
### Problem

BlueZ signals dispatching is done in `BluezEndpoint` and `ChipDeviceScanner`. This can be simplified by moving BlueZ management related logic to the `BluezObjectManager` class. Also, this way we can handle (in the future) adapter added/removed events.

### Changes

- Dispatch BlueZ signals in BluezObjectManager
- Notification facility for BLE adapter added/removed for future use (it will be used in the follow-up PR)

### Testing

Locally tested BLE commissioning on Linux with `chip-tool` and lighting app.

### DIFF
--- a/src/platform/Linux/CHIPDevicePlatformEvent.h
+++ b/src/platform/Linux/CHIPDevicePlatformEvent.h
@@ -45,6 +45,8 @@ enum PublicPlatformSpecificEventTypes
 enum InternalPlatformSpecificEventTypes
 {
     kPlatformLinuxEvent = kRange_InternalPlatformSpecific,
+    kPlatformLinuxBLEAdapterAdded,
+    kPlatformLinuxBLEAdapterRemoved,
     kPlatformLinuxBLECentralConnected,
     kPlatformLinuxBLECentralConnectFailed,
     kPlatformLinuxBLEWriteComplete,
@@ -67,6 +69,11 @@ struct ChipDevicePlatformEvent
 {
     union
     {
+        struct
+        {
+            unsigned int mAdapterId;
+            char mAdapterAddress[18];
+        } BLEAdapter;
         struct
         {
             BLE_CONNECTION_OBJECT mConnection;
--- a/src/platform/Linux/BLEManagerImpl.cpp
+++ b/src/platform/Linux/BLEManagerImpl.cpp
@@ -37,6 +37,7 @@
 
 #include <ble/BleError.h>
 #include <ble/CHIPBleServiceData.h>
+#include <lib/support/CHIPMemString.h>
 #include <lib/support/CodeUtils.h>
 #include <lib/support/SafeInt.h>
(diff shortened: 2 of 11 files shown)


yunhanw-google (approved):

src/platform/Linux/CHIPDevicePlatformEvent.h
```diff
@@ -45,6 +45,8 @@ enum PublicPlatformSpecificEventTypes
 enum InternalPlatformSpecificEventTypes
 {
     kPlatformLinuxEvent = kRange_InternalPlatformSpecific,
+    kPlatformLinuxBLEAdapterAdded,
```
yunhanw-google, line 48: could you replace the enum with the enum class, then specify uint8_t?
this way, we can reduce the scope leak here. thanks

src/platform/Linux/CHIPDevicePlatformEvent.h
```diff
@@ -67,6 +69,11 @@ struct ChipDevicePlatformEvent
 {
     union
     {
+        struct
+        {
+            unsigned int mAdapterId;
```
yunhanw-google, line 74: init this value.

src/platform/Linux/CHIPDevicePlatformEvent.h
arkq: I've been thinking about this but as a follow-up. There are more enums which could be changed to `enum class`. Also, for consistency this should be done in all platforms (as of now all platforms have `enum InternalPlatformSpecificEventTypes`). What do you think about another PR instead of adding complexity to this one?

src/platform/Linux/CHIPDevicePlatformEvent.h
arkq: This value is initialized here: https://github.com/arkq/connectedhomeip/blob/bluez-object-manager-signals/src/platform/Linux/BLEManagerImpl.cpp#L816

It's not possible to inline initialize this structure. It has to be trivial since it's used in the `ChipDeviceEvent`.

src/platform/Linux/CHIPDevicePlatformEvent.h
yunhanw-google: sure, could you file a ticket on it for follow-up? thanks

src/platform/Linux/CHIPDevicePlatformEvent.h
yunhanw-google: thanks

src/platform/Linux/CHIPDevicePlatformEvent.h
arkq: Issue for follow-up PR: https://github.com/project-chip/connectedhomeip/issues/32755

jmartinez-silabs (approved):

PR #32718, opened 2024-03-26
Python scripting: Add a test plan generator script

cecille:


ReneJosefsen (approved):

src/python_testing/test_plan_table_generator.py
```diff
@@ -0,0 +1,62 @@
...
+    test_class = getattr(module, classname)
+    config = generate_mobly_test_config(MatterTestConfig())
+    test_instance = test_class(config)
+    steps = test_instance.get_test_steps(test)
+    indent = 6
+    header_num = f'{"**#**":<{indent}}'
+    header_num_step = f'|{header_num} |*TestStep*   '
+    s = ('[cols="5%,45%,45%"]\n'
+         '|===\n'
+         f'{header_num_step}|*Expected Outcome*\n')
```
jamesharrow, line 48: @cecille I was considering in the existing test plans (in .adoc format manually written) - have 
<html>
<body>

# | Ref | PICS | Test Step | Expected Outcome
-- | -- | -- | -- | --



</body>
</html>

Whereas all of the python test scripts currently have a combine description / Expected outcome:

e.g.:
``` def steps_TC_EEM_2_1(self) -> list[TestStep]:
        steps = [
            TestStep("1", "Commissioning, already done", is_commissioning=True),
            TestStep("2", "TH reads Accuracy attribute. Verify that the DUT response contains a MeasurementAccuracyStruct value."),
            TestStep("3", "TH reads CumulativeEnergyImported attribute. Verify that the DUT response contains either null or an EnergyMeasurementStruct value."),
            TestStep("4", "TH reads CumulativeEnergyExported attribute. Verify that the DUT response contains either null or an EnergyMeasurementStruct value."),
            TestStep("5", "TH reads PeriodicEnergyImported attribute. Verify that the DUT response contains either null or an EnergyMeasurementStruct value."),
            TestStep("6", "TH reads PeriodicEnergyExported attribute. Verify that the DUT response contains either null or an EnergyMeasurementStruct value."),
            TestStep("7", "TH reads CumulativeEnergyReset attribute. Verify that the DUT response contains either null or an CumulativeEnergyResetStruct value."),
        ]
```
        
I was considering raising a PR to split the TestStep() class to have:
        Step (number), Test Step, ExpectedOutcome.
        
e.g. we then have:
``` 
         TestStep("3", "TH reads CumulativeEnergyImported attribute.", 
                       "Verify that the DUT response contains either null or an EnergyMeasurementStruct value."),`
```
[One positive side effect would be that the line length in python might not be scrolling so far off to the right and slightly easier to code as you can read more of it on the screen without scrolling left and right!]         
Then the code would autogen 3 columns instead of 2. Not sure how important that is.         
This is okay as is though - the content is there just in 1 big string.

jamesharrow (approved):

src/python_testing/test_plan_table_generator.py
cecille: That piece already landed, and I did it that way per your suggestion. 
See https://github.com/project-chip/connectedhomeip/blob/master/src/python_testing/matter_testing_support.py#L656

If you look at the line below here, that's the expectations. 

So now the new test plans are:
# | Test Step | Expected Outcome

The test step text comes from TestStep.description, the expected outcome comes from TestStep.expectation.

This script doesn't work for any tests with PICS, but that's ok because we don't need 'em anyway.

We have a couple new staff members that just started, and they're going to help with this transition .
issue here to move the test plans with no TestSteps to the new format: https://github.com/project-chip/matter-test-scripts/issues/231
Issue here to remove the pics from the energy stuff: https://github.com/project-chip/matter-test-scripts/issues/225

Didn't think to open an issue to move the existing TestStep impls to use the expectation as well, but that's also a good thought.

cecille: New issue to update current files:
https://github.com/project-chip/matter-test-scripts/issues/240

Before we do all that, it's probably worthwhile to see if I can get this script landed and make the test plan generation process a bit more automated. This cuts down on a BIT of the work, but it would be nice to just have the test plans update along with the script directly.

jamesharrow: Looks good to me!

arkq (approved):

src/python_testing/test_plan_table_generator.py
```diff
@@ -0,0 +1,62 @@
...
+    s = multiline.split('\n')
+    s = [(num_spaces * ' ' + line.lstrip()).rstrip() for line in s]
+    return '\n'.join(s).lstrip()
+
+
+@click.command()
+@click.argument('filename', type=click.Path(exists=True))
+@click.argument('classname', type=str)
+@click.argument('test', type=str)
+def main(filename, classname, test):
```
tcarmelveilleux, line 37: Documentation about what this script does is too sparse. Suggest adding more documentation about what this tool does.

tcarmelveilleux (approved):

src/python_testing/test_plan_table_generator.py
```diff
@@ -0,0 +1,62 @@
...
+    module = importlib.import_module(Path(os.path.basename(filename)).stem)
+    test_class = getattr(module, classname)
+    config = generate_mobly_test_config(MatterTestConfig())
+    test_instance = test_class(config)
+    steps = test_instance.get_test_steps(test)
```
tcarmelveilleux, line 42: Please make a warning happen if this list is missing. Also, an outer `try`/`except` to deal with failures to get the bits/pieces with a better message than the probably cryptic exception that will arise if any of the methods are not found via reflection.

src/python_testing/test_plan_table_generator.py
cecille: Added desc - this will now print on the help menu.

src/python_testing/test_plan_table_generator.py
cecille: done.

PR #32728, opened 2024-03-26
Remove unused public api

andy31415:
Some APIs declared in af.h were only ever used in attribute-storage.cpp

Move them into attribute-storage.cpp and mark them static (they are placed in an anonymous namespace).


src/app/util/attribute-storage.cpp
```diff
@@ -845,25 +870,6 @@ const EmberAfCluster * emberAfFindClusterIncludingDisabledEndpoints(EndpointId e
     return nullptr;
 }
 
-static uint16_t findIndexFromEndpoint(EndpointId endpoint, bool ignoreDisabledEndpoints)
```
cecille, line 848: isn't this used by the other index from endpoint functions?

cecille (approved):

yunhanw-google (approved):

bzbarsky-apple (approved):

PR #32730, opened 2024-03-26
Re-add Matter*Callback with STRONG linkage

andy31415:
This is to cause link errors if anyone relies on these functions to ever work because they do not anymore.

This was feedback from merging #32673 - making this a clear error that is seen at compile time rather than a silent "my callback that was previously invoked is not invoked anymore".

### DIFF
--- a/src/app/util/MatterCallbacks.cpp
+++ b/src/app/util/MatterCallbacks.cpp
@@ -54,4 +54,4 @@ CHIP_ERROR MatterPreCommandReceivedCallback(const chip::app::ConcreteCommandPath
 }
 void MatterPostCommandReceivedCallback(const chip::app::ConcreteCommandPath & commandPath,
                                        const chip::Access::SubjectDescriptor & subjectDescriptor)
-{} // namespace chip
+{}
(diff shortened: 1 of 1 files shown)


src/app/util/MatterCallbacks.cpp
```diff
@@ -34,3 +34,24 @@ DataModelCallbacks * DataModelCallbacks::SetInstance(DataModelCallbacks * newIns
...
+void MatterPreAttributeWriteCallback(const chip::app::ConcreteAttributePath & attributePath) {}
+void MatterPostAttributeWriteCallback(const chip::app::ConcreteAttributePath & attributePath) {}
+CHIP_ERROR MatterPreCommandReceivedCallback(const chip::app::ConcreteCommandPath & commandPath,
+                                            const chip::Access::SubjectDescriptor & subjectDescriptor)
+{
+    return CHIP_NO_ERROR;
+}
+void MatterPostCommandReceivedCallback(const chip::app::ConcreteCommandPath & commandPath,
+                                       const chip::Access::SubjectDescriptor & subjectDescriptor)
+{} // namespace chip
```
bzbarsky-apple, line 57:
```suggestion
{}
```

bzbarsky-apple (approved):

yunhanw-google (approved):

PR #32740, opened 2024-03-26
Add domain names matching the DnssdServices stored in Browse Context

### LINKED ISSUE #32739: Add domain names to the browse context corresponding to the Dnssdservice so it can be passed to the Resolve that is called afterwards
Currently there is no way to pass the domain returned as a result of the browse to the resolve call . The browse context needs to store the domain names so that we can achieve that.

nivi-apple:
This is needed to pass the domain returned from a call to Browse to the Resolve.

Fixes: #32739

### DIFF
--- a/src/platform/Darwin/DnssdImpl.h
+++ b/src/platform/Darwin/DnssdImpl.h
@@ -168,6 +168,7 @@ struct BrowseContext : public BrowseHandler
 {
     DnssdBrowseCallback callback;
     std::vector<DnssdService> services;
+    std::vector<std::pair<DnssdService, std::string>> domainNames;
 
     BrowseContext(void * cbContext, DnssdBrowseCallback cb, DnssdServiceProtocol cbContextProtocol);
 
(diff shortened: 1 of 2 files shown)


src/platform/Darwin/DnssdImpl.h
```diff
@@ -168,6 +168,7 @@ struct BrowseContext : public BrowseHandler
 {
     DnssdBrowseCallback callback;
     std::vector<DnssdService> services;
+    std::vector<std::pair<DnssdService, std::string>> domainNames;
```
bzbarsky-apple, line 171: This doesn't make sense.  We already have a vector of DnssdService.   Why are we having another vector of the same DnssdService?

What we talked about was having `services` be a vector of either std::pair<DnssdService, std::string>, or a vector of some struct that has DnssdService and std::string members, or a vector of some struct that inherits from DnssdService and has the std::string member in addition to that, no?

Or having just a vector of std::string that matches the vector of DnssdService, if we are creating Span<DnssdService> from the `services` vector....

bzbarsky-apple (changes requested):

src/platform/Darwin/DnssdImpl.h
nivi-apple: I fixed this already
. Have one vector of std::vector<std::pair<DnssdService, std::string>> services.

bzbarsky-apple (approved):

bzbarsky-apple:

PR #32748, opened 2024-03-27
Darwin CI: use macos-13 runners and other tweaks

ksperling-apple:
According to github docs these have 4 CPUs (vs 3 for macos-12 == latest). In practice compile steps seem to be about 7% faster.

Also use the pre-installed Python (3.12), remove the "aggregation" step, and run crash log collection after failure (not success).

Tweak bootstrap caching to save the cache right after bootstrap, so it doesn't matter if the build makes changes to the environment (like messing with zap-cli). Also allow us to split caches between GitHub and BuildJet (GitHub provides insight into what is actually cached but only 10G of storage).

### DIFF
--- a/.github/actions/bootstrap/action.yaml
+++ b/.github/actions/bootstrap/action.yaml
@@ -12,49 +12,66 @@ inputs:
 outputs:
   cache-hit:
     description: "Bootstrap environment was restored from cache"
-    value: ${{ fromJSON(steps.bootstrap-cache.outputs.outputs).cache-hit }} # retry returns all outputs in `outputs`
+    value: ${{ fromJSON(steps.restore.outputs.outputs).cache-hit }} # dynamic action returns all outputs in `outputs`
 
 runs:
   using: "composite"
   steps:
-    - name: Calculate bootstrap cache key
-      id: bootstrap-cache-key
+    - name: Determine bootstrap cache configuration
+      id: prepare
       shell: bash
       run: |
-        # Calculate bootstrap cache key
+        # Determine bootstrap cache configuration
         # In addition to the various setup files, the work directory matters as well,
         # because the bootstrapped Pigweed environment contains absolute paths.
+        echo "Calculating bootstrap cache key for '$PWD'"
         FILES_HASH="${{ hashFiles('scripts/setup/*', 'third_party/pigweed/**') }}"
         FINAL_HASH="$(echo "$PWD:$FILES_HASH" | shasum -a 256 | cut -d' ' -f1)"
-        echo "Calculated bootstrap cache key for '$PWD': $FINAL_HASH"
-        echo "hash=$FINAL_HASH" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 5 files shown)


andy31415 (dismissed):

arkq (approved):

andy31415 (approved):

.github/workflows/darwin.yaml
```diff
@@ -48,30 +48,28 @@ jobs:
...
               with:
                 platform: darwin
                 bootstrap-log-name: bootstrap-logs-framework-${{ matrix.options.flavor }}
             - name: Block zap-cli from being used
+              env:
+                PW_ENVSETUP_NO_BANNER: 1
               run: |
                 # Framework builds are NOT expected to require zap-cli
-                scripts/run_in_build_env.sh 'D=$(dirname $(which zap-cli)) && mv $D/zap-cli $D/zap-cli.moved'
+                scripts/run_in_build_env.sh 'rm -- "$(which zap-cli)"'
```
andy31415, line 61: @ksperling-apple this was made in this way to be able to undo the move, otherwise we had the cache saver save something without zap-cli.

is this safe?

andy31415: saw the comment and code that we save right after bootstrap. Seems ok then.

.github/actions/bootstrap/action.yaml
```diff
@@ -12,49 +12,66 @@ inputs:
...
         FILES_HASH="${{ hashFiles('scripts/setup/*', 'third_party/pigweed/**') }}"
         FINAL_HASH="$(echo "$PWD:$FILES_HASH" | shasum -a 256 | cut -d' ' -f1)"
-        echo "Calculated bootstrap cache key for '$PWD': $FINAL_HASH"
-        echo "hash=$FINAL_HASH" >> "$GITHUB_OUTPUT"
+        echo key="${RUNNER_OS}-${RUNNER_ARCH}-${{ inputs.platform }}-${FINAL_HASH}" | tee -a "$GITHUB_OUTPUT"
 
-    - uses: Wandalen/wretry.action@v1.4.10
-      name: Bootstrap from cache
-      id: bootstrap-cache
+        # Split caches across backends
```
bzbarsky-apple, line 32: This could use comments explaining _why_ we are doing that.   Because I don't understand why we're doing that.

.github/workflows/darwin.yaml
```diff
@@ -48,30 +48,28 @@ jobs:
...
               run: xcodebuild -target "Matter" ${{ matrix.options.arguments }}
-            # Now restore zap-cli, so that bootstrap caching does not cache the state when it's been renamed.
-            - name: Make zap-cli work again
-              run: scripts/run_in_build_env.sh 'D=$(dirname $(which zap-cli.moved)) && mv $D/zap-cli.moved $D/zap-cli'
 
     tests:
         name: Run framework tests
         if: github.actor != 'restyled-io[bot]'
-        runs-on: macos-latest
+        needs: [ framework ] # serialize to avoid running to many parallel macos runners
```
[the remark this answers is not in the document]
ksperling-apple: This is somewhat experimental... will see if this makes the overall queueing behavior worse or better.

PR #26718, opened 2023-05-22
Added Operational Node discovery to the mDNS resolver.

andersesbensen:
Added Interface for starting a general Operational Discovery. This is used for enumerating commissioned matter devices on the fabric. As the Operational discovery is a forever ongoing process, adding support for multiple delegates was also needed.

I also considered implementing this in the ResolverProxy, but honestly, I think implementing this could make the Resolver Proxy Obsolete.

PR #28301, opened 2023-07-26
[dnssd] Add Resolver::StartBrowse/StopBrowse/ResolveNode/NoneNameResolutionNoLongerNeeded

vivien-apple:
#### Problem

There is not backend independent API for browsing continuously for commissionables or operational nodes.

Darwin already has such mechanism but it seems like it is needed for other platforms or at least other mdns backend (minimal). This PR add methods to the `Resolver` class such that different backends can implement the same feature, and update the darwin code (in `chip-tool` and in `Matter.framework`) to use this API.

If someone wants to use such API for `minimal` for example, the `MinMdnsResolver::StartBrowse*`, `MinMdnsResolver::StopBrowse`, `MinMdnsResolver::ResolveNode` and `MinMdnsResolver::NodeNameResolutionNoLongerNeeded` are the methods to implement, which likely requires some change in the minimal source code.

This PR is still a WIP since it has been built on top of #27940 which has not landed yet and I also need to double check the Darwin implementation of `Resolver::ResolveNode` and `Resolver::NodeNameResolutionNoLongerNeeded` especially since there is currently no callback when resolving commissionable fails.

PR #32750, opened 2024-03-27
[Dns-sd] Added Operational Discovery and made browse delegate common

su-shanka:
Added support for operational node browse/discovery. Also, updated existing browse/discover interface in Resolver class to a common interface for all types of node browse and made the delegates common for all browse(commissionable, commissioner and operational)

This is continuation of work in #26718.

### DIFF
--- a/src/lib/dnssd/IncrementalResolve.cpp
+++ b/src/lib/dnssd/IncrementalResolve.cpp
@@ -349,10 +349,10 @@ CHIP_ERROR IncrementalResolver::Take(DiscoveredNodeData & outputData)
 
     IPAddressSorter::Sort(mCommonResolutionData.ipAddress, mCommonResolutionData.numIPs, mCommonResolutionData.interfaceId);
 
-    //Set nodeData with data specific to CommissionNodeData
+    // Set nodeData with data specific to CommissionNodeData
     outputData.Set<CommissionNodeData>(mSpecificResolutionData.Get<CommissionNodeData>());
 
-    //set the common resolution data to nodeData
+    // set the common resolution data to nodeData
     CommonResolutionData & resolutionData = outputData.Get<CommissionNodeData>();
     resolutionData                        = mCommonResolutionData;
 
--- a/src/lib/dnssd/IncrementalResolve.cpp
+++ b/src/lib/dnssd/IncrementalResolve.cpp
@@ -345,14 +345,14 @@ CHIP_ERROR IncrementalResolver::OnIpAddress(Inet::InterfaceId interface, const I
 }
 
 template <class NodeType>
-CHIP_ERROR IncrementalResolver::Set(DiscoveredNodeData &output)
+CHIP_ERROR IncrementalResolver::Set(DiscoveredNodeData & output)
 {
     VerifyOrReturnError(mSpecificResolutionData.Is<NodeType>(), CHIP_ERROR_INCORRECT_STATE);
     output.Set<NodeType>();
-    auto & nodeData = output.Get<NodeType>();
-    nodeData = mSpecificResolutionData.Get<NodeType>();
+    auto & nodeData                       = output.Get<NodeType>();
+    nodeData                              = mSpecificResolutionData.Get<NodeType>();
     CommonResolutionData & resolutionData = nodeData;
-    resolutionData = mCommonResolutionData;
+    resolutionData                        = mCommonResolutionData;
     return CHIP_NO_ERROR;
 }
 
@@ -361,7 +361,6 @@ CHIP_ERROR IncrementalResolver::Take(DiscoveredNodeData & outputData)
     VerifyOrReturnError(IsActive(), CHIP_ERROR_INCORRECT_STATE);
 
     IPAddressSorter::Sort(mCommonResolutionData.ipAddress, mCommonResolutionData.numIPs, mCommonResolutionData.interfaceId);
-
 
     if (IsActiveCommissionParse())
     {
--- a/src/lib/dnssd/IncrementalResolve.cpp
+++ b/src/lib/dnssd/IncrementalResolve.cpp
@@ -344,7 +344,7 @@ CHIP_ERROR IncrementalResolver::OnIpAddress(Inet::InterfaceId interface, const I
     return CHIP_NO_ERROR;
 }
 
-template <class NodeType> 
+template <class NodeType>
 CHIP_ERROR IncrementalResolver::Set(DiscoveredNodeData &output)
 {
     VerifyOrReturnError(mSpecificResolutionData.Is<NodeType>(), CHIP_ERROR_INCORRECT_STATE);
@@ -362,7 +362,7 @@ CHIP_ERROR IncrementalResolver::Take(DiscoveredNodeData & outputData)
 
     IPAddressSorter::Sort(mCommonResolutionData.ipAddress, mCommonResolutionData.numIPs, mCommonResolutionData.interfaceId);
 
-    
+
     if (IsActiveCommissionParse())
     {
         ReturnErrorOnFailure(Set<CommissionNodeData>(outputData));
--- a/src/lib/dnssd/IncrementalResolve.cpp
+++ b/src/lib/dnssd/IncrementalResolve.cpp
@@ -349,10 +349,12 @@ CHIP_ERROR IncrementalResolver::Take(DiscoveredNodeData & outputData)
 
     IPAddressSorter::Sort(mCommonResolutionData.ipAddress, mCommonResolutionData.numIPs, mCommonResolutionData.interfaceId);
 
-    // Set nodeData with data specific to CommissionNodeData
+    // Set the DiscoveredNodeData with CommissionNodeData info specific to commissionable/commisssioner 
+    // node available in mSpecificResolutionData.
     outputData.Set<CommissionNodeData>(mSpecificResolutionData.Get<CommissionNodeData>());
 
-    // set the common resolution data to nodeData
+    // IncrementalResolver stored CommonResolutionData separately in mCommonResolutionData hence copy the 
+    //  CommonResolutionData info from mCommonResolutionData, to CommissionNodeData within DiscoveredNodeData
     CommonResolutionData & resolutionData = outputData.Get<CommissionNodeData>();
     resolutionData                        = mCommonResolutionData;
 
--- a/src/lib/dnssd/IncrementalResolve.cpp
+++ b/src/lib/dnssd/IncrementalResolve.cpp
@@ -137,7 +137,8 @@ SerializedQNameIterator StoredServerName::Get() const
     return SerializedQNameIterator(BytesRange(mNameBuffer, mNameBuffer + sizeof(mNameBuffer)), mNameBuffer);
 }
 
-CHIP_ERROR IncrementalResolver::InitializeParsing(mdns::Minimal::SerializedQNameIterator name, const mdns::Minimal::SrvRecord & srv)
+CHIP_ERROR IncrementalResolver::InitializeParsing(mdns::Minimal::SerializedQNameIterator name, const uint64_t ttl,
+                                                  const mdns::Minimal::SrvRecord & srv)
 {
     AutoInactiveResetter inactiveReset(*this);
 
@@ -167,6 +168,7 @@ CHIP_ERROR IncrementalResolver::InitializeParsing(mdns::Minimal::SerializedQName
 
     switch (mServiceNameType)
     {
+    /* All known dnssd records for MATTER are filled into DnssdNodeData */
     case ServiceNameType::kOperational:
         mSpecificResolutionData.Set<OperationalNodeData>();
         {
@@ -183,6 +185,7 @@ CHIP_ERROR IncrementalResolver::InitializeParsing(mdns::Minimal::SerializedQName
             {
                 return err;
             }
+            mSpecificResolutionData.Get<OperationalNodeData>().hasZeroTTL = (ttl == 0);
         }
 
         LogFoundOperationalSrvRecord(mSpecificResolutionData.Get<OperationalNodeData>().peerId, mTargetHostName.Get());
@@ -304,7 +307,7 @@ CHIP_ERROR IncrementalResolver::OnTxtRecord(const ResourceData & data, BytesRang
         }
     }
 
-    if (IsActiveBrowseParse())
+    if (IsActiveCommissionParse())
     {
         TxtParser<CommissionNodeData> delegate(mSpecificResolutionData.Get<CommissionNodeData>());
         if (!ParseTxtRecord(data.GetData(), &delegate))
@@ -341,17 +344,37 @@ CHIP_ERROR IncrementalResolver::OnIpAddress(Inet::InterfaceId interface, const I
     return CHIP_NO_ERROR;
 }
 
+template <class NodeType> 
+CHIP_ERROR IncrementalResolver::Set(DiscoveredNodeData &output)
+{
+    VerifyOrReturnError(mSpecificResolutionData.Is<NodeType>(), CHIP_ERROR_INCORRECT_STATE);
+    output.Set<NodeType>();
+    auto & nodeData = output.Get<NodeType>();
+    nodeData = mSpecificResolutionData.Get<NodeType>();
+    CommonResolutionData & resolutionData = nodeData;
+    resolutionData = mCommonResolutionData;
+    return CHIP_NO_ERROR;
+}
+
 CHIP_ERROR IncrementalResolver::Take(DiscoveredNodeData & outputData)
 {
-    VerifyOrReturnError(IsActiveBrowseParse(), CHIP_ERROR_INCORRECT_STATE);
+    VerifyOrReturnError(IsActive(), CHIP_ERROR_INCORRECT_STATE);
 
     IPAddressSorter::Sort(mCommonResolutionData.ipAddress, mCommonResolutionData.numIPs, mCommonResolutionData.interfaceId);
 
-    outputData.Set<CommissionNodeData>();
-    CommissionNodeData & nodeData         = outputData.Get<CommissionNodeData>();
-    nodeData                              = mSpecificResolutionData.Get<CommissionNodeData>();
-    CommonResolutionData & resolutionData = nodeData;
-    resolutionData                        = mCommonResolutionData;
+    
+    if (IsActiveCommissionParse())
+    {
+        ReturnErrorOnFailure(Set<CommissionNodeData>(outputData));
+    }
+    else if (IsActiveOperationalParse())
+    {
+        ReturnErrorOnFailure(Set<OperationalNodeData>(outputData));
+    }
+    else
+    {
+        return CHIP_ERROR_INCORRECT_STATE;
+    }
 
     ResetToInactive();
 
--- a/src/lib/dnssd/IncrementalResolve.cpp
+++ b/src/lib/dnssd/IncrementalResolve.cpp
@@ -168,7 +168,6 @@ CHIP_ERROR IncrementalResolver::InitializeParsing(mdns::Minimal::SerializedQName
 
     switch (mServiceNameType)
     {
-    /* All known dnssd records for MATTER are filled into DnssdNodeData */
     case ServiceNameType::kOperational:
         mSpecificResolutionData.Set<OperationalNodeData>();
         {
@@ -344,36 +343,18 @@ CHIP_ERROR IncrementalResolver::OnIpAddress(Inet::InterfaceId interface, const I
     return CHIP_NO_ERROR;
 }
 
-template <class NodeType>
-CHIP_ERROR IncrementalResolver::Set(DiscoveredNodeData & output)
-{
-    VerifyOrReturnError(mSpecificResolutionData.Is<NodeType>(), CHIP_ERROR_INCORRECT_STATE);
-    output.Set<NodeType>();
-    auto & nodeData                       = output.Get<NodeType>();
-    nodeData                              = mSpecificResolutionData.Get<NodeType>();
-    CommonResolutionData & resolutionData = nodeData;
-    resolutionData                        = mCommonResolutionData;
-    return CHIP_NO_ERROR;
-}
-
 CHIP_ERROR IncrementalResolver::Take(DiscoveredNodeData & outputData)
 {
-    VerifyOrReturnError(IsActive(), CHIP_ERROR_INCORRECT_STATE);
+    VerifyOrReturnError(IsActiveCommissionParse(), CHIP_ERROR_INCORRECT_STATE);
 
     IPAddressSorter::Sort(mCommonResolutionData.ipAddress, mCommonResolutionData.numIPs, mCommonResolutionData.interfaceId);
 
-    if (IsActiveCommissionParse())
-    {
-        ReturnErrorOnFailure(Set<CommissionNodeData>(outputData));
-    }
-    else if (IsActiveOperationalParse())
-    {
-        ReturnErrorOnFailure(Set<OperationalNodeData>(outputData));
-    }
-    else
-    {
-        return CHIP_ERROR_INCORRECT_STATE;
-    }
+    //Set nodeData with data specific to CommissionNodeData
+    outputData.Set<CommissionNodeData>(mSpecificResolutionData.Get<CommissionNodeData>());
+
+    //set the common resolution data to nodeData
+    CommonResolutionData & resolutionData = outputData.Get<CommissionNodeData>();
+    resolutionData                        = mCommonResolutionData;
 
     ResetToInactive();
 
--- a/src/lib/dnssd/IncrementalResolve.cpp
+++ b/src/lib/dnssd/IncrementalResolve.cpp
@@ -349,11 +349,11 @@ CHIP_ERROR IncrementalResolver::Take(DiscoveredNodeData & outputData)
 
     IPAddressSorter::Sort(mCommonResolutionData.ipAddress, mCommonResolutionData.numIPs, mCommonResolutionData.interfaceId);
 
-    // Set the DiscoveredNodeData with CommissionNodeData info specific to commissionable/commisssioner 
+    // Set the DiscoveredNodeData with CommissionNodeData info specific to commissionable/commisssioner
     // node available in mSpecificResolutionData.
     outputData.Set<CommissionNodeData>(mSpecificResolutionData.Get<CommissionNodeData>());
 
-    // IncrementalResolver stored CommonResolutionData separately in mCommonResolutionData hence copy the 
+    // IncrementalResolver stored CommonResolutionData separately in mCommonResolutionData hence copy the
     //  CommonResolutionData info from mCommonResolutionData, to CommissionNodeData within DiscoveredNodeData
     CommonResolutionData & resolutionData = outputData.Get<CommissionNodeData>();
     resolutionData                        = mCommonResolutionData;
--- a/src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
+++ b/src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
@@ -213,7 +213,7 @@ void PacketParser::ParseSRVResource(const ResourceData & data)
             continue;
         }
 
-        CHIP_ERROR err = resolver.InitializeParsing(data.GetName(), srv);
+        CHIP_ERROR err = resolver.InitializeParsing(data.GetName(), data.GetTtlSeconds(), srv);
         if (err != CHIP_NO_ERROR)
         {
             // Receiving records that we do not need to parse is normal:
@@ -370,7 +370,7 @@ void MinMdnsResolver::AdvancePendingResolverStates()
 
         if (missing.Has(IncrementalResolver::RequiredInformationBitFlags::kIpAddress))
         {
-            if (resolver->IsActiveBrowseParse())
+            if (resolver->IsActiveCommissionParse())
             {
                 // Browse wants IP addresses
                 ScheduleIpAddressResolve(resolver->GetTargetHostName());
@@ -399,7 +399,7 @@ void MinMdnsResolver::AdvancePendingResolverStates()
         }
 
         // SUCCESS. Call the delegates
-        if (resolver->IsActiveBrowseParse())
+        if (resolver->IsActiveCommissionParse())
         {
             MATTER_TRACE_SCOPE("Active commissioning delegate call", "MinMdnsResolver");
             DiscoveredNodeData nodeData;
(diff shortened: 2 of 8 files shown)


src/lib/dnssd/Discovery_ImplPlatform.cpp
```diff
@@ -743,6 +743,27 @@ CHIP_ERROR DiscoveryImplPlatform::DiscoverCommissioners(DiscoveryFilter filter,
...
 }
 
+CHIP_ERROR DiscoveryImplPlatform::StartDiscovery(DiscoveryType type, DiscoveryFilter filter, DiscoveryContext & context)
+{
+    CHIP_ERROR err = CHIP_ERROR_INVALID_ARGUMENT;
+
+    switch (type)
+    {
+    case DiscoveryType::kCommissionableNode:
+        err = DiscoverCommissionableNodes(filter, context);
```
andy31415, line 753: let's do `return DiscoverCommissionableNodes(filter, context)` and similar below. Make is clearer and uses up less lines:

```cpp
switch (type)
{
    case DiscoveryType::kCommissionableNode:
        return DiscoverCommissionableNodes(filter, context);
    case DiscoveryType::kCommissionerNode:
        return DiscoverCommissioners(filter, context);
    case DiscoveryType::kOperational:
        return CHIP_ERROR_NOT_IMPLEMENTED;
    default:
}
```

src/lib/dnssd/Types.h
```diff
@@ -204,29 +204,31 @@ inline constexpr size_t kMaxDeviceNameLen         = 32;
...
+    size_t rotatingIdLen                  = 0;
+    uint32_t deviceType                   = 0;
+    uint16_t longDiscriminator            = 0;
+    uint16_t vendorId                     = 0;
+    uint16_t productId                    = 0;
+    uint16_t pairingHint                  = 0;
+    uint8_t commissioningMode             = 0;
+    uint8_t commissionerPasscode          = 0;
+    uint8_t rotatingId[kMaxRotatingIdLen] = {};
+    bool nodeStatus; // true is ttl > 0 false id ttl = 0
```
andy31415, line 220: This needs a better name. how about `hasZeroTTL` (but then condition is inverted) or something similar?

I should not need to read comments to understand what a value means.

andy31415 (approved):

src/lib/dnssd/IncrementalResolve.cpp
```diff
@@ -199,10 +182,11 @@ CHIP_ERROR IncrementalResolver::InitializeParsing(mdns::Minimal::SerializedQName
                 return CHIP_ERROR_INVALID_ARGUMENT;
             }
 
-            Platform::CopyString(mSpecificResolutionData.Get<CommissionNodeData>().instanceName, nameCopy.Value());
+            Platform::CopyString(mSpecificResolutionData.Get<DnssdNodeData>().instanceName, nameCopy.Value());
         }
+        mSpecificResolutionData.Get<DnssdNodeData>().nodeStatus = (ttl == 0) ? false : true;
```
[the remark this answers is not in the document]
su-shanka: Sure.

src/lib/dnssd/Types.h
su-shanka: That makes sense, application can decide how to use the info of ttl being 0. Will update it to `hasZeroTTL`.

src/lib/dnssd/Discovery_ImplPlatform.cpp
su-shanka: updated

bzbarsky-apple (changes requested):
Please break out the "pure rename, no functional changes" bits into a separate PR from the "now we are changing behavior" bits.  That will make it much easier to review both PRs.

su-shanka:
> Please break out the "pure rename, no functional changes" bits into a separate PR from the "now we are changing behavior" bits. That will make it much easier to review both PRs.

Moved the refactor and renaming changes to #32866. Will update this PR once #32866  is closed.

src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
```diff
@@ -418,6 +418,10 @@ void MinMdnsResolver::AdvancePendingResolverStates()
                 mActiveResolves.CompleteCommissionable(nodeData);
                 nodeData.nodeType = DiscoveryType::kCommissionableNode;
                 break;
+            case IncrementalResolver::ServiceNameType::kOperational:
```
bzbarsky-apple, line 421: Is there something tracking adding this to the "platform" dns-sd implementation?

bzbarsky-apple (changes requested):
Requesting changes for now until the issues in the underlying PR are resolved.

src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
su-shanka: Since the platform implementation of operational discovery in #28301 is blocked on some discussion, it is not considered here.

andy31415 (changes requested):
Blocking until https://github.com/project-chip/connectedhomeip/pull/32866/files#r1561854542 is resolved

We need to have resoltion data be "valid data" without a large structure with members that are a union of all possible members.

src/lib/dnssd/IncrementalResolve.cpp
```diff
@@ -201,6 +184,7 @@ CHIP_ERROR IncrementalResolver::InitializeParsing(mdns::Minimal::SerializedQName
 
             Platform::CopyString(mSpecificResolutionData.Get<DnssdNodeData>().instanceName, nameCopy.Value());
         }
+        mSpecificResolutionData.Get<DnssdNodeData>().hasZeroTTL = (ttl == 0) ? false : true;
```
andy31415, line 187: `(condition) ? false : true` is an antipattern (and in this case it seems reversed. it should be just `condition`

su-shanka: updated with proper check

src/lib/dnssd/IncrementalResolve.cpp
```diff
@@ -341,17 +344,36 @@ CHIP_ERROR IncrementalResolver::OnIpAddress(Inet::InterfaceId interface, const I
...
+template <class NodeType>
+CHIP_ERROR IncrementalResolver::Set(DiscoveredNodeData & output)
+{
+    VerifyOrReturnError(mSpecificResolutionData.Is<NodeType>(), CHIP_ERROR_INCORRECT_STATE);
+    output.Set<NodeType>();
+    auto & nodeData                       = output.Get<NodeType>();
+    nodeData                              = mSpecificResolutionData.Get<NodeType>();
+    CommonResolutionData & resolutionData = nodeData;
+    resolutionData                        = mCommonResolutionData;
+    return CHIP_NO_ERROR;
```
andy31415, line 356: There are several intermediate variables above. Could you re-visit the logic and see if it can be cleaned up? 

As it stands, this feels a bit too much magic in casting and operator=. Why do we have this complex of code?

andy31415: At a minimum `output.Set<NodeType>(mSpecificResolutionData.Get<NodeType>())` makes some of the assignments more explicity ... however I seem to get lost in what mSpecificResolutionData vs output are supposed to contain and their difference in types.

andy31415: I see this was copied & pasted from previous code, however previous code was also part of the refactor. It used to be:

```cpp
outputData.resolutionData = mCommonResolutionData;
outputData.nodeData       = mSpecificResolutionData.Get<DnssdNodeData>();
```

however now it seems significantly messier. This needs fixing to be readable.

src/lib/shell/commands/Dns.cpp
```diff
@@ -83,26 +83,47 @@ class DnsShellResolverDelegate : public Dnssd::DiscoverNodeDelegate, public Addr
...
-        Dnssd::CommissionNodeData nodeData = discNodeData.Get<Dnssd::CommissionNodeData>();
+        auto nodeData = discNodeData.Get<Dnssd::CommissionNodeData>();
+
         if (!nodeData.IsValid())
         {
             streamer_printf(streamer_get(), "DNS browse failed - not found valid services \r\n");
             return;
         }
 
+        streamer_printf(streamer_get(), "DNS browse succeeded: \r\n");
```
Damian-Nordic, line 121: Why was this moved?

src/lib/shell/commands/Dns.cpp
```diff
@@ -83,26 +83,47 @@ class DnsShellResolverDelegate : public Dnssd::DiscoverNodeDelegate, public Addr
...
+
+        if (discNodeData.Is<Dnssd::OperationalNodeData>())
         {
-            streamer_printf(streamer_get(), "DNS browse failed - not commission type node \r\n");
+            LogOperationalNodeDiscovered(discNodeData.Get<Dnssd::OperationalNodeData>());
             return;
         }
 
-        Dnssd::CommissionNodeData nodeData = discNodeData.Get<Dnssd::CommissionNodeData>();
+        auto nodeData = discNodeData.Get<Dnssd::CommissionNodeData>();
```
Damian-Nordic, line 113:
```suggestion
        const auto& nodeData = discNodeData.Get<Dnssd::CommissionNodeData>();
```

src/lib/dnssd/IncrementalResolve.cpp
```diff
@@ -167,6 +168,7 @@ CHIP_ERROR IncrementalResolver::InitializeParsing(mdns::Minimal::SerializedQName
 
     switch (mServiceNameType)
     {
+    /* All known dnssd records for MATTER are filled into DnssdNodeData */
```
bzbarsky-apple, line 171: What is DnssdNodeData?

src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
```diff
@@ -452,18 +452,49 @@ void MinMdnsResolver::AdvancePendingResolverStates()
...
+            ResolvedNodeData nodeResolvedData;
+            CHIP_ERROR err = CHIP_ERROR_UNINITIALIZED;
+
+            if (mActiveResolves.HasBrowseFor(chip::Dnssd::DiscoveryType::kCommissionableNode))
+            {
+                DiscoveredNodeData nodeData;
+                err = resolver->Take(nodeData);
+                if (!nodeData.Is<OperationalNodeData>())
+                {
+                    err = CHIP_ERROR_INVALID_ARGUMENT;
```
bzbarsky-apple, line 464: Isn't this losing the `err` that `Take` might have returned?

src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
```diff
@@ -452,18 +452,49 @@ void MinMdnsResolver::AdvancePendingResolverStates()
...
+                {
+                    mDiscoveryContext->OnNodeDiscovered(nodeData);
+                    auto opNodeData                  = nodeData.Get<OperationalNodeData>();
+                    nodeResolvedData.resolutionData  = static_cast<CommonResolutionData>(opNodeData);
+                    nodeResolvedData.operationalData = opNodeData;
+                }
+                else
+                {
+#if CHIP_MINMDNS_HIGH_VERBOSITY
+                    ChipLogError(Discovery, "No delegate to report commissioning node discovery");
```
bzbarsky-apple, line 480: This log does not look correct; we might have a delegate, but `err` could be a failure.

src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
```diff
@@ -452,18 +452,49 @@ void MinMdnsResolver::AdvancePendingResolverStates()
...
+                }
+                if (err != CHIP_NO_ERROR)
+                {
+                    ChipLogError(Discovery, "Failed to take discovery result: %" CHIP_ERROR_FORMAT, err.Format());
+                }
+                if (mDiscoveryContext != nullptr && err == CHIP_NO_ERROR)
+                {
+                    mDiscoveryContext->OnNodeDiscovered(nodeData);
+                    auto opNodeData                  = nodeData.Get<OperationalNodeData>();
+                    nodeResolvedData.resolutionData  = static_cast<CommonResolutionData>(opNodeData);
```
bzbarsky-apple, line 474: Why is this cast needed?

src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
```diff
@@ -452,18 +452,49 @@ void MinMdnsResolver::AdvancePendingResolverStates()
...
             MATTER_TRACE_SCOPE("Active operational delegate call", "MinMdnsResolver");
-            ResolvedNodeData nodeData;
+            ResolvedNodeData nodeResolvedData;
+            CHIP_ERROR err = CHIP_ERROR_UNINITIALIZED;
+
+            if (mActiveResolves.HasBrowseFor(chip::Dnssd::DiscoveryType::kCommissionableNode))
+            {
+                DiscoveredNodeData nodeData;
+                err = resolver->Take(nodeData);
+                if (!nodeData.Is<OperationalNodeData>())
```
bzbarsky-apple, line 462: I don't understand this.  We have a browse for kCommissionableNode; why are we suddenly checking for operational here?

src/lib/dnssd/Types.h
```diff
@@ -194,11 +194,15 @@ struct CommonResolutionData
 };
 
 /// Data that is specific to Operational Discovery of nodes
-struct OperationalNodeData
+struct OperationalNodeData : public CommonResolutionData
```
bzbarsky-apple, line 197: Hmm, so now `ResolvedNodeData` will basically have two copies of `CommonResolutionData`?  One direct, one via `OperationalNodeData`?

That's really confusing as an API consumer.  Do those copies have to be in sync?  If not, which one is the one to use?  If yes, why do we have both?

Seems like there's another intermediate change needed here which is to make this inheritance change and change `ResolvedNodeData` to either just have an `OperationalNodeData` member, or inherit from `OperationalNodeData` and have no data members, or just be an alias for `OperationalNodeData` (and figure out the logging situation).

src/lib/dnssd/Types.h
```diff
@@ -194,11 +194,15 @@ struct CommonResolutionData
...
 {
     PeerId peerId;
-
-    void Reset() { peerId = PeerId(); }
+    bool hasZeroTTL;
```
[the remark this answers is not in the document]
su-shanka: The zero TTL in operational indicates that it has been removed from or left the fabric, hence this helps detect when a operational node joins and leaves the respective fabric.

src/lib/dnssd/IncrementalResolve.cpp
su-shanka: Seems to have sneaked in during auto merge of rebase. Will correct it.

src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
su-shanka: It was supposed to be operational browse check, fixed

src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
su-shanka: Yes, but if there is no OperationalNodeData available then even if `Take` returned success we cannot use it. If the data is of type OperationalNodeData only then the `err` from `Take` has significance. But I guess I should do that type check only is `Take` is successful. I will update the logic a bit.

src/lib/dnssd/Types.h
su-shanka: I was actually wondering if we need to provide resolution data in delegate for Operational Discovery. From my understanding, the idea of Operational Browse is to detect new nodes in fabric, application follow up with a resolution request if resolution data is needed for any particular discovered node. Any thoughts on this?

src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
su-shanka: this part removed as no longer needed.

src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
su-shanka: fixed

src/lib/dnssd/Types.h
su-shanka: As agreed upon, removed inheritance of `CommonResolutionData` from `OperationalNodeData`  as it is not a necessity for Operational Browse.

src/lib/shell/commands/Dns.cpp
```diff
@@ -83,26 +83,47 @@ class DnsShellResolverDelegate : public Dnssd::DiscoverNodeDelegate, public Addr
...
+        streamer_printf(streamer_get(), "DNS browse operational succeeded: \r\n");
+        streamer_printf(streamer_get(), "   Node ID: " ChipLogFormatX64 "-" ChipLogFormatX64 "\r\n",
+                        ChipLogValueX64(nodeData.peerId.GetCompressedFabricId()), ChipLogValueX64(nodeData.peerId.GetNodeId()));
+        streamer_printf(streamer_get(), "   hasZeroTTL: %s\r\n", nodeData.hasZeroTTL ? "true" : "false");
+    }
+
     void OnNodeDiscovered(const Dnssd::DiscoveredNodeData & discNodeData) override
     {
-        if (!discNodeData.Is<Dnssd::CommissionNodeData>())
+        if (!discNodeData.Is<Dnssd::CommissionNodeData>() && !discNodeData.Is<Dnssd::OperationalNodeData>())
```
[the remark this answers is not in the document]
su-shanka: removed.

src/lib/shell/commands/Dns.cpp
su-shanka: done

src/lib/shell/commands/Dns.cpp
su-shanka: I moved it to ensure to have common code for both node types, which was reverted later on, missed to revert this moved.
Reverted it now.

src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
su-shanka: this logic was no longer needed with removal of inheritance. removed it.

src/lib/dnssd/IncrementalResolve.cpp
su-shanka: changed this a bit.

src/lib/dnssd/IncrementalResolve.cpp
su-shanka: removed the comment.

src/lib/dnssd/IncrementalResolve.cpp
```diff
@@ -343,14 +345,15 @@ CHIP_ERROR IncrementalResolver::OnIpAddress(Inet::InterfaceId interface, const I
...
     IPAddressSorter::Sort(mCommonResolutionData.ipAddress, mCommonResolutionData.numIPs, mCommonResolutionData.interfaceId);
 
-    outputData.Set<CommissionNodeData>();
-    CommissionNodeData & nodeData         = outputData.Get<CommissionNodeData>();
-    nodeData                              = mSpecificResolutionData.Get<CommissionNodeData>();
-    CommonResolutionData & resolutionData = nodeData;
+    // Set nodeData with data specific to CommissionNodeData
+    outputData.Set<CommissionNodeData>(mSpecificResolutionData.Get<CommissionNodeData>());
+
+    // set the common resolution data to nodeData
```
bzbarsky-apple, line 355: I don't understand this comment.  What is `nodeData` here?

In general, this "assign to the same struct twice" needs some clear documentation about the assumptions and ordering here.  

As a followup, the fact that we are keeping `mCommonResolutionData` separately from `mSpecificResolutionData.Get<CommissionNodeData>()` which _also_ holds the common resolution data is pretty confusing and should be fixed.  But not in this PR.

src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
```diff
@@ -452,18 +453,46 @@ void MinMdnsResolver::AdvancePendingResolverStates()
...
+                    opNodeData.hasZeroTTL = nodeResolvedData.operationalData.hasZeroTTL;
+                    nodeData.Set<OperationalNodeBrowseData>(opNodeData);
+                }
+                if (err != CHIP_NO_ERROR)
+                {
+                    ChipLogError(Discovery, "Failed to take discovery result: %" CHIP_ERROR_FORMAT, err.Format());
+                }
+                else if (mDiscoveryContext != nullptr)
+                {
+                    mDiscoveryContext->OnNodeDiscovered(nodeData);
```
bzbarsky-apple, line 477: I do have to ask: why are we not just setting up `DiscoveredNodeData` in this branch?  It's not used anywhere else...

src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
```diff
@@ -452,18 +453,46 @@ void MinMdnsResolver::AdvancePendingResolverStates()
         else if (resolver->IsActiveOperationalParse())
         {
             MATTER_TRACE_SCOPE("Active operational delegate call", "MinMdnsResolver");
-            ResolvedNodeData nodeData;
+            ResolvedNodeData nodeResolvedData;
+            CHIP_ERROR err = CHIP_ERROR_UNINITIALIZED;
+
+            err = resolver->Take(nodeResolvedData);
```
bzbarsky-apple, line 459:
```suggestion

            CHIP_ERROR err = resolver->Take(nodeResolvedData);
```

src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
```diff
@@ -452,18 +453,46 @@ void MinMdnsResolver::AdvancePendingResolverStates()
...
+                    ChipLogError(Discovery, "No delegate to report operational node discovery");
+#endif
+                }
+            }
 
-            CHIP_ERROR err = resolver->Take(nodeData);
             if (err != CHIP_NO_ERROR)
             {
-                ChipLogError(Discovery, "Failed to take discovery result: %" CHIP_ERROR_FORMAT, err.Format());
+                ChipLogError(Discovery, "Failed to take ResolvedNodeData - result: %" CHIP_ERROR_FORMAT, err.Format());
```
bzbarsky-apple, line 489: So if Take() fails we will log the error twice if `HasBrowseFor(chip::Dnssd::DiscoveryType::kOperational)`?  Seem confusing.

src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
```diff
@@ -452,18 +453,46 @@ void MinMdnsResolver::AdvancePendingResolverStates()
...
             }
 
-            mActiveResolves.Complete(nodeData.operationalData.peerId);
+            mActiveResolves.Complete(nodeResolvedData.operationalData.peerId);
```
bzbarsky-apple, line 492: It's really weird to be doing this if err is not success, but it predates your PR.  Please file a followup issue on this use of `nodeResolvedData` when it might contain arbitrary garbage.

src/lib/dnssd/Types.h
```diff
@@ -197,10 +197,22 @@ struct CommonResolutionData
...
     void Reset() { peerId = PeerId(); }
 };
 
+struct OperationalNodeBrowseData : public OperationalNodeData
+{
+    OperationalNodeBrowseData() { Reset(); };
+    void LogDetail() const
+    {
+        ChipLogDetail(Discovery, "Discovered Operational node:\r\n");
+        ChipLogDetail(Discovery, "\tNode ID: " ChipLogFormatX64 "-" ChipLogFormatX64 "\r\n",
```
bzbarsky-apple, line 210: This is really not the "Node ID"... Maybe "Node instance:"?

But also, this seems to be re-inventing ChipLogFormatPeerId/ChipLogValuePeerId.

src/lib/shell/commands/Dns.cpp
```diff
@@ -83,15 +83,24 @@ class DnsShellResolverDelegate : public Dnssd::DiscoverNodeDelegate, public Addr
 
     AddressResolve::NodeLookupHandle & Handle() { return mSelfHandle; }
 
+    void LogOperationalNodeDiscovered(const Dnssd::OperationalNodeBrowseData & nodeData)
+    {
+        streamer_printf(streamer_get(), "DNS browse operational succeeded: \r\n");
+        streamer_printf(streamer_get(), "   Node ID: " ChipLogFormatX64 "-" ChipLogFormatX64 "\r\n",
```
bzbarsky-apple, line 89: See comments above.

src/lib/dnssd/IncrementalResolve.cpp
su-shanka: Added more descriptive comment to explain the changes done.

src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
su-shanka: it was kept out as per previous logic, it no longer need to be branched that way. I have moved it to `err != CHIP_NO_ERROR` branch

src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
su-shanka: done

src/lib/dnssd/Types.h
su-shanka: renamed and used ChipLogFormatPeerId+ChipLogValuePeerId for logging.

src/lib/shell/commands/Dns.cpp
su-shanka: Renamed `Node ID` but this going streamer and not chip logging, so cannot use `ChipLogFormatPeerId` and `ChipLogValuePeerId` here

src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
su-shanka: thought it would be good idea to have separate error logs for browse/discovery and resolution.

src/lib/dnssd/IncrementalResolve.cpp
bzbarsky-apple: Much better, thank you!

src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
```diff
@@ -452,18 +453,42 @@ void MinMdnsResolver::AdvancePendingResolverStates()
...
+                    ChipLogError(Discovery, "Failed to take discovery result: %" CHIP_ERROR_FORMAT, err.Format());
+                }
+                else if (mDiscoveryContext != nullptr)
+                {
+                    DiscoveredNodeData nodeData;
+                    OperationalNodeBrowseData opNodeData;
+
+                    opNodeData.peerId     = nodeResolvedData.operationalData.peerId;
+                    opNodeData.hasZeroTTL = nodeResolvedData.operationalData.hasZeroTTL;
+                    nodeData.Set<OperationalNodeBrowseData>(opNodeData);
```
bzbarsky-apple, line 472: I wish we could just Set<>() with no args and then fill it in in-place instead of copying the struct, but if we can't this is fine of course.

src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
bzbarsky-apple: I think my point is that we will log both `Failed to take discovery result` and `Failed to take ResolvedNodeData` when `DiscoveryType::kOperational`.

Maybe this second `if (err != CHIP_NO_ERROR)` should just be `else if` so it does not get hit when `DiscoveryType::kOperational`?

src/lib/shell/commands/Dns.cpp
bzbarsky-apple: This seems to still be using "Node ID".

But also, I don't understand the "needs to go into streamer" distinction: `ChipLogFormatPeerId` is a macro that expands out exactly to `ChipLogFormatX64 "-" ChipLogFormatX64`, no?

bzbarsky-apple (approved):

src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
```diff
@@ -363,14 +363,15 @@ void MinMdnsResolver::AdvancePendingResolverStates()
     {
         if (!resolver->IsActive())
         {
+            ChipLogError(Discovery, "resolver inactive, continue to next");
```
Damian-Nordic, line 366: It's a normal situation in the case of parallel queries, not an error, right?

src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
Damian-Nordic: I suggest having a single log just after `Take` is executed. Also, after rebasing to https://github.com/project-chip/connectedhomeip/commit/881676de173e779b9471eaae944c33e5780b038c, the code will continue to the next record in such a case, which it should do.

src/lib/dnssd/Types.h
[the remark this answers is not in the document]
Damian-Nordic: OK, my point is that receiving records with TTL 0 makes sense for continuous queries, while for one-shot queries, which the common Matter's DNS-SD layer uses (except for Darwin that implements a custom interface for continuous queries - see their custom `ChipDnssdBrowse` interface), you rather not receive removed records at all than receive those with TTL 0.

Thus, I'm curious in which configuration do you test that.

src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
su-shanka: removed.

src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
su-shanka: Moved log just after Take() as suggested.

src/lib/shell/commands/Dns.cpp
su-shanka: My bad, I got confused. I have updated.

src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
su-shanka: Andy had a comment to use Set<> with args in another place, hence followed the same here.

src/lib/dnssd/Types.h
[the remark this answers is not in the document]
su-shanka: I didn't realize this. I was expecting continuous queries from matter's dnssd. Now that I tested for a longer duration I see that we stop querying after certain retries. However, this TTL 0 info is added for continuous queries.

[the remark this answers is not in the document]
Damian-Nordic: Resolving for now.

Damian-Nordic (approved):

andy31415 (approved):

Issue #31317, opened 2024-01-09
[Tooling] automated running of python tests in CI


cecille:
### Reproduction steps / Feature

We have a script that goes through all the yamls and runs them automatically. We should probably have something similar for the python tests because right now they need to be manually added.

### Platform

python

Apollon77:
If I could add a wish then it would be cool to allow to specify the started executables comparable to the yaml tooling ... in fact matter.js uses the same tooling to just start a nodejs  scropt as executable instead of the device.

cecille:
This is basically done, will track improvements as individual issues.

PR #32752, opened 2024-03-27
Added scripts to get metadata (runner arguments) associated with the python_testing scripts for automating the process 

vatsalghelani-csa:
Work for Issue #31317  

PR #3 - We have a Metadata_Reader object that will be initialized by passing the path to the environment file. The public method of this class can be used to get metadata associated with a script.

Files Worked:
- BUILD.gn
- scripts/tests/py/BUILD.gn
- scripts/tests/py/env_test.yaml
- scripts/tests/py/metadata.py
- scripts/tests/py/test_metadata.py
- scripts/tests/py/__init__.py
- scripts/tests/py/setup.py
- scripts/tests/py/setup.cfg
- scripts/tests/py/pyproject.toml


scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,160 @@
...
+    app: str = None
+    factoryreset: bool = False
+    comissioning_method: str = None
+    discriminator: int = None
+    kvs: str = None
+    storage_path: str = None
+    on_network_commission: str = None
+    passcode: int = None
+    trace_to: str = None
+    hex_arg: str = None
```
andy31415, line 34: hex_arg/bool_arg/int_arg seem off here.

Could we start with a reduced subset and add these as needed as we start updating tests?

This is a very long list right now and I am not sure we need it all:

- int/bool/hex arg are questionalbe: should probably not be here or should have different naming
- I believe `trace_to` are hardcoded and the same for every app. We could avoid having these.

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,160 @@
...
+    manual_code: str = None
+    bool_arg: str = None
+    PICS: str = None
+    tests: str = None
+    int_arg: str = None
+    trace_to_appjson: str = None
+    trace_to_testjson: str = None
+    trace_toperfetto: str = None
+
+    def print(self):
```
andy31415, line 45: this method does not seem needed: it replaces a one-line with another one-line.

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+    # builds the environment object 
+    def __build_env_object__(self, env_file_name):
+        
+        with open(env_file_name) as stream:
+            try:
+                env = yaml.safe_load(stream)
+            except yaml.YAMLError as yaml_exception:
+                raise yaml_exception
```
andy31415, line 66: you do `except e: raise e` which seems to be a noop. Remove the entire try/catch.

andy31415: This should probably be:

```py
with open(env_file_name) as stream:
    return yaml.safe_load(stream)
```

scripts/tests/py/test_metadata.py
```diff
@@ -0,0 +1,41 @@
...
+    #defines expected arguments associated with runs
+    def setUp(self):
+
+        # build the reader object
+        self.reader=Metadata_Reader("/src/python_testing", "/scripts/tests/py/env.yaml")
+        self.actual_runs_metadata = self.reader.get_runs_metadata()
+        self.expected_runs_metadata = {}
+
+        self.expected_runs_metadata["/src/python_testing/TC_RR_1_1.py+run1"] = Metadata(app="out/linux-x64-all-clusters-ipv6only-no-ble-no-wifi-tsan-clang-test/chip-all-clusters-app",
+                   discriminator=1234, py_script_path="/src/python_testing/TC_RR_1_1.py", run="run1", passcode=20202021)
```
andy31415, line 31: Please use test files for testing, not real tests. There should be separate test-only files (and maybe don't name them py, just txt for now) that validates different variants of items.

src/python_testing/TC_RR_1_1.py
```diff
@@ -13,7 +12,12 @@
 #    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
-#
+
+
+# run arguments:
+# test-runner-runs: run1
+# test-runner-run/run1/app: app/all-clusters discriminator KVS trace-to-appjson
+# test-runner-run/run1/script: storage-path commissioning-method discriminator passcode trace-to-testjson trace-to-testperfetto
```
andy31415, line 20: do not change this file yet. For testing purposes, create other files that have fixed content. That way we do not depend on tests never changing and we can test different variations without requiring tests to be changed/updated.

scripts/tests/py/test_metadata.py
```diff
@@ -0,0 +1,41 @@
...
+        
+    # gets the metadata associated with each run by parsing the test script and compares it to the expected metadata values
+    def test_case1(self):
```
andy31415, line 34: please name these in some different way ... `case1` is non-descriptive

scripts/tests/py/test_metadata.py
andy31415: These loads should be part of test case runs and not part of the setup. At most, the setup should prepare the reader.

scripts/tests/py/env.yaml
```diff
@@ -0,0 +1,32 @@
+# Copyright (c) 2023 Project CHIP Authors
```
andy31415, line 1: The file should probably be named "test_env.yaml" as it is for testing only.

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,160 @@
...
+
+from dataclasses import dataclass
+from pathlib import Path
+import re
+import yaml
+
+
+@dataclass
+class Metadata:
+    py_script_path: str = None
```
andy31415, line 23: given that these are initialized with None, all the types here are `Option[...]`

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,160 @@
...
+    # This class will be initialized with
+    # the name of the script folder and the environment
+    # file name
+    
+    def __init__(self, script_folder, env_file_name):
+        self.script_folder = script_folder
+        self.env = self.__build_env_object__(env_file_name)
+
+
+    # builds the environment object
```
andy31415, line 59: A method named "build_env_object" obviously `Builds the environment object`.
Please update comments for:
  - use python doc comments (i.e. strings after the declaration of the method)
  - add info in them that is not just re-stating the signature. For example reading this signature I could write `This function builds the environment object. It operates on self object and takes in the environment file name as an argument` which feels like a lot however it is 100% filler text without any help to the code reader.

for documentation you could have things like:
  - file name is a path to a YAML file (maybe the name should be env_yaml_file_path)
  - describe what an environment object is or where to find documentation about it

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,160 @@
...
+        with open(env_file_name) as stream:
+            try:
+                env = yaml.safe_load(stream)
+            except yaml.YAMLError as yaml_exception:
+                raise yaml_exception
+
+        return env
+
+    # resolves the run arguments associated with the environment  
+    def __resolve_env_vals__(self, metadata_dict):
```
andy31415, line 71: could you add types for both input and output for all methods? Is this `metadata_dict: Dict[str,str]` ? what does this return?

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,160 @@
...
+            if run_arg_val == None:
+                continue
+
+            sub_args = run_arg_val.split('/')
+                        
+            if len(sub_args)==1:
+                run_arg_val=self.env.get(sub_args[0])
+                
+            elif len(sub_args)==2:
+                run_arg_val=self.env.get(sub_args[0]).get(sub_args[1])
```
andy31415, line 90: we need some error checking ... what if len is 0 or 3 or 100 ?

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,160 @@
...
+                env = yaml.safe_load(stream)
+            except yaml.YAMLError as yaml_exception:
+                raise yaml_exception
+
+        return env
+
+    # resolves the run arguments associated with the environment  
+    def __resolve_env_vals__(self, metadata_dict):
+
+        for run_arg in metadata_dict:
```
andy31415, line 73: is this `for run_arg, run_arg_val in metadata_dict.items():` ?

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,160 @@
...
+
+            run_arg_val = metadata_dict[run_arg]
+
+            if not type(run_arg_val)==str or run_arg=="run":
+                metadata_dict[run_arg]=run_arg_val
+                continue
+            
+            if run_arg_val == None:
```
andy31415, line 81: use `is None` for this

scripts/tests/py/BUILD.gn
```diff
@@ -0,0 +1,37 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import("//build_overrides/build.gni")
+import("//build_overrides/chip.gni")
+
+import("//build_overrides/pigweed.gni")
+import("$dir_pw_build/python.gni")
+
+pw_python_package("metadata_parser") {
```
andy31415, line 21: This has to be referenced somewhere like `BUILD.gn`, like a `scripts/tests/py:metadata_parser.tests` to execute unit tests. Otherwise this does not get checked/executed.

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,160 @@
...
+                run_arg_val = True
+
+            if not self.__is_run_arg_valid__(run_arg_val):
+                raise Exception(str(run_arg_val)+" is not a valid value for "+str(run_arg))
+
+            metadata_dict[run_arg] = run_arg_val
+
+
+    # determines if the defined run arguments are valid
+    def __is_run_arg_valid__(self, run_arg_val):
```
andy31415, line 105: if this is not defined, please remove it for now. If we need it later, we can start adding it. Building infra in advance usually results in extra complex code.

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,160 @@
...
+            
+                runs_match = runs_def_ptrn.match(line.strip())
+                args_match = args_def_ptrn.match(line.strip())
+            
+                if runs_match:
+                    for run in runs_match.group(1).strip().split():
+                        runs_arg_lines[run]=[]
+
+                elif args_match:
+                    runs_arg_lines[args_match.group(1)].append(args_match.group(3))
```
andy31415, line 128: this loses the 2nd argument i.e. script or app. Why? If we do not need that part, how about not making it part of the syntax?

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,160 @@
...
+
+        for run in runs_arg_lines:
+            metadata = Metadata()
+            metadata_dict = vars(metadata)
+            metadata_dict['py_script_path'] = str(py_script_path)
+            metadata_dict['run'] = str(run)
+            self.read_args(runs_arg_lines[run], metadata_dict)
+            self.__resolve_env_vals__(metadata_dict)
+            runs_metadata[str(metadata.py_script_path)+"+"+str(metadata.run)] = metadata
```
andy31415, line 137: concatenation of keys seems off. Why not creating a dict inside a dict if we want a tree structure or a list of runs inside a dict?

Some description of formats may help here.

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,160 @@
...
+    trace_to_appjson: str = None
+    trace_to_testjson: str = None
+    trace_toperfetto: str = None
+
+    def print(self):
+        print(repr(self))
+
+
+class Metadata_Reader:
+    # This class will be initialized with
```
andy31415, line 50: Please use python docs (i.e. strings, usually multiline).

An example of usage would help.

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,160 @@
...
+    # the name of the script folder and the environment
+    # file name
+    
+    def __init__(self, script_folder, env_file_name):
+        self.script_folder = script_folder
```
andy31415, line 55: I think this tries too much: it both reads and parses metadata and tries to be aware of a `script folder` that it recurses over.

lets split: make a reader that is only capable to read metadata one by one. leave the "multiple script parsing" and "aggregating all scripts" as a concern for some separate piece of code.

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,160 @@
...
+            metadata = Metadata()
+            metadata_dict = vars(metadata)
+            metadata_dict['py_script_path'] = str(py_script_path)
+            metadata_dict['run'] = str(run)
+            self.read_args(runs_arg_lines[run], metadata_dict)
+            self.__resolve_env_vals__(metadata_dict)
+            runs_metadata[str(metadata.py_script_path)+"+"+str(metadata.run)] = metadata
+
+
+    # gets the run metadata associated with all the test scripts in a particular folder
```
andy31415, line 140: remove this capability from this class. it should be sufficient for now to have the ability to parse a single script. please expose that capability only.

I expect a `parse_metadata(script_path) -> List[Metadata]`

BUILD.gn
```diff
@@ -232,6 +232,7 @@ if (current_toolchain != "${dir_pw_toolchain}/default:default") {
         "//examples/chef:chef.tests",
         "//scripts/build:build_examples.tests",
         "//scripts/py_matter_idl:matter_idl.tests",
+        "//scripts/tests/py:py.tests",
```
andy31415, line 235: this is `//scripts/tests/py:metadata_parser.tests` based on your naming

src/python_testing/TC_RR_1_1.py
vatsalghelani-csa: Yes, I reverted back the changes to the original TC_RR_1_1.py script. Created test script TC_RR_1_1_Test.py in the same /scripts/tests/py folder

scripts/tests/py/TC_RR_1_1_Test.py
```diff
@@ -0,0 +1,880 @@
+#
+#    Copyright (c) 2022 Project CHIP Authors
```
andy31415, line 2: this should not be a copy of the full TC_RR_1_1. We want to focus on the content of run arguments.

this should be a 5-10 line file (i doubt we even need copyright blurb for test data) and not be a full 100s of lines file.

please delete all lines after line 24 and you can remove the license blurb too. This should be a minimal file, the same as we would have for test data.

We could even use make our parsers so that we can pass in raw strings (so tests have the strings in them)

andy31415: Naming should also not be `TC_RR_....`. It should be something that explains the test, like `simple_run_args.py` or `simple_run_args.txt` or similar.

scripts/tests/py/BUILD.gn
```diff
@@ -0,0 +1,45 @@
...
+  inputs = [
+    "env_test.yaml",
+    "simple_run_args.txt",
+  ]
+
+  sources = [
+    "__init__.py",
+    "metadata.py",
+    "env_test.yaml",
+    "simple_run_args.txt",
```
andy31415, line 37: are the yaml and txt file sources? This feels off ... I see these already exists in inputs.

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,223 @@
+#!/usr/local/bin/python3
+# Copyright (c) 2022 Project CHIP Authors
```
andy31415, line 2: nit: (across all files) we could do copyright 2024. The others have 2023 which felt off, however 2022 is really old.

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,223 @@
...
+from typing import Dict
+from typing import List
+from typing import Union
+
+@dataclass
+class Metadata:
+    py_script_path: Optional[str] = None
+    run: Optional[str] = None
+    app: Optional[str] = None
+    factoryreset: Optional[bool] = False
```
andy31415, line 29: since this has a default, it seems we do not need Optional on it. this seems to just be a bool.

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+class Metadata_Reader:
+    """
+    A class to parse run arguments from the test scripts and 
+    resolve them to environment specific values.
+
+    Attributes:
+    
+    env: str
```
andy31415, line 49: needs renames probably: env_file_name is the argument name and it represents a path rather than a dictionary.

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,223 @@
...
+    storage_path: Optional[str] = None
+    on_network_commission: Optional[str] = None
+    passcode: Optional[int] = None
+    endpoint: Optional[str] = None
+    manual_code: Optional[str] = None
+    PICS: Optional[str] = None
+    tests: Optional[str] = None
+
+
+class Metadata_Reader:
```
andy31415, line 42:
```suggestion
class MetadataReader:
```

Maybe as naming, I think we do not often use underscores and CamelCase is sufficient

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,223 @@
...
+        assocaited with the "all_cluster" variable and set the value for "app" option
+        to this raw value.
+
+        Parameter:
+        
+        metadata_dict: Dict[str, str]
+          Dictionary where each key represent a particular argument and its value represent
+          the value for that argument defined in the test script.
+
+        Return:
```
andy31415, line 98: remove this bit: the `Return: none` uses up 3 lines of text and it is obvious from the signature.

please try to make the code concise so that is easier to read.

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,223 @@
...
+            if run_arg_val is None:
+                run_arg_val = True
+
+            metadata_dict[run_arg] = run_arg_val
+
+                
+            
+    def __read_args__(self,run_args_lines: List[str],metadata_dict: Dict[str, str]) -> None:
+        """
+        Goes these each raw line in argument header of a test script and extracts argument
```
andy31415, line 141:
```suggestion
        Parses a list of lines and extracts argument
```

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,223 @@
...
+        values from it.
+
+        Parameters:
+        
+        run_args_lines: List[str]
```
andy31415, line 146: somehow these feel overspecified: I can read the types from the signature, I do not need them duplicated here.

Also please remove the return none bits.

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,223 @@
...
+          Dictionary where the extracted arguments will be stored.
+          This represents the side effect of this function.
+
+        Return:
+        None
+
+        """
+        for run_line in run_args_lines:
+            for run_arg_word in run_line.strip().split():
+                run_arg=run_arg_word.split('/',1)[0]
```
andy31415, line 159: Some examples of expected format would help here.

scripts/tests/py/metadata.py
andy31415: It is not clear here what a 'raw line' is

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,223 @@
...
+                if runs_match:
+                    for run in runs_match.group(1).strip().split():
+                        runs_arg_lines[run]=[]
+
+                elif args_match:
+                    runs_arg_lines[args_match.group(1)].append(args_match.group(2))
+
+        for run in runs_arg_lines:
+            metadata = Metadata()
+            metadata_dict = vars(metadata)
```
andy31415, line 205: this feels off: we lose any typesafety here and start operating on dictionaries. Please update the code to use metadata directly: we should enforce data format and not swith to fully dynamic code.

Somewhere in the codebase there should be a check agains expectet string names and metadata variable names.

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,223 @@
...
+                run_arg_val=self.env.get(sub_args[0])
+                
+            elif len(sub_args)==2:
+                run_arg_val=self.env.get(sub_args[0]).get(sub_args[1])
+
+            # if a argument has been specified in the comment header
+            # but can't be found in the env file, consider it to be
+            # boolean value.
+            if run_arg_val is None:
+                run_arg_val = True
```
andy31415, line 133: the type of metadata_dict is `Dict[str, str]`. This assigns a value of `bool` instead of `str`.

andy31415: Overall I think this should not take a Dict, instead please make it take in a Metadata and specifically change individual items.

scripts/tests/py/test_metadata.py
```diff
@@ -0,0 +1,44 @@
...
+    def setUp(self):
+
+        # build the reader object
+        self.reader=Metadata_Reader(path.join(path.dirname(__file__),"env_test.yaml"))
+            
+
+    def test_parse_single_run(self):
+
+        expected_runs_metadata = {}
+        path_under_test=path.join(path.dirname(__file__),"simple_run_args.txt")
```
andy31415, line 32: nit (maybe for future iterations): indirection of tests via a separate file is somewhat inconvenient: we should have more unit tests and if on every test I need to open the txt file it is hard to manage.

it would be easier to read something like:

```py

actual = parse_metadata("""
# not a real text
# runner arguments:
# test-runner-runs: run1
# test-runner-run/run1: app/all-clusters discriminator KVS storage-path commissioning-method discriminator passcode
more
txt
""")

self.assert_eq(
    actual,
    [
        Metadata(
            app="out/linux-x64-all-clusters-ipv6only-no-ble-no-wifi-tsan-clang-test/chip-all-clusters-app",
             discriminator=1234,
             py_script_path=path_under_test,
             run="run1",
             passcode=20202021
        )
    ]
)
```

scripts/tests/py/test_metadata.py
```diff
@@ -0,0 +1,44 @@
...
+        
+        expected_runs_metadata[path_under_test] = Metadata(app="out/linux-x64-all-clusters-ipv6only-no-ble-no-wifi-tsan-clang-test/chip-all-clusters-app",
```
andy31415, line 34: why are you building a dict? You should have a list.

scripts/tests/py/test_metadata.py
```diff
@@ -0,0 +1,44 @@
...
+                   discriminator=1234, py_script_path=path_under_test, run="run1", passcode=20202021)
+        
+                
+        for run in expected_runs_metadata:
```
andy31415, line 38: you should assertEqual on lists directly (or if order-independent then you could make them sets ... however for now lists seem ok)

scripts/tests/py/test_metadata.py
andy31415: You can either make your parser also accept txt objects besides file paths (seems trivial) or make some helper method that writes a text to a temporary file and parses that (then you don't need to have special worry about how to include txt files inside test builds)

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,225 @@
...
+
+class MetadataReader:
+    """
+    A class to parse run arguments from the test scripts and 
+    resolve them to environment specific values.
+
+    Attributes:
+    
+    env: str
+       A dictionary that reprsents a environment configuration in
```
andy31415, line 76:
```suggestion
       A dictionary that represents a environment configuration in
```

Also above, given that the type is "dictionary" the declaration of `:str` seems wrong.
Do we need these attributes to be public?

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,225 @@
...
+        values from it.
+
+        Parameters:
+        
+        run_args_lines:
+          Line in test script header that contains run argumment definition
+
+        metadata_dict:
+          Dictionary where the extracted arguments will be stored.
+          This represents the side effect of this function.
```
andy31415, line 154: why do we pass in these dictionaries? Why is the function not `__read_args__(self, lines) -> Dict[str,str]`

The format of the dictionary should be described: some example inputs and dictionary content. Right now I do not know what "run_args_lines` look like (there is a impl comment that says they should be name or name/val) or what the dictionary contains. What is the behaviour on duplicate lines?

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,225 @@
...
+                args_match = args_def_ptrn.match(line.strip())
+            
+                if runs_match:
+                    for run in runs_match.group(1).strip().split():
+                        runs_arg_lines[run]=[]
+
+                elif args_match:
+                    runs_arg_lines[args_match.group(1)].append(args_match.group(2))
+
+        for run in runs_arg_lines:
```
andy31415, line 210: nit: you could use `for run, line in run_arg_lines.items()` since you seem to need both key and values.

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,225 @@
...
+            metadata_dict = {}
+            self.__read_args__(runs_arg_lines[run], metadata_dict)
```
andy31415, line 212: this and the line above seems to indicate it would be more convenient to not pass in a metadata dict to read_args. lets make it `metadata_dict = self.__read_args(run_arg_lines[run])`

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,225 @@
...
+
+        metadata_dict:
+          Dictionary where each key represent a particular argument and its value represent
+          the value for that argument defined in the test script.
+        """
+
+        for run_arg, run_arg_val in metadata_dict.items():
+
+            if not type(run_arg_val) == str or run_arg == "run":
+                metadata_dict[run_arg] = run_arg_val
```
andy31415, line 104: why do we have this assignment? is this not a noop given that you get the arg and value ... and this just sets the arg and value again?

andy31415: A comment would be helpful here, to explain why non-strings and runs are skipped.

scripts/tests/py/test_metadata.py
```diff
@@ -0,0 +1,48 @@
...
+class TestMetadataReader(unittest.TestCase):
+    path_under_test = path_under_test = path.join(path.dirname(__file__), "simple_run_args.txt")
+
+    def setUp(self):
+
+        # build the reader object
+        self.reader = MetadataReader(path.join(path.dirname(__file__), "env_test.yaml"))
+        with open(self.path_under_test, 'w', encoding='utf8') as test_file:
+            test_file.writelines(
+                ["# test-runner-runs: run1", "\n# test-runner-run/run1: app/all-clusters discriminator KVS storage-path commissioning-method discriminator passcode"])
```
andy31415, line 31: This is hard to read. Use multiline strings

test_file.write('''
# test-runner-runs: run1
# test-runner-run/run1: app/all-clusters discriminator KVS storage-path commissioning-method discriminator passcode

non-parsed string
''')

scripts/tests/py/metadata.py
```diff
@@ -0,0 +1,226 @@
...
+                continue
+
+            if run_arg_val is None:
+                continue
+
+            sub_args = run_arg_val.split('/')
+
+            if len(sub_args) not in [1, 2]:
```
andy31415, line 104: key: a/b   => env['a']['b]
key: a  => env['a']
key:   => True


factoryreset
something

 # test-runner-runs: run1 
# test-runner-run/run1: app/all-clusters discriminator KVS storage-path factoryreset

scripts/tests/py/test_metadata.py
```diff
@@ -0,0 +1,51 @@
...
+    def setUp(self):
+
+        # build the reader object
+        self.reader = MetadataReader(path.join(path.dirname(__file__), "env_test.yaml"))
+        with open(self.path_under_test, 'w', encoding='utf8') as test_file:
+            test_file.writelines(''' 
+            # test-runner-runs: run1 
+            # test-runner-run/run1: app/all-clusters discriminator KVS storage-path commissioning-method discriminator passcode
+            ''')
+
```
andy31415, line 35:
```py
def assertMetadataParse(file_content: str, expected: List[Metadata]):
    with tempfile.TemporaryFile() as fp:
        fp.write(file_content)
        fp.close()
        actual = self.reader.parse_script(fp.name)
        self.assertEqual(actual, expected)
```

scripts/tests/py/test_metadata.py
```diff
@@ -0,0 +1,51 @@
...
+    def test_parse_single_run(self):
```
andy31415, line 36:
```py
self.assertMetadataParse(''' 
            # test-runner-runs: run1 
            # test-runner-run/run1: app/all-clusters discriminator KVS storage-path commissioning-method discriminator passcode
            ''',
[
Metadata(app="out/linux-x64-all-clusters-ipv6only-no-ble-no-wifi-tsan-clang-test/chip-all-clusters-app",
                                             discriminator=1234, py_script_path=self.path_under_test, run="run1", passcode=20202021)
]
)
```

scripts/tests/py/test_metadata.py
andy31415: Asking for this to not dirty the source tree with a fixed path.

andy31415: I see you have a teardown, so we can move this to a followup (lets make it the next PR immediately)

andy31415 (approved):
Approving for a V1, lets follow up with these updates:
  - use temporary files for tests instead of fixed name
  - `__resolve_env_vals__` change to return a new dict (and fix types: it returns a dict of string + bool instead of just string) instead of in-place modification
  -

PR #32753, opened 2024-03-27
Cloudbuild updates: subset submodule update & bootstrap, remove build-all

andy31415:
- Build-all has not been running for a while now, removing it
- Checkout of the submodule for full builds seems to checkout large things that cloudbuild does not compile. Reduce amount of boostrap/checkout


integrations/cloudbuild/smoke-test.yaml
```diff
@@ -5,14 +5,20 @@ steps:
           - "-c"
           - |
               git config --global --add safe.directory "*"
-              git submodule update --init --recursive
+              python scripts/checkout_submodules.py --shallow --recursive --platform esp32
```
Damian-Nordic, line 8:
```suggestion
              python scripts/checkout_submodules.py --shallow --recursive --platform esp32 nrfconnect silabs linux android
```
nit: `platform` is `nargs='+'` argument so you can init all platforms at once.

Damian-Nordic (approved):

integrations/cloudbuild/smoke-test.yaml
andy31415: I tried it at first and it failed for me. I may have used a wrong syntax ... still trying things out.
Generally it seems to me that cloudbuild is throttled on gecko_sdk checkouts and I am trying to find out what to do...

yunhanw-google (approved):

integrations/cloudbuild/smoke-test.yaml
andy31415: Updated

PR #32754, opened 2024-03-27
[Telink] Peripheral modules optimization & Update builds to docker version 46

s07641069:
**Change overview**

- Added obtaining peripheral HW configuration from device tree.
- Application is building with no matter of actual peripheral hardware.
- Update builds to docker version 44 https://github.com/project-chip/connectedhomeip/pull/32710
- Update builds to docker version 46

Device tree provides available HW resources. Which are provided as arrays for each device class. And accessing to instance of HW device is provided by integer index.
If linked real hw does not exist realtime warning message will be printed.

By default in .dts and linking:
- only available onboard buttons are defined.
- white LED is used for connectivity state indication. (red for Mars board since it's the only LED)
- RGB leds are used for RGB PWM fixture. If no more leds BLUE led is used as indication PWM LED
- Remaining LEDs are used for indication PWM LED and application specific PWM purposes.
- Defaults can be changed by .dts overlaying or overriding linking methods.


arkq:
Would it be possible to split docker update and other changes?

s07641069:
> Would it be possible to split docker update and other changes?

This changes require Telink docker update.
I can revert all docker updates excluding Telink, but usually @andy31415 ask to update images for all platforms.

arkq:
> This changes require Telink docker update.

But can you update docker without changes in the platform code, so the docker update could land in master before other changes?

s07641069:
> > This changes require Telink docker update.
> 
> But can you update docker without changes in the platform code, so the docker update could land in master before other changes?

@arkq It's not possible to update Telink docker without rest sw changes.
I can create separate PR where I will update all builds to docker 44 excluding Telink and in current PR will update only Telink Docker with rest changes. Is this what you mean?

arkq:
> It's not possible to update Telink docker without rest sw changes.

If it's tight coupled than I think that its OK to have it all in this PR.

examples/platform/telink/zephyr_ext/zephyr_pwm_pool.c
arkq, line 1: The whole project is C++ based, so when adding new files maybe create them as *.cpp ?

examples/platform/telink/zephyr_ext/zephyr_key_matrix.c
```diff
@@ -0,0 +1,128 @@
+/*
+ * Copyright (c) 2024 Telink Semiconductor
```
arkq, line 2: I'm not sure whether this is OK to add files without `Copyright (c) 2024 Project CHIP Authors` copyright notice to the matter SDK project...

arkq (changes requested):
Blocking due to copyright concerns. Can someone with a better legal knowledge dispel my doubts?

examples/platform/telink/zephyr_ext/zephyr_pwm_pool.c
s07641069: New Telink platform files which we added written on C.
There no problem to change extension, but I see lots of other vendors files as well as *.c in case if they written on C.
If this is a blocker I will change or we can wait comment reg this from one more reviewer.

examples/platform/telink/zephyr_ext/zephyr_key_matrix.c
tehampson: Is this file copied from somewhere else, or did you author this from scratch? If this is copied, I think you need to retain the old copyright and add the Project CHIP Authors copyright.

Although someone more familiar should confirm

s07641069: @arkq @tehampson 
All new files in this PR were written by Telink employee @andriy-bilynskyy

arkq: Apache 2.0 license should not have the "All rights reserved." phrase. I can see that some files in the project have that phrase, but I think that's actually an error (such license is not apache 2.0 anymore).

s07641069: @arkq Fixed

arkq (approved):

andy31415 (approved):
Approving as platform changes and direct docker image version update.
I could not review 5K worth of line changes.

PR #32760, opened 2024-03-27
[Python] Implement async friendly GetConnectedDevice

agners:
Currently GetConnectedDeviceSync() is blocking e.g. when a new session needs to be created. This is not asyncio friendly as it blocks the whole event loop.

Implement a asyncio friendly variant GetConnectedDevice() which is a co-routine function which can be awaited.

### DIFF
--- a/src/controller/python/chip/ChipDeviceCtrl.py
+++ b/src/controller/python/chip/ChipDeviceCtrl.py
@@ -824,7 +824,7 @@ class ChipDeviceControllerBase():
         if returnDevice.value is None:
             with deviceAvailableCV:
                 timeout = None
-                if (timeoutMs):
+                if timeoutMs is not None:
                     timeout = float(timeoutMs) / 1000
 
                 ret = deviceAvailableCV.wait(timeout)
@@ -878,7 +878,7 @@ class ChipDeviceControllerBase():
 
         # The callback might have been received synchronously (during self._ChipStack.Call()).
         # In that case the Future has already been set it will return immediately
-        if timeoutMs:
+        if timeoutMs is not None:
             timeout = float(timeoutMs) / 1000
             await asyncio.wait_for(future, timeout=timeout)
         else:
--- a/src/controller/python/chip/ChipDeviceCtrl.py
+++ b/src/controller/python/chip/ChipDeviceCtrl.py
@@ -787,8 +787,16 @@ class ChipDeviceControllerBase():
 
         return self._Cluster
 
-    def GetConnectedDeviceSync(self, nodeid, allowPASE=True, timeoutMs: int = None):
-        ''' Returns DeviceProxyWrapper upon success.'''
+    def GetConnectedDeviceSync(self, nodeid, allowPASE: bool = True, timeoutMs: int = None):
+        ''' Gets an OperationalDeviceProxy or CommissioneeDeviceProxy for the specified Node.
+
+        nodeId: Target's Node ID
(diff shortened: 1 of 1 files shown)


src/controller/python/chip/ChipDeviceCtrl.py
```diff
@@ -823,6 +823,56 @@ def deviceAvailable(self, device, err):
...
+                self._returnDevice = c_void_p(None)
+                self._returnErr = None
+                self._event_loop = loop
+                self._future = future
+
+            def _deviceAvailable(self):
+                if self._returnDevice.value is not None:
+                    self._future.set_result(self._returnDevice)
+                else:
+                    self._future.set_exception(self._returnErr.to_exception())
```
andy31415, line 852: for easier debugging, should we handle the case of `obj.deviceAvailable(None, None)` ever being called? In that case _returnErr would be none and we would not have to_exception. Should we add at least a log below in deviceAvailable or make the exception here something like generic `No device but no exception either` ?

src/controller/python/chip/ChipDeviceCtrl.py
```diff
@@ -823,6 +823,56 @@ def deviceAvailable(self, device, err):
...
+
+        closure = DeviceAvailableClosure(eventLoop, future)
+        ctypes.pythonapi.Py_IncRef(ctypes.py_object(closure))
+        self._ChipStack.Call(lambda: self._dmLib.pychip_GetConnectedDeviceByNodeId(
+            self.devCtrl, nodeid, ctypes.py_object(closure), _DeviceAvailableCallback),
+            timeoutMs).raise_on_error()
+
+        # The callback might have been received synchronously (during self._ChipStack.Call()).
+        # In that case the Future has already been set it will return immediately
+        if (timeoutMs):
```
andy31415, line 868: nit: remove brackets. Should we use `is not None` or are we saying 0 and None are equivalent of "no timeout" ? I would generally be more inclined to think 0 is "run and return immediately" (even though maybe that seems useless).

andy31415 (approved):

woody-apple (approved):

src/controller/python/chip/ChipDeviceCtrl.py
agners: As it stands now, the second argument will always be a valid object (generated by `ToPyChipError`). If we ever change the semantics, it would throw a `AttributeError` here indeed, but I don't think that is bad. In fact, it would help debugging quite nicely as it prints a nice stack trace :sweat_smile: I'd prefer to apply offensive programming here, and leave it as is.

src/controller/python/chip/ChipDeviceCtrl.py
agners: > Should we use is not None or are we saying 0 and None are equivalent of "no timeout" ? 

Yes. I followed `GetConnectedDeviceSync` semantic here.

I am fine changing this semantic, but it should be done in `GetConnectedDeviceSync` too. From what I can all users either leave the default or pass a positive value (even the YAML tests). So changing the semantic should be fine, at least for all uses within the Matter SDK.

agners: FWIW, changing this to 
```
        if timeoutMs is not None:
```

Would lead to a Timeout exception in almost all cases when passing 0. It could actually useful to test if/when the statement just above actually kicks in (callback received synchronously during self._ChipStack.Call()).

PR #32763, opened 2024-03-27
Remove TV app from full builds and smoke tests as they have been included in CI as a build example per issue Issue 30906

kliao-csa:


andy31415 (approved):

.github/workflows/full-android.yaml
```diff
@@ -73,30 +73,6 @@ jobs:
                     "./scripts/build/build_examples.py --target android-arm-chip-tool build"
             - name: Clean out build output
               run: rm -rf ./out examples/android/CHIPTool/app/libs/jniLibs/* examples/android/CHIPTool/app/libs/*.jar
-            - name: Build Android arm-tv-casting-app
-              run: |
-                  ./scripts/run_in_build_env.sh \
-                    "./scripts/build/build_examples.py --target android-arm-tv-casting-app build"
```
bzbarsky-apple, line 79: Why is this being removed?  Nothing else builds android-arm-tv-casting-app or android-arm-tv-server....  The "examples" thing mentioned only builds the arm64 versions.

andy31415: This is true ... it is likely worth it to build arm as it is 32-bit so int/long conversions will be different than the 64-bit builds.

kliao-csa:
New context from Yufeng indicates that integrating this smoketest as a standalone build is _not_ what we are looking for; instead we want runtime testing because failures were happening there rather than during builds.

Would be nice to have this context in the issue writeup next time...

I am reverting this for now.

Issue #27575, opened 2023-06-30
[1.2] Time sync - tracking 


cecille:
### Reproduction steps

PR inspection

### Bug prevalence

always

### GitHub hash of the SDK that was being used

abb7a802231f2b615b3030c5df43f5144871e09c

### Platform

core

### Type

Common Cluster Logic

### Anything else?

- [ ]  Figure out what happens if the TZ offset is out of range AND the code fails to load a time zone. See https://github.com/project-chip/connectedhomeip/pull/26082#discussion_r1240951613. In particular, which error code needs to be returned (I think this still needs to be a constraint error), AND can we do this check in a way that prevents making changes that necessitate reloading the time
- [ ] https://github.com/project-chip/connectedhomeip/pull/26082#discussion_r1240951626 - align errors between the ember command parser and the
- [ ] https://github.com/project-chip/connectedhomeip/pull/26082/files#r1212444174 - better overflow protection when going from chip epoch to unix epoch
- [ ] https://github.com/project-chip/connectedhomeip/pull/26082/files#r1247927627 - handlers for 3.5.6.1 SHOULD - If a Node has used the Last Known Good UTC Time, it SHOULD recheck its security materials and existing connections if it later achieves time synchronization.
- [ ] https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242621779 - time zone name copying
- [ ] document event generation on time zone name only change. https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242626060
- [ ] https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242627777 - change event flag from parameter to return
- [ ] https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242630331 - change size comparison
- [ ] follow up on ScheduleDelayedAction function https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242638706
- [ ] https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242689236 - adjust TZ / DST list on time change
- [ ] https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242639975 - propagate error for set or document why the error changes
- [ ] https://github.com/project-chip/connectedhomeip/pull/26082/files#r1241153735 - tighten up time get / set features to move them closer to the receive / send boundary
- [ ] https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242642964 - remove casts on TZ / DST offset
[...]

PR #32772, opened 2024-03-28
Fix issues in 27575

fessehaeve:
Description of issues fixed from #27575 :

- When setting time zone fails, this code was returning the error of LoadTimeZone(), if it failed, before returning ConstraintError. Fixed to return the correct error, ConstraintError, and ignore the error of LoadTimeZone(). https://github.com/project-chip/connectedhomeip/pull/26082#discussion_r1240951613
- Return ConstraintError instead of CHIP_ERROR_IM_MALFORMED_COMMAND_DATA_IB when checking time zone constraints. https://github.com/project-chip/connectedhomeip/pull/26082#discussion_r1240951626
- Improved time zone name copying as proposed by @bzbarsky-apple https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242621779
- generate time zone status event when time zone name changes in every aspect https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242626060
- removed event flag since implementation is changed https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242627777, https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242657708 
- use actual size of buffer instead of the macro for the attribute size when comparing to incoming list size https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242630331, https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242613876
- propagate error when setting TimeSource through attribute store https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242639975
- make sure that the dst list is kept updated or expired items are dropped in every situation https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242655854
- don't encode local time when GetLocalTime errors https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242660231
- use unknown value check for enums https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242667611, https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242668186, https://github.com/project-chip/connectedhomeip/pull/26082#discussion_r1242668186
- return ConstraintError for invalid enum values https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242673209
- tighten up granularity check & set to one lower value https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242671601, https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242676471
- handle event generation in update dst and time zone functions. Only exception is when the change of state can only be detected in the ember command callbacks - https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242683543, https://github.com/project-chip/connectedhomeip/pull/26082#discussion_r1242628037
- adjust revision in the lighting app example https://github.com/project-chip/connectedhomeip/pull/26082#discussion_r1241227274
- add check for name storage and incoming before copying https://github.com/project-chip/connectedhomeip/pull/26082#discussion_r1242607150
- use CHIP_IM_GLOBAL_STATUS instead of using random error code in SetDSTOffset https://github.com/project-chip/connectedhomeip/pull/26082#discussion_r1242631916
- minor code changes https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242658973, https://github.com/project-chip/connectedhomeip/pull/26082#discussion_r1242628882

Addition of delegate functions for better usability:

- TrustedTimeSourceAvailabilityChanged()
- NotifyTimeFailure()

### DIFF
--- a/src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
+++ b/src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
@@ -589,20 +589,14 @@ CHIP_ERROR TimeSynchronizationServer::SetTimeZone(const DataModel::DecodableList
         tzStore.timeZone.validAt = newTz.validAt;
         if (newTz.name.HasValue() && newTz.name.Value().size() > 0)
         {
-            size_t len = newTz.name.Value().size();
-            if (len > sizeof(tzStore.name))
-            {
-                LoadTimeZone();
-                return CHIP_IM_GLOBAL_STATUS(ConstraintError);
-            }
             memset(tzStore.name, 0, sizeof(tzStore.name));
-            chip::MutableCharSpan tempSpan(tzStore.name, len);
+            chip::MutableCharSpan tempSpan(tzStore.name);
             if (CHIP_NO_ERROR != CopyCharSpanToMutableCharSpan(newTz.name.Value(), tempSpan))
             {
                 LoadTimeZone();
                 return CHIP_IM_GLOBAL_STATUS(InvalidCommand);
             }
-            tzStore.timeZone.name.SetValue(CharSpan(tzStore.name, len));
+            tzStore.timeZone.name.SetValue(tempSpan);
         }
         else
         {
--- a/src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
+++ b/src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
@@ -570,20 +570,20 @@ CHIP_ERROR TimeSynchronizationServer::SetTimeZone(const DataModel::DecodableList
         const auto & newTz = newTzL.GetValue();
         if (newTz.offset < -43200 || newTz.offset > 50400)
         {
-            ReturnErrorOnFailure(LoadTimeZone());
-            return CHIP_ERROR_IM_MALFORMED_COMMAND_DATA_IB;
+            LoadTimeZone();
+            return CHIP_IM_GLOBAL_STATUS(ConstraintError);
         }
         // first element shall have validAt entry of 0
         if (i == 0 && newTz.validAt != 0)
         {
-            ReturnErrorOnFailure(LoadTimeZone());
-            return CHIP_ERROR_IM_MALFORMED_COMMAND_DATA_IB;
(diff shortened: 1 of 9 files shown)


src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
bzbarsky-apple, line 1: This seems like it should have been multiple sane-sized reviewable PRs, not a single giant blob....

fessehaeve: I could break this in to 2 or 3 PRs but more would be too much for me to maintain.
Would you prefer I do that @bzbarsky-apple ?

bzbarsky-apple: Is that because the dozens of issues listed in https://github.com/project-chip/connectedhomeip/issues/27575 involve touching the same bits of code?  Or something else?

Ideally, we would have one PR per issue or close to it.  Right now trying to review this and make sure it actually fixes the list of issues it's claiming is something like O(N*M) where N is size of PR and M is number of issues it claims to fix....

fessehaeve: To be exact, these are not issues. They are all comments (may be for 1 or 2 exceptions) that are tracked in 1 issue.

bzbarsky-apple: I know they are tracked in one github issue.  But they are independent bits that need to be fixed, and comparing the PR to that list is _very_ painful.

fessehaeve: I tried to break down the PR but at this point it needs too much work to align the changes without causing build errors.
Therefore, I will list down all the changes in this PR in the description. I hope that works for you.

bzbarsky-apple: The big problem is comparing the list in the description to the list of things this is supposed to be fixing and comparing one list or the other to the code....

That is, how do I check within sane time that this PR in fact fixes the things it claims to be fixing?

PR #32774, opened 2024-03-28
Revert "Remove TV app from full builds and smoke tests as they have been included in CI as a build example per issue Issue 30906"

kliao-csa:
Reverts project-chip/connectedhomeip#32763

Copy-paste from relevant comment:

New context from Yufeng indicates that integrating this smoketest as a standalone build is not what we are looking for; instead we want runtime testing because failures were happening there rather than during builds.

Would be nice to have this context in the issue writeup next time...

I am reverting this for now.


kliao-csa:
These changes were made with the assumption that "add TV app to CI" meant moving what was running in smoketest was stable and therefore no longer preliminary. It seems we may actually want some runtime testing for TV app; that would come in a later PR

yunhanw-google (approved):

bzbarsky-apple (approved):

PR #32775, opened 2024-03-28
Move declarations from `af.h` into the corresponding storage/table/... headers

andy31415:
Ember `af.h` was a catch-all that declared functions from various cpp implementations, making it harder to figure out dependencies. It seems the original intent was to declare "a public api" however the end result was that a lot of code was including "af.h" as a catch all of "use ember". Since ember is not really as stand-alone currently, looking to enforce "if foo.cpp definex X, then foo.h should declare X".

Here I am working towards first making dependencies clear and then later iterating to define the a common "cluster/attribute/command/... metadata" interface that we could use to decouple code-generated data models from the interfaces used by the IM and cluster implemnentations, so that things become injectable (ability for code-driven clusters) and existing cluster code becomes testable in unit tests.


yunhanw-google (approved):

andy31415:
Closing this one in favor of #32776 - after reimaging my machine, I messed up the origin/upstream and I was creating PRs on upstream, which results in double the CI for PRs. Moved to personal fork (the other PR)

PR #32776, opened 2024-03-28
Move declarations from `af.h` into the corresponding storage/table/... headers

andy31415:
Ember af.h was a catch-all that declared functions from various cpp implementations, making it harder to figure out dependencies. It seems the original intent was to declare "a public api" however the end result was that a lot of code was including "af.h" as a catch all of "use ember". Since ember is not really as stand-alone currently, looking to enforce "if foo.cpp definex X, then foo.h should declare X".

Here I am working towards first making dependencies clear and then later iterating to define the a common "cluster/attribute/command/... metadata" interface that we could use to decouple code-generated data models from the interfaces used by the IM and cluster implemnentations, so that things become injectable (ability for code-driven clusters) and existing cluster code becomes testable in unit tests.

### Changes

- remove af.h and move all declarations from af.h into the corresponding headers
- remove af.h includes, add attribute-table/attribute-storage/endpoint-config header includes to satisfy compilation
- af.h was transitively including more things. Some include fixes led me to fix some STL includes, specifically for adding `<string>` for `std::string` usage.
- dynamic_server overrides attribute-table functions, but without a `endpoint_config.h`, however since ota-requestor uses attribute-table.h, I created a separate mock endpoint_config.h and config specifically for dynamic server builds to fix java.


yunhanw-google (approved):

src/app/util/attribute-storage.h
bzbarsky-apple, line 1: Is this file now public API?  My understanding was that it wasn't really meant to be.

We need to be really careful about which things here are public API and which are not.

src/app/util/attribute-table.h
bzbarsky-apple, line 1: Again, is this meant to be public API now?  It's even less so than attribute-storage.h; really nothing outside this directory should be including this file.

src/app/util/attribute-storage.h
andy31415: I believe none of these should be public API at this point in the sense of guaranteeing they will be stable.

My expectation is that we probably have to export a set of APIs to be public, however I don't think the ember bits are what are supposed to be public. We need a set that is generally matter-specific and we can commit to maintaining. Ember is very tied to its data model (linear indexing within in-memory arrays) and has some ZCL legacy bits (e.g. calling things client-side and server-side clusters and various masks).

My intended approach is:
   - remove the approach of "header declaring things implemented across CPP - i.e. the af.h", this is this PR
   - start defining an API that is properly overridable (instead of compile-time linkage that depends on link order). That could be public API, although at this point it still seems ember-designed so I would say it should not yet be called public
   - define a real/final public API that has clear parallels with the matter data model, document it as "public"

src/app/util/attribute-table.h
andy31415: I believe all of attribute-table and attribute-storage should be "ember private". However the reality now is that "af.h" declares a subset of attribute-table and attribute-storage (and I believe util, compatibility functions and maybe more) considered as public. I am looking to trim that down: if you depend on attribute-table/attribute-storage it still feels like you depend on some implementation-specific bit, except you go through a header that makes coupling even harder (because now attribute-table, storage, bindings, util are all tied together with one header).

Right now I approached this by using "if foo.cpp implements it, then foo.h declares it".

Another approach possibility would be to make `foo.h` and `foo-private.h` if you think that is better (I would avoid `foo-public.h` because I believe these APIs are too tied to ZCL data model to call them public ... I would like to change them and have better maintained public API). Accessors.h/cpp is the only public thing I really want to keep. Relying on any other function seems wrong. Are you aware of other functions that seem really should be public?

Another thing we have public however I did not know how to even approach fixing is the defines for `FOO_..._CLUSTER_COUNT` and such. I think those should go away as well.

src/app/util/attribute-storage.h
bzbarsky-apple: OK.  I guess we already had `emberAfSetDynamicEndpoint` which is public SDK API (in the sense that you can't implement a bridge without using it) in this file...

src/app/util/attribute-table.h
bzbarsky-apple: > However the reality now is that "af.h" declares a subset of attribute-table and attribute-storage

It was declaring the subset that was nominally "public API".

> Are you aware of other functions that seem really should be public?

It depends on whether we consider the "ZAP-generated output" to be "public".  If not, then none of this stuff is really public API except maybe the ability to register AttributeAccessInterface implementations.

But the real question is: what are actual applications using in practice?

andy31415: I believe existing applications just used whatever thing was available, with some guidance from us to allow `emberAf*` and not allow `emAf*`, however we may have never defined "public stable API". Most applications at this point likely copied & adapted example code. Part of any updates in API will involve updating existing clusters and examples, so that should serve at least as a migration phase if required.

I expect some migrations will eventually be required:
   - inject the data model in "main"
   - any use of `emberAf*` or `emAf*` to be replaced with some equivalent methods
   - update usage of any `#defines` constants (like cluster counts) ... although these may be mostly in cluster implementations which apps do not touch.
   
Accessor and command/attribute access interfaces should stay the same.

src/app/util/attribute-storage.h
andy31415: Separated out things into a `-detail.h` header for the old declaration, kept only the seemingly publically-used functions in the main header (even though this exposes a LOT in attribute-storage.h).

src/app/util/attribute-table.h
andy31415: For now to limit exposure, I created separate `-detail.h` headers for attribute-table and attribute-storage.

I believe we will have to iterate a bit until we get a real public api.

bzbarsky-apple (approved):

src/app/util/attribute-storage-detail.h
```diff
@@ -0,0 +1,88 @@
+/**
+ *
+ *    Copyright (c) 2020 Project CHIP Authors
```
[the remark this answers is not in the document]
andy31415: I updated ... however I generally just copied and pasted declares, so unsure if this really qualifies as a copyright update... technically maybe 2020-2024 ... however :shrug:  ... it belongs to `Project CHIP Authors` and that is it.

Issue #32678, opened 2024-03-21
Address the comment about having the InterfaceKey in DnsSdImpl for Darwin as a struct rather than an std::pair 


nivi-apple:
Comment from here - https://github.com/project-chip/connectedhomeip/pull/32631/files#diff-d3d1d99d691d0aef90e9de721cad391450d0e28aa0b727d812022acc3755f248R551

[ksperling-apple](https://github.com/ksperling-apple) [yesterday](https://github.com/project-chip/connectedhomeip/pull/32631/files#r1533054058)
I think this code would be more readable if we had structs with 2 meaningfully-named fields, rather than using std::pair. If you use aggregate initialization it doesn't involve writing more code either.

Member
Author
@[nivi-apple](https://github.com/nivi-apple) nivi-apple [6 hours ago](https://github.com/project-chip/connectedhomeip/pull/32631/files#r1534253332)
I used std::pair from my conversation with Boris. I could add comments in code to say what first and second are if that helps.

@[Abhayakara](https://github.com/Abhayakara) Abhayakara [5 hours ago](https://github.com/project-chip/connectedhomeip/pull/32631/files#r1534328538)
I have to agree with Karsten here. Use a #define if you need to. :)

Member
Author
@[nivi-apple](https://github.com/nivi-apple) nivi-apple [now](https://github.com/project-chip/connectedhomeip/pull/32631/files#r1534771979)
so I tried to change this to a struct and running into issues where the map does some sorting based on keys and is looking for some operators on the struct. I spent some time adding < and + operator but seems like a non trivial fix here.

PR #32779, opened 2024-03-29
Add support to browse and resolve on both local and SRP domain

### LINKED ISSUE #32681: The dnssd changes that involves browsing on both local and srp domains has some major bugs that need to be fixed.
line - https://github.com/project-chip/connectedhomeip/pull/32631/files#diff-eba5148bba8ab5cd4e32114e5ddc8315f4f54f588b70a32039c09ba6bec34a8cR148

This should always be the last character of the string though? If it isn't then something else is going wrong.

Member
@[bzbarsky-apple](https://github.com/bzbarsky-apple) bzbarsky-apple [17 hours ago](https://github.com/project-chip/connectedhomeip/pull/32631/files#r1533254026)
Yes, exactly. This is making some assumptions, but then acting as if those assumptions are not being made, which means it will fail in bad ways if those assumptions happen to be violated.

Member
Author
@[nivi-apple](https://github.com/nivi-apple) nivi-apple [6 hours ago](https://github.com/project-chip/connectedhomeip/pull/32631/files#r1534272151)
added a check that last_pos should be equal to the string length.

Member
Author
@[nivi-apple](https://github.com/nivi-apple) nivi-apple [4 minutes ago](https://github.com/project-chip/connectedhomeip/pull/32631/files#r1534780515)
N/A. The API was changed to look for the first occurence of '.' after talking to Ted seems like the hostname with domain has hostname. after the first '.'. I will follow up on this after checking the matter spec on how the hostname could look like in different variations and will make any fixes as required.

Line - https://github.com/project-chip/connectedhomeip/pull/32631/files#diff-eba5148bba8ab5cd4e32114e5ddc8315f4f54f588b70a32039c09ba6bec34a8cL227

Can kDNSServiceFlagsMoreComing be relied upon to work this way when kDNSServiceFlagsShareConnection is in use? It seems like we're assuming here that it means the DNSServiceBrowse and DNSServiceGetAddrInfo calls for the domain of this last result is done for now.

https://developer.apple.com/documentation/dnssd/1823436-anonymous/kdnsserviceflagsshareconnection

Member
Author
@[nivi-apple](https://github.com/nivi-apple) nivi-apple [5 hours ago](https://github.com/project-chip/connectedhomeip/pull/32631/files#r1534304941)
When Boris, Ted and I discussed we didn't realize this can be an issue. I need to spend some time on this as I am not an expert in the domain. @bzbarsky-apple do you have an opinion here?

@[Abhayakara](https://github.com/Abhayakara) Abhayakara [5 hours ago](https://github.com/project-chip/connectedhomeip/pull/32631/files#r1534341130)
kDNSServiceFlagsMoreComing says that we have any more data on the shared connection. If you are doing multiple queries on the same connection, the response that doesn't set the MoreComing flag for any particular series of results can happen on any of the queries.

Member
Author
@[nivi-apple](https://github.com/nivi-apple) nivi-apple [3 minutes ago](https://github.com/project-chip/connectedhomeip/pull/32631/files#r1534783287)
This needs more thought and has to be fixed correctly. I need to check the API documentation and see how best to handle the kDNSServiceFlagsMoreComing for a shared connection. will do as a follow up

Line - https://github.com/project-chip/connectedhomeip/pull/32631/files#diff-eba5148bba8ab5cd4e32114e5ddc8315f4f54f588b70a32039c09ba6bec34a8cR367

[bzbarsky-apple](https://github.com/bzbarsky-apple) [17 hours ago](https://github.com/project-chip/connectedhomeip/pull/32631/files#r1533257366)
I don't understand this logic. We want to call GetAddrInfo any time we get OnResolve without kDNSServiceFlagsMoreComing, and only do the GetAddrInfo on the things we didn't do it on yet.

As in, this isResolveRequested flag needs to be per interface/domain pair, not per-context, no?

Member
Author
@[nivi-apple](https://github.com/nivi-apple) nivi-apple [3 minutes ago](https://github.com/project-chip/connectedhomeip/pull/32631/files#r1534784909)
I misunderstood this. This needs rework and I am going to do that as a follow up.

Line - https://github.com/project-chip/connectedhomeip/pull/32631/files#diff-eba5148bba8ab5cd4e32114e5ddc8315f4f54f588b70a32039c09ba6bec34a8cR386

This is incorrect. You should resolve the domain that you got in the browse callback. There is no promise that if you do a browse in local, all your responses will be in local. Same for default.service.arpa. There is no need to do two resolves here, and in fact it's incorrect because there might be totally unrelated data that happens to match the hostname you got for local in default.service.arpa, or vice versa.

Member
Author
@[nivi-apple](https://github.com/nivi-apple) nivi-apple [2 hours ago](https://github.com/project-chip/connectedhomeip/pull/32631/files#r1534602640)
We discussed this with you in our meeting with Boris. So we get a browse before resolve only for the pairing use cases. for operational reads/writes, we do not get a browse before a resolve and hence we had to do the same thing as browse. I know you re-iterated that what we get in the browse result can be different than what we requested. so what i could do is if the domain in the browse result is not null i.e if a browse preceded a resolve, i will use the domain from the browse result otherwise we will resolve on local domain and srp domain. would that work?

Member
Author
@[nivi-apple](https://github.com/nivi-apple) nivi-apple [3 minutes ago](https://github.com/project-chip/connectedhomeip/pull/32631/files#r1534786754)
This needs a discussion with @bzbarsky-apple. The code right now should work for the 2 domains we care about the local and srp domain but definitely has a bug. Need to be re-written as follow up

Line - https://github.com/project-chip/connectedhomeip/pull/32631/files#diff-d3d1d99d691d0aef90e9de721cad391450d0e28aa0b727d812022acc3755f248R722

This would have been a great place to use auto.

Member
Author
@nivi-apple nivi-apple 6 hours ago
filed - 32681

nivi-apple:
Fixes: #32681 #32678


src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -368,7 +362,12 @@ void BrowseContext::DispatchSuccess()
 void BrowseContext::DispatchPartialSuccess()
 {
     sContextDispatchingSuccess = this;
-    callback(context, services.data(), services.size(), false, CHIP_NO_ERROR);
+    std::vector<DnssdService> dnsServices;
```
nivi-apple, line 365: once this merges #32740, i need to rebase this. Right now I am keeping the changes here as well.

src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -494,7 +489,13 @@ ResolveContext::ResolveContext(CommissioningResolveDelegate * delegate, chip::In
     consumerCounter = std::move(consumerCounterToUse);
 }
 
-ResolveContext::~ResolveContext() {}
+ResolveContext::~ResolveContext()
+{
+    if (this->isSRPTimerRunning)
```
bzbarsky-apple, line 494:
```suggestion
    if (isSRPTimerRunning)
```

src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -602,15 +605,47 @@ bool ResolveContext::TryReportingResultsForInterfaceIndex(uint32_t interfaceInde
     return true;
 }
 
-CHIP_ERROR ResolveContext::OnNewAddress(uint32_t interfaceId, const struct sockaddr * address)
+bool ResolveContext::TryReportingResultsForInterfaceIndex(uint32_t interfaceIndex)
+{
+    if (interfaceIndex == 0)
+    {
+        // Not actually an interface we have.
```
bzbarsky-apple, line 612: Why bother with this check?  Either it's not in the map, or we will call the other overload of TryReportingResultsForInterfaceIndex and it will do the check...

src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -602,15 +605,47 @@ bool ResolveContext::TryReportingResultsForInterfaceIndex(uint32_t interfaceInde
...
+        return false;
+    }
+
+    for (auto & interface : interfaces)
+    {
+        if (interface.first.interfaceId == interfaceIndex)
+        {
+            auto & ips = interface.second.addresses;
```
bzbarsky-apple, line 620: Again, why are we checking `ips` here?  Just call the other overload; it does this check already.

src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -652,7 +687,7 @@ bool ResolveContext::HasAddress()
 }
 
 void ResolveContext::OnNewInterface(uint32_t interfaceId, const char * fullname, const char * hostnameWithDomain, uint16_t port,
-                                    uint16_t txtLen, const unsigned char * txtRecord)
+                                    uint16_t txtLen, const unsigned char * txtRecord, bool isSRPType)
```
bzbarsky-apple, line 690:
```suggestion
                                    uint16_t txtLen, const unsigned char * txtRecord, bool isFromSRPResolve)
```
perhaps?  Or just `isSRP`?

src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -566,16 +568,17 @@ void ResolveContext::DispatchSuccess()
     }
 }
 
-bool ResolveContext::TryReportingResultsForInterfaceIndex(uint32_t interfaceIndex)
+bool ResolveContext::TryReportingResultsForInterfaceIndex(uint32_t interfaceIndex, std::string hostname, bool isSRPType)
```
bzbarsky-apple, line 571:
```suggestion
bool ResolveContext::TryReportingResultsForInterfaceIndex(uint32_t interfaceIndex, const std::string & hostname, bool isSRPType)
```

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -126,11 +131,52 @@ std::shared_ptr<uint32_t> GetCounterHolder(const char * name)
     return std::make_shared<uint32_t>(0);
 }
 
+bool IsSRPType(const char * domain)
```
bzbarsky-apple, line 134:
```suggestion
bool IsSRPDomain(const char * domain)
```

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -126,11 +131,52 @@ std::shared_ptr<uint32_t> GetCounterHolder(const char * name)
...
+ * @brief Starts a timer to wait for the resolution on the kSRPDot domain to happen.
+ *
+ * @param[in] timeoutSeconds The timeout in seconds.
+ * @param[in] ResolveContext The resolve context.
+ */
+CHIP_ERROR StartSRPTimer(uint16_t timeoutInMSecs, ResolveContext * ctx)
+{
+    VerifyOrReturnValue(ctx != nullptr, CHIP_ERROR_INCORRECT_STATE);
+    return DeviceLayer::SystemLayer().StartTimer(System::Clock::Milliseconds16(timeoutInMSecs), SRPTimerExpiredCallback,
+                                                 reinterpret_cast<void *>(ctx));
```
bzbarsky-apple, line 167:
```suggestion
                                                 static_cast<void *>(ctx));
```
or change the cast the other way to be `reinterpret_cast`, but make the two match.  I'd personally prefer `static_cast` for both.

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -126,11 +131,52 @@ std::shared_ptr<uint32_t> GetCounterHolder(const char * name)
...
+}
+
+/**
+ * @brief Cancels the timer that was started to wait for the resolution on the kSRPDot domain to happen.
+ *
+ * @param[in] ResolveContext The resolve context.
+ */
+void CancelSRPTimer(ResolveContext * ctx)
+{
+    DeviceLayer::SystemLayer().CancelTimer(SRPTimerExpiredCallback, reinterpret_cast<void *>(ctx));
```
bzbarsky-apple, line 177:
```suggestion
    DeviceLayer::SystemLayer().CancelTimer(SRPTimerExpiredCallback, static_cast<void *>(ctx));
```

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -126,11 +131,52 @@ std::shared_ptr<uint32_t> GetCounterHolder(const char * name)
...
 namespace chip {
 namespace Dnssd {
 
+/**
+ * @brief Callback that is called when the timeout for resolving on the kSRPDot domain has expired.
+ *
+ * @param[in] systemLayer The system layer.
+ * @param[in] callbackContext The context passed to the timer callback.
+ */
+void SRPTimerExpiredCallback(System::Layer * systemLayer, void * callbackContext)
```
bzbarsky-apple, line 150: Why are these global functions?  Shouldn't they be in the anonymous namespace above?  Or inside the one below, either way.  But they need to be in an anonymous namespace.

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -215,20 +275,55 @@ static void OnGetAddrInfo(DNSServiceRef sdRef, DNSServiceFlags flags, uint32_t i
 {
     ChipLogProgress(Discovery, "Mdns: %s flags: %d, interface: %u, hostname: %s", __func__, flags, (unsigned) interfaceId,
                     StringOrNullMarker(hostname));
-    auto sdCtx = reinterpret_cast<ResolveContext *>(context);
+
+    auto contextWithType = reinterpret_cast<ResolveContextWithType *>(context);
+    VerifyOrReturn(contextWithType != nullptr, ChipLogError(Discovery, "ResolveContextWithType is null"));
+
+    auto sdCtx = reinterpret_cast<ResolveContext *>(contextWithType->context);
```
bzbarsky-apple, line 282: Why is there a cast here?  There should be no cast here.

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -215,20 +275,55 @@ static void OnGetAddrInfo(DNSServiceRef sdRef, DNSServiceFlags flags, uint32_t i
...
     ReturnOnFailure(MdnsContexts::GetInstance().Has(sdCtx));
     LogOnFailure(__func__, err);
 
     if (kDNSServiceErr_NoError == err)
     {
-        sdCtx->OnNewAddress(interfaceId, address);
+        InterfaceKey interfaceKey = { interfaceId, hostname, contextWithType->isSRPType };
```
bzbarsky-apple, line 288:
```suggestion
        InterfaceKey interfaceKey = { interfaceId, hostname, contextWithType->isSRP };
```

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -215,20 +275,55 @@ static void OnGetAddrInfo(DNSServiceRef sdRef, DNSServiceFlags flags, uint32_t i
...
+        sdCtx->OnNewAddress(interfaceKey, address);
+
+        // Set the flag to start the timer for resolve on SRP domain to complete if the key has the SRP type requested flag set to
+        // true.
+        if (interfaceKey.isSRPTypeRequested)
+        {
+            sdCtx->shoulStartSRPTimerForResolve = true;
```
bzbarsky-apple, line 295: This logic is making no sense to me.  interfaceKey.isSRPTypeRequested will be true if we have already gotten an IP address from some SRP thing (that's how we get into this function with this flag set).  Why would we do any timer-starting in that case?

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -215,20 +275,55 @@ static void OnGetAddrInfo(DNSServiceRef sdRef, DNSServiceFlags flags, uint32_t i
...
+    if (flags & kDNSServiceFlagsMoreComing)
+    {
+        return;
     }
 
-    if (!(flags & kDNSServiceFlagsMoreComing))
+    VerifyOrReturn(sdCtx->HasAddress(), sdCtx->Finalize(kDNSServiceErr_BadState));
+
+    // We should complete the resolve if we got a resolution on non SRP domain and no resolution on SRP domain was requested.
+    if (!sdCtx->shoulStartSRPTimerForResolve)
```
bzbarsky-apple, line 307: Seems to me like shouldStartSRPTimerForResolve (note the spelling) should be set when we are _starting the resolve_, based on whether we start a resolve on the SRP domain there.  It shouldn't be getting set above.

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -215,20 +275,55 @@ static void OnGetAddrInfo(DNSServiceRef sdRef, DNSServiceFlags flags, uint32_t i
...
     {
-        VerifyOrReturn(sdCtx->HasAddress(), sdCtx->Finalize(kDNSServiceErr_BadState));
         sdCtx->Finalize();
     }
+    else
+    {
+        // Since we started a resolve on the SRP domain we are going to start a timer to give the resolve some extra time to
```
bzbarsky-apple, line 313: Where are we checking for "we got an SRP answer already, no need to start the timer"?

All this timer-management logic seems wrong.

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -237,11 +332,19 @@ static void GetAddrInfo(ResolveContext * sdCtx)
...
+        auto interfaceId = interface.first.interfaceId;
         auto hostname    = interface.second.fullyQualifiedDomainName.c_str();
         auto sdRefCopy   = sdCtx->serviceRef; // Mandatory copy because of kDNSServiceFlagsShareConnection
-        auto err = DNSServiceGetAddrInfo(&sdRefCopy, kGetAddrInfoFlags, interfaceId, protocol, hostname, OnGetAddrInfo, sdCtx);
-        VerifyOrReturn(kDNSServiceErr_NoError == err, sdCtx->Finalize(err));
+
+        if (!interface.second.isDNSLookUpRequested)
+        {
+            ResolveContextWithType * contextWithType =
+                (interface.first.isSRPTypeRequested) ? &sdCtx->resolveContextWithSRPType : &sdCtx->resolveContextWithNonSRPType;
```
bzbarsky-apple, line 342:
```suggestion
                (interface.first.isSRPResult) ? &sdCtx->resolveContextWithSRPType : &sdCtx->resolveContextWithNonSRPType;
```

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -237,11 +332,19 @@ static void GetAddrInfo(ResolveContext * sdCtx)
...
     for (auto & interface : sdCtx->interfaces)
     {
-        auto interfaceId = interface.first;
+        auto interfaceId = interface.first.interfaceId;
         auto hostname    = interface.second.fullyQualifiedDomainName.c_str();
         auto sdRefCopy   = sdCtx->serviceRef; // Mandatory copy because of kDNSServiceFlagsShareConnection
-        auto err = DNSServiceGetAddrInfo(&sdRefCopy, kGetAddrInfoFlags, interfaceId, protocol, hostname, OnGetAddrInfo, sdCtx);
-        VerifyOrReturn(kDNSServiceErr_NoError == err, sdCtx->Finalize(err));
+
+        if (!interface.second.isDNSLookUpRequested)
```
bzbarsky-apple, line 339: Why is this check not the very first thing in the loop, with a `continue` if `isDNSLookUpRequested`?

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -251,13 +354,17 @@ static void OnResolve(DNSServiceRef sdRef, DNSServiceFlags flags, uint32_t inter
 {
     ChipLogProgress(Discovery, "Mdns: %s flags: %d, interface: %u, fullname: %s, hostname: %s, port: %u", __func__, flags,
                     (unsigned) interfaceId, StringOrNullMarker(fullname), StringOrNullMarker(hostname), ntohs(port));
-    auto sdCtx = reinterpret_cast<ResolveContext *>(context);
+
+    auto contextWithType = reinterpret_cast<ResolveContextWithType *>(context);
+    VerifyOrReturn(contextWithType != nullptr, ChipLogError(Discovery, "ResolveContextWithType is null"));
+
+    auto sdCtx = reinterpret_cast<ResolveContext *>(contextWithType->context);
```
bzbarsky-apple, line 361: There should be no cast here.

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -267,18 +374,40 @@ static void OnResolve(DNSServiceRef sdRef, DNSServiceFlags flags, uint32_t inter
...
-    err            = DNSServiceResolve(&sdRefCopy, kResolveFlags, interfaceId, name, type, kLocalDot, OnResolve, sdCtx);
-    VerifyOrReturnError(kDNSServiceErr_NoError == err, sdCtx->Finalize(err));
+    // If we have a single domain from a browse, we will use that for the Resolve.
+    // Otherwise we will try to resolve using both the local domain and the SRP domain.
+    if (domain != nullptr)
+    {
+        ReturnErrorOnFailure(
+            ResolveWithContext(sdCtx, interfaceId, type, name, domain,
+                               IsSRPType(domain) ? &sdCtx->resolveContextWithSRPType : &sdCtx->resolveContextWithNonSRPType));
+        sdCtx->shoulStartSRPTimerForResolve = false;
```
bzbarsky-apple, line 403: OK, so _this_ part makes sense....

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -267,18 +374,40 @@ static void OnResolve(DNSServiceRef sdRef, DNSServiceFlags flags, uint32_t inter
...
-    auto sdRefCopy = sdCtx->serviceRef; // Mandatory copy because of kDNSServiceFlagsShareConnection
-    err            = DNSServiceResolve(&sdRefCopy, kResolveFlags, interfaceId, name, type, kLocalDot, OnResolve, sdCtx);
-    VerifyOrReturnError(kDNSServiceErr_NoError == err, sdCtx->Finalize(err));
+    // If we have a single domain from a browse, we will use that for the Resolve.
+    // Otherwise we will try to resolve using both the local domain and the SRP domain.
+    if (domain != nullptr)
+    {
+        ReturnErrorOnFailure(
+            ResolveWithContext(sdCtx, interfaceId, type, name, domain,
+                               IsSRPType(domain) ? &sdCtx->resolveContextWithSRPType : &sdCtx->resolveContextWithNonSRPType));
```
bzbarsky-apple, line 402: Why not just use the "non SRP" context for everything here?  Then we don't need this IsSRPType() thing....

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -442,9 +571,23 @@ CHIP_ERROR ChipDnssdResolve(DnssdService * service, chip::Inet::InterfaceId inte
...
-    return Resolve(context, callback, interfaceId, service->mAddressType, regtype.c_str(), service->mName);
+    auto regtype        = GetFullType(service);
+    auto interfaceId    = GetInterfaceId(interface);
+    const char * domain = nullptr;
+
+    if (BrowseContext::sContextDispatchingSuccess != nullptr)
+    {
+        for (size_t i = 0; i < BrowseContext::sDispatchedServices->size(); ++i)
+        {
+            if (service == &BrowseContext::sDispatchedServices->at(i))
```
bzbarsky-apple, line 582:
```suggestion
            if (service == &BrowseContext::sDispatchedServices[i])
```

src/platform/Darwin/DnssdImpl.h
```diff
@@ -222,17 +225,50 @@ struct InterfaceInfo
...
+};
+
+struct InterfaceKey
+{
+    InterfaceKey()  = default;
+    ~InterfaceKey() = default;
+    inline bool operator<(const InterfaceKey & other) const
+    {
+        return (this->interfaceId < other.interfaceId) ||
+            ((this->interfaceId == other.interfaceId) && (this->hostname < other.hostname));
```
bzbarsky-apple, line 238: Needs to also test isSRPTypeRequested (or whatever it's called).

src/platform/Darwin/DnssdImpl.h
```diff
@@ -222,17 +225,50 @@ struct InterfaceInfo
...
+    }
+
+    uint32_t interfaceId;
+    std::string hostname;
+    bool isSRPTypeRequested = false;
```
bzbarsky-apple, line 243:
```suggestion
    bool isSRPResult = false;
```

src/platform/Darwin/DnssdImpl.h
```diff
@@ -222,17 +225,50 @@ struct InterfaceInfo
...
+};
+
+struct ResolveContextWithType
+{
+    ResolveContextWithType()  = default;
+    ~ResolveContextWithType() = default;
+
+    ResolveContext * context;
```
bzbarsky-apple, line 251:
```suggestion
    ResolveContext * const context;
```

src/platform/Darwin/DnssdImpl.h
```diff
@@ -222,17 +225,50 @@ struct InterfaceInfo
...
+    bool isSRPType = false;
```
bzbarsky-apple, line 252:
```suggestion
    const bool isSRPResolve;
```

src/platform/Darwin/DnssdImpl.h
```diff
@@ -222,17 +225,50 @@ struct InterfaceInfo
...
-    std::map<uint32_t, InterfaceInfo> interfaces;
+    std::map<InterfaceKey, InterfaceInfo> interfaces;
     DNSServiceProtocol protocol;
     std::string instanceName;
     std::shared_ptr<uint32_t> consumerCounter;
     BrowseContext * const browseThatCausedResolve; // Can be null
 
+    // Indicates whether the timer for 250 msecs should be started
+    // to give the resolve on SRP domain some extra time to complete.
+    bool shoulStartSRPTimerForResolve = false;
```
bzbarsky-apple, line 266:
```suggestion
    bool shouldStartSRPTimerForResolve = false;
```

src/platform/Darwin/DnssdImpl.h
```diff
@@ -222,17 +225,50 @@ struct InterfaceInfo
...
+    bool isSRPTimerRunning            = false;
+
+    ResolveContextWithType resolveContextWithSRPType;
+    ResolveContextWithType resolveContextWithNonSRPType;
```
bzbarsky-apple, line 270: Note that you could initialize them here, and then don't need to do it in both constructors....

src/platform/Darwin/DnssdImpl.h
```diff
@@ -244,11 +280,11 @@ struct ResolveContext : public GenericContext
     void DispatchFailure(const char * errorStr, CHIP_ERROR err) override;
     void DispatchSuccess() override;
 
-    CHIP_ERROR OnNewAddress(uint32_t interfaceId, const struct sockaddr * address);
+    CHIP_ERROR OnNewAddress(const InterfaceKey & interfaceKey, const struct sockaddr * address);
     bool HasAddress();
 
     void OnNewInterface(uint32_t interfaceId, const char * fullname, const char * hostname, uint16_t port, uint16_t txtLen,
-                        const unsigned char * txtRecord);
+                        const unsigned char * txtRecord, bool isSRPType);
```
bzbarsky-apple, line 287:
```suggestion
                        const unsigned char * txtRecord, bool isSRPResult);
```

src/platform/Darwin/DnssdImpl.h
```diff
@@ -258,6 +294,8 @@ struct ResolveContext : public GenericContext
      * Returns true if information was reported, false if not (e.g. if there
      * were no IP addresses, etc).
      */
+    bool TryReportingResultsForInterfaceIndex(uint32_t interfaceIndex, std::string hostname, bool isSRPType);
```
bzbarsky-apple, line 297:
```suggestion
    bool TryReportingResultsForInterfaceIndex(uint32_t interfaceIndex, const std::string & hostname, bool isSRPResult);
```

src/platform/Darwin/DnssdContexts.cpp
nivi-apple: i had removed it but then thought it will return early. will remove

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: The CancelTimer is called by the DnssdContexts and i can't call it from there if I put it in an anonymous namespace

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: This was not compiling. I can do a pointer->at or a &(*[value[i])

../../examples/chip-tool/third_party/connectedhomeip/src/platform/Darwin/DnssdImpl.cpp:582:25: error: comparison of distinct pointer types ('DnssdService *' and 'std::vector<DnssdService> *')
            if (service == &BrowseContext::sDispatchedServices[i])
                ~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
[401/761] objc++ obj/third_party/connectedhomeip/src/platform/Darwin/Darwin.BleConnectionDelegateImpl.mm.o

src/platform/Darwin/DnssdContexts.cpp
nivi-apple: done. removed.

src/platform/Darwin/DnssdContexts.cpp
nivi-apple: done removed.

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: removed.

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: done.

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: removed.

src/platform/Darwin/DnssdImpl.h
nivi-apple: done.

src/platform/Darwin/DnssdImpl.h
nivi-apple: done.

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: I fixed the timer logic based on what I think I understood by re-visiting the notes I took. I messed up before. For now the mental model is when calling ResolveWithContext based on whether we start a resolve on the SRP domain. In OnGetAddrInfo after successfully adding a new address if the address resolved was on SRP domain, i set shouldStartSRPTimerForResolve to false. hope that's correct.

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: done.

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: tried to fix it

src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -566,16 +566,17 @@ void ResolveContext::DispatchSuccess()
     }
 }
 
-bool ResolveContext::TryReportingResultsForInterfaceIndex(uint32_t interfaceIndex)
+bool ResolveContext::TryReportingResultsForInterfaceIndex(uint32_t interfaceIndex, const std::string & hostname, bool isSRPResolve)
```
bzbarsky-apple, line 569: Maybe make that third arg be isSRPResult to match the naming elsewhere?  I guess I suggested this naming before I realized where these booleans were coming from.

But note: in OnNewInterface below, isFromSRPResolve is the right name.  So don't change it there.  ;)

src/platform/Darwin/DnssdImpl.cpp
bzbarsky-apple: OK, but you can put SRPTimerExpiredCallback, StartSRPTimer in an anonymous namespace.

And CancelSRPTimer should just be a private member function on ResolveContext, not a global function.

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -215,20 +275,53 @@ static void OnGetAddrInfo(DNSServiceRef sdRef, DNSServiceFlags flags, uint32_t i
...
     ReturnOnFailure(MdnsContexts::GetInstance().Has(sdCtx));
     LogOnFailure(__func__, err);
 
     if (kDNSServiceErr_NoError == err)
     {
-        sdCtx->OnNewAddress(interfaceId, address);
+        InterfaceKey interfaceKey = { interfaceId, hostname, contextWithType->isSRPResolve };
+        CHIP_ERROR error          = sdCtx->OnNewAddress(interfaceKey, address);
+
+        // If we saw an address resolved on the SRP domain, set the shouldStartSRPTimerForResolve to false.
```
bzbarsky-apple, line 291:
```suggestion
        // If we saw an address resolved on the SRP domain, we don't need to wait
        // for SRP results, so don't bother with starting a timer to wait for those.
```

That is, document the _why_ of the code.  The fact that you are setting things to false is obvious.  It's why you are doing it that needs documentation.

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -215,20 +275,53 @@ static void OnGetAddrInfo(DNSServiceRef sdRef, DNSServiceFlags flags, uint32_t i
...
     }
+    else
+    {
+        if (!sdCtx->isSRPTimerRunning)
+        {
+            CHIP_ERROR error = StartSRPTimer(kSRPTimeoutInMsec, sdCtx);
+
+            if (error != CHIP_NO_ERROR)
+            {
+                sdCtx->Finalize();
```
bzbarsky-apple, line 319:
```suggestion
                // If we failed to start the timer, just go ahead and report whatever information
                // we have gotten so far.
                sdCtx->Finalize();
```

src/platform/Darwin/DnssdImpl.cpp
bzbarsky-apple: Oh, because sDispatchedServices is a pointer to the vector, right.  OK, `at()` is probably the best answer then.

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -267,18 +374,46 @@ static void OnResolve(DNSServiceRef sdRef, DNSServiceFlags flags, uint32_t inter
...
 
+static CHIP_ERROR ResolveWithContext(ResolveContext * sdCtx, uint32_t interfaceId, const char * type, const char * name,
+                                     const char * domain, ResolveContextWithType * contextWithType)
+{
+    auto sdRef = sdCtx->serviceRef; // Mandatory copy because of kDNSServiceFlagsShareConnection
+
+    auto err = DNSServiceResolve(&sdRef, kResolveFlags, interfaceId, name, type, domain, OnResolve, contextWithType);
+    VerifyOrReturnError(kDNSServiceErr_NoError == err, sdCtx->Finalize(err));
+
+    // Set the flag to start the timer for resolve on SRP domain to complete if a resolve has been requested on the SRP domain.
```
bzbarsky-apple, line 385: So this is a little silly, in that we set it true here but then one of the places where we call it immediately sets it false...

We should just set it false if we are starting one resolve, set it true if we are starting two resolves, at the callsites of this function.

src/platform/Darwin/DnssdImpl.cpp
bzbarsky-apple: Who cares?  The only reason we care about whether the results are SRP or not is for purposes of starting the timer, and in this case we are never going to start the timer, right?

src/platform/Darwin/DnssdContexts.cpp
nivi-apple: done.

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: that was silly. fixed it.

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: done.

src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -404,13 +401,12 @@ void BrowseContext::OnBrowseRemove(const char * name, const char * type, const c
...
 
     VerifyOrReturn(name != nullptr);
-    VerifyOrReturn(IsLocalDomain(domain));
 
     services.erase(std::remove_if(services.begin(), services.end(),
                                   [name, type, interfaceId, domain](const auto & service) {
                                       return strcmp(name, service.first.mName) == 0 && type == GetFullType(&service.first) &&
                                           service.first.mInterface == chip::Inet::InterfaceId(interfaceId) &&
-                                          strcmp(domain, service.second.c_str()) == 0;
+                                          domain != nullptr && strcmp(domain, service.second.c_str()) == 0;
```
bzbarsky-apple, line 409: How does `domain` get to be null here?

In any case, if it's null, shouldn't we treat it as empty string?  Otherwise if those entries got added with `std::string(nullptr)` as the domain, wouldn't we fail to remove them here?

All of which is to say that the simplest thing is to do something like:

```
std::string domain_str(domain);
```

before the `erase()` call, capture that by reference in the lambda, and use `service.second == domain_str` as part of the condition.

bzbarsky-apple: And note that https://github.com/project-chip/connectedhomeip/pull/32779#discussion_r1544845831 is still not addressed.

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: sounds good. sorry i missed this.

nivi-apple: done.

bzbarsky-apple (approved):

bzbarsky-apple:

PR #32780, opened 2024-03-29
lock all the add endpoint operations in a single chip stacklock of br…

jinbiaochenstayhungry:
When i worked on a bridge app based on the example bridge-app,i always encoutered endpoint lost in apple home app,or alexa app when re-start my bridge app , it turn out to be if i locked all the AddDeviceEndpoint call in a single stacklock of chip, all these problems will gone.
   It seems very import for these home app to read list in device start up, but anyway the don't read all the device in a time and will not always read correctly after that, but if we commit all the AddDeviceEndpoint in a lock, these controllers will absolutely retieved all the endpoint list if these read request arrived.
   Even this is not apparently a bug,i still consider that a more wide range of lock can be better.

### DIFF
--- a/examples/bridge-app/linux/main.cpp
+++ b/examples/bridge-app/linux/main.cpp
@@ -250,8 +250,12 @@ DataVersion gComposedTempSensor2DataVersions[ArraySize(bridgedTempSensorClusters
 
 // ---------------------------------------------------------------------------
 
+//If you add single endpoint in device after start up, it is excepted to lock all the actions of adding.
+//But when adding many devices in start up , we hope a lock must be applied to more wide range of operations,so that all the adding of all endpoints can be done in one lock,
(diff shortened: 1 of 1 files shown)


examples/bridge-app/linux/main.cpp
```diff
@@ -263,7 +267,10 @@ int AddDeviceEndpoint(Device * dev, EmberAfEndpointType * ep, const Span<const E
             while (true)
             {
                 // Todo: Update this to schedule the work rather than use this lock
-                DeviceLayer::StackLock lock;
+                if(lockInternal){
+                    DeviceLayer::StackLock lock;
```
bzbarsky-apple, line 271: This can't possibly work right.

bzbarsky-apple (changes requested):

PR #32809, opened 2024-04-02
CI: Add idf_target for esp_example.sh script

wqx6:
Add IDF target input for esp_example.sh script


examples/all-clusters-app/esp32/sdkconfig.defaults.esp32c3
andy31415, line 1: https://github.com/project-chip/connectedhomeip/blob/master/scripts/build/builders/esp32.py#L142 seems to use the naming of _c3devkit.defaults.

If we rename this, we should probably fix that file as well.
Also m5stack seems to use a similar naming convention. Should we change that too?

andy31415 (changes requested):
Needs fixes for build_examples.py as well.

examples/all-clusters-app/esp32/sdkconfig.defaults.esp32c3
wqx6: When the users run `idf.py set-target {target}`, IDF will choose `sdkconfig.defaults.{target}` as the default sdkconfig file. I renamed this name so that the users will not need to add `-D SDKCONFIG_DEFAULTS={sdkconfi_file}` when they set c3 target. For m5stack, I think we can keep the original name.

Changed scripts/build/builders/esp32.py.

andy31415 (approved):

PR #32817, opened 2024-04-02
Fix for arm64 build failure when bluetooth enabled on linux platform

sabollim-silabs:
Description of Problem: In the platform/Linux code for bluetooth, we are using "typeof" in ConnectDevice function. which throws an error ("typeof not found") for the arm64 build.

Description of Fix:
Changed the "typeof" to "decltype" which is a valid C++ Standard keyword.

Testing Done:
Locally tested, arm64 build successful


bzbarsky-apple (approved):

bzbarsky-apple:
Reopening, we should just make this change.

andy31415 (approved):

arkq (approved):

arkq:
I've added the same fix for Tizen platform.

PR #32818, opened 2024-04-02
Fix cloudbuilds of efr32.

andy31415:
Change #32773 changed `openthread_mtd` to `openthread-mtd`.

Apply this change to cloudbuild yaml files as well.


andy31415:
Hotfix because github CI cannot validate cloudbuild builds.

PR #32819, opened 2024-04-02
Make chef cloudbuild checkout submodules before building

andy31415:
smoketest did this, however chef seemed to rely on bootstrap doing the checkout.

Since #32638 it seems we do not do a submodule checkout on boostrap.


andy31415:
Hotfix as this is a cloudbuild-only change.

PR #32829, opened 2024-04-03
[dnssd] Allow selecting DNS-SD implementation at runtime

kkasperczyk-no:
Add `Resolver::SetInstance()` and `ServiceAdvertiser::SetInstance()` methods for dynamically changing the system-wide DNS-SD implementation used by Matter.

Also, allow for building "minimal" and "platform" DNS-SD implementations together.


src/lib/dnssd/Resolver.cpp
```diff
@@ -0,0 +1,41 @@
...
+namespace chip {
+namespace Dnssd {
+
+Resolver * Resolver::sInstance = nullptr;
+
+Resolver & Resolver::Instance()
+{
+    if (sInstance == nullptr)
+    {
+        sInstance = &GetDefaultResolver();
```
jmartinez-silabs, line 29: I might me missing something, but if you have both minimal and platform, doesn't both GetDefaultResolver conflict?

Same for advertiser.

Damian-Nordic: The implementations can be built independently using new `chip_mdns_minimal` and `chip_mdns_platform` GN args but there's a single default resolver/advertiser selected with the existing `chip_mdns` flag.

Damian-Nordic (approved):

src/lib/dnssd/Resolver.cpp
kkasperczyk-no: It won't, because there are two kinds of flags deciding about it. `chip_mdns_minimal` and `chip_mdns_platform` decide about including the files related to these implementations to the compilation, so we can add both. On the other hand about the `GetDefaultResolver` implementation decides `_chip_dnssd_default_platform` and `_chip_dnssd_default_minimal` flags that cannot be enabled at the same time.

In short we can compile both solutions, but only one can be set as a default one. The other one still can be set dynamically.

bzbarsky-apple (approved):

PR #32834, opened 2024-04-03
Issue/30754 dependabot ci runs for both push and pull

vatsalghelani-csa:
Work for Issue #30754

Disable CI runs for push on any branch except master. This means if we want CI runs on branches not from a fork of connectedhomeip, we will need to make pull requests. Dependabot already does this and I'm not aware of any process that doesn't, so this should be a good standard.


kliao-csa (approved):

.github/workflows/darwin.yaml
```diff
@@ -16,6 +16,8 @@ name: Darwin
 
 on:
     push:
+        branches:
+            - master
```
bzbarsky-apple, line 20: People absolutely trigger CI runs in their own forks (to avoid spamming the shared CI) by pushing to a non-master branch.  Does this break that?

bzbarsky-apple: Also, does this disable CI for Matter "release branches" that correspond to a release, SVE, etc?

andy31415 (approved):

jmartinez-silabs (approved):

PR #32843, opened 2024-04-03
tv-casting-app fixes: saving prevValue before callback

### LINKED ISSUE #32842: [BUG] example/tv-casting-app: Value from last/prev read operation not being set correctly
### Reproduction steps

1. Commission the tv-casting-app with the tv-app
2. Subscribe to MediaPlayback::CurrentState and change its value
3. The prev and current values of the the CurrentState are not returned as expected. Specifically, we are always returning a null previous value.

### Bug prevalence

Everytime

c1f823e78aace77d91e90a8f0fa3037361edb390

### Platform

other

### Platform Version(s)

_No response_

### Anything else?

_No response_

sharadb-amazon:
Fixes https://github.com/project-chip/connectedhomeip/issues/32842

### Change summary
1. In Attribute.h, saved the prevValue of the attribute if it already had a value (from the previous read/subscribe operation) and then set the attribute to its new value.
2. Fixed a file link issue in the APIs.md file
3. Also, increased tv-casting-app's MATTER_BINDING_TABLE_SIZE to 64. For casting, we need to allow for more binding table entries because the Casting App can connect to many Matter Casting Players, each with many Content Apps. Each Casting Player will set 1 binding per endpoint on it. ACasting Player will have 1 endpoint for
every Matter Content App installed on it + 1 endpoint representing the Casting Player + 1 endpoint representing a speaker.
4. Added some debug logging to indicate which version of the Linux tv-casting-app is being run, by outputing chip_casting_simplified = 0 or 1 from the respective files (main.cpp or simple-app.cpp).

### Testing
Tested with the tv-casting-app and tv-app

### DIFF
--- a/examples/tv-casting-app/APIs.md
+++ b/examples/tv-casting-app/APIs.md
@@ -158,9 +158,9 @@ client's lifecycle:
     the Matter specification's "Onboarding Payload" section for more details on
     commissioning data.
 
-    On Linux, define a function `InitCommissionableDataProvider` to initialize
-    initialize a `LinuxCommissionableDataProvider` that can provide the required
-    values to the `CastingApp`.
+    On Linux, define a function `InitCommissionableDataProvider` to initialize a
+    `LinuxCommissionableDataProvider` that can provide the required values to
+    the `CastingApp`.
 
     ```c
     CHIP_ERROR InitCommissionableDataProvider(LinuxCommissionableDataProvider & provider, LinuxDeviceOptions & options) {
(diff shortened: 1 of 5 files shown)


examples/tv-casting-app/APIs.md
pgregorr-amazon, line 1: This link also needs to be fixed?

pgregorr-amazon, line 1: This link also needs to be fixed?

sharadb-amazon: No, that link is fine because zzz\_generated is actually at the root of the SDK repo (unlike the tv-casting-app darwin links)

sharadb-amazon: You can see which links work v/s which don't by clicking on them on this page: [https://project-chip.github.io/connectedhomeip-doc/examples/tv-casting-app/APIs.html](https://project-chip.github.io/connectedhomeip-doc/examples/tv-casting-app/APIs.html)

pgregorr-amazon (approved):
Looks good to me.

examples/tv-casting-app/linux/main.cpp
```diff
@@ -106,6 +106,7 @@ CHIP_ERROR ProcessClusterCommand(int argc, char ** argv)
 
 int main(int argc, char * argv[])
 {
+    ChipLogProgress(AppServer, "chip_casting_simplified = 0");
```
andy31415, line 109: These logs seem hard-coded and tied to a BUILD.gn flag (which is a bool).
There seem to be no direct linkage between the gn flag and the log here ... we need something else probably, otherwise this may be a lie. This seems to just say "chip_casting_simplified = 0" whenever I run main.cpp .... they are not tied.

examples/tv-casting-app/tv-casting-common/include/CHIPProjectAppConfig.h
```diff
@@ -58,6 +58,9 @@
 #define CHIP_CONFIG_EXAMPLE_ACCESS_CONTROL_MAX_SUBJECTS_PER_ENTRY 20
 #define CHIP_CONFIG_EXAMPLE_ACCESS_CONTROL_MAX_ENTRIES_PER_FABRIC 20
 
+// For casting, we need to allow for more binding table entries
+#define MATTER_BINDING_TABLE_SIZE 64
```
andy31415, line 62: this change is not described in the PR summary: why is it tied to the other changes?

andy31415: Was looking at the tile ... the change is in the summary.

However could you please also describe the _why_ the change is done? The summary says "I changed it" and the comment here says "we need more" however it does not say why more or now. It may be things like "this runs on large devices, so having more bindings is possible and better" or "we discovered this use case that needs more bindings because it is common" or something like that.

Readers should understand why changes are done and see if assumptions change over time (i.e. can I make this larger or smaller over time? If I at some point decide to make it smaller again, what do I break?)

chrisdecenzo (approved):

examples/tv-casting-app/tv-casting-common/include/CHIPProjectAppConfig.h
sharadb-amazon: I've updated the config file and the PR summary with the reasons why this is required.

examples/tv-casting-app/linux/main.cpp
sharadb-amazon: I did not pull the value of chip\_casting\_simplified to log in these files, because as per the BUILD.gn, these files are only built/run if chip\_casting\_simplified is set to 0 or 1, respectively for main.cpp and simple-app.cpp. See https://github.com/project-chip/connectedhomeip/blob/master/examples/tv-casting-app/linux/BUILD.gn#L27-L39

I've updated the PR summary to call out this change and also added a comment inline in the code to state the above.

hawk248 (approved):

PR #32845, opened 2024-04-04
Darwin: Refactor stack startup logic when creating a controller

ksperling-apple:
Preparatory work for avoiding the stop/start of the stack we currently do.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
+++ b/src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
@@ -543,10 +543,9 @@ static void ShutdownOnExit()
         // state is brought up live on factory init, and not when it comes time
         // to actually start a controller, and does not actually clean itself up
         // until its refcount (which starts as 0) goes to 0.
-        if (!startingController) {
-            _controllerFactory->RetainSystemState();
-            _controllerFactory->ReleaseSystemState();
-        }
+        // TODO: Don't cause a stack shutdown and restart if startingController
+        _controllerFactory->RetainSystemState();
+        _controllerFactory->ReleaseSystemState();
 
         self->_advertiseOperational = startupParams.shouldStartServer;
         self->_running = YES;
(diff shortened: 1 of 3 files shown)


src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
```diff
@@ -536,8 +543,10 @@ - (BOOL)startControllerFactory:(MTRDeviceControllerFactoryParams *)startupParams
         // state is brought up live on factory init, and not when it comes time
         // to actually start a controller, and does not actually clean itself up
         // until its refcount (which starts as 0) goes to 0.
-        _controllerFactory->RetainSystemState();
-        _controllerFactory->ReleaseSystemState();
+        if (!startingController) {
+            _controllerFactory->RetainSystemState();
+            _controllerFactory->ReleaseSystemState();
```
bzbarsky-apple, line 548: So one concern:

If something fails after this point during controller startup, can we still end up in a situation where the system state is "running" but has a zero refcount?  Seems like we can....

Maybe what we should do is that we should unconditionally RetainSystemState() here, ReleaseSystemState() only if !startingController, and in `_startDeviceController` ensure that we ReleaseSystemState() before returning?  The hard part there would be that it has to happen on the Matter queue and that queue can be shut down partway through that method.... or indeed just on return from this method when we start doing the sanity checks.

Maybe we should do the "don't shut down the Matter queue" thing first?  That would make it a lot easier to ensure the invariants here....

ksperling-apple: Hm yeah makes sense.

ksperling-apple: I've backed the `if (startingController)` but left the other refactorings in place. I'll re-enable this optimization as part of a later PR.

bzbarsky-apple (approved):

PR #32846, opened 2024-04-04
Add DeviceController deleteFromFabricTableOnShutdown option

ksperling-apple:
This is analogous to removeFromFabricTableOnShutdown but does an actual Delete(), i.e. affects storage, rather than just a Forget().

Use this to simplify controller shutdown on Darwin.

### DIFF
--- a/src/controller/CHIPDeviceController.cpp
+++ b/src/controller/CHIPDeviceController.cpp
@@ -135,6 +135,7 @@ CHIP_ERROR DeviceController::Init(ControllerInitParams params)
     mState       = State::Initialized;
 
     mRemoveFromFabricTableOnShutdown = params.removeFromFabricTableOnShutdown;
+    mDeleteFromFabricTableOnShutdown = params.deleteFromFabricTableOnShutdown;
 
     if (GetFabricIndex() != kUndefinedFabricIndex)
     {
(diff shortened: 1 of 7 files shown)


src/controller/CHIPDeviceController.cpp
```diff
@@ -400,13 +401,13 @@ void DeviceController::Shutdown()
         // existing sessions too?
         mSystemState->SessionMgr()->ExpireAllSessionsForFabric(mFabricIndex);
 
-        if (mRemoveFromFabricTableOnShutdown)
+        if (mDeleteFromFabricTableOnShutdown)
         {
-            FabricTable * fabricTable = mSystemState->Fabrics();
-            if (fabricTable != nullptr)
```
ksperling-apple, line 406: We know the SystemState is valid here, so Fabrics() never returns null.

src/controller/CHIPDeviceController.h
```diff
@@ -393,7 +405,8 @@ class DLL_EXPORT DeviceController : public AbstractDnssdDiscoveryController
 
     FabricIndex mFabricIndex = kUndefinedFabricIndex;
 
-    bool mRemoveFromFabricTableOnShutdown = true;
+    bool mRemoveFromFabricTableOnShutdown;
```
ksperling-apple, line 408: These get initialized in `Init()`, so no need to repeat any defaults here.

src/controller/CHIPDeviceController.cpp
andy31415: Should this be in a code comment then?

src/controller/CHIPDeviceController.h
andy31415: Seems like good practice though .... please keep initializing them.

bzbarsky-apple (approved):
Approved, with @andy31415's comments addressed.

src/controller/CHIPDeviceController.h
ksperling-apple: I don’t particularly like having the defaults repeated in 3 places, but Ok.

src/controller/CHIPDeviceController.cpp
ksperling-apple: Added a comment

andy31415 (approved):

PR #32847, opened 2024-04-04
[Linux] Handle BLE adapter re-appearance due to BlueZ restart

arkq:
### Problem

In case of BlueZ restart (e.g. due to a crash and restart by systemd), all D-Bus proxy objects need to be refreshed (they keep unique name of watched D-Bus client). Otherwise, BLE functionality will be permanently broken in Matter SDK, until restart of the application.

### Changes

Implement logic for `kPlatformLinuxBLEAdapterAdded` and `kPlatformLinuxBLEAdapterRemoved` signals. In case of adapter disappearance, BlueZ integration is shutdown and BLE subsystem is disabled. Later when the adapter appears again, the BLE state machine re-registers with BlueZ in order to get new D-Bus unique name. The shutdown procedure does not stop timers, so the overall timeouts for BLE-related actions are not affected by adapter re-appearance.

### Testing

Manually tested LE advertising after BlueZ restart:
1. Restart BlueZ in the middle of advertising - overall advertising time is still 15 minutes.
2. Restart BlueZ after advertising is stopped - advertising is not started again.

### DIFF
--- a/src/platform/Linux/BLEManagerImpl.cpp
+++ b/src/platform/Linux/BLEManagerImpl.cpp
@@ -257,7 +257,8 @@ void BLEManagerImpl::HandlePlatformSpecificBLEEvent(const ChipDeviceEvent * apEv
                       apEvent->Platform.BLEAdapter.mAdapterAddress);
         if (apEvent->Platform.BLEAdapter.mAdapterId == mAdapterId)
         {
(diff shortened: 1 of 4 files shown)


yunhanw-google (approved):
thanks!

PR #32848, opened 2024-04-04
Timesync improvements - part 1

fessehaeve:
Fixes for issues form #27575 and more:

- made AttemptToGetTimeFromTrustedNode() public API
- free readclient memory after read is complete
- added delegate functions for better usability
- change cluster revision in light-switch-app

### DIFF
--- a/src/app/clusters/time-synchronization-server/DefaultTimeSyncDelegate.cpp
+++ b/src/app/clusters/time-synchronization-server/DefaultTimeSyncDelegate.cpp
@@ -78,3 +78,13 @@ void DefaultTimeSyncDelegate::UTCTimeAvailabilityChanged(uint64_t time)
 {
     // placeholder implementation
 }
+
+void DefaultTimeSyncDelegate::TrustedTimeSourceAvailabilityChanged(bool available, GranularityEnum granularity)
+{
+    // placeholder implementation
+}
+
+void DefaultTimeSyncDelegate::NotifyTimeFailure()
+{
+    // placeholder implementation
+}
--- a/src/app/clusters/time-synchronization-server/DefaultTimeSyncDelegate.cpp
+++ b/src/app/clusters/time-synchronization-server/DefaultTimeSyncDelegate.cpp
@@ -24,17 +24,6 @@
 using chip::TimeSyncDataProvider;
 using namespace chip::app::Clusters::TimeSynchronization;
 
-void DefaultTimeSyncDelegate::TimeZoneListChanged(const Span<TimeSyncDataProvider::TimeZoneStore> timeZoneList)
-{
-    // placeholder implementation
-}
-
-bool DefaultTimeSyncDelegate::HandleUpdateDSTOffset(chip::CharSpan name)
-{
-    // placeholder implementation
-    return false;
-}
-
 bool DefaultTimeSyncDelegate::IsNTPAddressValid(chip::CharSpan ntp)
 {
     // placeholder implementation
@@ -66,25 +55,4 @@ CHIP_ERROR DefaultTimeSyncDelegate::UpdateTimeFromPlatformSource(chip::Callback:
         return CHIP_NO_ERROR;
     }
     return CHIP_ERROR_NOT_IMPLEMENTED;
-}
-
-CHIP_ERROR DefaultTimeSyncDelegate::UpdateTimeUsingNTPFallback(const CharSpan & fallbackNTP,
-                                                               chip::Callback::Callback<OnFallbackNTPCompletion> * callback)
-{
-    return CHIP_ERROR_NOT_IMPLEMENTED;
-}
-
-void DefaultTimeSyncDelegate::UTCTimeAvailabilityChanged(uint64_t time)
-{
-    // placeholder implementation
-}
-
-void DefaultTimeSyncDelegate::TrustedTimeSourceAvailabilityChanged(bool available, GranularityEnum granularity)
-{
-    // placeholder implementation
-}
-
-void DefaultTimeSyncDelegate::NotifyTimeFailure()
-{
-    // placeholder implementation
 }
--- a/src/app/clusters/time-synchronization-server/time-synchronization-delegate.h
+++ b/src/app/clusters/time-synchronization-server/time-synchronization-delegate.h
@@ -56,7 +56,7 @@ public:
      *
      * @param timeZoneList new time zone list
      */
-    virtual void TimeZoneListChanged(const Span<TimeSyncDataProvider::TimeZoneStore> timeZoneList) = 0;
+    virtual void TimeZoneListChanged(const Span<TimeSyncDataProvider::TimeZoneStore> timeZoneList) {}
     /**
      * @brief Give the delegate the chance to call SetDSTOffset on the TimeSynchronizationServer with a list of
      * DST offsets based on the provided time zone name.  If the delegate does so, it should return true.
@@ -64,7 +64,7 @@ public:
      *
      * @param name name of active time zone
      */
-    virtual bool HandleUpdateDSTOffset(const CharSpan name) = 0;
+    virtual bool HandleUpdateDSTOffset(const CharSpan name) { return false; }
     /**
      * @brief Returns true if the provided string is a valid NTP address (either domain name or IPv6 address).
      *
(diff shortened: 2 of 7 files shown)


src/app/clusters/time-synchronization-server/DefaultTimeSyncDelegate.cpp
```diff
@@ -78,3 +78,13 @@ void DefaultTimeSyncDelegate::UTCTimeAvailabilityChanged(uint64_t time)
 {
     // placeholder implementation
 }
+
+void DefaultTimeSyncDelegate::TrustedTimeSourceAvailabilityChanged(bool available, GranularityEnum granularity)
+{
+    // placeholder implementation
```
andy31415, line 84: how about making them {} instead of `=0` in the base class to not need empty implementations always?

src/app/clusters/time-synchronization-server/time-synchronization-delegate.h
```diff
@@ -109,7 +109,9 @@ class Delegate
     /**
      * @brief Signals application that UTCTime has changed through the timesync cluster.
      */
-    virtual void UTCTimeAvailabilityChanged(uint64_t time) = 0;
+    virtual void UTCTimeAvailabilityChanged(uint64_t time)                                         = 0;
+    virtual void TrustedTimeSourceAvailabilityChanged(bool available, GranularityEnum granularity) = 0;
+    virtual void NotifyTimeFailure()                                                               = 0;
```
andy31415, line 114: how about passing the underlying CHIP_ERROR to this method, then the default implementation could log or do something, otherwise we do not know much besides "some failure occurred" which I would generally not know what to do with ... what do we expect implementations to do when they receive this? If I get a failure at some point, how do I know when the failure is gone? Some documentation on this would be useful.

fessehaeve: This is meant to be used as "time cannot be fetched from TrustedTimeSource". When time is available again, UTCTimeAvailabilityChanged() would be called.
But, passing why it failed could be done - just didn't see the importance of it for my use case.

src/app/clusters/time-synchronization-server/DefaultTimeSyncDelegate.cpp
fessehaeve: sounds good

src/app/clusters/time-synchronization-server/time-synchronization-delegate.h
andy31415: I think we should pass it in.

please also add documentation describing lifetime: that we have an error and you may get a UTCTimeAvailabilityChanged when the error condition goes away (and also describe if this is a backgroup automatic operation or if something needs to handle NotifyTimeFailurte in some way to trigger a retry)

src/app/clusters/time-synchronization-server/time-synchronization-delegate.h
```diff
@@ -104,12 +104,17 @@ class Delegate
...
+                                                  chip::Callback::Callback<OnFallbackNTPCompletion> * callback)
+    {
+        return CHIP_ERROR_NOT_IMPLEMENTED;
+    }
 
     /**
      * @brief Signals application that UTCTime has changed through the timesync cluster.
      */
-    virtual void UTCTimeAvailabilityChanged(uint64_t time) = 0;
+    virtual void UTCTimeAvailabilityChanged(uint64_t time) {}
```
cecille, line 115: I definitely like that this is in here now, but the delegate is now kind of a mix of service requests and notifications. There's likely only one place where we should be asking for time updates, but there are potentially LOTS of things that might want these notifications. Would it make sense to separate the platform delegate from a notifications class that can be registered by the apps?

src/app/clusters/time-synchronization-server/time-synchronization-delegate.h
```diff
@@ -104,12 +104,17 @@ class Delegate
...
+    virtual void TrustedTimeSourceAvailabilityChanged(bool available, GranularityEnum granularity) {}
```
cecille, line 116: What's the use case for this?

src/app/clusters/time-synchronization-server/time-synchronization-delegate.h
cecille, line 115: Is there a way for time to go from available -> unavailable? ie, does the time need to be optional or something?

Can you add in the description when this is expected to be called? ie, are you guaranteed to get a call to this function even if the time comes from an internal source?

src/app/clusters/time-synchronization-server/time-synchronization-server.h
```diff
@@ -123,6 +123,8 @@ class TimeSynchronizationServer : public FabricTable::Delegate
     // ReadClient::Callback functions
     void OnAttributeData(const ConcreteDataAttributePath & aPath, TLV::TLVReader * apData, const StatusIB & aStatus) override;
     void OnDone(ReadClient * apReadClient) override;
+
+    CHIP_ERROR AttemptToGetTimeFromTrustedNode();
```
cecille, line 127: why is this public now?

src/app/clusters/time-synchronization-server/time-synchronization-delegate.h
fessehaeve: That sounds like a good idea. This was added only to address the need for apps to get time when it becomes available. We can do the platform level additions in a separate effort.

src/app/clusters/time-synchronization-server/time-synchronization-delegate.h
fessehaeve: This is for the apps to be able to manage their time origin and decide if it needs to query for time.
Right now when TrustedTimeSource is set, the code attempt to get time from the source only if the current granularity is none. But this might be irrelevant for some apps.

src/app/clusters/time-synchronization-server/time-synchronization-delegate.h
fessehaeve: The name of the function might be the confusing aspect here since this gets called only when time is set through SetUTCTime(). I will add a description.

src/app/clusters/time-synchronization-server/time-synchronization-server.h
fessehaeve: It is so that apps can reuse this for periodic/random querying of the TrustedTimeSource. We use it to query time once a day. Just didn't want to write code which basically does the same thing.

src/app/clusters/time-synchronization-server/time-synchronization-delegate.h
fessehaeve: Had a look at this again and I only use this when we emit the time failure event. Therefore, I cannot put any error as a parameter here. I can add a description of what it means. Is there a way to be notified of events in the SDK internally for apps? If that exists I can remove this function.

src/app/clusters/time-synchronization-server/time-synchronization-delegate.h
fessehaeve: Also, I don't see a way to get rid of time once it is set. That is why I didn't consider making this optional. When there is a time failure, there is another function to notify this to the app - NotifyTimeFailure().

bzbarsky-apple (approved):
Approving, once @cecille's concerns are addressed.

Thank you for breaking this up into pieces!

src/app/clusters/time-synchronization-server/time-synchronization-delegate.h
andy31415: You have to plumb this more, however you do have different chip-errors. I see at least in the code:

```cpp
    if (GetDefaultNtp(span) != CHIP_NO_ERROR)
    {
        emitTimeFailureEvent(kRootEndpointId);
        return;
    }
    if (span.size() > kMaxDefaultNTPSize)
    {
        emitTimeFailureEvent(kRootEndpointId);
        return;
    }
    if (GetDelegate()->UpdateTimeUsingNTPFallback(span, &mOnFallbackNTPCompletion) != CHIP_NO_ERROR)
    {
        emitTimeFailureEvent(kRootEndpointId);
    }
```

so you have:
   - CHIP_ERROR from GetDefaultNtp
   - some custom error that should represent overly large span
   - CHIP_ERROR from UpdateTimeUsingNTPFallback

Right now this error is actually LOST so one cannot differentiate the type of error it gets and it is not even logged ... so if a time failure occurs I do not know if the span get failed, the span size is invalid or if the update fallback failed. We should not lose errors and when an error callback is called, the callback should be told what error happened to the best of our abilities. messages of "something failed" are hard to debug.

fessehaeve: I don't think this is a comment relevant to this PR. This code existed before and isn't even mine. Yes we should improve it but not in this PR.

bauerschwan (approved):

PR #32849, opened 2024-04-04
Timesync improvements - part 2

fessehaeve:
Fixes for issues form #27575 :
- simplify span usage
- encode mutablecharspan instead of creating a new charspan
- encode null when GetLocalTime() fails
- simplify granularity validity check
- change size comparison to compare against actual buffer size instead of attribute list size macro
- use unknown enum to check for invalid enums
- minor logging changes


src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
```diff
@@ -144,10 +144,10 @@ static bool emitDSTTableEmptyEvent(EndpointId ep)
 
     if (CHIP_NO_ERROR != error)
     {
-        ChipLogError(Zcl, "Unable to emit DSTTableEmpty event [ep=%d]", ep);
+        ChipLogError(Zcl, "DSTTableEmptyEvent failed");
```
andy31415, line 147: We should probably log error here as well for debugging

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
```diff
@@ -1090,19 +1088,18 @@ bool emberAfTimeSynchronizationClusterSetUTCTimeCallback(
     const auto & timeSource  = commandData.timeSource;
 
     auto currentGranularity = TimeSynchronizationServer::Instance().GetGranularity();
-    if (granularity < GranularityEnum::kNoTimeGranularity || granularity > GranularityEnum::kMicrosecondsGranularity)
+    if (granularity == GranularityEnum::kUnknownEnumValue)
```
andy31415, line 1091: is commandData.granularity guaranteed to have called `EnsureKnownEnumValue` to set things to unknown?

andy31415: same comment for below ... I would expect `EnsureKnownEnumValue` to be called whenever we validate against kUnknownEnumValue.

fessehaeve: How about this comment in the code?

```
    // All received enum values that are not listed above will be mapped
    // to kUnknownEnumValue. This is a helper enum value that should only
    // be used by code to process how it handles receiving and unknown
    // enum value. This specific should never be transmitted.
```

My understanding was that unknown enums that are received over the air are encoded to this value.

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
fessehaeve: Agree. Done.

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
```diff
@@ -564,6 +560,12 @@ CHIP_ERROR TimeSynchronizationServer::SetTimeZone(const DataModel::DecodableList
     uint8_t i   = 0;
     InitTimeZone();
 
+    if (items > mTimeZoneObj.timeZoneList.size())
```
cecille, line 563: Does this change mean that you're going to get changes to the time zone list even if there's an error returned?

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
```diff
@@ -589,20 +591,14 @@ CHIP_ERROR TimeSynchronizationServer::SetTimeZone(const DataModel::DecodableList
...
-                return CHIP_ERROR_IM_MALFORMED_COMMAND_DATA_IB;
-            }
             memset(tzStore.name, 0, sizeof(tzStore.name));
-            chip::MutableCharSpan tempSpan(tzStore.name, len);
+            chip::MutableCharSpan tempSpan(tzStore.name);
             if (CHIP_NO_ERROR != CopyCharSpanToMutableCharSpan(newTz.name.Value(), tempSpan))
             {
                 ReturnErrorOnFailure(LoadTimeZone());
-                return CHIP_IM_GLOBAL_STATUS(InvalidCommand);
+                return CHIP_IM_GLOBAL_STATUS(ConstraintError);
```
cecille, line 599: did we have a test failure here previously? If not, is there a test gap here?

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
fessehaeve: Since we do the load in the error case, the time zone won't be changing.

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
fessehaeve: This code was not working as expected and the tests never reached this corner case. However, with the string copying mechanism changes it was able to return InvalidCommand but that was wrong - it should have been ConstraintError.

bzbarsky-apple, line 599: So at this point we have already wiped out the value of `tzStore.name`, right?  Is that ok?

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
bzbarsky-apple: > is commandData.granularity guaranteed to have called EnsureKnownEnumValue to set things to unknown?

Yes.

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
```diff
@@ -1090,19 +1088,18 @@ bool emberAfTimeSynchronizationClusterSetUTCTimeCallback(
...
-    if (timeSource.HasValue() && (timeSource.Value() < TimeSourceEnum::kNone || timeSource.Value() > TimeSourceEnum::kGnss))
+    if (timeSource.HasValue() && timeSource.Value() == TimeSourceEnum::kUnknownEnumValue)
     {
         commandObj->AddStatus(commandPath, Status::InvalidCommand);
         return true;
     }
 
-    if (granularity != GranularityEnum::kNoTimeGranularity &&
-        (currentGranularity == GranularityEnum::kNoTimeGranularity || granularity >= currentGranularity) &&
+    if (granularity > currentGranularity &&
```
bzbarsky-apple, line 1102: So if `granularity == currentGranularity` we will ignore this call?  Why is that?  This code could use a comment explaining why the check is the way it is.

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
fessehaeve: We have the LoadTimeZone() to undo that. Or do you mean something else?

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
fessehaeve: Since we set `currentGranularity` to one lower than what is provided in the `SetUTCTime` command, if we let `granularity == currentGranularity` be valid then we will be lowering `currentGranularity` gradually. I don't think that is intended here. 
What is missing from this PR is setting Granularity to one lower value which is in the last PR of this series #32869 - a bad consequence of my attempt at breaking down the initial PR.

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
cecille: Functionally, what does this change do then? ie, why move this down here rather than having the early return at the top?

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
fessehaeve: @bzbarsky-apple any feedback?

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
```diff
@@ -679,6 +670,12 @@ CHIP_ERROR TimeSynchronizationServer::SetDSTOffset(const DataModel::DecodableLis
     size_t i     = 0;
     InitDSTOffset();
 
+    if (items > mDstOffsetObj.dstOffsetList.size())
```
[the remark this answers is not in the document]
cecille: OK, I get how that happened. It's definitely more complex. I suppose it's a more direct comparison, but messing with the size of the list, to check the incoming against the new size of the list, to then back out the changes seems odd.

@bzbarsky-apple - original comment was from you - WDYT here?

Does mDstOffsetObj always start from the first element of the mDST array?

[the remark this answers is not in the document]
cecille: also where is the long-name test that was previously failing? Is that running in the CI?

andy31415:
PR seems to have no author feedback on comments in a while now. Assuming stale and closing.

bzbarsky-apple:
This still needs to happen; this is addressing previous review issues...

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
```diff
@@ -145,10 +145,10 @@ static bool emitDSTTableEmptyEvent(EndpointId ep)
 
     if (CHIP_NO_ERROR != error)
     {
-        ChipLogError(Zcl, "Unable to emit DSTTableEmpty event [ep=%d]", ep);
+        ChipLogError(Zcl, "DSTTableEmptyEvent failed: %" CHIP_ERROR_FORMAT, error.Format());
```
bzbarsky-apple, line 148: Why are we no longer logging the endpoint?  This applies in various places.

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
```diff
@@ -593,20 +595,14 @@ CHIP_ERROR TimeSynchronizationServer::SetTimeZone(const DataModel::DecodableList
...
         {
-            size_t len = newTz.name.Value().size();
-            if (len > sizeof(tzStore.name))
-            {
-                ReturnErrorOnFailure(LoadTimeZone());
-                return CHIP_ERROR_IM_MALFORMED_COMMAND_DATA_IB;
-            }
             memset(tzStore.name, 0, sizeof(tzStore.name));
-            chip::MutableCharSpan tempSpan(tzStore.name, len);
+            chip::MutableCharSpan tempSpan(tzStore.name);
```
bzbarsky-apple, line 599:
```suggestion
            MutableCharSpan tempSpan(tzStore.name);
```

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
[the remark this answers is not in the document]
bzbarsky-apple: I think it's a lot safer to compare things to the actual buffer size you plan to work with, rather than some proxy for that buffer size that might stop being a correct proxy at some point...  That said, we could keep the check against the constant in the earlier position if we then also have a check against the same constant after the Init() call to make sure that our buffer size is what we think it should be.

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
bzbarsky-apple: The lowering argument makes sense, but again: this code could use a comment explaining why it's doing this.

bzbarsky-apple (approved):

andy31415:
Closing stale PR.

PR #32747, opened 2024-03-27
Add checks for localization-configuration cluster and time-format-localization cluster init callback

wqx6:
- `emberAfTimeFormatLocalizationClusterServerInitCallback` calls `ActiveCalendarType::Get()` without checking whether the cluster supports CalendarFormat feature.
- `emberAfLocalizationConfigurationClusterServerInitCallback` tries to set `ActiveLocale` without checking `validLocaleCached`.

PR #32850, opened 2024-04-04
Add check for calendar format feature when initializing time format localization cluster

shubhamdp:
`supported-calendar-types` and `active-calendar-type` from time format localization cluster should be enabled only if `Calendar Format` feature is enabled. Otherwise the cluster init routine try to read them and put up the error prints.

```
E (3766) chip[ZCL]: Failed to read calendar type with error: 0x86
```


shubhamdp:
This is duplicate of https://github.com/project-chip/connectedhomeip/pull/32747, closing!

PR #32852, opened 2024-04-04
[Darwin] MTRDeviceControllerStorageDelegate should support bulk store/fetch

jtung-apple:
This adds bulk store/fetch in the MTRDeviceControllerStorageDelegate, and the MTRDevice attribute storage implementation is changed to use it for bulk reads and writes if the storage delegate supports it.


jtung-apple:
Pushed wrong branch. closing this and will make new PR.

PR #32854, opened 2024-04-04
Fix initialization-order-fiasco

mwswartwout:
There is an initialization-order-fiasco between the static Server instance and the test overrides in messaging/ReliableMessageProtocolConfig.cpp (e.g. idleRetransTimeoutOverride). The call stack looks like this:

    #1 0x7f5995b4531f in chip::GetLocalMRPConfig() messaging/ReliableMessageProtocolConfig.cpp:108:36
    #2 0x7f59959a0699 in PairingSession protocols/secure_channel/PairingSession.h:247:53
    #3 0x7f59959a0699 in chip::CASESession::CASESession() protocols/secure_channel/CASESession.h:55:18
    #4 0x7f5995a0f2f9 in CASEServer protocols/secure_channel/CASEServer.h:34:5
    #5 0x7f5995a0f2f9 in chip::Server::Server() app/server/Server.h:392:5
    
    Fix this by making mLocalMRPConfig an Optional, so that we do not need to call GetLocalMRPConfig from PairingSession's constructor.


src/app/server/Server.h
```diff
@@ -386,13 +386,15 @@ class Server
         return System::SystemClock().GetMonotonicMicroseconds64() - mInitTimestamp;
     }
 
-    static Server & GetInstance() { return sServer; }
+    static Server & GetInstance()
+    {
+        static Server * sServer = new Server();
```
andy31415, line 391: do you need to use heap here? Could this be `static server sServer; return &sServer` ?

mwswartwout: There are potential issues with that approach regarding deinitialization: https://isocpp.org/wiki/faq/ctors#construct-on-first-use-v2

I don't feel qualified to say that they're not a problem here, but if you tell me they aren't a problem then I can change it :)

ksperling-apple: I don't think this is an initialization order problem as such... The problem is that `PairingSession` calls `GetLocalMRPConfig` during construction, rather than just before we actually start using it.

bzbarsky-apple, line 391: This is going to get detected as a leak by leak-detection tools, no?

mwswartwout: It's not a real leak, since the heap memory will still be reclaimed by the OS when the program exits (unless the Server destructor has some important side effects that will cause real leaks if they're not run). Our ASAN runs detect the intialization-order-fiasco but don't flag this.

mwswartwout: Ideally I would not have the static instance at all, and create a single that each platform must construct and destroy. But that could be a breaking change unless we have a place in common code to put the Server member.

mwswartwout: Ah yeah if we can move the GetLocalMRPConfig call out of the constructor that would also solve this.

bzbarsky-apple: Yeah, I'm not saying it's a real leak, just that it will lead to false positives in tools....

bzbarsky-apple: So the issue here is that when `CONFIG_BUILD_FOR_HOST_UNIT_TEST` is set the static constructor for `Server` can run before the static constructor for `idleRetransTimeoutOverride` and whatnot, and then we get use-before-initialization?

Realistically, calling GetLocalMRPConfig "too early" is just broken, because the whole point of not just those test overrides but SetLocalMRPConfig is that it can change.  And in practice, the subclasses of PairingSession do the following:

1. `CASESession::EstablishSession` (CASE initiator init) sets `mLocalMRPConfig`.
2. `CASESession::PrepareForSessionEstablishment` (CASE responder init) sets `mLocalMRPConfig`.
3. `PASESession::WaitForPairing` (PASE responder init) sets `mLocalMRPConfig`.
4. `PASESession::Pair` (PASE initiator init) sets `mLocalMRPConfig`.

So it really does not matter what `mLocalMRPConfig` is set to in the PairingSession constructor.  I would say we should just switch it to either default-initialization, of if that does not work because there is no default ctor just set it to GetDefaultMRPConfig(), with a comment explaining why it shouldn't use GetLocalMRPConfig() and that the value does not matter because actual pairing session instances always override it.

@mwswartwout thoughts?

arkq: I think that most tools (most that I know) will report that as not freed but still reachable memory because it's a static variable and such report in most cases is not fatal. There is a "ton" of allocations like that in glib library :D

Damian-Nordic: Or get rid of `mLocalMRPConfig` and just call `GetLocalMRPConfig()` inside `PairingSession::EncodeSessionParameters`?
The current behavior may be a bit counterintuitive that MRP config can be overriden but `PairingSession` may still use some cached value from the past (e.g. from when `PrepareForSessionEstablishment` was called). The main users of setting `mLocalMRPConfig`  upfront are probably unit tests but they could probably also set `idleRetransTimeoutOverride` if needed.

mwswartwout: @bzbarsky-apple @ksperling-apple @Damian-Nordic thoughts on making `mLocalMRPConfig` an `Optional` vs removing it entirely?

bzbarsky-apple: Sorry for the lag.  We could make it an Optional, but that has some storage cost (one extra word), right?   The benefit compared to what I suggested is that it would let us add those asserts, right?

I don't see how we can remove it without changing the API promise, that being that you can provide an override value when you start session establishment, right?  And maybe that's OK; there keeps being this tension between "inject everything so we don't have global functions" and "just get the thing when you need it", and if local MRP config is dynamically changeable (as it is) just getting it at the point when you need it makes the most sense.

mwswartwout: I don't see how removing `mLocalMRPConfig` changes the API promise of being able to provide override values. In fact what it does is allow you to provide the override later and still have it take effect. But if you continue to set the overrides before you start session establishment then the overrides will still be respected when they are fetched later.

bzbarsky-apple: Right now you are able to provide an override that is _not the same as what GetLocalMRPConfig() returns_.

bzbarsky-apple: > Where are those overrides stored

They're arguments to `CASESession::PrepareForSessionEstablishment`, `CASESession::EstablishSession`, and equivalent PASESession APIs.  If you're using those classes directly, you can just pass them in.

If you're using `CASESession` via `CASESessionManager`, say, then `CASESessionManager` can be initialized with a `CASESessionManagerConfig` which has `CASEClientInitParams` which allows passing in a `mrpLocalConfig`.  Generally speaking people shouldn't, but they can....

Now maybe we should in fact remove this API surface.  I'd be ok with that personally, but it might need some wider discussion.

mwswartwout: Ah okay yeah I got it now. Let me update mLocalMRPConfig to be an Optional and see what kind of size hit we take, and then removing it entirely can be a separate conversation.

bzbarsky-apple (approved):

andy31415 (approved):

PR #32801, opened 2024-04-01
Android - implement ChipDnssdStopBrowse

pgregorr-amazon:
This change adds the implementation for ChipDnssdStopBrowse in connectedhomeip/src/platform/android/DnssdImpl.cpp allowing DNS-SD discovery to be manally stopped before the timeout.

**Change summary**

1.	Implemented the ChipDnssdStopBrowse JNI function in connectedhomeip/src/platform/android/DnssdImpl.cpp which calls the NsdManagerServiceBrowser stopDiscover() Java method (in connectedhomeip/src/platform/android/java/chip/platform/NsdManagerServiceBrowser.java) This function stops ongoing browsing for services using DNS-SD.

**Testing**

Verified and tested locally with the Android tv-casting-app example app. By calling the stop discovery API form the tv-casting-app app GUI, we are able to stop NsdManagerDiscovery discovery, prior to timeout triggered stop discovery.

PR #32855, opened 2024-04-04
Android: Fix memory leak in DnssdImpl.cpp ChipDnssdStopBrowse()

pgregorr-amazon:
**Change Summary** 

Fixing context memory lean in connectedhomeip/src/platform/android/DnssdImpl.cpp ChipDnssdStopBrowse() pointed out by bzbarsky-apple in previous PR [32801](https://github.com/project-chip/connectedhomeip/pull/32801) after it merged. Leak would occur if the call to the Java method failed, threw an exception. 

**Testing**

Verified and tested locally with the Android tv-casting-app example app. By calling the stop discovery API form the tv-casting-app app GUI, we are able to stop NsdManagerDiscovery discovery, prior to timeout triggered stop discovery.

### DIFF
--- a/src/platform/android/DnssdImpl.cpp
+++ b/src/platform/android/DnssdImpl.cpp
@@ -224,6 +224,7 @@ CHIP_ERROR ChipDnssdStopBrowse(intptr_t browseIdentifier)
         ChipLogError(Discovery, "Java exception in ChipDnssdStopBrowse");
         env->ExceptionDescribe();
         env->ExceptionClear();
+        chip::Platform::Delete(ctx);
         return CHIP_JNI_ERROR_EXCEPTION_THROWN;
     }
     chip::Platform::Delete(ctx);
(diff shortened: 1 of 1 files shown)


sharadb-amazon (approved):

src/platform/android/DnssdImpl.cpp
```diff
@@ -224,6 +224,7 @@ CHIP_ERROR ChipDnssdStopBrowse(intptr_t browseIdentifier)
         ChipLogError(Discovery, "Java exception in ChipDnssdStopBrowse");
         env->ExceptionDescribe();
         env->ExceptionClear();
+        chip::Platform::Delete(ctx);
```
bzbarsky-apple, line 227: This seems ok, but wouldn't it be simpler to just delete it after the CallVoidMethod() before we do the ExceptionCheck() at all, since we are done with it at that point?

Either way, though.

bzbarsky-apple (approved):

src/platform/android/DnssdImpl.cpp
pgregorr-amazon: Right, at that point we used the callbackHandle in NsdManagerServiceBrowser stopDiscover() and returned so we don't need it anymore. Changed. Thanks!

andy31415 (approved):

PR #32858, opened 2024-04-04
[Darwin] MTRDeviceControllerStorageDelegate should support bulk store/fetch

jtung-apple:
This adds bulk store/fetch in the MTRDeviceControllerStorageDelegate, and the MTRDevice attribute storage implementation is changed to use it for bulk reads and writes if the storage delegate supports it.


src/darwin/Framework/CHIP/MTRDeviceControllerStorageDelegate.h
```diff
@@ -111,6 +111,34 @@ MTR_NEWLY_AVAILABLE
...
           sharingType:(MTRStorageSharingType)sharingType;
+
+@optional
+/**
+ * Return all keys and values stored, if any, for the provided controller in a
+ * dictionary. Returns nil if there is no stored value.
+ *
+ * securityLevel and dataType will always be the same for any given key value
+ * and are just present here to help locate the data if storage location is
+ * separated out by security level and data type.
```
bzbarsky-apple, line 122: I don't see how this works.  What happens if in fact data with different security levels and sharing types (not dataType? I guess we have that mis-naming in all the other comments too, and should fix it) is stored for the controller?

src/darwin/Framework/CHIP/MTRDeviceControllerStorageDelegate.h
```diff
@@ -111,6 +111,34 @@ MTR_NEWLY_AVAILABLE
     removeValueForKey:(NSString *)key
         securityLevel:(MTRStorageSecurityLevel)securityLevel
           sharingType:(MTRStorageSharingType)sharingType;
+
+@optional
+/**
+ * Return all keys and values stored, if any, for the provided controller in a
+ * dictionary. Returns nil if there is no stored value.
```
bzbarsky-apple, line 118:
```suggestion
 * Return all keys and values stored, if any, for the provided controller, in a
 * dictionary. Returns nil if there are no stored values.
```

src/darwin/Framework/CHIP/MTRDeviceControllerStorageDelegate.h
```diff
@@ -111,6 +111,34 @@ MTR_NEWLY_AVAILABLE
...
+- (nullable NSDictionary<NSString *, id<NSSecureCoding>> *)valuesForController:(MTRDeviceController *)controller
+                                                                 securityLevel:(MTRStorageSecurityLevel)securityLevel
+                                                                   sharingType:(MTRStorageSharingType)sharingType;
+
+/**
+ * Store a list of keys and values in the form of a dictionary. Returns whether
+ * the store succeeded.
+ *
+ * securityLevel and dataType will always be the same for any given key value
+ * and are present here as a hint to how the value should be stored.
```
bzbarsky-apple, line 136: I think the key part here is that the securityLevel and sharingType are the same across all the keys/values being stored in this store operation, right?

We need to very carefully document the expected behavior here if the stores fail partway through.  Is the callee expected to remove the already-stored things (roll back)?  Just leave things partially stored?  Something else?

bzbarsky-apple: The other question is whether this should be the _only_ store API.  The "one value" case can just be represented as a dictionary with one entry...  We have not shipped these APIs yet, so can make that change if we think it's the right one.

jtung-apple: Yes same securityLevel and sharingType for each bulk operation.

I think the failure behavior that makes sense is that the client rolls back. Allowing partial failure would mean introducing a fine-grained return, perhaps a NSSet of keys that failed to store? But that seems a bit too much to me.

And I think this being the only API could work, as the complexity we'll add for the API consumer is at worst a loop.

Currently I have the bulk APIs separate as optional, so that a consumer could choose to take on the complexity if they wish, or only implement the single store/fetch.

src/darwin/Framework/CHIP/MTRDeviceControllerStorageDelegate.h
jtung-apple: We are of course also stating that for any given key, the consumer should not expect the securityLevel and sharingType to change. But I'm not sure there is much to enforce/check here. The framework is making a promise and should not break it.

If we break the promise, and the API consumer in fact does have separate storage for each securityLevel/sharingType, then it'll be a framework bug that needs to be fixed.

bzbarsky-apple: My point is this: we are asking for all the keys.  But we are also providing the dataType and sharingType.

Does this actually mean that we are asking just for all the keys _for that dataType and sharingType_?  If so, we should clearly say that.

src/darwin/Framework/CHIP/MTRDeviceControllerDataStore.h
```diff
@@ -48,6 +48,13 @@ NS_ASSUME_NONNULL_BEGIN
                             storageDelegate:(id<MTRDeviceControllerStorageDelegate>)storageDelegate
                        storageDelegateQueue:(dispatch_queue_t)storageDelegateQueue;
 
+// clusterDataByNode a dictionary: nodeID => cluster data dictionary
+typedef void (^MTRDeviceControllerDataStoreClusterDataHandler)(NSDictionary<NSNumber *, NSDictionary<MTRClusterPath *, MTRDeviceClusterData *> *> * clusterDataByNode);
+/**
```
bzbarsky-apple, line 53:
```suggestion

/**
```

src/darwin/Framework/CHIP/MTRDeviceControllerDataStore.h
```diff
@@ -48,6 +48,13 @@ NS_ASSUME_NONNULL_BEGIN
...
+ * Asks the data store to load cluster data for nodes in bulk, if the storageDelegate supports it.
```
bzbarsky-apple, line 54: What's the contract around this?  Will clusterDataHandler not be called at all if the storageDelegate does not support bulk reads?  Something else?

Is clusterDataHandler called sync or async?  Caller certainly seems to asume sync, at least insofar as its lack of locking went.

src/darwin/Framework/CHIP/MTRDeviceControllerDataStore.mm
```diff
@@ -151,6 +151,17 @@ - (nullable instancetype)initWithController:(MTRDeviceController *)controller
     return self;
 }
 
+- (void)startUpWithClusterDataHandler:(MTRDeviceControllerDataStoreClusterDataHandler)clusterDataHandler
```
bzbarsky-apple, line 154: Seems to me like this is more like fetchAttributeDataForAllDevices or something?  This `startUpWithClusterDataHandler` is a bit odd, because we are not really starting up the controller data store...

src/darwin/Framework/CHIP/MTRDeviceControllerDataStore.mm
```diff
@@ -151,6 +151,17 @@ - (nullable instancetype)initWithController:(MTRDeviceController *)controller
...
+{
+    dispatch_sync(_storageDelegateQueue, ^{
+        if ([self->_storageDelegate respondsToSelector:@selector(valuesForController:securityLevel:sharingType:)]) {
+            NSDictionary<NSString *, id> * dataStoreSecureLocalValues = [self->_storageDelegate valuesForController:self->_controller securityLevel:MTRStorageSecurityLevelSecure sharingType:MTRStorageSharingTypeNotShared];
+
+            clusterDataHandler([self getClusterDataFromSecureLocalValues:dataStoreSecureLocalValues]);
```
bzbarsky-apple, line 160: So this calls `clusterDataHandler` while running on the delegate queue.  That seems a bit risky to me; what migt be better is if we do the dispatch_sync around just the `valuesForController` call, get the results into a `__block` `NSDictionary *`, and then do the `getClusterDataFromSecureLocalValues` and calling of the handler back on the original queue we got called on.

src/darwin/Framework/CHIP/MTRDeviceControllerDataStore.mm
```diff
@@ -519,6 +538,8 @@ - (BOOL)_deleteAttributeValueForNodeID:(NSNumber *)nodeID endpointID:(NSNumber *
 
 #pragma - Attribute Cache management
 
+#define ATTRIBUTE_CACHE_VERBOSE_LOGGING 1
```
bzbarsky-apple, line 541: I suspect this was not meant to be committed?

src/darwin/Framework/CHIP/MTRDeviceControllerDataStore.mm
```diff
@@ -968,6 +989,46 @@ - (void)clearAllStoredAttributes
     return clusterDataToReturn;
 }
 
+// Utility for constructing dictionary of nodeID to cluster data from dictionary of storage keys
+- (nullable NSDictionary<NSNumber *, NSDictionary<MTRClusterPath *, MTRDeviceClusterData *> *> *)getClusterDataFromSecureLocalValues:(NSDictionary<NSString *, id> *)secureLocalValues
```
bzbarsky-apple, line 993: Should this be named `_getClusterDataFromSecureLocalValues`?  I haven't figured out when we have the `_` prefix and when not...

src/darwin/Framework/CHIP/MTRDeviceControllerDataStore.mm
```diff
@@ -968,6 +989,46 @@ - (void)clearAllStoredAttributes
...
+        NSArray<NSNumber *> * endpointIndex = secureLocalValues[[self _endpointIndexKeyForNodeID:nodeID]];
+        for (NSNumber * endpointID in endpointIndex) {
+            NSArray<NSNumber *> * clusterIndex = secureLocalValues[[self _clusterIndexKeyForNodeID:nodeID endpointID:endpointID]];
+            for (NSNumber * clusterID in clusterIndex) {
+                MTRDeviceClusterData * clusterData = secureLocalValues[[self _clusterDataKeyForNodeID:nodeID endpointID:endpointID clusterID:clusterID]];
+                if (!clusterData) {
+                    continue;
+                }
+                if (![clusterData isKindOfClass:[MTRDeviceClusterData class]]) {
+                    continue;
```
bzbarsky-apple, line 1011: So why are we sanity-checking this but not sanity-checking that the various `NSArray<NSNumber *> *` we had above were actually arrays, and of numbers?

I'm really unclear on the trust model being used here.

src/darwin/Framework/CHIP/MTRDeviceControllerStorageDelegate.h
```diff
@@ -66,6 +66,10 @@ typedef NS_ENUM(NSUInteger, MTRStorageSharingType) {
  *    this delegate, apart from de-serializing and serializing the items being
  *    stored and calling MTRDeviceControllerStorageClasses(), is likely to lead
  *    to deadlocks.
+ *
+ * 3) Security level and sharing type will always be the same for any given key value
+ *    and are provided to describe the data should the storage delegate choose to
+ *    implement if separating storage location by security level and sharing type.
```
bzbarsky-apple, line 72:
```suggestion
 *    implement separating storage location by security level and sharing type.
```

src/darwin/Framework/CHIP/MTRDeviceControllerStorageDelegate.h
```diff
@@ -102,15 +99,38 @@ MTR_NEWLY_AVAILABLE
...
           sharingType:(MTRStorageSharingType)sharingType;
+
+@optional
+/**
+ * Return all keys and values stored, if any, for the provided controller, in a
+ * dictionary. Returns nil if there are no stored values.
+ *
+ * securityLevel and sharingType are provided as a hint for the storage delegate
+ * to load from the right security level and sharing type, if the implementation
+ * stores them separately. If the implementation includes keys from other
```
bzbarsky-apple, line 115:
```suggestion
 * stores them separately. If the implementation includes key/value pairs from other
```

src/darwin/Framework/CHIP/MTRDeviceControllerStorageDelegate.h
```diff
@@ -102,15 +99,38 @@ MTR_NEWLY_AVAILABLE
...
+ *
+ * The set of classes that might be decoded by this function is available by
+ * calling MTRDeviceControllerStorageClasses().
+ */
+- (nullable NSDictionary<NSString *, id<NSSecureCoding>> *)valuesForController:(MTRDeviceController *)controller
+                                                                 securityLevel:(MTRStorageSecurityLevel)securityLevel
+                                                                   sharingType:(MTRStorageSharingType)sharingType;
+
+/**
+ * Store a list of keys and values in the form of a dictionary. Returns whether
```
bzbarsky-apple, line 126:
```suggestion
 * Store a list of keys/value pairs in the form of a dictionary. Returns whether
```

src/darwin/Framework/CHIP/MTRDeviceControllerStorageDelegate.h
```diff
@@ -102,15 +99,38 @@ MTR_NEWLY_AVAILABLE
...
+ * the store succeeded. Specifically, if any keys in this dictionary fails to store,
```
bzbarsky-apple, line 127:
```suggestion
 * the store succeeded. Specifically, if any keys in this dictionary fail to store,
```

src/darwin/Framework/CHIPTests/MTRPerControllerStorageTests.m
```diff
@@ -1190,11 +1194,100 @@ - (void)test008_TestDataStoreDirect
...
+        [controller.controllerDataStore _clusterDataKeyForNodeID:@(3001) endpointID:@(1) clusterID:@(2)] : bulkTestClusterData12,
+        [controller.controllerDataStore _clusterDataKeyForNodeID:@(3001) endpointID:@(1) clusterID:@(3)] : bulkTestClusterData13,
+        [controller.controllerDataStore _clusterDataKeyForNodeID:@(3001) endpointID:@(2) clusterID:@(1)] : bulkTestClusterData21,
+        [controller.controllerDataStore _clusterDataKeyForNodeID:@(3001) endpointID:@(2) clusterID:@(2)] : bulkTestClusterData22,
+    };
+    // Bulk store with delegate
+    dispatch_sync(_storageQueue, ^{
+        [storageDelegate controller:controller storeValues:testBulkValues securityLevel:MTRStorageSecurityLevelSecure sharingType:MTRStorageSharingTypeNotShared];
+    });
+    // Verify that the store resluted in the correct values
```
bzbarsky-apple, line 1262:
```suggestion
    // Verify that the store resulted in the correct values
```

src/darwin/Framework/CHIPTests/MTRPerControllerStorageTests.m
```diff
@@ -1190,11 +1194,100 @@ - (void)test008_TestDataStoreDirect
...
+    dataStoreClusterData = [controller.controllerDataStore getStoredClusterDataForNodeID:@(3001)];
+    for (MTRClusterPath * path in dataStoreClusterData) {
```
bzbarsky-apple, line 1264: This will pass if dataStoreClusterData in fact returns an empty dictionary, no?

We should have a similar loop going over paths in `bulkTestclusterDataDictionary` and checking that they are all in dataStoreClusterData, I would think.

src/darwin/Framework/CHIPTests/MTRPerControllerStorageTests.m
```diff
@@ -1190,11 +1194,100 @@ - (void)test008_TestDataStoreDirect
...
+    };
+    MTRClusterPath * bulkTestClusterPath22 = [MTRClusterPath clusterPathWithEndpointID:@(2) clusterID:@(2)];
+    MTRDeviceClusterData * bulkTestClusterData22 = [[MTRDeviceClusterData alloc] init];
+    bulkTestClusterData22.dataVersion = @(22);
+    bulkTestClusterData22.attributes = @{
+        @(1) : @ { MTRTypeKey : MTRUnsignedIntegerValueType, MTRValueKey : @(221) },
+        @(2) : @ { MTRTypeKey : MTRUnsignedIntegerValueType, MTRValueKey : @(222) },
+        @(3) : @ { MTRTypeKey : MTRUnsignedIntegerValueType, MTRValueKey : @(223) },
+    };
+    NSDictionary<MTRClusterPath *, MTRDeviceClusterData *> * bulkTestclusterDataDictionary = @{
```
bzbarsky-apple, line 1238:
```suggestion
    NSDictionary<MTRClusterPath *, MTRDeviceClusterData *> * bulkTestClusterDataDictionary = @{
```

src/darwin/Framework/CHIPTests/MTRPerControllerStorageTests.m
```diff
@@ -1190,11 +1194,100 @@ - (void)test008_TestDataStoreDirect
...
+    // Verify that the store resluted in the correct values
+    dataStoreClusterData = [controller.controllerDataStore getStoredClusterDataForNodeID:@(3001)];
+    for (MTRClusterPath * path in dataStoreClusterData) {
+        XCTAssertEqualObjects(bulkTestclusterDataDictionary[path], dataStoreClusterData[path]);
+    }
+
+    // Now test bulk store through data store
+    [controller.controllerDataStore storeClusterData:bulkTestclusterDataDictionary forNodeID:@(3001)];
+    dataStoreClusterData = [controller.controllerDataStore getStoredClusterDataForNodeID:@(3001)];
+    for (MTRClusterPath * path in dataStoreClusterData) {
```
bzbarsky-apple, line 1271: See above.

src/darwin/Framework/CHIP/MTRDeviceController.mm
```diff
@@ -590,6 +590,22 @@ - (BOOL)startup:(MTRDeviceControllerStartupParamsInternal *)startupParams
...
+        [_controllerDataStore startUpWithClusterDataHandler:^(NSDictionary<NSNumber *, NSDictionary<MTRClusterPath *, MTRDeviceClusterData *> *> * _Nonnull clusterDataByNode) {
+            MTR_LOG_INFO("Loaded %lu nodes from storage", static_cast<unsigned long>(clusterDataByNode.count));
+            for (NSNumber * nodeID in clusterDataByNode) {
+                MTRDevice * device = [[MTRDevice alloc] initWithNodeID:nodeID controller:self];
+                NSDictionary * clusterData = clusterDataByNode[nodeID];
+                MTR_LOG_INFO("Loaded %lu cluster data from storage for %@", static_cast<unsigned long>(clusterData.count), device);
+                if (clusterData.count) {
+                    [device setClusterData:clusterData];
+                }
+                self->_nodeIDToDeviceMap[nodeID] = device;
```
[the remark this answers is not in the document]
jtung-apple: Okay I can certainly refactor this bit.

src/darwin/Framework/CHIP/MTRDeviceControllerDataStore.h
jtung-apple: I'll add these details.

src/darwin/Framework/CHIP/MTRDeviceControllerDataStore.mm
jtung-apple: I typically add a `_` if it's a method that is only used within the file, and not meant to be exposed. So you're right that this should get a `_` and I missed it. 👍

src/darwin/Framework/CHIP/MTRDeviceControllerDataStore.mm
jtung-apple: I think the trust model is that we don't trust what storage gives us what we stored. I'll add the sanity check above as well.

src/darwin/Framework/CHIP/MTRDeviceController.mm
```diff
@@ -590,6 +590,22 @@ - (BOOL)startup:(MTRDeviceControllerStartupParamsInternal *)startupParams
         return NO;
     }
 
+    if (_controllerDataStore) {
+        // If the storage delegate supports the bulk read API, then a dictionary of nodeID => cluster data dictionary would be passed to the handler. Otherwise this would be a no-op, and stored attributes for MTRDevice objects will be loaded lazily in -deviceForNodeID:.
+        [_controllerDataStore startUpWithClusterDataHandler:^(NSDictionary<NSNumber *, NSDictionary<MTRClusterPath *, MTRDeviceClusterData *> *> * _Nonnull clusterDataByNode) {
+            MTR_LOG_INFO("Loaded %lu nodes from storage", static_cast<unsigned long>(clusterDataByNode.count));
```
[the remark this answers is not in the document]
jtung-apple: I've added the logging of controller uuid. If it doesn't exist then the log will show "(null)", which shows that it's not set, and I feel that's alright?

src/darwin/Framework/CHIP/MTRDeviceController.mm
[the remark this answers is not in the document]
jtung-apple: Done.

src/darwin/Framework/CHIP/MTRDeviceControllerDataStore.mm
jtung-apple: Done.

src/darwin/Framework/CHIP/MTRDeviceControllerDataStore.mm
jtung-apple: Done.

src/darwin/Framework/CHIP/MTRDeviceControllerDataStore.mm
jtung-apple: Done.

src/darwin/Framework/CHIP/MTRDeviceControllerDataStore.mm
jtung-apple: Done.

src/darwin/Framework/CHIP/MTRDeviceControllerStorageDelegate.h
jtung-apple: I've consolidated the description and clarified a bit.

src/darwin/Framework/CHIP/MTRDeviceControllerStorageDelegate.h
jtung-apple: I've added that partial stores should just return NO.

src/darwin/Framework/CHIPTests/MTRPerControllerStorageTests.m
jtung-apple: Changed to just dictionary comparison.

src/darwin/Framework/CHIPTests/MTRPerControllerStorageTests.m
jtung-apple: Same as above.

src/darwin/Framework/CHIP/MTRDeviceController.mm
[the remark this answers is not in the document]
bzbarsky-apple: It's totally fine.  It's actually always set.  It's just the API that was not always available, but if you're using the ivar, everything is fine.  And in any case, the API is always available as of yesterday.

src/darwin/Framework/CHIP/MTRDeviceControllerDataStore.h
```diff
@@ -48,6 +48,14 @@ NS_ASSUME_NONNULL_BEGIN
                             storageDelegate:(id<MTRDeviceControllerStorageDelegate>)storageDelegate
                        storageDelegateQueue:(dispatch_queue_t)storageDelegateQueue;
 
+// clusterDataByNode a dictionary: nodeID => cluster data dictionary
+typedef void (^MTRDeviceControllerDataStoreClusterDataHandler)(NSDictionary<NSNumber *, NSDictionary<MTRClusterPath *, MTRDeviceClusterData *> *> * clusterDataByNode);
+
+/**
+ * Asks the data store to load cluster data for nodes in bulk. If the storageDelegate supports it, the handler will be called synchronously.
```
bzbarsky-apple, line 55:
```suggestion
 * Asks the data store to load cluster data for nodes in bulk. If the storageDelegate supports it, the handler will be called synchronously.
 * If the storageDelegate does not support it, the handler will not be called at all.
```

src/darwin/Framework/CHIP/MTRDeviceControllerDataStore.mm
```diff
@@ -151,6 +151,20 @@ - (nullable instancetype)initWithController:(MTRDeviceController *)controller
...
+    __block NSDictionary<NSNumber *, NSDictionary<MTRClusterPath *, MTRDeviceClusterData *> *> * clusterDataByNode = nil;
+    dispatch_sync(_storageDelegateQueue, ^{
+        if ([self->_storageDelegate respondsToSelector:@selector(valuesForController:securityLevel:sharingType:)]) {
+            NSDictionary<NSString *, id> * dataStoreSecureLocalValues = [self->_storageDelegate valuesForController:self->_controller securityLevel:MTRStorageSecurityLevelSecure sharingType:MTRStorageSharingTypeNotShared];
+
+            clusterDataByNode = [self _getClusterDataFromSecureLocalValues:dataStoreSecureLocalValues];
+        }
+    });
+
+    clusterDataHandler(clusterDataByNode);
```
bzbarsky-apple, line 165: Two issues here:

* There's no reason to run the conversion procedure on the delegate queue.
* We're calling our callback even if we have no data, right?
  
```suggestion
    __block NSDictionary<NSString *, id> * dataStoreSecureLocalValues = nil;
    dispatch_sync(_storageDelegateQueue, ^{
        if ([self->_storageDelegate respondsToSelector:@selector(valuesForController:securityLevel:sharingType:)]) {
            dataStoreSecureLocalValues = [self->_storageDelegate valuesForController:self->_controller securityLevel:MTRStorageSecurityLevelSecure sharingType:MTRStorageSharingTypeNotShared];

            clusterDataByNode = [self _getClusterDataFromSecureLocalValues:dataStoreSecureLocalValues];
        }
    });

    if (dataStoreSecureLocalValues) {
        clusterDataHandler([self _getClusterDataFromSecureLocalValues:dataStoreSecureLocalValues]);
    }
```

src/darwin/Framework/CHIP/MTRDeviceControllerStorageDelegate.h
```diff
@@ -102,15 +99,38 @@ MTR_NEWLY_AVAILABLE
...
+ *
+ * The set of classes that might be decoded by this function is available by
+ * calling MTRDeviceControllerStorageClasses().
+ */
+- (nullable NSDictionary<NSString *, id<NSSecureCoding>> *)valuesForController:(MTRDeviceController *)controller
+                                                                 securityLevel:(MTRStorageSecurityLevel)securityLevel
+                                                                   sharingType:(MTRStorageSharingType)sharingType;
+
+/**
+ * Store a list of keys/value pairs in the form of a dictionary. Returns whether
```
bzbarsky-apple, line 126:
```suggestion
 * Store a list of key/value pairs in the form of a dictionary. Returns whether
```

My typo last time....

bzbarsky-apple (approved):

src/darwin/Framework/CHIP/MTRDeviceControllerDataStore.mm
jtung-apple: Done.

PR #32859, opened 2024-04-04
Allow thermostat system mode to be set to dry and sleep modes

### LINKED ISSUE #31903: [BUG] Thermostat SystemModeEnum kDry and kSleep returns fail
### Reproduction steps

Using chip-tool to write system-mode with SystemModeEnum kDry or KSleep  (e.g., ./chip-tool thermostat write system-mode 8 1 1), returns 0x87 (CONSTRAINT_ERROR).

Seems that it is because of https://github.com/project-chip/connectedhomeip/blob/master/src/app/clusters/thermostat-server/thermostat-server.cpp#L447, where the Enum is larger than kFanOnly, it returns invalidvalue = constrainterror.

### Bug prevalence

always

83f8665f0177906a66b927b1a24e8ac038861c86

### Platform

ameba

### Platform Version(s)

_No response_

### Anything else?

_No response_

samryanuofl:
Fixes #31903 by allowing the thermostat mode to be written to `Dry` and `Sleep` modes

### DIFF
--- a/src/app/clusters/thermostat-server/thermostat-server.cpp
+++ b/src/app/clusters/thermostat-server/thermostat-server.cpp
@@ -440,7 +440,7 @@ MatterThermostatClusterServerPreAttributeChangedCallback(const app::ConcreteAttr
         }
         auto RequestedSystemMode = static_cast<SystemModeEnum>(*value);
         if (ControlSequenceOfOperation > ControlSequenceOfOperationEnum::kCoolingAndHeatingWithReheat ||
-            RequestedSystemMode > SystemModeEnum::kFanOnly)
+            RequestedSystemMode > SystemModeEnum::kSleep)
         {
             return imcode::InvalidValue;
         }


hasty:
We had some discussion about this in the HVAC group; no existing Matter thermostats have implemented FanOnly, Dry or Sleep. These values were copy/pasted from the Zigbee spec, and are probably a mistake, as these settings are better handled elsewhere (there was also confusion over what exactly "Sleep" meant in this context). We discussed deprecating them in an upcoming release.

So... I'd probably leave this functionality as is for now, as it's not a terrible thing that no one can implement these values.

samryanuofl:
@hasty Got it, thanks for responding. I'm working on a Room AC device type (which includes a Thermostat cluster) and it does have Dry and Fan Only modes. I guess for now I could map our dry mode to the System Mode `Cool` and map our fan only mode to System Mode `Off`, does that make sense? From a user interface perspective it seems like it would be better to have some way to display when the device is in those modes.

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -440,7 +440,7 @@ MatterThermostatClusterServerPreAttributeChangedCallback(const app::ConcreteAttr
         }
         auto RequestedSystemMode = static_cast<SystemModeEnum>(*value);
         if (ControlSequenceOfOperation > ControlSequenceOfOperationEnum::kCoolingAndHeatingWithReheat ||
-            RequestedSystemMode > SystemModeEnum::kFanOnly)
+            RequestedSystemMode > SystemModeEnum::kSleep)
```
bzbarsky-apple, line 443: Why are we doing this?  If this is meant to be checking for "valid system mode", then we should be checking that, not hardcoding values that will then break when a new value is added (like just happened).

bzbarsky-apple: And in particular, the right way to check would be `RequestedSystemMode != SystemModeEnum::kUnknownEnumValue`.

And the check for `ControlSequenceOfOperation` should be fixed too...

dusikasss:
> @hasty Понял, спасибо, что ответили. Я работаю над типом устройства для кондиционирования воздуха в помещении (которое включает в себя блок термостата), и у него есть режимы «Только осушение» и «Только вентилятор». Думаю, пока я могу сопоставить наш режим «Только осушение» с системным режимом `Cool`, имеет ли это смысл? Стоит ли мне избегать использования системного режима «Только вентилятор», если он будет устаревшим? Полагаю, его можно сопоставить с системным режимом `Off` С точки зрения пользовательского интерфейса, было бы лучше, если бы можно было как-то отображать, когда устройство находится в этих режимах.

I faced a similar dilemma in the implementation of a room air conditioner. I have a dry and fanonly mode. did you manage to find a solution in which attributes to implement this functionality?

dusikasss:
I've created issue  #36846. It proposes finding a method to detect supported operating modes for room air conditioner devices using the Thermostat cluster in Matter.

PR #32866, opened 2024-04-05
Refactored mDNS Browse Interface

su-shanka:
Refactored the existing browse/discover interface in Resolver class to a common interface for all types of node browse and made the delegates common for all browse. 
This also contains renaming changes across all files where resolver interfaces are used.

### DIFF
--- a/src/lib/dnssd/Types.h
+++ b/src/lib/dnssd/Types.h
@@ -205,7 +205,7 @@ inline constexpr size_t kMaxRotatingIdLen         = 50;
 inline constexpr size_t kMaxPairingInstructionLen = 128;
 
 /// Data that is specific to commisionable/commissioning node discovery
-struct CommissionNodeData
+struct DnssdNodeData
 {
     size_t rotatingIdLen                                      = 0;
     uint32_t deviceType                                       = 0;
@@ -220,13 +220,13 @@ struct CommissionNodeData
     char deviceName[kMaxDeviceNameLen + 1]                    = {};
     char pairingInstruction[kMaxPairingInstructionLen + 1]    = {};
 
-    CommissionNodeData() {}
+    DnssdNodeData() {}
 
     void Reset()
     {
         // Let constructor clear things as default
-        this->~CommissionNodeData();
-        new (this) CommissionNodeData();
+        this->~DnssdNodeData();
+        new (this) DnssdNodeData();
     }
 
     bool IsInstanceName(const char * instance) const { return strcmp(instance, instanceName) == 0; }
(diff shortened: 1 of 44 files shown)


andy31415 (approved):

src/lib/dnssd/Types.h
```diff
@@ -205,7 +205,7 @@ inline constexpr size_t kMaxRotatingIdLen         = 50;
 inline constexpr size_t kMaxPairingInstructionLen = 128;
 
 /// Data that is specific to commisionable/commissioning node discovery
-struct CommissionNodeData
+struct DnssdNodeData
```
Damian-Nordic, line 208: The struct name is quite vague as `OperationalNodeData` is technically also related to DNS-SD but I can't come up with a better alternative (though the structure seems to contain most PTR+SRV data so maybe something like `NodeServiceData` would make sense...) so treat it as a non-blocking comment.

Damian-Nordic (approved):

src/lib/dnssd/Types.h
su-shanka: There will be a follow-up PR where this structure will also be used for Operational Browse. That is the reason this was named as DnssdNodeData as it will be common for all node type browse delegates.

su-shanka: Combined PR for your reference : #32750

src/lib/dnssd/Types.h
```diff
@@ -220,13 +220,13 @@ struct CommissionNodeData
     char deviceName[kMaxDeviceNameLen + 1]                    = {};
     char pairingInstruction[kMaxPairingInstructionLen + 1]    = {};
 
-    CommissionNodeData() {}
+    DnssdNodeData() {}
```
bzbarsky-apple, line 223: So the problem is: most of the fields of this struct are meaningless for an operational discovery, no?  This struct is very much commissionee-discovery specific.  So naming it general "DnssdNodeData" (but leaving the "specific to commisionable/commissioning node discovery" comment above, and having all its fields be about commissionees) does not make sense.

Also, CommonResolutionData is _also_ "dnssd node data", but it's the data that is common to all discovery types, right?

src/lib/dnssd/Types.h
```diff
@@ -300,31 +300,33 @@ struct ResolvedNodeData
...
     }
 };
 
 /// Callbacks for discovering nodes advertising non-operational status:
 ///   - Commissioners
 ///   - Nodes in commissioning modes over IP (e.g. ethernet devices, devices already
 ///     connected to thread/wifi or devices with a commissioning window open)
-class CommissioningResolveDelegate
+class DiscoverNodeDelegate
 {
```
bzbarsky-apple, line 327: Also, the comments here no longer match the code...

su-shanka: As requested this PR does not have the functional changes of operational discovery support so comment is not updated here. However, I have updated the comment for the delegate in the functional PR.

src/lib/dnssd/Types.h
andy31415: I missed that during the first view - my thought on the PR was that this is common data, however this clearly is not, like `commissionerPasscode` is clearly not an operational discovery item.

PR #32869, opened 2024-04-05
Timesync improvements - part 3

fessehaeve:
Fixes for issues form https://github.com/project-chip/connectedhomeip/issues/27575 :

- improved time zone name changing check
- improved time zone name copying buffer size comparison
- removed EventFlag enum and its usage - https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242657708 
- changed mechanism by which time zone and dst tables are maintained - https://github.com/project-chip/connectedhomeip/pull/26082#discussion_r1242628037, https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242655854
- event generation is now part of maintaining time zone and dst tables. In some cases, event generation is handled by the caller of the functions that are updating the tables - https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242683543
- updated test that is checking validity of dst offset items
- store one less granularity in SetUTCTime command - https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242671601, https://github.com/project-chip/connectedhomeip/pull/26082/files#r1242676471
- use the correct error types - https://github.com/project-chip/connectedhomeip/pull/26082#discussion_r1242631916
- setting empty time zone list should result in ConstraintError
- simplify redundant code for loading TZ and DST
- https://github.com/project-chip/connectedhomeip/pull/26082#discussion_r1241227274
- make TimeSource attribute use AAI
- make it possible to initialise time sync server with any persistent storage implementation - rather than getting the default from Server

#### Testing 

This is tested by cert tests and unit tests


fessehaeve:
FYI - I am still figuring out a way to add unit tests to this.

fessehaeve:
> FYI - I am still figuring out a way to add unit tests to this.

The code can be tested right now with the changes I made. However, I still don't know what to test that is not covered by cert or yaml tests. Any tips would be appreciated.

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
```diff
@@ -551,15 +543,17 @@ CHIP_ERROR TimeSynchronizationServer::SetTimeZone(const DataModel::DecodableList
 
     if (lastTzState != TimeState::kInvalid)
     {
-        const auto & tzStore = GetTimeZone()[0];
-        lastTz.offset        = tzStore.timeZone.offset;
-        if (tzStore.timeZone.name.HasValue())
+        const TimeSyncDataProvider::TimeZoneStore & tzStore = GetTimeZone()[0];
+        lastTz.offset                                       = tzStore.timeZone.offset;
+        if (tzStore.timeZone.name.HasValue() && sizeof(name) >= sizeof(tzStore.name))
```
cecille, line 548: ok, so there's a resize(?) line in GetTimeZone()
    mTimeZoneObj.timeZoneList = mTimeZoneObj.timeZoneList.SubSpan(0, mTimeZoneObj.validSize);

Why not just move that re-size into where the list(s) are actually re-sized? Then the list is properly resized after the call to UpdateTimeZoneState and the getter function has no side effects. At that point, you're just directly using a class-scoped object with (I think) a class-owned buffer that won't go out of scope. And you can just directly use it as lastTz without making a copy.

Reasoning here - 1) Getters with side effects are misleading and 2) memcpy is a big bug source and makes me nervous.

If that's too much for this change, then can this at least be swapped over to use CopyCharSpanToMutableCharSpan?

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
```diff
@@ -570,20 +564,17 @@ CHIP_ERROR TimeSynchronizationServer::SetTimeZone(const DataModel::DecodableList
         const auto & newTz = newTzL.GetValue();
         if (newTz.offset < -43200 || newTz.offset > 50400)
         {
-            ReturnErrorOnFailure(LoadTimeZone());
-            return CHIP_ERROR_IM_MALFORMED_COMMAND_DATA_IB;
+            ExitNow(err = CHIP_IM_GLOBAL_STATUS(ConstraintError));
```
cecille, line 567: so some of these error messages have changed - do tests not complain about this? Are we not testing this? Or is this caught and changed elsewhere?

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
```diff
@@ -647,7 +648,9 @@ CHIP_ERROR TimeSynchronizationServer::LoadTimeZone()
 CHIP_ERROR TimeSynchronizationServer::ClearTimeZone()
 {
     InitTimeZone();
-    return mTimeSyncDataProvider.StoreTimeZone(GetTimeZone());
+    ReturnErrorOnFailure(mTimeSyncDataProvider.StoreTimeZone(GetTimeZone()));
+    emitTimeZoneStatusEvent(GetDelegate()->GetEndpoint());
```
cecille, line 652: I think the only place this function gets called right now is on Init() if the time zone fails to load. Is the intent to have that emit an event there? That seems odd. If that IS the intent, would it make more sense to move that up into the init?

This is a public method - can the app call this? Under what circumstances? If the intent is to allow this to be called by the application code, then it should probably push this though the SetTimeZone function so it can check if the current time zone is set.

examples/all-clusters-app/all-clusters-common/all-clusters-app.zap
```diff
@@ -4915,10 +4915,10 @@
               "side": "server",
               "type": "TimeSourceEnum",
               "included": 1,
-              "storageOption": "RAM",
+              "storageOption": "External",
```
cecille, line 4918: This is from mTimeSource now, right? So that's RAM.

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
fessehaeve: These errors are checked in the ember callbacks and that has been updated. So it doesn't bring change in the overall outcome.

examples/all-clusters-app/all-clusters-common/all-clusters-app.zap
fessehaeve: I don't think so. RAM means it is using attribute storage. Here it is using AAI and we don't need the accessors as well.

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
fessehaeve: Agree that emitting event during init is odd. Will remove it. 
My intention was that apps could decide to clear time zone for whatever reason. SetTimeZone can only be helpful if the app has the means to get a valid time zone setting without using timesync.

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
cecille: why would you need to re-size in multiple places if you move the resize into the update?

andy31415:
This is a 6-month old PR that has merge conflicts. Closing as stale.

bzbarsky-apple:
We still need to do this; it has been blocked on review from me because I could not figure out a sane way to review it, but I will get past that.

src/app/tests/suites/TestTimeSynchronization.yaml
```diff
@@ -127,15 +127,13 @@ tests:
               - name: "TimeZone"
                 value: []
       response:
-          values:
-              - name: "DSTOffsetRequired"
-                value: true
+          error: CONSTRAINT_ERROR
```
cecille, line 130: so this means the previous code wasn't returning the correct error value and the fact that this isn't failing any cert tests means we're not testing for this. 

Would you mind adding this to the certification tests as well? We'll have to be careful here since there are implementations in the field already that have the incorrect behavior so it probably need to be enforced from 1.5 onwards.

cecille (commented):
This generally looks good to me - good simplification. @andreilitvin - can you take a peek at the persistent storage stuff?

andy31415:
@fessehaeve this is now a very old PR ... could you update it? Should it be maintained?

src/app/tests/suites/TestTimeSynchronization.yaml
fessehaeve: Yes, this isn't tested in cert tests. That would need test plan change as well?

fessehaeve:
> @fessehaeve this is now a very old PR ... could you update it? Should it be maintained?

@andy31415 yes, I plan to finalise this and 1 other PR related to time-sync which has been hanging for a long time now.

src/app/clusters/time-synchronization-server/time-synchronization-server.h
```diff
@@ -104,13 +89,28 @@ class TimeSynchronizationServer : public FabricTable::Delegate
...
+     * Otherwise, it returns kActive since there is always a default time zone available.
+     * kStopped is not used in this function.
+     */
     TimeState UpdateTimeZoneState();
+    /**
+     * @brief Updates DST offset to the correct item from the DSTOffset list.
+     * If the DST offset has changed from the previous state or value, the function will emit a DSTStatus event.
+     *
+     * Returns kInvalid if an error occurrs or the DSTOffset list is empty.
+     * Returns kActive if a DST offset is being used.
```
bzbarsky-apple, line 111: Can this offset that's being used be 0?  Would be good to make that clear.

src/app/clusters/time-synchronization-server/time-synchronization-server.h
```diff
@@ -104,13 +89,28 @@ class TimeSynchronizationServer : public FabricTable::Delegate
...
+     * Returns kStopped if no DST offset is applicable at the current time.
```
bzbarsky-apple, line 112: I would have expected kStopped (i.e. "permanent item") to correspond to empty list, but maybe I am missing something? Spec certainly seems to allow empty list, right?

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
```diff
@@ -555,15 +547,18 @@ CHIP_ERROR TimeSynchronizationServer::SetTimeZone(const DataModel::DecodableList
 
     if (lastTzState != TimeState::kInvalid)
     {
-        const auto & tzStore = GetTimeZone()[0];
-        lastTz.offset        = tzStore.timeZone.offset;
-        if (tzStore.timeZone.name.HasValue())
+        const TimeSyncDataProvider::TimeZoneStore & tzStore = GetTimeZone()[0];
+        lastTz.offset                                       = tzStore.timeZone.offset;
+        chip::MutableCharSpan tempSpan(name, sizeof(name));
```
bzbarsky-apple:
```suggestion
        MutableCharSpan tempSpan(name);
```

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
```diff
@@ -555,15 +547,18 @@ CHIP_ERROR TimeSynchronizationServer::SetTimeZone(const DataModel::DecodableList
...
+        if (tzStore.timeZone.name.HasValue() &&
+            CHIP_NO_ERROR == CopyCharSpanToMutableCharSpan(tzStore.timeZone.name.Value(), tempSpan))
         {
-            lastTz.name.SetValue(CharSpan(name));
-            memcpy(name, tzStore.name, sizeof(tzStore.name));
+            lastTz.name.SetValue(CharSpan(tempSpan.data(), tempSpan.size()));
```
bzbarsky-apple:
```suggestion
            lastTz.name.SetValue(tempSpan);
```

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
```diff
@@ -616,30 +606,42 @@ CHIP_ERROR TimeSynchronizationServer::SetTimeZone(const DataModel::DecodableList
...
-        if ((tz.name.HasValue() && lastTz.name.HasValue()) && !(tz.name.Value().data_equal(lastTz.name.Value())))
+        if (tz.name.HasValue() != lastTz.name.HasValue() ||
+            ((tz.name.HasValue() && lastTz.name.HasValue()) && !(tz.name.Value().data_equal(lastTz.name.Value()))))
         {
             emit = true;
         }
         if (emit)
         {
-            mEventFlag = TimeSyncEventFlag::kTimeZoneStatus;
+            emitTimeZoneStatusEvent(GetDelegate()->GetEndpoint());
```
bzbarsky-apple, line 632: Pre-existing, but the right terminology is "generate", not "emit".  Followup to fix that, please?

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
```diff
@@ -616,30 +606,42 @@ CHIP_ERROR TimeSynchronizationServer::SetTimeZone(const DataModel::DecodableList
...
         }
     }
-    return mTimeSyncDataProvider.StoreTimeZone(GetTimeZone());
+
+exit:
+    if (err != CHIP_NO_ERROR)
+    {
+        // do not propagate if an error occurs here because it is not the user relevant error
```
bzbarsky-apple:
```suggestion
        // Restore our state to the last known-good one.
        //
        // Do not propagate failure from LoadTimeZone if an error occurs here because it is not the user relevant error
```

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
```diff
@@ -675,8 +678,10 @@ CHIP_ERROR TimeSynchronizationServer::SetDSTOffset(const DataModel::DecodableLis
         return ClearDSTOffset();
     }
 
-    auto newDstL = dstL.begin();
-    size_t i     = 0;
+    CHIP_ERROR err          = CHIP_NO_ERROR;
+    uint64_t lastValidUntil = 0;
```
bzbarsky-apple, line 686: The use of "goto exit" is pretty unfortunate.  It might have been better to just have two functions: one which returns on failures, and one which calls the other and rolls back the state if failure is returned...  But OK.

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
```diff
@@ -686,42 +691,44 @@ CHIP_ERROR TimeSynchronizationServer::SetDSTOffset(const DataModel::DecodableLis
...
-            return CHIP_ERROR_IM_MALFORMED_COMMAND_DATA_IB;
+            ExitNow(err = CHIP_IM_GLOBAL_STATUS(ConstraintError));
         }
     }
 
-    return mTimeSyncDataProvider.StoreDSTOffset(GetDSTOffset());
+exit:
+    if (err != CHIP_NO_ERROR)
+    {
+        // do not propagate if an error occurs here because it is not the user relevant error
```
bzbarsky-apple, line 727: Same suggestion here as for the timezone case.

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
```diff
@@ -779,13 +787,8 @@ CHIP_ERROR TimeSynchronizationServer::SetUTCTime(EndpointId ep, uint64_t utcTime
...
     }
     GetDelegate()->UTCTimeAvailabilityChanged(utcTime);
-    mGranularity  = granularity;
-    Status status = TimeSource::Set(ep, source);
-    if (!(status == Status::Success || status == Status::UnsupportedAttribute))
-    {
-        ChipLogError(Zcl, "Writing TimeSource failed.");
-        return CHIP_IM_GLOBAL_STATUS(Failure);
-    }
+    mGranularity = static_cast<GranularityEnum>(to_underlying(granularity) - 1);
```
bzbarsky-apple, line 794: I guess the other PR's strict "greater than" check is what keeps this from running off the bottom of the list?

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
```diff
@@ -381,12 +381,8 @@ void TimeSynchronizationServer::OnTimeSyncCompletionFn(TimeSourceEnum timeSource
...
         return;
     }
-    mGranularity  = granularity;
-    Status status = TimeSource::Set(kRootEndpointId, timeSource);
-    if (!(status == Status::Success || status == Status::UnsupportedAttribute))
-    {
-        ChipLogError(Zcl, "Writing TimeSource failed.");
-    }
+    mGranularity = granularity;
+    mTimeSource  = timeSource;
```
bzbarsky-apple: Changes to the value of the attribute need to mark it dirty, no?  This applies to a bunch of the attributes in this cluster, actually: it seems to be completely broken wrt subscriptions....

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
```diff
@@ -893,46 +900,69 @@ TimeState TimeSynchronizationServer::UpdateDSTOffsetState()
...
+        }
     }
-    if (!dstList[activeDstIndex].validUntil.IsNull() && dstList[activeDstIndex].validUntil.Value() <= chipEpochTime)
+    else if (dstList[activeDstIndex].validUntil.Value() <= chipEpochTime)
     {
-        if (activeDstIndex + 1 >= mDstOffsetObj.validSize) // last item in the list
+        if (activeDstIndex + 1 >= mDstOffsetObj.validSize)
         {
-            VerifyOrReturnValue(ClearDSTOffset() == CHIP_NO_ERROR, TimeState::kInvalid);
+            // this is the last item in the list that has already expired
```
bzbarsky-apple:
```suggestion
            // This is the last item in the list, and it has already expired.
```

src/app/clusters/time-synchronization-server/time-synchronization-server.cpp
```diff
@@ -1202,12 +1232,10 @@ bool emberAfTimeSynchronizationClusterSetTimeZoneCallback(
 
     if (response.DSTOffsetRequired)
     {
-        TimeState dstState = TimeSynchronizationServer::Instance().UpdateDSTOffsetState();
+        // attempt to emit DSTStatus event if DST was previously available
```
bzbarsky-apple, line 1254: "generate" throughout, not "emit".

andy31415:
Closing stale PR.

PR #32876, opened 2024-04-05
[Silabs] Reduce code size for the Lighting-app

jepenven-silabs:
Remove unused feature in the Lighting-app including 

- BLE features like scanner
- Fixed Label clusters
- READ_Client
- etc...

### DIFF
--- a/examples/lighting-app/silabs/build_for_wifi_args.gni
+++ b/examples/lighting-app/silabs/build_for_wifi_args.gni
@@ -18,6 +18,9 @@ silabs_sdk_target = get_label_info(":sdk", "label_no_toolchain")
 chip_enable_openthread = false
 import("${chip_root}/src/platform/silabs/wifi_args.gni")
 
+# Not needed for the Lighting-app
+chip_enable_read_client = false
+
 chip_enable_ota_requestor = true
 app_data_model =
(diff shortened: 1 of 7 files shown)


third_party/silabs/matter_support
andy31415, line 1: Are the silabs changes also expected? The summary says "remove unused bits" however not silabs matter support files which seem to contain ICD defines and config re-ordering

jmartinez-silabs: The changes in matter/si91x/siwx917/BRD4338A/autogen that you are pointing at are indeed changes from a commit before @jepenven-silabs lastest commit in that submodule. Those changes were not yet brought into usage here and are causing the CI build failure.

The changes are okay, but we must remove `sl_si91x_m4_ps.c` from the build.

jmartinez-silabs (approved):
Approved, but need to fix 917 build due to the unrelated changes brought up with the submodule update

third_party/silabs/matter_support
jepenven-silabs: Resolved ish ? I'll let a Silabs WiFi expert fix the behaviour on monday, for now the build should be fixed.

andy31415 (approved):

PR #32878, opened 2024-04-05
Detect some out of date submodules during bootstrap/activate

andy31415:
Since we removed auto submodule update from bootstrap, we got some complaints about not being aware that an update is required.

I added detection of a small subset of out of date modules:
  - pigweed since that really requires a new bootstrap (however on a fresh checkout things will be checked out already)
  - some small submodule subset that is always checked out

The intent here is to have a warning for folks doing `git checkout && source scripts/bootstrap.sh` that now this is not sufficient and you need submodules.


scripts/setup/bootstrap.sh
```diff
@@ -56,6 +56,29 @@ _install_additional_pip_requirements() {
...
 
+_submodules_need_updating() {
+  # Validates if a set o submodules that should always be checked out are up to date
+
+  # Pigweed will be up to date on an initial setup, however it may change over time
+  # the rest are a small subset of things that are always checked out (have no platform attachment)
+  _SUBMODULE_PATHS=(
+    "third_party/pigweed/repo"
+    "third_party/openthread/repo"
+    "third_party/editline/repo"
```
tcarmelveilleux, line 67: What about zap ?

andy31415: there is no zap submodule, we get zap via cipd.

arkq (approved):

scripts/setup/bootstrap.sh
```diff
@@ -56,6 +56,29 @@ _install_additional_pip_requirements() {
...
+  )
+
+  for path in "${_SUBMODULE_PATHS[@]}"; do
+    if git submodule status $path | grep -E '^-' >/dev/null 2>&1; then
```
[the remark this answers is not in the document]
andy31415: Updated.

Regarding shellharden, we explicitly ignore bootstrap: https://github.com/project-chip/connectedhomeip/blob/master/.restyled.yaml#L85 apparently because it wrongly tries to quote some things we do not want quoting.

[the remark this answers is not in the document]
arkq: > because it wrongly tries to quote some things we do not want quoting

Hmm... OK, I'll try to look at it, maybe it's not the fault of shellcheck but the looping pattern itself...

scripts/setup/bootstrap.sh
```diff
@@ -56,6 +56,29 @@ _install_additional_pip_requirements() {
     unset _SETUP_PLATFORM
 }
 
+_submodules_need_updating() {
+  # Validates if a set o submodules that should always be checked out are up to date
```
bzbarsky-apple, line 60: "of submodules"

scripts/setup/bootstrap.sh
```diff
@@ -56,6 +56,29 @@ _install_additional_pip_requirements() {
...
+
+  # Pigweed will be up to date on an initial setup, however it may change over time
+  # the rest are a small subset of things that are always checked out (have no platform attachment)
```
bzbarsky-apple, line 63: Uppercase "the", please, and probably periods at the ends of sentences in a multi-sentence comment?

PR #32880, opened 2024-04-05
Fix kotlin tests, remove unused code

ghost:
assertThat(false) doesn't actually do anything, it returns an object you can then make assertions against, use assertTrue instead

testQRCodeWithOptionalData relies upon an implicit ordering, however the underlying representation (HashMap) is unordered

### DIFF
--- a/src/controller/java/src/matter/onboardingpayload/QRCodeBasicOnboardingPayloadGenerator.kt
+++ b/src/controller/java/src/matter/onboardingpayload/QRCodeBasicOnboardingPayloadGenerator.kt
@@ -68,9 +68,6 @@ fun payloadBase38RepresentationWithTLV(
 
     // Copy the subBuffer back to the outBuffer
     subBuffer.copyInto(outBuffer, prefixLen)
-
-    // Reduce output buffer size to be the size of written data
-    outBuffer.copyOf(prefixLen + subBuffer.size)
   }
 }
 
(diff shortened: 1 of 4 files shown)


andy31415 (approved):

src/controller/java/src/matter/onboardingpayload/QRCodeBasicOnboardingPayloadGenerator.kt
```diff
@@ -68,9 +68,6 @@ fun payloadBase38RepresentationWithTLV(
 
     // Copy the subBuffer back to the outBuffer
     subBuffer.copyInto(outBuffer, prefixLen)
-
-    // Reduce output buffer size to be the size of written data
-    outBuffer.copyOf(prefixLen + subBuffer.size)
```
yufengwangca, line 73: outBuffer is used for output here, In Kotlin, the size of an array (CharArray in your case) cannot be changed once it is initialized because arrays have a fixed length, so 'outBuffer.copyOf(prefixLen + subBuffer.size)' just return a new copy, but does not change the original outBuffer.

It seems this is a bug, but I was wondering if we can simply remove this line without  effectively return the "size" or the number of elements that are actually used

yufengwangca: CharArray inherently handle the length, but it is better if we can add the test to test this, allocate more buffer to outBuffer than (prefixLen + subBuffer.size)

achaulk-goog: There are two codepaths that go through here. QRCodeBasicOnboardingPayloadGenerator takes a buffer but that doesn't seem to be used anywhere and payloadBase38Representation allocates a properly sized buffer itself

We could change this to return a new buffer instead of taking a provided one

yufengwangca (approved):

PR #32881, opened 2024-04-05
TC-TMP-2.1: Fix

cecille:
The constraints here are actually relatively complex because the min and max can be null and fall back to default values.

Updated the automation to python so we can do the appropriate tests, added tests for the various scenarios becuase there are a LOT and we don't cover any of them with our current examples.

Please see
https://github.com/CHIP-Specifications/chip-test-plans/pull/4096 for test plan updates


src/python_testing/TC_TMP_2_1.py
```diff
@@ -0,0 +1,97 @@
...
+        self.step(1)
+
+        self.step(2)
+        min_bound = -27315
+        max_bound = 32767
+
+        self.step(3)
+        min_measured_value = await self.read_single_attribute_check_success(cluster=cluster, attribute=attr.MinMeasuredValue)
+        if min_measured_value != NullValue:
+            min_bound = min_measured_value
```
tehampson, line 61: Should we check that `min_measured_value` is not lower than current `min_bound` as that is absolute 0 already.

src/python_testing/TC_TMP_2_1.py
```diff
@@ -0,0 +1,97 @@
...
+
+        self.step(4)
+        max_measured_value = await self.read_single_attribute_check_success(cluster=cluster, attribute=attr.MaxMeasuredValue)
+        if max_measured_value != NullValue:
+            max_bound = max_measured_value
```
tehampson, line 66: Similarly should we be checking that the value we get is not greater than `max_bound` which is currently max int16?

src/python_testing/TC_TMP_2_1.py
tehampson: NVM that is step 5

src/python_testing/TC_TMP_2_1.py
```diff
@@ -0,0 +1,97 @@
...
+            asserts.assert_greater(max_measured_value, min_bound,
+                                   "MaxMeasuredValue is not greater than the minimum bound")
+            asserts.assert_less_equal(
+                max_measured_value, 32767, "MaxMeasuredValue is not less than or equal to than 32767")
+
+        self.step(5)
+        if min_measured_value != NullValue:
+            asserts.assert_greater_equal(min_measured_value, -27315, "")
```
tehampson, line 74: Consider adding error log instead of empty string

tehampson (approved):

src/python_testing/TC_TMP_2_1.py
cecille: checked on line 69 below

src/python_testing/TC_TMP_2_1.py
cecille, line 74:
```suggestion
            asserts.assert_greater_equal(min_measured_value, -27315, "MinMeasuredValue is out of range")
```

src/python_testing/TC_TMP_2_1.py
```diff
@@ -0,0 +1,97 @@
...
+            asserts.assert_less(min_measured_value, max_bound)
```
cecille, line 75:
```suggestion
            asserts.assert_less(min_measured_value, max_bound, "MinMeasuredValue is out of range")
```

src/python_testing/TC_TMP_2_1.py
cecille: added.

PR #32883, opened 2024-04-06
Do not report attribute in _getAttributesToReportWithReportedValues i…

### LINKED ISSUE #32882: MTRDevice should not report values when there is an expected value and the expected value interval has not passed
Due to the above behavior, it is seen when we use slider on a fan control to change fan speed, The slider value on UI jumps between new value -> old value->new value. This can be fixed by not reporting a stale value if there is an expected value. The expected value interval when it expires should report the value.

nivi-apple:
…f an expected value is present

- The expiry timer for the expected value will take care of the reporting and we shouldn't report an attribute in the expected value interval if there is an existing expected value

Fixes: #32882

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -2144,13 +2144,8 @@ static BOOL AttributeHasChangesOmittedQuality(MTRAttributePath * attributePath)
             // if expected values exists, purge and update read cache
             NSArray * expectedValue = _expectedValueCache[attributePath];
             if (expectedValue) {
-                if (![self _attributeDataValue:attributeDataValue
-                            isEqualToDataValue:expectedValue[MTRDeviceExpectedValueFieldValueIndex]]) {
-                    shouldReportAttribute = YES;
-                    previousValue = expectedValue[MTRDeviceExpectedValueFieldValueIndex];
-                }
-                _expectedValueCache[attributePath] = nil;
                 _readCache[attributePath] = attributeDataValue;
+                shouldReportAttribute = NO;
             } else if (readCacheValueChanged) {
                 // otherwise compare and update read cache
                 previousValue = _readCache[attributePath];
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -2144,6 +2144,7 @@ static BOOL AttributeHasChangesOmittedQuality(MTRAttributePath * attributePath)
             // if expected values exists, purge and update read cache
             NSArray * expectedValue = _expectedValueCache[attributePath];
             if (expectedValue) {
+                previousValue = expectedValue[MTRDeviceExpectedValueFieldValueIndex];
                 _readCache[attributePath] = attributeDataValue;
                 shouldReportAttribute = NO;
             } else if (readCacheValueChanged) {


jtung-apple (approved):
Thank you for doing this work @nivi-apple ! This moves the expected value API to the original intention of optimizing for the success case, and keeping the value at the last set expected value, unless something unexpected happens, in which case the value will revert to the correct "last known value" either immediately after a write error, or at the expected value expiration time.

I've in the past put a TODO in the MTRDevice.h write API to better document this. Please either add some documentation as part of this PR, or file an issue for it. I thought I had done this but can't seem to find an issue for it.

nivi-apple:
One of the expected value test is failing. Investigating that. @jtung-apple will address your comment above as well.

nivi-apple:
Discussed offline with Jeff. Filed an issue and assigned to @jtung-apple https://github.com/project-chip/connectedhomeip/issues/32902. I think he would be the best person to tackle this.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -2144,13 +2144,9 @@ - (NSArray *)_getAttributesToReportWithReportedValues:(NSArray<NSDictionary<NSSt
             // if expected values exists, purge and update read cache
```
bzbarsky-apple, line 2144: This comment is now completely wrong.   There should be some comments here that actually explain what is going on, though, and those are very much missing.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -2144,13 +2144,9 @@ - (NSArray *)_getAttributesToReportWithReportedValues:(NSArray<NSDictionary<NSSt
...
             if (expectedValue) {
-                if (![self _attributeDataValue:attributeDataValue
-                            isEqualToDataValue:expectedValue[MTRDeviceExpectedValueFieldValueIndex]]) {
-                    shouldReportAttribute = YES;
-                    previousValue = expectedValue[MTRDeviceExpectedValueFieldValueIndex];
-                }
-                _expectedValueCache[attributePath] = nil;
+                previousValue = expectedValue[MTRDeviceExpectedValueFieldValueIndex];
                 _readCache[attributePath] = attributeDataValue;
+                shouldReportAttribute = NO;
```
bzbarsky-apple, line 2149: `shouldReportAttribute` is already NO at this point.  Why are we setting that again?

Similarly, the logging below in the !shouldReportAttribute case is completely wrong in the new setup, and will make people _extremely_ confused.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -2144,13 +2144,9 @@ - (NSArray *)_getAttributesToReportWithReportedValues:(NSArray<NSDictionary<NSSt
             // if expected values exists, purge and update read cache
             NSArray * expectedValue = _expectedValueCache[attributePath];
             if (expectedValue) {
-                if (![self _attributeDataValue:attributeDataValue
-                            isEqualToDataValue:expectedValue[MTRDeviceExpectedValueFieldValueIndex]]) {
-                    shouldReportAttribute = YES;
-                    previousValue = expectedValue[MTRDeviceExpectedValueFieldValueIndex];
-                }
-                _expectedValueCache[attributePath] = nil;
+                previousValue = expectedValue[MTRDeviceExpectedValueFieldValueIndex];
```
bzbarsky-apple, line 2147: Why are we setting `previousValue` to anything?  It should only be set (and is only used) when shouldReportAttribute == YES.

src/darwin/Framework/CHIP/MTRDevice.mm
nivi-apple: Filed https://github.com/project-chip/connectedhomeip/issues/33013

src/darwin/Framework/CHIP/MTRDevice.mm
nivi-apple: Filed https://github.com/project-chip/connectedhomeip/issues/33013

src/darwin/Framework/CHIP/MTRDevice.mm
nivi-apple: Filed https://github.com/project-chip/connectedhomeip/issues/33013

PR #32884, opened 2024-04-06
Temp fix for #32875 (network commissioning cluster check for SSID on Thread feature)

### LINKED ISSUE #32875: [1.3] NetworkCommissioning Cluster reads SSID field for Thread server instances
### Reproduction steps

There are some Matter Admins which may populate an empty SSID for the ScanNetworks command in NetworkCommissioning cluster in the field. The recent changes to the SDK now always checks if the SSID field exists regardless if the field is empty or not.

This will result in a Matter Accessory with these changes to always fail the ScanNetworks request.

The Matter specification states the following about the SSID FIeld:

```
11.9.7.1


SSID Field
This field, if present, SHALL contain the SSID for a directed scan of that particular Wi-Fi SSID. Otherwise, if the field is absent, or if it is null, this SHALL indicate scanning of all BSSID in range. This field SHALL be ignored for ScanNetworks invocations on non-WiFi server instances.
```

The last sentence indicates a requirement for the SSID field to be ignored for non-WiFi servers.

This was introduced in #32019 

Recommendation: 
Ignore the field entirely. Or at the very least, add check `if ssid.HasValue() && !ssid.Value().empty()`

### Bug prevalence

Always

660c928 

### Platform

core

### Platform Version(s)

1.3

### Type

Spec Compliance Issue

### Anything else?

_No response_

chrisdecenzo:
Fixes  #32875

As part of network commissioning cluster cleanup in 1.3, a check was added to the scan networks handler which checks for the SSID argument on a Thread device and returns an error when found.

There is confusing text in the spec which states that non-WiFi implementations should ignore this argument when null, and so the new check appears to not be conformant to that. But regardless, a popular commissioner was found to be passing a null SSID argument and this change breaks Thread device commissioning for users with that commissioner.

The spec needs to be clarified on this point, and the commissioner needs to be fixed. The commissioner fix is in the works but could take months to fully roll out to customers. In the meantime, this PR comments out the SSID check and a separate issue #32887 has been created to track that work.

Also, I have confirmed that this fix addresses the problem in #32875 (eg. no other fixes are needed).

### DIFF
--- a/src/app/clusters/network-commissioning/network-commissioning.cpp
+++ b/src/app/clusters/network-commissioning/network-commissioning.cpp
@@ -492,11 +492,11 @@ void Instance::HandleScanNetworks(HandlerContext & ctx, const Commands::ScanNetw
     else if (mFeatureFlags.Has(Feature::kThreadNetworkInterface))
     {
         // SSID present on Thread violates the `[WI]` conformance.
-        if (req.ssid.HasValue())
-        {
-            ctx.mCommandHandler.AddStatus(ctx.mRequestPath, Protocols::InteractionModel::Status::InvalidCommand);
-            return;
-        }
+        // if (req.ssid.HasValue())
+        // {
+        //     ctx.mCommandHandler.AddStatus(ctx.mRequestPath, Protocols::InteractionModel::Status::InvalidCommand);
+        //     return;
+        // }
 
         mCurrentOperationBreadcrumb = req.breadcrumb;
         mAsyncCommandHandle         = CommandHandler::Handle(&ctx.mCommandHandler);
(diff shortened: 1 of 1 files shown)


tcarmelveilleux (changes requested):
- Change description does not indicate what is being changed, or why, only links to an issue number, making tracing harder. Please describe the change.
- Change is removing a check that was deemed correct by spec, and spec interpretation should be validated with TSG before merging.

Note that TC-CNET-4.22 was missing the fix for TH == null (https://github.com/CHIP-Specifications/chip-test-plans/pull/3843), see TC-CNET-4.22 (https://github.com/CHIP-Specifications/chip-test-plans/blob/master/src/cluster/network_commissioning.adoc#4215-tc-cnet-422-thread-verification-for-scannetworks-command-dut-server), so step 3 implies that the test should pass. Given that the code at play was present at SVE, and nobody failed that test, it begs the question about whether the test was run at all!
  - See https://github.com/project-chip/connectedhomeip/blob/master/src/app/tests/suites/certification/Test_TC_CNET_4_22.yaml#L142

s-mcclain (approved):

tcarmelveilleux (approved):

robszewczyk (approved):

Damian-Nordic (approved):

PR #32889, opened 2024-04-08
Ble-WiFi commissioning test in CI

jlatusek:
## Problem
There is no testing of commissioning with ble-wifi in the CI

## Solution

Currently, there is no way to use virtual Bluetooth and wifi interfaces in CI. Github runner doesn't have
required modules to create virtual interfaces: [hci_vhci](https://www.kernelconfig.io/config_bt_hcivhci) and [mac80211_hwsim](https://wireless.wiki.kernel.org/en/users/drivers/mac80211_hwsim).
The proposed solution involves creating a custom qemu image with the required modules and using it in CI.

The created qemu image is based on [ubuntu:24.04](https://cloud-images.ubuntu.com/minimal/releases/focal/release/ubuntu-24.04-minimal-cloudimg-amd64.img) with a custom kernel and Bluez installed. The patch used to make the custom kernel is required to ensure that two devices can discover each other every time.

Tests use their dbus-daemon instance to separate them from the system dbus. This will allow developers
to run tests on their main system without needing a virtual machine. Unfortunately, there is no way to run two instances of Bluez simultaneously and select hci interfaces for each. That's why there is a proxy between the system and test buses, which forwards Bluez messages between them.

## Changes

1. Add Dockerfile with prepared qemu image (#32890).
2. Add CI job to run tests.
3. Modify testing scripts to prepare the environment for tests if ble-wifi is enabled.

PR #32890, opened 2024-04-08
Ble-WiFi commissioning test in CI - dockerfile

jlatusek:
This is a docker image required for related PR #32889


integrations/docker/images/stage-2/chip-build-linux-qemu/Dockerfile
```diff
@@ -0,0 +1,216 @@
...
+	&& apt-get update \
+	&& DEBIAN_FRONTEND=noninteractive apt-get install -fy \
+	cpu-checker \
+	qemu \
+	qemu-system-x86 \
+	&& apt-get clean \
+	&& rm -rf /var/lib/apt/lists/* \
+	&& rm -rf /var/cache/apt/* \
+	&& : # last line
+COPY --from=build-env ${UBUNTU_QEMU_DIR} ${UBUNTU_QEMU_DIR}
```
andy31415, line 214: We try to place every possible build and sdk into `chip-build-vscode`. Should this be done for the qemu image as well or is it too large or not viable to run in the chip-build-vscode image?

andy31415 (approved):
Approving as a stand-alone image creation. Maybe we can place the new image in chip-build-vscode in the future in case we want to run qemu tests in that image as well.

integrations/docker/images/stage-2/chip-build-linux-qemu/Dockerfile
jlatusek: The size of layers added to ghcr.io/project-chip/chip-build is about 400Mb. I feel like we can add it to chip-build-vscode. I just didn't think about it earlier.

PR #32893, opened 2024-04-08
[ESP32] Fix few attributes with fixed quality in DeviceInfoProvider

shubhamdp:
Fixed labels, supported locales, supported calendar types were being read from the nvs(flash) and during OTA its a hassle if one wants to upgrade these values.

Added few APIs to set the data for these attributes in ESP32DeviceInfoProvider.

#### Tests
- Read/write the attributes from time format localization, localization configuration, and fixed label attributes.

### DIFF
--- a/docs/guides/esp32/README.md
+++ b/docs/guides/esp32/README.md
@@ -18,3 +18,4 @@ example on ESP32 series of SoCs
 -   [Matter OTA](ota.md)
 -   [Generating and Using ESP Secure Cert Partition](secure_cert_partition.md)
 -   [BLE Settings](ble_settings.md)
+-   [Providers](providers.md)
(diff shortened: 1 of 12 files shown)


andy31415 (approved):

dhrishi (changes requested):
Does this also take care of backward compatibility?

If someone used the existing approach and now they want to do a OTA. Assuming that their fixed quality attributes have not changed, would they be mandated to implement these changes/APIs?

dhrishi (approved):

dhrishi:
@shubhamdp Please raise a backport PR on this branch https://github.com/espressif/connectedhomeip/tree/v1.3-branch

PR #32898, opened 2024-04-08
[ICD] Add TestEventTrigger support to the ICDManager

### LINKED ISSUE #31379: [ICD] Add TestEventTrigger for LIT that support enable/disable and specify specify the LIT time to maybe 30 seconds or so which can be used
In order to test LIT in certification test, we need add a capability in TestEventTrigger(https://github.com/project-chip/connectedhomeip/blob/caea3ec81b8c1f1e690323a00426301dd9c9a05a/src/app/TestEventTriggerDelegate.h#L27) that can control two flags, one flag can disable long idle time for non long idle time tests and another flag which can specify the LIT time to maybe 30 seconds or so which can be used for testing.

TestEventTrigger is an attribute in diagnostic cluster(https://github.com/CHIP-Specifications/connectedhomeip-spec/blob/0f07d0c7ad51aaf32c343d647dae8aa5ad768761/src/service_device_management/DiagnosticsGeneral.adoc#59-testeventtriggersenabled-attribute), tester can use TestEventTrigger to modify the above LIT things.

mkardous-silabs:
#### Description
PR Adds support for the TestEventTrigger to the ICDManager class

- Add infrastructure to support the TestEventTrigger
- Add Two TestEventTrigger events : AddActiveModeReq & RemoveActiveModeReq to be able to keep the device in ActiveMode until the remove command is sent

#### Tests

- Unit test to validate ICDManager::HandleTestEventTrigger function
- Python integration tests to validate that the ICDManager can process the incoming commands successfully

fixes: https://github.com/project-chip/connectedhomeip/issues/31379

Associated test plan PR : https://github.com/CHIP-Specifications/chip-test-plans/pull/4100


yunhanw-google (approved):

yunhanw-google (approved):

yunhanw-google:
Please follow-up with the below
1. Force the LIT behavior to be at a max sleep interval that is much lower than what the product would do in the field, so that test cases can be done with delays of >= ~30s (e.g. just out of SIT bounds), rather than waiting for real hours. (edited) 
2. Validate the ICD key refreshing bit via setting counter out of range in server via TestEventTrigger.
Thanks

src/app/icd/server/ICDManager.h
```diff
@@ -191,6 +200,12 @@ class ICDManager : public ICDListener
     uint8_t mOpenExchangeContextCount = 0;
 
 private:
+    enum class ICDTestEventTriggerEvent : uint64_t
+    {
+        kAddActiveModeReq    = 0x0046'0000'00000001,
+        kRemoveActiveModeReq = 0x0046'0000'00000002,
+    };
```
tcarmelveilleux, line 207: Move these constants to anonymous namespace in the .cpp file.

jmartinez-silabs (approved):

src/app/icd/server/ICDManager.h
tcarmelveilleux: Don't expose implementation details in the class when you can hide those details in the implementation. This enum is a private field, which means nobody else but the class implementation can use it. It can therefore live in the .cpp file anonymous namespace, and that reduces dependencies and API surface where the details don't need to be exposed.

mkardous-silabs: moved

PR #32911, opened 2024-04-09
Fix python evaluation of arithmetic statements with no spaces

tehampson:
Before yaml test would only properly evaluate `someVariable + 1`, but not `someVariable+1`. This improves evaluation capabilities to not rely on spaces to split up arithmetic operations

Test:
* CI Passes


yunhanw-google (approved):

scripts/py_matter_yamltests/matter_yamltests/parser.py
```diff
@@ -1223,7 +1223,8 @@ def _config_variable_substitution(self, value):
             # But some other tests were relying on the fact that the expression was put 'as if' in
             # the generated code and was resolved before being sent over the wire. For such
             # expressions (e.g 'myVar + 1') we need to compute it before sending it over the wire.
-            tokens = value.split()
+            delimiter_regex = "(\ |\+|\-|\*|\/|\%)"
```
andy31415, line 1226: is there any way for us to unit test some bits of these? It seems useful if we could have unit tests on yaml test parsing and interpreting.

cecille: We have some - maybe in test_yaml_parser.py?

tehampson: Added

bzbarsky-apple, line 1226: This should probably include `\(` and `\)` in the regex, so that things like `(myVariable + 4)/2` will work correctly.

scripts/tests/chiptest/__init__.py
```diff
@@ -164,6 +164,8 @@ def _GetChipToolUnsupportedTests() -> Set[str]:
     """Tests that fail in chip-tool for some reason"""
     return {
         "TestDiagnosticLogsDownloadCommand",  # chip-tool does not implement a bdx download command.
+        "Test_TC_TMP_2_1",  # Test incorrectly implemented as it relies on saving value that may be null, which maybe
```
bzbarsky-apple, line 167: If the test is being disabled _everywhere_, shouldn't it be moved out of CI to manual tests instead or something?  Otherwise we are left with the very incorrect impression that it actually runs in CI.

But also "may be" instead of "maybe", and the problem is not saving a value that may be null, it's doing arithmetic on a value that may be null, no?  And we should be able to fix the test to not do those checks if the value is null....  Followup for that, presumably.

tehampson: https://github.com/project-chip/connectedhomeip/pull/32881 Actually fixes the test.

I will wait for it to merge and merge mine afterwards. That way there will be no changed to this file or manualTests.json

Keeping it in for the time being so that I can confirm CI passes

andy31415 (approved):

scripts/py_matter_yamltests/test_yaml_parser.py
```diff
@@ -175,6 +178,81 @@
...
+          values:
+              - name: "arg"
+                value: myVariable/2
+
+    - label: "Arithmetic with parentheses"
+      command: "IntTest"
+      arguments:
+          values:
+              - name: "arg"
+                value: (myVariable -3)/ 2
```
bzbarsky-apple, line 253: I would suggest `(myVariable +3)/7` evaluating to 1...  The 0 makes it look too much like something that could happen by accident as a result of error handling.

bzbarsky-apple (approved):

yunhanw-google (approved):

PR #32914, opened 2024-04-09
Define a IM/DM decoupling interface for interaction model engine to interact with the data model bits

andy31415:
This starts implementing based on the proposal in https://hackmd.io/@6NmxK83qTm2u-ya_60bJEg/S1swIFbxR 

There is little implementation, mostly headers and virtual interface declarations.

The intent is to have changes reasonably small for review purposes rather than attempting one large PR at some future point in time. The downside on this is that we may find some things that don't map well and may need moving around as this is developed (especially command handling with batch support seems to potentially be complex).

### Changes

- define `app/interaction-model` (because data-model was taken)
- Most of the rest of the implementations are almost the same as proposed, except some changes to fix compilation
- Added build definitions and dependencies
- Added a single unit test for "event processing" to ensure that callbacks are called correctly and code compiles (since that bit is more template heavy than others)


bzbarsky-apple (commented):
I looked at some parts of this quickly, but in general this needs a very careful review which I would prefer to do once others have also had a chance to look at this....

src/app/interaction-model/paths.h
```diff
@@ -0,0 +1,37 @@
...
+ */
+#pragma once
+
+#include <app/AttributePathParams.h>
+
+namespace chip {
+namespace app {
+namespace InteractionModel {
+
+class Paths
```
bzbarsky-apple, line 25: InteractionModel::Paths says nothing about what this class is about....  What is it about?

src/app/interaction-model/request_context.h
```diff
@@ -0,0 +1,45 @@
...
+{
+public:
+    virtual ~RequestContext() = default;
+
+    /// Valid ONLY during synchronous handling of a Read/Write/Invoke
+    ///
+    /// Used sparringly, however some operations will require these:
+    ///   - Operational Credentials aborting communications on removed fabrics
+    ///   - General Commissioning failsafe handling
+    ///   - BDX operations and data transfer
```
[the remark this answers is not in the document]
andy31415: This was based on a grep on where the GetExchangeContext() was used. I can remove it and have shorter example lists.

src/app/interaction-model/invoke_responder.h
```diff
@@ -0,0 +1,108 @@
...
+
+#include <app/data-model/WrappedStructEncoder.h>
+#include <lib/core/CHIPError.h>
+
+namespace chip {
+namespace app {
+namespace InteractionModel {
+
+/// Aids in writing invoke replies:
+///   - Has access to the TLV writer of the response data content
```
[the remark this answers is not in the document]
andy31415: Will update to reference `ReplyEncoder`. I believe originally I wanted to have direct TLVWriter however then I saw that our typical repluies use a wrappedstructencoder.

I expect this may end up changing as we actually try to implement it. Overall I need something that I can cleanly say `Reply().Send(my_repy_struct)`

src/app/interaction-model/request_context.h
```diff
@@ -0,0 +1,45 @@
...
+
+// Context for a currently executing request
+class RequestContext
+{
+public:
+    virtual ~RequestContext() = default;
+
+    /// Valid ONLY during synchronous handling of a Read/Write/Invoke
+    ///
+    /// Used sparringly, however some operations will require these:
```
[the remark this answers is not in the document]
andy31415: Fixed `sparingly`.

src/app/interaction-model/actions.h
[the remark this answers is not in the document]
andy31415: Updated naming to `UpperCamelCase`

src/app/interaction-model/InvokeResponder.h
```diff
@@ -0,0 +1,192 @@
...
+        CHIP_ERROR err = mWriter->Complete(error);
+
+        if (err != CHIP_ERROR_BUFFER_TOO_SMALL)
+        {
+            return err;
+        }
+
+        // retry once. Failure to flush is permanent.
+        ReturnErrorOnFailure(mWriter->FlushPendingReplies());
+        return mWriter->Complete(error);
```
tehampson, line 133: How is data that was encoded into the buffer possibly moved to the new buffer here?

src/app/interaction-model/InvokeResponder.h
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    /// Sends the specified data structure as a response
+    ///
+    /// This version of the send has built-in RETRY and handles
+    /// Flush/Complete automatically.
+    ///
+    template <typename ReplyData>
+    CHIP_ERROR Send(const ReplyData & data)
+    {
```
tehampson, line 145: Should you check if `mCompleted` is already set to true here?

Also should you be setting it to true somewhere in the method?

src/app/interaction-model/InvokeResponder.h
```diff
@@ -0,0 +1,192 @@
...
+    InvokeResponder * mWriter;
+    bool mCompleted = false;
+};
+
+enum ReplyAsyncFlags
+{
+    // Some commmands that are expensive to process (e.g. crypto)
+    // should ack right away to improve reliability and reduce
+    // needless retries.
+    kFlushAckRightAway = 0x0001,
```
tehampson, line 171: In one of my recent PRs Boris wanted to move away from FlushAckRightAway to HandlingSlowCommand. That way apps don't care about underlying transport mechanism and simply provide a hint that they are going to take some time to handle slow command

src/app/interaction-model/IterationTypes.h
```diff
@@ -0,0 +1,93 @@
...
+#include <lib/core/DataModelTypes.h>
+#include <lib/support/BitFlags.h>
+
+namespace chip {
+namespace app {
+namespace InteractionModel {
+
+enum class ClusterQualityFlags : uint32_t
+{
+    kDiagnosticsData = 0x0001, // `K` quality, may be filtered out in subscriptions
```
tehampson, line 32: Clarification question: What is "K" quiality?

src/app/interaction-model/Paths.h
```diff
@@ -0,0 +1,48 @@
...
+
+#include <app/AttributePathParams.h>
+
+namespace chip {
+namespace app {
+namespace InteractionModel {
+
+/// Handles path attributes for interaction models.
+///
+/// It allows a user of the class to mark specific parths
```
tehampson, line 27: nit, :%s/parths/paths

src/app/interaction-model/tests/TestEventEmitting.cpp
```diff
@@ -0,0 +1,164 @@
...
+    CHIP_ERROR err = logOnlyEvents.DecodeLastEvent(decoded_event);
+
+    if (err != CHIP_NO_ERROR)
+    {
+        ChipLogError(EventLogging, "Decoding failed: %" CHIP_ERROR_FORMAT, err.Format());
+    }
+    ASSERT_EQ(err, CHIP_NO_ERROR);
+    ASSERT_EQ(decoded_event.softwareVersion, kFakeSoftwareVersion);
+
+    EventNumber n2 = events->GenerateEvent(event, 1);
```
tehampson, line 118: nit, `/* endpointId */  = 1`

src/app/interaction-model/tests/TestEventEmitting.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+    ASSERT_EQ(logOnlyEvents.LastOptions().mPath,
+              ConcreteEventPath(1 /* endpointId */, StartUpEventType::GetClusterId(), StartUpEventType::GetEventId()));
+}
+
+TEST(TestInteractionModelEventEmitting, TestFabricScoped)
+{
+    constexpr NodeId kTestNodeId           = 0x12ab;
+    constexpr uint16_t kTestPasscode       = 12345;
+    constexpr FabricIndex kTestFabricIndex = kMinValidFabricIndex + 10;
```
tehampson, line 130: nit, why `+ 10`

src/app/interaction-model/InvokeResponder.h
```diff
@@ -0,0 +1,192 @@
...
+    ///
+    /// MAY be called at most once.
+    ///
+    /// This function is intended to provided the ability to retry sending a reply
+    /// if a reply encoding fails due to insufficient buffer.
+    ///
+    /// Call this if `Complete(...)` returns CHIP_ERROR_BUFFER_TOO_SMALL and try
+    /// again. If reply data is needed, the complete ReplyEncoder + Complete
+    /// call chain MUST be re-run.
+    virtual CHIP_ERROR FlushPendingReplies() = 0;
```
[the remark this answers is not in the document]
andy31415: Yes, AutoCompleteInvokeResponder has automatic Flush handling it is Complete and Send calls.

src/app/interaction-model/InvokeResponder.h
```diff
@@ -0,0 +1,192 @@
...
+
+    /// Reply with a data payload.
+    ///
+    /// MUST be called at most once per reply.
+    /// Can be called a 2nd time after a `FlushPendingReplies()` call
+    ///
+    ///   - replyCommandId must correspond with the data encoded in the returned encoder
+    ///   - Complete(CHIP_NO_ERROR) MUST be called to flush the reply
+    virtual DataModel::WrappedStructEncoder & ReplyEncoder(CommandId replyCommandId) = 0;
```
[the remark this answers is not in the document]
andy31415: Updated all naming to use `Response` instead of `Reply`

src/app/interaction-model/InvokeResponder.h
andy31415: FlushPendingReplies is expected to clear all data corresponding to the "current reply" (i.e. it is a rollback after a backup on the TLV writer). So in this case I expect mWriter->Complete to re-do the entire "write path/ref & data".

this is why the `Send` side has both "data.Encode" and "Complete" after the flush.

src/app/interaction-model/InvokeResponder.h
```diff
@@ -0,0 +1,192 @@
...
+    /// Sends the specified data structure as a response
+    ///
+    /// This version of the send has built-in RETRY and handles
+    /// Flush/Complete automatically.
+    ///
+    template <typename ReplyData>
+    CHIP_ERROR Send(const ReplyData & data)
+    {
+        CHIP_ERROR err = data.Encode(ReplyEncoder(ReplyData::GetCommandId()));
+        LogErrorOnFailure(err);
```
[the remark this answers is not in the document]
andy31415: This was intentional saying that if encoding is bad then we should error out, however I think I missed the case of "insufficient buffer" where I do not need to complete with error just yet. Fixed.

src/app/interaction-model/InvokeResponder.h
```diff
@@ -0,0 +1,192 @@
...
+        err = mWriter->Complete(err);
+        if (err != CHIP_ERROR_BUFFER_TOO_SMALL)
+        {
+            return err;
+        }
+
+        // retry once. Failure to flush is permanent.
+        ReturnErrorOnFailure(mWriter->FlushPendingReplies());
+        err = data.Encode(ReplyEncoder(ReplyData::GetCommandId()));
+        LogErrorOnFailure(err);
```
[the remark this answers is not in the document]
andy31415: I believe this is correct: here we fail to encode data and it is a final failure, so in that case I cannot do anything better than reply "error" to the other side since they cannot receive the real data.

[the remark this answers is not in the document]
andy31415: Will update the code to return the "error" to the caller even if the complete of error returned success (and add comments about what is going on)

src/app/interaction-model/InvokeResponder.h
andy31415: Added checks and additional sets/resets.

src/app/interaction-model/Model.h
```diff
@@ -0,0 +1,117 @@
...
+    ///         - `UnsupportedWrite` for attempts to write read-only data
+    ///         - `UnsupportedAccess` for ACL failures
+    ///         - `NeedsTimedInteraction` for writes that are not timed however are required to be so
+    virtual CHIP_ERROR WriteAttribute(const WriteAttributeRequest & request, chip::TLV::TLVReader & attribute_data) = 0;
+
+    /// `responder` is used to send back the reply.
+    ///    - calling Reply() or ReplyAsync() will let the application control the reply
+    ///    - returning a CHIP_NO_ERROR without reply/reply_async implies a Status::Success reply without data
+    ///    - returning a CHIP_*_ERROR implies an error reply (error and data are mutually exclusive)
+    ///
```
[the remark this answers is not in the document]
andy31415: Added some documentation, hopefully in a better manner. In most cases it seems to me clients do not care sufficiently if a reply was fully sent to the other side since there is nothing they can actually do except logging, which the code would already do anyway.

src/app/interaction-model/tests/TestEventEmitting.cpp
andy31415: I have no reason for this, did not want to explicitly pick a min or a max. I felt that a constant with `Test` in name already meant `this is a magic number that I just picked` ... I was afraid that min may be too likely to hit special cases. Realistically it does not matter ... I can change it to anything if it seems better.

src/app/interaction-model/InvokeResponder.h
tehampson: My understanding is that from (Application + Data Model and Interaction Model) we should not need to know anything about the transport layer (exchange). The former should only care about indicating it is processing a command that will take and provide that hint to later layers. The later transport layers can translate that hint of processing a slow command to flushing a stand alone ack

src/app/interaction-model/InvokeResponder.h
```diff
@@ -103,23 +103,39 @@ class AutoCompleteInvokeResponder
...
+    }
 
     /// Call "Complete" without the automatic retries.
     ///
     /// Use this in conjunction with the other Raw* calls
-    CHIP_ERROR RawComplete(CHIP_ERROR error) { return mWriter->Complete(error); }
+    CHIP_ERROR RawComplete(CHIP_ERROR error)
+    {
+        VerifyOrReturnError(!mCompleted, CHIP_ERROR_INCORRECT_STATE);
+        mCompleted = true;
```
tehampson, line 126: Should we be setting `mCompleted` to true only if `mWriter->Complete(error)` was successful?

src/app/interaction-model/InvokeResponder.h
```diff
@@ -0,0 +1,227 @@
...
+
+class InvokeReply
+{
+public:
+    virtual ~InvokeReply() = default;
+
+    // reply with no data
+    CHIP_ERROR Reply(CHIP_ERROR e) { return this->Reply().Complete(e); }
+
+    // Send a reply "NOW" to the given invoke
```
tehampson, line 217: nit, this isn't exactly sending the reply right? This is adding the response to the reply. Sending is completely separate to this

src/app/interaction-model/InvokeResponder.h
```diff
@@ -0,0 +1,227 @@
...
+    virtual AutoCompleteInvokeResponder Reply() = 0;
+
+    // Reply "later" to the command. This allows async processing. A reply will be forced
+    // when the returned InvokeReply is destroyed.
+    virtual std::unique_ptr<InvokeReply> ReplyAsync(BitFlags<ReplyAsyncFlags> flags) = 0;
```
tehampson, line 222: I see this as a parallel to `CommandHandler::Handle`. With the `CommandHandler::Handle` mechanism, application code had the ability to reply to a command async without a heap. Do we want to attempt to do something similar here instead

src/app/interaction-model/InvokeResponder.h
```diff
@@ -0,0 +1,227 @@
...
+    /// MUST be called at most once per reply.
+    /// Can be called a 2nd time after a `FlushPendingResponses()` call
+    ///
+    ///   - responseCommandId must correspond with the data encoded in the returned encoder
+    ///   - Complete(CHIP_NO_ERROR) MUST be called to flush the reply
+    virtual DataModel::WrappedStructEncoder & ResponseEncoder(CommandId responseCommandId) = 0;
+
+    /// Signal completing of the reply.
+    ///
+    /// MUST be called exactly once to signal a reply to be sent.
```
tehampson, line 65: Nit, sent or added to the `InvokeResponseMessage`?

src/app/interaction-model/InvokeResponder.h
```diff
@@ -0,0 +1,227 @@
...
+    virtual ~InvokeResponder() = default;
+
+    // Copying not allowed since underlying requirement is that on deletion of this
+    // object, a reply will be sent.
+    InvokeResponder(const InvokeResponder &)             = delete;
+    InvokeResponder & operator=(const InvokeResponder &) = delete;
+
+    /// Flush any pending replies before encoding the current reply.
+    ///
+    /// MAY be called at most once.
```
tehampson, line 44: I am having a little trouble understanding this `MAY be called at most once` I can be called once by an instance of AutoCompleteInvokeResponder, but across multiple instance can this be called more than once?

Let's take a single InvokeRequestMessage that contains 3 batch commands. My understanding is that there will be:
* 1 Instance of `InvokeResponder`
* Up to 3 instance of `AutoCompleteInvokeResponder` all of which may need to call `FlushPendingResponses`

src/app/interaction-model/InvokeResponder.h
```diff
@@ -0,0 +1,227 @@
...
+    template <typename ReplyData>
+    CHIP_ERROR Send(const ReplyData & data)
+    {
+        VerifyOrReturnError(!mCompleted, CHIP_ERROR_INCORRECT_STATE);
+        mCompleted     = true;
+        CHIP_ERROR err = data.Encode(ResponseEncoder(ReplyData::GetCommandId()));
+        if (err != CHIP_ERROR_BUFFER_TOO_SMALL)
+        {
+            LogErrorOnFailure(err);
+            err = mWriter->Complete(err);
```
tehampson, line 168: If `FlushPendingReplies` is responsible for rolling back the buffer (as per the comment [here](https://github.com/project-chip/connectedhomeip/pull/32914/files#r1567540325)) what are we relying on here to rollback so that the `err` can be added here?

src/app/interaction-model/InvokeResponder.h
```diff
@@ -0,0 +1,227 @@
...
+#pragma once
+
+#include <app/data-model/WrappedStructEncoder.h>
+#include <lib/core/CHIPError.h>
+
+namespace chip {
+namespace app {
+namespace InteractionModel {
+
+/// Aids in writing invoke replies:
```
tehampson, line 26: Nit, This aids in writing and InvokeResponse for a single request.

Key aspect is that is it NOT replies (plural)

src/app/interaction-model/InvokeResponder.h
tehampson: My understanding was incorrect. There will be 3 instances of `InvokeResponder`

src/app/interaction-model/InvokeResponder.h
```diff
@@ -0,0 +1,192 @@
...
+    /// MUST be called exactly once to signal a reply to be sent.
+    /// If this returns CHIP_ERROR_BUFFER_TOO_SMALL, this can be called a 2nd time after
+    /// a FlushPendingReplies.
+    ///
+    /// Argument behavior:
+    ///  - If an ERROR is given (i.e. NOT CHIP_NO_ERROR) an error response will be given to the
+    ///    command.
+    ///  - If CHIP_NO_ERROR is given:
+    ///    - if a ReplyEncoder() was called, a data reply will be sent
+    ///    - if no ReplyEncoder() was called, a Success reply will be sent
```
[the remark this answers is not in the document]
tehampson: Just to confirm we are on the same page here the StatusIF and Cluster Errors would be in the `CHIP_ERROR error` parameter here. 

If so perhaps noting that in the comments might be helps for future implementer of handling a cluster command. Only reason I mention that is because today there is `AddClusterSpecificSuccess` and `AddStatus` that these callers have been using as of today and converting that may not be intuitive to someone that will be migrating code

src/app/interaction-model/InvokeResponder.h
tehampson: Now that I have clarified there is multiple InvokeResponse. I think I see the need for `std::unique_ptr`. But still might be good to see if there is a possible way we could do so without allocation.

src/app/interaction-model/InvokeResponder.h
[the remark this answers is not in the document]
andy31415: This was my intent, however looking at StatusIB::InitFromChipError there seems to be no way to say `Status::Success` but with a sdk error code, only Status::Failure seems to be covered.

This is quite unfortunate, however it seems we will need to handle it some way. Will take a look at alternatives.

src/app/interaction-model/InvokeResponder.h
andy31415: Updated the name.

src/app/interaction-model/InvokeResponder.h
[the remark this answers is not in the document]
andy31415: Ended up passing StatusIB for now. It has a clear conversion (even if explicit) from CHIP_ERROR so hopefully code will stay readable.

src/app/interaction-model/InvokeResponder.h
andy31415: The intent was to deny the other functions until you call RawFlush. However since this enforcement of logic using a single bool seemed not as robust, I now updated it to a full state machine that enforces the contract of "you may call complete ... and if that fails with buffer too small, you may call flush and complete again once".

src/app/interaction-model/InvokeResponder.h
andy31415: Updated comment.

src/app/interaction-model/InvokeResponder.h
andy31415: I have added a bit of comments. 

The difficulty I have here is that the calls are virtual since the intent is to be able to unit test these and create mocks. The closest equivalent we have are reference counting where we can use memory pools. I thought that using that path may increase complexity from the start. We can update logic to use ReferenceCounted maybe as a separate work stream.

src/app/interaction-model/InvokeResponder.h
andy31415: My expectation is that the writer is responsible to have a valid state, including rolling back data from commands in order to be able to encode an error. So in this case, `Encode` return an error so the writer is expected to send a status error.

The updated comments in the `Complete` of writers hopefully explain this more clearly.

src/app/interaction-model/InvokeResponder.h
andy31415: Updated comments significantly.

src/app/interaction-model/InvokeResponder.h
```diff
@@ -16,19 +16,44 @@
...
-///      - MUST be called exactly once
-///      - called with an error, makes TLV data written to be discarded (Invokes may only either
-///        succeed or fully fail)
+/// Handles encoding of an invoke response for a specific invoke request.
+///
+/// This class handles a single response (i.e. a CommandDataIB within the
+/// matter protocol) and is responsible for constructing its corresponding
+/// response (i.e. a InvokeResponseIB within the matter protocol)
+///
+/// Invoke responses MUST contain exacltly ONE of:
```
tehampson, line 33: nit, 
```suggestion
/// Invoke responses MUST contain exactly ONE of:
```

src/app/interaction-model/InvokeResponder.h
```diff
@@ -16,19 +16,44 @@
...
+///   - Complete with InteractionModel::Status::Success will respond with data
+///     some response data was written.
+///   - Any other case (including success with cluster specific codes) implies
+///     no response data and a status will be encoded instead
+///       - this includes the case when some response data was written already.
+///         In that case, the response data will be rolled back and only the status
+///         will be encoded.
+///
+/// Creating a response MUST be retried at most once, if and only if `Complete`
+/// returns CHIP_ERROR_BUFFER_TOO_SMALL:
```
tehampson, line 50: nit, is it required to always retry? If not perhaps something like the following

```suggestion
/// Creating a response MAY be retried at most once, if and only if `Complete`
/// returns CHIP_ERROR_BUFFER_TOO_SMALL. Retry attempts MUST not exceed 1:
```

src/app/interaction-model/InvokeResponder.h
```diff
@@ -181,29 +240,47 @@ class AutoCompleteInvokeResponder
...
+    // Contract says that complete may only be called twice:
+    //   - initial complete
+    //   - again after a `Flush`
+    // The states here expect we are in:
+    //
+    //    +----------------------------Flush---------|
+    //    |                                          v
+    //  NEVER --Complete--> F_EXPECTED --Flush--> FLUSHED --Complete--> COMPLETE
+    //             |                                                    ^
+    //             +-------------(success or permanent errror)----------|
```
tehampson, line 265: nit, 
```suggestion
    //             +-------------(success or permanent error)-----------|
```

src/app/interaction-model/InvokeResponder.h
```diff
@@ -181,29 +240,47 @@ class AutoCompleteInvokeResponder
...
+    CompleteState mCompleteState = CompleteState::kNeverCalled;
 };
 
 enum ReplyAsyncFlags
 {
-    // Some commmands that are expensive to process (e.g. crypto)
-    // should ack right away to improve reliability and reduce
-    // needless retries.
-    kFlushAckRightAway = 0x0001,
+    // Some commmands that are expensive to process (e.g. crypto).
```
tehampson, line 280: nit,
```suggestion
    // Some commands that are expensive to process (e.g. crypto).
```

src/app/interaction-model/Actions.h
```diff
@@ -0,0 +1,37 @@
...
+
+#include <app/interaction-model/Events.h>
+#include <app/interaction-model/Paths.h>
+#include <app/interaction-model/RequestContext.h>
+
+namespace chip {
+namespace app {
+namespace InteractionModel {
+
+/// Actual provided data for data models to interface with the interaction model environment
```
tehampson, line 27: nit, unclear what `Actual` means here and I think it can just be removed

src/app/interaction-model/OperationTypes.h
```diff
@@ -0,0 +1,90 @@
...
+enum class OperationFlags : uint32_t
+{
+    kInternal = 0x0001, // Internal request for data changes (can bypass checks/ACL etc.)
+};
+
+/// This information is available for ALL interactions: read/write/invoke
+struct OperationRequest
+{
+    OperationFlags operationFlags;
+    chip::Access::SubjectDescriptor subjectDescriptor; // Current authentication data
```
tehampson, line 41: In the case of an internal read/write/invoke is this actually provided?

Note I don't know too much about `Access::SubjectDescriptor`, but vaguely know that this comes from an exchange context so I just want to make sure that if we are considering the internal calls that we be careful with the comments about stating this is available to all (including internal)? We can make a comment on subjectDescriptor that is is only valid if not internal

src/app/interaction-model/OperationTypes.h
```diff
@@ -0,0 +1,90 @@
...
+enum class InvokeFlags : uint32_t
+{
+    kTimed = 0x0001, // Received as a 2nd command after a timed invoke
+};
+
+struct InvokeRequest : OperationRequest
+{
+    ConcreteCommandPath path;
+    Optional<GroupId> groupRequestId; // set if and only if this was a group request
+    BitFlags<InvokeFlags> writeFlags;
```
tehampson, line 85: nit, invokeFlags

tehampson (approved):

src/app/interaction-model/OperationTypes.h
andy31415: I think for internal it has no meaning. Will add a comment.

andy31415: Added comment and switched to optional, to at least have this consistently required to be validated.

I still do not like that we have this duality of both flags and descriptor ... if there are better suggestions let me know.

chrisdecenzo (approved):

src/app/interaction-model/OperationTypes.h
```diff
@@ -0,0 +1,94 @@
...
+struct ReadState
+{
+    // When reading lists, reading will start at this index.
+    // As list data is read, this index is incremented
+    ListIndex listEncodeStart = kInvalidListIndex;
+};
+
+enum class WriteFlags : uint32_t
+{
+    kTimed     = 0x0001, // Received as a 2nd command after a timed invoke
```
bzbarsky-apple, line 69: This doesn't make sense.  What do invokes have to do with WriteFlags?

Are there use cases for the data model knowing this information?

src/app/interaction-model/OperationTypes.h
```diff
@@ -0,0 +1,94 @@
...
+    kListBegin = 0x0002, // This is the FIRST list data element in a series of data
+    kListEnd   = 0x0004, // This is the LAST list element to write
```
bzbarsky-apple, line 71: This does not match documentation elsewhere.

src/app/interaction-model/OperationTypes.h
```diff
@@ -0,0 +1,94 @@
...
+{
+    // When reading lists, reading will start at this index.
+    // As list data is read, this index is incremented
+    ListIndex listEncodeStart = kInvalidListIndex;
+};
+
+enum class WriteFlags : uint32_t
+{
+    kTimed     = 0x0001, // Received as a 2nd command after a timed invoke
+    kListBegin = 0x0002, // This is the FIRST list data element in a series of data
```
bzbarsky-apple, line 70: It's the first list of data elements, no?

src/app/interaction-model/OperationTypes.h
```diff
@@ -0,0 +1,94 @@
...
+
+struct WriteAttributeRequest : OperationRequest
+{
+    ConcreteDataAttributePath path; // NOTE: this also contains LIST operation options (i.e. "data" path type)
+    BitFlags<WriteFlags> writeFlags;
+};
+
+enum class InvokeFlags : uint32_t
+{
+    kTimed = 0x0001, // Received as a 2nd command after a timed invoke
```
bzbarsky-apple, line 82: It's not a "2nd command".  Maybe you mean "2nd action"?  There is only one "command", or rather "list of commands", and it's in the "invoke".

Also, it's not "after a timed invoke".  A "timed invoke" is an interaction that involves a Timed Request action, then an Invoke Request action; this flag would mean this request is received as _part of_ a timed invoke.

src/app/interaction-model/OperationTypes.h
```diff
@@ -0,0 +1,94 @@
...
+};
+
+struct InvokeRequest : OperationRequest
+{
+    ConcreteCommandPath path;
+    std::optional<GroupId> groupRequestId; // set if and only if this was a group request
```
bzbarsky-apple, line 88: Why does the callee care whether this was a group request, much less what the group id involved was?  The data model should not care about that....

src/app/interaction-model/Paths.h
```diff
@@ -0,0 +1,48 @@
...
+ */
+#pragma once
+
+#include <app/AttributePathParams.h>
+
+namespace chip {
+namespace app {
+namespace InteractionModel {
+
+/// Handles path attributes for interaction models.
```
bzbarsky-apple, line 25: What are "path attributes"?

Was this supposed to be "attribute paths"?  But if so, the generic "Paths" naming is really odd.

src/app/interaction-model/Paths.h
```diff
@@ -0,0 +1,48 @@
...
+///
+/// It allows a user of the class to mark specific paths
+/// as having changed. The intended use is for some listener to
+/// perform operations as a result of something having changed,
+/// usually by forwarding updates (e.g. in case of subscriptions
+/// that cover that path).
+///
+/// Methods on this class MUCH be called from within the matter
+/// main loop as they will likely trigger interaction model
+/// internal updates and subscription event updates.
```
bzbarsky-apple, line 35: I don't know what "subscription event updates" are, especially as related to attribute paths.

This is back to the question of what this API is really about.  It seems we are very unclear on that.

src/app/interaction-model/Paths.h
```diff
@@ -0,0 +1,48 @@
...
+class Paths
+{
+public:
+    virtual ~Paths() = 0;
+
+    /// Mark some specific attributes dirty.
+    /// Wildcards are supported.
+    virtual void MarkDirty(const AttributePathParams & path) = 0;
```
bzbarsky-apple, line 43: "paths", not "path"?

src/app/interaction-model/RequestContext.h
```diff
@@ -0,0 +1,43 @@
...
+ */
+#pragma once
+
+#include <messaging/ExchangeContext.h>
+
+namespace chip {
+namespace app {
+namespace InteractionModel {
+
+// Context for a currently executing request
```
bzbarsky-apple, line 25: What is a "request", exactly?

This should be called ActionContext or something, to make sense to someone familiar with the spec.

src/app/interaction-model/RequestContext.h
```diff
@@ -0,0 +1,43 @@
...
+class RequestContext
+{
+public:
+    virtual ~RequestContext() = default;
+
+    /// Valid ONLY during synchronous handling of a Read/Write/Invoke
```
bzbarsky-apple, line 31: But not subscribe?  I'm sure it's valid during sync handling of a subscribe.  

I strongly suggest saying that this is only valid during synchronous handling of an action and leave it at that, without listing examples of actions.

src/app/interaction-model/Events.h
```diff
@@ -0,0 +1,130 @@
...
+
+#include <type_traits>
+
+namespace chip {
+namespace app {
+namespace InteractionModel {
+
+namespace internal {
+template <typename T>
+class SimpleEventLoggingDelegate : public EventLoggingDelegate
```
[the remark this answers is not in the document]
andy31415: I duplicated that one becasue EventLogging.h is already in-use and tied to `src/app:app` gn target. I tried to ensure that existing interaction-model classes are reasonably stand-alone because coupling in `src/app:app` and `src/app:interaction-model`  are very coupled and uncoupling them seemed hard. Ialso originally wanted a simple interface however it does seem I ended up with it being identical.

I can merge them together, however I was thinking I may want a merge at a later time once we start using the new interaction-model bits.

src/app/interaction-model/Events.h
```diff
@@ -0,0 +1,130 @@
...
+    eventOptions.mPath        = path;
+    eventOptions.mPriority    = aEventData.GetPriorityLevel();
+    eventOptions.mFabricIndex = aEventData.GetFabricIndex();
+
+    // this skips logging the event if it's fabric-scoped but no fabric association exists yet.
+
+    if (eventOptions.mFabricIndex == kUndefinedFabricIndex)
+    {
+        ChipLogError(EventLogging, "Event encode failure: no fabric index for fabric scoped event");
+        return kInvalidEventId;
```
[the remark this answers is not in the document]
andy31415: Mis-read the constants and did not make the association that `Id` is not the same as `number`. Switching it to return optional.

src/app/interaction-model/InvokeResponder.h
```diff
@@ -0,0 +1,316 @@
...
+/// Invoke responses MUST contain exactly ONE of:
+///   - response data (accessed via `ResponseEncoder`)
+///   - A status, which may be success or failure, both of which may
+///     contain a cluster-specific error code.
+///
+/// To encode a response, `Complete` MUST be called.
+///
+/// `Complete` requirements
+///   - Complete with InteractionModel::Status::Success will respond with data
+///     some response data was written.
```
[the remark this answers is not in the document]
andy31415: I think we will iterate a bit here: I wanted to have a separate class with a more ergonomic design, however it seems that it would require sigificant changes to integrate into the existing APIs.

As a result, I will move invoke to use some `CommandHandler` usage instead, so this class becomes obsolete. All implementation logic in it then will go away, so will not try hard to update it.

src/app/interaction-model/IterationTypes.h
```diff
@@ -0,0 +1,94 @@
...
+/// Provides metadata information for a data model
+///
+/// The data model can be viewed as a tree of endpoint/cluster/attribute
+/// where each element can be iterated through independently
+///
+/// Iteration rules:
+///   - kInvalidEndpointId will be returned when iteration ends (or generally kInvalid* for paths)
+///   - Any internal iteration errors are just logged (callers do not handle iteration CHIP_ERROR)
+///   - Iteration order is NOT guaranteed, however uniqueness and completeness is (must iterate
+///     over all possible distinct values as long as no internal structural changes occur)
```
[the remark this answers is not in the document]
andy31415: Will re-phrase. Intent is to allow any order of cluster ids inside an endpoint of attribute ids inside a acluster, however not switching between clusters/attributes while iterating.

I assume this is what you meant?

src/app/interaction-model/Model.h
```diff
@@ -0,0 +1,123 @@
...
+
+    // During the transition phase, we expect a large subset of code to require access to
+    // event emitting, path marking and other operations
+    virtual InteractionModelActions CurrentActions() { return mActions; }
+
+    /// List reading has specific handling logic:
+    ///   `state` contains in/out data about the current list reading. MUST start with kInvalidListIndex on first call
+    ///
+    /// Return codes:
+    ///   CHIP_ERROR_MORE_LIST_DATA_AVAILABLE (NOTE: new error defined for this purpose)
```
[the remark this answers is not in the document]
andy31415: I will revise this as I implement ReadAttribute. I assumed I may touch things around when implementing, so I did not define this yet.

src/app/interaction-model/OperationTypes.h
andy31415: I believe here I matched data that seemed to be used by our implementations, to make sure we can implement existing functionality using the new interface without providing typical access via Exchange/Session.

One example I have seen is https://github.com/project-chip/connectedhomeip/blob/master/src/app/clusters/on-off-server/on-off-server.cpp#L628 which does a:

```cpp

            GroupId groupId = ZCL_SCENES_GLOBAL_SCENE_GROUP_ID;
            if (commandObj->GetExchangeContext()->IsGroupExchangeContext())
            {
                groupId = commandObj->GetExchangeContext()->GetSessionHandle()->AsIncomingGroupSession()->GetGroupId();
            }
```

So I figured it seems we want information if we are a group request or not. If this is not that typical, we could indeed demand access through the ExchangeContext on this. Maybe we can iterate when integrating?

This seemed somewhat natural because group seemed to have some similarity to a path - this is how the request came in.

src/app/interaction-model/Paths.h
andy31415: It is a single path I believe, even if wildcard.

src/app/interaction-model/Events.h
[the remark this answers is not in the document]
bzbarsky-apple: Merging later is fine, I guess, as long as it happens.

src/app/interaction-model/Events.h
```diff
@@ -0,0 +1,130 @@
...
+        ChipLogError(EventLogging, "Failed to log event: %" CHIP_ERROR_FORMAT, err.Format());
+        return kInvalidEventId;
+    }
+
+    return eventNumber;
+}
+
+} // namespace internal
+
+class Events
```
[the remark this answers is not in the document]
bzbarsky-apple: I think "paths" is very weird because there are lots of types of paths.  Do "event paths" go under Paths or under Events?

src/app/interaction-model/IterationTypes.h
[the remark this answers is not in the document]
bzbarsky-apple: Yes.

src/app/interaction-model/Model.h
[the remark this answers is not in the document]
bzbarsky-apple: Then please make it clear that this is an error _to be defined_.  Because what the text says is that is has been defined, but it hasn't, and that really confused me.

src/app/interaction-model/OperationTypes.h
bzbarsky-apple: Um... That code is just wrong.  Spec says to use group id 0 in this case!  I'll check with the people who wrote it, but I am 90% sure it will get changed to not use the group id.

>  this is how the request came in.

That's already available in the subject.  It should not be attached to the command bits.

src/app/interaction-model/Paths.h
bzbarsky-apple: OK.

src/app/interaction-model/InvokeResponder.h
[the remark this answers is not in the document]
andy31415: I plan to change the invoke calls to be passed through a `CommandHandler` so I will not try to update InvokeResponder at all for now, will do that as a separate path towards CommandHandler.

src/app/interaction-model/InvokeResponder.h
```diff
@@ -0,0 +1,316 @@
...
+///   - Any other case (including success with cluster specific codes) implies
+///     no response data and a status will be encoded instead
+///       - this includes the case when some response data was written already.
+///         In that case, the response data will be rolled back and only the status
+///         will be encoded.
+///
+/// Creating a response MAY be retried at most once, if and only if `Complete`
+/// returns CHIP_ERROR_BUFFER_TOO_SMALL. Retry attempts MUST not exceed 1:
```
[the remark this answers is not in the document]
andy31415: Based on above, will not fix InvokeResponder yet (keeping only one comment open as a tracking reminder)

src/app/interaction-model/InvokeResponder.h
```diff
@@ -0,0 +1,316 @@
...
+    ///
+    /// Call this if `Complete(...)` returns CHIP_ERROR_BUFFER_TOO_SMALL and try
+    /// again. If reply data is needed, the complete ResponseEncoder + Complete
+    /// call chain MUST be re-run.
+    virtual CHIP_ERROR FlushPendingResponses() = 0;
+
+    /// Reply with a data payload.
+    ///
+    /// MUST be called at most once per reply.
+    /// Can be called a 2nd time after a `FlushPendingResponses()` call
```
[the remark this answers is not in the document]
andy31415: Based on above, will not fix InvokeResponder yet (keeping only one comment open as a tracking reminder)

src/app/interaction-model/InvokeResponder.h
```diff
@@ -0,0 +1,316 @@
...
+    ///
+    ///   - responseCommandId must correspond with the data encoded in the returned encoder
+    ///   - Complete(CHIP_NO_ERROR) MUST be called to flush the reply
+    ///
+    /// If encoder returns CHIP_ERROR_BUFFER_TOO_SMALL, FlushPendingResponses should be
+    /// used to attempt to free up buffer space then encoding should be tried again.
+    virtual DataModel::WrappedStructEncoder & ResponseEncoder(CommandId responseCommandId) = 0;
+
+    /// Signal completing of the reply.
```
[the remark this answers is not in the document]
andy31415: Based on above, will not fix InvokeResponder yet (keeping only one comment open as a tracking reminder)

src/app/interaction-model/InvokeResponder.h
```diff
@@ -0,0 +1,316 @@
...
+    /// a FlushPendingResponses.
+    ///
+    /// Argument behavior:
+    ///  - Commands can only be replied with ONE of the following (spec 8.9.4.4):
+    ///      - command data (i.e. ResponseEncoder contents)
+    ///      - A status (including success/error/cluster-specific-success-or-error )
+    ///  - As a result there are two possible paths:
+    ///      - IF a Status::Success is given (WITHOUT cluster specific status), then
+    ///        the data in ResponseEncoder is sent as a reply. If no data was sent,
+    ///        a invoke `Status::Success` with no cluster specific data is sent
```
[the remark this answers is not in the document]
andy31415: Based on above, will not fix InvokeResponder yet (keeping only one comment open as a tracking reminder)

src/app/interaction-model/InvokeResponder.h
```diff
@@ -0,0 +1,316 @@
...
+        CHIP_ERROR err = mWriter->Complete(status);
+
+        if (err != CHIP_ERROR_BUFFER_TOO_SMALL)
+        {
+            return err;
+        }
+
+        // retry once. Failure to flush is permanent.
+        ReturnErrorOnFailure(mWriter->FlushPendingResponses());
+        return mWriter->Complete(status);
```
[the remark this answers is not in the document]
andy31415: Based on above, will not fix InvokeResponder yet (keeping only one comment open as a tracking reminder)

src/app/interaction-model/InvokeResponder.h
```diff
@@ -0,0 +1,316 @@
...
+    ///        the data in ResponseEncoder is sent as a reply. If no data was sent,
+    ///        a invoke `Status::Success` with no cluster specific data is sent
+    ///      - OTHERWISE any previously encoded data via ResponseEncoder is discarded
+    ///        and the given reply (success with cluster status or failure) is sent
+    ///        as a reply to the invoke.
+    ///
+    ///
+    /// Returns success/failure state. One error code MUST be handled in particular:
+    ///
+    ///   - CHIP_ERROR_BUFFER_TOO_SMALL will return IF AND ONLY IF the responder was unable
```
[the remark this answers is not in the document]
andy31415: Based on above, will not fix InvokeResponder yet (keeping only one comment open as a tracking reminder)

src/app/interaction-model/InvokeResponder.h
```diff
@@ -0,0 +1,316 @@
...
+
+enum ReplyAsyncFlags
+{
+    // Some commands that are expensive to process (e.g. crypto).
+    // Implementations may choose to send an ack on the message right away to
+    // avoid MRP retransmits.
+    kSlowCommandHandling = 0x0001,
+};
+
+class InvokeReply
```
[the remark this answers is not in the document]
andy31415: Based on above, will not fix InvokeResponder yet (keeping only one comment open as a tracking reminder)

src/app/interaction-model/Model.h
```diff
@@ -0,0 +1,123 @@
...
+namespace InteractionModel {
+
+/// Represents operations against a matter-defined data model.
+///
+/// Class is SINGLE-THREADED:
+///   - operations are assumed to only be ever run in a single event-loop
+///     thread or equivalent
+///   - class is allowed to attempt to cache indexes/locations for faster
+///     lookups of things (e.g during iterations)
+class Model : public AttributeTreeIterator
```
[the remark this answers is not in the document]
andy31415: I want to call it `DataModel` and have it reside in `data-model` as paths, however data-model is already taken. As a result, I would prefer to have renames as a separate "this is renaming only" PR.

src/app/interaction-model/OperationTypes.h
andy31415: Timed flag applies to invoke and writes. This signals that a write has been received as part of a timed write transaction. The use is that currently ember-compatibility-functions do the timed write validation and I need the API to be drop-in-replace.

src/app/interaction-model/InvokeResponder.h
```diff
@@ -0,0 +1,316 @@
...
+    // Copying not allowed since underlying requirement is that on deletion of this
+    // object, a reply will be sent.
+    InvokeResponder(const InvokeResponder &)             = delete;
+    InvokeResponder & operator=(const InvokeResponder &) = delete;
+
+    /// Flush any pending replies before encoding the current reply.
+    ///
+    /// MAY be called at most once.
+    ///
+    /// This function is intended to provided the ability to retry sending a reply
```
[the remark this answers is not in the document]
andy31415: Based on above, will not fix InvokeResponder yet (keeping only one comment open as a tracking reminder)

src/app/interaction-model/InvokeResponder.h
```diff
@@ -0,0 +1,316 @@
...
+    ///
+    /// MUST be called exactly once to signal a response is to be recorded to be sent.
+    /// The error code (and the data encoded by ResponseEncoder) may be buffered for
+    /// sending among other batched responses.
+    ///
+    /// If this returns CHIP_ERROR_BUFFER_TOO_SMALL, this can be called a 2nd time after
+    /// a FlushPendingResponses.
+    ///
+    /// Argument behavior:
+    ///  - Commands can only be replied with ONE of the following (spec 8.9.4.4):
```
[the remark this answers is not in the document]
andy31415: Based on above, will not fix InvokeResponder yet (keeping only one comment open as a tracking reminder)

src/app/interaction-model/InvokeResponder.h
```diff
@@ -0,0 +1,316 @@
...
+    /// Reply with a data payload.
+    ///
+    /// MUST be called at most once per reply.
+    /// Can be called a 2nd time after a `FlushPendingResponses()` call
+    ///
+    ///   - responseCommandId must correspond with the data encoded in the returned encoder
+    ///   - Complete(CHIP_NO_ERROR) MUST be called to flush the reply
+    ///
+    /// If encoder returns CHIP_ERROR_BUFFER_TOO_SMALL, FlushPendingResponses should be
+    /// used to attempt to free up buffer space then encoding should be tried again.
```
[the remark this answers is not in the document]
andy31415: Based on above, will not fix InvokeResponder yet (keeping only one comment open as a tracking reminder)

src/app/interaction-model/Model.h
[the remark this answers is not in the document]
andy31415: I would like to do these updates when these functions get implemented and have a final validation pass on the API. It is not practical to try to fully design the API as even as it stands reviews took an extremely long time.

src/app/interaction-model/OperationTypes.h
andy31415: In particular: https://github.com/project-chip/connectedhomeip/blob/master/src/app/util/ember-compatibility-functions.cpp#L764

src/app/interaction-model/OperationTypes.h
andy31415: Could you be more precise here? I see documentation on `WriteAttribute` says:

```
- the last write wil have .... kListEnd
- the last list write ma have empy data (no list items)
```

which seems in-line with the documentation here. I updated it a bit in #33345 however if you noticed inconsistencies, please help me out by pointing out where those are located.

andy31415: Ended dropping list begin/end completely in 33345 since it seems during my write implementation I did not need to have it (since the list operation is part of the path itself)

src/app/interaction-model/OperationTypes.h
andy31415: Dropped the groupId as part of #33345

src/app/interaction-model/Model.h
```diff
@@ -0,0 +1,123 @@
...
+    /// List operation support:
+    ///    - the first list write will have `request.writeFlags.Has(WriteFlags::kListBegin)`
+    ///    - the last list write will have `request.writeFlags.Has(WriteFlags::kListEnd)`
+    ///    - the last list write MAY have empty data (no list items)
+    ///
+    /// When `request.writeFlags.Has(WriteFlags::kForceInternal)` the request is from an internal app update
+    /// and SHOULD bypass some internal checks (like timed enforcement, potentially read-only restrictions)
+    ///
+    /// Return codes
+    ///   CHIP_IM_GLOBAL_STATUS(code):
```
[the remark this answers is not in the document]
andy31415: It seems more convenient because all our macros are `VerifyOrReturnError` and not `VerifyOrReturnStatus`. The goal here should be to make the handler code easier to write and using CHIP_ERROR seems to make that easier.

src/app/interaction-model/OperationTypes.h
bzbarsky-apple: Right, but that's the IM (the part that can query attribute metadata) needing the information.  Cluster impls do not need to know it.

PR #32917, opened 2024-04-09
[Silabs] Moved nvm RebootCount and TotalOperationalHours keys to config

CuRahman:
**Description of Problem/Feature:** 

- TotalOperationalHours was being reset to 0 every power cycle, while the spec states to only reset on a factory reset.

**Description of Fix/Solution:**

- The update of TotalOperationalHours was not properly being handled within a shutdown function in the platform manager. I removed the update logic within the shutdown function and moved it within a function that is triggered hourly to increment and save it.

- The storage of both TotalOperationalHours and RebootCount was configured to not reset to 0 upon factory reset as well. I updated the memory region to be reset to 0 upon factory reset within src/platform/silabs/SilabsConfig.h.


examples/platform/silabs/SilabsDeviceAttestationCreds.cpp
```diff
@@ -195,14 +196,14 @@ void MigrateDacProvider(void)
...
 
-    MigrateUint32(kOldKey_Creds_KeyId, SilabsConfig::kConfigKey_Creds_KeyId);
-    MigrateUint32(kOldKey_Creds_Base_Addr, SilabsConfig::kConfigKey_Creds_Base_Addr);
-    MigrateUint32(kOldKey_Creds_DAC_Offset, SilabsConfig::kConfigKey_Creds_DAC_Offset);
-    MigrateUint32(kOldKey_Creds_DAC_Size, SilabsConfig::kConfigKey_Creds_DAC_Size);
-    MigrateUint32(kOldKey_Creds_PAI_Offset, SilabsConfig::kConfigKey_Creds_PAI_Offset);
-    MigrateUint32(kOldKey_Creds_PAI_Size, SilabsConfig::kConfigKey_Creds_PAI_Size);
-    MigrateUint32(kOldKey_Creds_CD_Offset, SilabsConfig::kConfigKey_Creds_CD_Offset);
-    MigrateUint32(kOldKey_Creds_CD_Size, SilabsConfig::kConfigKey_Creds_CD_Size);
+    MigrationManager::MigrateUint32(kOldKey_Creds_KeyId, SilabsConfig::kConfigKey_Creds_KeyId);
```
jmartinez-silabs, line 199: since you made a `MigrateUint32` in `MigrationManager`, delete the implementation of it in this file just above (line 176)

src/platform/silabs/efr32/PlatformManagerImpl.cpp
```diff
@@ -159,5 +160,16 @@ void PlatformManagerImpl::HandleWFXSystemEvent(wfx_event_base_t eventBase, sl_wf
 }
 #endif
 
+namespace Silabs {
+
+void MigrateCounterConfigs(void)
```
jmartinez-silabs, line 165: This should probably no be here and not duplicated for efr32 and 917

jmartinez-silabs (changes requested):

examples/platform/silabs/SilabsDeviceAttestationCreds.cpp
CuRahman: Removed.

src/platform/silabs/efr32/PlatformManagerImpl.cpp
CuRahman: Where would you reccomend this call be placed ? Not sure about a great spot for it..

CuRahman: I have moved this method to MigrationManager.cpp

jmartinez-silabs (approved):

andy31415 (approved):

PR #32919, opened 2024-04-09
Create test script to verify that the Linux tv-casting-app is able to discover the Linux tv-app.

### LINKED ISSUE #32918: Create test script to verify that the Linux tv-casting-app is able to discover the Linux tv-app.
As noted in #32787, there is currently no CI check in place to verify that the Linux tv-casting-app continues to work with the Linux tv-app whenever a PR is created and eventually merged in. This can result in the existing Matter casting experience breaking without knowing. 

This issue will cover the initial creation of the test script along with the implementation logic for testing the discovery between the Linux tv-casting-app and the Linux tv-app. The idea is to run the Linux tv-casting-app and the Linux tv-app as separate processes, write their respective output to separate files, iterate through the tv-casting-app logs to find strings of interest to determine if the discovery was successful or not. Lastly, the command to run the test script for testing the discovery between the tv-casting-app and the tv-app will also be added to the workflow yaml file.

In the case of a successful discovery, we will print the following to console:
- For each commissioner that we discover, we will print the discovered commissioner number, their corresponding Vendor ID, Product ID, and Device Type. Furthermore, each commissioner along with their information will be stored in a list.
- Print the number of commissioner(s) that was discovered.
- Print 'Discovery success!'

In the case of a failed discovery, we will print the following to console:
- If we find the string "No commissioner discovered" in the Linux tv-casting-app log file, then print this line along with 'Discovery failed!', dump the logs to console, and exit on error.
- In the rare instance where we don't find the aforementioned string "No commissioner discovered" and there is no entry for "Discovered Commissioner" (meaning the list of discovered commissioner(s) is empty), then something went wrong while running the Linux tv-casting-app and/or Linux tv-app. Print 'Discovery failed!', dump the logs to console, and exit on error.
Note: In either case of a failed discovery, the user should attempt to re-run the workflow and see if that allows for a successful discovery. If that does not pass, then the user should check their code to see how it is breaking the discovery process.

The next step is to test for the commissioning between the Linux tv-casting-app and the Linux tv-app. This will be created as a separate issue to be worked on.

shaoltan-amazon:
Fixes #32918.

**Problem**
There is no CI check in place to test that the discovery between the Linux tv-casting-app and the Linux tv-app continues to work whenever a PR is created.

**Solution Overview**
Create a test script to verify that the discovery between the Linux tv-casting-app and the Linux tv-app continue to work. The test script will be a part of the workflow yaml file that will run whenever a PR is created. 

The main idea is to run the Linux tv-casting-app and the Linux tv-app as separate processes and write their respective output to separate log files. The output of the tv-app process will be parsed in realtime for the string `Started commissioner` to indicate that the tv-app has started successfully. Next, the output of the tv-casting-app process will also be parsed in realtime for specific strings to indicate a successful or a failed discovery. If the strings of interest are found, then we print them to the console along with `Discovery success!`. Otherwise, we print the reason for the discovery failure along with `Discovery failed!`. Then, we will also dump the logs of both the tv-casting-app and the tv-app to the console and exit on error. 

In order for discovery to be considered as successful, the `Discovered Commissioner` in the tv-casting-app needs to have a `Vendor ID` value of `0xFFF1` (65521), `Product ID` value of `0x8001` (32769), and a `Device Type` value of `0x23` (35) as these values correspond to the Test TV App. Discovery fails when any of the entries of interest do not match the expected values or if "No commissioner discovered" string is found in the output of the tv-casting-app.

Note: The default relative paths to the tv-app and the tv-casting-app are `out/tv-app/chip-tv-app` and `out/tv-casting-app/chip-tv-casting-app` respectively. However, the user can pass their relative path via the command line by running:
`python3 run_tv_casting_test.py --tv-app-rel-path=path/to/tv-app --tv-casting-app-rel-path=path/to/tv-casting-app`

**Testing**
Tested the script locally by running `python3 ./scripts/tests/run_tv_casting_test.py` and observing the output. Also verified that the CI check works as expected by observing the workflow run on my local branch.

### DIFF
--- a/.github/workflows/examples-linux-tv-casting-app.yaml
+++ b/.github/workflows/examples-linux-tv-casting-app.yaml
@@ -17,7 +17,7 @@ name: Test TV Casting Example
 on:
     push:
         branches-ignore:
-            - 'dependabot/**'
+            - "dependabot/**"
     pull_request:
     merge_group:
 
@@ -63,10 +63,11 @@ jobs:
                   ./scripts/run_in_build_env.sh \
                     "scripts/examples/gn_build_example.sh examples/tv-casting-app/linux/ out/tv-casting-app"
 
-            - name: Test Discovery between Linux tv-casting-app and Linux tv-app
+            - name: Test casting from Linux tv-casting-app to Linux tv-app
               run: |
                   ./scripts/run_in_build_env.sh \
-                    "python3 ./scripts/tests/run_tv_casting_test.py test-discovery"
+                    "python3 ./scripts/tests/run_tv_casting_test.py test-casting"
+              timeout-minutes: 1
 
             - name: Uploading Size Reports
               uses: ./.github/actions/upload-size-reports
--- a/.github/workflows/examples-linux-tv-casting-app.yaml
+++ b/.github/workflows/examples-linux-tv-casting-app.yaml
@@ -63,6 +63,11 @@ jobs:
                   ./scripts/run_in_build_env.sh \
                     "scripts/examples/gn_build_example.sh examples/tv-casting-app/linux/ out/tv-casting-app"
 
+            - name: Test Discovery between Linux tv-casting-app and Linux tv-app
+              run: |
+                  ./scripts/run_in_build_env.sh \
+                    "python3 ./scripts/tests/run_tv_casting_test.py test-discovery"
+
             - name: Uploading Size Reports
               uses: ./.github/actions/upload-size-reports
               if: ${{ !env.ACT }}
--- a/.github/workflows/examples-linux-tv-casting-app.yaml
+++ b/.github/workflows/examples-linux-tv-casting-app.yaml
@@ -66,7 +66,7 @@ jobs:
             - name: Test casting from Linux tv-casting-app to Linux tv-app
               run: |
                   ./scripts/run_in_build_env.sh \
-                    "python3 ./scripts/tests/run_tv_casting_test.py test-casting"
+                    "python3 ./scripts/tests/run_tv_casting_test.py"
               timeout-minutes: 1
 
             - name: Uploading Size Reports
--- a/scripts/tests/run_tv_casting_test.py
+++ b/scripts/tests/run_tv_casting_test.py
@@ -1,0 +1,178 @@
+#!/usr/bin/env -S python3 -B
+
+# Copyright (c) 2024 Project CHIP Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import subprocess
+import sys
+import time
+
+import click
+
+
+# Dump the logs to the console in the case of an error.
+def dump_logs_to_console(log_file):
+    with open(log_file, 'r') as file:
+        logs = file.read()
+        print(logs)
+
+
+# Remove the log files once the script is done running.
+def remove_log_file(log_file):
+    if os.path.exists(log_file):
+        os.remove(log_file)
+    else:
+        print("The file does not exist.")
+
+
+# Read the logs from the Linux-tv-casting-app-logs.txt file.
+# The discovered commissioner(s) will be stored in a list along with their
+# vendor ID, product ID, and device type.
+def read_linux_tv_casting_app_logs(log_file):
+
+    with open(log_file, 'r') as file:
+        lines = file.readlines()
+
+    discovered_commissioners = []
+
+    print('Reading from Linux-tv-casting-app-logs.txt')
+
+    # Read through the Linux-tv-casting-app-logs.txt line by line
+    for i, line in enumerate(lines):
+
+        # If commissioner(s) are discovered, then the discovery process was successful.
+        if "commissioner(s) discovered" in line:
+            print(line)
+            print('Discovery success!')
+
+            remove_log_file('./scripts/tests/Linux-tv-casting-app-logs.txt')
+            remove_log_file('./scripts/tests/Linux-tv-app-logs.txt')
+
+            break
+
+        # If no commissioner was discovered, then something went wrong.
+        # Exit on error.
+        if "No commissioner discovered" in line:
+            print(line)
+            print('Discovery failed!')
+
+            dump_logs_to_console('./scripts/tests/Linux-tv-casting-app-logs.txt')
+
+            remove_log_file('./scripts/tests/Linux-tv-casting-app-logs.txt')
+            remove_log_file('./scripts/tests/Linux-tv-app-logs.txt')
+
+            sys.exit(1)
+
+        # Look for "Discovered Commissioner"
+        if "Discovered Commissioner" in line:
+            print(line)
+
+            # Extract the relevant part of the string
+            commissioner = line.split("Discovered Commissioner")[-1].strip()
+            commissioner = commissioner.replace('\x1b[0m', '')
+
+            # Initialize variables for Vendor ID, Product ID, and Device Type
+            vendor_id = None
+            product_id = None
+            device_type = None
+
+            # Iterate through the subsequent lines to find the strings of interest
+            for next_line in lines[i+1:]:
+
+                if "Vendor ID:" in next_line:
+                    print(next_line)
+
+                    vendor_id = next_line.split(":")[-1].strip()
+                    vendor_id = vendor_id.replace('\x1b[0m', '')
+
+                elif "Product ID:" in next_line:
+                    print(next_line)
+
+                    product_id = next_line.split(":")[-1].strip()
+                    product_id = product_id.replace('\x1b[0m', '')
+
+                elif "Device Type:" in next_line:
+                    print(next_line)
+
+                    device_type = next_line.split(":")[-1].strip()
+                    device_type = device_type.replace('\x1b[0m', '')
+
+                elif "commissioner(s) discovered" in next_line:
+                    break
+
+                # If the next line starts with "Discovered Commissioner", break the loop
+                if "Discovered Commissioner" in next_line:
+                    break
+
+            # Append the extracted information to the devices list
+            discovered_commissioners.append({
+                "discovered_commissioner": commissioner,
+                "vendor_id": vendor_id,
+                "product_id": product_id,
+                "device_type": device_type
+            })
+
+    # If the list of discovered commissioners is empty and we didn't find the "No commissioner discovered" string,
+    # then something went wrong. Exit on error.
+    if len(discovered_commissioners) == 0:
+        print('Discovery failed! No commissioner(s) discovered! The list of discovered commissioner(s) is empty!')
+
+        dump_logs_to_console('./scripts/tests/Linux-tv-casting-app-logs.txt')
+
+        remove_log_file('./scripts/tests/Linux-tv-casting-app-logs.txt')
+        remove_log_file('./scripts/tests/Linux-tv-app-logs.txt')
+
+        sys.exit(1)
+
+
+# Test if the Linux tv-casting-app is able to discover the Linux tv-app.
+# The Linux tv-casting-app and the tv-app will be run in separate processes.
+# Their corresponding output will be written to their respective log files.
+# The log file of the tv-casting-app will be parsed for strings of interest
+# which will be printed to the console.
+def test_discovery_fn():
+    with open('./scripts/tests/Linux-tv-app-logs.txt', 'w') as fd1, open('./scripts/tests/Linux-tv-casting-app-logs.txt', 'w') as fd2:
+
+        # Run the Linux tv-app and write the output to file
+        tv_app_rel_path = 'out/tv-app/chip-tv-app'
+        tv_app_abs_path = os.path.abspath(tv_app_rel_path)
+        p1 = subprocess.Popen(tv_app_abs_path, stdout=fd1, stderr=subprocess.PIPE, text=True)
+
+        time.sleep(5)
+
+        # Run the Linux tv-casting-app and write the output to file
+        tv_casting_app_rel_path = 'out/tv-casting-app/chip-tv-casting-app'
+        tv_casting_app_abs_path = os.path.abspath(tv_casting_app_rel_path)
+        p2 = subprocess.Popen(tv_casting_app_abs_path, stdout=fd2, stderr=subprocess.PIPE, text=True)
+
+    # Wait for the processes to finish writing before attempting to read
+    time.sleep(15)
+
+    read_linux_tv_casting_app_logs('./scripts/tests/Linux-tv-casting-app-logs.txt')
+
+
+@click.group()
+def main():
+    pass
+
+
+@main.command('test-discovery', help='Test if the Linux tv-casting-app is able to discover the Linux tv-app.')
+def test_discovery():
+    test_discovery_fn()
+
+
+if __name__ == '__main__':
+
+    main()
--- a/scripts/tests/run_tv_casting_test.py
+++ b/scripts/tests/run_tv_casting_test.py
@@ -21,6 +21,16 @@ import time
 
 import click
 
+LINUX_TV_APP_LOGS = './scripts/tests/Linux-tv-app-logs.txt'
+LINUX_TV_CASTING_APP_LOGS = './scripts/tests/Linux-tv-casting-app-logs.txt'
+
+RUN_INTERVAL = 5
+PARSE_INTERVAL = 15
+
+VENDOR_ID = 65521
+PRODUCT_ID = 32769
+DEVICE_TYPE = 35
+
 
 # Dump the logs to the console in the case of an error.
 def dump_logs_to_console(log_file):
@@ -37,10 +47,46 @@ def remove_log_file(log_file):
         print("The file does not exist.")
 
 
+# Whenever a failure is discovered, we should print 'Discovery failed!',
+# dump the logs, clean up the log files, exit on error.
+def handle_discovery_failure():
+    print('Discovery failed!')
+
+    dump_logs_to_console(LINUX_TV_CASTING_APP_LOGS)
+
+    remove_log_file(LINUX_TV_CASTING_APP_LOGS)
+    remove_log_file(LINUX_TV_APP_LOGS)
+
+    sys.exit(1)
+
+
+# Helper function to extract the integer value from a string.
+def extract_value_from_string(line):
+    value = line.split(":")[-1].strip().replace('\x1b[0m', '')
+    value = int(value)
+
+    return value
+
+
+# Check if the discovered value matches the expected value.
+def check_expected_value(line, expected_value, value_name):
+    # Extract the integer value from the string
+    value = extract_value_from_string(line)
+
+    # If the discovered value does not match the expected value,
+    # print the error and handle the discovery failure.
+    if value != expected_value:
+        print(f'{value_name} does not match the expected value!')
+        print(f'Discovered {value_name}: {value}')
+        print(f'Expected {value_name}: {expected_value}')
+
+        handle_discovery_failure()
+
+
 # Read the logs from the Linux-tv-casting-app-logs.txt file.
 # The discovered commissioner(s) will be stored in a list along with their
 # vendor ID, product ID, and device type.
-def read_linux_tv_casting_app_logs(log_file):
+def parse_linux_tv_casting_app_logs(log_file):
 
     with open(log_file, 'r') as file:
         lines = file.readlines()
@@ -57,23 +103,10 @@ def read_linux_tv_casting_app_logs(log_file):
             print(line)
             print('Discovery success!')
 
-            remove_log_file('./scripts/tests/Linux-tv-casting-app-logs.txt')
-            remove_log_file('./scripts/tests/Linux-tv-app-logs.txt')
+            remove_log_file(LINUX_TV_CASTING_APP_LOGS)
+            remove_log_file(LINUX_TV_APP_LOGS)
 
             break
-
-        # If no commissioner was discovered, then something went wrong.
-        # Exit on error.
-        if "No commissioner discovered" in line:
-            print(line)
-            print('Discovery failed!')
-
-            dump_logs_to_console('./scripts/tests/Linux-tv-casting-app-logs.txt')
-
-            remove_log_file('./scripts/tests/Linux-tv-casting-app-logs.txt')
-            remove_log_file('./scripts/tests/Linux-tv-app-logs.txt')
-
-            sys.exit(1)
 
         # Look for "Discovered Commissioner"
         if "Discovered Commissioner" in line:
@@ -93,21 +126,15 @@ def read_linux_tv_casting_app_logs(log_file):
 
                 if "Vendor ID:" in next_line:
                     print(next_line)
-
-                    vendor_id = next_line.split(":")[-1].strip()
-                    vendor_id = vendor_id.replace('\x1b[0m', '')
+                    vendor_id = extract_value_from_string(next_line)
 
                 elif "Product ID:" in next_line:
                     print(next_line)
-
-                    product_id = next_line.split(":")[-1].strip()
-                    product_id = product_id.replace('\x1b[0m', '')
+                    product_id = extract_value_from_string(next_line)
 
                 elif "Device Type:" in next_line:
                     print(next_line)
-
-                    device_type = next_line.split(":")[-1].strip()
-                    device_type = device_type.replace('\x1b[0m', '')
+                    device_type = extract_value_from_string(next_line)
 
                 elif "commissioner(s) discovered" in next_line:
                     break
@@ -127,14 +154,8 @@ def read_linux_tv_casting_app_logs(log_file):
     # If the list of discovered commissioners is empty and we didn't find the "No commissioner discovered" string,
     # then something went wrong. Exit on error.
     if len(discovered_commissioners) == 0:
-        print('Discovery failed! No commissioner(s) discovered! The list of discovered commissioner(s) is empty!')
-
-        dump_logs_to_console('./scripts/tests/Linux-tv-casting-app-logs.txt')
-
-        remove_log_file('./scripts/tests/Linux-tv-casting-app-logs.txt')
-        remove_log_file('./scripts/tests/Linux-tv-app-logs.txt')
-
-        sys.exit(1)
+        print('No commissioner(s) discovered! The list of discovered commissioner(s) is empty!')
+        handle_discovery_failure()
 
 
 # Test if the Linux tv-casting-app is able to discover the Linux tv-app.
@@ -143,24 +164,48 @@ def read_linux_tv_casting_app_logs(log_file):
 # The log file of the tv-casting-app will be parsed for strings of interest
 # which will be printed to the console.
 def test_discovery_fn():
-    with open('./scripts/tests/Linux-tv-app-logs.txt', 'w') as fd1, open('./scripts/tests/Linux-tv-casting-app-logs.txt', 'w') as fd2:
+
+    with open(LINUX_TV_APP_LOGS, 'w') as fd1, open(LINUX_TV_CASTING_APP_LOGS, 'w') as fd2:
 
         # Run the Linux tv-app and write the output to file
         tv_app_rel_path = 'out/tv-app/chip-tv-app'
         tv_app_abs_path = os.path.abspath(tv_app_rel_path)
         subprocess.Popen(tv_app_abs_path, stdout=fd1, stderr=subprocess.PIPE, text=True)
 
-        time.sleep(5)
+        time.sleep(RUN_INTERVAL)
 
         # Run the Linux tv-casting-app and write the output to file
         tv_casting_app_rel_path = 'out/tv-casting-app/chip-tv-casting-app'
         tv_casting_app_abs_path = os.path.abspath(tv_casting_app_rel_path)
-        subprocess.Popen(tv_casting_app_abs_path, stdout=fd2, stderr=subprocess.PIPE, text=True)
+        tv_casting_app_process = subprocess.Popen(
+            tv_casting_app_abs_path, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
+
+        for line in tv_casting_app_process.stdout:
+            # Write the line to the Linux tv-casting-app log file
+            fd2.write(line)
+
+            # Fail fast if "No commissioner discovered" string found
+            if "No commissioner discovered" in line:
+                print(line)
+                handle_discovery_failure()
+
+            # Check if the Vendor ID, Product ID, and Device Type match the expected values
+            if "Vendor ID:" in line:
+                check_expected_value(line, VENDOR_ID, "Vendor ID")
+
+            elif "Product ID:" in line:
+                check_expected_value(line, PRODUCT_ID, "Product ID")
+
+            elif "Device Type:" in line:
+                check_expected_value(line, DEVICE_TYPE, "Device Type")
+
+            if "commissioner(s) discovered" in line:
+                break
 
     # Wait for the processes to finish writing before attempting to read
-    time.sleep(15)
+    time.sleep(PARSE_INTERVAL)
 
-    read_linux_tv_casting_app_logs('./scripts/tests/Linux-tv-casting-app-logs.txt')
+    parse_linux_tv_casting_app_logs(LINUX_TV_CASTING_APP_LOGS)
 
 
 @click.group()
@@ -174,5 +219,7 @@ def test_discovery():
 
 
 if __name__ == '__main__':
+    # Start with a clean slate by removing any previously cached entries.
+    os.system('rm -f /tmp/chip_*')
 
     main()
--- a/scripts/tests/run_tv_casting_test.py
+++ b/scripts/tests/run_tv_casting_test.py
@@ -14,12 +14,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
 import os
 import subprocess
 import sys
 import time
 
 import click
+
+# Configure logging format.
+logging.basicConfig(level=logging.DEBUG, format='%(levelname)s - %(message)s')
 
 LINUX_TV_APP_LOGS = './scripts/tests/Linux-tv-app-logs.txt'
 LINUX_TV_CASTING_APP_LOGS = './scripts/tests/Linux-tv-casting-app-logs.txt'
@@ -33,28 +37,31 @@ DEVICE_TYPE = 35
 
 # Dump the logs to the console in the case of an error.
 def dump_logs_to_console(log_file):
+
     if log_file == LINUX_TV_CASTING_APP_LOGS:
-        print('Dumping Linux TV Casting App Logs to Console.')
+        logging.debug('Dumping Linux TV Casting App Logs to Console.')
     elif log_file == LINUX_TV_APP_LOGS:
-        print('Dumping Linux TV App Logs to Console.')
+        logging.debug('Dumping Linux TV App Logs to Console.')
 
     with open(log_file, 'r') as file:
         logs = file.read()
-        print(logs)
+        logging.debug(logs)
 
 
 # Remove the log files once the script is done running.
 def remove_log_file(log_file):
+
     if os.path.exists(log_file):
         os.remove(log_file)
     else:
-        print("The file does not exist.")
+        logging.error('The file does not exist.')
 
 
 # Whenever a failure is discovered, we should print 'Discovery failed!',
 # dump the logs, clean up the log files, exit on error.
 def handle_discovery_failure():
-    print('Discovery failed!\n')
+
+    logging.error('Discovery failed!\n')
 
     dump_logs_to_console(LINUX_TV_CASTING_APP_LOGS)
     dump_logs_to_console(LINUX_TV_APP_LOGS)
@@ -67,7 +74,8 @@ def handle_discovery_failure():
 
 # Helper function to extract the integer value from a string.
 def extract_value_from_string(line):
-    value = line.split(":")[-1].strip().replace('\x1b[0m', '')
+
+    value = line.split(':')[-1].strip().replace('\x1b[0m', '')
     value = int(value)
 
     return value
@@ -76,21 +84,30 @@ def extract_value_from_string(line):
 # Check if the discovered value matches the expected value.
 # Returns False if the value does not match, True otherwise.
 def validate_value(expected_value, line, value_name):
+
     # Extract the integer value from the string
     value = extract_value_from_string(line)
 
     # If the discovered value does not match the expected value,
     # print the error and return False.
     if value != expected_value:
-        print(f'{value_name} does not match the expected value!')
-        print(f'Expected {value_name}: {expected_value}')
+        logging.error(f'{value_name} does not match the expected value!')
+        logging.error(f'Expected {value_name}: {expected_value}')
         line = line.rstrip('\n')
-        print(line)
+        logging.error(line)
 
         return False
 
     # Return True if the value matches the expected value
     return True
+
+
+# Tear down the processes once done running.
+def tear_down(processes):
+
+    for process in processes:
+        process.terminate()
+        process.wait()
 
 
 # Test if the Linux tv-casting-app is able to discover the Linux tv-app.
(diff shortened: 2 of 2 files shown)


scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,178 @@
...
+
+    print('Reading from Linux-tv-casting-app-logs.txt')
+
+    # Read through the Linux-tv-casting-app-logs.txt line by line
+    for i, line in enumerate(lines):
+
+        # If commissioner(s) are discovered, then the discovery process was successful.
+        if "commissioner(s) discovered" in line:
+            print(line)
+            print('Discovery success!')
```
sharadb-amazon, line 58: It looks like we are marking the discovery as a "success" if "any" commissioner was discovered. Can we make the success criterion a bit more narrow so that we are validating that the commissioner we discovered matched the expected vendor\_id, product\_id and device\_type?

sharadb-amazon (changes requested):

sharadb-amazon (commented):
Is it possible to mark this test as "Required"?

shaoltan-amazon:
@sharadb-amazon I believe we can mark the test as "required" via the Settings tab of the repository but I don't have access to the Settings tab on my end.

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,178 @@
...
+
+# Test if the Linux tv-casting-app is able to discover the Linux tv-app.
+# The Linux tv-casting-app and the tv-app will be run in separate processes.
+# Their corresponding output will be written to their respective log files.
+# The log file of the tv-casting-app will be parsed for strings of interest
+# which will be printed to the console.
+def test_discovery_fn():
+    with open('./scripts/tests/Linux-tv-app-logs.txt', 'w') as fd1, open('./scripts/tests/Linux-tv-casting-app-logs.txt', 'w') as fd2:
+
+        # Run the Linux tv-app and write the output to file
```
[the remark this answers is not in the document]
shaoltan-amazon: Sure, I can add that.

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,178 @@
...
+
+    # If the list of discovered commissioners is empty and we didn't find the "No commissioner discovered" string,
+    # then something went wrong. Exit on error.
+    if len(discovered_commissioners) == 0:
+        print('Discovery failed! No commissioner(s) discovered! The list of discovered commissioner(s) is empty!')
+
+        dump_logs_to_console('./scripts/tests/Linux-tv-casting-app-logs.txt')
+
+        remove_log_file('./scripts/tests/Linux-tv-casting-app-logs.txt')
+        remove_log_file('./scripts/tests/Linux-tv-app-logs.txt')
```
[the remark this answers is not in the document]
shaoltan-amazon: Sounds good to me. I'll update to have the paths as constants.

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,178 @@
...
+# which will be printed to the console.
+def test_discovery_fn():
+    with open('./scripts/tests/Linux-tv-app-logs.txt', 'w') as fd1, open('./scripts/tests/Linux-tv-casting-app-logs.txt', 'w') as fd2:
+
+        # Run the Linux tv-app and write the output to file
+        tv_app_rel_path = 'out/tv-app/chip-tv-app'
+        tv_app_abs_path = os.path.abspath(tv_app_rel_path)
+        p1 = subprocess.Popen(tv_app_abs_path, stdout=fd1, stderr=subprocess.PIPE, text=True)
+
+        time.sleep(5)
```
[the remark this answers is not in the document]
shaoltan-amazon: Will update to make the wait times constants.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: So have a check to verify that the vendor_id, product_id, and device_type found in the Linux tv-casting-app logs matches with the vendor_id, product_id, and device_type entries from the Linux tv-app logs?

sharadb-amazon: You can check the values found in the tv-casting-app logs with corresponding constants declared in the script.

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,178 @@
...
+    if os.path.exists(log_file):
+        os.remove(log_file)
+    else:
+        print("The file does not exist.")
+
+
+# Read the logs from the Linux-tv-casting-app-logs.txt file.
+# The discovered commissioner(s) will be stored in a list along with their
+# vendor ID, product ID, and device type.
+def read_linux_tv_casting_app_logs(log_file):
```
sharadb-amazon, line 43: Since you are _searching_ through the logs here, maybe call this function "parse\_linux\_tv\_casting\_app\_logs"

.github/workflows/examples-linux-tv-casting-app.yaml
```diff
@@ -61,6 +61,11 @@ jobs:
                   ./scripts/run_in_build_env.sh \
                     "scripts/examples/gn_build_example.sh examples/tv-casting-app/linux/ out/tv-casting-app"
 
+            - name: Test Discovery between Linux tv-casting-app and Linux tv-app
+              run: |
+                  ./scripts/run_in_build_env.sh \
+                    "python3 ./scripts/tests/run_tv_casting_test.py test-discovery"
```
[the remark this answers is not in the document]
shaoltan-amazon: Yes, that was the initial idea that I was thinking of implementing. Do you have another approach in mind that you would prefer instead?

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Okay, I will update it.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Just to confirm, the constants would be the following?
- Vendor ID: 65521
- Product ID: 32769
- Device Type: 35

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,178 @@
...
+
+        time.sleep(5)
+
+        # Run the Linux tv-casting-app and write the output to file
+        tv_casting_app_rel_path = 'out/tv-casting-app/chip-tv-casting-app'
+        tv_casting_app_abs_path = os.path.abspath(tv_casting_app_rel_path)
+        subprocess.Popen(tv_casting_app_abs_path, stdout=fd2, stderr=subprocess.PIPE, text=True)
+
+    # Wait for the processes to finish writing before attempting to read
+    time.sleep(15)
```
sharadb-amazon, line 161: Is it possible to multitask i.e. parse the output as it is generated and fail-fast in cases there is a failure?

scripts/tests/run_tv_casting_test.py
sharadb-amazon: Yes

.github/workflows/examples-linux-tv-casting-app.yaml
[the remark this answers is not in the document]
sharadb-amazon: If we have them as separate tests, if I understanding this correctly, the "connection" test will re-run the "discovery" test. And the "command / attribute subscription" test will re-run "discovery" and "connection"? I don't see much value addition is separating them out, and would recommend adding one "step" for one test workflow. This is assuming we can dump out the reason for the failure in the CI test logs, like "Discovery failed," "Connection failed," "LaunchURL failed," etc. 

In the future we could add additional steps when we test other workflows (like the one where the casting app is reconnecting to a TV after a prior connection attempt)

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Sure, I can update to that approach.

.github/workflows/examples-linux-tv-casting-app.yaml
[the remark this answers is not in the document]
shaoltan-amazon: Sounds good to me. I will keep this in mind when I implement the remaining steps that test for connection and commands/attribute subscriptions later on.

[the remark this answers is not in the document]
sharadb-amazon: Can we update the name of this step accordingly to "Test casting fromLinux tv-casting-app to Linux tv-app"?

scripts/tests/run_tv_casting_test.py
sharadb-amazon: Does Line 104 stills seems to log "Discovery success" if any commissioner was discovered? Can we only log success for discovery if the discovered commissioner matched the above constants?

scripts/tests/run_tv_casting_test.py
sharadb-amazon: Are we still waiting for time.sleep(PARSE\_INTERVAL) i.e. 15 seconds before parsing the logs?

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Once we reach line 104, only commissioner(s) with valid `vendor_id`, `product_id`, and `device_type` will allow for "Discovery success!". The `check_expected_value` function handles when `vendor_id`, `product_id`, or `device_type` does not match the expected constants. This check is used in lines 192-200 before the Linux tv-casting-app logs are parsed. In other words, the logs should only contain commissioner(s) with valid `vendor_id`, `product_id`, and `device_type`. So "invalid" commissioner(s) would be caught early on and throw an error.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Yes.

.github/workflows/examples-linux-tv-casting-app.yaml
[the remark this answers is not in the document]
sharadb-amazon: Yes

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,225 @@
...
+
+        time.sleep(RUN_INTERVAL)
+
+        # Run the Linux tv-casting-app and write the output to file
+        tv_casting_app_rel_path = 'out/tv-casting-app/chip-tv-casting-app'
+        tv_casting_app_abs_path = os.path.abspath(tv_casting_app_rel_path)
+        tv_casting_app_process = subprocess.Popen(
+            tv_casting_app_abs_path, stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
+
+        for line in tv_casting_app_process.stdout:
```
sharadb-amazon, line 183: Just like we are failing fast, we could succeed fast as well here by looking for the commissioner with the expected vid, pid, dt and succeeding accordingly.

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,225 @@
...
+            elif "Device Type:" in line:
+                check_expected_value(line, DEVICE_TYPE, "Device Type")
+
+            if "commissioner(s) discovered" in line:
+                break
+
+    # Wait for the processes to finish writing before attempting to read
+    time.sleep(PARSE_INTERVAL)
+
+    parse_linux_tv_casting_app_logs(LINUX_TV_CASTING_APP_LOGS)
```
sharadb-amazon, line 208: See comment above about succeeding fast. We probably will not need this timeout or separate file handling logic at all.

sharadb-amazon (commented):
We should add a timeout for the overall step (60 seconds should be plenty)

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,225 @@
...
+# The log file of the tv-casting-app will be parsed for strings of interest
+# which will be printed to the console.
+def test_discovery_fn():
+
+    with open(LINUX_TV_APP_LOGS, 'w') as fd1, open(LINUX_TV_CASTING_APP_LOGS, 'w') as fd2:
+
+        # Run the Linux tv-app and write the output to file
+        tv_app_rel_path = 'out/tv-app/chip-tv-app'
+        tv_app_abs_path = os.path.abspath(tv_app_rel_path)
+        subprocess.Popen(tv_app_abs_path, stdout=fd1, stderr=subprocess.PIPE, text=True)
```
sharadb-amazon, line 173: I think we probably also need a tear-down stage where we kill the tv-app and the tv-casting-app after the test completes/times out.

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,228 @@
...
+                    remove_log_file(LINUX_TV_APP_LOGS)
+
+                    break
+
+        # Tear down the processes.
+        tv_app_process.terminate()
+        tv_app_process.wait()
+
+        tv_casting_app_process.terminate()
+        tv_casting_app_process.wait()
```
sharadb-amazon, line 211: Can you make this a `def tear_down`, that you just call from there?

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,225 @@
...
+# Helper function to extract the integer value from a string.
+def extract_value_from_string(line):
+    value = line.split(":")[-1].strip().replace('\x1b[0m', '')
+    value = int(value)
+
+    return value
+
+
+# Check if the discovered value matches the expected value.
+def check_expected_value(line, expected_value, value_name):
```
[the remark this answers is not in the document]
shaoltan-amazon: Agreed. The def has been updated per the recommendation mentioned.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Agreed, implementation has been updated to succeed fast after finding commissioner with expected vendor id, product id, and device type.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Agreed, by succeeding fast, we do not need the timeout and/or separate parsing logic. Parsing will be done in real-time as we get the output from the tv-casting-app process. The change has been implemented.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Agreed, a tear-down stage has now been added in the latest commit.

shaoltan-amazon:
@sharadb-amazon A timeout of 1 minute for the step has been set in the workflow yaml file in the latest commit.

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,228 @@
...
+    if os.path.exists(log_file):
+        os.remove(log_file)
+    else:
+        print("The file does not exist.")
+
+
+# Whenever a failure is discovered, we should print 'Discovery failed!',
+# dump the logs, clean up the log files, exit on error.
+def handle_discovery_failure():
+    print('Discovery failed!\n')
```
sharadb-amazon, line 57: Refer to [https://docs.python.org/3/howto/logging.html](https://docs.python.org/3/howto/logging.html)

`import logging` and then use `logging.error()` instead of `print()` here. Use logging in the rest of the file (with the appropriate log level)

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: `time.sleep(PARSE_INTERVAL)` has now been removed as we will be parsing the output as we receive them in real-time.

sharadb-amazon (approved):

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Okay, I will make the change.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Added `def tear_down` in the latest commit.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: The latest commit now uses the logging module per your suggestion.

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,245 @@
...
+        # Tear down the processes.
+        tear_down([tv_app_process, tv_casting_app_process])
+
+
+@click.group()
+def main():
+    pass
+
+
+@main.command('test-casting', help='Test casting from Linux tv-casting-app to Linux tv-app.')
```
andy31415, line 236: why do we have the separate group and command?
How about just have main that runs one test as today (and does not need extra arguments) and if we ever get more tests in the same file we can update.

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,245 @@
...
+import subprocess
+import sys
+import time
+
+import click
+
+# Configure logging format.
+logging.basicConfig(level=logging.DEBUG, format='%(levelname)s - %(message)s')
+
+LINUX_TV_APP_LOGS = './scripts/tests/Linux-tv-app-logs.txt'
```
andy31415, line 28: This seems to access paths inside scripts ... those are source files. you should not dirty the source tree, even teporarely. Can we use temporary files instead?

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,245 @@
...
+LINUX_TV_CASTING_APP_LOGS = './scripts/tests/Linux-tv-casting-app-logs.txt'
+
+RUN_INTERVAL = 2
+
+VENDOR_ID = 65521
+PRODUCT_ID = 32769
+DEVICE_TYPE = 35
```
andy31415, line 35: please add comments and explanation what 35 means and where the constant comes from.

VendorID in hex format may also be easier to read together with an exaplanation that this is a test vendor id.

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+# Dump the logs to the console in the case of an error.
```
andy31415, line 38: where is the "in case of an error" check?

if this function does not do the check and it is done somewhere else, the function description should not contain this detail.

This does not log to console ... instead it dumps the specified path as a `debug` entry to python standard logging.

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,245 @@
...
+    elif log_file == LINUX_TV_APP_LOGS:
+        logging.debug('Dumping Linux TV App Logs to Console.')
+
+    with open(log_file, 'r') as file:
+        logs = file.read()
+        logging.debug(logs)
+
+
+# Remove the log files once the script is done running.
+def remove_log_file(log_file):
```
andy31415, line 52: Use temporary files or directories for this.

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,245 @@
...
+    remove_log_file(LINUX_TV_CASTING_APP_LOGS)
+    remove_log_file(LINUX_TV_APP_LOGS)
+
+    sys.exit(1)
+
+
+# Helper function to extract the integer value from a string.
+def extract_value_from_string(line):
+
+    value = line.split(':')[-1].strip().replace('\x1b[0m', '')
```
andy31415, line 78: Give examples in comments of expected inputs and outputs.
I am unclear what `line` may contain ... it seems to be `<anything>:<maybe-some-color-code>number` ... however this is hard to understand here. Why do we even have the color encodings here?

This needs examples and explanations

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,245 @@
...
+    value = int(value)
+
+    return value
+
+
+# Check if the discovered value matches the expected value.
+# Returns False if the value does not match, True otherwise.
+def validate_value(expected_value, line, value_name):
```
andy31415, line 86: add types here to make things more clear. It seems expected_value is int, line is str, value_name is str and the function returns bool

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,245 @@
...
+
+import click
+
+# Configure logging format.
+logging.basicConfig(level=logging.DEBUG, format='%(levelname)s - %(message)s')
+
+LINUX_TV_APP_LOGS = './scripts/tests/Linux-tv-app-logs.txt'
+LINUX_TV_CASTING_APP_LOGS = './scripts/tests/Linux-tv-casting-app-logs.txt'
+
+RUN_INTERVAL = 2
```
andy31415, line 31: add unit of measurement to things ... this is `RUN_INTERVAL_SEC` ... but at the same time `run what?`.

It seems to be `TV_APP_STARTUP_TIME_SEC` or something like that. Please rename as appropriate to make the code easy to understand.

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,245 @@
...
+        logging.error(line)
+
+        return False
+
+    # Return True if the value matches the expected value.
+    return True
+
+
+# Tear down the processes once done running.
+def tear_down(processes):
```
andy31415, line 106: could we make these a context manager instead, so that exit is guaranteed to be called? a `with ...` seems clearer than trying to figure out that `tear_down` is called after a large series of nested items.

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,245 @@
...
+# The Linux tv-casting-app and the tv-app will be run in separate processes.
+# Their corresponding output will be written to their respective log files.
+# The output of the tv-casting-app will be parsed in realtime for strings of
+# interest which will be printed to the console.
+def test_discovery_fn():
+
+    with open(LINUX_TV_APP_LOGS, 'w') as fd1, open(LINUX_TV_CASTING_APP_LOGS, 'w') as fd2:
+
+        # Run the Linux tv-app and write the output to file.
+        tv_app_rel_path = 'out/tv-app/chip-tv-app'
```
andy31415, line 123: This seems to be highly specific to the unit test environment. When I compile things, I use build_examples and that generates `out/linux-x64-...` prefixes.

please make these application paths as arguments to the test program. you can have defaults, however at least allow others to run the test regardless of how they compile.

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,245 @@
...
+        valid_discovered_commissioner_str = ''
+
+        # Read the output as we get it from the tv-casting-app process.
+        for line in tv_casting_app_process.stdout:
+
+            # Write the line to the Linux tv-casting-app log file.
+            fd2.write(line)
+
+            # Fail fast if "No commissioner discovered" string found.
+            if "No commissioner discovered" in line:
```
andy31415, line 146: This seems to do quite a bit of parsing and "redirect to file" logic.

We already do this for integration tests in https://github.com/project-chip/connectedhomeip/blob/master/scripts/tests/chiptest/runner.py#L121 which also try to not allow buffering for the output streams (which this PR does not do).

I wonder if we could maybe re-use or make some things common.

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,245 @@
...
+def test_discovery_fn():
+
+    with open(LINUX_TV_APP_LOGS, 'w') as fd1, open(LINUX_TV_CASTING_APP_LOGS, 'w') as fd2:
+
+        # Run the Linux tv-app and write the output to file.
+        tv_app_rel_path = 'out/tv-app/chip-tv-app'
+        tv_app_abs_path = os.path.abspath(tv_app_rel_path)
+        tv_app_process = subprocess.Popen(tv_app_abs_path, stdout=fd1, stderr=subprocess.PIPE, text=True)
+
+        time.sleep(RUN_INTERVAL)
```
andy31415, line 127: Could we instead have some "wait max X seconds until condition is met ..." instead?

We seem to sleep and then assume everything is well. Generally Sleep() in tests is not a good idea. We should poll the output in this case it seems and check if it contains some signal of "program started successfully" and time out if we do not see the expected output.

A "sleep the maximum amount of program may take to start" will always be pessimistic and will make the tests as slow as possible.

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,245 @@
...
+            if "Discovered Commissioner" in line:
+
+                line = line.rstrip('\n')
+                valid_discovered_commissioner_str = line
+
+                # Continue parsing the content that belongs to the "Discovered Commissioner".
+                continue_parsing = True
+
+                # Initialize variables to store the information of interest.
+                valid_vendor_id = False
```
andy31415, line 162: Make the values as `Option[str]` and then None means "no value yet" and a valid string means a value. then you do not need both bool and str.

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,245 @@
...
+        os.remove(log_file)
+    else:
+        logging.error('The file does not exist.')
+
+
+# Whenever a failure is discovered, we should print 'Discovery failed!',
+# dump the logs, clean up the log files, exit on error.
+def handle_discovery_failure():
+
+    logging.error('Discovery failed!\n')
```
andy31415, line 64: this does not call teardown, so processes are not cleaned up.
Generally handling of cleanup via functions is error prone. Use context managers or try/finally (prefer context managers)

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Good point, I made the change in the latest commit.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Sure, I revised the code to use a temporary directory to store the log files.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Got it. I have revised this in the latest commit.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: As noted in a previous reply, I have made the change to use temporary directory in the latest commit.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: This is an example output from the Linux tv-casting-app which will be the provided string input (`line`):
`\x1b[0;34m[1713741926895] [7276:9521344] [DIS] Vendor ID: 65521\x1b[0m`

The value to be extracted from the string is `65521`.

I have added the relevant comments in the latest commit to address this.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: I have made the changes in the latest commit.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Noted. I renamed the constant in the latest commit.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Sure, that's a good suggestion! I made the change in the latest commit.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Good catch! The change has been made in the latest commit.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Noted! I changed to this approach in the latest commit.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: This is addressed in the latest commit.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Noted. Changed to use context managers to handle cleanup of the processes (and files) as suggested.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: For now, I added the following to the test script for dealing with stdout buffering:
```
# Configure command options to disable stdout buffering during tests
disable_stdout_buffering_cmd = []
# On Unix-like systems, use stdbuf to disable stdout buffering
if sys.platform == 'darwin' or sys.platform == 'linux':
     disable_stdout_buffering_cmd = ['stdbuf', '-o0', '-i0']
```

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,305 @@
...
+# The maximum amount of time to wait for the Linux tv-app to start before timeout.
+TV_APP_MAX_START_WAIT_SEC = 2
+
+# File names of logs for the Linux tv-casting-app and the Linux tv-app.
+LINUX_TV_APP_LOGS = 'Linux-tv-app-logs.txt'
+LINUX_TV_CASTING_APP_LOGS = 'Linux-tv-casting-app-logs.txt'
+
+# Values that identify the Linux tv-app and are noted in the 'Device Configuration' in the Linux tv-app output
+# as well as under the 'Discovered Commissioner' details in the Linux tv-casting-app output.
+VENDOR_ID = 65521   # Test vendor id
```
andy31415, line 39:
```suggestion
VENDOR_ID = 0xFFF1   # Spec 7.20.2.1 MEI code: test vendor IDs are 0xFFF1 to 0xFFF4
```

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,305 @@
...
+PRODUCT_ID = 32769  # Test product id
```
andy31415, line 40:
```suggestion
PRODUCT_ID = 0x7FF9  # Test product id
```

Why was this constant chosen? does it have any significance? It seems somewhat non-random as a rather large value close (but not quite) to the max_signed_16_bit

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,305 @@
...
+DEVICE_TYPE = 35    # Casting video player
```
andy31415, line 41: Using this naming, whenever we read `Device Type` I have to refer back to this line and read the comment. How about renaming and changing everywhere:

```suggestion
DEVICE_TYPE_CASTING_VIDEO_PLAYER = 35  # Device type library 10.3: Casting Video Player
```

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,305 @@
...
+    def __init__(self, log_file_path: str, mode: str = 'w+'):
+        self.log_file_path = log_file_path
+        self.mode = mode
+
+    # Enter the context manager to open and return the log file.
+    def __enter__(self):
+        try:
+            self.file = open(self.log_file_path, self.mode)
+        except FileNotFoundError:
+            # Handle file not found error
```
andy31415, line 65: These comments can be removed as they do not help in understanding the code better: it is obvious that the block within `except FileNotFoundError` contains the handling of a file not found error. Please apply this logic throughout: comments should add/help in understanding code, not re-stating it.

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,305 @@
...
+            raise FileNotFoundError(f"Log file '{self.log_file_path}' not found.")
+        except IOError:
+            # Handle IO error
+            raise IOError(f"Error opening log file '{self.log_file_path}'.")
```
andy31415, line 69: Why do we re-raise basically the same error but with a different text? When I tried it out I see python already contains the path:

```
>>> open("/invalid/path/x.txt", "rb")
  File "<stdin>", line 1, in <module>
FileNotFoundError: [Errno 2] No such file or directory: '/invalid/path/x.txt'
```

Catching an error just to re-throw the same error seems not as useful. The difference seems to be that this will say "log file" in the error text, however that will be seen in the stack trace if we fail anyway. Code would be shorter by not having these excepts.

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,305 @@
...
+    # Exit the context manager, terminating the subprocess.
+    #   exception_type: The type of exception that occurred, if any.
+    #   exception_value: The value of the exception, if any.
+    #   traceback: The traceback of the exception.
+    def __exit__(self, exception_type, exception_value, traceback):
+        self.process.terminate()
+        self.process.wait()
+
+
+# Dump the contents of a log file to the console.
```
andy31415, line 113: nit: it is generally more pythonic to put doc-style comments as strings after the function declaration.

https://peps.python.org/pep-0257/#one-line-docstrings
https://peps.python.org/pep-0257/#multi-line-docstrings

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,305 @@
...
+        return self.file
+
+    # Exit the context manager, closing and removing the log file created.
+    #   exception_type: The type of exception that occurred, if any.
+    #   exception_value: The value of the exception, if any.
+    #   traceback: The traceback of the exception.
+    def __exit__(self, exception_type, exception_value, traceback):
+        self.file.close()
+
+        if os.path.exists(self.log_file_path):
```
andy31415, line 79: This has a sideffect of the log files getting removed after dumping. This is not an intuitive sideffect of "dump files to console".

Unsure how to handle it, however maybe the name of the manager could be something that more closely reflects what it does. There seems nothing here that says "log file specific": this is a file context that removes files after processing. Maybe `RemoveAfterProcessing` or `OpenThenRemove` would be a better name than `LogFileManager`

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Sure, I will update with this logic throughout.

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,305 @@
...
+    #   exception_value: The value of the exception, if any.
+    #   traceback: The traceback of the exception.
+    def __exit__(self, exception_type, exception_value, traceback):
+        self.process.terminate()
+        self.process.wait()
+
+
+# Dump the contents of a log file to the console.
+#   log_file_path: The path to the log file.
+def dump_logs_to_console(log_file_path: str):
```
andy31415, line 115: Consider adding information that passed in logs are removed once processed. That may not be intuitive. Something like `dump_temporary_logs_to_console` and comment should say that the log file will get removed once the function exits.

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,305 @@
...
+
+# Log 'Discovery failed!' as an error, dump the contents of the log files
+# to the console, exit on error.
+#   log_file_paths: A list of paths to the log files, i.e. the path to the
+#                   Linux tv-casting-app logs and the tv-app logs.
+def handle_discovery_failure(log_file_paths: List[str]):
+    logging.error('Discovery failed!')
+
+    for log_file_path in log_file_paths:
+        dump_logs_to_console(log_file_path)
```
andy31415, line 131: nit: technically this will leave some files not removed if a dump of several files fails in the middle. For example if I dump files A,B,C and opening B fails for some reason, then we have:
  - dump A, remove A
  - start dump B ... throw
  
then once this handle exits, we both do not `sys.exit` and file C is not even attempted to be removed.

Maybe we should wrap these dumps into a try/except with a `logging.exception(f"Failed to dump {log_file_path}")` to prevent this.

scripts/tests/run_tv_casting_test.py
andy31415: You seem to already use LogFileManager inside the main test ... in that case does this still need to be a log file manager that removes files? Maybe yes because sys.exit, however then maybe add a comment explaining this.

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,305 @@
...
+                                valid_device_type: Optional[str] = None
+
+                            if continue_parsing:
+                                # Check if the Vendor ID, Product ID, and Device Type match the expected constant values.
+                                # If they do not match, then handle the discovery failure.
+                                if 'Vendor ID:' in line:
+                                    valid_vendor_id = validate_value(VENDOR_ID, line, 'Vendor ID')
+
+                                    if not valid_vendor_id:
+                                        handle_discovery_failure([linux_tv_app_log_path, linux_tv_casting_app_log_path])
```
andy31415, line 264: Indents here are too deep. I cannot follow such a large and indented method easily. Could you split it in separate functions so that things are easier read?

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,305 @@
...
+@click.option('--tv-casting-app-rel-path', type=str, default='out/tv-casting-app/chip-tv-casting-app', help='Path to the Linux tv-casting-app executable.')
+def test_discovery_fn(tv_app_rel_path, tv_casting_app_rel_path):
+
+    # Store the log files to a temporary directory.
+    with tempfile.TemporaryDirectory() as temp_dir:
+        linux_tv_app_log_path = os.path.join(temp_dir, LINUX_TV_APP_LOGS)
+        linux_tv_casting_app_log_path = os.path.join(temp_dir, LINUX_TV_CASTING_APP_LOGS)
+
+        # Open and write to the log file for the Linux tv-app.
+        with LogFileManager(linux_tv_app_log_path, 'w') as linux_tv_app_log_file:
```
andy31415, line 190: You could create a compund with for both log paths, so that things do not indent as much:

https://docs.python.org/3/reference/compound_stmts.html#with

https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory seems to indicate that `delete=True` is the default and cleanup will be done automatically. If so why do we need a separate class that cleans thigns up? Could we just do 

```py
with open(linux_tv_app_log_path, 'w') as linux_tv_app_log_file, open(linux_tv_casting_app_log_path, 'w') as linux_tv_casting_app_log_file:
```

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: This is updated in the latest commit.

Note: The vendor ID is set in [connectedhomeip/src/include/platform/CHIPDeviceConfig.h](https://github.com/project-chip/connectedhomeip/blob/1bc65f759e481c20d2510574f80e6376259c9ed4/src/include/platform/CHIPDeviceConfig.h#L180-L192)

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: The product ID is also set in the same file as the Vendor ID (mentioned in a previous reply): [connectedhomeip/src/include/platform/CHIPDeviceConfig.h](https://github.com/project-chip/connectedhomeip/blob/master/src/include/platform/CHIPDeviceConfig.h#L203-L215)

The constants of vendor ID=0xFFF1, product ID=0x8001, and device type=0x23 are what identifies the Linux tv app.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: This has been updated in the latest commit.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: This has been removed in the latest commit.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: The latest commit contains changes to follow the doc-style comment format.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Sure, separate functions have been created in the latest commit.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Good catch! I have removed the LogFileManager class as the temporary directory will take care of the cleanup.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: The LogFileManager class has been removed as the temporary directory takes care of the file clean up as noted in another comment.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: This function name has been updated per the suggestion.

As noted in another comment and reply, the LogFileManger has been removed as the temporary directory takes care of the file clean up.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Great suggestion! The change has been added to the latest commit.

scripts/tests/run_tv_casting_test.py
```diff
@@ -0,0 +1,224 @@
...
+        self.process = subprocess.Popen(self.command, stdout=self.stdout, stderr=self.stderr, text=True)
+        return self.process
+
+    def __exit__(self, exception_type, exception_value, traceback):
+        self.process.terminate()
+        self.process.wait()
+
+
+def dump_temporary_logs_to_console(log_file_path: str):
+    """Dump log file to the console; log file will be removed once the function exits."""
```
andy31415, line 65:
```suggestion
    """Write the entire content of `log_file_path` to the console."""
```

andy31415 (approved):

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Suggestion has been added to the latest commit.

PR #32921, opened 2024-04-09
TC-RR-1.1: Fix for pre-existing fabric removal

cecille:
This test assumed that the client ordering matched to the ordering of the fabric table, but this is not the case if there is a fabric on the device before the test starts. In this case, the initial fabric is in table slot 1 with index 1, the test starts in table slot 2, with fabric index 2. Then the initial fabric is removed from table slot 1, and the test adds a new fabric. The new fabric is allocated fabric index 3, but appears in slot 1 in the table, and the order between the controllers and the fabric table as read off the device is now out of sync.

Instead, name the controllers based on the fabric index rather than the index in the fabric table.

TEST: commissioned all-clusters using chip-tool then ran RR-1.1.
      Test now passes, whereas before there was a failure.

### DIFF
--- a/src/python_testing/TC_RR_1_1.py
+++ b/src/python_testing/TC_RR_1_1.py
@@ -37,7 +37,7 @@ from TC_SC_3_6 import AttributeChangeAccumulator, ResubscriptionCatcher
 
 
 def generate_controller_name(fabric_index: int, controller_index: int):
-    return f"RD{fabric_index}{str(chr(ord('A') + controller_index))}"
+    return f"RD{fabric_index}{string.ascii_uppercase[controller_index]))}"
 
 
 class TC_RR_1_1(MatterBaseTest):
--- a/src/python_testing/TC_RR_1_1.py
+++ b/src/python_testing/TC_RR_1_1.py
@@ -164,7 +164,6 @@ class TC_RR_1_1(MatterBaseTest):
                 dev_ctrl, node_id=self.dut_node_id, endpoint=0,
                 attribute=Clusters.OperationalCredentials.Attributes.Fabrics, fabricFiltered=False)
             current_fabric_index = await self.read_single_attribute_check_success(cluster=Clusters.OperationalCredentials, attribute=Clusters.OperationalCredentials.Attributes.CurrentFabricIndex)
-            print(fabrics)
             for fabric in fabrics:
                 if fabric.fabricIndex == current_fabric_index:
                     continue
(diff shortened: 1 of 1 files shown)


src/python_testing/TC_RR_1_1.py
```diff
@@ -36,6 +36,10 @@
 #
 
 
+def generate_controller_name(fabric_index: int, controller_index: int):
+    return f"RD{fabric_index}{str(chr(ord('A') + controller_index))}"
```
andy31415, line 40: nit:

```suggestion
    return f"RD{fabric_index}{string.ascii_uppercase[controller_index]))}"
```

assuming you `import string`

src/python_testing/TC_RR_1_1.py
```diff
@@ -163,11 +163,12 @@ async def test_TC_RR_1_1(self):
             fabrics: List[Clusters.OperationalCredentials.Structs.FabricDescriptorStruct] = await self.read_single_attribute(
                 dev_ctrl, node_id=self.dut_node_id, endpoint=0,
                 attribute=Clusters.OperationalCredentials.Attributes.Fabrics, fabricFiltered=False)
+            current_fabric_index = await self.read_single_attribute_check_success(cluster=Clusters.OperationalCredentials, attribute=Clusters.OperationalCredentials.Attributes.CurrentFabricIndex)
+            print(fabrics)
```
andy31415, line 167: do we want to keep this print in the test? It seems it is for debug mostly. Maybe logging.debug would be better.

cecille: nah, it snuck in there when I was trying to figure out what the deuce was going on. Will remove.

cecille, line 167:
```suggestion
```

src/python_testing/TC_RR_1_1.py
```diff
@@ -224,10 +225,9 @@ async def test_TC_RR_1_1(self):
         # Step 2: Set the Label field for each fabric and BasicInformation.NodeLabel to 32 characters
         logging.info("Step 2: Setting the Label field for each fabric and BasicInformation.NodeLabel to 32 characters")
 
-        for table_idx in range(len(fabric_table)):
+        for fabric in fabric_table:
             # Client is client A for each fabric to set the Label field
-            fabric = fabric_table[table_idx]
-            client_name = "RD%dA" % table_idx
+            client_name = "RD%dA" % fabric.fabricIndex
```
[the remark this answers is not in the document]
cecille: sure. changed

andy31415 (approved):

PR #32923, opened 2024-04-10
Darwin: Leave the work queue running

ksperling-apple:
Start the work queue when the MTRDeviceControllerFactory is initialized, and leave it running. This avoids having to conditionally dispatch to the work queue in various scenarios and lets us simplify controller startup and shutdown.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
+++ b/src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
@@ -215,6 +215,10 @@ static void ShutdownOnExit()
         return nil;
     }
 
+    // Start the work queue and leave it running. There is no performance
+    // cost to having an idle dispatch queue, and it simplifies our logic.
+    DeviceLayer::PlatformMgrImpl().StartEventLoopTask();
+
     _running = NO;
     _chipWorkQueue = DeviceLayer::PlatformMgrImpl().GetWorkQueue();
     _controllerFactory = &DeviceControllerFactory::GetInstance();
@@ -326,6 +330,7 @@ static void ShutdownOnExit()
 
 - (void)cleanupStartupObjects
 {
+    assertChipStackLockedByCurrentThread();
     MTR_LOG_INFO("Cleaning startup objects in controller factory");
 
     // Make sure the deinit order here is the reverse of the init order in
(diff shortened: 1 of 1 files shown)


src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
```diff
@@ -1060,60 +1039,45 @@ - (void)controllerShuttingDown:(MTRDeviceController *)controller
...
-            _otaProviderEndpoint = nil;
-        }
-
-        sharedCleanupBlock();
-    } else {
-        // Do the controller shutdown on the Matter work queue.
-        dispatch_sync(_chipWorkQueue, ^{
             if (_otaProviderDelegateBridge) {
-                _otaProviderDelegateBridge->ControllerShuttingDown(controller);
+                _otaProviderDelegateBridge->Shutdown();
```
bzbarsky-apple, line 1069: We used to do this before last-controller shutdown... But now we will do it after, no?  And that code uses the exchange manager and whatnot.  If the system state is shut down by then, won't those objects be dead after we do shutDownCppController?

src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
```diff
@@ -399,22 +404,13 @@ - (CHIP_ERROR)_initFabricTable:(FabricTable &)fabricTable
...
-        fillListBlock();
-    }
+    });
 
     if (listFilled == NO) {
         return nil;
     }
 
-    return [NSArray arrayWithArray:fabricList];
+    return fabricList;
```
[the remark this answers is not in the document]
ksperling-apple: Yes, we're not holding onto it, so making it actually immutable under the hood doesn't really do anything useful.

src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
ksperling-apple: The `_otaProviderDelegateBridge->ControllerShuttingDown(controller);` is still there above, before we call shutDownCppController. I think this is a case of the diff making it look confusing.

bzbarsky-apple: I realize the `ControllerShuttingDown()` bit is there.

But I am looking at the implementation of `Shutdown()` and it does:

```
        mExchangeMgr->UnregisterUnsolicitedMessageHandlerForProtocol(Protocols::BDX::Id);
```

ksperling-apple: Ah got it, fixed.

bzbarsky-apple (approved):

ksperling-apple:
Fast-tracking Darwin-only change

PR #32924, opened 2024-04-10
[crypto] PSA SPAKE2+ introduction

ArekBalysNordic:
Implement Matter's SPAKE2+ protocol using PSA crypto API.
The SPAKE2+ protocol requires the PSA crypto API in version v1.1 and all Spake2p class methods which are used in Matter Core.

You can use it by enabling the `CHIP_CRYPTO_PSA_SPAKE2P` config.

Aligned Spake2+ algorithm to the newest PSA Crypto API version that allows using the `psa_pake_get_shared_key` function.

- Use the PSA implementation in the PASE session for the nRF Connect platform.

### DIFF
--- a/config/nrfconnect/chip-module/CMakeLists.txt
+++ b/config/nrfconnect/chip-module/CMakeLists.txt
@@ -168,7 +168,8 @@ else()
 endif()
 
 if (CONFIG_CHIP_CRYPTO_PSA)
-    matter_add_gn_arg_string("chip_crypto" "psa")
+    matter_add_gn_arg_string("chip_crypto"             "psa")
+    matter_add_gn_arg_bool  ("chip_crypto_psa_spake2p" TRUE)
 endif()
 
 if (BOARD STREQUAL "native_posix")
--- a/config/nrfconnect/chip-module/CMakeLists.txt
+++ b/config/nrfconnect/chip-module/CMakeLists.txt
@@ -169,7 +169,7 @@ endif()
 
 if (CONFIG_CHIP_CRYPTO_PSA)
     matter_add_gn_arg_string("chip_crypto"             "psa")
-    matter_add_gn_arg_bool  ("chip_crypto_psa_spake2p" TRUE)
+    matter_add_gn_arg_bool  ("chip_crypto_psa_spake2p" CONFIG_PSA_WANT_ALG_SPAKE2P_MATTER)
 endif()
 
 if (BOARD STREQUAL "native_posix")
--- a/src/crypto/PSASpake2p.cpp
+++ b/src/crypto/PSASpake2p.cpp
@@ -32,14 +32,6 @@ CHIP_ERROR PSASpake2p_P256_SHA256_HKDF_HMAC::Init(const uint8_t * context, size_
 
     VerifyOrReturnError(context_len <= sizeof(mContext), CHIP_ERROR_BUFFER_TOO_SMALL);
 
-    psa_pake_cipher_suite_t cs = PSA_PAKE_CIPHER_SUITE_INIT;
-    psa_pake_cs_set_algorithm(&cs, PSA_ALG_SPAKE2P);
-    psa_pake_cs_set_primitive(&cs, PSA_PAKE_PRIMITIVE(PSA_PAKE_PRIMITIVE_TYPE_ECC, PSA_ECC_FAMILY_SECP_R1, 256));
-    psa_pake_cs_set_hash(&cs, PSA_ALG_SHA_256);
-
-    psa_status_t status = psa_pake_setup(&mOperation, &cs);
-    VerifyOrReturnError(status == PSA_SUCCESS, CHIP_ERROR_INTERNAL);
-
     memcpy(mContext, context, context_len);
     mContextLen = context_len;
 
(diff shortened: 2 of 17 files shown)


src/protocols/secure_channel/PASESession.h
```diff
@@ -198,7 +201,11 @@ class DLL_EXPORT PASESession : public Messaging::UnsolicitedMessageHandler,
     // mNextExpectedMsg is set when we are expecting a message.
     Optional<Protocols::SecureChannel::MsgType> mNextExpectedMsg;
 
+#if CHIP_CRYPTO_PSA_SPAKE2P
+    Crypto::PSASpake2p_P256_SHA256_HKDF_HMAC mSpake2p;
```
bzbarsky-apple, line 205: Shouldn't CryptoPAL handle this transparently, with `Crypto::Spake2p_P256_SHA256_HKDF_HMAC` being whatever type is needed for the current configuration?

ArekBalysNordic: We can handle it this way, but if we want to do it properly, we need to modify and split the Spake2p API into a separate file and then we can switch between its implementation. 

Currently, the Spake2p interface assumes that we will use raw-key implementation, whereas in PSA we don't use raw keys at all, so most of the API shouldn't be purely virtual.  The problem is, if I modify the Spake2p API and split it into separate files now the PR will be huge and hard to review. On the other hand, I can provide some #ifdefs in CryptoPAL files to switch between those backends but it will make the code less understandable. 

I guess, for now, we can switch it here and in the next step separate all spake2p backends into separate classes that inherit from the spake2p class.

What do you think about those approaches?

config/nrfconnect/chip-module/CMakeLists.txt
```diff
@@ -169,6 +169,10 @@ endif()
 
 if (CONFIG_CHIP_CRYPTO_PSA)
     matter_add_gn_arg_string("chip_crypto" "psa")
+    # PSA Spake2+ implementation is not ready for the POSIX target yet
+    if((NOT BOARD STREQUAL "native_posix") AND (NOT BOARD STREQUAL "native_posix_64"))
+        matter_add_gn_arg_bool  ("chip_crypto_psa_spake2p" TRUE)
+    endif()
```
Damian-Nordic, line 175:
```suggestion
    matter_add_gn_arg_bool  ("chip_crypto_psa_spake2p" NOT ARCH_POSIX)
```

or even:

```suggestion
    matter_add_gn_arg_bool  ("chip_crypto_psa_spake2p" CONFIG_PSA_WANT_ALG_SPAKE2P_MATTER)
```

just in case SPAKE2P is not available on some crypto configuration.

Damian-Nordic (approved):

src/protocols/secure_channel/PASESession.h
bzbarsky-apple: I think that's probably fine, as long as there is a concrete plan to remove the ifdef from here.

LuDuda (approved):

src/protocols/secure_channel/PASESession.h
ArekBalysNordic: I created a feature request here: https://github.com/project-chip/connectedhomeip/issues/32985. I think we can consider extracting this spake2p API and in this way, we will make the code clearer and allow using raw key and PSA key backends.  After that, we will be able to remove ifdef from here and switch between backends in CryptoPAL.

src/protocols/secure_channel/CASESession.h
```diff
@@ -146,7 +146,7 @@ class DLL_EXPORT CASESession : public Messaging::UnsolicitedMessageHandler,
      * @param session     Reference to the secure session that will be initialized once session establishment is complete
      * @return CHIP_ERROR The result of session derivation
      */
-    CHIP_ERROR DeriveSecureSession(CryptoContext & session) const override;
```
bzbarsky-apple, line 149: Why do these various things become non-const?  Is this because PSA mutates some state of things it uses when doing key derivation?  Do the things it uses need to be members?  If so, would it be appropriate for them to be `mutable` members (much as I hesitate to suggest that)?

Damian-Nordic: It's because [psa_pake_get_shared_key()](https://github.com/ARM-software/psa-api/pull/114/files#diff-c5fda0d6f379353d5f5defd8e7e8fb3aec52e0d9806698104021dc7e18817f99R48) function takes a non-const pointer to the operation that is a member of `PSASpake2p_P256_SHA256_HKDF_HMAC` (which in turn is a member of `PASESession`). Turning the methods to non-const is my fault as I suggested @ArekBalysNordic that it would be a clear solution than `const_cast` or `mutable` :) (from my experience, some devs don't like `mutable` for things other than mutexes as it hides the fact of a method having a side effect but it's probably acceptable in this case).

src/crypto/PSASpake2p.cpp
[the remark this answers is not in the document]
Damian-Nordic: Sure thing!

jmartinez-silabs (approved):

src/protocols/secure_channel/CASESession.h
ArekBalysNordic: @bzbarsky-apple so it looks like we only need a non-const pointer for the PSA stuff, but not for the raw-key. We've got a few ways to tackle this: like I did before, by taking out the const from the "parent" methods, or we could try using `const_cast` or maybe even mutable. But As @Damian-Nordic wrote, it seemed like those options weren't really popular with the devs. What do you think? Is it a no-go to drop the `const` from those methods when we're working with the new Spake2+ stuff that can't be const?

bzbarsky-apple: I don't think it's necessarily a problem; I am just trying to understand the situation.

So getting a shared key with PSA requires passing in some mutable state in addition to the place to put the key id?  And that mutable state needs to be in some way persistent (as in, the mutated version needs to be kept around)?  How long does that mutated state need to be kept around?

Damian-Nordic: That mutable state is `psa_pake_operation_t` that acts as a context for the entire PAKE operation:
1. First, you feed this context with subsequent input values like passwords, salt etc.
2. Then you use the same context to output the results. PSA APIs often require non-const pointers to the operation even for getting the results, probably to enable implementations to complete some calculations at the last moment.

bzbarsky-apple: OK.  And I assume psa_pake_operation_t is non-copyable, in the sense that you couldn't have mutable methods that update its state with salt/etc and then a const method that copies the state, gets the key with the copied state (letting PSA do whatever computations on that copied state), and so on?

(Ignoring for the moment whether that would likely be inefficient; just trying to wrap my head around the problem space.)

Damian-Nordic: I can't find anything about copyability of the operation on the spec but given that there's no official API other than `memcpy` for that I doubt we can rely that it will always work. I guess if you think it's better not to remove the consts for this method just because PSA requires that, I would say adding `mutable` is the best workaround in this case.

ArekBalysNordic: What do you think @bzbarsky-apple ?

bzbarsky-apple: Let's just make these non-const, if there are crypto backends where this really is a conceptually non-const operation.

PR #32927, opened 2024-04-10
[Silabs] Refactor Silabs Test Event Trigger impl to be compliant with its requirements

mkardous-silabs:
#### Description
PR refactors the Silabs Test event triggers to be inline with the enableKey handling requirements

- Adds a DataProvider function to get the enableKey from nvm3 instead of storing it in RAM
- Adds a default enableKey for testing purpose that is removed in a release build (gated by NDEBUG)

PR also changes the smoke&co app to be an ICD app by default. Since the ICDM cluster was present the zap configuration, the app could not link succesfully without building Matter with the ICD build flag to true.

#### Notes to reviewers
PR doesn't touch `emberAfHandleEventTrigger` that acts like the "global" silabs test event trigger handler. This will be refactored in a follow up PR.

#### Tests 
Manual tests to validate that the smoke&co test event trigger still works with the new implementation

- Validate default behavior with the test event trigger happy and sad path (correct and wrong key)
- Validate default behavior with nvm3 prescense with a valid and invalid key in storage
- Validate release build behavior to confirm that the test key isn't usable


jmartinez-silabs (approved):

examples/platform/silabs/efr32/BUILD.gn
```diff
@@ -41,11 +41,11 @@ declare_args() {
   chip_default_wifi_psk = ""
 
   # Enable TestEventTrigger in GeneralDiagnostics cluster
-  silabs_test_event_trigger_enabled = false
+  sl_enable_test_event_trigger = false
```
tcarmelveilleux, line 44: If this is build-configured, it could lead to the device appearing in the wild with the feature enabled in firmware. Consider making this configurable in factory partition/settings if possible.

mkardous-silabs: Unless i misunderstood the test event triger use, isn't that the comprimise for the test event trigger? If it is required for certification, it ends up being dead code in release since you can't recompile your app.

At release, since the key won't be present in nvm3, a client won't be able to trigger test behavior.

examples/smoke-co-alarm-app/silabs/build_for_wifi_args.gni
```diff
@@ -21,3 +21,17 @@ import("${chip_root}/src/platform/silabs/wifi_args.gni")
 chip_enable_ota_requestor = true
 app_data_model =
     "${chip_root}/examples/smoke-co-alarm-app/smoke-co-alarm-common"
+
+sl_enable_test_event_trigger = true
+
+# ICD Default configurations
+chip_enable_icd_server = true
+chip_subscription_timeout_resumption = false
+sl_use_subscription_synching = true
```
tcarmelveilleux, line 30: `synching` --> typo on syncing

examples/platform/silabs/SilabsDeviceDataProvider.cpp
```diff
@@ -510,6 +367,43 @@ CHIP_ERROR SilabsDeviceDataProvider::GetProductLabel(char * buf, size_t bufSize)
     return SilabsConfig::ReadConfigValueStr(SilabsConfig::KConfigKey_ProductLabel, buf, bufSize, productLabelLen);
 }
 
+#ifdef SL_MATTER_TEST_EVENT_TRIGGER_ENABLED
```
tcarmelveilleux, line 370: This should be runtime...

examples/smoke-co-alarm-app/silabs/build_for_wifi_args.gni
mkardous-silabs: done

examples/platform/silabs/efr32/BUILD.gn
mkardous-silabs: Just to make sure i understand, ` sl_enable_test_event_trigger` and `SL_MATTER_TEST_EVENT_TRIGGER_ENABLED` should just not be present in the code. In other words, test event trigger delagate is always present and wether handlers are registers and an enableKey provide be application specific.

mkardous-silabs: Synced on slack - resolving the comment

examples/platform/silabs/SilabsDeviceDataProvider.cpp
mkardous-silabs: Synced on slack - resolving the comment

examples/platform/silabs/SiWx917/BUILD.gn
```diff
@@ -80,6 +77,30 @@ config("chip_examples_project_config") {
   ]
 }
 
+config("test-event-trigger-config") {
+  defines = [ "SL_MATTER_TEST_EVENT_TRIGGER_ENABLED" ]
```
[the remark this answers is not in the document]
mkardous-silabs: Agreed, we have an action item to do this clean up at some point. For now, i'm goign to leave it as is until i get some time to clean up our defines. PR is blocking follow up work which is getting urgent.

andy31415 (approved):

PR #32928, opened 2024-04-10
Fix up mock ember implementation (use after free and missing function)

andy31415:
MockEndpointConfig is self-referential because the ember structure contains a pointer to an internal vector of data. This means the default copy constructor cannot be used.

Also provide the `emberAfDataVersionStorage` which is a public API that is supposed to be available to the IM.

### Changes

- provide data version function as described by attribute-storage.h
- fix self-referencing structure copying for mock node configurations

### DIFF
--- a/src/app/util/mock/MockNodeConfig.h
+++ b/src/app/util/mock/MockNodeConfig.h
@@ -61,6 +61,10 @@ struct MockEndpointConfig
 {
     MockEndpointConfig(EndpointId aId, std::initializer_list<MockClusterConfig> aClusters = {});
 
+    // Cluster-config is self-referntial: mEmberCluster.clusters references  mEmberClusters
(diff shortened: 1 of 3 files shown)


jmartinez-silabs (approved):

yunhanw-google (approved):

src/app/util/mock/MockNodeConfig.h
```diff
@@ -61,6 +61,10 @@ struct MockEndpointConfig
 {
     MockEndpointConfig(EndpointId aId, std::initializer_list<MockClusterConfig> aClusters = {});
 
+    // Cluster-config is self-referntial: mEmberCluster.clusters references  mEmberClusters
```
bzbarsky-apple, line 64: "referential"

PR #32930, opened 2024-04-10
Updating the existing xml with features per cluster and features per device type

brdandu:
JIRA: ZAPP-1346

No direct chip repo ticket associated to the changes here. However, these changes are associated to populating the feature map attribute based on the device type selected in ZAP. See ZAP PR: https://github.com/project-chip/zap/pull/1300 .

Note: The ZAP changes here only populate the feature map attribute for mandatory features for now during the creation of an endpoint. Existing apps will also get a warning during the .zap file open if the feature map attribute is not correctly set as per the device type specification for features.

### DIFF
--- a/src/app/zap-templates/zcl/data-model/chip/administrator-commissioning-cluster.xml
+++ b/src/app/zap-templates/zcl/data-model/chip/administrator-commissioning-cluster.xml
@@ -69,9 +69,4 @@ limitations under the License.
     </command>
 
   </cluster>
-
-  <bitmap name="Feature" type="bitmap32">
-    <cluster code="0x003c" />
-    <field name="Basic" mask="0x01" />
-  </bitmap>
 </configurator>
--- a/src/app/zap-templates/zcl/data-model/chip/administrator-commissioning-cluster.xml
+++ b/src/app/zap-templates/zcl/data-model/chip/administrator-commissioning-cluster.xml
@@ -38,6 +38,11 @@ limitations under the License.
     <define>ADMINISTRATOR_COMMISSIONING_CLUSTER</define>
     <description>Commands to trigger a Node to allow a new Administrator to commission it.</description>
 
+    <features>
+      <feature bit="0" code="BC" name="Basic" summary="Node supports Basic Commissioning Method.">
+        <optionalConform/>
+      </feature>
+    </features>
     <attribute side="server" code="0x0000" define="WINDOW_STATUS" type="CommissioningWindowStatusEnum" writable="false" optional="false">WindowStatus</attribute>
     <attribute side="server" code="0x0001" define="ADMIN_FABRIC_INDEX" type="fabric_idx" writable="false" isNullable="true" optional="false">AdminFabricIndex</attribute>
     <attribute side="server" code="0x0002" define="ADMIN_VENDOR_ID" type="vendor_id" writable="false" isNullable="true" optional="false">AdminVendorId</attribute>
--- a/src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
+++ b/src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
@@ -107,6 +107,24 @@ limitations under the License.
 
     <globalAttribute side="either" code="0xFFFD" value="6"/>
 
+    <features>
+      <feature bit="0" code="HS" name="Hue/Saturation" summary="Supports color specification via hue/saturation.">
+        <optionalConform/>
+      </feature>
+      <feature bit="1" code="EHUE" name="Enhanced Hue" summary="Enhanced hue is supported.">
+        <optionalConform/>
+      </feature>
+      <feature bit="2" code="CL" name="Color loop" summary="Color loop is supported.">
+        <optionalConform/>
+      </feature>
+      <feature bit="3" code="XY" name="XY" summary="Supports color specification via XY.">
+        <optionalConform/>
+      </feature>
+      <feature bit="4" code="CT" name="Color temperature" summary="Supports specification of color temperature.">
+        <optionalConform/>
+      </feature>
+    </features>
+
     <attribute side="server" code="0x0000" define="COLOR_CONTROL_CURRENT_HUE" type="int8u" min="0x00" max="0xFE" writable="false" reportable="true" default="0x00" optional="true">CurrentHue</attribute>
     <!-- CURRENT_HUE -->
     <attribute side="server" code="0x0001" define="COLOR_CONTROL_CURRENT_SATURATION" type="int8u" min="0x00" max="0xFE" writable="false" reportable="true" default="0x00" optional="true">CurrentSaturation</attribute>
--- a/src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
+++ b/src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
@@ -15,14 +15,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 <configurator>
-  <bitmap name="Feature" type="bitmap32">
-    <cluster code="0x0300" />
-    <field name="HueAndSaturation" mask="0x01" />
-    <field name="EnhancedHue" mask="0x02" />
-    <field name="ColorLoop" mask="0x04" />
-    <field name="XY" mask="0x08" />
-    <field name="ColorTemperature" mask="0x10" />
-  </bitmap>
+
 
   <enum name="HueDirection" type="enum8">
     <cluster code="0x0300"/>
--- a/src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
+++ b/src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
@@ -101,7 +101,7 @@ limitations under the License.
     <globalAttribute side="either" code="0xFFFD" value="6"/>
 
     <features>
-      <feature bit="0" code="HS" name="Hue/Saturation" summary="Supports color specification via hue/saturation.">
+      <feature bit="0" code="HS" name="Hue And Saturation" summary="Supports color specification via hue/saturation.">
         <optionalConform/>
       </feature>
       <feature bit="1" code="EHUE" name="Enhanced Hue" summary="Enhanced hue is supported.">
--- a/src/app/zap-templates/zcl/data-model/chip/dishwasher-alarm-cluster.xml
+++ b/src/app/zap-templates/zcl/data-model/chip/dishwasher-alarm-cluster.xml
@@ -37,6 +37,12 @@ limitations under the License.
     <client tick="false" init="false">true</client>
     <server tick="false" init="false">true</server>
 
+    <features>
+      <feature bit="0" code="RESET" name="Reset" summary="Supports the ability to reset alarms">
+        <optionalConform/>
+      </feature>
+    </features>
+
     <attribute side="server" code="0x0000" define="MASK"      type="AlarmBitmap" default="0" writable="false" optional="false">Mask</attribute>
     <attribute side="server" code="0x0001" define="LATCH"     type="AlarmBitmap" default="0" writable="false" optional="true">Latch</attribute>
     <attribute side="server" code="0x0002" define="STATE"     type="AlarmBitmap" default="0" writable="false" optional="false">State</attribute>
--- a/src/app/zap-templates/zcl/data-model/chip/dishwasher-alarm-cluster.xml
+++ b/src/app/zap-templates/zcl/data-model/chip/dishwasher-alarm-cluster.xml
@@ -60,9 +60,4 @@ limitations under the License.
         <field id="3" name="Mask" type="AlarmBitmap" />
     </event>
   </cluster>
-
-  <bitmap name="Feature" type="bitmap32">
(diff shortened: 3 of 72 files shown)


src/app/zap-templates/zcl/data-model/chip/administrator-commissioning-cluster.xml
```diff
@@ -38,6 +38,11 @@ limitations under the License.
     <define>ADMINISTRATOR_COMMISSIONING_CLUSTER</define>
     <description>Commands to trigger a Node to allow a new Administrator to commission it.</description>
 
+    <features>
```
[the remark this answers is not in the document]
brdandu: From what I understand the new xml comes from https://github.com/project-chip/connectedhomeip/tree/master/data_model . I have used the new xml to update the existing xml since @tecimovic is still working on updating ZAP to the new xml from what I know. The only reason I updated the existing xml is so that we can start having the warnings for existing apps without having to wait for the new xml. Made sure the formatting comes from the new xml so functionality remains the same as much as possible when we make permanent switch.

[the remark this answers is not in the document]
andy31415: I would have expected a zap update to be included together with this metadata update, so that we can verify that the new information takes effect. I think when we do that, we should remove the old feature bitmaps and use these new items instead. At that time we should verify that code generation gives no differences.

As it stands, this duplicates information and having two "sources of truth" is unlikely to stay in sync (in this case I think these new "feature" entries will just be ignored by all tooling we have)

andy31415 (changes requested):
Having this delta early is bound to become out of sync. This change should only happen when zap can understand the new layout (at which point we should delete the old feature bitmaps)

src/app/zap-templates/zcl/data-model/chip/administrator-commissioning-cluster.xml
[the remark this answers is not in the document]
hasty: I can pretty easily add a parameter to alchemy to auto-generate these and remove the old ones, when we're ready on the ZAP side.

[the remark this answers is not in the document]
bzbarsky-apple: @brdandu My point was that we should not have both this bit and the Feature bitmaps in the "old" XML (which is the only XML used for anything at the moment).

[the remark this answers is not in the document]
hasty: OK, released v0.4.8 of alchemy, which has a new "featureXML" parameter that will generate these directly from the spec, and automatically remove the old feature bitmap. It also includes the new feature XML format on every cluster, not just the first one in each file. This parameter is off by default.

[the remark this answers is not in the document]
brdandu: I can check on the removal of the old feature bit information in the xml. Not sure if I will break something or not by removing that information. Also do not know where all in the generation that information is currently used.
However I do want to say that we should not withhold features which wait on the new xml. If we have a feature why not enable it via the old xml?
I can start looking into removing the old feature bit data from the xml and see what happens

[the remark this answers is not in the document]
hasty: Oh, it'll 100% break; if the zap tool hasn't been updated to look for this new feature schema, and the old feature schema has been removed, it'll fail. That's why I put it in as a flag to use once whatever has to happen in the zap tool has happened.

One weird aspect of the new approach here is that it's just pulling over the conformance schema from the data model side. With the existing approach the zap tool doesn't get into the weeds on feature-based conformance; everything boils down to "is this mandatory or not mandatory?" via an "optional" attribute.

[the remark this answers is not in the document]
brdandu: Ok I see that <bitmap name="Feature"...> has the same data as <features> but as of now <bitmap> is considered as one of the data types required by the cluster and it is not equivalent to <features>. features is not considered as a data type but a separate entity in the ZAP schema since it is not like any other bitmap relationship we have. Could we keep them separate for now and deal with combining the two in a separate PR?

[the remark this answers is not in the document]
hasty: I think Boris' point was: "why add this to the XML now when nothing can use it?" We should update the ZAP tool to understand the new schema (as well as the old as a fallback), and then we can add the new schema to the XML while removing the old and bumping the min ZAP version. Having both just means something else that could fall out of sync.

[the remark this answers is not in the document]
bzbarsky-apple: @brdandu I think the basic point is: we do not want the manual maintenance burden of maintaining these two sets of data.  And since only one of them leads to visible anything on the Matter side, that means that the other will break as soon as someone adds a new feature bit to a spec, because no one will realize it should be added to that part of the XML.

If there is a concrete commitment with a timeline to not having the two things around at once, and some urgency to the UI behavior you describe, we can have the conversation about doing things this way and then making it right in a followup.  But we would need to understand the timelines involved to even start that conversation.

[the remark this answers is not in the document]
brdandu: Alright I added the ZAP changes required for this in https://github.com/project-chip/zap/pull/1300 and have removed the bitmap name='feature' from the xml. however the zap PR needs to be updated before this can be merged in.

bzbarsky-apple (changes requested):
There should be no codegen changes from this change.  Any codegen changes are bugs in either ZAP or the new XML bits.

src/app/zap-templates/zcl/data-model/chip/network-commissioning-cluster.xml
```diff
@@ -105,6 +97,18 @@ limitations under the License.
...
 
+        <features>
+            <feature bit="0" code="WI" name="WiFiNetworkInterface" summary="Wi-Fi related features">
+            <optionalConform choice="a"/>
+            </feature>
+            <feature bit="1" code="TH" name="ThreadNetworkInterface" summary="Thread related features">
+            <optionalConform choice="a"/>
+            </feature>
+            <feature bit="2" code="ET" name="EthernetNetworkInterface" summary="Ethernet related features">
+            <optionalConform choice="a"/>
```
brdandu, line 108: PerDeviceCredentials feature is not present in features

brdandu:
> There should be no codegen changes from this change. Any codegen changes are bugs in either ZAP or the new XML bits.

I have updated the xml the best I could. The remaining diffs seem to be issues with the old xml bitmap features which do not exist in the new xml.

src/app/zap-templates/zcl/data-model/chip/rvc-clean-mode-cluster.xml
```diff
@@ -61,10 +61,5 @@ limitations under the License.
     </command>
   </cluster>
 
-  <bitmap name="Feature" type="bitmap32">
-    <cluster code="0x0055"/>
-    <field name="NoFeatures" mask="0x0"/>
-    <!-- Feature OnOff with code 0x01 is deprecated -->
-  </bitmap>
```
brdandu, line 68: I see no features for this cluster in new xml

src/app/zap-templates/zcl/data-model/chip/rvc-run-mode-cluster.xml
```diff
@@ -68,10 +68,5 @@ limitations under the License.
     </command>
   </cluster>
 
-  <bitmap name="Feature" type="bitmap32">
-    <cluster code="0x0054"/>
-    <field name="NoFeatures" mask="0x0"/>
-    <!-- Feature OnOff with code 0x01 is deprecated -->
-  </bitmap>
```
brdandu, line 75: I see no features for this cluster in new xml

src/app/zap-templates/zcl/data-model/chip/dishwasher-alarm-cluster.xml
```diff
@@ -60,9 +60,4 @@ limitations under the License.
         <field id="3" name="Mask" type="AlarmBitmap" />
     </event>
   </cluster>
-
-  <bitmap name="Feature" type="bitmap32">
-    <cluster code="0x005D" />
-    <field name="Reset" mask="0x01" />
-  </bitmap>
```
brdandu, line 67: No features for this in new xml

src/app/zap-templates/zcl/data-model/chip/thermostat-cluster.xml
```diff
@@ -290,6 +277,45 @@ limitations under the License.
...
+        <optionalConform/>
+      </feature>
+      <feature bit="5" code="AUTO" name="AutoMode" summary="Supports a System Mode of Auto">
+        <optionalConform/>
+      </feature>
+      <feature bit="6" code="LTNE" name="LocalTemperatureNotExposed" summary="Thermostat does not expose the LocalTemperature Value in the LocalTemperature attribute">
+        <optionalConform/>
+      </feature>
+    </features>
+
```
brdandu, line 318: This is missing the following in the new xml
<field name="MatterScheduleConfiguration" mask="0x80"/>
     <field name="Presets" mask="0x100"/>
     <field name="Setpoints" mask="0x200"/>
     <field name="QueuedPresetsSupported" mask="0x400"/>

src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
```diff
@@ -107,6 +100,24 @@ limitations under the License.
 
     <globalAttribute side="either" code="0xFFFD" value="6"/>
 
+    <features>
+      <feature bit="0" code="HS" name="Hue/Saturation" summary="Supports color specification via hue/saturation.">
```
brdandu, line 104: HueAndSaturation became Hue/Saturation

brdandu:
> There should be no codegen changes from this change. Any codegen changes are bugs in either ZAP or the new XML bits.

Code gen changes are caused by the new xml since it does not match bitmap feature which existed in the xml. I do not see any zap issues here. If the new xml is the source of truth for features/bitmap feature then we should merge this PR.

src/app/zap-templates/zcl/data-model/chip/rvc-clean-mode-cluster.xml
bzbarsky-apple: Yes, this cluster has no features, but the bitmap can't be empty and the bitmap needs to exist.  Hence the NoFeatures thing.

We need a way to continue supporting that.

andy31415: Yes, we had a feature that disappeared, so at some point we had a bitmap. We could not completely remove the bitmap because backwards compatibility and an empty bitmap generates code that refuses to compile (python and I believe even obj-c compiler complains). Hence we created a fake `NoFeatures` value.

src/app/zap-templates/zcl/data-model/chip/dishwasher-alarm-cluster.xml
bzbarsky-apple: Has this shipped anywhere?  If so, it needs to stay, however we accomplish it.  I agree this is not in the spec right now, but clearly it was at some point, since the XML is here.  When was it removed, and why was the XML not updated?

src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
bzbarsky-apple: That's wrong and needs to be fixed.  This is long since shipped as API and cannot change.

src/app/zap-templates/zcl/data-model/chip/dishwasher-alarm-cluster.xml
brdandu: Added it back

src/app/zap-templates/zcl/data-model/chip/rvc-clean-mode-cluster.xml
brdandu: Added it back

src/app/zap-templates/zcl/data-model/chip/dishwasher-alarm-cluster.xml
bzbarsky-apple: To answer my own question: This is a derived cluster, and the feature is on the derived-from cluster.  The spec scraper is not understanding that structure, apparently.

bzbarsky-apple (approved):
Thank you; this is much better!

Please do make sure there is a followup tracking the duplicated bits for the alias clusters....

brdandu:
> Thank you; this is much better!
> 
> Please do make sure there is a followup tracking the duplicated bits for the alias clusters....

Np. Will follow up on the duplicate features across clusters

mkardous-silabs (approved):

jmartinez-silabs (approved):

andy31415 (approved):

PR #32931, opened 2024-04-10
[Silabs] Level Control revision 5

lpbeliveau-silabs:
LC revision was mistakenly bumped up, bumping it down.


bzbarsky-apple (approved):

jmartinez-silabs (approved):

mkardous-silabs (approved):
Does the wifi zap file need the same change?

PR #32935, opened 2024-04-10
Add std::source_location to ChipError for C++20 builds

mwswartwout:
Add std::source_location to ChipError when compiling with C++20.

This has the side effect of making the ChipError default constructor non-trivial, which made ChipDeviceEvent also not trivial. Switching ChipDeviceEvent construction to using the designated initializer syntax fixed this.


src/lib/core/CHIPError.h
```diff
@@ -159,10 +183,17 @@ class ChipError
      * @note
      *  This is intended to be used only in foreign function interfaces.
      */
+#if __cplusplus >= 202002L
+    explicit constexpr ChipError(StorageType error) : mError(error) CHIP_INITIALIZE_ERROR_SOURCE(nullptr, 0, std::source_location()) {}
```
tcarmelveilleux, line 187: Please add a unit test for this behavior in the chip error case, and include ifdef guards for c++20

src/lib/core/CHIPError.h
```diff
@@ -123,23 +131,39 @@ class ChipError
      * @note
      *  The result is valid only if CanEncapsulate() is true.
      */
+#if __cplusplus >= 202002L
+    constexpr ChipError(Range range, ValueType value) :
+        mError(MakeInteger(range, (value & MakeMask(0, kValueLength)))) CHIP_INITIALIZE_ERROR_SOURCE(nullptr, 0, std::source_location())
```
bzbarsky-apple, line 136: Won't this fail to compile if `__cplusplus >= 202002L` and `!CHIP_CONFIG_ERROR_SOURCE`?

src/lib/core/CHIPError.h
```diff
@@ -123,23 +131,39 @@ class ChipError
...
+    {}
+    constexpr ChipError(Range range, ValueType value, const char * file, unsigned int line, std::source_location location = std::source_location::current()) :
```
bzbarsky-apple, line 138: So how does this handle situations where the ChipError() constructor is called via a macro defined at location A but used at location B?  Which location ends up in `location`?

mwswartwout: It should always show location B: https://godbolt.org/z/oThxMKKM4

bzbarsky-apple: Excellent, thanks!

src/lib/core/CHIPError.h
mwswartwout: Good catch, fixed by the adding the loc argument to all instances of CHIP_INITIALIZE_ERROR_SOURCE

src/app/FailSafeContext.cpp
```diff
@@ -85,12 +85,11 @@ void FailSafeContext::ScheduleFailSafeCleanup(FabricIndex fabricIndex, bool addN
 
     SetFailSafeArmed(false);
 
-    ChipDeviceEvent event;
-    event.Type                                                = DeviceEventType::kFailSafeTimerExpired;
-    event.FailSafeTimerExpired.fabricIndex                    = fabricIndex;
-    event.FailSafeTimerExpired.addNocCommandHasBeenInvoked    = addNocCommandInvoked;
-    event.FailSafeTimerExpired.updateNocCommandHasBeenInvoked = updateNocCommandInvoked;
-    CHIP_ERROR status                                         = PlatformMgr().PostEvent(&event);
+    ChipDeviceEvent event{ .Type                 = DeviceEventType::kFailSafeTimerExpired,
```
bzbarsky-apple, line 88: How are these changes related to this PR?

src/app/clusters/bindings/bindings.cpp
```diff
@@ -259,9 +259,8 @@ CHIP_ERROR BindingTableAccess::WriteBindingTable(const ConcreteDataAttributePath
 
 CHIP_ERROR BindingTableAccess::NotifyBindingsChanged()
 {
-    DeviceLayer::ChipDeviceEvent event;
-    event.Type                        = DeviceLayer::DeviceEventType::kBindingsChangedViaCluster;
-    event.BindingsChanged.fabricIndex = mAccessingFabricIndex;
+    DeviceLayer::ChipDeviceEvent event{ .Type            = DeviceLayer::DeviceEventType::kBindingsChangedViaCluster,
```
bzbarsky-apple, line 262: How are these changes related to this PR?

src/include/platform/CHIPDeviceEvent.h
```diff
@@ -530,7 +530,6 @@ struct ChipDeviceEvent final
         } OtaStateChanged;
     };
 
-    void Clear() { memset(this, 0, sizeof(*this)); }
```
bzbarsky-apple, line 533: This seems unrelated?

bzbarsky-apple (approved):
Sorry for the lag here....

andy31415 (approved):

PR #32936, opened 2024-04-11
Improve MTRDevice work item logging.

bzbarsky-apple:
Stop including the node ID in the description passed to enqueueWorkItem, because the work queue includes that already.

Use hex for node IDs, cluster IDs, command/attribute IDs (but keep using decimal for endpoint IDs), to be consistent with other logging.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -1509,14 +1509,14 @@ static BOOL AttributeHasChangesOmittedQuality(MTRAttributePath * attributePath)
             while (readRequestsNext.count) {
                 // Can only read up to 9 paths at a time, per spec
                 if (readRequestsCurrent.count >= 9) {
-                    MTR_LOG_INFO("Batching read attribute work item [%llu]: cannot add more work, item is full [%@:%@:%@:%@]", workItemID, nodeID, endpointID, clusterID, attributeID);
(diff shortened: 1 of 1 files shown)


src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -1509,23 +1509,23 @@ static BOOL AttributeHasChangesOmittedQuality(MTRAttributePath * attributePath)
             while (readRequestsNext.count) {
                 // Can only read up to 9 paths at a time, per spec
                 if (readRequestsCurrent.count >= 9) {
-                    MTR_LOG_INFO("Batching read attribute work item [%llu]: cannot add more work, item is full [%@:%@:%@:%@]", workItemID, nodeID, endpointID, clusterID, attributeID);
+                    MTR_LOG_INFO("Batching read attribute work item [%llu]: cannot add more work, item is full [0x%016llX:%@:0x%llx:0x%llx]", workItemID, nodeID.unsignedLongLongValue, endpointID, clusterID.unsignedLongLongValue, attributeID.unsignedLongLongValue);
```
jtung-apple, line 1512: I thought we typically used `unsignedLongValue` for clusterID and attributeID, given they are uint32_t?

jtung-apple (approved):

src/darwin/Framework/CHIP/MTRDevice.mm
bzbarsky-apple: We could do that.  In practice it doesn't really matter too much and this way we know we are not dropping any information, but if you prefer I can do unsignedLongValue.

bzbarsky-apple:

PR #32948, opened 2024-04-11
Cancel the SRP timer if we have resolved on the SRP domain but the kD…

### LINKED ISSUE #32947: There is an issue with the DNS code where we start a timer to resolve on the SRP domain but do not check if more results are coming
We need to fix this issue by cancelling the timer when we get a resolve on srp and more results are coming.

nivi-apple:
…NSServiceFlagsMoreComing is set indicating there are more results coming

Fixes: #32947

### DIFF
--- a/src/platform/Darwin/DnssdContexts.cpp
+++ b/src/platform/Darwin/DnssdContexts.cpp
@@ -492,10 +492,7 @@ ResolveContext::ResolveContext(DiscoverNodeDelegate * delegate, chip::Inet::IPAd
 
 ResolveContext::~ResolveContext()
 {
-    if (isSRPTimerRunning)
-    {
-        CancelSRPTimer();
-    }
+    CancelSRPTimerIfRunning();
 }
 
 void ResolveContext::DispatchFailure(const char * errorStr, CHIP_ERROR err)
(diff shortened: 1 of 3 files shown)


src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -272,6 +272,12 @@ static void OnGetAddrInfo(DNSServiceRef sdRef, DNSServiceFlags flags, uint32_t i
 
     if (flags & kDNSServiceFlagsMoreComing)
     {
+        // If we have a timer running and we have resolved on the SRP domain but the kDNSServiceFlagsMoreComing
+        // is set, we need to cancel the timer so we don't pre-emptively call finalize().
+        if (sdCtx->isSRPTimerRunning && !sdCtx->shouldStartSRPTimerForResolve)
+        {
+            sdCtx->CancelSRPTimer();
```
bzbarsky-apple, line 279: This leaves isSRPTimerRunning set to true, yes?  We should probably have CancelSRPTimer() set it to false.

We should probably also move the `isSRPTimerRunning` check inside CancelSRPTimer, which should probably be called CancelSRPTimerIfRunning.

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -272,6 +272,12 @@ static void OnGetAddrInfo(DNSServiceRef sdRef, DNSServiceFlags flags, uint32_t i
 
     if (flags & kDNSServiceFlagsMoreComing)
     {
+        // If we have a timer running and we have resolved on the SRP domain but the kDNSServiceFlagsMoreComing
+        // is set, we need to cancel the timer so we don't pre-emptively call finalize().
```
bzbarsky-apple, line 276:
```suggestion
        // If we now don't need to have a timer while we wait for SRP results, ensure that there is no such
        // timer running.  Otherwise the timer could fire before we get the rest of the results that flags
        // say are coming, and trigger a finalize before we have all the data that is already available.
```

src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -647,10 +644,14 @@ void ResolveContext::SRPTimerExpiredCallback(chip::System::Layer * systemLayer,
...
 
-void ResolveContext::CancelSRPTimer()
+void ResolveContext::CancelSRPTimerIfRunning()
 {
-    DeviceLayer::SystemLayer().CancelTimer(SRPTimerExpiredCallback, static_cast<void *>(this));
-    ChipLogProgress(Discovery, "SRP resolve timer for %s cancelled; resolve timed out", instanceName.c_str());
+    if (isSRPTimerRunning)
+    {
+        DeviceLayer::SystemLayer().CancelTimer(SRPTimerExpiredCallback, static_cast<void *>(this));
+        ChipLogProgress(Discovery, "SRP resolve timer for %s cancelled; resolve timed out", instanceName.c_str());
```
bzbarsky-apple, line 652: This is no longer the right log message if we are canceling because we got results, right?

bzbarsky-apple (approved):

src/platform/Darwin/DnssdContexts.cpp
nivi-apple: sorry i didn't pay much attention. i will create a follow up PR.

nivi-apple: Filed #32955

PR #32949, opened 2024-04-11
Replace #if FOO with #if defined(FOO) && FOO

ghost:
Prevents warnings about using undefined macros

### DIFF
--- a/src/app/AttributePathParams.h
+++ b/src/app/AttributePathParams.h
@@ -25,7 +25,7 @@
 
 namespace chip {
 namespace app {
-#if CHIP_CONFIG_ENABLE_READ_CLIENT
+#if defined(CHIP_CONFIG_ENABLE_READ_CLIENT) && CHIP_CONFIG_ENABLE_READ_CLIENT
 class ReadClient;
 #endif // CHIP_CONFIG_ENABLE_READ_CLIENT
 struct AttributePathParams
--- a/src/lib/core/TLVReader.cpp
+++ b/src/lib/core/TLVReader.cpp
@@ -345,7 +345,7 @@ CHIP_ERROR TLVReader::Get(CharSpan & v) const
     }
 
     v = CharSpan(Uint8::to_const_char(bytes), len);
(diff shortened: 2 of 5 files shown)


src/lib/core/TLVReader.cpp
```diff
@@ -345,7 +345,7 @@ CHIP_ERROR TLVReader::Get(CharSpan & v) const
     }
 
     v = CharSpan(Uint8::to_const_char(bytes), len);
-#if CHIP_CONFIG_TLV_VALIDATE_CHAR_STRING_ON_READ
+#if defined(CHIP_CONFIG_TLV_VALIDATE_CHAR_STRING_ON_READ) && CHIP_CONFIG_TLV_VALIDATE_CHAR_STRING_ON_READ
```
andy31415, line 348: This one is defined by lib/core/CHIPConfig.h (actually buildconfig in gn).

We should never have it undefined. Not having this means that the chip config is old or not properly defined.

andy31415 (changes requested):
I believe we currently ensure these are specifically set - we are intentionally using `#if` to trigger compile errors in case something is not properly defined.

Rather than trying to use ifdef, we should figure out why these are undefined and fix that.

src/app/AttributePathParams.h
```diff
@@ -25,7 +25,7 @@
 
 namespace chip {
 namespace app {
-#if CHIP_CONFIG_ENABLE_READ_CLIENT
+#if defined(CHIP_CONFIG_ENABLE_READ_CLIENT) && CHIP_CONFIG_ENABLE_READ_CLIENT
```
bzbarsky-apple, line 28: If this is warning, that means that the right config headers are not being included, no?  The right fix is to include those headers.

bzbarsky-apple (changes requested):

bauerschwan (changes requested):
I would also highly vote against hiding these essential warnings. It would make the code prone to misconfigurations.

andy31415:
OLD PR with changes requested. Closing

PR #32952, opened 2024-04-11
More IWYU across app/ & src/

ghost:
Misc missing includes

### DIFF
--- a/src/protocols/bdx/DiagnosticLogs.h
+++ b/src/protocols/bdx/DiagnosticLogs.h
@@ -18,6 +18,8 @@
 
 #pragma once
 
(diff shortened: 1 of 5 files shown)


src/protocols/bdx/DiagnosticLogs.h
```diff
@@ -18,6 +18,8 @@
 
 #pragma once
 
+#include <stdint.h>
```
arkq, line 21: Since the project is C++ based it's better to use <c...> versions of includes (if available).

arkq (approved):

andy31415 (approved):

PR #32954, opened 2024-04-11
Separate out attribute access interface registry/cache, remove free functions from InteractionModelEngine.h

andy31415:
AttributeAccessInterface is very stand-alone and has no reason to be tied to ember directly (as its use is mostly inside the compatibility functions as a hook and the only callbacks ember needs is a cleanup).

### Changes

- create `AttributeAccessInterfaceRegistry.h/cpp`  to contain the old functions that existed in `attribute-storage.cpp`.
  - as a sideffect, mocks do not need to return nullptr for all attribute access interfaces
  - as a sideffect, dynamic server would in theory also support attributeaccessinterface
- moved the declaration of the `DispatchSingleClusterCommand` method into its own `src/app/util/IMClusterCommandHandler.h` to represent that we generate this code into `IMClusterCommandHandler.cpp`
- removed invalid/unused platform includes: it turns out platform code was including `attributeaccessinterface` but not using it ... and as a sideffect this made a `using chip::TLV` needing removal (that was also there due to copy and paste, but no TLV is used in platform)
- Fixed a few other platform includes that we transitively assumed due to AttributeAccessInterface (added some safecasts and cluster-enums includes).

### DIFF
--- a/src/app/AttributeAccessInterfaceRegistry.cpp
+++ b/src/app/AttributeAccessInterfaceRegistry.cpp
@@ -66,7 +66,7 @@ void unregisterAttributeAccessOverride(AttributeAccessInterface * attrOverride)
     UnregisterMatchingAttributeAccessInterfaces([attrOverride](AttributeAccessInterface * entry) { return entry == attrOverride; });
 }
 
-void unregisterAttributeAccessOverrideForEndpoint(EmberAfDefinedEndpoint * definedEndpoint)
+void unregisterAllAttributeAccessOverridesForEndpoint(EmberAfDefinedEndpoint * definedEndpoint)
 {
     UnregisterMatchingAttributeAccessInterfaces(
         [endpoint = definedEndpoint->endpoint](AttributeAccessInterface * entry) { return entry->MatchesEndpoint(endpoint); });
--- a/src/app/AttributeAccessInterfaceRegistry.cpp
+++ b/src/app/AttributeAccessInterfaceRegistry.cpp
@@ -1,0 +1,126 @@
+/**
+ *    Copyright (c) 2024 Project CHIP Authors
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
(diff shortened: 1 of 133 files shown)


src/app/AttributeAccessInterfaceRegistry.h
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Unregister an attribute access override (for example if the object
+ * implementing AttributeAccessInterface is being destroyed).
+ */
+void unregisterAttributeAccessOverride(chip::app::AttributeAccessInterface * attrOverride);
+
+/**
+ * Unregister all attribute access interfaces that match this given endpoint.
+ */
+void unregisterAttributeAccessOverrideForEndpoint(EmberAfDefinedEndpoint * definedEndpoint);
```
bzbarsky-apple, line 43:
```suggestion
void unregisterAllAttributeAccessOverridesForEndpoint(EmberAfDefinedEndpoint * definedEndpoint);
```

src/darwin/Framework/CHIP/ServerEndpoint/MTRServerCluster.mm
```diff
@@ -36,6 +36,7 @@
 #include <protocols/interaction_model/StatusCode.h>
 
 // TODO: These attribute-*.h bits are a hack that should eventually go away.
```
bzbarsky-apple, line 38:
```suggestion
// TODO: These attribute-*.h and AttributeAccessInterfaceRegistry.h bits are a hack that should eventually go away.
```

bzbarsky-apple (approved):

src/app/AttributeAccessInterfaceRegistry.cpp
```diff
@@ -0,0 +1,126 @@
...
+}
+
+bool registerAttributeAccessOverride(AttributeAccessInterface * attrOverride)
+{
+    gAttributeAccessInterfaceCache.Invalidate();
+    for (auto * cur = gAttributeAccessOverrides; cur; cur = cur->GetNext())
+    {
+        if (cur->Matches(*attrOverride))
+        {
+            ChipLogError(Zcl, "Duplicate attribute override registration failed");
```
yunhanw-google, line 82: for logging, we should consider to use DataManagement instead of zcl?

andy31415: Was copy & paste ... unsure. I believe it may have been `zcl` because it was residing in src/app/util . I will update (although will do that a bit later as CI restart is painful).

andy31415: Thinking maybe even InteractionModel ... I am generally unclear what each is supposed to cover.

andy31415: I see:

```
src/app/InteractionModelEngine.cpp:            ChipLogError(InteractionModel, "Duplicate command handler registration failed");
```

so for consistency since this is the same (just attribute instead of command) it seems InteractionModel is correct or at least consistent.

yunhanw-google (approved):

PR #32956, opened 2024-04-11
Remove the MTR_PER_CONTROLLER_STORAGE_ENABLED define.

bzbarsky-apple:
This is now enabled unconditionally.  Also allows us to remove the wrapper headers we had to allow us to use the per-controller storage bits internally.


ksperling-apple (approved):

ksperling-apple:

PR #32960, opened 2024-04-12
Darwin: MTRDeviceControllerFactory cleanup

ksperling-apple:
Internal methods don't need a named category
Use ivars directly (most internal properties were never used as properties)
Use MTR_DIRECT_MEMBERS for internals / implementation
No need to pre-declare internal methods used only with the same file

Reorder ivars a little to group related things

Allocate C++ objects that are created at init time directly as ivars instead of
as pointers. This avoids the need for manual cleanup code

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
+++ b/src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
@@ -84,33 +84,43 @@ static void ShutdownOnExit()
     [[MTRDeviceControllerFactory sharedInstance] stopControllerFactory];
 }
 
-@interface MTRDeviceControllerFactory () {
+@interface MTRDeviceControllerFactoryParams ()
+
+// Flag to keep track of whether our .storage is real consumer-provided storage
+// or just the fake thing we made up.
+@property (nonatomic, assign) BOOL hasStorage;
+
+@end
+
+MTR_DIRECT_MEMBERS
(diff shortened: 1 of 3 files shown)


src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
```diff
@@ -1012,9 +932,6 @@ - (void)resetOperationalAdvertising
         });
     }
 }
-@end
-
-@implementation MTRDeviceControllerFactory (InternalMethods)
```
bzbarsky-apple, line 1017: I think we were trying to indicate which things were not really part of the API... But maybe we should just move to using `_` prefixes for that?  Anyway, that's all followup bits.

bzbarsky-apple (approved):

src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
ksperling-apple: Hm yeah I'm not sure trying to keep all the internal methods in an actual category implementation is worth the hassle. Similarly, the underscore convention can become tedious if something changes from being internal to public means having to rename it. I think in practice it's often a case of just having to look which header (if any) a method is declared in (hopefully the IDE provides some quick way of jumping between implementation and declaration of a method -- Cmd-Click in Xcode).

ksperling-apple:

PR #32962, opened 2024-04-12
[nrfconnect] Implemented WiFiNetworkDiagnostics events generation

kkasperczyk-no:
Added generation of optional events from WiFiNetworkDiagnostics cluster.

### DIFF
--- a/src/platform/nrfconnect/wifi/WiFiManager.cpp
+++ b/src/platform/nrfconnect/wifi/WiFiManager.cpp
@@ -25,6 +25,7 @@
 #include <crypto/RandUtils.h>
 #include <lib/support/logging/CHIPLogging.h>
 #include <platform/CHIPDeviceLayer.h>
+#include <platform/DiagnosticDataProvider.h>
 #include <platform/Zephyr/InetUtils.h>
 
 #include <zephyr/kernel.h>
@@ -151,7 +152,7 @@ void WiFiManager::WifiMgmtEventHandler(net_mgmt_event_callback * cb, uint32_t mg
         Platform::UniquePtr<uint8_t> eventData(new uint8_t[cb->info_length]);
         VerifyOrReturn(eventData);
         memcpy(eventData.get(), cb->info, cb->info_length);
-        sEventHandlerMap[mgmtEvent](std::move(eventData));
+        sEventHandlerMap[mgmtEvent](std::move(eventData), cb->info_length);
     }
 }
 
@@ -292,8 +293,11 @@ CHIP_ERROR WiFiManager::GetNetworkStatistics(NetworkStatistics & stats) const
     return CHIP_NO_ERROR;
 }
 
-void WiFiManager::ScanResultHandler(Platform::UniquePtr<uint8_t> data)
+void WiFiManager::ScanResultHandler(Platform::UniquePtr<uint8_t> data, size_t length)
 {
+    // Validate that input data size matches the expected one.
+    VerifyOrReturn(length == sizeof(wifi_scan_result));
+
     // Contrary to other handlers, offload accumulating of the scan results from the CHIP thread to the caller's thread
     const wifi_scan_result * scanResult = reinterpret_cast<const wifi_scan_result *>(data.get());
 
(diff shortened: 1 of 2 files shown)


src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -432,6 +433,36 @@ void WiFiManager::ConnectHandler(Platform::UniquePtr<uint8_t> data)
...
+            WiFiDiagnosticsDelegate * delegate = GetDiagnosticDataProvider().GetWiFiDiagnosticsDelegate();
+            if (delegate)
+            {
+                uint16_t reason = Instance().GetLastDisconnectReason();
+                uint8_t associationFailureCause;
+
+                switch (connStatus)
+                {
+                case WIFI_STATUS_CONN_WRONG_PASSWORD:
+                    associationFailureCause = chip::to_underlying(
```
bzbarsky-apple, line 446: Isn't this all inside `namespace chip`?  Why all the `chip::` prefixes here?

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -432,6 +433,36 @@ void WiFiManager::ConnectHandler(Platform::UniquePtr<uint8_t> data)
...
+                        chip::app::Clusters::WiFiNetworkDiagnostics::AssociationFailureCauseEnum::kAuthenticationFailed);
```
bzbarsky-apple, line 447: Might be worth it to have a `using app::Clusters::WiFiNetworkDiagnostics::AssociationFailureCauseEnum` at the top of this function so these bits can be shorter.

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -469,13 +507,52 @@ void WiFiManager::ConnectHandler(Platform::UniquePtr<uint8_t> data)
     }
 }
 
-void WiFiManager::DisconnectHandler(Platform::UniquePtr<uint8_t>)
+void WiFiManager::DisconnectHandler(Platform::UniquePtr<uint8_t> data)
 {
-    SystemLayer().ScheduleLambda([] {
+    CHIP_ERROR err = SystemLayer().ScheduleLambda([capturedData = data.get()] {
```
bzbarsky-apple, line 512: We can't just capture the UniquePtr by move here?

src/platform/nrfconnect/wifi/WiFiManager.cpp
```diff
@@ -469,13 +507,52 @@ void WiFiManager::ConnectHandler(Platform::UniquePtr<uint8_t> data)
...
+        Platform::UniquePtr<uint8_t> safePtr(capturedData);
+        uint8_t * rawData          = safePtr.get();
+        const wifi_status * status = reinterpret_cast<const wifi_status *>(rawData);
```
bzbarsky-apple, line 515: This seems very fragile, in that we have no checks to make sure the data size is correct, etc..... And nothing even documents what the expected data type here is.

As a followup: if the input to this stuff were at least ScopedMemoryBufferWithSize then you could check the sizing up front (outside the lambda)....

bzbarsky-apple (approved):

src/platform/nrfconnect/wifi/WiFiManager.cpp
Damian-Nordic: Our lambdas must be trivially copyable since they are passed via `ChipDeviceEvent`, which prevents capturing `UniquePtr`, right?

bzbarsky-apple: Ah, ok, makes sense.

yunhanw-google (approved):

src/platform/nrfconnect/wifi/WiFiManager.cpp
kkasperczyk-no: Fair point, thanks for the catch. I passed data size to the handlers API and added checks validating if size matches the expectation.

PR #32970, opened 2024-04-12
Use RAII for group session iteration

andy31415:
This is to ensure `Release()` is always called on the iterator, even in case of early returns.

### DIFF
--- a/src/transport/SessionManager.cpp
+++ b/src/transport/SessionManager.cpp
@@ -58,6 +58,31 @@ using Transport::SecureSession;
 namespace {
 Global<GroupPeerTable> gGroupPeerTable;
 
+/// RAII class for iterators that guarantees that Release() will be called
+/// on the underlying type
+template <typename ReleasableIterator>
+class AutoReleaseGroupIterator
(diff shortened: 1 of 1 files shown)


src/transport/SessionManager.cpp
```diff
@@ -58,6 +58,31 @@ using Transport::SecureSession;
...
+
+    void Release()
+    {
+        VerifyOrReturn(mIter != nullptr);
+        mIter->Release();
+        mIter = nullptr;
+    }
+
+private:
+    Releasable * mIter;
```
yunhanw-google, line 83: init with nullptr

yunhanw-google (approved):

bzbarsky-apple (approved):

src/transport/SessionManager.cpp
andy31415: Updated, however existing constructor initialises this member, so I think the nullptr bit does not do anything. Probably useful only in case the class changes in the future in some way to allow a default constructor or if we add more constructors.

PR #32972, opened 2024-04-12
DM XML scrape for 1.3

cecille:
Scrape was taken at the 1.3 spec tag.


data_model/clusters/BooleanStateConfiguration.xml
```diff
@@ -104,31 +104,31 @@ Davis, CA 95616, USA
     </bitmap>
   </dataTypes>
   <attributes>
-    <attribute id="0x0000" name="CurrentSensitivityLevel" type="uint8">
+    <attribute id="0x0000" name="CurrentSensitivityLevel" type="uint8" default="MS">
```
robszewczyk, line 107: It is super weird to have a Manufacturer Specific default as an item, but I guess that's the scraper: I would expect that this could lead to unexpected results -- e.g. someone specifies a string MS as a default value for some cluster somewhere and that leads to unexpected results.  A Manufacturer Specific default is the same as no default, so the preferred option would be to omit.

robszewczyk (approved):

data_model/clusters/BooleanStateConfiguration.xml
robszewczyk: Non blocking comment to be clear; in actuality the right call is to remove that notation from the spec.

andy31415 (approved):

data_model/clusters/BooleanStateConfiguration.xml
ReneJosefsen: Keep in mind that the default here is for client, to use in case the element is not present. 

My assumption and understanding, since this predates my knowledge, is that the intention here is to indicate that it is up to the client to determine how to address this not being present, hence MS but from the client perspective.

PR #32974, opened 2024-04-12
Fix crash if commissioning is canceled while waiting on an attestation delegate.

bzbarsky-apple:
When we extend the fail-safe after device attestation before calling into the delegate, we ended up with a dangling mInvokeCancelFn after the command finished and until CommissioningStageComplete().  That last would not happen until our delegate called back into us to continue commissioning.

If an attempt was made to cancel commissioning during that time interval, we would end up crashing.


woody-apple (approved):

woody-apple:
This is causing crashes when pairing in CSG tests, marking hotfix

PR #32976, opened 2024-04-12
Splitting example and core workflows

woody-apple:
This example workflow seems to fail a lot, but also more importantly should be separated out as an example, not the core build

### DIFF
--- a/.github/workflows/darwin.yaml
+++ b/.github/workflows/darwin.yaml
@@ -17,13 +17,16 @@ name: Darwin
 on:
     push:
         branches-ignore:
-            - 'dependabot/**'
+            - "dependabot/**"
     pull_request:
     merge_group:
     workflow_dispatch:
 
 concurrency:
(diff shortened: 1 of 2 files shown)


.github/workflows/darwin.yaml
```diff
@@ -51,30 +54,33 @@ jobs:
...
+        needs: [framework] # serialize to avoid running to many parallel macos runners
         runs-on: macos-13
         strategy:
             matrix:
                 options: # We don't need a full matrix
                     - flavor: asan
-                      arguments: -enableAddressSanitizer YES -enableUndefinedBehaviorSanitizer YES
+                      arguments:
+                          -enableAddressSanitizer YES
+                          -enableUndefinedBehaviorSanitizer YES
```
bzbarsky-apple, line 83: This reformat is odd.  It's making it look like a YAML array, but it's just a string, no?

PR #32984, opened 2024-04-15
Split out `AttributeAccessInterface.h` into separate headers/cpp for each underlying class

andy31415:
### Changes

- separate out `AttributeValueEncoder`, `AttributeValueDecoder` and `AttributeReportBuilder`
- Update includes for files that only generally want the encoder and not the entire attribute access interface


src/app/AttributeReportBuilder.h
```diff
@@ -0,0 +1,79 @@
+/*
+ *
+ *    Copyright (c) 2021-2022 Project CHIP Authors
```
bzbarsky-apple, line 3: Why is the .cpp copyright 2024 but this is 2021-2022?  Seems like they should have more or less consistent copyright dates (probably whatever year this bit was added to 2024).

src/app/AttributeValueDecoder.h
```diff
@@ -0,0 +1,74 @@
+/*
+ *    Copyright (c) 2024 Project CHIP Authors
```
bzbarsky-apple, line 2: Should be `N-2024` where N is the year this class was added, I expect.

src/app/AttributeValueEncoder.cpp
```diff
@@ -1,6 +1,5 @@
 /*
- *
- *    Copyright (c) 2021 Project CHIP Authors
+ *    Copyright (c) 2024 Project CHIP Authors
```
bzbarsky-apple, line 2: Should be `N-2024`?

src/app/AttributeValueEncoder.h
```diff
@@ -0,0 +1,286 @@
+/*
+ *    Copyright (c) 2024 Project CHIP Authors
```
bzbarsky-apple, line 2: See comments above.

bzbarsky-apple (approved):
Thank you for doing this!  Makes things much more readable, if nothing else.

yunhanw-google (approved):

yunhanw-google:
Thanks for this split for better readability!

PR #32987, opened 2024-04-15
[Silabs] Add adv change event to the silabs platforms

mkardous-silabs:
#### Description
PR adds the support for the `kCHIPoBLEAdvertisingChange` to the silabs platform.

- Event is published when we start and stop advertising. Restarting when we are already advertising does not publish the event.
- Add the event to the EFR32 BleManager
- Add the event to the RS91x Ble Manater

#### Tests 
- Manual test to make sure the device posts the event when it is booted when uncommissionned - Start event
- Manual test to make sure the device posts the event when commissionning starts - Stop event
- Manual test to make sure the device posts the event when Ble Adv time out is triggered - Stop event

### DIFF
--- a/src/platform/silabs/efr32/BLEManagerImpl.cpp
+++ b/src/platform/silabs/efr32/BLEManagerImpl.cpp
@@ -532,11 +532,12 @@ exit:
 
 CHIP_ERROR BLEManagerImpl::StartAdvertising(void)
 {
-    CHIP_ERROR err;
-    sl_status_t ret;
-    uint32_t interval_min;
-    uint32_t interval_max;
+    CHIP_ERROR err           = CHIP_NO_ERROR;
+    sl_status_t ret          = SL_STATUS_OK;
+    uint32_t interval_min    = 0;
+    uint32_t interval_max    = 0;
     uint16_t numConnectionss = NumConnections();
+    bool postAdvChangeEvent  = false;
     uint8_t connectableAdv =
         (numConnectionss < kMaxConnections) ? sl_bt_advertiser_connectable_scannable : sl_bt_advertiser_scannable_non_connectable;
 
@@ -548,6 +549,7 @@ CHIP_ERROR BLEManagerImpl::StartAdvertising(void)
     else
     {
         ChipLogDetail(DeviceLayer, "Start BLE advertisement");
+        postAdvChangeEvent = true;
     }
 
     err = ConfigureAdvertisingData();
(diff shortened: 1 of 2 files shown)


src/platform/silabs/efr32/BLEManagerImpl.cpp
```diff
@@ -606,23 +616,29 @@ CHIP_ERROR BLEManagerImpl::StartAdvertising(void)
 CHIP_ERROR BLEManagerImpl::StopAdvertising(void)
 {
     CHIP_ERROR err = CHIP_NO_ERROR;
-    sl_status_t ret;
 
     if (mFlags.Has(Flags::kAdvertising))
     {
+        sl_status_t ret = SL_STATUS_OK;
```
jmartinez-silabs, line 622: i don't see `ret` used.

jmartinez-silabs (approved):

src/platform/silabs/efr32/BLEManagerImpl.cpp
mkardous-silabs: Varialbe is used at line 627 which does show in the PR diff. Variable was moved from the top of the function to inside the if since it is only used there.

src/platform/silabs/efr32/BLEManagerImpl.cpp
```diff
@@ -586,18 +588,26 @@ CHIP_ERROR BLEManagerImpl::StartAdvertising(void)
...
+        StartBleAdvTimeoutTimer(CHIP_DEVICE_CONFIG_BLE_ADVERTISING_INTERVAL_CHANGE_TIME);
     }
+    mFlags.Set(Flags::kAdvertising);
 
-    err = MapBLEError(ret);
+    if (postAdvChangeEvent)
+    {
+        // Post CHIPoBLEAdvertisingChange event.
+        ChipDeviceEvent advChange;
+        advChange.Type                             = DeviceEventType::kCHIPoBLEAdvertisingChange;
```
cecille, line 606: Change looks reasonable, but ... isn't CHIPoBLE a BLE operational network? We don't support that. So is this just a weirdly named event?

mkardous-silabs: I guess so... Event was named 4 years ago based on the blame. I can create a follow up ticket to remove the `CHIPo` part from the name event name.

cecille: okie dokie.

cecille (approved):

yunhanw-google (approved):

PR #32990, opened 2024-04-15
Stop doing read-throughs on MTRDevice when subscription is being established.

bzbarsky-apple:
The key goal is to stop doing read-throughs when we expect to start getting data soon: between sending the Sigma1 for our initial subscribe and getting the first priming read data packets.  Any reads dispatched during that time "because we don't have a subscription yet" are just wasted network traffic, because chances are we're about to have a subscription and get that data as part of the priming read anyway.

The main fix is just to stop doing read-throughs altogether for any attributes known to not be C and if we have a subscription delegate.  In that state we should be subscribed and should be getting updates for attribute values, and there's no reason to read through.

The rest of the changes are related to a side-effect of the stopping of read-throughs: if we had a subscription drop and have backed off quite a bit, a read-through that succeeded could prod us to retry subscribing.  We don't want to do a read just to (maybe) trigger that behavior, though.  Instead, treat a skipped read-through as a direct signal to retry our subscription attempt.  But not too often.

This "try to resubscribe" heuristic can probably use some tweaking.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -28,6 +28,7 @@
 #import "MTRCommandTimedCheck.h"
 #import "MTRConversion.h"
 #import "MTRDefines_Internal.h"
+#import "MTRDeviceControllerOverXPC.h"
 #import "MTRDeviceController_Internal.h"
 #import "MTRDevice_Internal.h"
 #import "MTRError_Internal.h"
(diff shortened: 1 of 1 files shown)


yunhanw-google (approved):

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -651,13 +664,26 @@ - (void)invalidate
...
     MTR_LOG_DEFAULT("%@ saw new operational advertisement", self);
 
+    [self _triggerResubscribeWithReason:"operational advertisement seen"
+                    nodeLikelyReachable:YES];
+}
+
+// Trigger a resubscribe as needed.  nodeLikelyReachable should be YES if we
+// have reason to suspect the node is now reachable, NO if we have no idea
+// whether it might be.
+- (void)_triggerResubscribeWithReason:(const char *)reason nodeLikelyReachable:(BOOL)nodeLikelyReachable
```
jtung-apple, line 678: If `nodeLikelyReachable` should this also reset `_lastSubscriptionAttemptWait` to 0?

bzbarsky-apple: @jtung-apple  That might be reasonable, yes, but I'd like to do that in a separate PR, since it's a change to existing behavior....

jtung-apple (approved):

bzbarsky-apple:

PR #32993, opened 2024-04-15
Fix LevelControl Move when no motion requested (#32539)

cecille:
* Fix LevelControl Move when no motion requested

- rate == 0 means "do not move", so handle it more efficiently without moving.

Testing done:
- Added an integration test for the behavior.

* Restyled by prettier-yaml

### DIFF
--- a/src/app/clusters/level-control/level-control.cpp
+++ b/src/app/clusters/level-control/level-control.cpp
@@ -18,6 +18,8 @@
 // clusters specific header
 #include "level-control.h"
 
+#include <algorithm>
+
 // this file contains all the common includes for clusters in the util
 #include <app-common/zap-generated/attributes/Accessors.h>
 #include <app-common/zap-generated/cluster-objects.h>
@@ -904,7 +906,7 @@ static Status moveToLevelHandler(EndpointId endpoint, CommandId commandId, uint8
 
     // The duration between events will be the transition time divided by the
     // distance we must move.
-    state->eventDurationMs = state->transitionTimeMs / actualStepSize;
+    state->eventDurationMs = state->transitionTimeMs / std::max(static_cast<uint8_t>(1u), actualStepSize);
     state->elapsedTimeMs   = 0;
 
     state->storedLevel = storedLevel;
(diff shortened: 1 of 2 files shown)


andy31415 (approved):

src/app/clusters/level-control/level-control.cpp
```diff
@@ -958,6 +960,14 @@ static void moveHandler(app::CommandHandler * commandObj, const app::ConcreteCom
         goto send_default_response;
     }
 
+    if (!rate.IsNull() && (rate.Value() == 0))
```
Apollon77, line 963: @cecille Even if merged already ... is this really the right place?

There is another second case for a comparable check for the defaultRate fallback which is way later in the code flow https://github.com/cecille/connectedhomeip/blob/97d435b8a498fc935f04f16743c657f2b660f04b/src/app/clusters/level-control/level-control.cpp#L1042 ... when I understood the code right now "rate===0" would not cause "withOnOff" cases to turn on/off with this logic, but a defaultRate === 0 would ... so that's a different behavior for "Kind of the same". Was that intended?

cecille: @tennessee-google

jmartinez-silabs: @Apollon77 I believe you are right. Both rate =0 or DefaultRate = 0 when rate is Null should lead to no action taken.

I think a small correction would be to evaluate the rate is Null and DefaultRate > 0 earlier in the handler

cecille: issue here - https://github.com/CHIP-Specifications/connectedhomeip-spec/issues/9125

Apollon77: @jmartinez-silabs Thank you for your check. One question for detail: "No action taken at all"? or "Do nothing but process potential onoff logic"?
Case level  is 50% and onoff=off ... should be a call here leave onof===off or turn on as onoff cluster dependency logic would require it.
I ask because there are some other cases where - even without any action taken - the onoff dependecy logic is executed ... and this is also included in tests ...

PR #32997, opened 2024-04-16
Darwin: MTRDeviceControllerFactory improvements

ksperling-apple:
Handle cleanup after failed factory start in the same dispatch_sync
Also remove some unnecessary OOM handling at the framework level, treat allocation failure as fatal.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
+++ b/src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
@@ -330,15 +330,12 @@ MTR_DIRECT_MEMBERS
         InitializeServerAccessControl();
 
         if (startupParams.hasStorage) {
-            _persistentStorageDelegate = new (std::nothrow) MTRPersistentStorageDelegateBridge(startupParams.storage);
-            VerifyOrExit(_persistentStorageDelegate != nullptr, err = CHIP_ERROR_NO_MEMORY);
+            _persistentStorageDelegate = new MTRPersistentStorageDelegateBridge(startupParams.storage);
             _sessionResumptionStorage = nullptr;
             _usingPerControllerStorage = NO;
         } else {
-            _persistentStorageDelegate = new (std::nothrow) MTRDemuxingStorage(self);
-            VerifyOrExit(_persistentStorageDelegate != nullptr, err = CHIP_ERROR_NO_MEMORY);
-            _sessionResumptionStorage = new (std::nothrow) MTRSessionResumptionStorageBridge(self);
-            VerifyOrExit(_sessionResumptionStorage != nullptr, err = CHIP_ERROR_NO_MEMORY);
+            _persistentStorageDelegate = new MTRDemuxingStorage(self);
+            _sessionResumptionStorage = new MTRSessionResumptionStorageBridge(self);
             _usingPerControllerStorage = YES;
         }
 
@@ -350,12 +347,10 @@ MTR_DIRECT_MEMBERS
 
         // TODO: Allow passing a different keystore implementation via startupParams.
         _keystore = new PersistentStorageOperationalKeystore();
-        VerifyOrExit(_keystore != nullptr, err = CHIP_ERROR_NO_MEMORY);
         SuccessOrExit((err = _keystore->Init(_persistentStorageDelegate)));
 
         // TODO Allow passing a different opcert store implementation via startupParams.
         _opCertStore = new Credentials::PersistentStorageOpCertStore();
-        VerifyOrExit(_opCertStore != nullptr, err = CHIP_ERROR_NO_MEMORY);
         SuccessOrExit((err = _opCertStore->Init(_persistentStorageDelegate)));
 
         _productAttestationAuthorityCertificates = [startupParams.productAttestationAuthorityCertificates copy];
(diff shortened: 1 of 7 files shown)


src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
```diff
@@ -452,47 +395,44 @@ - (BOOL)_startControllerFactory:(MTRDeviceControllerFactoryParams *)startupParam
...
     }
-
     return YES;
 }
 
 - (void)stopControllerFactory
 {
     [self _assertCurrentQueueIsNotMatterQueue];
 
-    if (![self isRunning]) {
```
woody-apple, line 476: Is this safe to remove?

src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
```diff
@@ -907,12 +838,6 @@ - (MTRDeviceController * _Nullable)maybeInitializeOTAProvider:(MTRDeviceControll
 
 - (void)resetOperationalAdvertising
 {
-    if (![self checkIsRunning:nil]) {
```
woody-apple, line 910: Is this safe to remove? Maybe it is...

src/darwin/Framework/CHIP/MTROperationalCredentialsDelegate.mm
```diff
@@ -65,15 +65,8 @@
     // Make copies of the certificates, just in case the API consumer
     // has them as MutableData.
     mRootCert = [NSData dataWithData:rootCert];
-    if (mRootCert == nil) {
```
woody-apple, line 68: It seems like this can still fail, but agree CHIP_ERROR_NO_MEMORY is the wrong error, should this be invalid data or something?

src/darwin/Framework/CHIP/MTROperationalCredentialsDelegate.mm
```diff
@@ -65,15 +65,8 @@
...
-        return CHIP_ERROR_NO_MEMORY;
-    }
-
     if (icaCert != nil) {
         mIntermediateCert = [NSData dataWithData:icaCert];
-        if (mIntermediateCert == nil) {
```
woody-apple, line 74: Same as above

src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
```diff
@@ -383,63 +347,42 @@ - (BOOL)_startControllerFactory:(MTRDeviceControllerFactoryParams *)startupParam
...
-            int ret = atexit(ShutdownOnExit);
-            if (ret != 0) {
-                MTR_LOG_ERROR("Error registering exit handler: %d", ret);
-                return;
+            if (atexit(ShutdownOnExit) != 0) {
+                char error[128];
+                strerror_r(errno, error, sizeof(error));
+                MTR_LOG_ERROR("Warning: Failed to register atexit handler: %s", error);
             }
+            sExitHandlerRegistered = true;
```
bzbarsky-apple, line 385: We never set that to true?  Go me...

src/darwin/Framework/CHIP/MTROperationalCredentialsDelegate.mm
bzbarsky-apple: It's not really "invalid data".  It's "we tried to copy the NSData and failed", which really is just OOM, I woudl think.

(And this can probably use `[rootCert copy]` as far as that goes.)

src/darwin/Framework/CHIP/MTROperationalCredentialsDelegate.mm
```diff
@@ -65,15 +65,8 @@
     // Make copies of the certificates, just in case the API consumer
     // has them as MutableData.
     mRootCert = [NSData dataWithData:rootCert];
-    if (mRootCert == nil) {
-        return CHIP_ERROR_NO_MEMORY;
-    }
-
     if (icaCert != nil) {
         mIntermediateCert = [NSData dataWithData:icaCert];
```
bzbarsky-apple, line 69: This could likely become `mIntermediateCert = [icaCert copy];` without even nil-checking icaCert....

src/darwin/Framework/CHIPTests/MTRControllerTests.m
```diff
@@ -62,6 +62,10 @@ - (void)testFactoryLifecycle
     XCTAssertTrue([factory startControllerFactory:factoryParams error:nil]);
     XCTAssertTrue([factory isRunning]);
 
+    // Starting again with identical params is a no-op
```
bzbarsky-apple, line 65: It's a no-op even with different params, no?

bzbarsky-apple (approved):

src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
ksperling-apple: I'm removing a couple of the double checks if favor of checking just within the dispatch_sync. We want to work towards having the factory API be fully thread-safe (currently it assumes some level of external synchronization for some operations.)

src/darwin/Framework/CHIP/MTROperationalCredentialsDelegate.mm
ksperling-apple: Yeah `copy` would work too (though technically someone could pass some sort of custom sub-class that overrides `copy`, so I stuck with the existing `dataWithData:`)

ksperling-apple: The return type of `dataWithData:` is declared as non null, so this is just treating it as such.

src/darwin/Framework/CHIP/MTROperationalCredentialsDelegate.mm
ksperling-apple: Yes, I can do those... we probably don't need to worry about the hypothetically overridden `copy` method

src/darwin/Framework/CHIPTests/MTRControllerTests.m
ksperling-apple: At the moment it is, but I think we may want to make that return an INVALID_STATE error instead?

src/darwin/Framework/CHIP/MTROperationalCredentialsDelegate.mm
ksperling-apple: done

src/darwin/Framework/CHIPTests/MTRControllerTests.m
bzbarsky-apple: Yeah, that's fair.

ksperling-apple:

PR #32998, opened 2024-04-16
DeviceControllerFactory: Add InitAndRetainSystemState

ksperling-apple:
This does a RetainSystemState, but re-initializes if necessary.

Also rename no-arg InitSystemState() to ReinitSystemStateIfNecessary() and check for the case where no work is necessary before creating the FactoryInitParams and calling InitSystemState(params).

This means InitSystemState(params) doesn't have to handle that case anymore since it's only called from Init() or after ReinitSystemStateIfNecessary() has already done the check.

### DIFF
--- a/src/controller/CHIPDeviceControllerFactory.h
+++ b/src/controller/CHIPDeviceControllerFactory.h
@@ -210,7 +210,7 @@ public:
     bool ReleaseSystemState();
 
     // Like RetainSystemState(), but will re-initialize the system state first if necessary.
-    CHIP_ERROR InitAndRetainSystemState();
+    CHIP_ERROR EnsureAndRetainSystemState();
 
     //
     // Retrieve a read-only pointer to the system state object that contains pointers to key stack
--- a/src/controller/CHIPDeviceControllerFactory.h
+++ b/src/controller/CHIPDeviceControllerFactory.h
@@ -209,6 +209,9 @@ public:
     //
     bool ReleaseSystemState();
 
+    // Like RetainSystemState(), but will re-initialize the system state first if necessary.
+    CHIP_ERROR InitAndRetainSystemState();
+
     //
     // Retrieve a read-only pointer to the system state object that contains pointers to key stack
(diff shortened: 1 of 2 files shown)


saurabhst (approved):

src/controller/CHIPDeviceControllerFactory.h
```diff
@@ -209,6 +209,9 @@ class DeviceControllerFactory
     //
     bool ReleaseSystemState();
 
+    // Like RetainSystemState(), but will re-initialize the system state first if necessary.
```
bzbarsky-apple, line 212: It's "necessary" if it's been shut down, right?

Should this be called something like EnsureRetainedSystemState()?  Feels like the important part here is that once this function returns success we have a "live" system state that has been retained, right?

And separately, should we have a `EnsureSystemState()` which does the "init if necessary" bits and then consumers just call retain separately?  It's not 100% clear why we are rolling both operations into one function.  Maybe more comments explaining how this should be used would help?

ksperling-apple: I'm mainly adding this because I didn't want to change the behavior of the existing `RetainSystemState` to potentially re-initialize. `EnsureRetainedSystemState` is maybe a bit ambiguous... if the retain count is already > 0 then the system state is already retained and it sounds like it would then do nothing?

We have the equivalent of `EnsureSystemState` internally, but I don't think we want to publicly expose the ability to do that without also taking a retain count at the same time -- after all the retain count is what is meant to manage the lifecycle of the system state, and the `+0` state of being initialized but not retained yet is something that only exists transiently while bringing up a controller.

jmartinez-silabs (approved):

src/controller/CHIPDeviceControllerFactory.h
bzbarsky-apple: That makes sense.

bzbarsky-apple (approved):

PR #31776, opened 2024-01-30
Adjust storage and processing in SystemPacketBuffer and associated code for large payloads.

pidarped:
Extend uint16_t type variables to uint32_t wherever applicable.

Fixes #31115

PR #33001, opened 2024-04-16
Bring PR #31776(Adjust storage for large payloads) from tcp branch to master

pidarped:
This PR is to bring PR#31776(in tcp branch) to master with the later review comments addressed.

### DIFF
--- a/src/transport/SecureMessageCodec.cpp
+++ b/src/transport/SecureMessageCodec.cpp
@@ -43,13 +43,10 @@ CHIP_ERROR Encrypt(const CryptoContext & context, CryptoContext::ConstNonceView 
     VerifyOrReturnError(!msgBuf->HasChainedBuffer(), CHIP_ERROR_INVALID_MESSAGE_LENGTH);
     VerifyOrReturnError(msgBuf->TotalLength() <= kMaxAppMessageLen, CHIP_ERROR_MESSAGE_TOO_LONG);
 
-    static_assert(std::is_same<decltype(msgBuf->TotalLength()), size_t>::value,
-                  "Addition to generate payloadLength might overflow");
-
     ReturnErrorOnFailure(payloadHeader.EncodeBeforeData(msgBuf));
 
-    uint8_t * data    = msgBuf->Start();
-    uint32_t totalLen = static_cast<uint32_t>(msgBuf->TotalLength());
+    uint8_t * data  = msgBuf->Start();
+    size_t totalLen = msgBuf->TotalLength();
 
     MessageAuthenticationCode mac;
     ReturnErrorOnFailure(context.Encrypt(data, totalLen, data, nonce, packetHeader, mac));
@@ -57,8 +54,7 @@ CHIP_ERROR Encrypt(const CryptoContext & context, CryptoContext::ConstNonceView 
     uint16_t taglen = 0;
     ReturnErrorOnFailure(mac.Encode(packetHeader, &data[totalLen], msgBuf->AvailableDataLength(), &taglen));
 
-    VerifyOrReturnError(CanCastTo<size_t>(totalLen + taglen), CHIP_ERROR_INTERNAL);
-    msgBuf->SetDataLength(static_cast<size_t>(totalLen + taglen));
+    msgBuf->SetDataLength(totalLen + taglen);
 
     return CHIP_NO_ERROR;
 }
@@ -69,7 +65,7 @@ CHIP_ERROR Decrypt(const CryptoContext & context, CryptoContext::ConstNonceView 
     ReturnErrorCodeIf(msg.IsNull(), CHIP_ERROR_INVALID_ARGUMENT);
 
     uint8_t * data = msg->Start();
-    uint32_t len   = static_cast<uint32_t>(msg->DataLength());
+    size_t len     = msg->DataLength();
 
     PacketBufferHandle origMsg;
 #if CHIP_SYSTEM_CONFIG_USE_LWIP
@@ -89,7 +85,7 @@ CHIP_ERROR Decrypt(const CryptoContext & context, CryptoContext::ConstNonceView 
     ReturnErrorOnFailure(mac.Decode(packetHeader, &data[len - footerLen], footerLen, &taglen));
     VerifyOrReturnError(taglen == footerLen, CHIP_ERROR_INTERNAL);
 
-    len = static_cast<uint16_t>(len - taglen);
+    len = len - taglen;
     msg->SetDataLength(len);
 
     uint8_t * plainText = msg->Start();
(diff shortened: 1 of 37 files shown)


src/transport/SecureMessageCodec.cpp
```diff
@@ -43,22 +43,27 @@ CHIP_ERROR Encrypt(const CryptoContext & context, CryptoContext::ConstNonceView
     VerifyOrReturnError(!msgBuf->HasChainedBuffer(), CHIP_ERROR_INVALID_MESSAGE_LENGTH);
     VerifyOrReturnError(msgBuf->TotalLength() <= kMaxAppMessageLen, CHIP_ERROR_MESSAGE_TOO_LONG);
 
-    static_assert(std::is_same<decltype(msgBuf->TotalLength()), uint16_t>::value,
+    static_assert(std::is_same<decltype(msgBuf->TotalLength()), size_t>::value,
                   "Addition to generate payloadLength might overflow");
```
bzbarsky-apple, line 47: This static assert makes no sense.  It _used_ to make sense because if TotalLength() is 16-bit and tagLen is 16-bit then the addition promotes to 32-bit signed int on 32-bit-or-more systems and can't possibly overflow.

But in the new setup, what does TotalLength being a size_t tell us about overflow?  Absolutely nothing.

src/transport/SecureMessageCodec.cpp
```diff
@@ -43,22 +43,27 @@ CHIP_ERROR Encrypt(const CryptoContext & context, CryptoContext::ConstNonceView
...
 
     MessageAuthenticationCode mac;
     ReturnErrorOnFailure(context.Encrypt(data, totalLen, data, nonce, packetHeader, mac));
 
     uint16_t taglen = 0;
     ReturnErrorOnFailure(mac.Encode(packetHeader, &data[totalLen], msgBuf->AvailableDataLength(), &taglen));
 
-    VerifyOrReturnError(CanCastTo<uint16_t>(totalLen + taglen), CHIP_ERROR_INTERNAL);
-    msgBuf->SetDataLength(static_cast<uint16_t>(totalLen + taglen));
+    VerifyOrReturnError(CanCastTo<size_t>(totalLen + taglen), CHIP_ERROR_INTERNAL);
```
bzbarsky-apple, line 60: This check is just checking for `totalLen + taglen >= 0`, effectively, right?  But won't that always be the case?  It's an unsigned value.

The overflow checks here are broken and need to be fixed.  I brought this up in https://github.com/project-chip/connectedhomeip/pull/31776#discussion_r1509682211 too, and unlike the other comments on that PR it doesn't look like that was addressed.

bzbarsky-apple (changes requested):

src/transport/SecureMessageCodec.cpp
pidarped: This check should have been removed because I now have bound checks for totalLen + taglen below. There were a few bits here and there that I needed to change and was doing another pass today. Had pushed the PR to check/fix builds on all platforms for CI.

src/transport/SecureMessageCodec.cpp
pidarped: Agreed that the static_assert does not make sense anymore. Will remove it.

src/transport/SecureMessageCodec.cpp
```diff
@@ -43,22 +43,23 @@ CHIP_ERROR Encrypt(const CryptoContext & context, CryptoContext::ConstNonceView
...
     uint16_t taglen = 0;
     ReturnErrorOnFailure(mac.Encode(packetHeader, &data[totalLen], msgBuf->AvailableDataLength(), &taglen));
 
-    VerifyOrReturnError(CanCastTo<uint16_t>(totalLen + taglen), CHIP_ERROR_INTERNAL);
-    msgBuf->SetDataLength(static_cast<uint16_t>(totalLen + taglen));
+#if CHIP_SYSTEM_CONFIG_USE_LWIP
+    VerifyOrReturnError(totalLen + taglen <= UINT16_MAX, CHIP_ERROR_INTERNAL);
+#else  // CHIP_SYSTEM_CONFIG_USE_LWIP
+    VerifyOrReturnError(totalLen + taglen <= UINT32_MAX, CHIP_ERROR_INTERNAL);
+#endif // CHIP_SYSTEM_CONFIG_USE_LWIP
```
bzbarsky-apple, line 61: Why are we ending up with ifdefs here?  What situation are we actually trying to guard against, and why does it need an ifdef?

SetDataLength() takes a size_t, yes?  Why can't we just do the addition and pass in our size_t?  The packet buffer will clamp the incoming length itself as needed, so if we want to be 100% sure that the set succeeded, we could just check that the data length is what we expect after the set call.

Again, the old code was trying to handle two things:

1. Overflow on the `totalLen + taglen` addition.  The new code just assumes that will never happen, and maybe that's OK, given that `taglen` is small and `totalLen` fits in sane-sized memory?  But if we ever start having 4GB packet buffers we need to worry about this.
2. Making sure the cast to uint16_t did not drop data.  That's not an issue here, because we are no longer casting to a smaller type.   The only place data might get dropped now is if SetDataLength() clamps, which was an unhandled case in the old code....

pidarped: I guess the checks here are pre-emptive sanity checks for the length. The reason for the ifdef is because LwIP uses u16 throughout. For Linux sockets, UINT32_MAX is a reasonable loose upper bound. I could remove these but it seemed better to have the bound check apriori rather than comparing against the size after SetDataLength(), no? I noticed that SetDataLength() internally has the UINT16_MAX check for LwIP. Maybe, I can add a UINT32_MAX check for Linux sockets there and drop these checks?

pidarped: Removed the checks from here.

src/transport/SecureMessageCodec.cpp
```diff
@@ -43,22 +43,18 @@ CHIP_ERROR Encrypt(const CryptoContext & context, CryptoContext::ConstNonceView
...
 
     MessageAuthenticationCode mac;
     ReturnErrorOnFailure(context.Encrypt(data, totalLen, data, nonce, packetHeader, mac));
 
     uint16_t taglen = 0;
     ReturnErrorOnFailure(mac.Encode(packetHeader, &data[totalLen], msgBuf->AvailableDataLength(), &taglen));
 
-    VerifyOrReturnError(CanCastTo<uint16_t>(totalLen + taglen), CHIP_ERROR_INTERNAL);
-    msgBuf->SetDataLength(static_cast<uint16_t>(totalLen + taglen));
+    msgBuf->SetDataLength(static_cast<size_t>(totalLen + taglen));
```
bzbarsky-apple, line 57: Why does this need a cast?  I would expect the type of `totalLen + taglen` to already be `size_t`, since it's the size of a `size_t` and `uint16_t`, both are unsigned, and `size_t` is at least 16 bits.

pidarped: Yes, this cast is redundant. Have removed it.

bzbarsky-apple (approved):

yunhanw-google (approved):

tcarmelveilleux (commented):
LGTM

PR #33002, opened 2024-04-16
Make test setup functions terminate on error

mbknust:
I changed test setup and teardown functions to take no arguments and return void, because that's what pw_unit_test expects. If an error occurs in the initialization / deinitialization code, I call `VerifyOrDie` instead of returning an error value.

**Followup**: `VerifyOrDie(...)` should be replaced by `ASSERT_SOMETHING(...)` once transition to pw_unit_test is completed.

EDIT: Following suggestions by @andy31415, the scope was reduced - I will make the necessary methods static and rename some context classes in followup PRs.

### DIFF
--- a/src/app/tests/TestEventLoggingNoUTCTime.cpp
+++ b/src/app/tests/TestEventLoggingNoUTCTime.cpp
@@ -84,11 +84,10 @@ class TestContext : public chip::Test::AppContext
 {
 public:
     // Performs shared setup for all tests in the test suite
-    CHIP_ERROR SetUpTestSuite() override
+    void SetUpTestSuite() override
     {
-        ReturnErrorOnFailure(chip::Test::AppContext::SetUpTestSuite());
+        chip::Test::AppContext::SetUpTestSuite();
         mClock.Emplace(chip::System::SystemClock());
-        return CHIP_NO_ERROR;
     }
 
     // Performs shared teardown for all tests in the test suite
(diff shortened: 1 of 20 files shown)


arkq (approved):

src/transport/tests/LoopbackTransportManager.h
```diff
@@ -27,26 +27,26 @@ class LoopbackTransportManager
 {
 public:
     /// Initialize the underlying layers.
-    CHIP_ERROR Init()
+    static CHIP_ERROR Init()
```
andy31415, line 30: This class was not meant to be static. If you need static methods, please separate them somewhere else. Do not use a class as a `namespace` to place everything static inside it.

andy31415 (changes requested):
LoopackTransportManager changes are suspect: we should not convert a class to an odd singleton with all methods static.

src/app/tests/TestEventLoggingNoUTCTime.cpp
```diff
@@ -80,57 +80,56 @@ class MockClock : public chip::System::Clock::ClockBase
     chip::System::Clock::ClockBase & mRealClock;
 };
 
-class TestContext : public chip::Test::AppContext
+class TestContext : public chip::Test::AppTestFixture
 {
 public:
     // Performs shared setup for all tests in the test suite
-    CHIP_ERROR SetUpTestSuite() override
+    static void SetUpTestSuite()
```
andy31415, line 87: make this call a static method if that is what is needed. Converting a subset of methods to be static is odd ... there is no clear reason of why some are static and some are not.

Maybe we should break down this PR into smaller chunks to define a good pattern on splitting things.

andy31415: I see googletest style wants these static. Marking resolved.

src/transport/tests/LoopbackTransportManager.h
mbknust: Ok, I will keep this in mind when switching to static in followup PR

cecille:
@Alami-Amine - FYI

andy31415 (approved):

PR #33005, opened 2024-04-16
[zephyr] Added Bluetooth LE Extended Advertisement option

### LINKED ISSUE #33020: [BUG] MaxCommissioningTimeout returned by CommissioningWindowManager overflows 16-bit type
### Reproduction steps

The `CommissioninWindowManager` supports `CHIP_DEVICE_CONFIG_BLE_EXT_ADVERTISING` what makes it return `MaxCommissioningTimeout` equal to 48 hours (172800 seconds). The method returns `Seconds16` type, and the returned value will not fit in the range of 16 bit value, so it's going to overflow.

### Bug prevalence

Always

35ced57c3828238d2d125cd3116a7b0e0410e286

### Platform

core

### Platform Version(s)

_No response_

### Anything else?

_No response_

kkasperczyk-no:
This commit implements platform solution for a Bluetooth LE extended advertising.

Additionally, for the CommissioningWindowManager types were changed from Seconds16 to Seconds32, because the current implementation overflows for 48h duration.

Fixed: #33020


src/app/server/CommissioningWindowManager.h
```diff
@@ -58,21 +58,21 @@ class CommissioningWindowManager : public Messaging::UnsolicitedMessageHandler,
         return CHIP_NO_ERROR;
     }
 
-    static constexpr System::Clock::Seconds16 MaxCommissioningTimeout()
+    static constexpr System::Clock::Seconds32 MaxCommissioningTimeout()
     {
 #if CHIP_DEVICE_CONFIG_BLE_EXT_ADVERTISING
         // Specification section 2.3.1 - Extended Announcement Duration up to 48h
-        return System::Clock::Seconds16(60 * 60 * 48);
+        return System::Clock::Seconds32(60 * 60 * 48);
```
Damian-Nordic, line 65: Consider filing an issue for this integer overflow and associating it with this PR so that it's not overseen as the PR is prefixed with `[zephyr]`.

Damian-Nordic (approved):

woody-apple (approved):

src/app/server/CommissioningWindowManager.h
kkasperczyk-no: Ok, I've updated the PR description with an issue link.

PR #33023, opened 2024-04-17
Force including <ble/Ble.h> umbrella header instead of specific headers

arkq:
### Problem

In the `src/ble` library there is a dedicated `<ble/Ble.h>` umbrella header which ought to be a dedicated public header for the BLE library. However, not everyone is accessing BLE library API via this header.

### Changes

- force everyone to use `<ble/Ble.h>` (one exception is `<ble/BleConfig.h>` which is included in the `CHIPConfig.h`)
- fix includes in `src/ble` according to IWYU output
- improve `iwyu-check.py` helper script a bit

### Testing

CI should detect potential build breaks.


Damian-Nordic:
Actually, why is it better to have a single header for everything instead of including only what is needed?

arkq:
> Actually, why is it better to have a single header for everything instead of including only what is needed?

Actually it depends. But I thought that the intent of the `ble/Ble.h` header was to define a single header for the BLE internal library, so any refactoring in the src/ble will not imply changes in dependent components. However, since this lib is a static lib used (only ?) internally, maybe this `ble/Ble.h` should be dropped in favor of actually including what is needed. So, there are two ways of doing things and I think that we should choose one of these. Personally, I would favor your way - treat this BLE library as an internal library only and do not define any public include for it. But since there was already this umbrella header, I've tried to use only it in the rest of the code.

Damian-Nordic:
> > Actually, why is it better to have a single header for everything instead of including only what is needed?
> 
> Actually it depends.
> 
> I thought that the intent of the `ble/Ble.h` header was to define a single header for the BLE internal library, so any refactoring in the src/ble will not imply changes in dependent components. However, since this lib is a static lib used (only ?) internally, maybe this `ble/Ble.h` should be dropped in favor of actually including what is needed. So, there are two ways of doing things and I think that we should choose one of these. Personally, I would favor your way - treat this BLE library as an internal library only and do not define any public include for it. But since there was already this umbrella header, I've tried to use only it in the rest of the code.

Thanks for the clarification. Right, I don't have a strong opinion which option is better (though subjectively, I prefer if headers explicitly include what they need instead of relying on some outer header) and your PR is an improvement in a way it at least enforces the correct usage of the headers, but I was just curious about you reasons.

yunhanw-google (approved):

Damian-Nordic (approved):

PR #33024, opened 2024-04-17
TC-CNET-4.13: Update cnet 4.13 test as per spec

Kshitijjain21:
The test plan states that the _'TH reads the network attribute and saves the list 'OriginalNetworkList''_ after successfully sending an ArmFailSafe and AddOrUpdateWiFiNetwork commands. Later on in the test (step 14 and 15) the TH send another ArmFailSafe with the expiry length set to 0 and then reads the network. The verification in the test plans states that the networks attribute list should be same as the _'OriginalNetworkList'._
In the matter spec under _'Behavior on expiry of Fail-Safe timer'_ it mentions to _'Reset the configuration of all Network Commissioning Networks attribute to their state prior to the Fail-Safe being armed'_.
Based on the steps mentioned in the test plan for CNET 4.13 and matter specs there should be a commission complete command after adding new network info which should be the  _'OriginalNetworkList'._
Note: The DUT's max network needs  to be greater than 1.

Test plan PR - https://github.com/CHIP-Specifications/chip-test-plans/pull/4125

Link to slack - https://csamembers.slack.com/archives/C02723N0C7N/p1713277617971359


src/app/tests/suites/certification/Test_TC_CNET_4_13.yaml
```diff
@@ -122,20 +123,28 @@ tests:
           ( Pre-Condition)
       disabled: true
 
+    - label: "Step 7: TH sends the CommissioningComplete command to the DUT"
+      verification: |
+          ./chip-tool generalcommissioning CommissioningComplete 1 0
+
+          The test case is not verifiable in RPI platform. As MaxNetworks value is 1 but expected is 4
+          ( Pre-Condition)
+      disabled: true
```
andy31415, line 132: This is a new disabled (i.e. manual) test step. We generally do not allow new manual test steps anymore.

Kshitijjain21: Thanks for the feedback. The change done here is allow this test to align with the matter spec. Since the test was a manual test and has not been changed to an automated/semi automated test what would be the correct course of action to get it to align with the spec?

cjandhyala: @andy31415 since this is a old cert test, we can make updates to the manual test. 
@Kshitijjain21 But we can not merge this change unless test plan PR is approved.

andy31415:
@Kshitijjain21 @cjandhyala this is a very old PR with merge conflicts. Since this was considering adding manual tests, what do we do with it? Can we update/automate or should it be closed?

Kshitijjain21:
> @Kshitijjain21 @cjandhyala this is a very old PR with merge conflicts. Since this was considering adding manual tests, what do we do with it? Can we update/automate or should it be closed?

@andy31415 Happy to close and reopen this after the test plan change has been merged in. Although this test require that device can support up to 4 networks. Currently in the matter code its hard coded to 1 network per device and would require changes to support 4 network. Will wait on @cjandhyala thought before closing this PR.

andy31415:
> > @Kshitijjain21 @cjandhyala this is a very old PR with merge conflicts. Since this was considering adding manual tests, what do we do with it? Can we update/automate or should it be closed?
> 
> @andy31415 Happy to close and reopen this after the test plan change has been merged in. Although this test require that device can support up to 4 networks. Currently in the matter code its hard coded to 1 network per device and would require changes to support 4 network. Will wait on @cjandhyala thought before closing this PR.

We can close and re-open, however I would note that we will still reject any manual tests addition in the SDK (including extra tests because existing tests are already manual ... we can update existing manual tests because that makes things not worse, but new tests should not be manual, even for clusters that already have manual tests as that makes the problem worse).

Do we have a plan to automate things?

src/app/tests/suites/certification/Test_TC_CNET_4_13.yaml
andy31415: This adds several new steps that are manual ... so it kind of makes the problem worse. I understand the justification of "we add a test step in a long list of test steps that is already manual" however I also am concerned that this does increase the count of manual stuff invocation. We should not do that.

Kshitijjain21:
> > > @Kshitijjain21 @cjandhyala this is a very old PR with merge conflicts. Since this was considering adding manual tests, what do we do with it? Can we update/automate or should it be closed?
> > 
> > 
> > @andy31415 Happy to close and reopen this after the test plan change has been merged in. Although this test require that device can support up to 4 networks. Currently in the matter code its hard coded to 1 network per device and would require changes to support 4 network. Will wait on @cjandhyala thought before closing this PR.
> 
> We can close and re-open, however I would note that we will still reject any manual tests addition in the SDK (including extra tests because existing tests are already manual ... we can update existing manual tests because that makes things not worse, but new tests should not be manual, even for clusters that already have manual tests as that makes the problem worse).
> 
> Do we have a plan to automate things?

Hi @andy31415, I am currently not sure on the plan of automating this test.

PR #33028, opened 2024-04-17
Some interface cleanup and make AttributeValueEncoder/Decoder more similar

andy31415:
### Changes

- Decouple the Encoding state from the AttributeValueEncoder to not be a subclass anymore (reduce usage of friend classes)
- Make both encoder and decoder expose a `SubjectDescriptor`. Expectation is that in the future we may want to act based on these descriptors (e.g. if our interfaces are also required to validate access permissions).
- Re-arrange some of the members to better use space, especially since I added the subject descriptor so the sturct size may have become slightly larger.


src/app/AttributeValueEncoder.h
```diff
@@ -226,7 +248,7 @@ class AttributeValueEncoder
         }
 
         mCurrentEncodingListIndex++;
-        mEncodeState.mCurrentEncodingListIndex++;
+        mEncodeState.SetCurrentEncodingListIndex(mCurrentEncodingListIndex);
```
andy31415, line 251: Here I believe the previous intent was to increase both and they seem to be kept in sync since line 220 seems to increment until they are equal.

The exact equivalent is `mEncodeState.SetCurrentEncodingListIndex(mEncodeState.CurrentEncodingListIndex() + 1)` however current "they are equal" seems clearer.

Unit and integration tests should detect if something is off - we do have tests for chunking.

src/app/util/ember-compatibility-functions.cpp
```diff
@@ -425,16 +425,20 @@ CHIP_ERROR GlobalAttributeReader::EncodeCommandList(const ConcreteClusterPath &
 // Helper function for trying to read an attribute value via an
 // AttributeAccessInterface.  On failure, the read has failed.  On success, the
 // aTriedEncode outparam is set to whether the AttributeAccessInterface tried to encode a value.
-CHIP_ERROR ReadViaAccessInterface(FabricIndex aAccessingFabricIndex, bool aIsFabricFiltered,
+CHIP_ERROR ReadViaAccessInterface(SubjectDescriptor subjectDescriptor, bool aIsFabricFiltered,
```
bzbarsky-apple, line 428: Why are we passing the SubjectDecriptor by value?

src/app/util/mock/attribute-storage.cpp
```diff
@@ -351,9 +350,15 @@ CHIP_ERROR ReadSingleMockClusterData(FabricIndex aAccessingFabricIndex, const Co
...
-        AttributeValueEncoder valueEncoder(aAttributeReports, aAccessingFabricIndex, aPath, dataVersion, false, state);
+        AttributeEncodeState state;
+        if (apEncoderState != nullptr)
+        {
+            state = *apEncoderState;
+        }
+        Access::SubjectDescriptor subject;
+        subject.fabricIndex = aAccessingFabricIndex;
+
+        AttributeValueEncoder valueEncoder(aAttributeReports, subject, aPath, dataVersion, false /* isFabricFiltered */, state);
```
bzbarsky-apple, line 361:
```suggestion
        AttributeValueEncoder valueEncoder(aAttributeReports, subject, aPath, dataVersion, /* isFabricFiltered = */ false , state);
```

src/controller/tests/data_model/TestRead.cpp
```diff
@@ -109,10 +109,13 @@ CHIP_ERROR ReadSingleClusterData(const Access::SubjectDescriptor & aSubjectDescr
...
-            AttributeValueEncoder::AttributeEncodeState state =
-                (apEncoderState == nullptr ? AttributeValueEncoder::AttributeEncodeState() : *apEncoderState);
-            AttributeValueEncoder valueEncoder(aAttributeReports, aSubjectDescriptor.fabricIndex, path,
-                                               kDataVersion /* data version */, aIsFabricFiltered, state);
+            AttributeEncodeState state;
+            if (apEncoderState != nullptr)
+            {
+                state = *apEncoderState;
+            }
+            AttributeValueEncoder valueEncoder(aAttributeReports, aSubjectDescriptor, path, kDataVersion /* data version */,
```
bzbarsky-apple, line 117: Pre-existing, but I don't think this comment adds much here.

Same in a few places below.

src/controller/tests/data_model/TestRead.cpp
```diff
@@ -141,19 +147,22 @@ CHIP_ERROR ReadSingleClusterData(const Access::SubjectDescriptor & aSubjectDescr
...
             aPath.mAttributeId == app::Clusters::UnitTesting::Attributes::Int16u::Id)
         {
-            AttributeValueEncoder::AttributeEncodeState state =
-                (apEncoderState == nullptr ? AttributeValueEncoder::AttributeEncodeState() : *apEncoderState);
-            AttributeValueEncoder valueEncoder(aAttributeReports, aSubjectDescriptor.fabricIndex, aPath,
-                                               kDataVersion /* data version */, aIsFabricFiltered, state);
+            AttributeEncodeState state;
+            if (apEncoderState != nullptr)
+            {
+                state = *apEncoderState;
```
bzbarsky-apple, line 153: OK, we have a _lot_ of copies of this pattern.  Is it worth having a `static AttributeEncodeState CopyState(AttributeEncodeState *)` or something which will do this in one place?  Don't feel strongly about this one.

bzbarsky-apple (changes requested):
The "this makes no sense" bit and the AccessingFabricIndex bit are why I'm requesting changes...

src/app/AttributeValueDecoder.h
```diff
@@ -46,19 +46,14 @@ class AttributeValueDecoder
...
+        aArg.SetFabricIndex(GetSubjectDescriptor().fabricIndex);
         return CHIP_NO_ERROR;
     }
 
     bool TriedDecode() const { return mTriedDecode; }
 
-    /**
-     * The accessing fabric index for this write interaction.
-     */
-    FabricIndex AccessingFabricIndex() const { return mSubjectDescriptor.fabricIndex; }
```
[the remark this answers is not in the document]
andy31415: The difficulty I have is separation of concerns: this class is an `AttributeValueDecoder` so it should decode attribute values. Yet we started moving along subject descriptors and have an accessing fabric index. My path is to split out the decoding/encoding part (which is generally pure value) and the request data which is generally the subjectdescriptor.

I was looking to minimise the number of calls that are not `encode/decode` related, in this case we have `subject_descriptor`  and `accessing_fabric_index` and the latter can be derived from the first. 

I am claiming that `AttributeValueDecoder` should only decode values and `AttributeValueEncoder` should only encode values ... and they should not contain data about subject descriptors, however pulling that out seems to need to be done stepwise.

Is this goal reasonable or is your opinion that the existing layout is the more correct one?

src/app/util/mock/attribute-storage.cpp
andy31415: I did not do this because AttributeValueEncoder uses the `aCamelCase` convention for arguments and clang-tidy enforces that `/* ...  = */` arguments match ... and `aIsFabricFiltered` felt uglyer than `isFabricFiltered` .

Updated to the `=` style with appropriate name, however overall I am not a fan of `a...` naming style that we inherited.

src/controller/tests/data_model/TestRead.cpp
andy31415: Removed some of the useless comments.

src/controller/tests/data_model/TestRead.cpp
andy31415: I created a constructor from a pointer, since this seems to be the common use case.

src/app/util/ember-compatibility-functions.cpp
andy31415: Search and replace. Replaced with const reference.

src/app/AttributeValueEncoder.h
```diff
@@ -27,6 +28,60 @@
 namespace chip {
 namespace app {
 
+/**
+ * Maintains the internal state of list encoding
+ */
+class AttributeEncodeState
```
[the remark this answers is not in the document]
andy31415: Moved it to a separate file and also updated some comments and added a constructor from a pointer.

src/app/AttributeValueDecoder.h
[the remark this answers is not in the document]
bzbarsky-apple: Fundamentally, _how you encode a value for a read depends on the accessing fabric index_ in the spec, yes?  So I am not sure how you want to separate those bits.  At some point, something involved in the encoding needs to know the accessing fabric index.

Similarly, how you decode a value depends on the accessing fabric index.  Arguably that could be handled by the decode consumers, but it's extremely error prone and security-critical, so we want to make it impossible to get wrong.  See "7.19.1.9. Struct" which says:

> When a write interaction creates a fabric-scoped struct entry (in a fabric-scoped list), the server SHALL implicitly load the accessing fabric-index into the global FabricIndex field of the struct.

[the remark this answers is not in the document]
bzbarsky-apple: And I agree that the subject descriptor was sort of smuggled along in there and conceptually you could have a "write context" of some sort that has an AttributeValueDecoder and a a subject descriptor.  Or something.  But the decode and encode really do need the fabric index pretty organically as far as I can tell.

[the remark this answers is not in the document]
andy31415: Added it back - for now left this PR then to just split out the state in a stand alone class and make subject descriptor symmetrical.

I believe we have to figure out how to make decoder/encoder actually only do encode/decode, however we can figure that out separately.

src/app/AttributeValueDecoder.h
bzbarsky-apple, line 1: At this point, there should be no changes to this file: no need to remove the API docs and no need to re-order the APIs.

src/app/AttributeValueEncoder.h
```diff
@@ -169,17 +140,16 @@ class AttributeValueEncoder
...
-            mEncodeState = AttributeEncodeState();
+            mEncodeState.Reset();
         }
         return err;
     }
 
     bool TriedEncode() const { return mTriedEncode; }
 
-    /**
-     * The accessing fabric index for this read or subscribe interaction.
```
bzbarsky-apple, line 180: Again, why removing the API docs?  The key part is not documenting that this is the accessing fabric index (which is obvious from the name, sure), but _what it's the accessing fabric index for_, in spec terms.

bzbarsky-apple (approved):

tcarmelveilleux (approved):

PR #33036, opened 2024-04-17
Make chip::Dnssd::kMaxDeviceNameLen configurable

### LINKED ISSUE #33034: [BUG] Discovered deviceName is cutoff or malformed
### Reproduction steps

1. On MacOS, run the following to advertise a Matter commissioner:
`dns-sd -R DD200C20D25AE5F7 _matterd._udp,_V123,_T35 . 33333 VP=123+456 DT=35 DN="シャラドのエコーショー15"`

2. Build and run the iOS tv-casting-app.

3. Click "Skip to commissioner discovery" and then click "Start discovery"

4. Note that the deviceName is surfaced as null.

### Bug prevalence

Whenever I do the above

035d3023e21615f12bd730bbe197af5d96e511a4

### Platform

core

### Platform Version(s)

_No response_

### Anything else?

This seems to be happening because chip::Dnssd::kMaxDeviceNameLen is set and hardcoded to 32 bytes. This may not be enough for some applications and languages where deviceNames can exceed this length.

sharadb-amazon:
Fixes https://github.com/project-chip/connectedhomeip/issues/33034

### Change Summary
1. Added a gn variable chip_max_device_name_len and set its default value to the existing value of chip::Dnssd::kMaxDeviceNameLen.
2. Used chip_max_device_name_len to set a C++ macro CHIP_DEVICE_CONFIG_MAX_DEVICE_NAME_LEN in src/platform/BUILD.gn
3. Overrode chip_max_device_name_len to 128 for the android/darwin/linux tv-casting-apps in their args.gni files. For TV Casting use cases, TV names can be strings that are longer than 32 bytes.

### Testing
Tested with the iOS tv-casting-app and saw the full deviceName appearing in discovery results: シャラドのエコーショー15


bzbarsky-apple (changes requested):
Spec says, in "4.3.1.9. TXT key for device name (DN)":

> The optional key DN MAY provide a device advertisement name. If present, it SHALL be encoded as a valid UTF-8 string with a maximum length of 32 bytes (matching the maximum length of the NodeLabel string in the Basic Information Cluster).

So if somethiing has a longer name than that, it's violating the spec and should stop doing that.

sharadb-amazon:
> Spec says, in "4.3.1.9. TXT key for device name (DN)":
> 
> > The optional key DN MAY provide a device advertisement name. If present, it SHALL be encoded as a valid UTF-8 string with a maximum length of 32 bytes (matching the maximum length of the NodeLabel string in the Basic Information Cluster).
> 
> So if somethiing has a longer name than that, it's violating the spec and should stop doing that.

Understood. Canceling this PR for now.

PR #33038, opened 2024-04-17
v1.3 Commissioner Passcode field for Linux and Android tv-casting-apps

pgregorr-amazon:
Exposed the new Matter v1.3 Commissioner Passcode (CP) field for discovered Casting Players for Linux and Android example tv-casting-apps.

**Change summary**

1.	Updated CastingPlayer.h/cpp with the new Commissioner Passcode field for Linux. Added getter function.
2.	Updated MatterCastingPlayer.java for Android. Updated the DiscoveryExampleFragment.java to display the new field in the UX. 
3. Added Commissioner Passcode field to CastingStore.h/cpp WriteAll() and ReadAll()so that all the details about a connected CastingPLayer can be cached.

**Testing**

Verified and tested locally with the Linux and Android tv-casting-app example apps. The example apps discover CastingPlayers and display the Commissioner Passcode field for each discovered CastingPlayer in the terminal and Android mobile device tv-casting-app UX respectively.

### DIFF
--- a/examples/tv-casting-app/android/App/app/src/main/java/com/matter/casting/DiscoveryExampleFragment.java
+++ b/examples/tv-casting-app/android/App/app/src/main/java/com/matter/casting/DiscoveryExampleFragment.java
@@ -350,6 +350,10 @@ class CastingPlayerArrayAdapter extends ArrayAdapter<CastingPlayer> {
             ? (aux.isEmpty() ? "" : ", ") + "Device Type: " + player.getDeviceType()
             : "";
     aux += (aux.isEmpty() ? "" : ", ") + "Resolved IP?: " + (player.getIpAddresses().size() > 0);
+    aux +=
+        (aux.isEmpty() ? "" : ", ")
+            + "Commissioner Passcode Supported?: "
+            + (player.isCommissionerPasscodeSupported());
 
     aux = aux.isEmpty() ? aux : "\n" + aux;
     return main + aux;
--- a/examples/tv-casting-app/android/App/app/src/main/java/com/matter/casting/DiscoveryExampleFragment.java
+++ b/examples/tv-casting-app/android/App/app/src/main/java/com/matter/casting/DiscoveryExampleFragment.java
@@ -352,8 +352,8 @@ class CastingPlayerArrayAdapter extends ArrayAdapter<CastingPlayer> {
     aux += (aux.isEmpty() ? "" : ", ") + "Resolved IP?: " + (player.getIpAddresses().size() > 0);
     aux +=
         (aux.isEmpty() ? "" : ", ")
-            + "Commissioner Passcode Supported?: "
-            + (player.isCommissionerPasscodeSupported());
+            + "Commissioner Passcode: "
+            + (player.getCommissionerPasscode());
 
     aux = aux.isEmpty() ? aux : "\n" + aux;
     return main + aux;
--- a/examples/tv-casting-app/android/App/app/src/main/java/com/matter/casting/DiscoveryExampleFragment.java
+++ b/examples/tv-casting-app/android/App/app/src/main/java/com/matter/casting/DiscoveryExampleFragment.java
@@ -352,8 +352,8 @@ class CastingPlayerArrayAdapter extends ArrayAdapter<CastingPlayer> {
     aux += (aux.isEmpty() ? "" : ", ") + "Resolved IP?: " + (player.getIpAddresses().size() > 0);
     aux +=
         (aux.isEmpty() ? "" : ", ")
-            + "CommissionerPasscode: "
-            + (player.isCommissionerPasscodeSupported());
+            + "Supports Commissioner Generated Passcode: "
+            + (player.getSupportsCommissionerGeneratedPasscode());
 
     aux = aux.isEmpty() ? aux : "\n" + aux;
     return main + aux;
--- a/examples/tv-casting-app/android/App/app/src/main/java/com/matter/casting/DiscoveryExampleFragment.java
+++ b/examples/tv-casting-app/android/App/app/src/main/java/com/matter/casting/DiscoveryExampleFragment.java
@@ -352,8 +352,8 @@ class CastingPlayerArrayAdapter extends ArrayAdapter<CastingPlayer> {
     aux += (aux.isEmpty() ? "" : ", ") + "Resolved IP?: " + (player.getIpAddresses().size() > 0);
     aux +=
         (aux.isEmpty() ? "" : ", ")
-            + "Commissioner Passcode: "
-            + (player.getCommissionerPasscode());
+            + "CommissionerPasscode: "
+            + (player.isCommissionerPasscodeSupported());
 
     aux = aux.isEmpty() ? aux : "\n" + aux;
     return main + aux;
--- a/examples/tv-casting-app/android/App/app/src/main/jni/com/matter/casting/core/MatterCastingPlayer.java
+++ b/examples/tv-casting-app/android/App/app/src/main/jni/com/matter/casting/core/MatterCastingPlayer.java
@@ -47,6 +47,7 @@ public class MatterCastingPlayer implements CastingPlayer {
   private int productId;
   private int vendorId;
   private long deviceType;
+  private boolean commissionerPasscodeSupported;
   protected long _cppCastingPlayer;
 
   public MatterCastingPlayer(
@@ -59,7 +60,8 @@ public class MatterCastingPlayer implements CastingPlayer {
       int port,
       int productId,
       int vendorId,
-      long deviceType) {
+      long deviceType,
+      boolean commissionerPasscodeSupported) {
     this.connected = connected;
     this.deviceId = deviceId;
     this.hostName = hostName;
@@ -70,6 +72,7 @@ public class MatterCastingPlayer implements CastingPlayer {
     this.productId = productId;
     this.vendorId = vendorId;
     this.deviceType = deviceType;
+    this.commissionerPasscodeSupported = commissionerPasscodeSupported;
   }
 
   /**
@@ -129,6 +132,11 @@ public class MatterCastingPlayer implements CastingPlayer {
   @Override
   public long getDeviceType() {
     return this.deviceType;
+  }
+
+  @Override
+  public boolean isCommissionerPasscodeSupported() {
+    return this.commissionerPasscodeSupported;
   }
 
   @Override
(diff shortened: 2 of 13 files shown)


examples/tv-casting-app/tv-casting-common/core/CastingPlayer.h
```diff
@@ -56,13 +56,14 @@ class CastingPlayerAttributes
...
     char instanceName[chip::Dnssd::Commission::kInstanceNameMaxLength + 1] = {};
-    unsigned int numIPs; // number of valid IP addresses
+    unsigned int numIPs; // Number of valid IP addresses
     chip::Inet::IPAddress ipAddresses[chip::Dnssd::CommonResolutionData::kMaxIPAddresses];
     chip::Inet::InterfaceId interfaceId;
     uint16_t port;
     uint16_t productId;
     uint16_t vendorId;
     uint32_t deviceType;
+    uint8_t commissionerPasscode = 0; // Indicates whether a CastingPlayer supports the Commissioner-Generated Passcode feature.
```
sharadb-amazon, line 66: Can we add code to the CastingStore to read/write this value.

examples/tv-casting-app/tv-casting-common/core/CastingPlayer.h
```diff
@@ -113,6 +114,11 @@ class CastingPlayer : public std::enable_shared_from_this<CastingPlayer>
      */
     bool IsConnected() const { return mConnectionState == CASTING_PLAYER_CONNECTED; }
 
+    /**
+     * @return true if this CastingPlayer supports the Commissioner-Generated Passcode feature.
+     */
+    bool IsCommissionerPasscodeSupported() const { return mAttributes.commissionerPasscode != 0; }
```
sharadb-amazon, line 120: It looks like the commissionerPasscode returned by the underlying SDK is a uint8\_t. To future proof the casting APIs, perhaps we should just align with that?

```suggestion
    uint8_t GetCommissionerPasscode() const { return mAttributes.commissionerPasscode; }
```

examples/tv-casting-app/tv-casting-common/core/CastingPlayer.cpp
```diff
@@ -281,6 +282,7 @@ void CastingPlayer::LogDetail() const
     {
         ChipLogDetail(AppServer, "\tDevice Type: %" PRIu32, mAttributes.deviceType);
     }
+    ChipLogDetail(AppServer, "\tCommissioner Passcode feature supported: %u", mAttributes.commissionerPasscode);
```
sharadb-amazon, line 285:
```suggestion
    ChipLogDetail(AppServer, "\tCommissioner Passcode: %u", mAttributes.commissionerPasscode);
```

examples/tv-casting-app/android/App/app/src/main/java/com/matter/casting/DiscoveryExampleFragment.java
```diff
@@ -351,6 +351,10 @@ private String getCastingPlayerButtonText(CastingPlayer player) {
             ? (aux.isEmpty() ? "" : ", ") + "Device Type: " + player.getDeviceType()
             : "";
     aux += (aux.isEmpty() ? "" : ", ") + "Resolved IP?: " + (player.getIpAddresses().size() > 0);
+    aux +=
+        (aux.isEmpty() ? "" : ", ")
+            + "Commissioner Passcode Supported?: "
+            + (player.isCommissionerPasscodeSupported());
```
sharadb-amazon, line 357:
```suggestion
            + "Commissioner Passcode: "
            + (player.getCommissionerPasscode());
```

examples/tv-casting-app/tv-casting-common/core/CastingPlayer.cpp
pgregorr-amazon: okay

examples/tv-casting-app/android/App/app/src/main/java/com/matter/casting/DiscoveryExampleFragment.java
pgregorr-amazon: okay

examples/tv-casting-app/tv-casting-common/core/CastingPlayer.h
pgregorr-amazon: Okay, will pass as a short to the Java layer

examples/tv-casting-app/tv-casting-common/core/CastingPlayer.h
pgregorr-amazon: Okay.

examples/tv-casting-app/tv-casting-common/support/CastingStore.h
```diff
@@ -83,6 +83,7 @@ class CastingStore : public chip::FabricTable::Delegate
         kCastingPlayerPortTag,
         kCastingPlayerInstanceNameTag,
         kCastingPlayerDeviceNameTag,
+        kCastingCommissionerPasscodeTag,
```
sharadb-amazon, line 86:
```suggestion
        kCastingPlayerCommissionerPasscodeTag,
```

sharadb-amazon (approved):

examples/tv-casting-app/tv-casting-common/support/CastingStore.h
pgregorr-amazon: fixed

examples/tv-casting-app/tv-casting-common/core/CastingPlayer.h
andy31415, line 66: I am unclear based on the comment how a uint8_t `indicates weather a cstingplayer supports ....`.

Is it `if not 0, ....` ? Maybe we need a constant for this 0 constant to be `kInvalidComissionerPasscode` ? Does the spec explicitly say that 0 is not a valid value or should this be some form of optional?

andy31415 (changes requested):
My spec copy says "CommissionerPasscode" is a flag to instruct the commissioner to use the commissioner-generated passcode.

This seems to indicate we need some form of `bool` rather than a uint8_t.

examples/tv-casting-app/tv-casting-common/support/CastingStore.h
```diff
@@ -83,6 +83,7 @@ class CastingStore : public chip::FabricTable::Delegate
         kCastingPlayerPortTag,
         kCastingPlayerInstanceNameTag,
         kCastingPlayerDeviceNameTag,
+        kCastingPlayerCommissionerPasscodeTag,
```
andy31415, line 86: This re-orders/re-names all the other tags. Is this expected ? I would expect TLV tags to be fixed as long as there is a possiblity of old data to be stored or received (e.g. by combining new and old client/server applications).

Is this not a concern here? If so there should be a comment associated with this enum otherwise you have backwards compatibility issues.

examples/tv-casting-app/tv-casting-common/core/CastingPlayer.h
andy31415: Why? The spec seems to say "it is a boolean".

andy31415: I would generally try to not "change this for future proofing" unless you have clear path (i.e. happening in O(days)) that a change is needed. In 99% of the cases the "for the future" never happens and just results in more complex code.

examples/tv-casting-app/tv-casting-common/support/CastingStore.h
pgregorr-amazon: I'll move it to the end in the follow up commit.

examples/tv-casting-app/tv-casting-common/core/CastingPlayer.h
pgregorr-amazon: Per chrisdecenzo this should be a boolean. Will update.

examples/tv-casting-app/tv-casting-common/core/CastingPlayer.h
pgregorr-amazon: changing to a boolean

examples/tv-casting-app/android/App/app/src/main/jni/com/matter/casting/core/MatterCastingPlayer.java
```diff
@@ -47,6 +47,12 @@ public class MatterCastingPlayer implements CastingPlayer {
   private int productId;
   private int vendorId;
   private long deviceType;
+  /**
+   * The CommissionerPasscode field indicates whether a CastingPlayer supports the
+   * Commissioner-Generated Passcode feature.
+   */
+  private boolean commissionerPasscode;
```
andy31415, line 54: how about naming this `supportsCommissionerGeneratedPasscode` and then you do not need the lengthy comment?

examples/tv-casting-app/tv-casting-common/core/CastingPlayer.h
```diff
@@ -56,13 +56,17 @@ class CastingPlayerAttributes
...
     chip::Inet::IPAddress ipAddresses[chip::Dnssd::CommonResolutionData::kMaxIPAddresses];
     chip::Inet::InterfaceId interfaceId;
     uint16_t port;
     uint16_t productId;
     uint16_t vendorId;
     uint32_t deviceType;
+    /**
+     * The CommissionerPasscode field indicates whether a CastingPlayer supports the Commissioner-Generated Passcode feature.
+     */
+    bool commissionerPasscode;
```
andy31415, line 69: same comment here: `supportsCommissionerGeneratedPasscode` or similar makes it less likely that you need a comment that indicates that `supports` is they thing here. Otherwise `commissionerPasscode` generally sounds like a non-bool (like a number/char passcode). Booleans generally have a prefix like `is/has/supports/contains`.

andy31415 (approved):

examples/tv-casting-app/android/App/app/src/main/jni/com/matter/casting/core/MatterCastingPlayer.java
pgregorr-amazon: Okay.

examples/tv-casting-app/tv-casting-common/core/CastingPlayer.h
pgregorr-amazon: Changing in the next commit: "Addressed comments by andy31415, part 2 member variable name change"

PR #29342, opened 2023-09-20
Testing using pw_unit_test

mbknust:
### Problem

Matter currently uses [nlunit-test](https://github.com/nestlabs/nlunit-test) as a testing framework, which isn't maintained anymore. Moreover, the output it gives when some assertion fails is not very informative, eg. when asserting the value of some expression, it doesn't print what the expected and actual values were.

### Changes

All unit tests have been rewritten to use the [pw_unit_test](https://pigweed.dev/pw_unit_test/) framework instead. Since pigweed is already a dependency of Matter, and was developed with embedded devices in mind, it seemed a good choice. It also aims to be source-compatible with googletest, so most developers will feel right at home using it. 

#### Build system

-   support for `pw_unit_test` added to `./.gn`
-   the GN template in `./build/chip/chip_test_suite.gni` changed, so it invokes `pw_test`
-   minor changes in `build/chip/chip_test.gni` and `build/chip/chip_test_group.gni`, to keep target naming consistent
-   small adjustments in `**/BUILD.gn`

#### Unit test files

The overwhelming majority of changes were made semi-automatically using a simple regex-based rewriting script. This crude method worked quite well, though most files needed some small manual intervention afterwards. The actual testing logic was unchanged - the rewritten files ought to behave in exactly the same way as before. These are the kinds of changes made:

1. including `<gtest/gtest.h>` instead of `<nlunit-test.h>`
2. replacing `NL_TEST_ASSERT(suite, condition)` with `EXPECT_TRUE(condition)`
3. removing the redundant `nlTestSuite * apSuite` parameter from functions
4. creating test fixture classes (for test suites requiring setup/teardown)
5. turning individual unit test functions into `TEST`, `TEST_F`, or `STATIC_TEST`
6. moving `TestContext`s into test fixture classes as static members
7. deleting `CHIP_REGISTER_TEST_SUITE`, `nlTest[]`, `nlTestSuite`, etc.

Unlike `nlunit-test`, testcases are run by `pw_unit_test` in the order in which they were declared in the source code. To keep the run-time order of testcases unchanged, sometimes their source-code declaration order had to be reshuffled (causing thousand-line-long diffs, even though nothing substantial changed!).

### Testing

Running `./scripts/tests/all_tests.sh` shows that all test suites compile and run successfully.

### Followups

The following problems were encountered, but couldn't be fixed "properly", because that would require changing the testing code (and risk breaking it). Instead, provisional workarounds were applied:

[...]

PR #33042, opened 2024-04-18
Replace nl-unit-test for pigweed for crypto

jlatusek:
## Problem

## Changes
1. These changes are another piece of replacing the nl-unit test with the pigweed-unit test. I changed tests for the Crypto. I didn't introduce any changes to the logic of tests; I only replaced one framework with another. 

2. Changes to the `messenger` folder were required to make old nlunit tests work. There are dependencies between tests. Changing the test unit framework changed the order of the tests, and `psa_crypto_init` was not called before the messaging tests.


src/crypto/tests/TestChipCryptoPAL.cpp
```diff
@@ -121,15 +119,15 @@ class HeapChecker
 class HeapChecker
 {
 public:
-    explicit HeapChecker(nlTestSuite *) {}
+    explicit HeapChecker() {}
```
andy31415, line 122: It seems we need to change above line 95 as well.
Do we never use MBEDTLS_MEMORY_DEBUG in CI?

src/crypto/tests/TestChipCryptoPAL.cpp
```diff
@@ -1226,66 +1238,65 @@ static void TestECDH_EstablishSecret(nlTestSuite * inSuite, void * inContext)
...
         (void) DRBG_get_bytes(buffer, sizeof(buffer));
     }
-    NL_TEST_ASSERT(inSuite, gs_test_entropy_source_called > test_entropy_source_call_count);
+    EXPECT_GT(gs_test_entropy_source_called, test_entropy_source_call_count);
 }
 #endif
 
 #if CHIP_CRYPTO_PSA
-static void TestAddEntropySources(nlTestSuite * inSuite, void * inContext) {}
+TEST_F(TestChipCryptoPAL, TestAddEntropySources)
```
andy31415, line 1286: can we just delete these? I think they may have existed to make sure that the NL-unittest list contained some values that are valid always, however for gtest it looks like declaration via macro is marking the test as existing, so we do not need empty/noop entries.

src/crypto/tests/TestChipCryptoPAL.cpp
jlatusek: Fixed

src/crypto/tests/TestChipCryptoPAL.cpp
jlatusek: You are right; I've removed empty tests

src/crypto/tests/TestPersistentStorageOpKeyStore.cpp
```diff
@@ -130,7 +129,17 @@ class TestOperationalKeystore final : public Crypto::OperationalKeystore
     FabricIndex mUsedFabricIndex = 0;
 };
 
-void TestBasicLifeCycle(nlTestSuite * inSuite, void * inContext)
+struct TestPersistentStorageOpKeyStore : public ::testing::Test
+{
+    void SetUp() override
```
arkq, line 134: I think that it might be better to initialize memory per entire testsuite, not per test. Also, use `ASSERT_*` in SetUp* functions.

src/messaging/tests/TestExchangeMgr.cpp
```diff
@@ -40,14 +40,28 @@
...
 
 using namespace chip;
 using namespace chip::Inet;
 using namespace chip::Transport;
 using namespace chip::Messaging;
 
-using TestContext = Test::LoopbackMessagingContext;
+struct TestContext : Test::LoopbackMessagingContext
+{
+    virtual CHIP_ERROR SetUp()
```
arkq, line 56: Why there is a `SetUp` without `TearDown` ?

arkq (approved):

src/messaging/tests/TestExchangeMgr.cpp
arkq: Adding a `TODO` might be nice, though.

src/crypto/tests/TestPersistentStorageOpKeyStore.cpp
jlatusek: Fixed

andy31415 (approved):

PR #33043, opened 2024-04-18
ESP32: Add UAT button for lit-icd-app

wqx6:
Add UAT button for ESP32 lit-icd-app.

When pressing the UAT button, the sleep device will be waked up and transition to Active Mode for `CONFIG_ICD_ACTIVE_MODE_THRESHOLD_MS` ms and then return to Idle Mode again.

### DIFF
--- a/examples/lit-icd-app/esp32/main/main.cpp
+++ b/examples/lit-icd-app/esp32/main/main.cpp
@@ -15,6 +15,7 @@
  *    limitations under the License.
  */
 
+#include "app/icd/server/ICDNotifier.h"
 #include "esp_log.h"
 #include "esp_netif.h"
 #include "esp_system.h"
@@ -84,7 +85,7 @@ static AppDeviceCallbacks EchoCallbacks;
 static void UatButtonHandler(UatButton * button)
 {
     DeviceLayer::PlatformMgr().ScheduleWork([](intptr_t) {
-        Server::GetInstance().GetICDManager().UpdateOperationState(app::ICDManager::OperationalState::ActiveMode);
+        app::ICDNotifier::GetInstance().NotifyNetworkActivityNotification();
     });
 }
 
--- a/examples/lit-icd-app/esp32/main/main.cpp
+++ b/examples/lit-icd-app/esp32/main/main.cpp
@@ -24,6 +24,8 @@
 #include "nvs_flash.h"
 
 #include <DeviceCallbacks.h>
+#include <IcdUatButton.h>
+#include <app/icd/server/ICDManager.h>
 #include <app/server/Server.h>
 #include <common/CHIPDeviceManager.h>
 #include <common/Esp32AppServer.h>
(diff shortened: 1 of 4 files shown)


andy31415 (approved):

examples/platform/esp32/common/Esp32AppServer.cpp
```diff
@@ -158,17 +158,19 @@ void Esp32AppServer::Init(AppDelegate * sAppDelegate)
...
+#if CONFIG_TEST_EVENT_TRIGGER_ENABLED
     if (hex_string_to_binary(CONFIG_TEST_EVENT_TRIGGER_ENABLE_KEY, sTestEventTriggerEnableKey,
                              sizeof(sTestEventTriggerEnableKey)) == 0)
     {
         ESP_LOGE(TAG, "Failed to convert the EnableKey string to octstr type value");
         memset(sTestEventTriggerEnableKey, 0, sizeof(sTestEventTriggerEnableKey));
     }
     static SimpleTestEventTriggerDelegate sTestEventTriggerDelegate{};
-    static OTATestEventTriggerHandler sOtaTestEventTriggerHandler{};
     VerifyOrDie(sTestEventTriggerDelegate.Init(ByteSpan(sTestEventTriggerEnableKey)) == CHIP_NO_ERROR);
```
mkardous-silabs, line 169: as a side comment, this isn't spec compliant.

examples/lit-icd-app/esp32/main/main.cpp
```diff
@@ -66,6 +81,13 @@ extern const char TAG[] = "lit-icd-app";
 
 static AppDeviceCallbacks EchoCallbacks;
 
+static void UatButtonHandler(UatButton * button)
+{
+    DeviceLayer::PlatformMgr().ScheduleWork([](intptr_t) {
+        Server::GetInstance().GetICDManager().UpdateOperationState(app::ICDManager::OperationalState::ActiveMode);
+    });
+}
```
mkardous-silabs, line 89: This isn't the right way to interact with the ICDManager. We should instead use the `ICDNotifier` which allows to notify the ICDManager of outside events.

`NotifyNetworkActivityNotification` is the function to use until the UAT one is created.

yunhanw-google (approved):

examples/lit-icd-app/esp32/main/main.cpp
wqx6: Changed.

examples/platform/esp32/common/Esp32AppServer.cpp
wqx6: https://github.com/project-chip/connectedhomeip/blob/master/src/app/server/Server.cpp#L353 uses the `testEventTriggerDelegate` in the `SeviceInitParams`. So I have to enable test event trigger to use ICD feature. Otherwise the device will crash as `mTestEventTriggerDelegate` will be NULL.

mkardous-silabs: I was reffering to the key being in RAM and set this way. Spec requires the device to get the key from storage for certification as to not require the device to recomiled after certification since that is prohibited.

Havin the test event trigger enable for the ICD use-case is correct.

mkardous-silabs (approved):

examples/platform/esp32/common/Esp32AppServer.cpp
wqx6: OK ,I will raise another PR for it.

PR #33046, opened 2024-04-18
Replace nl-unit-test with pigweed for system

jlatusek:
## Problem

## Changes
These changes are another piece of replacing the nl-unit test with the pigweed-unit test. I changed tests for the Crypto. I didn't introduce any changes to the logic of tests; I only replaced one framework with another. 

There was a problem with adding pigweed-unit-test as dependencies for `src/system/SystemPacketBuffer.h`. For now, I've manually created the proper friend class name instead of using `FRIEND_TEST` from Pigweed framework


src/system/SystemPacketBuffer.h
```diff
@@ -392,7 +390,31 @@ class DLL_EXPORT PacketBuffer : private pbuf
     const uint8_t * ReserveStart() const;
 
     friend class PacketBufferHandle;
-    friend class ::PacketBufferTest;
+    friend class TestSystemPacketBuffer;
```
andy31415, line 393: Can we do something else here? This is terrible... maybe some test class that is a friend and exposes any getters that others need. listing every single test class as a friend is ugly.

jlatusek: ok, I've used macro `TEST_F_FROM_FIXTURE`, which allows the implementation of test functionality in the base class. It will limit the number of friend classes.

arkq (approved):

andy31415 (approved):

PR #33048, opened 2024-04-18
converting 3 minimal mDNS unit tests from NL_unit to pw_unit

Alami-Amine:
This is based on the effort of issue #29342


src/lib/dnssd/minimal_mdns/core/tests/TestQName.cpp
```diff
@@ -40,299 +38,263 @@ static SerializedQNameIterator AsSerializedQName(const uint8_t (&v)[N])
...
     {
         static const uint8_t kOneItem[] = "\04test\00";
         SerializedQNameIterator it      = AsSerializedQName(kOneItem);
 
-        NL_TEST_ASSERT(inSuite, it.Next());
-        NL_TEST_ASSERT(inSuite, strcmp(it.Value(), "test") == 0);
-        NL_TEST_ASSERT(inSuite, !it.Next());
-        NL_TEST_ASSERT(inSuite, it.IsValid());
+        EXPECT_TRUE(it.Next());
+        EXPECT_FALSE(strcmp(it.Value(), "test"));
```
andy31415, line 49: although tehcnically true, this reeds ugly as "expect false it compared with test" when the actual test is "expect value to be test"). Prefer for strcmp to use "EXPECT_EQ(...., 0)" so that we are not pulled into reading the "FALSE" since it is not as fast to realize "false means they are equal".

we could maybe even create a macro or function so we can write `EXPECT_STREQUAL(it.Value(), "test")` which would be significantly more readable.

andy31415: https://pigweed.dev/pw_unit_test/#c.EXPECT_STREQ exists

Please update things to `EXPECT_STREQ(....)`

src/lib/dnssd/minimal_mdns/core/tests/TestRecordWriter.cpp
```diff
@@ -17,16 +17,19 @@
  */
 
 #include <lib/dnssd/minimal_mdns/core/RecordWriter.h>
-#include <lib/support/UnitTestRegistration.h>
+//#include <lib/support/UnitTestRegistration.h>
```
andy31415, line 20: delete commented out text instead of just commenting it out. If we need history, we can use git history. Applies throughout.

andy31415 (approved):

bzbarsky-apple (approved):

andy31415:
#33069 seems to be the equivalent full conversion. Since this PR seems to have timed-out in some CI, closing this and keeping the other complete conversion for review.

PR #33050, opened 2024-04-18
Replace some usages of chip::Optional with std::optional

andy31415:
Trying to use the standard library as much as possible instead of writing our own (especially since our own seems to have some bugs regarding assignment: chip::Optional supports = even if the underlying type does not support it).

Looking to see the impact on code size from making a change.

### DIFF
--- a/src/lib/core/Optional.h
+++ b/src/lib/core/Optional.h
@@ -214,10 +214,7 @@ public:
 
     std::optional<T> std_optional() const
     {
-        if (!HasValue())
-        {
-            return std::optional<T>();
-        }
+        VerifyOrReturnValue(HasValue(), std::nullopt);
         return std::make_optional(Value());
     }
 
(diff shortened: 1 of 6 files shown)


src/lib/core/Optional.h
```diff
@@ -211,6 +212,12 @@ class Optional
     bool operator==(const T & other) const { return HasValue() && Value() == other; }
     bool operator!=(const T & other) const { return !(*this == other); }
 
+    std::optional<T> std_optional() const
```
andy31415, line 215: not an operator on purpose to make all casts explicit, hoping that in time we are able to convert everything and do not rely on implicit casts (and then we get full incremental updates).

cecille (approved):

jmartinez-silabs (approved):

Damian-Nordic (approved):

PR #33053, opened 2024-04-18
[TI] Update Number of LWIP Buffers, Fix ICD GPIO build error

adabreuti:
Update memory interface to support realloc, decrease num message buffers Fix ICD LED Build failures

Validated correct response from scan-network command and that all CC13x4 platforms builds.


src/platform/cc13xx_26xx/memory.c
```diff
@@ -0,0 +1,357 @@
...
+ */
+typedef union Header
+{
+    struct
+    {
+        void * actualBuf;
+        size_t size;
+    } header;
+    int pad[2]; /* 4 words on 28L, 8 bytes on most others */
+} Header;
```
andy31415, line 94: would some static_assert regarding sizes make sense and enforce limits here?

andy31415 (approved):

src/platform/cc13xx_26xx/memory.c
adabreuti: I don't think it's necessary since it's only for the 13x4 platform and the advanced feature set isn't currently being exercised so this additional header is currently unused.

jmartinez-silabs (approved):

PR #33058, opened 2024-04-19
Make ClusterStateCache use less memory when not storing attribute values.

bzbarsky-apple:
ClusterStateCache has a "no storing values" mode, but in that mode it still uses 24 bytes per attribute to store just the data size.  For a typical device with a few hundred attributes, this adds up to multiple KB of RAM; across many devices it really starts to add up.

Introduce a compile-time switch that lets us optimize the storage when we know we're not storing the data.

This still keeps the old "enable storing data at compile time but turn it off at runtime" API, both for API compat and because it might reduce codesize for some consumers who care about that more than RAM.

### DIFF
--- a/src/app/ClusterStateCache.cpp
+++ b/src/app/ClusterStateCache.cpp
@@ -27,14 +27,14 @@ namespace app {
 namespace {
 
 // Determine how much space a StatusIB takes up on the wire.
-size_t SizeOfStatusIB(const StatusIB & aStatus)
+uint32_t SizeOfStatusIB(const StatusIB & aStatus)
 {
     // 1 byte: anonymous tag control byte for struct.
     // 1 byte: control byte for uint8 value.
     // 1 byte: context-specific tag for uint8 value.
     // 1 byte: the uint8 value.
     // 1 byte: end of container.
-    size_t size = 5;
+    uint32_t size = 5;
 
     if (aStatus.mClusterStatus.HasValue())
     {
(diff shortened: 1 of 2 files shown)


woody-apple (approved):

src/app/ClusterStateCache.cpp
```diff
@@ -418,23 +464,32 @@ void ClusterStateCache::OnEventData(const EventHeader & aEventHeader, TLV::TLVRe
     mCallback.OnEventData(aEventHeader, apData ? &dataSnapshot : nullptr, apStatus);
 }
 
-CHIP_ERROR ClusterStateCache::GetStatus(const ConcreteAttributePath & path, StatusIB & status) const
+template <bool CanEnableDataCaching>
```
andy31415, line 467: This whole function is one big constexpr ifdef. Would it be maybe easier to read (one less indent) to specialize directly like:

```cpp
template<>
CHIP_ERROR ClusterStateCacheT<false>::GetStatus(const ConcreteAttributePath & path, StatusIB & status) const
{
}

template<>
CHIP_ERROR ClusterStateCacheT<true>::GetStatus(const ConcreteAttributePath & path, StatusIB & status) const
{
    CHIP_ERROR err;
    auto attributeState = GetAttributeState(path.mEndpointId, path.mClusterId, path.mAttributeId, err);
    ReturnErrorOnFailure(err);
    // ...
}
```

bzbarsky-apple: Yes, I can do that.

bzbarsky-apple: Done.

andy31415 (approved):

PR #33060, opened 2024-04-19
ESP32: Drop Invalid RA packets

wqx6:
When `ra_recv_handler` receives an invalid message with an option length equal to 0, it will enter into an infinite loop, consequently causing the LwIP task to become unresponsive.


andy31415 (approved):

andy31415:
@wqx6 you may want to cherrypick this to v1.3-branch or similar - we do not auto-sync with master for them, however a potential infinite loop seems important to fix.

wqx6:
> @wqx6 you may want to cherrypick this to v1.3-branch or similar - we do not auto-sync with master for them, however a potential infinite loop seems important to fix.

Yes, we will raise the cherry-pick PR to cherrypick this to the previous released branches.

PR #33065, opened 2024-04-19
[Infineon] Update CYW30739 to support CYW30739B2-P5-EVK-XX boards.

hsusid:
* Replace CYW30739 submodules with the matter-wpan-sdk submodule.
* Update README files for CYW30739 apps.
* Add CYW30739B2-P5-EVK-01, CYW30739B2-P5-EVK-02, and CYW30739B2-P5-EVK-03 boards for CYW30739.
* Add development DAC files for CYW30739.
* Add cyw30739_example, cyw30739_platform, and ifx_openthread template for board dependant targets.
* Update GN files for CYW30739 apps.
* Update CYW30739 builder scripts.
* Update checkout_submodules.py script to support the recursive configuration in the .gitmodules file. (split out to
  https://github.com/project-chip/connectedhomeip/pull/33089)
* Update the CYW30739 flashing script.
* Update the workflow to build all CYW30739-supported targets.
* Move shared codes to the matter_config.cpp file.
* Implemented FactoryDataProvider for CYW30739 to load factory data.
* Remove the CYW30739 ota-requestor-app.


scripts/checkout_submodules.py
```diff
@@ -101,9 +105,16 @@ def checkout_modules(modules: list, shallow: bool, force: bool, recursive: bool,
     cmd += ['--force'] if force else []
     cmd += ['--recursive'] if recursive else []
     cmd += ['--jobs', f'{jobs}'] if jobs else []
-    cmd += [module.path for module in modules]
+    module_paths = [module.path for module in modules]
 
-    subprocess.check_call(cmd)
+    subprocess.check_call(cmd + module_paths)
+
+    if not recursive:
```
andy31415, line 112: could we split out this change separately? I would like to review it and merge it as a stand-alone.

My first instinct reading just this like is "why do we have if not recursive add recursive" .... seems somehow reversed.

hsusid: I split this change out to https://github.com/project-chip/connectedhomeip/pull/33089 and changed some expressions to make it easier to read.

hsusid:
Rebase to clear conflicts.

jmartinez-silabs (approved):

.github/workflows/examples-infineon.yaml
```diff
@@ -107,41 +107,82 @@ jobs:
                       out/artifacts/infineon-psoc6-light/chip-psoc6-lighting-example.out \
                       /tmp/bloat_reports/
 
-            - name: Build example CYW30739 Apps
+            - name: Build some CYW30739 Lighting Apps
```
andy31415, line 110:
```suggestion
            - name: Build CYW30739 Lighting App
```

.github/workflows/examples-infineon.yaml
```diff
@@ -107,41 +107,82 @@ jobs:
...
+                  .environment/pigweed-venv/bin/python3 scripts/tools/memory/gh_sizes.py \
+                    cyw30739 CYW30739B2-P5-EVK-02 light \
+                    out/artifacts/cyw30739-light/lighting_app-CYW30739B2-P5-EVK-02.elf \
                     /tmp/bloat_reports/
-            - name: Get lock size stats
-              run: |
                   .environment/pigweed-venv/bin/python3 scripts/tools/memory/gh_sizes.py \
-                    cyw30739 cyw930739m2evb_01 lock \
-                    out/artifacts/cyw30739-cyw930739m2evb_01-lock/chip-cyw30739-lock-example.elf \
+                    cyw30739 CYW30739B2-P5-EVK-03 light \
```
andy31415, line 129: does a single target generated several build outputs? why don't we have `cyw30739-p5-evk-03-light` target to build specific things? I find it surprising that a single build will generate so many elf files.

Could you explain the build process a bit to help me understand?

examples/light-switch-app/infineon/cyw30739/BUILD.gn
```diff
@@ -12,71 +12,54 @@
...
     ]
+
+    include_dirs = [ "include" ]
   }
 }
 
-group("cyw30739") {
-  deps = [ ":light_switch_app" ]
+foreach(board, matter_wpan_sdk_board_list) {
+  matter_wpan_app("${app_name}-${board}") {
```
andy31415, line 56: why do we build several apps in parallel? How about specifying the board itself as part of the build configuration?

andy31415 (changes requested):
Having each target compile every possible supported board seems off. We can do that as part of CI, however auto-selecting all boards as part of the build seems strange.

.github/workflows/examples-infineon.yaml
hsusid: Thank you for having reviewed it.
A brief explanation of the build process would be:
1. GN uses the `matter_wpan_sdk_board_list` variable in `examples/<app-name>/infineon/cyw30739/args.gni` to generate build rules for all supported boards into ninja files.
2. A `ninja -C <out>` command would build all supported boards.

In this change, I make the CYW30739 builder build all supported boards without specifying the board name. I expected CI to build all CYW30739-supported boards by default.

examples/light-switch-app/infineon/cyw30739/BUILD.gn
hsusid: My goal is to minimize the increase in CI building time when I add more supported boards on the CYW30739 platform.
If we specified the board as part of the build configuration, the ninja would repeatedly build the whole Matter stack across every supported board.
But if we use the board as one of the GN targets, the GN would help us finger out the dependencies among app-specific codes, Matter stack codes, and board-specific codes. As a result, the ninja would only build board-independent parts of codes once, reducing the build time considerably.
In the CYW30739 case, building an app for a single board requires about 5 minutes, building an app for three different boards separately requires about 15 minutes, and building an app with three different boards simultaneously only requires about 7 minutes.

hsusid:
> Having each target compile every possible supported board seems off. We can do that as part of CI, however auto-selecting all boards as part of the build seems strange.

We didn't remove the ability to let the developer build only a specified board. We documented how to do it in README files. The below instruction is a part of the CYW30739 app README file:
```bash
$ cd ~/connectedhomeip
$ source scripts/activate.sh
$ gn gen --root=examples/lighting-app/infineon/cyw30739 out/cyw30739-light
$ ninja -C out/cyw30739-light [lighting_app-BOARD_NAME]
```

I also noticed that the `NINJA_ARGS` environment in `scripts/examples/gn_build_example.sh` seems to provide the same capability of specifying the ninja build target.

.github/workflows/examples-infineon.yaml
hsusid: Fixed.

.github/workflows/examples-infineon.yaml
hsusid: updated.

andy31415 (approved):

examples/light-switch-app/infineon/cyw30739/BUILD.gn
andy31415: I think long term this may not be workable - if you have more boards they would at some point differ under `/platform` in some way more than just peripherals seen by the app. This is why I would like top-level to keep board selection clear even if a single target can build multiple boards.

PR #33073, opened 2024-04-19
Convert a few more items to std::optional

andy31415:
To assess the impact of such a conversion.

I expect at some point we have to do some different mass conversion by aliasing chip::Optional to std::optional and see the sideffects (and keep a legacy copy for incompatible items).

### DIFF
--- a/examples/light-switch-app/asr/src/BindingHandler.cpp
+++ b/examples/light-switch-app/asr/src/BindingHandler.cpp
@@ -280,7 +280,7 @@ void BindingHandler::PrintBindingTable()
             \t+ ClusterId %d \n \
             \t+ RemoteEndpointId %d \n \
             \t+ NodeId %d",
-                    (int) entry.fabricIndex, (int) entry.local, (int) entry.clusterId.value_or(-1), (int) entry.remote,
+                    (int) entry.fabricIndex, (int) entry.local, (int) entry.clusterId.value_or(kInvalidClusterId), (int) entry.remote,
                     (int) entry.nodeId);
             break;
         case MATTER_MULTICAST_BINDING:
--- a/examples/light-switch-app/asr/src/BindingHandler.cpp
+++ b/examples/light-switch-app/asr/src/BindingHandler.cpp
@@ -280,8 +280,8 @@ void BindingHandler::PrintBindingTable()
             \t+ ClusterId %d \n \
             \t+ RemoteEndpointId %d \n \
             \t+ NodeId %d",
-                    (int) entry.fabricIndex, (int) entry.local, (int) entry.clusterId.value_or(kInvalidClusterId), (int) entry.remote,
-                    (int) entry.nodeId);
+                    (int) entry.fabricIndex, (int) entry.local, (int) entry.clusterId.value_or(kInvalidClusterId),
+                    (int) entry.remote, (int) entry.nodeId);
             break;
         case MATTER_MULTICAST_BINDING:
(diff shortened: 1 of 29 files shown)


arkq (approved):

examples/light-switch-app/asr/src/BindingHandler.cpp
```diff
@@ -280,7 +280,7 @@ void BindingHandler::PrintBindingTable()
             \t+ ClusterId %d \n \
             \t+ RemoteEndpointId %d \n \
             \t+ NodeId %d",
-                    (int) entry.fabricIndex, (int) entry.local, (int) entry.clusterId.Value(), (int) entry.remote,
+                    (int) entry.fabricIndex, (int) entry.local, (int) entry.clusterId.value_or(-1), (int) entry.remote,
```
cecille, line 283: so previously, did this just crash if there was no value?

cecille (approved):

examples/light-switch-app/asr/src/BindingHandler.cpp
bzbarsky-apple: Sure looks like it....

bzbarsky-apple, line 283:
```suggestion
                    (int) entry.fabricIndex, (int) entry.local, (int) entry.clusterId.value_or(kInvalidClusterId), (int) entry.remote,
```
because `-1` is not even the right type for a `ClusterId`.

examples/light-switch-app/infineon/cyw30739/src/BindingHandler.cpp
```diff
@@ -306,7 +306,7 @@ void BindingHandler::PrintBindingTable()
             \t+ ClusterId %d \n \
             \t+ RemoteEndpointId %d \n \
             \t+ NodeId %d \n",
-                   (int) entry.fabricIndex, (int) entry.local, (int) entry.clusterId.Value(), (int) entry.remote,
+                   (int) entry.fabricIndex, (int) entry.local, (int) entry.clusterId.value_or(-1), (int) entry.remote,
```
bzbarsky-apple, line 309:
```suggestion
                   (int) entry.fabricIndex, (int) entry.local, (int) entry.clusterId.value_or(kInvalidClusterId), (int) entry.remote,
```

examples/light-switch-app/nrfconnect/main/BindingHandler.cpp
```diff
@@ -284,7 +284,7 @@ void BindingHandler::PrintBindingTable()
             \t+ ClusterId %d \n \
             \t+ RemoteEndpointId %d \n \
             \t+ NodeId %d",
-                    (int) entry.fabricIndex, (int) entry.local, (int) entry.clusterId.Value(), (int) entry.remote,
+                    (int) entry.fabricIndex, (int) entry.local, (int) entry.clusterId.value_or(-1), (int) entry.remote,
```
bzbarsky-apple, line 287:
```suggestion
                    (int) entry.fabricIndex, (int) entry.local, (int) entry.clusterId.value_or(kInvalidClusterId), (int) entry.remote,
```

examples/light-switch-app/asr/src/BindingHandler.cpp
andy31415: It looks like yes ... we maybe knew this will always have a value, however I did not check. I figured I may as well replace with value_or.

bzbarsky-apple (approved):

PR #33080, opened 2024-04-19
Use std::optional for chip::Nullable

andy31415:
This makes chip::Nullable use std::optional instead of chip::Optional.

Api is compatible, however move destructor calling is slightly different. Tests pass, however move constructor/assignment tests were updated.


mkardous-silabs (approved):

src/app/data-model/Nullable.h
```diff
@@ -37,29 +37,37 @@ inline constexpr auto NullNullable = NullOptional;
...
-    using Optional<T>::ValueOr;
+    using std::optional<T>::optional;
+    using std::optional<T>::value_or;
+    using std::optional<T>::value;
+    using std::optional<T>::has_value;
+    using std::optional<T>::operator*;
+    using std::optional<T>::operator->;
+
+    // backwards compatibiltiyt with old chip::Optional functionality
+    // NOTE: as we transition to std::optional, these should be removed
```
bzbarsky-apple, line 55: No, I don't think they in fact should be.  We want to be very clear on where things are optional and where they are nullable, especially for things that are both optional and nullable.

This is why Nullable exists at all as a separate type.

src/app/data-model/Nullable.h
```diff
@@ -96,22 +104,29 @@ struct Nullable : protected Optional<T>
...
+    {
+        if (!has_value())
+        {
+            return !other.has_value();
+        }
+        return other.has_value() && (*other == **this);
+    }
+    bool operator!=(const Nullable<T> & other) const { return !(*this == other); }
+    bool operator==(const T & other) const { return has_value() && (**this == other); }
+    bool operator!=(const T & other) const { return !has_value() || (**this != other); }
```
bzbarsky-apple, line 117: Why not implement this as `!(*this == other)` like the other operator!=?

src/app/tests/TestNullable.cpp
```diff
@@ -165,24 +169,26 @@ static void TestMove(nlTestSuite * inSuite, void * inContext)
     CtorDtorCounter::ResetCounter();
 
     {
-        auto testSrc = MakeNullable<MovableCtorDtorCounter>(400);
-        Nullable<MovableCtorDtorCounter> testDst(std::move(testSrc));
-        NL_TEST_ASSERT(inSuite, CtorDtorCounter::created == 2 && CtorDtorCounter::destroyed == 1);
+        auto testSrc = MakeNullable<MovableCtorDtorCounter>(400);     // contstruct
```
bzbarsky-apple, line 172:
```suggestion
        auto testSrc = MakeNullable<MovableCtorDtorCounter>(400);     // construct
```

bzbarsky-apple (changes requested):

src/app/data-model/Nullable.h
```diff
@@ -37,29 +37,37 @@ inline constexpr auto NullNullable = NullOptional;
...
-    using Optional<T>::Optional;
-
-    // Pull in APIs that make sense on Nullable with the same names as on
-    // Optional.
-    using Optional<T>::Value;
-    using Optional<T>::ValueOr;
+    using std::optional<T>::optional;
+    using std::optional<T>::value_or;
+    using std::optional<T>::value;
+    using std::optional<T>::has_value;
```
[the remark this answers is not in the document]
bzbarsky-apple: And I'm not sure whether we should try to be API-compatible with `std::optional` here in terms of `value_or` and `has_value`, for that matter.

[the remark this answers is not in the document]
andy31415: sure, updated. I think I had this to make code shorter in the operator==. Agree not needed since we seem to hide the optional bits on purpose.

src/app/data-model/Nullable.h
andy31415: I think we should keep `NullOptionalType` hnowever for the rest I think we should replace `Value` with `value` and `ValueOr` with `value_or` as they seem to make sense and this is what we did with optional as well.

I will move the constructor to not be attached to this comment.

bzbarsky-apple: > and this is what we did with optional as well

We did it with optional because we want to be API-compatible with std::optional there, with an eye to replacing it, yes?

We are not planning to replace Nullable with std::optional.  We shouldn't churn things just to make it look like std::optional when they are not in fact the same thing... and if it's a Matter-defined type, its API should look like other Matter types.

src/app/data-model/Nullable.h
```diff
@@ -96,22 +98,39 @@ struct Nullable : protected Optional<T>
...
+        return other.has_value() && (*other == **this);
+    }
+    bool operator!=(const Nullable<T> & other) const { return !(*this == other); }
+    bool operator==(const T & other) const { return std::optional<T>::has_value() && (**this == other); }
+    bool operator!=(const T & other) const { return !(*this == other); }
+
+    // backwards compatibility with old chip::Nullable method names
+    //
+    // NOTE: as we transition to std::optional, these should be removed
+    // We expect only `value` and `value_or` to remain as standard names
```
bzbarsky-apple, line 116: Again, I disagree with that expectation.  Why is that the expectation?

andy31415: We seem to have taken `Value` and `ValueOr` as being useful to be taken from chip::Optional. If we are to convert chip::Optional to std::optional it seems to make sense to say "we want to preserve value and value_or" as useful from std::optional.

I am thinking:
   - value and valueor (regardless of syntax) are useful and should exist
   - a `using std::optional<T>::value` seems to likely bring in a more complete declaration of value/value_or than us trying to expose them with a different name
   
 So I have to do extra code to try to expose `Value` instead of `value` and I do not see the benefit: why is it important to name it `Value/ValueOr` instead of `value/value_or`? I agree on "IsNull" is clearer than a "has_value/HasValue" ... however I do not understand the value/valueor bit.
 
 Could you explain your line of thinking?

andy31415: I guess TLDR is I prefer:

```cpp
using std::optional<T>::value_or;
```

over

```cpp
    template <typename... Args>
    auto ValueOr(Args &&... args) -> decltype(std::optional<T>::value_or(std::forward<Args>(args)...))
    {
        return std::optional<T>::value_or(std::forward<Args>(args)...);
    }

    template <typename... Args>
    auto ValueOr(Args &&... args) const -> decltype(std::optional<T>::value_or(std::forward<Args>(args)...))
    {
        return std::optional<T>::value_or(std::forward<Args>(args)...);
    }
```

maybe it can be simplified, however it is ugly and even then I am not sure I covered all the cases transparrently. the using is much simpler.

bzbarsky-apple: It can definitely be simplified, in that the whole decltype thing is not needed due to type deduction for functions with auto return, and it  covers all cases transparently due to perfect forwarding (though you have to match the actual signatures of value_or, which does not have a non-const version for obvious reasons, but does have a move version which we may or may not care about).  That is you can do:

```
    template <typename... Args>
    constexpr auto ValueOr(Args &&... args) const
    {
        return std::optional<T>::value_or(std::forward<Args>(args)...);
    }
```

and that's all you need for the use cases we have in our tree right now.  Similar for `Value()`.

That said, there is an interesting and subtle difference between `value_or` for `std::optional` and `value_or` for `chip::Optional` and `Nullable` which I just noticed when I was looking into this: `std::optional`'s `value_or` returns by value (makes a copy), whereas `chip::Optional`'s `ValueOr` returns by reference.  Maybe that's an OK change in practice, but we need to watch out for it carefully for non-primitive types, where we might start copying a bunch of stuff we didn't use to.

Back to the "why do we want this?" question: it's a matter of API consistency, basically.  Why `value()` but `IsNull()`?  The inconsistency just looks odd, and to me ironing out the oddness is worth the admittedly Perl-like line of perfect forwarding variadic template.

Since for now we need the aliases anyway, here is my proposal: let's not add the public `value` and `value_or` for now, not add comments about how the existing bits should be removed (again, for now), and not block this PR on this discussion, which we can continue as desired.

bzbarsky-apple: Also, importantly: I _really_ want the fact that we are using `std::optional` here to be an implementation detail that does not leak through the API, if possible...

andy31415: Updated to only expose Value/ValueOr as the public API and not value/value_or.

I kept the expose of `->` and `*` as unchecked value versions as those seem to be capable to provide slight flash and readability benefits since most of our code does contain `IsNull` guards.

src/app/data-model/Nullable.h
andy31415: The main difficulty will be that we will change a lot of files, so I expect a sed of `%s/ValueOr/value_or` and if the API surface between nullable and optional differs, that replacement will be a pain to only apply to optional but not nullable. 

We can defer that to the chip::Optional replacement though. PR is now updated.

src/app/data-model/Nullable.h
andy31415: Had to make `Value` use operator* so that we get a reference rather than a copy. This has the sideffect of making Value unchecked (likely less flash, however it could be used unsafely).

src/app/data-model/Nullable.h
```diff
@@ -37,31 +37,40 @@ inline constexpr auto NullNullable = NullOptional;
...
     {
-        return Optional<T>::Emplace(std::forward<Args>(args)...);
+        return std::optional<T>::emplace(std::forward<Args>(args)...);
+    }
+
+    template <typename... Args>
+    constexpr auto ValueOr(Args &&... args) const
+    {
+        return std::optional<T>::value_or(std::forward<Args>(args)...);
     }
```
Damian-Nordic, line 69:
```suggestion
    template <typename Arg>
    constexpr auto ValueOr(Arg && arg) const
    {
        return std::optional<T>::value_or(std::forward<Arg>(arg));
    }
```
value_or takes a single argument, right?

src/app/tests/TestNullable.cpp
```diff
@@ -165,24 +169,26 @@ static void TestMove(nlTestSuite * inSuite, void * inContext)
     CtorDtorCounter::ResetCounter();
 
     {
-        auto testSrc = MakeNullable<MovableCtorDtorCounter>(400);
-        Nullable<MovableCtorDtorCounter> testDst(std::move(testSrc));
-        NL_TEST_ASSERT(inSuite, CtorDtorCounter::created == 2 && CtorDtorCounter::destroyed == 1);
+        auto testSrc = MakeNullable<MovableCtorDtorCounter>(400);     // construct
+        Nullable<MovableCtorDtorCounter> testDst(std::move(testSrc)); // move construct
+        NL_TEST_ASSERT(inSuite, CtorDtorCounter::created == 2 && CtorDtorCounter::destroyed == 0);
```
Damian-Nordic, line 174: Hmm. That's quite significant behavior change between `std::optional` and `Optional`. The former doesn't null out the moved optional/nullable while the latter does. I hope our code doesn't rely on this :). We could override this behavior for our Optional/Nullable but now that we started recommending `std::optional` it's probably better to align to the standard...

Damian-Nordic (approved):

src/app/tests/TestNullable.cpp
andy31415: I think std::optional does the right thing. After a move the guarantee is "valid state, however undefined" so code should not rely that the value is usable after move.

src/app/data-model/Nullable.h
andy31415: I was unsure if this would trigger perfect forwarding or not, hence the entire thing. I can apply the change though and see if things compile still.

src/app/data-model/Nullable.h
andy31415: Was looking at the wrong compiler options ... can use `value` just fine. Updated back to it instead of operator*.

src/app/clusters/ota-requestor/ota-requestor-server.cpp
bzbarsky-apple, line 1: Why the random whitespace change to this file?

src/app/data-model/Nullable.h
```diff
@@ -37,31 +37,40 @@ inline constexpr auto NullNullable = NullOptional;
...
+struct Nullable : protected std::optional<T>
 {
+
     //
     // The following 'using' statement is needed to make visible
     // all constructors of the base class within this derived class.
     //
-    using Optional<T>::Optional;
+    using std::optional<T>::optional;
+    using std::optional<T>::operator*;
```
bzbarsky-apple, line 48: Do we want to pull in `operator*`?  I guess we can...  I worry about people misusing the unchecked thing.

src/app/data-model/Nullable.h
```diff
@@ -37,31 +37,40 @@ inline constexpr auto NullNullable = NullOptional;
...
     constexpr T & SetNonNull(Args &&... args)
     {
-        return Optional<T>::Emplace(std::forward<Args>(args)...);
+        return std::optional<T>::emplace(std::forward<Args>(args)...);
+    }
+
+    template <typename Arg>
+    constexpr auto ValueOr(Arg && arg) const
+    {
+        return std::optional<T>::value_or(std::forward<Arg>(arg));
```
bzbarsky-apple, line 68:
```suggestion
    template <typename... Args>
    constexpr auto ValueOr(Args &&... args) const
    {
        return std::optional<T>::value_or(std::forward<Args>(args)...);
```

bzbarsky-apple (approved):

src/app/data-model/Nullable.h
andy31415: at least for using std::optional directly, clang-tidy was super aggressive on complaining about unchecked accesses. Beyond that however, this seems to have flash benefits.

src/app/clusters/ota-requestor/ota-requestor-server.cpp
andy31415: I do not know ... reverted.

src/app/data-model/Nullable.h
andy31415: I would lean towards keeping it: std::optional provides this, so it seems others assumed it reasonable enough to have, so we may follow suite.

src/app/data-model/Nullable.h
Damian-Nordic: I actually asked Andy to do the opposite to be clear this takes a single argument :)

bzbarsky-apple: Ah, I see.  I guess I don't feel super-strongly about it....

PR #33081, opened 2024-04-19
[ICD] Update and Document  ICDManager interface

mkardous-silabs:
#### Description
The PRs goal is finalize the public interface provided by the ICDManager. During developmment, most of the functions in the ICDManager were made public without good reason and gave access to internal functions and states that should not have been accessible.

PR also has a bug fix to reset the `OnTransitionToIdle` flag when we enter ActiveMode and not when the IdleMode timer expires. This is the only real functionnal changes. ICDManager.cpp line #468 & #510

- Move all internal function to `private` to prevent consummers from modyfing internal states without going through standard and tested methods.
- Fix when `OnTransitionToIdle` flag was being set / cleared
- Document public APIs to use to get notified of internal changes in the ICDManager
- Fix platforms that were using internal functions

__Note to reviewers__  :

- Transition from nlTest to pw_unit_test will be done in a follow up
- Transition from chip::Optionnal to std::optionnal will be done in a follow up
- OnEnterIdleMode was added even if it is not used in this PR. Utilisation will be done in a follow up
- Two issues were found by doing this work, fix will be done in a follow up. Issue were created as trackers

#### Tests
- Existing tests to validate APIs from public to private
- Added unit tests to validate the ICDStateObserver class

### DIFF
--- a/src/app/icd/server/ICDManager.cpp
+++ b/src/app/icd/server/ICDManager.cpp
@@ -378,12 +378,13 @@ void ICDManager::UpdateICDMode()
     if (ICDConfigurationData::GetInstance().GetICDMode() != tempMode)
     {
         ICDConfigurationData::GetInstance().SetICDMode(tempMode);
-        postObserverEvent(ObserverEventType::ICDModeChange);
 
         // Can't use attribute accessors/Attributes::OperatingMode::Set in unit tests
 #if !CONFIG_BUILD_FOR_HOST_UNIT_TEST
         Attributes::OperatingMode::Set(kRootEndpointId, static_cast<OperatingModeEnum>(tempMode));
 #endif
+
+        postObserverEvent(ObserverEventType::ICDModeChange);
     }
 
     // When in SIT mode, the slow poll interval SHOULDN'T be greater than the SIT mode polling threshold, per spec.
@@ -433,6 +434,8 @@ void ICDManager::UpdateOperationState(OperationalState state)
         {
             ChipLogError(AppServer, "Failed to set Slow Polling Interval: err %" CHIP_ERROR_FORMAT, err.Format());
         }
+
+        postObserverEvent(ObserverEventType::EnterIdleMode);
     }
     else if (state == OperationalState::ActiveMode)
     {
@@ -447,7 +450,7 @@ void ICDManager::UpdateOperationState(OperationalState state)
 
             if (activeModeDuration == kZero && !mKeepActiveFlags.HasAny())
             {
-                // A Network Activity triggered the active mode and activeModeDuration is 0.
+                // Network Activity triggered the active mode and activeModeDuration is 0.
                 // Stay active for at least Active Mode Threshold.
                 activeModeDuration = ICDConfigurationData::GetInstance().GetActiveModeThreshold();
             }
@@ -455,9 +458,14 @@ void ICDManager::UpdateOperationState(OperationalState state)
             DeviceLayer::SystemLayer().StartTimer(activeModeDuration, OnActiveModeDone, this);
 
             Milliseconds32 activeModeJitterInterval = Milliseconds32(ICD_ACTIVE_TIME_JITTER_MS);
+            // TODO(#33074): Edge case when we transition to IdleMode with this conditio being true
+            // (activeModeDuration == kZero && !mKeepActiveFlags.HasAny())
             activeModeJitterInterval =
                 (activeModeDuration >= activeModeJitterInterval) ? activeModeDuration - activeModeJitterInterval : kZero;
 
+            // Reset this flag when we enter ActiveMode to avoid having a feedback loop that keeps us indefinitly in
+            // ActiveMode.
+            mTransitionToIdleCalled = false;
             DeviceLayer::SystemLayer().StartTimer(activeModeJitterInterval, OnTransitionToIdle, this);
 
             CHIP_ERROR err =
@@ -504,10 +512,6 @@ void ICDManager::OnIdleModeDone(System::Layer * aLayer, void * appState)
 {
     ICDManager * pICDManager = reinterpret_cast<ICDManager *>(appState);
     pICDManager->UpdateOperationState(OperationalState::ActiveMode);
-
-    // We only reset this flag when idle mode is complete to avoid re-triggering the check when an event brings us back to active,
-    // which could cause a loop.
-    pICDManager->mTransitionToIdleCalled = false;
 }
 
 void ICDManager::OnActiveModeDone(System::Layer * aLayer, void * appState)
@@ -532,10 +536,10 @@ void ICDManager::OnTransitionToIdle(System::Layer * aLayer, void * appState)
 }
 
 /* ICDListener functions. */
+
 void ICDManager::OnKeepActiveRequest(KeepActiveFlags request)
 {
     assertChipStackLockedByCurrentThread();
-
     VerifyOrReturn(request < KeepActiveFlagsValues::kInvalidFlag);
 
     if (request.Has(KeepActiveFlag::kExchangeContextOpen))
@@ -560,7 +564,6 @@ void ICDManager::OnKeepActiveRequest(KeepActiveFlags request)
 void ICDManager::OnActiveRequestWithdrawal(KeepActiveFlags request)
 {
     assertChipStackLockedByCurrentThread();
-
     VerifyOrReturn(request < KeepActiveFlagsValues::kInvalidFlag);
 
     if (request.Has(KeepActiveFlag::kExchangeContextOpen))
@@ -695,6 +698,10 @@ void ICDManager::postObserverEvent(ObserverEventType event)
         {
         case ObserverEventType::EnterActiveMode: {
             obs->mObserver->OnEnterActiveMode();
+            return Loop::Continue;
+        }
+        case ObserverEventType::EnterIdleMode: {
+            obs->mObserver->OnEnterIdleMode();
             return Loop::Continue;
         }
         case ObserverEventType::TransitionToIdle: {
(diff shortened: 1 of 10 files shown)


jmartinez-silabs (approved):

src/app/icd/server/ICDManager.cpp
```diff
@@ -447,17 +450,22 @@ void ICDManager::UpdateOperationState(OperationalState state)
...
-                // A Network Activity triggered the active mode and activeModeDuration is 0.
+                // Network Activity triggered the active mode and activeModeDuration is 0.
                 // Stay active for at least Active Mode Threshold.
                 activeModeDuration = ICDConfigurationData::GetInstance().GetActiveModeThreshold();
             }
 
             DeviceLayer::SystemLayer().StartTimer(activeModeDuration, OnActiveModeDone, this);
 
             Milliseconds32 activeModeJitterInterval = Milliseconds32(ICD_ACTIVE_TIME_JITTER_MS);
+            // TODO(#33074): Edge case when we transition to IdleMode with this conditio being true
```
bzbarsky-apple, line 461:
```suggestion
            // TODO(#33074): Edge case when we transition to IdleMode with this condition being true
```

src/app/icd/server/ICDManager.h
```diff
@@ -71,11 +73,14 @@ class ICDManager : public ICDListener, public TestEventTriggerHandler
...
-    // mStateObserverPool for the ICDManager.
+    /**
+     * @brief This enum class represents to all ICDStateObserver callbacks available from the
+     *        mStateObserverPool for the ICDManager.
+     */
     enum class ObserverEventType : uint8_t
     {
         EnterActiveMode,
+        EnterIdleMode,
         TransitionToIdle,
```
bzbarsky-apple, line 84: The difference between EnterIdleMode and TransitionToIdle is extremely non-obvious.  Should be documented.

Similarly, how does ICDModeChange interact with the other events?  Instead of them?  In addition to them?  Something else?  Why does it exist at all?

src/app/icd/server/ICDManager.h
```diff
@@ -85,22 +90,31 @@ class ICDManager : public ICDListener, public TestEventTriggerHandler
...
-    void SetKeepActiveModeRequirements(KeepActiveFlags flag, bool state);
-    bool IsKeepActive() { return mKeepActiveFlags.HasAny(); }
+
+    /**
+     * @brief SupportsFeature verifies if a given FeatureMap bit is enabled
+     *
+     * @param[in] feature FeatureMap bit to verify
+     *
+     * @return true: if the FeatureMap bit is enabled in the ICDM cluster attribute.
+     *         false: ff the FeatureMap bit is not enabled in the ICDM cluster attribute.
```
bzbarsky-apple, line 111:
```suggestion
     *         false: if the FeatureMap bit is not enabled in the ICDM cluster attribute.
```

src/app/icd/server/ICDManager.h
```diff
@@ -160,47 +167,106 @@ class ICDManager : public ICDListener, public TestEventTriggerHandler
...
     void OnKeepActiveRequest(KeepActiveFlags request) override;
     void OnActiveRequestWithdrawal(KeepActiveFlags request) override;
     void OnICDManagementServerEvent(ICDManagementEvents event) override;
     void OnSubscriptionReport() override;
 
-protected:
+private:
     friend class TestICDManager;
+    /**
+     * @brief UpdateICDMode checks in which ICDMode (SIT or LIT) the ICD can go to and updates the mode if necessary.
```
bzbarsky-apple, line 187: I am not sure what "in which mode ... can go to" means.

src/app/icd/server/ICDManager.h
```diff
@@ -160,47 +167,106 @@ class ICDManager : public ICDListener, public TestEventTriggerHandler
...
+     *        For a SIT ICD, the function does nothing.
+     *        For a LIT ICD, the function checks if the ICD has a registration in the ICDMonitoringTable to determine in which
```
bzbarsky-apple, line 189:
```suggestion
     *        For a LIT ICD, the function checks if the ICD has a registration in the ICDMonitoringTable to determine which
```

src/app/icd/server/ICDManager.h
```diff
@@ -160,47 +167,106 @@ class ICDManager : public ICDListener, public TestEventTriggerHandler
...
+     *        ICDMode the ICD must be in.
+     */
+    void UpdateICDMode();
+
+    /**
+     * @brief UpdateOperationState updates the OperationState of the ICD to the requested one.
+     *        IdleMode -> IdleMode     : No actions are necessary, do nothing.
+     *        IdleMode -> ActiveMode   : Transition the device to ActiveMode, start the  ActiveMode timer and triggers all necessary
```
bzbarsky-apple, line 197:
```suggestion
     *        IdleMode -> ActiveMode   : Transition the device to ActiveMode, start the  ActiveMode timer and trigger all necessary
```

Well, or the first part should be "transitions" and "starts", but this seems better.

src/app/icd/server/ICDManager.h
```diff
@@ -160,47 +167,106 @@ class ICDManager : public ICDListener, public TestEventTriggerHandler
...
+     *                                   operations. These operations could be : Send Check-In messages
+     *                                                                           Send subscription reports
+     *                                                                           Process user actions
+     *        ActiveMode -> ActiveMode : Increase remaining ActiveMode timer to one ActiveModeThreshold.
+     *                                   If ActiveModeThreshold is 0, do nothing.
+     *        ActiveMode -> IdleMode   : Transition ICD to IdleMode and starts the IdleMode timer.
```
bzbarsky-apple, line 203:
```suggestion
     *        ActiveMode -> IdleMode   : Transition ICD to IdleMode and start the IdleMode timer.
```

src/app/icd/server/ICDManager.h
```diff
@@ -160,47 +167,106 @@ class ICDManager : public ICDListener, public TestEventTriggerHandler
...
+     *
+     * @param state requested OperationalState for the ICD to transition to
+     */
+    void UpdateOperationState(OperationalState state);
+
+    /**
+     * @brief Set or Remove a keep ActiveMode requirement for the given flag
+     *        If state is true and the ICD is in IdleMode, transition the ICD to ActiveMode
+     *        If state is false and the ICD is in ActiveMode, verifies if we can transition the ICD to IdleMode.
```
bzbarsky-apple, line 212:
```suggestion
     *        If state is false and the ICD is in ActiveMode, check whether we can transition the ICD to IdleMode.
```

src/app/icd/server/ICDStateObserver.h
```diff
@@ -27,13 +27,43 @@
 namespace chip {
 namespace app {
 
+/**
+ * @brief Public API used my the ICDManager to expose when different events occurs.
```
bzbarsky-apple, line 31:
```suggestion
 * @brief Public API used by the ICDManager to expose when different events occur.
```

src/app/icd/server/ICDStateObserver.h
```diff
@@ -27,13 +27,43 @@
...
+ *        ICDManager::RegisterObserver can be used to register as an Observer to be notified when these events occurs.
```
bzbarsky-apple, line 32:
```suggestion
 *        ICDManager::RegisterObserver can be used to register as an Observer to be notified when these events occur.
```

yunhanw-google (approved):

mkardous-silabs:
@bzbarsky-apple resolving the two remaining comments to unblock follow up PR. If the commit doesn't fully address them, i'll open a follow up PR.

src/app/icd/server/ICDManager.h
```diff
@@ -71,11 +73,31 @@ class ICDManager : public ICDListener, public TestEventTriggerHandler
         ActiveMode,
     };
 
-    // This enum class represents to all ICDStateObserver callbacks available from the
-    // mStateObserverPool for the ICDManager.
+    /**
+     * @brief This enum class represents to all ICDStateObserver callbacks available from the
```
bzbarsky-apple, line 77: "represents all"?

src/app/icd/server/ICDManager.h
```diff
@@ -71,11 +73,31 @@ class ICDManager : public ICDListener, public TestEventTriggerHandler
...
+     *
+     *        EnterActiveMode, TransitionToIdle and EnterIdleMode will always be called as a trio in the same order.
+     *        Each event will only be called once per cycle.
+     *        EnterActiveMode will always be called first, when the ICD has transitioned to ActiveMode.
+     *        TransitionToIdle will always be second. This event will only be called the first time there is
+     *        `ICD_ACTIVE_TIME_JITTER_MS` remaining to the ActiveMode timer.
+     *         When this event is called, the ICD is still in ActiveMode.
+     *        If the ActiveMode timer is increased due to the TransitionToIdle event, the event will not be called a second time in
+     *        a given cycle.
+     *        OnEnterIdleMode will always the third when the ICD has transitioned to IdleMode.
```
bzbarsky-apple, line 88: "OnEnterIdleMode will always the third event and indictats that the ICD has transitioned to IdleMode."

src/app/icd/server/ICDManager.h
```diff
@@ -71,11 +73,31 @@ class ICDManager : public ICDListener, public TestEventTriggerHandler
...
+     *
+     *        The ICDModeChange event can occur independently from the EnterActiveMode, TransitionToIdle and EnterIdleMode.
+     *        It will typpically hapen at the ICDManager init when a client is already registered with the ICD before the
```
bzbarsky-apple, line 91: "happen"

src/app/icd/server/ICDManager.h
```diff
@@ -71,11 +73,31 @@ class ICDManager : public ICDListener, public TestEventTriggerHandler
...
+     *        OnEnterIdleMode event or when a client send a register command after the OnEnterActiveMode event. Nothing prevents the
```
bzbarsky-apple, line 92: "sends"

src/app/icd/server/ICDManager.h
```diff
@@ -71,11 +73,31 @@ class ICDManager : public ICDListener, public TestEventTriggerHandler
...
+     *        ICDModeChange event to happen multiple times per cycle or while the ICD is in IdleMode.
```
bzbarsky-apple, line 93: "from happening"

PR #33089, opened 2024-04-22
Update the checkout_submodules.py script to support the recursive configuration in the .gitmodules file.

hsusid:
I split this PR out from https://github.com/project-chip/connectedhomeip/pull/33065 for a standalone change.

### DIFF
--- a/scripts/checkout_submodules.py
+++ b/scripts/checkout_submodules.py
@@ -52,7 +52,7 @@ ALL_PLATFORMS = set([
     'silabs_docker',
 ])
 
-Module = namedtuple('Module', 'name path platforms')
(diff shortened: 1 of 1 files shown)


hsusid:
> I wonder, can't the same be achieved with standard git's https://git-scm.com/docs/gitmodules#Documentation/gitmodules.txt-submoduleltnamegtfetchRecurseSubmodules?

I had wondered the same thing, tried it, and it didn't work as I expected. After having read some documents, it seems that the 'fetchRecurseSubmodules' only affects the 'fetch' behavior, but it doesn't help you to update submodules recursively for your working space.

Damian-Nordic (approved):

jmartinez-silabs (approved):

PR #33091, opened 2024-04-22
pw_unit_test migration: lib support batch #1

Alami-Amine:
On-going movement towards pw_unit_test
issue https://github.com/project-chip/connectedhomeip/pull/29342


src/lib/support/tests/TestBitMask.cpp
```diff
@@ -37,68 +36,68 @@ enum class TestEnum : uint16_t
...
 
-void TestBitMaskInvalid(nlTestSuite * inSuite, void * inContext)
+TEST(TestBitMask, TestBitMaskInvalid)
 {
     BitMask<TestEnum> mask;
 
     // This generally tests for no infinite loops. Nothing to set here
     mask.SetField(TestEnum::kZero, 0x01);
-    NL_TEST_ASSERT(inSuite, mask.Raw() == 0);
+    EXPECT_FALSE(mask.Raw());
```
andy31415, line 93: I think in this case equality shows things better. I would generally consider making `== 0` translate to _EQ instead of false in general as within our C++ code we tend to use `bool` for actual booleans and we rarely do the 1/0 true/false equivalence.

In this case it is a mask as a binary number, so saying mask is false feels awkward.

src/lib/support/tests/TestBitMask.cpp
```diff
@@ -37,68 +36,68 @@ enum class TestEnum : uint16_t
     kBits_High8 = 0xFF00,
 };
 
-void TestBitMaskOperations(nlTestSuite * inSuite, void * inContext)
+TEST(TestBitMask, TestBitMaskOperations)
 {
     BitMask<TestEnum> mask;
 
-    NL_TEST_ASSERT(inSuite, mask.Raw() == 0);
+    EXPECT_FALSE(mask.Raw());
```
andy31415, line 43: All mask.Raw should be eq tested rather than treating it as a bool.

andy31415 (approved):

PR #33093, opened 2024-04-22
[mrp] Make GetBackoff() use Timeout instead of Timestamp type

Damian-Nordic:
All users of `ReliableMessageMgr::GetBackoff()` seem to assume it takes and returns 32-bit `Timeout` while it actually takes and returns 64-bit `Timestamp`. Hence all the users must do implicit casts.

Replace Timestamp with Timeout in the function's signature and only use 64-bit type for internal calculations to prevent overflowing the underlying integer type.

### DIFF
--- a/src/messaging/ReliableMessageMgr.cpp
+++ b/src/messaging/ReliableMessageMgr.cpp
@@ -465,7 +465,7 @@ CHIP_ERROR ReliableMessageMgr::MapSendError(CHIP_ERROR error, uint16_t exchangeI
 
 void ReliableMessageMgr::CalculateNextRetransTime(RetransTableEntry & entry)
 {
-    System::Clock::Timestamp baseTimeout = System::Clock::Milliseconds64(0);
+    System::Clock::Timeout baseTimeout = System::Clock::Timeout(0);
 
     // Check if we have received at least one application-level message
     if (entry.ec->HasReceivedAtLeastOneMessage())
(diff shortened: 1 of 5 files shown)


src/messaging/tests/TestReliableMessageProtocol.cpp
```diff
@@ -225,92 +225,97 @@ struct BackoffComplianceTestVector
...
+                                                                            .sendCount   = 1,
+                                                                            .backoffBase = System::Clock::Timeout(300),
+                                                                            .backoffMin  = System::Clock::Timeout(330),
+                                                                            .backoffMax  = System::Clock::Timeout(413),
+                                                                        },
+                                                                        {
+                                                                            .sendCount   = 2,
+                                                                            .backoffBase = System::Clock::Timeout(300),
+                                                                            .backoffMin  = System::Clock::Timeout(528),
+                                                                            .backoffMax  = System::Clock::Timeout(661),
```
andy31415, line 244: Why did these change with off-by-one?

Damian-Nordic: Because I realized that some maximums were not correct given that we approximate 1.1 with 1127/1024 for faster arithmetics.

andy31415 (approved):

src/messaging/ReliableMessageMgr.cpp
bzbarsky-apple, line 1: I'm a little confused: how does `ReliableMessageMgr::CalculateNextRetransTime` compile with this change?  It's passing in a `System::Clock::Timestamp`....

Damian-Nordic: Yeah, I was surprised by that as well but it appears that casting from `Milliseconds64` to `Milliseconds32` is implicit and doesn't even raise a warning. According to the description of the 4th constructor in https://en.cppreference.com/w/cpp/chrono/duration/duration, only certain casts are prohibited like float->int or if the new period does not divide the old one. But thanks, looks like `CalculateNextRetransTime` should be aligned as well.

bzbarsky-apple (approved):

yunhanw-google (approved):

PR #33105, opened 2024-04-23
Add missing logging handler for embedded platforms for testing with Pigweed

jlatusek:
## Problem

There was a missing initialisation of `LoggingEventHandler` for embedded platforms. This caused tests for the nrfconnect to run and report an error, but didn't print the correct messages.

## Changes
Add initialization like in [pigweed main](https://github.com/google/pigweed/blob/c6e83c173d13b52d9542700bf40743d2548fe9a2/pw_unit_test/logging_main.cc#L18)

### DIFF
--- a/src/lib/support/UnitTest.cpp
+++ b/src/lib/support/UnitTest.cpp
@@ -1,12 +1,16 @@
 #include "UnitTest.h"
 
 #include "pw_unit_test/framework.h"
+#include "pw_unit_test/logging_event_handler.h"
 
 namespace chip {
 namespace test {
 
 int RunAllTests()
 {
+    testing::InitGoogleTest(nullptr, nullptr);
+    pw::unit_test::LoggingEventHandler handler;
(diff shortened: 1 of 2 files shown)


arkq (approved):

src/lib/support/UnitTest.cpp
```diff
@@ -1,12 +1,16 @@
...
 
 #include "pw_unit_test/framework.h"
+#include "pw_unit_test/logging_event_handler.h"
 
 namespace chip {
 namespace test {
 
 int RunAllTests()
 {
+    testing::InitGoogleTest(nullptr, nullptr);
```
arkq, line 11: I think that this will fails with real google test.... please check that.

arkq, line 11:
```suggestion
    testing::InitGoogleTest(nullptr, static_cast<char **>(nullptr));
```

arkq (approved):

andy31415 (approved):

PR #33106, opened 2024-04-23
Create a DNSSD test program for custom packet broadcast

andy31415:
This is a split from #33095 - it provides a program that for now just broadcasts a `PTR + SRV` announcement without any AAAA or TXT records.

### DIFF
--- a/examples/minimal-mdns/tester.cpp
+++ b/examples/minimal-mdns/tester.cpp
@@ -1,0 +1,303 @@
+/*
+ *
+ *    Copyright (c) 2020-2021 Project CHIP Authors
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


examples/minimal-mdns/tester.cpp
```diff
@@ -0,0 +1,303 @@
...
+HelpOptions helpOptions("minimal-mdns-tester", "Usage: minimal-mdns-tester [options]", "1.0");
+
+OptionSet * allOptions[] = { &cmdLineOptions, &helpOptions, nullptr };
+
+class ReportDelegate : public mdns::Minimal::ServerDelegate
+{
+public:
+    void OnQuery(const mdns::Minimal::BytesRange & data, const chip::Inet::IPPacketInfo * info) override
+    {
+        char addr[32];
```
tcarmelveilleux, line 140: This is not always enough. There is `IPAddress::kMaxStringLength`. Also, I think +1 is needed here. Same for all address buffers.

tcarmelveilleux (approved):

examples/minimal-mdns/tester.cpp
andy31415: This was a copy & paste. Adjusted all 3 example files (tester, client and server). Comments say it includes space for NUL so I did not add +1.

tehampson (approved):

mspang (approved):

woody-apple (approved):

PR #33109, opened 2024-04-23
[ESP32] Implement BLE Manager Shutdown for nimble host

shubhamdp:
- Replace ble deinit impl in Esp32AppServer with BLEMgr().Shutdown()
- Replace few ESP_LOG with ChipLog in Esp32AppServer
- Move ble deinit kCommissioningComplete switch case
- Make USE_BLE_ONLY_FOR_COMMISSIONING depends on BT_ENABLED

There has been intermittent crashes during ble deinitialization due to accessing the BLE memory which was freed.
Here, before freeing the BLE memory and adding it to heap we make sure that the nimble host stack is stopped.

#### Test
- verified that memory frees up after first commissioning and on boot up if device is commissioned.


examples/platform/esp32/common/Esp32AppServer.cpp
```diff
@@ -111,47 +111,12 @@ static size_t hex_string_to_binary(const char * hex_string, uint8_t * buf, size_
 
 void Esp32AppServer::DeInitBLEIfCommissioned(void)
 {
-#if CONFIG_BT_ENABLED && CONFIG_USE_BLE_ONLY_FOR_COMMISSIONING
+#if CONFIG_USE_BLE_ONLY_FOR_COMMISSIONING
```
dhrishi, line 114: Why is this removed? As `CONFIG_USE_BLE_ONLY_FOR_COMMISSIONING` depends on `CONFIG_BT_ENABLED` ?

src/platform/ESP32/nimble/BLEManagerImpl.cpp
```diff
@@ -252,6 +252,22 @@ CHIP_ERROR BLEManagerImpl::_Init()
     return err;
 }
 
+void BLEManagerImpl::_Shutdown()
+{
+    BleLayer::Shutdown();
+    xTimerDelete(sbleAdvTimeoutTimer, portMAX_DELAY);
+
+    // selectively setting kGATTServiceStarted flag, in order to notify the state machine to stop the CHIPoBLE gatt service
```
dhrishi, line 260: Nit: GATT

src/platform/ESP32/nimble/BLEManagerImpl.cpp
```diff
@@ -969,6 +986,56 @@ CHIP_ERROR BLEManagerImpl::InitESPBleLayer(void)
...
+void BLEManagerImpl::ClaimBLEMemory(System::Layer *, void *)
+{
+    TaskHandle_t handle = xTaskGetHandle("nimble_host");
+    if (handle)
+    {
+        ChipLogDetail(DeviceLayer, "Schedule ble memory reclaiming since nimble host is still running");
+
+        // Rescheduling it for later, 2 seconds is an arbitrary value, keeping it a bit more so that
+        // we dont have to reschedule it again
+        SystemLayer().StartTimer(System::Clock::Seconds32(2), ClaimBLEMemory, nullptr);
```
dhrishi, line 1004: The first execution of the function is through the caller of `DeinitESPBleLayer`, right? Can it be any task?

Also, the subsequent executions are through the timer task?

src/platform/ESP32/nimble/BLEManagerImpl.cpp
```diff
@@ -969,6 +986,56 @@ CHIP_ERROR BLEManagerImpl::InitESPBleLayer(void)
...
+#if CONFIG_IDF_TARGET_ESP32
+        err = esp_bt_mem_release(ESP_BT_MODE_BTDM);
+#elif CONFIG_IDF_TARGET_ESP32C2 || CONFIG_IDF_TARGET_ESP32C3 || CONFIG_IDF_TARGET_ESP32S3 || CONFIG_IDF_TARGET_ESP32H2 ||          \
+    CONFIG_IDF_TARGET_ESP32C6
+        err = esp_bt_mem_release(ESP_BT_MODE_BLE);
+#endif
+
+        VerifyOrReturn(err == ESP_OK, ChipLogError(DeviceLayer, "BLE deinit failed"));
+        ChipLogProgress(DeviceLayer, "BLE deinit successful and memory reclaimed");
+        // TODO: PostEvent(DeviceEventType::kBLEDeinitialized);
```
dhrishi, line 1020: Why is this commented?

examples/platform/esp32/common/Esp32AppServer.cpp
shubhamdp: Yes.

src/platform/ESP32/nimble/BLEManagerImpl.cpp
shubhamdp: done!

src/platform/ESP32/nimble/BLEManagerImpl.cpp
shubhamdp: Its a TODO, I'll implement it in subsequent PR. Rephrased it.

src/platform/ESP32/nimble/BLEManagerImpl.cpp
shubhamdp: The trigger for this is `_Shutdown()`, it sets some flags which tells the state machine to deinit and reclaim memory. Every action from there will always run in the chip thread context.

woody-apple (approved):

dhrishi (approved):

Issue #31571, opened 2024-01-20
[Tooling] Enforce no changes to data_model/ directory without a sha update


cecille:
### Reproduction steps / Feature

see title

### Platform

core (please add to version below)

### Type

Manually tested with SDK

PR #33110, opened 2024-04-23
Check for changes to data_model directory without a sha update - Issue #31571

vatsalghelani-csa:
Work for Issue #31571 

Workflow that runs on pull request - Enforce no changes to the data_model directory without a sha update, else result in failure of the check.

Files Worked:
- .github/workflows/check-data-model-directory-updates.yaml

### DIFF
--- a/.github/workflows/check-data-model-directory-updates.yaml
+++ b/.github/workflows/check-data-model-directory-updates.yaml
@@ -26,6 +26,6 @@ jobs:
     if: "git diff --name-only HEAD^..HEAD data_model/ | grep -q spec_sha"
     steps:
       - name: Error Message
-        run: echo This pull request attempts to update data_model directory, but is missing updates to spec_sha file with the latest version of the sha.
+        run: echo This pull request attempts to update data_model directory, but is missing updates to spec_sha file with the latest version of the sha. Files in the data_model directory are generated automatically and should not be updated manually.
       - name: Fail Job
         run: exit 1
--- a/.github/workflows/check-data-model-directory-updates.yaml
+++ b/.github/workflows/check-data-model-directory-updates.yaml
@@ -12,3 +12,22 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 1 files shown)


.github/workflows/check-data-model-directory-updates.yaml
```diff
@@ -0,0 +1,33 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Check for changes to data_model directory without a sha update 
+
+on:
+  pull_request:
+    branches-ignore:
+      - 'dependabot/**'
```
andy31415, line 20: There is no reason dependabot should be allowed to update data-model. I think we should remove this restriction. It made sense for submodules (which is the purpose of dependaboty) however it does not make sense for data_model.

andy31415 (approved):

.github/workflows/check-data-model-directory-updates.yaml
```diff
@@ -0,0 +1,31 @@
...
+      - "data_model/**"
+
+jobs:
+  check-submodule-update-label:
+    name: Check for changes to data_model directory without a sha update
+    runs-on: ubuntu-latest
+    if: "git diff --name-only HEAD^..HEAD data_model/ | grep -q spec_sha"
+    steps:
+      - name: Error Message
+        run: echo This pull request attempts to update data_model directory, but is missing updates to spec_sha file with the latest version of the sha.
```
cecille, line 29:
```suggestion
        run: echo This pull request attempts to update data_model directory, but is missing updates to spec_sha file with the latest version of the sha. Files in the data_model directory are generated automatically and should not be updated manually.
```

cecille: Tying to make it a bit more obvious that people shouldn't just update the sha too.

cecille (approved):

woody-apple (approved):

PR #33112, opened 2024-04-23
pw_unit_test migration: setup_payload batch #2 (with fixtures)

Alami-Amine:
Continuation of nl_test to pw_unit_test effort of issue #29342
batch 2 of issue #33079


src/setup_payload/tests/TestAdditionalDataPayload.cpp
```diff
@@ -22,22 +22,20 @@
  *
  */
 
+#include <gtest/gtest.h>
```
arkq, line 25: I would suggest to follow google c++ codestyle with include grouping:

1. standard includes
2. 3rd party includes
3. project includes

arkq (approved):

andy31415 (approved):

src/setup_payload/tests/TestAdditionalDataPayload.cpp
Alami-Amine: done, thanks!

Alami-Amine: - I took a look at the google[ C++ code style](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes), it seems that to strictly follow it I would also need to seperate the "related header" and have it before standard includes, this does not seem to be strictly enforced anyways, any thoughts?

- from my understanding, the google C++ code style will need the includes to be as below, which might overcomplicate things:

```
#include <setup_payload/AdditionalDataPayloadGenerator.h>
#include <setup_payload/AdditionalDataPayloadParser.h>
#include <setup_payload/SetupPayload.h>

#include <math.h>
#include <memory>
#include <stdio.h>

#include <gtest/gtest.h>

#include <lib/support/BytesToHex.h>
#include <lib/support/CHIPMem.h>
#include <lib/support/CHIPMemString.h>
#include <lib/support/CHIPPlatformMemory.h>

#include <system/SystemPacketBuffer.h>
#include <lib/support/verhoeff/Verhoeff.h>
```

PR #33114, opened 2024-04-23
Switch MTRDevice to using MTRDeviceClusterData for its attribute cache.

bzbarsky-apple:
Use these to store attribute values too, not just data versions.


jtung-apple (approved):

bzbarsky-apple:

PR #33117, opened 2024-04-23
TC-DA-1.7: Use steps_function, add missing checks

cecille:
Also adding in a new flag that will let us use this test in post-cert (just introducing flag now, checks will be updated in an upcoming PR).

### DIFF
--- a/src/python_testing/TC_DA_1_7.py
+++ b/src/python_testing/TC_DA_1_7.py
@@ -113,7 +113,7 @@ class TC_DA_1_7(MatterBaseTest):
         self.allow_sdk_dac = self.user_params.get("allow_sdk_dac", False)
         self.post_cert_test = self.user_params.get("post_cert_test", False)
 
-    def expected_number_of_DUTs() -> int:
+    def expected_number_of_DUTs(self) -> int:
         return 1 if (self.allow_sdk_dac or self.post_cert_test) else 2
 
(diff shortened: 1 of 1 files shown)


src/python_testing/TC_DA_1_7.py
```diff
@@ -107,20 +108,66 @@ class TC_DA_1_7(MatterBaseTest):
...
+        # We don't currently support this in cert because the base doesn't support multiple QR/manual
+        num = 0
+        if self.matter_test_config.discriminators:
+            num += len(self.matter_test_config.discriminators)
+        if self.matter_test_config.qr_code_content:
+            num += 1
+        if self.matter_test_config.manual_code:
+            num += 1
+        asserts.assert_equal(num, self.expected_number_of_DUTs(),
+                             "Unexpected number of devices specified - this test expects two DUTs at cert, one device for CI")
```
andy31415, line 162: we seem to compare `matter_test_config` data and the `user_params` derived data there. 

Could we re-phrase this message to reference what we actually check and should fix? The `test expects two DUTs at cert and one for CI` does not seem as obvious on where that info comes from.

cecille: Do you mean you want an explanation of the command line flags?

cecille: kk, updated the error messages to be more verbose here.

andy31415 (approved):

PR #33118, opened 2024-04-23
MinMDNS - do not ask for IP addresses for every SRV record that is seen

andy31415:
Cherrypick of #33095 into the 1.2 branch
Also cherrypick #30987 to fix some compile errors due to prompt toolkit updates.
Also cherrypick #32474 to fix some compile errors for esp32 compilation
Also cherrypick #32687 to fix some compile errors due to module named `proto` conflict in python modules

### DIFF
--- a/src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
+++ b/src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
@@ -17,8 +17,6 @@
 
 #include "Resolver.h"
 
-#include <limits>
-
 #include <lib/core/CHIPConfig.h>
 #include <lib/dnssd/ActiveResolveAttempts.h>
 #include <lib/dnssd/IncrementalResolve.h>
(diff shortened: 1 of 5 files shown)


src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
```diff
@@ -17,8 +17,6 @@
 
 #include "Resolver.h"
 
-#include <limits>
```
tcarmelveilleux, line 20: Why this removal?

tcarmelveilleux (approved):

mspang (approved):

src/lib/dnssd/Resolver_ImplMinimalMdns.cpp
andy31415: Unused header.

tcarmelveilleux (approved):

PR #33122, opened 2024-04-23
Use RAII for group session iteration

andy31415:
This cherrypicks #32970 into the 1.3 branch.

### DIFF
--- a/src/transport/SessionManager.cpp
+++ b/src/transport/SessionManager.cpp
@@ -58,6 +58,31 @@ using Transport::SecureSession;
 namespace {
 Global<GroupPeerTable> gGroupPeerTable;
 
+/// RAII class for iterators that guarantees that Release() will be called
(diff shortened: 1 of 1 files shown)


yunhanw-google (approved):

woody-apple:
Everything now is 1.3... We shouldn't cherry pick stuff in at this point. Will bring this up at the next SW standup (which will have to move times, sorry!)

woody-apple:
There are known issues in master right now for pairing issues, and failures, we should be encouraging (like all releases), master vs 1.3-branch.

andy31415:
> There are known issues in master right now for pairing issues, and failures, we should be encouraging (like all releases), master vs 1.3-branch.

Based on comments from various members, it seems that people would like stable branches. While I agree that master is likely to have more fixes than a fixed branch in time, we cannot assume everyone uses master.

PR #33125, opened 2024-04-23
Use NSCache for the MTRDevice attribute cache when possible.

bzbarsky-apple:
When we are storing our cluster data persistently, we don't have to ensure it's pinned in memory all the time.  Use NSCache, and reload from storage as needed.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -368,7 +368,7 @@ static NSString * const sAttributesKey = @"attributes";
     // don't have persistence enabled, this is our only data store.  Nil if we
     // currently have nothing that could need persisting.
     NSMutableDictionary<MTRClusterPath *, MTRDeviceClusterData *> * _Nullable _clusterDataToPersist;
-    // _existingClusters stores the set of "valid" keys into _persistedClusterData.
+    // _persistedClusters stores the set of "valid" keys into _persistedClusterData.
     // These are keys that could have values in _persistedClusterData even if they don't
     // right now (because they have been evicted).
     NSMutableSet<MTRClusterPath *> * _persistedClusters;
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -361,7 +361,7 @@ static NSString * const sAttributesKey = @"attributes";
 #endif
     BOOL _delegateDeviceCachePrimedCalled;
 
-    // _clusterData stores data that we have already persisted (when we have
+    // _persistedClusterData stores data that we have already persisted (when we have
     // cluster data persistence enabled).  Nil when we have no persistence enabled.
     NSCache<MTRClusterPath *, MTRDeviceClusterData *> * _Nullable _persistedClusterData;
     // _clusterDataToPersist stores data that needs to be persisted.  If we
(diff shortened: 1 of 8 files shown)


src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -361,10 +361,17 @@ @implementation MTRDevice {
 #endif
     BOOL _delegateDeviceCachePrimedCalled;
 
-    // With MTRDeviceClusterData now able to hold attribute data, the plan is to move to using it
-    // as the read cache, should testing prove attribute storage by cluster is the better solution.
-    NSMutableDictionary<MTRClusterPath *, MTRDeviceClusterData *> * _clusterData;
-    NSMutableSet<MTRClusterPath *> * _clustersToPersist;
+    // _clusterData stores data that we have already persisted (when we have
```
jtung-apple, line 364:
```suggestion
    // _persistedClusterData stores data that we have already persisted (when we have
```

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -361,10 +361,17 @@ @implementation MTRDevice {
...
+    // cluster data persistence enabled).  Nil when we have no persistence enabled.
+    NSCache<MTRClusterPath *, MTRDeviceClusterData *> * _Nullable _persistedClusterData;
+    // _clusterDataToPersist stores data that needs to be persisted.  If we
+    // don't have persistence enabled, this is our only data store.  Nil if we
+    // currently have nothing that could need persisting.
+    NSMutableDictionary<MTRClusterPath *, MTRDeviceClusterData *> * _Nullable _clusterDataToPersist;
+    // _existingClusters stores the set of "valid" keys into _persistedClusterData.
```
jtung-apple, line 371:
```suggestion
    // _persistedClusters stores the set of "valid" keys into _persistedClusterData.
```
Comment to match the code

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -1205,13 +1232,75 @@ - (void)_handleEventReport:(NSArray<NSDictionary<NSString *, id> *> *)eventRepor
...
+    NSAssert(_deviceController.controllerDataStore != nil,
+        @"How can _persistedClusters have an entry if we have no persistence?");
+    NSAssert(_persistedClusterData != nil,
+        @"How can _persistedClusterData not exist if we have persisted clusters?");
+
+    // Page in the stored value for the data.
+    MTRDeviceClusterData * data = [_deviceController.controllerDataStore getStoredClusterDataForNodeID:_nodeID endpointID:clusterPath.endpoint clusterID:clusterPath.cluster];
+    if (data != nil) {
+        [_persistedClusterData setObject:data forKey:clusterPath];
+    }
```
jtung-apple, line 1278: If at this point data is nil (e.g. storage data loss, or previous failure to store), should we trigger some kind of recovery mechanism?

bzbarsky-apple: In general, yes, we should.  I am not sure what form that can/should take, yet.  But we should probably have a followup to figure that out and do that.

That said, we should also trigger that mechanism if the data that comes back is not valid, etc, right?  So might be a bigger task than just "we got nothing".

jtung-apple: Yup I don't see this as blocking merge, and definite more of a follow up task.

jtung-apple (approved):

bzbarsky-apple:

PR #33128, opened 2024-04-24
Prevent unclean dealloc of SystemState

ksperling-apple:
Per current documentation calling Shutdown while any controller is alive is already prohibited. But at the moment, calling `Shutdown` while there are unbalanced `RetainSystemState` calls outstanding is broken in that we deallocate the SystemState without calling its `Shutdown` method, i.e. various parts of the stack don't get stopped correctly. This change explicitly prohibits this.

Also rename the recently-introduced isShutdown method to isShutDown.

### DIFF
--- a/src/controller/CHIPDeviceControllerFactory.cpp
+++ b/src/controller/CHIPDeviceControllerFactory.cpp
@@ -76,7 +76,7 @@ CHIP_ERROR DeviceControllerFactory::Init(FactoryInitParams params)
 CHIP_ERROR DeviceControllerFactory::ReinitSystemStateIfNecessary()
 {
     VerifyOrReturnError(mSystemState != nullptr, CHIP_ERROR_INCORRECT_STATE);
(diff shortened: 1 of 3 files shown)


jmartinez-silabs (approved):

saurabhst (approved):

andy31415 (approved):

src/controller/CHIPDeviceControllerFactory.cpp
```diff
@@ -404,6 +404,7 @@ void DeviceControllerFactory::Shutdown()
 {
     if (mSystemState != nullptr)
     {
+        VerifyOrDie(mSystemState->IsShutDown());
```
andy31415, line 407: should we have a comment saying why this verification exists? That way if some code path reaches this, people can understand why this check is done and `mSystemState` nullability check is insufficient (i.e. explain why delete on a system state does not implicitly mean shut down).

PR #33142, opened 2024-04-24
Fix builds to macos-13 rather than latest due to bootstrap python incompatibilities

andy31415:
Builds seem to fail with `AttributeError: cython_sources` which seems to point at an incompatibility between PyYaml and python 3.12 (pyyaml seems to claim compatibility up to python 3.11 so far)

Preliminary investigation of failures seem to indicate:
  - CI passed when `macos-latest` was pointing to `macos-12` (odd, since 13 also existed)
  - When github moved the label to point to `macos-14-arm64` we started having failures

So this may be python version or architecture version or both.

Unclear if this is the correct fix since we presumably want to ensure darwin code compiles on the latest release of the platform, however for now validating the claim that runner version is at fault for build failures.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -371,7 +371,7 @@ jobs:
 
     build_darwin:
         name: Build on Darwin (clang, python_lib, simulated)
-        runs-on: macos-latest
+        runs-on: macos-13
         if: github.actor != 'restyled-io[bot]'
 
         steps:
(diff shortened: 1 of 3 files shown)


andy31415:
Marking as hotfix and will merge as without this entire CI is blocked on darwin and no PRs can be checked.

@woody-apple @bzbarsky-apple @ksperling-apple please be aware and update with other fixes as appropriate for Darwin. For now I just patched it to unblock CI.

PR #33143, opened 2024-04-24
Make sure reference counting is correct when calling abort on a failed Paring handshake.

andy31415:
ExchangeContext::Abort() already calls Release(). Since ExchangeHandles are used, maintain the ReferenceCount the same by increasing it before calling abort. Cleanup is to be handled by ExchangeHandles.

### DIFF
--- a/src/protocols/secure_channel/PairingSession.cpp
+++ b/src/protocols/secure_channel/PairingSession.cpp
@@ -242,7 +242,7 @@ void PairingSession::Clear()
         // ExchangeContext::Abort will auto-release the underlying values. Since
         // we are using reference counted handles, we Retain to increase the
         // reference count before Abort, so that Abort() just closes without
-        // invalidagint context reference counts
+        // invalidating handle reference counts
         const uint32_t referenceCount = mExchangeCtxt.Value()->GetReferenceCount();
         mExchangeCtxt.Value()->Retain();
         mExchangeCtxt.Value()->Abort();
--- a/src/protocols/secure_channel/PairingSession.cpp
+++ b/src/protocols/secure_channel/PairingSession.cpp
@@ -16,6 +16,7 @@
  *    limitations under the License.
  */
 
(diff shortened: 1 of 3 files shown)


src/protocols/secure_channel/PairingSession.cpp
```diff
@@ -237,7 +238,16 @@ void PairingSession::Clear()
         // middle of our handshake. In that case, there is no point in trying to
         // do MRP resends of the last message we sent. So, abort the exchange
         // instead of just closing it.
+
+        // ExchangeContext::Abort will auto-release the underlying values. Since
+        // we are using reference counted handles, we Retain to increase the
+        // reference count before Abort, so that Abort() just closes without
+        // invalidating handle reference counts
```
bzbarsky-apple, line 245: The reference counted handle is supposed to be a separate ref from the ref that Abort() here is decrementing.

andy31415: Do you happen to know the code paths which use non-reference-counted handles vs the ones that the reference counted handles see?

When this piece of code triggers, I do see the reference count being `1` when this code enters, so both abort and the handle deletion tries to call Release, calling it 2x for a count of 1, which is no good.

andy31415:
Marking as draft ... apparently we have some unit tests fail as a result, so it seems we have some code paths that rely on different aquire/release logic.

andy31415:
Closing: this may be a different logic error ... I am not familiar with the lifetime bits on this so my first instinct was probably wrong.

src/protocols/secure_channel/PairingSession.cpp
bzbarsky-apple: That's a really broad question.  When an exchange is created, it starts with a refcount of 1, and the responsibility for that refcount is transferred between the network stack and the exchange delegate in a somewhat complicated dance....

PR #33148, opened 2024-04-24
Fix handling of non-permitted messages in ExchangeContext.

bzbarsky-apple:
They should not count as responses, since we don't notify our delegate about them.

### DIFF
--- a/src/messaging/ExchangeContext.cpp
+++ b/src/messaging/ExchangeContext.cpp
@@ -596,21 +596,26 @@ CHIP_ERROR ExchangeContext::HandleMessage(uint32_t messageCounter, const Payload
     // Set kFlagReceivedAtLeastOneMessage to true since we have received at least one new application level message
     SetHasReceivedAtLeastOneMessage(true);
 
-    if (IsResponseExpected())
-    {
-        // Since we got the response, cancel the response timer.
-        CancelResponseTimer();
-
-        // If the context was expecting a response to a previously sent message, this message
-        // is implicitly that response.
-        SetResponseExpected(false);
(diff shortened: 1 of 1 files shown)


src/messaging/ExchangeContext.cpp
```diff
@@ -596,21 +596,26 @@ CHIP_ERROR ExchangeContext::HandleMessage(uint32_t messageCounter, const Payload
     // Set kFlagReceivedAtLeastOneMessage to true since we have received at least one new application level message
     SetHasReceivedAtLeastOneMessage(true);
 
-    if (IsResponseExpected())
+    // Don't send messages on to our delegate if our dispatch does not allow
+    // those messages.  Those messages should also not be treated as responses,
+    // since if our delegate is expecting a response we will not notify it about
+    // these messages.
+    if (mDispatch.MessagePermitted(payloadHeader.GetProtocolID(), payloadHeader.GetMessageType()))
```
andy31415, line 603: Is this change something we can unit test? It would be great if we could ensure we never regress again on this.

andy31415: Discussed offline, we can add the test as we work on test infra, so it will be a separate PR.

andy31415 (approved):

src/messaging/ExchangeContext.cpp
bzbarsky-apple: It should be, but will take some work.  I'll do the test as a followup.

yunhanw-google (approved):

woody-apple (approved):

PR #33152, opened 2024-04-25
[Android] Implement LIT ICD for subscription

### LINKED ISSUE #31580: [Android-chip-tool] Support LIT ICD for subscriptions
Assume android chip-tool has commissioned one LIT, if it wanna subscribe to this LIT, it would check whether ICDClientStore has the related entry inside and automatically go with ICD subscription via setting mIsPeerLIT in ReadPrepareParams.

The reference c++ chip-tool work is in https://github.com/project-chip/connectedhomeip/pull/31576, thanks

joonhaengHeo:
Fix #31580 

For supporting LIT ICD subscribe, this code request to merge.


src/controller/java/AndroidICDClient.cpp
```diff
@@ -0,0 +1,99 @@
...
+ *
+ */
+
+#include "AndroidICDClient.h"
+
+#include <app/icd/client/ICDClientInfo.h>
+
+chip::app::DefaultICDClientStorage sICDClientStorage;
+
+jobject getICDClientInfo(JNIEnv * env, const char * icdClientInfoSign, jint jFabricIndex)
```
yunhanw-google, line 31: although it is good to have getICDClientInfo to dump all lit client info for disgnostic purpose as we previous discussed, it may be better to introduce another getICDClientInfo with fabric index and nodeId for "isPeerICDClient usage" without stream match.

joonhaengHeo: @yunhanw-google 
The API providing ICDClientInfo already exists, so I don't think it's necessary to create a new one. (Previously, it was implemented in the ChipDeviceController class, and it was created as a separate class to use in InteractionClient.)

yunhanw-google: ok, thanks

yunhanw-google (approved):

src/controller/java/AndroidICDClient.cpp
```diff
@@ -0,0 +1,99 @@
...
+ *    @file
+ *      Implementation of ICD Client API for Android Platform
+ *
+ */
+
+#include "AndroidICDClient.h"
+
+#include <app/icd/client/ICDClientInfo.h>
+
+chip::app::DefaultICDClientStorage sICDClientStorage;
```
yunhanw-google, line 29: Later android tv-app may act as the hub and get the ICD key and offset from cloud and save these icd metadata into ICDClientStorage for later check-in message handling. I think we can further expose the Java API related to https://github.com/project-chip/connectedhomeip/blob/master/src/app/icd/client/ICDClientStorage.h#L38  for ICDClientStorage.

This may be follow-up, not blocking.

joonhaengHeo: thank you for telling me. I think this part can be implemented additionally in this file once the exact requirements are determined in tv-app.

sharadb-amazon (approved):

PR #33155, opened 2024-04-25
Updated unit tests in src/transport/tests/ to use PW instead of NL.

feasel0:
These changes are replacing NL unit tests with pigweed unit tests. I updated the tests in src/transport/tests/


andy31415 (approved):

src/transport/tests/TestSecureSessionTable.cpp
```diff
@@ -131,10 +128,25 @@ void TestSecureSessionTable::CreateSessionTable(std::vector<SessionParameters> &
     }
 }
 
-void TestSecureSessionTable::ValidateSessionSorting(nlTestSuite * inSuite, void * inContext)
+class SecureSessionTableTest : public ::testing::Test
```
jlatusek, line 131: Merge with `TestSecureSessionTable`

src/transport/tests/TestSessionManager.cpp
```diff
@@ -106,51 +103,53 @@ class TestSessMgrCallback : public SessionMessageDelegate
...
         ReceiveHandlerCallCount++;
     }
 
-    nlTestSuite * mSuite        = nullptr;
     int ReceiveHandlerCallCount = 0;
     bool LargeMessageSent       = false;
 };
 
-void CheckSimpleInitTest(nlTestSuite * inSuite, void * inContext)
+class TestSessionManager : public ::testing::Test
```
jlatusek, line 120: Merge it with `TestContext`

src/transport/tests/TestSessionManager.cpp
```diff
@@ -979,31 +945,29 @@ static void SessionShiftingTest(nlTestSuite * inSuite, void * inContext)
...
     // Now shift the 1st session to the 3rd one, after shifting, holders should be:
     // 1. alice -> bob: stickyAliceToBobSession
     // 2. alice <- bob: bobToAliceSession
     // 3. alice -> bob: aliceToBobSession, newAliceToBobSession
     sessionManager.GetSecureSessions().NewerSessionAvailable(newAliceToBobSession.Get().Value()->AsSecureSession());
 
-    NL_TEST_ASSERT(inSuite, aliceToBobSession);
-    NL_TEST_ASSERT(inSuite, stickyAliceToBobSession);
-    NL_TEST_ASSERT(inSuite, newAliceToBobSession);
+    EXPECT_TRUE(!!aliceToBobSession);
```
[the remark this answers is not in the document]
andy31415: It seems we only have operator bool on these. I guess we could do static_cast<bool> or Get().HasValue() however both looked ugly as well. the `!!` does feel javascript-like however I had no good suggestions. Do you have any?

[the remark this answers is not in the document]
jlatusek: I've tried it locally, and the compiler doesn't have a problem with automatic conversion to a bool value in the `EXPECT_TRUE` expression. In [doc](https://en.cppreference.com/w/cpp/language/implicit_conversion), I found out that `Contextual conversions` should be made for `the controlling expression of if, while, for;`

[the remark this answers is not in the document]
feasel0: Ok this seems to work.  Removing `!!`.

src/transport/tests/TestSessionManager.cpp
```diff
@@ -137,8 +134,8 @@ TEST_F(TestSessionManager, CheckSimpleInitTest)
 
     EXPECT_EQ(CHIP_NO_ERROR, fabricTableHolder.Init());
     EXPECT_EQ(CHIP_NO_ERROR,
-              sessionManager.Init(&inContext->GetSystemLayer(), &inContext->GetTransportMgr(), &gMessageCounterManager,
-                                  &deviceStorage, &fabricTableHolder.GetFabricTable(), sessionKeystore));
+              sessionManager.Init(&this->GetSystemLayer(), &this->GetTransportMgr(), &gMessageCounterManager, &deviceStorage,
```
jlatusek, line 137: You can just call functions, you don't have to use `&this->`

jlatusek (changes requested):

src/transport/tests/TestSessionManager.cpp
feasel0: Updated in all places.

src/controller/tests/data_model/BUILD.gn
```diff
@@ -40,4 +40,4 @@ chip_test_suite_using_nltest("data_model") {
     "${chip_root}/src/transport/raw/tests:helpers",
     "${nlunit_test_root}:nlunit-test",
   ]
-}
+}
```
andy31415, line 43: some of these files seem to have lost the trailing newline. Could we add it back?

PR #33156, opened 2024-04-25
Add equality operator to StatusIB

mwswartwout:
Add equality (and inequality) operator to StatusIB

### DIFF
--- a/src/app/MessageDef/StatusIB.h
+++ b/src/app/MessageDef/StatusIB.h
@@ -117,5 +117,15 @@ struct StatusIB
 
 }; // struct StatusIB
 
+constexpr bool operator==(const StatusIB & one, const StatusIB & two)
+{
+    return one.mStatus == two.mStatus && one.mClusterStatus == two.mClusterStatus;
+}
+
+constexpr bool operator!=(const StatusIB & one, const StatusIB & two)
+{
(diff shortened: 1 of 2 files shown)


tcarmelveilleux (approved):

mspang (approved):

src/app/MessageDef/StatusIB.h
```diff
@@ -117,5 +117,15 @@ struct StatusIB
 
 }; // struct StatusIB
 
+constexpr bool operator==(const StatusIB & one, const StatusIB & two)
+{
+    return one.mStatus == two.mStatus && one.mClusterStatus == two.mClusterStatus;
```
bzbarsky-apple, line 122: mClusterStatus is meaningless unless mStatus has one of the two "generic" values, right?

mwswartwout: If true that's not enforced or validated by the constructor in any way, you can set an arbitrary mStatus with an arbitrary mClusterStatus.

bzbarsky-apple: Yes, that is true.  This is probably fine, but I can think of weird situations where this can lead to "wrong" results....

bzbarsky-apple (approved):

PR #33168, opened 2024-04-25
Add python checkout for arm64 to be fixed as well to fix macos-latest to arm64

andy31415:
Looking to see if fixing python makes Darwin compile.


andy31415:
Closing: does not seem to actually work. Will cherrypick the fixing of the image instead in #33142

PR #33170, opened 2024-04-25
[OpenThread] Disable IPv6 interface during erasing persistent info.

ArekBalysNordic:
Apart from disabling Thread, and erasing Persistent Info, we should disable the IPv6 interface as well and block receiving further packets. Otherwise, if we don't reboot the device after erasing Thread persistent data, we can get a packet and as a result, it causes an assert during processing AES_ECB (because crypto keys have been removed).

### DIFF
--- a/src/platform/OpenThread/GenericThreadStackManagerImpl_OpenThread.hpp
+++ b/src/platform/OpenThread/GenericThreadStackManagerImpl_OpenThread.hpp
@@ -1212,6 +1212,7 @@ void GenericThreadStackManagerImpl_OpenThread<ImplClass>::_ErasePersistentInfo(v
     ChipLogProgress(DeviceLayer, "Erasing Thread persistent info...");
     Impl()->LockThreadStack();
     otThreadSetEnabled(mOTInst, false);
+    otIp6SetEnabled(mOTInst, false);
     otInstanceErasePersistentInfo(mOTInst);
     Impl()->UnlockThreadStack();
 }


LuDuda (approved):

arkq (approved):

jmartinez-silabs (approved):

src/platform/OpenThread/GenericThreadStackManagerImpl_OpenThread.hpp
```diff
@@ -1212,6 +1212,7 @@ void GenericThreadStackManagerImpl_OpenThread<ImplClass>::_ErasePersistentInfo(v
     ChipLogProgress(DeviceLayer, "Erasing Thread persistent info...");
     Impl()->LockThreadStack();
     otThreadSetEnabled(mOTInst, false);
+    otIp6SetEnabled(mOTInst, false);
```
jmartinez-silabs, line 1215: It doesn't hurt to add it, but I'm surprised a packet can still be received after disabling thread. I might be unaware of some underlying process.

LuDuda: @jmartinez-silabs `otThreadSetEnabled` itself does not disable the radio, so yeah, `ifconfig down` helps here.

jmartinez-silabs: Thank you both for the explanation

PR #33174, opened 2024-04-25
[ICD] Add counter invalidation events to the ICD Test Event trigger handler

mkardous-silabs:
#### Description
The goal of the PR is to add two new ICD test event trigger events that increase the ICDCounter value to enable client behavior testing. The two events are :

1. kInvalidateHalfCounterValues  which increases the ICDCounter by half of its maximum size to enable testing the client refresh key behavior.
2. kInvalidateAllCounterValues which invalidates all the ICDCounter values to verify Check-In message protection mechanisms against replay attacks.

**IMPORTANT** : PR also modifies the PersistedCounter check when updating the counter value because it prevented the counter from using the last counter value (max counter size value - UINTXX_MAX) and would return an error anytime the counter is increased afterwards. - https://github.com/project-chip/connectedhomeip/issues/33175

**PR adds / changes :**

- Adds AdvanceBy(...) function to the CHIPCounter interface to enable advance by a given value
- Adds AdvanceBy(...) to the PersistedCounter class to enable advance by a given value and updating the persistent storage epoch
- Adds the InvalidateHalfCheckInCouterValues & InvalidateAllCheckInCounterValues functions to the CheckInCouter class to modify counter value
- Adds the Test event trigger events to the ICDManager handler

**Drive by changes :**

- Convert TestCHIPCounter to gtest
- Convert TestPersistedCounter to gtest

#### Note to revewiers:
TestICDManager will be converted to gtest in a follow up since it is a pretty large test file

#### Tests

- Existing unit tests
- Newly added unit tests for each class that has new code
- Newly added integration test steps to validate end-to-end behavior


jmartinez-silabs (approved):

yunhanw-google (approved):

src/lib/support/CHIPCounter.h
```diff
@@ -45,20 +45,27 @@ class Counter
...
     virtual T GetValue() = 0;
+
+    /**
+     * @brief Set counter to a new value
+     *
+     * @param value new counter value
+     *
+     * @return CHIP_ERROR A CHIP error code if anything failed, CHIP_NO_ERROR otherwise.
+     */
+    virtual CHIP_ERROR SetValue(T value) = 0;
```
bzbarsky-apple, line 68: If this is meant for tests only, should it only be supported when compiling for unit tests, perhaps?  Either entirely ifdefed out, or just implemented here to return CHIP_ERROR_NOT_IMPLEMENTED and the subclass impls are conditional?

It's rather worrying to have something that can violate a counter's invariants like this, in general.

If the goal is to act as if the counter has advanced a bunch, should we have an "advance by N" method instead?

src/lib/support/PersistedCounter.h
```diff
@@ -110,6 +110,45 @@ class PersistedCounter : public MonotonicallyIncreasingCounter<T>
         return MonotonicallyIncreasingCounter<T>::Init(startValue);
     }
 
+    CHIP_ERROR SetValue(T value) override
+    {
+        VerifyOrReturnError(mStorage != nullptr, CHIP_ERROR_INCORRECT_STATE);
+        VerifyOrReturnError(mKey.IsInitialized(), CHIP_ERROR_INCORRECT_STATE);
+
+        // Calculate previous Epoch from value
+        T previousEpoch = value - (value % mEpoch);
```
bzbarsky-apple, line 119: I'm not sure this makes sense.  Nothing ensures that the stored epoch values are multiples of mEpoch.  This `previousEpoch` value has nothing to do with anything "previous".

What we are really trying to compute here is a place to restart the counter next time, no?  We can just as easily use `value + mEpoch` (just like in `Init()` we use "value read from storage + mEpoch").  If we are doing something else, we need a clear explantion for why.

src/lib/support/PersistedCounter.h
```diff
@@ -110,6 +110,45 @@ class PersistedCounter : public MonotonicallyIncreasingCounter<T>
...
+#endif
+
+        // Set next Epoch value
+        ReturnErrorOnFailure(PersistNextEpochStart(previousEpoch + mEpoch));
+
+        // Set new mCounterValue
+        ReturnErrorOnFailure(MonotonicallyIncreasingCounter<T>::SetValue(value));
+
+        // Advancing the epoch should have ensured that the current value is valid
+        VerifyOrReturnError(MonotonicallyIncreasingCounter<T>::GetValue() + mEpoch >= mNextEpoch, CHIP_ERROR_INTERNAL);
```
bzbarsky-apple, line 142: This has the same rollover problems as the commented-out checks here, no?  That is, if we are in the `>` case, not `==`, then we could test false if `mNextEpoch` is the mac counter value....

It seems like if these counters can roll over we need to be _very_ clear about what our `>` and `<` mean, or just use equality checks....

src/lib/support/PersistedCounter.h
```diff
@@ -130,9 +169,15 @@ class PersistedCounter : public MonotonicallyIncreasingCounter<T>
             // Ensure that a new starting point is persisted.
             ReturnErrorOnFailure(PersistNextEpochStart(mNextEpoch + mEpoch));
 
-            // Advancing the epoch should have ensured that the current value
-            // is valid
-            VerifyOrReturnError(MonotonicallyIncreasingCounter<T>::GetValue() < mNextEpoch, CHIP_ERROR_INTERNAL);
+            // Advancing the epoch should have ensured that the current value is valid
+            VerifyOrReturnError(MonotonicallyIncreasingCounter<T>::GetValue() + mEpoch == mNextEpoch, CHIP_ERROR_INTERNAL);
```
bzbarsky-apple, line 173: It's sort of odd to fix this bit but not the `GetValue() >= mNextEpoch` check... Or is 33175 tracking fixing this in general?

src/protocols/secure_channel/CheckInCounter.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+namespace chip {
+namespace Protocols {
+namespace SecureChannel {
+
+CHIP_ERROR CheckInCounter::InvalidateHalfCheckInCouterValues()
+{
+    uint32_t currentCounterValue = GetValue();
+
+    // Double current counter value and update underlying counter storage. CheckInCounter is allowed to rollover.
```
bzbarsky-apple, line 33: This function is not doubling....

src/protocols/secure_channel/CheckInCounter.h
```diff
@@ -45,6 +46,24 @@ class CheckInCounter : public PersistedCounter<uint32_t>
      */
     inline uint32_t GetNextCheckInCounterValue() { return (mCounterValue + 1); };
 
+    /**
+     * @brief Doubles the current Check-In counter value to invalidate at least half of the values.
```
bzbarsky-apple, line 50: Again, not sure "doubles" is the right word here.

src/lib/support/PersistedCounter.h
mkardous-silabs: Good point! I'll simplify the logic around this.

src/lib/support/PersistedCounter.h
mkardous-silabs: Reason for the difference in `>=` and `==` was due to this bit of logic
```
// Calculate previous Epoch from value
T previousEpoch = value - (value % mEpoch);
```

which created a situation where it was possible for `==` would be false. By simplifying the logic around that calculation, both can have the same check.

src/lib/support/CHIPCounter.h
bzbarsky-apple: Ah, good point about test event trigger.

src/app/tests/TestICDManager.cpp
```diff
@@ -742,6 +744,40 @@ class TestICDManager
...
+    {
+        TestContext * ctx = static_cast<TestContext *>(aContext);
+
+        constexpr uint32_t startValue    = 105;
+        constexpr uint32_t expectedValue = 104;
+
+        // Set starting value for test
+        ICDConfigurationData::GetInstance().GetICDCounter().SetValue(startValue);
+
+        // Trigger ICD kInvalidateHalfCounterValues event
```
bzbarsky-apple, line 774:
```suggestion
        // Trigger ICD kInvalidateAllCounterValues event
```

src/lib/support/PersistedCounter.h
```diff
@@ -104,12 +104,22 @@ class PersistedCounter : public MonotonicallyIncreasingCounter<T>
...
     }
 
+    CHIP_ERROR AdvanceBy(T value) override
+    {
+        VerifyOrReturnError(mStorage != nullptr, CHIP_ERROR_INCORRECT_STATE);
+        VerifyOrReturnError(mKey.IsInitialized(), CHIP_ERROR_INCORRECT_STATE);
+
+        ReturnErrorOnFailure(MonotonicallyIncreasingCounter<T>::AdvanceBy(value));
+
+        return VerifyAndPersistNextEpochStart(MonotonicallyIncreasingCounter<T>::GetValue());
```
bzbarsky-apple, line 120: This could use a comment explaining why we are using current value as the reference epoch here.

src/lib/support/PersistedCounter.h
```diff
@@ -124,20 +134,32 @@ class PersistedCounter : public MonotonicallyIncreasingCounter<T>
...
         ReturnErrorOnFailure(MonotonicallyIncreasingCounter<T>::Advance());
 
+        return VerifyAndPersistNextEpochStart(mNextEpoch);
+    }
+
+private:
+    CHIP_ERROR VerifyAndPersistNextEpochStart(T refEpoch)
+    {
+
         if (MonotonicallyIncreasingCounter<T>::GetValue() >= mNextEpoch)
```
bzbarsky-apple, line 144: So fixing this check (which does not seem to be right in the "counter rolled over when doing AdvanceBy" case) is a followup, right?

src/protocols/secure_channel/CheckInCounter.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include <protocols/secure_channel/CheckInCounter.h>
+#include <stdint.h>
+
+namespace chip {
+namespace Protocols {
+namespace SecureChannel {
+
+CHIP_ERROR CheckInCounter::InvalidateHalfCheckInCouterValues()
+{
+    // Increases the current counter value by half of its maximum range and update underlying counter storage.
```
bzbarsky-apple, line 31:
```suggestion
    // Increases the current counter value by half of its maximum range and updates underlying counter storage.
```
Or "increase" and "update", but just be consistent.

src/protocols/secure_channel/CheckInCounter.cpp
```diff
@@ -0,0 +1,47 @@
...
+    // CheckInCounter is allowed to rollover.
```
bzbarsky-apple, line 32:
```suggestion
    // CheckInCounter is allowed to roll over.
```

src/protocols/secure_channel/CheckInCounter.cpp
```diff
@@ -0,0 +1,47 @@
...
+    ReturnErrorOnFailure(AdvanceBy((UINT32_MAX / 2)));
+    return CHIP_NO_ERROR;
+}
+
+CHIP_ERROR CheckInCounter::InvalidateAllCheckInCounterValues()
+{
+    // Increase current counter value by its maximum value to mimick N -> UINT32_MAX -> N-1 to invalidate all Check-In counter
+    // values. CheckInCounter is allowed to rollover.
```
bzbarsky-apple, line 40:
```suggestion
    // values. CheckInCounter is allowed to roll over.
```

src/protocols/secure_channel/CheckInCounter.h
```diff
@@ -45,6 +46,24 @@ class CheckInCounter : public PersistedCounter<uint32_t>
...
+     *        Function adds UINT32_MAX / 2 to the current counter value. Next time a Check-In message is sent, 1 is added to the
+     *        current value.
+     *
+     * @return CHIP_ERROR Any error returned by a write to persisted storage.
```
bzbarsky-apple, line 54:
```suggestion
     * @brief Advances the current Check-In counter value to invalidate half of the values.
     *        Function adds UINT32_MAX / 2 to the current counter value. Next time a Check-In message is sent, 1 is added to the
     *        current value.
     *
     * @return CHIP_ERROR Any error returned by a write to persisted storage.
```

src/protocols/secure_channel/CheckInCounter.h
```diff
@@ -45,6 +46,24 @@ class CheckInCounter : public PersistedCounter<uint32_t>
...
+     */
+    CHIP_ERROR InvalidateHalfCheckInCouterValues();
+
+    /**
+     * @brief Invalidates all the Check-In counter values. After function call, new Check-In counter value will be equal to current
```
bzbarsky-apple, line 59:
```suggestion
     * @brief Invalidates all the Check-In counter values. After this function is called, the new Check-In counter value will be equal to current
```

src/protocols/secure_channel/CheckInCounter.cpp
```diff
@@ -0,0 +1,47 @@
...
+{
+    // Increases the current counter value by half of its maximum range and update underlying counter storage.
+    // CheckInCounter is allowed to rollover.
+    ReturnErrorOnFailure(AdvanceBy((UINT32_MAX / 2)));
+    return CHIP_NO_ERROR;
+}
+
+CHIP_ERROR CheckInCounter::InvalidateAllCheckInCounterValues()
+{
+    // Increase current counter value by its maximum value to mimick N -> UINT32_MAX -> N-1 to invalidate all Check-In counter
```
bzbarsky-apple, line 39:
```suggestion
    // Increase current counter value by its maximum value to mimic N -> UINT32_MAX -> N-1 to invalidate all Check-In counter
```

src/protocols/secure_channel/CheckInCounter.h
```diff
@@ -45,6 +46,24 @@ class CheckInCounter : public PersistedCounter<uint32_t>
...
+     *        value minus 1 (with rollover) to mimick N -> UINT32_MAX -> N-1. Next time a Check-In message is sent, 1 is added to
```
bzbarsky-apple, line 60:
```suggestion
     *        value minus 1 (with rollover) to mimic N -> UINT32_MAX -> N-1. Next time a Check-In message is sent, 1 is added to
```

src/protocols/secure_channel/tests/TestCheckInCounter.cpp
```diff
@@ -0,0 +1,129 @@
...
+namespace {
+
+enum class CheckInCounterOperations : uint8_t
+{
+    kInvalidateHalf = 1,
+    kInvalidateAll  = 2,
+};
+
+/**
+ * @brief Helpter function that validates CheckInCounter value after an operation for configurable input values
```
bzbarsky-apple, line 37:
```suggestion
 * @brief Helper function that validates CheckInCounter value after an operation for configurable input values
```

bzbarsky-apple (approved):

src/lib/support/PersistedCounter.h
mkardous-silabs: done

src/lib/support/PersistedCounter.h
mkardous-silabs: Fixed the check and added an extra unit test for `AdvanceBy` to validate the behavior.

src/lib/support/PersistedCounter.h
```diff
@@ -104,16 +105,50 @@ class PersistedCounter : public MonotonicallyIncreasingCounter<T>
...
+        bool shouldDoEpochUpdate = ((MonotonicallyIncreasingCounter<T>::GetValue() + value) >= mNextEpoch) ||
+            (MonotonicallyIncreasingCounter<T>::GetValue() > std::numeric_limits<T>::max() - value);
+
+        ReturnErrorOnFailure(MonotonicallyIncreasingCounter<T>::AdvanceBy(value));
+
+        if (shouldDoEpochUpdate)
+        {
+            // Since AdvanceBy allows the counter to be increase by an arbitrary value, it is possible that the new counter value is
+            // greater than mNextEpoch + mEpoch. As such, we want the next Epoch value to be calculate from the new current
+            // value.
```
bzbarsky-apple, line 143:
```suggestion
            // Since AdvanceBy allows the counter to be increased by an arbitrary value, it is possible that the new counter value is
            // greater than mNextEpoch + mEpoch. As such, we want the next Epoch value to be calculated from the new current
            // value.
```

bzbarsky-apple (approved):

src/protocols/secure_channel/CheckInCounter.cpp
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+#include <protocols/secure_channel/CheckInCounter.h>
+#include <stdint.h>
+
+namespace chip {
+namespace Protocols {
+namespace SecureChannel {
+
+CHIP_ERROR CheckInCounter::InvalidateHalfCheckInCouterValues()
```
tcarmelveilleux, line 29: "Couter" is a typo

src/protocols/secure_channel/CheckInCounter.h
```diff
@@ -45,6 +46,24 @@ class CheckInCounter : public PersistedCounter<uint32_t>
...
     inline uint32_t GetNextCheckInCounterValue() { return (mCounterValue + 1); };
 
+    /**
+     * @brief Advances the current Check-In counter value to invalidate half of the values.
+     *        Function adds UINT32_MAX / 2 to the current counter value. Next time a Check-In message is sent, 1 is added to the
+     *        current value.
+     *
+     * @return CHIP_ERROR Any error returned by a write to persisted storage.
+     */
+    CHIP_ERROR InvalidateHalfCheckInCouterValues();
```
tcarmelveilleux, line 56: "Couter" is a typo

src/app/icd/server/ICDManager.cpp
```diff
@@ -666,6 +668,14 @@ CHIP_ERROR ICDManager::HandleEventTrigger(uint64_t eventTrigger)
     case ICDTestEventTriggerEvent::kRemoveActiveModeReq:
         SetKeepActiveModeRequirements(KeepActiveFlag::kTestEventTriggerActiveMode, false);
         break;
+#if CHIP_CONFIG_ENABLE_ICD_CIP
```
tcarmelveilleux, line 671: When is this not enabled?

mkardous-silabs: For SIT devices that still use the TestEventTrigger but might not have the Check-In Protocol support.

PR #33184, opened 2024-04-26
Darwin: Move preWarmCommissioningSession to MTRDeviceControllerFactory

ksperling-apple:
This makes it possible to start pre-warming before a MTRDeviceController has been created, since doing that requires the client to have obtained various configuration details.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDeviceController.h
+++ b/src/darwin/Framework/CHIP/MTRDeviceController.h
@@ -178,7 +178,7 @@ MTR_AVAILABLE(ios(16.1), macos(13.0), watchos(9.1), tvos(16.1))
                                                         error:(NSError * __autoreleasing *)error
     MTR_AVAILABLE(ios(16.4), macos(13.3), watchos(9.4), tvos(16.4));
 
-- (void)preWarmCommissioningSession MTR_NEWLY_DEPRECATED("-[MTRDeviceControllerFactory preWarmCommissioningSession]"); // MTR_AVAILABLE(ios(16.4), macos(13.3), watchos(9.4), tvos(16.4));
+- (void)preWarmCommissioningSession MTR_NEWLY_DEPRECATED("-[MTRDeviceControllerFactory preWarmCommissioningSession]") MTR_AVAILABLE(ios(16.4), macos(13.3), watchos(9.4), tvos(16.4));
 
 /**
  * Set the Delegate for the device controller  as well as the Queue on which the Delegate callbacks will be triggered
--- a/src/darwin/Framework/CHIP/MTRDeviceController.h
+++ b/src/darwin/Framework/CHIP/MTRDeviceController.h
@@ -178,16 +178,7 @@ MTR_AVAILABLE(ios(16.1), macos(13.0), watchos(9.1), tvos(16.1))
                                                         error:(NSError * __autoreleasing *)error
     MTR_AVAILABLE(ios(16.4), macos(13.3), watchos(9.4), tvos(16.4));
 
-/**
- * Optionally pre-warm the controller for setting up a commissioning session.
- * This may be called before setupCommissioningSessionWithPayload if it's known
- * that a commissioning attempt will soon take place but the commissioning
- * payload is not known yet.
- *
- * For example this may do a BLE scan in advance so results are ready earlier
- * once the discriminator is known.
- */
-- (void)preWarmCommissioningSession MTR_AVAILABLE(ios(16.4), macos(13.3), watchos(9.4), tvos(16.4));
+- (void)preWarmCommissioningSession MTR_NEWLY_DEPRECATED("-[MTRDeviceControllerFactory preWarmCommissioningSession]"); // MTR_AVAILABLE(ios(16.4), macos(13.3), watchos(9.4), tvos(16.4));
 
 /**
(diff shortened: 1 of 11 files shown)


src/darwin/Framework/CHIP/MTRDeviceController.h
```diff
@@ -178,16 +178,7 @@ MTR_AVAILABLE(ios(16.1), macos(13.0), watchos(9.1), tvos(16.1))
...
- * Optionally pre-warm the controller for setting up a commissioning session.
- * This may be called before setupCommissioningSessionWithPayload if it's known
- * that a commissioning attempt will soon take place but the commissioning
- * payload is not known yet.
- *
- * For example this may do a BLE scan in advance so results are ready earlier
- * once the discriminator is known.
- */
-- (void)preWarmCommissioningSession MTR_AVAILABLE(ios(16.4), macos(13.3), watchos(9.4), tvos(16.4));
+- (void)preWarmCommissioningSession MTR_NEWLY_DEPRECATED("-[MTRDeviceControllerFactory preWarmCommissioningSession]"); // MTR_AVAILABLE(ios(16.4), macos(13.3), watchos(9.4), tvos(16.4));
```
bzbarsky-apple, line 181:
```suggestion
- (void)preWarmCommissioningSession MTR_NEWLY_DEPRECATED("-[MTRDeviceControllerFactory preWarmCommissioningSession]") MTR_AVAILABLE(ios(16.4), macos(13.3), watchos(9.4), tvos(16.4));
```
so we are ok even if we somehow ship this without fixing the `MTR_NEWLY_DEPRECATED` bit.  You can have both annotations present, especiall since MTR_NEWLY_DEPRECATED expands to nothing, right?

src/darwin/Framework/CHIP/MTRDeviceControllerFactory.h
```diff
@@ -177,6 +177,17 @@ MTR_AVAILABLE(ios(16.4), macos(13.3), watchos(9.4), tvos(16.4))
...
                                                          error:(NSError * __autoreleasing *)error;
 
+/**
+ * If possible, pre-warm the Matter stack for setting up a commissioning session.
+ *
+ * This may be called before -[MTRDeviceController setupCommissioningSessionWithPayload:]
+ * if it is known that a commissioning attempt will soon take place, but the commissioning
+ * payload is not known yet.
+ *
+ * The controller factory must be running for pre-warming to take place.
```
bzbarsky-apple, line 187:
```suggestion
 * The controller factory must be running for pre-warming to take place.  Pre-warming can take place
 * before any controllers are started.
```

src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
```diff
@@ -742,6 +755,32 @@ - (MTRDeviceController * _Nullable)createControllerOnNewFabric:(MTRDeviceControl
                                   error:error];
 }
 
+- (void)preWarmCommissioningSession
+{
+    dispatch_async(_chipWorkQueue, ^{
+        CHIP_ERROR err = CHIP_ERROR_INCORRECT_STATE;
+        if (!self->_running) {
+            MTR_LOG_ERROR("Can't pre-warm, Matter controller factory is not running");
```
bzbarsky-apple, line 763: Do we still want to record the metric in this case?

src/platform/Darwin/PlatformManagerImpl.h
```diff
@@ -49,9 +50,8 @@ class PlatformManagerImpl final : public PlatformManager, public Internal::Gener
     dispatch_queue_t GetWorkQueue() { return mWorkQueue; }
     bool IsWorkQueueCurrentQueue() const;
 
-    CHIP_ERROR StartBleScan(BleScannerDelegate * delegate = nullptr);
+    CHIP_ERROR StartBleScan(BleScannerDelegate * delegate, BleScanMode mode = BleScanMode::kDefault);
```
bzbarsky-apple, line 53: The invariant about how null delegate is allowed if mode is kPreWarm should be documented here, right?

That said, at this point do we have any non-delegate callers of this API?  If not, removing that ability in a follow-up would be good.

bzbarsky-apple (approved):

src/darwin/Framework/CHIP/MTRDeviceController.h
ksperling-apple: Ah yes, it's just `_MTR_IMPLICIT_EXPORT` so it's fine to have both.

src/platform/Darwin/PlatformManagerImpl.h
ksperling-apple: I didn't come to any great conclusions on this... maybe I'll just remove the precondition for now. We currently handle null delegate in the code, and all delegate methods now have an empty default implementation, so functionally there isn't really a difference between null delegate and one that does nothing.

src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
ksperling-apple: We'll get the error code in the metric, so I think I'll leave it in for now.

ksperling-apple:

PR #33189, opened 2024-04-26
[zephyr] Added several improvements to the BLEMgr implementation

kkasperczyk-no:
Added several improvements to the Zephyr BLEMgr, that allows to:

- Support Bluetooth LE bonding and more than one Bluetooth LE identity, what enables using different type of Bluetooth LE address than static random (for non-Matter related BLE services)
- Filtering Zephyr BLE events to not handle them in the BLEMgr if there are not related to CHIPoBLE service, but they come from the other services that can be used on the device
- Restart Bluetooth LE advertising in case of failures, as so far if the first attempt fails the API does not allow to try again later (returned state is enabled, but it is actually not working).

### DIFF
--- a/src/platform/Zephyr/BLEAdvertisingArbiter.cpp
+++ b/src/platform/Zephyr/BLEAdvertisingArbiter.cpp
@@ -29,6 +29,9 @@ namespace {
 // List of advertising requests ordered by priority
 sys_slist_t sRequests;
 
+bool sIsInitialized;
+uint8_t sBtId;
+
 // Cast an intrusive list node to the containing request object
 const BLEAdvertisingArbiter::Request & ToRequest(const sys_snode_t * node)
 {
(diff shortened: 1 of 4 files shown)


markaj-nordic (approved):

src/platform/Zephyr/BLEAdvertisingArbiter.cpp
```diff
@@ -29,6 +29,9 @@ namespace {
 // List of advertising requests ordered by priority
 sys_slist_t sRequests;
 
+bool sIsInitialized;
+uint8_t sBtId;
```
andy31415, line 33: could we initialize these with something (false and 0 maybe)?

src/platform/Zephyr/BLEAdvertisingArbiter.h
```diff
@@ -61,6 +61,17 @@ struct Request : public sys_snode_t
     OnAdvertisingStopped onStopped;       ///< (Optional) Callback invoked when the request stops being top-priority.
 };
 
+/**
+ * @brief Initialize BLE advertising arbiter
+ *
+ * @note This method must be called before trying to insert or cancel any requests.
+ *
+ * @param btId   Local Bluetooth LE identifier to be used for the advertising parameters.
```
andy31415, line 69: is there any `Uninit/Shutdown` that will set "initialized" to false?

or if not, it seems that the btId is sticky and forever.  Maybe this should be explained in the documentation and why this lack of shutdown is correct.

kkasperczyk-no: Well, similarly to many other components (like BLEManager) it was not designed to be uninitialized and the assumption is it will be run once and work until reboot. 

> it seems that the btId is sticky and forever
Yes, indeed I also spotted it and I'm going to fix this in the next PR (I can create issue for improvement to track this). I believe that Bluetooth ID could be a part of `Request` structure that is passed to the `InsertRequest` method, so different bluetooth ids could be used for the following requests.

kkasperczyk-no: I updated docs with the information that for now id is fixed and created issue to improve it: https://github.com/project-chip/connectedhomeip/issues/33310

LuDuda (approved):
Thanks :+1:

andy31415 (approved):

PR #33191, opened 2024-04-26
[thread] Enhance/Optimize attach to thread network

kkasperczyk-no:
Enhance attaching to Thread network by verifying if the current dataset is the same as provided.
In the current solution implementation will not reset the Thread interface while the dataset
are equal to the active dataset.


Damian-Nordic (approved):
Worth mentioning in the PR description that this is to avoid interface reset on fail-safe timeout (if network credentials have not been changed since the fail-safe was armed). At least I believe that was the purpose.

bzbarsky-apple (approved):

PR #33197, opened 2024-04-26
Include what you use for some chef files

andy31415:
Got chef compilation failures like:

```
...
examples/chef/common/clusters/target-navigator/TargetNavigatorManager.h:31:37: error: expected ‘)’ before ‘<’ token
31 |     TargetNavigatorManager(std::list<std::string> targets, uint8_t currentTarget);
```

and it turns out chef was using a lot of STL without validating that those includes actually were made. Applied a few that I found during a simple grep. Hopefully this at least makes things compile.


tehampson (approved):

andy31415:
Fast track chef-specific (test-only) code.

PR #33200, opened 2024-04-26
[chore] Replace all usages of chip::Optional with std::optional in src/lib/dnssd

andy31415:
Replace remaining usages of chip optional in src/lib/dnssd with std::optional.
Expecting some minor flash savings.


src/lib/dnssd/Advertiser_ImplMinimalMdns.cpp
```diff
@@ -670,10 +671,10 @@ CHIP_ERROR AdvertiserMinMdns::Advertise(const CommissionAdvertisingParameters &
         }
     }
 
-    if (params.GetVendorId().HasValue())
+    const auto vendorId = params.GetVendorId();
```
bzbarsky-apple, line 674: Doesn't this make a copy?  Do we want that?  Or should this be `const auto &`?

src/lib/dnssd/Advertiser_ImplMinimalMdns.cpp
```diff
@@ -688,10 +689,10 @@ CHIP_ERROR AdvertiserMinMdns::Advertise(const CommissionAdvertisingParameters &
         }
     }
 
-    if (params.GetDeviceType().HasValue())
+    const auto deviceType = params.GetDeviceType();
```
bzbarsky-apple, line 692: Again, should this be a ref?

src/lib/dnssd/Advertiser_ImplMinimalMdns.cpp
```diff
@@ -812,28 +813,34 @@ FullQName AdvertiserMinMdns::GetCommissioningTxtEntries(const CommissionAdvertis
                                                                                            : &mQueryResponderAllocatorCommissioner;
 
     char txtVidPid[chip::Dnssd::kKeyVendorProductMaxLength + 4];
-    if (params.GetProductId().HasValue() && params.GetVendorId().HasValue())
+
+    const auto productId = params.GetProductId();
+    const auto vendorId  = params.GetVendorId();
```
bzbarsky-apple, line 818:
```suggestion
    const auto & productId = params.GetProductId();
    const auto & vendorId  = params.GetVendorId();
```

src/lib/dnssd/Advertiser_ImplMinimalMdns.cpp
```diff
@@ -858,27 +865,30 @@ FullQName AdvertiserMinMdns::GetCommissioningTxtEntries(const CommissionAdvertis
         snprintf(txtCommissioningMode, sizeof(txtCommissioningMode), "CM=%d", static_cast<int>(params.GetCommissioningMode()));
         txtFields[numTxtFields++] = txtCommissioningMode;
 
-        if (params.GetRotatingDeviceId().HasValue())
+        const auto rotatingDeviceId = params.GetRotatingDeviceId();
```
bzbarsky-apple, line 868:
```suggestion
        const auto & rotatingDeviceId = params.GetRotatingDeviceId();
```

src/lib/dnssd/Advertiser_ImplMinimalMdns.cpp
```diff
@@ -858,27 +865,30 @@ FullQName AdvertiserMinMdns::GetCommissioningTxtEntries(const CommissionAdvertis
...
+        if (rotatingDeviceId.has_value())
         {
-            snprintf(txtRotatingDeviceId, sizeof(txtRotatingDeviceId), "RI=%s", params.GetRotatingDeviceId().Value());
+            snprintf(txtRotatingDeviceId, sizeof(txtRotatingDeviceId), "RI=%s", *rotatingDeviceId);
             txtFields[numTxtFields++] = txtRotatingDeviceId;
         }
 
-        if (params.GetPairingHint().HasValue())
+        const auto pairingHint = params.GetPairingHint();
```
bzbarsky-apple, line 875: I'll stop commenting on this, but fix them all up, please, if the copies are unintended.

src/lib/dnssd/Discovery_ImplPlatform.cpp
```diff
@@ -72,14 +72,16 @@ static void HandleNodeBrowse(void * context, DnssdService * services, size_t ser
         discoveryContext->Retain();
         // For some platforms browsed services are already resolved, so verify if resolve is really needed or call resolve callback
 
+        auto ipAddress = services[i].mAddress;
```
bzbarsky-apple, line 75: Again, is this copy intended?

src/lib/dnssd/Advertiser_ImplMinimalMdns.cpp
andy31415: The return value was `chip::optional<...>` and now it is std::optional ... both without reference. I think we can treat small optionals the same as just simple values and copy by value. Only structures may be worth it to reference. Looking for tnose now, assuming MRPConfig falls under that only.

src/lib/dnssd/Advertiser_ImplMinimalMdns.cpp
andy31415: All of these are like `Optional<uint16_t> GetVendorId() const { return mVendorId; }` today and the same (by value) after std::optional conversion.

I can make them return `const std::optional<uint16_t>&` however does that really save space? I am not sure they will ....

src/lib/dnssd/Advertiser_ImplMinimalMdns.cpp
andy31415: I think they are intended or at least they were pre-existing, however I updated the code.

The main reason for this was that clang-tidy was complaining of "unchecked optional access" when we did has_value followed by value when the optional is from a method call (since the method call may return different values each time).

src/lib/dnssd/Advertiser_ImplMinimalMdns.cpp
```diff
@@ -670,39 +673,43 @@ CHIP_ERROR AdvertiserMinMdns::Advertise(const CommissionAdvertisingParameters &
...
-        FullQName vendorServiceName =
-            allocator->AllocateQName(nameBuffer, kSubtypeServiceNamePart, serviceType, kCommissionProtocol, kLocalDomain);
-        ReturnErrorCodeIf(vendorServiceName.nameCount == 0, CHIP_ERROR_NO_MEMORY);
-
-        if (!allocator->AddResponder<PtrResponder>(vendorServiceName, instanceName)
-                 .SetReportAdditional(instanceName)
-                 .SetReportInServiceListing(true)
-                 .IsValid())
+        const auto & vendorId = params.GetVendorId();
+        if (vendorId.has_value())
```
Damian-Nordic, line 678: Super nit but in C++17, you can actually get rid off those scopes with:
```suggestion
        if (const auto & vendorId = params.GetVendorId(); vendorId.has_value())
```

src/lib/dnssd/Advertiser_ImplMinimalMdns.cpp
```diff
@@ -812,29 +819,39 @@ FullQName AdvertiserMinMdns::GetCommissioningTxtEntries(const CommissionAdvertis
...
+            txtFields[numTxtFields++] = txtDeviceType;
+        }
     }
 
     char txtDeviceName[chip::Dnssd::kKeyDeviceNameMaxLength + 4];
-    if (params.GetDeviceName().HasValue())
     {
-        snprintf(txtDeviceName, sizeof(txtDeviceName), "DN=%s", params.GetDeviceName().Value());
-        txtFields[numTxtFields++] = txtDeviceName;
+        const auto deviceName = params.GetDeviceName();
```
Damian-Nordic, line 849:
```suggestion
        const auto& deviceName = params.GetDeviceName();
```
to make it consistent with other fields.

src/lib/dnssd/Advertiser_ImplMinimalMdns.cpp
```diff
@@ -858,27 +875,36 @@ FullQName AdvertiserMinMdns::GetCommissioningTxtEntries(const CommissionAdvertis
         snprintf(txtCommissioningMode, sizeof(txtCommissioningMode), "CM=%d", static_cast<int>(params.GetCommissioningMode()));
         txtFields[numTxtFields++] = txtCommissioningMode;
 
-        if (params.GetRotatingDeviceId().HasValue())
         {
-            snprintf(txtRotatingDeviceId, sizeof(txtRotatingDeviceId), "RI=%s", params.GetRotatingDeviceId().Value());
-            txtFields[numTxtFields++] = txtRotatingDeviceId;
+            const auto rotatingDeviceId = params.GetRotatingDeviceId();
```
Damian-Nordic, line 879:
```suggestion
            const auto& rotatingDeviceId = params.GetRotatingDeviceId();
```

src/lib/dnssd/Advertiser_ImplMinimalMdns.cpp
```diff
@@ -858,27 +875,36 @@ FullQName AdvertiserMinMdns::GetCommissioningTxtEntries(const CommissionAdvertis
...
+                snprintf(txtPairingHint, sizeof(txtPairingHint), "PH=%d", *pairingHint);
+                txtFields[numTxtFields++] = txtPairingHint;
+            }
         }
 
-        if (params.GetPairingInstruction().HasValue())
         {
-            snprintf(txtPairingInstr, sizeof(txtPairingInstr), "PI=%s", params.GetPairingInstruction().Value());
-            txtFields[numTxtFields++] = txtPairingInstr;
+            const auto pairingInstruction = params.GetPairingInstruction();
```
Damian-Nordic, line 897:
```suggestion
            const auto& pairingInstruction = params.GetPairingInstruction();
```

Damian-Nordic (approved):

src/lib/dnssd/Advertiser.h
```diff
@@ -95,20 +95,20 @@ class BaseAdvertisingParams
...
 
     // NOTE: The SetTcpSupported API is deprecated and not compliant with 1.3. T flag should not be set.
-    Derived & SetTcpSupported(Optional<bool> tcpSupported)
+    Derived & SetTcpSupported(std::optional<bool> tcpSupported)
     {
         mTcpSupported = tcpSupported;
         return *reinterpret_cast<Derived *>(this);
     }
-    Optional<bool> GetTcpSupported() const { return mTcpSupported; }
+    const std::optional<bool> & GetTcpSupported() const { return mTcpSupported; }
```
bzbarsky-apple, line 111: Why are we changing this, and the vendor id, etc, to return references?

I guess it's OK, but I was just talking about whether the _callers_ should force copies, not whether the _callees_ should be changed to not copy.  I guess this is fine, probably....

src/lib/dnssd/Discovery_ImplPlatform.cpp
```diff
@@ -760,16 +762,15 @@ CHIP_ERROR DiscoveryImplPlatform::StartDiscovery(DiscoveryType type, DiscoveryFi
 
 CHIP_ERROR DiscoveryImplPlatform::StopDiscovery(DiscoveryContext & context)
 {
-    if (!context.GetBrowseIdentifier().HasValue())
+    const auto browseIdentifier = context.GetBrowseIdentifier();
```
bzbarsky-apple, line 765:
```suggestion
    const std::optional<intptr_t> browseIdentifier = context.GetBrowseIdentifier();
```
is probably clearer in terms of "this will copy here, so the clear below is ok".

bzbarsky-apple (approved):

src/lib/dnssd/Advertiser_ImplMinimalMdns.cpp
andy31415: This is neat! I will do so.

src/lib/dnssd/Advertiser.h
andy31415: Reverted. I thought it odd for us to do `auto &x = foo.GetTcpSupported()` when the `GetTcpSupported()` returns by value rather than reference since then it is somewhat unclear why we even ask for a reference to a temporary. I.e. I assumed this is what you asked for.

If you did not ask for it, I am fine to do minimal code change instead.

PR #33204, opened 2024-04-26
[LVL control] Return invalid command on move when rate is 0 or rate is null  and de…

jmartinez-silabs:
…faultRate is 0

This PR syncs the SDK in line with this spec PR
https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/9140

As a move of 0 doesn't make much sense.
Details about the decision can be found in this Slack thread
https://csamembers.slack.com/archives/C01EWUSBT0C/p1713794720672179

I also update the test plan to change the expected command response:
https://github.com/CHIP-Specifications/chip-test-plans/pull/4139

**Edit: Latest updates**
- Added checks for stepSize != 0 also since spec pr also update this now. 
- set a Min value of 1 for Default Move Rate attribute
- Bumped the cluster revision + updated zap file that were using the cluster + regen


rcasallas-silabs (approved):

rcasallas-silabs (approved):

src/app/clusters/level-control/level-control.cpp
```diff
@@ -960,18 +960,56 @@ static void moveHandler(app::CommandHandler * commandObj, const app::ConcreteCom
...
+        app::DataModel::Nullable<uint8_t> defaultMoveRate;
+        status = Attributes::DefaultMoveRate::Get(endpoint, defaultMoveRate);
+        if (status != Status::Success || defaultMoveRate.IsNull())
+        {
+            ChipLogProgress(Zcl, "ERR: reading default move rate %x", to_underlying(status));
+            eventDuration = FASTEST_TRANSITION_TIME_MS;
+        }
+        else
+        {
+            // The defaultMoveRate cannot be 0
```
mkardous-silabs, line 986:
```suggestion
            // Error out if DefaultMoveRate is equal to 0
```
since it technically can be equal to 0

src/app/clusters/level-control/level-control.cpp
```diff
@@ -960,18 +960,56 @@ static void moveHandler(app::CommandHandler * commandObj, const app::ConcreteCom
...
-        ChipLogProgress(Zcl, "Immediate success due to move rate of 0 (would move at no rate).");
-        status = Status::Success;
+        // Providing rate of 0 is not allowed.
+        status = Status::InvalidCommand;
         goto send_default_response;
     }
 
+    uint8_t eventDuration; // use this local var so state->eventDurationMs is only set once the command is validated.
+#ifndef IGNORE_LEVEL_CONTROL_CLUSTER_TRANSITION
+    // If the Rate field is null, the device should move at the default move rate, if available,
```
mkardous-silabs, line 973: I think this should be outisde the ifndef since it describes the #if-else behavior, no?

mkardous-silabs (approved):

src/app/clusters/level-control/level-control.cpp
jmartinez-silabs: no, this only describes the condition we use the default rate if the rate is null. Outside the `#ifndef IGNORE_LEVEL_CONTROL_CLUSTER_TRANSITION`, it ignores transition and runs at max speed regardless of rate or default rate setting

src/app/clusters/level-control/level-control.cpp
```diff
@@ -960,18 +960,56 @@ static void moveHandler(app::CommandHandler * commandObj, const app::ConcreteCom
...
     }
 
+    // Always validate the rate parameter received. A rate of 0 is invalid.
     if (!rate.IsNull() && (rate.Value() == 0))
     {
-        // Move at a rate of zero is no move at all. Immediately succeed without touching anything.
-        ChipLogProgress(Zcl, "Immediate success due to move rate of 0 (would move at no rate).");
-        status = Status::Success;
+        // Providing rate of 0 is not allowed.
+        status = Status::InvalidCommand;
```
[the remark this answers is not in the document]
jmartinez-silabs: i'll bump the release version too then. It shall be done on the spec pr also.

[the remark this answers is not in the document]
Apollon77: Would this then also require to splitup/duplicate the test files to have one for current revision that tests "as is" and one for other revision with changed behavior? In fact when I see it right would be the first "breaking" change,right?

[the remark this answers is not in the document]
cecille: This spec change would apply to the 1.4 spec, so the 1.4 test cases would be for the new revision. Other ones wouldn't change barring a specific CCB.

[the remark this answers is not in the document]
jmartinez-silabs: This will apply to 1.4 and up. not 1.3 and prior

src/app/clusters/level-control/level-control.cpp
```diff
@@ -960,18 +960,56 @@ static void moveHandler(app::CommandHandler * commandObj, const app::ConcreteCom
         goto send_default_response;
     }
 
+    // Always validate the rate parameter received. A rate of 0 is invalid.
```
bzbarsky-apple, line 963: I don't think it's the intent of the spec changes that rate == 0 plus mode == stop should stop working...  But maybe it is?  I asked these questions on the spec PR, and we should make sure we don't make changes here until the spec is sorted.

src/app/clusters/level-control/level-control.cpp
```diff
@@ -960,18 +960,56 @@ static void moveHandler(app::CommandHandler * commandObj, const app::ConcreteCom
...
+        app::DataModel::Nullable<uint8_t> defaultMoveRate;
+        status = Attributes::DefaultMoveRate::Get(endpoint, defaultMoveRate);
+        if (status != Status::Success || defaultMoveRate.IsNull())
+        {
+            ChipLogProgress(Zcl, "ERR: reading default move rate %x", to_underlying(status));
+            eventDuration = FASTEST_TRANSITION_TIME_MS;
+        }
+        else
+        {
+            // Error out if DefaultMoveRate is equal to 0
```
bzbarsky-apple, line 986: The spec has no provisions for this; another question that was raised on the spec PR.

bzbarsky-apple (changes requested):

src/app/clusters/level-control/level-control.cpp
jmartinez-silabs: The level control movemode enum has no stop. The level control has a separate stop command to stop transitions.

https://github.com/CHIP-Specifications/connectedhomeip-spec/blob/master/src/app_clusters/LevelControl.adoc#52-movemodeenum-type

Your comment does applied to color control which I am also working on for a separate pr.

src/app/clusters/level-control/level-control.cpp
jmartinez-silabs: You are correct, the spec must clarify this. This implementation comes from the Slack conversation where this was the agreed behaviour.

I'll wait for spec to change or merge this

jmartinez-silabs: So the spec pr changes the min value for DefaultMoveRate to 1. I updated the level control XML with this change. Going forward, it will not be possible to set a DefaultMoveRate of 0.

On the off chance that someone configured the DefaultMoveRate to 0 with a previous matter version, this code would return an invalid command.

I am not sure if I should leave this behaviour as is, since 0 is an invalid value for the attribute, or rather move at the fastest rate possible as if it was null since it was previously possible to set it to 0

src/app/clusters/level-control/level-control.cpp
bzbarsky-apple: Ah, sorry.  Too many things named similarly.  :(

src/app/clusters/level-control/level-control.cpp
bzbarsky-apple: That sounds like a spec question, really; they should talk about the backwards compat behavior in there.

src/app/clusters/level-control/level-control.cpp
```diff
@@ -950,14 +950,21 @@ static void moveHandler(app::CommandHandler * commandObj, const app::ConcreteCom
...
     Status status;
-    app::DataModel::Nullable<uint8_t> currentLevel;
     uint8_t difference;
+    EmberAfLevelControlState * state;
+    app::DataModel::Nullable<uint8_t> currentLevel;
 
+    EndpointId endpoint = commandPath.mEndpointId;
+    CommandId commandId = commandPath.mCommandId;
+    // Validate the received rate and moveMode first.
+    if ((!rate.IsNull() && (rate.Value() == 0)) || moveMode == MoveModeEnum::kUnknownEnumValue)
```
andy31415, line 961:
```suggestion
    if (rate.ValueOr(1) == 0) || moveMode == MoveModeEnum::kUnknownEnumValue)
```

Although I believe even `rate == 0` may actually do the right thing.

andy31415 (approved):

src/app/clusters/level-control/level-control.cpp
bzbarsky-apple: Yes, `rate == 0` should do the right thing.

src/app/clusters/level-control/level-control.cpp
```diff
@@ -970,18 +977,48 @@ static void moveHandler(app::CommandHandler * commandObj, const app::ConcreteCom
...
+        app::DataModel::Nullable<uint8_t> defaultMoveRate;
+        status = Attributes::DefaultMoveRate::Get(endpoint, defaultMoveRate);
+        if (status != Status::Success || defaultMoveRate.IsNull())
+        {
+            ChipLogProgress(Zcl, "ERR: reading default move rate %x", to_underlying(status));
+            eventDuration = FASTEST_TRANSITION_TIME_MS;
+        }
+        else
+        {
+            // Error out if DefaultMoveRate is equal to 0
```
bzbarsky-apple, line 995: Please document that per spec that is not supposed to happen but we may have an old (now invalid) value stored?

src/app/clusters/level-control/level-control.cpp
```diff
@@ -970,18 +977,48 @@ static void moveHandler(app::CommandHandler * commandObj, const app::ConcreteCom
...
+            if (defaultMoveRate.Value() == 0)
+            {
+                status = Status::InvalidCommand;
```
bzbarsky-apple, line 998: This is not ideal, since it claims something is wrong with the command that was sent, but something is actually wrong with the stored state of the device.

Ideally the spec would define the behavior here.  Until it does, I'd suggest a more relevant error code might be InvalidInState or Failure or something.

bzbarsky-apple (approved):

src/app/clusters/level-control/level-control.cpp
jmartinez-silabs: Is it new that nullable variables have a direct the == operator to the value? Anyway I'll test/change that then

jmartinez-silabs: So rate==0 errors out with
`error: ambiguous overload for 'operator==' (operand types are 'chip::app::DataModel::Nullable<unsigned char>' and 'int')`

casting the 0 as uin8_t here works. I don't know if there is a cleaner way to ensure the == operator is with the proper type.

bzbarsky-apple: > Is it new that nullable variables have a direct the == operator to the value?

Yes.  That said, the fact that for integers it's got those overload issues is annoying.  :(

PR #33208, opened 2024-04-29
[Ameba] Add Ameba crypto implementation

pankore:
* Add Ameba cryto implementation for security purposes


src/platform/Ameba/BUILD.gn
```diff
@@ -71,4 +71,14 @@ static_library("Ameba") {
       "AmebaOTAImageProcessor.h",
     ]
   }
+
+  if (chip_crypto == "platform") {
```
Damian-Nordic, line 75: Do you actually need platform crypto if you're building `mbedtls` anyway? I mean even if you use `chip_crypto = "mbedtls"`, you can still inject a custom operational keystore.

Damian-Nordic (approved):

andy31415 (approved):

src/platform/Ameba/BUILD.gn
pankore: You are right, its not necessary to utilize platform crypto in this case. I have made the changes.

PR #33214, opened 2024-04-29
[K32W] Fix apply action corner case in OTATlvProcessor interface

marius-alex-tache:
`OTATlvProcessor::ApplyAction` now has a default implementation, but derived classes are still able to overwrite it.
 
The flag is used by the default `ApplyAction` implementation. If something goes wrong during `ExitAction` of the TLV processor,
then `mShouldNotApply` should be set to `true` and the image processor  should abort. In this case, the BDX transfer was already finished and calling `CancelImageUpdate` will not abort the transfer, hence the device will reboot even though it should not have. If `ApplyAction` fails during `HandleApply`, then the process will be aborted.

### DIFF
--- a/src/platform/nxp/k32w/common/OTATlvProcessor.cpp
+++ b/src/platform/nxp/k32w/common/OTATlvProcessor.cpp
@@ -31,6 +31,12 @@ namespace chip {
 #if OTA_ENCRYPTION_ENABLE
 constexpr uint8_t au8Iv[] = { 0x00, 0x00, 0x00, 0x10, 0x11, 0x12, 0x13, 0x14, 0x15, 0x16, 0x17, 0x18, 0x00, 0x00, 0x00, 0x00 };
 #endif
+
+CHIP_ERROR OTATlvProcessor::ApplyAction()
+{
+    return mShouldNotApply ? CHIP_OTA_PROCESSOR_SHOULD_NOT_APPLY : CHIP_NO_ERROR;
+}
+
 CHIP_ERROR OTATlvProcessor::Process(ByteSpan & block)
 {
     CHIP_ERROR status     = CHIP_NO_ERROR;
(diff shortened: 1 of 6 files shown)


src/platform/nxp/k32w/common/OTATlvProcessor.cpp
```diff
@@ -31,6 +31,12 @@ namespace chip {
 #if OTA_ENCRYPTION_ENABLE
 constexpr uint8_t au8Iv[] = { 0x00, 0x00, 0x00, 0x10, 0x11, 0x12, 0x13, 0x14, 0x15, 0x16, 0x17, 0x18, 0x00, 0x00, 0x00, 0x00 };
 #endif
+
+CHIP_ERROR OTATlvProcessor::ApplyAction()
+{
+    return mShouldNotApply ? CHIP_OTA_PROCESSOR_SHOULD_NOT_APPLY : CHIP_NO_ERROR;
```
andy31415, line 37: the name `mShouldNotApply` feels like a default negation in it, which is inconvenient for booleans.

Do we need a boolean? How about a `enum class` with values like `kApply` and `kDoNoApply` instead?

src/platform/nxp/k32w/common/OTATlvProcessor.h
```diff
@@ -41,6 +41,7 @@ namespace chip {
 #define CHIP_OTA_PROCESSOR_EEPROM_OFFSET CHIP_ERROR_TLV_PROCESSOR(0x0C)
 #define CHIP_OTA_PROCESSOR_EXTERNAL_STORAGE CHIP_ERROR_TLV_PROCESSOR(0x0D)
 #define CHIP_OTA_PROCESSOR_START_IMAGE CHIP_ERROR_TLV_PROCESSOR(0x0E)
+#define CHIP_OTA_PROCESSOR_SHOULD_NOT_APPLY CHIP_ERROR_TLV_PROCESSOR(0x0F)
```
andy31415, line 44: consider `CHIP_OTA_PROCESOR_DO_NOT_APPLY` as `SHOULD` seems to have some optionality in it.

WHY are these not named as `CHIP_ERROR_....` or `CHIP_OTA_ERROR` ? Otherwise they seem to feel like no-error values.

andy31415 (approved):

PR #33225, opened 2024-04-29
[ICD] Fix comment mistakes from post-merge reviews

mkardous-silabs:
#### Description
These two PRs has post-merge reviews. PR fixes the comments from these PRs.
https://github.com/project-chip/connectedhomeip/pull/33081
https://github.com/project-chip/connectedhomeip/pull/33174

#### Tests
CI since only comment changes and a rename

### DIFF
--- a/src/app/icd/server/ICDManager.h
+++ b/src/app/icd/server/ICDManager.h
@@ -85,7 +85,7 @@ public:
      *         When this event is called, the ICD is still in ActiveMode.
      *        If the ActiveMode timer is increased due to the TransitionToIdle event, the event will not be called a second time in
      *        a given cycle.
-     *        OnEnterIdleMode will always the third event and indictats that the ICD has transitioned to IdleMode.
+     *        OnEnterIdleMode will always the third event and indicates that the ICD has transitioned to IdleMode.
      *
      *        The ICDModeChange event can occur independently from the EnterActiveMode, TransitionToIdle and EnterIdleMode.
      *        It will typpically happen at the ICDManager init when a client is already registered with the ICD before the
--- a/src/app/icd/server/ICDManager.h
+++ b/src/app/icd/server/ICDManager.h
@@ -88,7 +88,7 @@ public:
      *        OnEnterIdleMode will always the third event and indicates that the ICD has transitioned to IdleMode.
      *
      *        The ICDModeChange event can occur independently from the EnterActiveMode, TransitionToIdle and EnterIdleMode.
-     *        It will typpically happen at the ICDManager init when a client is already registered with the ICD before the
+     *        It will typically happen at the ICDManager init when a client is already registered with the ICD before the
      *        OnEnterIdleMode event or when a client sends a register command after the OnEnterActiveMode event. Nothing prevents
      *        the ICDModeChange event from happening multiple times per cycle or while the ICD is in IdleMode.
      *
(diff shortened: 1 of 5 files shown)


jmartinez-silabs (approved):

andy31415 (approved):

src/app/icd/server/ICDManager.h
```diff
@@ -85,12 +85,12 @@ class ICDManager : public ICDListener, public TestEventTriggerHandler
      *         When this event is called, the ICD is still in ActiveMode.
      *        If the ActiveMode timer is increased due to the TransitionToIdle event, the event will not be called a second time in
      *        a given cycle.
-     *        OnEnterIdleMode will always the third when the ICD has transitioned to IdleMode.
+     *        OnEnterIdleMode will always the third event and indictats that the ICD has transitioned to IdleMode.
```
tcarmelveilleux, line 88:
```suggestion
     *        OnEnterIdleMode will always the third event and indicates that the ICD has transitioned to IdleMode.
```

Is it true it's always the 3rd call?

Also "OnEnterIdleMode does not exist anywhere.

mkardous-silabs: Yes, it is always the 3rd call.
You might have a version before it was used. We use it in our example.

src/app/icd/server/ICDManager.h
```diff
@@ -85,12 +85,12 @@ class ICDManager : public ICDListener, public TestEventTriggerHandler
...
      *
      *        The ICDModeChange event can occur independently from the EnterActiveMode, TransitionToIdle and EnterIdleMode.
-     *        It will typpically hapen at the ICDManager init when a client is already registered with the ICD before the
-     *        OnEnterIdleMode event or when a client send a register command after the OnEnterActiveMode event. Nothing prevents the
-     *        ICDModeChange event to happen multiple times per cycle or while the ICD is in IdleMode.
+     *        It will typpically happen at the ICDManager init when a client is already registered with the ICD before the
```
tcarmelveilleux, line 91:
```suggestion
     *        It will typically happen at the ICDManager init when a client is already registered with the ICD before the
```

src/app/icd/server/ICDManager.h
```diff
@@ -85,12 +85,12 @@ class ICDManager : public ICDListener, public TestEventTriggerHandler
...
+     *        OnEnterIdleMode event or when a client sends a register command after the OnEnterActiveMode event. Nothing prevents
+     *        the ICDModeChange event from happening multiple times per cycle or while the ICD is in IdleMode.
```
tcarmelveilleux, line 93: I am having a hard time parsing this whole paragraph. There's a lot of gotchas and MAYs here. What is the actual public API guarantees here...

tcarmelveilleux (approved):

yunhanw-google (approved):

src/app/icd/server/ICDManager.h
mkardous-silabs: 1. OnEnterActiveMode will be called each time we transition from Idle to Active mode. The event is not called when the ActiveMode timer is extended.
2. OnTransitionToIdle will be called the first time we are about to enter IdleMode. The event is only called once per cycle. The event is not called a second time if the ActiveMode timer is increased because of the event.
3. OnEnterIdleMode will be called each time we transition from Active to Idle mode. 
4. onICDModeChange is independent from the first three. Called whenever ICD goes from SIT <-> LIT modes.

PR #33231, opened 2024-04-29
Add some better logging around PASE session setup in Matter.framework.

bzbarsky-apple:
Log the device ID being used.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDeviceController.mm
+++ b/src/darwin/Framework/CHIP/MTRDeviceController.mm
@@ -616,6 +616,8 @@ static inline void emitMetricForSetupPayload(MTRSetupPayload * payload)
                                    newNodeID:(NSNumber *)newNodeID
                                        error:(NSError * __autoreleasing *)error
 {
+    MTR_LOG_DEFAULT("Setting up commissioning session for device ID 0x%016llX with setup payload %@", newNodeID.unsignedLongLongValue, payload);
+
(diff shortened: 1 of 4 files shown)


Damian-Nordic (approved):

yunhanw-google (approved):

src/darwin/Framework/CHIP/MTRDeviceController.mm
```diff
@@ -912,7 +916,10 @@ - (MTRBaseDevice *)deviceBeingCommissionedWithNodeID:(NSNumber *)nodeID error:(N
         return [[MTRBaseDevice alloc] initWithPASEDevice:deviceProxy controller:self];
     };
 
-    return [self syncRunOnWorkQueueWithReturnValue:block error:error];
+    MTRBaseDevice * device = [self syncRunOnWorkQueueWithReturnValue:block error:error];
+    MTR_LOG_DEFAULT("Getting device being commissioned with node ID 0x%016llX: %@ (error: %@)",
```
ksperling-apple, line 920: I don't think what's effectively a getter should be logging this (or at least not at DEFAULT level). A client that's interested in this can easily log the return value of the call themselves.

bzbarsky-apple: My issue is that I have clients who claim they are calling this function and I don't think they are, so I need log evidence for that....

PR #33239, opened 2024-04-30
[ICD] Convert TestICDMonitoringTable to gTest

mkardous-silabs:
#### Description
PR starts the transition of the ICD unit tests to gtest - TestICDMonitoringTable is transitionned to gtest in this PR.

- Move ICD server tests out `app/tests` to `icd/server/tests` since they don't on anything from apps.
- Remove the dependency in TestICDManager on the Apps helpers
- Convert TestICDMonitoringTable to gTest

PR has no underlying behavior changes.

#### Note to reviewers:
TestICDManager will be converted in a follow up because we need to modify the MessagingContext first. `LoopbackMessagingContext` has `CHIP_ERROR SetUpTestSuite()` in its class definition that conflicts with the gtest signature that returns void.

Added Check-In specific tests behind a define guard for TestICDManager. This allows some of the unit tests to always be run as part of the "main" unit test runs.

TestICDMonitoringTable is now part of the main unit test build since it doesn't need any specific build configuration.

#### Tests
Unit test themselves to make sure they still pass


src/BUILD.gn
```diff
@@ -66,6 +66,8 @@ if (chip_build_tests) {
       "${chip_root}/src/protocols/interaction_model/tests",
       "${chip_root}/src/protocols/user_directed_commissioning/tests",
       "${chip_root}/src/transport/retransmit/tests",
+      "${chip_root}src/app/icd/server/tests",
```
andy31415, line 69: Should these tests appear in openiot cmake settings now that these start being visible in the main test list? like https://github.com/project-chip/connectedhomeip/blob/master/src/test_driver/openiotsdk/unit-tests/test_components_nl.txt and https://github.com/project-chip/connectedhomeip/blob/master/src/test_driver/openiotsdk/unit-tests/test_components.txt ?

src/app/BUILD.gn
```diff
@@ -150,6 +150,18 @@ source_set("test-event-trigger") {
   sources = [ "TestEventTriggerDelegate.h" ]
 }
 
+source_set("timer-delegates") {
+  sources = [
+    "TimerDelegates.cpp",
+    "TimerDelegates.h",
```
andy31415, line 156: What are these?

Looking at the content, it looks like they specifically only implement reporting::ReeportScheduler::TimerDelegate and specifically contain a `DefaultTimerDelegate`

Should they maybe reside in `app/reporting` and be named `DefaultTimerDelegate` instead? There is no plural delegates right now and the placement in app is unclear to me. Or alternatively, if this is specifically for icd, should these reside in ICD? Are there plans to add more to this which would make `Default` in the name inappropriate?

mkardous-silabs: Removed this change. I had doen this initially solve a dependency problem. In the end, i was able to remove the full app dependecy from the ICD tests.

src/BUILD.gn
mkardous-silabs: Thanks for the reminder! added the libs to the correct files

andy31415 (approved):

jmartinez-silabs (approved):

yunhanw-google (approved):

Issue #32735, opened 2024-03-26
[CI/CD] Run slow workflows only if fast workflows pass


kliao-csa:
### Workflow issues

I plan to create a master workflow file as the only workflow that runs on pull request, which manages other workflows to run in waves so that slow workflows don't start if fast ones fail in seconds

### Platform

all

kliao-csa:
Getting back to this ticket. The plan:

1. For starters, create a new callable workflow that use built-in Powershell scripting to execute a Github CLI call for workflow statuses and filter that to the status of the associated spellcheck action.
2. If that fails, fail this yaml.
3. Insert a call to this new yaml as a job on all heavy duty on-PR runs; fail instantly if this fails.

kliao-csa:
Initial YAML created

kliao-csa:
https://github.com/project-chip/connectedhomeip/actions/runs/8897889850/job/24433825293?pr=33241 Found a way to access "restyled" and other non-workflow checks that `gh run list` does not see. Seems to error out and there's not an easy way to parse this output, so I will be working on putting this in a Python script.

PR #33241, opened 2024-04-30
Await Fast CI

kliao-csa:
#32735

Created as a draft to prototype functionality


yunhanw-google (approved):

yunhanw-google (changes requested):
Sorry, click wrong approve button, may I get to know the goal for this PR?

kliao-csa:
> Sorry, click wrong approve button, may I get to know the goal for this PR?

This PR will insert a callable job into all CI that triggers on pull request (I'm using Build example - ASR for debug, but this will go into all Build examples) and runs a Python script which polls for the completion of short preliminary checks. restyled and spell check are the primary ones, but this script could be used in other situations to block builds depending on the completion status of other builds.

Use case example:
You may have noticed that the last commit in this PR has failed restyled. I will definitely need to make another commit to fix this issue; in the meantime, all builds currently being run are useless because they will be re-triggered after the style cleanup. My PR would cancel them as soon as restyled is detected to have failed, saving a lot of resources on CI.

kliao-csa (commented):
This PR will insert a callable job into all CI that triggers on pull request (I'm using Build example - ASR for debug, but this will go into all Build examples) and runs a Python script which polls for the completion of short preliminary checks. restyled and spell check are the primary ones, but this script could be used in other situations to block builds depending on the completion status of other builds.

Use case example:
You may have noticed that the last commit in this PR has failed restyled. I will definitely need to make another commit to fix this issue; in the meantime, all builds currently being run are useless because they will be re-triggered after the style cleanup. My PR would cancel them as soon as restyled is detected to have failed, saving a lot of resources on CI.

scripts/helpers/poll_quick_ci.py
```diff
@@ -0,0 +1,40 @@
...
+                    break
+                elif "pass" in line:
+                    print(f"Fast CI {args.check} has passed; if all fast CI passes, builds and tests may commence.")
+                    exit(0)
+                elif "fail" in line:
+                    print(f"Fast CI {args.check} has failed; please resolve this issue before running builds and tests.")
+                    exit(1)
+
+    print(
+        f"Polling for completion of fast CI {args.check} failed. Please ensure the name of the check was entered correctly and verify that it should take less than {poll_max} seconds to run.")
```
cecille, line 35: Does this imply that if the fast jobs don't complete within 300s that this will fail and we'll cancel all the remaining CI jobs?

I think in this case maybe we should just let the complete? Thoughts @andy31415 ?

scripts/helpers/poll_quick_ci.py
```diff
@@ -0,0 +1,40 @@
...
+    parser.add_argument("check", help="Pull request check to poll for")
+    args = parser.parse_args()
+
+    print(f"Gathering info on the {args.check} check being run for the current pull request.")
+    polling = True
+    poll_count = 0
+    poll_max = 300
+    while (polling):
+        for line in subprocess.run(f"gh pr checks -R project-chip/connectedhomeip {args.pr}", stdout=subprocess.PIPE, shell=True).stdout.decode("utf-8").splitlines():
+            if args.check in line:
```
cecille, line 18: This feels fragile - you're parsing the log line, but if the log line changes and you get back something that doesn't have args.check, or "pending", "pass" or "fail" then you'll loop forever. 

If you're going to use polling here, the count increment needs to be outside of the nested if.

cecille:
@yunhanw-google - please see attached issue. The goal is to cancel the slow-running CI targets when the fast targets fail, so we don't spend multiple hours of CI time running jobs when we know the PR will need to be updated.

arkq:
> The goal is to cancel the slow-running CI targets when the fast targets fail, so we don't spend multiple hours of CI time running jobs when we know the PR will need to be updated.

And what about the case when someone would like to know all failing platforms (workflows) in advance? In case of fast failure, CI will report only the first one. Then one will fix it and learn in another x minutes that another platform (workflow) fails... I know that preferably it would be nice to test everything locally before pushing to upstream, but it's not that simple with constrained development environment (storage, RAM, CPU) and so many platforms to test. Other possibility is to test CI on local fork and then create PR. However, such approach will double the CI time in most cases (there was already a case reported by GitHub, that this project uses too much GitHub Actions resources dedicated for ALL open source projects). I hope that we will not end up in a case similar to Darwin matrix workflow were one flaky workflow stops others, so instead of rerunning one job it is required to rerun two or three.

arkq:
For simple cases like CLA or restyled there is already an action that can pause a job before other is completed: https://github.com/marketplace/actions/wait-on-check

IMHO it's better not to spawn any jobs if restyle (or other simple check like that) does not pass. So, for such approach see https://github.com/marketplace/actions/wait-on-check#alternatives maybe one of these will be useful.

kliao-csa:
> For simple cases like CLA or restyled there is already an action that can pause a job before other is completed: https://github.com/marketplace/actions/wait-on-check
> 
> IMHO it's better not to spawn any jobs if restyle (or other simple check like that) does not pass. So, for such approach see https://github.com/marketplace/actions/wait-on-check#alternatives maybe one of these will be useful.

Seems these options run into the same issues as the stop on fail action when I tested them, where permissions are missing and appear to be unobtainable with even an allow-everything token.

scripts/helpers/poll_quick_ci.py
kliao-csa: Moved increment outside.

scripts/helpers/poll_quick_ci.py
kliao-csa: Changed the timeout to 1 hour - if it takes longer, that implies the job is hanging.

andy31415:
Closing this as obsolete - we have a separate implementation since #36000

PR #33243, opened 2024-04-30
[shell] Reduce memory usage and boileplate code for subcommands

Damian-Nordic:
1. Add lightweight `CommandSet` class to prevent instantiating `Shell::Engine`, which allows for dynamic command registration, just to create a subshell with a fixed number of subcommands.

   This slightly reduces the RAM usage. Additionally, the command set automatically prints the help texts if no command or "help" command has been submitted.
2. Add SubShellCommand templatized command handler to remove the boilerplate associated with creating a subshell.

### DIFF
--- a/src/lib/shell/Command.h
+++ b/src/lib/shell/Command.h
@@ -1,0 +1,62 @@
+/*
+ *    Copyright (c) 2024 Project CHIP Authors
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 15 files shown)


src/lib/shell/Command.h
```diff
@@ -0,0 +1,62 @@
...
+ */
+struct Command
+{
+    /**
+     * Shell command handler function type.
+     *
+     * @param argc  Number of arguments in argv.
+     * @param argv  Array of arguments in the tokenized command line to execute.
+     */
+    using Handler = CHIP_ERROR (*)(int argc, char * argv[]);
```
andy31415, line 50: should argv be maybe `const char *` to not be modifiable?

andy31415 (approved):

src/lib/shell/Command.h
Damian-Nordic: That would make sense to me though I would have to modify dozens of examples that define handlers using the current signature, so I'd prefer not doing that in this PR :)

yufengwangca (approved):

jmartinez-silabs (approved):

PR #33246, opened 2024-04-30
Add python platform for perfetto submodule

cecille:
We need this to compile the python wheel, which is not strictly one of the host platforms. Adding a python platform designation so we can specify this submodule without pulling in the other host platform dependencies.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -309,7 +309,7 @@
 	path = third_party/perfetto/repo
 	url = https://github.com/google/perfetto.git
 	branch = master
-	platforms = linux,android,darwin
+	platforms = linux,android,darwin,python
 [submodule "third_party/asr/components"]
 	path = third_party/asr/components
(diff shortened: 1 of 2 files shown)


bzbarsky-apple (approved):

andy31415 (approved):

andy31415:
@cecille is the python wheel not compiling C++ code, so it would need either darwin or linux dependencies (e.g. our wheel accesses blutooth and dnssd so it may need those underlying libraries)

andy31415 (changes requested):
Marking request changes for now: I am unclear how we compile python without native bits. This needs to be explained.

Until now we only ever compiled python whl with native bits and then we depend on linux or darwin submodules.

cecille:
This is submodule specific, trying not to pull in the world. I'd rather not shave this particular yak today, so closing.

PR #33247, opened 2024-04-30
[Metrics] Metrics in core commissioning phase

anush-apple:
- Added metrics tracing in core parts of the commissionning flow:
  - PASE
  - OperationalSessionSetup
  - Different stages of commissioning

### DIFF
--- a/src/app/OperationalSessionSetup.cpp
+++ b/src/app/OperationalSessionSetup.cpp
@@ -38,11 +38,13 @@
 #include <lib/support/logging/CHIPLogging.h>
 #include <system/SystemClock.h>
 #include <system/SystemLayer.h>
+#include <tracing/metric_event.h>
 
 using namespace chip::Callback;
 using chip::AddressResolve::NodeLookupRequest;
 using chip::AddressResolve::Resolver;
 using chip::AddressResolve::ResolveResult;
+using namespace chip::Tracing;
 
 namespace chip {
 
@@ -246,6 +248,8 @@ void OperationalSessionSetup::UpdateDeviceData(const Transport::PeerAddress & ad
         return;
     }
 
+    MATTER_LOG_METRIC_END(kMetricDeviceOperationalDiscovery, CHIP_NO_ERROR);
+
     CHIP_ERROR err = EstablishConnection(config);
     LogErrorOnFailure(err);
     if (err == CHIP_NO_ERROR)
@@ -305,9 +309,11 @@ CHIP_ERROR OperationalSessionSetup::EstablishConnection(const ReliableMessagePro
     mCASEClient = mClientPool->Allocate();
     ReturnErrorCodeIf(mCASEClient == nullptr, CHIP_ERROR_NO_MEMORY);
 
+    MATTER_LOG_METRIC_BEGIN(kMetricDeviceCASESession);
     CHIP_ERROR err = mCASEClient->EstablishSession(mInitParams, mPeerId, mDeviceAddress, config, this);
     if (err != CHIP_NO_ERROR)
     {
+        MATTER_LOG_METRIC_END(kMetricDeviceCASESession, err);
         CleanupCASEClient();
         return err;
     }
@@ -503,6 +509,7 @@ void OperationalSessionSetup::OnSessionEstablishmentError(CHIP_ERROR error, Sess
 #endif // CHIP_DEVICE_CONFIG_ENABLE_AUTOMATIC_CASE_RETRIES
     }
 
+    MATTER_LOG_METRIC_END(kMetricDeviceCASESession, error);
     DequeueConnectionCallbacks(error, stage);
     // Do not touch `this` instance anymore; it has been destroyed in DequeueConnectionCallbacks.
 }
@@ -520,6 +527,8 @@ void OperationalSessionSetup::OnSessionEstablished(const SessionHandle & session
 {
     VerifyOrReturn(mState == State::Connecting,
                    ChipLogError(Discovery, "OnSessionEstablished was called while we were not connecting"));
+
+    MATTER_LOG_METRIC_END(kMetricDeviceCASESession, CHIP_NO_ERROR);
 
     if (!mSecureSession.Grab(session))
     {
@@ -591,6 +600,7 @@ CHIP_ERROR OperationalSessionSetup::LookupPeerAddress()
     {
         --mResolveAttemptsAllowed;
     }
+    MATTER_LOG_METRIC(kMetricDeviceOperationalDiscoveryAttemptCount, mAttemptsDone);
 #endif // CHIP_DEVICE_CONFIG_ENABLE_AUTOMATIC_CASE_RETRIES
 
     // NOTE: This is public API that can be used to update our stored peer
@@ -604,6 +614,8 @@ CHIP_ERROR OperationalSessionSetup::LookupPeerAddress()
                         mPeerId.GetFabricIndex(), ChipLogValueX64(mPeerId.GetNodeId()));
         return CHIP_NO_ERROR;
     }
+
+    MATTER_LOG_METRIC_BEGIN(kMetricDeviceOperationalDiscovery);
 
     auto const * fabricInfo = mInitParams.fabricTable->FindFabricWithIndex(mPeerId.GetFabricIndex());
     VerifyOrReturnError(fabricInfo != nullptr, CHIP_ERROR_INVALID_FABRIC_INDEX);
@@ -676,6 +688,8 @@ void OperationalSessionSetup::OnNodeAddressResolutionFailed(const PeerId & peerI
             --mAttemptsDone;
         }
 
+        MATTER_LOG_METRIC(kMetricDeviceOperationalDiscoveryAttemptCount, mAttemptsDone);
+
         CHIP_ERROR err = LookupPeerAddress();
         if (err == CHIP_NO_ERROR)
         {
@@ -689,6 +703,8 @@ void OperationalSessionSetup::OnNodeAddressResolutionFailed(const PeerId & peerI
         }
     }
 #endif
+
+    MATTER_LOG_METRIC_END(kMetricDeviceOperationalDiscovery, reason);
 
     // No need to modify any variables in `this` since call below releases `this`.
     DequeueConnectionCallbacks(reason);
--- a/src/app/OperationalSessionSetup.cpp
+++ b/src/app/OperationalSessionSetup.cpp
@@ -307,9 +307,6 @@ CHIP_ERROR OperationalSessionSetup::EstablishConnection(const ReliableMessagePro
     mCASEClient = mClientPool->Allocate();
     ReturnErrorCodeIf(mCASEClient == nullptr, CHIP_ERROR_NO_MEMORY);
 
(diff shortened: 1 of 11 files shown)


src/controller/CommissioningDelegate.cpp
```diff
@@ -138,5 +138,124 @@ const char * StageToString(CommissioningStage stage)
...
+        return "core_cm_stg_secure_pairing";
+
+    case kReadCommissioningInfo:
+        return "core_cm_stg_rd_comm_info";
+
+    case kReadCommissioningInfo2:
+        return "core_cm_stg_rd_comm_info2";
+
+    case kArmFailsafe:
+        return "core_cm_stg_afs";
```
tcarmelveilleux, line 159: Why these cryptic strings? Is there a limit to length?

src/app/OperationalSessionSetup.cpp
```diff
@@ -242,6 +244,8 @@ void OperationalSessionSetup::UpdateDeviceData(const Transport::PeerAddress & ad
         return;
     }
 
+    MATTER_LOG_METRIC_END(kMetricDeviceOperationalDiscovery, CHIP_NO_ERROR);
```
bzbarsky-apple, line 247: So how will this work in practice when we get multiple UpdateDeviceData for a single discovery start (e.g. if we found multiple IPs and the first one we tried did not work)?

src/app/OperationalSessionSetup.cpp
```diff
@@ -591,6 +601,8 @@ CHIP_ERROR OperationalSessionSetup::LookupPeerAddress()
         return CHIP_NO_ERROR;
     }
 
+    MATTER_LOG_METRIC_BEGIN(kMetricDeviceOperationalDiscovery);
```
bzbarsky-apple, line 604: Similar question here: it looks like we can reach this point multiple times before ever hitting a `MATTER_LOG_METRIC_END(kMetricDeviceOperationalDiscovery)`, no?

src/controller/AutoCommissioner.cpp
```diff
@@ -24,12 +24,14 @@
 #include <controller/CHIPDeviceController.h>
 #include <credentials/CHIPCert.h>
 #include <lib/support/SafeInt.h>
+#include <tracing/metric_event.h>
```
bzbarsky-apple, line 27: Why the header and namespace change in this file?  I don't see any metrics being recorded in this file....

src/controller/CHIPDeviceController.cpp
```diff
@@ -736,6 +744,7 @@ CHIP_ERROR DeviceCommissioner::EstablishPASEConnection(NodeId remoteDeviceId, Re
                     // We already have an open secure session to this device, call the callback immediately and early return.
                     mPairingDelegate->OnPairingComplete(CHIP_NO_ERROR);
                 }
+                MATTER_LOG_METRIC_END(kMetricDeviceCommissionerPASESession, CHIP_NO_ERROR);
                 return CHIP_NO_ERROR;
             }
             if (current->IsSessionSetupInProgress())
```
bzbarsky-apple, line 750: Do we need an end metric inside this block too?

Or perhaps: should we only have the "start" metric when we get around to actually doing the `Pair()` call?  I don't quite understand what the `kMetricDeviceCommissionerPASESession` represents given the code here right now.

src/controller/CHIPDeviceController.cpp
```diff
@@ -908,6 +920,7 @@ CHIP_ERROR DeviceCommissioner::Commission(NodeId remoteDeviceId)
     mDefaultCommissioner->SetOperationalCredentialsDelegate(mOperationalCredentialsDelegate);
     if (device->IsSecureConnected())
     {
+        MATTER_LOG_METRIC_BEGIN(kMetricDeviceCommissionerCommission);
```
bzbarsky-apple, line 923: So is kMetricDeviceCommissionerCommission a metric that gets begin/end?  How do the `ReturnErrorOnFailureWithMetric`/`VerifyOrDoWithMetric` bits interact with begin/end?

src/controller/CHIPDeviceController.cpp
```diff
@@ -1960,6 +1982,7 @@ void DeviceCommissioner::OnDeviceConnectedFn(void * context, Messaging::Exchange
 void DeviceCommissioner::OnDeviceConnectionFailureFn(void * context, const ScopedNodeId & peerId, CHIP_ERROR error)
 {
     // CASE session establishment failed.
+    MATTER_LOG_METRIC_END(kMetricDeviceOperationalSetup, error);
```
bzbarsky-apple, line 1985: This seems like a very generic metric name, for a place which is not involved in most "operational setup". This bit is specific to operational session setup at the end of commissioning.

src/controller/SetUpCodePairer.cpp
```diff
@@ -89,10 +93,16 @@ CHIP_ERROR SetUpCodePairer::PairDevice(NodeId remoteId, const char * setUpCode,
...
 
-    ReturnErrorOnFailure(Connect(payload));
-    return mSystemLayer->StartTimer(System::Clock::Milliseconds32(kDeviceDiscoveredTimeout), OnDeviceDiscoveredTimeoutCallback,
-                                    this);
+    ReturnErrorOnFailureWithMetric(kMetricSetupCodePairerPairDevice, Connect(payload));
+    auto errorCode =
+        mSystemLayer->StartTimer(System::Clock::Milliseconds32(kDeviceDiscoveredTimeout), OnDeviceDiscoveredTimeoutCallback, this);
+    if (CHIP_NO_ERROR == errorCode)
+    {
+        MATTER_LOG_METRIC_BEGIN(kMetricSetupCodePairerPairDevice);
```
bzbarsky-apple, line 101: Again, I don't quite understand how MATTER_LOG_METRIC_BEGIN and ReturnErrorOnFailureWithMetric work when using the same metric name....

src/protocols/secure_channel/CASESession.cpp
```diff
@@ -47,6 +47,7 @@
 #include <system/SystemClock.h>
 #include <system/TLVPacketBufferBackingStore.h>
 #include <tracing/macros.h>
+#include <tracing/metric_event.h>
```
bzbarsky-apple, line 50: Where is this used in here?

src/controller/CHIPDeviceController.cpp
anush-apple: Actually I don't think we want to end here. As the comment says, the delegate will be called and that is where the end of the PASE session will be marked

src/controller/CHIPDeviceController.cpp
anush-apple: The ReturnErrorOnFailureWithMetric/VerifyOrDoWithMetric are used to handle any failure that happen synchronously before the log async operation to start commissioning.

src/controller/SetUpCodePairer.cpp
anush-apple: The begin and end are used to make longer sync stages. The ReturnErrorOnFailureWithMetric is used for any errors that happen before the begin is marked. If there are errors, they are just marked with the same even name but not begin/end modes

src/app/OperationalSessionSetup.cpp
anush-apple: Yes, that is fine. Backend needs to handle this

src/protocols/secure_channel/CASESession.cpp
woody-apple: Resolved by my eye

src/controller/SetUpCodePairer.cpp
woody-apple: Yup

src/controller/CHIPDeviceController.cpp
woody-apple: Yup

src/controller/CHIPDeviceController.cpp
woody-apple: Agree

src/app/OperationalSessionSetup.cpp
woody-apple: Yup

src/app/OperationalSessionSetup.cpp
```diff
@@ -305,9 +307,14 @@ CHIP_ERROR OperationalSessionSetup::EstablishConnection(const ReliableMessagePro
     mCASEClient = mClientPool->Allocate();
     ReturnErrorCodeIf(mCASEClient == nullptr, CHIP_ERROR_NO_MEMORY);
 
+    // Session has been established. This is when discovery is also stopped
```
bzbarsky-apple, line 310: I don't get this comment.  The session has not been established at this point....

And as we discussed discovery is sill ongoing, in parallel with the CASE bits.  So I'm a little confused by this part.

src/app/OperationalSessionSetup.cpp
```diff
@@ -605,6 +616,10 @@ CHIP_ERROR OperationalSessionSetup::LookupPeerAddress()
         return CHIP_NO_ERROR;
     }
 
+    // This call be called multiple times for each discovered address.The metric backend can handle this 
+    // and always pick the earliest occurance as the start of the event.
```
bzbarsky-apple, line 620:
```suggestion
    // This code can be reached multiple times, if we discover multiple addresses or do retries.
    // The metric backend can handle this and always picks the earliest occurrence as the start of the event.
```

bzbarsky-apple (approved):

src/controller/CHIPDeviceController.cpp
```diff
@@ -666,16 +668,20 @@ CHIP_ERROR DeviceCommissioner::PairDevice(NodeId remoteDeviceId, const char * se
...
 
 CHIP_ERROR DeviceCommissioner::PairDevice(NodeId remoteDeviceId, RendezvousParameters & rendezvousParams,
                                           CommissioningParameters & commissioningParams)
 {
     MATTER_TRACE_SCOPE("PairDevice", "DeviceCommissioner");
-    ReturnErrorOnFailure(EstablishPASEConnection(remoteDeviceId, rendezvousParams));
-    return Commission(remoteDeviceId, commissioningParams);
+    ReturnErrorOnFailureWithMetric(kMetricDeviceCommissionerCommission, EstablishPASEConnection(remoteDeviceId, rendezvousParams));
+    auto errorCode = Commission(remoteDeviceId, commissioningParams);
+    VerifyOrDoWithMetric(kMetricDeviceCommissionerCommission, CHIP_NO_ERROR == errorCode, errorCode);
```
tcarmelveilleux, line 683: Sounds like you need a `VerifySuccessOrDoWithMetric` for the common case of checking errorCode == CHIP_NO_ERROR.

tcarmelveilleux (approved):

src/controller/CHIPDeviceController.cpp
anush-apple: Thanks for the suggestion. I can add that next

anush-apple: Filed https://github.com/project-chip/connectedhomeip/issues/33508

PR #33249, opened 2024-05-01
Updated unit tests in src/transport/tests/ to use PW instead of NL.

feasel0:
These changes are replacing NL unit tests with pigweed unit tests. I updated the tests in src/transport/tests/

This PR replaces a previous PR [#33155](https://github.com/project-chip/connectedhomeip/pull/33155) which is now closed.  The previous PR had some files accidentally included in it which were not supposed to be part of it.
All suggestions from that PR have been addressed.

Note that in `TestSecureSessionTable.cpp`, the test function `ValidateSessionSorting` has to call another function `TestSecureSessionTable::ValidateSessionSorting` because the tests make reference to `SecureSession::State` which is private.  Putting the tests inside the scope `TestSecureSessionTable` allows `State` to be accessed since `SecureSession` explicitly names `TestSecureSessionTable` as a friend class.


andy31415:
@feasel0 this seems to need merge conflict fixing and restyle

andy31415 (approved):

PR #33250, opened 2024-05-01
Fix handling of short discriminator in QRCodeSetupPayloadGenerator

ksperling-apple:
If a SetupPayload contains a short discriminator then
- isValidQRCodePayload() should return false
- trying to generate a QR Code should return INVALID_ARGUMENT
- generating with SetAllowInvalidPayload(true) should work (not die)

Also make `IsCommonTag`, `IsVendorTag`, and `getOptionalVendorData(tag, &info)` public.

The first two are just encoding spec rules that are useful for clients, and the
latter allows clients to read vendor data by tag instead of having to read the
whole list.

Also use default values instead of explicit constructors for OptionalQRCodeInfo
and fix up some doc comments to correctly reference the vendor tag range.

### DIFF
--- a/src/setup_payload/QRCodeSetupPayloadGenerator.cpp
+++ b/src/setup_payload/QRCodeSetupPayloadGenerator.cpp
@@ -174,10 +174,12 @@ static CHIP_ERROR generateBitSet(PayloadContents & payload, MutableByteSpan & bi
     ReturnErrorOnFailure(populateBits(bits.data(), offset, payload.rendezvousInformation.Value().Raw(),
                                       kRendezvousInfoFieldLengthInBits, kTotalPayloadDataSizeInBits));
 
-    // If the payload is valid then discriminator will always be long, but we shouldn't die if we're
-    // encoding a short one due to mAllowInvalidPayload being set. Just use the short value in that case.
+    // isValidQRCodePayload() ensures that we have a long discriminator, but if AllowInvalidPayload is set
+    // we might have only a short one, and calling GetLongValue() would die. So check for this case here
+    // and use a placeholder value of 0. Encoding an invalid (or partially valid) payload is useful for
+    // clients that need to be able to serialize and deserialize partially populated or invalid payloads.
     auto const & pd        = payload.discriminator;
-    uint16_t discriminator = (pd.IsShortDiscriminator()) ? pd.GetShortValue() : pd.GetLongValue();
+    uint16_t discriminator = !pd.IsShortDiscriminator() ? pd.GetLongValue() : 0;
     ReturnErrorOnFailure(
         populateBits(bits.data(), offset, discriminator, kPayloadDiscriminatorFieldLengthInBits, kTotalPayloadDataSizeInBits));
     ReturnErrorOnFailure(
--- a/src/setup_payload/QRCodeSetupPayloadGenerator.cpp
+++ b/src/setup_payload/QRCodeSetupPayloadGenerator.cpp
@@ -173,8 +173,13 @@ static CHIP_ERROR generateBitSet(PayloadContents & payload, MutableByteSpan & bi
     VerifyOrReturnError(payload.rendezvousInformation.HasValue(), CHIP_ERROR_INVALID_ARGUMENT);
     ReturnErrorOnFailure(populateBits(bits.data(), offset, payload.rendezvousInformation.Value().Raw(),
                                       kRendezvousInfoFieldLengthInBits, kTotalPayloadDataSizeInBits));
-    ReturnErrorOnFailure(populateBits(bits.data(), offset, payload.discriminator.GetLongValue(),
-                                      kPayloadDiscriminatorFieldLengthInBits, kTotalPayloadDataSizeInBits));
+
+    // If the payload is valid then discriminator will always be long, but we shouldn't die if we're
+    // encoding a short one due to mAllowInvalidPayload being set. Just use the short value in that case.
+    auto const & pd        = payload.discriminator;
+    uint16_t discriminator = (pd.IsShortDiscriminator()) ? pd.GetShortValue() : pd.GetLongValue();
+    ReturnErrorOnFailure(
+        populateBits(bits.data(), offset, discriminator, kPayloadDiscriminatorFieldLengthInBits, kTotalPayloadDataSizeInBits));
     ReturnErrorOnFailure(
         populateBits(bits.data(), offset, payload.setUpPINCode, kSetupPINCodeFieldLengthInBits, kTotalPayloadDataSizeInBits));
     ReturnErrorOnFailure(populateBits(bits.data(), offset, 0, kPaddingFieldLengthInBits, kTotalPayloadDataSizeInBits));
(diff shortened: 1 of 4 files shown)


src/setup_payload/QRCodeSetupPayloadGenerator.cpp
```diff
@@ -173,8 +173,13 @@ static CHIP_ERROR generateBitSet(PayloadContents & payload, MutableByteSpan & bi
     VerifyOrReturnError(payload.rendezvousInformation.HasValue(), CHIP_ERROR_INVALID_ARGUMENT);
     ReturnErrorOnFailure(populateBits(bits.data(), offset, payload.rendezvousInformation.Value().Raw(),
                                       kRendezvousInfoFieldLengthInBits, kTotalPayloadDataSizeInBits));
-    ReturnErrorOnFailure(populateBits(bits.data(), offset, payload.discriminator.GetLongValue(),
-                                      kPayloadDiscriminatorFieldLengthInBits, kTotalPayloadDataSizeInBits));
+
+    // If the payload is valid then discriminator will always be long, but we shouldn't die if we're
+    // encoding a short one due to mAllowInvalidPayload being set. Just use the short value in that case.
+    auto const & pd        = payload.discriminator;
+    uint16_t discriminator = (pd.IsShortDiscriminator()) ? pd.GetShortValue() : pd.GetLongValue();
```
andy31415, line 180:
```suggestion
    uint16_t discriminator = pd.IsShortDiscriminator() ? pd.GetShortValue() : pd.GetLongValue();
```

src/setup_payload/QRCodeSetupPayloadGenerator.cpp
```diff
@@ -173,8 +173,13 @@ static CHIP_ERROR generateBitSet(PayloadContents & payload, MutableByteSpan & bi
     VerifyOrReturnError(payload.rendezvousInformation.HasValue(), CHIP_ERROR_INVALID_ARGUMENT);
     ReturnErrorOnFailure(populateBits(bits.data(), offset, payload.rendezvousInformation.Value().Raw(),
                                       kRendezvousInfoFieldLengthInBits, kTotalPayloadDataSizeInBits));
-    ReturnErrorOnFailure(populateBits(bits.data(), offset, payload.discriminator.GetLongValue(),
-                                      kPayloadDiscriminatorFieldLengthInBits, kTotalPayloadDataSizeInBits));
+
+    // If the payload is valid then discriminator will always be long, but we shouldn't die if we're
+    // encoding a short one due to mAllowInvalidPayload being set. Just use the short value in that case.
```
andy31415, line 178: should we in this case validate that mAllowInvalidPayload is set before trying to use the short value?

Please add a description of the use case of when we would like to encode invalid/incomplete data to the comment here. This just says "we shoudn't die" (replace this "shouldn't return error" probably) but it does not explain why. It should explain why.

src/setup_payload/SetupPayload.h
```diff
@@ -153,29 +153,19 @@ enum optionalQRCodeInfoType
...
-        int32  = 0;
-        int64  = 0;
-        uint32 = 0;
-        uint64 = 0;
-    }
-
-    int64_t int64;
-    uint64_t uint32;
-    uint64_t uint64;
+    int64_t int64   = 0; /**< the integer value if type is optionalQRCodeInfoTypeInt64, otherwise should not be set */
```
andy31415, line 166: not for your PR, however I think we should start using variants here or something like that. I imagine std::string could not be placed in a union with int32 above, so now we have i64/u32/u64 added on top of that again without a union even though only a subset of them can be valid. This feels very brittle.

src/setup_payload/QRCodeSetupPayloadGenerator.cpp
bzbarsky-apple: "shouldn't die" means "shouldn't die": `GetLongValue()` involves a VerifyOrDie that the value is long.

src/setup_payload/QRCodeSetupPayloadGenerator.cpp
bzbarsky-apple, line 180: So here's the thing: this payload is not just "invalid", it's literally nonsensical.  Does anything (including tests, etc, given that needs explicit opt-in) actually rely on encoding such payloads with short discriminators as QR codes?  If not, I don't think we should support it and should just error out.  The resulting QR code is not "invalid" in any sense, so it can't be used to test handling of invalid QR codes by other entities; it's just not a valid QR code that matches any reality in any way.

src/setup_payload/QRCodeSetupPayloadGenerator.cpp
ksperling-apple: > should we in this case validate that mAllowInvalidPayload is set before trying to use the short value

If `mAllowInvalidPayload` is not set, then we won't get here with a short discriminator because this is now checked by `isValidQRCodePayload`.

My use case here is to allow maybe-invalid payloads to be serialized, for the purpose of deserializing them again. "AllowInvalidPayload" already describes this pretty well?

src/setup_payload/QRCodeSetupPayloadGenerator.cpp
ksperling-apple: My specific use case is to use the QR code representation to serialize MTRSetupPayload objects that may not be valid, and be able to deserialize them again. Using the QR code TLV representation seems like a much more robust way to handle the extension data than coming up with some other encoding for it.

ksperling-apple: As discussed in Slack, I'm going to just encode the long discriminator as 0, to avoid a client making any undue assumptions about being able to roundtrip a short discriminator here.

bzbarsky-apple (approved):

src/setup_payload/QRCodeSetupPayloadGenerator.cpp
```diff
@@ -170,10 +175,11 @@ static CHIP_ERROR generateBitSet(PayloadContents & payload, MutableByteSpan & bi
         populateBits(bits.data(), offset, payload.productID, kProductIDFieldLengthInBits, kTotalPayloadDataSizeInBits));
     ReturnErrorOnFailure(populateBits(bits.data(), offset, static_cast<uint64_t>(payload.commissioningFlow),
                                       kCommissioningFlowFieldLengthInBits, kTotalPayloadDataSizeInBits));
-    VerifyOrReturnError(payload.rendezvousInformation.HasValue(), CHIP_ERROR_INVALID_ARGUMENT);
-    ReturnErrorOnFailure(populateBits(bits.data(), offset, payload.rendezvousInformation.Value().Raw(),
-                                      kRendezvousInfoFieldLengthInBits, kTotalPayloadDataSizeInBits));
-    ReturnErrorOnFailure(populateBits(bits.data(), offset, payload.discriminator.GetLongValue(),
+    auto const & ri = payload.rendezvousInformation;
+    ReturnErrorOnFailure(populateBits(bits.data(), offset, (ri.HasValue() ? ri.Value().Raw() : 0), kRendezvousInfoFieldLengthInBits,
```
andy31415, line 179: I wonder if this is more readable:

```cpp
    ReturnErrorOnFailure(populateBits(bits.data(), offset, ri.ValueOr(RendezvousInformationFlag::kNone).Raw(),
                                      kRendezvousInfoFieldLengthInBits, kTotalPayloadDataSizeInBits));
```

or 

```cpp
    ReturnErrorOnFailure(populateBits(bits.data(), offset,
                                      payload.rendezvousInformation.ValueOr(RendezvousInformationFlag::kNone).Raw(),
                                      kRendezvousInfoFieldLengthInBits, kTotalPayloadDataSizeInBits));
```

andy31415 (approved):

src/setup_payload/QRCodeSetupPayloadGenerator.cpp
ksperling-apple: Done

PR #33256, opened 2024-05-01
Post certification checks script

cecille:
This test is used to evaluate that all the proper post-certification work has been done to make a Matter device production ready. This test ensure that:
- DAC chain is valid and spec compliant, and chains up to a PAA that is registered in the main net DCL
- CD is valid and, signed by one of the known CSA signing certs and is marked as a production CD
- DCL entries for this device and vendor have all been registered
- TestEventTriggers have been turned off

This test is performed over PASE on a factory reset device.

To run this test, first build and install the python chip wheel files, then add the extra dependencies. From the root:

./scripts/build_python.sh -i py
source py/bin/activate
pip install opencv-python requests click_option_group

Note that right now I'm pulling the PAA and CD signing certs on every run. This is a bit aggressive, but I want to ensure we're not running on stale content. Pulling daily might be a later optimization.


src/python_testing/post_certification_tests/post-cert-checks.py
```diff
@@ -0,0 +1,348 @@
...
+        asserts.assert_true(key in entry.keys(),
+                            f"Unable to find compliance entry for {self.vid:04x} {self.pid:04x} {self.software_version}")
+        logging.info(
+            f'Found compliance info for vid=0x{self.vid:04X} pid=0x{self.pid:04X} software version={self.software_version} in the DCL:')
+        logging.info(f'{entry[key]}')
+
+    def steps_CertifiedModel(self):
+        return [TestStep(1, "Check if device VID/PID/SoftwareVersion are listed in the DCL certified model schema", "Listing found")]
+
+    def test_CertifiedModel(self):
```
tpmanley, line 192: We've seen a few cases where a software update is posted in the DCL and the `OtaChecksum` is not base64 encoded. What do you think about adding a test that verifies at least that all the software versions for this device model have a correctly formatted checksum? Could potentially also verify other fields as well but this one in particular feels especially important.

credentials/fetch_paa_certs_from_dcl.py
```diff
@@ -133,7 +139,38 @@ def use_dcld(dcld, production, cmdlist):
...
+        subject_key_id = signer['subjectKeyId']
+        certificate, subject = get_cert_from_rest(rest_node_url, subject, subject_key_id)
+
+        print(f"Downloaded CD signing cert with subject: {subject}")
+        write_cert(certificate, subject)
+
+    os.chdir(original_dir)
+
+
+def fetch_paa_certs(use_main_net_dcld, use_test_net_dcld, use_main_net_http, use_test_net_http, paa_trust_store_path):
```
jcps07, line 173: This will include CD Signing CAs in the PAA trust store, but hose are not used as PAA resource in the DCL. The DCL stores Root CAs in the PKI schema, not PAA (only the CD Signing CAs are non-PAA though).

I'd recommend hardcoding the CD Signing Cert in the script as those are not intended to change. There is no need to query them every time. And remove them from the PAA trust store.

src/python_testing/post_certification_tests/post-cert-checks.py
```diff
@@ -0,0 +1,348 @@
...
+        self.connect_over_pase(self.default_controller)
+        bi = Clusters.BasicInformation
+        self.vid = await self.read_single_attribute_check_success(cluster=bi, attribute=bi.Attributes.VendorID)
+        self.pid = await self.read_single_attribute_check_success(cluster=bi, attribute=bi.Attributes.ProductID)
+        self.software_version = await self.read_single_attribute_check_success(cluster=bi, attribute=bi.Attributes.SoftwareVersion)
+
+    def steps_Vendor(self):
+        return [TestStep(1, "Check if device VID is listed in the DCL vendor schema", "Listing found")]
+
+    def test_Vendor(self):
```
jcps07, line 157: The following should be optional or good to show as informative to be missing since they are not mandatory to be public/available at Certification Testing stage. (specially the Compliance-related ones, as those require the cert tests to be completed in order to get).
- test_Vendor
- test_Model  <- Missing "test_ModelVersion"
- test_Compliance
- test_CertifiedModel

The only one that could make sense to confirm as mandatory is the existence of the Root CA in the DCL (to define if Main-Net or Test-Net).

src/python_testing/post_certification_tests/post-cert-checks.py
cecille: Yeah, that seems like a good addition. Let me see how to go about doing that - right now I'm just pulling the entry for this specific device version, and I'm not totally sure how to get the index. Give me like a day or two.

credentials/fetch_paa_certs_from_dcl.py
```diff
@@ -37,6 +37,10 @@
 PRODUCTION_NODE_URL_REST = "https://on.dcl.csa-iot.org"
 TEST_NODE_URL_REST = "https://on.test-net.dcl.csa-iot.org"
 
+# TODO: really? We can't just get this by name from the DCL?
```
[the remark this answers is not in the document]
cecille: will the current subject/key style be deprecated? or is this a pure addition?

credentials/fetch_paa_certs_from_dcl.py
cecille: PAA query shouldn't include the CD certs, as far as I can tell. This is index'd off the pki/root-certificates approvedRootCertificates list. So that should be strictly PAAs, right?

For CDs...I was torn on that. The seem like a static set, but can the CSA not mint new intermediates? ie, the root should never change, do we know the set of intermediates will not?

src/python_testing/post_certification_tests/post-cert-checks.py
cecille: So this test is going to be run after cert on production devices at the interop lab. For production devices, these are all mandatory, right?

what's test_Model/Version?

src/python_testing/post_certification_tests/post-cert-checks.py
cecille: I also wonder if this is something that the DCL should be checking before accepting a new entry....

@jcps07 - do you know if it would be possible to pre-check entries for this as they're uploaded?

jcps07: I'll bring it up to the DCL-TT and create an issue.

credentials/fetch_paa_certs_from_dcl.py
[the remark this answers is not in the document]
jcps07: I believe it's addition, no deprecation.

src/python_testing/post_certification_tests/post-cert-checks.py
cecille: awesome - thanks

src/python_testing/post_certification_tests/post-cert-checks.py
jcps07: Still not mandatory, a company may days or even months to decide going public in the DCL after they get their certification approved... But I do believe it's good to have for members to be aware of what they are missing.

In DCL, there is Model and Model-Version (that contains the Software Version) under the Model. 
I believe the test is only checking for the Model, not the Model-Version... which is the entry that actually gets the Compliance entry associated with it.

When a Product gets a certification, it is granted to an specific software version (VID/PID/SoftwareVersion/SoftwareVersionString) that translates to a Model-Version in the DCL. Sometimes members just enter the Model to the DCL, and they miss to enter the Model-Version, so they won't get their Compliance entry in the DCL until they add it. Or sometimes the information they add in the Model-Version mismatches with the information in the Certification Tool, for example:
Software Version String in Cert Tool: "v1.0.0"
Software Version String in DCL: "1.0" or "1.0.0"

That kind of discrepancy triggers troubleshooting with the Certification team to make sure all data is aligned. This could be caught at early stage.

Some info: [HowTo - Writing to the DCL.pdf](https://groups.csa-iot.org/wg/members-all/document/27881)

credentials/fetch_paa_certs_from_dcl.py
jcps07: From what I saw, `approvedRootCertificates` gets the data from `{rest_node_url}/dcl/pki/root-certificates`

https://on.dcl.csa-iot.org/dcl/pki/root-certificates

That one includes Root CAs, not filtered by PAA neccessarily.... so yes, it includes the CD Signing Cert (see first item in response):
> {
>   "approvedRootCertificates": {
>     "certs": [
>       {
>         "subject": "MFIxDDAKBgNVBAoMA0NTQTEsMCoGA1UEAwwjTWF0dGVyIENlcnRpZmljYXRpb24gYW5kIFRlc3RpbmcgQ0ExFDASBgorBgEEAYKifAIBDARDNUEw",
>         "subjectKeyId": "97:E4:69:D0:C5:04:14:C2:6F:C7:01:F7:7E:94:77:39:09:8D:F6:A5"
>       },
>       {
>         "subject": "MFMxCzAJBgNVBAYTAlVTMRcwFQYDVQQKEw5EaWdpQ2VydCwgSW5jLjErMCkGA1UEAxMiRGlnaUNlcnQgUm9vdCBDQSBmb3IgTUFUVEVSIFBLSSBHMQ==",
>         "subjectKeyId": "32:50:45:19:33:44:59:9B:46:65:D4:59:FD:3A:15:F1:C1:16:CC:BD"
>       },
>       {
>         "subject": "MDYxIDAeBgNVBAMMF1N0cm9uZ0tleSBNYXR0ZXIgRzEgUEFBMRIwEAYDVQQKDAlTdHJvbmdLZXk=",
>         "subjectKeyId": "37:4E:29:5A:C5:51:73:0B:AF:65:D5:29:83:9B:4E:E0:9D:39:58:8A"
>       },
>       {
>         "subject": "MDMxGzAZBgNVBAMMElRQLUxpbmsgTWF0dGVyIFBBQTEUMBIGCisGAQQBgqJ8AgEMBDExODg=",
>         "subjectKeyId": "C5:85:31:41:9E:E0:3E:E4:A1:DC:F5:51:E0:34:40:DC:2B:BE:C2:0A"
>       },
>       {

src/python_testing/post_certification_tests/post-cert-checks.py
jcps07: [Issue #570](https://github.com/zigbee-alliance/distributed-compliance-ledger/issues/570) created

credentials/fetch_paa_certs_from_dcl.py
cecille: actually, I'll just test it myself and see what's up.

src/python_testing/post_certification_tests/post-cert-checks.py
cecille: OK added a more complete check where it verifies the OTA image fully. PTAL if you have a sec.

credentials/fetch_paa_certs_from_dcl.py
cecille, line 40:
```suggestion
```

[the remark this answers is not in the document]
cecille: OK, for now, going to leave this - as long as this method isn't being deprecated, this is fine. Removed the TODO.

credentials/fetch_paa_certs_from_dcl.py
cecille: Yup - CD root was in there. Removed it now. Thanks, JC.

jcps07: I had the understanding that the SDK used the CD Signing Keys hardcoded within the SDK repo: 

- https://github.com/project-chip/connectedhomeip/tree/master/credentials/development/cd-certs
- https://github.com/project-chip/connectedhomeip/blob/master/src/credentials/attestation_verifier/DefaultDeviceAttestationVerifier.cpp

src/python_testing/post_certification_tests/post-cert-checks.py
jcps07: Yes, sorry I'm not being clear. 
Your script is validating that a Model is in the DCL, which is fine. But it still requires checking that the corresponding Model-Version is also in the DCL. That is the actual product in the DCL, not the Model.
So, the script will confirm that the data from the Basic Information Cluster matches the information from the DCL, and that is with the Model (contains the VID, PID and Device Type) and Model-Version (contains the VID, PID, Software Version and Software Version String), not just the Model.

cecille: Is the model version not required to be in the DCL once the product is released? This test is meant to be run on launched products in the interop lab - ie things the lab purchased from a store. At that point, there's no confidentiality problem because it's a commercial product, and the commissioners should be able to get information about the device version from the production DCL. Or is the model version stuff completely optional (ie, ecosystems shouldn't be checking that).

cecille: spoke offline - these checks are correct for devices in production. Renamed script and updated the top of file readme to reflect that.

tpmanley (approved):
My coworker @samadDotDev tested this out with a device with some bad checksums in testnet, and it correctly identified those issues and passed with mainnet.

ReneJosefsen (approved):

credentials/fetch_paa_certs_from_dcl.py
```diff
@@ -133,7 +138,38 @@ def use_dcld(dcld, production, cmdlist):
...
+    subject = response["subjectAsText"]
+    return certificate, subject
+
+
+def fetch_cd_signing_certs(store_path):
+    ''' Only supports using main net http currently.'''
+    rest_node_url = PRODUCTION_NODE_URL_REST
+    os.makedirs(store_path, exist_ok=True)
+    original_dir = os.getcwd()
+    os.chdir(store_path)
```
andy31415, line 157: why do you need to change the directory? Can you pass in paths with the store_path into write_cert instead?

andy31415: Functions assuming local directory feel brittle. I believe we should pass in paths and not rely on local dir .

credentials/fetch_paa_certs_from_dcl.py
```diff
@@ -148,36 +184,43 @@ def main(use_main_net_dcld, use_test_net_dcld, use_main_net_http, use_test_net_h
     rest_node_url = PRODUCTION_NODE_URL_REST if production else TEST_NODE_URL_REST
 
     os.makedirs(paa_trust_store_path, exist_ok=True)
+    original_dir = os.getcwd()
```
andy31415, line 187: let's update code to not rely on cwd or chdir.

andy31415 (approved):

credentials/fetch_paa_certs_from_dcl.py
andy31415: Leaving for future PR, so that we do not mix new functionality with "make code not rely on cwd"

credentials/fetch_paa_certs_from_dcl.py
andy31415: Leaving for future PR, so that we do not mix new functionality with "make code not rely on cwd"

credentials/fetch_paa_certs_from_dcl.py
cecille: discussed offline - this is from pre-existing code, and orthogonal to this change, so will address in a followup to clean both this and the other code in this file.

PR #33258, opened 2024-05-01
Enable chef robotic vacuum cleaner device

stingchang:
Add a chef device - Robotic Vacuum Cleaner.
this device supports rvc  run mode cluster, rvc clean mode cluster, rvc operational state cluster, and power supply cluster.


examples/chef/common/chef-rvc-mode-delegate.cpp
```diff
@@ -106,17 +108,61 @@ void RvcRunMode::Shutdown()
     }
 }
 
+chip::Protocols::InteractionModel::Status chefRvcRunModeWriteCallback(chip::EndpointId endpointId, chip::ClusterId clusterId,
+                                                                      const EmberAfAttributeMetadata * attributeMetadata,
+                                                                      uint8_t * buffer)
+{
+    chip::Protocols::InteractionModel::Status ret = Protocols::InteractionModel::Status::Success;
+
+    VerifyOrDie(endpointId == 1); // this cluster is only enabled for endpoint 1.
```
andy31415, line 117: is verifyOrDier correct or should we just return some `Status::Failure` instead?

examples/chef/common/chef-rvc-mode-delegate.cpp
```diff
@@ -106,17 +108,61 @@ void RvcRunMode::Shutdown()
...
+    switch (attributeId)
+    {
+    case chip::app::Clusters::RvcRunMode::Attributes::CurrentMode::Id: {
+        uint8_t m                                        = static_cast<uint8_t>(buffer[0]);
+        chip::Protocols::InteractionModel::Status status = clusterInstance->UpdateCurrentMode(m);
+        if (chip::Protocols::InteractionModel::Status::Success == status)
+        {
+            break;
+        }
+        ret = chip::Protocols::InteractionModel::Status::UnsupportedWrite;
```
andy31415, line 132: what kind of status may UpdateCurrentMode return that we convert always to UnsupportedWrite? Why are we not doing `ret = clusterInstance->UpdateCurrentMode(m)` ? a comment should explain

examples/chef/common/chef-rvc-mode-delegate.cpp
```diff
@@ -106,17 +108,61 @@ void RvcRunMode::Shutdown()
...
+
+    return ret;
+}
+
+chip::Protocols::InteractionModel::Status chefRvcRunModeReadCallback(chip::EndpointId endpointId, chip::ClusterId clusterId,
+                                                                     const EmberAfAttributeMetadata * attributeMetadata,
+                                                                     uint8_t * buffer, uint16_t maxReadLength)
+{
+
+    return chip::Protocols::InteractionModel::Status::Success;
```
andy31415, line 150: is this guaranteed to be always correct? We do not alter `buffer` at all. Should we at least memset to 0? A comment should explain why a read that is a noop is to be considered a success.

examples/chef/common/chef-rvc-operational-state-delegate.cpp
```diff
@@ -17,17 +17,31 @@
...
+#include <platform/CHIPDeviceLayer.h>
 
 using namespace chip;
 using namespace chip::app;
 using namespace chip::app::Clusters;
 using namespace chip::app::Clusters::OperationalState;
 using namespace chip::app::Clusters::RvcOperationalState;
+using chip::Protocols::InteractionModel::Status;
+
+#ifdef MATTER_DM_PLUGIN_RVC_OPERATIONAL_STATE_SERVER
```
andy31415, line 30: why was this moved?

Why do we ifdef the entire file? can we instead use the build system to either include the cpp file or not?

examples/chef/common/chef-rvc-operational-state-delegate.cpp
```diff
@@ -46,100 +60,115 @@ CHIP_ERROR GenericOperationalStateDelegateImpl::GetOperationalPhaseAtIndex(size_
...
+        static_cast<OperationalState::OperationalStateEnum>(instance->GetCurrentOperationalState());
 
-static OperationalState::Instance * gOperationalStateInstance = nullptr;
-static OperationalStateDelegate * gOperationalStateDelegate   = nullptr;
+    auto countdown_time = delegate->GetCountdownTime();
 
-void OperationalState::Shutdown()
-{
-    if (gOperationalStateInstance != nullptr)
+    if (countdown_time.IsNull() || (!countdown_time.IsNull() && countdown_time.Value() > 0))
```
andy31415, line 150:
```suggestion
    if (countdown_time.ValueOr(1) > 0)
```

may be shorter.

examples/chef/common/stubs.cpp
```diff
@@ -56,6 +64,21 @@ Protocols::InteractionModel::Status emberAfExternalAttributeReadCallback(Endpoin
     case chip::app::Clusters::RadonConcentrationMeasurement::Id:
     case chip::app::Clusters::TotalVolatileOrganicCompoundsConcentrationMeasurement::Id:
         return chefConcentrationMeasurementReadCallback(endpoint, clusterId, attributeMetadata, buffer, maxReadLength);
+#endif
+#ifdef MATTER_DM_PLUGIN_RVC_RUN_MODE_SERVER
+    case chip::app::Clusters::RvcRunMode::Id:
+        return chefRvcRunModeReadCallback(endpoint, clusterId, attributeMetadata, buffer, maxReadLength);
+        break;
```
andy31415, line 71: we should remove the break after return

examples/chef/devices/rootnode_roboticvacuumcleaner_1807ff0c49.matter
```diff
@@ -1417,6 +1677,22 @@ endpoint 1 {
     callback attribute clusterRevision;
   }
 
+  server cluster PowerSource {
```
andy31415, line 1680: is power source required for RVC? The spec seems to require run-more and operational state only. Did we make this change on purpose? If so it should be described in the PR summary.

tcarmelveilleux: Having power source is relevant since RVCs are battery operated

examples/chef/common/chef-rvc-operational-state-delegate.cpp
stingchang: moved back

examples/chef/common/chef-rvc-mode-delegate.cpp
stingchang: Right, makes sense.

examples/chef/common/chef-rvc-mode-delegate.cpp
stingchang: I'm not sure whether it's neccessary since it's a call back register which does nothing.

examples/chef/common/chef-rvc-mode-delegate.cpp
stingchang: good point. thanks. Updated

examples/chef/common/chef-rvc-operational-state-delegate.cpp
stingchang: done.

examples/chef/common/stubs.cpp
stingchang: done

examples/chef/common/chef-rvc-operational-state-delegate.cpp
stingchang: right. it's not necessary as includes already specified in stubs.cpp. removed

stingchang: actually it caused dimmablelight build failed so i'm adding it back

examples/chef/common/chef-rvc-mode-delegate.cpp
andy31415: The "does nothing" is the bit that is unclear to me. The way I understand it is that we ask "please fill buffer with the data" and we return success, basically saying "success".

It seems odd that we say success but we leave buffer in an unfilled/undefined state. Would a `memset(buffer, 0, maxReadLength)` be correct? Should we fill buffer with a specific run more instead?

Just leaving it to uninitialized (I imagine callers will just allocate stack space for this) seems wrong and non-deterministic.

examples/chef/common/chef-rvc-mode-delegate.cpp
```diff
@@ -197,14 +246,58 @@ void RvcCleanMode::Shutdown()
...
+    }
+
+    return ret;
+}
+
+chip::Protocols::InteractionModel::Status chefRvcCleanModeReadCallback(chip::EndpointId endpointId, chip::ClusterId clusterId,
+                                                                       const EmberAfAttributeMetadata * attributeMetadata,
+                                                                       uint8_t * buffer, uint16_t maxReadLength)
+{
+    return chip::Protocols::InteractionModel::Status::Success;
```
andy31415, line 291: Here as well: the contract on this function seems to be "fill this buffer with data". The method does not do that yet it returns success.

andy31415 (dismissed):
All read callbacks seem wrong: The requirement from `emberAfExternalAttributeReadCallback` at https://github.com/project-chip/connectedhomeip/blob/master/src/app/util/generic-callbacks.h#L81 says

```
 * If the application was successfully able to
 * read the attribute and write it into the passed buffer, it should return a
 * value of InteractionModel::Status::Success
```

you must fill/read the buffer not just return uninitialized data

examples/chef/common/chef-rvc-mode-delegate.cpp
stingchang: Done

stingchang:
> All read callbacks seem wrong: The requirement from `emberAfExternalAttributeReadCallback` at https://github.com/project-chip/connectedhomeip/blob/master/src/app/util/generic-callbacks.h#L81 says
> 
> ```
>  * If the application was successfully able to
>  * read the attribute and write it into the passed buffer, it should return a
>  * value of InteractionModel::Status::Success
> ```
> 
> you must fill/read the buffer not just return uninitialized data

Done

stingchang:
@andy31415 can you please take a look at your convenience, thank you

examples/chef/common/chef-rvc-mode-delegate.cpp
```diff
@@ -106,17 +108,68 @@ void RvcRunMode::Shutdown()
...
+    return ret;
+}
+
+chip::Protocols::InteractionModel::Status chefRvcRunModeReadCallback(chip::EndpointId endpointId, chip::ClusterId clusterId,
+                                                                     const EmberAfAttributeMetadata * attributeMetadata,
+                                                                     uint8_t * buffer, uint16_t maxReadLength)
+{
+    if (sizeof(*attributeMetadata) <= sizeof(buffer))
+    {
+        memcpy(buffer, &attributeMetadata, sizeof(attributeMetadata));
```
andy31415, line 154: why is this copying the metadata into the buffer? 

I do not quite understand this code. I thought the intent of `read` is to read the specified attribute (and given ember, the attribute would be in ember format ... like raw bytes for integers, pascal strings for strings and binary data).

Why is copying the metadata into the buffer the correct way to handle a read?

examples/chef/common/chef-rvc-mode-delegate.cpp
```diff
@@ -197,14 +250,63 @@ void RvcCleanMode::Shutdown()
...
+    return ret;
+}
+
+chip::Protocols::InteractionModel::Status chefRvcCleanModeReadCallback(chip::EndpointId endpointId, chip::ClusterId clusterId,
+                                                                       const EmberAfAttributeMetadata * attributeMetadata,
+                                                                       uint8_t * buffer, uint16_t maxReadLength)
+{
+    if (sizeof(*attributeMetadata) <= sizeof(buffer))
+    {
+        memcpy(buffer, &attributeMetadata, sizeof(attributeMetadata));
```
andy31415, line 297: This I do not understand either ....

Realistically if this is a "we do not care about the value" you can maybe 

```cpp
// NOTE: we do NOT care about a real implementation here at this point
memset(buffer, 0, maxReadLength)
```

examples/chef/common/chef-rvc-mode-delegate.cpp
```diff
@@ -106,17 +108,68 @@ void RvcRunMode::Shutdown()
...
+    }
+
+    return ret;
+}
+
+chip::Protocols::InteractionModel::Status chefRvcRunModeReadCallback(chip::EndpointId endpointId, chip::ClusterId clusterId,
+                                                                     const EmberAfAttributeMetadata * attributeMetadata,
+                                                                     uint8_t * buffer, uint16_t maxReadLength)
+{
+    if (sizeof(*attributeMetadata) <= sizeof(buffer))
```
andy31415, line 152: so if `sizeof a struct <= sizeof(pointer)` ? that seems off .... the space in buffer is maxReadLength presumably.

stingchang: right, removed

examples/chef/common/chef-rvc-mode-delegate.cpp
```diff
@@ -197,14 +248,61 @@ void RvcCleanMode::Shutdown()
...
+    switch (attributeId)
+    {
+    case chip::app::Clusters::RvcCleanMode::Attributes::CurrentMode::Id: {
+        uint8_t m                                  = static_cast<uint8_t>(buffer[0]);
+        Protocols::InteractionModel::Status status = clusterInstance->UpdateCurrentMode(m);
+        if (Protocols::InteractionModel::Status::Success == status)
+        {
+            break;
+        }
+        ret = chip::Protocols::InteractionModel::Status::UnsupportedWrite;
```
andy31415, line 276: why do you replace the status from above with unsupportedwrite? why is this not `ret = status` ? A comment explaining why any failure status should go to unsupported write is needed here...

examples/chef/common/chef-rvc-mode-delegate.cpp
```diff
@@ -197,14 +248,61 @@ void RvcCleanMode::Shutdown()
...
+
+    return ret;
+}
+
+chip::Protocols::InteractionModel::Status chefRvcCleanModeReadCallback(chip::EndpointId endpointId, chip::ClusterId clusterId,
+                                                                       const EmberAfAttributeMetadata * attributeMetadata,
+                                                                       uint8_t * buffer, uint16_t maxReadLength)
+{
+    uint8_t m = gRvcCleanModeInstance->GetCurrentMode();
+    memcpy(buffer, &m, sizeof(m));
```
andy31415, line 294:
```cpp
VerifyOrReturnValue(maxReadLength > 0, chip::Protocols::InteractionModel::Status::ResourceExhausted);
buffer[0] = gRvcCleanModeInstance->GetCurrentMode();  // data type is u8
return chip::Protocols::InteractionModel::Status::Success;
```

Seems to be the concise implementation with some verification as well.

examples/chef/common/chef-rvc-operational-state-delegate.cpp
```diff
@@ -154,12 +183,66 @@ void RvcOperationalState::Shutdown()
...
+    }
+
+    return ret;
+}
+
+chip::Protocols::InteractionModel::Status chefRvcOperationalStateReadCallback(chip::EndpointId endpoint, chip::ClusterId clusterId,
+                                                                              const EmberAfAttributeMetadata * attributeMetadata,
+                                                                              uint8_t * buffer, uint16_t maxReadLength)
+{
+    app::DataModel::Nullable<uint8_t> m = gRvcOperationalStateInstance->GetCurrentPhase();
```
andy31415, line 234: Is this supposed to be phase? You read an "operational state".

andy31415: You could read anyting: current phase or operationalstate or maybe even state list or error struct. I am unsure about naming .. the name seems to imply you can read anything.

examples/chef/common/chef-rvc-operational-state-delegate.cpp
```diff
@@ -17,17 +17,31 @@
...
 
-CHIP_ERROR GenericOperationalStateDelegateImpl::GetOperationalStateAtIndex(size_t index, GenericOperationalState & operationalState)
+static void onOperationalStateTimerTick(System::Layer * systemLayer, void * data);
+
+DataModel::Nullable<uint32_t> RvcOperationalStateDelegate::GetCountdownTime()
+{
+    if (mCountDownTime.IsNull())
+        return DataModel::NullNullable;
+
+    return DataModel::MakeNullable((uint32_t) (mCountDownTime.Value() - mRunningTime));
```
tcarmelveilleux, line 41: `mCountDownTime` should be called `mPhaseDuration`, since it's the total duration of the phase, which is the initial value of countdown time.

Also, we must prevent underrun if mRunningTIme >= mPhaseDuration.

examples/chef/common/chef-rvc-operational-state-delegate.cpp
```diff
@@ -46,100 +60,115 @@ CHIP_ERROR GenericOperationalStateDelegateImpl::GetOperationalPhaseAtIndex(size_
...
+void RvcOperationalStateDelegate::HandleStartStateCallback(GenericOperationalError & err)
 {
+    OperationalState::GenericOperationalError current_err(to_underlying(OperationalState::ErrorStateEnum::kNoError));
+    GetInstance()->GetCurrentOperationalError(current_err);
+
+    if (current_err.errorStateID != to_underlying(OperationalState::ErrorStateEnum::kNoError))
+    {
+        err.Set(to_underlying(OperationalState::ErrorStateEnum::kUnableToStartOrResume));
+        return;
+    }
```
tcarmelveilleux, line 100: Why do we override this to UnableToStartOrResume here, and not in cluster logic?

examples/chef/common/chef-rvc-operational-state-delegate.cpp
```diff
@@ -46,100 +60,115 @@ CHIP_ERROR GenericOperationalStateDelegateImpl::GetOperationalPhaseAtIndex(size_
...
-        delete gOperationalStateInstance;
-        gOperationalStateInstance = nullptr;
+        if (state == OperationalState::OperationalStateEnum::kRunning)
+        {
+            delegate->mRunningTime++;
+        }
+        else if (state == OperationalState::OperationalStateEnum::kPaused)
+        {
+            delegate->mPausedTime++;
+        }
```
tcarmelveilleux, line 159: Why do we keep track of paused time?

examples/chef/common/chef-rvc-operational-state-delegate.cpp
```diff
@@ -154,12 +183,66 @@ void RvcOperationalState::Shutdown()
...
+    {
+    case chip::app::Clusters::RvcOperationalState::Attributes::CurrentPhase::Id: {
+        uint8_t m = static_cast<uint8_t>(buffer[0]);
+        DataModel::Nullable<uint8_t> aPhase(m);
+        CHIP_ERROR err = gRvcOperationalStateInstance->SetCurrentPhase(aPhase);
+        if (CHIP_NO_ERROR == err)
+        {
+            break;
+        }
+        ret = chip::Protocols::InteractionModel::Status::UnsupportedWrite;
```
tcarmelveilleux, line 206: UnsupportedWRite is the wrong error for all RPCs. Suggest ConstraintError  to match spec.

examples/chef/common/chef-rvc-operational-state-delegate.cpp
tcarmelveilleux: This line is reading the wrong field. This should always be null if the list is empty.

examples/chef/common/chef-rvc-operational-state-delegate.cpp
```diff
@@ -154,12 +183,66 @@ void RvcOperationalState::Shutdown()
...
+    return chip::Protocols::InteractionModel::Status::Success;
+}
+
 void emberAfRvcOperationalStateClusterInitCallback(chip::EndpointId endpointId)
 {
     VerifyOrDie(endpointId == 1); // this cluster is only enabled for endpoint 1.
     VerifyOrDie(gRvcOperationalStateInstance == nullptr && gRvcOperationalStateDelegate == nullptr);
 
-    gRvcOperationalStateDelegate        = new RvcOperationalStateDelegate;
+    gRvcOperationalStateDelegate        = new chip::app::Clusters::RvcOperationalState::RvcOperationalStateDelegate();
```
tcarmelveilleux, line 245: Do not new here. Instead, have a static instance. Otherwise there is a leak at shutdown. This is true of gRvcOperationalStateInstance  as well.

examples/chef/common/chef-rvc-operational-state-delegate.cpp
erwinpan1: I think this is respecting the spec of 1.15.6.3 Start Command 
```
A device that is unable to honor the Start command for whatever reason SHALL respond 
with an "OperationalCommandResponse" command 
with an ErrorStateID of UnableToStartOrResume 
but take no further action.
```
image

examples/chef/common/chef-rvc-mode-delegate.cpp
```diff
@@ -106,17 +108,78 @@ void RvcRunMode::Shutdown()
...
+}
+
 void emberAfRvcRunModeClusterInitCallback(chip::EndpointId endpointId)
 {
     VerifyOrDie(endpointId == 1); // this cluster is only enabled for endpoint 1.
     VerifyOrDie(gRvcRunModeDelegate == nullptr && gRvcRunModeInstance == nullptr);
     gRvcRunModeDelegate = new RvcRunMode::RvcRunModeDelegate;
-    gRvcRunModeInstance =
-        new ModeBase::Instance(gRvcRunModeDelegate, 0x1, RvcRunMode::Id, chip::to_underlying(RvcRunMode::Feature::kOnOff));
+    gRvcRunModeInstance = new ModeBase::Instance(gRvcRunModeDelegate, endpointId, RvcRunMode::Id,
```
tcarmelveilleux, line 175: This is still using a bare new.

examples/chef/common/chef-rvc-operational-state-delegate.cpp
```diff
@@ -46,125 +60,203 @@ CHIP_ERROR GenericOperationalStateDelegateImpl::GetOperationalPhaseAtIndex(size_
...
+void RvcOperationalStateDelegate::HandleStartStateCallback(GenericOperationalError & err)
 {
+    OperationalState::GenericOperationalError current_err(to_underlying(OperationalState::ErrorStateEnum::kNoError));
+    GetInstance()->GetCurrentOperationalError(current_err);
+
+    if (current_err.errorStateID != to_underlying(OperationalState::ErrorStateEnum::kNoError))
+    {
+        err.Set(to_underlying(OperationalState::ErrorStateEnum::kUnableToStartOrResume));
+        return;
+    }
```
tcarmelveilleux, line 100: My point is that the base class version should have something sane that avoids needing to override with boilerplate. For now you can leave as-is.

tcarmelveilleux (approved):

examples/chef/common/chef-rvc-mode-delegate.cpp
```diff
@@ -197,14 +260,76 @@ void RvcCleanMode::Shutdown()
...
+    return ret;
+}
+
+chip::Protocols::InteractionModel::Status chefRvcCleanModeReadCallback(chip::EndpointId endpointId, chip::ClusterId clusterId,
+                                                                       const EmberAfAttributeMetadata * attributeMetadata,
+                                                                       uint8_t * buffer, uint16_t maxReadLength)
+{
+    VerifyOrReturnValue(maxReadLength > 0, chip::Protocols::InteractionModel::Status::ResourceExhausted);
+    buffer[0] = gRvcCleanModeInstance->GetCurrentMode();
+    return chip::Protocols::InteractionModel::Status::Success;
```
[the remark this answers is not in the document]
stingchang: Removed, thank you

examples/chef/common/chef-rvc-mode-delegate.cpp
stingchang: Updated

PR #33259, opened 2024-05-01
Replace `DispatchSessionEvent` with `NotifySessionHang` on sessions.

andy31415:
Only a single event was ever dispatched and the pattern of passing in pointers to member methods was not used anywhere else and resulted in fairly unique code.

Spelling out the `NotifySessionHang` explicitly seems to simplify maintainability.

### DIFF
--- a/src/transport/Session.h
+++ b/src/transport/Session.h
@@ -114,7 +114,7 @@ public:
 
     Transport::Session * operator->() const { return &mSession.Value().Get(); }
 
-    // There is not delegate, nothing to do here
+    // There is no delegate, nothing to do here
     virtual void OnSessionHang() {}
 
 protected:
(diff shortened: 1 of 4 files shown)


src/transport/Session.h
```diff
@@ -115,7 +115,7 @@ class SessionHolder : public IntrusiveListNodeBase<>
     Transport::Session * operator->() const { return &mSession.Value().Get(); }
 
     // There is not delegate, nothing to do here
```
mkardous-silabs, line 117:
```suggestion
    // There is no delegate, nothing to do here
```

mkardous-silabs (approved):

cecille (approved):

bzbarsky-apple (approved):
Probably OK... we used to have more events before, I thought.

PR #32632, opened 2024-03-19
[Scenes] Missing Tests Cases

lpbeliveau-silabs:
Implemented missing yaml tests cases for points 1 to 3 of https://github.com/CHIP-Specifications/chip-test-plans/issues/4060 and server checks in code to verify validity of ExtensionFieldSets attributes.

Fixes: https://github.com/CHIP-Specifications/chip-test-plans/issues/4060

Also added a MockNodeConfig for Scenes, 
Fixes: https://github.com/project-chip/connectedhomeip/issues/30696

Added test to check for failure on wrong attribute id value,
fixes: https://github.com/project-chip/connectedhomeip/issues/33821

Fixes: https://github.com/CHIP-Specifications/chip-test-plans/issues/4060

PR #33260, opened 2024-05-01
[Scenes] AttributeValue refactor

lpbeliveau-silabs:
Updated scene.xml AttributeValuePairStruct to match the spec.
Updated scene handlers and test scripts to use the new valueUnsigned* instead of AttributeValue.


src/app/tests/suites/certification/Test_TC_S_2_3.yaml
```diff
@@ -395,16 +395,16 @@ tests:
           [1708071897.294574][7731:7733] CHIP:TOO:         ClusterID: 6
           [1708071897.294580][7731:7733] CHIP:TOO:         AttributeValueList: 1 entries
           [1708071897.294589][7731:7733] CHIP:TOO:           [1]: {
-          [1708071897.294593][7731:7733] CHIP:TOO:             AttributeID: 1
-          [1708071897.294596][7731:7733] CHIP:TOO:             AttributeValue: 1
+          [1708071897.294593][7731:7733] CHIP:TOO:             AttributeID: 0
```
jmartinez-silabs, line 398: was this a error before or why does the attribute id changes?

jmartinez-silabs (approved):

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -371,11 +395,26 @@ class DefaultColorControlSceneHandler : public scenes::DefaultSceneHandlerImpl
...
+    /// @param attributeCount number of attributes in the list, incremented by this function, used to keep track of how many
+    /// attributes from the array are being used for the list to encode
+    template <typename Type>
+    void AddAttributeValuePair(ScenesManagement::Structs::AttributeValuePair::Type * pairs, AttributeId id, Type value,
                                size_t & attributeCount)
     {
-        pairs[attributeCount].attributeID    = id;
-        pairs[attributeCount].attributeValue = value;
+        pairs[attributeCount].attributeID = id;
+        if constexpr (sizeof(Type) == sizeof(uint8_t))
```
andy31415, line 410: however we also have signed/unsigned diffs, so actual selection seems to need more than just sizeof.

Would we need explicit template specialization instead? Or some other approach?

src/app/tests/suites/certification/Test_TC_S_2_3.yaml
lpbeliveau-silabs: Yes, there was an error in the comment, it should have been attribute ID 0 from the start.

src/app/clusters/color-control-server/color-control-server.cpp
lpbeliveau-silabs: This helper is made specifically for the the level control scene handler, which only has unsigned scenable attributes. Since handler implementation differs a lot between clusters, I do not think we need to generalise more for this.

src/app/zap-templates/zcl/data-model/chip/scene.xml
```diff
@@ -32,7 +32,30 @@ limitations under the License.
   <struct name="AttributeValuePair">
```
bzbarsky-apple, line 32: This name does not match the spec.

Please go through this XML and make it match the spec.

Please also speak with @naveenkommareddi and @robszewczyk about whether it's OK to make the changes here and to the cert tests after SVE and whatnot.

src/app/tests/suites/certification/Test_TC_S_2_2.yaml
bzbarsky-apple, line 1: We need to have tests exercising the non-uint8 bits.  At least ColorX/Y for uint16.  If not cert tests then just CI tests.

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -237,52 +237,76 @@ class DefaultColorControlSceneHandler : public scenes::DefaultSceneHandlerImpl
...
+                        colorHueTransitionState->finalEnhancedHue = decodePair.valueUnsigned16.Value();
+                    }
                 }
                 break;
             case Attributes::CurrentSaturation::Id:
                 if (SupportsColorMode(endpoint, ColorControl::EnhancedColorMode::kCurrentHueAndCurrentSaturation))
                 {
-                    colorSaturationTransitionState->finalValue =
-                        std::min(static_cast<uint16_t>(decodePair.attributeValue), colorSaturationTransitionState->highLimit);
+                    if (decodePair.valueUnsigned16.HasValue())
```
bzbarsky-apple, line 267: This does not make sense to me.  Why valueUnsigned16?

src/app/clusters/level-control/level-control.cpp
```diff
@@ -173,19 +173,21 @@ class DefaultLevelControlSceneHandler : public scenes::DefaultSceneHandlerImpl
         pairs[0].attributeID = Attributes::CurrentLevel::Id;
         if (!level.IsNull())
         {
-            pairs[0].attributeValue = level.Value();
+            pairs[0].valueUnsigned8.SetValue(level.Value());
         }
         else
         {
-            chip::app::NumericAttributeTraits<uint32_t>::SetNull(pairs[0].attributeValue);
+            uint8_t nullValue = 0;
```
bzbarsky-apple, line 180:
```suggestion
            NumericAttributeTraits<uint8_t>::StorageType nullValue = 0;
```

bzbarsky-apple (changes requested):
Changes requested because the XML does not match the spec.

src/app/zap-templates/zcl/data-model/chip/scene.xml
lpbeliveau-silabs: Scenes going back to provisional means these changes can be made.

src/app/clusters/color-control-server/color-control-server.cpp
lpbeliveau-silabs: We are putting the cluster back to provisional, so backwards compatibility is not required anymore.

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -371,11 +395,28 @@ class DefaultColorControlSceneHandler : public scenes::DefaultSceneHandlerImpl
...
+    /// @param value attribute value
+    /// @param attributeCount number of attributes in the list, incremented by this function, used to keep track of how many
+    /// attributes from the array are being used for the list to encode
+    template <typename Type>
+    void AddAttributeValuePair(ScenesManagement::Structs::AttributeValuePairStruct::Type * pairs, AttributeId id, Type value,
                                size_t & attributeCount)
     {
-        pairs[attributeCount].attributeID    = id;
-        pairs[attributeCount].attributeValue = value;
+        static_assert(sizeof(Type) == sizeof(uint8_t) || sizeof(Type) == sizeof(uint16_t), "Type must be uint8_t or uint16_t");
```
bzbarsky-apple, line 409: This should be an `is_same_v` assert, no?  Again, otherwise if something signed gets added this function will silently do the wrong thing.....

Ideally, you would just static_assert in the "else" of the if constexpr cascade, but C++ does not seem to allow that.  :(

src/app/zap-templates/zcl/data-model/chip/scene.xml
```diff
@@ -63,7 +86,7 @@ limitations under the License.
     <attribute side="server" code="0x0002" define="FABRIC_SCENE_INFO" type="array" entryType="SceneInfoStruct" writable="false" optional="false">FabricSceneInfo</attribute>
     <command source="client" code="0x00" name="AddScene" response="AddSceneResponse" isFabricScoped="true" optional="false" cli="chip scenes add">
       <description>
-        Add a scene to the scene table. Extension field sets are supported, and are inputed as '{"ClusterID": VALUE, "AttributeValueList":[{"AttributeId": VALUE, "AttributeValue": VALUE}]}'
+        Add a scene to the scene table. Extension field sets are supported, and are inputed as '{"ClusterID": VALUE, "AttributeValueList":[{"AttributeId": VALUE, "ValueX": VALUE}]}'
```
bzbarsky-apple, line 89:
```suggestion
        Add a scene to the scene table. Extension field sets are supported, and are inputed as '{"ClusterID": VALUE, "AttributeValueList":[{"AttributeID": VALUE, "Value*": VALUE}]}'
```

src/app/clusters/on-off-server/on-off-server.cpp
```diff
@@ -193,8 +193,8 @@ class DefaultOnOffSceneHandler : public scenes::DefaultSceneHandlerImpl
         {
             auto & decodePair = pair_iterator.GetValue();
             VerifyOrReturnError(decodePair.attributeID == Attributes::OnOff::Id, CHIP_ERROR_INVALID_ARGUMENT);
-            ReturnErrorOnFailure(
-                mSceneEndpointStatePairs.InsertPair(OnOffEndPointPair(endpoint, static_cast<bool>(decodePair.attributeValue))));
+            ReturnErrorOnFailure(mSceneEndpointStatePairs.InsertPair(
+                OnOffEndPointPair(endpoint, static_cast<bool>(decodePair.valueUnsigned8.Value()))));
```
bzbarsky-apple, line 197: What if someone sends malformed input where decodePair.valueUnsigned8 is missing?  Won't this crash?

src/app/clusters/level-control/level-control.cpp
```diff
@@ -225,11 +225,11 @@ class DefaultLevelControlSceneHandler : public scenes::DefaultSceneHandlerImpl
             switch (decodePair.attributeID)
             {
             case Attributes::CurrentLevel::Id:
-                level = static_cast<uint8_t>(decodePair.attributeValue);
+                level = decodePair.valueUnsigned8.Value();
```
bzbarsky-apple, line 228: What ensures that this does not crash (i.e. that decodePair.valueUnsigned8 has a value)?

src/app/clusters/level-control/level-control.cpp
```diff
@@ -240,7 +240,7 @@ class DefaultLevelControlSceneHandler : public scenes::DefaultSceneHandlerImpl
         // TODO : Implement action on frequency when frequency not provisional anymore
         // if(LevelControlHasFeature(endpoint, LevelControl::Feature::kFrequency)){}
 
-        if (!chip::app::NumericAttributeTraits<uint8_t>::IsNullValue(level))
+        if (!app::NumericAttributeTraits<uint8_t>::IsNullValue(level))
```
bzbarsky-apple, line 243: This doesn't seem to match spec.  Spec says that anything outside [MinLevel, MaxLevel] should be treated as null, no?

Followup is best for fixing this, but please don't lose track of it.

src/app/tests/suites/certification/Test_TC_S_2_2.yaml
```diff
@@ -463,7 +463,7 @@ tests:
                                     [
                                         {
                                             AttributeID: 0x0000,
-                                            AttributeValue: 0x01,
+                                            ValueUnsigned8: 0x01,
```
bzbarsky-apple, line 466: We need tests that send the wrong fields for the attribute ID.  Followup is best for this, since it might need test plan updates, but we absolutely must have them.

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -237,52 +237,76 @@ class DefaultColorControlSceneHandler : public scenes::DefaultSceneHandlerImpl
             case Attributes::CurrentX::Id:
                 if (SupportsColorMode(endpoint, ColorControl::EnhancedColorMode::kCurrentXAndCurrentY))
                 {
-                    if (decodePair.attributeValue)
+                    if (decodePair.valueUnsigned16.HasValue())
```
bzbarsky-apple, line 240: Spec says "SHALL be considered invalid" in this case.  Doesn't that mean there should be some errors returned somewhere when such an invalid struct is presented?

Followup to figure that out.

bzbarsky-apple (commented):
The crash issues should get fixed here, the rest in followups.

src/app/tests/suites/certification/Test_TC_S_2_2.yaml
lpbeliveau-silabs: Added a comment in https://github.com/CHIP-Specifications/chip-test-plans/issues/4060, this is the issue for adding more test cases in the Scenes test plans.

src/app/clusters/level-control/level-control.cpp
lpbeliveau-silabs: The check as described in spec happens on the AddScene command (implementation ongoing here: https://github.com/project-chip/connectedhomeip/pull/32632), converting any values outside bounds to NULL. Therefore, on a RecallScene that would call this handler, the value should either valid or null at this point no?

src/app/clusters/level-control/level-control.cpp
lpbeliveau-silabs: As above, added Added a check on `valueUnsigned8.HasValue()` before, preventing storage of invalid attributes to be implemented in the future.

src/app/clusters/on-off-server/on-off-server.cpp
lpbeliveau-silabs: Added a check on `valueUnsigned8.HasValue()` right before.

Note that this is something the PR https://github.com/project-chip/connectedhomeip/pull/32632 will be addressing once the `ValidateAttributePath` method implements the attribute refactor from this PR.

src/app/clusters/color-control-server/color-control-server.cpp
lpbeliveau-silabs: Just to make sure, you mean in the case where we do not have a value in valueUnsigned16? 

If so I added a `VerifyOrReturnError` check for `HasValue`.

src/app/clusters/level-control/level-control.cpp
bzbarsky-apple: So the thing is, what happens if a scene is stored, then OTA happens and Min/MaxLevel change and make the stored scene invalid?

It sounds like the spec needs to make it much more clear how this situation is handled for scenes in general.

bzbarsky-apple (approved):

src/app/clusters/level-control/level-control.cpp
lpbeliveau-silabs: Since we are using the same helper function than the MoveToLevel command `moveToLevelHandler` I assumed the level control cluster would be able to handle it the same way it handles any commands outside of its bounds. 

The idea was that this section of the code should handle it:
```
    if (state->maxLevel <= level)
    {
        state->moveToLevel = state->maxLevel;
    }
    else if (level <= state->minLevel)
    {
        state->moveToLevel = state->minLevel;
    }
    else
    {
        state->moveToLevel = level;
    }
```
But now I see that the this clashes with the spec, the ExtensionFieldSet behavior states this should behave like null, whereas the level control cluster constrains the value to the closest maximum since it assumes a uint8 in the MoveToLevel command.

I added a check in the handler, this will be something to take in consideration in the other PR as well, we cannot only rely on storage to catch out of bounds values

PR #33261, opened 2024-05-01
YAML checker for CI - ensure no unit testing cluster

cecille:
Adding a checker for YAML tests that reference the unit testing cluster. 

It looks like some of the tests were using this cluster to get around YAML limitations, but this can't be done in cert because it then requires devices to carry a unit testing cluster in their prod image.

This workflow can then serve as a landing place for other quirks we find as we're doing cert stuff.

NOTE that we do have two cert tests currently that do this, and they are NOT being fixed as part of this PR.


.github/workflows/cert_test_checks.yaml
```diff
@@ -0,0 +1,35 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Checkout submodules & Bootstrap
+        uses: ./.github/actions/checkout-submodules-and-bootstrap
+        with:
+          platform: linux
+      - name: Run checks
+        run: |
+          ./scripts/run_in_build_env.sh "python3 scripts/tests/matter_yaml_linter.py"
```
andy31415, line 35: please add trailing newline

andy31415 (dismissed):

.github/workflows/cert_test_checks.yaml
```diff
@@ -0,0 +1,35 @@
...
+      - "src/app/tests/suites/certification/**"
+
+jobs:
+  check-certification-tests:
+    name: Check for common problems in certification tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Checkout submodules & Bootstrap
```
andy31415, line 29: bootstrap and submodules seems painful. Would running on the `chip-build` image (which includes python) or using https://github.com/actions/setup-python be faster?

.github/workflows/cert_test_checks.yaml
cecille: done.

.github/workflows/cert_test_checks.yaml
cecille: done. yes.

bzbarsky-apple (approved):

mkardous-silabs (approved):

PR #33263, opened 2024-05-01
Fix reference cycle between controller and data store in Matter.framework.

bzbarsky-apple:
Also adds some leak detection to unit tests in CI that would have caught this.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDeviceControllerDataStore.mm
+++ b/src/darwin/Framework/CHIP/MTRDeviceControllerDataStore.mm
@@ -22,6 +22,7 @@
 
 #include <lib/core/CASEAuthTag.h>
 #include <lib/core/NodeId.h>
+#include <lib/support/CodeUtils.h>
 #include <lib/support/SafeInt.h>
 
 // FIXME: Are these good key strings? https://github.com/project-chip/connectedhomeip/issues/28973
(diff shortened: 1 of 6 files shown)


src/darwin/Framework/CHIP/MTRDeviceControllerDataStore.mm
```diff
@@ -154,9 +157,15 @@ - (nullable instancetype)initWithController:(MTRDeviceController *)controller
 - (void)fetchAttributeDataForAllDevices:(MTRDeviceControllerDataStoreClusterDataHandler)clusterDataHandler
 {
     __block NSDictionary<NSString *, id> * dataStoreSecureLocalValues = nil;
+    MTRDeviceController * controller = _controller;
+    if (controller == nil) {
+        // Not expected; no way to call delegate without controller.
+        return;
+    }
```
ksperling-apple, line 164:
```suggestion
    VerifyOrReturn(controller != nil); // controller has gone away, no way to call delegate
```

Nit: VerifyOrReturn[Value] would make these checks look tidier.

ksperling-apple (approved):

src/darwin/Framework/CHIP/MTRDeviceControllerDataStore.mm
bzbarsky-apple: Done.

bzbarsky-apple:

PR #33266, opened 2024-05-02
Remove the kDNSServiceFlagsTimeout flag when calling DNSServiceGetAdd…

### LINKED ISSUE #33051: [Darwin] Figure out a better setup to have longer-running DNS resolves instead of doing them as-needed
### Reproduction steps

Right now the basic process for operational resolve on Darwin looks like this:

1. We start with the operational instance name.
2. We do a `DNSServiceResolve` to get the hostname and TXT records.
3. We do a bunch of `DNSServiceGetAddrInfo` (one per interface id we got results on), asking the for IPs for that hostname.
4. Once we get results back from `DNSServiceGetAddrInfo` we pass that info up to the caller.

Importantly: when we do `DNSServiceGetAddrInfo` we pass the `kDNSServiceFlagsTimeout` flag, so if we get nothing quickly we fail out and stop trying, even though our caller is willing to wait for 45 seconds for answers.

We should consider a few changes here:

1. Removing the `kDNSServiceFlagsTimeout` flag. This would be especially helpful with https://github.com/project-chip/connectedhomeip/issues/32512 also resolved, so we can stream the results back as they come in.
2. Generally having a longer-running `DNSServiceGetAddrInfo` for nodes we "care about" instead of kicking off things "when we need the info", so that we don't keep starting/stopping the DNS lookups.

### Platform

darwin

### Platform Version(s)

_No response_

### Type

Manually tested with SDK

_No response_

### Anything else?

_No response_

nivi-apple:
…rInfo

- We should not be terminating the queries if we do not get anything back quickly The queries should be running until we get something back

Fixes: #33051

### DIFF
--- a/src/platform/Darwin/DnssdImpl.cpp
+++ b/src/platform/Darwin/DnssdImpl.cpp
@@ -173,7 +173,7 @@ CHIP_ERROR Register(void * context, DnssdPublishCallback callback, uint32_t inte
     ChipLogProgress(Discovery, "Registering service %s on host %s with port %u and type: %s on interface id: %" PRIu32,
                     StringOrNullMarker(name), StringOrNullMarker(hostname), port, StringOrNullMarker(type), interfaceId);
 
-    const DNSServiceFlags registerFlags = kDNSServiceFlagsNoAutoRename;
+    constexpr DNSServiceFlags registerFlags = kDNSServiceFlagsNoAutoRename;
     RegisterContext * sdCtx             = nullptr;
     if (CHIP_NO_ERROR == MdnsContexts::GetInstance().GetRegisterContextOfTypeAndName(type, name, &sdCtx))
     {
--- a/src/platform/Darwin/DnssdImpl.cpp
+++ b/src/platform/Darwin/DnssdImpl.cpp
@@ -174,7 +174,7 @@ CHIP_ERROR Register(void * context, DnssdPublishCallback callback, uint32_t inte
                     StringOrNullMarker(name), StringOrNullMarker(hostname), port, StringOrNullMarker(type), interfaceId);
 
     const DNSServiceFlags registerFlags = kDNSServiceFlagsNoAutoRename;
-    RegisterContext * sdCtx = nullptr;
+    RegisterContext * sdCtx             = nullptr;
     if (CHIP_NO_ERROR == MdnsContexts::GetInstance().GetRegisterContextOfTypeAndName(type, name, &sdCtx))
     {
         auto err = DNSServiceUpdateRecord(sdCtx->serviceRef, nullptr, registerFlags, record.size(), record.data(), 0 /* ttl */);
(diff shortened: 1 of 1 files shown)


src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -43,7 +43,7 @@ constexpr uint16_t kSRPTimeoutInMsec = 250;
 
 constexpr DNSServiceFlags kRegisterFlags        = kDNSServiceFlagsNoAutoRename;
 constexpr DNSServiceFlags kBrowseFlags          = kDNSServiceFlagsShareConnection;
```
nivi-apple, line 45: Do we need these constants kBrowseFlags, kGetAddrInfoFlags? could just use the kDNSServiceFlagsShareConnection in the api calls? we might want to keep them for future use cases where we might add to the flags

ksperling-apple: I agree, I feel like seeing the flags inline at the call site would be more helpful than having these constants.

kiel-apple: It looks like `kRegisterFlags` is the only one of these constants that ends up being used more than once, and it's reused within a single function.

A middle ground would be to scope these constants to the functions where they are used rather than putting them in global scope, and that seems like the right solution for `kRegisterFlags` - the others are probably more clear with the flags inline.

nivi-apple: I would also add the flag for register and reconfirm records as well in the call site itself to be consistent since it doesn't have multiple flags there. If folks disagree I will revert that

kiel-apple (approved):

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -179,10 +173,11 @@ CHIP_ERROR Register(void * context, DnssdPublishCallback callback, uint32_t inte
     ChipLogProgress(Discovery, "Registering service %s on host %s with port %u and type: %s on interface id: %" PRIu32,
                     StringOrNullMarker(name), StringOrNullMarker(hostname), port, StringOrNullMarker(type), interfaceId);
 
+    DNSServiceFlags registerFlags = kDNSServiceFlagsNoAutoRename;
```
kiel-apple, line 176: this can be `const` here to avoid changes within scope

kiel-apple (approved):

src/platform/Darwin/DnssdImpl.cpp
```diff
@@ -179,10 +173,11 @@ CHIP_ERROR Register(void * context, DnssdPublishCallback callback, uint32_t inte
     ChipLogProgress(Discovery, "Registering service %s on host %s with port %u and type: %s on interface id: %" PRIu32,
                     StringOrNullMarker(name), StringOrNullMarker(hostname), port, StringOrNullMarker(type), interfaceId);
 
-    RegisterContext * sdCtx = nullptr;
+    const DNSServiceFlags registerFlags = kDNSServiceFlagsNoAutoRename;
```
bzbarsky-apple, line 176: Why not constexpr?

bzbarsky-apple (approved):

src/platform/Darwin/DnssdImpl.cpp
nivi-apple: done.

andy31415 (approved):

PR #33269, opened 2024-05-02
[ICD] Convert TestICDManager to gtest

mkardous-silabs:
#### Description
Goal of the PR is to convert the TestICDManager file to gtest.

PR changes : 
- TestICDManager to use gtest 
- Removes the inhenritance of LoopbackMessagingContext to instead use it as a external object. We use dynamic allocation to avoid increase the flash size for the unit tests to be able to run on hardware.

#### Note to reviewers
Created `ICDConfigurationDataTestAccess` to be able to access a private method from the ICDConfigurationData class without making all tests that need API friends with it.

#### Tests
Unit test themselves

### DIFF
--- a/src/app/icd/server/ICDConfigurationData.h
+++ b/src/app/icd/server/ICDConfigurationData.h
@@ -28,19 +28,13 @@ namespace chip {
 namespace app {
 // Forward declaration of ICDManager to allow it to be friend with ICDConfigurationData.
 class ICDManager;
-
-// Forward declaration of TestICDManager tests to allow it to be friend with the ICDConfigurationData.
-// Used in unit tests
-class TestICDManager_TestICDStateObserverOnTransitionToIdleModeGreaterActiveModeDuration_Test;
-class TestICDManager_TestICDStateObserverOnTransitionToIdleModeEqualActiveModeDuration_Test;
-class TestICDManager_TestICDStateObserverOnTransitionToIdleMode0ActiveModeDurationWithoutReq_Test;
-class TestICDManager_TestICDStateObserverOnTransitionToIdleModeEqualActiveModeDuration_Test;
-class TestICDManager_TestICDStateObserverOnTransitionToIdleMode0ActiveModeDurationWithoutReq_Test;
-class TestICDManager_TestICDStateObserverOnTransitionToIdleMode0ActiveModeDurationWittReq_Test;
-class TestICDManager_TestICDModeDurationsWith0ActiveModeDurationWithoutActiveSub_Test;
-class TestICDManager_TestICDModeDurationsWith0ActiveModeDurationWithActiveSub_Test;
-
 } // namespace app
+
+namespace Test {
+// Forward declaration of ICDConfigurationDataTestAccess tests to allow it to be friend with the ICDConfigurationData.
+// Used in unit tests
+class ICDConfigurationDataTestAccess;
+} // namespace Test
 
 /**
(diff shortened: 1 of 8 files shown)


src/messaging/tests/BUILD.gn
```diff
@@ -38,6 +38,9 @@ static_library("helpers") {
     "${chip_root}/src/transport",
     "${chip_root}/src/transport/tests:helpers",
   ]
+
+  # TODO(#29342): Dependency to enable building tests converted to gtest with the messaging context
```
andy31415, line 42: The TODO issue is a closed pull request and does not seem to explain this extra dependency.

Either the header/cpp files depend on it or they do not. Given that the code compiles it seems that they do not need to depend on this ... could you explain and maybe create a separate open issue here?

andy31415: actually I see that MesagingContext.h has `#include nlunit-test` so this dependency should have been here. 

However why does it incldue it? Can we just drop it? if not, then the comment could say this can be removed once MessagingContext does not include nlunit-test.h

andy31415: I see we include it for the constants of SUCCESS/FAILURE. that is really unfortunate.

I would comment as

```cpp
# MessagingContext exposes nl-test compatible setup/teardown functions, specifically
# they return nltest specific SUCCESS/FAILURE constants hence this dependency.
# 
# Once all tests are moved to pw_unittest/gtest, this dependency should be removed
```

mkardous-silabs: done

src/app/icd/server/ICDConfigurationData.h
```diff
@@ -29,9 +29,16 @@ namespace app {
 // Forward declaration of ICDManager to allow it to be friend with ICDConfigurationData.
 class ICDManager;
 
-// Forward declaration of TestICDManager to allow it to be friend with the ICDConfigurationData.
+// Forward declaration of TestICDManager tests to allow it to be friend with the ICDConfigurationData.
 // Used in unit tests
-class TestICDManager;
+class TestICDManager_TestICDStateObserverOnTransitionToIdleModeGreaterActiveModeDuration_Test;
```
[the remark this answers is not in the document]
mkardous-silabs: Tests need to be able to Modify the Active/Idle mode durations to trigger specfic edges cases.
The API itself is private because we only want the ICDManager to be able to access it to trigger these updates and any associated buisness logic. 

Since this logic isn't there yet, we use the `friend` option for now.

andy31415 (approved):

jmartinez-silabs (approved):

src/app/icd/server/tests/TestICDManager.cpp
```diff
@@ -118,1061 +122,1007 @@ class TestSubscriptionsInfoProvider : public SubscriptionsInfoProvider
...
+    }
 
-        return CHIP_NO_ERROR;
+    // Performs shared setup for all tests in the test suite
+    static void SetUpTestSuite()
+    {
+        if (pMockClock == nullptr)
+        {
+            pMockClock = new System::Clock::Internal::MockClock();
+            VerifyOrDie(pMockClock != nullptr);
```
Alami-Amine, line 157: shouldn't we use ASSERT_NE instead to stay true to pw_unit format?

mkardous-silabs: Applied

PR #33271, opened 2024-05-02
Add comments in SessionHandle explaining its intent

andy31415:
Why we actually have both ReferenceCountedHandle and SessionHandle was unclear to me. Adding some comments explaining now that I believe I understand it better.

### DIFF
--- a/src/transport/Session.h
+++ b/src/transport/Session.h
@@ -45,6 +45,43 @@ class Session;
  *    as pending removal but not actually removed yet. During this period, the handle is functional, but the underlying session
  *    won't be able to be grabbed by any SessionHolder. SessionHandle->IsActiveSession can be used to check if the session is
  *    active.
+ *
+ *    A `SessionHandle` exists to guarante that the object it points to will not be released while allowing passing
+ *    the handle as a reference, to not incur refeference-counted Retain/Release extra calls.
+ *
+ *    Example code that breaks assumptions (hard to reason about/maintain):
+ *
+ *       void Process(ReferenceCountedHandle<Session> &handle);
+ *       class Foo {
+ *          ReferenceCountedHandle<Session> mSession;
+ *          void ResetSession() { mSession = createNewSession(); }
+ *          void DoProcessing() {
+ *             Process(mSession);
+ *          }
+ *
+ *          static Foo& GetInstance();
+ *       };
+ *
+ *       void Process(ReferenceCountedHandle<Session> &handle) {
+ *          Foo::GetInstance()->ResetSession(); // this changes the passed in handle
+ *          // trying to use "&handle" here may point to something else alltogether.
+ *       }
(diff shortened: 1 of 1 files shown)


src/transport/Session.h
```diff
@@ -45,6 +45,46 @@ class Session;
  *    as pending removal but not actually removed yet. During this period, the handle is functional, but the underlying session
  *    won't be able to be grabbed by any SessionHolder. SessionHandle->IsActiveSession can be used to check if the session is
  *    active.
+ *
+ *    A `SessionHandle` exists to guarante that the object it points to will not be released while allowing passing
```
bzbarsky-apple, line 49:
```suggestion
 *    A `SessionHandle` exists to guarantee that the object it points to will not be released while allowing passing
```

src/transport/Session.h
```diff
@@ -45,6 +45,46 @@ class Session;
...
+ *    the handle as a reference, to not incur refeference-counted Retain/Release extra calls.
```
bzbarsky-apple, line 50:
```suggestion
 *    the handle as a reference, to not incur extra reference-counted Retain/Release calls.
```

src/transport/Session.h
```diff
@@ -45,6 +45,46 @@ class Session;
...
+ *          void DoProcessing() {
+ *             Process(mSession);
+ *          }
+ *
+ *          static Foo& GetInstance();
+ *       };
+ *
+ *       void Process(ReferenceCountedHandle<Session> &handle) {
+ *          Foo::GetInstance()->ResetSession(); // this changes the passed in handle
+ *          // trying to use "&handle" here may point to something else alltogether.
```
bzbarsky-apple, line 67:
```suggestion
 *          // trying to use "&handle" here may point to something else altogether.
```

src/transport/Session.h
```diff
@@ -45,6 +45,46 @@ class Session;
...
+ *      void Process(SessionHandle &handle);
+ *
+ *      ....
+ *      void Foo::DoProcessing() {
+ *         SessionHandle handle(mSession);  // retains the session and mSession can be independently changed
+ *         Process(&handle);  // reference is now safe to use.
+ *      }
+ *
+ *   To meet the requirements of "you should not store this", the Handle has additional restrictions
+ *   preventing modification (no assignment or copy constructor) and allows only move.
```
bzbarsky-apple, line 86: Technically, it should not allow move either....  But that makes it hard to return it by value from methods.  Unfortunately there's not great solution for that.  But might be worth pointing out in case someone comes up with one.

bzbarsky-apple (approved):

yufengwangca (approved):

src/transport/Session.h
andy31415: Added description about this as a note.

PR #33274, opened 2024-05-02
[Silabs] Merge PlatformManagerImpl.cpp for efr32 and si917 platforms

jmartinez-silabs:
Both MCU series had a PlatformManagerImpl file and those were, for the most part, identical.

this pr 
- moves one of the files to `src/platform/silabs`, which is common for both MCU series, and deletes the other.
- #if code that was unique to si917 + small cleanup.
- update path to the file in each mcu series build.gn


src/platform/silabs/PlatformManagerImpl.cpp
```diff
@@ -39,30 +40,27 @@
...
-
     sys_mutex_lock(&rngMutexHandle);
-    res = (chip::Crypto::DRBG_get_bytes(dest, size) == CHIP_NO_ERROR) ? size : 0;
+    int res = (chip::Crypto::DRBG_get_bytes(dest, size) == CHIP_NO_ERROR) ? size : 0;
     sys_mutex_unlock(&rngMutexHandle);
 
     return res;
 }
-#endif
+#endif // TINYCRYPT_PRIMITIVES
```
mkardous-silabs, line 63: I don't think this comment is still accurate since the TINYCRYPT_PRIMITIVES check was removed.

src/platform/silabs/PlatformManagerImpl.cpp
```diff
@@ -39,30 +40,27 @@
...
 
 static void app_get_random(uint8_t * aOutput, size_t aLen)
```
mkardous-silabs, line 65: This function is weird...

mkardous-silabs (approved):

src/platform/silabs/PlatformManagerImpl.cpp
jmartinez-silabs: It is not remove, it is still wrapped in the same condition, I just cleanup the block by 
removing the `#endif` at line 52 (origin file) and additional `#if defined(TINYCRYPT_PRIMITIVES)` at line 54(origin).

the `#if defined(TINYCRYPT_PRIMITIVES)` block starts at line 52 (updated file) and ends at line 63(updated file)

jepenven-silabs (approved):

rcasallas-silabs (approved):

tcarmelveilleux (approved):

PR #33275, opened 2024-05-02
Fix leak in MTRBaseDevice tests.

bzbarsky-apple:
Our test code in deregisterReportHandlersWithQueue would directly delete the ReadClient, but that never called OnDone, and hence never deleted the BufferedReadClientCallback<MTRDataValueDictionaryDecodableType> that we had also allocated.

The fix is to put all the things we need to deallocate in the MTRReadClientContainer and consistently use it to clean those things up, including on all the various failure paths we have.

Removes the CauseReadClientFailure function which did not in fact cause a failure, and instead just has the relevant test trigger a subscription drop.


nivi-apple (approved):

kiel-apple (approved):

bzbarsky-apple:

PR #33285, opened 2024-05-03
Implement separate Produce and Consume SetupPayload validation modes

ksperling-apple:
In Consume mode, unknown rendevouz flags are allowed.

Also (separate commit) simplify the validation code by using VerifyOrReturnValue. `setUpPINCode` range is checked in IsValidSetupPIN() via CheckPayloadCommonConstraints() so no separate check is needed.

### DIFF
--- a/src/setup_payload/SetupPayload.cpp
+++ b/src/setup_payload/SetupPayload.cpp
@@ -36,7 +36,7 @@ namespace chip {
 // Check the Setup Payload for validity
 //
 // `vendor_id` and `product_id` are allowed all of uint16_t
-bool PayloadContents::isValidQRCodePayload() const
+bool PayloadContents::isValidQRCodePayload(ValidationMode mode) const
 {
     // 3-bit value specifying the QR code payload version.
     if (version >= 1 << kVersionFieldLengthInBits)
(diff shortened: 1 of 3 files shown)


andy31415 (approved):

src/setup_payload/SetupPayload.cpp
```diff
@@ -36,60 +36,43 @@ namespace chip {
...
+    VerifyOrReturnValue(version < (1 << kVersionFieldLengthInBits), false);
 
-    if (static_cast<uint8_t>(commissioningFlow) > static_cast<uint8_t>((1 << kCommissioningFlowFieldLengthInBits) - 1))
-    {
-        return false;
-    }
+    VerifyOrReturnValue(static_cast<uint8_t>(commissioningFlow) < (1 << kCommissioningFlowFieldLengthInBits), false);
 
     // Device Commissioning Flow
+    // Even in ValidatoinMode::kConsume we can only handle modes that we understand.
```
bzbarsky-apple, line 47:
```suggestion
    // Even in ValidationMode::kConsume we can only handle modes that we understand.
```

bzbarsky-apple (approved):

PR #33288, opened 2024-05-03
pkg-config.py: fix IndexError when --variable=prefix is empty

shr-project:
sometimes (e.g. with openssl-3.3.0) the pkg-config returns just new-line and prefix[-4] then fails with IndexError.

Fixes:

```
Command: python3 build/config/linux/pkg-config.py -s recipe-sysroot --system_libdir lib openssl Returned 1.
stderr:

  File "build/config/linux/pkg-config.py", line 259, in <module>
    sys.exit(main())
             ^^^^^^
  File "build/config/linux/pkg-config.py", line 156, in main
    prefix = GetPkgConfigPrefixToStrip(options, args)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "build/config/linux/pkg-config.py", line 102, in GetPkgConfigPrefixToStrip
    if prefix[-4] == '/usr':
       ~~~~~~^^^^
IndexError: string index out of range
```

### DIFF
--- a/build/config/linux/pkg-config.py
+++ b/build/config/linux/pkg-config.py
@@ -99,7 +99,7 @@ def GetPkgConfigPrefixToStrip(options, args):
     # from pkg-config's |prefix| variable.
     prefix = subprocess.check_output([options.pkg_config,
                                       "--variable=prefix"] + args, env=os.environ).decode('utf-8')
-    if prefix[-4] == '/usr':
+    if prefix[:4] == '/usr':
         return prefix[4:]
     return prefix
 


build/config/linux/pkg-config.py
```diff
@@ -99,7 +99,7 @@ def GetPkgConfigPrefixToStrip(options, args):
     # from pkg-config's |prefix| variable.
     prefix = subprocess.check_output([options.pkg_config,
                                       "--variable=prefix"] + args, env=os.environ).decode('utf-8')
-    if prefix[-4] == '/usr':
+    if len(prefix) > 4 and prefix[-4] == '/usr':
```
andy31415, line 102: `prefix[-4]` makes no sense to me: check_output should return a string so [-4] is a character and it will never equal `/usr`.

Tried it on my machine:

```
❯ ptpython
>>> import subprocess
>>> p = subprocess.check_output(['pkg-config', '--variable=prefix'] + ['openssl']).decode('utf-8')
>>> print('%r' % p)
'/usr\n'
>>> print(p[-4])
u
>>> print(p[-4] == '/usr')
False
```

Should we maybe fix the code instead? I am questioning if this method should be here anymore.

shr-project: Right, this whole if doesn't make much sense to me. It was introduced with initial changes for building with gn in efa630b58e514fa7a3240dcbeb2ffebd5adc68c4 will check if upstream gn still has this chunk or removed it already.

andy31415: we should either fix it like chromium or drop the usage of this method alltogether.

andy31415 (approved):

build/config/linux/pkg-config.py
andy31415: I see your update did the `[:4]` change. Assuming that maybe OK (CI to verify) however since we never got this error, I wonder if we will ever actually run any CI/compilation that hits this...

shr-project: > I wonder if we will ever actually run any CI/compilation that hits this...

openssl-3.3.0 was the only pkg-config I've seen this and this version isn't very widespread yet (e.g. openSUSE Tumbleweed is on 3.1.4, ubuntu 24.04 on 3.0.13) so it might take a while until such Distribution gets included in your CI matrix.

I was seeing this with custom Distribution built with OpenEmbedded where openssl-3.3.0 is included only just couple days since:
https://git.openembedded.org/openembedded-core/commit/?id=1d6cb1592d6883cc504ff0776810312f732664ae

PR #33299, opened 2024-05-03
TC-IDM-4.2 Use the min/max correctly and account for network delay

raul-marquez-csa:
Updates TC-IDM-4.2 as per test plan changes

Category:
Functional

Description:
Validates Interaction Data Model (IDM), specifically subscription responses. Some example of tests run:
  - Subscriptions with varying MaxIntervalCeiling
  - Checks for `InvalidAction` results when subscribing to clusters and attributes without access rights
  - Checks that subscription is not established for invalid MinIntervalFloor
  - Validates that only correctly filtered data is received when a subscription is established

Full test plan link for details:
[[TC-IDM-4.2] Subscription Response Messages from DUT Test Cases.](https://github.com/CHIP-Specifications/chip-test-plans/blob/master/src/interactiondatamodel.adoc#442-tc-idm-42-subscription-response-messages-from-dut-test-cases-dut_server---removed)


src/python_testing/TC_IDM_4_2.py
```diff
@@ -42,6 +44,23 @@
 https://github.com/CHIP-Specifications/chip-test-plans/blob/master/src/interactiondatamodel.adoc#tc-idm-4-2-subscription-response-messages-from-dut-test-cases-dut_server
 '''
 
+class AttributeChangeCallback:
```
cecille: If you can move this into the base, that would be good - we have stuff there for events already, but I'm worried we're starting to get a lot of copy/paste for subscriptions.

src/python_testing/TC_IDM_4_2.py
```diff
@@ -151,18 +193,18 @@ async def test_TC_IDM_4_2(self):
                 "CR1 reads from the DUT the IdleModeDuration attribute and sets SUBSCRIPTION_MAX_INTERVAL_PUBLISHER_LIMIT_SEC = IdleModeDuration")
 
             idleModeDuration = await self.get_idle_mode_duration_sec(CR1)
-
             SUBSCRIPTION_MAX_INTERVAL_PUBLISHER_LIMIT_SEC = idleModeDuration
```
cecille, line 204: nit: i know the test plan has this in caps, but please just use normal python style. variable names don't need to match the test plan exactly.

src/python_testing/TC_IDM_4_2.py
raul-marquez-csa: Done

src/python_testing/TC_IDM_4_2.py
```diff
@@ -376,7 +418,7 @@ async def test_TC_IDM_4_2(self):
         data_version_filter = [(0, Clusters.BasicInformation, data_version)]
 
         # Subscribe to attribute with provided DataVersion
-        sub_cr1_provided_dvf = await CR1.ReadAttribute(
+        sub_cr1_step8 = await CR1.ReadAttribute(
```
mkardous-silabs, line 421:
```suggestion
        sub_cr1_step7 = await CR1.ReadAttribute(
```

src/python_testing/TC_IDM_4_2.py
```diff
@@ -385,41 +427,67 @@ async def test_TC_IDM_4_2(self):
         )
 
         # Verify that the subscription is activated between CR1 and DUT
-        asserts.assert_true(sub_cr1_provided_dvf.subscriptionId, "Subscription not activated")
+        asserts.assert_true(sub_cr1_step8.subscriptionId, "Subscription not activated")
```
mkardous-silabs, line 430:
```suggestion
        asserts.assert_true(sub_cr1_step7.subscriptionId, "Subscription not activated")
```

src/python_testing/TC_IDM_4_2.py
```diff
@@ -385,41 +427,67 @@ async def test_TC_IDM_4_2(self):
...
 
-        sub_cr1_provided_dvf.Shutdown()
+        sub_cr1_step8.Shutdown()
```
mkardous-silabs, line 432:
```suggestion
        sub_cr1_step7.Shutdown()
```

src/python_testing/TC_IDM_4_2.py
```diff
@@ -385,41 +427,67 @@ async def test_TC_IDM_4_2(self):
...
 
         # *** Step 8 ***
-        self.print_step(8, "CR1 sends a subscription request action for an attribute and sets the MinIntervalFloor value to be same as MaxIntervalCeiling. Activate the Subscription between CR1 and DUT. Modify the attribute which has been subscribed to on the DUT.")
+        self.print_step(8, "CR1 sends a subscription request action for an attribute and sets the MinIntervalFloor to min_interval_floor_sec and MaxIntervalCeiling to 10. Activate the Subscription between CR1 and DUT and record the time when the priming ReportDataMessage is received as t_report_sec. Save the returned MaxInterval from the SubscribeResponseMessage as max_interval_sec.")
 
         # Subscribe to attribute
-        same_min_max_interval_sec = 3
+        min_interval_floor_sec = 3
```
mkardous-silabs, line 438:
```suggestion
        min_interval_floor_sec = 10
```

src/python_testing/TC_IDM_4_2.py
```diff
@@ -385,41 +427,67 @@ async def test_TC_IDM_4_2(self):
...
 
-        # Modify attribute value
+        # Record the time when the priming ReportDataMessage is received
+        t_report_sec = time.time()
+
+        # *** Step 9 ***
+        self.print_step(9, "CR1 modifies the attribute which has been subscribed to on the DUT and waits for an incoming ReportDataMessage")
+
+        # Saving the returned MaxInterval from the SubscribeResponseMessage
+        sub_cr1_step8_intervals = sub_cr1_update_value.GetReportingIntervalsSeconds()
```
mkardous-silabs, line 454:
```suggestion
        sub_cr1_step9_intervals = sub_cr1_update_value.GetReportingIntervalsSeconds()
```

src/python_testing/TC_IDM_4_2.py
```diff
@@ -385,41 +427,67 @@ async def test_TC_IDM_4_2(self):
...
+        min_interval_floor_sec, max_interval_sec = sub_cr1_step8_intervals
```
mkardous-silabs, line 455:
```suggestion
        min_interval_floor_sec, max_interval_sec = sub_cr1_step9_intervals
```

mkardous-silabs (approved):

src/python_testing/TC_IDM_4_2.py
```diff
@@ -191,16 +218,16 @@ async def test_TC_IDM_4_2(self):
         asserts.assert_true(self.is_valid_uint32_value(sub_cr1_step1_max_interval_ceiling_sec),
                             "MaxInterval is not of uint32 type.")
 
-        # Verify MaxInterval is less than or equal to MaxIntervalCeiling
+        # Verify MaxInterval is less than or equal to subscription_max_interval_publisher_limit_sec
         asserts.assert_less_equal(sub_cr1_step1_max_interval_ceiling_sec, max_interval_ceiling_sec,
-                                  "MaxInterval is not less than or equal to MaxIntervalCeiling")
+                                  "MaxInterval is not less than or equal to subscription_max_interval_publisher_limit_sec")
```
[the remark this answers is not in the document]
raul-marquez-csa: Updated

src/python_testing/TC_IDM_4_2.py
raul-marquez-csa: Updated

src/python_testing/TC_IDM_4_2.py
raul-marquez-csa: Updated

src/python_testing/TC_IDM_4_2.py
raul-marquez-csa: Updated

src/python_testing/TC_IDM_4_2.py
raul-marquez-csa: Updated

src/python_testing/TC_IDM_4_2.py
raul-marquez-csa: Updated

src/python_testing/TC_IDM_4_2.py
cecille: can't do this and it doens't match the test plan. If you do floor == ceiling the device is being asked to send a message before 10 seconds, but also not before 10 seconds.

mkardous-silabs: My mistake - misread the test steps

src/python_testing/TC_IDM_4_2.py
```diff
@@ -385,41 +427,67 @@ async def test_TC_IDM_4_2(self):
...
+
+        # Saving the returned MaxInterval from the SubscribeResponseMessage
+        sub_cr1_step9_intervals = sub_cr1_update_value.GetReportingIntervalsSeconds()
+        min_interval_floor_sec, max_interval_sec = sub_cr1_step9_intervals
+
+        # Get subscription timeout
+        subscription_timeout_sec = sub_cr1_update_value.GetSubscriptionTimeoutMs() / 1000
+
+        # Set Attribute Update Callback
+        node_label_queue = queue.Queue()
```
cecille, line 461: can you move this to be a member of the AttributeChangeCallback? I think that'll make it easier to use. Then you can just make a wait function on that class and we can re-use this whole things quite easily.

ie instead of wait_for_attribute_report, you could then do node_label_update_cb.wait_for_report()

src/python_testing/TC_IDM_4_2.py
```diff
@@ -437,88 +505,88 @@ async def test_TC_IDM_4_2(self):
         except Exception:
```
cecille, line 505: above - I'm pretty sure this is used incorrectly. In my reading, that should be wrapping the function call itself.

except Exception is liable to cause a lint error. It's probably easier to just let that exception do its thing - it will cause the test to fail.

But also...I'm pretty sure that ReadAttribute does not throw a ChipStackError on a read failure. You should get a value decode failure. See read_single_attribute_expect_error function in the matter_testing_support file for the implementation of this.

cecille (changes requested):

src/python_testing/TC_IDM_4_2.py
```diff
@@ -67,6 +66,29 @@ async def get_idle_mode_duration_sec(self, ctrl, ep=ROOT_NODE_ENDPOINT_ID):
             attribute=Clusters.IcdManagement.Attributes.IdleModeDuration
         )
 
+    def get_attribute_value_wait(self, sub, typed_attr_path):
```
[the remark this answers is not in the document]
raul-marquez-csa: removed

src/python_testing/TC_IDM_4_2.py
raul-marquez-csa: **ValueDecodeFailure** states this:
https://github.com/project-chip/connectedhomeip/blob/6bd80b080eff659ceb3b77e5eb08992b69be505e/src/controller/python/chip/clusters/Attribute.py#L276-L280

It doesn't appear to be an actual exception type.

The error in step 10 is expected to come from invalid report intervals, not invalid clusters/attributes.

Maybe that's why **ValueDecodeFailure** isn't being summoned, **InteractionModelError** isn't being thrown either, even though it's what ReadAttribute claims it throws in the function description.

https://github.com/project-chip/connectedhomeip/blob/6bd80b080eff659ceb3b77e5eb08992b69be505e/src/controller/python/chip/ChipDeviceCtrl.py#L1418-L1419

Both **InteractionModelError** and **ChipStackError** inherit from **ChipStackException** 

https://github.com/project-chip/connectedhomeip/blob/6bd80b080eff659ceb3b77e5eb08992b69be505e/src/controller/python/chip/interaction_model/__init__.py#L92

https://github.com/project-chip/connectedhomeip/blob/6bd80b080eff659ceb3b77e5eb08992b69be505e/src/controller/python/chip/exceptions/__init__.py#L34

So for now I wrapped the ReadAttribute call in a **ChipStackError** exception assert, which is the exception it's actually throwing.

src/python_testing/TC_IDM_4_2.py
raul-marquez-csa: Done

src/python_testing/TC_IDM_4_2.py
```diff
@@ -385,41 +427,67 @@ async def test_TC_IDM_4_2(self):
...
 
-        # Modify attribute value
+        # Record the time when the priming ReportDataMessage is received
+        t_report_sec = time.time()
+
+        # *** Step 9 ***
+        self.print_step(9, "CR1 modifies the attribute which has been subscribed to on the DUT and waits for an incoming ReportDataMessage")
+
+        # Saving the returned MaxInterval from the SubscribeResponseMessage
+        sub_cr1_step9_intervals = sub_cr1_update_value.GetReportingIntervalsSeconds()
```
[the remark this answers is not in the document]
raul-marquez-csa: Updated

src/python_testing/TC_IDM_4_2.py
```diff
@@ -198,9 +225,9 @@ async def test_TC_IDM_4_2(self):
         sub_cr1_step1.Shutdown()
 
         # *** Step 2 ***
-        self.print_step(2, "CR1 sends a subscription message to the DUT with MaxIntervalCeiling set to a value less than SUBSCRIPTION_MAX_INTERVAL_PUBLISHER_LIMIT_SEC. DUT sends a report data action to the CR1. CR1 sends a success status response to the DUT. DUT sends a Subscribe Response Message to the CR1 to activate the subscription.")
+        self.print_step(2, "CR1 sends a subscription message to the DUT with MaxIntervalCeiling set to a value less than subscription_max_interval_publisher_limit_sec. DUT sends a report data action to the CR1. CR1 sends a success status response to the DUT. DUT sends a Subscribe Response Message to the CR1 to activate the subscription.")
         min_interval_floor_sec = 1
-        max_interval_ceiling_sec = max(2, SUBSCRIPTION_MAX_INTERVAL_PUBLISHER_LIMIT_SEC - 5)
+        max_interval_ceiling_sec = max(2, subscription_max_interval_publisher_limit_sec - 5)
```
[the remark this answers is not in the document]
raul-marquez-csa: Need some clarification:

An assert to prevent idleModeDuration to be 1 and fail the test if so?

If I do _max(0, ...)_ and the max turns out to be the 0 itself, that's what **max_interval_ceiling_sec** will be, but then **min_interval_floor_sec** is set to 1, which will make the **MIN** greater than the **MAX**, causing an error. Maybe I'm missing something.

src/python_testing/TC_IDM_4_2.py
```diff
@@ -306,7 +346,11 @@ async def test_TC_IDM_4_2(self):
             targets=[Clusters.AccessControl.Structs.AccessControlTargetStruct(endpoint=1)],
```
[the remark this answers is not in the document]
raul-marquez-csa: Done

src/python_testing/TC_IDM_4_2.py
```diff
@@ -140,30 +167,30 @@ async def test_TC_IDM_4_2(self):
             paaTrustStorePath=str(self.matter_test_config.paa_trust_store_path),
         )
 
-        # Read ServerList attribute
-        self.print_step("0a", "CR1 reads the Descriptor cluster ServerList attribute from EP0")
+        # *** Step 0 ***
+        self.print_step("0", "CR1 reads the Descriptor cluster ServerList attribute from EP0")
```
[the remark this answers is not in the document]
raul-marquez-csa: Done

src/python_testing/TC_IDM_4_2.py
```diff
@@ -376,7 +418,7 @@ async def test_TC_IDM_4_2(self):
         data_version_filter = [(0, Clusters.BasicInformation, data_version)]
 
         # Subscribe to attribute with provided DataVersion
-        sub_cr1_provided_dvf = await CR1.ReadAttribute(
+        sub_cr1_step7 = await CR1.ReadAttribute(
             nodeid=self.dut_node_id,
             attributes=node_label_attr_path,
             reportInterval=(10, 20),
```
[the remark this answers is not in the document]
raul-marquez-csa: Changed reportInterval to min_interval_floor_sec: 3 and max_interval_ceiling_sec: 10 like in step 8 for steps 3-8 and 11-13.

Can you clarify the list/tuple point?

src/python_testing/TC_IDM_4_2.py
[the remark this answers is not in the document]
cecille: good point. Let's do the following then:
- add an assert that SUBSCRIPTION_MAX_INTERVAL_PUBLISHER_LIMIT_SEC >=1
- change this step to less than or equal to SUBSCRIPTION_MAX_INTERVAL_PUBLISHER_LIMIT_SEC. It's just trying to verify that the subscription maxinterval is correct.
- set the floor to 0.

src/python_testing/TC_IDM_4_2.py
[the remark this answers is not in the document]
cecille: you have reportInterval=(10,20) here, and before it was reportInteval=[min,max]. So now it's a tuple, before it was a list. It looks like the tuple is what you want, so I think this is probably ok.

src/python_testing/TC_IDM_4_2.py
```diff
@@ -287,26 +359,43 @@ async def test_TC_IDM_4_2(self):
                 # Cluster controller 2 has no access to
                 attributes=[(0, Clusters.BasicInformation)],
                 keepSubscriptions=False,
-                reportInterval=[3, 3],
+                reportInterval=(min_interval_floor_sec, max_interval_ceiling_sec),
                 autoResubscribe=False
             )
             raise ValueError("Expected exception not thrown")
         except ChipStackError as e:
+            print('INVALID_ACTION_ERROR_CODE')
```
cecille, line 367: remove print.

src/python_testing/TC_IDM_4_2.py
```diff
@@ -287,26 +359,43 @@ async def test_TC_IDM_4_2(self):
                 # Cluster controller 2 has no access to
                 attributes=[(0, Clusters.BasicInformation)],
                 keepSubscriptions=False,
-                reportInterval=[3, 3],
+                reportInterval=(min_interval_floor_sec, max_interval_ceiling_sec),
                 autoResubscribe=False
             )
             raise ValueError("Expected exception not thrown")
         except ChipStackError as e:
```
cecille, line 366: If you check for InteractionModelError here, you should be able to use Status.InvalidAction rather than redefining it, I think.

cecille (commented):
Can you please re-add this to tests.yaml so it's running in the CI?

src/python_testing/TC_IDM_4_2.py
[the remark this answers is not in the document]
raul-marquez-csa: Double checked for all, all are tuples now.

src/python_testing/TC_IDM_4_2.py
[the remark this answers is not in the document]
raul-marquez-csa: Done.

src/python_testing/TC_IDM_4_2.py
raul-marquez-csa: Done.

src/python_testing/TC_IDM_4_2.py
raul-marquez-csa: I thought so too, but it only catches ChipStackError 🤷🏼.

src/python_testing/TC_IDM_4_2.py
```diff
@@ -151,19 +197,32 @@ async def test_TC_IDM_4_2(self):
...
+        # SUBSCRIPTION_MAX_INTERVAL_PUBLISHER_LIMIT_SEC. DUT sends a report data action to the TH. CR1 sends
+        # a success status response to the DUT. DUT sends a Subscribe Response Message to the CR1 to activate
+        # the subscription.
+        self.step(1)
+
+        # Adding 5 to subscription_max_interval_publisher_limit_sec and setting result to MaxIntervalCeiling,
+        # as per test step instructions.
+        # The edge cases here are:
+        #       min_interval_floor_sec = 0, max_interval_ceiling_sec = 6
+        #       min_interval_floor_sec = 3, max_interval_ceiling_sec = 3605
```
cecille, line 224: What does this comment mean?

src/python_testing/TC_IDM_4_2.py
```diff
@@ -244,18 +319,23 @@ async def test_TC_IDM_4_2(self):
             targets=[Clusters.AccessControl.Structs.AccessControlTargetStruct(cluster=Clusters.BasicInformation.id)],
             subjects=[CR2_nodeid])
 
-        self.add_ace_to_dut_acl(CR1, CR2_limited_ace)
+        # Restore original DUT ACL
```
cecille, line 322: add_ace_to_dut_acl writes the ACL, right? So why do you need to write the original first?

src/python_testing/TC_IDM_4_2.py
raul-marquez-csa: I will remove this, it's probably for an older version of my code

src/python_testing/TC_IDM_4_2.py
raul-marquez-csa: Correct, will be removed.

raul-marquez-csa (commented):
Done.

.github/workflows/tests.yaml
```diff
@@ -477,109 +471,92 @@ jobs:
...
+                  scripts/run_in_python_env.sh out/venv './scripts/tests/run_python_test.py --app out/linux-x64-lit-icd-ipv6only-no-ble-no-wifi-tsan-clang-test/lit-icd-app  --factoryreset --app-args "--discriminator 1234 --KVS kvs1 --trace-to json:out/trace_data/app-{SCRIPT_BASE_NAME}.json" --script "src/python_testing/TC_ICDM_2_1.py" --script-args "--storage-path admin_storage.json --commissioning-method on-network --discriminator 1234 --passcode 20202021 --PICS src/app/tests/suites/certification/ci-pics-values --trace-to json:out/trace_data/test-{SCRIPT_BASE_NAME}.json --trace-to perfetto:out/trace_data/test-{SCRIPT_BASE_NAME}.perfetto"'
```
andy31415, line 508: please do not undo these changes

andy31415 (dismissed):
Changes to tests.yaml is too large. We should only have potential changes to a line for TC_IDM_4_2.py (or no changes if execution logic is not changed)

raul-marquez-csa:
> Changes to tests.yaml is too large. We should only have potential changes to a line for TC_IDM_4_2.py (or no changes if execution logic is not changed)

Updated tests.yaml

src/python_testing/TC_IDM_4_2.py
andy31415, line 1: Needs coments for execution. See other files with `# test-runner-...` contents for examples.

raul-marquez-csa: Done.

src/python_testing/TC_IDM_4_2.py
```diff
@@ -229,13 +296,19 @@ async def test_TC_IDM_4_2(self):
...
 
         sub_cr1_step2.Shutdown()
 
         # *** Step 3 ***
-        self.print_step(3, "Setup CR2 such that it does not have access to a specific cluster. CR2 sends a subscription message to subscribe to an attribute on that cluster for which it does not have access.")
+        # Setup CR2 such that it does not have access to a specific cluster. CR2 sends a subscription
+        # message to subscribe to an attribute on that cluster for which it does not have access.
+        self.step(3)
+
+        # Setting reportInterval values for steps 3-8
```
cecille, line 309: why are these being changed for these steps?

src/python_testing/TC_IDM_4_2.py
```diff
@@ -244,18 +317,20 @@ async def test_TC_IDM_4_2(self):
             targets=[Clusters.AccessControl.Structs.AccessControlTargetStruct(cluster=Clusters.BasicInformation.id)],
             subjects=[CR2_nodeid])
 
-        self.add_ace_to_dut_acl(CR1, CR2_limited_ace)
+        # Add limited ACE
+        await self.add_ace_to_dut_acl(CR1, CR2_limited_ace, dut_acl_original)
```
cecille, line 321: I think you've run into a situation where CR2 doesn't have access to anything and you're failing for that reason. Should we maybe add a step to confirm that CR2 CAN subscribe to something?

src/python_testing/TC_IDM_4_2.py
```diff
@@ -376,153 +474,188 @@ async def test_TC_IDM_4_2(self):
...
+        # Verify no subscription was established
+        with asserts.assert_raises(AttributeError):
+            sub_cr1_invalid_intervals.subscriptionId
+
+        # *** Step 11 ***
+        # CR1 sends a subscription request to subscribe to a specific global attribute from
+        # all clusters on all endpoints.
+        self.step(11)
+
+        # Setting reportInterval values for steps 11-13
```
cecille, line 571: why?

src/python_testing/matter_testing_support.py
```diff
@@ -266,6 +266,39 @@ def wait_for_event_report(self, expected_event: ClusterObjects.ClusterEvent, tim
...
+        self._expected_attribute = expected_attribute
+
+    def __call__(self, path: TypedAttributePath, transaction: SubscriptionTransaction):
+        """This is the subscription callback when an attribute is updated.
+           It checks the passed in attribute is the same as the subscribed to attribute and
+           then posts it into the queue for later processing."""
+
+        asserts.assert_equal(path.AttributeType, self._expected_attribute,
+                             f"[AttributeChangeCallback] Attribute mismatch. Expected: {self._expected_attribute}, received: {path.AttributeType}")
+        logging.info(f"[AttributeChangeCallback] Attribute update callback for {path.AttributeType}")
```
cecille, line 281: Please remove log here - this could get distracting. Or downgrade to debug

cecille (commented):
General comment - not going to ask you to change it in this PR, but you may want to reconsider naming the variables after the step numbers - if the test plan ever gets steps added, you'll end up having to change a lot or the code will get confusing.

src/python_testing/TC_IDM_4_2.py
raul-marquez-csa: For steps 3-8 and 11-13 the reportInterval values are not part of the testing per se, they just need to be there. The other steps update those values, so I'm just resetting them to "normal".

src/python_testing/TC_IDM_4_2.py
raul-marquez-csa: For steps 3-8 and 11-13 the reportInterval values are not part of the testing per se, they just need to be there. The other steps update those values, so I'm just resetting them to "normal".

src/python_testing/matter_testing_support.py
raul-marquez-csa: Downgraded to debug.

src/python_testing/TC_IDM_4_2.py
cecille: Can you leave the min at least as-is? The mins set at the top make sense and account for ICDs.

src/python_testing/TC_IDM_4_2.py
cecille: as above - please leave min as-is.

src/python_testing/TC_IDM_4_2.py
raul-marquez-csa: Updated to only set the max interval

src/python_testing/TC_IDM_4_2.py
raul-marquez-csa: Updated to only set the max interval

cecille (approved):

PR #33306, opened 2024-05-04
Refactored transport/raw/tests to more cleanly handle access to private members of TCPBase

feasel0:
TestTCP requires access to private members of TCPBase.  Previously this was done by friending TCPTest and then putting entire test functions into that class.

As suggested by [#33269](https://github.com/project-chip/connectedhomeip/pull/33269), I've isolated the operations that require private access and moved them to an accessor class TCPBaseTestAccess in a separate header file that may be used by any future test that also requires this access.

TCPBase now friends TCPBaseTestAccess instead of the less-clearly-named `chip::Transport::TCPTest`, and thus many of the references to the name `TCPTest` are now gone.  Once `src/inet/tests/TestInetLayer.cpp` has been refactored to use a new TestAccess class, the similarly named `chip::Inet::TCPTest` can be removed from `src/inet/TCPEndPoint.h`, the last place that references that name.

TestUDP.cpp was also refactored so as to match the way TestTCP.cpp defines its fixtures, since the two unit tests are very similar in structure.


src/transport/raw/tests/TestMessageHeader.cpp
```diff
@@ -23,14 +23,13 @@
  *
  */
 
+#include <gtest/gtest.h>
```
Damian-Nordic, line 26: Why these changes? It's OK to have a few include groups if the headers come from different projects.

Damian-Nordic (approved):

src/transport/raw/tests/TestMessageHeader.cpp
feasel0: Good point.  I updated the groupings to comply with the order "Related header, C system headers, C++ standard library headers, other libraries' headers, your project's headers"

andy31415 (approved):

pidarped (approved):

PR #33308, opened 2024-05-05
Changes for large Packetbuffer allocation to support TCP payloads

### LINKED ISSUE #31779: Max buffer allocation size needs to take into account differing max size caps for TCP and MRP
### Feature description

Allocating a new PacketBuffer(using PacketBuffer::New(..)) currently has internal checks against kMaxSizeWithoutReserve(set to CHIP_SYSTEM_CONFIG_PACKETBUFFER_CAPACITY_MAX), that is a hardcoded maximum buffer capacity based on network constraints like MTU. This needs to be updated to utilize a slightly more flexible approach during allocation for systems that support both MRP and TCP so that buffer allocation can check against a larger capacity for TCP but keep the existing one for MRP. 

### Platform

core

### Platform Version(s)

_No response_

### Anything else?

_No response_

pidarped:
Changes to internal checks in SystemPacketBuffer.

Update the length encoding for TCP payloads during send and receive.

Max size config for large packetbuffer size limit in SystemPacketBuffer.h.
    
Test modifications for chainedbuffer receives for TCP.
    - Add test for Buffer length greater than MRP max size.
Disable TCP on nrfconnect platform because of resource requirements.
   Stack allocations for large buffer with default size is exceeding
   limits. Disabling the Test file altogether for this platform would
   prevent all tests from running. Instead, only disabling TCP on
   nrfConnect for now, since it is unused.

Fixes #31779.

### DIFF
--- a/config/nrfconnect/chip-module/CMakeLists.txt
+++ b/config/nrfconnect/chip-module/CMakeLists.txt
@@ -129,7 +129,7 @@ matter_add_gn_arg_bool  ("chip_enable_nfc"                        CONFIG_CHIP_NF
 matter_add_gn_arg_bool  ("chip_enable_ota_requestor"              CONFIG_CHIP_OTA_REQUESTOR)
 matter_add_gn_arg_bool  ("chip_persist_subscriptions"             CONFIG_CHIP_PERSISTENT_SUBSCRIPTIONS)
 matter_add_gn_arg_bool  ("chip_monolithic_tests"                  CONFIG_CHIP_BUILD_TESTS)
-matter_add_gn_arg_bool  ("chip_inet_config_enable_tcp_endpoint"   CONFIG_CHIP_BUILD_TESTS)
+matter_add_gn_arg_bool  ("chip_inet_config_enable_tcp_endpoint"   FALSE)
 matter_add_gn_arg_bool  ("chip_error_logging"                     CONFIG_MATTER_LOG_LEVEL GREATER_EQUAL 1)
 matter_add_gn_arg_bool  ("chip_progress_logging"                  CONFIG_MATTER_LOG_LEVEL GREATER_EQUAL 3)
 matter_add_gn_arg_bool  ("chip_detail_logging"                    CONFIG_MATTER_LOG_LEVEL GREATER_EQUAL 4)
--- a/src/inet/TCPEndPoint.h
+++ b/src/inet/TCPEndPoint.h
@@ -522,7 +522,7 @@ public:
     /**
      * Size of the largest TCP packet that can be received.
      */
-    constexpr static size_t kMaxReceiveMessageSize = System::PacketBuffer::kMaxSizeWithoutReserve;
+    constexpr static size_t kMaxReceiveMessageSize = System::PacketBuffer::kMaxAllocSize;
 
 protected:
     friend class TCPTest;
--- a/src/inet/TCPEndPointImplSockets.cpp
+++ b/src/inet/TCPEndPointImplSockets.cpp
@@ -947,10 +947,9 @@ void TCPEndPointImplSockets::ReceiveData()
         {
             VerifyOrDie(rcvLen > 0);
             size_t newDataLength = rcvBuf->DataLength() + static_cast<size_t>(rcvLen);
-            VerifyOrDie(CanCastTo<uint16_t>(newDataLength));
             if (isNewBuf)
             {
-                rcvBuf->SetDataLength(static_cast<uint16_t>(newDataLength));
+                rcvBuf->SetDataLength(newDataLength);
                 rcvBuf.RightSize();
                 if (mRcvQueue.IsNull())
                 {
@@ -963,7 +962,7 @@ void TCPEndPointImplSockets::ReceiveData()
             }
             else
             {
-                rcvBuf->SetDataLength(static_cast<uint16_t>(newDataLength), mRcvQueue);
+                rcvBuf->SetDataLength(newDataLength, mRcvQueue);
             }
         }
     }
--- a/src/system/SystemConfig.h
+++ b/src/system/SystemConfig.h
@@ -788,3 +788,21 @@ struct LwIPEvent;
 #define CHIP_SYSTEM_CONFIG_USE_ZEPHYR_EVENTFD 0
 #endif
 #endif // CHIP_SYSTEM_CONFIG_USE_ZEPHYR_EVENTFD
+
+/**
+ * @def CHIP_SYSTEM_CONFIG_MAX_LARGE_BUFFER_SIZE_BYTES
+ *
+ * @brief Maximum buffer allocation size of a 'Large' message
+ *
+ * This is the default maximum capacity (including both data and reserve
+ * space) of a large PacketBuffer(exceeding the IPv6 MTU of 1280 bytes).
+ * This shall be used over transports, such as TCP, that support large
+ * payload transfers. Fetching of large command responses or wildcard
+ * subscription responses may leverage this increased bandwidth transfer.
+ * Individual systems may override this size based on their requirements.
+ * Data transfers over MRP should not be using this size for allocating
+ * buffers as they are restricted by the IPv6 MTU.
+ */
+#ifndef CHIP_SYSTEM_CONFIG_MAX_LARGE_BUFFER_SIZE_BYTES
+#define CHIP_SYSTEM_CONFIG_MAX_LARGE_BUFFER_SIZE_BYTES (64000)
+#endif
--- a/src/system/SystemPacketBuffer.cpp
+++ b/src/system/SystemPacketBuffer.cpp
@@ -565,7 +565,7 @@ PacketBufferHandle PacketBufferHandle::New(size_t aAvailableSize, uint16_t aRese
     if (sumOfAvailAndReserved > UINT16_MAX)
     {
         ChipLogError(chipSystemLayer,
-                     "LwIP based systems require total buffer size to be less than UINT16_MAX!."
+                     "LwIP based systems require total buffer size to be less than UINT16_MAX!"
                      "Attempted allocation size = " ChipLogFormatX64,
                      ChipLogValueX64(sumOfAvailAndReserved));
         return PacketBufferHandle();
--- a/src/system/SystemPacketBuffer.cpp
+++ b/src/system/SystemPacketBuffer.cpp
@@ -515,6 +515,21 @@ PacketBufferHandle PacketBufferHandle::New(size_t aAvailableSize, uint16_t aRese
     // Setting a static upper bound on the maximum buffer size allocation for regular sized messages (not large).
     static_assert(PacketBuffer::kMaxSizeWithoutReserve <= UINT16_MAX, "kMaxSizeWithoutReserve should not exceed UINT16_MAX.");
 
+#if INET_CONFIG_ENABLE_TCP_ENDPOINT
+    // Setting a static upper bound on the maximum buffer size allocation for
+    // large messages.
+#if CHIP_SYSTEM_CONFIG_USE_LWIP
+    // LwIP based systems are internally limited to using a u16_t type as the size of a buffer.
+    static_assert(PacketBuffer::kLargeBufMaxSizeWithoutReserve <= UINT16_MAX,
+                  "In LwIP, max size for Large payload buffers cannot exceed UINT16_MAX!");
+#else
+    // Messages over TCP are framed using a length field that is 32 bits in
+    // length.
+    static_assert(PacketBuffer::kLargeBufMaxSizeWithoutReserve <= UINT32_MAX,
+                  "Max size for Large payload buffers cannot exceed UINT32_MAX");
+#endif // CHIP_SYSTEM_CONFIG_USE_LWIP
+#endif // INET_CONFIG_ENABLE_TCP_ENDPOINT
+
     // Ensure that aAvailableSize is bound within a max and is not big enough to cause overflow during
     // subsequent addition of all the sizes.
     if (aAvailableSize > UINT32_MAX)
@@ -547,7 +562,14 @@ PacketBufferHandle PacketBufferHandle::New(size_t aAvailableSize, uint16_t aRese
 #if CHIP_SYSTEM_CONFIG_USE_LWIP
     // LwIP based APIs have a maximum buffer size of UINT16_MAX. Ensure that
     // limit is met during allocation.
-    VerifyOrDieWithMsg(sumOfAvailAndReserved < UINT16_MAX, chipSystemLayer, "LwIP based systems can handle only up to UINT16_MAX!");
+    if (sumOfAvailAndReserved > UINT16_MAX)
+    {
+        ChipLogError(chipSystemLayer,
+                     "LwIP based systems require total buffer size to be less than UINT16_MAX!."
+                     "Attempted allocation size = " ChipLogFormatX64,
+                     ChipLogValueX64(sumOfAvailAndReserved));
+        return PacketBufferHandle();
+    }
 #endif // CHIP_SYSTEM_CONFIG_USE_LWIP
 
     // sumOfAvailAndReserved is no larger than sumOfSizes, which we checked can be cast to
@@ -557,7 +579,7 @@ PacketBufferHandle PacketBufferHandle::New(size_t aAvailableSize, uint16_t aRese
 
     CHIP_SYSTEM_FAULT_INJECT(FaultInjection::kFault_PacketBufferNew, return PacketBufferHandle());
 
-    if (lAllocSize > PacketBuffer::kMaxSizeWithoutReserve)
+    if (lAllocSize > PacketBuffer::kMaxAllocSize)
     {
         ChipLogError(chipSystemLayer, "PacketBuffer: allocation exceeding buffer capacity limits.");
         return PacketBufferHandle();
@@ -621,11 +643,6 @@ PacketBufferHandle PacketBufferHandle::New(size_t aAvailableSize, uint16_t aRese
 PacketBufferHandle PacketBufferHandle::NewWithData(const void * aData, size_t aDataSize, size_t aAdditionalSize,
                                                    uint16_t aReservedSize)
 {
-    if (aDataSize > UINT16_MAX)
-    {
-        ChipLogError(chipSystemLayer, "PacketBuffer: allocation too large.");
-        return PacketBufferHandle();
-    }
     // Since `aDataSize` fits in uint16_t, the sum `aDataSize + aAdditionalSize` will not overflow.
     // `New()` will only return a non-null buffer if the total allocation size does not overflow.
     PacketBufferHandle buffer = New(aDataSize + aAdditionalSize, aReservedSize);
@@ -633,6 +650,8 @@ PacketBufferHandle PacketBufferHandle::NewWithData(const void * aData, size_t aD
     {
         memcpy(buffer.mBuffer->payload, aData, aDataSize);
 #if CHIP_SYSTEM_CONFIG_USE_LWIP
+        // Checks in the New() call catch buffer allocations greater
+        // than UINT16_MAX for LwIP based platforms.
         buffer.mBuffer->len = buffer.mBuffer->tot_len = static_cast<uint16_t>(aDataSize);
 #else
         buffer.mBuffer->len = buffer.mBuffer->tot_len = aDataSize;
@@ -755,18 +774,18 @@ PacketBufferHandle PacketBufferHandle::CloneData() const
         size_t originalDataSize       = original->MaxDataLength();
         uint16_t originalReservedSize = original->ReservedSize();
 
-        if (originalDataSize + originalReservedSize > PacketBuffer::kMaxSizeWithoutReserve)
+        if (originalDataSize + originalReservedSize > PacketBuffer::kMaxAllocSize)
         {
             // The original memory allocation may have provided a larger block than requested (e.g. when using a shared pool),
             // and in particular may have provided a larger block than we are able to request from PackBufferHandle::New().
             // It is a genuine error if that extra space has been used.
-            if (originalReservedSize + original->DataLength() > PacketBuffer::kMaxSizeWithoutReserve)
+            if (originalReservedSize + original->DataLength() > PacketBuffer::kMaxAllocSize)
             {
                 return PacketBufferHandle();
             }
             // Otherwise, reduce the requested data size. This subtraction can not underflow because the above test
-            // guarantees originalReservedSize <= PacketBuffer::kMaxSizeWithoutReserve.
-            originalDataSize = PacketBuffer::kMaxSizeWithoutReserve - originalReservedSize;
+            // guarantees originalReservedSize <= PacketBuffer::kMaxAllocSize.
+            originalDataSize = PacketBuffer::kMaxAllocSize - originalReservedSize;
         }
 
         PacketBufferHandle clone = PacketBufferHandle::New(originalDataSize, originalReservedSize);
--- a/src/system/SystemPacketBuffer.h
+++ b/src/system/SystemPacketBuffer.h
@@ -119,7 +119,7 @@ private:
 
 public:
     /**
-     * The maximum size buffer an application can allocate with no protocol header reserve.
+     * The maximum size of a regular buffer an application can allocate with no protocol header reserve.
      */
 #if CHIP_SYSTEM_CONFIG_USE_LWIP
     static constexpr size_t kMaxSizeWithoutReserve = LWIP_MEM_ALIGN_SIZE(PBUF_POOL_BUFSIZE);
@@ -134,9 +134,29 @@ public:
     static constexpr uint16_t kDefaultHeaderReserve = CHIP_SYSTEM_CONFIG_HEADER_RESERVE_SIZE;
 
     /**
-     * The maximum size buffer an application can allocate with the default protocol header reserve.
+     * The maximum size of a regular buffer an application can allocate with the default protocol header reserve.
      */
     static constexpr size_t kMaxSize = kMaxSizeWithoutReserve - kDefaultHeaderReserve;
+
+    /**
+     * The maximum size of a large buffer(> IPv6 MTU) that an application can allocate with no protocol header reserve.
+     */
+    static constexpr size_t kLargeBufMaxSizeWithoutReserve = CHIP_SYSTEM_CONFIG_MAX_LARGE_BUFFER_SIZE_BYTES;
+
+    /**
+     * The maximum size of a large buffer(> IPv6 MTU) that an application can allocate with the default protocol header reserve.
+     */
+    static constexpr size_t kLargeBufMaxSize = kLargeBufMaxSizeWithoutReserve - kDefaultHeaderReserve;
+
+    /**
+     * Unified constant(both regular and large buffers) for the maximum size that an application can allocate with no
+     * protocol header reserve.
+     */
+#if INET_CONFIG_ENABLE_TCP_ENDPOINT
+    static constexpr size_t kMaxAllocSize = kLargeBufMaxSizeWithoutReserve;
+#else
+    static constexpr size_t kMaxAllocSize          = kMaxSizeWithoutReserve;
+#endif // INET_CONFIG_ENABLE_TCP_ENDPOINT
 
     /**
      * Return the size of the allocation including the reserved and payload data spaces but not including space
--- a/src/system/tests/TestSystemPacketBuffer.cpp
+++ b/src/system/tests/TestSystemPacketBuffer.cpp
@@ -309,7 +309,7 @@ TEST_F_FROM_FIXTURE(TestSystemPacketBuffer, CheckNew)
     {
         const PacketBufferHandle buffer = PacketBufferHandle::New(0, config.reserved_size);
 
-        if (config.reserved_size > PacketBuffer::kMaxSizeWithoutReserve)
+        if (config.reserved_size > PacketBuffer::kMaxAllocSize)
         {
             EXPECT_TRUE(buffer.IsNull());
             continue;
@@ -1605,7 +1605,8 @@ TEST_F_FROM_FIXTURE(TestSystemPacketBuffer, CheckHandleRightSize)
 
 TEST_F_FROM_FIXTURE(TestSystemPacketBuffer, CheckHandleCloneData)
 {
-    uint8_t lPayload[2 * PacketBuffer::kMaxSizeWithoutReserve];
+    uint8_t lPayload[2 * PacketBuffer::kMaxAllocSize];
+
     for (uint8_t & payload : lPayload)
     {
         payload = static_cast<uint8_t>(random());
@@ -1684,7 +1685,7 @@ TEST_F_FROM_FIXTURE(TestSystemPacketBuffer, CheckHandleCloneData)
     // This is only testable on heap allocation configurations, where pbuf records the allocation size and we can manually
     // construct an oversize buffer.
 
-    constexpr uint16_t kOversizeDataSize = PacketBuffer::kMaxSizeWithoutReserve + 99;
+    constexpr size_t kOversizeDataSize = PacketBuffer::kMaxAllocSize + 99;
     PacketBuffer * p = reinterpret_cast<PacketBuffer *>(chip::Platform::MemoryAlloc(kStructureSize + kOversizeDataSize));
     ASSERT_NE(p, nullptr);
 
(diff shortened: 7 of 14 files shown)


Damian-Nordic (commented):
I have a general question before finishing the review, how are you going to support allocation of large messages on platforms that use buffer pools?

I wonder if we should support configurations in which normal messages are allocated from pools but large messages are allocated on the heap. In other words, I'm curious if having a single `PacketBufferHandle::New()` method for allocating both normal and large messages is a good and sufficient way forward.

pidarped:
> I have a general question before finishing the review, how are you going to support allocation of large messages on platforms that use buffer pools?
> 
> I wonder if we should support configurations in which normal messages are allocated from pools but large messages are allocated on the heap. In other words, I'm curious if having a single `PacketBufferHandle::New()` method for allocating both normal and large messages is a good and sufficient way forward.

@Damian-Nordic, platforms that use buffer pools and require TCP need to adapt the implementation towards that. The current model of supporting large messages uses a heap-based allocation scheme and the choice of a heap-based model is done at compile-time. At the moment it seemed an added complexity to make the heap based allocation for TCP dynamic, and allow regular messages on such platforms to use buffer pools. What is the issue in using heaps for both regular and large messages on TCP-enabled systems?

Damian-Nordic:
> @Damian-Nordic, platforms that use buffer pools and require TCP need to adapt the implementation towards that. The current model of supporting large messages uses a heap-based allocation scheme and the choice of a heap-based model is done at compile-time. At the moment it seemed an added complexity to make the heap based allocation for TCP dynamic, and allow regular messages on such platforms to use buffer pools. What is the issue in using heaps for both regular and large messages on TCP-enabled systems?

Buffer (and other) pools are used in embedded devices to secure memory for the most critical functionalities and reduce the heap fragmentation so I was just curious about the plan.

src/inet/TCPEndPointImplSockets.cpp
```diff
@@ -947,10 +947,10 @@ void TCPEndPointImplSockets::ReceiveData()
         {
             VerifyOrDie(rcvLen > 0);
             size_t newDataLength = rcvBuf->DataLength() + static_cast<size_t>(rcvLen);
-            VerifyOrDie(CanCastTo<uint16_t>(newDataLength));
+            VerifyOrDie(newDataLength <= UINT32_MAX);
```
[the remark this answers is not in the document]
pidarped: The reason I kept it around was to have some check that the length does not exceed the 4 byte length field max size that message payloads are framed with over TCP.

src/system/SystemPacketBuffer.cpp
```diff
@@ -593,7 +600,11 @@ PacketBufferHandle PacketBufferHandle::New(size_t aAvailableSize, uint16_t aRese
 PacketBufferHandle PacketBufferHandle::NewWithData(const void * aData, size_t aDataSize, size_t aAdditionalSize,
                                                    uint16_t aReservedSize)
 {
+#if INET_CONFIG_ENABLE_TCP_ENDPOINT
+    if (aDataSize > UINT32_MAX)
```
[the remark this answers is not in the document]
pidarped: Kept this check to ensure data fits within 4 bytes which is the size of the frame length field for messages over TCP.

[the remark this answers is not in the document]
Damian-Nordic: Sure, but you already check in `New()` that the size is not greater than `CHIP_CONFIG_MAX_LARGE_PAYLOAD_SIZE_BYTES`, which is ensured to be within `uint32_t` range. Those checks were needed before only because of the cast to `uint16_t` before calling `New()`.

andy31415:
@pidarped it seems NRF builds complain about `../../../../src/system/tests/TestSystemPacketBuffer.cpp:1610:21: error: stack usage is 128160 bytes `

src/system/SystemPacketBuffer.cpp
```diff
@@ -530,12 +530,19 @@ PacketBufferHandle PacketBufferHandle::New(size_t aAvailableSize, uint16_t aRese
 
     CHIP_SYSTEM_FAULT_INJECT(FaultInjection::kFault_PacketBufferNew, return PacketBufferHandle());
 
-    // TODO: Change the max to a lower value
-    if (aAvailableSize > UINT32_MAX || lAllocSize > PacketBuffer::kMaxSizeWithoutReserve || lBlockSize > UINT32_MAX)
+#if INET_CONFIG_ENABLE_TCP_ENDPOINT
+    if (lAllocSize > CHIP_CONFIG_MAX_LARGE_PAYLOAD_SIZE_BYTES || lBlockSize > UINT32_MAX)
```
andy31415, line 534: should we static-assert that `CHIP_CONFIG_MAX_LARGE_PAYLOAD_SIZE_BYTES >= PacketBuffer::kMaxSizeWithoutReserve` to preserve previous/non-tcp restrictions?

src/system/tests/TestSystemPacketBuffer.cpp
```diff
@@ -309,7 +309,11 @@ TEST_F_FROM_FIXTURE(TestSystemPacketBuffer, CheckNew)
     {
         const PacketBufferHandle buffer = PacketBufferHandle::New(0, config.reserved_size);
 
+#if INET_CONFIG_ENABLE_TCP_ENDPOINT
+        if (config.reserved_size > CHIP_CONFIG_MAX_LARGE_PAYLOAD_SIZE_BYTES)
```
andy31415, line 313: I see this pattern of ifdef repeated several times where we switch back and forth between either PacketBuffer::kMaxSizeWithoutReserve or CHIP_CONFIG_MAX_LARGE_PAYLOAD_SIZE_BYTES

How about having a single place where we define a constant of like `kMaxPacketBufferPayloadSize` or similar name and have this ifdef only once.

src/transport/raw/TCP.cpp
```diff
@@ -40,10 +40,10 @@ namespace {
 using namespace chip::Encoding;
 
 // Packets start with a 16-bit size
-constexpr size_t kPacketSizeBytes = 2;
+constexpr size_t kPacketSizeBytes = sizeof(uint32_t);
```
andy31415, line 43: nit: comment above now seems off

src/inet/TCPEndPointImplSockets.cpp
bzbarsky-apple, line 950: On a 32-bit system, this check is completely meaningless, yes?

We need to figure out what we are _actually_ checking here and why.

src/system/SystemPacketBuffer.cpp
```diff
@@ -530,12 +530,19 @@ PacketBufferHandle PacketBufferHandle::New(size_t aAvailableSize, uint16_t aRese
 
     CHIP_SYSTEM_FAULT_INJECT(FaultInjection::kFault_PacketBufferNew, return PacketBufferHandle());
 
-    // TODO: Change the max to a lower value
-    if (aAvailableSize > UINT32_MAX || lAllocSize > PacketBuffer::kMaxSizeWithoutReserve || lBlockSize > UINT32_MAX)
+#if INET_CONFIG_ENABLE_TCP_ENDPOINT
```
bzbarsky-apple, line 533: So... this got missed in the PR that made this thing take `size_t` for `aAvailableSize`, but all the static_asserts and overflow checks in this function are now broken.  The comment on lines 511/512 no longer has any bearing on reality, etc.

Before we make more changes to this function, can we please fix it to actually be overflow-safe again?

bzbarsky-apple, line 533: Also, its seems weird to condition this on INET_CONFIG_ENABLE_TCP_ENDPOINT.  Seems to me like we should have CHIP_CONFIG_MAX_LARGE_PAYLOAD_SIZE_BYTES as a value.  That value might be 0 (or might be something very large?) if large payloads are not supported, and then we can have non-ifdefed code. I guess this goes to @andy31415 's comment about the logic duplication here too.  What's here right now seems really fragile and is very hard to read, which is a huge red flag for security-critical size checks.

src/system/SystemPacketBuffer.h
```diff
@@ -44,6 +44,19 @@
 #include <lwip/pbuf.h>
 #endif // CHIP_SYSTEM_CONFIG_USE_LWIP
 
+/**
+ * @def CHIP_CONFIG_MAX_LARGE_PAYLOAD_SIZE_BYTES
+ *
+ * @brief Maximum payload size of a 'Large' message
+ */
+#ifndef CHIP_CONFIG_MAX_LARGE_PAYLOAD_SIZE_BYTES
+#define CHIP_CONFIG_MAX_LARGE_PAYLOAD_SIZE_BYTES (64000)
```
bzbarsky-apple, line 53: Config bits should go in config headers, no?

src/transport/raw/TCP.cpp
```diff
@@ -40,10 +40,10 @@ namespace {
...
 
 // TODO: Actual limit may be lower (spec issue #2119)
-constexpr uint16_t kMaxMessageSize = static_cast<uint16_t>(System::PacketBuffer::kMaxSizeWithoutReserve - kPacketSizeBytes);
+constexpr uint32_t kMaxTCPMessageSize = static_cast<uint32_t>(CHIP_CONFIG_MAX_LARGE_PAYLOAD_SIZE_BYTES - kPacketSizeBytes);
```
bzbarsky-apple, line 46: I don't follow this.  we're saying the max message size is smaller than the max payload size?  Note that the old code _did_ make sense, in that it had the message size being smaller than the max size of a buffer we might be stuffing the message + length into.

src/transport/raw/TCP.cpp
```diff
@@ -202,7 +202,7 @@ CHIP_ERROR TCPBase::SendMessage(const Transport::PeerAddress & address, System::
 
     VerifyOrReturnError(address.GetTransportType() == Type::kTcp, CHIP_ERROR_INVALID_ARGUMENT);
     VerifyOrReturnError(mState == TCPState::kInitialized, CHIP_ERROR_INCORRECT_STATE);
-    VerifyOrReturnError(kPacketSizeBytes + msgBuf->DataLength() <= std::numeric_limits<uint16_t>::max(),
+    VerifyOrReturnError(kPacketSizeBytes + msgBuf->DataLength() <= std::numeric_limits<uint32_t>::max(),
```
bzbarsky-apple, line 205: This no longer does the right things with overflow, yes?  (That breakage may predate this PR, and maybe we are not worried about overflow... but on a 32-bit system this test here is always true which is again a huge red flag for me: either the test should not be there because we never hit this case, or the handling of this case is broken on 32-bit systems.)

src/transport/raw/TCPConfig.h
```diff
@@ -69,7 +69,7 @@ namespace chip {
  * @brief Maximum payload size of a message over a TCP connection
  */
 #ifndef CHIP_CONFIG_MAX_TCP_PAYLOAD_SIZE_BYTES
-#define CHIP_CONFIG_MAX_TCP_PAYLOAD_SIZE_BYTES 1000000
+#define CHIP_CONFIG_MAX_TCP_PAYLOAD_SIZE_BYTES (CHIP_CONFIG_MAX_LARGE_PAYLOAD_SIZE_BYTES)
```
bzbarsky-apple, line 72: I don't understand how this is expected to work.  Can you explain, please?

bzbarsky-apple (changes requested):

src/inet/TCPEndPointImplSockets.cpp
pidarped: Removed this check from here.

nicolas17:
Doesn't this break backwards compatibility? Or was *nobody* using TCP with 16-bit message lengths in production yet?

src/transport/raw/TCP.cpp
pidarped: Renamed CHIP_CONFIG_MAX_LARGE_PAYLOAD_SIZE_BYTES to CHIP_SYSTEM_CONFIG_MAX_LARGE_BUFFER_SIZE_BYTES

src/inet/TCPEndPointImplSockets.cpp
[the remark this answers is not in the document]
pidarped: Removed this check.

src/transport/raw/TCPConfig.h
pidarped: The Large buffer define is in SystemConfig.h

src/system/SystemPacketBuffer.cpp
pidarped: Defined constants in SystemPacketBuffer.h for large payload values that have been used in some of the static_asserts here.

src/transport/raw/TCP.cpp
pidarped: Changed the check to be a tighter bound based on the SystemConfig definition of the max large buffer size.

pidarped:
> Doesn't this break backwards compatibility? Or was _nobody_ using TCP with 16-bit message lengths in production yet?
@nicolas17, yes that is right. TCP was not being used in production yet which gave us the opportunity to increase the message length field to 4 bytes to include large messages.

config/nrfconnect/chip-module/CMakeLists.txt
```diff
@@ -129,7 +129,7 @@ matter_add_gn_arg_bool  ("chip_enable_nfc"                        CONFIG_CHIP_NF
 matter_add_gn_arg_bool  ("chip_enable_ota_requestor"              CONFIG_CHIP_OTA_REQUESTOR)
 matter_add_gn_arg_bool  ("chip_persist_subscriptions"             CONFIG_CHIP_PERSISTENT_SUBSCRIPTIONS)
 matter_add_gn_arg_bool  ("chip_monolithic_tests"                  CONFIG_CHIP_BUILD_TESTS)
-matter_add_gn_arg_bool  ("chip_inet_config_enable_tcp_endpoint"   CONFIG_CHIP_BUILD_TESTS)
+matter_add_gn_arg_bool  ("chip_inet_config_enable_tcp_endpoint"   FALSE)
```
andy31415, line 132: This change should be explained in the PR summary: why did we change this from generally "yes for tests" to "never". Describe why this is a better path than removing the test for NRF (e.g. splitting tests into 2 and have one of the cpp files not included for nrf)

src/inet/TCPEndPoint.h
```diff
@@ -522,7 +522,7 @@ class DLL_EXPORT TCPEndPoint : public EndPointBasis<TCPEndPoint>
     /**
      * Size of the largest TCP packet that can be received.
      */
-    constexpr static size_t kMaxReceiveMessageSize = System::PacketBuffer::kMaxSizeWithoutReserve;
+    constexpr static size_t kMaxReceiveMessageSize = System::PacketBuffer::kMaxAllocSize;
```
andy31415, line 525: Could you also explain this change? I believe that the "withoutreserve" vs maxsize is a bit messy, however overall this seems to change a constant with something else, so it is a change in existing behaviourt even for UDP. I would like to make sure it is correct.

src/system/SystemConfig.h
```diff
@@ -788,3 +788,12 @@ struct LwIPEvent;
 #define CHIP_SYSTEM_CONFIG_USE_ZEPHYR_EVENTFD 0
 #endif
 #endif // CHIP_SYSTEM_CONFIG_USE_ZEPHYR_EVENTFD
+
+/**
+ * @def CHIP_SYSTEM_CONFIG_MAX_LARGE_BUFFER_SIZE_BYTES
+ *
+ * @brief Maximum buffer allocation size of a 'Large' message
```
andy31415, line 795: please update a comment to help reading/understanding the code. It is obvious from the name that `CONFIG_MAX_LARGE_BUFFER_SIZE_BYTES` is `configures the maximum size in bytes of a large buffer`.

It should answer things like: what is the buffer/message in this context? where is it used?

Also if this is tied to TCP only this should be explained in the comments.

src/system/SystemPacketBuffer.cpp
```diff
@@ -515,6 +515,23 @@ PacketBufferHandle PacketBufferHandle::New(size_t aAvailableSize, uint16_t aRese
...
+    // large messages.
+#if CHIP_SYSTEM_CONFIG_USE_LWIP
+    static_assert(PacketBuffer::kLargeBufMaxSizeWithoutReserve <= UINT16_MAX,
+                  "In LwIP, max size for Large payload buffers cannot exceed UINT16_MAX!");
+#else
+    static_assert(PacketBuffer::kLargeBufMaxSizeWithoutReserve <= UINT32_MAX,
+                  "Max size for Large payload buffers cannot exceed UINT32_MAX");
+#endif // CHIP_SYSTEM_CONFIG_USE_LWIP
+
+    // Ensure that the definition of the max buffer allocation for regular
```
andy31415, line 529: This seems to re-read the code. Could you update it to say why this is needed? Why is this a sctrict inequality rather than "<="? What does the extra 1 byte give us?

src/system/SystemPacketBuffer.cpp
```diff
@@ -515,6 +515,23 @@ PacketBufferHandle PacketBufferHandle::New(size_t aAvailableSize, uint16_t aRese
     // Setting a static upper bound on the maximum buffer size allocation for regular sized messages (not large).
     static_assert(PacketBuffer::kMaxSizeWithoutReserve <= UINT16_MAX, "kMaxSizeWithoutReserve should not exceed UINT16_MAX.");
 
+#if INET_CONFIG_ENABLE_TCP_ENDPOINT
+    // Setting a static upper bound on the maximum buffer size allocation for
+    // large messages.
+#if CHIP_SYSTEM_CONFIG_USE_LWIP
```
andy31415, line 521: we need some "why" comments here: why is LwIP limited to u16 (presumably some internals?) and why is bsd sockets limited to u32?

src/system/SystemPacketBuffer.cpp
```diff
@@ -557,7 +574,7 @@ PacketBufferHandle PacketBufferHandle::New(size_t aAvailableSize, uint16_t aRese
 
     CHIP_SYSTEM_FAULT_INJECT(FaultInjection::kFault_PacketBufferNew, return PacketBufferHandle());
 
-    if (lAllocSize > PacketBuffer::kMaxSizeWithoutReserve)
+    if (lAllocSize > PacketBuffer::kMaxAllocSize)
```
andy31415, line 577: why this change? Did we update meanings for some constants?

src/system/SystemPacketBuffer.cpp
```diff
@@ -755,18 +769,20 @@ PacketBufferHandle PacketBufferHandle::CloneData() const
         size_t originalDataSize       = original->MaxDataLength();
         uint16_t originalReservedSize = original->ReservedSize();
 
-        if (originalDataSize + originalReservedSize > PacketBuffer::kMaxSizeWithoutReserve)
+        uint32_t maxSize = PacketBuffer::kMaxAllocSize;
```
andy31415, line 772: do we need a separate variable? Should this be const?

To have minimal change we should probably not create a separate variable.

src/system/SystemPacketBuffer.h
```diff
@@ -138,6 +138,18 @@ class DLL_EXPORT PacketBuffer : private pbuf
      */
     static constexpr size_t kMaxSize = kMaxSizeWithoutReserve - kDefaultHeaderReserve;
 
+#if INET_CONFIG_ENABLE_TCP_ENDPOINT
+    static constexpr size_t kLargeBufMaxSizeWithoutReserve = CHIP_SYSTEM_CONFIG_MAX_LARGE_BUFFER_SIZE_BYTES;
+
+    static constexpr size_t kLargeBufMaxSize = kLargeBufMaxSizeWithoutReserve - kDefaultHeaderReserve;
+#endif // INET_CONFIG_ENABLE_TCP_ENDPOINT
+
+#if INET_CONFIG_ENABLE_TCP_ENDPOINT
```
andy31415, line 147: can we merge these? you have `#if tcp ... #endif #if tcp ... ` which seems redundant.

src/system/SystemPacketBuffer.h
```diff
@@ -138,6 +138,18 @@ class DLL_EXPORT PacketBuffer : private pbuf
...
+    static constexpr size_t kMaxAllocSize = kLargeBufMaxSizeWithoutReserve;
+#else
+    static constexpr size_t kMaxAllocSize          = kMaxSizeWithoutReserve;
```
andy31415, line 150: I find this confusing: we still have kMaxSizeWithoutReserve however then kMaxSizeWithoutReserve is not used much except in some asserts and in computing these things. Why do we keep the asserts then?

These sizes need some form of diagram/explanation. I do not know what a `reserve` is and I do not understand the difference between "large buffer" vs other.


https://github.com/project-chip/connectedhomeip/blob/master/src/system/SystemPacketBuffer.h#L94 has a diagram in terms of `Member()` calls ... can we have something similar explaining these constants, probably closer to this area? I cannot seem to just read the PR and see that everything is fine and because of the renaming of kMaxSizeWithoutReserve to kMaxAllocSize I worry that we may change behavior in unintended ways.

pidarped: kMaxSizeWithoutReserve is used as the max allocation size for MRP based logic at several places, including Test code. The reserve space is to contain the encapsulation header fields(Matter message and exchange header). It is defined in detail in SystemConfig.h
The renaming to kMaxAllocSize as a unified constant actually aligns with the diagram in https://github.com/project-chip/connectedhomeip/blob/master/src/system/SystemPacketBuffer.h#L94, which is why I chose it. It is, essentially, the max of AllocSize() that encompasses both Reserve and Data space in the buffer.

config/nrfconnect/chip-module/CMakeLists.txt
pidarped: Added some justification in the PR summary. A future approach would be to split the file up but for now, since the platform is not using TCP, it is disabled.

src/system/SystemPacketBuffer.cpp
```diff
@@ -621,18 +636,15 @@ PacketBufferHandle PacketBufferHandle::New(size_t aAvailableSize, uint16_t aRese
...
-    }
     // Since `aDataSize` fits in uint16_t, the sum `aDataSize + aAdditionalSize` will not overflow.
     // `New()` will only return a non-null buffer if the total allocation size does not overflow.
     PacketBufferHandle buffer = New(aDataSize + aAdditionalSize, aReservedSize);
     if (buffer.mBuffer != nullptr)
     {
         memcpy(buffer.mBuffer->payload, aData, aDataSize);
 #if CHIP_SYSTEM_CONFIG_USE_LWIP
+        // The VerifyOrDie() in the New() call catches buffer allocations greater
+        // than UINT16_MAX for LwIP based platforms.
```
bzbarsky-apple, line 647: This comment doesn't make sense.  Doesn't `New()` just fail (return empty handle) if the requested size is too big?  I don't see a VerifyOrDie there.

src/transport/SessionManager.cpp
```diff
@@ -201,6 +201,17 @@ CHIP_ERROR SessionManager::PrepareMessage(const SessionHandle & sessionHandle, P
         packetHeader.SetSecureSessionControlMsg(true);
     }
 
+    if (sessionHandle->AllowsMRP())
+    {
+        VerifyOrReturnError(message->TotalLength() <= kMaxAppMessageLen, CHIP_ERROR_MESSAGE_TOO_LONG);
+    }
+#if INET_CONFIG_ENABLE_TCP_ENDPOINT
+    else if (sessionHandle->AllowsLargePayload())
```
bzbarsky-apple, line 209: So a few questions here:

1. Should we not enforce `<= kMaxAppMessageLen` for _all_ transports that don't allow large payloads, whether they allow MRP or not?
2. Why is this conditioned on `INET_CONFIG_ENABLE_TCP_ENDPOINT`?   Is that just an attempt to save some codesize when TCP is not supported?  If so, I would prefer that were handled by having `AllowsLargePayload` compile-time conditional or something (done in a followup), not by sprinkling ifdefs on its consumers.

src/transport/raw/MessageHeader.h
```diff
@@ -74,6 +73,18 @@ static constexpr size_t kMaxAppMessageLen = detail::kMaxApplicationPayloadAndMIC
 
 static constexpr uint16_t kMsgUnicastSessionIdUnsecured = 0x0000;
 
+#if INET_CONFIG_ENABLE_TCP_ENDPOINT
```
bzbarsky-apple, line 76: Why does this need to be conditioned on INET_CONFIG_ENABLE_TCP_ENDPOINT?  These are just constexpr constants; they cost nothing....

I guess the packetbuffer constants are conditioned on it, but why is that?

src/transport/raw/TCP.cpp
```diff
@@ -39,11 +39,11 @@ namespace {
 
 using namespace chip::Encoding;
 
-// Packets start with a 16-bit size
-constexpr size_t kPacketSizeBytes = 2;
+// Packets start with a 32-bit size field.
+constexpr size_t kPacketSizeBytes = sizeof(uint32_t);
```
bzbarsky-apple, line 43: Please leave this as "4".  Nothing in theory guarantees that sizeof() is measured in units of 8 bits in C/C++ (though of course a lot of our code would break if it wasn't).

src/transport/raw/TCP.cpp
```diff
@@ -39,11 +39,11 @@ namespace {
...
 
-// TODO: Actual limit may be lower (spec issue #2119)
-constexpr uint16_t kMaxMessageSize = static_cast<uint16_t>(System::PacketBuffer::kMaxSizeWithoutReserve - kPacketSizeBytes);
+constexpr uint32_t kMaxTCPMessageSize =
+    static_cast<uint32_t>(System::PacketBuffer::kLargeBufMaxSizeWithoutReserve - kPacketSizeBytes);
```
bzbarsky-apple, line 46: This is making two assumptions that need to be static_asserted:

1. kLargeBufMaxSizeWithoutReserve fits in uint32_t.
2. kLargeBufMaxSizeWithoutReserve is not smaller than kPacketSizeBytes.

src/transport/raw/TCP.cpp
```diff
@@ -197,12 +197,12 @@ ActiveTCPConnectionState * TCPBase::FindInUseConnection(const Inet::TCPEndPoint
 CHIP_ERROR TCPBase::SendMessage(const Transport::PeerAddress & address, System::PacketBufferHandle && msgBuf)
 {
     // Sent buffer data format is:
-    //    - packet size as a uint16_t
+    //    - packet size as a uint32_t
     //    - actual data
 
     VerifyOrReturnError(address.GetTransportType() == Type::kTcp, CHIP_ERROR_INVALID_ARGUMENT);
     VerifyOrReturnError(mState == TCPState::kInitialized, CHIP_ERROR_INCORRECT_STATE);
-    VerifyOrReturnError(kPacketSizeBytes + msgBuf->DataLength() <= std::numeric_limits<uint16_t>::max(),
```
bzbarsky-apple, line 205: There's a cast of kPacketSizeBytes to uint16_t that was definitely safe because of this check.  But now, we should probably be static_asserting here right before the relevant cast that kPacketSizeBytes <= UINT16_MAX.

src/transport/raw/TCP.cpp
```diff
@@ -324,11 +324,11 @@ CHIP_ERROR TCPBase::ProcessReceivedBuffer(Inet::TCPEndPoint * endPoint, const Pe
...
         }
-        uint16_t messageSize = LittleEndian::Get16(messageSizeBuf);
-        if (messageSize >= kMaxMessageSize)
+        uint32_t messageSize = LittleEndian::Get32(messageSizeBuf);
+        if (messageSize >= kMaxTCPMessageSize)
         {
-
-            // This message is too long for upper layers.
+            // Message is too big for node to process. Disconnect with peer.
+            CloseConnectionInternal(state, CHIP_ERROR_MESSAGE_TOO_LONG, SuppressCallback::No);
```
bzbarsky-apple, line 331: Hmm, this is a behavior change that does not seem directly related to the "replace uint16_t with uint32_t" bits going on elsewhere here.  

And this behavior change was _definitely_ not in the PR description.  This change should probably be a separate PR with a clear explanation of why it's being made.

src/system/SystemPacketBuffer.cpp
pidarped: There is a VerifyOrDie() in line 565 for LwIP. But I think, I will change that to a return of an empty handle instead.

src/system/SystemPacketBuffer.cpp
```diff
@@ -547,7 +562,11 @@ PacketBufferHandle PacketBufferHandle::New(size_t aAvailableSize, uint16_t aRese
 #if CHIP_SYSTEM_CONFIG_USE_LWIP
     // LwIP based APIs have a maximum buffer size of UINT16_MAX. Ensure that
     // limit is met during allocation.
-    VerifyOrDieWithMsg(sumOfAvailAndReserved < UINT16_MAX, chipSystemLayer, "LwIP based systems can handle only up to UINT16_MAX!");
+    if (sumOfAvailAndReserved > UINT16_MAX)
+    {
+        ChipLogError(chipSystemLayer, "LwIP based systems can handle only up to UINT16_MAX!");
```
bzbarsky-apple, line 567: Should make it clear that this is about the total size of a packetbuffer, and perhaps log what the attempted-to-be-allocated size was.

src/system/SystemPacketBuffer.cpp
```diff
@@ -621,18 +640,15 @@ PacketBufferHandle PacketBufferHandle::New(size_t aAvailableSize, uint16_t aRese
...
-        return PacketBufferHandle();
-    }
     // Since `aDataSize` fits in uint16_t, the sum `aDataSize + aAdditionalSize` will not overflow.
     // `New()` will only return a non-null buffer if the total allocation size does not overflow.
     PacketBufferHandle buffer = New(aDataSize + aAdditionalSize, aReservedSize);
     if (buffer.mBuffer != nullptr)
     {
         memcpy(buffer.mBuffer->payload, aData, aDataSize);
 #if CHIP_SYSTEM_CONFIG_USE_LWIP
+        // Checks in the New() call catches buffer allocations greater
```
bzbarsky-apple, line 650:
```suggestion
        // Checks in the New() call catche buffer allocations greater
```

src/transport/raw/TCP.cpp
bzbarsky-apple: > The static_assert for (1) is there in PacketBufferHandle::New().

How does that help _this_ code?  This code is making an assumption.  Someone could easily change the value of kLargeBufMaxSizeWithoutReserve, change the assert in PacketBufferHandle::New and not realize that _this_ code is then broken.

Which is why _this_ code should have asserts about the assumptions it's making, even if other places in the system happen to be making the same assumptions and asserting them already.

src/transport/raw/TCP.cpp
```diff
@@ -324,11 +326,11 @@ CHIP_ERROR TCPBase::ProcessReceivedBuffer(Inet::TCPEndPoint * endPoint, const Pe
...
```
bzbarsky-apple, line 333: Again: this needs to be in a separate PR.  One PR for the mechanical "allow larger sizes" changes that affect all codepaths.  A separate PR for a significant change to the TCP behavior.  Please don't roll fundamentally unrelated changes all into the same PR.

bzbarsky-apple (changes requested):

src/transport/raw/TCP.cpp
pidarped: Ok. Added a static_assert here for 2.

src/transport/raw/TCP.cpp
pidarped: Removed the connection closure. Will float a separate PR after this is merged.

src/system/SystemPacketBuffer.cpp
pidarped: Added text and logged the attempted size.

src/transport/raw/TCP.cpp
pidarped: Removed from this PR.Will float a separate one after this is merged.

src/system/SystemPacketBuffer.cpp
```diff
@@ -547,7 +562,14 @@ PacketBufferHandle PacketBufferHandle::New(size_t aAvailableSize, uint16_t aRese
 #if CHIP_SYSTEM_CONFIG_USE_LWIP
     // LwIP based APIs have a maximum buffer size of UINT16_MAX. Ensure that
     // limit is met during allocation.
-    VerifyOrDieWithMsg(sumOfAvailAndReserved < UINT16_MAX, chipSystemLayer, "LwIP based systems can handle only up to UINT16_MAX!");
+    if (sumOfAvailAndReserved > UINT16_MAX)
+    {
+        ChipLogError(chipSystemLayer,
+                     "LwIP based systems require total buffer size to be less than UINT16_MAX!."
```
bzbarsky-apple, line 568:
```suggestion
                     "LwIP based systems require total buffer size to be less than UINT16_MAX!"
```

src/transport/raw/TCP.cpp
```diff
@@ -39,11 +39,16 @@ namespace {
...
 
-// Packets start with a 16-bit size
-constexpr size_t kPacketSizeBytes = 2;
+// Packets start with a 32-bit size field.
+constexpr size_t kPacketSizeBytes = 4;
 
-// TODO: Actual limit may be lower (spec issue #2119)
-constexpr uint16_t kMaxMessageSize = static_cast<uint16_t>(System::PacketBuffer::kMaxSizeWithoutReserve - kPacketSizeBytes);
+static_assert(System::PacketBuffer::kLargeBufMaxSizeWithoutReserve <= UINT32_MAX,
+              "Max size for Large payload buffers cannot exceed UINT32_MAX");
```
bzbarsky-apple, line 46:
```suggestion
              "Cast below could truncate the value");
```

bzbarsky-apple (approved):

yufengwangca (approved):

andy31415 (approved):

Issue #33047, opened 2024-04-18
[Feature] MRP retry interval tuning for Thread


Damian-Nordic:
### Feature description

**TLDR**
We hear from some members that the current MRP retry intervals for Thread devices are insufficient, resulting in too eager retransmissions and a positive feedback loop that exacerbates network congestion.

An overloaded Thread router that is an intermediate hop for many parallel conversations may become a bottleneck and then it may be burdened with even more traffic due to spurious retransmissions.

**Some history**
1. Initially, Matter used 300ms as default SAI/SII for all types of devices.
2. At the beginning of 2023, we [increased default SAI/SII for Thread devices to 800ms](https://github.com/project-chip/connectedhomeip/pull/24549). The choice of 800ms was made based on measurements done using the following line topology:
   
   ![image](https://github.com/project-chip/connectedhomeip/assets/66667989/2374ac99-048e-4195-9508-995940a8ba7a)

   This PR has also introduced `CHIP_CONFIG_MRP_RETRY_INTERVAL_SENDER_BOOST` configuration that allowed a Thread sender to add an extra time to the MRP backoff. This was necessary to address scenarios in which a Thread device sends a message to a Wi-Fi device that uses low SAI/SII (300ms/500ms):

   ![image](https://github.com/project-chip/connectedhomeip/assets/66667989/22e961d3-5343-4524-856c-f5eec741892c)

3. [Quick search](https://github.com/search?q=repo%3Aproject-chip%2Fconnectedhomeip%20CHIP_CONFIG_MRP_LOCAL_ACTIVE_RETRY_INTERVAL&type=code) reveals that other silicon vendors increased these parameters as well in the past.

**Proposal**
Let's change the default SAI/SII for Thread devices to 2s.

It might be said that this configuration is device-specific and doesn't have to be aligned. However, my concern is that if it's not unified to some degree, controller makers might compensate for too low values used by some accessories, which could lead to excessively long retry intervals in the case of other accessories. Therefore, I believe it's beneficial to align these settings to a certain degree to assure better interoperability.

There's also a question what to do with `CHIP_CONFIG_MRP_RETRY_INTERVAL_SENDER_BOOST`. The most simple answer would be to also increase it to, say, `1500ms`. But that would cause that two Thread devices talking to each other would use pretty long retry intervals: `MRP_backoff(2s) + 1.5s`.

[...]

PR #33314, opened 2024-05-06
[mrp] Increase default retry interval for Thread

Damian-Nordic:
The current 800ms is not enough in real setups, where Thread routers must serve as intermediate hops for many parallel conversations. Bump this to 2s.

Make the additional MRP backoff time (aka CHIP_CONFIG_MRP_RETRY_INTERVAL_SENDER_BOOST) dynamic for all build configurations to remove the need for adjusting timeouts in unit tests whenever this parameter changes.

In messaging tests, by default, set this parameter to 0, except for tests that explicitly verify its meaning.
By the way, fix tests increasing the MRP backoff time by the slow-polling instead of fast-polling interval.

Related to: #33047.


jmartinez-silabs (approved):

src/messaging/ReliableMessageMgr.h
```diff
@@ -220,11 +219,7 @@ class ReliableMessageMgr
      * set this before actually bringing up the stack and having access to a
      * ReliableMessageMgr.
      */
-    static void SetAdditionalMRPBackoffTime(const Optional<System::Clock::Milliseconds64> & additionalTime)
-    {
-        sAdditionalMRPBackoffTime = additionalTime;
-    }
-#endif // CHIP_DEVICE_CONFIG_ENABLE_DYNAMIC_MRP_CONFIG
+    static void SetAdditionalMRPBackoffTime(Optional<System::Clock::Timeout> additionalTime);
```
bzbarsky-apple, line 222: Is there a reason to pass that by value, not reference?

bzbarsky-apple (approved):

src/messaging/ReliableMessageMgr.h
Damian-Nordic: I thought the type is small enough to be passed by value, but this doesn't matter in this case. Reverted.

PR #33319, opened 2024-05-06
Matter.framework should explicitly look for things in the SRP domain.

bzbarsky-apple:
The operational browse and connectivity monitoring in Matter.framework should happen on both the SRP domain and the default domain, so we are sure we'll notice Thread devices properly.

### DIFF
--- a/src/darwin/Framework/CHIP/MTROperationalBrowser.mm
+++ b/src/darwin/Framework/CHIP/MTROperationalBrowser.mm
@@ -21,13 +21,16 @@
 
 #include <cinttypes>
 #include <lib/dnssd/ServiceNaming.h>
+#include <lib/support/CodeUtils.h>
 #include <lib/support/logging/CHIPLogging.h>
 #include <platform/LockTracker.h>
 
 namespace {
-constexpr char kLocalDot[] = "local.";
+constexpr const char * kBrowseDomains[] = {
+    "default.service.arpa.", // SRP
+    "local.",
+};
 constexpr char kOperationalType[] = "_matter._tcp";
-constexpr DNSServiceFlags kBrowseFlags = 0;
 }
 
 MTROperationalBrowser::MTROperationalBrowser(MTRDeviceControllerFactory * aFactory, dispatch_queue_t aQueue)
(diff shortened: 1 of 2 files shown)


woody-apple (approved):

src/darwin/Framework/CHIP/MTROperationalBrowser.mm
```diff
@@ -43,23 +46,30 @@
...
-        ChipLogError(Controller, "Failed to set up dispatch queue properly");
+        ChipLogError(Controller, "Failed to set up dispatch queue properly for persistent operational browse: %" PRId32, err);
         DNSServiceRefDeallocate(mBrowseRef);
         return;
     }
 
     mInitialized = true;
+
+    for (auto domain : kBrowseDomains) {
+        auto browseRef = mBrowseRef; // Mandatory copy because of kDNSServiceFlagsShareConnection.
```
jtung-apple, line 67: My understanding is that the DNSServiceRefs copied/created from a shared connection also need to have `DNSServiceRefDeallocate` called separately when they are done, or we'll be leaking some memory.

bzbarsky-apple: Hmm.  We are very much not doing that in our src/platform code.  My understanding was that when the shared connection is closed, all the resources associated with it are cleaned up.  But let me double-check that.

yunhanw-google (approved):

src/darwin/Framework/CHIP/MTROperationalBrowser.mm
bzbarsky-apple: So https://developer.apple.com/documentation/dnssd/1823436-anonymous/kdnsserviceflagsshareconnection says:

> Calling DNSServiceRefDeallocate(ref) for a particular operation's DNSServiceRef terminates just that operation. Calling DNSServiceRefDeallocate(ref) for the main shared DNSServiceRef (the parent DNSServiceRef, originally created by DNSServiceCreateConnection(&ref)) automatically terminates the shared connection and all operations that were still using it. After doing this, DO NOT then attempt to deallocate any remaining subordinate DNSServiceRef's. The memory used by those subordinate DNSServiceRef's has already been freed, so any attempt to do a DNSServiceRefDeallocate (or any other operation) on them will result in accesses to freed memory, leading to crashes or other equally undesirable results.

So yes, a least as per the API documentation what we have here is fine: once we deallocate `mBrowseRef` all the things go away.

bzbarsky-apple:
CI failure is real, looking into what we should do there.

bzbarsky-apple:

PR #33330, opened 2024-05-07
Make test suite setup and teardown functions static

mbknust:
In order to migrate to pw_unit_test, `SetUpTestSuite` and `TearDownTestSuite` ought to be static member functions of a test fixture class.

But nltest still expects them to take a `void *` argument and return an int, so I wrap them with [macros](https://github.com/mbknust/connectedhomeip/blob/487741605af00e5420c260bcf01c7ffdaa638288/src/lib/support/UnitTestRegistration.h#L51-L63) each time they're used inside a `nlTestSuite` declaration.

**Followup**: Remove `NL_TEST_WRAP_FUNCTION` and `NL_TEST_WRAP_METHOD`, when they are no longer needed.

These changes used to be part of #33002, but that PR was split into smaller pieces.


src/transport/tests/LoopbackTransportManager.h
```diff
@@ -27,26 +27,26 @@ class LoopbackTransportManager
 {
 public:
     /// Initialize the underlying layers.
-    CHIP_ERROR Init()
+    static CHIP_ERROR Init()
```
arkq, line 30: Could you change the name of this function, if it needs to be static. By convention ([almost all](https://github.com/search?q=repo%3Aproject-chip%2Fconnectedhomeip+%22static+CHIP_ERROR+Init%28%29%22&type=code)) `Init()` functions are not static in whole project. Making this `Init()` as static might confuse someone...

mbknust: Fixed; LoopbackTransportManager is no longer a singleton, instead it's a static member of LoopbackMessagingContext

arkq (approved):

andy31415 (approved):

feasel0 (approved):
This looks good.

src/transport/tests/UDPTransportManager.h
```diff
@@ -29,33 +29,33 @@ class UDPTransportManager
 {
 public:
     /// Initialize the underlying layers.
-    CHIP_ERROR Init()
+    static CHIP_ERROR Init()
```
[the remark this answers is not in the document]
mbknust: I changed this code to be written analogously to `LoopbackTransportManager`

PR #33334, opened 2024-05-07
[mrp] Make sure not all packet buffers are used for retransmissions

Damian-Nordic:
On constrained devices, the number of available packet buffers can be very low. During stress testing a device that has 8 buffers available, which is the default on many platforms, it was observed that the device would receive a lot of parallel read requests and then it would use all its buffers for responses. The responses would then be stored in the retransmission table,
awaiting an ACK, but the ACK would never arrive because of lack of available buffers.

Configure the retransmission table size to be less than the number of packet buffers to make sure that not all buffers are held by the retransmission entries, in which case the device is unable to receive an ACK and hence it becomes unavailable until any message times out.


bzbarsky-apple (approved):

LuDuda (approved):
LGTM :+1:

PR #33337, opened 2024-05-07
[Python] Eliminate ZCLSubscribeAttribute

agners:
This eliminates use of ZCLSubscribeAttribute and improves some tests using asyncio.

### DIFF
--- a/src/controller/python/test/test_scripts/base.py
+++ b/src/controller/python/test/test_scripts/base.py
@@ -1284,9 +1284,18 @@ class BaseTestHelper:
                         "Failed to receive subscription update")
                     break
 
-            # thread changes 5 times, and sleeps for 3 seconds in between.
-            # Add an additional 3 seconds of slack. Timeout is in seconds.
-            await asyncio.wait_for(taskAttributeChange, 3)
+            # At this point the task should really have done the three attribute,
+            # otherwise something is wrong. Wait for just 1s in case of a race
+            # condition between the last attribute update and the callback.
+            try:
+                await asyncio.wait_for(taskAttributeChange, 1)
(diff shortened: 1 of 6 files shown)


yunhanw-google (approved):

src/controller/python/test/test_scripts/base.py
```diff
@@ -1199,122 +1201,106 @@ def TestReadBasicAttributes(self, nodeid: int, endpoint: int, group: int):
...
+                # is really wrong and bail out here with some information.
+                try:
+                    await asyncio.wait_for(updateEvent.wait(), 10)
+                    updateEvent.clear()
+                except TimeoutError:
+                    self.logger.error(
+                        "Failed to receive subscription update")
+                    break
 
             # thread changes 5 times, and sleeps for 3 seconds in between.
```
tehampson, line 1287: nit, I think this comment should be updated as we are not waiting for 18s and now only waiting for 3

tehampson (commented):
I would like to take another look after CI Cirque is passing, so just ping me when that is passing

src/controller/python/test/test_scripts/base.py
agners: Hm right :thinking: I guess we were waiting 18s before for the worst case, just to make sure the Thread cleanly exited in case of a problem.

I guess in theory this could happen still, if the test doesn't work, we would fail after 10 seconds, but the task would still run longer. I use a task cancel here to make sure things are cleaned up in any case. Might be a bit overkill as it is a test failure anyways at that point but doesn't hurt either :man_shrugging:

tehampson (approved):

PR #33338, opened 2024-05-07
[Linux] Add mechanism to redirect CHIP log other than stdout

yufengwangca:
Currently, both shell command and CHIP logs are printed out to the same console on Linux, make it very difficult to interact with the example app via shell CLI.

```
[1714771193.157408][264120:264121] CHIP:DL: Got IP address on interface: wlp3s0 IP: 192.168.86.250
[1714771193.157474][264120:264122] CHIP:DL: HandlePlatformSpecificBLEEvent 32770
[1714771194.179146][264120:264121] CHIP:DL: Got IP address on interface: wlp3s0 IP: 192.168.86.250
[1714771194.179241][264120:264122] CHIP:DL: HandlePlatformSpecificBLEEvent 32770 (edited) 
```

Event though I can turn off those logs, but it is better to have a way to redirect logging to another place so that I can only see CLI message in console and check CHIP log in another place

### DIFF
--- a/src/platform/Linux/Logging.cpp
+++ b/src/platform/Linux/Logging.cpp
@@ -32,6 +32,27 @@ void __attribute__((weak)) OnLogOutput() {}
 namespace Logging {
 namespace Platform {
 
+// File pointer for the log file
+FILE * logFile = nullptr;
+
+void OpenLogFile(const char * filePath)
+{
+    logFile = fopen(filePath, "a");
+    if (logFile == nullptr)
+    {
+        perror("Failed to open log file");
+    }
+}
+
+void CloseLogFile()
+{
+    if (logFile != nullptr)
+    {
+        fclose(logFile);
+        logFile = nullptr;
+    }
+}
+
 /**
  * CHIP log output functions.
  */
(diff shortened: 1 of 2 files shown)


yunhanw-google (approved):

src/platform/Linux/Logging.cpp
```diff
@@ -32,6 +32,27 @@ void __attribute__((weak)) OnLogOutput() {}
...
 
+// File pointer for the log file
+FILE * logFile = nullptr;
+
+void OpenLogFile(const char * filePath)
+{
+    logFile = fopen(filePath, "a");
+    if (logFile == nullptr)
+    {
+        perror("Failed to open log file");
```
tcarmelveilleux, line 43: This must VerifyOrDie.

src/platform/logging/LogV.h
```diff
@@ -30,6 +30,25 @@ namespace Platform {
...
+ *                      messages will be appended to it.
+ */
+void OpenLogFile(const char * filePath);
+
+/**
+ * Close the log file if it is open. This function should be called
+ * at the end of the application to ensure that all log messages are
+ * flushed to disk and the file is properly closed.
+ */
+void CloseLogFile();
```
tcarmelveilleux, line 50: Where are these called?

src/platform/logging/LogV.h
```diff
@@ -30,6 +30,25 @@ namespace Platform {
...
+ * Open a log file for appending log messages. This function opens
+ * the specified file and ensures that subsequent logs are written to
+ * this file. It should be called at the start of the application
+ * before any logging is done.
+ *
+ * @param[in] filePath  The path to the log file. If the file does not
+ *                      exist, it will be created. If it exists, log
+ *                      messages will be appended to it.
+ */
+void OpenLogFile(const char * filePath);
```
tcarmelveilleux, line 43: We should have a list of log sinks (one of which is stdout by default) rather than assume a single place where logs go. Otherwise it is impossible to have stdout + output to file logs.

tcarmelveilleux (changes requested):
- [ ] Introduce a log sink interface (base class) with LogV. Concrete implementations can have their own init, and are responsiblity of the owner/caller
- [ ] Make stdout a default log sink for linux
- [ ] Do not add new file methods to src/platform/logging --> not portable or relevant everywhere. AddLogSink/RemoveLogSink should be enough
- [ ] Ensure >1 log sink can be set, at least on Linux, so that when we update chip-tool interactive we have a way to not lose the logs, even on interactive, if needed.

src/platform/Linux/Logging.cpp
bzbarsky-apple, line 1: Why is this Linux-specific?  The PR description sounds like "we want to do something to fix a chip-tool problem", but the changes are to one platform logging implementation, not chip-tool....

src/platform/logging/LogV.h
bzbarsky-apple: We do currently allow setting one log sink (the logging redirector setup), which can then be used to multiplex as desired without changing any platform implementations....

andy31415: https://github.com/project-chip/connectedhomeip/blob/master/src/lib/support/logging/TextOnlyLogging.h#L359 contains a `SetLogRedirectCallback` where you can move your logging to go anywhere. This is what python uses to redirect to python logging instead of just console.

Could we use this for the CLI bits instead? One could even imagine some term-linke thing where you can both do cli-shell-interact and view the log in a separate pane...

yufengwangca: Yes, this method works.

yufengwangca:
Close this PR since we already have the mechanism to redirect the logging.

PR #33340, opened 2024-05-07
Add some unit tests for ExchangeContext and ExchangeMessageDispatch.

bzbarsky-apple:


src/messaging/tests/TestExchange.cpp
```diff
@@ -0,0 +1,266 @@
...
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+/**
+ *    @file
+ *      This file implements unit tests for the ExchangeContext implementation.
```
andy31415, line 20: nit: I think we can drop this file-level comment. Maybe if we find it useful we could name it `TestExchangeContext` to make it clear that we test exchange contexts.

andy31415 (approved):

yunhanw-google (approved):

src/messaging/tests/TestExchange.cpp
bzbarsky-apple: Done.

kiel-apple (approved):

PR #33341, opened 2024-05-07
Make sure to include ember defines for door lock.

andy31415:
Chef was using MATTER_DM_PLUGIN_DOOR_LOCK_SERVER without including relevant headers (and assuming that data model would transitively define them)

This does not work anymore after #32776 , so include app config instead.


tcarmelveilleux (approved):

pgregorr-amazon (commented):
Looks good

PR #33351, opened 2024-05-08
Converted src/controller/tests from NL to PW unit tests.

feasel0:
Fairly straightforward conversion, albeit lengthy.  This PW includes the contents of `src/controller/tests/` but not the subdir `src/controller/tests/data_model/` which will have its own separate PR once I finish those.


src/test_driver/openiotsdk/unit-tests/test_components.txt
```diff
@@ -18,4 +18,5 @@ SupportTests
 UserDirectedCommissioningTests
 SecureChannelTests
 ICDServerTests
+ControllerTests
```
andy31415, line 21: previously it seems controllertests were not part of openiot (not sure why) ... since openiot fails, should we add them here?

If they cannot run on openiot, it would be nice to add that info to BUILD.gn

feasel0: Ok.  I've removed ControllerTests from test_components.txt

andy31415 (approved):

PR #33359, opened 2024-05-08
Unit tests for BLE layer class

arkq:
### Problem

There are no unit tests for functionality in BleLayer.cpp

### Changes

- Added unit tests for `BleLayer` class
- Returning `false` in case of error in all `Handle*` functions

### Testing

Manually verified linux-linux BLE commissioning after changes.


yunhanw-google (approved):
Thanks for the error return fix in Handle* including HandleWriteReceived

woody-apple (approved):

PR #33361, opened 2024-05-08
TC-IDM-10.2: Add test for provisional clusters

cecille:
Adds an explicit test that the provisional scenes cluster isn't included. Because this cluster is included in the SDK at the point where the TH was cut, we need a specific test because the later tests that check the cluster from the SDK impl won't fail. Also adding the content control cluster for completeness.

Once the scraper is updated, we can do this automatically from the DM XMLs, but for now, this is just a manually curated list.

Test plan: https://github.com/CHIP-Specifications/chip-test-plans/pull/4158


src/python_testing/TC_DeviceConformance.py
```diff
@@ -76,17 +76,23 @@ def record_warning(location, problem):
 
         success = True
         allow_provisional = self.user_params.get("allow_provisional", False)
+        # TODO: automate this once https://github.com/csa-data-model/projects/issues/454 is done.
+        provisional_cluster_ids = [0x050F, 0x0062]
```
andy31415, line 80:
```suggestion
        provisional_cluster_ids = [
            0x050F, # Content Control 
            0x0062, # Scenes Management
        ]
```

andy31415 (approved):

src/python_testing/TC_DeviceConformance.py
cecille: I added them from the objects file, since they're there.

src/python_testing/TC_DeviceConformance.py
```diff
@@ -76,17 +76,23 @@ def record_warning(location, problem):
 
         success = True
         allow_provisional = self.user_params.get("allow_provisional", False)
+        # TODO: automate this once https://github.com/csa-data-model/projects/issues/454 is done.
+        provisional_cluster_ids = [Clusters.ContentControl.id, Clusters.ScenesManagement.id]
```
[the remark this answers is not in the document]
cecille: sure. When I originally was doing this, I didn't think it would be a problem because those other clusters didn't have any backing implementation in the TH. Let me just check and see what landed in the sve branch and I'll add the provisional ones that can be added.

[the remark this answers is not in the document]
cecille: added. Checked on the 1.3-sve branch and that also looks OK.

ReneJosefsen (approved):

src/python_testing/TC_DeviceConformance.py
```diff
@@ -76,17 +76,24 @@ def record_warning(location, problem):
 
         success = True
         allow_provisional = self.user_params.get("allow_provisional", False)
+        # TODO: automate this once https://github.com/csa-data-model/projects/issues/454 is done.
+        provisional_cluster_ids = [Clusters.ContentControl.id, Clusters.ScenesManagement.id, Clusters.BallastConfiguration.id,
```
cjandhyala, line 80: you will need to add Dish washer Alarm cluster to this list

cecille: Cluster itself is not provisional, it just has provisional features. This will be captured by the conformance tests for the elements.

cjandhyala: it did not pass the rule of 3 platforms in SVE, hence this was requested to be marked as provisional and all cert tests are removed.

ReneJosefsen: The cluster is not marked as provisional in the spec? https://github.com/CHIP-Specifications/connectedhomeip-spec/blob/master/src/app_clusters/DishwasherAlarm.adoc

cecille: Discussed this on slack. Cluster is not provisional - there is one feature that can be certified. The provisional features will be caught by the element-level tests. 

Going to resolve this conversation now, since there is nothing to do here.

PR #33371, opened 2024-05-08
Converted unit tests in src/controller/tests/data_model from NL to PW.

feasel0:
I converted the unit tests in `src/controller/tests/data_model` to use PW.
(For the tests in the parent directory `src/controller/tests`, see [PR #33351](https://github.com/project-chip/connectedhomeip/pull/33351))

In the NL version the tests were executed in a different order than they were defined (whereas in PW they execute in the order of definition).  Normally this shouldn't matter, so I opted to leave the order of definition unchanged so as to reduce the code delta.  But there was one exception; one of the tests seems to be order-dependent.  `TestReadSubscribeAttributeResponseWithVersionOnlyCache` must execute before `TestReadSubscribeAttributeResponseWithCache` or else it fails one of its EXPECTs.  In particular this one:

`EXPECT_TRUE(version1.HasValue() && (version1.Value() == 0));`

`version1.Value()` returns `2` if this test is done after `TestReadSubscribeAttributeResponseWithCache` or `0` if this test is done before it.

I added a comment to the code noting this order-dependence, but we should eventually take a look at why this is happening.


src/controller/tests/data_model/TestCommands.cpp
```diff
@@ -180,25 +179,44 @@ InteractionModel::Status ServerClusterCommandExists(const ConcreteCommandPath &
...
-    static void TestFailure(nlTestSuite * apSuite, void * apContext);
-    static void TestMultipleFailures(nlTestSuite * apSuite, void * apContext);
-    static void TestSuccessNoDataResponseWithClusterStatus(nlTestSuite * apSuite, void * apContext);
-    static void TestFailureWithClusterStatus(nlTestSuite * apSuite, void * apContext);
-
-private:
+    // Performs shared setup for all tests in the test suite
+    static void SetUpTestSuite()
+    {
+        if (mpContext == nullptr)
```
andy31415, line 188: are these conditions needed here and in teardown as well?

src/controller/tests/data_model/TestCommands.cpp
```diff
@@ -180,25 +179,44 @@ InteractionModel::Status ServerClusterCommandExists(const ConcreteCommandPath &
...
+    }
+
+protected:
+    // Performs setup for each individual test in the test suite
+    void SetUp() { mpContext->SetUp(); }
+
+    // Performs teardown for each individual test in the test suite
+    void TearDown() { mpContext->TearDown(); }
+
+    static TestContext * mpContext;
```
andy31415, line 214:
```suggestion
    static TestContext * mpContext = nullptr;
```

feasel0: I thought non-const static members had to be initialized outside the class.

andy31415: You are right ... I missed the init below. Sorry for the noise.

andy31415 (approved):

PR #33372, opened 2024-05-08
Improve server logging on subscription failures.

bzbarsky-apple:
1) When logging "session inactive", log which session it is. 
2) When a subscription ReadHandler closes, log the relevant subscription ID and
   peer node.

### DIFF
--- a/src/messaging/ExchangeMgr.cpp
+++ b/src/messaging/ExchangeMgr.cpp
@@ -106,8 +106,13 @@ ExchangeContext * ExchangeManager::NewContext(const SessionHandle & session, Exc
 {
     if (!session->IsActiveSession())
     {
+#if CHIP_ERROR_LOGGING
         // Disallow creating exchange on an inactive session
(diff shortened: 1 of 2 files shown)


arkq (approved):

src/messaging/ExchangeMgr.cpp
```diff
@@ -106,8 +106,13 @@ ExchangeContext * ExchangeManager::NewContext(const SessionHandle & session, Exc
 {
     if (!session->IsActiveSession())
     {
+#if CHIP_ERROR_LOGGING
         // Disallow creating exchange on an inactive session
```
andy31415, line 110: nit: maybe comment should be moved up or tied to `return nullptr` ?

andy31415 (approved):

src/messaging/ExchangeMgr.cpp
bzbarsky-apple: Will do.

bzbarsky-apple: Done.

PR #33376, opened 2024-05-08
[Silabs] Fix window-app sync issue with 917 soc

jmartinez-silabs:
**Problem**
The use of chip stack locks in timers callback to handle data model reads and write for the window tilt and lift actions was causing synchronization issues on 917 soc. 

**Description changes**
The base of this PR is to remove all chip stack locks from the window app tilt and lift updates that are timer based. We pivot to fully utilizing the ScheduleWork API. We provide a callback function (worker) that will be run on the chip task (chip event queue) and therefore can access the data model directly.
The window app was already using that logic from some elements but not all.
Rework and regroup all Lift and Tilt update functions into 1 worker callback per action.
Create one `ScheduleControlAction` to schedule the identified, based on the passed arguments, worker on the chip event queue.

### DIFF
--- a/examples/window-app/silabs/src/WindowManager.cpp
+++ b/examples/window-app/silabs/src/WindowManager.cpp
@@ -273,20 +273,28 @@ void WindowManager::Cover::ScheduleControlAction(ControlAction action, bool setN
     VerifyOrReturn(action <= ControlAction::Tilt);
     // Allocate a CoverWorkData. It will be freed by the Worker callback
     CoverWorkData * data = new CoverWorkData(this, setNewTarget);
+    VerifyOrDie(data != nullptr);
 
     AsyncWorkFunct workFunct = (action == ControlAction::Lift) ? LiftUpdateWorker : TiltUpdateWorker;
-    chip::DeviceLayer::PlatformMgr().ScheduleWork(workFunct, reinterpret_cast<intptr_t>(data));
+    if (PlatformMgr().ScheduleWork(workFunct, reinterpret_cast<intptr_t>(data)) != CHIP_NO_ERROR)
+    {
+        ChipLogError(AppServer, "Failed to schedule cover control action");
+        delete data;
+    }
 }
 
 void WindowManager::Cover::LiftUpdateWorker(intptr_t arg)
 {
-    CoverWorkData * data = reinterpret_cast<CoverWorkData *>(arg);
-    Cover * cover        = data->cover;
+    // Use a unique_prt so it's freed when leaving this context
+    std::unique_ptr<CoverWorkData> data(reinterpret_cast<CoverWorkData *>(arg));
+    Cover * cover = data->cover;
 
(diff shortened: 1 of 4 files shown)


mkardous-silabs (approved):

examples/window-app/silabs/src/WindowManager.cpp
```diff
@@ -268,143 +268,136 @@ void WindowManager::Cover::Init(chip::EndpointId endpoint)
...
-    {
-        percent100ths = WC_PERCENT100THS_MIDDLE; // set at middle by default
-    }
+    VerifyOrReturn(action <= ControlAction::Tilt);
+    // Allocate a CoverWorkData. It will be freed by the Worker callback
+    CoverWorkData * data = new CoverWorkData(this, setNewTarget);
 
-    LiftSchedulePositionSet(percent100ths);
+    AsyncWorkFunct workFunct = (action == ControlAction::Lift) ? LiftUpdateWorker : TiltUpdateWorker;
+    chip::DeviceLayer::PlatformMgr().ScheduleWork(workFunct, reinterpret_cast<intptr_t>(data));
```
andy31415, line 278: nit: PlatformManager.h seems to say this returns a CHIP_ERROR.

Should we VerifyOrDie that this returns CHIP_NO_ERROR (otherwise we should log and make sure data is freed).

examples/window-app/silabs/src/WindowManager.cpp
```diff
@@ -268,143 +268,136 @@ void WindowManager::Cover::Init(chip::EndpointId endpoint)
...
 }
 
-void WindowManager::Cover::LiftUpdate(bool newTarget)
+void WindowManager::Cover::LiftUpdateWorker(intptr_t arg)
 {
-    NPercent100ths current, target;
+    CoverWorkData * data = reinterpret_cast<CoverWorkData *>(arg);
```
andy31415, line 283: would you consider 

```cpp
std::unique_ptr<CoverWorkData> data(reinterpret_cast<CoverWorkData *>(arg));
```

to not worry about calling `delete` at the right place?

examples/window-app/silabs/src/WindowManager.cpp
```diff
@@ -268,143 +268,136 @@ void WindowManager::Cover::Init(chip::EndpointId endpoint)
...
+    Cover * cover        = data->cover;
 
-    chip::DeviceLayer::PlatformMgr().LockChipStack();
+    NPercent100ths current, target;
 
-    Attributes::TargetPositionLiftPercent100ths::Get(mEndpoint, target);
-    Attributes::CurrentPositionLiftPercent100ths::Get(mEndpoint, current);
+    Attributes::TargetPositionLiftPercent100ths::Get(cover->mEndpoint, target);
```
andy31415, line 288: Do these getters return errors that we should/could handle?

andy31415 (approved):

PR #33382, opened 2024-05-09
[Infineon] Add a test trigger for CYW30739 to fill up event logging buffers.

hsusid:
#### Problem
The CYW30739 thermostat app doesn't have a way to trigger chunked event data, which is required by the TC-IDM-6.1 test.

#### Change overview
* Implement the TestEventTriggerDelegate class to provide a vendor trigger `kFillUpEventLoggingBuffer` as `0xffffffff13880000` (-3967287296) to fill up event logging buffers with HardwareFaultChange and SoftwareFault events.
* Implement the OnSoftwareFaultEventHandler method to build and log a SoftwareFault event.

#### Testing
* chip-tool generaldiagnostics test-event-trigger <EnableKey> -3967287296 <destination-id> 0
* chip-tool any read-event-by-id 0xFFFFFFFF 0xFFFFFFFF 2 0xFFFF

### DIFF
--- a/examples/platform/infineon/cyw30739/SoftwareDiagnostics.cpp
+++ b/examples/platform/infineon/cyw30739/SoftwareDiagnostics.cpp
@@ -39,15 +39,15 @@ void OnSoftwareFaultEventHandler(const char * faultRecordString)
 
     if (DeviceLayer::PlatformMgrImpl().IsCurrentTask())
     {
-        softwareFault.name.SetValue(CharSpan::fromCharString("Matter"));
+        softwareFault.name.SetValue("Matter"_span);
     }
     else if (DeviceLayer::ThreadStackMgrImpl().IsCurrentTask())
     {
-        softwareFault.name.SetValue(CharSpan::fromCharString("Thread"));
+        softwareFault.name.SetValue("Thread"_span);
     }
     else
     {
-        softwareFault.name.SetValue(CharSpan::fromCharString("App"));
+        softwareFault.name.SetValue("App"_span);
     }
 
     softwareFault.faultRecording.SetValue(ByteSpan(Uint8::from_const_char(faultRecordString), strlen(faultRecordString)));
--- a/examples/platform/infineon/cyw30739/SoftwareDiagnostics.cpp
+++ b/examples/platform/infineon/cyw30739/SoftwareDiagnostics.cpp
@@ -1,0 +1,62 @@
+/*
+ *
+ *    Copyright (c) 2024 Project CHIP Authors
+ *    All rights reserved.
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+#include "SoftwareDiagnostics.h"
+
+#include <app/clusters/software-diagnostics-server/software-diagnostics-server.h>
+#include <app/util/config.h>
+#include <platform/CHIPDeviceLayer.h>
+
+namespace chip {
+namespace DeviceLayer {
+namespace Infineon {
+namespace CYW30739 {
+
+using namespace chip::app::Clusters;
+
+void OnSoftwareFaultEventHandler(const char * faultRecordString)
+{
+#ifdef MATTER_DM_PLUGIN_SOFTWARE_DIAGNOSTICS_SERVER
+    SoftwareDiagnostics::Events::SoftwareFault::Type softwareFault;
+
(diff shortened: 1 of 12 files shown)


examples/platform/infineon/cyw30739/SoftwareDiagnostics.cpp
```diff
@@ -0,0 +1,62 @@
...
+{
+#ifdef MATTER_DM_PLUGIN_SOFTWARE_DIAGNOSTICS_SERVER
+    SoftwareDiagnostics::Events::SoftwareFault::Type softwareFault;
+
+    /* Unable to access thread ID in the application layer. */
+    softwareFault.id = 0;
+
+    if (DeviceLayer::PlatformMgrImpl().IsCurrentTask())
+    {
+        softwareFault.name.SetValue(CharSpan::fromCharString("Matter"));
```
andy31415, line 42: documentation seems to say:   
```
    // Note that for string literals, the user-defined `_span` string
    // literal operator should be used instead, e.g. `"Hello"_span`.
```

```suggestion
        softwareFault.name.SetValue("Matter"_span);
```

same for below

examples/platform/infineon/cyw30739/TestEventTriggerHandler.h
```diff
@@ -0,0 +1,42 @@
...
+
+namespace chip {
+namespace DeviceLayer {
+namespace Infineon {
+namespace CYW30739 {
+
+class TestEventTriggerHandler : public chip::TestEventTriggerHandler
+{
+public:
+    static constexpr uint64_t kFillUpEventLoggingBuffer = 0xffff'ffff'0000'0000;
```
andy31415, line 31: Is there a standard naming that we could use here? I could not directly find it, however https://github.com/project-chip/connectedhomeip/blob/master/examples/platform/linux/AppMain.cpp#L337 seems to suggest:

```
    /// NOTE: If you copy this for NON-STANDARD CLUSTERS OR USAGES, please use the reserved range FFFF_FFFF_<VID_HEX>_xxxx for your
    /// trigger codes. NOTE: Standard codes are <CLUSTER_ID_HEX>_xxxx_xxxx_xxxx.
```

so maybe `0xffff'ffff'1388'0000` ?

examples/platform/infineon/cyw30739/SoftwareDiagnostics.cpp
hsusid: Fixed. Thank you for the suggestion.

examples/platform/infineon/cyw30739/TestEventTriggerHandler.h
hsusid: Fixed. Thank you for the review.

examples/platform/infineon/cyw30739/matter_config.cpp
```diff
@@ -191,8 +192,10 @@ void CYW30739MatterConfig::InitApp(void)
     /* Start CHIP datamodel server */
     static chip::SimpleTestEventTriggerDelegate sTestEventTriggerDelegate{};
     static chip::OTATestEventTriggerHandler sOtaTestEventTriggerHandler{};
+    static Infineon::CYW30739::TestEventTriggerHandler sCYW30739TestEventTriggerHandler{};
     VerifyOrDie(sTestEventTriggerDelegate.Init(chip::ByteSpan(sTestEventTriggerEnableKey)) == CHIP_NO_ERROR);
```
mkardous-silabs, line 196: Just as an fyi, the `SimpleTestEventTriggerDelegate` isn't a spec compliant implementation of the TestEventTrigger.
The enable key should be in non-volatile storage to avoid needing to recompile the binary after certification since that is prohibited.

hsusid: Thank you for the information.
We will store the enable key as a field of factory data in another change.

examples/platform/infineon/cyw30739/TestEventTriggerHandler.h
```diff
@@ -0,0 +1,42 @@
...
+
+namespace chip {
+namespace DeviceLayer {
+namespace Infineon {
+namespace CYW30739 {
+
+class TestEventTriggerHandler : public chip::TestEventTriggerHandler
+{
+public:
+    static constexpr uint64_t kFillUpEventLoggingBuffer = 0xffff'ffff'1388'0000;
```
cecille, line 31: I'm definitely not blocking the PR on this, but I almost wonder if this could apply more generally. Other platforms may well find this useful, and the general diagnostics cluster is a REALLY good choice here - it's mandatory, and it looks like you're using a standard API.

andy31415 (approved):

examples/platform/infineon/cyw30739/TestEventTriggerHandler.h
hsusid: How about I create a class `GenericEventManagementTestEventTriggerHandler` in `src/app/GenericEventManagementTestEventTriggerHandler.[hc]` for a generic implementation to fill event logging buffers?

hsusid: I moved common trigger logic to the `GenericEventManagementTestEventTriggerHandler` class.

cecille: Thanks - this is great.

cecille (approved):

PR #33390, opened 2024-05-09
Converted unit tests in src/tracing/tests from NL to PW.

feasel0:
I converted the unit tests in `src/tracing/tests` to use PW.

These tests use the multiplexed version of tracing, so there are enabling conditions in `src/BUILD.gn` and `src/tracing/tests/BUILD.gn` which cause the tests to only run if tracing is enabled and multiplexed is the configuration.  However these conditions were looking for the wrong string -- they were looking for just the word "multiplexed" whereas now we are using full paths in the config string, so this has been fixed.


src/test_driver/openiotsdk/unit-tests/test_components.txt
```diff
@@ -13,6 +13,7 @@ PlatformTests
 RawTransportTests
 RetransmitTests
 TestShell
+TracingTests
```
andy31415, line 16: we did not have these in nl to start with, I think we can skip them here as well for now.

feasel0: Removed.

andy31415 (approved):

PR #33396, opened 2024-05-09
Ember compatibility layer - split out GlobalAttributeAccessInterface and the I/O buffer

andy31415:
These are code sharing changes to support future development for a ember-backed data provider, implementing the interface defined in #32914 

### Changes

- Moved `MandatoryGlobalAttributeReader` and `GlobalAttributeReader` into a separate cpp/header: `ember-global-attribute-access-interface` since these are ember-backed
- Moved the storage buffer named `attributeData` into `ember-io-storage.h/cpp` and exposed it as a `MutableByteSpan` (so have access to size, however no sizeof usage anymore ... this loses some compile-time asserts, however that should be safe as the minimum size of this span is 8 and compile-time asserts were previously done for numeric types only)
- Moved `BaseType` function into ember-io-storage as well and renamed to `AttributeBaseType` with documentation to make things clearer about usage
- Added/Moved `privilege-storagec.pp` into ember mocking as this seems to be very tightly coupled with ember in general. Removed the corresponding stubs from integration tests (as they can now use the mocks)
- added a few methods to mock attribute storage - items that were missing but implemented and semi-public api (at least as ember-compatibility functions is concerned):
  - `emberAfLocateAttributeMetadata`
  - `emberAfFindClusterInType`

For mocks, also provide a ATTRIBUTE_LARGEST which I initialized to the all-clusters-app size (1003).


src/app/util/ember-io-storage.h
```diff
@@ -0,0 +1,46 @@
...
+#include <app/util/attribute-metadata.h>
+#include <lib/support/Span.h>
+
+namespace chip {
+namespace app {
+namespace Compatibility {
+namespace Internal {
+
+/// A buffer guaranteed to be sized to contain any individual value from
+/// ember (i.e. a buffer that can be used to read ember data into).
```
bzbarsky-apple, line 30: What does "ember data" mean here?  In practice, this is sized to be able to fit any attribute value that's handled via the ember data store (including external attributes handled via the external callbacks).

src/app/util/ember-io-storage.h
```diff
@@ -0,0 +1,46 @@
...
+extern MutableByteSpan gEmberAttributeIOBufferSpan;
+
+/// Maps a generic attribute to a basic int(8|16|32|64)(s|u) type
```
bzbarsky-apple, line 33:
```suggestion
/// Maps an attribute type that is not an integer but can be represented as an integer to the corresponding basic int(8|16|32|64)(s|u) type
```

src/app/util/ember-io-storage.h
```diff
@@ -0,0 +1,46 @@
...
+///
+/// For example:
+///    ZCL_ENUM8_ATTRIBUTE_TYPE maps to ZCL_INT8U_ATTRIBUTE_TYPE
+///    ZCL_VENDOR_ID_ATTRIBUTE_TYPE maps to ZCL_INT16U_ATTRIBUTE_TYPE
+///    ZCL_BITMAP32_ATTRIBUTE_TYPE maps to ZCL_INT32U_ATTRIBUTE_TYPE
+///    ZCL_VOLTAGE_MV_ATTRIBUTE_TYPE maps to ZCL_INT64S_ATTRIBUTE_TYPE
+///    ...
```
bzbarsky-apple, line 40: Should mention that types that are not "could be represented as integer" are left as-is, yes?

src/app/util/ember-io-storage.h
andy31415: I expanded the text a bit. PTAL

yufengwangca (approved):

bzbarsky-apple (approved):

PR #33398, opened 2024-05-09
[Linux] Add test logic to verify the subscription state in the Linux tv-casting-app.

### LINKED ISSUE #33397: [Linux] Add test logic to verify the subscription state in the Linux tv-casting-app.
Continuing on from #32787, #32918, #33300, and #33343.

There is no CI check to verify that the Linux tv-casting-app is still correctly subscribing to the `MediaPlayback` Cluster and `CurrentState` Attribute whenever a PR is created. Therefore, this issue will add the logic needed to the `run_tv_casting_test.py` file to check the subscription state of the tv-casting-app by finding a `ReportDataMessage` block with the relevant cluster and attribute values.

Essentially, we will be parsing the output of the Linux tv-casting-app for the relevant `ReportDataMessage` block and confirm that the tv-casting-app is subscribed to the `MediaPlayback` cluster and the `CurrentState` (playback state) attribute. If the relevant `ReportDataMessage` block is not found within a given timeout, then we will log the error, dump the logs to the console, and exit on error.

shaoltan-amazon:
Fixes #33397 

**Problem**
There is currently no CI check to verify that the Linux tv-casting-app is still subscribed to the `MediaPlayback` cluster and the `CurrentState` (playback state) attribute whenever a PR is created.

**Solution Overview**
Add test logic to the `run_tv_casting_test.py` file to verify the subscription state of the Linux tv-casting-app. The output will be written to the tv-casting-app log file.

The following has been added to the test script:
- Parse the Linux tv-casting-app output for `ReportDataMessage` and validate the `Cluster` and `Attribute` values against the `MediaPlayback` cluster and `CurrentState` attribute. The cluster and attribute values are `0x506` and `0x0000_0000` respectively. For reference, these values are set [here](https://github.com/project-chip/connectedhomeip/blob/550b1c13608c2bbcbaca6e4375b19997cec4418b/examples/tv-app/tv-common/tv-app.zap#L5190-L5193) for the cluster value and [here](https://github.com/project-chip/connectedhomeip/blob/master/examples/tv-app/tv-common/tv-app.zap#L5320-L5321) for the attribute value.
- Log the first instance of the `ReportDataMessage` with valid cluster and attribute values to console.
- If we are not able to find a `ReportDataMessage` with valid cluster and attribute values within the timeout, then we print an error message, dump the logs to console, and exit on error.
- Also cleaned up logging messages to improve readability and maintain uniformity.

**Testing**
Tested the script locally by running `python3 ./scripts/tests/run_tv_casting_test.py` and observing the output. Also verified that the CI check works as expected.

### DIFF
--- a/scripts/tests/run_tv_casting_test.py
+++ b/scripts/tests/run_tv_casting_test.py
@@ -36,6 +36,9 @@ COMMISSIONING_STAGE_MAX_WAIT_SEC = 10
 # The maximum amount of time to test that the launchURL is sent from the Linux tv-casting-app and received on the tv-app before timeout.
 TEST_LAUNCHURL_MAX_WAIT_SEC = 10
 
+# The maximum amount of time to verify the subscription state in the Linux tv-casting-app output before timeout.
+VERIFY_SUBSCRIPTION_STATE_MAX_WAIT_SEC = 10
+
 # File names of logs for the Linux tv-casting-app and the Linux tv-app.
 LINUX_TV_APP_LOGS = 'Linux-tv-app-logs.txt'
 LINUX_TV_CASTING_APP_LOGS = 'Linux-tv-casting-app-logs.txt'
@@ -45,6 +48,10 @@ LINUX_TV_CASTING_APP_LOGS = 'Linux-tv-casting-app-logs.txt'
 VENDOR_ID = 0xFFF1   # Spec 7.20.2.1 MEI code: test vendor IDs are 0xFFF1 to 0xFFF4
 PRODUCT_ID = 0x8001  # Test product id
 DEVICE_TYPE_CASTING_VIDEO_PLAYER = 0x23    # Device type library 10.3: Casting Video Player
+
+# Values to verify the subscription state against from the `ReportDataMessage` in the Linux tv-casting-app output.
+CLUSTER_MEDIA_PLAYBACK = '0x506'  # Application Cluster Spec 6.10.3 Cluster ID: Media Playback
+ATTRIBUTE_CURRENT_PLAYBACK_STATE = '0x0000_0000'  # Application Cluster Spec 6.10.6 Attribute ID: Current State of Playback
 
 
 class ProcessManager:
@@ -96,13 +103,19 @@ def extract_value_from_string(line: str) -> str:
 
     The string is expected to be in the following format as it is received
     from the Linux tv-casting-app output:
+    \x1b[0;34m[1715206773402] [20056:2842184] [DMG]  Cluster = 0x506,\x1b[0m
+    The substring to be extracted here is '0x506'.
+    Or:
     \x1b[0;34m[1713741926895] [7276:9521344] [DIS] Vendor ID: 65521\x1b[0m
     The integer value to be extracted here is 65521.
     Or:
     \x1b[0;34m[1714583616179] [7029:2386956] [SVR] 	device Name: Test TV casting app\x1b[0m
     The substring to be extracted here is 'Test TV casting app'.
     """
-    value = line.split(':')[-1].strip().replace('\x1b[0m', '')
+    if '=' in line:
+        value = line.split('=')[-1].strip().replace(',\x1b[0m', '')
+    else:
+        value = line.split(':')[-1].strip().replace('\x1b[0m', '')
 
     return value
 
@@ -211,7 +224,8 @@ def validate_identification_declaration_message_on_tv_app(tv_app_info: Tuple[sub
     while True:
         # Check if we exceeded the maximum wait time for validating the device information from the Linux tv-app to the corresponding values from the Linux tv-app.
         if time.time() - start_wait_time > COMMISSIONING_STAGE_MAX_WAIT_SEC:
-            logging.erro('The device information from the Linux tv-app output was not validated against the corresponding values from the Linux tv-casting-app output within the timeout.')
+            logging.error(
+                'The device information from the Linux tv-app output was not validated against the corresponding values from the Linux tv-casting-app output within the timeout.')
             return False
 
         tv_app_line = tv_app_process.stdout.readline()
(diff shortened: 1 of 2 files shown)


scripts/tests/run_tv_casting_test.py
```diff
@@ -310,6 +323,56 @@ def validate_commissioning_success(tv_casting_app_info: Tuple[subprocess.Popen,
...
+    continue_parsing = False
+    report_data_message = []
+
+    start_wait_time = time.time()
+
+    while True:
+        # Check if we exceeded the maximum wait time to parse the Linux tv-casting-app output for `ReportDataMessage` block.
+        if time.time() - start_wait_time > VERIFY_SUBSCRIPTION_STATE_MAX_WAIT_SEC:
+            logging.error(
+                'The relevant `ReportDataMessage` block was not found in the Linux tv-casting-app process within the timeout.')
```
sharadb-amazon, line 340:
```suggestion
                'The relevant `ReportDataMessage` block for the MediaPlayback:CurrentState subscription was not found in the Linux tv-casting-app process within the timeout.')
```

scripts/tests/run_tv_casting_test.py
```diff
@@ -96,13 +103,19 @@ def extract_value_from_string(line: str) -> str:
...
     The integer value to be extracted here is 65521.
     Or:
     \x1b[0;34m[1714583616179] [7029:2386956] [SVR] 	device Name: Test TV casting app\x1b[0m
     The substring to be extracted here is 'Test TV casting app'.
     """
-    value = line.split(':')[-1].strip().replace('\x1b[0m', '')
+    if '=' in line:
+        value = line.split('=')[-1].strip().replace(',\x1b[0m', '')
+    else:
+        value = line.split(':')[-1].strip().replace('\x1b[0m', '')
```
sharadb-amazon, line 118: It probably makes sense to check if we found ":" in line before trying to split by it and then add an else case where we did not find either separator, which should be considered an error case.

sharadb-amazon (approved):

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Addressed in latest commit.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Addressed in latest commit.

chrisdecenzo (approved):

scripts/tests/run_tv_casting_test.py
```diff
@@ -80,36 +87,53 @@ def dump_temporary_logs_to_console(log_file_path: str):
...
+    Or:
     \x1b[0;34m[1713741926895] [7276:9521344] [DIS] Vendor ID: 65521\x1b[0m
-    The integer value to be extracted here is 65521.
+    The integer value to be extracted here is '65521'.
     Or:
     \x1b[0;34m[1714583616179] [7029:2386956] [SVR] 	device Name: Test TV casting app\x1b[0m
     The substring to be extracted here is 'Test TV casting app'.
     """
-    value = line.split(':')[-1].strip().replace('\x1b[0m', '')
+    if ':' in line:
```
andy31415, line 118: This seems odd a bit to me: the example we have a above is `\x1b[0;34m[1715206773402] [20056:2842184] [DMG] Cluster = 0x506,\x1b[0m` ... which does match the `:` but it looks like by chance since the log line staring with `[DMG]` does not actually contain the ':'. Same comment on the `#`.

It looks to me like everything seems to match.
Maybe we should have a different parsing logic. We seem to maybe search for `if value_name in line` as being a more accurate search.

andy31415: I wonder if we should have some form of regex given the format that I see: 
  - some prefix (like time, color changing)
  - ` \[[A-Z]{1,3}\]` which represents the logging source. The 1-3 characters is guaranteed by https://github.com/project-chip/connectedhomeip/blob/master/src/lib/support/logging/TextOnlyLogging.cpp#L111
   - then the log with an optional suffix of `\x1b[0m`
   Seems maybe valuable to extract the log text from a formatted log line, and then processing the log text by extracting `#,=,:`

scripts/tests/run_tv_casting_test.py
```diff
@@ -187,12 +213,12 @@ def extract_device_info_from_tv_casting_app(tv_casting_app_info: Tuple[subproces
         linux_tv_casting_app_log_file.flush()
 
         if 'device Name' in line:
-            device_name = extract_value_from_string(line)
+            device_name = extract_value_from_string(line, 'device Name', casting_state, log_paths)
```
andy31415, line 216: this repeats `device Name` in two places. I wonder if this should become:

```py
if value := extract_value_from_log_line(line, 'device name', casting_state, log_paths):
    device_name = value
```

and same below.
I also wonder about passing casting state and log paths every single time. It seems it would be more natural to have a `log_value_extractor = LogValueExtractor(casting_state, log_paths)` and then

```py
if value := log_value_extractor.extract_from(line, 'device Name'):
    device_name = value
```

Judgement call though as I am afraid I ask for too many changes. Could be a followup.

scripts/tests/run_tv_casting_test.py
```diff
@@ -80,36 +87,53 @@ def dump_temporary_logs_to_console(log_file_path: str):
...
+            delimiter = '='
+        elif '#' in line:
+            delimiter = '#'
+        else:
+            delimiter = ':'
+
+        value = line.split(delimiter)[-1].strip().replace('\x1b[0m', '').rstrip(',')
+    else:
+        logging.error(f'Could not extract {value_name} from the following line: {line}')
+        handle_casting_failure(casting_state, log_paths)
```
andy31415, line 129: how about throwing an exception and letting the caller handle it? Passing in casting_state and log_paths just for failure handling seems inconvenient - many args which are not relevant to the functioning of this function.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Good idea! Addressed in the latest commit.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Great suggestion! Addressed in the latest commit.

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Good call. Addressed in the latest commit.

hawk248 (approved):

PR #33399, opened 2024-05-10
Darwin: promote Cooking Appliances to Introduced

kiel-apple:
Continuing the theme of v1.3 promotions.  Added comments about `WATTS` and affected provisional bits.


src/darwin/Framework/CHIP/templates/availability.yaml
```diff
@@ -8823,6 +8827,52 @@
...
               - FeatureMap
               - ClusterRevision
+          MicrowaveOvenControl:
+              - CookTime
+              - MaxCookTime
+              - PowerSetting
+              - MinPower
+              - MaxPower
+              - PowerStep
+              # SupportedWatts and SelectedWattIndex are Provisional in Matter 1.3
```
bzbarsky-apple, line 8837: Then they need explicit provisional annotations, right?

kiel-apple: Yep - is [this](https://github.com/project-chip/connectedhomeip/pull/33399/files#diff-94ddfb24d1680d06578d1e18f05664162d7c7e9f05958813917cde261c7e9541R9765-R9767) correct?

bzbarsky-apple: Generally, yes, but I would add a comment explaining why those are marked provisional.

andy31415:
@kiel-apple this PR seems quite old and has merge conflicts. Is it supposed to go in?

kiel-apple:
@andy31415 let me take a stab at bringing it back up to date today, I think it won't be awful.

src/darwin/Framework/CHIP/templates/availability.yaml
```diff
@@ -9685,13 +9685,156 @@
...
+              - GeneratedCommandList
+              - AcceptedCommandList
+              - AttributeList
+              - FeatureMap
+              - ClusterRevision
+          MicrowaveOvenMode:
+              - SupportedModes
+              - CurrentMode
+              - StartUpMode
+              - OnMode
```
bzbarsky-apple, line 9713: These are marked X in the spec, no?  We shouldn't have them here.

src/darwin/Framework/CHIP/templates/availability.yaml
```diff
@@ -9685,13 +9685,156 @@
...
+              - GeneratedCommandList
+              - AcceptedCommandList
+              - AttributeList
+              - FeatureMap
+              - ClusterRevision
+          OvenMode:
+              - SupportedModes
+              - CurrentMode
+              - StartUpMode
+              - OnMode
```
bzbarsky-apple, line 9723: Again, StartUpMode and OnMode are X in the spec for Oven Mode.

src/darwin/Framework/CHIP/templates/availability.yaml
```diff
@@ -9685,13 +9685,156 @@
...
+              - AttributeList
+              - FeatureMap
+              - ClusterRevision
+      commands:
+          MicrowaveOvenControl:
+              - SetCookingParameters
+              - AddMoreTime
+          MicrowaveOvenMode:
+              - ChangeToMode
+              - ChangeToModeResponse
```
bzbarsky-apple, line 9747: These are set as X in the spec, no?

src/darwin/Framework/CHIP/templates/availability.yaml
```diff
@@ -9685,13 +9685,156 @@
...
+          OvenMode:
+              - ChangeToMode
+              - ChangeToModeResponse
+          OvenCavityOperationalState:
+              - Pause
+              - Stop
+              - Start
+              - Resume
```
bzbarsky-apple, line 9755: Pause and Resume are X in the spec.

src/darwin/Framework/CHIP/templates/availability.yaml
```diff
@@ -9742,6 +9918,18 @@
...
                   - UpdateStartHue
+          MicrowaveOvenControl:
+              Feature:
+                  - PowerAsNumber
+                  # PowerInWatts is provisional because WATTS is provisional in Matter 1.3
+                  # WATTS is provisional in 1.3 because no one has submitted a DUT for it
+                  - PowerNumberLimits
+          MicrowaveOvenMode:
+              Feature:
+                  - OnOff
```
bzbarsky-apple, line 9929: This feature is disallowed for this cluster, no?

src/darwin/Framework/CHIP/templates/availability.yaml
```diff
@@ -9742,6 +9918,18 @@
...
+          OvenMode:
+              Feature:
+                  - OnOff
```
bzbarsky-apple, line 9932: Likewise, this is disallowed....

bzbarsky-apple (changes requested):
I would recommend removing the provisional annotations for things being added; that would also make it clearer what's still marked as provisional.

kiel-apple:
@bzbarsky-apple thanks for checking the spec, I could have done better there given how long it has been since I opened this

kiel-apple:
> I would recommend removing the provisional annotations for things being added; that would also make it clearer what's still marked as provisional.

I still need to go back and do this.  Coming soon!

kiel-apple:
need to move all these entries forward to new Future since Future from back when this started has been released

src/darwin/Framework/CHIP/templates/availability.yaml
```diff
@@ -9689,13 +9685,144 @@
       watchOS: "11.2"
       tvOS: "18.2"
   introduced:
+      clusters:
+          - MicrowaveOvenControl
+          - MicrowaveOvenMode
```
bzbarsky-apple, line 9690: So... now when I run codegen with this PR as of right now (rev cd2fc6b3f63aeeb051a6b99b461427609f95fc08) I see a bunch of warnings about thigns that are not annotated as either provisional, removed, or introduced, right?

woody-apple (approved):

src/darwin/Framework/CHIP/templates/availability.yaml
```diff
@@ -10084,9 +10080,245 @@
...
   provisional:
       device types:
           - BatteryStorage
           - HeatPump
           - SolarPower
           - WaterHeater
+      attributes:
+          MicrowaveOvenControl:
+              - SupportedWatts
+              - SelectedWattIndex
```
bzbarsky-apple, line 10293: The comment about why these are provisional should be here.

src/darwin/Framework/CHIP/templates/availability.yaml
```diff
@@ -10084,9 +10080,245 @@
...
+          - OvenCavityOperationalState
+      attributes:
+          MicrowaveOvenControl:
+              - CookTime
+              - MaxCookTime
+              - PowerSetting
+              - MinPower
+              - MaxPower
+              - PowerStep
+              # SupportedWatts and SelectedWattIndex are Provisional because WATTS is Provisional in Matter 1.3
```
bzbarsky-apple, line 10097: We're looking at Matter 1.4 now, no?  Please fix comments to reflect 1.4 status?

src/darwin/Framework/CHIP/templates/availability.yaml
```diff
@@ -10084,9 +10080,245 @@
...
+              - AttributeList
+              - FeatureMap
+              - ClusterRevision
+      commands:
+          MicrowaveOvenControl:
+              - SetCookingParameters
+              - AddMoreTime
+          MicrowaveOvenMode:
+              - ChangeToMode
+              - ChangeToModeResponse
```
bzbarsky-apple, line 10139: These are X in the spec, and presumably do not exist?  Why are they here?

src/darwin/Framework/CHIP/templates/availability.yaml
```diff
@@ -10084,9 +10080,245 @@
...
+                  - wattSettingIndex
+      bitmap values:
+          MicrowaveOvenControl:
+              Feature:
+                  - PowerInWatts
+  removed:
+      attributes:
+          OvenMode:
+              - StartUpMode
+              - OnMode
```
bzbarsky-apple, line 10306: Please document why these are "removed" even though they are still in the XML?

And please file an issue to get the XML fixed.  Same for OvenCavityOperationalState.

src/darwin/Framework/CHIP/templates/availability.yaml
```diff
@@ -10084,9 +10080,245 @@
...
+              - Pause
+              - Resume
+      bitmaps:
+          # MicrowaveOvenMode and OvenMode Feature bitmaps removed because
+          # an NS_ENUM must have at least one value and without ONOFF they have none
+          MicrowaveOvenMode:
+              - Feature
+          OvenMode:
+              - Feature
+      bitmap values:
```
bzbarsky-apple, line 10318: Why do you need these?  if the bitmap is removed, then the bits inside it are removed automatically....  I guess the bitmap values removals got added first to implement the X from the spec, and then that failed to compile and the bitmap removals above were added?

In any case, either just remove the bitmaps or clearly document why the redundant bits are here (e.g. so that if the bitmaps ever get un-removed due to growing new bits we will not get incorrect bits in them).

src/darwin/Framework/CHIP/templates/availability.yaml
kiel-apple: added.

kiel-apple:
incorporated into #36793

PR #33402, opened 2024-05-10
Operational Discovery with Continuous Query Support for Linux 

su-shanka:
Added operational node discovery/browse support to linux platform mdns implementation. Also, enhanced browse implementation in linux mdns to support continuous query browse instead single shot query.

This is a follow up from #32750 for linux platform.

### DIFF
--- a/src/lib/dnssd/Discovery_ImplPlatform.cpp
+++ b/src/lib/dnssd/Discovery_ImplPlatform.cpp
@@ -93,12 +93,11 @@ static void HandleNodeBrowse(void * context, DnssdService * services, size_t ser
         discoveryContext->Retain();
         // For some platforms browsed services are already resolved, so verify if resolve is really needed or call resolve callback
 
-        auto & ipAddress         = services[i].mAddress;
+        auto & ipAddress = services[i].mAddress;
 
         // mType(service name) exactly matches with operational service name
         bool isOperationalBrowse = (strncmp(services[i].mType, kOperationalServiceName, sizeof(services[i].mType)) == 0 &&
                                     strlen(services[i].mType) == strlen(kOperationalServiceName));
-
 
         // For operational browse result we currently don't need IP address hence skip resolution and handle differently.
         if (isOperationalBrowse)
--- a/src/lib/dnssd/Discovery_ImplPlatform.cpp
+++ b/src/lib/dnssd/Discovery_ImplPlatform.cpp
@@ -96,15 +96,14 @@ static void HandleNodeBrowse(void * context, DnssdService * services, size_t ser
         auto & ipAddress = services[i].mAddress;
 
         // mType(service name) exactly matches with operational service name
-        bool isOperationalBrowse = (strncmp(services[i].mType, kOperationalServiceName, sizeof(services[i].mType)) == 0 &&
-                                    strlen(services[i].mType) == strlen(kOperationalServiceName));
+        bool isOperationalBrowse = strcmp(services[i].mType, kOperationalServiceName) == 0;
 
         // For operational browse result we currently don't need IP address hence skip resolution and handle differently.
         if (isOperationalBrowse)
         {
             HandleNodeOperationalBrowse(context, &services[i], error);
         }
-        // if SRV, TXT and AAAA records were received in DNS responses
+        // check whether SRV, TXT and AAAA records were received in DNS responses
         else if (strlen(services[i].mHostName) == 0 || services[i].mTextEntrySize == 0 || !ipAddress.has_value())
         {
             ChipDnssdResolve(&services[i], services[i].mInterface, HandleNodeResolve, context);
--- a/src/lib/dnssd/Discovery_ImplPlatform.cpp
+++ b/src/lib/dnssd/Discovery_ImplPlatform.cpp
@@ -50,7 +50,17 @@ static void HandleNodeResolve(void * context, DnssdService * result, const Span<
     }
 
     DiscoveredNodeData nodeData;
-    result->ToDiscoveredNodeData(addresses, nodeData);
+
+    if (0 == strncmp(result->mType, kOperationalServiceName, sizeof(kOperationalServiceName)))
+    {
+        result->ToDiscoveredOperationalNodeBrowseData(nodeData);
+
+        nodeData.Get<OperationalNodeBrowseData>().LogDetail();
+        discoveryContext->OnNodeDiscovered(nodeData);
+        return;
+    }
+
+    result->ToDiscoveredCommissionNodeData(addresses, nodeData);
 
     nodeData.Get<CommissionNodeData>().LogDetail();
     discoveryContext->OnNodeDiscovered(nodeData);
@@ -75,8 +85,9 @@ static void HandleNodeBrowse(void * context, DnssdService * services, size_t ser
 
         auto & ipAddress = services[i].mAddress;
 
-        // Check if SRV, TXT and AAAA records were received in DNS responses
-        if (strlen(services[i].mHostName) == 0 || services[i].mTextEntrySize == 0 || !ipAddress.has_value())
+        // Check that this is not operational browse result and if SRV, TXT and AAAA records were received in DNS responses
+        if (strncmp(services[i].mType, kOperationalServiceName, sizeof(kOperationalServiceName)) != 0
+            && (strlen(services[i].mHostName) == 0 || services[i].mTextEntrySize == 0 || !ipAddress.has_value()))
         {
             ChipDnssdResolve(&services[i], services[i].mInterface, HandleNodeResolve, context);
         }
@@ -340,7 +351,15 @@ void DiscoveryImplPlatform::HandleNodeIdResolve(void * context, DnssdService * r
     impl->mOperationalDelegate->OnOperationalNodeResolved(nodeData);
 }
 
-void DnssdService::ToDiscoveredNodeData(const Span<Inet::IPAddress> & addresses, DiscoveredNodeData & nodeData)
+void DnssdService::ToDiscoveredOperationalNodeBrowseData(DiscoveredNodeData & nodeData)
+{
+    nodeData.Set<OperationalNodeBrowseData>();
+
+    ExtractIdFromInstanceName(mName, &nodeData.Get<OperationalNodeBrowseData>().peerId);
+    nodeData.Get<OperationalNodeBrowseData>().hasZeroTTL = (mTtlSeconds == 0);
+}
+
+void DnssdService::ToDiscoveredCommissionNodeData(const Span<Inet::IPAddress> & addresses, DiscoveredNodeData & nodeData)
 {
     nodeData.Set<CommissionNodeData>();
     auto & discoveredData = nodeData.Get<CommissionNodeData>();
@@ -746,6 +765,31 @@ CHIP_ERROR DiscoveryImplPlatform::DiscoverCommissioners(DiscoveryFilter filter, 
     return error;
 }
 
+CHIP_ERROR DiscoveryImplPlatform::DiscoverOperational(DiscoveryFilter filter, DiscoveryContext & context)
+{
+    ReturnErrorOnFailure(InitImpl());
+    StopDiscovery(context);
+
+    char serviceName[kMaxCommissionerServiceNameSize];
+    ReturnErrorOnFailure(MakeServiceTypeName(serviceName, sizeof(serviceName), filter, DiscoveryType::kOperational));
+
+    intptr_t browseIdentifier;
+    // Increase the reference count of the context to keep it alive until HandleNodeBrowse is called back.
+    CHIP_ERROR error = ChipDnssdBrowse(serviceName, DnssdServiceProtocol::kDnssdProtocolTcp, Inet::IPAddressType::kAny,
+                                       Inet::InterfaceId::Null(), HandleNodeBrowse, context.Retain(), &browseIdentifier);
+
+    if (error == CHIP_NO_ERROR)
+    {
+        context.SetBrowseIdentifier(browseIdentifier);
+    }
+    else
+    {
+        context.Release();
+    }
+
+    return error;
+}
+
 CHIP_ERROR DiscoveryImplPlatform::StartDiscovery(DiscoveryType type, DiscoveryFilter filter, DiscoveryContext & context)
 {
     switch (type)
@@ -755,7 +799,7 @@ CHIP_ERROR DiscoveryImplPlatform::StartDiscovery(DiscoveryType type, DiscoveryFi
     case DiscoveryType::kCommissionerNode:
         return DiscoverCommissioners(filter, context);
     case DiscoveryType::kOperational:
-        return CHIP_ERROR_NOT_IMPLEMENTED;
+        return DiscoverOperational(filter, context);
     default:
         return CHIP_ERROR_INVALID_ARGUMENT;
     }
--- a/src/lib/dnssd/Discovery_ImplPlatform.cpp
+++ b/src/lib/dnssd/Discovery_ImplPlatform.cpp
@@ -99,7 +99,7 @@ static void HandleNodeBrowse(void * context, DnssdService * services, size_t ser
         bool isOperationalBrowse = (strncmp(services[i].mType, kOperationalServiceName, sizeof(services[i].mType)) == 0 &&
                                     strlen(services[i].mType) == strlen(kOperationalServiceName));
 
-        
+
         // For operational browse result we currently don't need IP address hence skip resolution and handle differently.
         if (isOperationalBrowse)
         {
--- a/src/lib/dnssd/Discovery_ImplPlatform.cpp
+++ b/src/lib/dnssd/Discovery_ImplPlatform.cpp
@@ -788,7 +788,7 @@ CHIP_ERROR DiscoveryImplPlatform::DiscoverOperational(DiscoveryFilter filter, Di
     ReturnErrorOnFailure(InitImpl());
     StopDiscovery(context);
 
-    char serviceName[kMaxCommissionerServiceNameSize];
+    char serviceName[kMaxOperationalServiceNameSize];
     ReturnErrorOnFailure(MakeServiceTypeName(serviceName, sizeof(serviceName), filter, DiscoveryType::kOperational));
 
     intptr_t browseIdentifier;
--- a/src/lib/dnssd/Discovery_ImplPlatform.cpp
+++ b/src/lib/dnssd/Discovery_ImplPlatform.cpp
@@ -86,8 +86,8 @@ static void HandleNodeBrowse(void * context, DnssdService * services, size_t ser
         auto & ipAddress = services[i].mAddress;
 
         // Check that this is not operational browse result and if SRV, TXT and AAAA records were received in DNS responses
-        if (strncmp(services[i].mType, kOperationalServiceName, sizeof(kOperationalServiceName)) != 0
-            && (strlen(services[i].mHostName) == 0 || services[i].mTextEntrySize == 0 || !ipAddress.has_value()))
+        if (strncmp(services[i].mType, kOperationalServiceName, sizeof(kOperationalServiceName)) != 0 &&
+            (strlen(services[i].mHostName) == 0 || services[i].mTextEntrySize == 0 || !ipAddress.has_value()))
         {
             ChipDnssdResolve(&services[i], services[i].mInterface, HandleNodeResolve, context);
         }
--- a/src/lib/dnssd/Discovery_ImplPlatform.cpp
+++ b/src/lib/dnssd/Discovery_ImplPlatform.cpp
@@ -51,18 +51,28 @@ static void HandleNodeResolve(void * context, DnssdService * result, const Span<
 
     DiscoveredNodeData nodeData;
 
-    if (0 == strncmp(result->mType, kOperationalServiceName, sizeof(kOperationalServiceName)))
-    {
-        result->ToDiscoveredOperationalNodeBrowseData(nodeData);
-
-        nodeData.Get<OperationalNodeBrowseData>().LogDetail();
-        discoveryContext->OnNodeDiscovered(nodeData);
-        return;
-    }
-
     result->ToDiscoveredCommissionNodeData(addresses, nodeData);
 
     nodeData.Get<CommissionNodeData>().LogDetail();
+    discoveryContext->OnNodeDiscovered(nodeData);
+    discoveryContext->Release();
+}
+
+static void HandleNodeOperationalBrowse(void * context, DnssdService * result, CHIP_ERROR error)
+{
+    DiscoveryContext * discoveryContext = static_cast<DiscoveryContext *>(context);
+
+    if (error != CHIP_NO_ERROR)
+    {
+        discoveryContext->Release();
+        return;
+    }
+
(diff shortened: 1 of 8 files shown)


src/lib/shell/commands/Dns.cpp
```diff
@@ -290,13 +297,14 @@ void RegisterDnsCommands()
         { &BrowseCommissionerHandler, "commissioner",
           "Browse Matter commissioner nodes. Usage: dns browse commissioner [subtype]" },
         { &BrowseOperationalHandler, "operational", "Browse Matter operational nodes. Usage: dns browse operational" },
+        { &BrowseStopHandler, "stop", "Stop ongoing browse. USage: dns browse stop" },
```
Damian-Nordic, line 300:
```suggestion
        { &BrowseStopHandler, "stop", "Stop ongoing browse. Usage: dns browse stop" },
```

src/lib/dnssd/Discovery_ImplPlatform.cpp
```diff
@@ -75,8 +85,9 @@ static void HandleNodeBrowse(void * context, DnssdService * services, size_t ser
 
         auto & ipAddress = services[i].mAddress;
 
-        // Check if SRV, TXT and AAAA records were received in DNS responses
-        if (strlen(services[i].mHostName) == 0 || services[i].mTextEntrySize == 0 || !ipAddress.has_value())
+        // Check that this is not operational browse result and if SRV, TXT and AAAA records were received in DNS responses
+        if (strncmp(services[i].mType, kOperationalServiceName, sizeof(kOperationalServiceName)) != 0 &&
```
Damian-Nordic, line 89: This may cause that null `ipAddress` is dereferenced if the address has not been returned for the browse query.

src/lib/dnssd/Discovery_ImplPlatform.cpp
```diff
@@ -50,7 +50,17 @@ static void HandleNodeResolve(void * context, DnssdService * result, const Span<
...
 
     DiscoveredNodeData nodeData;
-    result->ToDiscoveredNodeData(addresses, nodeData);
+
+    if (0 == strncmp(result->mType, kOperationalServiceName, sizeof(kOperationalServiceName)))
+    {
+        result->ToDiscoveredOperationalNodeBrowseData(nodeData);
+
+        nodeData.Get<OperationalNodeBrowseData>().LogDetail();
+        discoveryContext->OnNodeDiscovered(nodeData);
```
Damian-Nordic, line 59: You have to call `discoveryContext->Release()` before returning.

src/platform/Linux/DnssdImpl.cpp
```diff
@@ -772,7 +796,24 @@ void MdnsAvahi::HandleBrowse(AvahiServiceBrowser * browser, AvahiIfIndex interfa
...
+            CopyTypeWithoutProtocol(service.mType, type);
+            service.mProtocol      = GetProtocolInType(type);
+            service.mAddressType   = context->mAddressType;
+            service.mTransportType = ToAddressType(protocol);
+            service.mInterface     = Inet::InterfaceId::Null();
+            if (interface != AVAHI_IF_UNSPEC)
+            {
+                service.mInterface = static_cast<chip::Inet::InterfaceId>(interface);
+            }
+            service.mType[kDnssdTypeMaxSize] = 0;
```
Damian-Nordic, line 812: `CopyTypeWithoutProtocol` already takes care of that.

src/platform/Linux/DnssdImpl.cpp
```diff
@@ -743,20 +748,39 @@ void MdnsAvahi::HandleBrowse(AvahiServiceBrowser * browser, AvahiIfIndex interfa
...
         ChipLogProgress(DeviceLayer, "Avahi browse: all for now");
         bool needRetries = context->mBrowseRetries++ < kMaxBrowseRetries && !context->mStopped.load();
+
         // If we were already asked to stop, no need to send a callback - no one is listening.
         if (!context->mStopped.load())
         {
             context->mCallback(context->mContext, context->mServices.data(), context->mServices.size(), !needRetries,
                                CHIP_NO_ERROR);
+
+            // Clearing records/services already passed to application through delegate. Keeping it may cause
```
Damian-Nordic, line 758: But the application will anyway receive duplicates if we do a retry, right?
The question is, should we even do any retries if we do a continuous query? Shouldn't we let avahi just keep listening without sending more active queries?

src/lib/dnssd/Discovery_ImplPlatform.cpp
su-shanka: This is filled into OperationalNodeBrowseData for operational which doesn't have any member for ip address, hence for operational browse ipAddress is not needed or accessed.

src/lib/dnssd/Discovery_ImplPlatform.cpp
su-shanka: That would release the context, and the following browse results will not have context or delegate. I think releasing makes sense for one shot query, does it makes sense to release for continuous query? 
When Stop is called the context gets released anyways right?

Damian-Nordic: This is reference counted object and `Release()` decrements the reference count which was increased for each service before this function is called (otherwise, there will be a context leak). Until `HandleNodeBrowse` is called with `finalBrowse=true`, the object is not actually released.

src/lib/dnssd/Discovery_ImplPlatform.cpp
Damian-Nordic: I understand but the `else` code assumes that the address exists so by adding this code you make it possible to dereference a null optional.

src/platform/Linux/DnssdImpl.cpp
su-shanka: Here in same delegates we were sending duplicate records. delegate get called first time with record x without duplicates, the next delegate call with one duplicate and the following with 2 and so on. Also, its almost like we are caches once without library and in platform implementation which seemed redundant.
However, need for retry seems like a valid question. But should we not do any further active queries at all? Or should we skip the exponential retries and have periodic active queries with larger intervals?

Damian-Nordic: I was wondering if Avahi doesn't do retries on its own. At least this is what I observe on wireshark when I run `avahi-browse` command. The retries must follow an exponential backoff per mDNS specification.

The problem with killing and restarting the browser is that there's a short period of time Avahi will not listen. Additionally, if it has some services cached, they will be provided to the application for each retry - is that right or am I missing something? Thus, if I understand correctly that Avahi does retries on its own, it would be easier and better to just keep it running.

src/lib/dnssd/Discovery_ImplPlatform.cpp
bzbarsky-apple: @Damian-Nordic is correct: all codepaths through here need to call Release().

src/lib/dnssd/Discovery_ImplPlatform.cpp
bzbarsky-apple, line 89: Again, I am not sure strncmp is the right thing here.

src/lib/dnssd/Discovery_ImplPlatform.cpp
```diff
@@ -75,8 +85,9 @@ static void HandleNodeBrowse(void * context, DnssdService * services, size_t ser
 
         auto & ipAddress = services[i].mAddress;
 
-        // Check if SRV, TXT and AAAA records were received in DNS responses
-        if (strlen(services[i].mHostName) == 0 || services[i].mTextEntrySize == 0 || !ipAddress.has_value())
+        // Check that this is not operational browse result and if SRV, TXT and AAAA records were received in DNS responses
```
bzbarsky-apple, line 88: This comment should be explaining why we are doing what we are doing, not what we are doing....  Why is this the right thing to check for?  Do we not care about IP addresses for operational browse?  If so, _that_ is what we should document here.

src/lib/dnssd/Discovery_ImplPlatform.cpp
bzbarsky-apple: This line:
```
Inet::IPAddress * address = &(*ipAddress);
```
will have undefined behavior (not necessarily crash, since `operator*` does not do checks) `ipAddress` has no value.

The logic here needs to be revamped, and the assumptions of `HandleNodeResolve` need to be audited.

Alternately, we should have two different versions of `HandleNodeResolve` for operational and commissionable, since they seem to have quite different logic anyway.

src/lib/dnssd/Discovery_ImplPlatform.cpp
```diff
@@ -746,6 +765,31 @@ CHIP_ERROR DiscoveryImplPlatform::DiscoverCommissioners(DiscoveryFilter filter,
     return error;
 }
 
+CHIP_ERROR DiscoveryImplPlatform::DiscoverOperational(DiscoveryFilter filter, DiscoveryContext & context)
+{
+    ReturnErrorOnFailure(InitImpl());
+    StopDiscovery(context);
+
+    char serviceName[kMaxCommissionerServiceNameSize];
```
bzbarsky-apple, line 773: Why is kMaxCommissionerServiceNameSize the right thing here?  That doesn't seem related to operational anything.

src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -607,7 +607,7 @@ bool ResolveContext::TryReportingResultsForInterfaceIndex(uint32_t interfaceInde
     {
         auto delegate = static_cast<DiscoverNodeDelegate *>(context);
         DiscoveredNodeData nodeData;
-        service.ToDiscoveredNodeData(addresses, nodeData);
+        service.ToDiscoveredCommissionNodeData(addresses, nodeData);
```
bzbarsky-apple, line 610: So this does not make much sense to me.  This could be an operational browse, no?  `DiscoverNodeDelegate` claims to support operational, commissionable, etc.

I think we are fundamentally making an odd assumption that "operational browse means just finding the instance name" or something like that, but that's not what DiscoverNodeDelegate consumers want or get....

src/platform/Linux/DnssdImpl.cpp
bzbarsky-apple, line 1: I did not look at the Linux-specific bits.

src/lib/dnssd/Discovery_ImplPlatform.cpp
su-shanka: fixed it.

src/lib/dnssd/Discovery_ImplPlatform.cpp
su-shanka: fixed it.

src/lib/dnssd/Discovery_ImplPlatform.cpp
```diff
@@ -50,7 +50,17 @@ static void HandleNodeResolve(void * context, DnssdService * result, const Span<
     }
 
     DiscoveredNodeData nodeData;
-    result->ToDiscoveredNodeData(addresses, nodeData);
+
+    if (0 == strncmp(result->mType, kOperationalServiceName, sizeof(kOperationalServiceName)))
```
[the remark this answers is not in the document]
su-shanka: Yes. I don't see, how else can we differentiate between the browse results for Operational. Any other thoughts on how we can differentiate?

src/platform/Darwin/DnssdContexts.cpp
su-shanka: I was not sure if Darwin supports operational browse yet. I might have incorrectly assumed that it doesn't. 
We ideally need to call either `ToDiscoveredCommissionNodeData()` or `ToDiscoveredOperationalNodeBrowseData()` based on the browse type. This is what I have done in `Discovery_ImplPlatform.cpp`.
Does that make sense?

src/platform/Linux/DnssdImpl.cpp
su-shanka: Yes, Avahi does have services cached which is provided for each retry. That is exactly why I wanted to clear the services cached in platform implementation here. 
I am not quite sure if avahi does retry on its own. Let me explore on this and get back.

src/lib/dnssd/Discovery_ImplPlatform.cpp
su-shanka: You are right. I will fix this.

src/lib/dnssd/Discovery_ImplPlatform.cpp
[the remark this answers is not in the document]
bzbarsky-apple: So if `result->mType` is `_matterc`, what happens on this line of code?

src/platform/Darwin/DnssdContexts.cpp
bzbarsky-apple: > I was not sure if Darwin supports operational browse yet.

I am telling you that it does.

And what it expects out of an operational browse in this case is complete information including the IPs and whatnot, no?

bzbarsky-apple: Well, my point is that there is existing operational browse code, and that code is expected to produce IPs and whatnot, and it's used by other code...

Maybe that means we need to figure out what our data structures and their naming look like better, but that ResolveContext code is absolutely reached for any kind of browse, and DiscoverNodeDelegate explicitly supports any kind of browse and produces full information for it.

If we have to do things this way for now just to unblock you we can, but the naming is pretty broken and we should at least have a followup to fix it...

su-shanka: `DiscoverNodeDelegate` previously only delivered `CommissionNodeData`, even if the darwin implementation supported operational browse, the user of `DiscoverNodeDelegate` would probably not be expecting it. Also, during the refactor we have made sure all existing users of `DiscoverNodeDelegate` check that `DiscoverNodeData` has `CommissionNodeData` instead of `OperationalNodeBrowseData`.

Anyways, I will need this to get unblocked. So, please suggest how to proceed. Do I need to have check for serviceType and fill the right structure (i.e. either `CommissionNodeData` or `OperationalNodeBrowseData`) here? or does the current changes looks good for now?

src/platform/Linux/DnssdImpl.cpp
su-shanka: I commented all retry and re-attempt to querying. Looks like avahi takes care of exp. retries on its own and then some. The retries goes up exponentially till 64s and then queries every 97s +/- 1s.
As discussed offline,  removes all sorts of retries and letting avahi take care of retries.

src/lib/dnssd/Discovery_ImplPlatform.cpp
su-shanka: fixed the length to use kMaxOperationalServiceNameSize.

src/lib/dnssd/Discovery_ImplPlatform.cpp
su-shanka: fixed

src/lib/dnssd/Discovery_ImplPlatform.cpp
```diff
@@ -73,13 +93,22 @@ static void HandleNodeBrowse(void * context, DnssdService * services, size_t ser
         discoveryContext->Retain();
         // For some platforms browsed services are already resolved, so verify if resolve is really needed or call resolve callback
 
-        auto & ipAddress = services[i].mAddress;
+        auto & ipAddress         = services[i].mAddress;
+        bool isOperationalBrowse = (strncmp(services[i].mType, kOperationalServiceName, sizeof(kOperationalServiceName)) == 0 &&
+                                    strlen(services[i].mType) == strlen(kOperationalServiceName));
```
Damian-Nordic, line 98:
```suggestion
        bool isOperationalBrowse = (strncmp(services[i].mType, kOperationalServiceName, sizeof(services[i].mType));
```
It doesn't make sense to use `strncmp` with max size set to the length of a string literal (which we know is null-terminated). It's `mType` buffer that needs the bounds check.

src/lib/dnssd/Discovery_ImplPlatform.cpp
```diff
@@ -73,13 +93,22 @@ static void HandleNodeBrowse(void * context, DnssdService * services, size_t ser
...
 
-        // Check if SRV, TXT and AAAA records were received in DNS responses
-        if (strlen(services[i].mHostName) == 0 || services[i].mTextEntrySize == 0 || !ipAddress.has_value())
+        // if SRV, TXT and AAAA records were received in DNS responses, also for operational browse result we currently
+        // don't need IP address hence skip resolution.
+        if (!isOperationalBrowse &&
```
Damian-Nordic, line 102: You can make the `if (isOperationalBrowse)` the first one to avoid checking if `!isOperationalBrowse`.

src/platform/Linux/DnssdImpl.cpp
```diff
@@ -738,41 +744,52 @@ void MdnsAvahi::HandleBrowse(AvahiServiceBrowser * browser, AvahiIfIndex interfa
...
     case AVAHI_BROWSER_REMOVE:
         ChipLogProgress(DeviceLayer, "Avahi browse: remove");
         if (strcmp("local", domain) == 0)
         {
-            context->mServices.erase(
-                std::remove_if(context->mServices.begin(), context->mServices.end(), [name, type](const DnssdService & service) {
-                    return strcmp(name, service.mName) == 0 && type == GetFullType(service.mType, service.mProtocol);
-                }));
+            // don't attempt to erase if vector has been cleared
+            if (context->mServices.size())
```
Damian-Nordic, line 765: nit: this check is redundant as if the vector is empty, `erase` is a no-op anyway.

src/platform/Linux/DnssdImpl.cpp
```diff
@@ -738,41 +744,52 @@ void MdnsAvahi::HandleBrowse(AvahiServiceBrowser * browser, AvahiIfIndex interfa
...
+            {
+                context->mServices.erase(std::remove_if(
+                    context->mServices.begin(), context->mServices.end(), [name, type](const DnssdService & service) {
+                        return strcmp(name, service.mName) == 0 && type == GetFullType(service.mType, service.mProtocol);
+                    }));
+            }
+
+            DnssdService service = {};
```
Damian-Nordic, line 773: This service initialization could be in the `if (mReceivedAllCached)` block as it's not needed otherwise.

Damian-Nordic (approved):

src/platform/Linux/DnssdImpl.cpp
su-shanka: There was actually an issue without the check hence it was added. without the check the item_count or vector size is decremented to -1. 
**Observation:** clear() seems to be just decrementing the count and not really remove the contents, so remove_if still finds the content and returns element and erase clears and is decrementing size to -1.

src/platform/Linux/DnssdImpl.cpp
su-shanka: Damian has reviewed the linux specific bits.

src/platform/Darwin/DnssdContexts.cpp
su-shanka: As discussed on slack, have added check for serviceType to fill the right structure before calling delegate.

src/lib/dnssd/Discovery_ImplPlatform.cpp
```diff
@@ -75,8 +95,17 @@ static void HandleNodeBrowse(void * context, DnssdService * services, size_t ser
 
         auto & ipAddress = services[i].mAddress;
 
-        // Check if SRV, TXT and AAAA records were received in DNS responses
-        if (strlen(services[i].mHostName) == 0 || services[i].mTextEntrySize == 0 || !ipAddress.has_value())
+        // mType(service name) exactly matches with operational service name
+        bool isOperationalBrowse = (strncmp(services[i].mType, kOperationalServiceName, sizeof(services[i].mType)) == 0 &&
+                                    strlen(services[i].mType) == strlen(kOperationalServiceName));
```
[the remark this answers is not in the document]
bzbarsky-apple: In particular:

1. `kOperationalServiceName` is obviously null-terminated, and we are not using its length to constrain the `strncmp` anyway.
2. If `services[i].mType` is not null-terminated, the `strncmp` will avoid running off the end of the buffer, but then the `strlen` will happily go off the end.

So if the intent is to handle a not-terminated `services[i].mType`, then this code does not do it right.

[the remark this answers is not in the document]
su-shanka: changed to strcmp

src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -607,7 +607,17 @@ bool ResolveContext::TryReportingResultsForInterfaceIndex(uint32_t interfaceInde
     {
         auto delegate = static_cast<DiscoverNodeDelegate *>(context);
         DiscoveredNodeData nodeData;
-        service.ToDiscoveredNodeData(addresses, nodeData);
+
+        // mType(service name) exactly matches with operational service name
+        if (strncmp(services[i].mType, kOperationalServiceName, sizeof(services[i].mType)) == 0 &&
+            strlen(services[i].mType) == strlen(kOperationalServiceName))
```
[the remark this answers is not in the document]
su-shanka: changed to strcmp

src/lib/dnssd/Discovery_ImplPlatform.cpp
```diff
@@ -75,8 +95,17 @@ static void HandleNodeBrowse(void * context, DnssdService * services, size_t ser
...
+
+        // For operational browse result we currently don't need IP address hence skip resolution and handle differently.
+        if (isOperationalBrowse)
+        {
+            HandleNodeOperationalBrowse(context, &services[i], error);
+        }
+        // if SRV, TXT and AAAA records were received in DNS responses
```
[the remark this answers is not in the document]
su-shanka: changed

src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -607,7 +607,16 @@ bool ResolveContext::TryReportingResultsForInterfaceIndex(uint32_t interfaceInde
     {
         auto delegate = static_cast<DiscoverNodeDelegate *>(context);
         DiscoveredNodeData nodeData;
-        service.ToDiscoveredNodeData(addresses, nodeData);
+
+        // mType(service name) exactly matches with operational service name
+        if (strncmp(service[i].mType, kOperationalServiceName) == 0)
```
bzbarsky-apple, line 612:
```suggestion
        if (strcmp(service[i].mType, kOperationalServiceName) == 0)
```

src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -607,7 +607,16 @@ bool ResolveContext::TryReportingResultsForInterfaceIndex(uint32_t interfaceInde
     {
         auto delegate = static_cast<DiscoverNodeDelegate *>(context);
         DiscoveredNodeData nodeData;
-        service.ToDiscoveredNodeData(addresses, nodeData);
+
+        // mType(service name) exactly matches with operational service name
```
bzbarsky-apple, line 611:
```suggestion
        // Check whether mType (service name) exactly matches with operational service name
```

bzbarsky-apple (approved):

PR #33408, opened 2024-05-10
Darwin:  MTRDevice should not persist non-priming attribute values for attributes with Changes Omitted Quality

kiel-apple:
Some devices are reporting these attributes outside the priming report.  Avoid repeated writes to persistent storage by only updating in-memory cache.

A subsequent PR will add logging and metrics to help understand what devices currently have this behavior.

**TODO in Followup PRs**
- [ ] add a test
- [ ] remove convenience overload and force callers to decide if they are for a subscription or not

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -1400,7 +1400,7 @@ static NSString * const sAttributesKey = @"attributes";
     }
 
     [clusterData storeValue:value forAttribute:path.attribute];
-    
+
     if (value != nil
         && isFromSubscription
         && !_receivingPrimingReport
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -1163,12 +1163,16 @@ static NSString * const sAttributesKey = @"attributes";
     }
 }
 
-- (void)_handleAttributeReport:(NSArray<NSDictionary<NSString *, id> *> *)attributeReport
-{
+- (void)_handleAttributeReport:(NSArray<NSDictionary<NSString *, id> *> *)attributeReport fromSubscription:(BOOL)isFromSubscription {
     std::lock_guard lock(_lock);
 
     // _getAttributesToReportWithReportedValues will log attribute paths reported
-    [self _reportAttributes:[self _getAttributesToReportWithReportedValues:attributeReport]];
+    [self _reportAttributes:[self _getAttributesToReportWithReportedValues:attributeReport fromSubscription:isFromSubscription]];
+}
+
+- (void)_handleAttributeReport:(NSArray<NSDictionary<NSString *, id> *> *)attributeReport
+{
+    [self _handleAttributeReport:attributeReport fromSubscription:NO];
 }
 
 #ifdef DEBUG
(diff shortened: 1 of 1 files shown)


ksperling-apple (approved):

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -1368,6 +1373,16 @@ - (void)_setCachedAttributeValue:(MTRDeviceDataValueDictionary _Nullable)value f
 
     [clusterData storeValue:value forAttribute:path.attribute];
 
+    if (value != nil
+        && isFromSubscription
+        && !_receivingPrimingReport
+        && AttributeHasChangesOmittedQuality(path)) {
+        // do not cache new values for Changes Omitted Quality attributes unless
```
bzbarsky-apple, line 1380:
```suggestion
        // Do not persist new values for Changes Omitted Quality attributes unless
```

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -1134,12 +1134,17 @@ - (void)_reportAttributes:(NSArray<NSDictionary<NSString *, id> *> *)attributes
...
+- (void)_handleAttributeReport:(NSArray<NSDictionary<NSString *, id> *> *)attributeReport fromSubscription:(BOOL)isFromSubscription
 {
     std::lock_guard lock(_lock);
 
     // _getAttributesToReportWithReportedValues will log attribute paths reported
-    [self _reportAttributes:[self _getAttributesToReportWithReportedValues:attributeReport]];
+    [self _reportAttributes:[self _getAttributesToReportWithReportedValues:attributeReport fromSubscription:isFromSubscription]];
+}
+
+- (void)_handleAttributeReport:(NSArray<NSDictionary<NSString *, id> *> *)attributeReport
```
bzbarsky-apple, line 1145: I think we should remove this overload, so that callers have to clearly decide whether they are "from subscription" or not.

For example, the unit test thing should be doing fromSubscription:YES, I think, or making that an argument or something.

bzbarsky-apple (approved):

src/darwin/Framework/CHIP/MTRDevice.mm
kiel-apple: Agree - thought about it, but decided to go with the "smallest possible change" for the first pass.  I'll see how many call sites that change affects - might need some context from the team updating some of them.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -1368,6 +1373,16 @@ - (void)_setCachedAttributeValue:(MTRDeviceDataValueDictionary _Nullable)value f
 
     [clusterData storeValue:value forAttribute:path.attribute];
 
+    if (value != nil
+        && isFromSubscription
+        && !_receivingPrimingReport
```
jtung-apple, line 1378: If https://github.com/project-chip/connectedhomeip/pull/33462/files merges first, we'll need change the `_receivingPrimingReport` check to a `HadSubscriptionEstablishedOnce(_internalDeviceState)` one.

jtung-apple (approved):

src/darwin/Framework/CHIP/MTRDevice.mm
bzbarsky-apple: It affects 3 callsites: subscriptions, reads, and the unit test bit.

kiel-apple: I'm going to make a new PR for these changes in the coming hours.

kiel-apple: PR to remove convenience method: https://github.com/project-chip/connectedhomeip/pull/33489

PR #33411, opened 2024-05-13
Fixes and additions of some of the bounds checks in SystemPacketBuffer::New()

pidarped:
With the data length types changed to size_t, the bounds checks for the lengths need to be appropriately modified so that we ensure that overflow does not happen and the buffer allocation is performed correctly as requested by the caller of the API.

### DIFF
--- a/src/system/SystemPacketBuffer.cpp
+++ b/src/system/SystemPacketBuffer.cpp
@@ -510,13 +510,13 @@ PacketBufferHandle PacketBufferHandle::New(size_t aAvailableSize, uint16_t aRese
 {
     // Sanity check for kStructureSize to ensure that it matches the PacketBuffer size.
     static_assert(PacketBuffer::kStructureSize == sizeof(PacketBuffer), "PacketBuffer size mismatch");
-    // Setting a static upper bound on kStructureSize to ensure the summation of all the sizes do not overflow.
+    // Setting a static upper bound on kStructureSize to ensure the summation of all the sizes does not overflow.
     static_assert(PacketBuffer::kStructureSize <= UINT16_MAX, "kStructureSize should not exceed UINT16_MAX.");
-    // Setting a static upper bound on the maximum buffer size allocation for regular sized messages(not large).
+    // Setting a static upper bound on the maximum buffer size allocation for regular sized messages (not large).
     static_assert(PacketBuffer::kMaxSizeWithoutReserve <= UINT16_MAX, "kMaxSizeWithoutReserve should not exceed UINT16_MAX.");
 
     // Ensure that aAvailableSize is bound within a max and is not big enough to cause overflow during
-    // subsequent addtion of all the sizes.
+    // subsequent addition of all the sizes.
     if (aAvailableSize > UINT32_MAX)
     {
         ChipLogError(chipSystemLayer,
--- a/src/system/SystemPacketBuffer.cpp
+++ b/src/system/SystemPacketBuffer.cpp
@@ -508,37 +508,64 @@ void PacketBuffer::AddRef()
 
 PacketBufferHandle PacketBufferHandle::New(size_t aAvailableSize, uint16_t aReservedSize)
 {
-    // Adding three 16-bit-int sized numbers together will never overflow
-    // assuming int is at least 32 bits.
-    static_assert(INT_MAX >= INT32_MAX, "int is not big enough");
+    // Sanity check for kStructureSize to ensure that it matches the PacketBuffer size.
     static_assert(PacketBuffer::kStructureSize == sizeof(PacketBuffer), "PacketBuffer size mismatch");
-    static_assert(PacketBuffer::kStructureSize < UINT16_MAX, "Check for overflow more carefully");
-    static_assert(SIZE_MAX >= INT_MAX, "Our additions might not fit in size_t");
-    static_assert(PacketBuffer::kMaxSizeWithoutReserve <= UINT32_MAX, "PacketBuffer may have size not fitting uint32_t");
-#if CHIP_SYSTEM_CONFIG_USE_LWIP
-    // LwIP based APIs have a maximum buffer size of UINT16_MAX. Ensure that
-    // limit is met during allocation.
-    VerifyOrDieWithMsg(aAvailableSize + aReservedSize < UINT16_MAX, chipSystemLayer,
-                       "LwIP based systems can handle only up to UINT16_MAX!");
-#endif // CHIP_SYSTEM_CONFIG_USE_LWIP
+    // Setting a static upper bound on kStructureSize to ensure the summation of all the sizes do not overflow.
+    static_assert(PacketBuffer::kStructureSize <= UINT16_MAX, "kStructureSize should not exceed UINT16_MAX.");
+    // Setting a static upper bound on the maximum buffer size allocation for regular sized messages(not large).
+    static_assert(PacketBuffer::kMaxSizeWithoutReserve <= UINT16_MAX, "kMaxSizeWithoutReserve should not exceed UINT16_MAX.");
 
-    // When `aAvailableSize` fits in uint16_t (as tested below) and size_t is at least 32 bits (as asserted above),
-    // these additions will not overflow.
-    const size_t lAllocSize = aReservedSize + aAvailableSize;
-    const size_t lBlockSize = PacketBuffer::kStructureSize + lAllocSize;
-    PacketBuffer * lPacket;
-
-    CHIP_SYSTEM_FAULT_INJECT(FaultInjection::kFault_PacketBufferNew, return PacketBufferHandle());
-
-    // TODO: Change the max to a lower value
-    if (aAvailableSize > UINT32_MAX || lAllocSize > PacketBuffer::kMaxSizeWithoutReserve || lBlockSize > UINT32_MAX)
+    // Ensure that aAvailableSize is bound within a max and is not big enough to cause overflow during
+    // subsequent addtion of all the sizes.
+    if (aAvailableSize > UINT32_MAX)
     {
-        ChipLogError(chipSystemLayer, "PacketBuffer: allocation too large.");
+        ChipLogError(chipSystemLayer,
+                     "PacketBuffer: AvailableSize of a buffer cannot exceed UINT32_MAX. aAvailableSize = 0x" ChipLogFormatX64,
+                     ChipLogValueX64(static_cast<uint64_t>(aAvailableSize)));
+        return PacketBufferHandle();
+    }
+
+    // Cast all to uint64_t and add. This cannot overflow because we have
+    // ensured that the maximal value of the summation is
+    // UINT32_MAX + UINT16_MAX + UINT16_MAX, which should always fit in
+    // a uint64_t variable.
+    uint64_t sumOfSizes = static_cast<uint64_t>(aAvailableSize) + static_cast<uint64_t>(aReservedSize) +
+        static_cast<uint64_t>(PacketBuffer::kStructureSize);
+    uint64_t sumOfAvailAndReserved = static_cast<uint64_t>(aAvailableSize) + static_cast<uint64_t>(aReservedSize);
+
+    // Ensure that the sum fits in a size_t so that casting into size_t variables,
+    // viz., lBlockSize and lAllocSize, is safe.
+    if (!CanCastTo<size_t>(sumOfSizes))
+    {
+        ChipLogError(chipSystemLayer,
+                     "PacketBuffer: Sizes of allocation request are invalid. (aAvailableSize = " ChipLogFormatX64
+                     ", aReservedSize = " ChipLogFormatX64 ")",
+                     ChipLogValueX64(static_cast<uint64_t>(aAvailableSize)), ChipLogValueX64(static_cast<uint64_t>(aReservedSize)));
         return PacketBufferHandle();
     }
 
 #if CHIP_SYSTEM_CONFIG_USE_LWIP
+    // LwIP based APIs have a maximum buffer size of UINT16_MAX. Ensure that
+    // limit is met during allocation.
+    VerifyOrDieWithMsg(sumOfAvailAndReserved < UINT16_MAX, chipSystemLayer, "LwIP based systems can handle only up to UINT16_MAX!");
+#endif // CHIP_SYSTEM_CONFIG_USE_LWIP
 
+    // sumOfAvailAndReserved is no larger than sumOfSizes, which we checked can be cast to
+    // size_t.
+    const size_t lAllocSize = static_cast<size_t>(sumOfAvailAndReserved);
+    PacketBuffer * lPacket;
+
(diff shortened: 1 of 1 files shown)


src/system/SystemPacketBuffer.cpp
```diff
@@ -508,45 +508,63 @@ void PacketBuffer::AddRef()
 
 PacketBufferHandle PacketBufferHandle::New(size_t aAvailableSize, uint16_t aReservedSize)
 {
-    // Adding three 16-bit-int sized numbers together will never overflow
-    // assuming int is at least 32 bits.
+    // Ensure that SIZE_MAX is at least 32 bits
```
bzbarsky-apple, line 511: Why do we care about ensuring this?  Where do we use this information?

As far as I can tell, the INT_MAX and SIZE_MAX asserts are not checking anything we care about anymore....

src/system/SystemPacketBuffer.cpp
```diff
@@ -508,45 +508,63 @@ void PacketBuffer::AddRef()
...
     static_assert(INT_MAX >= INT32_MAX, "int is not big enough");
+    static_assert(SIZE_MAX >= INT_MAX, "Our additions may not fit in size_t");
+
+    // Sanity check for kStructureSize and putting an upper bound.
```
bzbarsky-apple, line 515:
```suggestion
    // Sanity check for kStructureSize and make sure our constant sizes are not going to lead to overflow.
```

src/system/SystemPacketBuffer.cpp
```diff
@@ -508,45 +508,63 @@ void PacketBuffer::AddRef()
...
     static_assert(PacketBuffer::kStructureSize == sizeof(PacketBuffer), "PacketBuffer size mismatch");
-    static_assert(PacketBuffer::kStructureSize < UINT16_MAX, "Check for overflow more carefully");
-    static_assert(SIZE_MAX >= INT_MAX, "Our additions might not fit in size_t");
-    static_assert(PacketBuffer::kMaxSizeWithoutReserve <= UINT32_MAX, "PacketBuffer may have size not fitting uint32_t");
+    static_assert(PacketBuffer::kStructureSize < UINT16_MAX, "kStructureSize should not exceed UINT16_MAX");
+    static_assert(PacketBuffer::kMaxSizeWithoutReserve <= UINT16_MAX, "PacketBuffer may have size not fitting uint16_t");
+
+    // Have a bounds check for aAvailableSize to ensure that
+    // the next overflow check is done correctly.
```
bzbarsky-apple, line 521:
```suggestion
    // Ensure that aAvailableSize is not big enough to cause overflow.
```

src/system/SystemPacketBuffer.cpp
```diff
@@ -508,45 +508,63 @@ void PacketBuffer::AddRef()
...
+    // Ensure that SIZE_MAX is at least 32 bits
     static_assert(INT_MAX >= INT32_MAX, "int is not big enough");
+    static_assert(SIZE_MAX >= INT_MAX, "Our additions may not fit in size_t");
+
+    // Sanity check for kStructureSize and putting an upper bound.
     static_assert(PacketBuffer::kStructureSize == sizeof(PacketBuffer), "PacketBuffer size mismatch");
-    static_assert(PacketBuffer::kStructureSize < UINT16_MAX, "Check for overflow more carefully");
-    static_assert(SIZE_MAX >= INT_MAX, "Our additions might not fit in size_t");
-    static_assert(PacketBuffer::kMaxSizeWithoutReserve <= UINT32_MAX, "PacketBuffer may have size not fitting uint32_t");
+    static_assert(PacketBuffer::kStructureSize < UINT16_MAX, "kStructureSize should not exceed UINT16_MAX");
```
bzbarsky-apple, line 517: Is there a reason this is `<`, not `<=`?

src/system/SystemPacketBuffer.cpp
```diff
@@ -508,45 +508,63 @@ void PacketBuffer::AddRef()
...
+    static_assert(PacketBuffer::kMaxSizeWithoutReserve <= UINT16_MAX, "PacketBuffer may have size not fitting uint16_t");
```
bzbarsky-apple, line 518: I don't understand the message on this static assert....

src/system/SystemPacketBuffer.cpp
```diff
@@ -508,45 +508,63 @@ void PacketBuffer::AddRef()
...
+
+    // Have a bounds check for aAvailableSize to ensure that
+    // the next overflow check is done correctly.
+    if (aAvailableSize > UINT32_MAX)
+    {
+        ChipLogError(chipSystemLayer, "PacketBuffer: AvailableSize of a buffer cannot exceed UINT32_MAX");
```
bzbarsky-apple, line 524: Worth logging the invalid value that was provided?

src/system/SystemPacketBuffer.cpp
```diff
@@ -508,45 +508,63 @@ void PacketBuffer::AddRef()
...
+        return PacketBufferHandle();
+    }
+
+    // Cast all to uint64_t and add to consider the widest possible result.
```
bzbarsky-apple, line 528:
```suggestion
    // Cast all to uint64_t and add.  This can't overflow, because we have ensured that the maximal sum here is
    // UINT32_MAX + UINT16_MAX + UINT16_MAX, which always fits in a 64-bit unsigned integer. 
```

src/system/SystemPacketBuffer.cpp
```diff
@@ -508,45 +508,63 @@ void PacketBuffer::AddRef()
...
+    uint64_t sumOfSizes = static_cast<uint64_t>(aAvailableSize) + static_cast<uint64_t>(aReservedSize) +
+        static_cast<uint64_t>(PacketBuffer::kStructureSize);
+    uint64_t sumOfAvailAndReserved = static_cast<uint64_t>(aAvailableSize) + static_cast<uint64_t>(aReservedSize);
+
+    // Ensure that the sum does not overflow a size_t whose max value is at
+    // least INT_MAX.
+    if (sumOfSizes > std::numeric_limits<size_t>::max())
```
bzbarsky-apple, line 535: Is there a reason this is not a `CanCastTo<size_t>(sumOfSizes)` check?  That's what we really care about, right?

src/system/SystemPacketBuffer.cpp
```diff
@@ -508,45 +508,63 @@ void PacketBuffer::AddRef()
...
+    {
+        ChipLogError(chipSystemLayer, "PacketBuffer: Sizes of allocation request are invalid");
```
bzbarsky-apple, line 537: This should probably log the values involved?

src/system/SystemPacketBuffer.cpp
```diff
@@ -508,45 +508,63 @@ void PacketBuffer::AddRef()
...
-                       "LwIP based systems can handle only up to UINT16_MAX!");
+    VerifyOrDieWithMsg(sumOfAvailAndReserved < UINT16_MAX, chipSystemLayer, "LwIP based systems can handle only up to UINT16_MAX!");
 #endif // CHIP_SYSTEM_CONFIG_USE_LWIP
 
-    // When `aAvailableSize` fits in uint16_t (as tested below) and size_t is at least 32 bits (as asserted above),
-    // these additions will not overflow.
-    const size_t lAllocSize = aReservedSize + aAvailableSize;
-    const size_t lBlockSize = PacketBuffer::kStructureSize + lAllocSize;
+    // The above check against the size_t limit ensures that the below sum will not
+    // overflow.
```
bzbarsky-apple, line 548:
```suggestion
    // sumOfAvailAndReserved is no larger than sumOfSizes, which we checked can be cast to size_t.
```

src/system/SystemPacketBuffer.cpp
```diff
@@ -508,45 +508,63 @@ void PacketBuffer::AddRef()
...
     {
-        ChipLogError(chipSystemLayer, "PacketBuffer: allocation too large.");
+        ChipLogError(chipSystemLayer, "PacketBuffer: allocation exceeding buffer capacity limits.");
         return PacketBufferHandle();
     }
 
 #if CHIP_SYSTEM_CONFIG_USE_LWIP
-
     lPacket = static_cast<PacketBuffer *>(
         pbuf_alloc(PBUF_RAW, static_cast<uint16_t>(lAllocSize), CHIP_SYSTEM_PACKETBUFFER_LWIP_PBUF_TYPE));
```
bzbarsky-apple, line 562: Should have a comment that this case is safe because we lAllocSize is no larger than kMaxSizeWithoutReserve, which fits in uint16_t.

src/system/SystemPacketBuffer.cpp
```diff
@@ -565,7 +583,7 @@ PacketBufferHandle PacketBufferHandle::New(size_t aAvailableSize, uint16_t aRese
     UNLOCK_BUF_POOL();
 
 #elif CHIP_SYSTEM_PACKETBUFFER_FROM_CHIP_HEAP
-
+    const size_t lBlockSize = PacketBuffer::kStructureSize + lAllocSize;
```
bzbarsky-apple, line 586: So this is safe because that addition on the right is just `sumOfSizes` which we already checked fits in size_t.

But it might be clearer to just do `static_cast<size_t>(sumOfSizes)` with a comment about how that's kStructureSize + lAllocSize.

pidarped: Yes, they are the same. Will change it to sumOfSizes.

src/system/SystemPacketBuffer.cpp
pidarped: Added comment.

src/system/SystemPacketBuffer.cpp
andy31415: It may help to say __why__ this enforcement is done.  Right now it reads like `assert foo < 1234, "Foo should not exceed 1234"` which does not help users understand the sideffects of why this is being enforced. What happesn if I change the `UINT16_MAX` here with `UINT32_MAX` like it was before? why is uint16 max correct but uint32 max was wrong before?

src/system/SystemPacketBuffer.cpp
andy31415: This info should be in the comment. Otherwise the comment just restatest the code.

There should be an explanation why each of these assumptions are required/necessary for our code to work correctly.

src/system/SystemPacketBuffer.cpp
```diff
@@ -508,45 +508,71 @@ void PacketBuffer::AddRef()
...
 PacketBufferHandle PacketBufferHandle::New(size_t aAvailableSize, uint16_t aReservedSize)
 {
-    // Adding three 16-bit-int sized numbers together will never overflow
-    // assuming int is at least 32 bits.
+    // Ensure that SIZE_MAX is at least 32 bits
     static_assert(INT_MAX >= INT32_MAX, "int is not big enough");
+    static_assert(SIZE_MAX >= INT_MAX, "Our additions may not fit in size_t");
+
+    // Sanity check for kStructureSize and ensure that the constant sizes are not going to lead
+    // to overflow.
```
andy31415, line 516: please add some why here - what is the UINT16 that controls size of things (given that pbuf sizes are size_t)? When trying to read this, I am wondering if I set max sizes, I have `kMaxSizeWithoutReserve` being UINT16_MAX and then we reserve space for something ... so the entire pbuf size will be over UINT16_MAX ... is that ok? It seeems hard to follow how these size choices were made.

andy31415, line 516: please add some why here - what is the UINT16 that controls size of things (given that pbuf sizes are size_t)? When trying to read this, I am wondering if I set max sizes, I have `kMaxSizeWithoutReserve` being UINT16_MAX and then we reserve space for something ... so the entire pbuf size will be over UINT16_MAX ... is that ok? It seeems hard to follow how these size choices were made.

https://github.com/project-chip/connectedhomeip/blob/master/src/system/SystemPacketBuffer.h#L94 seems to describe the data layout however that does not immediately indicate why uint16 is important.

src/system/SystemPacketBuffer.cpp
andy31415: we also seem to use uint64_t now for additions ... do we ever care about size_t sizing now?

src/system/SystemPacketBuffer.cpp
```diff
@@ -508,45 +508,71 @@ void PacketBuffer::AddRef()
...
+    // Cast all to uint64_t and add. This cannot overflow because we have
+    // ensured that the maximal value of the summation is
+    // UINT32_MAX + UINT16_MAX + UINT16_MAX, which should always fit in
+    // a uint64_t variable.
+    uint64_t sumOfSizes = static_cast<uint64_t>(aAvailableSize) + static_cast<uint64_t>(aReservedSize) +
+        static_cast<uint64_t>(PacketBuffer::kStructureSize);
+    uint64_t sumOfAvailAndReserved = static_cast<uint64_t>(aAvailableSize) + static_cast<uint64_t>(aReservedSize);
+
+    // Ensure that the sum does not overflow a size_t whose max value is at
+    // least INT_MAX.
```
andy31415, line 538: why do we care about INT_MAX here? packet header sizes seem to be size_t already.

andy31415 (approved):

src/system/SystemPacketBuffer.cpp
pidarped: The uint64_t cast was to ensure that the addition result does not overflow given the max sizes of the corresponding addends. The check for size_t was to ensure that the cast back to a size_t fits. I have a run-time check for CanCastTo<size_t>, so that makes this static_assert for SIZE_MAX redundant. Have removed it.

src/system/SystemPacketBuffer.cpp
pidarped: This is a relatively tight upper bound on the max buffer allocation for regular messages(i.e., not large) to ensure a corresponding upper bound on the sum, so that it does not overflow. UINT32_MAX is a much more loose upper bound for kMaxSizeWithoutReserve( which is equal to CHIP_SYSTEM_CONFIG_PACKETBUFFER_CAPACITY_MAX ).
Have added some text in the comment to qualify.

src/system/SystemPacketBuffer.cpp
pidarped: kMaxSizeWithoutReserve could have been named better IMO. But it essentially represents the max buffer size of a regular message(MRP-based) when the Reserve is 0. So, it includes the AvailableSize + Reserve. The UINT16_MAX is used as an upper bound from the perspective of not overflowing the addition result. I have added some more text to the comment.

pidarped: Modified comment slightly.

src/system/SystemPacketBuffer.cpp
pidarped: Removed INT_MAX.

bzbarsky-apple (approved):

PR #33434, opened 2024-05-14
[openthread] Fixed potential usage fault

kkasperczyk-no:
Pointer to the ThreadDiagnosticsDelegate is used without a null check. It leads to the usage fault, if ThreadNetworkDiagnostics cluster is disabled and pointer is set to nullptr.


arkq (approved):

markaj-nordic (approved):

Damian-Nordic (approved):

LuDuda (commented):
LGTM 👍

PR #33446, opened 2024-05-14
Add log tag for fabric synchronization

yufengwangca:
Add log tag for fabric synchronization

### DIFF
--- a/src/lib/support/logging/Constants.h
+++ b/src/lib/support/logging/Constants.h
@@ -60,6 +60,7 @@ enum LogModule
     kLogModule_Automation,
     kLogModule_CASESessionManager,
     kLogModule_ICD,
+    kLogModule_FabricSync,
 
     kLogModule_Max
 };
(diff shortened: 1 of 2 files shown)


src/lib/support/logging/Constants.h
```diff
@@ -60,6 +60,7 @@ enum LogModule
     kLogModule_Automation,
     kLogModule_CASESessionManager,
     kLogModule_ICD,
+    kLogModule_FabricSync,
```
andy31415, line 63: both this and ICD seem to need adding to https://github.com/project-chip/connectedhomeip/blob/master/src/lib/support/logging/TextOnlyLogging.cpp#L152

yufengwangca: Acked

bzbarsky-apple (approved):

arkq (approved):

PR #33449, opened 2024-05-14
Using strncmp instead of strcmp to make the function robust and safe from buffer overruns.

yufengwangca:
We use strncmp throughout examples/chip-tool/commands/common/Command.cpp for argument comparisons, but in one instance, we use strcmp(argValue, "null").

strncmp would only be necessary if argValue was not null-terminated and had a length of exactly 4, making strcmp unsafe. However, since argValue is always null-terminated in this context, it is safe to use either strncmp(argValue, "null", 4) or strcmp(argValue, "null").

This change is intended to standardize the method used for argument comparison in the code.

### DIFF
--- a/examples/chip-tool/commands/common/Command.cpp
+++ b/examples/chip-tool/commands/common/Command.cpp
@@ -171,7 +171,7 @@ bool HandleNullableOptional(Argument & arg, char * argValue, std::function<bool(
     if (arg.isNullable())
     {
         auto * nullable = reinterpret_cast<chip::app::DataModel::Nullable<T> *>(arg.value);
-        if (strcmp(argValue, "null") == 0)
+        if (argValue != nullptr && strncmp(argValue, "null", 4) == 0)
         {
             nullable->SetNull();
             return true;


jmartinez-silabs (approved):

kiel-apple (approved):

examples/chip-tool/commands/common/Command.cpp
```diff
@@ -171,7 +171,7 @@ bool HandleNullableOptional(Argument & arg, char * argValue, std::function<bool(
     if (arg.isNullable())
     {
         auto * nullable = reinterpret_cast<chip::app::DataModel::Nullable<T> *>(arg.value);
-        if (strcmp(argValue, "null") == 0)
+        if (argValue != nullptr && strncmp(argValue, "null", 4) == 0)
```
Damian-Nordic, line 174: How can `argValue` be not null-terminated? So how do we then know the size of it?

yufengwangca: Yes, it is impossible in this case, we use strncmp for argument comparisons in other cases, but in one instance, we use strcmp(argValue, "null").

strncmp would only be necessary if argValue was not null-terminated and had a length of exactly 4, making strcmp unsafe. However, since argValue is always null-terminated in this context, it is safe to use either strncmp(argValue, "null", 4) or strcmp(argValue, "null").

This change is intended to standardize the method used for argument comparison in the code.

bzbarsky-apple: Just to follow up on the Slack thread about this: "standardizing" breaks the code.  The places that use strncmp are doing _prefix_ checks, but this one is supposed to be an _equality_ check.   We don't want to treat all strings starting with "null" as the null value.

PR #33458, opened 2024-05-14
[Darwin] Disable metric collector debug logging

anush-apple:
- The debug logging in metrics collector is very noisy and debug is enabled by default.
- Disabling the metrics debug logging since it is not really useful except for developer development mode debugging, which is recompile and test.


yufengwangca (approved):

ksperling-apple (approved):

ksperling-apple:

woody-apple (approved):

PR #33462, opened 2024-05-15
Fix historical event detection in MTRDevice.

bzbarsky-apple:
A few fixes here:

* Fix MTRDeviceTests to clean up MTRDevice between tests, so each test starts with a known clean slate.
* Fix the "are we getting a priming report?" detection in MTRDevice.  Relying on reachability state is not correct, because unsolicited reports can mark us reachable even while we are in the middle of a priming report.


kiel-apple (approved):

woody-apple (approved):

bzbarsky-apple:

jtung-apple (approved):

jtung-apple (approved):

PR #33468, opened 2024-05-15
Increase dnssd kHostNameMaxLength to 40 bytes

yunhanw-google:
In pixel 7, we see nsdManager is advertising host name with 40bytes, in android nsdManager, there is no API to set this hostname.
The below is avahi-browse -r _matter._tcp
=  wlan0 IPv6 B6071280809E4E17-000000000001B669             _matter._tcp         local
   hostname = [Android_25101c8afe6a479387b1d63318378d56.local]
   address = [192.168.0.102]
   port = [58586]

During ICD check-in test, we see the mdns advertisment from android cannot be resolved in lit icd board because of the the above length issue so that lit cannot send the check-in message,
In order to resolve this issue, we expand the kHostNameMaxLength to 40 in dnssd.

### DIFF
--- a/src/lib/dnssd/Constants.h
+++ b/src/lib/dnssd/Constants.h
@@ -29,7 +29,12 @@ namespace Dnssd {
  * Matter DNS host settings
  */
 
-inline constexpr size_t kHostNameMaxLength = 16; // MAC or 802.15.4 Extended Address in hex
+// Matter spec expects hostname to be MAC or 802.15.4 Extended Address in hex.
+// But in latest android nsdManager, it would set hostname with 40 bytes with prefix as android_,
+// and there is no existing API to update the hostname, therefore we put temporary workaround with 40 bytes.
+// Follow-up with ticket issue https://github.com/project-chip/connectedhomeip/issues/33474
+inline constexpr size_t kHostNameMaxLength = 40;
+//
 
 /*
  * Matter DNS service subtypes


joonhaengHeo (approved):

mkardous-silabs (approved):

src/lib/dnssd/Constants.h
```diff
@@ -29,7 +29,7 @@ namespace Dnssd {
  * Matter DNS host settings
  */
 
-inline constexpr size_t kHostNameMaxLength = 16; // MAC or 802.15.4 Extended Address in hex
+inline constexpr size_t kHostNameMaxLength = 40; // The max host name length for dnssd in android
```
Damian-Nordic, line 32: I guess that's OK as a workaround, but this means that Android implementation is not spec-compliant, right? Will this eventually be fixed in Android or should the spec be updated to make Android devices discoverable by all compliant Matter devices?

yunhanw-google: I would discuss with our android team for this issue in NSDManager, which may involve a new API surface, since android is pretty fragmented, even though we make this change happen, I think it would be hard to spread this change uniformly across. I would raise this issue with TSG and discuss whether matter the spec can be updated as well.

yunhanw-google: I would work on this for the follow-up https://github.com/project-chip/connectedhomeip/issues/33474, thanks

Damian-Nordic: Thanks for filing the follow up!

andy31415: Let's add this as a comment - why this is a workaround as well as a link to the open issue.
Otherwise the spec seems to explicitly say that we have a different limit and this causes some slight increase in RAM usage (looks like around 100 bytes ... we probably repeat this in 4 places)

andy31415: For comment I mean both keeping the spec description and describing that android is a workaround. Otherwise it is not obvious that a chip dnssd constant is `The max host name length for dnssd in android` is correct ... why was android picked against all other implementations.

andy31415 (approved):

PR #33470, opened 2024-05-15
Update Matter SDK to show Authorisation dialog only for account login flow. 

osama10:
[Problem]
Authorisation dialogue should only be shown to user if user enabled it from the secure settings. `matter_show_authorisation_dialog` show the user selection .

[Solution]
Check the value of `matter_show_authorisation_dialog` before showing the dialogue. If it's 1 only then we show the dialogue.

[Test]
Tested-by: Lazar Kovacic <lkovacic@amazon.com>


lazarkov (approved):

andy31415:
@osama10  could you sign the CLA? Robot says "osamabb" did not sign.

andy31415 (approved):

PR #33472, opened 2024-05-15
[Darwin] MTRDeviceController to limit concurrent subscriptions to Thread-enabled devices

jtung-apple:
This patch adds the ability to throttle concurrent subscriptions to Thread-enabled devices, to avoid flooding the Thread network.

Specific changes:
* Added MTRDeviceControllerParameters property to specify the number of concurrent subscriptions allowed to Thread devices (defaults to 3 per controller)
* Moved the `OnResubscriptionNeeded` retry count etc., from `MTRBaseSubscriptionCallback` to the local `SubscriptionCallback` in `MTRDevice.m`.
* Added logic in MTRDevice to enqueue the subscription work into the concurrent subscriptions pool.

Currently working on unit tests.

### DIFF
--- a/.github/workflows/darwin.yaml
+++ b/.github/workflows/darwin.yaml
@@ -115,6 +115,11 @@ jobs:
                   echo "This is a simple log" > /tmp/darwin/framework-tests/end_user_support_log.txt
                   ../../../out/debug/chip-all-clusters-app --interface-id -1 --end_user_support_log /tmp/darwin/framework-tests/end_user_support_log.txt > >(tee /tmp/darwin/framework-tests/all-cluster-app.log) 2> >(tee /tmp/darwin/framework-tests/all-cluster-app-err.log >&2) &
                   ../../../out/debug/chip-all-clusters-app --interface-id -1 --dac_provider ../../../credentials/development/commissioner_dut/struct_cd_origin_pid_vid_correct/test_case_vector.json --product-id 32768 --discriminator 3839 --secured-device-port 5539 --KVS /tmp/chip-all-clusters-app-kvs2 > >(tee /tmp/darwin/framework-tests/all-cluster-app-origin-vid.log) 2> >(tee /tmp/darwin/framework-tests/all-cluster-app-origin-vid-err.log >&2) &
+                  ../../../out/debug/chip-all-clusters-app --interface-id -1 --discriminator 101 --passcode 1001 --KVS /tmp/chip-all-clusters-app-kvs101 --secured-device-port 5531 &
+                  ../../../out/debug/chip-all-clusters-app --interface-id -1 --discriminator 102 --passcode 1002 --KVS /tmp/chip-all-clusters-app-kvs102 --secured-device-port 5532 &
+                  ../../../out/debug/chip-all-clusters-app --interface-id -1 --discriminator 103 --passcode 1003 --KVS /tmp/chip-all-clusters-app-kvs103 --secured-device-port 5533 &
+                  ../../../out/debug/chip-all-clusters-app --interface-id -1 --discriminator 104 --passcode 1004 --KVS /tmp/chip-all-clusters-app-kvs104 --secured-device-port 5534 &
+                  ../../../out/debug/chip-all-clusters-app --interface-id -1 --discriminator 105 --passcode 1005 --KVS /tmp/chip-all-clusters-app-kvs105 --secured-device-port 5535 &
 
                   export TEST_RUNNER_ASAN_OPTIONS=__CURRENT_VALUE__:detect_stack_use_after_return=1
 
--- a/src/darwin/Framework/CHIP/MTRBaseSubscriptionCallback.h
+++ b/src/darwin/Framework/CHIP/MTRBaseSubscriptionCallback.h
@@ -100,9 +100,6 @@ public:
         mClusterStateCache = std::move(aClusterStateCache);
     }
 
-    // Used to reset Resubscription backoff on events that indicate likely availability of device to come back online
-    void ResetResubscriptionBackoff() { mResubscriptionNumRetries = 0; }
-
 protected:
     // Report an error, which may be due to issues in our own internal state or
     // due to the OnError callback happening.
@@ -147,6 +144,10 @@ protected:
     NSMutableArray * _Nullable mAttributeReports = nil;
     NSMutableArray * _Nullable mEventReports = nil;
 
+    void CallResubscriptionScheduledHandler(NSError * error, NSNumber * resubscriptionDelay);
+    // Copied from ReadClient and customized for MTRDevice resubscription time reset
+    uint32_t ComputeTimeTillNextSubscription();
+
 private:
     DataReportCallback _Nullable mAttributeReportCallback = nil;
     DataReportCallback _Nullable mEventReportCallback = nil;
@@ -181,10 +182,6 @@ private:
     bool mHaveQueuedDeletion = false;
     OnDoneHandler _Nullable mOnDoneHandler = nil;
     dispatch_block_t mInterimReportBlock = nil;
-
-    // Copied from ReadClient and customized for
-    uint32_t ComputeTimeTillNextSubscription();
-    uint32_t mResubscriptionNumRetries = 0;
 };
 
 NS_ASSUME_NONNULL_END
--- a/src/darwin/Framework/CHIP/MTRBaseSubscriptionCallback.h
+++ b/src/darwin/Framework/CHIP/MTRBaseSubscriptionCallback.h
@@ -145,8 +145,6 @@ protected:
     NSMutableArray * _Nullable mEventReports = nil;
 
     void CallResubscriptionScheduledHandler(NSError * error, NSNumber * resubscriptionDelay);
-    // Copied from ReadClient and customized for MTRDevice resubscription time reset
-    uint32_t ComputeTimeTillNextSubscription();
 
 private:
     DataReportCallback _Nullable mAttributeReportCallback = nil;
--- a/src/darwin/Framework/CHIP/MTRBaseSubscriptionCallback.mm
+++ b/src/darwin/Framework/CHIP/MTRBaseSubscriptionCallback.mm
@@ -125,57 +125,29 @@ void MTRBaseSubscriptionCallback::OnDeallocatePaths(ReadPrepareParams && aReadPr
 
 void MTRBaseSubscriptionCallback::OnSubscriptionEstablished(SubscriptionId aSubscriptionId)
 {
-    // ReadClient resets it at ProcessSubscribeResponse after calling OnSubscriptionEstablished, so this is equivalent
-    mResubscriptionNumRetries = 0;
     if (mSubscriptionEstablishedHandler) {
         auto subscriptionEstablishedHandler = mSubscriptionEstablishedHandler;
         subscriptionEstablishedHandler();
     }
 }
 
-uint32_t MTRBaseSubscriptionCallback::ComputeTimeTillNextSubscription()
-{
-    uint32_t maxWaitTimeInMsec = 0;
-    uint32_t waitTimeInMsec = 0;
-    uint32_t minWaitTimeInMsec = 0;
-
-    if (mResubscriptionNumRetries <= CHIP_RESUBSCRIBE_MAX_FIBONACCI_STEP_INDEX) {
-        maxWaitTimeInMsec = GetFibonacciForIndex(mResubscriptionNumRetries) * CHIP_RESUBSCRIBE_WAIT_TIME_MULTIPLIER_MS;
-    } else {
-        maxWaitTimeInMsec = CHIP_RESUBSCRIBE_MAX_RETRY_WAIT_INTERVAL_MS;
-    }
-
-    if (maxWaitTimeInMsec != 0) {
-        minWaitTimeInMsec = (CHIP_RESUBSCRIBE_MIN_WAIT_TIME_INTERVAL_PERCENT_PER_STEP * maxWaitTimeInMsec) / 100;
-        waitTimeInMsec = minWaitTimeInMsec + (Crypto::GetRandU32() % (maxWaitTimeInMsec - minWaitTimeInMsec));
-    }
-
-    return waitTimeInMsec;
-}
-
 CHIP_ERROR MTRBaseSubscriptionCallback::OnResubscriptionNeeded(ReadClient * apReadClient, CHIP_ERROR aTerminationCause)
 {
-    // No need to check ReadClient internal state is Idle because ReadClient only calls OnResubscriptionNeeded after calling ClearActiveSubscriptionState(), which sets the state to Idle.
+    CHIP_ERROR err = ClusterStateCache::Callback::OnResubscriptionNeeded(apReadClient, aTerminationCause);
+    ReturnErrorOnFailure(err);
 
-    // This part is copied from ReadClient's DefaultResubscribePolicy:
-    auto timeTillNextResubscription = ComputeTimeTillNextSubscription();
-    ChipLogProgress(DataManagement,
-        "Will try to resubscribe to %02x:" ChipLogFormatX64 " at retry index %" PRIu32 " after %" PRIu32
-        "ms due to error %" CHIP_ERROR_FORMAT,
-        apReadClient->GetFabricIndex(), ChipLogValueX64(apReadClient->GetPeerNodeId()), mResubscriptionNumRetries, timeTillNextResubscription,
-        aTerminationCause.Format());
-    ReturnErrorOnFailure(apReadClient->ScheduleResubscription(timeTillNextResubscription, NullOptional, aTerminationCause == CHIP_ERROR_TIMEOUT));
+    auto error = [MTRError errorForCHIPErrorCode:aTerminationCause];
+    auto delayMs = @(apReadClient->ComputeTimeTillNextSubscription());
+    CallResubscriptionScheduledHandler(error, delayMs);
+    return CHIP_NO_ERROR;
+}
 
-    // Not as good a place to increment as when resubscription timer fires, but as is, this should be as good, because OnResubscriptionNeeded is only called from ReadClient's Close() while Idle, and nothing should cause this to happen
-    mResubscriptionNumRetries++;
-
+void MTRBaseSubscriptionCallback::CallResubscriptionScheduledHandler(NSError * error, NSNumber * resubscriptionDelay)
+{
     if (mResubscriptionCallback != nil) {
         auto callback = mResubscriptionCallback;
-        auto error = [MTRError errorForCHIPErrorCode:aTerminationCause];
-        auto delayMs = @(apReadClient->ComputeTimeTillNextSubscription());
-        callback(error, delayMs);
+        callback(error, resubscriptionDelay);
     }
-    return CHIP_NO_ERROR;
 }
 
 void MTRBaseSubscriptionCallback::OnUnsolicitedMessageFromPublisher(ReadClient *)
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -980,7 +980,7 @@ static NSString * const sAttributesKey = @"attributes";
     // Bit 2 Ethernet
     uint32_t networkCommissioningClusterFeatureMapValue = static_cast<uint32_t>(networkCommissioningClusterFeatureMapValueNumber.unsignedLongValue);
 
-    return (networkCommissioningClusterFeatureMapValue & (1 << 1));
+    return (networkCommissioningClusterFeatureMapValue & (1 << 1)) ? YES : NO;
 }
 
 - (void)_clearSubscriptionPoolWork
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -940,7 +940,7 @@ static NSString * const sAttributesKey = @"attributes";
 
 // This method is used for signaling whether to use the subscription pool. This functions as
 // a heuristic for whether to throttle subscriptions to the device via a pool of subscriptions.
-// If products appear that have both Thread and Wifi enabled but is primarily on wifi, this
+// If products appear that have both Thread and Wifi enabled but are primarily on wifi, this
 // method will need to be updated to reflect that.
 - (BOOL)_deviceUsesThread
 {
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -979,9 +979,6 @@ static NSString * const sAttributesKey = @"attributes";
         return NO;
     }
 
-    // Bit 0 WiFi
-    // Bit 1 Thread
-    // Bit 2 Ethernet
     uint32_t networkCommissioningClusterFeatureMapValue = static_cast<uint32_t>(networkCommissioningClusterFeatureMapValueNumber.unsignedLongValue);
 
     return (networkCommissioningClusterFeatureMapValue & MTRNetworkCommissioningFeatureThreadNetworkInterface) != 0 ? YES : NO;
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -42,6 +42,7 @@
 #include "lib/core/CHIPError.h"
 #include "lib/core/DataModelTypes.h"
 #include <app/ConcreteAttributePath.h>
+#include <lib/support/FibonacciUtils.h>
 
 #include <app/AttributePathParams.h>
 #include <app/BufferedReadCallback.h>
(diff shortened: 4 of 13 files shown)


src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -919,9 +938,79 @@ - (void)_handleSubscriptionError:(NSError *)error
...
+// This method is used for signaling whether to use the subscription pool. This functions as
+// a heuristic for whether to throttle subscriptions to the device via a pool of subscriptions.
+// If products appear that have both Thread and Wifi enabled but is primarily on wifi, this
+// method will need to be updated to reflect that.
+- (BOOL)_deviceUsesThread
 {
-    std::lock_guard lock(_lock);
+    os_unfair_lock_assert_owner(&self->_lock);
+
+    // Device is thread-enabled if there is a Thread Network Diagnostics cluster on endpoint 0
```
bzbarsky-apple, line 949: Thread Network Diagnostics is optional, no?

What we should really be doing is checking for the FeatureMap of the network commissioning cluster on the root endpoint, I would think.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -919,9 +938,79 @@ - (void)_handleSubscriptionError:(NSError *)error
...
+{
+    os_unfair_lock_assert_owner(&self->_lock);
+    MTRAsyncWorkCompletionBlock completion = self->_subscriptionPoolWorkCompletionBlock;
+    if (completion) {
+        self->_subscriptionPoolWorkCompletionBlock = nil;
+        completion(MTRAsyncWorkComplete);
+    }
+}
+
+- (void)_scheduleSubscriptionPoolWork:(void (^)(void))workBlock inNanoseconds:(int64_t)inNanoseconds description:(NSString *)description
```
bzbarsky-apple, line 973: Any reason to not write that as `(dispatch_block_t)workBlock`?

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -919,9 +938,79 @@ - (void)_handleSubscriptionError:(NSError *)error
...
+{
+    os_unfair_lock_assert_owner(&self->_lock);
+
+    // Sanity check we are not scheduling for this device multiple times in the pool
+    if (_subscriptionPoolWorkCompletionBlock) {
+        MTR_LOG_DEFAULT("%@ already scheduled in subscription pool for this device - ignoring: %@", self, description);
```
bzbarsky-apple, line 979: Should this be an error log?

src/darwin/Framework/CHIP/MTRDeviceControllerParameters.h
```diff
@@ -77,6 +77,11 @@ MTR_AVAILABLE(ios(17.6), macos(14.6), watchos(10.6), tvos(17.6))
  */
 - (void)setOTAProviderDelegate:(id<MTROTAProviderDelegate>)otaProviderDelegate queue:(dispatch_queue_t)queue;
 
+/**
+ * Sets the maximum subscriptions allowed for devices on Thread. This defaults to 3.
```
bzbarsky-apple, line 81: 1. This makes it sound like we can only subscribe to at most 3 Thread devices.  Needs to be rephrased to make it clear what this is really controlling.
2. The behavior of setting this to 0 needs to be documented.
3. Do we want to make this thread-specific?  Or make it more generic in some way where this is the cap on the parallel subscriptions on low-bandwidth links and we get to decide internally that "low-bandwidth" means "Thread"?

src/darwin/Framework/CHIP/MTRDeviceControllerParameters.h
```diff
@@ -77,6 +77,11 @@ MTR_AVAILABLE(ios(17.6), macos(14.6), watchos(10.6), tvos(17.6))
...
+ */
+@property (nonatomic, assign) NSUInteger concurrentSubscriptionsAllowedOnThread;
```
bzbarsky-apple, line 83:
```suggestion
@property (nonatomic, assign) NSUInteger concurrentSubscriptionsAllowedOnThread MTR_NEWLY_AVAILABLE;
```

src/darwin/Framework/CHIP/MTRBaseSubscriptionCallback.h
```diff
@@ -147,6 +144,10 @@ class MTRBaseSubscriptionCallback : public chip::app::ClusterStateCache::Callbac
     NSMutableArray * _Nullable mAttributeReports = nil;
     NSMutableArray * _Nullable mEventReports = nil;
 
+    void CallResubscriptionScheduledHandler(NSError * error, NSNumber * resubscriptionDelay);
+    // Copied from ReadClient and customized for MTRDevice resubscription time reset
+    uint32_t ComputeTimeTillNextSubscription();
```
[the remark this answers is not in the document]
jtung-apple: Will remove. This was leftover while I was shuffling things around.

src/darwin/Framework/CHIP/MTRDevice.mm
woody-apple: Not for thread devices, no.

woody-apple: In the spec it is required for Thread devices, same as WiFi Diagnostics for Wi-Fi Devices.

src/darwin/Framework/CHIP/MTRDeviceControllerParameters.h
jtung-apple: Will update the comments to be clearer and make it clamp to minimum of 1.

src/darwin/Framework/CHIP/MTRDevice.mm
bzbarsky-apple: Where is it required?  The conformance for it on Root Node device type is `[Thread]`, which means "optionally allowed if Thread".

src/darwin/Framework/CHIP/MTRDeviceControllerParameters.h
bzbarsky-apple: I guess the question is whether we want a "use whatever the default is, even if it changes later" value.  Right now 0 is that value, effectively....  Which is fine, but that should just be documented.  At least if we want to support it.

src/darwin/Framework/CHIP/MTRDevice.mm
LuDuda: I do agree. Thread Diagnostic cluster is optional, we should not check whether the device operates on Thread based on this cluster.

woody-apple: My bad! I read it wrong, Jeff is moving it to another thing.

src/darwin/Framework/CHIP/MTRDeviceControllerParameters.h
jtung-apple: I've changed it so that the parameter object initializes it to a default value. And if it's forcibly changed to 0, the real value is clamped to 1, as documented.

src/darwin/Framework/CHIP/MTRBaseSubscriptionCallback.mm
```diff
@@ -125,57 +125,29 @@
...
+    ReturnErrorOnFailure(err);
 
-    if (maxWaitTimeInMsec != 0) {
-        minWaitTimeInMsec = (CHIP_RESUBSCRIBE_MIN_WAIT_TIME_INTERVAL_PERCENT_PER_STEP * maxWaitTimeInMsec) / 100;
-        waitTimeInMsec = minWaitTimeInMsec + (Crypto::GetRandU32() % (maxWaitTimeInMsec - minWaitTimeInMsec));
-    }
-
-    return waitTimeInMsec;
+    auto error = [MTRError errorForCHIPErrorCode:aTerminationCause];
+    auto delayMs = @(apReadClient->ComputeTimeTillNextSubscription());
```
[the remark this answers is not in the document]
jtung-apple: I'm not sure it matters right now. We could reexamine this later?

src/darwin/Framework/CHIP/MTRDevice.mm
jtung-apple: I've changed this logic to look at NetworkCommissioning cluster feature map, and look for the thread bit.

src/darwin/Framework/CHIP/MTRBaseSubscriptionCallback.mm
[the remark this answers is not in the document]
jtung-apple: As far as I understand, as is this wouldn't cause an issue.

woody-apple (approved):

src/darwin/Framework/CHIP/MTRDeviceControllerStartupParams.mm
```diff
@@ -254,6 +254,8 @@ - (instancetype)_initInternal
 }
 @end
 
+constexpr NSUInteger kDefaultConcurrentSubscriptionPoolSize = 3;
```
woody-apple, line 257:
```suggestion
constexpr NSUInteger kDefaultConcurrentSubscriptionPoolSize = 100;
```

woody-apple (approved):

.github/workflows/darwin.yaml
```diff
@@ -115,6 +115,11 @@ jobs:
                   echo "This is a simple log" > /tmp/darwin/framework-tests/end_user_support_log.txt
                   ../../../out/debug/chip-all-clusters-app --interface-id -1 --end_user_support_log /tmp/darwin/framework-tests/end_user_support_log.txt > >(tee /tmp/darwin/framework-tests/all-cluster-app.log) 2> >(tee /tmp/darwin/framework-tests/all-cluster-app-err.log >&2) &
                   ../../../out/debug/chip-all-clusters-app --interface-id -1 --dac_provider ../../../credentials/development/commissioner_dut/struct_cd_origin_pid_vid_correct/test_case_vector.json --product-id 32768 --discriminator 3839 --secured-device-port 5539 --KVS /tmp/chip-all-clusters-app-kvs2 > >(tee /tmp/darwin/framework-tests/all-cluster-app-origin-vid.log) 2> >(tee /tmp/darwin/framework-tests/all-cluster-app-origin-vid-err.log >&2) &
+                  ../../../out/debug/chip-all-clusters-app --interface-id -1 --discriminator 101 --passcode 1001 --KVS /tmp/chip-all-clusters-app-kvs101 --secured-device-port 5531 &
```
bzbarsky-apple, line 118: For what it's worth, you don't really need different passcodes, but I guess that's a sunk cost at this point since you've generated the relevant QR codes...

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -940,43 +950,66 @@ - (void)_handleSubscriptionError:(NSError *)error
...
     os_unfair_lock_assert_owner(&self->_lock);
 
-    // Device is thread-enabled if there is a Thread Network Diagnostics cluster on endpoint 0
-    for (MTRClusterPath * path in [self _knownClusters]) {
-        if (path.endpoint.unsignedShortValue != 0) {
-            continue;
+#ifdef DEBUG
+    id testDelegate = _weakDelegate.strongObject;
+    if (testDelegate) {
+        if ([testDelegate respondsToSelector:@selector(unitTestPretendThreadEnabled:)]) {
```
bzbarsky-apple, line 962: Might be worth a comment explaining why it's OK to not do this on the delegate queue.  Especially because other unit test callbacks we _do_ go out of our way to call on that queue.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -940,43 +950,66 @@ - (void)_handleSubscriptionError:(NSError *)error
...
+                return YES;
+            }
         }
+    }
+#endif
 
-        if (path.cluster.unsignedLongValue == MTRClusterIDTypeThreadNetworkDiagnosticsID) {
-            return YES;
-        }
+    MTRClusterPath * networkCommissioningClusterPath = [MTRClusterPath clusterPathWithEndpointID:@(0) clusterID:@(MTRClusterIDTypeNetworkCommissioningID)];
```
bzbarsky-apple, line 970:
```suggestion
    MTRClusterPath * networkCommissioningClusterPath = [MTRClusterPath clusterPathWithEndpointID:@(kRootEndpointId) clusterID:@(MTRClusterIDTypeNetworkCommissioningID)];
```

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -940,43 +950,66 @@ - (void)_handleSubscriptionError:(NSError *)error
...
+    MTRDeviceClusterData * networkCommissioningClusterData = [self _clusterDataForPath:networkCommissioningClusterPath];
+    NSNumber * networkCommissioningClusterFeatureMapValueNumber = networkCommissioningClusterData.attributes[@(MTRClusterGlobalAttributeFeatureMapID)][MTRValueKey];
```
bzbarsky-apple, line 972: Will this do the right things when `networkCommissioningClusterData` is nil or `networkCommissioningClusterData.attributes[@(MTRClusterGlobalAttributeFeatureMapID)]` is nil?  That is, when we don't have the information yet?

Seems like at best it will make `networkCommissioningClusterFeatureMapValueNumber` nil which will cause `isKindOfClass` to be false, and then we get the MTR_LOG_ERROR... but that's not really an error condition, right?

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -940,43 +950,66 @@ - (void)_handleSubscriptionError:(NSError *)error
...
+        return NO;
     }
 
-    return NO;
+    // Bit 0 WiFi
+    // Bit 1 Thread
+    // Bit 2 Ethernet
+    uint32_t networkCommissioningClusterFeatureMapValue = static_cast<uint32_t>(networkCommissioningClusterFeatureMapValueNumber.unsignedLongValue);
+
+    return (networkCommissioningClusterFeatureMapValue & (1 << 1)) ? YES : NO;
```
bzbarsky-apple, line 983: We should be using `MTRNetworkCommissioningFeatureThreadNetworkInterface` here, I would think.

src/darwin/Framework/CHIP/MTRDeviceControllerParameters.h
```diff
@@ -79,8 +79,10 @@ MTR_AVAILABLE(ios(17.6), macos(14.6), watchos(10.6), tvos(17.6))
 
 /**
  * Sets the maximum subscriptions allowed for devices on Thread. This defaults to 3.
+ *
+ * If this value is 0, the maximum subscriptinos allowed will be set to 1.
```
bzbarsky-apple, line 83:
```suggestion
 * If this value is 0, the maximum subscriptions allowed will be set to 1.
```
But again: I feel like we should have an explicit way to signal "use your default" or "use your best guess" or whatever...

src/darwin/Framework/CHIP/MTRDeviceControllerParameters.h
```diff
@@ -79,8 +79,10 @@ MTR_AVAILABLE(ios(17.6), macos(14.6), watchos(10.6), tvos(17.6))
 
 /**
  * Sets the maximum subscriptions allowed for devices on Thread. This defaults to 3.
```
bzbarsky-apple, line 81: Again:

> This makes it sound like we can only subscribe to at most 3 Thread devices. Needs to be rephrased to make it clear what this is really controlling.

src/darwin/Framework/CHIPTests/MTRPerControllerStorageTests.m
```diff
@@ -2004,4 +2010,147 @@ - (void)testControllerServer
...
+    NSNumber * nodeID = @(555);
+    NSNumber * fabricID = @(555);
+
+    NSError * error;
+
+    NSUserDefaults * defaults = [[NSUserDefaults alloc] initWithSuiteName:kLocalTestUserDefaultDomain];
+    NSNumber * subscriptionPoolSizeOverrideOriginalValue = [defaults objectForKey:kLocalTestUserDefaultSubscriptionPoolSizeOverrideKey];
+
+    // Test DeviceController with a Subscription pool of 2
+    [defaults setInteger:subscriptionPoolSize forKey:kLocalTestUserDefaultSubscriptionPoolSizeOverrideKey];
```
bzbarsky-apple, line 2041: So the big problem here is that if this test ever gets interrupted now that machine is in a bad state forever, with that override stored, unless the user figures out that's a problem and removes it.   We already have this problem with the storage stuff, and it's been causing a lot of problems and wasted time in practice.  :(

I would really prefer it if the tests did some in-memory-only override.  Followup is OK for this, but needs to happen ASAP before we get people's machines into bad states.

src/darwin/Framework/CHIPTests/MTRPerControllerStorageTests.m
```diff
@@ -2004,4 +2010,147 @@ - (void)testControllerServer
...
+
+    NSNumber * nodeID = @(555);
+    NSNumber * fabricID = @(555);
+
+    NSError * error;
+
+    NSUserDefaults * defaults = [[NSUserDefaults alloc] initWithSuiteName:kLocalTestUserDefaultDomain];
+    NSNumber * subscriptionPoolSizeOverrideOriginalValue = [defaults objectForKey:kLocalTestUserDefaultSubscriptionPoolSizeOverrideKey];
+
+    // Test DeviceController with a Subscription pool of 2
```
bzbarsky-apple, line 2040: Where is this "2" coming from?  We are testing with whatever size was provided.

src/darwin/Framework/CHIPTests/MTRPerControllerStorageTests.m
```diff
@@ -2004,4 +2010,147 @@ - (void)testControllerServer
...
+        delegate.onSubscriptionPoolWorkComplete = ^{
+            // Stop counting subscribing right before calling work item completion
+            os_unfair_lock_lock(&counterLock);
+            subscriptionRunningCount--;
+            os_unfair_lock_unlock(&counterLock);
+        };
+        __weak __auto_type weakDelegate = delegate;
+        delegate.onReportEnd = ^{
+            [subscriptionExpectations[deviceID] fulfill];
+            // reset callback so expectation not fulfilled twice, given the run time of this can be long due to subscription pool
```
bzbarsky-apple, line 2118: I guess we don't have an actual "subscription established" callback, right?

src/darwin/Framework/CHIPTests/MTRPerControllerStorageTests.m
```diff
@@ -2004,4 +2010,147 @@ - (void)testControllerServer
...
+        __auto_type * device = [MTRDevice deviceWithNodeID:deviceID controller:controller];
+        [device setDelegate:deviceDelegates[deviceID] queue:queue];
+    }
+
+    // Make the wait time depend on pool size and device count (can expand number of devices in the future)
+    [self waitForExpectations:subscriptionExpectations.allValues timeout:(kSubscriptionPoolBaseTimeoutInSeconds * orderedDeviceIDs.count / subscriptionPoolSize)];
+
+    XCTAssertEqual(subscriptionDequeueCount, orderedDeviceIDs.count);
+
+    // Reset our commissionee.
```
bzbarsky-apple, line 2134:
```suggestion
    // Reset our commissionees.
```

src/darwin/Framework/CHIPTests/MTRPerControllerStorageTests.m
```diff
@@ -2004,4 +2010,147 @@ - (void)testControllerServer
...
+    for (NSNumber * deviceID in deviceOnboardingPayloads) {
```
bzbarsky-apple, line 2135:
```suggestion
    for (NSNumber * deviceID in orderedDeviceIDs) {
```

src/darwin/Framework/CHIP/MTRDeviceControllerParameters.h
jtung-apple: When the parameters object is created, this value is set to the default value (of 3).

src/darwin/Framework/CHIP/MTRDeviceControllerParameters.h
jtung-apple: Ah I see what you mean. I'll add clarification.

src/darwin/Framework/CHIPTests/MTRPerControllerStorageTests.m
jtung-apple: Will fix this comment

src/darwin/Framework/CHIP/MTRDevice.mm
woody-apple, line 983:
```suggestion
    return (networkCommissioningClusterFeatureMapValue & MTRNetworkCommissioningFeatureThreadNetworkInterface) != 0 ? YES : NO;
```

src/darwin/Framework/CHIP/MTRDevice.mm
woody-apple, line 972:
```suggestion
    NSNumber * networkCommissioningClusterFeatureMapValueNumber = networkCommissioningClusterData.attributes[@(MTRClusterGlobalAttributeFeatureMapID)][MTRValueKey];
    
    if ( networkCommissioningClusterFeatureMapValueNumber == nil )
        return NO;
```

src/darwin/Framework/CHIP/MTRDevice.mm
woody-apple, line 962:
```suggestion
        // Note: This is a hack to allow our unit tests to test the subscription pooling behavior we have implemented for thread, so we mock devices to be a thread device
        if ([testDelegate respondsToSelector:@selector(unitTestPretendThreadEnabled:)]) {
```

.github/workflows/darwin.yaml
woody-apple: Resolving, doesn't really affect anything

src/darwin/Framework/CHIP/MTRDeviceControllerParameters.h
woody-apple:
```suggestion
/**
 * Sets the maximum simultaneous subscriptions allowed to be established at one time for devices on Thread. This defaults to a large number.
```

woody-apple (approved):

src/darwin/Framework/CHIPTests/MTRPerControllerStorageTests.m
woody-apple, line 2040:
```suggestion
    // Test DeviceController with a Subscription pool
```

src/darwin/Framework/CHIP/MTRDeviceControllerParameters.h
```diff
@@ -77,6 +77,13 @@ MTR_AVAILABLE(ios(17.6), macos(14.6), watchos(10.6), tvos(17.6))
  */
 - (void)setOTAProviderDelegate:(id<MTROTAProviderDelegate>)otaProviderDelegate queue:(dispatch_queue_t)queue;
 
+/**
+ * Sets the maximum subscriptions allowed for devices on Thread. This defaults to 3.
```
woody-apple, line 81:
```suggestion
 * Sets the maximum simultaneous subscriptions allowed to be established at one time for devices on Thread. This defaults to a large number.
```

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -919,9 +948,116 @@ - (void)_handleSubscriptionError:(NSError *)error
...
+    if (networkCommissioningClusterFeatureMapValueNumber == nil)
+        return NO;
+    if (![networkCommissioningClusterFeatureMapValueNumber isKindOfClass:[NSNumber class]]) {
+        MTR_LOG_ERROR("%@ Unexpected NetworkCommissioning FeatureMap value %@", self, networkCommissioningClusterFeatureMapValueNumber);
+        return NO;
+    }
+
+    // Bit 0 WiFi
+    // Bit 1 Thread
+    // Bit 2 Ethernet
```
bzbarsky-apple, line 984:
```suggestion
```

src/darwin/Framework/CHIP/MTRDeviceControllerParameters.h
```diff
@@ -77,6 +77,13 @@ MTR_AVAILABLE(ios(17.6), macos(14.6), watchos(10.6), tvos(17.6))
  */
 - (void)setOTAProviderDelegate:(id<MTROTAProviderDelegate>)otaProviderDelegate queue:(dispatch_queue_t)queue;
 
+/**
+ * Sets the maximum simultaneous subscriptions allowed to be established at one time for devices on Thread. This defaults to a large number.
```
bzbarsky-apple, line 81:
```suggestion
 * Sets the maximum simultaneous subscription establishments that can be happening
 * at one time for devices on Thread. This defaults to a large number.
```

src/darwin/Framework/CHIP/MTRDeviceControllerParameters.h
```diff
@@ -77,6 +77,13 @@ MTR_AVAILABLE(ios(17.6), macos(14.6), watchos(10.6), tvos(17.6))
...
+ *
+ * If this value is 0, the maximum subscriptions allowed will be set to 1.
```
bzbarsky-apple, line 83:
```suggestion
 * If this value is 0, the maximum subscription establishments allowed at a time will be set to 1.
```

src/darwin/Framework/CHIP/MTRDeviceControllerParameters.h
```diff
@@ -77,6 +77,13 @@ MTR_AVAILABLE(ios(17.6), macos(14.6), watchos(10.6), tvos(17.6))
...
+ */
+@property (nonatomic, assign) NSUInteger concurrentSubscriptionsAllowedOnThread MTR_NEWLY_AVAILABLE;
```
bzbarsky-apple, line 85:
```suggestion
@property (nonatomic, assign) NSUInteger concurrentSubscriptionEstablishmentsAllowedOnThread MTR_NEWLY_AVAILABLE;
```

bzbarsky-apple (approved):

PR #33479, opened 2024-05-16
Linux tv-casting-app v1.3 Commissioner-Generated passcode flow

pgregorr-amazon:
Linux tv-casting-app example app, implemented new Matter v1.3 Matter Casting User Directed Commissioning (UDC) feature commissioner-generated passcode flow.

**Change summary**

1.	Added UpdateCommissionableDataProvider() function to connectedhomeip/examples/tv-casting-app/tv-casting-common/core/CastingApp.cpp/h which updates the Provided in the DeviceLayer. Also updated the VerifyOrEstablishConnection() function call with the new connectionCallbacks parameter.
2.	Added a SetCommissionableDataProvider() function to  connectedhomeip/examples/tv-casting-app/tv-casting-common/support/AppParameters.h
3.	Created connectedhomeip/examples/tv-casting-app/tv-casting-common/core/ConnectionCallbacks.h to hold the two UDC callbacks which are now passed to VerifyOrEstablishConnection() function
4.	Updated /examples/tv-casting-app/tv-casting-common/core/IdentificationDeclarationOptions.h to set the CommissioneeInstanceName when building an IdentificationDeclaration message.
5.	Updated connectedhomeip/examples/tv-casting-app/tv-casting-common/core/CommissionerDeclarationHandler.h with SetCommissionerDeclarationCallback() to set the client defined callback. Also updated OnCommissionerDeclarationMessage() to close the commissioning window when a CommissionerDeclaration message is received. This allows us to regenerate the PAKE verifier before opening the next commissioning window and sending the next IdentificationDeclaration message to the commissioner. 
6.	Updated connectedhomeip/examples/tv-casting-app/tv-casting-common/core/CastingPlayer.cpp VerifyOrEstablishConnection() with the new ConnectionCallbacks parameter. VerifyOrEstablishConnection() now sets the CommissionerDeclarationCallback passed in ConnectionCallbacks by the client. Added error checking to confirm the UDC status. 
7.	Updated connectedhomeip/examples/tv-casting-app/tv-casting-common/core/CastingPlayer.cpp with ContinueConnecting() function to send a second IdentificationDeclaration to the commissioner once the user has entered the commissioner-generated passcode in the tv-casting-app UX.
8.	Updated connectedhomeip/examples/tv-casting-app/linux/simple-app-helper.cpp/h with CommissionerDeclarationCallback() to handle the commissioners messages. Added the “setcommissionerpasscode <passcode>” command which allows the user to enter the commissioner-generated passcode. This command also calls InitCommissionableDataProvider() and UpdateCommissionableDataProvider() to update the commissioning session’s PAKE verifier and then calls ContinueConnecting().
9.	Updated VerifyOrEstablishConnection() function with the new param in MatterCastingPlayer-JNI.cpp and MatterTvCastingBridge/MCCastingPlayer.mm
10.	Moved InitCommissionableDataProvider() from connectedhomeip/examples/tv-casting-app/linux/simple-app.cpp to simple-app-helper.cpp/h so that it can be reused.

**Testing**

Verified and tested locally with the Linux, Android and iOS tv-casting-app example mobile apps, and the Linux tv-app (CastingPlayer). With the Linux example tv-casting-app app, we are able to successfully commission using the commissioner generated passcode flow as follows:

**Usage**

1.	Start tv-app:                connectedhomeip % ./out/tv-app/chip-tv-app
2.	Start tv-casting-app:   connectedhomeip % ./out/tv-casting-app/chip-tv-casting-app
3.	In tv-casting-app:       > cast request 0 commissioner-generated-passcode
4.	In tv-app:                    > controller ux ok
5.	Tv-app displays the commissioner-generated-passcode in hex: Casting passcode: [0x00BC_614E] or 12345678 in decimal.
6.	In tv-casting-app:       > cast setcommissionerpasscode 12345678
7.	The logs display “simple-app-helper.cpp::ConnectionHandler(): Successfully connected to CastingPlayer (ID: 88665A3712045650)”

### DIFF
--- a/examples/tv-casting-app/android/App/app/src/main/jni/cpp/core/MatterCastingPlayer-JNI.cpp
+++ b/examples/tv-casting-app/android/App/app/src/main/jni/cpp/core/MatterCastingPlayer-JNI.cpp
@@ -101,6 +101,8 @@ JNI_METHOD(jobject, verifyOrEstablishConnection)
                             "MatterCastingPlayer-JNI::verifyOrEstablishConnection() ConnectCallback() Connected to Casting Player "
                             "with device ID: %s",
                             playerPtr->GetId());
+            // The Java jSuccessCallback is expecting a Void v callback parameter which translates to a nullptr. When calling the
+            // Java method from C++ via JNI, passing nullptr is equivalent to passing a Void object in Java.
             MatterCastingPlayerJNIMgr().mConnectionSuccessHandler.Handle(nullptr);
         }
         else
@@ -118,6 +120,7 @@ JNI_METHOD(jobject, verifyOrEstablishConnection)
     matter::casting::core::ConnectionCallbacks connectionCallbacks;
     connectionCallbacks.mOnConnectionComplete = connectCallback;
 
+    // TODO: Verify why commissioningWindowTimeoutSec is a "unsigned long long int" type. Seems too big.
     castingPlayer->VerifyOrEstablishConnection(connectionCallbacks,
                                                static_cast<unsigned long long int>(commissioningWindowTimeoutSec), idOptions);
     return support::convertMatterErrorFromCppToJava(CHIP_NO_ERROR);
--- a/examples/tv-casting-app/android/App/app/src/main/jni/cpp/core/MatterCastingPlayer-JNI.cpp
+++ b/examples/tv-casting-app/android/App/app/src/main/jni/cpp/core/MatterCastingPlayer-JNI.cpp
@@ -24,6 +24,7 @@
 #include "core/CastingApp.h"             // from tv-casting-common
 #include "core/CastingPlayer.h"          // from tv-casting-common
 #include "core/CastingPlayerDiscovery.h" // from tv-casting-common
+#include "core/ConnectionCallbacks.h"    // from tv-casting-common
 
 #include <app/clusters/bindings/BindingManager.h>
 #include <app/server/Server.h>
@@ -92,24 +93,33 @@ JNI_METHOD(jobject, verifyOrEstablishConnection)
     MatterCastingPlayerJNIMgr().mConnectionSuccessHandler.SetUp(env, jSuccessCallback);
     MatterCastingPlayerJNIMgr().mConnectionFailureHandler.SetUp(env, jFailureCallback);
 
-    // TODO: In the following PRs. Add optional CommissionerDeclarationHandler callback parameter.
-    castingPlayer->VerifyOrEstablishConnection(
-        [](CHIP_ERROR err, CastingPlayer * playerPtr) {
-            ChipLogProgress(AppServer, "MatterCastingPlayer-JNI::verifyOrEstablishConnection() ConnectCallback called");
-            if (err == CHIP_NO_ERROR)
-            {
-                ChipLogProgress(AppServer, "MatterCastingPlayer-JNI:: Connected to Casting Player with device ID: %s",
-                                playerPtr->GetId());
-                MatterCastingPlayerJNIMgr().mConnectionSuccessHandler.Handle(nullptr);
-            }
-            else
-            {
-                ChipLogError(AppServer, "MatterCastingPlayer-JNI:: ConnectCallback, connection error: %" CHIP_ERROR_FORMAT,
-                             err.Format());
-                MatterCastingPlayerJNIMgr().mConnectionFailureHandler.Handle(err);
-            }
-        },
-        static_cast<unsigned long long int>(commissioningWindowTimeoutSec), idOptions);
+    auto connectCallback = [](CHIP_ERROR err, CastingPlayer * playerPtr) {
+        ChipLogProgress(AppServer, "MatterCastingPlayer-JNI::verifyOrEstablishConnection() ConnectCallback()");
+        if (err == CHIP_NO_ERROR)
+        {
+            ChipLogProgress(AppServer,
+                            "MatterCastingPlayer-JNI::verifyOrEstablishConnection() ConnectCallback() Connected to Casting Player "
+                            "with device ID: %s",
+                            playerPtr->GetId());
+            MatterCastingPlayerJNIMgr().mConnectionSuccessHandler.Handle(nullptr);
+        }
+        else
+        {
+            ChipLogError(
+                AppServer,
+                "MatterCastingPlayer-JNI::verifyOrEstablishConnection() ConnectCallback() Connection error: %" CHIP_ERROR_FORMAT,
+                err.Format());
+            MatterCastingPlayerJNIMgr().mConnectionFailureHandler.Handle(err);
+        }
+    };
+
+    // TODO: In the following PRs. Add optional CommissionerDeclarationHandler callback parameter for the Commissioner-Generated
+    // passcode commissioning flow.
+    matter::casting::core::ConnectionCallbacks connectionCallbacks;
+    connectionCallbacks.mOnConnectionComplete = connectCallback;
+
+    castingPlayer->VerifyOrEstablishConnection(connectionCallbacks,
+                                               static_cast<unsigned long long int>(commissioningWindowTimeoutSec), idOptions);
     return support::convertMatterErrorFromCppToJava(CHIP_NO_ERROR);
 }
 
--- a/examples/tv-casting-app/linux/simple-app-helper.cpp
+++ b/examples/tv-casting-app/linux/simple-app-helper.cpp
@@ -35,7 +35,7 @@
 const uint16_t kDesiredEndpointVendorId = 65521;
 
 DiscoveryDelegateImpl * DiscoveryDelegateImpl::_discoveryDelegateImpl = nullptr;
-bool gAwaitingCommissionerPasscodeInput                                = false;
+bool gAwaitingCommissionerPasscodeInput                               = false;
 std::shared_ptr<matter::casting::core::CastingPlayer> targetCastingPlayer;
 
 DiscoveryDelegateImpl * DiscoveryDelegateImpl::GetInstance()
--- a/examples/tv-casting-app/linux/simple-app-helper.cpp
+++ b/examples/tv-casting-app/linux/simple-app-helper.cpp
@@ -192,7 +192,8 @@ CHIP_ERROR InitCommissionableDataProvider(LinuxCommissionableDataProvider & prov
         VerifyOrDie(TestOnlyCommissionableDataProvider.GetSetupPasscode(defaultTestPasscode) == CHIP_NO_ERROR);
 
         ChipLogError(Support,
-                     "*** WARNING: Using temporary passcode %u due to no neither --passcode or --spake2p-verifier-base64 "
+                     "InitCommissionableDataProvider() *** WARNING: Using temporary passcode %u due to no neither --passcode or "
+                     "--spake2p-verifier-base64 "
                      "given on command line. This is temporary and will be deprecated. Please update your scripts "
                      "to explicitly configure onboarding credentials. ***",
                      static_cast<unsigned>(defaultTestPasscode));
@@ -201,21 +202,43 @@ CHIP_ERROR InitCommissionableDataProvider(LinuxCommissionableDataProvider & prov
     }
     else
     {
-        // Passcode is 0, so will be ignored, and verifier will take over. Onboarding payload
-        // printed for debug will be invalid, but if the onboarding payload had been given
-        // properly to the commissioner later, PASE will succeed.
+        ChipLogError(Support,
+                     "InitCommissionableDataProvider() *** WARNING: Passcode is 0, so will be ignored, and verifier will take "
+                     "over. Onboarding payload printed for debug will be invalid, but if the onboarding payload had been given "
+                     "properly to the commissioner later, PASE will succeed. ***");
     }
 
-    // Default to the minimum PBKDF iterations (1000) for this example implementation. For TV devices and TV casting app production
+    // Default to the minimum PBKDF iterations (1,000) for this example implementation. For TV devices and TV casting app production
     // implementations, you should use a higher number of PBKDF iterations to enhance security. The default minimum iterations are
-    // not sufficient against brute-force and rainbow table attacks.
-    uint32_t spake2pIterationCount = chip::Crypto::kSpake2p_Min_PBKDF_Iterations;
-    if (options.spake2pIterations != 0)
+    // not sufficient against brute-force and rainbow table attacks. Increasing the number of iterations will increase the
+    // computational time required to derive the key. This can slow down the authentication process, especially on devices with
+    // limited processing power like a Raspberry Pi 4. For a production implementation, you should measure the actual performance on
+    // the target device.
+    uint32_t spake2pIterationCount =
+        chip::Crypto::kSpake2p_Min_PBKDF_Iterations; // 1,000 - Hypothetical key derivation time: ~20 milliseconds (ms).
+    // uint32_t spake2pIterationCount = chip::Crypto::kSpake2p_Max_PBKDF_Iterations; // 100,000 - Hypothetical key derivation time:
+    // ~2 seconds.
+    if (options.spake2pIterations == 1000)
     {
         spake2pIterationCount = options.spake2pIterations;
+        ChipLogError(Support,
+                     "InitCommissionableDataProvider() *** WARNING: PASE PBKDF iterations provided are the minimum allowable: %u. "
+                     "Increase for production use to enhance security. ***",
+                     static_cast<unsigned>(spake2pIterationCount));
     }
-    ChipLogError(Support, "PASE PBKDF iterations set to %u. Should be set higher for a production app implementation.",
-                 static_cast<unsigned>(spake2pIterationCount));
+    else if ((options.spake2pIterations > 1000))
+    {
+        spake2pIterationCount = options.spake2pIterations;
+        ChipLogProgress(Support, "InitCommissionableDataProvider() PASE PBKDF iterations set to: %u.",
+                        static_cast<unsigned>(spake2pIterationCount));
+    }
+    else
+    {
+        ChipLogError(Support,
+                     "InitCommissionableDataProvider() *** WARNING: PASE PBKDF iterations set to the minimum allowable: %u. "
+                     "Increase for production use to enhance security. ***",
+                     static_cast<unsigned>(spake2pIterationCount));
+    }
 
     return provider.Init(options.spake2pVerifier, options.spake2pSalt, spake2pIterationCount, setupPasscode,
                          options.payload.discriminator.GetLongValue());
@@ -278,7 +301,7 @@ void CommissionerDeclarationCallback(const chip::Transport::PeerAddress & source
     {
         ChipLogProgress(AppServer, "---- Awaiting user input ----");
         ChipLogProgress(AppServer, "Input the Commissioner-Generated passcode displayed on the CastingPlayer UX.");
-        ChipLogProgress(AppServer, "Input 1245678 to use the defualt passcode.");
+        ChipLogProgress(AppServer, "Input 1245678 to use the default passcode.");
         ChipLogProgress(AppServer, "Example:     cast setcommissionerpasscode 12345678");
         ChipLogProgress(AppServer, "---- Awaiting user input ----");
         gAwaitingCommissionerPasscodeInput = true;
--- a/examples/tv-casting-app/linux/simple-app-helper.cpp
+++ b/examples/tv-casting-app/linux/simple-app-helper.cpp
@@ -35,7 +35,7 @@
 const uint16_t kDesiredEndpointVendorId = 65521;
 
 DiscoveryDelegateImpl * DiscoveryDelegateImpl::_discoveryDelegateImpl = nullptr;
-bool awaitingCommissionerPasscodeInput                                = false;
+bool gAwaitingCommissionerPasscodeInput                                = false;
 std::shared_ptr<matter::casting::core::CastingPlayer> targetCastingPlayer;
 
 DiscoveryDelegateImpl * DiscoveryDelegateImpl::GetInstance()
@@ -206,13 +206,16 @@ CHIP_ERROR InitCommissionableDataProvider(LinuxCommissionableDataProvider & prov
         // properly to the commissioner later, PASE will succeed.
     }
 
-    // Default to minimum PBKDF iterations
+    // Default to the minimum PBKDF iterations (1000) for this example implementation. For TV devices and TV casting app production
+    // implementations, you should use a higher number of PBKDF iterations to enhance security. The default minimum iterations are
+    // not sufficient against brute-force and rainbow table attacks.
     uint32_t spake2pIterationCount = chip::Crypto::kSpake2p_Min_PBKDF_Iterations;
     if (options.spake2pIterations != 0)
     {
         spake2pIterationCount = options.spake2pIterations;
     }
-    ChipLogError(Support, "PASE PBKDF iterations set to %u", static_cast<unsigned>(spake2pIterationCount));
+    ChipLogError(Support, "PASE PBKDF iterations set to %u. Should be set higher for a production app implementation.",
+                 static_cast<unsigned>(spake2pIterationCount));
 
     return provider.Init(options.spake2pVerifier, options.spake2pSalt, spake2pIterationCount, setupPasscode,
                          options.payload.discriminator.GetLongValue());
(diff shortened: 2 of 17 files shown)


examples/tv-casting-app/linux/simple-app-helper.cpp
```diff
@@ -168,22 +175,72 @@ void SubscribeToMediaPlaybackCurrentState(matter::casting::memory::Strong<matter
...
+        options.payload.setUpPINCode = defaultTestPasscode;
+    }
+    else
+    {
+        // Passcode is 0, so will be ignored, and verifier will take over. Onboarding payload
+        // printed for debug will be invalid, but if the onboarding payload had been given
+        // properly to the commissioner later, PASE will succeed.
+    }
+
+    // Default to minimum PBKDF iterations
```
tcarmelveilleux, line 209: For TVs, this should be more if possible, as the minimum is quite easy to rainbow-table at almost no cost.

pgregorr-amazon: Hi. Right. Production implementations should set their PBKDF iteration count higher to enhance security. I'll add a note stating that.
I moved this InitCommissionableDataProvider() helper method form connectedhomeip/examples/tv-casting-app/linux/simple-app.cppto to this file without modifying the content. 
Thanks for pointing that out!

examples/tv-casting-app/linux/simple-app-helper.cpp
```diff
@@ -28,11 +29,14 @@
 #include <lib/support/CHIPMem.h>
 #include <lib/support/CodeUtils.h>
 #include <platform/CHIPDeviceLayer.h>
+#include <platform/TestOnlyCommissionableDataProvider.h>
 
 // VendorId of the Endpoint on the CastingPlayer that the CastingApp desires to interact with after connection
 const uint16_t kDesiredEndpointVendorId = 65521;
 
 DiscoveryDelegateImpl * DiscoveryDelegateImpl::_discoveryDelegateImpl = nullptr;
+bool awaitingCommissionerPasscodeInput                                = false;
```
andy31415, line 38: nit: given that these are globals, we may want to prefix them as `g` like `gAwaitingCommissionerPasscodeInput` to make code slightly easier to read

examples/tv-casting-app/linux/simple-app-helper.cpp
```diff
@@ -168,22 +175,72 @@ void SubscribeToMediaPlaybackCurrentState(matter::casting::memory::Strong<matter
...
+        // properly to the commissioner later, PASE will succeed.
+    }
+
+    // Default to minimum PBKDF iterations
+    uint32_t spake2pIterationCount = chip::Crypto::kSpake2p_Min_PBKDF_Iterations;
+    if (options.spake2pIterations != 0)
+    {
+        spake2pIterationCount = options.spake2pIterations;
+    }
+    ChipLogError(Support, "PASE PBKDF iterations set to %u", static_cast<unsigned>(spake2pIterationCount));
```
andy31415, line 215: Is this logging as error intentional? It seems like this should be debug or progress.... if intentional to "do not forget to use" maybe we should mark it as such.

This seems to be "happy-path" so logging as error feels inappropriate.

examples/tv-casting-app/tv-casting-common/core/CastingApp.cpp
```diff
@@ -85,45 +86,61 @@ CHIP_ERROR CastingApp::Initialize(const AppParameters & appParameters)
     return CHIP_NO_ERROR;
 }
 
+CHIP_ERROR CastingApp::UpdateCommissionableDataProvider(chip::DeviceLayer::CommissionableDataProvider * commissionableDataProvider)
+{
+    ChipLogProgress(Discovery, "CastingApp::UpdateCommissionableDataProvider()");
+    chip::DeviceLayer::SetCommissionableDataProvider(commissionableDataProvider);
+    CHIP_ERROR err = CHIP_NO_ERROR;
+    err            = mAppParameters->SetCommissionableDataProvider(commissionableDataProvider);
+    return err;
```
andy31415, line 95: This is overly long. Do:

```cpp
return mAppParameters->SetCommissionableDataProvider(commissionableDataProvider);
```

examples/tv-casting-app/tv-casting-common/core/CastingPlayer.cpp
```diff
@@ -29,7 +29,8 @@ namespace core {
 
 CastingPlayer * CastingPlayer::mTargetCastingPlayer = nullptr;
 
-void CastingPlayer::VerifyOrEstablishConnection(ConnectCallback onCompleted, unsigned long long int commissioningWindowTimeoutSec,
+void CastingPlayer::VerifyOrEstablishConnection(ConnectionCallbacks connectionCallbacks,
+                                                unsigned long long int commissioningWindowTimeoutSec,
```
andy31415, line 33: this should be an explicit size instead of this long size.... uint64_t (but again... why do we need so many seconds? I understand ms, but not seconds).

examples/tv-casting-app/tv-casting-common/core/CastingPlayer.h
```diff
@@ -119,22 +119,39 @@ class CastingPlayer : public std::enable_shared_from_this<CastingPlayer>
...
+     * point it is assumed that the following have occurred:
+     * 1. Client has handled the Commissioner's CommissionerDecelration message with PasscodeDialogDisplayed and
+     * CommissionerPasscode set to true.
+     * 2. Client prompted user to input Passcode from Commissioner.
+     * 3. Client has updated the commissioning session's PAKE verifier using the user input Passcode by updating the CastingApps
+     * CommissionableDataProvider
+     * (matter::casting::core::CastingApp::GetInstance()->UpdateCommissionableDataProvider(CommissionableDataProvider)).
+     *
+     * @param connectionCallbacks contains the ConnectCallback and CommissionerDeclarationCallback (Optional).
+     * @param commissioningWindowTimeoutSec (Optional) time (in sec) to keep the commissioning window open, if commissioning is
```
andy31415, line 149: who is the commissioningWindowTimeoutSec  marked as optional? It is not std::optional or chip::Optional. Is there some specific optional value? It just seems to have a default

examples/tv-casting-app/tv-casting-common/core/CastingPlayer.h
```diff
@@ -206,6 +223,11 @@ class CastingPlayer : public std::enable_shared_from_this<CastingPlayer>
     unsigned long long int mCommissioningWindowTimeoutSec = kCommissioningWindowTimeoutSec;
     ConnectCallback mOnCompleted                          = {};
 
+    /**
+     * @brief resets this CastingPlayer's state and calls mOnCompleted with the CHIP_ERROR.
```
andy31415, line 227: it also clears mOnCompleted ... not sure if this is obvious from the description. May want to expand this.

examples/tv-casting-app/tv-casting-common/core/CastingPlayer.cpp
```diff
@@ -130,15 +173,66 @@ void CastingPlayer::VerifyOrEstablishConnection(ConnectCallback onCompleted, uns
...
+    }
+}
+
+void CastingPlayer::resetState(CHIP_ERROR err)
+{
+    support::ChipDeviceEventHandler::SetUdcStatus(false);
+    mConnectionState               = CASTING_PLAYER_NOT_CONNECTED;
+    mCommissioningWindowTimeoutSec = kCommissioningWindowTimeoutSec;
+    mTargetCastingPlayer           = nullptr;
+    mOnCompleted(err, nullptr);
```
andy31415, line 232: should we check that mOnCompleted is not null already? Or is that fine since we would segfault on invalid usage anyway?

examples/tv-casting-app/tv-casting-common/core/ConnectionCallbacks.h
```diff
@@ -0,0 +1,62 @@
...
+ */
+using ConnectCallback = std::function<void(CHIP_ERROR err, CastingPlayer * castingPlayer)>;
+
+/**
+ * @brief Called when a Commissioner Declaration UDC message has been received.
+ * @param[in] source The source of the Commissioner Declaration message.
+ * @param[in] cd The Commissioner Declaration message.
+ */
+using CommissionerDeclarationCallback = std::function<void(const chip::Transport::PeerAddress & source,
+                                                           chip::Protocols::UserDirectedCommissioning::CommissionerDeclaration cd)>;
```
andy31415, line 42: should we pass cd as a const reference as well given that it is a struct ?

examples/tv-casting-app/tv-casting-common/core/ConnectionCallbacks.h
```diff
@@ -0,0 +1,62 @@
...
+namespace core {
+
+class CastingPlayer;
+
+/**
+ * @brief Called when the User Directed Commissioning (UDC) process succeeds or fails.
+ * @param[in] err For success, called back with CHIP_NO_ERROR. For failure, called back with an error.
+ * @param[in] castingPlayer For success, called back with a CastingPlayer *. For failure, called back with a nullptr.
+ */
+using ConnectCallback = std::function<void(CHIP_ERROR err, CastingPlayer * castingPlayer)>;
```
andy31415, line 34: not sure how popular, however it sounds like CHIP_ERROR and CastingPlayer are mutually exclusive, so I wonder if `std::variant<CHIP_ERROR, CastingPlayer *>` would make invalid state harder to represent (CHIP_NO_ERROR would still be invalid state, but at least less logic).

examples/tv-casting-app/tv-casting-common/core/ConnectionCallbacks.h
```diff
@@ -0,0 +1,62 @@
...
+ */
+using CommissionerDeclarationCallback = std::function<void(const chip::Transport::PeerAddress & source,
+                                                           chip::Protocols::UserDirectedCommissioning::CommissionerDeclaration cd)>;
+
+/**
+ * @brief A container class for User Directed Commissioning (UDC) callbacks.
+ */
+class ConnectionCallbacks
+{
+public:
```
andy31415, line 49: a class with all public members should be a `struct`.

examples/tv-casting-app/tv-casting-common/core/ConnectionCallbacks.h
```diff
@@ -0,0 +1,62 @@
...
+    ConnectCallback mOnConnectionComplete = nullptr;
+    /**
+     * The callback called when the Commissionee receives a CommissionerDeclaration message from the Commissioner.
+     */
+    CommissionerDeclarationCallback mCommissionerDeclarationCallback = nullptr;
+};
+
+}; // namespace core
+}; // namespace casting
+}; // namespace matter
```
andy31415, line 62: please remove extra `;` after namespaces

examples/tv-casting-app/android/App/app/src/main/jni/cpp/core/MatterCastingPlayer-JNI.cpp
```diff
@@ -92,24 +93,33 @@ JNI_METHOD(jobject, verifyOrEstablishConnection)
...
-        static_cast<unsigned long long int>(commissioningWindowTimeoutSec), idOptions);
+    auto connectCallback = [](CHIP_ERROR err, CastingPlayer * playerPtr) {
+        ChipLogProgress(AppServer, "MatterCastingPlayer-JNI::verifyOrEstablishConnection() ConnectCallback()");
+        if (err == CHIP_NO_ERROR)
+        {
+            ChipLogProgress(AppServer,
+                            "MatterCastingPlayer-JNI::verifyOrEstablishConnection() ConnectCallback() Connected to Casting Player "
+                            "with device ID: %s",
+                            playerPtr->GetId());
+            MatterCastingPlayerJNIMgr().mConnectionSuccessHandler.Handle(nullptr);
```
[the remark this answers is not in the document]
pgregorr-amazon: Yes, looks like nullptr is expected here since this callback on the Java side is defined as: 
public void handle(Void v). 

The code codes not compile with: MatterCastingPlayerJNIMgr().mConnectionSuccessHandler.Handle(/*empty/*);

I'll add a comment to clarify. 

In Java targetCastingPlayer.VerifyOrEstablishConnection() is called on the targetCastingPlayer so we don't need to call back with playerPtr. Calling back on the successCallback() is enough to know that we have connected to the target CastingPlayer. 

Next, I'll be working on the examples/tv-casting-app Android Commissioner-Generated passcode implementation so I'll be modifying this and the other Java related files some more. For now I just moved the lambda definition out of castingPlayer->VerifyOrEstablishConnection() since the first param is now ConnectionCallbacks.

examples/tv-casting-app/android/App/app/src/main/jni/cpp/core/MatterCastingPlayer-JNI.cpp
```diff
@@ -92,24 +93,33 @@ JNI_METHOD(jobject, verifyOrEstablishConnection)
...
+        }
+    };
+
+    // TODO: In the following PRs. Add optional CommissionerDeclarationHandler callback parameter for the Commissioner-Generated
+    // passcode commissioning flow.
+    matter::casting::core::ConnectionCallbacks connectionCallbacks;
+    connectionCallbacks.mOnConnectionComplete = connectCallback;
+
+    castingPlayer->VerifyOrEstablishConnection(connectionCallbacks,
+                                               static_cast<unsigned long long int>(commissioningWindowTimeoutSec), idOptions);
```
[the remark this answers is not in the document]
pgregorr-amazon: Yeah. I'll leave a TODO note and check with sharadb-amazon when he's back next week. In connectedhomeip/examples/tv-casting-app/tv-casting-common/core/CastingPlayer.h VerifyOrEstablishConnection() is expecting a unsigned long long int commissioningWindowTimeoutSec but that's definitely overkill. I would like to defer changing this for now since i simply moved the connect callback definition out of the VerifyOrEstablishConnection() call without changing the logic/types.

examples/tv-casting-app/tv-casting-common/core/CastingApp.cpp
pgregorr-amazon: Sounds good.

examples/tv-casting-app/tv-casting-common/core/CastingPlayer.cpp
pgregorr-amazon: Leaving a TODO for this.

examples/tv-casting-app/tv-casting-common/core/CastingPlayer.cpp
pgregorr-amazon: Adding check for null to avoid a potential segmentation fault since it's good practice. It is somewhat redundant since we check if the passed in connectionCallbacks.mOnConnectionComplete is not null above in:

    VerifyOrExit(
        connectionCallbacks.mOnConnectionComplete != nullptr,
        ChipLogError(AppServer, "CastingPlayer::ContinueConnecting() ConnectionCallbacks.mOnConnectionComplete was not provided"));

examples/tv-casting-app/tv-casting-common/core/CastingPlayer.h
pgregorr-amazon: It is optional because it has a default value if not passed in; like in VerifyOrEstablishConnection() above: 
unsigned long long int commissioningWindowTimeoutSec = kCommissioningWindowTimeoutSec,

defined in CastingPlayer.h as:
const unsigned long long int kCommissioningWindowTimeoutSec = 3 * 60; // 3 minutes
Going to figure out why sharadb-amazon made it so big.

examples/tv-casting-app/tv-casting-common/core/CastingPlayer.h
pgregorr-amazon: Sounds good.

examples/tv-casting-app/tv-casting-common/core/ConnectionCallbacks.h
pgregorr-amazon: This could be a future improvement, right now I'd like to avoid changing this since that would require changing the Linux, iOS and Android tv-casting-app example apps.

examples/tv-casting-app/tv-casting-common/core/ConnectionCallbacks.h
pgregorr-amazon: yes

examples/tv-casting-app/tv-casting-common/core/ConnectionCallbacks.h
pgregorr-amazon: right, thanks!

examples/tv-casting-app/tv-casting-common/core/ConnectionCallbacks.h
pgregorr-amazon: ok!

examples/tv-casting-app/linux/simple-app-helper.cpp
pgregorr-amazon: I believe so, since this is an example app.  For a production app implementation PBKDF iterations should be set higher to mitigate against against brute-force and rainbow table attacks. I moved this function from another file.

examples/tv-casting-app/linux/simple-app-helper.cpp
pgregorr-amazon: That will be good, changing.

examples/tv-casting-app/linux/simple-app-helper.cpp
```diff
@@ -168,22 +175,75 @@ void SubscribeToMediaPlaybackCurrentState(matter::casting::memory::Strong<matter
...
+                     "given on command line. This is temporary and will be deprecated. Please update your scripts "
+                     "to explicitly configure onboarding credentials. ***",
+                     static_cast<unsigned>(defaultTestPasscode));
+        setupPasscode.SetValue(defaultTestPasscode);
+        options.payload.setUpPINCode = defaultTestPasscode;
+    }
+    else
+    {
+        // Passcode is 0, so will be ignored, and verifier will take over. Onboarding payload
+        // printed for debug will be invalid, but if the onboarding payload had been given
```
chrisdecenzo, line 205: should we print a warning that onboarding payload printed for debug is invalid?

examples/tv-casting-app/linux/simple-app-helper.cpp
```diff
@@ -168,22 +175,75 @@ void SubscribeToMediaPlaybackCurrentState(matter::casting::memory::Strong<matter
...
+        // properly to the commissioner later, PASE will succeed.
+    }
+
+    // Default to the minimum PBKDF iterations (1000) for this example implementation. For TV devices and TV casting app production
```
chrisdecenzo, line 209: does increasing the number in the example have a downside? Seems like this is easy for a developer to miss.

examples/tv-casting-app/linux/simple-app-helper.cpp
```diff
@@ -201,7 +261,27 @@ void ConnectionHandler(CHIP_ERROR err, matter::casting::core::CastingPlayer * ca
...
+                                     chip::Protocols::UserDirectedCommissioning::CommissionerDeclaration cd)
+{
+    ChipLogProgress(AppServer,
+                    "simple-app-helper.cpp::CommissionerDeclarationCallback() called with CommissionerDeclaration message:");
+    cd.DebugLog();
+    if (cd.GetCommissionerPasscode())
+    {
+        ChipLogProgress(AppServer, "---- Awaiting user input ----");
+        ChipLogProgress(AppServer, "Input the Commissioner-Generated passcode displayed on the CastingPlayer UX.");
+        ChipLogProgress(AppServer, "Input 1245678 to use the defualt passcode.");
```
chrisdecenzo, line 281:
```suggestion
        ChipLogProgress(AppServer, "Input 1245678 to use the default passcode.");
```

chrisdecenzo (approved):

examples/tv-casting-app/tv-casting-common/core/ConnectionCallbacks.h
andy31415: remove the `public` bit as well once we switched to a struct

andy31415 (approved):
Adding approve for consensus and some general readability. I am insufficiently familiar with TV apps logic and leaving it up to TV maintainers to review logic here.

yunhanw-google (approved):

examples/tv-casting-app/linux/simple-app-helper.cpp
pgregorr-amazon: Increasing the number of iterations will increase the computational time required to derive the key. This can slow down the authentication process, especially on devices with limited processing power. For a TV app running on a Raspberry Pi 4, 1,000 Iterations might take ~20 ms and the max allowed 100,000 iterations might take ~2 seconds.

Ideally the developers should measure the actual performance on the target device and pick a number that works best for their use case.
I'll make it a clear *** WARNING: log and add some notes about performance.

examples/tv-casting-app/tv-casting-common/core/CastingPlayer.h
```diff
@@ -119,22 +119,39 @@ class CastingPlayer : public std::enable_shared_from_this<CastingPlayer>
...
+     * 3. Client has updated the commissioning session's PAKE verifier using the user input Passcode by updating the CastingApps
+     * CommissionableDataProvider
+     * (matter::casting::core::CastingApp::GetInstance()->UpdateCommissionableDataProvider(CommissionableDataProvider)).
+     *
+     * @param connectionCallbacks contains the ConnectCallback and CommissionerDeclarationCallback (Optional).
+     * @param commissioningWindowTimeoutSec (Optional) time (in sec) to keep the commissioning window open, if commissioning is
+     * required. Needs to be >= kCommissioningWindowTimeoutSec.
+     */
+    void ContinueConnecting(ConnectionCallbacks connectionCallbacks,
+                            unsigned long long int commissioningWindowTimeoutSec = kCommissioningWindowTimeoutSec);
```
sharadb-amazon, line 153: Why do we need to pass commissioningWindowTimeoutSec in ContinueConnecting() at all, it was already passed in to VerifyOrEstablishConnection. Similarly connectionCallbacks should not be a parameter here either.

examples/tv-casting-app/tv-casting-common/core/CommissionerDeclarationHandler.cpp
```diff
@@ -36,14 +38,36 @@ CommissionerDeclarationHandler * CommissionerDeclarationHandler::GetInstance()
 }
 
 // TODO: In the following PRs. Implement setHandler() for CommissionerDeclaration messages and expose messages to higher layers for
-// Linux, Android and iOS.
+// Linux(DONE), Android(pending) and iOS(pending).
```
sharadb-amazon, line 41: For future reference, task tracking commentary should just be added in the PR description.

examples/tv-casting-app/tv-casting-common/support/ChipDeviceEventHandler.cpp
```diff
@@ -105,17 +105,19 @@ void ChipDeviceEventHandler::HandleFailSafeTimerExpired()
     }
 
     // if UDC was NOT in progress (when the Fail-Safe timer expired), start UDC
-    ChipLogProgress(AppServer, "ChipDeviceEventHandler::HandleFailSafeTimerExpired when sUdcInProgress: %d, starting UDC",
+    ChipLogProgress(AppServer, "ChipDeviceEventHandler::HandleFailSafeTimerExpired() when sUdcInProgress: %d, starting UDC",
```
sharadb-amazon, line 108: It doesn't look like the new commissioning flow is implemented in the case of this HandleFailSafeTimerExpired() callback?

examples/tv-casting-app/tv-casting-common/core/CastingPlayer.h
pgregorr-amazon: For flexibility in their implementation, the client might want to specify a diferent timeout or callbacks for ContinueConnecting() depending on how they implement their UX. They might have transitioned to a different page in the UX so using a  new callback might be easier to use than the old one. In Java for example callbacks can be defined in the in the function call.

examples/tv-casting-app/tv-casting-common/support/ChipDeviceEventHandler.cpp
pgregorr-amazon: I realized that no changes we're necessary here since we need to call OpenBasicCommissioningWindow() for both Commissionee-Generated passcode (AccountLogin) and Commissioner-Generated passcode flows.

PR #33480, opened 2024-05-16
Add ServerCluster base class and Ember adapter

ksperling-apple:
This adds a `ServerCluster` base class that extends `AttributeAccessInterface` and `CommandHandlerInterface`, and automatically registers the handlers in `Init`.

Additionally, a `DECLARE_SERVER_CLUSTER` macro is provided that defines a global holder object with storage for `MATTER_DM_##NAME##_SERVER_ENDPOINT_COUNT` instances of the concrete `ServerCluster` sub-class, and defines the `emberAf...ClusterServerInitCallback` and `Matter...ClusterServerShutdownCallback` functions to initialize those cluster instances based on ZAP configuration.


src/app/BUILD.gn
```diff
@@ -328,6 +328,8 @@ static_library("app") {
     "OTAUserConsentCommon.h",
     "ReadHandler.cpp",
     "SafeAttributePersistenceProvider.h",
+    "ServerCluster.cpp",
+    "ServerCluster.h",
```
andy31415, line 332: This wraps AAI ... seems better to split it out of "app" ... maybe it should be tied to `attribute-access` instead. We should stop having "app" being a container of everything.

src/app/util/BUILD.gn
```diff
@@ -23,6 +23,7 @@ source_set("types") {
     "attribute-metadata.h",
     "attribute-storage-null-handling.h",
     "basic-types.h",
+    "ember-server-cluster.h",
```
andy31415, line 26: Types generally should not depend on generated code. This depends on `app-common/zap-generated/callback.h` so it seems it should not be here.

andy31415:
@ksperling-apple please provide a description on include dependencies for these things. We still have the ember and generated code dependencies very hard to untangle, so it looks like we need to have a solid plan on how things include each-other. GN does not seem to enforce `.h` include as well as one would want. It complans sometimes, but not always.

src/app/util/BUILD.gn
ksperling-apple: Yeah these dependencies are tricky... my goal is to have the ember adapters statically allocated based on the `MATTER_DM_*_SERVER_ENDPOINT_COUNT` without having to manually write "stub" files for each cluster.

A different possible approach would be to address this is in the zap templates directly, maybe by listing the clusters that are based on `ServerCluster` in config-data.yaml and then have the code generation do the right thing automatically.

PR #30392, opened 2023-11-09
Plumb Peer DNS-SD advertisement for TCP support into SessionSetup context

hnnajh:
Currently, the TCP support advertisement from the peer is not correctly consumed and stored. This needs to be parsed properly during operational session setup and stored in the session context(given that the TCP support advertisement is a bit field in the 1.3 spec). This would be used to set the PeerAddress type to TCP or otherwise, when attempting to establish CASE with the peer.

Fixes https://github.com/project-chip/connectedhomeip/issues/29539

PR #33481, opened 2024-05-16
 Plumb Peer DNS-SD advertisement for TCP support into SessionSetup context #30392 

### LINKED ISSUE #29539: Plumb Peer DNS-SD advertisement for TCP support into SessionSetup context
### Feature description

Currently, the TCP support advertisement from the peer is not correctly consumed and stored. This needs to be parsed properly during operational session setup and stored in the session context(given that the TCP support advertisement is a bit field in the 1.3 spec). This would be used to set the PeerAddress type to TCP or otherwise, when attempting to establish CASE with the peer.

### Platform

core

### Platform Version(s)

_No response_

### Anything else?

_No response_

hnnajh:
Currently, the TCP support advertisement from the peer is not correctly consumed and stored. This needs to be parsed properly during operational session setup and stored in the session context(given that the TCP support advertisement is a bit field in the 1.3 spec). This would be used to set the PeerAddress type to TCP or otherwise, when attempting to establish CASE with the peer.

Fixes https://github.com/project-chip/connectedhomeip/issues/29539


src/lib/address_resolve/AddressResolve_DefaultImpl.cpp
```diff
@@ -305,6 +306,8 @@ void Resolver::OnOperationalNodeResolved(const Dnssd::ResolvedNodeData & nodeDat
             }
 #endif
             result.address.SetIPAddress(nodeData.resolutionData.ipAddress[i]);
+            result.address.SetTransportType((result.supportsTcpClient || result.supportsTcpServer) ? Transport::Type::kTcp
```
bzbarsky-apple, line 309: This does not seem right.  Just because the other side supports TCP, doesn't mean we will in fact use TCP....

src/lib/address_resolve/tool.cpp
```diff
@@ -56,7 +56,8 @@ class PrintOutNodeListener : public chip::AddressResolve::NodeListener
         result.address.ToString(addr_string);
 
         ChipLogProgress(Discovery, "Resolve completed: %s", addr_string);
-        ChipLogProgress(Discovery, "   Supports TCP:                  %s", result.supportsTcp ? "YES" : "NO");
+        ChipLogProgress(Discovery, "   Supports TCP:                  %s",
+                        (result.supportsTcpServer || result.supportsTcpClient) ? "YES" : "NO");
```
bzbarsky-apple, line 60: Shouldn't this log the two booleans separately?

src/lib/dnssd/tests/TestTxtFields.cpp
```diff
@@ -646,25 +647,29 @@ void DiscoveredTxtFieldTcpSupport()
 
     // True
```
bzbarsky-apple, line 648: This comment no longer makes sense.

src/lib/dnssd/tests/TestTxtFields.cpp
```diff
@@ -646,25 +647,29 @@ void DiscoveredTxtFieldTcpSupport()
...
+    EXPECT_TRUE(nodeData.Get<NodeData>().supportsTcpClient);
+    EXPECT_FALSE(nodeData.Get<NodeData>().supportsTcpServer);
 
     // Test no other fields were populated
-    nodeData.Get<NodeData>().supportsTcp = false;
+    nodeData.Get<NodeData>().supportsTcpClient = false;
+    nodeData.Get<NodeData>().supportsTcpServer = false;
     EXPECT_TRUE(NodeDataIsEmpty(nodeData.Get<NodeData>()));
 
     // False
```
bzbarsky-apple, line 660: This comment no longer makes sense.

src/lib/dnssd/tests/TestTxtFields.cpp
```diff
@@ -646,25 +647,29 @@ void DiscoveredTxtFieldTcpSupport()
...
     strcpy(key, "T");
     strcpy(val, "0");
     FillNodeDataFromTxt(GetSpan(key), GetSpan(val), resolutionData);
-    EXPECT_TRUE(nodeData.Get<NodeData>().supportsTcp == false);
+    EXPECT_TRUE(nodeData.Get<NodeData>().supportsTcpServer == false);
+    EXPECT_TRUE(nodeData.Get<NodeData>().supportsTcpClient == false);
```
bzbarsky-apple, line 665: Should just be EXPECT_FALSE, no?

src/lib/dnssd/tests/TestTxtFields.cpp
```diff
@@ -646,25 +647,29 @@ void DiscoveredTxtFieldTcpSupport()
...
 
     // Invalid value, stil false
     strcpy(key, "T");
     strcpy(val, "asdf");
     FillNodeDataFromTxt(GetSpan(key), GetSpan(val), resolutionData);
-    EXPECT_TRUE(nodeData.Get<NodeData>().supportsTcp == false);
+    EXPECT_TRUE(nodeData.Get<NodeData>().supportsTcpClient == false);
+    EXPECT_TRUE(nodeData.Get<NodeData>().supportsTcpServer == false);
```
bzbarsky-apple, line 672: Should be EXPECT_FALSE.

src/lib/dnssd/tests/TestTxtFields.cpp
```diff
@@ -646,25 +647,29 @@ void DiscoveredTxtFieldTcpSupport()
...
 
     // Test no other fields were populated
-    nodeData.Get<NodeData>().supportsTcp = false;
+    nodeData.Get<NodeData>().supportsTcpClient = false;
+    nodeData.Get<NodeData>().supportsTcpServer = false;
     EXPECT_TRUE(NodeDataIsEmpty(nodeData.Get<NodeData>()));
 
     // False
     strcpy(key, "T");
     strcpy(val, "0");
```
bzbarsky-apple, line 662: Needs tests for T=2, T=4, T=6.

src/lib/dnssd/tests/TestTxtFields.cpp
```diff
@@ -983,23 +988,27 @@ void TxtFieldTcpSupport()
...
+    EXPECT_FALSE(nodeData.resolutionData.supportsTcpServer);
+    EXPECT_FALSE(nodeData.resolutionData.supportsTcpClient);
 
     // Test no other fields were populated
-    nodeData.resolutionData.supportsTcp = false;
+    nodeData.resolutionData.supportsTcpServer = false;
+    nodeData.resolutionData.supportsTcpClient = false;
     EXPECT_TRUE(NodeDataIsEmpty(nodeData));
 
     // False
```
bzbarsky-apple, line 999: Again, this comment makes no sense anymore.

src/lib/dnssd/tests/TestTxtFields.cpp
```diff
@@ -983,23 +988,27 @@ void TxtFieldTcpSupport()
...
     strcpy(key, "T");
     strcpy(val, "0");
     FillNodeDataFromTxt(GetSpan(key), GetSpan(val), nodeData.resolutionData);
-    EXPECT_EQ(nodeData.resolutionData.supportsTcp, false);
+    EXPECT_EQ(nodeData.resolutionData.supportsTcpServer, false);
+    EXPECT_EQ(nodeData.resolutionData.supportsTcpClient, false);
```
bzbarsky-apple, line 1004: EXPECT_FALSE?

src/lib/dnssd/tests/TestTxtFields.cpp
```diff
@@ -983,23 +988,27 @@ void TxtFieldTcpSupport()
...
 
     // Invalid value, stil false
     strcpy(key, "T");
     strcpy(val, "asdf");
     FillNodeDataFromTxt(GetSpan(key), GetSpan(val), nodeData.resolutionData);
-    EXPECT_EQ(nodeData.resolutionData.supportsTcp, false);
+    EXPECT_EQ(nodeData.resolutionData.supportsTcpClient, false);
+    EXPECT_EQ(nodeData.resolutionData.supportsTcpServer, false);
```
bzbarsky-apple, line 1011: EXPECT_FALSE?

src/lib/dnssd/tests/TestTxtFields.cpp
```diff
@@ -983,23 +988,27 @@ void TxtFieldTcpSupport()
...
 
     // Test no other fields were populated
-    nodeData.resolutionData.supportsTcp = false;
+    nodeData.resolutionData.supportsTcpServer = false;
+    nodeData.resolutionData.supportsTcpClient = false;
     EXPECT_TRUE(NodeDataIsEmpty(nodeData));
 
     // False
     strcpy(key, "T");
     strcpy(val, "0");
```
bzbarsky-apple, line 1001: Again, seems like we need more coverage of different values here in the test.

src/lib/shell/commands/Dns.cpp
```diff
@@ -51,7 +51,8 @@ class DnsShellResolverDelegate : public Dnssd::DiscoverNodeDelegate, public Addr
         result.address.ToString(addr_string);
 
         streamer_printf(streamer_get(), "Resolve completed: %s\r\n", addr_string);
-        streamer_printf(streamer_get(), "   Supports TCP:                  %s\r\n", result.supportsTcp ? "YES" : "NO");
+        streamer_printf(streamer_get(), "   Supports Client TCP:  %s\r\n", result.supportsTcpClient ? "YES" : "NO");
+        streamer_printf(streamer_get(), "   Supports Server TCP:  %s\r\n", result.supportsTcpServer ? "YES" : "NO");
```
bzbarsky-apple, line 55: Why are these talking about "Client TCP" and "Server TCP" while other logs talk about "TCP Client" and "TCP Server"?  Why the inconsistency?

bzbarsky-apple (changes requested):
Changes requested mainly for the stamping of the transport type on the address....

src/lib/address_resolve/AddressResolve_DefaultImpl.cpp
hnnajh: that's correct, that PR is not intended to gaurantee the usage of TCP, @pidarped  has other PRs that would depend on that plumbing to use TCP transport

src/lib/dnssd/tests/TestTxtFields.cpp
hnnajh: done

src/lib/dnssd/tests/TestTxtFields.cpp
hnnajh: done

src/lib/dnssd/tests/TestTxtFields.cpp
hnnajh: correct, fixed

src/lib/dnssd/tests/TestTxtFields.cpp
hnnajh: fixed

src/lib/dnssd/tests/TestTxtFields.cpp
hnnajh: added bunch of more tests cases to cover all the values

src/lib/dnssd/tests/TestTxtFields.cpp
hnnajh: done

src/lib/dnssd/tests/TestTxtFields.cpp
hnnajh: yes, done

src/lib/shell/commands/Dns.cpp
hnnajh: good point, fixed

src/lib/address_resolve/AddressResolve_DefaultImpl.cpp
pidarped: Yes, agree with @bzbarsky-apple. I think there is no need to set the address type here. As long as we are collecting the info that the peer supports TcpClient/Server, that can be used subsequently during session establishment phase. I also see that OperationalSessionSetup::UpdateDeviceData() takes the peerAddress and MRPConfig. I think we need to pass the ResolveResult so that OperationalSessionSetup can capture the TCP support bits for the peer and take the informed decision during EstablishConnection() to set the TransportType.

hnnajh: sounds fair, i modified the code to route the result object to EstablishConnection method where the decision is made to support large payload session with a specific client

pidarped (approved):
Approved with one minor comment.

src/app/OperationalSessionSetup.cpp
```diff
@@ -290,12 +291,14 @@ void OperationalSessionSetup::UpdateDeviceData(const Transport::PeerAddress & ad
...
-CHIP_ERROR OperationalSessionSetup::EstablishConnection(const ReliableMessageProtocolConfig & config)
+CHIP_ERROR OperationalSessionSetup::EstablishConnection(const ResolveResult & result)
 {
+    auto config = result.mrpRemoteConfig;
 #if INET_CONFIG_ENABLE_TCP_ENDPOINT
     // TODO: Combine LargePayload flag with DNS-SD advertisements from peer.
     // Issue #32348.
-    if (mTransportPayloadCapability == TransportPayloadCapability::kLargePayload)
+    if (mTransportPayloadCapability == TransportPayloadCapability::kLargePayload && result.supportsTcpServer &&
+        result.supportsTcpClient)
```
bzbarsky-apple, line 301: Why do we care whether it supports TCP client?  We're going to be the client; as long as it supports server we are fine, no?

src/app/OperationalSessionSetup.cpp
```diff
@@ -290,12 +291,14 @@ void OperationalSessionSetup::UpdateDeviceData(const Transport::PeerAddress & ad
     // Do not touch `this` instance anymore; it has been destroyed in DequeueConnectionCallbacks.
 }
 
-CHIP_ERROR OperationalSessionSetup::EstablishConnection(const ReliableMessageProtocolConfig & config)
+CHIP_ERROR OperationalSessionSetup::EstablishConnection(const ResolveResult & result)
 {
+    auto config = result.mrpRemoteConfig;
 #if INET_CONFIG_ENABLE_TCP_ENDPOINT
     // TODO: Combine LargePayload flag with DNS-SD advertisements from peer.
     // Issue #32348.
```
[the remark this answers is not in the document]
hnnajh: no, removed

src/app/OperationalSessionSetup.cpp
hnnajh: good point, updated

src/app/OperationalSessionSetup.cpp
```diff
@@ -292,12 +294,11 @@ void OperationalSessionSetup::UpdateDeviceData(const Transport::PeerAddress & ad
...
 
-CHIP_ERROR OperationalSessionSetup::EstablishConnection(const ReliableMessageProtocolConfig & config)
+CHIP_ERROR OperationalSessionSetup::EstablishConnection(const ResolveResult & result)
 {
+    auto & config = result.mrpRemoteConfig;
 #if INET_CONFIG_ENABLE_TCP_ENDPOINT
-    // TODO: Combine LargePayload flag with DNS-SD advertisements from peer.
-    // Issue #32348.
-    if (mTransportPayloadCapability == TransportPayloadCapability::kLargePayload)
+    if (mTransportPayloadCapability == TransportPayloadCapability::kLargePayload && result.supportsTcpServer)
```
bzbarsky-apple, line 301: If `result.supportsTcpServer` is false but `mTransportPayloadCapability == TransportPayloadCapability::kLargePayload`, shouldn't this error out?

hnnajh: done

src/app/OperationalSessionSetup.cpp
```diff
@@ -292,12 +294,17 @@ void OperationalSessionSetup::UpdateDeviceData(const Transport::PeerAddress & ad
...
 {
+    auto & config = result.mrpRemoteConfig;
 #if INET_CONFIG_ENABLE_TCP_ENDPOINT
-    // TODO: Combine LargePayload flag with DNS-SD advertisements from peer.
-    // Issue #32348.
-    if (mTransportPayloadCapability == TransportPayloadCapability::kLargePayload)
+    if (mTransportPayloadCapability == TransportPayloadCapability::kLargePayload && !result.supportsTcpServer)
+    {
+        // we should not set the large payload while the TCP support is not enabled
+        ChipLogError(Discovery, "LargePayload support is enabled while TCP server support is not enabled");
```
pidarped, line 304: Please change the error log to something like "LargePayload session requested but peer does not support TCP server" to clarify that one came down from the local app, while the other is a parameter of the peer.

src/app/OperationalSessionSetup.cpp
```diff
@@ -292,15 +294,23 @@ void OperationalSessionSetup::UpdateDeviceData(const Transport::PeerAddress & ad
...
-        mDeviceAddress.SetTransportType(chip::Transport::Type::kTcp);
+        if (result.supportsTcpServer)
+        {
+            // Set the transport type for carrying large payloads
+            mDeviceAddress.SetTransportType(chip::Transport::Type::kTcp);
+        }
+        else
+        {
+            // we should not set the large payload while the TCP support is not enabled
+            ChipLogError(Discovery, "LargePayload support is enabled while TCP server support is not enabled");
```
bzbarsky-apple, line 311: This log line needs to include some indication of the peer identity.  Otherwise it's impossible to tell what device it's talking about.

src/app/OperationalSessionSetup.cpp
hnnajh: done

src/app/OperationalSessionSetup.cpp
hnnajh: done

bzbarsky-apple (approved):

PR #33482, opened 2024-05-16
Revert using strncmp instead of strcmp to make the function robust and safe from buffer overruns. #33449

yufengwangca:
The previous fix in [PR #33449](https://github.com/project-chip/connectedhomeip/pull/33449) used strncmp instead of strcmp for equality checks. After further offline discussions, we concluded that this change was not well-considered and could introduce additional thread safety issues.

This PR reverts the previous fix and adds tests for the Nullable String attribute to ensure proper handling and validation.

### DIFF
--- a/src/app/tests/suites/TestCluster.yaml
+++ b/src/app/tests/suites/TestCluster.yaml
@@ -707,6 +707,20 @@ tests:
       response:
           value: 0
 
+    # Tests for Nullable String attribute
+
+    - label: "Write attribute NULLABLE_CHAR_STRING - Value starting with null"
+      command: "writeAttribute"
+      attribute: "nullable_char_string"
+      arguments:
+          value: "nulla"
+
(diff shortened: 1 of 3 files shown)


src/app/tests/suites/TestCluster.yaml
```diff
@@ -707,6 +707,20 @@ tests:
       response:
           value: 0
 
+    # Tests for Nullable String attribute
```
bzbarsky-apple, line 710: Put this in the existing section with tests for nullable_char_string?  That would help notice why the test is failing with this change.

bzbarsky-apple (approved):

arkq (approved):

bzbarsky-apple:
Just to be clear: the change being reverted does not "introduce additional thread safety issues".  It just replaces an equality test with a prefix test, which completely changes the logic.

PR #33488, opened 2024-05-16
[Python] Drop chip-device-ctrl

agners:
Already a while ago (with #23724) `chip-device-ctrl` got marked as unmaintained. And it previously broke and got fixed numerous times.

It currently is in a broken state again, leading to segfault on startup.

Get finally rid of it. Almost all features are covered by `chip-repl`.


docs/guides/python_chip_controller_building.md
```diff
@@ -255,429 +258,195 @@ the following stages:
...
+│   optionsOverride=0
+)
 ```
 
-Note: The format of the value is the same as the format of argument values for
-ZCL cluster commands.
+### `ReadAttribute(<nodeid>: int, [(<endpoint id>: int, Clusters.<cluster>.Attributes.<attribute>)])`
 
-**REPL Commands**
+Read the value of ZCL attribute. For example:
```
cecille, line 404:
```suggestion
Read the value of an attribute. For example:
```

docs/guides/python_chip_controller_building.md
```diff
@@ -255,429 +258,195 @@ the following stages:
...
-await devCtrl.WriteAttribute(1, [(1, Clusters.UnitTesting.Attributes.OctetString(value=b'123123\x00'))])
-await devCtrl.WriteAttribute(1, [(1, Clusters.UnitTesting.Attributes.CharString(value='233233'))])
+await devCtrl.ReadAttribute(1234, [(0, Clusters.BasicInformation.Attributes.VendorName)])
 ```
 
-### `zclsubscribe <Cluster> <Attribute> <Nodeid> <Endpoint> <MinInterval> <MaxInterval>`
+### `WriteAttribute(<nodeid>: int, [(<endpointid>: int, Clusters.<cluster>.Attributes.<attribute>(value=<attribute value>))])`
 
-Configure ZCL attribute reporting settings. For example:
+Write the value to a ZCL attribute. For example:
```
cecille, line 412:
```suggestion
Write a value to an attribute. For example:
```

docs/guides/python_chip_controller_building.md
```diff
@@ -255,429 +258,195 @@ the following stages:
...
 
 ```
-chip-device-ctrl > set-pairing-wifi-credential TESTSSID P455W4RD
+devCtrl.SetWiFiCredentials('TESTSSID', 'P455W4RD')
 ```
 
-**REPL Commands:** `devCtrl.SetWiFiCredentials('TESTSSID', 'P455W4RD')`
-
-### `connect -ip <address> <SetUpPinCode> [<nodeid>]`
+### `CommissionIP(<ip address>: str, <setup pin code>: int, <nodeid>: int)`
```
cecille, line 320: This function is marked as deprecated in the API.

docs/guides/python_chip_controller_building.md
```diff
@@ -255,429 +258,195 @@ the following stages:
...
-
-> Use the `zcl ? BasicInformation` command to list all available commands for
-> Basic Information Cluster.
->
 > In REPL, you can type `Clusters.BasicInformation.Attributes.` and then use the
 > TAB key.
 
 <hr>
 
 ## List of commands
```
cecille, line 299: We now have a full API description from the ChipDeviceCtrl documentation here (https://project-chip.github.io/connectedhomeip-doc/testing/ChipDeviceCtrlAPI.html) - it's generated from the function comments.

I think this documentation was written before some of the functions were updated, and before some of the new commissioning stuff landed, so there are a few things in here that are a bit out of date. WDYT about just getting rid of this section and making a note that the devCtrl is a device controller and you can use any of the normal functions, and then just linking over to that documentation?

docs/guides/python_chip_controller_building.md
tehampson, line 1: nit, if we do plan on keeping this file around with it's content, should we rename the file to reflect the content a little more. It seems like it has expanded greatly from building the python chip controller

docs/guides/python_chip_controller_building.md
tehampson: +1, this is replaced with CommissionOnNetwork.

agners: Ok, since it is currently marked deprecated, we should drop it from here. But I do disagree on the deprecation of this API (see discussion in #tsg-tt-python-repl). But that is a different discussion.

docs/guides/python_chip_controller_building.md
agners: The current name doesn't seem too far of to me, no?

There are two things: The Python CHIP Controller (library), which is essentially almost all the Python code which allows to be a CHIP Controller using Python. And then there is a (rather small) REPL built on-top of it. While the docs document mostly the API as it can be used from the REPL, I think just documenting how-to build the Python stuff makes sense.

tehampson (approved):

cecille (approved):

arkq (approved):

PR #33489, opened 2024-05-16
Darwin: remove `_handleAttributeReport` convenience method

kiel-apple:
Per previous PR discussion, force callers to explicitly declare whether or not the report is from a subscription.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -1367,11 +1367,6 @@ static NSString * const sLastInitialSubscribeLatencyKey = @"lastInitialSubscribe
     [self _reportAttributes:[self _getAttributesToReportWithReportedValues:attributeReport fromSubscription:isFromSubscription]];
 }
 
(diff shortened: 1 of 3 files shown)


src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -1188,7 +1183,7 @@ - (void)unitTestInjectAttributeReport:(NSArray<NSDictionary<NSString *, id> *> *
 {
     dispatch_async(self.queue, ^{
         [self _handleReportBegin];
-        [self _handleAttributeReport:attributeReport];
+        [self _handleAttributeReport:attributeReport fromSubscription:NO];
```
bzbarsky-apple, line 1186: The unit test injection is supposed to look like a subscription, so we should pass YES here, imo.  Or make it configurable by caller.

kiel-apple: Made `unitTestInjectAttributeReport` configurable by caller and set the two current callsites to pass `fromSubscription:YES`.

bzbarsky-apple (approved):

nivi-apple (approved):

andy31415 (approved):

PR #33490, opened 2024-05-16
Provide an estimate of subscription latency in MTRDevice.

bzbarsky-apple:
The idea is to track how long subscriptions took in the past and surface an estimate based on that to the API consumer.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -213,6 +213,7 @@ typedef NS_ENUM(NSUInteger, MTRDeviceWorkItemDuplicateTypeID) {
 
 static NSString * const sDataVersionKey = @"dataVersion";
 static NSString * const sAttributesKey = @"attributes";
+static NSString * const sLastInitialSubscribeLatencyKey = @"lastInitialSubscribeLatency";
 
 - (void)storeValue:(MTRDeviceDataValueDictionary _Nullable)value forAttribute:(NSNumber *)attribute
 {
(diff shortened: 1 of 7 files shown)


src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -925,6 +931,18 @@ - (void)_handleSubscriptionEstablished
...
 
+    auto initialSubscribeStart = _initialSubscribeStart;
+    // We no longer need to track subscribe latency for this device.
+    _initialSubscribeStart = nil;
+
+    if (initialSubscribeStart != nil) {
+        // We want time interval from initialSubscribeStart to now, not the other
+        // way around.
+        NSTimeInterval subscriptionLatency = -[initialSubscribeStart timeIntervalSinceNow];
+        _estimatedSubscriptionLatency = @(subscriptionLatency);
```
jtung-apple, line 942: Not necessarily now but we can consider doing an exponential moving average (when a previous value exists), so that no single super-slow establishment skews the estimate too much.

Suggesting exponential moving average since it's probably best to not store a lot of history. But we could also alternatively keep "last 10" if we are okay with that.

Again, not a blocking suggestion but we should consider this in a follow up.

kiel-apple: Yeah, "some sort of average of multiple points" is what I expected when I saw the name as well.  Fortunately we have plenty of flexibility in how it could be implemented behind the scenes, and a single datapoint is a fine start.

kiel-apple (approved):

src/darwin/Framework/CHIP/MTRDevice.mm
bzbarsky-apple: > we can consider doing an exponential moving average (when a previous value exists)

Yes, that is the idea.  Trying to keep it simple for the initial impl, but improve the heuristic later.

PR #33492, opened 2024-05-16
Updated the unit test documentation to reflect the transition to PW.

feasel0:
- Updated the documentation in `docs/testing/unit_testing.md` to reflect the transition from NL to PW unit tests.
- Added example code for a fixture-less test, a test with fixtures, and a test that derives from AppContext.
- Added a "best practices" section, which I encourage people to add their wisdom to.
- Also made minor updates to `docs/guides/openiotsdk_unit_tests.md` where NL was mentioned briefly.

The documentation assumes that some refactoring has been done to AppContext, MessagingContext, and its subclasses after the last NL tests have been replaced.  See [Draft PR #33497](https://github.com/project-chip/connectedhomeip/pull/33497) for a proposal what those changes might look like.

PR #33497, opened 2024-05-17
Proposed refactoring of AppContext, MessagingContext, and its subclasses after PW test migration.

feasel0:
This draft has been replaced by PR [#34036](https://github.com/project-chip/connectedhomeip/pull/34036)
-----------------------------------------------------------------------

In order to update the unit testing documentation (See PR [#33492](https://github.com/project-chip/connectedhomeip/pull/33492)) I needed to make some assumptions about what `AppContext`, `LoopbackMessagingContext`, and `UDPMessagingContext` will look like after the transition to PW is complete.  This draft is an example of how those classes might be refactored, as well as how they might be used in a test.

I assume we want `LoopbackMessagingContext` to inherit from `::testing::Test` (and then `AppContext` inherits from `LoopbackMessagingContext`).  So people writing unit tests will have their test contexts inherit from `AppContext` if they need the loopback messaging, or they will inherit directly from `::testing::Test` if they don't need the loopback messaging but do need other fixtures.

I propose several changes to `AppTestContext.h/cpp` and `MessagingContext.h/cpp`.
- `AppTestContext`'s SetUp and TearDown functions are made virtual.
- `LoopbackMessagingContext` and `UDPMessagingContext` inherit from `::testing::Test`.
- `VerifyOrDieWithMsg` in the SetUp/SetUpTestSuite functions are replaced by ASSERT_EQ.
- `sLoopbackTransportManager` and `sUDPTransportManager` are made into pointers.  This is done to reduce the static footprint.

Assuming the above is correct and what we want, a number of tests in `src/app/tests`, `src/controller/tests`, and other places can be updated so that their test context inherits from `AppContext` instead of defining a member variable that holds a pointer to an AppContext instance that is created dynamically.

As an example of how these tests might look after refactoring, I have updated `src/controller/tests/TestEventCaching.cpp` and `src/controller/tests/TestWriteChunking.cpp` so that the test context inherits from `AppContext`.


arkq:
@feasel0  why this PR is still a draft?
The documentation which is based on this PR is already merged. So, I think that we should go forward with these changes as well.

PR #33498, opened 2024-05-17
Remove final keyword from the class contains virtual functions

yufengwangca:
If the class is declared as final, it cannot be inherited from. Therefore, the virtual member functions are unnecessary. 
 
 We should either remove the final keyword from class declaration or the virtual keyword from the member functions of the final class.
 
I choose to remove the final keyword from the class declaration to  align with other RPC service interfaces

### DIFF
--- a/examples/common/pigweed/rpc_services/BooleanState.h
+++ b/examples/common/pigweed/rpc_services/BooleanState.h
@@ -28,7 +28,7 @@
 namespace chip {
 namespace rpc {
 
-class BooleanState final : public pw_rpc::nanopb::BooleanState::Service<BooleanState>
+class BooleanState : public pw_rpc::nanopb::BooleanState::Service<BooleanState>
 {
 public:
     virtual ~BooleanState() = default;
--- a/examples/common/pigweed/rpc_services/Locking.h
+++ b/examples/common/pigweed/rpc_services/Locking.h
@@ -27,7 +27,7 @@
 namespace chip {
 namespace rpc {
 
-class Locking final : public pw_rpc::nanopb::Locking::Service<Locking>
+class Locking : public pw_rpc::nanopb::Locking::Service<Locking>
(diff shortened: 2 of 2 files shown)


andy31415 (dismissed):

examples/common/pigweed/rpc_services/BooleanState.h
```diff
@@ -28,7 +28,7 @@
 namespace chip {
 namespace rpc {
 
-class BooleanState final : public pw_rpc::nanopb::BooleanState::Service<BooleanState>
```
andy31415, line 31: Is this correct thoug? could these be actually meant to be `override` instead rather than virtual that is supposed to be overriden later?

If these are override, it does sound reasonable for these classes to be final.

yufengwangca: I did't get this, a final class cannot be inherited from, typically does not need member functions with the virtual keyword. The virtual keyword is used to allow derived classes to override functions in the base class. Since a final class cannot be inherited, there will be no derived classes to override its member functions.

After looking at other interface definitions, the interface definitions within these header files might be override by different platforms, such as https://github.com/project-chip/connectedhomeip/blob/master/examples/platform/esp32/Rpc.cpp#L148.

If the member functions are not meant to be overridden, they should not be marked with the virtual keyword. Conversely, if the functions are marked with virtual, the class should not be declared as final.

andy31415: My question is: in what situations will we override `BooleanState` ?

andy31415: Ok, I thought that these were implementations overriding the BooleanState::Service ... looking at the impl that does not seem to be what is happening here.

andy31415 (approved):

PR #33500, opened 2024-05-17
Replace nl-unit-test with pigweed for messaging

jlatusek:
## Problem

## Changes
These changes are another piece of replacing the nl-unit test with the pigweed-unit test. I changed tests for the `messaging`. 

I've swapped the order of calling tests on nrfconnect because there is a problem with running messaging tests after calling the test with the nlunitest framework(probably the nlunittest does not clean everything properly). The nlunittest will be removed in the future so I don't want to resolve the related problem right now.

### DIFF
--- a/src/messaging/tests/TestAbortExchangesForFabric.cpp
+++ b/src/messaging/tests/TestAbortExchangesForFabric.cpp
@@ -21,8 +21,9 @@
  *      one) for a fabric.
  */
 
-#include <app/icd/server/ICDServerConfig.h>
 #include <gtest/gtest.h>
+
+#include <app/icd/server/ICDServerConfig.h>
 #include <messaging/ExchangeContext.h>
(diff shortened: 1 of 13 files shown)


src/messaging/tests/TestAbortExchangesForFabric.cpp
```diff
@@ -226,23 +233,23 @@ void CommonCheckAbortAllButOneExchange(nlTestSuite * inSuite, TestContext & ctx,
         //
         auto waitTimeout = System::Clock::Milliseconds32(1000);
 
-#if CHIP_CONFIG_ENABLE_ICD_SERVER
+#if CHIP_CONFIG_ENABLE_ICD_SERVER == 1
```
arkq, line 236: Why such change?

arkq (approved):

andy31415 (approved):

PR #33510, opened 2024-05-17
Python testing: id range check functions

cecille:
Adding some convenience functions for range checks on the IDs. This should make some of the IDM tests a bit easier to understand because now they're using magic numbers.

now with tests!


src/python_testing/global_attribute_ids.py
```diff
@@ -38,3 +39,95 @@ def to_name(self) -> str:
             return "FeatureMap"
         if self == GlobalAttributeIds.CLUSTER_REVISION_ID:
             return "ClusterRevision"
+
+
+@dataclass
+class IdRange:
```
andy31415, line 45: Could we use `range(start, stop)` as a python object instead of having a separate min/max class with contains?

src/python_testing/global_attribute_ids.py
```diff
@@ -38,3 +39,95 @@ def to_name(self) -> str:
...
+    min: int
+    max: int
+
+    def _contains(self, id: int):
+        return id in range(self.min, self.max+1)
+
+
+@dataclass
+class PrefixIdRange(IdRange):
```
andy31415, line 54: why is this a dataclass but SuffixIdRange is not?

cecille: copy/paste error.

src/python_testing/global_attribute_ids.py
cecille: sure. I did it like basically so I could have the contains function. Are you thinking of making the range the datamember? or using something like STANDARD_PREFIX = range(0x0000, 0x0000+1) and having a prefix_contains and suffix_contains functions?

andy31415: I thought complete replacing it. If this was done to have 'includes the end' range, then maybe a comment saying that would help (to make it clear why the python range object did not directly fit the bar).

cecille: added a comment.

andy31415 (approved):

PR #33516, opened 2024-05-18
Make TC_IDM_4_2.py a bit more resistent to timing issues in Step 8

### LINKED ISSUE #33515: [CERT-TEST-FAILURE] TC_IDM_4_2.py has timing issue in step 8 and should wait a bit longer to verify subscription response
### Feature Area

Other

### Test Case

TC_IDM_4_2.py

### Reproduction steps

The test in Step 8 requests a minimum subscription update interval of 3s. Then it changes an attribute and exactly waits for 3s and expects that after this the subscription change already arrived.

https://github.com/project-chip/connectedhomeip/blob/443bb9c775a09c28b3b6a15bfd2bf7e91cfec5ad/src/python_testing/TC_IDM_4_2.py#L412

Giving timings on device nd also network side - and especially very fast local message processing this can lead to weird timing issues during test that lead to failing tests.

My experiments show that already adding 100ms to the sleep time makes these issues in tests go away. Given the fact that the devcie could in real world cases have also other things to handle the expectation of such an accurateness might be too much ;-)

### Bug prevalence

alwys with matter,js in local testing

master

### Platform

core

### Anything else?

_No response_

Apollon77:
fixes #33515

PS: yes i know that this is a bit "hacky" because the additional waiting time is not accurately determined, but I would not know how. So in my eyes this is a pragmatic "better then the former too exact approach" solution :-)

I know that this test was removed from test plans ifor Matter 1.3, but for me it still has it's value to verify matter.js in CI, so it would be cool to get this small optimization in here. Alternatively I can wait until https://github.com/project-chip/connectedhomeip/pull/33299 lands and you can close the issue and the PR.

### DIFF
--- a/src/python_testing/TC_IDM_4_2.py
+++ b/src/python_testing/TC_IDM_4_2.py
@@ -408,8 +408,8 @@ class TC_IDM_4_2(MatterBaseTest):
             [(0, node_label_attr(value=new_node_label_write))]
         )
 
-        # Wait MinIntervalFloor seconds before reading updated attribute value
-        time.sleep(same_min_max_interval_sec)
+        # Wait MinIntervalFloor seconds (plus a bit longer for timing reasons) before reading updated attribute value
+        time.sleep(same_min_max_interval_sec + 0.1)
         new_node_label_read = sub_cr1_update_value.GetAttribute(node_label_attr_typed_path)
 
         # Verify new attribute value after MinIntervalFloor time


andy31415 (approved):

Apollon77:
I got aware of other issues in the test and the follwo up PR, so I will close here becaus ethe test is - as it is today - not really usable in CI.

PR #33519, opened 2024-05-20
Add Wi-Fi Network Management Cluster and use correct NIM device type id

ksperling-apple:
Add Wi-Fi Network Management Cluster XML and implementation.
Implement the cluster in the network-manager-app example (with hard-coded data for now).
Use the correct assigned device type id for the NIM.
Change the minimal-build job to build the network-manage-app.

### DIFF
--- a/docs/zap_clusters.md
+++ b/docs/zap_clusters.md
@@ -127,6 +127,7 @@ Generally regenerate using one of:
 |       1295 |      0x50F | ContentControl                                          |
 |       1296 |      0x510 | ContentAppObserver                                      |
 |       2820 |      0xB04 | ElectricalMeasurement                                   |
+|       3072 |      0xC00 | WiFiNetworkManagement                                   |
 | 4294048773 | 0xFFF1FC05 | UnitTesting                                             |
 | 4294048774 | 0xFFF1FC06 | FaultInjection                                          |
 | 4294048800 | 0xFFF1FC20 | SampleMei                                               |
--- a/docs/zap_clusters.md
+++ b/docs/zap_clusters.md
@@ -112,6 +112,7 @@ Generally regenerate using one of:
 |       1069 |      0x42D | Pm10ConcentrationMeasurement                            |
 |       1070 |      0x42E | TotalVolatileOrganicCompoundsConcentrationMeasurement   |
 |       1071 |      0x42F | RadonConcentrationMeasurement                           |
+|       1105 |      0x451 | WiFiNetworkManagement                                   |
 |       1283 |      0x503 | WakeOnLan                                               |
 |       1284 |      0x504 | Channel                                                 |
 |       1285 |      0x505 | TargetNavigator                                         |
@@ -127,7 +128,6 @@ Generally regenerate using one of:
 |       1295 |      0x50F | ContentControl                                          |
 |       1296 |      0x510 | ContentAppObserver                                      |
 |       2820 |      0xB04 | ElectricalMeasurement                                   |
-|       3072 |      0xC00 | WiFiNetworkManagement                                   |
 | 4294048773 | 0xFFF1FC05 | UnitTesting                                             |
 | 4294048774 | 0xFFF1FC06 | FaultInjection                                          |
 | 4294048800 | 0xFFF1FC20 | SampleMei                                               |
--- a/examples/network-manager-app/linux/main.cpp
+++ b/examples/network-manager-app/linux/main.cpp
@@ -39,8 +39,8 @@ int main(int argc, char * argv[])
         return -1;
     }
 
-    WiFiNetworkManagement::Server::Instance().SetNetworkCredentials(ByteSpanFromCharSpan("MatterAP"_span),
-                                                                    ByteSpanFromCharSpan("Setec Astronomy"_span));
+    WiFiNetworkManagementServer::Instance().SetNetworkCredentials(ByteSpanFromCharSpan("MatterAP"_span),
+                                                                  ByteSpanFromCharSpan("Setec Astronomy"_span));
 
     ChipLinuxAppMainLoop();
     return 0;
--- a/examples/network-manager-app/linux/main.cpp
+++ b/examples/network-manager-app/linux/main.cpp
@@ -16,12 +16,20 @@
  */
 
 #include <AppMain.h>
+#include <app/clusters/wifi-network-management-server/wifi-network-management-server.h>
+#include <lib/support/Span.h>
 
 using namespace chip;
 using namespace chip::app;
+using namespace chip::app::Clusters;
 
 void ApplicationInit() {}
 void ApplicationShutdown() {}
+
+ByteSpan ByteSpanFromCharSpan(CharSpan span)
+{
+    return ByteSpan(reinterpret_cast<const uint8_t *>(span.data()), span.size());
+}
 
 int main(int argc, char * argv[])
 {
@@ -29,6 +37,9 @@ int main(int argc, char * argv[])
     {
         return -1;
     }
+
+    WiFiNetworkManagement::Server::Instance().SetNetworkCredentials(ByteSpanFromCharSpan("MatterAP"_span),
+                                                                    ByteSpanFromCharSpan("Seatec Astronomy"_span));
 
     ChipLinuxAppMainLoop();
     return 0;
--- a/examples/network-manager-app/linux/main.cpp
+++ b/examples/network-manager-app/linux/main.cpp
@@ -17,6 +17,7 @@
 
 #include <AppMain.h>
 #include <app/clusters/wifi-network-management-server/wifi-network-management-server.h>
+#include <lib/core/CHIPSafeCasts.h>
 #include <lib/support/Span.h>
 
 using namespace chip;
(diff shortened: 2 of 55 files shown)


docs/zap_clusters.md
```diff
@@ -127,6 +127,7 @@ Generally regenerate using one of:
 |       1295 |      0x50F | ContentControl                                          |
 |       1296 |      0x510 | ContentAppObserver                                      |
 |       2820 |      0xB04 | ElectricalMeasurement                                   |
+|       3072 |      0xC00 | WiFiNetworkManagement                                   |
```
bzbarsky-apple, line 130: This does not match the cluster ID I see in the spec?  That seems to be 0x0451.

examples/network-manager-app/linux/main.cpp
```diff
@@ -16,20 +16,31 @@
...
 using namespace chip;
 using namespace chip::app;
+using namespace chip::app::Clusters;
 
 void ApplicationInit() {}
 void ApplicationShutdown() {}
 
+ByteSpan ByteSpanFromCharSpan(CharSpan span)
+{
+    return ByteSpan(reinterpret_cast<const uint8_t *>(span.data()), span.size());
```
bzbarsky-apple, line 31: `Uint8::from_const_char`, please.

src/app/clusters/wifi-network-management-server/wifi-network-management-server.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+    VerifyOrReturnError(!SsidSpan().data_equal(ssid) || !PassphraseSpan().data_equal(passphrase), CHIP_NO_ERROR);
+
+    memcpy(mSsid, ssid.data(), ssid.size());
+    mSsidLen = static_cast<decltype(mSsidLen)>(ssid.size());
+
+    ReturnErrorOnFailure(mPassphrase.SetLength(passphrase.size()));
+    memcpy(mPassphrase.Bytes(), passphrase.data(), passphrase.size());
+
+    MatterReportingAttributeChangeCallback(mEndpointId, Id, Ssid::Id); // report SSID change even if only passphrase changed
```
bzbarsky-apple, line 83: Why?  That's a very strange thing to do....  Please explain why that's happening.

src/app/clusters/wifi-network-management-server/wifi-network-management-server.cpp
```diff
@@ -0,0 +1,142 @@
...
+    case Commands::NetworkPassphraseRequest::Id:
+        HandleCommand<Commands::NetworkPassphraseRequest::DecodableType>(
+            ctx, std::bind(&Server::HandleNetworkPassphraseRequest, this, _1, _2));
+        return;
+    }
+}
+
+void Server::HandleNetworkPassphraseRequest(HandlerContext & ctx, const Commands::NetworkPassphraseRequest::DecodableType & req)
+{
+    if (mPassphrase.Length() > 0)
```
bzbarsky-apple, line 112: I don't see this in the spe.  The spec says you just return whatever the passphrase is.  There is no mention of InvalidInState in the spec that I can see...

Am I looking at the wrong spec?  I'm looking at what's on tip of `master` right now.

src/app/zap-templates/zcl/data-model/chip/wifi-network-management-cluster.xml
```diff
@@ -0,0 +1,44 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<configurator>
+    <domain name="CHIP"/>
+
+    <cluster>
+        <domain>Network Infrastructure</domain>
+        <name>Wi-Fi Network Management</name>
+        <code>0x0C00</code>
```
bzbarsky-apple, line 23: This does not seem to match the spec.

src/app/clusters/wifi-network-management-server/wifi-network-management-server.cpp
ksperling-apple: The spec currently allows the SSID to be null (e.g. when the network is not set up yet). In that case we also don't have a passphrase to return, so I think we need to amend the spec to define what the status code should be in that case. I've filed https://github.com/CHIP-Specifications/connectedhomeip-spec/issues/9234 to track this.

ksperling-apple: Also to clarify further, I'm using `mSsidLen == 0` and `mPassphrase.Length() == 0` to represent the state where no network is configured yet. Possibly this check should be for `mSsidLen == 0` if we want to allow a passphrase length of 0 as a valid state, however according to the interpretation in the Network Commissioning Cluster ("Valid Credentials length are:" ...) this implies an unsecured / open network, which I don't think we want to allow here.

I will add this to the spec issue... We should really only be allowing WPA2-Personal or better credentials for the primary Wi-Fi Network for the home.

src/app/clusters/wifi-network-management-server/wifi-network-management-server.cpp
ksperling-apple: The spec has no event defined currently for reporting a change to the passphrase, so this was intended to be something of a placeholder until we tidy up the spec, but maybe it's better to just take it out. I've included this issue in https://github.com/CHIP-Specifications/connectedhomeip-spec/issues/9234.

src/app/zap-templates/zcl/data-model/chip/wifi-network-management-cluster.xml
ksperling-apple: Ah good catch, I thought I had updated the ID in the xml but apparently I didn't.

src/app/clusters/wifi-network-management-server/wifi-network-management-server.cpp
ksperling-apple: Actually the spec says "WPA-Personal" so the credential should always be 8..63 bytes or 64 hex characters (but this could be made more explicit in the spec). `SetNetworkCredentials` should explicitly check for this.

docs/zap_clusters.md
ksperling-apple: Addressed in the XML below (this one is generated)

examples/network-manager-app/linux/main.cpp
ksperling-apple: TIL

src/app/clusters/wifi-network-management-server/wifi-network-management-server.cpp
ksperling-apple: Added a `HaveNetworkCredentials()` helper to make the code clearer, as well as the described validation in `SetNetworkCredentials()`

src/app/clusters/wifi-network-management-server/wifi-network-management-server.cpp
bzbarsky-apple: Sounds good, thanks!

src/app/clusters/wifi-network-management-server/wifi-network-management-server.cpp
bzbarsky-apple: Looks great, thanks!

bzbarsky-apple (approved):

src/app/clusters/wifi-network-management-server/wifi-network-management-server.h
```diff
@@ -0,0 +1,75 @@
...
+#include <lib/support/Span.h>
+
+void emberAfWiFiNetworkManagementClusterServerInitCallback(chip::EndpointId);
+
+namespace chip {
+namespace app {
+namespace Clusters {
+namespace WiFiNetworkManagement {
+
+class Server : private AttributeAccessInterface, private CommandHandlerInterface
```
andy31415, line 35: nit: Would you consider merging namespace and class name into `WiFiNetworkManagementServer` ?

Reasoning being that `Server` is a generic name and if this pattern gets repeated, grepping for `Server` will start having a lot of hits.

src/app/clusters/wifi-network-management-server/wifi-network-management-server.cpp
```diff
@@ -0,0 +1,169 @@
...
+using namespace chip::app::Clusters;
+using namespace chip::app::Clusters::WiFiNetworkManagement::Attributes;
+using namespace std::placeholders;
+
+namespace chip {
+namespace app {
+namespace Clusters {
+namespace WiFiNetworkManagement {
+
+Global<Server> gServerInstance;
```
andy31415, line 41: should this be in an anonymous namespace so that access can only happen via the Intance call?

andy31415 (approved):

src/app/clusters/wifi-network-management-server/wifi-network-management-server.h
ksperling-apple: Yes, I think that's much better... (I was trying to follow an existing pattern here but we should probably standardize on something like `<cluster name>Server` instead)

ksperling-apple: Done

PR #33520, opened 2024-05-20
dac-revocation: Some refactoring to revocation set generation script

shubhamdp:
- Implemented few TODOs
- Added `DCLDRequestsHelper` for interacting with DCL
- And few minor fixes

#### Tests
Verified by running below commands
```
python3 generate-revocation-set.py --use-test-net-http
python3 generate-revocation-set.py --use-main-net-dcld ~/go/bin/dcld
```

### DIFF
--- a/credentials/generate-revocation-set.py
+++ b/credentials/generate-revocation-set.py
@@ -22,6 +22,7 @@
 
 import base64
 import json
+import logging
 import subprocess
 import sys
 from enum import Enum
@@ -31,6 +32,15 @@ import requests
 from click_option_group import RequiredMutuallyExclusiveOptionGroup, optgroup
 from cryptography import x509
 from cryptography.hazmat.primitives.asymmetric import ec
+
+# Supported log levels, mapping string values required for argument
+# parsing into logging constants
+__LOG_LEVELS__ = {
+    'debug': logging.DEBUG,
+    'info': logging.INFO,
+    'warn': logging.WARN,
+    'fatal': logging.FATAL,
+}
 
 
 class RevocationType(Enum):
@@ -191,9 +201,19 @@ class DCLDClient:
 @optgroup.option('--use-main-net-http', is_flag=True, type=str, help="Use RESTful API with HTTPS against public MainNet observer.")
 @optgroup.option('--use-test-net-http', is_flag=True, type=str, help="Use RESTful API with HTTPS against public TestNet observer.")
 @optgroup.group('Optional arguments')
-@optgroup.option('--output', default='sample_revocation_set_list.json', type=str, metavar='FILEPATH', help="Output filename (default: sample_revocation_set_list.json)")
-def main(use_main_net_dcld, use_test_net_dcld, use_main_net_http, use_test_net_http, output):
+@optgroup.option('--output', default='sample_revocation_set_list.json', type=str, metavar='FILEPATH',
+                 help="Output filename (default: sample_revocation_set_list.json)")
+@optgroup.option('--log-level', default='INFO', show_default=True, type=click.Choice(__LOG_LEVELS__.keys(),
+                                                                                     case_sensitive=False), callback=lambda c, p, v: __LOG_LEVELS__[v],
+                 help='Determines the verbosity of script output')
+def main(use_main_net_dcld: str, use_test_net_dcld: str, use_main_net_http: bool, use_test_net_http: bool, output: str, log_level: str):
     """Tool to construct revocation set from DCL"""
+
+    logging.basicConfig(
+        level=log_level,
+        format='%(asctime)s %(name)s %(levelname)-7s %(message)s',
+        datefmt='%Y-%m-%d %H:%M:%S'
+    )
 
     production = False
     dcld = use_test_net_dcld
@@ -217,7 +237,7 @@ def main(use_main_net_dcld, use_test_net_dcld, use_main_net_http, use_test_net_h
     for revocation_point in revocation_point_list:
         # 1. Validate Revocation Type
         if revocation_point["revocationType"] != RevocationType.CRL.value:
-            print("Revocation Type is not CRL, continue...")
+            logging.warning("Revocation Type is not CRL, continue...")
             continue
 
         # 2. Parse the certificate
@@ -236,15 +256,15 @@ def main(use_main_net_dcld, use_test_net_dcld, use_main_net_http, use_test_net_h
             if crl_vid is not None:
                 if vid != crl_vid:
                     # TODO: Need to log all situations where a continue is called
-                    print("VID is not CRL VID, continue...")
+                    logging.warning("VID is not CRL VID, continue...")
                     continue
         else:
             if crl_vid is None or vid != crl_vid:
-                print("VID is not CRL VID, continue...")
+                logging.warning("VID is not CRL VID, continue...")
                 continue
             if crl_pid is not None:
                 if pid != crl_pid:
-                    print("PID is not CRL PID, continue...")
+                    logging.warning("PID is not CRL PID, continue...")
                     continue
 
         # 5. Validate the certification path containing CRLSignerCertificate.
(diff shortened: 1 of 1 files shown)


credentials/generate-revocation-set.py
```diff
@@ -44,20 +45,61 @@ class RevocationType(Enum):
...
 
 
+class DCLDRequestsHelper:
+    def __init__(self, use_rest, dcld, production, rest_node_url):
+        self.use_rest = use_rest
+        self.dcld = dcld
+        self.production = production
+        self.rest_node_url = rest_node_url
+
+    def use_dcld(self, dcld, production, cmdlist):
```
tcarmelveilleux, line 64: Suggest: `build_cld_command_line`

credentials/generate-revocation-set.py
```diff
@@ -44,20 +45,61 @@ class RevocationType(Enum):
...
+        return [dcld] + cmdlist + (['--node', PRODUCTION_NODE_URL] if production else [])
+
+    def get_dcld_cmd_output(self, cmdlist):
```
tcarmelveilleux, line 67: - Please add basic description of arguments to all methods.
- Consider adding type annotations (e.g. `cmdlist: list[str]`)

credentials/generate-revocation-set.py
```diff
@@ -44,20 +45,61 @@ class RevocationType(Enum):
...
+
+    def get_revocation_points(self):
+        if self.use_rest:
+            response = requests.get(f"{self.rest_node_url}/dcl/pki/revocation-points").json()
+        else:
+            response = self.get_dcld_cmd_output(['query', 'pki', 'all-revocation-points'])
+
+        return response["PkiRevocationDistributionPoint"]
+
+    def get_paa_cert_for_crl_issuer(self, crl_signer_issuer_name, crl_signer_authority_key_id):
```
tcarmelveilleux, line 83: issuer_name is base 64, right? Suggest `_b64` suffix for each base 64 args.

credentials/generate-revocation-set.py
```diff
@@ -83,28 +125,20 @@ def main(use_main_net_dcld, use_test_net_dcld, use_main_net_http, use_test_net_h
...
 
-    # TODO: Extract this to a helper function
-    if use_rest:
-        revocation_point_list = requests.get(f"{rest_node_url}/dcl/pki/revocation-points").json()["PkiRevocationDistributionPoint"]
-    else:
-        cmdlist = ['config', 'output', 'json']
-        subprocess.Popen([dcld] + cmdlist)
-
-        cmdlist = ['query', 'pki', 'all-revocation-points']
+    dcld_helper = DCLDRequestsHelper(use_rest, dcld, production, rest_node_url)
```
tcarmelveilleux, line 128: Suggest `DCLClient` as a class name, and `dcl_client` instead of `dcld_helper`

credentials/generate-revocation-set.py
```diff
@@ -133,99 +170,119 @@ def main(use_main_net_dcld, use_test_net_dcld, use_main_net_http, use_test_net_h
...
 
-        paa_certificate_object = x509.load_pem_x509_certificate(paa_certificate)
+        paa_certificate_object = x509.load_pem_x509_certificate(bytes(paa_certificate, 'utf-8'))
 
+        # This requires cryptography 40.0.0 and this haults on 40.0.0
+        # try:
+        #     crl_signer_certificate.verify_directly_issued_by(paa_certificate_object)
+        # except Exception as e:
+        #     print("CRL Signer Certificate is not issued by PAA Certificate, continue...")
+        #     continue
```
tcarmelveilleux, line 191: Not sure I understand. Are you saying we need to bump the requested module version?

credentials/generate-revocation-set.py
```diff
@@ -133,99 +170,119 @@ def main(use_main_net_dcld, use_test_net_dcld, use_main_net_http, use_test_net_h
...
+            # try:
+            #     revoked_cert_authority_key_id = revoked_cert.extensions.get_extension_for_oid(
+            #         x509.OID_AUTHORITY_KEY_IDENTIFIER).value.key_identifier
+            #
+            #     if revoked_cert_authority_key_id is None or revoked_cert_authority_key_id != crl_signer_subject_key_id:
+            #         print("CRL Authority Key ID is not CRL Signer Subject Key ID, continue...")
+            #         continue
+            # except Exception:
+            #     print("CRL Authority Key ID not found, continue...")
+            #     continue
```
tcarmelveilleux, line 285: Why is this commented?

credentials/generate-revocation-set.py
```diff
@@ -133,99 +170,119 @@ def main(use_main_net_dcld, use_test_net_dcld, use_main_net_http, use_test_net_h
...
                 continue
 
         # 9. Assign CRL File Issuer
         certificate_authority_name = base64.b64encode(crl_file.issuer.public_bytes()).decode('utf-8')
+        print(f"CRL File Issuer: {certificate_authority_name}")
 
         serialnumber_list = []
         # 10. Iterate through the Revoked Certificates List
         for revoked_cert in crl_file:
+            print(revoked_cert)
```
tcarmelveilleux, line 262: This is too verbose. Please remove, or add a `verbose_logging` bool member to self

credentials/generate-revocation-set.py
```diff
@@ -44,20 +45,61 @@ class RevocationType(Enum):
...
     if len(attribute_list) == 1:
-        if attribute_list[0].value.isdigit():
-            return int(attribute_list[0].value)
+        return int(attribute_list[0].value, 16)
 
     return None
 
 
+class DCLDRequestsHelper:
+    def __init__(self, use_rest, dcld, production, rest_node_url):
```
andy31415, line 58: if python3, could we add some typing here? it would be nice to know which are bool and which are string and maybe even some info about them.

Also consider if we can improve on naming. Generally helper/support/util/common are filler words that should be avoided if possible.

What ius the purpose of the class? As a `helper` it feels like potentially a grab all of anything...

credentials/generate-revocation-set.py
andy31415: Also please use doccomments for descriptions (basically strings below declaration) to make the API feel pythonic.

Comment feels off off as it just seems to restate the arguments below. Maybe something more descriptive is useful?

credentials/generate-revocation-set.py
andy31415: `dcld` maybe should be renamed as `dcld_executable` or `dcld_binary`

credentials/generate-revocation-set.py
```diff
@@ -44,20 +45,144 @@ class RevocationType(Enum):
...
+        Returns
+        -------
+        dict
+            The JSON output from the command
+        '''
+
+        # Set the output as JSON
+        subprocess.Popen([self.dcld_exe, 'config', 'output', 'json'])
+
+        cmdpipe = subprocess.Popen(self.build_dcld_command_line(cmdlist),
```
andy31415, line 117: should we have `check=True` here?

credentials/generate-revocation-set.py
```diff
@@ -119,12 +236,15 @@ def main(use_main_net_dcld, use_test_net_dcld, use_main_net_http, use_test_net_h
...
                     # TODO: Need to log all situations where a continue is called
+                    print("VID is not CRL VID, continue...")
                     continue
         else:
             if crl_vid is None or vid != crl_vid:
+                print("VID is not CRL VID, continue...")
                 continue
             if crl_pid is not None:
                 if pid != crl_pid:
+                    print("PID is not CRL PID, continue...")
```
andy31415, line 247: Consider using `logging` instead of print. This could be info or debug.

tcarmelveilleux: +1: logging.warn

tcarmelveilleux (approved):

credentials/generate-revocation-set.py
shubhamdp: [Popen()](https://docs.python.org/3/library/subprocess.html#subprocess.Popen) do not have the `check=` argument

credentials/generate-revocation-set.py
shubhamdp: using the logging module in this script.

PR #33527, opened 2024-05-21
Darwin: Make MTRSetMessageReliabilityParameters thread-safe

ksperling-apple:
This can be called from any thread, but the underyling SDK calls need to be made from the Matter queue. Dispatching again in resetOperationalAdvertising was also prone to racing.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
+++ b/src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
@@ -883,23 +883,19 @@ MTR_DIRECT_MEMBERS
 {
     assertChipStackLockedByCurrentThread();
 
-    if (!_advertiseOperational) {
-        // No need to reset anything; we are not advertising the things that
-        // would need to get reset.
-        return;
+    // If we're not advertising, then there's no need to reset anything.
+    VerifyOrReturn(_advertiseOperational);
+
+    // If there are no running controllers there will be no advertisements to reset.
+    {
+        std::lock_guard lock(_controllersLock);
+        VerifyOrReturn(_controllers.count > 0);
     }
 
-    std::lock_guard lock(_controllersLock);
-    if (_controllers.count != 0) {
-        // We have a running controller.  That means we likely need to reset
-        // operational advertising for that controller.
-
-        // StartServer() is the only API we have for resetting DNS-SD
-        // advertising.  It sure would be nice if there were a "restart"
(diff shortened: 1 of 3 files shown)


woody-apple (approved):

src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
```diff
@@ -1165,6 +1166,45 @@ - (MTRDeviceController * _Nullable)initializeController:(MTRDeviceController *)c
                                   error:error];
 }
 
+- (void)setMessageReliabilityProtocolIdleRetransmitMs:(nullable NSNumber *)idleRetransmitMs
+                                   activeRetransmitMs:(nullable NSNumber *)activeRetransmitMs
+                                    activeThresholdMs:(nullable NSNumber *)activeThresholdMs
+                          additionalRetransmitDelayMs:(nullable NSNumber *)additionalRetransmitDelayMs
+{
+    [self _assertCurrentQueueIsNotMatterQueue];
+    dispatch_sync(_chipWorkQueue, ^{
```
bzbarsky-apple, line 1175: Does this need to be a dispatch_sync?  The MRP params are not observable by any code that is not running on the Matter queue, after this change, right?  Message sending is on that queue, and so is our `GetLocalMRPConfig()` in this function after these changes.

So it seems to me that this can be an async dispatch just fine.

bzbarsky-apple (approved):

src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
```diff
@@ -891,13 +893,12 @@ - (void)resetOperationalAdvertising
...
-            // that was a no-op if the DNS-SD server was not already
-            // running.
-            app::DnssdServer::Instance().StartServer();
-        });
+
+        // StartServer() is the only API we have for resetting DNS-SD
+        // advertising.  It sure would be nice if there were a "restart"
+        // that was a no-op if the DNS-SD server was not already
+        // running.
+        app::DnssdServer::Instance().StartServer();
```
bzbarsky-apple, line 901: Though I would be a bit happier if we did a "we have a running controller" check under the lock, then released the lock before we call into arbitrary code via StartServer() here.  We know the "have a running controller" state can't change while we are in this function, even if we release the controllers lock, because nothing else can run on the Matter queue while we are here, with the dispatch now gone.

PR #33533, opened 2024-05-21
[python] Add ICD support to Python binding

erjiaqing:
This PR adds ICD support to Python binding, including:

1. ICD Registration during commissioning
2. Wait for active device

APIs added:

```
devCtrl.EnableICDRegistration(Optional[ICDRegistrationParameters])
async devCtrl.WaitForActive(nodeid, stayActiveDurationMilliseconds)

chip.icd.RegisterOnActiveCallback(fabricIndex, nodeId, callback)
chip.icd.UnregisterOnActiveCallback(fabricIndex, nodeId, callback)
```

ICD functions requires running chip-repl with `-s`.


src/controller/python/chip/ChipDeviceCtrl.py
```diff
@@ -1979,6 +1984,27 @@ def SetCheckMatchingFabric(self, check: bool):
             lambda: self._dmLib.pychip_DeviceController_SetCheckMatchingFabric(check)
         ).raise_on_error()
 
+    def SetIcdRegistrationParameters(self, symmetricKey: bytes = None, checkInNodeId: int = 0, monitoredSubject: int = 0, stayActiveMs: int = 0):
```
andy31415, line 1987: would it make maybe sense to have a `@dataclass` for parameters instead of several values? Or mark these as `**, symmetricKey: bytes = None, checkInNodeId: int = 0, monitoredSubject: int = 0, stayActiveMs: int = 0` to force to always name these arguments?

would a `stayActiveMs` being a timeDelta make sense?
A comment should maybe explain what a 0 means for stayActiveMs.

src/controller/python/chip/ChipDeviceCtrl.py
```diff
@@ -1979,6 +1984,27 @@ def SetCheckMatchingFabric(self, check: bool):
...
+        if symmetricKey is not None:
+            if len(symmetricKey) != 16:
+                raise ValueError("symmetricKey should be 16 bytes")
+        if not checkInNodeId:
```
andy31415, line 1991: should these default to None and be Optional[int] instead of 0 to distingush between set/not-set?

src/controller/python/ChipDeviceController-ScriptDevicePairingDelegate.cpp
```diff
@@ -180,5 +185,40 @@ ScriptDevicePairingDelegate::GetOpenWindowCallback(Controller::CommissioningWind
...
+
+    if (err != CHIP_NO_ERROR)
+    {
+        sICDClientStorage.RemoveKey(clientInfo);
+        ChipLogError(chipTool, "Failed to persist symmetric key for " ChipLogFormatX64 ": %s", ChipLogValueX64(nodeId),
+                     err.AsString());
+        return;
+    }
+
+    ChipLogProgress(Zcl, "Saved ICD Symmetric key for " ChipLogFormatX64, ChipLogValueX64(nodeId));
```
yunhanw-google, line 209: please use controller instead of zcl for log in this folder

src/controller/python/ChipDeviceController-ScriptDevicePairingDelegate.cpp
```diff
@@ -180,5 +185,40 @@ ScriptDevicePairingDelegate::GetOpenWindowCallback(Controller::CommissioningWind
...
+    CHIP_ERROR err = sICDClientStorage.SetKey(clientInfo, ByteSpan(sICDSymmetricKey));
+    if (err == CHIP_NO_ERROR)
+    {
+        err = sICDClientStorage.StoreEntry(clientInfo);
+    }
+
+    if (err != CHIP_NO_ERROR)
+    {
+        sICDClientStorage.RemoveKey(clientInfo);
+        ChipLogError(chipTool, "Failed to persist symmetric key for " ChipLogFormatX64 ": %s", ChipLogValueX64(nodeId),
```
yunhanw-google, line 204: @[yunhanw-google](https://github.com/yunhanw-google) yunhanw-google [now](https://github.com/project-chip/connectedhomeip/pull/33533/files#r1610447077)
please use controller instead of zcl for log in this folder

src/controller/python/chip/ChipDeviceCtrl.py
yunhanw-google: I would prefer stayActiveMs or StayActiveDurationMs that could explicitly know the functionality about this parameteer which is to send stayActive request in the end of commissioning flow.

erjiaqing: Makes sense

I feel `SetIcdRegistrationParameters` which can accept 0 arguemtns a bit weird. So I renamed it to EnableICDRegistration

src/controller/python/ChipDeviceController-ScriptDevicePairingDelegate.cpp
erjiaqing: Done

src/controller/python/chip/ChipDeviceCtrl.py
erjiaqing: Done

yunhanw-google (approved):

yunhanw-google (approved):

src/controller/python/ChipDeviceController-ScriptBinding.cpp
```diff
@@ -291,6 +304,12 @@ PyChipError pychip_DeviceController_StackInit(Controller::Python::StorageAdapter
     //
     DeviceControllerFactory::GetInstance().RetainSystemState();
 
+    auto engine = chip::app::InteractionModelEngine::GetInstance();
+    PyReturnErrorOnFailure(ToPyChipError(PyChipCheckInDelegate::GetInstance().Init(&sICDClientStorage, engine)));
```
tcarmelveilleux, line 308: The interaction model engine is too big an interface for check-in delegate registration. Why is the whole InteractionModelEngine needed?

src/controller/python/ChipDeviceController-ScriptBinding.cpp
```diff
@@ -291,6 +304,12 @@ PyChipError pychip_DeviceController_StackInit(Controller::Python::StorageAdapter
...
+    PyReturnErrorOnFailure(
+        ToPyChipError(sCheckInHandler.Init(DeviceControllerFactory::GetInstance().GetSystemState()->ExchangeMgr(),
```
tcarmelveilleux, line 310: Why does the InteractionModelEngine need to be passed to check-in handler?

src/controller/python/ChipDeviceController-ScriptBinding.cpp
```diff
@@ -555,6 +580,46 @@ PyChipError pychip_DeviceController_SetCheckMatchingFabric(bool check)
     return ToPyChipError(CHIP_NO_ERROR);
 }
 
+PyChipError pychip_DeviceController_SetIcdRegistrationParameters(chip::Controller::DeviceCommissioner * devCtrl, bool enabled,
```
tcarmelveilleux, line 583: Can we use a struct instead to reduce the number of arguments to register?

src/controller/python/ChipDeviceController-ScriptBinding.cpp
```diff
@@ -555,6 +580,46 @@ PyChipError pychip_DeviceController_SetCheckMatchingFabric(bool check)
...
+                                                                 uint8_t * icdSymmetricKeyOrNull, uint64_t icdCheckInNodeIdOrZero,
```
tcarmelveilleux, line 584: - icdSymmetricKeyOrNull --> icdTokenOrNull
- token has to be `const uint8_t *`
- Please have a size_t for token size, rather than assume size, for safety.

src/controller/python/ChipDeviceController-ScriptBinding.cpp
```diff
@@ -555,6 +580,46 @@ PyChipError pychip_DeviceController_SetCheckMatchingFabric(bool check)
...
+
+    sCommissioningParameters.SetICDRegistrationStrategy(ICDRegistrationStrategy::kBeforeComplete);
+
+    if (icdSymmetricKeyOrNull != nullptr)
+    {
+        memcpy(sICDSymmetricKey, icdSymmetricKeyOrNull, sizeof(sICDSymmetricKey));
+    }
+    else
+    {
+        chip::Crypto::DRBG_get_bytes(sICDSymmetricKey, sizeof(sICDSymmetricKey));
```
tcarmelveilleux, line 600: Please do not have a method that auto-generates from C++. Just don't allow null and let the caller always manage it.

src/controller/python/ChipDeviceController-ScriptBinding.cpp
```diff
@@ -555,6 +580,46 @@ PyChipError pychip_DeviceController_SetCheckMatchingFabric(bool check)
...
+    }
+    if (icdCheckInNodeIdOrZero == 0)
+    {
+        icdCheckInNodeIdOrZero = devCtrl->GetNodeId();
+    }
+    if (icdMonitoredSubjectOrZero == 0)
+    {
+        icdMonitoredSubjectOrZero = icdCheckInNodeIdOrZero;
+    }
```
tcarmelveilleux, line 609: What's the use case for those paths?

src/controller/python/ChipDeviceController-ScriptBinding.cpp
```diff
@@ -555,6 +580,46 @@ PyChipError pychip_DeviceController_SetCheckMatchingFabric(bool check)
...
+    // These Optionals must have values now.
+    // The commissioner will verify these values.
+    sCommissioningParameters.SetICDSymmetricKey(ByteSpan(sICDSymmetricKey));
```
tcarmelveilleux, line 612: Can we avoid accessing this global state? Does the `DeviceCommissioner` have setter/getters for this?

src/controller/python/ChipDeviceController-ScriptDevicePairingDelegate.cpp
```diff
@@ -180,5 +185,40 @@ ScriptDevicePairingDelegate::GetOpenWindowCallback(Controller::CommissioningWind
     return &mOpenWindowCallback;
 }
 
+void ScriptDevicePairingDelegate::OnICDRegistrationComplete(NodeId nodeId, uint32_t icdCounter)
```
tcarmelveilleux, line 188: Is it possible to pass the fabric index as argument, rather than relying on state?

src/controller/python/ChipDeviceController-ScriptDevicePairingDelegate.cpp
```diff
@@ -180,5 +185,40 @@ ScriptDevicePairingDelegate::GetOpenWindowCallback(Controller::CommissioningWind
...
+
+    ChipLogProgress(Controller, "Saved ICD Symmetric key for " ChipLogFormatX64, ChipLogValueX64(nodeId));
+    ChipLogProgress(Controller,
+                    "ICD Registration Complete for device " ChipLogFormatX64 " / Check-In NodeID: " ChipLogFormatX64
+                    " / Monitored Subject: " ChipLogFormatX64 " / ICDCounter %u",
+                    ChipLogValueX64(nodeId), ChipLogValueX64(sCommissioningParameters.GetICDCheckInNodeId().Value()),
+                    ChipLogValueX64(clientInfo.monitored_subject), icdCounter);
+}
+
+void ScriptDevicePairingDelegate::OnICDStayActiveComplete(NodeId deviceId, uint32_t promisedActiveDuration)
```
tcarmelveilleux, line 217: There should be a `ScopedNodeId` here rather than `NodeId`

src/controller/python/ChipDeviceController-ScriptDevicePairingDelegate.h
```diff
@@ -68,11 +68,14 @@ class ScriptDevicePairingDelegate final : public Controller::DevicePairingDelega
...
                                 Optional<Credentials::AttestationVerificationResult> additionalErrorInfo) override;
     void OnCommissioningStatusUpdate(PeerId peerId, CommissioningStage stageCompleted, CHIP_ERROR error) override;
+    void OnICDRegistrationComplete(NodeId deviceId, uint32_t icdCounter) override;
+    void OnICDStayActiveComplete(NodeId deviceId, uint32_t promisedActiveDuration) override;
     void OnFabricCheck(NodeId matchingNodeId) override;
     Callback::Callback<Controller::OnOpenCommissioningWindow> *
     GetOpenWindowCallback(Controller::CommissioningWindowOpener * context);
     void OnOpenCommissioningWindow(NodeId deviceId, CHIP_ERROR status, SetupPayload payload);
     void SetExpectingPairingComplete(bool value) { expectingPairingComplete = value; }
+    void SetFabricIndex(FabricIndex fabricIndex) { mFabricIndex = fabricIndex; }
```
tcarmelveilleux, line 78: Please remove and replace with having `deviceId` be a ScopedNodeId

src/controller/python/chip/ChipDeviceCtrl.py
```diff
@@ -844,6 +867,18 @@ def deviceAvailable(self, device, err):
...
+    async def WaitForActive(self, nodeid, stayActiveDurationMs=30000):
+        ''' Waits a LIT ICD device to become active. Will send a StayActive command to the device on active to allow human operations.
+
+        nodeId: Node ID of the LID ICD
+        stayActiveDurationMs: The duration in the StayActive command, in milliseconds
+
+        Returns:
+            - StayActiveResponse on success
+        '''
+        await icd.WaitForCheckIn(self._fabricIndex, nodeid)
```
tcarmelveilleux, line 879: This could pend for a really really long time. There needs to be a time-out for this overall API

src/controller/python/chip/ChipDeviceCtrl.py
```diff
@@ -1979,6 +2022,35 @@ def SetCheckMatchingFabric(self, check: bool):
             lambda: self._dmLib.pychip_DeviceController_SetCheckMatchingFabric(check)
         ).raise_on_error()
 
+    def EnableICDRegistration(self, parameters: typing.Optional[ICDRegistrationParameters] = None):
```
tcarmelveilleux, line 2025: Why allow "default' arguments? Suggest requiring explicit values to avoid "hidden" behavior

src/controller/python/chip/icd/PyChipCheckInDelegate.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+    if (mCallback != nullptr)
+    {
+        mCallback(clientInfo.peer_node.GetFabricIndex(), clientInfo.peer_node.GetNodeId());
+    }
+}
+
+extern "C" void pychip_CheckInDelegate_SetOnCheckInCompleteCallback(PyChipCheckInDelegate::OnCheckInCompleteCallback * callback)
+{
+    PyChipCheckInDelegate::GetInstance().SetOnCheckInCompleteCallback(callback);
```
tcarmelveilleux, line 38: This is not thread-safe, as the callback is always called from Matter stack context, but this call could be done in Python interpreter context. Consider taking the chip lock.

src/controller/python/chip/icd/PyChipCheckInDelegate.h
```diff
@@ -0,0 +1,40 @@
...
+public:
+    using OnCheckInCompleteCallback = void(uint8_t fabricIndex, uint64_t nodeId);
+
+    virtual ~PyChipCheckInDelegate() = default;
+
+    void OnCheckInComplete(const chip::app::ICDClientInfo & clientInfo) override;
+
+    void SetOnCheckInCompleteCallback(OnCheckInCompleteCallback * callback) { mCallback = callback; }
+
+    static PyChipCheckInDelegate & GetInstance() { return sInstance; }
```
tcarmelveilleux, line 34: Is there a reason this must be a singleton? Is there no clean way to inject a single instance without needing to access globally?

src/controller/python/chip/icd/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+
+from chip import native
+
+_OnCheckInCompleteFunct = CFUNCTYPE(None, c_uint8, c_uint64)
+
+_OnCheckInCompleteWaitListLock = Lock()
+_OnCheckInCompleteWaitList = dict()
+
+
+@dataclass
```
tcarmelveilleux, line 32: Please don't add complex logic and globals to `__init__.py`. Use `__init__.py` just to manage exports for `from xxx import *`. This logic can be put in a properly named file, preferably inside python/chip. I don't see a need to sub-package this under `python/chip/icd`.

src/controller/python/chip/icd/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+class OnCheckInCompleteParams:
+    fabricIndex: int
+    nodeId: int
+
+
+@_OnCheckInCompleteFunct
+def _OnCheckInComplete(fabricIndex: int, nodeId: int):
+    callbacks = list()
```
tcarmelveilleux, line 40: `= []`

src/controller/python/chip/icd/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+    if _initialized:
+        return
+
+    libraryHandle = native.GetLibraryHandle()
+    libraryHandle.pychip_CheckInDelegate_SetOnCheckInCompleteCallback.restype = None
+    libraryHandle.pychip_CheckInDelegate_SetOnCheckInCompleteCallback.argtypes = [_OnCheckInCompleteFunct]
+
+    libraryHandle.pychip_CheckInDelegate_SetOnCheckInCompleteCallback(_OnCheckInComplete)
+
+    _initialized = True
```
tcarmelveilleux, line 63: Please do not separately register this feature and include it in basic python controller support if at all possible.

src/controller/python/ChipDeviceController-ScriptBinding.cpp
yunhanw-google: it is used to re-trigger the inactive subscription(mpImEngine->OnActiveModeNotification(mICDClientInfo.peer_node);) via DefaultCheckInDelegate/RefreshKeySender, we need look for the right readClient for subscription based upon the mpActiveReadClientList, this list is coupled with IM engine at this moment. 

void InteractionModelEngine::OnActiveModeNotification(ScopedNodeId aPeer)
{
    for (ReadClient * pListItem = mpActiveReadClientList; pListItem != nullptr;)
    {
        auto pNextItem = pListItem->GetNextClient();
        // It is possible that pListItem is destroyed by the app in OnActiveModeNotification.
        // Get the next item before invoking `OnActiveModeNotification`.
        if (ScopedNodeId(pListItem->GetPeerNodeId(), pListItem->GetFabricIndex()) == aPeer)
        {
            pListItem->OnActiveModeNotification();
        }
        pListItem = pNextItem;
    }
}

https://github.com/project-chip/connectedhomeip/blob/master/src/app/icd/client/DefaultCheckInDelegate.cpp#L58
https://github.com/project-chip/connectedhomeip/blob/master/src/app/icd/client/RefreshKeySender.cpp#L67

src/controller/python/ChipDeviceController-ScriptBinding.cpp
yunhanw-google: it is used to re-trigger the inactive subscription(mpImEngine->OnActiveModeNotification(mICDClientInfo.peer_node);) via DefaultCheckInDelegate/RefreshKeySender, we need look for the right readClient for subscription based upon the mpActiveReadClientList, this list is coupled with IM engine at this moment.

void InteractionModelEngine::OnActiveModeNotification(ScopedNodeId aPeer)
{
for (ReadClient * pListItem = mpActiveReadClientList; pListItem != nullptr;)
{
auto pNextItem = pListItem->GetNextClient();
// It is possible that pListItem is destroyed by the app in OnActiveModeNotification.
// Get the next item before invoking OnActiveModeNotification.
if (ScopedNodeId(pListItem->GetPeerNodeId(), pListItem->GetFabricIndex()) == aPeer)
{
pListItem->OnActiveModeNotification();
}
pListItem = pNextItem;
}
}

https://github.com/project-chip/connectedhomeip/blob/master/src/app/icd/client/DefaultCheckInDelegate.cpp#L58
https://github.com/project-chip/connectedhomeip/blob/master/src/app/icd/client/RefreshKeySender.cpp#L67

src/controller/python/ChipDeviceController-ScriptBinding.cpp
yunhanw-google: updated.

src/controller/python/ChipDeviceController-ScriptBinding.cpp
yunhanw-google: updated.

src/controller/python/chip/icd/PyChipCheckInDelegate.h
yunhanw-google: yes, i think we can initialize this inside ChipDeviceController-ScriptDevicePairingDelegate.cpp

src/controller/python/ChipDeviceController-ScriptBinding.cpp
erjiaqing: removed.

src/controller/python/ChipDeviceController-ScriptBinding.cpp
erjiaqing: removed.

src/controller/python/chip/icd/PyChipCheckInDelegate.h
erjiaqing: Done by moving to ChipDeviceController-ScriptBinding.cpp

I think we currently don't have a global objects holder for python, but anyway the global object are at a single place for now.

src/controller/python/ChipDeviceController-ScriptDevicePairingDelegate.h
erjiaqing: Same as above.

src/controller/python/ChipDeviceController-ScriptBinding.cpp
yunhanw-google: Follow-up: https://github.com/project-chip/connectedhomeip/issues/33723

src/controller/python/ChipDeviceController-ScriptDevicePairingDelegate.cpp
erjiaqing: Updated.

src/controller/python/chip/ChipDeviceCtrl.py
```diff
@@ -98,6 +99,21 @@ class NOCChain:
...
+    checkInNodeId: typing.Optional[int]
+    monitoredSubject: typing.Optional[int]
+    stayActiveMs: typing.Optional[int]
+
+    class CStruct(Structure):
+        _fields_ = [('symmetricKey', c_char_p), ('symmetricKeyLength', c_size_t), ('checkInNodeId',
+                                                                                   c_uint64), ('monitoredSubject', c_uint64), ('stayActiveMsec', c_uint32)]
+
+    def to_c(self):
+        return ICDRegistrationParameters.CStruct(self.symmetricKey, len(self.symmetricKey), self.checkInNodeId, self.monitoredSubject, self.stayActiveMs)
```
tcarmelveilleux, line 114: If the value is None for any of the fields, we have to use default value that is not zero, but rather the invalid value.

src/controller/python/ChipDeviceController-ScriptDevicePairingDelegate.cpp
```diff
@@ -19,12 +19,17 @@
...
+#include <app/icd/client/DefaultICDClientStorage.h>
 #include <controller/python/chip/native/PyChipError.h>
 #include <setup_payload/ManualSetupPayloadGenerator.h>
 #include <setup_payload/QRCodeSetupPayloadGenerator.h>
 
 #include <string>
 
+extern chip::app::DefaultICDClientStorage sICDClientStorage;
+extern chip::Controller::CommissioningParameters sCommissioningParameters;
+extern uint8_t sICDSymmetricKey[chip::Crypto::kAES_CCM128_Key_Length];
```
tcarmelveilleux, line 31: Direct global access should not be done. There should be a struct with the necessary state and a way to get a pointer to it. This goes across the wall. It can't be called `sXxx` if it's not static. It would have needed to be `gXxx`

tcarmelveilleux (approved):

src/controller/python/chip/ChipDeviceCtrl.py
yunhanw-google: https://github.com/project-chip/connectedhomeip/issues/33732, immediate follow-up with ticket this week here.

src/controller/python/ChipDeviceController-ScriptDevicePairingDelegate.cpp
yunhanw-google: https://github.com/project-chip/connectedhomeip/issues/33733

PR #33535, opened 2024-05-21
[Darwin] MTRDevice should throttle writes to the attribute storage

jtung-apple:
This change makes MTRDevice throttle writing to the attribute storage. It has three parts:
1. Add a delay between receiving a report and writing to the storage. If a report comes in during that delay, the writing is further pushed out, until a larger limit, at which point the change set is written to storage, and the logic repeats.
2. If a device is detected to be reporting frequently, the above delay thresholds are multiplied by an increasing multiplier.
3. If a device is detected to be reporting excessively, all writes to the storage is suspended until the device is reporting at a reasonable interval again.

Currently working on unit tests, and fixing the unit tests that this broke.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -360,6 +360,7 @@ static NSString * const sLastInitialSubscribeLatencyKey = @"lastInitialSubscribe
 - (BOOL)unitTestPretendThreadEnabled:(MTRDevice *)device;
 - (void)unitTestSubscriptionPoolDequeue:(MTRDevice *)device;
 - (void)unitTestSubscriptionPoolWorkComplete:(MTRDevice *)device;
+- (void)unitTestClusterDataPersisted:(MTRDevice *)device;
 @end
 #endif
 
@@ -400,6 +401,23 @@ static NSString * const sLastInitialSubscribeLatencyKey = @"lastInitialSubscribe
     // Tracking of initial subscribe latency.  When _initialSubscribeStart is
     // nil, we are not tracking the latency.
     NSDate * _Nullable _initialSubscribeStart;
+
+    // Storage behavior configuration and variables to keep track of the logic
+    //  _clusterDataPersistenceFirstScheduledTime is used to track the start time of the delay between
+    //      report and persistence.
+    //  _mostRecentReportTimes is a list of the most recent report timestamps used for calculating
+    //      the running average time between reports.
+    //  _deviceReportingExcessivelyStartTime tracks when a device starts reporting excessively.
+    //  _reportToPersistenceDelayCurrentMultiplier is the current multiplier that is calculated when a
+    //      report comes in.
+    MTRDeviceStorageBehaviorConfiguration * _storageBehaviorConfiguration;
+    NSDate * _Nullable _clusterDataPersistenceFirstScheduledTime;
+    NSMutableArray<NSDate *> * _mostRecentReportTimes;
+    NSDate * _Nullable _deviceReportingExcessivelyStartTime;
+    double _reportToPersistenceDelayCurrentMultiplier;
+
+    // System time change observer reference
+    id _systemTimeChangeObserverToken;
 }
 
 - (instancetype)initWithNodeID:(NSNumber *)nodeID controller:(MTRDeviceController *)controller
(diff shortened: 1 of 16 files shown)


woody-apple (approved):

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -423,11 +441,27 @@ - (instancetype)initWithNodeID:(NSNumber *)nodeID controller:(MTRDeviceControlle
         }
         _clusterDataToPersist = nil;
         _persistedClusters = [NSMutableSet set];
+
+        // If there is a data store, make sure we have an observer to
```
bzbarsky-apple, line 445: "to" what?

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -1280,37 +1314,262 @@ - (void)_handleReportBegin
...
+- (void)_resetStorageBehaviorState
+{
+    os_unfair_lock_assert_owner(&self->_lock);
+
+    _clusterDataPersistenceFirstScheduledTime = nil;
+    _mostRecentReportTimes = nil;
+    _deviceReportingExcessivelyStartTime = nil;
+    _reportToPersistenceDelayCurrentMultiplier = 1;
+
+    if (_persistedClusters) {
```
bzbarsky-apple, line 1550: This is an odd check.  Why is this not just unconditional (and then no-op if there is nothing to persist right now)?  Needs comments, at least.

src/darwin/Framework/CHIP/MTRDeviceStorageBehaviorConfiguration.h
```diff
@@ -0,0 +1,98 @@
...
+ *    limitations under the License.
+ */
+
+#import <Foundation/Foundation.h>
+#import <Matter/MTRDefines.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ * Class that configures how MTRDevice objects persist its attributes to storage, so as to not
```
bzbarsky-apple, line 23: "their" not "its", since it's "objects"?

src/darwin/Framework/CHIP/MTRDeviceStorageBehaviorConfiguration.h
```diff
@@ -0,0 +1,98 @@
...
+ * MTRDevice will also track recentReportTimesMaxCount number of report times. If the running
+ * average time between reports dips below timeBetweenReportsTooShortThreshold, a portion of the
+ * reportToPersistenceDelayMaxMultiplier will be applied to both the reportToPersistenceDelayTime
+ * and reportToPersistenceDelayTimeMax. The multiplier will reach the max when the average time
+ * between reports reach timeBetweenReportsTooShortMinThreshold.
+ *
+ * When the running average time between reports dips below timeBetweenReportsTooShortMinThreshold
+ * for the first time, the time will be noted. If the device remains in this state for longer than
+ * deviceReportingExcessivelyIntervalThreshold, persistence will stop until the average time between
+ * reports go back above timeBetweenReportsTooShortMinThreshold.
```
bzbarsky-apple, line 68: "goes back".

In terms of the implementation, if the device is reporting excessively for a bit, and then stops reporting entirely for a long time, shouldn't we persist at some point?  Right now, I don't think we do....

Seems to me like we should compute when the average time would go back above if we got no reports until then and schedule something to persist at that point or something.

src/darwin/Framework/CHIP/MTRDeviceStorageBehaviorConfiguration.h
```diff
@@ -0,0 +1,98 @@
...
+ * is disabled.
+ */
+@property (nonatomic, assign) BOOL disableStorageBehaviorOptimization;
+
+/**
+ * If any of these properties are set to be out of the documented limits, these default values will
+ * be used to replace all of them:
+ *
+ * reportToPersistenceDelayTimeDefault (15)
+ * reportToPersistenceDelayTimeMaxDefault (20 * kReportToPersistenceDelayTimeDefault)
```
bzbarsky-apple, line 82: kReportToPersistenceDelayTimeDefault is not a concept in this API... Should just say `20 * 15`, I think.

PR #33537, opened 2024-05-21
Updating logs in darwin framework to enable more by default

woody-apple:
Need to log more :)

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -118,8 +118,8 @@ public:
         UnsolicitedMessageFromPublisherHandler unsolicitedMessageFromPublisherHandler, ReportBeginHandler reportBeginHandler,
         ReportEndHandler reportEndHandler)
         : MTRBaseSubscriptionCallback(attributeReportCallback, eventReportCallback, errorCallback, resubscriptionCallback,
-            subscriptionEstablishedHandler, onDoneHandler, unsolicitedMessageFromPublisherHandler, reportBeginHandler,
-            reportEndHandler)
+              subscriptionEstablishedHandler, onDoneHandler, unsolicitedMessageFromPublisherHandler, reportBeginHandler,
+              reportEndHandler)
     {
     }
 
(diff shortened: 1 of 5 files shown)


kiel-apple (approved):

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -2099,22 +2099,22 @@ static BOOL AttributeHasChangesOmittedQuality(MTRAttributePath * attributePath)
             while (readRequestsNext.count) {
                 // Can only read up to 9 paths at a time, per spec
                 if (readRequestsCurrent.count >= 9) {
-                    MTR_LOG_INFO("Batching read attribute work item [%llu]: cannot add more work, item is full [0x%016llX:%@:0x%llx:0x%llx]", workItemID, nodeID.unsignedLongLongValue, endpointID, clusterID.unsignedLongLongValue, attributeID.unsignedLongLongValue);
+                    MTR_LOG_ERROR("Batching read attribute work item [%llu]: cannot add more work, item is full [0x%016llX:%@:0x%llx:0x%llx]", workItemID, nodeID.unsignedLongLongValue, endpointID, clusterID.unsignedLongLongValue, attributeID.unsignedLongLongValue);
```
bzbarsky-apple, line 2102: This is not an error condition.  This will happen any time there are more than 9 reads queued up, and all that happens is the next read will happen as a separate request.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -2099,22 +2099,22 @@ static BOOL AttributeHasChangesOmittedQuality(MTRAttributePath * attributePath)
...
                     return outcome;
                 }
 
                 // if params don't match then they cannot be merged
                 if (![readRequestsNext[0][MTRDeviceReadRequestFieldParamsIndex]
                         isEqual:readRequestsCurrent[0][MTRDeviceReadRequestFieldParamsIndex]]) {
-                    MTR_LOG_INFO("Batching read attribute work item [%llu]: cannot add more work, parameter mismatch [0x%016llX:%@:0x%llx:0x%llx]", workItemID, nodeID.unsignedLongLongValue, endpointID, clusterID.unsignedLongLongValue, attributeID.unsignedLongLongValue);
+                    MTR_LOG_ERROR("Batching read attribute work item [%llu]: cannot add more work, parameter mismatch [0x%016llX:%@:0x%llx:0x%llx]", workItemID, nodeID.unsignedLongLongValue, endpointID, clusterID.unsignedLongLongValue, attributeID.unsignedLongLongValue);
```
bzbarsky-apple, line 2109: This is not an error condition.  Just two reads that could not be combined into a single read request.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -2099,22 +2099,22 @@ static BOOL AttributeHasChangesOmittedQuality(MTRAttributePath * attributePath)
...
                     return outcome;
                 }
 
                 // merge the next item's first request into the current item's list
                 auto readItem = readRequestsNext.firstObject;
                 [readRequestsNext removeObjectAtIndex:0];
                 [readRequestsCurrent addObject:readItem];
-                MTR_LOG_INFO("Batching read attribute work item [%llu]: added %@ (now %tu requests total) [0x%016llX:%@:0x%llx:0x%llx]",
+                MTR_LOG_ERROR("Batching read attribute work item [%llu]: added %@ (now %tu requests total) [0x%016llX:%@:0x%llx:0x%llx]",
```
bzbarsky-apple, line 2117: This is _definitely_ not an error condition.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -2252,7 +2252,7 @@ - (void)writeAttributeWithEndpointID:(NSNumber *)endpointID
             // with the later one.
             if (![writeRequestsNext[0][MTRDeviceWriteRequestFieldPathIndex]
                     isEqual:writeRequestsCurrent[0][MTRDeviceWriteRequestFieldPathIndex]]) {
-                MTR_LOG_INFO("Batching write attribute work item [%llu]: cannot replace with next work item due to path mismatch", workItemID);
+                MTR_LOG_ERROR("Batching write attribute work item [%llu]: cannot replace with next work item due to path mismatch", workItemID);
```
bzbarsky-apple, line 2255: Also not an error.

src/darwin/Framework/CHIP/MTRLogging_Internal.h
```diff
@@ -22,5 +22,5 @@
 
 #define MTR_LOG_ERROR(msg, ...) ChipLogError(NotSpecified, msg, ##__VA_ARGS__)
 #define MTR_LOG_DEFAULT(msg, ...) ChipLogProgress(NotSpecified, msg, ##__VA_ARGS__)
-#define MTR_LOG_INFO(msg, ...) ChipLogDetail(NotSpecified, msg, ##__VA_ARGS__)
+#define MTR_LOG_INFO(msg, ...) ChipLogProgress(NotSpecified, msg, ##__VA_ARGS__)
```
bzbarsky-apple, line 25: Thing is, these log levels are _supposed_ to correspond to the error/default/info/debug values of `os_log_type_t` and have the same semantics.

What we should probably do is remove uses of MTR_LOG_INFO (and maybe the macro?), and replace with default/debug, if we think it's being misused.

Note that right now MTR_LOG_DEBUG in fact ends up using OS_LOG_TYPE_INFO (just like MTR_LOG_INFO does), but that's because it's being squeezed through the ChipLog* bits that have a different number of levels.

In particular, I am not sure whether we want the "got attribute report" logs in MTRDevice to be TYPE_INFO or TYPE_DEFAULT; those can be pretty spammy.

That said, I do agree that a bunch of stuff we log as INFO should have been DEFAULT; one option is to take this PR for now (with the problems above fixed), but then do a followup to get rid of MTR_LOG_INFO and audit all the remaining callers....

bzbarsky-apple (dismissed):

PR #33542, opened 2024-05-22
[ESP32] Remove the Lambda function from ConnectivityManager implementation.

shripad621git:
**Change Overview:**
- Removed  ScheduleLambda function  in ESP32 ConnectivityManager to decrease the size of event queue on the platform.
- Both the ScheduleLambda and OnNetworkStatusChange run in the CHIP context  irrespective of the lambda function.

**Testing**
- Tested the lighting-app esp32 and commissioned and controlled the device successfully.
- Tested for the functionality for OnNetworkStatusChange with and without scheduling the lamda.

### DIFF
--- a/src/platform/ESP32/ConnectivityManagerImpl_WiFi.cpp
+++ b/src/platform/ESP32/ConnectivityManagerImpl_WiFi.cpp
@@ -780,7 +780,7 @@ void ConnectivityManagerImpl::ChangeWiFiStationState(WiFiStationState newState)
         ChipLogProgress(DeviceLayer, "WiFi station state change: %s -> %s", WiFiStationStateToStr(mWiFiStationState),
                         WiFiStationStateToStr(newState));
         mWiFiStationState = newState;
-        SystemLayer().ScheduleLambda([]() { NetworkCommissioning::ESPWiFiDriver::GetInstance().OnNetworkStatusChange(); });
+        NetworkCommissioning::ESPWiFiDriver::GetInstance().OnNetworkStatusChange();
     }
 }
 


src/platform/ESP32/ConnectivityManagerImpl_WiFi.cpp
```diff
@@ -780,7 +780,7 @@ void ConnectivityManagerImpl::ChangeWiFiStationState(WiFiStationState newState)
         ChipLogProgress(DeviceLayer, "WiFi station state change: %s -> %s", WiFiStationStateToStr(mWiFiStationState),
                         WiFiStationStateToStr(newState));
         mWiFiStationState = newState;
-        SystemLayer().ScheduleLambda([]() { NetworkCommissioning::ESPWiFiDriver::GetInstance().OnNetworkStatusChange(); });
+        NetworkCommissioning::ESPWiFiDriver::GetInstance().OnNetworkStatusChange();
```
andy31415, line 783: Is the change worth it? I am unsure about sideffects, however this may have been done on purpose, usually to decrease stack usage or remove recursive calls (can OnNetworkStatusChange eventually call ChangeWiFiStationState?). 

I am wondering if the overhead saved by not scheduling the lambda is worth it in investigating if this is a correct change or not.

shripad621git: I have checked that there ate no recursive calls i.e this scenario of  `OnNetworkStatusChange eventually call ChangeWiFiStationState does not hold`. Although the ScheduleLamda might decrease the stack usage to some extent, but scheduling it also may fill up the platform queue due to lambda events  and the application crashes due to PostEventOrDie in case of failure to post event to the platfiorm queue .
@wqx6, can you please vaildate this as well?

andy31415: Ok, please add this to the summary as to "why". It says "overhead" however it does not explain what overhead. I believe saying that this is done to decrease the size of the event queue on the platform would be useful.

bzbarsky-apple, line 783: Is this code already running on the Matter event queue or with the Matter lock otherwise held?  If not, is it ok to call OnNetworkStatusChange() without the Matter lock held?

shripad621git: The scheduled function `OnNetworkingStatusChange` and the function in which the lambda is scheduled i.e `ChangeWifiStation` state run in the CHIP context i,e with task name as 'chip' irrespective of the lambda.

shripad621git: Done.

andy31415 (approved):

src/platform/ESP32/ConnectivityManagerImpl_WiFi.cpp
bzbarsky-apple: Great, thanks!

bzbarsky-apple (approved):

PR #33546, opened 2024-05-22
Fix for test hanging when no ProductID/VendorID is present

adamb-q:
TestChipCryptoPAL is hanging when no Product ID or Vendor ID is present during TestVIDPID_x509Extraction. Wrong value is checked


src/crypto/tests/TestChipCryptoPAL.cpp
```diff
@@ -2808,13 +2808,13 @@ TEST_F(TestChipCryptoPAL, TestVIDPID_x509Extraction)
         EXPECT_EQ(vidpid.mProductId.HasValue(), testCase.expectedPidPresent);
 
         // If present, make sure the VID matches expectation.
-        if (testCase.expectedVidPresent)
+        if (vidpid.mVendorId.HasValue())
```
andy31415, line 2811: There is an expectation above of `EXPECT_EQ(vidpid.mVendorId.HasValue(), testCase.expectedVidPresent);`

If that expect passes, the logic here should be identical. If the test is hanging due to using of `Value`, we should probably use ASSERT_EQ instead?

woody-apple (approved):

PR #33556, opened 2024-05-22
Use a valid command id in TC_IDM_1_4.py

Apollon77:
The test TC_IDM_1_4.py used 0xffff_ffff as test command id. In fact this is not valid MEI according to specification.

While chip seems to validate this fact later the test works.

In matter.js such datatype validations are handled earlier and thats why we throw an constraint error because the value is wrong data type wise. 

I did not found any place in the specification where it is defined when such "semantic" data value checks should be done (unless some very specific error cases in some adapters).

I think this is a theoretical topic in this case because it will hopefully never happen in real live devices 8and when it does the exact error should be irrelevant) for this test it matters because the test fails.

With the change of the command ID to be a still unknown, but valid, value the issue is gone. I also think that tests should still use basically "correct" values. ;-)

Thank you for considering this change.

### DIFF
--- a/src/python_testing/TC_IDM_1_4.py
+++ b/src/python_testing/TC_IDM_1_4.py
@@ -74,7 +74,11 @@ class TC_IDM_1_4(MatterBaseTest):
         cap_for_batch_commands = 100
         number_of_commands_to_send = min(max_paths_per_invoke + 1, cap_for_batch_commands)
 
+        # Use a valid (according to MEI definition) command-id (in this case belonging to Test Vendor MC with prefix 0xfff4)
+        # which should never be existing on a prodiction device. We use this decodable id to prevent hitting issues with the
+        # specification being not clearly defined in respect to decoding vs processing the invoke requests.
         invalid_command_id = 0xfff4_00ff
+
         list_of_commands_to_send = []
         for endpoint_index in range(number_of_commands_to_send):
             # Using Toggle command to form the base as it is a command that doesn't take
--- a/src/python_testing/TC_IDM_1_4.py
+++ b/src/python_testing/TC_IDM_1_4.py
@@ -74,7 +74,7 @@ class TC_IDM_1_4(MatterBaseTest):
         cap_for_batch_commands = 100
         number_of_commands_to_send = min(max_paths_per_invoke + 1, cap_for_batch_commands)
 
-        invalid_command_id = 0xffff_ffff
+        invalid_command_id = 0xfff4_00ff
         list_of_commands_to_send = []
         for endpoint_index in range(number_of_commands_to_send):
             # Using Toggle command to form the base as it is a command that doesn't take


woody-apple (approved):

src/python_testing/TC_IDM_1_4.py
```diff
@@ -74,7 +74,7 @@ async def test_TC_IDM_1_4(self):
         cap_for_batch_commands = 100
         number_of_commands_to_send = min(max_paths_per_invoke + 1, cap_for_batch_commands)
 
-        invalid_command_id = 0xffff_ffff
+        invalid_command_id = 0xfff4_00ff
```
tehampson, line 77: Please add a comment above saying that the prefix 0xfff4 is a reserved as "Test Vendor MC", and that from the perspective of a test device that is an invalid command id.

src/python_testing/TC_IDM_1_4.py
```diff
@@ -74,7 +74,7 @@ async def test_TC_IDM_1_4(self):
...
         list_of_commands_to_send = []
         for endpoint_index in range(number_of_commands_to_send):
```
tehampson, line 79: In general I would like to understand, are you having issues on the device receiving this command or the device sending the command?

If you are having issues receiving the command on the server, then I think you are not conforming to the spec. My understanding is that "8.8.2.3. Incoming Invoke Request Action" is intended to be followed in the order stated. You are supposed to check that:

```
If this action contains more CommandDataIB elements in the InvokeRequests list than are
supported by the device (see MaxPathsPerInvoke), then a Status Response action with the
INVALID_ACTION Status Code SHALL be submitted to the message layer and this interaction
SHALL terminate.
```

Before even looking at any of the requests and their contents

tehampson (dismissed):
Adding Request changes for the moment. If my understanding of section "8.8.2.3. Incoming Invoke Request Action" is correct, the order of operations are intended to be followed in that order. Adjusting the test here because a device is failing in an unexpected way, I don't think is the right way to fix this.

From my perspective we should really be able to fuzz any sort of invalid command path for the values and we should expect the same non-path specific error. It just so happens that this value we are sending here is tripping up a device, but some other invalid path is not.

Before I remove my request for change I would like confirmation that 8.8.2.3 is not intended to be followed in that order otherwise I think this is catching a real out of order spec violating issue.

Note: Lots of the parts of this test are intentionally sending out spec violating thing making sure the DUT behaves appropriately

Apollon77:
@tehampson Hi and thank you for looking at my change.
Ok, when I get your comment right then we hit implementation specific topics or unclear specification requirements ;-) (no offense!)

Your understanding of the specification as you reference relies on an iterative Tlv decoding/determination, so first you decode just the "action field" listed in 8.8.2.1 (matter spec 1.3) and not looking into the list content at all ... then you look into the list content when handling the values.

That leads to the followup up question how an "semantically invalid MEI" should be handled afterwards? Falls it into the "Valid Command path" check (spec do not tell anything here about invalid values, but field combinations) or should it be "just" a non-existing path (so completely ignoring that the value is invalid semantically regarding the datamodel definition)?
Because there agsin it would be a difference in the returned responsecodes and response at all.

On a pure "data type" level it is correct: eg command-id is just defined as a uint32, not mentioning any semantics like MEI value range limitations which lead to the question above: Where and when such validation should be done and with which results/consequences?

One example: ACL test TC_ACL_2_4 step 39 requires to decline an invalid cluster id (0xFFFFFFFF) with a Constraint error" ... and spec just tells very generically:

> The Access Control Cluster SHALL fail to write, and return an appropriate error, if an attempt is made to create or update an Access Control Entry or Access Control Extension such that it would have invalid contents.

So in this case "invalid contents" is also considered an invalid MEI value which is like an invalid value.

Thank you,

Ingo

src/python_testing/TC_IDM_1_4.py
tehampson: Please also include in the comment that the spec is undefined with respect to decoding vs processing the invoke request so we use a decodable, so we used a valid id that should not be implemented on a production device as it is a Test Vendor MC for prefix 0xfff4

Apollon77:
@tehampson I looked into it. In fact it would be awesome to have options to handle such validations generically, but in fact it's not. ACL cluster, as already told, expects a contraint error when any invalid cluster/group/deviceType id is provided in acl data on a write interaction.

I adjusted my side already to be like the spec seems to be interpreted: The decoding mainly checks only data types, so decoding is not checking values on a semantic level. All more deep checks need to be done later in the code flows.

So with this topic clarified I would mainly misss in the specs how an "semantically invalid command id (or clusterid)" should be handled fpor invokes (or other for read/write/subscribe. Here, again as I interpret the specs because it is not defined, this is simply irrelevant because for device side such invalid values should never be used for cluster-ids or such and so this is simply an "unsupported cluster" (in that example) error which would happen. means: no explicit validation required in this case.

Sure, anything just assumptions on how I think the specs want to get interpreted ;-)

If you do not see that different we could close here because in fact it is just another unusual edge case checked here which could be relevant to just make sure noone else stumbles over it with a differnt understanding.

if the spec could use an enhancement to clarify such things is not on me to decide.

WDYT?

Thank you very much for the open and constructive chat about the topic!

Apollon77:
Alright, before we iterate in code commits ... would that comment fulfill your requirements:

# Use a valid (according to MEI definition) command-id (in this case belonging to Test Vendor MC with prefix 0xfff4)
# which should never be existing on a prodiction device. We use this decodable id to prevent hitting issues with the
# specification being not clearly defined in respect to decoding vs processing the invoke requests.

WDYT? Any better ideas?

tehampson:
> Alright, before we iterate in code commits ... would that comment fulfill your requirements:
> 
> ```
> # Use a valid (according to MEI definition) command-id (in this case belonging to Test Vendor MC with prefix 0xfff4)
> # which should never be existing on a prodiction device. We use this decodable id to prevent hitting issues with the
> # specification being not clearly defined in respect to decoding vs processing the invoke requests.
> ```
> 
> WDYT? Any better ideas?

Looks good to me. Upload that and I will approve

src/python_testing/TC_IDM_1_4.py
Apollon77: added

tehampson (approved):

src/python_testing/TC_IDM_1_4.py
tehampson: Thanks!

PR #33559, opened 2024-05-22
Do not set up subscription when setting a delegate for a MTRDevice wi…

### LINKED ISSUE #33557: Do not set up subscription for remote controllers when delegate is set
Currently when we set delegate on a MTRDevice, we always set up subscriptions by default. However for remote controllers (over XPC) we should not be setting up subscriptions.

nivi-apple:
…th a remote controller over XPC

- Add tests for testing the behavior when a MTRDevice is created with an XPC controller

Fixes: #33557

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -690,7 +690,8 @@ static NSString * const sLastInitialSubscribeLatencyKey = @"lastInitialSubscribe
     MTR_LOG_INFO("%@ setDelegate %@", self, delegate);
 
     // We should not set up a subscription for device controllers over XPC.
-    BOOL setUpSubscription = ![_deviceController isKindOfClass:MTRDeviceControllerOverXPC.class];;
+    BOOL setUpSubscription = ![_deviceController isKindOfClass:MTRDeviceControllerOverXPC.class];
+    ;
 
     // For unit testing only
 #ifdef DEBUG
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -689,7 +689,8 @@ static NSString * const sLastInitialSubscribeLatencyKey = @"lastInitialSubscribe
 {
     MTR_LOG_INFO("%@ setDelegate %@", self, delegate);
 
-    BOOL setUpSubscription = YES;
+    // We should not set up a subscription for device controllers over XPC.
+    BOOL setUpSubscription = ![_deviceController isKindOfClass:MTRDeviceControllerOverXPC.class];;
 
     // For unit testing only
 #ifdef DEBUG
@@ -932,6 +933,14 @@ static NSString * const sLastInitialSubscribeLatencyKey = @"lastInitialSubscribe
         MTR_LOG_DEFAULT("%@ internal state change %lu => %lu", self, static_cast<unsigned long>(lastState), static_cast<unsigned long>(state));
     }
 }
+
+#ifdef DEBUG
+- (MTRInternalDeviceState)_getInternalState
+{
+    std::lock_guard lock(self->_lock);
+    return _internalDeviceState;
+}
+#endif
 
 // First Time Sync happens 2 minutes after reachability (this can be changed in the future)
 #define MTR_DEVICE_TIME_UPDATE_INITIAL_WAIT_TIME_SEC (60 * 2)
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -685,7 +685,8 @@ static NSString * const sLastInitialSubscribeLatencyKey = @"lastInitialSubscribe
 #define MTR_DEVICE_SUBSCRIPTION_MAX_INTERVAL_MIN (10 * 60) // 10 minutes (for now)
 #define MTR_DEVICE_SUBSCRIPTION_MAX_INTERVAL_MAX (60 * 60) // 60 minutes
 
--(BOOL)_subscriptionsAllowed {
+- (BOOL)_subscriptionsAllowed
+{
     return ![_deviceController isKindOfClass:MTRDeviceControllerOverXPC.class];
 }
 
@@ -1693,8 +1694,7 @@ static NSString * const sLastInitialSubscribeLatencyKey = @"lastInitialSubscribe
 {
     os_unfair_lock_assert_owner(&self->_lock);
 
-    if (![self _subscriptionsAllowed])
-    {
+    if (![self _subscriptionsAllowed]) {
         MTR_LOG("_setupSubscription: Subscriptions not allowed. Do not set up subscription");
         return;
     }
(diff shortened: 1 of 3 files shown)


src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -689,7 +689,9 @@ - (void)setDelegate:(id<MTRDeviceDelegate>)delegate queue:(dispatch_queue_t)queu
 {
     MTR_LOG_INFO("%@ setDelegate %@", self, delegate);
 
-    BOOL setUpSubscription = YES;
+    // We should not set up a subscription for device controllers over XPC.
+    BOOL setUpSubscription = ![_deviceController isKindOfClass:MTRDeviceControllerOverXPC.class];
+    ;
```
jtung-apple, line 694: Extraneous `;`

nivi-apple: :)

woody-apple (approved):

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -685,11 +685,17 @@ - (void)_setDSTOffsets:(NSArray<MTRTimeSynchronizationClusterDSTOffsetStruct *>
...
+{
+    return ![_deviceController isKindOfClass:MTRDeviceControllerOverXPC.class];
+}
+
 - (void)setDelegate:(id<MTRDeviceDelegate>)delegate queue:(dispatch_queue_t)queue
 {
     MTR_LOG_INFO("%@ setDelegate %@", self, delegate);
 
-    BOOL setUpSubscription = YES;
+    // We should not set up a subscription for device controllers over XPC.
```
bzbarsky-apple, line 697: Shouldn't this comment be in _subscriptionsAllowed?

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -685,11 +685,17 @@ - (void)_setDSTOffsets:(NSArray<MTRTimeSynchronizationClusterDSTOffsetStruct *>
 #define MTR_DEVICE_SUBSCRIPTION_MAX_INTERVAL_MIN (10 * 60) // 10 minutes (for now)
 #define MTR_DEVICE_SUBSCRIPTION_MAX_INTERVAL_MAX (60 * 60) // 60 minutes
 
+- (BOOL)_subscriptionsAllowed
+{
+    return ![_deviceController isKindOfClass:MTRDeviceControllerOverXPC.class];
```
bzbarsky-apple, line 690: Shouldn't we use this helper in the other place where we are doing this check?

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -685,11 +685,17 @@ - (void)_setDSTOffsets:(NSArray<MTRTimeSynchronizationClusterDSTOffsetStruct *>
...
+}
+
 - (void)setDelegate:(id<MTRDeviceDelegate>)delegate queue:(dispatch_queue_t)queue
 {
     MTR_LOG_INFO("%@ setDelegate %@", self, delegate);
 
-    BOOL setUpSubscription = YES;
+    // We should not set up a subscription for device controllers over XPC.
+    BOOL setUpSubscription = [self _subscriptionsAllowed];
```
bzbarsky-apple, line 698: This is using `_deviceController` without locking.  That might be OK, but have you checked that?

src/darwin/Framework/CHIPTests/MTRDeviceTests.m
```diff
@@ -3618,6 +3618,36 @@ - (void)test034_TestMTRDeviceHistoricalEvents
...
+    __auto_type * delegate = [[MTRDeviceTestDelegate alloc] init];
+
+    XCTAssertTrue([device _getInternalState] == MTRInternalDeviceStateUnsubscribed);
+
+    delegate.onAttributeDataReceived = ^(NSArray<NSDictionary<NSString *, id> *> * attributeReport) {
+        [subscriptionExpectation fulfill];
+    };
+
+    [device setDelegate:delegate queue:queue];
+    [self waitForExpectations:@[ subscriptionExpectation ] timeout:30];
```
bzbarsky-apple, line 3646: So this test will always take 30 seconds to run?  That is pretty painful, in terms of actually running the tests locally....  Why do we need this bit at all, given the internal state check above?

src/darwin/Framework/CHIPTests/MTRDeviceTests.m
```diff
@@ -3618,6 +3618,36 @@ - (void)test034_TestMTRDeviceHistoricalEvents
...
+    __auto_type * device = [MTRDevice deviceWithNodeID:kDeviceId deviceController:remoteController];
+    dispatch_queue_t queue = dispatch_get_main_queue();
+
+    // We should not set up a subscription when creating a MTRDevice with a remote controller.
+    XCTestExpectation * subscriptionExpectation = [self expectationWithDescription:@"Subscription has been set up"];
+    subscriptionExpectation.inverted = YES;
+
+    __auto_type * delegate = [[MTRDeviceTestDelegate alloc] init];
+
+    XCTAssertTrue([device _getInternalState] == MTRInternalDeviceStateUnsubscribed);
```
bzbarsky-apple, line 3639: Why isn't this XCTAssertEqual?

src/darwin/Framework/CHIPTests/TestHelpers/MTRTestDeclarations.h
```diff
@@ -59,10 +59,31 @@ NS_ASSUME_NONNULL_BEGIN
 @end
 
 @interface MTRDevice (TestDebug)
+typedef NS_ENUM(NSUInteger, MTRInternalDeviceState) {
```
bzbarsky-apple, line 62: Why is this here?  Seems like we should just move the "internal device state" to a header that both MTRDevice.mm and the tests include instead of duplicating the enum and having problems when our duplication gets out of sync.

nivi-apple: done.

src/darwin/Framework/CHIPTests/MTRDeviceTests.m
nivi-apple: done.

src/darwin/Framework/CHIPTests/MTRDeviceTests.m
nivi-apple: i wanted to wait a few seconds before checking the internal state again and also make sure the inverted exception is not called. would 5 secs be ok or you are saying we do not need to wait just check the state after setting delegate?

src/darwin/Framework/CHIP/MTRDevice.mm
nivi-apple: done.

src/darwin/Framework/CHIP/MTRDevice.mm
nivi-apple: done.

PR #33560, opened 2024-05-23
Darwin: add helpers for essential attributes for logging; log unexpected C-quality attributes

kiel-apple:
Log a metric event when we see an unexpected update of a C-quality attribute

future work:
add these attributes to other Darwin-side logs? (later PR)

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -3653,7 +3653,7 @@ static BOOL AttributeHasChangesOmittedQuality(MTRAttributePath * attributePath)
 
 #pragma mark Log Help
 
-- (NSNumber *)_informationalNumberAtAttributePath:(MTRAttributePath *)attributePath
+- (nullable NSNumber *)_informationalNumberAtAttributePath:(MTRAttributePath *)attributePath
 {
     auto * cachedData = [self _cachedAttributeValueForPath:attributePath];
 
@@ -3662,12 +3662,11 @@ static BOOL AttributeHasChangesOmittedQuality(MTRAttributePath * attributePath)
         MTRDataKey : cachedData,
     }
                                                                     error:nil];
-    // REVIEWERS:  is it worth logging the `error` above?
 
     return attrReport.value;
 }
 
-- (NSNumber *)_informationalVendorID
+- (nullable NSNumber *)_informationalVendorID
 {
     auto * vendorIDPath = [MTRAttributePath attributePathWithEndpointID:@(kRootEndpointId)
                                                               clusterID:@(MTRClusterIDTypeBasicInformationID)
@@ -3676,7 +3675,7 @@ static BOOL AttributeHasChangesOmittedQuality(MTRAttributePath * attributePath)
     return [self _informationalNumberAtAttributePath:vendorIDPath];
 }
 
-- (NSNumber *)_informationalProductID
+- (nullable NSNumber *)_informationalProductID
 {
     auto * productIDPath = [MTRAttributePath attributePathWithEndpointID:@(kRootEndpointId)
                                                                clusterID:@(MTRClusterIDTypeBasicInformationID)
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -93,17 +93,17 @@ NSString * const MTRDataVersionKey = @"dataVersion";
 }
 @end
 
-// Stores essential-for-logging attributes immutably for use in logs
-@interface MTRDeviceEssentialAttributes : NSObject
+// convenience object for commonly-logged device attributes
+@interface MTRDeviceInformationalAttributes : NSObject
 @property (readonly) UInt16 vendorID;
 @property (readonly) UInt16 productID;
 @property (readonly) BOOL usesThread;
 
-- (void)addEssentialAttributesToCurrentMetricScope;
+- (void)addInformationalAttributesToCurrentMetricScope;
 
 @end
 
-@implementation MTRDeviceEssentialAttributes
+@implementation MTRDeviceInformationalAttributes
 
 - (instancetype)initWithVendorID:(UInt16)vendorID productID:(UInt16)productID usesThread:(BOOL)usesThread {
     self = [super init];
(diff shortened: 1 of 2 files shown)


src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -91,6 +93,39 @@ - (id)strongObject
 }
 @end
 
+// Stores essential-for-logging attributes immutably for use in logs
+@interface MTRDeviceEssentialAttributes : NSObject
+@property (readonly) UInt16 vendorID;
```
woody-apple, line 98: Why not allow these to be mutable? Also I forget, is atomic, assigned present by default? Note... we also should be using NSUIntegers, or NSNumber*

woody-apple: (Preferably NSNumber *, to follow other stuff)

kiel-apple: `atomic` is default, i think `assign` is default as well.  `NSNumber *` sounds fine.

kiel-apple: As far as mutability:  I was hoping to set this up as a set of attributes that are unchanging throughout the `MTRDevice`'s lifetime and therefore only have to be pulled from their respective clusters _once_ at a predictable point in the lifecycle.  Another valid option is to make the properties computed and pull them (from cache?) every time we want to log them.

I don't have a comprehensive picture of how often we want to use these properties in logs or how often those logs will be called.  I also am not sure about all scenarios where our attribute caches get cleared. I erred on the conservative side here, but that could be premature optimization on my part.  Thoughts?

kiel-apple: (Discussed earlier today - probably will want to frequently log some changeable properties soon, so immutability isn't going to work out in the long run.)

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -1426,6 +1465,8 @@ - (void)_handleEventReport:(NSArray<NSDictionary<NSString *, id> *> *)eventRepor
         BOOL isStartUpEvent = (eventPath.cluster.unsignedLongValue == MTRClusterIDTypeBasicInformationID)
             && (eventPath.event.unsignedLongValue == MTREventIDTypeClusterBasicInformationEventStartUpID);
         if (isStartUpEvent) {
+            // REVIEWERS:  this seems like a good place to set up / cache
+            // the essential device attributes - is it?
```
bzbarsky-apple, line 1469: This isn't really a great place to do it; ideally we would make this work along with the "device cache primed" bits, but those need some fixing first (on my plate to do ASAP).

That said, we could do this in `_handleSubscriptionEstablished` and `setPersistedClusterData`, which are two places we know our data is "best known at this time".... But again, if the device changes these values and reports new ones, shouldn't we update our concept of them?

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -3247,6 +3297,22 @@ - (void)removeClientDataForKey:(NSString *)key endpointID:(NSNumber *)endpointID
     [self.temporaryMetaDataCache removeObjectForKey:[NSString stringWithFormat:@"%@:%@", key, endpointID]];
 }
 
+#pragma mark Log Help
+
+- (MTRDeviceEssentialAttributes *)_essentialAttributesForCurrentState {
+    MTRClusterPath * basicInfoClusterPath = [MTRClusterPath clusterPathWithEndpointID:@(kRootEndpointId) clusterID:@(MTRClusterIDTypeBasicInformationID)];
+    MTRDeviceClusterData * basicInfoClusterData = [self _clusterDataForPath:basicInfoClusterPath];
+    
+    NSNumber * vidObj = basicInfoClusterData.attributes[@(MTRAttributeIDTypeClusterBasicInformationAttributeVendorIDID)][MTRValueKey];
```
bzbarsky-apple, line 3306: This is not safe.  If the server happened to server non-spec-compliant TLV we could have some other type of object for the value, and now you get bad behavior.

I strongly suggest using `[[MTRAttributeReport alloc] initWithResponseValue:...]` which for attributes we know about (which these are) will verify that the types are correct and put the value (In this case `NSNumber *`) into the `.value` of the `MTRAttributeReport`.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -3493,6 +3502,49 @@ - (void)removeClientDataForKey:(NSString *)key endpointID:(NSNumber *)endpointID
...
+- (NSNumber *)_informationalNumberAtAttributePath:(MTRAttributePath *)attributePath
+{
+    auto * cachedData = [self _cachedAttributeValueForPath:attributePath];
+
+    auto * attrReport = [[MTRAttributeReport alloc] initWithResponseValue:@{
+        MTRAttributePathKey : attributePath,
+        MTRDataKey : cachedData,
+    }
+                                                                    error:nil];
+    // REVIEWERS:  is it worth logging the `error` above?
```
bzbarsky-apple, line 3516: Error here would be "we expected a number but what we got is not a number", right?

I think this method should return `nullable NSNumber *` to make that clear.

Past that, logging it might not be worth it.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -3493,6 +3502,49 @@ - (void)removeClientDataForKey:(NSString *)key endpointID:(NSNumber *)endpointID
...
+
+    return attrReport.value;
+}
+
+- (NSNumber *)_informationalVendorID
```
bzbarsky-apple, line 3521: nullable.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -3493,6 +3502,49 @@ - (void)removeClientDataForKey:(NSString *)key endpointID:(NSNumber *)endpointID
...
+{
+    auto * vendorIDPath = [MTRAttributePath attributePathWithEndpointID:@(kRootEndpointId)
+                                                              clusterID:@(MTRClusterIDTypeBasicInformationID)
+                                                            attributeID:@(MTRClusterBasicAttributeVendorIDID)];
+
+    return [self _informationalNumberAtAttributePath:vendorIDPath];
+}
+
+- (NSNumber *)_informationalProductID
```
bzbarsky-apple, line 3530: nullable.

bzbarsky-apple (approved):

PR #33573, opened 2024-05-23
Address follow up review comments for not setting up subscription for…

### LINKED ISSUE #33572: Address followup comments from https://github.com/project-chip/connectedhomeip/pull/33559
There are some followup comments from PR - https://github.com/project-chip/connectedhomeip/pull/33559 that need to be addressed

nivi-apple:
… XPC controllers

Fixes: #33572

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -141,25 +141,6 @@ private:
 } // anonymous namespace
 
 #pragma mark - MTRDevice
-typedef NS_ENUM(NSUInteger, MTRInternalDeviceState) {
-    // Unsubscribed means we do not have a subscription and are not trying to set one up.
-    MTRInternalDeviceStateUnsubscribed = 0,
-    // Subscribing means we are actively trying to establish our initial subscription (e.g. doing
-    // DNS-SD discovery, trying to establish CASE to the peer, getting priming reports, etc).
-    MTRInternalDeviceStateSubscribing = 1,
-    // InitialSubscriptionEstablished means we have at some point finished setting up a
-    // subscription.  That subscription may have dropped since then, but if so it's the ReadClient's
-    // responsibility to re-establish it.
-    MTRInternalDeviceStateInitialSubscriptionEstablished = 2,
-    // Resubscribing means we had established a subscription, but then
-    // detected a subscription drop due to not receiving a report on time. This
-    // covers all the actions that happen when re-subscribing (discovery, CASE,
-    // getting priming reports, etc).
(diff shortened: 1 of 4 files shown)


woody-apple (approved):

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -687,14 +668,18 @@ - (void)_setDSTOffsets:(NSArray<MTRTimeSynchronizationClusterDSTOffsetStruct *>
...
+    // We should not allow a subscription for device controllers over XPC.
     return ![_deviceController isKindOfClass:MTRDeviceControllerOverXPC.class];
 }
 
 - (void)setDelegate:(id<MTRDeviceDelegate>)delegate queue:(dispatch_queue_t)queue
 {
     MTR_LOG("%@ setDelegate %@", self, delegate);
 
-    // We should not set up a subscription for device controllers over XPC.
+    std::lock_guard lock(_lock);
```
bzbarsky-apple, line 681: Ideally we would not hold the lock across the call into testDelegate below...  It might be OK, though, since it's debug-only code.  Please add a comment down in that code, though, that if it ever stops being debug-only it needs to happen outside the lock.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -830,12 +813,8 @@ - (BOOL)_subscriptionAbleToReport
 #endif
 
     // Unfortunately, we currently have no subscriptions over our hacked-up XPC
-    // setup.  Try to detect that situation.
-    if ([_deviceController isKindOfClass:MTRDeviceControllerOverXPC.class]) {
-        return NO;
-    }
-
-    return YES;
+    // setup. Based on whether subscriptions are allowed or not, return YES or NO.
```
bzbarsky-apple, line 816: How about:
```suggestion
    // Subscriptions are not able to report if they are not allowed.
```

and then the XPC bits are just an impl detail of `_subscriptionsAllowed`.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -830,12 +813,8 @@ - (BOOL)_subscriptionAbleToReport
 #endif
 
     // Unfortunately, we currently have no subscriptions over our hacked-up XPC
```
bzbarsky-apple, line 815:
```suggestion
```

bzbarsky-apple (approved):

src/darwin/Framework/CHIP/MTRDevice.mm
nivi-apple: done.

PR #33574, opened 2024-05-23
[ICD] Reset mNeedIcdRegistration to false when starting the commision flow

### LINKED ISSUE #33517: [BUG] chip-tool commissioning ICD device then commissioning non-ICD device will fail
### Reproduction steps

1.run chip-tool under interactive mode
2.pair a ICD device -> OK
3.then pair a non-ICD device -> FAIL

### Bug prevalence

always

matter 1.3

### Platform

core

### Platform Version(s)

all platform that implement commissioner

### Anything else?

From my anaysis, the root cause is no chance to set mNeedIcdRegistration to false (controller/AutoCommissioner.cpp)
When commissioning a ICD device, because ICD cluster exists, mNeedIcdRegistration will be set to true. Following a non-ICD device, mNeedIcdRegistration is still true, then commissioner will go to ICD registration stage. But the regisitration will fail for status UNSUPPORTED_CLUSTER.

One changes to fix maybe:

```
--- a/src/controller/AutoCommissioner.cpp
+++ b/src/controller/AutoCommissioner.cpp
@@ -752,6 +752,7 @@ CHIP_ERROR AutoCommissioner::CommissioningStepFinished(CHIP_ERROR err, Commissio
 
             if (mParams.GetICDRegistrationStrategy() != ICDRegistrationStrategy::kIgnore)
             {
+                mNeedIcdRegistration = false;
                 if (mDeviceCommissioningInfo.icd.isLIT && mDeviceCommissioningInfo.icd.checkInProtocolSupport)
                 {
                     mNeedIcdRegistration = true;
```

yunhanw-google:
When commissioning a LIT device,  mNeedIcdRegistration will be set to true. Following a non-LIT device, mNeedIcdRegistration is still true, then commissioner will continue with LIT registration.
We should reset the mNeedIcdRegistration to false before check whether the device is LIT or not.

### DIFF
--- a/src/controller/AutoCommissioner.cpp
+++ b/src/controller/AutoCommissioner.cpp
@@ -125,6 +125,8 @@ CHIP_ERROR AutoCommissioner::SetCommissioningParameters(const CommissioningParam
 
     mParams = params;
 
+    mNeedIcdRegistration = false;
+
     if (haveMaybeDanglingBufferPointers)
     {
         mParams.ClearExternalBufferDependentValues();
@@ -760,7 +762,6 @@ CHIP_ERROR AutoCommissioner::CommissioningStepFinished(CHIP_ERROR err, Commissio
                 }
             }
 
-            mNeedIcdRegistration = false;
             if (mParams.GetICDRegistrationStrategy() != ICDRegistrationStrategy::kIgnore)
             {
                 if (mDeviceCommissioningInfo.icd.isLIT && mDeviceCommissioningInfo.icd.checkInProtocolSupport)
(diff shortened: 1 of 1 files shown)


src/controller/AutoCommissioner.cpp
```diff
@@ -760,6 +760,7 @@ CHIP_ERROR AutoCommissioner::CommissioningStepFinished(CHIP_ERROR err, Commissio
                 }
             }
 
+            mNeedIcdRegistration = false;
```
bzbarsky-apple, line 763: This is probably OK, but conceptually it would make more sense to either store this in mParams or reset it in `AutoCommissioner::SetCommissioningParameters` like we do other things we get from the device (e.g. the failsafe timer above).

bzbarsky-apple (approved):

src/controller/AutoCommissioner.cpp
yunhanw-google: yes, it make sense, move it inside AutoCommissioner::SetCommissioningParameters

yufengwangca (approved):

PR #33577, opened 2024-05-23
Remove use of IMEngine Singleton in EventManagement

yyzhong-g:
This PR is to decouple EventManagement with IMEngine singleton.
 - Inject EventManagement into ReportingEngine so it doesn't look for the global instance.
 - Creates EventScheduler interface and inject into EventManagement to schedule events.
 
Created the EventScheduler to avoid both classes depends on each others. Also both classes are injected during Init call and takes the global one if not provided.

This is beneficial and allow us to have in-process testing (e.g. one client and one server running in parallel as different IM engines).


src/app/BUILD.gn
```diff
@@ -340,7 +339,8 @@ static_library("app") {
     # "CommandResponseSender._h"
     # "CommandHandler._h"
     # "ReadHandler._h",
-    # "WriteHandler._h"
+    # "WriteHandler._h",
+    # "EventManagement._h"
```
andy31415, line 343: Ideally if cpp is in a target, I would expect the .h to be there as well.

Something seems off here ... do we need more splitting maybe? We should ideally not add more of these mis-matches.

src/app/EventManagement.h
```diff
@@ -189,6 +189,18 @@ struct LogStorageResources
         PriorityLevel::Invalid; // Log priority level associated with the resources provided in this structure.
 };
 
+/**
+ * @brief
+ *  A EventScheduler to schedule deliveries of events.
```
andy31415, line 194: Can we describe this somehow else to help read the code? It seems obvious that a class named `Event scheduler` is used to `schedule events`.

Maybe some background of intent would be useful.

andy31415 (changes requested):
I think this couples reporting and src/app through EventManagement.h. Could we also clean up dependencies and targets? Making `Eventmanagement.h` a free target without it being managed by gn layering seems a step backwards.

src/app/EventManagement.h
yyzhong-g: Separate EventScheduler to its own header file solves the problem

src/app/EventManagement.h
```diff
@@ -189,6 +189,18 @@ struct LogStorageResources
...
+ */
+
+class EventScheduler
+{
+public:
+    virtual ~EventScheduler()                                                                   = default;
+    CHIP_ERROR virtual ScheduleEventDelivery(ConcreteEventPath & aPath, uint32_t aBytesWritten) = 0;
```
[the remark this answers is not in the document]
yyzhong-g: Added.

src/app/BUILD.gn
yyzhong-g: This is not needed anymore with Event Scheduler in its own file.

PR #33581, opened 2024-05-23
Building some internal hooks for automation

woody-apple:
Sadly we need some strange SPI here to support automation, these shouldn't be public API, so marking them as such for now.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -3302,7 +3302,7 @@ static BOOL AttributeHasChangesOmittedQuality(MTRAttributePath * attributePath)
 
     MTR_LOG("%@ _deviceMayBeReachable called", self);
 
-    [self _triggerResubscribeWithReason:"SPI client indicated the device may now be online"
+    [self _triggerResubscribeWithReason:"SPI client indicated the device may now be reachable"
                     nodeLikelyReachable:YES];
 }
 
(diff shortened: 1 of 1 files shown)


src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -936,6 +936,14 @@ - (void)_changeInternalState:(MTRInternalDeviceState)state
...
     if (lastState != state) {
         MTR_LOG("%@ internal state change %lu => %lu", self, static_cast<unsigned long>(lastState), static_cast<unsigned long>(state));
+
+        /* BEGIN DRAGONS: This is a huge hack for a specific use case, do not rename, remove or modify behavior here */
+        if ([delegate respondsToSelector:@selector(_deviceInternalStateChanged:)]) {
+            dispatch_async(_delegateQueue, ^{
+                [(id) delegate _deviceInternalStateChanged:self];
+            });
+        }
+        /* END DRAGONS */
```
kiel-apple, line 946: eagerly await a future of nested dragons

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -3268,6 +3285,31 @@ - (void)removeClientDataForKey:(NSString *)key endpointID:(NSNumber *)endpointID
...
+    return HaveSubscriptionEstablishedRightNow(_internalDeviceState);
+}
+
+- (void)_deviceMayBeReachable
+{
+    assertChipStackLockedByCurrentThread();
+
+    MTR_LOG("%@ _deviceMayBeReachable called", self);
+
+    [self _triggerResubscribeWithReason:"SPI client indicated the device may now be online"
```
bzbarsky-apple, line 3305:
```suggestion
    [self _triggerResubscribeWithReason:"SPI client indicated the device may now be reachable"
```

bzbarsky-apple (approved):
Why is it always dragons?  Can't it be dangerous squirrels for once?

PR #33583, opened 2024-05-24
Add Compare IPAddress in OndiscoveredNode

### LINKED ISSUE #32576: Properly check that IP addresses are the same in AbstractDnssdDiscoveryController::OnNodeDiscovered
https://github.com/project-chip/connectedhomeip/pull/32459 Introduced issue where it was comparing address of ipaddress array, and not ipaddress themsevles.

This caused a compilation error. https://github.com/project-chip/connectedhomeip/pull/32575 was introduced just to fix the compilation error, but left comparing IP addresses properly to the original PR author.

We must NOT assume that the ordering is consistent between the two lists.

Also we should be able to add tests in src/controller/tests/TestCommissionableNodeController.cpp

joonhaengHeo:
Fix #32576 
I add to compare IPAddress in onDiscoveredNode.
And, I added test code in "TestCommissionableNodeController.cpp".


src/controller/AbstractDnssdDiscoveryController.cpp
```diff
@@ -25,6 +25,32 @@
 namespace chip {
 namespace Controller {
 
+bool compareIpAddresses(const size_t sourceNumIPs, const size_t destinationNumIPs, const Inet::IPAddress * source,
```
bzbarsky-apple, line 28: Why isn't this taking two `const Span<Inet::IPAddress> &`?

bzbarsky-apple, line 28: Why is this a global function not scoped to this file?

bzbarsky-apple, line 28: Should probably be called `AddressListsSameExceptOrder` or something like that.

joonhaengHeo: I didn't think I would ever use it anywhere else.
The purpose of this function is to compare the IPAddress List in this file, and the code length would be too long to implement in OnNodeDiscovered.

joonhaengHeo: Inet::IPAddress is not pointer in OnNodeDiscovered.
When implemented in Span format, a Span variable must be created in OnNodeDiscovered, so it was implemented like this.

joonhaengHeo: OK. I see. I changed function name.

src/controller/AbstractDnssdDiscoveryController.cpp
```diff
@@ -25,6 +25,33 @@
 namespace chip {
 namespace Controller {
 
+bool AddressListsSameExceptOrder(const size_t sourceNumIPs, const size_t destinationNumIPs,
+                                 const Inet::IPAddress source[Dnssd::CommissionNodeData::kMaxIPAddresses],
+                                 const Inet::IPAddress destination[Dnssd::CommissionNodeData::kMaxIPAddresses])
+{
+    size_t sameIpAddress                                                 = 0;
+    bool addressUsed[chip::Dnssd::CommonResolutionData::kMaxIPAddresses] = { false };
```
andy31415, line 33: `std::bitset` may be more compact maybe

src/controller/AbstractDnssdDiscoveryController.cpp
joonhaengHeo: Ah.. I will add "static" keyword in this function.

joonhaengHeo: I modified using "Span" class.

src/controller/AbstractDnssdDiscoveryController.cpp
```diff
@@ -25,6 +25,33 @@
 namespace chip {
 namespace Controller {
 
+bool AddressListsSameExceptOrder(const size_t sourceNumIPs, const size_t destinationNumIPs,
```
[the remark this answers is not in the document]
joonhaengHeo: I modified.

src/controller/AbstractDnssdDiscoveryController.cpp
joonhaengHeo: Modified.

src/controller/AbstractDnssdDiscoveryController.cpp
```diff
@@ -22,9 +22,34 @@
 #include <lib/core/CHIPEncoding.h>
 #include <lib/support/logging/CHIPLogging.h>
 
+#include <bitset>
+
 namespace chip {
 namespace Controller {
 
+static bool SameExceptOrder(const chip::Span<const Inet::IPAddress> & source, const chip::Span<const Inet::IPAddress> & destination)
```
andy31415, line 30: source/destination names seem off ... those are generally used for 'copy' or equivalent. 

Please use `a,b` or `v1,v2` or anything that makes these symmetrical enough to imply they are not of special usage.

src/controller/AbstractDnssdDiscoveryController.cpp
```diff
@@ -22,9 +22,34 @@
...
+{
+    std::bitset<chip::Dnssd::CommonResolutionData::kMaxIPAddresses> addressUsed;
+    if (source.size() != destination.size())
```
andy31415, line 33: please add a verify that `size() <= kMaxIpAddresses` since otherwise the indexing below may go out of bounds.

You could VerifyOrDie or return false if the sizes are out of bounds.

src/controller/AbstractDnssdDiscoveryController.cpp
joonhaengHeo: I see. Thanks.

src/controller/AbstractDnssdDiscoveryController.cpp
joonhaengHeo: I added. Thanks.

yunhanw-google (approved):

src/controller/AbstractDnssdDiscoveryController.cpp
```diff
@@ -22,9 +22,36 @@
...
+        return false;
+    }
+
+    for (size_t s = 0; s < v1.size(); s++)
+    {
+        for (size_t d = 0; d < v2.size(); d++)
+        {
+            if (!addressUsed[d] && v1[s] == v2[d])
+            {
+                // Change the user flag so that the compared target is no longer used
```
andy31415, line 46:
```suggestion
                // Change the used flag so that the compared target is no longer used
```

PR #33594, opened 2024-05-24
Remove racy Darwin test.

bzbarsky-apple:
We are going to change how the estimated latency is computed anyway, at which point this test will no longer be valid.


jtung-apple (approved):

bzbarsky-apple:

yufengwangca (approved):

PR #33605, opened 2024-05-27
dac revocation: Fallback method to parse VID/PID from crl signer

shubhamdp:
#### Test
Tested method using test certs in the specification

### DIFF
--- a/credentials/generate-revocation-set.py
+++ b/credentials/generate-revocation-set.py
@@ -64,6 +64,26 @@ def extract_single_integer_attribute(subject, oid):
     return None
 
 
+def parse_vid_pid_from_distinguished_name(distinguished_name):
+    # VID/PID encoded using Matter specific RDNs
+    vid = extract_single_integer_attribute(distinguished_name, OID_VENDOR_ID)
+    pid = extract_single_integer_attribute(distinguished_name, OID_PRODUCT_ID)
+
+    # Fallback method to get the VID/PID, encoded in CN as "Mvid:FFFF Mpid:1234"
+    if vid is None and pid is None:
+        cn = distinguished_name.get_attributes_for_oid(x509.ObjectIdentifier("2.5.4.3"))[0].value
+
+        vid_start = cn.find('Mvid:')
+        if vid_start != -1:
+            vid = int(cn[vid_start + 5:vid_start + 9], 16)
+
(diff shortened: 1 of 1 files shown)


credentials/generate-revocation-set.py
```diff
@@ -64,6 +64,31 @@ def extract_single_integer_attribute(subject, oid):
     return None
 
 
+def extract_single_attribute_from_cn(cn, marker):
```
tcarmelveilleux, line 67: - Suggest: `extract_fallback_tag_from_common_name`

credentials/generate-revocation-set.py
```diff
@@ -64,6 +64,31 @@ def extract_single_integer_attribute(subject, oid):
...
+    val_len = 4
+    start_idx = cn.find(marker)
+
+    if start_idx != -1:
+        val_start_idx = start_idx + len(marker)
+        return int(cn[val_start_idx:val_start_idx + val_len], 16)
```
tcarmelveilleux, line 73: Extract the substring before calling `int` and return `None` if the substring length `!= len('FFFF')`. Current code can fail on a string ending in `Mvid:` or if there are insufficient characters for the fallback encoding.

credentials/generate-revocation-set.py
```diff
@@ -64,6 +64,31 @@ def extract_single_integer_attribute(subject, oid):
...
+
+
+def parse_vid_pid_from_distinguished_name(distinguished_name):
+    # VID/PID encoded using Matter specific RDNs
+    vid = extract_single_integer_attribute(distinguished_name, OID_VENDOR_ID)
+    pid = extract_single_integer_attribute(distinguished_name, OID_PRODUCT_ID)
+
+    # Fallback method to get the VID/PID, encoded in CN as "Mvid:FFFF Mpid:1234"
+    if vid is None and pid is None:
+        cn = distinguished_name.get_attributes_for_oid(x509.ObjectIdentifier("2.5.4.3"))[0].value
```
tcarmelveilleux, line 85: Add `from cryptography.x509.oid import NameOID` at the top, and you can replace `"2.5.4.3"` with `NameOID.COMMON_NAME`

tcarmelveilleux (approved):

dhrishi (approved):

Issue #30359, opened 2023-11-09
[1.3] Color Control Cluster XML has to be updated as per spec


Ashwinigrl:
### Reproduction steps

- [ ]  **"EnhancedColorMode" attribute max range is wrong in XML. It should be 3 instead of 0xFF.**
 
![image](https://github.com/project-chip/connectedhomeip/assets/98016634/4713c188-9e54-4ff9-8136-2f7ceaae13f0)
![image](https://github.com/project-chip/connectedhomeip/assets/98016634/c6494e02-4d47-4493-9e45-cedeb3b906a0)

- [ ]  **Bit 0 Feature name has to be updated from “HueAndSaturation” to “Hue or Saturation” in XML.**
 
![image](https://github.com/project-chip/connectedhomeip/assets/98016634/c4a4ee2b-8b46-40c4-8d70-849e86bec2a8)
![image](https://github.com/project-chip/connectedhomeip/assets/98016634/53ca2b31-3586-4ea4-8c0f-927e78ceb18c)

- [ ]  **"CoupleColorTempToLevelMinMireds" attribute Max range value is wrong in XML.**
![image](https://github.com/project-chip/connectedhomeip/assets/98016634/049c6978-35e5-4e80-8dd8-498efdf8a4b3)
![image](https://github.com/project-chip/connectedhomeip/assets/98016634/b8dddc3e-25ac-4938-a5eb-a054dc268409)

- [ ]  **"EnhancedColorMode" and "DriftCompensation" Attributes eum values are missing in XML.**
![image](https://github.com/project-chip/connectedhomeip/assets/98016634/30df2770-472b-4edd-9e5f-dbe6420ed247)

- [ ]  **"Options" attribute bit value is missing in XML.**
![image](https://github.com/project-chip/connectedhomeip/assets/98016634/85268fb5-769e-4712-826b-7ff924f08331)

### Bug prevalence

Whenever I do

### GitHub hash of the SDK that was being used

f4735790763692462c0803be35c808d52741ffb4

### Platform

raspi

### Platform Version(s)

1.3

### Type

Test Improvement

### Anything else?

**Spec Reference:** https://github.com/CHIP-Specifications/connectedhomeip-spec/blob/master/src/app_clusters/ColorControl.adoc

**XML Reference:** https://github.com/project-chip/connectedhomeip/blob/master/src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml

bzbarsky-apple:
[...]

Issue #30303, opened 2023-11-08
[1.3] Door Lock Cluster XML has to be updated as per spec


KishokG:
### Reproduction steps

**XML has to be updated as per spec:**

- The RID feature is been implemented in SDK and sample app. So the RID feature has to be updated from False to True in Feature map(Row No: 50).
- DRLK.S.A0025(OperatingMode) - Default value has to be added as per spec(Row No: 139).
- UserTypeEnum has to be updated as per spec(Row No: 820).
- ActuatorEnabled, LEDSettings and SoundVolume Attribute data type has to be added in Auxiliary data types.
- **Below Non matter features data type has to be removed(From Row No: 598):**
     - KeypadOperationEventMask, 
     - RemoteOperationEventMask , 
     - ManualOperationEventMask, 
     - RFIDOperationEventMask,  
    - KeypadProgrammingEventMask, 
    - RemoteProgrammingEventMask, 
    - RFIDProgrammingEventMask
- **Below mentioned data types is not available in spec. But it is been mentioned in XML so it has to be removed in XML(From Row No: 774):** 
    - DoorLockDayOfWeek bitmap
    - DoorLockProgrammingEventCode
    - DoorLockOperationEventCode
    - DoorLockSetPinOrIdStatus
    - DoorLockUserType
    - DoorLockUserStatus
-  SendPINOverTheAir attribute has to be removed in XML(Non Matter).
- DlStatus enum has to be updated as per spec(Row No:746).

**XML Reference:**
https://github.com/project-chip/connectedhomeip/blob/master/src/app/zap-templates/zcl/data-model/chip/door-lock-cluster.xml

**Specification Reference:**
https://github.com/CHIP-Specifications/connectedhomeip-spec/blob/master/src/app_clusters/DoorLock.adoc

### Bug prevalence

Whenever I do this 

### GitHub hash of the SDK that was being used

https://github.com/project-chip/connectedhomeip/commit/eb2f8f6ca95c181577415bc4b3e7cc240d5b2edd

### Platform

raspi

### Platform Version(s)

RPI - 4, 8GB RAM

### Type

Test Improvement

bzbarsky-apple:
Please see https://github.com/CHIP-Specifications/connectedhomeip-spec/issues/2448 which needs to be addressed before we change anything in the XML here.

PR #33612, opened 2024-05-27
Colorcontrol xml file sync with specs

Thirsrin:
Description: 

Added types for ColorControlOptions,EnhancedColorMode,DriftCompensation attributes in colorcontrol cluster(colorcontrol.xml)

ColorCapabilities renamed to ColorCapabilitiesBitmap and used as type for the attribute ColorCapabilities
UpdateFlags changed to UpdateFlagsBitmap Type and used in ColorLoopSet command argument
ColorMode renamed to ColorModeEnum
HueDirection renamed to DirectionEnum and ShortestDistance/LongestDistance values renamed to Shortest/Longest
HueMoveMode renamed to MoveModeEnum
HueStepMode renamed to StepModeEnum
ColorLoopAction renamed to ColorLoopActionEnum
ColorLoopDirection renamed to ColorLoopDirectionEnum and values renamed to Decrement and Increment
removed SaturationMoveMode as it is regrouped in MoveModeEnum
removed SaturationStepMode as it is regrouped in StepModeEnum
after generated zap.


src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
```diff
@@ -82,6 +82,28 @@ limitations under the License.
...
+
+  <enum name="EnhancedColorMode" type="enum8">
+    <cluster code="0x0300"/>
+    <item name="CurrentHueAndCurrentSaturation" value="0x00"/>
+    <item name="CurrentXAndCurrentY" value="0x01"/>
+    <item name="ColorTemperature" value="0x02"/>
+    <item name="EnhancedCurrentHueAndCurrentSaturation" value="0x03"/>
+  </enum>
+
+  <enum name="DriftCompensation" type="enum8">
```
bzbarsky-apple, line 98: This does not match the spec.

src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
```diff
@@ -82,6 +82,28 @@ limitations under the License.
     <item name="ActivateFromEnhancedCurrentHue" value="0x02"/>
   </enum>
 
+  <bitmap name="ColorControlOptions" type="bitmap8">
+    <cluster code="0x0300"/>
+    <field name="ExecuteIfOff" mask="0x1"/>
+  </bitmap>
+
+  <enum name="EnhancedColorMode" type="enum8">
```
bzbarsky-apple, line 90: This does not exist in the spec.

src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
```diff
@@ -82,6 +82,28 @@ limitations under the License.
     <item name="ActivateFromEnhancedCurrentHue" value="0x02"/>
   </enum>
 
+  <bitmap name="ColorControlOptions" type="bitmap8">
```
bzbarsky-apple, line 85: This does not exist in the spec with this name.

bzbarsky-apple (changes requested):
Please don't make XML changes that don't match the spec...

src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
```diff
@@ -82,6 +82,28 @@ limitations under the License.
...
+    <item name="CurrentHueAndCurrentSaturation" value="0x00"/>
+    <item name="CurrentXAndCurrentY" value="0x01"/>
+    <item name="ColorTemperature" value="0x02"/>
+    <item name="EnhancedCurrentHueAndCurrentSaturation" value="0x03"/>
+  </enum>
+
+  <enum name="DriftCompensation" type="enum8">
+    <cluster code="0x0300"/>
+    <item name="None" value="0x00"/>
+    <item name="Other/Unknown" value="0x01"/>
```
jmartinez-silabs, line 101: OtherOrUnknown

src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
jmartinez-silabs: Use EnhancedColorModeEnum

src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
jmartinez-silabs: use OptionsBitmap

src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
Thirsrin: updated with DriftCompensationEnum

src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
Thirsrin: updated with EnhancedColorModeEnum

src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
Thirsrin: updated with OptionsBitmap

src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
Thirsrin: updated with OtherOrUnknown

src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
jmartinez-silabs, line 1: Your changes and additions look good, but while comparing this xml with the spec, I noticed a few names and type that aren't spec-compliant with the latest spec version. Could you address them in this PR while you are at it?

`ColorCapabilities` should be `ColorCapabilitiesBitmap` and used as type for the attribute ColorCapabilities

`UpdateFlagsBitmap` Type is missing and should be used in `ColorLoopSet` command argument
`ColorMode` renamed to `ColorModeEnum`
`HueDirection` renamed to `DirectionEnum` and `ShortestDistance`/`LongestDistance` renamed to `Shortest`/`Longest`
`HueMoveMode` renamed to `MoveModeEnum`
`HueStepMode` renamed to `StepModeEnum`
`ColorLoopAction` renamed to `ColorLoopActionEnum`
`ColorLoopDirection` renamed to `ColorLoopDirectionEnum` and values renamed to `Decrement` and `Increment`


remove `SaturationMoveMode` as it is regrouped in `MoveModeEnum`
remove `SaturationStepMode` as it is regrouped in `StepModeEnum`

all type changes affect command arguments too.

I believe that is all. 
Let me know if you can take care of it

Thirsrin: updated as per your suggestions @jmartinez-silabs

jmartinez-silabs (approved):

src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
jmartinez-silabs: Thanks. You could put the pr out of draft and the build issues seems due to the renames of attributes in a few platform examples

bzbarsky-apple (changes requested):
First, thank you very much for doing this!  I would expect that this aims to fix https://github.com/project-chip/connectedhomeip/issues/30359 and https://github.com/project-chip/connectedhomeip/issues/30303?

That said, I would really appreciate the door lock and color control bits being split up into separate PRs, because this needs very careful review and finding time to review the whole thing as a big chunk is pretty hard.

I would also appreciate understanding what set of things are being fixed here and what set of things are not being fixed here and why.  For example, for Color Control some things got aligned with spec but not everything did (e.g. ColorModeEnum values, ColorCapabilitiesBitmap values, I stopped looking at that point).  Why is that?  If we're going to break APIs by changing names (which to be clear is what this PR is doing) we should not be doing it a few names at a time.  Communicating an API break as "we had to do this to align with spec" is a lot saner for API consumers than "oh, we broke it again in this release", "oh, we broke it again", etc.  This is why the review takes a while, by the way: a reviewer literally has to go through the cluster identifier-by-identifier comparing it with the spec....

Given the API changes here, for each one we need to decide:

1. Is this an API change we want to make, or does the spec need to change here instead?
2. If we make the API change, how do we minimize problems for API consumers?  For Darwin myself or @kiel-apple will handle this part.  For the C++ API you may want to add things to `src/app/common/CompatEnumNames.h` (which would incidentally fix some of the CI failures).  For Python and Java APIs, please work with the relevant platform maintainers as needed.

src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
Thirsrin: Thanks @jmartinez-silabs, I will move out from draft

src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
```diff
@@ -17,48 +17,35 @@ limitations under the License.
...
     <cluster code="0x0300"/>
     <item name="CurrentHueAndCurrentSaturation" value="0x00"/>
     <item name="CurrentXAndCurrentY" value="0x01"/>
     <item name="ColorTemperature" value="0x02"/>
   </enum>
 
-  <bitmap name="ColorCapabilities" type="bitmap16">
+  <bitmap name="ColorCapabilitiesBitmap" type="bitmap16">
     <cluster code="0x0300"/>
     <field name="HueSaturationSupported" mask="0x0001"/>
```
jmartinez-silabs, line 50: all *Supported have been removed from the names
HueSaturation
EnhancedHue
ColorLoop
XY
ColorTemperature

src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
```diff
@@ -67,25 +54,47 @@ limitations under the License.
...
+  <bitmap name="OptionsBitmap" type="bitmap8">
+    <cluster code="0x0300"/>
+    <field name="ExecuteIfOff" mask="0x1"/>
+  </bitmap>
+
+  <enum name="EnhancedColorModeEnum" type="enum8">
+    <cluster code="0x0300"/>
+    <item name="CurrentHueAndCurrentSaturation" value="0x00"/>
+    <item name="CurrentXAndCurrentY" value="0x01"/>
+    <item name="ColorTemperature" value="0x02"/>
```
jmartinez-silabs, line 81:
```suggestion
    <item name="ColorTemperatureMireds" value="0x02"/>
```

src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
```diff
@@ -17,48 +17,35 @@ limitations under the License.
...
-    <item name="Up" value="0x1"/>
-    <item name="Down" value="0x3"/>
-  </enum>
-
-  <enum name="ColorMode" type="enum8">
+  <enum name="ColorModeEnum" type="enum8">
     <cluster code="0x0300"/>
     <item name="CurrentHueAndCurrentSaturation" value="0x00"/>
     <item name="CurrentXAndCurrentY" value="0x01"/>
     <item name="ColorTemperature" value="0x02"/>
```
jmartinez-silabs, line 45:
```suggestion
    <item name="ColorTemperatureMireds" value="0x02"/>
```

src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
```diff
@@ -368,13 +377,13 @@ limitations under the License.
...
-    <attribute side="server" code="0x4001" define="COLOR_CONTROL_ENHANCED_COLOR_MODE" type="enum8" min="0x00" max="0xFF" writable="false" default="0x01">EnhancedColorMode</attribute>
+    <attribute side="server" code="0x4001" define="COLOR_CONTROL_ENHANCED_COLOR_MODE" type="EnhancedColorModeEnum" min="0x00" max="0xFF" writable="false" default="0x01">EnhancedColorMode</attribute>
     <attribute side="server" code="0x4002" define="COLOR_CONTROL_COLOR_LOOP_ACTIVE" type="int8u" min="0x00" max="0xFF" writable="false" default="0x00" optional="true">ColorLoopActive</attribute>
-    <attribute side="server" code="0x4003" define="COLOR_CONTROL_COLOR_LOOP_DIRECTION" type="int8u" min="0x00" max="0xFF" writable="false" default="0x00" optional="true">ColorLoopDirection</attribute>
+    <attribute side="server" code="0x4003" define="COLOR_CONTROL_COLOR_LOOP_DIRECTION" type="ColorLoopDirectionEnum" min="0x00" max="0xFF" writable="false" default="0x00" optional="true">ColorLoopDirection</attribute>
```
bzbarsky-apple, line 382: Just a note: this seems to not match the spec?  Spec says "uint8" with a "max 1" constraint....

If we think the spec should say `ColorLoopDirectionEnum`, then the spec needs to be updated accordingly, yes?  And the test plan (which just says to check that the value is a uint8, without even checking for the "max 1" constraint).

And if the spec is still in flux here, we should, imo, not touch this XML until the spec stabilizes.

@jmartinez-silabs @ReneJosefsen

ReneJosefsen: Good catch, I need to do some research on this to verify if this has always been the case or if this is a mistake on the spec side.

ReneJosefsen: Looks like the editorial https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/7808 changed this to max 1 🤔 

I can see the PR does state: "Adjusted constraints of ColorLoopDirection and ColorLoopActive" so it seems intentional.

ReneJosefsen: A bit more concerning that COLOR_CONTROL_COLOR_LOOP_ACTIVE and COLOR_CONTROL_COLOR_LOOP_DIRECTION have been marked as int8 but the spec has since 1.0 been uint8?

bzbarsky-apple: "int8u" is what the spec calls "uint8", so that part is all fine.

tcarmelveilleux (dismissed):
These renames add no value, but cause significant API breakage for all SDK users on common device types. Please justify the reason of all these renames.

src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
ReneJosefsen: Oh yeah, that was on me, read it as int8t. Oops. All good.

ReneJosefsen:
> These renames add no value, but cause significant API breakage for all SDK users on common device types. Please justify the reason of all these renames.

As was done with the other clusters, there has been a huge editorial change done to the cluster (https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/7808), due to the huge size of the ColorControl and DoorLock clusters the editorial landed very late due to lack of reviewers, which it totally understandable.

Given the spec side have landed late, it has not been feasible to address this on the SDK side. While it is annoying and causes pain, I would still hope we can manage to find a path for this, if not we properly need to revert a lot of this on the spec side if we want to keep the spec and SDK aligned 😔

jmartinez-silabs:
@tcarmelveilleux The spec was updated this way and uses enum type for all those. Shouldn't we sync the XML with the spec? 

It should be done at some point. What do you suggest?

src/app/zap-templates/zcl/data-model/chip/color-control-cluster.xml
jmartinez-silabs: @bzbarsky-apple I believe the Spec PR was finally merged. There is a new PR from Rene for some small bits that were still discussed. Since a Enum is used now. Do we still have to define a min and Max? Is using kUnknownEnumValue, in SDK,  to ensure proper limits enough?

Test plan should also be updated if we proceed with this pr.

ReneJosefsen: The spec does not use an enum for ColorLoopDirection right now. We can certainly change that, but currently the ColorLoopDirection attribute is still marked as uint8 in the attribute table.

Thirsrin: @ReneJosefsen  Thanks, ColorLoopDirection attribute changed to uint8.

bzbarsky-apple (changes requested):
Given that this continues to not match the spec...  I suggest that once you fix the XML to what you think is spec-compliant then you ask @hasty how to use Alchemy, and run it on top of this PR, then make sure there are no codegen changes with the Alchemy output.

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -223,7 +199,7 @@ class DefaultColorControlSceneHandler : public scenes::DefaultSceneHandlerImpl
 #endif
 
         // Initialize action attributes to default values in case they are not in the scene
-        uint8_t targetColorMode    = 0x00;
+        auto targetColorMode       = EnhancedColorMode::kCurrentHueAndCurrentSaturation;
```
bzbarsky-apple, line 202:
```suggestion
        auto targetColorMode       = EnhancedColorModeEnum::kCurrentHueAndCurrentSaturation;
```
since this is new, why not just use the right name?

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -290,7 +266,8 @@ class DefaultColorControlSceneHandler : public scenes::DefaultSceneHandlerImpl
                 if (decodePair.valueUnsigned8.Value() <=
                     static_cast<uint8_t>(ColorControl::EnhancedColorMode::kEnhancedCurrentHueAndCurrentSaturation))
                 {
-                    targetColorMode = decodePair.valueUnsigned8.Value();
+                    targetColorMode =
+                        static_cast<chip::app::Clusters::ColorControl::EnhancedColorModeEnum>(decodePair.valueUnsigned8.Value());
```
bzbarsky-apple, line 270:
```suggestion
                        static_cast<ColorControl::EnhancedColorModeEnum>(decodePair.valueUnsigned8.Value());
```
Please stop with all the extra prefixes.  I've said that a number of times on this PR already....

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -303,7 +280,7 @@ class DefaultColorControlSceneHandler : public scenes::DefaultSceneHandlerImpl
         // Switch to the mode saved in the scene
         if (SupportsColorMode(endpoint, targetColorMode))
         {
-            ColorControlServer::Instance().handleModeSwitch(endpoint, targetColorMode);
+            ColorControlServer::Instance().handleModeSwitch(endpoint, static_cast<uint8_t>(targetColorMode));
```
bzbarsky-apple, line 283:
```suggestion
            ColorControlServer::Instance().handleModeSwitch(endpoint, to_underlying(targetColorMode));
```
Again: please do not use static_cast to convert from enums to integers.  That loses size information that can catch bugs.

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -486,7 +463,7 @@ Status ColorControlServer::stopAllColorTransitions(EndpointId endpoint)
 }
 
 bool ColorControlServer::stopMoveStepCommand(app::CommandHandler * commandObj, const app::ConcreteCommandPath & commandPath,
-                                             uint8_t optionsMask, uint8_t optionsOverride)
+                                             chip::BitMask<OptionsBitmap> optionsMask, chip::BitMask<OptionsBitmap> optionsOverride)
```
bzbarsky-apple, line 466:
```suggestion
                                             BitMask<OptionsBitmap> optionsMask, BitMask<OptionsBitmap> optionsOverride)
```

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -525,7 +502,8 @@ bool ColorControlServer::stopMoveStepCommand(app::CommandHandler * commandObj, c
     return true;
 }
 
-bool ColorControlServer::shouldExecuteIfOff(EndpointId endpoint, uint8_t optionMask, uint8_t optionOverride)
+bool ColorControlServer::shouldExecuteIfOff(EndpointId endpoint, chip::BitMask<OptionsBitmap> optionMask,
+                                            chip::BitMask<OptionsBitmap> optionOverride)
```
bzbarsky-apple, line 506:
```suggestion
                                            BitMask<OptionsBitmap> optionOverride)
```

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -540,7 +518,7 @@ bool ColorControlServer::shouldExecuteIfOff(EndpointId endpoint, uint8_t optionM
         return true;
     }
 
-    uint8_t options = 0x00;
+    chip::BitMask<OptionsBitmap> options = 0x00;
```
bzbarsky-apple, line 521:
```suggestion
    BitMask<OptionsBitmap> options = 0x00;
```

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -567,23 +545,23 @@ bool ColorControlServer::shouldExecuteIfOff(EndpointId endpoint, uint8_t optionM
     // ---------- The following order is important in decision making -------
     // -----------more readable ----------
     //
-    if (optionMask == 0xFF && optionOverride == 0xFF)
+    if (optionMask == static_cast<OptionsBitmap>(0xFF) && optionOverride == static_cast<OptionsBitmap>(0xFF))
```
bzbarsky-apple, line 548: Followup (separate PR): we should properly express "not present" here (e.g. via Optionals).

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -599,26 +577,28 @@ bool ColorControlServer::shouldExecuteIfOff(EndpointId endpoint, uint8_t optionM
...
+    Attributes::EnhancedColorMode::Set(endpoint, newEnhanceColorModeEnum);
+    if (newEnhanceColorModeEnum == ColorControl::EnhancedColorMode::kEnhancedCurrentHueAndCurrentSaturation)
     {
         // Transpose COLOR_MODE_EHSV to ColorControl::EnhancedColorMode::kCurrentHueAndCurrentSaturation after setting
         // EnhancedColorMode
-        newColorMode = ColorControl::EnhancedColorMode::kCurrentHueAndCurrentSaturation;
+        newEnhanceColorModeEnum = ColorControl::EnhancedColorMode::kCurrentHueAndCurrentSaturation;
     }
-    Attributes::ColorMode::Set(endpoint, newColorMode);
+    Attributes::ColorMode::Set(endpoint, newColorModeEnum);
```
bzbarsky-apple, line 599: This makes no sense, sorry.  Isn't it potentially setting an invalid value?

I think my previous comment on this function was mis-implemented.  Maybe it's fine if the input is `EnhancedColorModeEnum`, if that's clearly documented and the callers handle it properly.  Certainly having it be uint8_t but then directly casting to `EnhancedColorModeEnum` is no better than passing in `EnhancedColorModeEnum` to start with.

But the mapping from that to `ColorModeEnum` in this function needs to be handled correctly.  And the callers need to map from non-enhanced to enhanced color mode correctly as needed.

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -599,26 +577,28 @@ bool ColorControlServer::shouldExecuteIfOff(EndpointId endpoint, uint8_t optionM
...
 
-    colorModeTransition = static_cast<uint8_t>((newColorMode << 4) + oldColorMode);
+    colorModeTransition = static_cast<uint8_t>((to_underlying(newColorModeEnum) << 4) + to_underlying(oldColorMode));
```
bzbarsky-apple, line 601: Yeah, this is clearly broken now too, with the above logic that never changes newColorModeEnum to be a legal non-enhanced value.  The fact that no tests fail on this is very bad.  @jmartinez-silabs

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -599,26 +577,28 @@ bool ColorControlServer::shouldExecuteIfOff(EndpointId endpoint, uint8_t optionM
...
 void ColorControlServer::handleModeSwitch(EndpointId endpoint, uint8_t newColorMode)
 {
-    uint8_t oldColorMode = 0;
+    auto oldColorMode = ColorMode::kCurrentHueAndCurrentSaturation;
     Attributes::ColorMode::Get(endpoint, &oldColorMode);
 
     uint8_t colorModeTransition;
 
-    if (oldColorMode == newColorMode)
+    ColorControl::ColorMode newColorModeEnum                = static_cast<ColorControl::ColorMode>(newColorMode);
```
bzbarsky-apple, line 585: Again, this cast is not valid if the input is actually an `EnhancedColorModeEnum` value.

bzbarsky-apple, line 585: Please use the non-legacy names of the enums in new code.

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -1389,8 +1369,8 @@ Status ColorControlServer::moveToHueAndSaturation(uint16_t hue, uint8_t saturati
  * @return false Failed
  */
 bool ColorControlServer::moveHueCommand(app::CommandHandler * commandObj, const app::ConcreteCommandPath & commandPath,
-                                        HueMoveMode moveMode, uint16_t rate, uint8_t optionsMask, uint8_t optionsOverride,
-                                        bool isEnhanced)
+                                        HueMoveMode moveMode, uint16_t rate, chip::BitMask<OptionsBitmap> optionsMask,
+                                        chip::BitMask<OptionsBitmap> optionsOverride, bool isEnhanced)
```
bzbarsky-apple, line 1373:
```suggestion
                                        BitMask<OptionsBitmap> optionsOverride, bool isEnhanced)
```

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -1496,8 +1476,9 @@ bool ColorControlServer::moveHueCommand(app::CommandHandler * commandObj, const
  * @return false Failed
  */
 bool ColorControlServer::moveToHueCommand(app::CommandHandler * commandObj, const app::ConcreteCommandPath & commandPath,
-                                          uint16_t hue, HueDirection moveDirection, uint16_t transitionTime, uint8_t optionsMask,
-                                          uint8_t optionsOverride, bool isEnhanced)
+                                          uint16_t hue, HueDirection moveDirection, uint16_t transitionTime,
+                                          chip::BitMask<OptionsBitmap> optionsMask, chip::BitMask<OptionsBitmap> optionsOverride,
```
bzbarsky-apple, line 1480:
```suggestion
                                          BitMask<OptionsBitmap> optionsMask, BitMask<OptionsBitmap> optionsOverride,
```

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -1631,8 +1612,9 @@ bool ColorControlServer::moveToHueCommand(app::CommandHandler * commandObj, cons
  */
 bool ColorControlServer::moveToHueAndSaturationCommand(app::CommandHandler * commandObj,
                                                        const app::ConcreteCommandPath & commandPath, uint16_t hue,
-                                                       uint8_t saturation, uint16_t transitionTime, uint8_t optionsMask,
-                                                       uint8_t optionsOverride, bool isEnhanced)
+                                                       uint8_t saturation, uint16_t transitionTime,
+                                                       chip::BitMask<OptionsBitmap> optionsMask,
+                                                       chip::BitMask<OptionsBitmap> optionsOverride, bool isEnhanced)
```
bzbarsky-apple, line 1617:
```suggestion
                                                       BitMask<OptionsBitmap> optionsMask,
                                                       BitMask<OptionsBitmap> optionsOverride, bool isEnhanced)
```
I thought you said you fixed all of these....

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -1671,8 +1653,9 @@ bool ColorControlServer::moveToHueAndSaturationCommand(app::CommandHandler * com
  * @return false Failed
  */
 bool ColorControlServer::stepHueCommand(app::CommandHandler * commandObj, const app::ConcreteCommandPath & commandPath,
-                                        HueStepMode stepMode, uint16_t stepSize, uint16_t transitionTime, uint8_t optionsMask,
-                                        uint8_t optionsOverride, bool isEnhanced)
+                                        HueStepMode stepMode, uint16_t stepSize, uint16_t transitionTime,
+                                        chip::BitMask<OptionsBitmap> optionsMask, chip::BitMask<OptionsBitmap> optionsOverride,
```
bzbarsky-apple, line 1657:
```suggestion
                                        BitMask<OptionsBitmap> optionsMask, BitMask<OptionsBitmap> optionsOverride,
```

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -1869,14 +1852,14 @@ bool ColorControlServer::stepSaturationCommand(app::CommandHandler * commandObj,
...
-    uint8_t optionsMask       = commandData.optionsMask;
-    uint8_t optionsOverride   = commandData.optionsOverride;
-    EndpointId endpoint       = commandPath.mEndpointId;
-    Status status             = Status::Success;
-    uint8_t currentSaturation = 0;
+    auto stepMode                                = commandData.stepMode;
+    uint8_t stepSize                             = commandData.stepSize;
+    uint8_t transitionTime                       = commandData.transitionTime;
+    chip::BitMask<OptionsBitmap> optionsMask     = commandData.optionsMask;
+    chip::BitMask<OptionsBitmap> optionsOverride = commandData.optionsOverride;
```
bzbarsky-apple, line 1859:
```suggestion
    BitMask<OptionsBitmap> optionsMask     = commandData.optionsMask;
    BitMask<OptionsBitmap> optionsOverride = commandData.optionsOverride;
```

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -1933,17 +1916,17 @@ bool ColorControlServer::colorLoopCommand(app::CommandHandler * commandObj, cons
...
-    Status status             = Status::Success;
-    uint8_t isColorLoopActive = 0;
-    uint8_t deactiveColorLoop = 0;
+    auto updateFlags                             = commandData.updateFlags;
+    auto action                                  = commandData.action;
+    auto direction                               = commandData.direction;
+    uint16_t time                                = commandData.time;
+    uint16_t startHue                            = commandData.startHue;
+    chip::BitMask<OptionsBitmap> optionsMask     = commandData.optionsMask;
+    chip::BitMask<OptionsBitmap> optionsOverride = commandData.optionsOverride;
```
bzbarsky-apple, line 1925:
```suggestion
    BitMask<OptionsBitmap> optionsMask     = commandData.optionsMask;
    BitMask<OptionsBitmap> optionsOverride = commandData.optionsOverride;
```

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -2268,12 +2251,12 @@ bool ColorControlServer::moveToColorCommand(app::CommandHandler * commandObj, co
...
 {
-    int16_t rateX           = commandData.rateX;
-    int16_t rateY           = commandData.rateY;
-    uint8_t optionsMask     = commandData.optionsMask;
-    uint8_t optionsOverride = commandData.optionsOverride;
-    EndpointId endpoint     = commandPath.mEndpointId;
-    Status status           = Status::Success;
+    int16_t rateX                                = commandData.rateX;
+    int16_t rateY                                = commandData.rateY;
+    chip::BitMask<OptionsBitmap> optionsMask     = commandData.optionsMask;
```
bzbarsky-apple, line 2256: I am stopping here.  Please actually address the issues that I keep pointing out, so I can stop pointing them out.

bzbarsky-apple (changes requested):

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -287,10 +263,9 @@ class DefaultColorControlSceneHandler : public scenes::DefaultSceneHandlerImpl
                 break;
             case Attributes::EnhancedColorMode::Id:
                 VerifyOrReturnError(decodePair.valueUnsigned8.HasValue(), CHIP_ERROR_INVALID_ARGUMENT);
-                if (decodePair.valueUnsigned8.Value() <=
-                    static_cast<uint8_t>(ColorControl::EnhancedColorMode::kEnhancedCurrentHueAndCurrentSaturation))
+                if (decodePair.valueUnsigned8.Value() <= to_underlying(EnhancedColorMode::kEnhancedCurrentHueAndCurrentSaturation))
                 {
-                    targetColorMode = decodePair.valueUnsigned8.Value();
+                    targetColorMode = static_cast<EnhancedColorMode>(decodePair.valueUnsigned8.Value());
```
bzbarsky-apple, line 268:
```suggestion
                    targetColorMode = static_cast<EnhancedColorModeEnum>(decodePair.valueUnsigned8.Value());
```

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -597,28 +573,28 @@ bool ColorControlServer::shouldExecuteIfOff(EndpointId endpoint, uint8_t optionM
  * @param endpoint
  * @param newColorMode
  */
-void ColorControlServer::handleModeSwitch(EndpointId endpoint, uint8_t newColorMode)
+void ColorControlServer::handleModeSwitch(EndpointId endpoint, EnhancedColorMode newColorMode)
```
bzbarsky-apple, line 576:
```suggestion
void ColorControlServer::handleModeSwitch(EndpointId endpoint, EnhancedColorModeEnum newColorMode)
```

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -597,28 +573,28 @@ bool ColorControlServer::shouldExecuteIfOff(EndpointId endpoint, uint8_t optionM
...
 {
-    uint8_t oldColorMode = 0;
+    auto oldColorMode = ColorModeEnum::kCurrentHueAndCurrentSaturation;
     Attributes::ColorMode::Get(endpoint, &oldColorMode);
 
     uint8_t colorModeTransition;
 
-    if (oldColorMode == newColorMode)
+    if (oldColorMode == static_cast<ColorModeEnum>(newColorMode))
```
bzbarsky-apple, line 583: It's not safe to cast from EnhancedColorModeEnum to ColorModeEnum without checking range first.  The cast needs to be in the other direction (that is, cast oldColorMode to EnhancedColorModeEnum, no?

But also (followup for this, since it would be a logic change, not just naming change), having this check before we coerce the value to a value value for ColorMode doesn't really make sense either: this check will always test false when `newColorMode == EnhancedColorMode::kEnhancedCurrentHueAndCurrentSaturation`...

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -2629,11 +2607,10 @@ void ColorControlServer::startUpColorTempCommand(EndpointId endpoint)
                 if (status == Status::Success)
                 {
                     // Set ColorMode attributes to reflect ColorTemperature.
-                    uint8_t updateColorMode = ColorControl::EnhancedColorMode::kColorTemperature;
+                    auto updateColorMode = ColorModeEnum::kColorTemperatureMireds;
                     Attributes::ColorMode::Set(endpoint, updateColorMode);
 
-                    updateColorMode = ColorControl::EnhancedColorMode::kColorTemperature;
-                    Attributes::EnhancedColorMode::Set(endpoint, updateColorMode);
+                    Attributes::EnhancedColorMode::Set(endpoint, static_cast<EnhancedColorMode>(updateColorMode));
```
bzbarsky-apple, line 2613:
```suggestion
                    Attributes::EnhancedColorMode::Set(endpoint, static_cast<EnhancedColorModeEnum>(updateColorMode));
```

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -2961,10 +2938,10 @@ void ColorControlServer::levelControlColorTempChangeCommand(EndpointId endpoint)
         return;
     }
 
-    uint8_t colorMode = 0;
+    auto colorMode = ColorModeEnum::kCurrentHueAndCurrentSaturation;
     Attributes::ColorMode::Get(endpoint, &colorMode);
 
-    if (colorMode == ColorControl::EnhancedColorMode::kColorTemperature)
+    if (static_cast<EnhancedColorMode>(colorMode) == EnhancedColorMode::kColorTemperatureMireds)
```
bzbarsky-apple, line 2944:
```suggestion
    if (static_cast<EnhancedColorModeEnum>(colorMode) == EnhancedColorMode::kColorTemperatureMireds)
```

bzbarsky-apple (approved):

jmartinez-silabs:
The old enum names were mapped to the new enum names in `src/app/common/CompatEnumNames.h`.
This is currently the best option we have to maintain a backward compatibility option.

Another option could've been to also keep the old enum in the current XML so they get generated too. However, that would work once the new XML generated by the spec parser is used.

PR #33614, opened 2024-05-27
Add basic support for reporting application installation status

lazarkov:
Adding basic support for reporting application's installation status

#### Problem
In case app is installed, there's no mechanism to report the current installation status.

#### Change overview
- Added Method to App Installation Service to Retrieve App's installation Status
- Hooked the `GetInstallationStatusOfApp` with ContentAppFactory 
- Added Shell support so developers can test the method

#### Tests
1. Build the TV app using command:
`scripts/examples/gn_build_example.sh examples/tv-app/linux out/debug/`

2. Build the TV casting app using command:
`scripts/examples/gn_build_example.sh examples/tv-casting-app/linux out/debug/`

3. Start TV App
`out/debug/chip-tv-app`

4. Start TV Casting App
`out/debug/chip-tv-casting-app`

5. On TV App: Uninstall the app 
`app uninstall 65521 32769`

6. On TV Casting App: Send casting request to the tv-app using 
`cast request 0`

7. On TV App: Accepted the Cast request from the tv-casting app
`controller ux ok`

_Note that you'll see the error 13, which stands for `kAppInstallConsentPending`_

8. On TV App: Once TV app confirmed there is no Content App, try and set other app installation statuses for example:
`app setinstallerrorstatus 65521 32769 15`

9. Try to connect to the TV app again and after you enter `controller ux ok` you'll see error :
`[1716811631946] [62904:4025232] [SVR] ---- Commissioner Declaration Start ----`
`[1716811631946] [62904:4025232] [SVR] 	error code: 15`
`[1716811631946] [62904:4025232] [SVR] ---- Commissioner Declaration End ----`


src/controller/CommissionerDiscoveryController.h
```diff
@@ -235,6 +235,19 @@ class DLL_EXPORT AppInstallationService
      */
     virtual bool LookupTargetContentApp(uint16_t vendorId, uint16_t productId) = 0;
 
+    /**
+     * @brief
+     *   Called to check app's installation status.
+     *
+     * This will be called by the main chip thread so any blocking work should be moved to a separate thread.
```
chrisdecenzo, line 242: since this method is synchronous, is it reasonable to expect blocking work to be moved to a separate thread?

src/controller/CommissionerDiscoveryController.h
```diff
@@ -235,6 +235,19 @@ class DLL_EXPORT AppInstallationService
...
+     *
+     *  @param[in]    vendorId           The vendorId in the DNS-SD advertisement of the requesting commissionee.
+     *  @param[in]    productId          The productId in the DNS-SD advertisement of the requesting commissionee.
+     *
+     */
+    virtual chip::Protocols::UserDirectedCommissioning::CommissionerDeclaration::CdError
```
chrisdecenzo, line 248: It's not intuitive to have this return a CdError. So if we are re-using that type then we need to provide explanation for how to interpret the different values like NoError.

src/controller/CommissionerDiscoveryController.cpp
```diff
@@ -244,6 +239,18 @@ void CommissionerDiscoveryController::InternalOk()
             mUserPrompter->PromptForAppInstallOKPermission(client->GetVendorId(), client->GetProductId(), client->GetDeviceName());
         }
         ChipLogDetail(Controller, "------Via Shell Enter: app install <pid> <vid>");
+
+        CommissionerDeclaration::CdError appInstallStatus =
+            mAppInstallationService->GetInstallationStatusOfApp(client->GetVendorId(), client->GetProductId());
+
+        if (appInstallStatus != CommissionerDeclaration::CdError::kNoError)
```
chrisdecenzo, line 246: We need to return something to the casting client, even when there is no entry in the install status table. At the very least it should be the kAppInstallConsentPending

examples/tv-app/tv-common/src/AppTv.cpp
```diff
@@ -627,6 +633,31 @@ bool ContentAppFactoryImpl::UninstallContentApp(uint16_t vendorId, uint16_t prod
     return false;
 }
 
+void ContentAppFactoryImpl::SetAppInstallationStatus(uint16_t vendorId, uint16_t productId, CommissionerDeclaration::CdError status)
+{
+    std::map<uint16_t, Protocols::UserDirectedCommissioning::CommissionerDeclaration::CdError>::iterator it;
+    it = mAppInstallationStatus.find(vendorId);
+    if (it != mAppInstallationStatus.end())
+    {
+        mAppInstallationStatus[vendorId] = status;
```
chrisdecenzo, line 642: in theory, a single vid could have multiple apps and so it seems like we should index this on vendorId and productId

src/controller/CommissionerDiscoveryController.cpp
lazarkov: Removed the if statement.

examples/tv-app/tv-common/src/AppTv.cpp
lazarkov: Created a search index <pid>:<vid>

src/controller/CommissionerDiscoveryController.h
lazarkov: Updated.

examples/tv-app/tv-common/shell/AppTvShellCommands.cpp
```diff
@@ -205,6 +205,16 @@ static CHIP_ERROR PrintAllCommands()
     streamer_printf(sout,
                     "  add-admin-vendor <vid>         Add vendor ID to list which will receive admin privileges. Usage: app "
                     "add-admin-vendor 65521\r\n");
+    streamer_printf(sout,
+                    "  app install <vid> <pid>              Install app with given vendor ID  and product ID. Usage: app install "
+                    "65521 32768\r\n");
+    streamer_printf(sout,
+                    "  app uinstall <vid> <pid>              Uinstall app at given vendor ID  and product ID. Usage: app uninstall "
```
chrisdecenzo, line 212:
```suggestion
                    "  app uninstall <vid> <pid>              Uinstall app at given vendor ID  and product ID. Usage: app uninstall "
```

examples/tv-app/tv-common/src/AppTv.cpp
```diff
@@ -627,6 +633,40 @@ bool ContentAppFactoryImpl::UninstallContentApp(uint16_t vendorId, uint16_t prod
...
 }
 
+std::string createSearchIndex(uint16_t vendorId, uint16_t productId)
+{
+    // Format the IDs into a string
+    std::string formattedString = std::to_string(vendorId) + ":" + std::to_string(productId);
+    return formattedString;
+}
+
+void ContentAppFactoryImpl::SetAppInstallationStatus(uint16_t vendorId, uint16_t productId, CommissionerDeclaration::CdError status)
```
chrisdecenzo, line 643: should we be calling this from somewhere else in this file to indicate that that app is now installed? And when that happens, remove the entry from the mAppInstallationStatus map? Otherwise, that map just grows, which seems unnecessary.

lazarkov: Addressed offline.

src/controller/CommissionerDiscoveryController.h
```diff
@@ -235,6 +235,21 @@ class DLL_EXPORT AppInstallationService
      */
     virtual bool LookupTargetContentApp(uint16_t vendorId, uint16_t productId) = 0;
 
+    /**
+     * @brief
+     *   Called to check app's installation status.
+     *
+     * CdError status is designed for CommissionerDeclaration and should be set by using SetErrorCode() and
+     * sent back to the client as a CDC Message. It is expected that app installation can have following statuses:
```
andy31415, line 243: could we review the text of this comment and see if it can be improved to help reading/using the code?

At a first read: 
  - method is named `GetInstallationStatusOfApp` and description si `checks app installation status` ... is it a get or a check?
  - constants for CdError are listed here ... why and why not just leave it as CdError? Are there specific constants that are never returned? Maybe mentioning that `CdError` has a constant of `kNoError` is important, however listing the rest seems extra.
  
  Overall this seems to be a `gets the status previously set by SetErrorCode, typically during commissioning` based on the description. However then why did we not name it `GetErrorCode` ? Why is this `Application installation status`?

examples/tv-app/tv-common/src/AppTv.cpp
```diff
@@ -684,7 +724,7 @@ CHIP_ERROR AppTvInit()
     ContentAppPlatform::GetInstance().SetupAppPlatform();
     ContentAppPlatform::GetInstance().SetContentAppFactory(&gFactory);
     gFactory.InstallContentApp((uint16_t) 1, (uint16_t) 11);
-    gFactory.InstallContentApp((uint16_t) 65521, (uint16_t) 32768);
+    gFactory.InstallContentApp((uint16_t) 65521, (uint16_t) 32769);
```
andy31415, line 727: could we name these constants here? what is 1, 11, 65521, 32769 and such and why is 32769 better than 32768?

examples/tv-app/tv-common/src/AppTv.cpp
andy31415, line 643: naming wise, a parameter of type `CdError` should probably be named error or cdError

src/controller/CommissionerDiscoveryController.h
lazarkov: Yes, your comment make sense. I've updated the method name and comment of the method.

However as CdError contains more errors that are not related to the app installation feature, I've noted only the ones that are necessary for that.

examples/tv-app/tv-common/src/AppTv.cpp
lazarkov: created variables that describe what's going on.

examples/tv-app/tv-common/src/AppTv.cpp
lazarkov: Updated the method name and parameter name.

src/controller/CommissionerDiscoveryController.h
```diff
@@ -235,6 +235,21 @@ class DLL_EXPORT AppInstallationService
      */
     virtual bool LookupTargetContentApp(uint16_t vendorId, uint16_t productId) = 0;
 
+    /**
+     * @brief
+     *   Called to get app's installation error code.
+     *
+     * CdError status is designed for CommissionerDeclaration and should be set by using SetErrorCode() and
+     * sent back to the client as a CDC Message. It is expected that app during installation can return following errors:
+     * kNoError, kAppInstallConsentPending, kAppInstalling, kAppInstallFailed, kAppInstalledRetryNeeded
```
andy31415, line 244: If this is a subset of errors, the comment should explain why only this subset will be returned here (and that it is a subset).

If this is an exhaustive set, then we should NOT have this list as this would mean we have to maintain the list in sync. 

Overall listing anything except kNoError or special one-offs adds a maintenance burden to update the comment if the underlying code changes (and worse, you return a generic CdError yet your api claims to never return the entire set of errors and if clients rely on that, they may break on updates).

How about something like:

```
Returns a subset of CDError that are specific during application instalation lifecycle. In particular:
  - kNoError means no error
  - kAppInstallConsentPending is returned if no application is available
  - other kApp* constants for lifecycle managemnt (kAppInstalling, kAppInstallFailed, kAppInstallRetryNeeded)
  - other CdError constants are EXPLICITLY DISALLOWED
```

unsure if the above comment is true ... however please see how it can be phrased to make expectations clear. Is for example `kUnexpectedCommissionerPasscodeReady` (random pick) possible here? it is a valid CdError value. How about `kConfigurationFailed`?

Should we maybe consider having a separate enum that is more restricted and that can be directly converted to a CdError instead?

lazarkov:
I will follow up with new PR soon.

PR #33618, opened 2024-05-27
[Fabric-Bridge] Refactor device management from main

yufengwangca:
1. Refactor device management from main.
2. Remove unused functions.

### DIFF
--- a/examples/fabric-bridge-app/linux/DeviceManager.cpp
+++ b/examples/fabric-bridge-app/linux/DeviceManager.cpp
@@ -1,0 +1,221 @@
+/*
+ *
+ *    Copyright (c) 2024 Project CHIP Authors
+ *    All rights reserved.
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+#include "DeviceManager.h"
+
+#include <app-common/zap-generated/ids/Attributes.h>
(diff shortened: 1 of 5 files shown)


examples/fabric-bridge-app/linux/DeviceManager.cpp
```diff
@@ -0,0 +1,221 @@
...
+// REVISION DEFINITIONS:
+// =================================================================================
+
+#define ZCL_DESCRIPTOR_CLUSTER_REVISION (1u)
+#define ZCL_BRIDGED_DEVICE_BASIC_INFORMATION_CLUSTER_REVISION (2u)
+#define ZCL_BRIDGED_DEVICE_BASIC_INFORMATION_FEATURE_MAP (0u)
+
+// ---------------------------------------------------------------------------
+
+int AddDeviceEndpoint(Device * dev, EmberAfEndpointType * ep, const Span<const EmberAfDeviceType> & deviceTypeList,
```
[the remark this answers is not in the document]
yufengwangca: Acked, currently we don't have unit tests framework setup for example apps

examples/fabric-bridge-app/linux/DeviceManager.cpp
```diff
@@ -0,0 +1,130 @@
...
+            mDevices[index] = dev;
+            CHIP_ERROR err;
+            int retryCount = 0;
+            while (retryCount < kMaxRetries)
+            {
+                DeviceLayer::StackLock lock;
+                dev->SetEndpointId(mCurrentEndpointId);
+                dev->SetParentEndpointId(parentEndpointId);
+                err =
+                    emberAfSetDynamicEndpoint(index, mCurrentEndpointId, ep, dataVersionStorage, deviceTypeList, parentEndpointId);
```
jmartinez-silabs, line 76: wouldn't be better to look for a free DynamicEndpointId and set it to that endpoint rather This retry loop?

jmartinez-silabs (approved):

yunhanw-google (approved):

examples/fabric-bridge-app/linux/DeviceManager.cpp
yufengwangca: I agree, but we first need to implement a new ember function emberAfDynamicEndpointIdExists to check if a endpoint is free. At this moment, the only way to see if an endpoint is free is try to set it via  emberAfSetDynamicEndpoint and check the result.

PR #33620, opened 2024-05-27
Minimize API surface that is templated for `CommandHandler`

### LINKED ISSUE #31627: Investigate removing `template <typename CommandData>` for `TryAddResponseData`
It should be possible to use a base class instead of the template "duck-typing" we are using when calling `DataModel::Encode(*writer, TLV::ContextTag(CommandDataIB::Tag::kFields), aData)`.

This will help reduce code compiled code size as well as make code more readable in `CommandHandler`

#### Alternative Solution to still explore:
* `LambdaBridge` as [mentioned here](https://github.com/project-chip/connectedhomeip/pull/31631#discussion_r1463535688)

andy31415:
This is in preparation to simplify CommandHandler usage in the CommandHandlerInterface: even though command handler has 50+ methods and 20+ data members, the external interface seems to be limited to 6 calls: addstatus/response/responsedata and some getters (fabricindex and subjectedescriptor) and an early flush call:

```sh
❯ rg '\.mCommandHandler\.\w*' -oIN | sort | uniq -c | sort -n
      1 .mCommandHandler.GetSubjectDescriptor
      2 .mCommandHandler.FlushAcksRightAwayOnSlowCommand
     32 .mCommandHandler.GetAccessingFabricIndex
     38 .mCommandHandler.AddResponse
     97 .mCommandHandler.AddStatus
```

As a result, it seems highly desirable to have some `AddRespose` implementation that is non-templated and this also addresses a subset #31627 (and may save some flash as a result as the surface that is templated may become smaller even if that is not a primary goal of this PR).

### DIFF
--- a/src/app/CommandHandler.h
+++ b/src/app/CommandHandler.h
@@ -358,13 +358,13 @@ public:
     }
 
     /**
-     * API for adding a data response.  The template parameter T is generally
-     * expected to be a ClusterName::Commands::CommandName::Type struct, but any
-     * object that can be encoded using the DataModel::Encode machinery and
-     * exposes the right command id will work.
+     * API for adding a data response.  The encoded is generally expected to encode
+     * a ClusterName::Commands::CommandName::Type struct, but any
+     * object should work.
      *
      * @param [in] aRequestCommandPath the concrete path of the command we are
      *             responding to.
+     * @param [in] commandId the command whose content is being encoded.
      * @param [in] aData the data for the response.
      */
     CHIP_ERROR AddResponseDataViaEncoder(const ConcreteCommandPath & aRequestCommandPath, CommandId commandId,
@@ -399,6 +399,9 @@ public:
 
     /**
      * API for adding a response with a given encoder of TLV data.
+     *
+     * The encoder would generally encode a ClusterName::Commands::CommandName::Type with
+     * the corresponding `GetCommandId` call.
      */
     void AddResponseViaEncoder(const ConcreteCommandPath & aRequestCommandPath, CommandId commandId, EncoderToTLV & encoder)
     {
--- a/src/app/CommandHandler.h
+++ b/src/app/CommandHandler.h
@@ -57,6 +57,30 @@
 namespace chip {
 namespace app {
 
+/// Defines an abstract class of something that can be encoded
+/// into a TLV with a given data tag
+class EncoderToTLV
+{
+public:
+    virtual ~EncoderToTLV() = default;
+
+    virtual CHIP_ERROR Encode(TLV::TLVWriter &, TLV::Tag tag) = 0;
+};
+/// Defines an abstract class of something that can be encoded
+/// into a TLV with a given data tag
+
+template <typename T>
+class DataModelEncoderToTLV : public EncoderToTLV
+{
+public:
+    DataModelEncoderToTLV(const T & value) : mValue(value) {}
+
+    virtual CHIP_ERROR Encode(TLV::TLVWriter & writer, TLV::Tag tag) { return DataModel::Encode(writer, tag, mValue); }
+
+private:
+    const T & mValue;
+};
+
 class CommandHandler
 {
 public:
@@ -327,12 +351,29 @@ public:
      * @param [in] aData the data for the response.
      */
     template <typename CommandData>
-    CHIP_ERROR AddResponseData(const ConcreteCommandPath & aRequestCommandPath, const CommandData & aData)
+    inline CHIP_ERROR AddResponseData(const ConcreteCommandPath & aRequestCommandPath, const CommandData & aData)
+    {
+        DataModelEncoderToTLV<CommandData> encoder(aData);
+        return AddResponseDataViaEncoder(aRequestCommandPath, CommandData::GetCommandId(), encoder);
+    }
+
+    /**
+     * API for adding a data response.  The template parameter T is generally
+     * expected to be a ClusterName::Commands::CommandName::Type struct, but any
+     * object that can be encoded using the DataModel::Encode machinery and
+     * exposes the right command id will work.
+     *
+     * @param [in] aRequestCommandPath the concrete path of the command we are
+     *             responding to.
+     * @param [in] aData the data for the response.
+     */
+    CHIP_ERROR AddResponseDataViaEncoder(const ConcreteCommandPath & aRequestCommandPath, CommandId commandId,
+                                         EncoderToTLV & encoder)
     {
         // Return early when response should not be sent out.
         VerifyOrReturnValue(ResponsesAccepted(), CHIP_NO_ERROR);
-
-        return TryAddingResponse([&]() -> CHIP_ERROR { return TryAddResponseData(aRequestCommandPath, aData); });
+        return TryAddingResponse(
+            [&]() -> CHIP_ERROR { return TryAddResponseDataViaEncoder(aRequestCommandPath, commandId, encoder); });
     }
 
     /**
@@ -350,9 +391,18 @@ public:
      * @param [in] aData the data for the response.
      */
     template <typename CommandData>
-    void AddResponse(const ConcreteCommandPath & aRequestCommandPath, const CommandData & aData)
+    inline void AddResponse(const ConcreteCommandPath & aRequestCommandPath, const CommandData & aData)
     {
-        if (AddResponseData(aRequestCommandPath, aData) != CHIP_NO_ERROR)
+        DataModelEncoderToTLV<CommandData> encoder(aData);
+        return AddResponseViaEncoder(aRequestCommandPath, CommandData::GetCommandId(), encoder);
+    }
+
+    /**
+     * API for adding a response with a given encoder of TLV data.
+     */
+    void AddResponseViaEncoder(const ConcreteCommandPath & aRequestCommandPath, CommandId commandId, EncoderToTLV & encoder)
+    {
+        if (AddResponseDataViaEncoder(aRequestCommandPath, commandId, encoder) != CHIP_NO_ERROR)
         {
             AddStatus(aRequestCommandPath, Protocols::InteractionModel::Status::Failure);
         }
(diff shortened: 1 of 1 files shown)


jmartinez-silabs (approved):

src/app/CommandHandler.h
```diff
@@ -57,6 +57,30 @@
...
+/// Defines an abstract class of something that can be encoded
+/// into a TLV with a given data tag
+class EncoderToTLV
+{
+public:
+    virtual ~EncoderToTLV() = default;
+
+    virtual CHIP_ERROR Encode(TLV::TLVWriter &, TLV::Tag tag) = 0;
+};
+/// Defines an abstract class of something that can be encoded
```
cecille, line 69: Does this comment go with the DataModelEncoderToTlv class? I think this is meant to say concrete (or just not abstract)

src/app/CommandHandler.h
```diff
@@ -327,12 +351,29 @@ class CommandHandler
...
+     * API for adding a data response.  The encoded is generally expected to encode
+     * a ClusterName::Commands::CommandName::Type struct, but any
+     * object should work.
+     *
+     * @param [in] aRequestCommandPath the concrete path of the command we are
+     *             responding to.
+     * @param [in] commandId the command whose content is being encoded.
+     * @param [in] aData the data for the response.
+     */
+    CHIP_ERROR AddResponseDataViaEncoder(const ConcreteCommandPath & aRequestCommandPath, CommandId commandId,
```
cecille, line 370: This is a new public function - is the intent that devs start moving away from AddResponseData? ie, if you're a dev, which one of these should you choose?

cecille (approved):

tehampson (approved):

src/app/CommandHandler.h
andy31415: Ah, copy and paste got me (even though I was convinced I typed a proper comment). Updated.

src/app/CommandHandler.h
andy31415: Added comments. I think `AddResponse/AddResponseData` is what people will use because of the auto `GetCommandId()` calls.

I kept this public because I expect I will split this to become virtual while the templated bits will stay templated (and non-virtual).

src/app/CommandHandler.h
```diff
@@ -57,6 +57,32 @@
...
 namespace app {
 
+/// Defines an abstract class of something that can be encoded
+/// into a TLV with a given data tag
+class EncoderToTLV
+{
+public:
+    virtual ~EncoderToTLV() = default;
+
+    virtual CHIP_ERROR Encode(TLV::TLVWriter &, TLV::Tag tag) = 0;
```
tcarmelveilleux, line 67: Suggest naming the writer arg

src/app/CommandHandler.h
```diff
@@ -57,6 +57,32 @@
...
+
+/// An `EncoderToTLV` the uses `DataModel::Encode` to encode things.
+///
+/// Generally useful to encode things like <ClusterName>::Commands::<CommandName>::Type
+/// structures.
+template <typename T>
+class DataModelEncoderToTLV : public EncoderToTLV
+{
+public:
+    DataModelEncoderToTLV(const T & value) : mValue(value) {}
```
tcarmelveilleux, line 78: This has an ownership time consideration. Consider documenting.

src/app/CommandHandler.h
```diff
@@ -57,6 +57,32 @@
...
+
+    virtual CHIP_ERROR Encode(TLV::TLVWriter & writer, TLV::Tag tag) { return DataModel::Encode(writer, tag, mValue); }
```
tcarmelveilleux, line 80: Please remove virtual and specify `override

src/app/CommandHandler.h
```diff
@@ -350,9 +399,21 @@ class CommandHandler
      * @param [in] aData the data for the response.
      */
     template <typename CommandData>
-    void AddResponse(const ConcreteCommandPath & aRequestCommandPath, const CommandData & aData)
+    inline void AddResponse(const ConcreteCommandPath & aRequestCommandPath, const CommandData & aData)
```
tcarmelveilleux, line 402: Why inline?

src/app/CommandHandler.h
```diff
@@ -57,6 +57,32 @@
 namespace chip {
 namespace app {
 
+/// Defines an abstract class of something that can be encoded
+/// into a TLV with a given data tag
+class EncoderToTLV
```
bzbarsky-apple, line 62: Why is this CommandHandler-specific?  And if it is, why is it in the chip::app namespace?

src/app/CommandHandler.h
```diff
@@ -57,6 +57,32 @@
...
+{
+public:
+    virtual ~EncoderToTLV() = default;
+
+    virtual CHIP_ERROR Encode(TLV::TLVWriter &, TLV::Tag tag) = 0;
+};
+
+/// An `EncoderToTLV` the uses `DataModel::Encode` to encode things.
```
bzbarsky-apple, line 70: "the uses"?

src/app/CommandHandler.h
```diff
@@ -325,14 +351,37 @@ class CommandHandler
...
     template <typename CommandData>
-    CHIP_ERROR AddResponseData(const ConcreteCommandPath & aRequestCommandPath, const CommandData & aData)
+    inline CHIP_ERROR AddResponseData(const ConcreteCommandPath & aRequestCommandPath, const CommandData & aData)
+    {
+        DataModelEncoderToTLV<CommandData> encoder(aData);
+        return AddResponseDataViaEncoder(aRequestCommandPath, CommandData::GetCommandId(), encoder);
+    }
+
+    /**
+     * API for adding a data response.  The encoded is generally expected to encode
```
bzbarsky-apple, line 365: What is "the encoded"?

src/app/CommandHandler.h
```diff
@@ -325,14 +351,37 @@ class CommandHandler
...
+     * @param [in] aRequestCommandPath the concrete path of the command we are
+     *             responding to.
+     * @param [in] commandId the command whose content is being encoded.
+     * @param [in] encoder - an encoder that places the command data structure for `commandId`
+     *             into a TLV Writer.
+     *
+     * Most applications are likely to use `AddResponseData` as a more convenient
+     * one-call that auto-sets command ID and creates the underlying encoders.
+     */
+    CHIP_ERROR AddResponseDataViaEncoder(const ConcreteCommandPath & aRequestCommandPath, CommandId commandId,
```
bzbarsky-apple, line 378: Why does this need the "ViaEncoder" suffix?  Why not just call this `AddResponseData`?  It has different arguments from the template version, so confusion should not be an issue.

src/app/CommandHandler.h
```diff
@@ -350,9 +399,21 @@ class CommandHandler
...
+        return AddResponseViaEncoder(aRequestCommandPath, CommandData::GetCommandId(), encoder);
+    }
+
+    /**
+     * API for adding a response with a given encoder of TLV data.
+     *
+     * The encoder would generally encode a ClusterName::Commands::CommandName::Type with
+     * the corresponding `GetCommandId` call.
+     */
+    void AddResponseViaEncoder(const ConcreteCommandPath & aRequestCommandPath, CommandId commandId, EncoderToTLV & encoder)
```
bzbarsky-apple, line 414: Again, why the "ViaEncoder" suffix?

src/app/CommandHandler.h
```diff
@@ -640,26 +700,14 @@ class CommandHandler
      *             responding to.
      * @param [in] aData the data for the response.
      */
-    template <typename CommandData>
-    CHIP_ERROR TryAddResponseData(const ConcreteCommandPath & aRequestCommandPath, const CommandData & aData)
+    CHIP_ERROR TryAddResponseDataViaEncoder(const ConcreteCommandPath & aRequestCommandPath, CommandId commandId,
```
bzbarsky-apple, line 703: Again, why the ViaEncoder suffix?

src/app/CommandHandler.h
```diff
@@ -640,26 +700,14 @@ class CommandHandler
...
-        // from CommandData with as little code as possible.
-        //
-        // Previously, non-essential code was unnecessarily templated, leading to
-        // compilation and duplication N times. By isolating only the code segments
-        // that genuinely require templating, minimizes duplicate compiled code.
-        ConcreteCommandPath responseCommandPath = { aRequestCommandPath.mEndpointId, aRequestCommandPath.mClusterId,
-                                                    CommandData::GetCommandId() };
+        ConcreteCommandPath responseCommandPath = { aRequestCommandPath.mEndpointId, aRequestCommandPath.mClusterId, commandId };
         ReturnErrorOnFailure(TryAddResponseDataPreEncode(aRequestCommandPath, responseCommandPath));
         TLV::TLVWriter * writer = GetCommandDataIBTLVWriter();
```
bzbarsky-apple, line 708: Since this stuff is no longer templated:

1. We no longer need this fine-grained slicing into things that do or do not depend on the now-nonexistent template argument.
2. We do not need this thing in the header.  It shouldn't be.

src/app/CommandHandler.h
andy31415: I had it without `ViaEncoder` at first however then `<template typename T>void AddResponse(const ConcreteCommandPath&, T>` and `void AddResponse(const ConcreteCommandPath&, EncoderToTLV &encoder)` seemed ambigous.

In the mean time I had to add commandId, so I could rename them as well I guess. This was iterated a few times, can be more.

src/app/CommandHandler.h
andy31415: Was the minimal delta. The correct thing is probably for this to live close to DataModel::Encode since it ends up being like that. I will move it as a followup.

src/app/CommandHandler.h
andy31415: It did not originally. I will change them back.

PR #33621, opened 2024-05-27
TC-DA-1.2: Fix for numberic PIDs with fallback

### LINKED ISSUE #268: [TC-DA-1.2] Validation when using `parse_ids_from_common_name` failed when VID or PID are digits only.
### Describe the bug

In my test run, it uses `parse_ids_from_common_name`, and it failed at this line ([#59](https://github.com/project-chip/connectedhomeip/blob/master/src/python_testing/TC_DA_1_2.py#L59)).
From my test, I get
```
(Pdb) print(commonName)
Mvid:1049 Mpid:8005 001b000005b5000a
```
and this `not s.isupper())` will always cause it to return None.
```
(Pdb) print (s)
1049
(Pdb) print (not s.isupper())
True
(Pdb) print (not s.islower())
True
```

### Steps to reproduce the behavior

_No response_

### Expected behavior

_No response_

### Log files

_No response_

### PICS file

_No response_

### Screenshots

_No response_

### Environment

_No response_

### Additional Information

_No response_

cecille:
Apparently isupper is false if there are no letters at all. Adds a fix for TC-DA-1.2 and test vector that exposes this condition.

Also adds the pid as a variable to the test vectors.

Test: test_TC_DA_1_2.py

Fixes: https://github.com/project-chip/certification-tool/issues/268


src/python_testing/TC_DA_1_2.py
```diff
@@ -56,7 +56,7 @@ def parse_single_vidpid_from_common_name(commonName: str, tag_str: str) -> str:
         return None
 
     s = sp[1][:4]
-    if not s.isupper() or len(s) != 4:
+    if not (s.isnumeric() or s.isupper()) or len(s) != 4:
```
tcarmelveilleux, line 59: Overall, the entire if could be replaced with `if re.match("[0-9A-F]{4}", s) is None:`

tcarmelveilleux (approved):

src/python_testing/TC_DA_1_2.py
cecille: Changed, added new test vector and confirmed in test that this now fails on line 81 at this function call, rather than on line 90 in the int conversion.

yinyi-hu (approved):

PR #33622, opened 2024-05-27
Python testing: Conformance support for device type conditions

cecille:
This adds conformance functions for device type conformances.

The conformances in the device library are subtly different than the conformances in the cluster libraries because the "features" are just arbitrary strings that are not exposed anywhere on the device.

Future work on this will look at passing in those strings and evaluating this as a group, but for now we consider all device type feature conformances to be optional. 

Test: Please see attached tests.

### DIFF
--- a/src/python_testing/TestConformanceSupport.py
+++ b/src/python_testing/TestConformanceSupport.py
@@ -17,14 +17,21 @@
 
 import xml.etree.ElementTree as ElementTree
 
-from conformance_support import ConformanceDecision, ConformanceException, ConformanceParseParameters, parse_callable_from_xml
-from matter_testing_support import MatterBaseTest, async_test_body, default_matter_test_main
+from conformance_support import (ConformanceDecision, ConformanceException, ConformanceParseParameters, deprecated, disallowed,
+                                 mandatory, optional, parse_basic_callable_from_xml, parse_callable_from_xml,
+                                 parse_device_type_callable_from_xml, provisional, zigbee)
+from matter_testing_support import MatterBaseTest, default_matter_test_main
 from mobly import asserts
 
 
+def basic_test(xml: str, cls: callable) -> None:
+    et = ElementTree.fromstring(xml)
+    xml_callable = parse_basic_callable_from_xml(et)
+    asserts.assert_true(isinstance(xml_callable, cls), "Unexpected class parsed from basic conformance")
+
+
 class TestConformanceSupport(MatterBaseTest):
-    @async_test_body
-    async def setup_class(self):
+    def setup_class(self):
         super().setup_class()
         # a small feature map
         self.feature_names_to_bits = {'AB': 0x01, 'CD': 0x02}
@@ -46,8 +53,7 @@ class TestConformanceSupport(MatterBaseTest):
         self.params = ConformanceParseParameters(
             feature_map=self.feature_names_to_bits, attribute_map=self.attribute_names_to_values, command_map=self.command_names_to_values)
 
-    @async_test_body
-    async def test_conformance_mandatory(self):
+    def test_conformance_mandatory(self):
         xml = '<mandatoryConform />'
         et = ElementTree.fromstring(xml)
         xml_callable = parse_callable_from_xml(et, self.params)
@@ -55,8 +61,7 @@ class TestConformanceSupport(MatterBaseTest):
             asserts.assert_equal(xml_callable(f, [], []), ConformanceDecision.MANDATORY)
         asserts.assert_equal(str(xml_callable), 'M')
 
-    @async_test_body
-    async def test_conformance_optional(self):
+    def test_conformance_optional(self):
         xml = '<optionalConform />'
         et = ElementTree.fromstring(xml)
         xml_callable = parse_callable_from_xml(et, self.params)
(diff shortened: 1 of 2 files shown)


src/python_testing/TestConformanceSupport.py
```diff
@@ -671,6 +663,92 @@ def test_conformance_greater(self):
...
+        xml = '<feature name="CD" />'
+        et = ElementTree.fromstring(xml)
+        try:
+            parse_basic_callable_from_xml(et)
+            asserts.fail("Unexpected success parsing non-basic conformance")
+        except ConformanceException:
+            pass
+
+        # mandatory tag is basic, but this one is a wrapper, so we should get a TypeError
+        xml = ('<mandatoryConform>'
```
andy31415, line 684: nit: maybe use multiline strings here and then brackets are not needed?

src/python_testing/conformance_support.py
```diff
@@ -151,6 +162,19 @@ def __str__(self):
...
+    ''' This is different than element feature because device types use "features" that aren't reported anywhere'''
+
+    def __init__(self, feature: str):
+        self.feature = feature
+
+    def __call__(self, feature_map: uint = 0, attribute_list: list[uint] = [], all_command_list: list[uint] = []) -> ConformanceDecision:
+        return ConformanceDecision.OPTIONAL
+
+    def __str__(self):
+        return f'{self.feature}'
```
andy31415, line 175: self.feature is already a string, so 
```suggestion
        return self.feature
```
?
Or do we need a clone here? I guess same comment to existing code ... it seems it does the same.

src/python_testing/conformance_support.py
```diff
@@ -393,3 +395,59 @@ def parse_callable_from_xml(element: ElementTree.Element, params: ConformancePar
...
+            return parse_basic_callable_from_xml(element)
+        # For device types ONLY, there are conformances called "attributes" that are essentially just placeholders for conditions in the device library.
+        # For example, temperature controlled cabinet has conditions called "heating" and "cooling". The cluster conditions are dependent on them, but they're not
+        # actually exposed anywhere ON the device other than through the presence of the cluster. So for now, treat any attribute conditions that are cluster conditions
+        # as just optional, because it's optional to implement any device type feature.
+        # Device types also have some marked as "condition" that are similarly optional
+        except ConformanceException as e:
+            if element.tag == ATTRIBUTE_TAG or element.tag == CONDITION_TAG or element.tag == FEATURE_TAG:
+                return device_feature(element.attrib['name'])
+            raise e
```
andy31415, line 413:
```suggestion
            raise
```

src/python_testing/conformance_support.py
cecille: nah, it was just because the ones below used that style, so it ended up like this. Removed these in the next commit.

src/python_testing/conformance_support.py
cecille: fixed in anotehr commit

src/python_testing/conformance_support.py
```diff
@@ -393,3 +395,59 @@ def parse_callable_from_xml(element: ElementTree.Element, params: ConformancePar
...
+        except ConformanceException:
+            pass
+        if element.tag == FEATURE_TAG:
+            try:
+                return feature(params.feature_map[element.get('name')], element.get('name'))
+            except KeyError:
+                raise ConformanceException(f'Conformance specifies feature not in feature table: {element.get("name")}')
+        elif element.tag == ATTRIBUTE_TAG:
+            # Some command conformance tags are marked as attribute, so if this key isn't in attribute, try command
+            name = element.get('name')
```
andy31415, line 435: why is keyerror handled above for `feature` but not here?

andy31415 (approved):

src/python_testing/conformance_support.py
```diff
@@ -393,3 +395,59 @@ def parse_callable_from_xml(element: ElementTree.Element, params: ConformancePar
...
+
+    return parse_wrapper_callable_from_xml(element, ops)
+
+
+def parse_callable_from_xml(element: ElementTree.Element, params: ConformanceParseParameters) -> Callable:
+    if not list(element):
+        try:
+            return parse_basic_callable_from_xml(element)
+        except ConformanceException:
+            pass
```
tcarmelveilleux, line 427: Please comment on why it's OK to `pass` here without any logs or warning.

src/python_testing/conformance_support.py
```diff
@@ -393,3 +395,59 @@ def parse_callable_from_xml(element: ElementTree.Element, params: ConformancePar
...
+        # as just optional, because it's optional to implement any device type feature.
+        # Device types also have some marked as "condition" that are similarly optional
+        except ConformanceException:
+            if element.tag == ATTRIBUTE_TAG or element.tag == CONDITION_TAG or element.tag == FEATURE_TAG:
+                return device_feature(element.attrib['name'])
+            raise
+
+    ops = []
+    for sub in element:
+        ops.append(parse_device_type_callable_from_xml(sub))
```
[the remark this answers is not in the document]
tcarmelveilleux: Similar on line 450.

tcarmelveilleux (approved):

ReneJosefsen (approved):

src/python_testing/conformance_support.py
cecille: added comment in next commit.

src/python_testing/conformance_support.py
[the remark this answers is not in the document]
cecille: done.

fessehaeve (approved):

PR #33623, opened 2024-05-27
test

cecille:
Testing a theory - if the yaml step is timing out before the session expires, then this might happen


cecille:
test was a failure because the timeout is hardcoded for darwin. Fixed by Boris here: https://github.com/project-chip/connectedhomeip/pull/33633. He is looking at piping through the timeout to darwin as well.

PR #33625, opened 2024-05-27
[thread] fix the crash issue when the SRP server discovery failure

ZhangLe2016:
When the SRP server rejects the end device's mDNS request, the end device cannot discover the SRP server's address. In this case, the value of the parameter `aServerSockAddr`is `nullptr`, and printing it will cause a system crash

### DIFF
--- a/src/platform/OpenThread/GenericThreadStackManagerImpl_OpenThread.hpp
+++ b/src/platform/OpenThread/GenericThreadStackManagerImpl_OpenThread.hpp
@@ -1353,7 +1353,7 @@ template <class ImplClass>
 void GenericThreadStackManagerImpl_OpenThread<ImplClass>::OnSrpClientStateChange(const otSockAddr * aServerSockAddr,
                                                                                  void * aContext)
 {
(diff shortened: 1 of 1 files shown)


src/platform/OpenThread/GenericThreadStackManagerImpl_OpenThread.hpp
```diff
@@ -1353,7 +1353,7 @@ template <class ImplClass>
 void GenericThreadStackManagerImpl_OpenThread<ImplClass>::OnSrpClientStateChange(const otSockAddr * aServerSockAddr,
                                                                                  void * aContext)
 {
-    if (aServerSockAddr)
```
superwhd, line 1356: I think this already checks the nullity of the pointer.

ZhangLe2016: Thanks for your correction, A null pointer is a special pointer value (usually 0) that indicates the pointer isn't pointing to any valid object.

PR #33634, opened 2024-05-28
[Silabs ] Bugfix re-enable ICD in lock-app

lpbeliveau-silabs:
Re-enable ICD in lock-app openthread.gni

### DIFF
--- a/examples/lock-app/silabs/openthread.gni
+++ b/examples/lock-app/silabs/openthread.gni
@@ -28,7 +28,7 @@ openthread_external_platform =
 sl_enable_test_event_trigger = true
 
 # ICD Default configurations
-chip_enable_icd_server = false
+chip_enable_icd_server = true
 chip_subscription_timeout_resumption = false
 sl_use_subscription_syncing = true
 


mkardous-silabs (approved):

andy31415 (approved):

tehampson (approved):

examples/lock-app/silabs/openthread.gni
```diff
@@ -28,7 +28,7 @@ openthread_external_platform =
 sl_enable_test_event_trigger = true
 
 # ICD Default configurations
-chip_enable_icd_server = false
+chip_enable_icd_server = true
```
cecille, line 31: Q, but perhaps not one that's directed at you - what happens when you have the cluster defined in zap but this flag turned off? It seems like that would always lead to bad behaviour. So does having this this flag at all make sense?

cecille (approved):

examples/lock-app/silabs/openthread.gni
lpbeliveau-silabs: We are currently using `#if CHIP_CONFIG_ENABLE_ICD_SERVER` and `#if CHIP_CONFIG_ENABLE_ICD_CIP`, (which has `assert(chip_enable_icd_server)` in its BUILD.gn) in the `icd-management-server.cpp`. So including the cluster in Zap without enabling the ICD server creates an empty shell that is not spec compliant, but that will work for CI examples that build this way.

This was discussed here: https://github.com/project-chip/connectedhomeip/pull/32247#discussion_r1498505642

And an issue was opened here for The Lock-App, but I suppose we could generalize this issue for any ICD app:
https://github.com/project-chip/connectedhomeip/issues/32321

PR #33635, opened 2024-05-28
Do not generate exec link map files by default

arkq:
### Problem

Link map files can be really huge. The problem is with CI where the overall storage is capped to 14GB. Currently Tizen CI QEMU check is almost over the top and with changes in https://github.com/project-chip/connectedhomeip/pull/33358 CI can no longer hold all the files.

### Changes

- do not generate map files by default
- if required it is possible to generate it by adding `chip_generate_link_map_file=true` to `gn gen` args, or by adding `--enable-link-map-file` to `./scripts/build/build_examples.py`

### Testing

Tested with `tizen-arm-tests-no-ble-no-thread`, and after this change the build directory is 6.7GB, where it was 9.8GB before.


build/toolchain/gcc_toolchain.gni
```diff
@@ -48,11 +46,7 @@ template("gcc_toolchain") {
       cxx = invoker.cxx
     }
 
-    if (defined(invoker.link_generate_map_file)) {
```
mspang, line 51: It seems like the goal of the patch can be accomplished with just the one line change defaulting chip_generate_link_map_file to false.

What was the reason to change the way that argument works?

arkq: > It seems like the goal of the patch can be accomplished with just the one line change defaulting chip_generate_link_map_file to false.

The `chip_generate_link_map_file` was introduced in `d15daf8a02b97` to allow disabling map files for Python builds. Also, the `link_generate_map_file` is here to allow disabling map files for Android. So, there are two mechanisms which basically do the same but in a different way. This PR unifies this by removing one of these (the less generic one, which is `link_generate_map_file` defined by the toolchain). Also, it changes the default behavior since map files are rather big and if not required they should be disabled.

andy31415 (approved):

build/toolchain/gcc_toolchain.gni
andy31415: Assuming this explanation resolves the comment (unifying mechanisms and also fixing up builder logic results in a larger PR).

PR #33637, opened 2024-05-28
Update Dockerfiles to `ubuntu:24.04`

andy31415:
Switch images to ubuntu 24.04

### Changes

- update base image to `ubuntu 24.04`
- Add some extra `do not rebuild` flags to build-all to help me with my testing
- to make the image smaller, use `--no-install-recommends` for apt installs
- Fix various versions of installed packages
- prefer system packages for `gn` (i.e. `generate-ninja`), `node` and various python packages
- python 3.12 does not like system pip installs, so added `--break-system-packages` for now to allow such installations
- Update make run argument as `-j$(nproc)` instead of `-j8`
- Some formatting fixes
- Case-by-case updates for several images, generally installing package prerequisites
- ESP32 qemu checkout tag update (we were using a 3-year old version)
- Remove cirque hardcoding of a specific python version

### Testing

- Ran locally `DOCKER_BUILD_VERSION=test-ubuntu-24.04 ./build-all.sh`
- Observed most images are 300MB smaller, EXCEPT chip-build-minimal which is 70MB larger (however this should not be significant as the image itself is around 500 to start with)
- Validated that ALL images compiled successfully
- ran `bootstra.sh` successfully on the vscode image
- Tested in vscode the following image compilations (using build_examples.py build command):
   - linux-x64-chip-tool
   - linux-arm64-chip-tool-clang
   - ameba-amebad-light
   - android-x64-tv-server
   - bouffalolab-bl602-night-light-light
   - cc32xx-lock
   - ti-cc13x4_26x4-lighting
   - efr32-brd4161a-switch
   - imx-lighting-app
   - nxp-k32w1-contact-sensor
   - mw320-all-clusters-app
   - qpg-qpg6105-thermostat (needed `pip install -r scripts/setup/requirements.qpg.txt`)
   - stm32-stm32wb5mm-dk-light
   - tizen-arm-light
   - telink-tlsr9528a-pump
   - openiotsdk-lock (error on imgtool not found, however this seems post-link and expected)
   
Failed builds:
  - genio-lighting-app (FAILS today in ToT on vscode as well with seemingly the same error)
  - mbed-cy8cproto_062_4343w-light (mbed-tools not installed in vscode image, same error on ToT)
  - rw61x-laundry-washer (complains about missing "lwipopts.h" ... identical error in ToT)
  - nrf-nrf52840dk-window-covering 
      - required `pip3 install -r ./scripts/setup/requirements.nrfconnect.txt)`
      - Needs more fixes as /workspace/scripts/tools/nrfconnect/generate_nrfconnect_chip_factory_data.py has trouble importing qrcode (type 'typing.TypeVar' is not an acceptable base type)
  - nuttx-x64-light - says `/bin/sh: 1: patch: not found`, added patch to the chip-build image
      - still fails, likely compiler version errors (e.g. `error: conflicting declaration ‘typedef struct __mbstate_t mbstate_t’`)


integrations/docker/images/stage-3/chip-build-esp32-qemu/Dockerfile
```diff
@@ -13,9 +13,9 @@ RUN set -x \
 WORKDIR /opt/espressif/qemu
 # Setup QEMU emulator for ESP32 platform
 RUN set -x \
-    && git clone --depth 1 -b esp-develop-20210826 https://github.com/espressif/qemu.git ../qemu-src \
+    && git clone --depth 1 -b v9.0.0 https://github.com/espressif/qemu.git ../qemu-src \
```
kiel-apple, line 16: ❤️

integrations/docker/images/stage-3/chip-build-esp32-qemu/Dockerfile
```diff
@@ -13,9 +13,9 @@ RUN set -x \
...
     && ../qemu-src/configure --target-list=xtensa-softmmu --enable-debug --disable-strip --disable-user --disable-capstone --disable-vnc --disable-sdl --disable-gtk \
-    && make -j8 \
+    && make -j$(nproc) \
```
kiel-apple, line 18: ❤️

integrations/docker/build-all.sh
```diff
@@ -25,19 +25,33 @@ set -e
...
 SUB_PATH="*"
 ALL_ARGS=""
+SKIP_BUILT=0
 
 for i in "$@"; do
     case $i in
+    -h | --help)
+       echo "Supported arguments:"
+       echo "  -h/--help       This help text"
+       echo "  -p/--path       TODO: document this"
```
kiel-apple, line 35: i like docs :)

integrations/docker/images/base/chip-build-minimal/Dockerfile
```diff
@@ -1,27 +1,19 @@
 # This minimal build image is intentionally not based on chip-build
-FROM ubuntu:focal
+FROM ubuntu:24.04
```
kiel-apple, line 2: I openly wonder:  is there a reason to use CalVer (`24.04`) vs the nickname (`noble`)?

I prefer CalVer so +1 from me on the change, but when it was the nickname in the past I wonder if there was a reason why.  (It's certainly possible that there was no particular reason.)

integrations/docker/build-all.sh
andy31415: I wish I had an idea what -p/--path actually does, but I could not figure it out at a quick glance. I just wanted at least the thing that I added documented :)

integrations/docker/images/base/chip-build-minimal/Dockerfile
andy31415: Names seem to hide away how old of a release we are talking about. If I would see 20, I would know 4 years old, but focal does not seem as friendly. did this change on purpose.

integrations/docker/build-all.sh
andy31415: I will try to figure out what it does ... It was a miss from our side not documenting things to start.

integrations/docker/images/base/chip-build/Dockerfile
```diff
@@ -183,22 +139,23 @@ RUN set -x \
 RUN set -x \
     && apt-get update \
     # Install build and runtime requirements for IWYU
-    && DEBIAN_FRONTEND=noninteractive apt-get install -fy --fix-missing clang-12 libclang-12-dev \
```
arkq, line 186: I guess that the IWYU is new enough in 24.04, so it's not required to install it from source any more.

arkq (approved):

integrations/docker/images/base/chip-build/Dockerfile
andy31415: Updated. We do get version 21 instead of 22, however hopefully good enough.

jmartinez-silabs (approved):

tcarmelveilleux (approved):

PR #33638, opened 2024-05-28
pw_unit_test migration: apps-1 and stub-refactoring

Alami-Amine:
Continuation of migration from nl_test to pw_unit_test, effort first discussed in issue https://github.com/project-chip/connectedhomeip/pull/29342

### In this PR:

1. consolidation and refactoring of stub functions that are used by Unit Tests and are related to Interaction Model and ember.
   - the files with the stubs are `test-interaction-model-api` and `test-ember-api`
    - these stub functions were previously defined in the unit test scripts.
   - decoupling them is necessary to allow us to convert tests one by one (to NOT need to convert all tests at once) or we will get linking errors
   
   - The functions to decouple are in TestReadInteraction.cpp , TestWritenteraction.cpp , TestCommandInteraction.cpp, TestAclAttribute.cpp as well as in `TestPowerSourceCluster` 
   - Functions from TestCommandInteraction.cpp weren't moved yet as they use Asserts specific to the unit-test-framework. So i defined them with a weak link.

2. conversion of  a bunch of Tests in `src/app/tests` (Interaction Layer)


src/app/tests/AppTestContext.cpp
```diff
@@ -49,6 +49,12 @@ void AppContext::SetUpTestSuite()
 
 void AppContext::TearDownTestSuite()
 {
+    // Adding a DrainAndServiceIO call fixes the teardown for some tests (TestAclAttribute and TestReportScheduler); these were
+    // triggering failures in tests that follow them when running on nRF CI (Zephyr native_posix where all Unit Tests are compiled
+    // into a single file)
+    // TODO: understand this more and solve underlying issue
+    LoopbackMessagingContext::DrainAndServiceIO();
```
bzbarsky-apple, line 56: Is this required as part of this change?  Seems like an unrelated fix that could use very careful review not in the context of a mass-change....

src/app/tests/TestAclAttribute.cpp
```diff
@@ -46,18 +47,13 @@ namespace {
...
-chip::EndpointId kTestEndpointId      = 4;
-
 class TestAccessControlDelegate : public AccessControl::Delegate
 {
 public:
     CHIP_ERROR Check(const SubjectDescriptor & subjectDescriptor, const chip::Access::RequestPath & requestPath,
                      Privilege requestPrivilege) override
     {
-        if (requestPath.cluster == kTestDeniedClusterId2)
+        if (requestPath.cluster == chip::Test::kTestDeniedClusterId2)
```
bzbarsky-apple, line 56: Why the extra `chip::`?

src/app/tests/TestAclAttribute.cpp
```diff
@@ -166,12 +147,12 @@ void TestAclAttribute::TestACLDeniedAttribute(nlTestSuite * apSuite, void * apCo
                                    chip::app::ReadClient::InteractionType::Subscribe);
 
         chip::app::AttributePathParams attributePathParams[2];
-        attributePathParams[0].mEndpointId  = kTestEndpointId;
-        attributePathParams[0].mClusterId   = kTestDeniedClusterId1;
+        attributePathParams[0].mEndpointId  = chip::Test::kTestEndpointId;
+        attributePathParams[0].mClusterId   = chip::Test::kTestDeniedClusterId1;
```
bzbarsky-apple, line 151: Why alll the extra `chip::`?  This applies to a bunch of places here.

src/app/tests/TestReadInteraction.cpp
```diff
@@ -301,53 +303,54 @@ using ReadHandlerNode     = chip::app::reporting::ReportScheduler::ReadHandlerNo
...
-    }
-
-    return false;
-}
-
-bool IsDeviceTypeOnEndpoint(DeviceTypeId deviceType, EndpointId endpoint)
-{
-    return false;
-}
+// CHIP_ERROR ReadSingleClusterData(const Access::SubjectDescriptor & aSubjectDescriptor, bool aIsFabricFiltered,
```
bzbarsky-apple, line 306: I assume the commented-out bits will go away before this is un-drafted?

src/app/tests/ember-test-compatibility.cpp
```diff
@@ -0,0 +1,127 @@
...
+// was previously in TestWriteInteraction.cpp
+const EmberAfAttributeMetadata * GetAttributeMetadata(const ConcreteAttributePath & aConcreteClusterPath)
+{
+    // Note: This test does not make use of the real attribute metadata.
+    static EmberAfAttributeMetadata stub = { .defaultValue = EmberAfDefaultOrMinMaxAttributeValue(uint32_t(0)) };
+    return &stub;
+}
+
+// was previously in TestWriteInteraction.cpp
+CHIP_ERROR WriteSingleClusterData(const Access::SubjectDescriptor & aSubjectDescriptor, const ConcreteDataAttributePath & aPath,
```
bzbarsky-apple, line 27: This has nothing to do with "ember".  Same for a lot of the other functions being added here.

src/app/tests/AppTestContext.cpp
Alami-Amine: well, I only need it in order to be able to run `TestAclAttribute.cpp` and `TestTestReportScheduler.cpp` before the other tests. 
I have just put it in a seperate PR: https://github.com/project-chip/connectedhomeip/pull/33654

src/app/tests/TestAclAttribute.cpp
Alami-Amine: ok will review them all

src/app/tests/ember-test-compatibility.cpp
Alami-Amine: if I understood correctly, the original form of this function (the one defined in `ember-compatibility-functions.cpp`) interacts directly with ember layer. and this is a stub definition to override it, no?
 what do you suggest I call it?

src/app/tests/TestAclAttribute.cpp
Alami-Amine: because the pigweed namespace is also called `Test`... putting only Test chooses `pw::unit_test::internal::Test`

src/app/tests/TestAclAttribute.cpp
Alami-Amine: because the pigweed namespace is also called `Test`... putting only Test chooses `pw::unit_test::internal::Test`

src/app/tests/ember-test-compatibility.cpp
bzbarsky-apple: I suggest you call this `test-interaction-model-api.cpp` or something like that.  `WriteSingleClusterData` is a backend-agnostic API that the IM uses.  The implementation in ember-compatibility implements `WriteSingleClusterData` on top of ember bits, but your implementation has nothing to do with ember....

Alami-Amine: FYI, it was suggested by @bzbarsky-apple to create another file for the ember specific bit: since I just added a new stub `emberAfGetClusterServerEndpointIndex()` which is in fact more ember related

src/app/tests/TestAttributeValueEncoder.cpp
```diff
@@ -18,10 +18,15 @@
 
 /**
  *    @file
- *      This file implements unit tests for CommandPathParams
+ *      This file implements unit tests for AttributeValueEncoder
```
andy31415, line 21: nit: instead of fixing I would consider dropping this entire comment block: it often has negative benefit:

- it is obvious `a file named testAttributeValueEncoder obviously tessts the AttributeValueEncoder`
- tests are copied and pasted for license and header include boilerplate ... and then this comment becomes copy and pasted and sometimes not updated, making us spend time to update it.

src/app/tests/TestBindingTable.cpp
```diff
@@ -15,122 +15,122 @@
...
 
-void TestEmptyBindingTable(nlTestSuite * aSuite, void * aContext)
+void VerifyTableSame(BindingTable & table, const std::vector<EmberBindingTableEntry> & expected)
+{
+    EXPECT_EQ(table.Size(), expected.size());
+    auto iter1 = table.begin();
+    auto iter2 = expected.begin();
+    while (iter2 != expected.end())
+    {
+        EXPECT_NE(iter1, table.end());
```
andy31415, line 37: if we use EXPECT then processing continues and the line below will deference invalid data probably.

Should we maybe do ASSERT_EQ on line 32 above and then not need this?

src/app/tests/TestReadInteraction.cpp
```diff
@@ -54,14 +56,14 @@ uint8_t gDebugEventBuffer[128];
...
-chip::EventId kTestEventIdCritical      = chip::Test::MockEventId(2);
-uint8_t kTestFieldValue1                = 1;
+chip::ClusterId kTestClusterId        = 6;
+chip::ClusterId kTestEventClusterId   = chip::Test::MockClusterId(1);
+chip::ClusterId kInvalidTestClusterId = 7;
+chip::EndpointId kTestEndpointId      = 1;
+chip::EndpointId kTestEventEndpointId = chip::Test::kMockEndpoint1;
+chip::EventId kTestEventIdDebug       = chip::Test::MockEventId(1);
+chip::EventId kTestEventIdCritical    = chip::Test::MockEventId(2);
+// uint8_t kTestFieldValue1                = 1;
```
andy31415, line 66: shoudl we delete things instead of commenting them out? Where was this used before?

src/app/tests/TestTimeSyncDataProvider.cpp
```diff
@@ -201,98 +207,65 @@ void TestDSTOffset(nlTestSuite * inSuite, void * inContext)
...
-    VerifyOrReturnError(CHIP_NO_ERROR == chip::Platform::MemoryInit(), FAILURE);
-    return SUCCESS;
-}
-
-int TestTearDown(void * inContext)
-{
-    chip::Platform::MemoryShutdown();
-    return SUCCESS;
+    EXPECT_EQ(CHIP_ERROR_PERSISTED_STORAGE_VALUE_NOT_FOUND, timeSyncDataProv.LoadDSTOffset(dstObj));
+    EXPECT_FALSE(dstObj.dstOffsetList.begin());
```
andy31415, line 267: this reads awakward. Does `EXPECT_TRUE(dstObj.dstOffsetList.empty())` work ? Or maybe:

`EXPECT_EQ(dstObj.dstOffsetList.begin(), dstObj.dstOffsetList.end())`

or is this really wanting to check for nullability? maybe then we should compare with nullptr, however a compare by converting an iterator to a bool feels off.

src/app/tests/test-ember-api.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+#include <app/tests/test-ember-api.h>
+#include <app/util/attribute-storage.h>
+#include <app/util/basic-types.h>
+#include <app/util/mock/Constants.h>
+#include <app/util/mock/Functions.h>
+
+chip::EndpointId chip::Test::numEndpoints = 0;
+
+// was previously in TestPowerSourceCluster.cpp
```
andy31415, line 25: I think the `previously` may lose context as time goes by. How about:
```suggestion
// Used by code in TestPowerSourceCluster.cpp (and generally things using mock ember functions may need this).
```

src/app/tests/test-ember-api.h
```diff
@@ -0,0 +1,29 @@
...
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+#include <app/util/basic-types.h>
+
+/// test-ember-api was created to consolidate and centralize stub functions that are related to ember and are run for unit-tests
+/// These were previously defined in the unit test-scripts themselves.
```
andy31415, line 19: I would rephrase to lose `previously` to make the comment not reference potentially very old code. We have context now but it will be lost in months/years.

src/app/tests/test-interaction-model-api.cpp
```diff
@@ -0,0 +1,144 @@
...
+#include <lib/core/CHIPCore.h>
+#include <messaging/ReliableMessageContext.h>
+
+namespace chip {
+uint8_t Test::attributeDataTLV[CHIP_CONFIG_DEFAULT_UDP_MTU_SIZE];
+size_t Test::attributeDataTLVLen = 0;
+
+namespace app {
+
+// was previously in TestWriteInteraction.cpp
```
andy31415, line 35: same here and maybe more instances. I will stop commenting on this particular thing.

src/app/tests/test-interaction-model-api.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+// was previously in TestReadInteraction.cpp
+bool IsClusterDataVersionEqual(const ConcreteClusterPath & aConcreteClusterPath, DataVersion aRequiredVersion)
+{
+    if (Test::kTestDataVersion1 == aRequiredVersion)
+    {
+        return true;
+    }
+
+    return false;
```
andy31415, line 99:
```suggestion
    return (Test::kTestDataVersion1 == aRequiredVersion);
```

andy31415 (approved):

src/app/tests/TestStatusResponseMessage.cpp
```diff
@@ -22,130 +22,92 @@
  *
  */
 
+#include <gtest/gtest.h>
```
feasel0, line 25:
```suggestion
#include <pw_unit_test/framework.h>
```

src/app/tests/TestStatusIB.cpp
```diff
@@ -15,193 +15,150 @@
  *    limitations under the License.
  */
 
+#include <gtest/gtest.h>
```
feasel0, line 18:
```suggestion
#include <pw_unit_test/framework.h>
```

src/app/tests/TestPendingResponseTrackerImpl.cpp
```diff
@@ -15,121 +15,99 @@
...
 
-    NL_TEST_ASSERT(inSuite, std::numeric_limits<uint16_t>::max() == pendingResponseTracker.Count());
+    EXPECT_EQ(std::numeric_limits<uint16_t>::max(), pendingResponseTracker.Count());
 
     for (uint16_t commandRef = 0; commandRef < std::numeric_limits<uint16_t>::max(); commandRef++)
     {
-        NL_TEST_ASSERT(inSuite, CHIP_NO_ERROR == pendingResponseTracker.Remove(commandRef));
-        NL_TEST_ASSERT(inSuite, false == pendingResponseTracker.IsTracked(commandRef));
+        EXPECT_EQ(CHIP_NO_ERROR, pendingResponseTracker.Remove(commandRef));
+        EXPECT_EQ(false, pendingResponseTracker.IsTracked(commandRef));
```
feasel0, line 44:
```suggestion
        EXPECT_FALSE(pendingResponseTracker.IsTracked(commandRef));
```

src/app/tests/TestBuilderParser.cpp
```diff
@@ -16,6 +16,8 @@
  *    limitations under the License.
  */
 
+#include <gtest/gtest.h>
```
feasel0, line 19:
```suggestion
#include <pw_unit_test/framework.h>
```

bzbarsky-apple (approved):

src/app/tests/TestMessageDef.cpp
```diff
@@ -22,6 +22,8 @@
  *
  */
 
+#include <gtest/gtest.h>
```
feasel0, line 25:
```suggestion
#include <pw_unit_test/framework.h>
```

feasel0 (approved):

src/app/tests/TestAttributeValueEncoder.cpp
Alami-Amine: totally Agree, this same exact comment is already in half the tests

src/app/tests/TestReadInteraction.cpp
Alami-Amine: i will delete it, this was here before, and now I moved it to `test-interaction-model-api` where it is used by the function `ReadSingleClusterData`

src/app/tests/TestStatusResponseMessage.cpp
Alami-Amine: thanks, I also added `#include <lib/core/StringBuilderAdapters.h>` as a bonus

src/app/tests/TestPendingResponseTrackerImpl.cpp
Alami-Amine: good catch

src/app/tests/TestBindingTable.cpp
Alami-Amine: yes sure, assuming that iter1 will not have an invalid index?

src/app/tests/TestTimeSyncDataProvider.cpp
Alami-Amine: yes emtpy works fine, done!

src/app/tests/test-ember-api.cpp
Alami-Amine: done

PR #33641, opened 2024-05-28
[Scenes] Fixing Scenes Ifdef

lpbeliveau-silabs:
Trying to build apps without the scenes cluster revealed that the cluster wasn't properly ifedefed.

Since the cluster is now provisional, we need this fix to allow building the onoff, colorcontrol and level control without scenes.

### DIFF
--- a/src/app/clusters/color-control-server/color-control-server.cpp
+++ b/src/app/clusters/color-control-server/color-control-server.cpp
@@ -431,15 +431,16 @@ ColorControlServer & ColorControlServer::Instance()
     return instance;
 }
 
+#ifdef MATTER_DM_PLUGIN_SCENES_MANAGEMENT
 chip::scenes::SceneHandler * ColorControlServer::GetSceneHandler()
 {
-
-#if defined(MATTER_DM_PLUGIN_SCENES_MANAGEMENT) && CHIP_CONFIG_SCENES_USE_DEFAULT_HANDLERS
+#if CHIP_CONFIG_SCENES_USE_DEFAULT_HANDLERS
     return &sColorControlSceneHandler;
 #else
     return nullptr;
-#endif // defined(MATTER_DM_PLUGIN_SCENES_MANAGEMENT) && CHIP_CONFIG_SCENES_USE_DEFAULT_HANDLERS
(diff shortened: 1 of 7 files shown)


src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -431,15 +431,16 @@ ColorControlServer & ColorControlServer::Instance()
     return instance;
 }
 
+#if defined(MATTER_DM_PLUGIN_SCENES_MANAGEMENT)
```
jmartinez-silabs, line 434: nit: just for consistency. you use #ifdef most of the time but for chip::scenes::SceneHandler * declarations you use #if defined()

jmartinez-silabs (approved):

src/app/clusters/color-control-server/color-control-server.h
```diff
@@ -20,14 +20,17 @@
...
 #include <app/ConcreteCommandPath.h>
-#include <app/clusters/scenes-server/SceneTable.h>
 #include <app/util/af-types.h>
 #include <app/util/attribute-storage.h>
 #include <app/util/basic-types.h>
 #include <app/util/config.h>
 #include <platform/CHIPDeviceConfig.h>
 #include <protocols/interaction_model/StatusCode.h>
 
+#ifdef MATTER_DM_PLUGIN_SCENES_MANAGEMENT
```
bzbarsky-apple, line 30: This is extremely fragile, since it relies on util/config.h being included before this part, but absolutely nothing checks that it is.

We need to figure out a better approach here.  Followup for that, please....  Possible options:

1. Use 1/0 vs defined/not and change the codegen accordingly, then use `#if`, which will fail if it's not defined at all.
2. static_assert something that is defined in config.h any time we rely on it being included.
3. Others?

bzbarsky-apple (approved):

src/app/clusters/color-control-server/color-control-server.h
lpbeliveau-silabs: My preference leans towards solution 1. since it doesn't cost much to have every clusters defined by default to 0 and it won't be too different from what we are currently doing.

Opened a followup here:
https://github.com/project-chip/connectedhomeip/issues/33659

PR #33642, opened 2024-05-28
Streamline bloat reports, part 1

kpschoedel:
- Remove ‘Increases’ and ‘Decreases’ tables from the GitHub comment reports, leaving only the full table and the large-increase highlights.

- Add region groups to platform configurations. Reports will switch to summary flash and RAM totals rather than reporting each section. Some platform report configurations did not already indicate which section(s) belong in which reportable region.

- Add region sizes to size report artifacts.

- Record area type (section, region, read/write) in the size database.

A followup will remove sizes other than regions from the GitHub comment reports.

### DIFF
--- a/scripts/tools/memory/gh_report.py
+++ b/scripts/tools/memory/gh_report.py
@@ -316,9 +316,6 @@ class V1Comment:
             threshold_df = df[df['% change'] > threshold]
             if threshold_df.empty:
                 threshold_df = None
-        decrease_df = df[df['change'] < 0]
-        if decrease_df.empty:
-            decrease_df = None
 
         with io.StringIO() as md:
             md.write(df.attrs['title'])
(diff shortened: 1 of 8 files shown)


andy31415 (approved):

Damian-Nordic:
@kpschoedel What's the reason for removing Increases and Decreases tables? I know an increase by < 1kB doesn't seem like a lot but it was still worth seeking for savings in the case of such changes. I am worried now more code size increases will slip through unnoticed.

kpschoedel:
There were requests to dramatically simplify the reports, on the grounds that there was too much information that it routinely ignored. It would be easy to restore the Increase table, I think, by undoing 322–388 in gh_report.py. (What I originally wanted was a central filterable database, but GitHub has no provision for workflows to write to one.)

PR #33646, opened 2024-05-29
Add ability to add pending fabric at a chosen fabric index

mspang:
Currently it's only possible to allocate fabric indexes sequentially starting from 1.

Sparse fabric tables could be produced by adding & removing fabrics, but not directly.

This allow the application to control the assignment of fabric ids directly by providing a function that overrides the next index to use.

eg If an application has 3 fabrics, it can recreate the fabric table from scratch while keeping consistent indexes.

### DIFF
--- a/src/credentials/FabricTable.h
+++ b/src/credentials/FabricTable.h
@@ -1003,6 +1003,13 @@ public:
      */
     CHIP_ERROR PeekFabricIndexForNextAddition(FabricIndex & outIndex);
 
+    /**
+     * Set the fabric index that will be used fo the next fabric added.
+     *
+     * Returns an error if the |fabricIndex| is already in use.
+     */
+    CHIP_ERROR SetFabricIndexForNextAddition(FabricIndex fabricIndex);
+
 private:
     enum class StateFlags : uint16_t
     {
--- a/src/credentials/tests/TestFabricTable.cpp
+++ b/src/credentials/tests/TestFabricTable.cpp
@@ -183,6 +183,18 @@ static CHIP_ERROR LoadTestFabric_Node02_01(FabricTable & fabricTable, bool doCom
     return err;
 }
 
+const FabricInfo * FindFabric(FabricTable & fabricTable, ByteSpan rootPublicKey, FabricId fabricId)
+{
+    Crypto::P256PublicKey key;
+    EXPECT_GE(key.Length(), rootPublicKey.size());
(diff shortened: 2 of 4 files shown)


src/credentials/FabricTable.h
```diff
@@ -1003,6 +1003,13 @@ class DLL_EXPORT FabricTable
      */
     CHIP_ERROR PeekFabricIndexForNextAddition(FabricIndex & outIndex);
 
+    /**
+     * Set the fabric index that will be used fo the next fabric added.
```
bzbarsky-apple, line 1007:
```suggestion
     * Set the fabric index that will be used for the next fabric added.
```

bzbarsky-apple, line 1007: This should document that use of this API might well be a spec violation (spec requires sequential assignment) and hence it must be used with extreme care.

mspang: If there's a spec violation but nobody can prove it, does it matter? Removing fabrics is permitted.

mspang: Added a clarifying comment that this may affect conformance.

bzbarsky-apple (approved):

tcarmelveilleux (approved):

src/credentials/tests/TestFabricTable.cpp
```diff
@@ -79,6 +79,11 @@ class ScopedFabricTable
     chip::Credentials::PersistentStorageOpCertStore mOpCertStore;
 };
 
+// These two fabric ids are supposed to be the same; only the trust root is
+// different.
```
andy31415, line 83: why two constants though? why not a shared fabric id name? Otherwise reading through code if I see `x = kNode1Constant, y = kNode2Constant` I naturally assume they are different because different names ... and then I could go back up to see "equal on purpose".

Why not name this `constexpr FabricId kSharedFabricId` or something like that?

mspang: These tests do not care that they are the same, it's not relevant. We just need a way to read the fabric id for a test NOC.

mspang: Basically we just need a constant for the fabric id for test NOC AA_BB, instead of directly pasting this number all over the tests.

And I think it's preferable for tests that aren't concerned with the root of trust to also not have to concern themselves with whether or not the root of the trust is "shared" between test NOCs, or not.

PR #33649, opened 2024-05-29
Make StringToUUID constexpr

jlatusek:
## Problem
* The current constexpr declaration of ChipBleUUID requires manual input of 16 UUID bytes, which is error-prone and hard to read
* The same UUIDs are defined in code for specific platforms.

## Changes
I've changed the implementation of StringToUUID to make it possible to use it during the constexpr initialization of ChipBleUUID.
For Tizen and Linux platforms, I've replaced the usage of locally defined UUIDs in favour of using global constexpr variables.


src/ble/BleLayer.cpp
```diff
@@ -140,18 +140,6 @@ static BleEndPointPool sBLEEndPointPool;
 
 // UUIDs used internally by BleLayer:
```
arkq, line 141: Comment no longer necessary

arkq (approved):

src/ble/BleLayer.cpp
jlatusek: Fixed

andy31415 (approved):

PR #33651, opened 2024-05-29
dac revocation: Default implementation to check if DAC chain is revoked

### LINKED ISSUE #33124: Implement default version of CheckForRevokedDACChain
### Reproduction steps

Implement default version of CheckForRevokedDACChain

### Bug prevalence

NA

11838c6616a08cc527771748a884bef809ff0920

### Platform

other

### Platform Version(s)

_No response_

### Type

Common Cluster Logic

### Anything else?

_No response_

### LINKED ISSUE #33652: dac revocation - Unit tests to for default implementation of CheckForRevokedDACChain()
Need some unit test along with #33651

shubhamdp:
Fixes #33124
Fixes #33652

#### Change overview
- Impl for parsing revocation set generated using `credentials/generate-revocation-set.py` using jsoncpp.
- Impl for parsing AKID, serial number, issuer name from cert in hex string, and base64 format
- Added an option to chip-tool to take json file as an input revoked set

#### Testing
- Verified commissioning with lighting-app with/without `--dac-revocation-set-path` option.
- When using `--dac-revocation-set-path`, verified with revoked as well as not revoked test dataset.
- Added unit tests


src/credentials/attestation_verifier/DefaultDeviceAttestationVerifier.cpp
```diff
@@ -26,10 +26,16 @@
...
 #include <lib/core/Global.h>
+#include <lib/support/Base64.h>
+#include <lib/support/BytesToHex.h>
 #include <lib/support/CodeUtils.h>
 #include <lib/support/ScopedBuffer.h>
 #include <lib/support/Span.h>
 
+#include <fstream>
+#include <json/json.h>
+#include <string.h>
```
tcarmelveilleux, line 37: Use `#include <string>` for STL.

src/credentials/attestation_verifier/DefaultDeviceAttestationVerifier.cpp
```diff
@@ -607,12 +613,147 @@ CHIP_ERROR DefaultDACVerifier::VerifyNodeOperationalCSRInformation(const ByteSpa
...
+
+    if (!parsingSuccessful)
+    {
+        return false;
+    }
+
+    for (const auto & revokedSet : jsonData)
+    {
+        if (strncmp(revokedSet["issuer_name"].asCString(), issuerNameBase64Str.data(), issuerNameBase64Str.size()) == 0 &&
+            strncmp(revokedSet["issuer_subject_key_id"].asCString(), akidHexStr.data(), akidHexStr.size()) == 0)
```
tcarmelveilleux, line 655: Don't use strncmp to do these checks. Construct std::string from .data()/.size()

src/credentials/attestation_verifier/DefaultDeviceAttestationVerifier.cpp
```diff
@@ -607,12 +613,147 @@ CHIP_ERROR DefaultDACVerifier::VerifyNodeOperationalCSRInformation(const ByteSpa
...
+    uint8_t serialNumberBuf[kMaxCertificateSerialNumberLength] = { 0 };
+    MutableByteSpan serialNumber(serialNumberBuf);
+
+    CHIP_ERROR err = ExtractSerialNumberFromX509Cert(certDer, serialNumber);
+    VerifyOrReturnError(err == CHIP_NO_ERROR, err);
+    VerifyOrReturnError(outSerialNumberHexStr.size() > serialNumber.size() * 2, CHIP_ERROR_BUFFER_TOO_SMALL);
+
+    Encoding::HexFlags flags = Encoding::HexFlags::kUppercaseAndNullTerminate;
+    err = BytesToHex(serialNumber.data(), serialNumber.size(), outSerialNumberHexStr.data(), outSerialNumberHexStr.size(), flags);
+    VerifyOrReturnError(err == CHIP_NO_ERROR, err);
```
tcarmelveilleux, line 697: Extract hex-ification to a helper, with ByteSpan input and returning an optional<std::string>

tcarmelveilleux (changes requested):
- [ ] Need to introduce an interface for revocation checking and injection of implementation.
- [ ] Need to avoid JSON and files being added at all to common code outside client-provided implementation.
- [ ] Need `CheckForRevokedDACChain` to be able to push exact errors
- [ ] Need unit tests for the additions

examples/chip-tool/commands/common/CredentialIssuerCommands.h
```diff
@@ -62,6 +62,15 @@ class CredentialIssuerCommands
     virtual CHIP_ERROR SetupDeviceAttestation(chip::Controller::SetupParams & setupParams,
                                               const chip::Credentials::AttestationTrustStore * trustStore) = 0;
 
+    /**
+     * @brief
+     *   This function is used to set the path to Device Attestation revocation set JSON file.
+     *
+     * @param[in] path Path to the JSON file containing list of revoked DACs or PAIs.
```
bzbarsky-apple, line 69: Apparently passing null is allowed?  This needs to be documented, along with what the semantics of passing null are.

examples/chip-tool/commands/common/CredentialIssuerCommands.h
```diff
@@ -62,6 +62,15 @@ class CredentialIssuerCommands
...
+     *                 It can be generated using credentials/generate-revocation-set.py script
+     */
+    virtual void SetupDeviceAttestationRevocationSetPath(const char * path) = 0;
```
bzbarsky-apple, line 72:
```suggestion
    virtual void SetDeviceAttestationRevocationSetPath(const char * path) = 0;
```

examples/chip-tool/commands/example/ExampleCredentialIssuerCommands.h
```diff
@@ -44,6 +44,17 @@ class ExampleCredentialIssuerCommands : public CredentialIssuerCommands
 
         return CHIP_NO_ERROR;
     }
+
+    void SetupDeviceAttestationRevocationSetPath(const char * path) override
+    {
+        if (path)
+        {
+            // As we know that we are using DefaultDACVerifier, we can downcast from
```
bzbarsky-apple, line 52: If we know that, why is `mDacVerifier` not just of type `DefaultDACVerifier`?  Either it can be (and then this cast is not needed), or it can't be, and then this cast is not valid.  I think it's the latter, since nothing says `GetDefaultDACVerifier` in fact returns an instance of `DefaultDACVerifier` right now, and its implementation could easily change to return something else in some cases.

Either way, there should be no cast here.

examples/chip-tool/commands/example/ExampleCredentialIssuerCommands.h
```diff
@@ -44,6 +44,17 @@ class ExampleCredentialIssuerCommands : public CredentialIssuerCommands
 
         return CHIP_NO_ERROR;
     }
+
+    void SetupDeviceAttestationRevocationSetPath(const char * path) override
+    {
+        if (path)
```
bzbarsky-apple, line 50: OK, so if path is null then we keep using whatever the previous path used to be?  Again, this all needs to be clearly documented in the API, and then we can have a discussion about what the right behavior is.

src/credentials/attestation_verifier/DefaultDeviceAttestationVerifier.cpp
```diff
@@ -607,12 +613,147 @@ CHIP_ERROR DefaultDACVerifier::VerifyNodeOperationalCSRInformation(const ByteSpa
...
+//    "type": "revocation_set",
+//    "issuer_subject_key_id": "63540E47F64B1C38D13884A462D16C195D8FFB3C",
+//    "issuer_name": "MD0xJTAjBgNVBAMMHE1hdHRlciBEZXYgUEFJIDB4RkZGMSBubyBQSUQxFDASBgorBgEEAYKifAIBDARGRkYx",
+//    "revoked_serial_numbers": [
+//      "69CDF10DE9E54ED1"
+//    ]
+//  }
+// ]
+//
+bool DefaultDACVerifier::IsEntryExistsInRevocationSet(const CharSpan & akidHexStr, const CharSpan & issuerNameBase64Str,
```
bzbarsky-apple, line 628:
```suggestion
bool DefaultDACVerifier::IsEntryInRevocationSet(const CharSpan & akidHexStr, const CharSpan & issuerNameBase64Str,
```

src/credentials/attestation_verifier/DefaultDeviceAttestationVerifier.cpp
```diff
@@ -607,12 +613,147 @@ CHIP_ERROR DefaultDACVerifier::VerifyNodeOperationalCSRInformation(const ByteSpa
...
+    std::string errs;
+
+    bool parsingSuccessful = Json::parseFromStream(readerBuilder, file, &jsonData, &errs);
+
+    // Close the file as it's no longer needed
+    file.close();
+
+    if (!parsingSuccessful)
+    {
+        return false;
```
bzbarsky-apple, line 649: Please log the errors; otherwise this is a completely inscrutable failure.

src/credentials/attestation_verifier/DefaultDeviceAttestationVerifier.cpp
bzbarsky-apple: 💯 

The checks you have here are:

> does `revokedSet["issuer_name"]` start with `issuerNameBase64Str`?

This is not the right check, and we should have tests that would have caught this bug, please.

src/credentials/attestation_verifier/DefaultDeviceAttestationVerifier.cpp
```diff
@@ -607,12 +613,147 @@ CHIP_ERROR DefaultDACVerifier::VerifyNodeOperationalCSRInformation(const ByteSpa
...
+    }
+
+    for (const auto & revokedSet : jsonData)
+    {
+        if (strncmp(revokedSet["issuer_name"].asCString(), issuerNameBase64Str.data(), issuerNameBase64Str.size()) == 0 &&
+            strncmp(revokedSet["issuer_subject_key_id"].asCString(), akidHexStr.data(), akidHexStr.size()) == 0)
+        {
+            for (const auto & revokedSerialNumber : revokedSet["revoked_serial_numbers"])
```
bzbarsky-apple, line 657: Consider reversing the sense of the issuer name/subject checks, doing `continue` if either of those does not match, and outdenting.

src/credentials/attestation_verifier/DefaultDeviceAttestationVerifier.cpp
```diff
@@ -607,12 +613,147 @@ CHIP_ERROR DefaultDACVerifier::VerifyNodeOperationalCSRInformation(const ByteSpa
...
+    return false;
+}
+
+CHIP_ERROR DefaultDACVerifier::GetAKIDHexStr(const ByteSpan & certDer, MutableCharSpan & outAKIDHexStr)
+{
+    uint8_t akidBuf[kAuthorityKeyIdentifierLength];
+    MutableByteSpan akid(akidBuf);
+
+    CHIP_ERROR err = ExtractAKIDFromX509Cert(certDer, akid);
+    VerifyOrReturnError(err == CHIP_NO_ERROR, err);
```
bzbarsky-apple, line 675:
```suggestion
    ReturnErrorOnFailure(err);
```

src/credentials/attestation_verifier/DefaultDeviceAttestationVerifier.cpp
```diff
@@ -607,12 +613,147 @@ CHIP_ERROR DefaultDACVerifier::VerifyNodeOperationalCSRInformation(const ByteSpa
...
+    VerifyOrReturnError(outAKIDHexStr.size() > akid.size() * 2, CHIP_ERROR_BUFFER_TOO_SMALL);
+
+    Encoding::HexFlags flags = Encoding::HexFlags::kUppercaseAndNullTerminate;
```
bzbarsky-apple, line 678: Why are we null-terminating?  We shouldn't need to null-terminate if we are writing into a span, which knows its own length.  And then we can make the length check `>=`, not `>`.

That said, doesn't `BytesToHex` do its own size check?  Why are we doing a separate one?

src/credentials/attestation_verifier/DefaultDeviceAttestationVerifier.cpp
```diff
@@ -607,12 +613,147 @@ CHIP_ERROR DefaultDACVerifier::VerifyNodeOperationalCSRInformation(const ByteSpa
...
+    err                      = BytesToHex(akid.data(), akid.size(), outAKIDHexStr.data(), outAKIDHexStr.size(), flags);
+    VerifyOrReturnError(err == CHIP_NO_ERROR, err);
```
bzbarsky-apple, line 680:
```suggestion
    ReturnErrorOnFailure(err);
```

src/credentials/attestation_verifier/DefaultDeviceAttestationVerifier.cpp
```diff
@@ -607,12 +613,147 @@ CHIP_ERROR DefaultDACVerifier::VerifyNodeOperationalCSRInformation(const ByteSpa
...
+
+    outAKIDHexStr.reduce_size(strlen(outAKIDHexStr.data()));
```
bzbarsky-apple, line 682: This makes no sense, but does explain why we were null-terminating...

The size must be `akid.size() * 2`, no?

src/credentials/attestation_verifier/DefaultDeviceAttestationVerifier.cpp
```diff
@@ -607,12 +613,147 @@ CHIP_ERROR DefaultDACVerifier::VerifyNodeOperationalCSRInformation(const ByteSpa
...
+    return CHIP_NO_ERROR;
+}
+
+CHIP_ERROR DefaultDACVerifier::GetSerialNumberHexStr(const ByteSpan & certDer, MutableCharSpan & outSerialNumberHexStr)
+{
+    uint8_t serialNumberBuf[kMaxCertificateSerialNumberLength] = { 0 };
+    MutableByteSpan serialNumber(serialNumberBuf);
+
+    CHIP_ERROR err = ExtractSerialNumberFromX509Cert(certDer, serialNumber);
+    VerifyOrReturnError(err == CHIP_NO_ERROR, err);
```
bzbarsky-apple, line 692: A bunch of the GetAKIDHexStr comments apply here.

src/credentials/attestation_verifier/DefaultDeviceAttestationVerifier.cpp
```diff
@@ -607,12 +613,147 @@ CHIP_ERROR DefaultDACVerifier::VerifyNodeOperationalCSRInformation(const ByteSpa
...
+    return CHIP_NO_ERROR;
+}
+
+CHIP_ERROR DefaultDACVerifier::GetIssuerNameBase64Str(const ByteSpan & certDer, MutableCharSpan & outIssuerNameBase64String)
+{
+    uint8_t issuerBuf[kMaxCertificateDistinguishedNameLength] = { 0 };
+    MutableByteSpan issuer(issuerBuf);
+
+    CHIP_ERROR err = ExtractIssuerFromX509Cert(certDer, issuer);
+    VerifyOrReturnError(CHIP_NO_ERROR == err, err);
```
bzbarsky-apple, line 709: See comments above.

src/credentials/attestation_verifier/DefaultDeviceAttestationVerifier.cpp
```diff
@@ -607,12 +613,147 @@ CHIP_ERROR DefaultDACVerifier::VerifyNodeOperationalCSRInformation(const ByteSpa
...
+    VerifyOrReturnError(outIssuerNameBase64String.size() > BASE64_ENCODED_LEN(issuer.size()), CHIP_ERROR_BUFFER_TOO_SMALL);
```
bzbarsky-apple, line 710: Why is this a `>`, not a `>=` check?  I would expect passing a buffer of size `BASE64_ENCODED_LEN(dataToEncode)` to work....

src/credentials/attestation_verifier/DefaultDeviceAttestationVerifier.cpp
```diff
@@ -607,12 +613,147 @@ CHIP_ERROR DefaultDACVerifier::VerifyNodeOperationalCSRInformation(const ByteSpa
...
+    char issuerNameBuffer[maxIssuerBase64Len]                                = { 0 };
+    char serialNumberHexStrBuffer[2 * kMaxCertificateSerialNumberLength + 1] = { 0 };
+    char akidHexStrBuffer[2 * kAuthorityKeyIdentifierLength + 1]             = { 0 };
+
+    MutableCharSpan issuerName(issuerNameBuffer);
+    MutableCharSpan serialNumber(serialNumberHexStrBuffer);
+    MutableCharSpan akid(akidHexStrBuffer);
+
+    CHIP_ERROR err = GetIssuerNameBase64Str(certDer, issuerName);
+    VerifyOrReturnValue(err == CHIP_NO_ERROR, false);
```
bzbarsky-apple, line 730: ReturnErrorOnFailure various places here.

src/credentials/attestation_verifier/DefaultDeviceAttestationVerifier.cpp
```diff
@@ -607,12 +613,147 @@ CHIP_ERROR DefaultDACVerifier::VerifyNodeOperationalCSRInformation(const ByteSpa
...
-    // TODO(#33124): Implement default version of CheckForRevokedDACChain
+    if (mDeviceAttestationRevocationSetPath != nullptr)
+    {
+        if (IsCertificateRevoked(info.dacDerBuffer))
+        {
+            attestationError = AttestationVerificationResult::kDacRevoked;
+        }
+        if (IsCertificateRevoked(info.paiDerBuffer))
+        {
+            attestationError = AttestationVerificationResult::kPaiRevoked;
```
bzbarsky-apple, line 754: So if both are revoked we consider the PAI being revoked more important than the DAC being revoked, in terms of the errors that get reported?

src/credentials/attestation_verifier/DefaultDeviceAttestationVerifier.cpp
```diff
@@ -607,12 +613,147 @@ CHIP_ERROR DefaultDACVerifier::VerifyNodeOperationalCSRInformation(const ByteSpa
     return CHIP_NO_ERROR;
 }
 
+// This method parses the below JSON Scheme
```
bzbarsky-apple, line 616: So right now, you can provide a trust store to a DefaultDACVerifier and it will work with that via API, agnostic to how the data is actually stored or gotten.

We should probably do something similar here; there are existing deployments of DefaultDACVerifier that cannot in fact create a file on disk to point it to, so that detail should be hidden behind an API boundary, with some TestSomething class that implements the relevant API and has the JSON thing as an implementation.

The API to that thing should ideally be async, just like the API to the DefaultDACVerifier itself.

bzbarsky-apple (changes requested):

src/credentials/attestation_verifier/DefaultDeviceAttestationVerifier.cpp
shubhamdp: Added another error code `kPaiAndDacRevoked` and using that one.

shubhamdp:
@tcarmelveilleux I have address the comments, PTAL for more details below

> * [x]  Need to introduce an interface for revocation checking and injection of implementation.

Added the `DeviceAttestationRevocationDelegate` and methods to inject the implementation into DeviceAttestationVerifier.
> * [x]  Need to avoid JSON and files being added at all to common code outside client-provided implementation.

Not using JSON impl in the DefaultDeviceAttestionVerifier and added TestDACRevocationDelegateImpl.
> * [x]  Need `CheckForRevokedDACChain` to be able to push exact errors

Added another error code if PAI and DAC both are revoked.
> * [x]  Need unit tests for the additions

Added unit tests

shubhamdp:
@bzbarsky-apple I've address the review-comments PTAL!

examples/chip-tool/commands/common/CredentialIssuerCommands.h
```diff
@@ -62,6 +62,17 @@ class CredentialIssuerCommands
     virtual CHIP_ERROR SetupDeviceAttestation(chip::Controller::SetupParams & setupParams,
                                               const chip::Credentials::AttestationTrustStore * trustStore) = 0;
 
+    /**
+     * @brief
+     *   This function is used to set the path to Device Attestation revocation set JSON file.
```
bzbarsky-apple, line 67:
```suggestion
     *   This function is used to set the path to the Device Attestation revocation set JSON file.
```

src/credentials/attestation_verifier/DeviceAttestationVerifier.h
```diff
@@ -79,6 +79,8 @@ enum class AttestationVerificationResult : uint16_t
 
     kInternalError = 900,
 
+    kPaiAndDacRevoked = 1000,
```
bzbarsky-apple, line 82: I'd expect this to be up near the other Pai/Dac bits, with a similar value.  Not sure whether to put it with the kDac or kPai constants; which of those is the more immediate problem if both are revoked?

src/credentials/attestation_verifier/DeviceAttestationVerifier.h
```diff
@@ -409,13 +413,39 @@ class DeviceAttestationVerifier
...
+
 protected:
     CHIP_ERROR ValidateAttestationSignature(const Crypto::P256PublicKey & pubkey, const ByteSpan & attestationElements,
                                             const ByteSpan & attestationChallenge, const Crypto::P256ECDSASignature & signature);
 
     // Default to support the "development" test key for legacy purposes (since the DefaultDACVerifier)
     // always supported development keys.
     bool mEnableCdTestKeySupport = true;
+
+    DeviceAttestationRevocationDelegate * mRevocationDelegate = nullptr;
```
bzbarsky-apple, line 426: Why is this on the interface here?  Shouldn't it be on whatever class uses DeviceAttestationRevocationDelegate to implement CheckForRevokedDACChain?

I guess the issue is that some consumer (e.g. chip-tool?) just does GetDefaultDACVerifier() and then wants to set this delegate on it. But this part doesn't make sense for other verifier impls....

What should probably happen is that `GetDefaultDACVerifier` grows either an optional argument or a two-argument overload or something.

shubhamdp: Even I didn't want it here, but wasn't sure where to put it. Thanks for the suggestion. Moved to `DefaultDACVerifier`. and adjusted the bits in chip-tool to set it up accordingly. PTAL

src/credentials/attestation_verifier/DeviceAttestationVerifier.h
shubhamdp: Moved closer to 'kPai` bits, as the name starts with 'kPai` and  revoked PAI is a bigger concern, revoked DAC.

examples/chip-tool/commands/common/CHIPCommand.cpp
```diff
@@ -450,7 +453,14 @@ CHIP_ERROR CHIPCommand::InitializeCommissioner(CommissionerIdentity & identity,
     std::unique_ptr<ChipDeviceCommissioner> commissioner = std::make_unique<ChipDeviceCommissioner>();
     chip::Controller::SetupParams commissionerParams;
 
-    ReturnLogErrorOnFailure(mCredIssuerCmds->SetupDeviceAttestation(commissionerParams, sTrustStore));
+    if (mDacRevocationSetPath.HasValue())
+    {
+        static chip::Credentials::TestDACRevocationDelegateImpl testDacRevocationDelegate;
```
bzbarsky-apple, line 458: So in interactive mode, if `mDacRevocationSetPath` has no value this will just keep using the revocation set from last time?  Trying to understand the intended behavior here.

shubhamdp: Yes that is the intended behaviour here, keeping it similar to trust store path. Also, factored out this code into similar method as trust store path.

should we honour the argument provided in interactive mode? If yes, I'll do it in follow up.

bzbarsky-apple: This is quite different from the trust store path behavior.  The behavior for trust store path is:

1. The first time a command is executed in interactive mode with a trust store path argument, create a FileAttestationTrustStore for that path.
2. Any time after that a command is executed with a trust store path argument, use that FileAttestationTrustStore (with the old path, ignoring the new value).
3. Any command executed without a trust store path argument uses `Credentials::GetTestAttestationTrustStore()`. 
4. Some complications around the env var bits, but let's ignore those.

The behavior that is being implemented here, on the other hand is:

1. The first time a command is executed in interactive mode with a trust store path argument, create a TestDACRevocationDelegateImpl for that path.
2. Any time after that a command is executed, use that same TestDACRevocationDelegateImpl no matter what the revocation path for the command was or whether one was present.

Neither behavior is terribly sane, honestly, but they are also different and I am not sure whether the behavior differences are intentional or not.

src/credentials/attestation_verifier/DeviceAttestationVerifier.h
```diff
@@ -47,6 +47,7 @@ enum class AttestationVerificationResult : uint16_t
     kPaiVendorIdMismatch  = 205,
     kPaiAuthorityNotFound = 206,
     kPaiMissing           = 207,
+    kPaiAndDacRevoked     = 208,
```
tcarmelveilleux, line 50: Have a different code for PAI revoked, and DAC revoked, in the correct ranges (200 for PAI, 300 for DAC)

src/credentials/attestation_verifier/DeviceAttestationVerifier.h
```diff
@@ -259,6 +260,8 @@ class ArrayAttestationTrustStore : public AttestationTrustStore
     const size_t mNumCerts;
 };
 
+class DeviceAttestationRevocationDelegate;
```
tcarmelveilleux, line 263: This would not be needed if DeviceAttestationRevocationDelegate was a subclass of DeviceAttestationVerifier

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
```diff
@@ -0,0 +1,204 @@
...
+    mDeviceAttestationRevocationSetPath = path;
+    return CHIP_NO_ERROR;
+}
+
+// This method parses the below JSON Scheme
+// [
+//  {
+//    "type": "revocation_set",
+//    "issuer_subject_key_id": "63540E47F64B1C38D13884A462D16C195D8FFB3C",
+//    "issuer_name": "MD0xJTAjBgNVBAMMHE1hdHRlciBEZXYgUEFJIDB4RkZGMSBubyBQSUQxFDASBgorBgEEAYKifAIBDARGRkYx",
```
tcarmelveilleux, line 55: What constitutes the issuer name? Which bytes?

src/credentials/attestation_verifier/DeviceAttestationVerifier.h
bzbarsky-apple: Those already exist.  This is the code for "they are both revoked".

examples/chip-tool/commands/common/CHIPCommand.cpp
shubhamdp: Thanks for the explanation. Yes, they are not exactly similar. May be once we have default revocation set built up, I can align the behaviour to attestation trust store. I have added the comment in the code, about how it behaves. Also raised #34434

examples/chip-tool/commands/common/CHIPCommand.h
```diff
@@ -222,11 +224,16 @@ class CHIPCommand : public Command
...
     chip::Optional<bool> mUseMaxSizedCerts;
     chip::Optional<bool> mOnlyAllowTrustedCdKeys;
+    chip::Optional<char *> mDacRevocationSetPath;
 
     // Cached trust store so commands other than the original startup command
     // can spin up commissioners as needed.
     static const chip::Credentials::AttestationTrustStore * sTrustStore;
 
+    // Cached DAC revocation delegate, this can be set using "--dac-revocation-set-path" argument
+    // Once set this will be used by all commands.
```
bzbarsky-apple, line 234: This information should be in the AddArgument help message, no?  People should not have to read the source to find out how arguments behave.

bzbarsky-apple (approved):

examples/chip-tool/commands/common/CHIPCommand.h
shubhamdp: Added it to the argument help!

src/credentials/attestation_verifier/DeviceAttestationVerifier.h
shubhamdp: Logged an issue to make it a subclass, will do it in follow-up https://github.com/project-chip/connectedhomeip/issues/34495

dhrishi:
@tcarmelveilleux @andy31415 Can you please review and approve this if it looks okay? We are hoping that this gets merged before the 1.4.TE2 deadline

src/credentials/BUILD.gn
```diff
@@ -157,6 +158,8 @@ static_library("default_attestation_verifier") {
     "attestation_verifier/DefaultDeviceAttestationVerifier.cpp",
     "attestation_verifier/DefaultDeviceAttestationVerifier.h",
     "attestation_verifier/DeviceAttestationDelegate.h",
+    "attestation_verifier/TestDACRevocationDelegateImpl.cpp",
+    "attestation_verifier/TestDACRevocationDelegateImpl.h",
```
tcarmelveilleux, line 162: Remove these files from here. and the `jsoncpp_root` dependency in default_attestation_verifier

src/credentials/attestation_verifier/DefaultDeviceAttestationVerifier.h
```diff
@@ -57,7 +57,11 @@ class CsaCdKeysTrustStore : public WellKnownKeysTrustStore
 class DefaultDACVerifier : public DeviceAttestationVerifier
 {
 public:
-    DefaultDACVerifier(const AttestationTrustStore * paaRootStore) : mAttestationTrustStore(paaRootStore) {}
+    DefaultDACVerifier(const AttestationTrustStore * paaRootStore,
+                       DeviceAttestationRevocationDelegate * revocationDelegate = nullptr) :
```
tcarmelveilleux, line 61: Instead of having a default parameter, just add a second constructor with both arguments. This will keep backwards compatibility.

src/credentials/attestation_verifier/DefaultDeviceAttestationVerifier.h
```diff
@@ -84,6 +88,7 @@ class DefaultDACVerifier : public DeviceAttestationVerifier
 
     CsaCdKeysTrustStore mCdKeysTrustStore;
     const AttestationTrustStore * mAttestationTrustStore;
+    DeviceAttestationRevocationDelegate * mRevocationDelegate = nullptr;
```
tcarmelveilleux, line 91: Suggest having a setter for the revocation delegate.

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
```diff
@@ -0,0 +1,205 @@
...
+CHIP_ERROR TestDACRevocationDelegateImpl::GetIssuerNameBase64Str(const ByteSpan & certDer,
+                                                                 MutableCharSpan & outIssuerNameBase64String)
+{
+    uint8_t issuerBuf[kMaxCertificateDistinguishedNameLength] = { 0 };
+    MutableByteSpan issuer(issuerBuf);
+
+    ReturnErrorOnFailure(ExtractIssuerFromX509Cert(certDer, issuer));
+    VerifyOrReturnError(outIssuerNameBase64String.size() >= BASE64_ENCODED_LEN(issuer.size()), CHIP_ERROR_BUFFER_TOO_SMALL);
+
+    uint32_t encodedLen = Base64Encode32(issuer.data(), static_cast<uint32_t>(issuer.size()), outIssuerNameBase64String.data());
```
tcarmelveilleux, line 142: This can be simply `Base64Encode`.  I don't know what the `32` suffix is about.

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
```diff
@@ -0,0 +1,205 @@
...
+    MutableCharSpan akid(akidHexStrBuffer);
+
+    VerifyOrReturnValue(CHIP_NO_ERROR == GetIssuerNameBase64Str(certDer, issuerName), false);
+    ChipLogDetail(NotSpecified, "Issuer: %.*s", static_cast<int>(issuerName.size()), issuerName.data());
+
+    VerifyOrReturnValue(CHIP_NO_ERROR == GetSerialNumberHexStr(certDer, serialNumber), false);
+    ChipLogDetail(NotSpecified, "Serial Number: %.*s", static_cast<int>(serialNumber.size()), serialNumber.data());
+
+    VerifyOrReturnValue(CHIP_NO_ERROR == GetAKIDHexStr(certDer, akid), false);
+    ChipLogDetail(NotSpecified, "AKID: %.*s", static_cast<int>(akid.size()), akid.data());
```
tcarmelveilleux, line 166: Add a TODO (and create associated issue) to cross-validate the CRLSignerCertificate and CRLSignerDelegator per spec:

```
. If a matching `RevocationSet` is found, determined if `serialNumber` is a member of the set.
    .. If the entity type whose revocation status is being verified is a PAI, then the subject PAI's issuer (a PAA) SHALL fulfill one of the following two cases, otherwise return immediately:
      ... The Subject and Subject Key of the PAI certificate's issuer matches exactly the CRLSignerCertificate's subject (i.e. the CRLSignerCertificate is a PAA).
      ... The Subject and Subject Key of the PAI certificate's issuer matches exactly the PAA which is the issuer of the CRLSignerCertificate (i.e. the CRLSignerCertificate is a CRL signer delegated by a PAA).
    .. If the entity type whose revocation status is being verified is a DAC, then:
      ... If the `CRLSignerDelegator` is present, then the subject DAC's issuer (a PAI) SHALL match the `CRLSignerDelegator` in both Subject Key and Subject, otherwise return immediately.
      ... If the `CRLSignerDelegator` is absent, then the subject DAC's issuer (a PAI) SHALL match the `CRLSignerCertificate` in both Subject Key and Subject, otherwise return immediately.
```

Those checks are not currently do, and MUST be done later once this is supported in revocation set construction.

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
```diff
@@ -0,0 +1,205 @@
...
+    return IsEntryInRevocationSet(akid, issuerName, serialNumber);
+}
+
+void TestDACRevocationDelegateImpl::CheckForRevokedDACChain(
+    const DeviceAttestationVerifier::AttestationInfo & info,
+    Callback::Callback<DeviceAttestationVerifier::OnAttestationInformationVerification> * onCompletion)
+{
+    AttestationVerificationResult attestationError = AttestationVerificationResult::kSuccess;
+
+    if (mDeviceAttestationRevocationSetPath != nullptr)
```
tcarmelveilleux, line 177: Suggest an early exit with call of the callback if `mDeviceAttestationRevocationSetPath == nullptr`, to de-nest this next part.

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
```diff
@@ -0,0 +1,205 @@
...
+    {
+        ChipLogDetail(NotSpecified, "Checking for revoked DAC in %s", mDeviceAttestationRevocationSetPath);
```
tcarmelveilleux, line 179: Suggeste using std::string for the mDeviceAttestationRevocationSetPath

src/credentials/tests/TestDeviceAttestationCredentials.cpp
tcarmelveilleux, line 1: The test needs to now have a dependency on `:test_dac_revocation_delegate` target.

examples/chip-tool/commands/common/CredentialIssuerCommands.h
```diff
@@ -57,10 +57,14 @@ class CredentialIssuerCommands
...
      *
+     * @param[in] revocationDelegate A pointer to the Device Attestation Revocation Delegate for checking revoked DACs and PAIs.
+     *                               Default is nullptr.
+     *
      * @return CHIP_ERROR CHIP_NO_ERROR on success, or corresponding error code.
      */
-    virtual CHIP_ERROR SetupDeviceAttestation(chip::Controller::SetupParams & setupParams,
-                                              const chip::Credentials::AttestationTrustStore * trustStore) = 0;
+    virtual CHIP_ERROR
+    SetupDeviceAttestation(chip::Controller::SetupParams & setupParams, const chip::Credentials::AttestationTrustStore * trustStore,
```
tcarmelveilleux, line 66: Please do not use default params for this, just require the value to always be passed.

tcarmelveilleux (approved):

examples/chip-tool/commands/common/CredentialIssuerCommands.h
shubhamdp: not using the default value.

src/credentials/BUILD.gn
shubhamdp: Thanks for pointing it out, removed. Thats why tests were passing!

src/credentials/attestation_verifier/DefaultDeviceAttestationVerifier.h
shubhamdp: Added one more constructor.

src/credentials/attestation_verifier/DefaultDeviceAttestationVerifier.h
shubhamdp: Added the setter API

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
```diff
@@ -0,0 +1,205 @@
...
+CHIP_ERROR BytesToHexStr(const ByteSpan & bytes, MutableCharSpan & outHexStr)
+{
+    Encoding::HexFlags flags = Encoding::HexFlags::kUppercase;
+    ReturnErrorOnFailure(BytesToHex(bytes.data(), bytes.size(), outHexStr.data(), outHexStr.size(), flags));
+    outHexStr.reduce_size(2 * bytes.size());
+    return CHIP_NO_ERROR;
+}
+} // anonymous namespace
+
+CHIP_ERROR TestDACRevocationDelegateImpl::SetDeviceAttestationRevocationSetPath(const char * path)
```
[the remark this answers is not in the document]
shubhamdp: Using `std::string_view` for this.

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
shubhamdp: 32 suffix returns the 32 bit value. `Base64Encode` returns 16 bit value. Using `Base64Encode`.

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
shubhamdp: Added a TODO and an issue: #34587

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
shubhamdp: Since we have to make a call at the end with attestation result, that why it was like this. Inverted the check and returning early.

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
shubhamdp: Using earlier suggestion, `std::string_view`

src/credentials/tests/TestDeviceAttestationCredentials.cpp
shubhamdp: Added!

src/credentials/tests/TestDeviceAttestationCredentials.cpp
```diff
@@ -413,3 +416,158 @@ TEST_F(TestDeviceAttestationCredentials, TestAttestationTrustStore)
...
 }
+
+static void WriteTestRevokedData(const char * jsonData, const char * fileName)
+{
+    // write data to /tmp/sample_revoked_set.json using fstream APIs
+    std::ofstream file;
+    file.open(fileName, std::ofstream::out | std::ofstream::trunc);
+    file << jsonData;
+    file.close();
+}
```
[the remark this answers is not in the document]
shubhamdp: Added TODO for now, will fix in follow-up #34588

tcarmelveilleux:
Broke python tests on HEAD

Issue #27976, opened 2023-07-14
[Platform] Implement event queu cleanup on unit test teardown.


lpbeliveau-silabs:
### Reproduction steps

When adding a new test to the CI unit test, if unprocessed events remain in the platform manager's event queue, they will persist into the following tests. This is will impact the time `DrainAndServiceIO()` takes to process events in the queue and leads to failure when executing CI unit test for OpenIoT SDK as the tests are ran in a single file.

To reproduce the steps, the TestReportScheduler on the CI and it will cause TestAclAttribute to fail on:
`src/app/tests/TestAclAttribute.cpp:251: assertion failed: "delegate.mGotReport"`

Which is expecting the report to be processed after  `DrainAndServiceIO()` was called.

It is unclear what exactly is causing this behaviour but it is reproducible by having unprocessed events on in the queue at the end of previously executed tests. 

### Platform

other

### Type

Unit tested, CI tested

Alami-Amine:
a fix (workaround?) was found to this issue, just putting it here for reference. If we think this is ok we can close the issue.

PR #33654, opened 2024-05-29
Fix issue related to teardown of some App Tests

Alami-Amine:
Is this required as part of this change?  Seems like an unrelated fix that could use very careful review not in the context of a mass-change....

_Originally posted by @bzbarsky-apple in https://github.com/project-chip/connectedhomeip/pull/33638#discussion_r1617960964_
            
## Context

- `TestAclAttribute.cpp` and `TestReportScheduler.cpp` are triggering failures in (some) tests that follow them when being run on Nordic CI (on the Zephyr Native POSIX target, which compiles all unit tests into a single binary).
- At the moment, `TestReportScheduler.cpp` is not being run for Nordic, and `TestAclAttribute.cpp` is being configured to run last, right after all the tests that it triggers failures.
- That tests on which the failures happen are `TestAclEvent` and `TestReadInteraction`: it seems that the problem is related to `Report Data` that comes after a Read Request --> the Callback that will call `BuildAndSendSingleReportData` is not being called.
- A solution (workaround?) which solves this problem, however, is to just call `DrainAndServiceIO()` right in the teardown of these Tests (`TestAclAttribute.cpp` and `TestReportScheduler.cpp` ) --> However, I thought I might just put it in the teardown of AppTestContext incase we add more tests in the future that might trigger the same problem

UPDATE: There was already an issue for this here:  [#27976](https://github.com/project-chip/connectedhomeip/issues/27976)

### DIFF
--- a/src/app/tests/AppTestContext.cpp
+++ b/src/app/tests/AppTestContext.cpp
@@ -49,6 +49,12 @@ void AppContext::SetUpTestSuite()
 
 void AppContext::TearDownTestSuite()
 {
+    // Adding a DrainAndServiceIO call fixes the teardown for some tests (TestAclAttribute and TestReportScheduler); these were
+    // triggering failures in tests that follow them when running on nRF CI (Zephyr native_posix where all Unit Tests are compiled
+    // into a single file)
+    // TODO: understand this more and solve underlying issue
+    LoopbackMessagingContext::DrainAndServiceIO();
+
     chip::DeviceLayer::PlatformMgr().Shutdown();
     LoopbackMessagingContext::TearDownTestSuite();
 }
--- a/src/app/tests/AppTestContext.cpp
+++ b/src/app/tests/AppTestContext.cpp
@@ -49,10 +49,16 @@ void AppContext::SetUpTestSuite()
 
 void AppContext::TearDownTestSuite()
 {
-    // Adding a DrainAndServiceIO call fixes the teardown for some tests (TestAclAttribute and TestReportScheduler); these were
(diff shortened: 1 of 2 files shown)


src/app/tests/AppTestContext.cpp
```diff
@@ -49,6 +49,12 @@ void AppContext::SetUpTestSuite()
 
 void AppContext::TearDownTestSuite()
 {
+    // Adding a DrainAndServiceIO call fixes the teardown for some tests (TestAclAttribute and TestReportScheduler); these were
+    // triggering failures in tests that follow them when running on nRF CI (Zephyr native_posix where all Unit Tests are compiled
+    // into a single file)
+    // TODO: understand this more and solve underlying issue
```
bzbarsky-apple, line 55: The comment that was removed in the .gn talks about the details of the underlying issue, no:

> unprocessed reported engine runs are remaining

(Should be "reporting engine", probably).

How about:

```suggestion
    // Some test suites finish with unprocessed work left in the platform manager event queue.
    // This can particularly be a problem when this unprocessed work involves reporting engine runs,
    // since those can take a while and cause later tests to not reach their queued work before
    // their timeouts hit.  This is only an issue in setups where all unit tests are compiled into
    // a single file (e.g. nRF CI (Zephyr native_posix)).
    //
    // Work around this issue by doing a DrainAndServiceIO() here to attempt to flush out any queued-up work.
    //
    // TODO: Solve the underlying issue where test suites leave unprocessed work.  Or is this actually
    // the right solution?
```

bzbarsky-apple (approved):

andy31415 (approved):

PR #33666, opened 2024-05-29
[Darwin] MTRDeviceConnectivityMonitor stopMonitoring crash fix

jtung-apple:
This fixes a crash in a shutdown / cleanup scenario where MTRDevice goes away and calls `stopMonitoring` when the shared resolver connection/queue have already cleaned up.

Also added some logging for when resubscription happens in the MTRDevice layer.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -395,8 +395,8 @@ static NSString * const sLastInitialSubscribeLatencyKey = @"lastInitialSubscribe
     NSMutableSet<MTRClusterPath *> * _persistedClusters;
 
     // When we last failed to subscribe to the device (either via
-    // _setupSubscription or via the auto-resubscribe behavior of the
-    // ReadClient).  Nil if we have had no such failures.
+    // _setupSubscriptionWithReason or via the auto-resubscribe behavior
+    // of the ReadClient).  Nil if we have had no such failures.
     NSDate * _Nullable _lastSubscriptionFailureTime;
     MTRDeviceConnectivityMonitor * _connectivityMonitor;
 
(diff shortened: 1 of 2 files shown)


src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -745,10 +745,10 @@ - (void)setDelegate:(id<MTRDeviceDelegate>)delegate queue:(dispatch_queue_t)queu
         if ([self _deviceUsesThread]) {
             [self _scheduleSubscriptionPoolWork:^{
                 std::lock_guard lock(self->_lock);
-                [self _setupSubscription];
+                [self _setupSubscriptionWithReason:@"delegate is set and subscription is scheduled"];
             } inNanoseconds:0 description:@"MTRDevice setDelegate first subscription"];
         } else {
-            [self _setupSubscription];
+            [self _setupSubscriptionWithReason:@"delegate is set and subscription is needed"];
```
kiel-apple, line 751: on one hand we could make these constants

on the other hand we could we could make it more clear that the method is just using the string for logging  (`_setupSubscriptionWithLoggedReason:`)

kiel-apple (approved):

src/darwin/Framework/CHIP/MTRDeviceConnectivityMonitor.mm
```diff
@@ -258,7 +258,7 @@ - (void)_stopMonitoring
             dispatch_after(dispatch_time(DISPATCH_TIME_NOW, kSharedConnectionLingerIntervalSeconds * NSEC_PER_SEC), sSharedResolverQueue, ^{
                 std::lock_guard lock(sConnectivityMonitorLock);
 
-                if (!sConnectivityMonitorCount) {
+                if (!sConnectivityMonitorCount && sSharedResolverConnection) {
```
kiel-apple, line 261: i take it this is the "juicy part"?

bzbarsky-apple (approved):

src/darwin/Framework/CHIP/MTRDevice.mm
jtung-apple: Hmm, I like the second version of naming the methods better. Right now I'm using the same convention as `_triggerResubscribeWithReason:` and so I think we can stick with this for now, and rename everything in a separate patch.

src/darwin/Framework/CHIP/MTRDeviceConnectivityMonitor.mm
jtung-apple: While this is also "juicy", the fix below in `stopMonitoring` is what I'm fixing for a particular crash report. So both are needed.

bzbarsky-apple:

PR #32182, opened 2024-02-16
TC-OPCREDS-3.2: Add

gvargas-csa:
Corresponding test plan: [3.2.2. [TC-OPCREDS-3.2] Attribute-CurrentFabricIndex validation [DUT-Server]](https://github.com/CHIP-Specifications/chip-test-plans/blob/38f6fade945ea5e63efdc4c5afa71b5fba8b6ff5/src/cluster/node_operational_credentials.adoc#tc-opcreds-3-2-attribute-currentfabricindex-validation-dut-server)

PR #33667, opened 2024-05-29
Updated TestSteps to add expectation strings and match the descriptions in the current test plan.

feasel0:
Updated the TestSteps for several of the python tests, as per [#240](https://github.com/project-chip/matter-test-scripts/issues/240), now that expectations are supported.

Descriptions and expectations now match what's currently in the test plan.  I ran each of these tests through the generator script and compared the output to the current test plan to confirm it matches.

Tests updated:
`TC_ACE_1_3`
`TC_ACL_2_2`
`TC_BOOLCFG_*`

Some notes:
- The current test plans contain adoc comments which separate out certain groups of steps.  I preserved those comments as python comments, so they can be seen in the python code, however they are not part of the dataclass so they won't get rendered by the generator script and make their way back into the adoc.
- [FIXED] There seems to be a mistake in the test plan for `TC_BOOLCFG_4_4` step 14.  Looks like the expectation is either missing or it got rolled into the description.  [See here](https://github.com/CHIP-Specifications/chip-test-plans/blob/1709d736d85b2c622b4f8e718ef57b5e1c93f0a8/src/cluster/boolean_state_configuration.adoc?plain=1#L594).  This should probably be figured out before we merge these changes.
- I added some functions and strings to `src/python_testing/test_plan_support.py` to take the place of the adoc macros.  This has currently been done in `TC_BOOLCFG_2_2`, but I will update the others if we're happy with this solution.


src/python_testing/TC_BOOLCFG_2_1.py
```diff
@@ -34,16 +34,16 @@ def desc_TC_BOOLCFG_2_1(self) -> str:
...
-            TestStep(2, "Read attribute list to determine supported attributes"),
-            TestStep(3, "Read SupportedSensitivityLevels attribute, if supported"),
-            TestStep(4, "Read CurrentSensitivityLevel attribute, if supported"),
-            TestStep(5, "Read DefaultSensitivityLevel attribute, if supported"),
-            TestStep(6, "Read AlarmsActive attribute, if supported"),
-            TestStep(7, "Read AlarmsSuppressed attribute, if supported"),
-            TestStep(8, "Read AlarmsEnabled attribute, if supported"),
-            TestStep(9, "Read AlarmsSupported attribute, if supported"),
-            TestStep(10, "Read SensorFault attribute, if supported"),
+            TestStep(1, "{comDutTH}.", "", is_commissioning=True),
```
cecille, line 37: things I should have mentioned earlier - things get a bit weird if you pull these directly from the adoc files because all the stuff in the curly braces are adoc macros (er...they don't call them macros, but basically macros). 

So this will render correctly in the adoc (I think?), but it'll also show up directly like this in the test harness.

The solution Rene and I were kicking around was using actual python functions to return the text. ex like here: https://github.com/project-chip/connectedhomeip/pull/32182/files#diff-8bafa8e5179589abd80fe1365c30897596b32305372c4ee3986f1062b9afa0fd

The bonus there is you can consolidate all of this nicely by doing ex.
```
def constraints(min: typing.Union[int, str], max: stepnum: typing.Union[int, str], type: typing.Union[int, str, None):
  type_str = '' if type is none else f'dut_reply() a {type} value. '
  return f'{type_str}Value has to be between a range of {min} and {max}'

TestStep(3, read_attribute('SupporedSensitivitiyLevels'), constraints('uint8', 2, 10)) 
```

aside...man I'd love to get rid of these attribute constraint tests. Maybe in 1.5 we can start picking them off from the easy DM XMLs. This seems like SO much manual copy-paste for no reason.

Anyway @ReneJosefsen - what are your thoughts here? Do we like this style better than what's in the adoc currently? Is this going to be easier for people to keep in sync if we can generate the test plans automatically?

ReneJosefsen: From just an initial view, this seems like a move in the wrong direction. As you say, this does render fine in adoc once the import is there, but the macros/document attributes are not going to be adjusted accordingly here so it will look very weird when running this on TH.

So for this to work we need to have a similar text replacement in place first IMO.

ReneJosefsen:
Good catch on step 14, I will get that sorted. It does look like it does not even render properly because it are missing the last column in that row.

src/python_testing/TC_BOOLCFG_2_1.py
feasel0: I updated `TC_BOOLCFG_2_2.py` to use macros.  See [here](https://github.com/project-chip/connectedhomeip/blob/f251be2951df7d6e856aa1f26c17c1ccaeac60d0/src/python_testing/TC_BOOLCFG_2_1.py#L38).

If we like this, I will update the rest of the `TC_BOOLCFG_*.py` files.

PR #243, opened 2020-04-03
Add HAP Crypto to src/crypto

bhaskar-apple:
#### Problem
No APIs to support symmetric key based payload encryption available in CHIP

 #### Summary of Changes
Adds new files from [HAP ADK](https://github.com/apple/HomeKitADK). These have been scrubbed to replace occurrences of `HAP` with `ADK`

```
src/crypto/CHIPBase+CompilerAbstraction.h
src/crypto/CHIPBase+Crypto.c
src/crypto/CHIPCrypto.h
src/crypto/CHIPOpenSSL.c
src/crypto/Makefile.am
src/crypto/tests/CHIPCryptoTest.c
src/crypto/tests/Makefile.am
```

fixes #147

PR #33672, opened 2024-05-29
Worked on issue 243; replaced input with self.wait_for_user_input; other changes

austina-csa:
Worked on issue #243 that was assigned to me.

This includes the following:

- Replaced input with self.wait_for_user_input in the following Python scripts:
  - TC_OpstateCommon.py
  - TC_RVCCLEANM_2_1.py
  - TC_RVCCLEANM_2_2.py
  - TC_RVCOPSTATE_2_1.py
  - TC_RVCOPSTATE_2_3.py
  - TC_RVCOPSTATE_2_4.py
  - TC_RVCRUNM_2_1.py
  - TC_RVCRUNM_2_2.py
  - matter_testing_support.py
- Removed input_msg argument from wait_for_user_input
- Accordingly, passed test steps into prompt_msg, and appended "Press Enter when ready".
- Adjusted the return value accordingly, taking into account that it needs to end with an \n for the text to print properly in the test framework.

### DIFF
--- a/src/python_testing/TC_OpstateCommon.py
+++ b/src/python_testing/TC_OpstateCommon.py
@@ -122,7 +122,7 @@ class TC_OPSTATE_BASE():
             self.write_to_app_pipe(command)
             time.sleep(0.1)
         else:
-            input("Press Enter when done.\n")
+            self.wait_for_user_input(prompt_msg="Press Enter when ready.\n")
 
     def send_manual_or_pipe_command(self, device: str, name: str, operation: str, param: Any = None):
         command = {
@@ -1113,7 +1113,7 @@ class TC_OPSTATE_BASE():
             # In CI environment, the STOP coommand (step 8) already resets the variables. Only ask for
             # reboot outside CI environment.
             if not self.is_ci:
-                input("Press Enter when done.\n")
+                self.wait_for_user_input(prompt_msg="Press Enter when ready.\n")
 
             # STEP 12: TH waits for {PIXIT.WAITTIME.REBOOT}
             self.step(12)
--- a/src/python_testing/TC_OpstateCommon.py
+++ b/src/python_testing/TC_OpstateCommon.py
@@ -1113,7 +1113,7 @@ class TC_OPSTATE_BASE():
             # In CI environment, the STOP coommand (step 8) already resets the variables. Only ask for
             # reboot outside CI environment.
             if not self.is_ci:
-                self.wait_for_user_input(prompt_msg="Press Enter when ready.\n")
+                self.wait_for_user_input(prompt_msg="Restart DUT. Press Enter when ready.\n")
 
             # STEP 12: TH waits for {PIXIT.WAITTIME.REBOOT}
             self.step(12)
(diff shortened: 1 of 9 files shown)


src/python_testing/TC_OpstateCommon.py
```diff
@@ -122,7 +122,7 @@ def send_raw_manual_or_pipe_command(self, command):
             self.write_to_app_pipe(command)
             time.sleep(0.1)
         else:
-            input("Press Enter when done.\n")
+            self.wait_for_user_input(prompt_msg="Press Enter when ready.\n")
```
cecille, line 125: can you please add the test step text in here? Otherwise I'm not sure quite what the TH shows.

src/python_testing/TC_OpstateCommon.py
```diff
@@ -1113,7 +1113,7 @@ async def TEST_TC_OPSTATE_BASE_2_5(self, endpoint=1):
             # In CI environment, the STOP coommand (step 8) already resets the variables. Only ask for
             # reboot outside CI environment.
             if not self.is_ci:
-                input("Press Enter when done.\n")
+                self.wait_for_user_input(prompt_msg="Press Enter when ready.\n")
```
cecille, line 1116: same

cecille (approved):

plauric (approved):

rquidute (approved):

mhazley (approved):

hicklin (approved):

andy31415:
@austina-csa this seems to need more CLA signing

src/python_testing/TC_OpstateCommon.py
austina-csa: It's unclear what the test step is based on the code I'm looking at, since this actually seems to vary.

The method send_raw_manual_or_pipe_command only passes a command as an argument (which is a dictionary containing the following keys: Name, Device, Operation, and Param, all of which are strings.)

The method send_raw_manual_or_pipe_command is wrapped by send_manual_or_pipe_command -- nothing seems to call it directly (based on a grep search). send_manual_or_pipe_command sends these arguments in a JSON format.

send_manual_or_pipe_command is called multiple times, but the steps differ. (For example, one step is "# STEP 6a: Manually put the device in the Stopped(0x00) operational state", another is "# STEP 6c: Manually put the device in the Running(0x01) operational state", etc., and all of these are indicated in _comments_ before this method is called.) There doesn't seem to be anything that prints the steps -- they just call self.step(int) beforehand.

There might have to be a non-trivial rework of this file to improve this method send_raw_manual_or_pipe_command. I can do that if needed, but I don't want to jump the gun without a prior discussion.

src/python_testing/TC_OpstateCommon.py
austina-csa: This one is more straightforward than the comment above since this appears to be tied to a specific step (11 - Restart DUT). I added this change (improving the string) to my local repo but haven't committed it yet, since it is in the same file as the comment above.

src/python_testing/TC_OpstateCommon.py
cecille: nah, let's just leave this for now then. If the RVC folks want to re-work it after seeing it in the TH, they can handle it themselves.

austina-csa: That leaves me with a couple of questions:

1. Should an issue be file to address the needed changes?
2. Should I revert this one-line change or keep it?

cecille: This is fine. RVC can open an issue if they find it's a problem when they go to test.

PR #33674, opened 2024-05-29
Darwin: Keep MTRCommissionableBrowser around until OnBleScanStopped

ksperling-apple:
Because the BLE platform implementation uses a separate dispatch queue, StopBleScan() does not synchrnously stop any further use of the current BleScannerDelegate. Keep the MTRCommissionableBrowser that contains the delegate alive until OnBleScanStopped to avoid a UAF.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRCommissionableBrowser.mm
+++ b/src/darwin/Framework/CHIP/MTRCommissionableBrowser.mm
@@ -65,7 +65,7 @@ public:
     id mBleScannerDelegateOwner;
 #endif // CONFIG_NETWORK_LAYER_BLE
 
-    CHIP_ERROR Start(id owner, id<MTRCommissionableBrowserDelegate> delegate, MTRDeviceController * controller, dispatch_queue_t queue)
+    CHIP_ERROR Start(id<MTRCommissionableBrowserDelegate> delegate, MTRDeviceController * controller, dispatch_queue_t queue)
     {
         assertChipStackLockedByCurrentThread();
 
@@ -81,10 +81,7 @@ public:
         ResetCounters();
 
 #if CONFIG_NETWORK_LAYER_BLE
-        mBleScannerDelegateOwner = owner; // retain the owner until OnBleScanStopped is called
         ReturnErrorOnFailure(PlatformMgrImpl().StartBleScan(this));
-#else
-        (void) owner;
 #endif // CONFIG_NETWORK_LAYER_BLE
 
         ReturnErrorOnFailure(Resolver::Instance().Init(chip::DeviceLayer::UDPEndPointManager()));
@@ -97,7 +94,7 @@ public:
             chip::Inet::InterfaceId::Null(), this);
     }
 
-    CHIP_ERROR Stop()
+    CHIP_ERROR Stop(id owner)
     {
         assertChipStackLockedByCurrentThread();
 
@@ -116,7 +113,14 @@ public:
         mDiscoveredResults = nil;
 
 #if CONFIG_NETWORK_LAYER_BLE
-        ReturnErrorOnFailure(PlatformMgrImpl().StopBleScan());
+        mBleScannerDelegateOwner = owner; // retain the owner until OnBleScanStopped is called
+        CHIP_ERROR err = PlatformMgrImpl().StopBleScan();
+        if (err != CHIP_NO_ERROR) {
+            mBleScannerDelegateOwner = nil;
+            return err;
+        }
+#else
+        (void) owner;
 #endif // CONFIG_NETWORK_LAYER_BLE
 
         return ChipDnssdStopBrowse(this);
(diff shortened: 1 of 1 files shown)


src/darwin/Framework/CHIP/MTRCommissionableBrowser.mm
```diff
@@ -77,7 +81,10 @@ CHIP_ERROR Start(id<MTRCommissionableBrowserDelegate> delegate, MTRDeviceControl
         ResetCounters();
 
 #if CONFIG_NETWORK_LAYER_BLE
+        mBleScannerDelegateOwner = owner; // retain the owner until OnBleScanStopped is called (see rdar://128723029)
```
anush-apple, line 84:
```suggestion
        mBleScannerDelegateOwner = owner; // retain the owner until OnBleScanStopped is called
```

src/darwin/Framework/CHIP/MTRCommissionableBrowser.mm
```diff
@@ -77,7 +81,10 @@ CHIP_ERROR Start(id<MTRCommissionableBrowserDelegate> delegate, MTRDeviceControl
         ResetCounters();
 
 #if CONFIG_NETWORK_LAYER_BLE
+        mBleScannerDelegateOwner = owner; // retain the owner until OnBleScanStopped is called
```
anush-apple, line 84: What happens is Start is called twice ?

ksperling-apple: That's prevented by the `VerifyOrReturnError` statements above.

anush-apple (approved):

src/darwin/Framework/CHIP/MTRCommissionableBrowser.mm
bzbarsky-apple, line 1: "synchrnously" is mis-spelled in the commit message.

src/darwin/Framework/CHIP/MTRCommissionableBrowser.mm
```diff
@@ -77,7 +81,10 @@ CHIP_ERROR Start(id<MTRCommissionableBrowserDelegate> delegate, MTRDeviceControl
...
         ReturnErrorOnFailure(PlatformMgrImpl().StartBleScan(this));
```
bzbarsky-apple, line 85: Won't we leak due to the cycle if StartBleScan() fails?

Also, what happens if StartBleScan succeeds but then `Resolver::Instance().Init()` below fails?  Is the expectation that someone will still call `stop` if `start` returns `NO`, or that something else will at least trigger a `Stop()` call on us?

It might be safer to take the reference to the relevant owner in `Stop`.  We would still leak if `StopBleScan` returns failure, but that's saner than having a use-after-free if that happens (though we should document that that's the setup).

Separately, if `StopBleScan()` returns failure, shouldn't we still `ChipDnssdStopBrowse` anyway?  That part is pre-existing....

ksperling-apple: All these corner cases are handled pretty badly in the existing code, but I don't want to address these issues in this fix. I'll follow your suggestion of only taking the reference in `Stop()`

ksperling-apple: Also `StopBleScan` doesn't actually fail in practice, unless the stack is not set up yet.

src/darwin/Framework/CHIP/MTRCommissionableBrowser.mm
ksperling-apple: artistic license ;-)

src/darwin/Framework/CHIP/MTRCommissionableBrowser.mm
```diff
@@ -109,7 +113,14 @@ CHIP_ERROR Stop()
         mDiscoveredResults = nil;
 
 #if CONFIG_NETWORK_LAYER_BLE
-        ReturnErrorOnFailure(PlatformMgrImpl().StopBleScan());
+        mBleScannerDelegateOwner = owner; // retain the owner until OnBleScanStopped is called
+        CHIP_ERROR err = PlatformMgrImpl().StopBleScan();
+        if (err != CHIP_NO_ERROR) {
+            mBleScannerDelegateOwner = nil;
```
bzbarsky-apple, line 119: I think we should just leave mBleScannerDelegateOwner in this case, with a comment that a leak is better than use-after-free...

bzbarsky-apple (approved):

src/darwin/Framework/CHIP/MTRCommissionableBrowser.mm
bzbarsky-apple: Or I guess we could document why if StopBleScan fails we can assume there is in fact no scan and hence that no one will try calling us.

PR #33677, opened 2024-05-30
Stop using human-readable release names in Darwin availability.yaml.

bzbarsky-apple:
People are making assumptions based on those, and those assumptions will come back to bite them.

Just switch to using UUIDs for the release names.


woody-apple (approved):

mkardous-silabs (approved):
Didn't check that all the renames matched - assumed find/replace

yunhanw-google (approved):

bzbarsky-apple:
> Didn't check that all the renames matched - assumed find/replace

Yep, I did `git sed`.  But also, if the generated code did not change, they match.  ;)

PR #33681, opened 2024-05-31
WIP: generate PICS tables

cecille:
Do not review - design sketch only


cecille:
Let's use alchemy instead.

PR #33684, opened 2024-05-31
[ESP32] Unregistering wifi and IP event handlers before factory reset to avoid random crashes due to PostEventQueue failure.

shripad621git:
**Problem**
- Some random crashes are observed in case the event queue is full and some events like WIFI_EVENT_STA_DISCONNECT gets posted to the platform queue at the time of factory reset.
- This leads to crash due to `PostEventOrDie` on failure to post the `wifi event.`

**Change Overview**
- Unregistered the `wifi and IP `event handlers just before the factory reset flow in `DoFactoryReset` to avoid posting esp events to the esp default event loop.

**Testing**
- Tested the changes in the factory reset flow by adding some debug prints.
- Tested the behaviour for esp32 lighting-app with these changes.

### DIFF
--- a/src/platform/ESP32/ConfigurationManagerImpl.cpp
+++ b/src/platform/ESP32/ConfigurationManagerImpl.cpp
@@ -29,6 +29,7 @@
 #include <lib/support/CodeUtils.h>
 #include <platform/ConfigurationManager.h>
 #include <platform/ESP32/ESP32Config.h>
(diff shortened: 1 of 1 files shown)


andy31415 (approved):

bzbarsky-apple (approved):
Rubber-stamping, assuming this gets proper review from a platform owner.

PR #33685, opened 2024-05-31
Added endpoint id for Thread network commissioning instance

DejinChen:
Currently, the generic thread network commissioning instance can only be added at endpoint 0. To support multiple network interfaces, a configurable endpoint ID is necessary.
- Added a definition endpoint ID for Thread network commissioning instance.

### DIFF
--- a/src/platform/BUILD.gn
+++ b/src/platform/BUILD.gn
@@ -85,6 +85,9 @@ if (chip_device_platform != "none" && chip_device_platform != "external") {
     # devices with multiple radios that have different sleep behavior for
     # different radios.
     chip_device_config_enable_dynamic_mrp_config = false
+
+    # Define the default endpoint id for the generic Thread network commissioning instance
+    chip_device_config_thread_network_endpoint_id = 0
   }
 
   if (chip_stack_lock_tracking == "auto") {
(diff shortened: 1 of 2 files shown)


src/platform/OpenThread/GenericThreadStackManagerImpl_OpenThread.hpp
```diff
@@ -75,8 +75,12 @@ static_assert(OPENTHREAD_API_VERSION >= 219, "OpenThread version too old");
 // Network commissioning
 namespace {
 #ifndef _NO_NETWORK_COMMISSIONING_DRIVER_
+#ifndef CHIP_DEVICE_CONFIG_THREAD_NETWORK_ENDPOINT_ID
+#define CHIP_DEVICE_CONFIG_THREAD_NETWORK_ENDPOINT_ID (0)
+#endif
```
bzbarsky-apple, line 80: This should be in a config header (either one that includes some sort of consumer-provided config, or one generated based on gn args).

In particular, it should be _very_ clear to someone using the SDK how to set this properly; right now the only "proper" way to do it seems to be "ensure it's in the command line when this file is being compiled or something".

bzbarsky-apple (changes requested):

bauerschwan (approved):

src/platform/OpenThread/GenericThreadStackManagerImpl_OpenThread.hpp
DejinChen: Added a gn arg to define the endpoint id.

src/platform/BUILD.gn
```diff
@@ -85,6 +85,9 @@ if (chip_device_platform != "none" && chip_device_platform != "external") {
     # devices with multiple radios that have different sleep behavior for
     # different radios.
     chip_device_config_enable_dynamic_mrp_config = false
+
+    # Define the default endpoint id for the generic Thread network commissioning instance
+    chip_device_config_thread_network_endpoint_id = 0
```
chshu, line 90: There should be similar requirement for Wi-Fi, shall we have a common solution for multiple networks use case.

DejinChen: This definition is for the generic Thread driver instance in Platform, while most Wi-Fi driver instances are in examples. Maybe a further followup should do for this.

chshu (approved):

bzbarsky-apple (approved):

PR #33687, opened 2024-05-31
Update network-commissioning to use `AddResponse` for responding to scan responses 

andy31415:
This removes some usages of seemingly internal API for CommandHandler. This makes the logic more safe because it starts taking advantage of CommandHandler built-in behavior:

- Auto snapshot/rollback on failures and setting error codes on failures to write
- Simpler overall code

Also fixed oddities in the code:
- Release of iterators was inconsistent: thread was not checking for null, wifi was
- Release of iterators was not done on early exit or if internal defines were not set (but then maybe also those callbacks are never called, however unfortunately everything is common code)

### Tested:

- validated that I get Wifi results from a linux app and thread results from a nrfconnect board

### DIFF
--- a/src/app/clusters/network-commissioning/network-commissioning.cpp
+++ b/src/app/clusters/network-commissioning/network-commissioning.cpp
@@ -224,7 +224,8 @@ private:
     /// Fills up scanResponseArray with valid and de-duplicated thread responses from mNetworks.
     /// Handles sorting and keeping only larger rssi
     ///
-    /// Returns the valid list of scan responses into `validResponses`
+    /// Returns the valid list of scan responses into `validResponses`, which is only valid
+    /// as long as scanResponseArray is valid.
     CHIP_ERROR LoadResponses(Platform::ScopedMemoryBuffer<ThreadScanResponse> & scanResponseArray,
                              Span<ThreadScanResponse> & validResponses) const;
 };
--- a/src/app/clusters/network-commissioning/network-commissioning.cpp
+++ b/src/app/clusters/network-commissioning/network-commissioning.cpp
@@ -215,8 +215,8 @@ private:
     CharSpan mDebugText;
     ThreadScanResponseIterator * mNetworks;
 
-    /// parses mNetworks into `buffer` and performs post-processing (sort and dedup)
-    /// on them.
+    /// Fills up scanResponseArray with valid and de-duplicated thread responses from mNetworks.
+    /// Handles sorting and keeping only largers rssi
     ///
     /// Returns the valid list of scan responses into `validResponses`
     CHIP_ERROR LoadResponses(Platform::ScopedMemoryBuffer<ThreadScanResponse> & scanResponseArray,
@@ -267,10 +267,9 @@ CHIP_ERROR ThreadScanResponseToTLV::LoadResponses(Platform::ScopedMemoryBuffer<T
             scanResponseArrayLength++;
         }
         scanResponseArray[scanResponseArrayLength - 1] = scanResponse;
+        Sorting::InsertionSort(scanResponseArray.Get(), scanResponseArrayLength,
+                               [](const ThreadScanResponse & a, const ThreadScanResponse & b) -> bool { return a.rssi > b.rssi; });
     }
-
-    Sorting::InsertionSort(scanResponseArray.Get(), scanResponseArrayLength,
-                           [](const ThreadScanResponse & a, const ThreadScanResponse & b) -> bool { return a.rssi > b.rssi; });
 
     validResponses = Span<ThreadScanResponse>(scanResponseArray.Get(), scanResponseArrayLength);
 
--- a/src/app/clusters/network-commissioning/network-commissioning.cpp
+++ b/src/app/clusters/network-commissioning/network-commissioning.cpp
@@ -226,11 +226,11 @@ private:
     ///
     /// Returns the valid list of scan responses into `validResponses`
     CHIP_ERROR LoadResponses(Platform::ScopedMemoryBuffer<ThreadScanResponse> & scanResponseArray,
-                             Span<ThreadScanResponse> &validResponses) const;
+                             Span<ThreadScanResponse> & validResponses) const;
 };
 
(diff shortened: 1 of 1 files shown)


src/app/clusters/network-commissioning/network-commissioning.cpp
```diff
@@ -115,6 +115,224 @@ BitFlags<Feature> WiFiFeatures(WiFiDriver * driver)
...
+};
+
+CHIP_ERROR ThreadScanResponseToTLV::LoadResponses(Platform::ScopedMemoryBuffer<ThreadScanResponse> & scanResponseArray,
+                                                  Span<ThreadScanResponse> & validResponses) const
+{
+    VerifyOrReturnError(scanResponseArray.Alloc(chip::min(mNetworks->Count(), kMaxNetworksInScanResponse)), CHIP_ERROR_NO_MEMORY);
+
+    ThreadScanResponse scanResponse;
+    size_t scanResponseArrayLength = 0;
+    for (; mNetworks != nullptr && mNetworks->Next(scanResponse);)
```
jmartinez-silabs, line 240: as discussed this de-duplication by keeping the best RSSI and then sorting the final list of networks can be optimized and probably brought to the source of the scan callback in the thread network driver.
To be done in a separate pr.

jmartinez-silabs (approved):

src/app/clusters/network-commissioning/network-commissioning.cpp
```diff
@@ -115,6 +115,225 @@ BitFlags<Feature> WiFiFeatures(WiFiDriver * driver)
...
+
+private:
+    T * mValue;
+};
+
+/// Convenience macro to auto-create a variable for you to release the given name at
+/// the exit of the current scope.
+#define DEFER_AUTO_RELEASE(name) AutoRelease autoRelease##__COUNTER__(name)
+
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFI_STATION || CHIP_DEVICE_CONFIG_ENABLE_WIFI_AP || CHIP_DEVICE_CONFIG_ENABLE_THREAD
```
bzbarsky-apple, line 141: Why would we include wifi code just because CHIP_DEVICE_CONFIG_ENABLE_THREAD?

src/app/clusters/network-commissioning/network-commissioning.cpp
```diff
@@ -115,6 +115,225 @@ BitFlags<Feature> WiFiFeatures(WiFiDriver * driver)
...
+        ReturnErrorOnFailure(writer.StartContainer(TLV::ContextTag(Commands::ScanNetworksResponse::Fields::kWiFiScanResults),
+                                                   TLV::kTLVType_Array, listContainerType));
+
+        if ((mStatus == Status::kSuccess) && (mNetworks != nullptr))
+        {
+            WiFiScanResponse scanResponse;
+            size_t networksEncoded = 0;
+            while (mNetworks->Next(scanResponse))
+            {
+                // NOTE: it is assumed that scanResponse.ssid/bssid lifetime is larger than just within the `next` call
```
bzbarsky-apple: This comment is very confusing. What we are really assuming is that those lifetimes are ok until the next time we call `Next()` or until mNetworks goes away, right?

src/app/clusters/network-commissioning/network-commissioning.cpp
```diff
@@ -115,6 +115,225 @@ BitFlags<Feature> WiFiFeatures(WiFiDriver * driver)
...
+
+    CHIP_ERROR EncodeTo(TLV::TLVWriter & writer, TLV::Tag tag) const override;
+
+private:
+    Status mStatus;
+    CharSpan mDebugText;
+    ThreadScanResponseIterator * mNetworks;
+
+    /// Fills up scanResponseArray with valid and de-duplicated thread responses from mNetworks.
+    /// Handles sorting and keeping only largers rssi
```
bzbarsky-apple, line 225:
```suggestion
    /// Handles sorting and keeping only larger rssi
```

src/app/clusters/network-commissioning/network-commissioning.cpp
```diff
@@ -115,6 +115,225 @@ BitFlags<Feature> WiFiFeatures(WiFiDriver * driver)
...
+    ///
+    /// Returns the valid list of scan responses into `validResponses`
```
bzbarsky-apple, line 227:
```suggestion
    /// Returns the valid list of scan responses into `validResponses`, which is only valid
    /// as long as scanResponseArray is valid.
```

src/app/clusters/network-commissioning/network-commissioning.cpp
```diff
@@ -115,6 +115,225 @@ BitFlags<Feature> WiFiFeatures(WiFiDriver * driver)
...
+        }
+
+        if (scanResponseArrayLength < kMaxNetworksInScanResponse)
+        {
+            scanResponseArrayLength++;
+        }
+        scanResponseArray[scanResponseArrayLength - 1] = scanResponse;
+
+        // TODO: this is a sort (insertion sort even, so O(n^2)) in a O(n) loop,
+        ///      maybe we should find something better than a O(n^3) to preserve the highest rssi only
```
bzbarsky-apple, line 278: We're preserving the "N highest", not the "highest".  This comment is very confusing.

src/app/clusters/network-commissioning/network-commissioning.cpp
andy31415: Removed ... I am not sure why I had added it in the first place. We encode so lifetime should not matter.

src/app/clusters/network-commissioning/network-commissioning.cpp
andy31415: Updated ... I had copied from where it is used, not sure how I got the thread bit. I blame autocomplete in the IDE.

src/app/clusters/network-commissioning/network-commissioning.cpp
andy31415: Rephrased ... generally we have to find what we want to do and not use a O(n^3) loop to do that.

bzbarsky-apple (approved):

PR #33690, opened 2024-05-31
[chip-tool][ICD] Store ICDClientInfo when running icd registration independently

yunhanw-google:


examples/chip-tool/commands/clusters/ModelCommand.cpp
```diff
@@ -85,6 +85,22 @@ void ModelCommand::ClearICDEntry(const chip::ScopedNodeId & nodeId)
...
+    {
+        err = CHIPCommand::sICDClientStorage.StoreEntry(clientInfo);
+    }
+
+    if (err != CHIP_NO_ERROR)
+    {
+        CHIPCommand::sICDClientStorage.RemoveKey(clientInfo);
+        ChipLogError(chipTool, "Failed to persist symmetric key with error: %" CHIP_ERROR_FORMAT, err.Format());
+        return;
+    }
```
mkardous-silabs, line 101: this logic can be simplified to only have 1 conditionnal block. either with an if else or with if(...return)...

examples/chip-tool/commands/clusters/ClusterCommand.h
```diff
@@ -117,7 +127,28 @@ class ClusterCommand : public InteractionModelCommands, public ModelCommand, pub
...
+                    ChipLogError(chipTool, "Failed to decode ICD counter: %" CHIP_ERROR_FORMAT, err.Format());
+                    return;
+                }
+
+                chip::app::ICDClientInfo clientInfo;
+                clientInfo.peer_node         = mScopedNodeId;
+                clientInfo.monitored_subject = mMonitoredSubject;
+                clientInfo.start_icd_counter = value.ICDCounter;
+
+                ModelCommand::AddICDEntry(clientInfo, chip::ByteSpan(mICDSymmetricKey));
```
bzbarsky-apple, line 148: Why does this need the `ModelCommand` prefix?

examples/chip-tool/commands/clusters/ModelCommand.cpp
```diff
@@ -85,6 +85,22 @@ void ModelCommand::ClearICDEntry(const chip::ScopedNodeId & nodeId)
     }
 }
 
+void ModelCommand::AddICDEntry(chip::app::ICDClientInfo & clientInfo, chip::ByteSpan key)
```
bzbarsky-apple, line 88: Why is that first arg non-const?

examples/chip-tool/commands/clusters/ModelCommand.h
```diff
@@ -69,6 +69,7 @@ class ModelCommand : public CHIPCommand
 
     virtual void ClearICDEntry(const chip::ScopedNodeId & nodeId);
 
+    virtual void AddICDEntry(chip::app::ICDClientInfo & clientinfo, chip::ByteSpan key);
```
bzbarsky-apple, line 72: Why is this virtual?  It doesn't need to be virtual.

It looks like ClearICDEntry is also incorrectly virtual....

bzbarsky-apple (changes requested):

examples/chip-tool/commands/clusters/ModelCommand.cpp
yunhanw-google: sICDClientStorage.SetKey is happening inside this function,
just rename this function to StoreICDEntryWithKey to reflect this. thanks

examples/chip-tool/commands/clusters/ClusterCommand.h
yunhanw-google: removed, thanks

examples/chip-tool/commands/clusters/ModelCommand.h
yunhanw-google: removed, thanks

examples/chip-tool/commands/clusters/ModelCommand.cpp
yunhanw-google: we need check the error from CHIPCommand::sICDClientStorage.StoreEntry, and we still need two blocks?

bzbarsky-apple (approved):

joonhaengHeo (approved):

PR #33692, opened 2024-05-31
Make BLE UUID string constants lowercase

andy31415:
I  was failing to provision ble-wifi from linux after #33649. It seems that turning string constants to lowercase makes it work (previous code had it lowercase).

I could not really find documentation in bluez about the case requirement though. All I see is that:

```cpp
    GAutoPtr<GVariant> dataValue(g_variant_lookup_value(serviceData, Ble::CHIP_BLE_SERVICE_LONG_UUID_STR, nullptr));
    VerifyOrReturnError(dataValue != nullptr, false);
```

fails with uppercase and succeeds with lowercase, so it seems to me that service data is a map with lowercase keys.

### DIFF
--- a/src/ble/BleUUID.h
+++ b/src/ble/BleUUID.h
@@ -98,7 +98,7 @@ constexpr std::pair<bool, ChipBleUUID> StringToUUID(const char (&str)[N])
 // NOTE: lower-case string seems to be required at least by bluez when
 //       executing g_variant_lookup_value
 //
-//       BlueZ API https://git.kernel.org/pub/scm/bluetooth/bluez.git/tree/doc/org.bluez.Device.rst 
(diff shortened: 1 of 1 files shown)


harimau-qirex (approved):

yunhanw-google (approved):

jlatusek (approved):

arkq (approved):
Maybe this should be quoted somehow: https://datatracker.ietf.org/doc/html/rfc4122#section-3:

> Each field is treated as an integer and has its value printed as a
> zero-filled hexadecimal digit string with the most significant
> digit first.  The hexadecimal values "a" through "f" are output as
> lower case characters and are case insensitive on input.

EDIT: UUID in the docstring on line 58 should also be converted to lowercase.

tehampson (approved):

PR #33693, opened 2024-05-31
Use EncodableToTLV in im_responder

andy31415:
This makes the API cleaner and safer (have snapshot & rollback capabilities, less use of prepare/finish).

Goal is to start treating PrepareCommand/GetCommandDataIBTLVWriter/FinishCommand as private APIs.

### DIFF
--- a/src/app/tests/integration/chip_im_responder.cpp
+++ b/src/app/tests/integration/chip_im_responder.cpp
@@ -52,7 +52,7 @@ namespace app {
 
 namespace {
 
-class TestTLVDataEncoder : public chip::app::DataModel::EncodableToTLV
+class TestTLVDataEncoder : public DataModel::EncodableToTLV
 {
 public:
     CHIP_ERROR EncodeTo(TLV::TLVWriter & writer, TLV::Tag tag) const override
--- a/src/app/tests/integration/chip_im_responder.cpp
+++ b/src/app/tests/integration/chip_im_responder.cpp
@@ -50,6 +50,25 @@
 namespace chip {
 namespace app {
 
+namespace {
+
+class TestTLVDataEncoder : public chip::app::DataModel::EncodableToTLV
+{
+public:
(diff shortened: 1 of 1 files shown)


yunhanw-google (approved):

src/app/tests/integration/chip_im_responder.cpp
```diff
@@ -50,6 +50,25 @@
 namespace chip {
 namespace app {
 
+namespace {
+
+class TestTLVDataEncoder : public chip::app::DataModel::EncodableToTLV
```
bzbarsky-apple, line 55:
```suggestion
class TestTLVDataEncoder : public DataModel::EncodableToTLV
```
This is all inside chip::app.

src/app/tests/integration/chip_im_responder.cpp
```diff
@@ -104,17 +123,8 @@ void DispatchSingleClusterCommand(const ConcreteCommandPath & aRequestCommandPat
...
-        ReturnOnFailure(apCommandObj->PrepareInvokeResponseCommand(path, prepareParams));
-
-        writer = apCommandObj->GetCommandDataIBTLVWriter();
-        ReturnOnFailure(writer->Put(chip::TLV::ContextTag(kTestFieldId1), kTestFieldValue1));
-
-        ReturnOnFailure(writer->Put(chip::TLV::ContextTag(kTestFieldId2), kTestFieldValue2));
-
-        ReturnOnFailure(apCommandObj->FinishCommand());
+        TestTLVDataEncoder testData;
+        apCommandObj->AddResponse(path, kTestCommandId, testData);
```
bzbarsky-apple, line 127: So here's a question....  AddResponse will always need to take a thing that encodes a struct start/end, no?  Seems like we should consider a `StructEncodableToTLV` thing which has a single impl of those and then calls some other virtual method it calls that can be overridden to encode the struct fields...  Not part of this PR, but both this and the network commissioning bits could use it.

bzbarsky-apple (approved):

PR #33694, opened 2024-05-31
Use EncodableToTLV in TestCommandInteraction

andy31415:
This replaces some usages of direct prepare/GetIBTLVWriter/FinishCommand on CommandHandler.

I only kept a single instance of this where the direct API is being tested without it being invoked through a CommandHandler dispatch.

Intent is to treat these functions as more internal API as using AddStatus/AddResponse (and equivalent) seem to be the public API that we want everyone to use since those include snapsho/rollback/retry capability.

### DIFF
--- a/src/app/tests/TestCommandInteraction.cpp
+++ b/src/app/tests/TestCommandInteraction.cpp
@@ -88,7 +88,7 @@ constexpr CommandId kTestNonExistCommandId                = 0;
 
 const app::CommandSender::TestOnlyMarker kCommandSenderTestOnlyMarker;
 
-class SimpleTLVPayload : public chip::app::DataModel::EncodableToTLV
+class SimpleTLVPayload : public app::DataModel::EncodableToTLV
 {
 public:
     CHIP_ERROR EncodeTo(TLV::TLVWriter & writer, TLV::Tag tag) const override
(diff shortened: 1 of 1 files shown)


src/app/tests/TestCommandInteraction.cpp
```diff
@@ -87,6 +87,19 @@ constexpr CommandId kTestCommandIdFillResponseMessage     = 7;
 constexpr CommandId kTestNonExistCommandId                = 0;
 
 const app::CommandSender::TestOnlyMarker kCommandSenderTestOnlyMarker;
+
+class SimpleTLVPayload : public chip::app::DataModel::EncodableToTLV
```
bzbarsky-apple, line 91:
```suggestion
class SimpleTLVPayload : public app::DataModel::EncodableToTLV
```
This is inside namespace chip.

src/app/tests/TestCommandInteraction.cpp
```diff
@@ -87,6 +87,19 @@ constexpr CommandId kTestCommandIdFillResponseMessage     = 7;
...
+{
+public:
+    CHIP_ERROR EncodeTo(TLV::TLVWriter & writer, TLV::Tag tag) const override
+    {
+        TLV::TLVType outerType;
+        ReturnErrorOnFailure(writer.StartContainer(tag, TLV::kTLVType_Structure, outerType));
```
bzbarsky-apple, line 97: See https://github.com/project-chip/connectedhomeip/pull/33693#discussion_r1623072622 for a suggestion for this part, to do as a followup.

bzbarsky-apple (approved):

arkq (approved):

yufengwangca (approved):

PR #33701, opened 2024-06-03
Mark AutoRelockTime attribute of the Door Lock cluster as optional in the XML

Thirsrin:
Description:

AutoRelockTime attribute is optional attribute for doorlock cluster(doorlock.xml) as per matter specs(sec 5.2.9).


bzbarsky-apple:
So this is not doing a "sync" with the spec like the PR subject says, it's just marking the one attribute as optional?

src/app/zap-templates/zcl/data-model/chip/door-lock-cluster.xml
bzbarsky-apple, line 1: The change here is fine, but the commit message needs to be fixed to describe what's actually changing.  This is not a general "make door lock XML follow the spec" change, yes?

jmartinez-silabs (approved):

bzbarsky-apple (approved):

andy31415:
@Thirsrin this is a backwards-incompatible change (API change). may need a bit more review if/why this would be safe.

andy31415 (dismissed):
Marking this as hold a bit, until we are able to discuss (likely offline from this PR) about sideffects of API changes.

What we wish we had is full API stability, however in practice this is hard if the spec says one thing and the SDK implements something else.

Looking to figure out what to do here.

PR #33715, opened 2024-06-03
Fix advertised device type for all-clusters-app to match data model.

### LINKED ISSUE #33697: [Test Failed] TestDiscovery expects announcement for device type 0xffff which is no valid MEI
### Test issue(s)

I know that TestDiscovery.yaml is a chip internal test but it expects that the announced device type is 0xffff which is not any valid MEI definition for a Device type according to chip specs. The highest allowed non-manufacturerspecific devicetype should be 0xbfff (aka 49151).

Even tho the chip test applications are just reference implementations I think they should not use out of range data and formally invalid IDs

### Platform

_No response_

### Anything else?

_No response_

bzbarsky-apple:
We were claiming a device type of 0x0101 in the Descriptor DeviceTypeList on endpoint 1, but claiming a device type of 0xFFFF (not even a valid value) via DNS-SD advertising.  Align our advertising behavior with the data model values.

Fixes https://github.com/project-chip/connectedhomeip/issues/33697

### DIFF
--- a/examples/all-clusters-app/linux/include/CHIPProjectAppConfig.h
+++ b/examples/all-clusters-app/linux/include/CHIPProjectAppConfig.h
@@ -47,3 +47,7 @@
 #define CHIP_CONFIG_MAX_PATHS_PER_INVOKE 5
 
 #define CHIP_CONFIG_ENABLE_BDX_LOG_TRANSFER 1
+
+// Claim a device type while advertising that matches the device type on our
+// endpoint 1.
+#define CHIP_DEVICE_CONFIG_DEVICE_TYPE 0x0101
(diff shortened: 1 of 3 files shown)


andy31415 (approved):

examples/all-clusters-app/linux/include/CHIPProjectAppConfig.h
```diff
@@ -47,3 +47,7 @@
 #define CHIP_CONFIG_MAX_PATHS_PER_INVOKE 5
 
 #define CHIP_CONFIG_ENABLE_BDX_LOG_TRANSFER 1
+
+// Claim a device type while advertising that matches the device type on our
```
andy31415, line 51: should we say that 0x0101 is "dimmable light" ?

bzbarsky-apple: Yeah, I can do that.

Apollon77:
I know it does not count, but LGTM :-)

Apollon77 (approved):

fessehaeve (approved):

PR #33716, opened 2024-06-03
Streamline bloat reports, part 2 of 2

kpschoedel:
Remove the early attempt to aggregate FLASH vs RAM based on whether ELF segments are writable or not, which turned out not to be useful because several platforms mark flash segments as writable. Now that explicit section groups are in use (#33642), there is no additional value to segment aggregation even on platforms that mark them accurately.


kpschoedel:
Do we still want the individual sections in the reports, or _only_ the FLASH and RAM summaries?

andy31415 (approved):

arkq (approved):

PR #33719, opened 2024-06-03
Marking signed types as signed

paulr34:
Marking signed types as signed. This modification enables ZAP to accurately consume and interpret this data.


src/app/zap-templates/zcl/data-model/chip/chip-types.xml
bzbarsky-apple, line 1: 1. Commit message should talk about how signed types should be marked as signed.

2. PR description needs to not claim that this fixes an issue it does not (on its own) fix (and PR must not be linked to that issue).

bzbarsky-apple (changes requested):

bzbarsky-apple (approved):

lpbeliveau (approved):

PR #33721, opened 2024-06-03
Linux: Tidy up ifdefs around app options

ksperling-apple:
Use consistent ifdef conditions between member definitions, option definition, option handling code, and help text.


cecille (approved):

examples/platform/linux/Options.cpp
```diff
@@ -90,7 +91,6 @@ enum
     kOptionCSRResponseAttestationSignatureInvalid       = 0x101d,
     kOptionCSRResponseCSRExistingKeyPair                = 0x101e,
     kDeviceOption_TestEventTriggerEnableKey             = 0x101f,
-    kCommissionerOption_FabricID                        = 0x1020,
```
arkq, line 93: Since these values grows downward, maybe instead of moving the option simply wrap it with `#if`? If all these options were mixed it would be hard to track which value is next.

arkq (approved):

examples/platform/linux/Options.cpp
ksperling-apple: I could also just renumber it? There's no external meaning to these numbers right?

arkq: Renumbering should also be OK, I think.

ksperling-apple: There doesn't seem any reason for these to be manually allocated in the first place, so I've changed it to start at `0x1000` and let the compiler handle the rest.

PR #33729, opened 2024-06-04
Setup bloat reports for Tizen CI builds

arkq:
This PR adds blot reports for Tizen examples built in CI on GitHub Actions. Since Tizen is very similar to Linux, no Tizen-specific configuration was created, instead, bloat report uses Linux configuration.

### DIFF
--- a/.github/workflows/examples-tizen.yaml
+++ b/.github/workflows/examples-tizen.yaml
@@ -71,12 +71,12 @@ jobs:
             - name: Bloat report - chip-tool
               run: |
                   .environment/pigweed-venv/bin/python3 scripts/tools/memory/gh_sizes.py \
-                    linux arm chip-tool-ubsan out/tizen-arm-chip-tool-ubsan/chip-tool \
+                    tizen arm chip-tool-ubsan out/tizen-arm-chip-tool-ubsan/chip-tool \
                     /tmp/bloat_reports/
             - name: Bloat report - all-clusters-app
               run: |
                   .environment/pigweed-venv/bin/python3 scripts/tools/memory/gh_sizes.py \
-                    linux arm all-clusters-app out/tizen-arm-all-clusters/chip-all-clusters-app \
+                    tizen arm all-clusters-app out/tizen-arm-all-clusters/chip-all-clusters-app \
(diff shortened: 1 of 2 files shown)


.github/workflows/examples-tizen.yaml
```diff
@@ -61,3 +67,20 @@ jobs:
                           build \
                           --copy-artifacts-to out/artifacts \
                       "
+
+            - name: Bloat report - chip-tool
+              run: |
+                  .environment/pigweed-venv/bin/python3 scripts/tools/memory/gh_sizes.py \
+                    linux arm chip-tool-ubsan out/tizen-arm-chip-tool-ubsan/chip-tool \
```
andy31415, line 74: should we have tizen somewhere in the platforms? Linux seems to clash with the "ubuntu" builds of things.

I believe these triplets are used to uniquely identify things for reports. 

```suggestion
                    tizen arm chip-tool-ubsan out/tizen-arm-chip-tool-ubsan/chip-tool \
```

arkq: I think that this is for the configuration which should be selected, but it might be also for some ID... So, maybe you are right that it would be safer to use platform specific name here.

arkq:
@kpschoedel, since it seems that you know exactly how the bloat check works, do I need to create explicit configuration for Tizen in `scripts/tools/memory/platform/tizen.cfg`, or using default should be good enough?

.github/workflows/examples-tizen.yaml
kpschoedel: The first three arguments correspond to the **platform**, **config**, and **target** headings in the comment tables. The latter two are used to match up corresponding runs, but the first (platform) _also_ identifies a config file under `scripts/tools/memory/platform`. You can probably just copy `linux.cfg` to `tizen.cfg`. For the GitHub bloat reports now, the important part is defining the sections that go into the FLASH and RAM totals.

kpschoedel:
Yes (we just crossed paths) — you should be able to just copy `linux.cfg` to `tizen.cfg` under `scripts/tools/memory/platform/`.

kpschoedel (approved):

PR #33730, opened 2024-06-04
Post-merge updates for IM/DM separation

andy31415:
Going through review comments post-submit in #32914

### Changes:
- several renames and comment updates

Not addressed:
  - renaming "interactionModel" to "dataModel" (larger change, should probably be stand-alone)
  - InvokeRequest not touched as I am moving towards using CommandHandler, in which case all this code is obsolete

### DIFF
--- a/src/app/interaction-model/Events.h
+++ b/src/app/interaction-model/Events.h
@@ -107,7 +107,7 @@ std::optional<EventNumber> GenerateEvent(G & generator, const T & aEventData, En
 /// Exposes event access capabilities.
 ///
 /// Allows callers to "generate events" which effectively notifies of an event having
-/// ocurred. 
+/// ocurred.
 class Events
 {
 public:
--- a/src/app/interaction-model/Events.h
+++ b/src/app/interaction-model/Events.h
@@ -33,10 +33,10 @@ namespace InteractionModel {
 
 namespace internal {
 template <typename T>
-class SimpleEventLoggingDelegate : public EventLoggingDelegate
+class SimpleEventPayloadWriter : public EventLoggingDelegate
 {
 public:
-    SimpleEventLoggingDelegate(const T & aEventData) : mEventData(aEventData){};
+    SimpleEventPayloadWriter(const T & aEventData) : mEventData(aEventData){};
     CHIP_ERROR WriteEvent(chip::TLV::TLVWriter & aWriter) final override
     {
         return DataModel::Encode(aWriter, TLV::ContextTag(EventDataIB::Tag::kData), mEventData);
@@ -49,7 +49,7 @@ private:
 template <typename G, typename T, std::enable_if_t<DataModel::IsFabricScoped<T>::value, bool> = true>
 std::optional<EventNumber> GenerateEvent(G & generator, const T & aEventData, EndpointId aEndpoint)
 {
-    internal::SimpleEventLoggingDelegate<T> eventPayloadWriter(aEventData);
+    internal::SimpleEventPayloadWriter<T> eventPayloadWriter(aEventData);
     ConcreteEventPath path(aEndpoint, aEventData.GetClusterId(), aEventData.GetEventId());
     EventOptions eventOptions;
     eventOptions.mPath        = path;
@@ -86,7 +86,7 @@ std::optional<EventNumber> GenerateEvent(G & generator, const T & aEventData, En
 template <typename G, typename T, std::enable_if_t<!DataModel::IsFabricScoped<T>::value, bool> = true>
 std::optional<EventNumber> GenerateEvent(G & generator, const T & aEventData, EndpointId endpointId)
 {
-    internal::SimpleEventLoggingDelegate<T> eventPayloadWriter(aEventData);
+    internal::SimpleEventPayloadWriter<T> eventPayloadWriter(aEventData);
     ConcreteEventPath path(endpointId, aEventData.GetClusterId(), aEventData.GetEventId());
     EventOptions eventOptions;
     eventOptions.mPath     = path;
--- a/src/app/interaction-model/Events.h
+++ b/src/app/interaction-model/Events.h
@@ -25,6 +25,7 @@
 #include <lib/support/logging/CHIPLogging.h>
 
 #include <type_traits>
+#include <optional>
 
 namespace chip {
 namespace app {
@@ -45,8 +46,8 @@ private:
     const T & mEventData;
 };
 
-template <typename E, typename T, std::enable_if_t<DataModel::IsFabricScoped<T>::value, bool> = true>
-EventNumber GenerateEvent(E & emittor, const T & aEventData, EndpointId aEndpoint)
+template <typename G, typename T, std::enable_if_t<DataModel::IsFabricScoped<T>::value, bool> = true>
+std::optional<EventNumber> GenerateEvent(G & generator, const T & aEventData, EndpointId aEndpoint)
 {
     internal::SimpleEventLoggingDelegate<T> eventData(aEventData);
     ConcreteEventPath path(aEndpoint, aEventData.GetClusterId(), aEventData.GetEventId());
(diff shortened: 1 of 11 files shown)


src/app/interaction-model/Events.h
```diff
@@ -45,22 +46,22 @@ class SimpleEventLoggingDelegate : public EventLoggingDelegate
...
     ConcreteEventPath path(aEndpoint, aEventData.GetClusterId(), aEventData.GetEventId());
     EventOptions eventOptions;
     eventOptions.mPath        = path;
     eventOptions.mPriority    = aEventData.GetPriorityLevel();
     eventOptions.mFabricIndex = aEventData.GetFabricIndex();
 
-    // this skips logging the event if it's fabric-scoped but no fabric association exists yet.
-
+    // this skips generating the event if it is fabric-scoped however the event does not seem
+    // associated with any fabric.
```
bzbarsky-apple, line 60:
```suggestion
    // this skips generating the event if it is fabric-scoped but the provided event data is not
    // associated with any fabric.
```

src/app/interaction-model/Events.h
```diff
@@ -72,37 +73,41 @@ EventNumber GenerateEvent(E & emittor, const T & aEventData, EndpointId aEndpoin
...
     return eventNumber;
 }
 
 } // namespace internal
 
+/// Exposes event access capabilities.
+///
+/// Allows callers to "generate events" which effectively notifies of an event having
+/// ocurred.
 class Events
```
bzbarsky-apple, line 111: I think we should consider calling this EventGenerator until we need to do something else.  Then we can think about renaming, but we will have a better idea for the name based on what we actually want it to do.

src/app/interaction-model/InvokeResponder.h
bzbarsky-apple, line 1: This seems to be missing fixes for a bunch of the comments from https://github.com/project-chip/connectedhomeip/pull/32914

src/app/interaction-model/IterationTypes.h
```diff
@@ -70,8 +70,13 @@ struct AttributeEntry
 /// Iteration rules:
 ///   - kInvalidEndpointId will be returned when iteration ends (or generally kInvalid* for paths)
 ///   - Any internal iteration errors are just logged (callers do not handle iteration CHIP_ERROR)
```
bzbarsky-apple, line 72: This is missing some comments from https://github.com/project-chip/connectedhomeip/pull/32914

src/app/interaction-model/IterationTypes.h
```diff
@@ -70,8 +70,13 @@ struct AttributeEntry
...
-///   - Iteration order is NOT guaranteed, however uniqueness and completeness is (must iterate
-///     over all possible distinct values as long as no internal structural changes occur)
+///   - Iteration order is NOT guaranteed globally. Only the following is guaranteed:
+///     - when iterating over an endpoint, ALL clusters of that endpoint will be iterated first, before
+///       switching the endpoint (order of clusters ids themselves not guaranteed)
```
bzbarsky-apple, line 75:
```suggestion
///       switching the endpoint (order of clusters themselves not guaranteed)
```

src/app/interaction-model/Model.h
```diff
@@ -43,17 +43,17 @@ class Model : public AttributeTreeIterator
 public:
     virtual ~Model() = default;
```
bzbarsky-apple, line 44: Again: why is this called InteractionModel::Model when it's a "data model"?

src/app/interaction-model/OperationTypes.h
```diff
@@ -67,7 +67,7 @@ struct ReadState
 enum class WriteFlags : uint32_t
 {
     kTimed     = 0x0001, // Received as a 2nd command after a timed invoke
```
bzbarsky-apple, line 69: Unaddressed issue here.

src/app/interaction-model/OperationTypes.h
```diff
@@ -67,7 +67,7 @@ struct ReadState
...
-    kListBegin = 0x0002, // This is the FIRST list data element in a series of data
+    kListBegin = 0x0002, // This is the FIRST list of data elements
     kListEnd   = 0x0004, // This is the LAST list element to write
```
bzbarsky-apple, line 71: Unaddressed issue here.

src/app/interaction-model/Paths.h
```diff
@@ -22,23 +22,21 @@ namespace chip {
 namespace app {
 namespace InteractionModel {
 
-/// Handles path attributes for interaction models.
+/// Notification listener for attribute changes.
```
bzbarsky-apple, line 25: If it's for "attribute changes", why is it not called "Attributes"?

src/app/interaction-model/Paths.h
```diff
@@ -22,23 +22,21 @@ namespace chip {
...
 /// main loop as they will likely trigger interaction model
-/// internal updates and subscription event updates.
+/// internal updates and subscription data reporting.
 class Paths
 {
 public:
     virtual ~Paths() = 0;
 
-    /// Mark some specific attributes dirty.
+    /// Mark some specific attribute dirty (or several of attributes when using wildcards).
```
bzbarsky-apple, line 38:
```suggestion
    /// Mark all attributes matching the given path (which may be a wildcard) dirty.
```

bzbarsky-apple (commented):
I resolved the comments on https://github.com/project-chip/connectedhomeip/pull/32914 that are fixed by this PR so far, but a bunch are still unresolved.

src/app/interaction-model/Model.h
andy31415: I would like to do renames as part of a separate PR that just renames. It seems to need file and directory shuffling and combining that with documentation changes is rough.

src/app/interaction-model/InvokeResponder.h
andy31415: I kept unaddressed comments open - I would like to do a few passes on this rather than one large CL because one large CL takes too much time for review.

andy31415: Overall I did not want to really touch InvokeResponder as part of this update because I will switch it to use CommandHandler and at that point the API would be finalized. Me guessing what is needed in advance seems not as effective.

src/app/interaction-model/OperationTypes.h
andy31415: I did not address because for a drop-in repleacement for ember-compatibility-functions we have https://github.com/project-chip/connectedhomeip/blob/master/src/app/util/ember-compatibility-functions.cpp#L764

It is also true that as we update the code we may choose to update some requirements (e.g. timed write can be decided by IM based on attribute flags). However the start is a drop-replacement.

src/app/interaction-model/IterationTypes.h
andy31415: Updated in #33345

Because that PR also does renaming, it starts being hard to both rename and update comments.

src/app/interaction-model/OperationTypes.h
andy31415: Removed lists as part of #33345. Final cleanup (e.g. figuring out proper responses like the extra CHIP_ERROR) should be done when implementing the individual actions and/or at a final pass before looking to start using these interfaces.

src/app/interaction-model/Paths.h
andy31415: We need some better name. i think bare `Attributes` is insufficient. Probably some `AttributeChangeListener` or observer or similar would be better. I would like to make renaming as a subsequent pass.

src/app/interaction-model/Events.h
andy31415: I would like to do renames after I have this and #33345 (or just 33345 because that includes all these changes) merged. Managing both renames and code updates as part of the same git branch chains and PRs is very hard.

bzbarsky-apple: That's fine, please just don't lose track of the rename request.

src/app/interaction-model/OperationTypes.h
bzbarsky-apple: My point was that this is _not an invoke_ and _not a command_.  The documentation just makes no sense.

src/app/interaction-model/Paths.h
bzbarsky-apple: OK, please just don't lose track of it.

bzbarsky-apple (approved):
Approving assuming the remaining issues will get fixed in followups.  But please don't lose track of them.  I do not have the bandwidth to correlate all the comments across 4 different PRs and make sure nothing got missed...

src/app/interaction-model/OperationTypes.h
andy31415: I totally missed what you meant. Updated the comment.

andy31415:
> Approving assuming the remaining issues will get fixed in followups. But please don't lose track of them. I do not have the bandwidth to correlate all the comments across 4 different PRs and make sure nothing got missed...

I will do my best, but obviously no guarantees. What I also expect is that we will have a final pass before we turn this on to replace existing emeber-compatibility-functions so we will have another chance to sanity-review then as a whole.

Thank you for the deep reviews so far!

robszewczyk (approved):

yunhanw-google (approved):

PR #33737, opened 2024-06-04
Fix Darwin server cluster checks for wildcard read to check received values.

bzbarsky-apple:
This caught a bug where the synthesized descriptor clusters were not spec-compliant in the sense of requiring admin privileges to read their attributes, and the synthesized global attributes likewise required admin privileges.  The changes to MTRServerAccessControl fix that.


jtung-apple (approved):

bzbarsky-apple:

PR #33740, opened 2024-06-04
Python test runner: Add non-printing CI option

cecille:
The CI files for the REPL tests are HUGE and they make it nearly
impossible to find the error you want. This PR adds a CI flag
where the output is only sent in case of a test failure.

If you look at the earlier runs of the REPL on this pr (here: https://github.com/project-chip/connectedhomeip/actions/runs/9375514343/job/25813769232?pr=33740), I have an example of an intentional failure. Error is visible from the main screen.


scripts/tests/run_python_test.py
```diff
@@ -87,7 +88,8 @@ def DumpProgramOutputToQueue(thread_list: typing.List[threading.Thread], tag: st
               help='Script arguments, can use placeholders like {SCRIPT_BASE_NAME}.')
 @click.option("--script-gdb", is_flag=True,
               help='Run script through gdb')
-def main(app: str, factoryreset: bool, factoryreset_app_only: bool, app_args: str, script: str, script_args: str, script_gdb: bool):
+@click.option("--ci", is_flag=True, help="Use this flag in CI to suppress output from passing tests")
```
andy31415, line 91: let's use some different name then "CI". If we want to suppress output from passing tests, how about "silent" ?

cecille: changed to "quiet" because it's not strictly silent. Updated the msg to match.

andy31415 (approved):

PR #33757, opened 2024-06-05
pigweed unit_test migration: App 3rd batch

Alami-Amine:
- Continuation of migration from nl_test to pw_unit_test, effort first discussed in issue https://github.com/project-chip/connectedhomeip/pull/29342
- 3rd batch of src/app tests (related to Interaction Model) first started in https://github.com/project-chip/connectedhomeip/pull/33638

- `TestClusterInfo` was removed and its tests were split between `TestAttributePathParams` and `TestEventPathParams`


src/app/tests/TestExtensionFieldSets.cpp
```diff
@@ -68,118 +76,118 @@ void TestInsertExtensionFieldSet(nlTestSuite * aSuite, void * aContext)
 
     memset(double_size_buffer, static_cast<uint8_t>(1), sizeof(double_size_buffer));
 
-    NL_TEST_ASSERT(aSuite, true == EFS->IsEmpty());
+    EXPECT_EQ(true, EFS->IsEmpty());
```
feasel0, line 79:
```suggestion
    EXPECT_TRUE(EFS->IsEmpty());
```

src/app/tests/TestExtensionFieldSets.cpp
```diff
@@ -68,118 +76,118 @@ void TestInsertExtensionFieldSet(nlTestSuite * aSuite, void * aContext)
...
-    NL_TEST_ASSERT(aSuite, CHIP_NO_ERROR == testEFS3.InsertFieldSet(EFS2));
+    EXPECT_EQ(CHIP_NO_ERROR, testEFS3.InsertFieldSet(EFS1));
+    EXPECT_EQ(CHIP_NO_ERROR, testEFS3.InsertFieldSet(EFS2));
 
     tempTestEFS = testEFS1;
-    NL_TEST_ASSERT(aSuite, tempTestEFS == testEFS1);
-    NL_TEST_ASSERT(aSuite, !(tempTestEFS == testEFS2));
-    NL_TEST_ASSERT(aSuite, !(tempTestEFS == testEFS3));
+    EXPECT_EQ(tempTestEFS, testEFS1);
+    EXPECT_FALSE(tempTestEFS == testEFS2);
```
feasel0, line 153: Will EXPECT_NE work here?

Alami-Amine: no it won't because the operator != is not overloaded for that class

src/app/tests/TestExtensionFieldSets.cpp
```diff
@@ -203,148 +211,105 @@ void TestSerializeDerializeExtensionFieldSet(nlTestSuite * aSuite, void * aConte
...
+    EXPECT_EQ(tempEFS, EFS2);
+    EXPECT_EQ(CHIP_NO_ERROR, EFS->GetFieldSetAtPosition(tempEFS, 1));
+    EXPECT_EQ(tempEFS, EFS1);
+    EXPECT_EQ(CHIP_NO_ERROR, EFS->GetFieldSetAtPosition(tempEFS, 2));
+    EXPECT_EQ(tempEFS, EFS3);
 
     // Emptying the table
     EFS->Clear();
-    NL_TEST_ASSERT(aSuite, true == EFS->IsEmpty());
+    EXPECT_EQ(true, EFS->IsEmpty());
```
feasel0, line 312:
```suggestion
    EXPECT_TRUE(EFS->IsEmpty());
```

src/app/tests/TestSceneTable.cpp
```diff
@@ -437,10 +477,10 @@ void ResetSceneTable(SceneTable * sceneTable)
     sceneTable->RemoveFabric(kFabric3);
 }
 
-void TestHandlerRegistration(nlTestSuite * aSuite, void * aContext)
+TEST_F(TestSceneTable, TestHandlerRegistration)
 {
     SceneTable * sceneTable = scenes::GetSceneTableImpl(kTestEndpoint1, defaultTestTableSize);
-    NL_TEST_ASSERT(aSuite, nullptr != sceneTable);
+    EXPECT_NE(nullptr, sceneTable);
     VerifyOrReturn(nullptr != sceneTable);
```
feasel0, line 484: Remove VerifyOrReturn and change the previous line to ASSERT

src/app/tests/TestSceneTable.cpp
```diff
@@ -462,38 +502,38 @@ void TestHandlerRegistration(nlTestSuite * aSuite, void * aContext)
...
+    EXPECT_TRUE(sceneTable->HandlerListEmpty());
 }
 
-void TestHandlerFunctions(nlTestSuite * aSuite, void * aContext)
+TEST_F(TestSceneTable, TestHandlerFunctions)
 {
     SceneTable * sceneTable = scenes::GetSceneTableImpl(kTestEndpoint1, defaultTestTableSize);
-    NL_TEST_ASSERT(aSuite, nullptr != sceneTable);
+    EXPECT_NE(nullptr, sceneTable);
     VerifyOrReturn(nullptr != sceneTable);
```
feasel0, line 537: Remove VerifyOrReturn and change the previous line to ASSERT

feasel0 (approved):

src/app/tests/TestExtensionFieldSets.cpp
andy31415: we have operator == but not != ... that is terrible!

src/app/tests/TestDataModelSerialization.cpp
```diff
@@ -16,70 +16,53 @@
...
-    static void TestDataModelSerialization_InvalidListType(nlTestSuite * apSuite, void * apContext);
-
-    static void NullablesOptionalsStruct(nlTestSuite * apSuite, void * apContext);
-    static void NullablesOptionalsCommand(nlTestSuite * apSuite, void * apContext);
-
-    void Shutdown();
+    static void SetUpTestSuite() { ASSERT_EQ(chip::Platform::MemoryInit(), CHIP_NO_ERROR); }
+    static void TearDownTestSuite()
+    {
+        Shutdown();
```
arkq, line 42: Maybe just calls `mStore.Release();` here without any "Shutdown" wrapper. Also, I wonder why `mStore` is static, is it shared between tests (if that's the case, that's somehow not ideal, because unit tests might depend on each other)?

src/app/tests/TestSceneTable.cpp
```diff
@@ -425,10 +425,50 @@ class TestSceneTableImpl : public SceneTableImpl
...
+    }
+
+    static void TearDownTestSuite()
+    {
+        // Terminate Scene Table
+        SceneTable * sceneTable = scenes::GetSceneTableImpl();
+        ASSERT_NE(sceneTable, nullptr);
+        sceneTable->Finish();
+
+        if (mpTestStorage != nullptr)
```
arkq, line 452: This `if` is not required, `delete` can be called on `nullptr`.

arkq (approved):

src/app/tests/TestExtensionFieldSets.cpp
Alami-Amine: this is the case for a lot of the classes that we tested actually!

src/app/tests/TestEventPathParams.cpp
```diff
@@ -16,65 +16,120 @@
...
-} // namespace chip
 
-namespace {
-const nlTest sTests[] = { NL_TEST_DEF("TestSamePath", chip::app::TestEventPathParams::TestSamePath),
-                          NL_TEST_DEF("TestDifferentEndpointId", chip::app::TestEventPathParams::TestDifferentEndpointId),
-                          NL_TEST_DEF("TestDifferentClusterId", chip::app::TestEventPathParams::TestDifferentClusterId),
-                          NL_TEST_DEF("TestDifferentEventId", chip::app::TestEventPathParams::TestDifferentEventId),
-                          NL_TEST_SENTINEL() };
+/*
+{kInvalidEndpointId, kInvalidClusterId, kInvalidEventId},
```
jlatusek, line 71: I think that this comment should be removed

jlatusek (approved):

jlatusek (approved):

src/app/tests/TestExtensionFieldSets.cpp
Alami-Amine: should I add it for that class?

src/app/tests/TestEventPathParams.cpp
Alami-Amine: I Think it is relevant though; makes what happened in InitEventPaths() more readable,
it shows that for example:
for validEventpaths[0] we have {kInvalidEndpointId, kInvalidClusterId, kInvalidEventId}, 
for validEventpaths[1] we have {kInvalidEndpointId, MockClusterId(1), kInvalidEventId},

src/app/tests/TestDataModelSerialization.cpp
Alami-Amine: - well in order to be able to call `mStore.Release();` in TearDownTestSuite it has to be static.
- also the reason I called Shutdown instead of directly putting `mStore.Release();` is:
     1. `mStore.Release();` has a return of `CHECK_RETURN_VALUE` so  it needs to be assigned to the `System::PacketBufferHandle  buf` object
     2. the `buf` object will be destructed when we reach the end of scope of `TearDownTestSuite()` just after `chip::Platform::MemoryShutdown();`  --> this results in an error because the destruction of `buf` will attempt to free the buffer, but since `memoryInitialized` of `CHIPMem-Malloc.cpp` is Zero due to `chip::Platform::MemoryShutdown();` , we hit a VerifyorDie and exception happens....maybe I can explicitly Free the buffer before memoryshutdown...

Alami-Amine: ok I added `buf = nullptr;` just before MemoryShutdown to avoid the error that happens

arkq: > well in order to be able to call mStore.Release(); in TearDownTestSuite it has to be static.

That's exactly the point. Is there any particular reason why it's called in the `TearDownTestSuite()` and not in the `TearDown()`. Like I've said, it smells like the buffer is shared between unit test cases, which might be not good.

Alami-Amine: done.

andy31415 (approved):

PR #33769, opened 2024-06-05
[Fabric-Bridge] Don't track device changes for synced devices

yufengwangca:
For the bridged device for fabric sync, they are not the physical devices proxy-ed by the bridge, but the Matter devices to be synced.

The bridge can not catch the changes of those devices, and there is nothing to do for those changes.

### DIFF
--- a/examples/fabric-bridge-app/linux/DeviceManager.cpp
+++ b/examples/fabric-bridge-app/linux/DeviceManager.cpp
@@ -43,8 +43,13 @@ using namespace chip::Transport;
 using namespace chip::DeviceLayer;
 using namespace chip::app::Clusters;
 
+class DeviceManager;
+
 namespace {
+
 constexpr uint8_t kMaxRetries = 10;
+DeviceManager gDeviceManager;
+
 } // namespace
 
 DeviceManager::DeviceManager()
@@ -127,4 +132,9 @@ Device * DeviceManager::GetDevice(uint16_t index) const
         return mDevices[index];
     }
     return nullptr;
+}
+
+DeviceManager * GetDeviceManager()
+{
+    return &gDeviceManager;
 }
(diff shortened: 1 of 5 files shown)


pidarped (approved):

examples/fabric-bridge-app/linux/DeviceManager.cpp
```diff
@@ -43,8 +43,13 @@ using namespace chip::Transport;
 using namespace chip::DeviceLayer;
 using namespace chip::app::Clusters;
 
+class DeviceManager;
```
andy31415, line 46: why do we need this forward declaration here?

examples/fabric-bridge-app/linux/include/DeviceManager.h
```diff
@@ -66,3 +66,12 @@ class DeviceManager
...
 };
+
+/**
+ * Instance getter for the global DeviceManager.
+ *
+ * Callers have to externally synchronize usage of this function.
+ *
+ * @return The global DeviceManager. Assume never null.
+ */
+DeviceManager * GetDeviceManager();
```
andy31415, line 77: What parts of the code do we expect to use this global?

Wondering if we could have a path to inject the device manager instead of creating singletons. I think for one-off maybe globals are convenient, however if we expect integrations of this code into other applications, then not using singletons seems better and more testable.

yufengwangca: DeviceManager is used to manage the list of devices to be synced. It is utilized by all entities where the Fabric Sync protocol is implemented, such as the RPC Client, RPC Server, and all relevant ChipCommands.

Its role is very similar to PlatformMgr, which is why I have adopted a similar access pattern to PlatformMgr.

joonhaengHeo (approved):

PR #33771, opened 2024-06-06
Update device-energy-management-cluster.xml based on latest cluster spec

PeterC1965:
Update device-energy-management-cluster.xml based on latest cluster spec. This is the first of several pull-requests that implement the DEM cluster in Matter 1.4. 

Note DEM did not make Matter 1.3 SVE and is only provisional. 

The spec can be found at: https://github.com/CHIP-Specifications/connectedhomeip-spec/blob/master/src/app_clusters/DeviceEnergyManagement.adoc 

Follow-up PRs will include adding:
1. SDK cluster support.
2. Python test cases.


src/app/clusters/device-energy-management-server/device-energy-management-server.cpp
```diff
@@ -665,15 +847,47 @@ void Instance::HandleRequestConstraintBasedForecast(HandlerContext & ctx,
...
+    {
+        status = mDelegate.CancelRequest();
+    }
 
-    status = mDelegate.CancelRequest();
     ctx.mCommandHandler.AddStatus(ctx.mRequestPath, status);
-    if (status != Status::Success)
+}
+
+Status Instance::GetCurrentUtcTime(uint32_t & currentUtcTime) const
```
jamesharrow, line 871: Maybe worth calling this GetCurrentUtcTimeEpoch or just GetCurrentTimeEpochS since I think EPOCH is by definition UTC to be clearer that it is using Matter Epoch time.

src/app/clusters/device-energy-management-server/device-energy-management-server.cpp
```diff
@@ -629,7 +667,54 @@ void Instance::HandleModifyForecastRequest(HandlerContext & ctx, const Commands:
         return;
     }
 
-    status = mDelegate.ModifyForecastRequest(forecastId, slotAdjustments, adjustmentCause);
+    // Check the various values in the slot structures
+    auto iterator = slotAdjustments.begin();
+    while (iterator.Next())
```
bzbarsky-apple, line 672: I don't see where we are checking status at the end of the list, so if the TLV in there is invalid we will happily successfully process things up to there and claim everything is good.  We need to be checking the status after the loop ends.

src/app/clusters/device-energy-management-server/device-energy-management-server.cpp
```diff
@@ -629,7 +667,54 @@ void Instance::HandleModifyForecastRequest(HandlerContext & ctx, const Commands:
...
+
+        // Check for an invalid slotIndex
+        if (slotAdjustment.slotIndex > forecast.Value().slots.size())
+        {
+            ChipLogError(Zcl, "DEM: Bad slot index %d", slotAdjustment.slotIndex);
+            ctx.mCommandHandler.AddStatus(ctx.mRequestPath, Status::ConstraintError);
+            return;
+        }
+
+        // Check to see if trying to modify a slot which have already been run
```
bzbarsky-apple, line 684:
```suggestion
        // Check to see if trying to modify a slot which has already been run
```

src/app/clusters/device-energy-management-server/device-energy-management-server.cpp
```diff
@@ -654,6 +746,96 @@ void Instance::HandleRequestConstraintBasedForecast(HandlerContext & ctx,
         return;
     }
 
+    uint32_t currentUtcTime = 0;
+    status                  = GetCurrentUtcTime(currentUtcTime);
+    if (status != Status::Success)
+    {
+        ChipLogError(Zcl, "DEM: Forecast is Null");
```
bzbarsky-apple, line 753: This log message does not look right.

src/app/clusters/device-energy-management-server/device-energy-management-server.cpp
```diff
@@ -654,6 +746,96 @@ void Instance::HandleRequestConstraintBasedForecast(HandlerContext & ctx,
...
+        ctx.mCommandHandler.AddStatus(ctx.mRequestPath, status);
+        return;
+    }
+
+    // Check for invalid power levels and whether the constraint time/duration is in the past
+    {
+        auto iterator = constraints.begin();
+        if (iterator.Next())
```
bzbarsky-apple, line 761: Again, where is the required-when-iterating-lists status check?

src/app/clusters/device-energy-management-server/device-energy-management-server.cpp
```diff
@@ -654,6 +746,96 @@ void Instance::HandleRequestConstraintBasedForecast(HandlerContext & ctx,
...
+                {
+                    ChipLogError(Zcl, "DEM: RequestConstraintBasedForecast no nominalPower");
+                    ctx.mCommandHandler.AddStatus(ctx.mRequestPath, Status::InvalidCommand);
+                    return;
+                }
+
+                if (constraint.nominalPower.Value() < mDelegate.GetAbsMinPower() ||
+                    constraint.nominalPower.Value() > mDelegate.GetAbsMaxPower())
+                {
+                    ChipLogError(Zcl, "DEM: RequestConstraintBasedForecast nominalPower out of range [absMinPower, absMaxPower]");
```
bzbarsky-apple, line 784: Is it worth logging the incoming value, or all three?

src/app/clusters/device-energy-management-server/device-energy-management-server.cpp
```diff
@@ -654,6 +746,96 @@ void Instance::HandleRequestConstraintBasedForecast(HandlerContext & ctx,
...
+                    return;
+                }
+            }
+        }
+    }
+
+    // Check for overlappping elements
+    {
+        auto iterator = constraints.begin();
+        if (iterator.Next())
```
bzbarsky-apple, line 819: Where is the status check?

src/app/clusters/device-energy-management-server/device-energy-management-server.cpp
```diff
@@ -654,6 +746,96 @@ void Instance::HandleRequestConstraintBasedForecast(HandlerContext & ctx,
...
+        {
+            // Get the first constraint
+            const Structs::ConstraintsStruct::DecodableType & prevConstraint = iterator.GetValue();
+
+            // Start comparing next vs prev constraints
+            while (iterator.Next())
```
bzbarsky-apple, line 825: Status check?

But also, after this call, `prevConstraint` has been overwritten.  So I would expect the tests below to always fail.  Was this tested?

src/app/clusters/device-energy-management-server/device-energy-management-server.cpp
```diff
@@ -654,6 +746,96 @@ void Instance::HandleRequestConstraintBasedForecast(HandlerContext & ctx,
...
+            {
+                const Structs::ConstraintsStruct::DecodableType & constraint = iterator.GetValue();
+                if (constraint.startTime < prevConstraint.startTime ||
+                    prevConstraint.startTime + prevConstraint.duration >= constraint.startTime)
```
bzbarsky-apple, line 829: This makes no sense, since we never change prevConstraint.

src/app/clusters/device-energy-management-server/device-energy-management-server.cpp
```diff
@@ -665,15 +847,47 @@ void Instance::HandleRequestConstraintBasedForecast(HandlerContext & ctx,
...
+    if (err != CHIP_NO_ERROR)
     {
-        ChipLogError(Zcl, "DEM: CancelRequest FAILURE");
-        return;
+        ChipLogError(Zcl, "DEM: Unable to get current time - err:%" CHIP_ERROR_FORMAT, err.Format());
+        return Status::Failure;
+    }
+
+    auto unixEpoch = std::chrono::duration_cast<System::Clock::Seconds32>(cTMs).count();
+    if (!UnixEpochToChipEpochTime(unixEpoch, currentUtcTime))
```
bzbarsky-apple, line 884: It's very much not obvious based on naming that this function returns Matter epoch times.  Please rename accordingly.

bzbarsky-apple (changes requested):

src/app/clusters/device-energy-management-server/device-energy-management-server.cpp
PeterC1965: I believe UnixEpochToChipEpochTime is defined in connectedhomeip/src/lib/support/TimeUtils.h and used in a few places in connectedhomeip (see below) -> do you still want me to rename it?

connectedhomeip/examples/chef/common/clusters/door-lock/chef-lock-endpoint.cpp
connectedhomeip/examples/lock-app/lock-common/src/LockEndpoint.cpp
connectedhomeip/src/credentials/CHIPCertificateSet.h
connectedhomeip/src/tools/chip-cert/Cmd_ValidateCert.cpp

If I do rename it, would GetMatterEpochTimeFromUnixTime() be ok?

src/app/clusters/device-energy-management-server/device-energy-management-server.cpp
PeterC1965: Function renamed.

src/app/clusters/device-energy-management-server/device-energy-management-server.cpp
PeterC1965: I have added a check.

src/app/clusters/device-energy-management-server/device-energy-management-server.cpp
PeterC1965: Changed.

src/app/clusters/device-energy-management-server/device-energy-management-server.cpp
PeterC1965: I have changed the message accordingly.

src/app/clusters/device-energy-management-server/device-energy-management-server.cpp
PeterC1965: I have added a check.

src/app/clusters/device-energy-management-server/device-energy-management-server.cpp
PeterC1965: I have added a check.

src/app/clusters/device-energy-management-server/device-energy-management-server.cpp
PeterC1965: See above comment.

src/app/clusters/device-energy-management-server/device-energy-management-server.cpp
bzbarsky-apple: UnixEpochToChipEpochTime is fine as far as it goes.  I meant the function here, which calls that and returns a Matter epoch time, but does not make that clear.

src/app/clusters/device-energy-management-server/device-energy-management-server.cpp
```diff
@@ -654,6 +752,114 @@ void Instance::HandleRequestConstraintBasedForecast(HandlerContext & ctx,
...
+                if (!constraint.loadControl.HasValue())
+                {
+                    ChipLogError(Zcl, "DEM: RequestConstraintBasedForecast no loadControl");
+                    ctx.mCommandHandler.AddStatus(ctx.mRequestPath, Status::InvalidCommand);
+                    return;
+                }
+
+                if (constraint.loadControl.Value() < -100 || constraint.loadControl.Value() > 100)
+                {
+                    ChipLogError(Zcl, "DEM: RequestConstraintBasedForecast bad loadControl %d", constraint.loadControl.HasValue());
```
bzbarsky-apple, line 818:
```suggestion
                    ChipLogError(Zcl, "DEM: RequestConstraintBasedForecast bad loadControl %d", constraint.loadControl.Value());
```

src/app/clusters/device-energy-management-server/device-energy-management-server.cpp
```diff
@@ -654,6 +752,114 @@ void Instance::HandleRequestConstraintBasedForecast(HandlerContext & ctx,
...
+                const Structs::ConstraintsStruct::DecodableType & constraint = iterator.GetValue();
+                if (constraint.startTime < prevConstraint.startTime ||
+                    prevConstraint.startTime + prevConstraint.duration >= constraint.startTime)
+                {
+                    ChipLogError(Zcl, "DEM: RequestConstraintBasedForecast overlapping constraint times");
+                    ctx.mCommandHandler.AddStatus(ctx.mRequestPath, Status::ConstraintError);
+                    return;
+                }
+
+                prevConstraint = iterator.GetValue();
```
bzbarsky-apple, line 852: Might be clearer as:
```suggestion
                prevConstraint = constraint;
```

bzbarsky-apple (approved):

examples/energy-management-app/energy-management-common/energy-management-app.matter
```diff
@@ -1583,7 +1594,7 @@ provisional cluster DeviceEnergyManagement = 152 {
   readonly attribute ESAStateEnum ESAState = 2;
   readonly attribute power_mw absMinPower = 3;
   readonly attribute power_mw absMaxPower = 4;
-  readonly attribute optional nullable PowerAdjustStruct powerAdjustmentCapability[] = 5;
+  readonly attribute optional nullable PowerAdjustCapabilityStruct powerAdjustmentCapability = 5;
```
andy31415, line 1597: this changed the type of a structure. This is not backwards compatible.

What is the history here? 

Also API-wise several changes here are not backwards-compatible. Is this still under development? If this was already released we cannot change it anymore.

PeterC1965: This has not been released yet (it didn't go into 1.3 SVE) and is only provisional/ok to change. It is now in the Matter 1.4 Spec 0.7 ballot in the changed format above.

jamesharrow (approved):

andy31415 (approved):

PR #33734, opened 2024-06-04
pw_unit_test migration: apps-2

Alami-Amine:
- Continuation of migration from nl_test to pw_unit_test, effort first discussed in issue https://github.com/project-chip/connectedhomeip/pull/29342
- 2nd batch of `src/app` tests (related to Interaction Model) first started in #33638

### In this PR:
- Converting Tests from nltest to pw_unit.
- new  `*TestAccess` classes to allow tests to access private class members
- declaring these `*TestAccess` classes as `friend` to the classes they need access to.
  - this approach was chosen over gtest's `FRIEND_TEST` api which needs to be applied to EVERY testcase that needs access to a private class member.

**- UPDATE: It was decided to not use `*TestAccess.h` classes as they overcomplicate the testscript, instead, I re-used the Macro `TEST_F_FROM_FIXTURE` from [ExtraPwTestMacros.h](https://github.com/project-chip/connectedhomeip/blob/master/src/lib/support/tests/ExtraPwTestMacros.h) and created a new PR : https://github.com/project-chip/connectedhomeip/pull/33772**

PR #33772, opened 2024-06-06
pw_unit_test migration: apps final batch

Alami-Amine:
- This PR is the same as https://github.com/project-chip/connectedhomeip/pull/33734, but without the `*TestAccess.h` classes and instead, private tested-class members were accessed using the Macro `TEST_F_FROM_FIXTURE` from [ExtraPwTestMacros.h](https://github.com/project-chip/connectedhomeip/blob/master/src/lib/support/tests/ExtraPwTestMacros.h)

- final batch of src/app tests (related to Interaction Model) first started in https://github.com/project-chip/connectedhomeip/pull/33638

- pw_unit conversion was first discussed in:  https://github.com/project-chip/connectedhomeip/pull/29342

### In the PR

1. This is the final PR of unit tests to be migrated 
2. some NL Test dependencies in  OpenIoT SDK were removed.
3. the `TestCommissioningWindowManager` was renamed and the sleep timeouts were replaced by `RunEventLoop` and a call to  StopEventLoop for each task

- NOTE:  `NL_TEST_ASSERT` API behavior is closest to `EXPECT` than to  `ASSERT` which is why `EXPECT` has been used more widely

- Follow-up: remove other NL_TEST dependencies.


src/app/tests/TestCommandInteraction.cpp
```diff
@@ -850,424 +739,512 @@ void TestCommandInteraction::ValidateCommandHandlerEncodeInvokeResponseMessage(n
...
+    EXPECT_EQ(mockCommandSenderDelegate.onResponseCalledTimes, 0);
+    EXPECT_EQ(mockCommandSenderDelegate.onFinalCalledTimes, 0);
+    EXPECT_EQ(mockCommandSenderDelegate.onErrorCalledTimes, 0);
 
-    NL_TEST_ASSERT(apSuite, GetNumActiveCommandResponderObjects() == 0);
+    EXPECT_EQ(GetNumActiveCommandResponderObjects(), 0u);
 
     System::PacketBufferHandle msgBuf = System::PacketBufferHandle::New(kMaxSecureSduLengthBytes);
-    NL_TEST_ASSERT(apSuite, !msgBuf.IsNull());
+    EXPECT_FALSE(msgBuf.IsNull());
```
jlatusek, line 919: Use ASSERT

src/app/tests/TestTimedHandler.cpp
```diff
@@ -93,189 +115,153 @@ class TestExchangeDelegate : public Messaging::ExchangeDelegate
 
 } // anonymous namespace
 
-void TestTimedHandler::GenerateTimedRequest(nlTestSuite * aSuite, uint16_t aTimeoutValue, System::PacketBufferHandle & aPayload)
+void TestTimedHandler::GenerateTimedRequest(uint16_t aTimeoutValue, System::PacketBufferHandle & aPayload)
 {
     aPayload = System::PacketBufferHandle::New(System::PacketBuffer::kMaxSize);
-    NL_TEST_ASSERT(aSuite, !aPayload.IsNull());
+    EXPECT_FALSE(aPayload.IsNull());
```
jlatusek, line 121:
```suggestion
    ASSERT_FALSE(aPayload.IsNull());
```

src/app/tests/TestTimedHandler.cpp
```diff
@@ -93,189 +115,153 @@ class TestExchangeDelegate : public Messaging::ExchangeDelegate
...
+    pTestContext->DrainAndServiceIO();
+    EXPECT_TRUE(delegate.mNewMessageReceived);
+    EXPECT_TRUE(delegate.mLastMessageWasStatus);
+    EXPECT_EQ(delegate.mError, CHIP_NO_ERROR);
 
     // Send an empty payload, which will error out but not with the
     // UNSUPPORTED_ACCESS status we expect if we miss our timeout.
     payload = MessagePacketBuffer::New(0);
-    NL_TEST_ASSERT(aSuite, !payload.IsNull());
+    EXPECT_FALSE(payload.IsNull());
```
jlatusek, line 163:
```suggestion
    ASSERT_FALSE(payload.IsNull());
```

jlatusek (changes requested):

src/app/tests/TestAclAttribute.cpp
```diff
@@ -121,29 +109,65 @@ class MockInteractionModelApp : public chip::app::ReadClient::Callback
...
 
-class TestAclAttribute
+class TestAclAttribute : public ::testing::Test
 {
 public:
-    static void TestACLDeniedAttribute(nlTestSuite * apSuite, void * apContext);
+    static void SetUpTestSuite()
+    {
+
+        mpTestContext = new chip::Test::AppContext;
```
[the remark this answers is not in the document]
Alami-Amine: This was done on a lot of tests as of now, I prefer to change that on a seperate future PR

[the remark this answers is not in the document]
jlatusek: Ok, If so I think you should create an issue to keep it in mind.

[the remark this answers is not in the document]
Alami-Amine: yup will do

src/app/tests/TestCommandInteraction.cpp
Alami-Amine: done for 4 occurances

src/app/tests/TestTimedHandler.cpp
```diff
@@ -93,189 +115,153 @@ class TestExchangeDelegate : public Messaging::ExchangeDelegate
...
 
     System::PacketBufferHandle payload;
-    GenerateTimedRequest(aSuite, 50, payload);
+    GenerateTimedRequest(50, payload);
 
     TestExchangeDelegate delegate;
-    ExchangeContext * exchange = ctx.NewExchangeToAlice(&delegate);
-    NL_TEST_ASSERT(aSuite, exchange != nullptr);
+    ExchangeContext * exchange = pTestContext->NewExchangeToAlice(&delegate);
+    EXPECT_NE(exchange, nullptr);
```
feasel0, line 196:
```suggestion
    ASSERT_NE(exchange, nullptr);
```
since exchange gets dereferenced after this

src/app/tests/TestTimedHandler.cpp
```diff
@@ -16,21 +16,19 @@
  *    limitations under the License.
  */
 
+#include <gtest/gtest.h>
```
feasel0, line 19: change to the new include

src/app/tests/TestInteractionModelEngine.cpp
```diff
@@ -115,249 +137,248 @@ void TestInteractionModelEngine::TestAttributePathParamsPushRelease(nlTestSuite
...
     InteractionModelEngine * engine = InteractionModelEngine::GetInstance();
 
     NodeId bobNodeId           = 0x12344321ull;
     FabricIndex bobFabricIndex = 1;
 
     // Create ExchangeContext
-    Messaging::ExchangeContext * exchangeCtx1 = ctx.NewExchangeToBob(nullptr, false);
-    NL_TEST_ASSERT(apSuite, exchangeCtx1);
+    Messaging::ExchangeContext * exchangeCtx1 = mpTestContext->NewExchangeToBob(nullptr, false);
+    EXPECT_TRUE(exchangeCtx1);
```
feasel0, line 299: Does exchangeCtx1 get dereferenced inside `engine->GetReadHandlerPool().CreateObject()` without being checked for nullness first?  If so then we'd want ASSERT here.

src/app/tests/TestInteractionModelEngine.cpp
```diff
@@ -115,249 +137,248 @@ void TestInteractionModelEngine::TestAttributePathParamsPushRelease(nlTestSuite
...
     InteractionModelEngine * engine = InteractionModelEngine::GetInstance();
 
     NodeId bobNodeId           = 0x12344321ull;
     FabricIndex bobFabricIndex = 1;
 
     // Create ExchangeContexts
-    Messaging::ExchangeContext * exchangeCtx1 = ctx.NewExchangeToBob(nullptr, false);
-    NL_TEST_ASSERT(apSuite, exchangeCtx1);
+    Messaging::ExchangeContext * exchangeCtx1 = mpTestContext->NewExchangeToBob(nullptr, false);
+    EXPECT_TRUE(exchangeCtx1);
```
feasel0, line 344: Same as my previous comment regarding `exchangeCtx1`

src/app/tests/TestCommandInteraction.cpp
```diff
@@ -95,8 +123,8 @@ class SimpleTLVPayload : public app::DataModel::EncodableToTLV
     CHIP_ERROR EncodeTo(TLV::TLVWriter & writer, TLV::Tag tag) const override
     {
         TLV::TLVType outerType;
-        ReturnErrorOnFailure(writer.StartContainer(tag, TLV::kTLVType_Structure, outerType));
-        ReturnErrorOnFailure(writer.PutBoolean(chip::TLV::ContextTag(1), true));
+        EXPECT_EQ(writer.StartContainer(tag, TLV::kTLVType_Structure, outerType), CHIP_NO_ERROR);
```
tehampson, line 126: Don't you want ASSERT_EQ here? My understanding is that ASSERT ends the test when it is unsafe to proceed further in the test and EXPECT will fail the test but continue running. I am guessing the old reason for `ReturnErrorOnFailure` was because we didn't want the next line of code to run if this like failed.

This comment applies to all instances of the coversions of a `ReturnErrorOnFailure` to and `EXPECT_EQ` in this PR not just here

src/app/tests/TestCommandInteraction.cpp
```diff
@@ -453,162 +502,141 @@ CommandPathParams MakeTestCommandPath(CommandId aCommandId = kTestCommandIdWithD
...
     writer.Init(std::move(aPayload));
 
-    err = invokeRequestMessageBuilder.Init(&writer);
-    NL_TEST_ASSERT(apSuite, err == CHIP_NO_ERROR);
+    EXPECT_EQ(invokeRequestMessageBuilder.Init(&writer), CHIP_NO_ERROR);
 
     invokeRequestMessageBuilder.SuppressResponse(true).TimedRequest(aIsTimedRequest);
     InvokeRequests::Builder & invokeRequests = invokeRequestMessageBuilder.CreateInvokeRequests();
-    NL_TEST_ASSERT(apSuite, invokeRequestMessageBuilder.GetError() == CHIP_NO_ERROR);
+    EXPECT_EQ(invokeRequestMessageBuilder.GetError(), CHIP_NO_ERROR);
```
tehampson, line 517: This is another instance where I think we should be using ASSERT_EQ instead of EXPECT_EQ

src/app/tests/TestReadInteraction.cpp
```diff
@@ -2150,165 +2119,155 @@ void TestReadInteraction::TestSubscribeEarlyReport(nlTestSuite * apSuite, void *
...
+        EXPECT_EQ(err, CHIP_NO_ERROR);
 
-        ctx.DrainAndServiceIO();
+        mpTestContext->DrainAndServiceIO();
         System::Clock::Timestamp startTime = gMockClock.GetMonotonicTimestamp();
 
-        NL_TEST_ASSERT(apSuite, engine->GetNumActiveReadHandlers() == 1);
-        NL_TEST_ASSERT(apSuite, engine->ActiveHandlerAt(0) != nullptr);
+        EXPECT_EQ(engine->GetNumActiveReadHandlers(), 1u);
+        EXPECT_NE(engine->ActiveHandlerAt(0), nullptr);
```
feasel0, line 2133:
```suggestion
        ASSERT_NE(engine->ActiveHandlerAt(0), nullptr);
```

src/app/tests/TestReadInteraction.cpp
```diff
@@ -2150,165 +2119,155 @@ void TestReadInteraction::TestSubscribeEarlyReport(nlTestSuite * apSuite, void *
...
+                  gMockClock.GetMonotonicTimestamp() + Seconds16(readPrepareParams.mMinIntervalFloorSeconds));
+        EXPECT_EQ(gReportScheduler->GetMaxTimestampForHandler(delegate.mpReadHandler),
+                  gMockClock.GetMonotonicTimestamp() + Seconds16(maxInterval));
 
         // Confirm that the node is scheduled to run
-        NL_TEST_ASSERT(apSuite, gReportScheduler->IsReportScheduled(delegate.mpReadHandler));
+        EXPECT_TRUE(gReportScheduler->IsReportScheduled(delegate.mpReadHandler));
         ReportScheduler::ReadHandlerNode * node = gReportScheduler->GetReadHandlerNode(delegate.mpReadHandler);
-        NL_TEST_ASSERT(apSuite, node != nullptr);
+        EXPECT_NE(node, nullptr);
```
feasel0, line 2150:
```suggestion
        ASSERT_NE(node, nullptr);
```

src/app/tests/TestReadInteraction.cpp
```diff
@@ -2150,165 +2119,155 @@ void TestReadInteraction::TestSubscribeEarlyReport(nlTestSuite * apSuite, void *
...
+        EXPECT_EQ(gReportScheduler->GetMinTimestampForHandler(delegate.mpReadHandler),
+                  gMockClock.GetMonotonicTimestamp() + Seconds16(readPrepareParams.mMinIntervalFloorSeconds));
+        EXPECT_EQ(gReportScheduler->GetMaxTimestampForHandler(delegate.mpReadHandler),
+                  gMockClock.GetMonotonicTimestamp() + Seconds16(maxInterval));
 
         // Confirm that the node is scheduled to run
-        NL_TEST_ASSERT(apSuite, gReportScheduler->IsReportScheduled(delegate.mpReadHandler));
-        NL_TEST_ASSERT(apSuite, node != nullptr);
+        EXPECT_TRUE(gReportScheduler->IsReportScheduled(delegate.mpReadHandler));
+        EXPECT_NE(node, nullptr);
```
feasel0, line 2220:
```suggestion
        ASSERT_NE(node, nullptr);
```

src/app/tests/TestReadInteraction.cpp
```diff
@@ -2331,39 +2290,41 @@ void TestReadInteraction::TestSubscribeUrgentWildcardEvent(nlTestSuite * apSuite
...
+        EXPECT_EQ(err, CHIP_NO_ERROR);
 
-        ctx.DrainAndServiceIO();
+        mpTestContext->DrainAndServiceIO();
         System::Clock::Timestamp startTime = gMockClock.GetMonotonicTimestamp();
 
-        NL_TEST_ASSERT(apSuite, engine->GetNumActiveReadHandlers() == 2);
-        NL_TEST_ASSERT(apSuite, engine->ActiveHandlerAt(0) != nullptr);
+        EXPECT_EQ(engine->GetNumActiveReadHandlers(), 2u);
+        EXPECT_NE(engine->ActiveHandlerAt(0), nullptr);
```
feasel0, line 2310: ASSERT

src/app/tests/TestReadInteraction.cpp
```diff
@@ -2331,39 +2290,41 @@ void TestReadInteraction::TestSubscribeUrgentWildcardEvent(nlTestSuite * apSuite
...
         nonUrgentDelegate.mpReadHandler = engine->ActiveHandlerAt(0);
-        NL_TEST_ASSERT(apSuite, engine->ActiveHandlerAt(1) != nullptr);
+        EXPECT_NE(engine->ActiveHandlerAt(1), nullptr);
```
feasel0, line 2312: ASSERT

src/app/tests/TestReadInteraction.cpp
```diff
@@ -2628,10 +2581,10 @@ void TestReadInteraction::TestSubscribeWildcard(nlTestSuite * apSuite, void * ap
...
         constexpr size_t kExpectedAttributeResponse = 29 * 2 + (kMockAttribute4ListLength - 4) + (kMockAttribute4ListLength - 4);
 #endif
-        NL_TEST_ASSERT(apSuite, delegate.mNumAttributeResponse == kExpectedAttributeResponse);
-        NL_TEST_ASSERT(apSuite, delegate.mNumArrayItems == 12);
-        NL_TEST_ASSERT(apSuite, engine->GetNumActiveReadHandlers(ReadHandler::InteractionType::Subscribe) == 1);
-        NL_TEST_ASSERT(apSuite, engine->ActiveHandlerAt(0) != nullptr);
+        EXPECT_EQ((unsigned) delegate.mNumAttributeResponse, kExpectedAttributeResponse);
+        EXPECT_EQ(delegate.mNumArrayItems, 12);
+        EXPECT_EQ(engine->GetNumActiveReadHandlers(ReadHandler::InteractionType::Subscribe), 1u);
+        EXPECT_NE(engine->ActiveHandlerAt(0), nullptr);
```
feasel0, line 2587: I'll stop individually marking these, but i think most of the cases where we have `EXPECT_NE\(.*nullptr\);` are going to be testing pointers that we're about to dereference

src/app/tests/TestTimedHandler.cpp
```diff
@@ -93,189 +115,153 @@ class TestExchangeDelegate : public Messaging::ExchangeDelegate
...
-    NL_TEST_ASSERT(aSuite, !aPayload.IsNull());
+    ASSERT_FALSE(aPayload.IsNull());
 
     System::PacketBufferTLVWriter writer;
     writer.Init(std::move(aPayload));
 
     TimedRequestMessage::Builder builder;
     CHIP_ERROR err = builder.Init(&writer);
-    NL_TEST_ASSERT(aSuite, err == CHIP_NO_ERROR);
+    EXPECT_EQ(err, CHIP_NO_ERROR);
```
jlatusek, line 128: In the other files, there is still a separate assignment and verification of the error, please fix that for the rest of the cases.

``` cpp
    EXPECT_EQ(builder.Init(&writer), CHIP_NO_ERROR);
```

src/app/tests/TestCommandInteraction.cpp
Alami-Amine: done for two occurrences

src/app/tests/TestCommandInteraction.cpp
Alami-Amine: so technically we can not have an `ASSERT_EQ` because this function doesn't return a Void and using ASSERT will make it fail (since the macro will not know what to return) 
However I think we can just leave a ReturnErrorOnFailure as was done before.

src/app/tests/TestReadInteraction.cpp
Alami-Amine: done for many  occurances

src/app/tests/TestTimedHandler.cpp
Alami-Amine: done for all the files in this PR, 
FYI, there are many other converted tests where this was not done

src/app/tests/TestAclAttribute.cpp
```diff
@@ -121,29 +109,63 @@ class MockInteractionModelApp : public chip::app::ReadClient::Callback
...
+
+        mpTestContext->SetUpTestSuite();
+    }
+    static void TearDownTestSuite()
+    {
+        mpTestContext->TearDownTestSuite();
+        if (mpTestContext != nullptr)
```
andy31415, line 125: you are using mpTestContext above ... so it is not nullptr or it would segfault.

Should we move the tear down here or just not have the check and trust that teardown follows setup?

src/app/tests/TestAclAttribute.cpp
```diff
@@ -121,29 +109,63 @@ class MockInteractionModelApp : public chip::app::ReadClient::Callback
...
+        {
+            delete mpTestContext;
+        }
+    }
+
+    void SetUp() override
+    {
+
+        if (mpTestContext != nullptr)
```
andy31415, line 134: under what circumstances will mpTestContext be null? Both here and in TearDown?

src/app/tests/TestEventOverflow.cpp
```diff
@@ -52,9 +51,23 @@ static uint8_t gInfoEventBuffer[2048];
...
 public:
+    static void SetUpTestSuite()
+    {
+        mpTestContext = new chip::Test::AppContext;
+        mpTestContext->SetUpTestSuite();
+    }
+    static void TearDownTestSuite()
+    {
+        mpTestContext->TearDownTestSuite();
+        if (mpTestContext != nullptr)
```
andy31415, line 65: same comment here: mpTestContext is used already, so check seems redundant. I assume this may be throughout the PR. 

I wonder if given this is test code, maybe we can assume that "new" does not really fail so we can skip all these nullability checks. We have to trust the framework that TearDown is called only after SetUp.

src/app/tests/TestCommandInteraction.cpp
andy31415: ReturnErrorOnFailure seems correct. There should be no assert/expect changes here.

src/app/tests/TestEventOverflow.cpp
Alami-Amine: maybe I have trust issues....done

src/app/tests/TestTimedHandler.cpp
```diff
@@ -93,189 +100,146 @@ class TestExchangeDelegate : public Messaging::ExchangeDelegate
...
 {
-    TestFollowingMessageTooSlow(aSuite, aContext, MsgType::WriteRequest);
+    TestFollowingMessageTooSlow(MsgType::WriteRequest);
 }
 
-void TestTimedHandler::TestInvokeNeverComes(nlTestSuite * aSuite, void * aContext)
+TEST_F(TestTimedHandler, TestInvokeNeverComes)
 {
-    TestContext & ctx = *static_cast<TestContext *>(aContext);
+    // TestContext & ctx = *static_cast<TestContext *>(aContext);
```
andy31415, line 222: Let''s delete commented out code, here and above

andy31415 (approved):

tehampson (approved):
My focus was more on TestCommandInteraction, with a quick glance at other tests

src/app/tests/TestWriteInteraction.cpp
```diff
@@ -25,19 +28,50 @@
...
+#include <lib/support/tests/ExtraPwTestMacros.h>
 #include <messaging/ExchangeContext.h>
 #include <messaging/Flags.h>
+#include <pw_unit_test/framework.h>
 
-#include <memory>
-#include <nlunit-test.h>
-#include <utility>
+/**
+ * Helper macro we can use to pretend we got a reply from the server in cases
```
jlatusek, line 43: It seems there is 3 definition of the same macro. Can you refactor it and make only one?

jlatusek (changes requested):

PR #33773, opened 2024-06-06
Log errors if AdvertiseOperational fails

andy31415:
Existing code would silently fail, so advertising errors are hard to debug.

> [!IMPORTANT]
> SRP advertisement seems to rely on being in state "initialized" until openthread connectivity is possible, so
> applications will likely log errors at boot time before commissioning.
> This change is probably still good as this would encourage us to clean up "error is expected" paths.

### DIFF
--- a/src/app/clusters/operational-credentials-server/operational-credentials-server.cpp
+++ b/src/app/clusters/operational-credentials-server/operational-credentials-server.cpp
@@ -703,7 +703,7 @@ bool emberAfOperationalCredentialsClusterAddNOCCallback(app::CommandHandler * co
     needRevert = false;
 
     // We might have a new operational identity, so we should start advertising it right away.
-    app::DnssdServer::Instance().AdvertiseOperational();
+    LogErrorOnFailure(app::DnssdServer::Instance().AdvertiseOperational());
 
     // Notify the attributes containing fabric metadata can be read with new data
     MatterReportingAttributeChangeCallback(commandPath.mEndpointId, OperationalCredentials::Id,
--- a/src/app/clusters/operational-credentials-server/operational-credentials-server.cpp
+++ b/src/app/clusters/operational-credentials-server/operational-credentials-server.cpp
@@ -703,7 +703,11 @@ bool emberAfOperationalCredentialsClusterAddNOCCallback(app::CommandHandler * co
     needRevert = false;
 
     // We might have a new operational identity, so we should start advertising it right away.
-    LogErrorOnFailure(app::DnssdServer::Instance().AdvertiseOperational());
+    err = app::DnssdServer::Instance().AdvertiseOperational();
+    if (err != CHIP_NO_ERROR)
+    {
+        ChipLogError(AppServer, "Operational advertising failed: %" CHIP_ERROR_FORMAT, err.Format());
+    }
 
     // Notify the attributes containing fabric metadata can be read with new data
     MatterReportingAttributeChangeCallback(commandPath.mEndpointId, OperationalCredentials::Id,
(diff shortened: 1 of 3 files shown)


src/app/server/CommissioningWindowManager.cpp
```diff
@@ -84,7 +84,7 @@ void CommissioningWindowManager::OnPlatformEvent(const DeviceLayer::ChipDeviceEv
     }
     else if (event->Type == DeviceLayer::DeviceEventType::kOperationalNetworkEnabled)
     {
-        app::DnssdServer::Instance().AdvertiseOperational();
+        LogErrorOnFailure(app::DnssdServer::Instance().AdvertiseOperational());
```
Damian-Nordic, line 87: And what about the progress log below?

src/app/clusters/operational-credentials-server/operational-credentials-server.cpp
```diff
@@ -703,7 +703,7 @@ bool emberAfOperationalCredentialsClusterAddNOCCallback(app::CommandHandler * co
     needRevert = false;
 
     // We might have a new operational identity, so we should start advertising it right away.
-    app::DnssdServer::Instance().AdvertiseOperational();
+    LogErrorOnFailure(app::DnssdServer::Instance().AdvertiseOperational());
```
Damian-Nordic, line 706: I noticed it started before your PR but we shouldn't really use logging that refers to `__FILE__` in the code for constrained devices. In general, I don't know why LogErrorOnFailure includes the error location while we already have `CHIP_CONFIG_ERROR_SOURCE` exactly for that. I think if we want to start using these macros more widely, LogErrorOnFailure should not directly refer to `__FILE__` and `__LINE__` macros (and rely on `CHIP_CONFIG_ERROR_SOURCE`).

src/app/server/CommissioningWindowManager.cpp
andy31415: I am actually also re-considering this PR:  apparently for thread it is normal to get an error here because SRP is still being initialized. This may make spammy logs, unsure if good (it should help us not treat failures as normal) or annoying.

Will update logic though to not use LogErrorOnFailure due to usage of __FILE__

yufengwangca (approved):

bzbarsky-apple (approved):

src/app/server/CommissioningWindowManager.cpp
```diff
@@ -84,8 +84,15 @@ void CommissioningWindowManager::OnPlatformEvent(const DeviceLayer::ChipDeviceEv
     }
     else if (event->Type == DeviceLayer::DeviceEventType::kOperationalNetworkEnabled)
     {
-        app::DnssdServer::Instance().AdvertiseOperational();
-        ChipLogProgress(AppServer, "Operational advertising enabled");
+        CHIP_ERROR err = app::DnssdServer::Instance().AdvertiseOperational();
```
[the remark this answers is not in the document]
andy31415: I wonder if we ever will have some form of fallback ... since if you cannot advertise operational you probably are not a very useful device. maybe it would be an idea for some device implementations to reboot if this fails or flash some lights as "we are in error state". I think just logging and being void is insufficient.

Or are you suggesting to log and return a CHIP_ERROR? In that case I wonder if we double-handle as it seems not as expected that if you get an error, it was already logged for you.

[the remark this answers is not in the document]
bzbarsky-apple: I meant the latter (log and return error), but I don't feel very strongly about it.

PR #33775, opened 2024-06-06
[Python] Remove Python Bluetooth and ChipStack event loop integration

agners:
The Python Bluetooth implementation for Linux (`BluezManager` in ChipBluezMgr.py) and macOS (`CoreBluetoothManager` in ChipCoreBluetoothMgr.py) integrate with ChipStack to pump their event loops on long running operations such as commissioning (through `CallAsyncWithCompleteCallback()`). From what I can tell this is no longer used as the Bluetooth communication is entirely handled by the respective native implementations for Linux and macOS today. It seems the Python Bluetooth managers are only used for some mbed integration tests. Specifically through `scan_chip_ble_devices()` in src/test_driver/mbed/integration_tests/common/utils.py. This operation anyway busy wait during scanning and don't need any event loop integration.

So as a first step, this PR simply breaks this tie and removes the event loop integration with the Device ChipStack/ChipDeviceController.

### DIFF
--- a/src/controller/python/chip/ChipStack.py
+++ b/src/controller/python/chip/ChipStack.py
@@ -165,8 +165,6 @@ class ChipStack(object):
             callback()
 
         self.cbHandleChipThreadRun = HandleChipThreadRun
-        # set by other modules(BLE) that require service by thread while thread blocks.
-        self.blockingCB = None
 
         #
         # Storage has to be initialized BEFORE initializing the stack, since the latter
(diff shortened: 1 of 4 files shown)


agners:
Note that the event loop integration is rather hacky anyways. Especially on the Linux side this could be done much nicer with Python asyncio and GLib. [This StackOverflow](https://stackoverflow.com/a/45299702/184207) answer has the basics: Essentially instead of creating a MainLoop on GLib side, simply `pending()` and `iteration(may_block=False)` regularly from asyncio tasks.

src/controller/python/chip/ChipStack.py
```diff
@@ -255,11 +255,7 @@ def CallAsyncWithCompleteCallback(self, callFunct):
         if not res.is_success:
             self.completeEvent.set()
             raise res.to_exception()
-        while not self.completeEvent.isSet():
-            if self.blockingCB:
```
arkq, line 259: How about removing this `blockingCB` completely? This variable and the setter.

agners: Right, of course, something went wrong here while gitting. Now it should be fine.

arkq (approved):

andy31415 (approved):

PR #33782, opened 2024-06-06
Allow CommandSender request to be built using DataModel::EncodableToTLV

tehampson:
Add new AddRequestData API so that clients don't need to call raw CommandSender APIs, and we can move towards removing access to TLVWriter.

The newly AddRequestData does bypass check that time invoke does in fact provide a timeout, the reality is that things outside of SDK heavily use the raw CommandSender APIs (PrepareCommand/FinishCommand) to build their own requests bypassing this check already. We still allow clients to call the older API for building the request.

### DIFF
--- a/src/app/CommandSender.cpp
+++ b/src/app/CommandSender.cpp
@@ -537,8 +537,8 @@ CHIP_ERROR CommandSender::FinishCommand(FinishCommandParameters & aFinishCommand
     return FinishCommandInternal(aFinishCommandParams);
 }
 
-CHIP_ERROR CommandSender::AddRequestData(const CommandPathParams & aCommandPath, const DataModel::EncodableToTLV & aEncodable,
-                                         AddRequestDataParameters & aAddRequestDataParams)
+CHIP_ERROR CommandSender::AddRequest(const CommandPathParams & aCommandPath, const DataModel::EncodableToTLV & aEncodable,
+                                     AddRequestDataParameters & aAddRequestDataParams)
 {
     PrepareCommandParameters prepareCommandParams(aAddRequestDataParams);
     ReturnErrorOnFailure(PrepareCommand(aCommandPath, prepareCommandParams));
--- a/src/app/CommandSender.cpp
+++ b/src/app/CommandSender.cpp
@@ -537,8 +537,8 @@ CHIP_ERROR CommandSender::FinishCommand(FinishCommandParameters & aFinishCommand
     return FinishCommandInternal(aFinishCommandParams);
 }
 
-CHIP_ERROR CommandSender::AddRequest(const CommandPathParams & aCommandPath, const DataModel::EncodableToTLV & aEncodable,
-                                     AddRequestDataParameters & aAddRequestDataParams)
+CHIP_ERROR CommandSender::AddRequestData(const CommandPathParams & aCommandPath, const DataModel::EncodableToTLV & aEncodable,
+                                         AddRequestDataParameters & aAddRequestDataParams)
 {
     PrepareCommandParameters prepareCommandParams(aAddRequestDataParams);
     ReturnErrorOnFailure(PrepareCommand(aCommandPath, prepareCommandParams));
--- a/src/app/CommandSender.cpp
+++ b/src/app/CommandSender.cpp
@@ -537,7 +537,8 @@ CHIP_ERROR CommandSender::FinishCommand(FinishCommandParameters & aFinishCommand
     return FinishCommandInternal(aFinishCommandParams);
 }
 
-CHIP_ERROR CommandSender::AddRequestDataInternal(const CommandPathParams & aCommandPath, DataModel::EncodableToTLV & aEncodable,
+CHIP_ERROR CommandSender::AddRequestDataInternal(const CommandPathParams & aCommandPath,
+                                                 const DataModel::EncodableToTLV & aEncodable,
                                                  AddRequestDataParameters & aAddRequestDataParams)
 {
     PrepareCommandParameters prepareCommandParams(aAddRequestDataParams);
--- a/src/app/CommandSender.cpp
+++ b/src/app/CommandSender.cpp
@@ -537,8 +537,7 @@ CHIP_ERROR CommandSender::FinishCommand(FinishCommandParameters & aFinishCommand
     return FinishCommandInternal(aFinishCommandParams);
 }
 
-CHIP_ERROR CommandSender::AddRequestData(const CommandPathParams & aCommandPath,
-                                         const DataModel::EncodableToTLV & aEncodable,
+CHIP_ERROR CommandSender::AddRequestData(const CommandPathParams & aCommandPath, const DataModel::EncodableToTLV & aEncodable,
                                          AddRequestDataParameters & aAddRequestDataParams)
 {
     PrepareCommandParameters prepareCommandParams(aAddRequestDataParams);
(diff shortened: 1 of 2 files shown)


src/app/CommandSender.cpp
```diff
@@ -537,6 +537,18 @@ CHIP_ERROR CommandSender::FinishCommand(FinishCommandParameters & aFinishCommand
     return FinishCommandInternal(aFinishCommandParams);
 }
 
+CHIP_ERROR CommandSender::AddRequestDataInternal(const CommandPathParams & aCommandPath, DataModel::EncodableToTLV & aEncodable,
```
andy31415, line 540: nit: EncodeToTLV is const, so we could make aEncodable const ref.
Can we make aAddRequestDataparams also const?

andy31415 (approved):

src/app/CommandSender.cpp
tehampson: Added const to `EncodeToTLV`

aAddRequestDataParams was intentionally left as non-const. At one point we did have CommandSender supply the commandRef, and because the API is trying to make it easier for things to be added in the future (including possible out parameters) we should leave it as non-const IMO

src/app/CommandSender.h
```diff
@@ -375,6 +376,26 @@ class CommandSender final : public Messaging::ExchangeDelegate
 
     TLV::TLVWriter * GetCommandDataIBTLVWriter();
 
+    /**
+     * API for adding request data.  The `aEncodable` is generally expected to encode
+     * a ClusterName::Commands::CommandName::Type struct, however any object should work.
+     *
+     * @param [in] aCommandPath  The path of the command being requested.
+     * @param [in] aEncodable - an encodable that places the command data structure
+     *             for `aResponseCommandId` into a TLV Writer.
```
bzbarsky-apple, line 385: There is no aResponseCommandId here.

src/app/CommandSender.h
```diff
@@ -375,6 +376,26 @@ class CommandSender final : public Messaging::ExchangeDelegate
...
+     * @param [in] aAddRequestDataParams parameters associated with building the
+     *             InvokeRequestMessage that are associated with this request.
+     *
+     * This API does not validate if the command provided requires interaction use a timed
+     * invoke interaction. If the caller wants that certainty they should call the templated
+     * version of AddRequestData.
```
bzbarsky-apple, line 391:
```suggestion
     * This API will not fail if this is an untimed invoke but the command provided requires a timed
     * invoke interaction. If the caller wants that to fail before sending the command, they should call
     * the templated version of AddRequestData.
```

src/app/CommandSender.h
```diff
@@ -375,6 +376,26 @@ class CommandSender final : public Messaging::ExchangeDelegate
 
     TLV::TLVWriter * GetCommandDataIBTLVWriter();
 
+    /**
+     * API for adding request data.  The `aEncodable` is generally expected to encode
+     * a ClusterName::Commands::CommandName::Type struct, however any object should work.
```
bzbarsky-apple, line 381: This doesn't really make sense.  If you have a struct, you'd just call the templated API...

src/app/CommandSender.h
```diff
@@ -375,6 +376,26 @@ class CommandSender final : public Messaging::ExchangeDelegate
...
+     *             InvokeRequestMessage that are associated with this request.
+     *
+     * This API does not validate if the command provided requires interaction use a timed
+     * invoke interaction. If the caller wants that certainty they should call the templated
+     * version of AddRequestData.
+     */
+    CHIP_ERROR AddRequestData(const CommandPathParams & aCommandPath, const DataModel::EncodableToTLV & aEncodable,
+                              AddRequestDataParameters & aAddRequestDataParams)
+    {
+        return AddRequestDataInternal(aCommandPath, aEncodable, aAddRequestDataParams);
```
bzbarsky-apple, line 396: Why do we need the separate AddRequestDataInternal() function at all?  Just implement `AddRequestData` out of line with these parameters, and call it from the other AddRequestData.

src/app/CommandSender.h
```diff
@@ -375,6 +376,22 @@ class CommandSender final : public Messaging::ExchangeDelegate
 
     TLV::TLVWriter * GetCommandDataIBTLVWriter();
 
+    /**
+     * API for adding request data using DataModel::EncodableToTLV.
+     *
+     * @param [in] aCommandPath  The path of the command being requested.
+     * @param [in] aEncodable - an encodable that places the command data structure
+     *             for `aAddRequestDataParams` into a TLV Writer.
```
bzbarsky-apple, line 384: That's not what it does, though....  aEncodable never even sees aAddRequestDataParams.

src/app/CommandSender.h
```diff
@@ -398,8 +415,10 @@ class CommandSender final : public Messaging::ExchangeDelegate
         VerifyOrReturnError(!CommandDataT::MustUseTimedInvoke() || aAddRequestDataParams.timedInvokeTimeoutMs.HasValue(),
                             CHIP_ERROR_INVALID_ARGUMENT);
 
-        return AddRequestDataInternal(aCommandPath, aData, aAddRequestDataParams);
+        DataModel::EncodableType<CommandDataT> encodable(aData);
+        return AddRequest(aCommandPath, encodable, aAddRequestDataParams);
```
bzbarsky-apple, line 419: So why is the new API named `AddRequest`, not `AddRequestData` like it was in the original version of this PR?

src/app/CommandSender.h
tehampson: Fixed. I am sorry, about this I should have been better here.

src/app/CommandSender.h
tehampson: Perhaps this is code smell, but when `AddRequest` is named `AddRequestData` there is a similarity in parameters that causes issues:

```
    CHIP_ERROR AddRequestData(const CommandPathParams & aCommandPath, const DataModel::EncodableToTLV & aEncodable,
                          AddRequestDataParameters & aAddRequestDataParams);

    template <typename CommandDataT>
    CHIP_ERROR AddRequestData(const CommandPathParams & aCommandPath, const CommandDataT & aData,
                              AddRequestDataParameters & aAddRequestDataParams)
```

Say there is a `SimpleTLVPayload` that implements `app::DataModel::EncodableToTLV`, when you have code like :
```
SimpleTLVPayload simple_tlv_payload;
command_sender.AddResponseData(path, simple_tlv_payload, request_data_params);
```
Code fails to compile because it deduces the templated version of AddResponseData. I thought that having clients implement boiler plate code like this:
```
SimpleTLVPayload simple_tlv_payload;
app::DataModel::EncodableToTLV & encodable(simple_tlv_payload);
command_sender.AddResponseData(path, encodable, request_data_params);
```
was too cumbersome, so I went with rename for the non-templated version of the call

bzbarsky-apple: Ah, overload resolution ends up not doing the right thing for us here?  OK.

Was this problem there also with the previous PR version?

bzbarsky-apple: Sounds like it was.  How do you feel about this:

```
    template <typename CommandDataT, typename std::enable_if_t<!std::is_base_of_v<DataModel::EncodableToTLV, CommandDataT>, int> = 0>
    CHIP_ERROR AddRequestData(const CommandPathParams & aCommandPath, const CommandDataT & aData,
                              AddRequestDataParameters & aAddRequestDataParams)
```

and then calling the new thing also AddRequestData?  This should work, I believe (tested locally).

tehampson: Made the change in this PR. It seem to work for me on my local changes thus far.

bzbarsky-apple (approved):

PR #33786, opened 2024-06-06
Proposed changes to split tests into separate binaries for efr32 [Version 1]

feasel0:
REPLACED BY:  [34769](https://github.com/project-chip/connectedhomeip/pull/34769)

The approach here is to keep the `silabs_executable` targets inside `src/test_driver/efr32/BUILD.gn` and to leave `chip_test_suite` alone without any knowledge of efr32.

This creates a `.s37` file for each test suite.  I have tested the build process and it seems to generate the correct files, with each `.s37` file only having the tests for that particular suite.  However without the hardware I have not yet flashed and run them on a device, so I can only judge by the `.map` file what's inside them.  But in general I believe this PR is functional and does what it's supposed to do.

Note that even though there's a binary for each test directory (rather than a binary for each individual test source within that directory), each of the binaries is able to build without a complaint about flash size limitations.  So that means so far none of our test directories are too large to fit on the device.  So we could get away with 20-something binaries instead of 200-something if we wanted to go this route.

The drawback to this approach is that it requires `src/test_driver/efr32/BUILD.gn` to have a list of all test directories. This is not ideal since it will need to be updated any time a new test directory is created.  To avoid this I have made another version which moves the `silabs_executable` target into `chip_test_suite`.  See [Version 2](https://github.com/project-chip/connectedhomeip/pull/33790).

### DIFF
--- a/third_party/silabs/silabs_executable.gni
+++ b/third_party/silabs/silabs_executable.gni
@@ -61,14 +61,16 @@ template("silabs_executable") {
     flashing_image_name = output_base_name + ".rps"
   }
 
-  flashing_runtime_target = target_name + ".flashing_runtime"
   flashing_script_inputs = [
     "${chip_root}/scripts/flashing/silabs_firmware_utils.py",
     "${chip_root}/scripts/flashing/firmware_utils.py",
   ]
-  copy(flashing_runtime_target) {
-    sources = flashing_script_inputs
-    outputs = [ "${root_out_dir}/{{source_file_part}}" ]
+  if (defined(invoker.copy_py_files) && invoker.copy_py_files) {  #++++ added this
+    flashing_runtime_target = target_name + ".flashing_runtime"
+    copy(flashing_runtime_target) {
+      sources = flashing_script_inputs
+      outputs = [ "${root_out_dir}/{{source_file_part}}" ]
+    }
   }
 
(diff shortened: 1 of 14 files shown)


third_party/silabs/silabs_executable.gni
```diff
@@ -61,14 +61,16 @@ template("silabs_executable") {
...
 
-  flashing_runtime_target = target_name + ".flashing_runtime"
   flashing_script_inputs = [
     "${chip_root}/scripts/flashing/silabs_firmware_utils.py",
     "${chip_root}/scripts/flashing/firmware_utils.py",
   ]
-  copy(flashing_runtime_target) {
-    sources = flashing_script_inputs
-    outputs = [ "${root_out_dir}/{{source_file_part}}" ]
+  if (defined(invoker.copy_py_files) && invoker.copy_py_files) {  #++++ added this
```
feasel0, line 68: The "copy" action only needs to be done once per build, however now that `silabs_executable` is being called many times it ends up trying to copy these same python files repeatedly which causes an error.  I could not find a clean way of making it only copy once, so as a workaround I use a variable `copy_py_files` that controls whether the files get copied or not.  This variable is set to false after silabs_executable is invoked so the files won't be copied next time.

PR #33787, opened 2024-06-06
Proposed changes to split tests into separate binaries for efr32 [Version 1.5]

feasel0:
The approach here is to move the `silabs_executable` target into `chip_test_suite`.  The advantage is that we don't need to update anything if we add a new directory of tests, since it will make a new efr32 binary target each time `chip_test_suite` gets invoked.  The disadvantage is that it moves a lot of efr32-specific stuff into `chip_test_suite`, which may not be desirable.

In this version I experimented with creating a binary for each test suite (e.g. `src/messaging/tests`) versus creating a binary for each test source (e.g. `src/messaging/tests/TestExchange.cpp`).  There are some debug flags at the top of `build/chip/chip_test_suite.gni` that control which one is being implemented.  It is currently set to create a binary for each test source.

So far I have not gotten this to build successfully, probably due to `nl_test_service` being defined in the wrong place.  Solving this issue is the next step, however I wanted to stop and get some feedback about the overall approach before going further.

### DIFF
--- a/build/chip/chip_test_suite.gni
+++ b/build/chip/chip_test_suite.gni
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+#++++ DEBUG: These flags control whether we're creating a binary for each test suite or each individual test source.  This is a temporary solution while debugging.
 enable_efr32_build_per_suite = false  #++++ TODO: Figure out how to enable things based on `chip_device_platform == "efr32"` which is not currently in scope.
 enable_efr32_build_per_testsource = true  #++++ TODO: Figure out how to enable things based on `chip_device_platform == "efr32"` which is not currently in scope.
 
--- a/build/chip/chip_test_suite.gni
+++ b/build/chip/chip_test_suite.gni
@@ -12,12 +12,29 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+enable_efr32_build_per_suite = false  #++++ TODO: Figure out how to enable things based on `chip_device_platform == "efr32"` which is not currently in scope.
+enable_efr32_build_per_testsource = true  #++++ TODO: Figure out how to enable things based on `chip_device_platform == "efr32"` which is not currently in scope.
+
 import("//build_overrides/build.gni")
 import("//build_overrides/chip.gni")
+if (enable_efr32_build_per_suite || enable_efr32_build_per_testsource) {
+  import("//build_overrides/efr32_sdk.gni")  # Defines efr32_sdk_build_root
+  import("//build_overrides/nlunit_test.gni")  # Defines nlunit_test_root
+}
 
 import("${chip_root}/build/chip/chip_test.gni")
 import("${chip_root}/build/chip/tests.gni")
 import("${dir_pw_unit_test}/test.gni")
+if (enable_efr32_build_per_suite || enable_efr32_build_per_testsource) {
+  import("${efr32_sdk_build_root}/efr32_sdk.gni")
+  import("${efr32_sdk_build_root}/silabs_executable.gni")
+  #import("$dir_pw_protobuf_compiler/proto.gni")
+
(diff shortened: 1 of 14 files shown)


build/chip/chip_test_suite.gni
```diff
@@ -12,12 +12,30 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+#++++ DEBUG: These flags control whether we're creating a binary for each test suite or each individual test source.  This is a temporary solution while debugging.
+enable_efr32_build_per_suite = false  #++++ TODO: Figure out how to enable things based on `chip_device_platform == "efr32"` which is not currently in scope.
+enable_efr32_build_per_testsource = true  #++++ TODO: Figure out how to enable things based on `chip_device_platform == "efr32"` which is not currently in scope.
```
feasel0, line 17: Having these flags is just a temporary solution, since I didn't yet figure out how to access the value of `chip_device_platform` in this scope in order to test if we're building for efr32.

build/chip/chip_test_suite.gni
```diff
@@ -12,12 +12,30 @@
...
+if (enable_efr32_build_per_suite || enable_efr32_build_per_testsource) {
+  import("${efr32_sdk_build_root}/efr32_sdk.gni")
+  import("${efr32_sdk_build_root}/silabs_executable.gni")
+  #import("$dir_pw_protobuf_compiler/proto.gni")
+
+  #++++ TODO: Figure out how to get access to these variables from src/test_driver/efr32/BUILD.gn so we don't have to set them explicitly here.
+  examples_plat_dir = "${chip_root}/examples/platform/silabs/efr32"
+  examples_common_plat_dir = "${chip_root}/examples/platform/silabs"
+  #efr32_project_dir = "${chip_root}/src/test_driver/efr32"
+}
```
feasel0, line 38: A bunch of stuff needs importing into here since it is used by `silabs_executable`.  Eventually I'll find a cleaner way to do this.

third_party/silabs/silabs_executable.gni
```diff
@@ -68,7 +68,7 @@ template("silabs_executable") {
   ]
   copy(flashing_runtime_target) {
     sources = flashing_script_inputs
-    outputs = [ "${root_out_dir}/{{source_file_part}}" ]
+    outputs = [ "${root_out_dir}/${target_name}--{{source_file_part}}" ]
```
feasel0, line 71: I believe the "copy" action only needs to be done once, however now that `silabs_executable` is being called many times it ends up trying to copy these same python files repeatedly which causes an error.  I could not find a clean way of making it only do the copy once, so as a temporary fix I changed the names of the files to be something unique.  However ultimately this is not what we want, so I'll need to circle back and fix this properly.

build/chip/chip_test_suite.gni
```diff
@@ -78,6 +96,115 @@ template("chip_test_suite") {
...
+      # Get it from the suite's output_name.
+      _unique_suite_name = invoker.output_name
+    } else {
+      # Some test suites don't have output_name defined. Use the first source file.
+      _unique_suite_name = string_replace(string_join("-", invoker.test_sources), ".cpp", "")
+    }
+  }
+
+  print("++++ SUITE ${_unique_suite_name}")
+  if (enable_efr32_build_per_suite) {
```
feasel0, line 111: This `if` branch is not executed unless we change the debug flags to enable per-suite binaries instead of per-test binaries.

build/chip/chip_test_suite.gni
```diff
@@ -93,13 +220,17 @@ template("chip_test_suite") {
     }
 
     deps = [ dir_pw_unit_test ]
+    if (enable_efr32_build_per_suite) {
```
feasel0, line 223: This `if` branch is not executed unless we change the debug flags to enable per-suite binaries instead of per-test binaries.

PR #33788, opened 2024-06-06
Fix download log test

### LINKED ISSUE #32636: TestDiagnosticLogsDownloadCommand fails during darwin CI runs quite often
The diagnostic log tests seem to fail quite frequently. Logs - https://github.com/project-chip/connectedhomeip/actions/runs/8347127817/job/22854062448?pr=32631 for an instance

nivi-apple:
Fix the download log test that expects a failure by waiting for the status report before continuing to the next test

Fixes: #32636

### DIFF
--- a/src/app/tests/suites/TestDiagnosticLogsDownloadCommand.yaml
+++ b/src/app/tests/suites/TestDiagnosticLogsDownloadCommand.yaml
@@ -295,6 +295,18 @@ tests:
                     minLength: 9238
                     maxLength: 9238
 
+    - label: "BDX: Wait for 'Diagnostic logs transfer: Success' message"
+      cluster: "DelayCommands"
+      command: "WaitForMessage"
+      arguments:
+          values:
+              - name: "registerKey"
+                value: "default"
+              - name: "message"
+                value: "Diagnostic logs transfer: Success"
+              - name: "timeoutInSeconds"
+                value: 60
+
     - label: "Read Crash log intent"
       command: "Download"
       arguments:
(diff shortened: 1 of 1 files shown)


src/app/clusters/diagnostic-logs-server/BDXDiagnosticLogsProvider.cpp
```diff
@@ -230,14 +230,14 @@ void BDXDiagnosticLogsProvider::OnInternalError()
...
-    Reset();
+    mTransfer.AbortTransfer(GetBdxStatusCodeFromChipError(CHIP_ERROR_INTERNAL));
 }
 
 void BDXDiagnosticLogsProvider::OnTimeout()
 {
     ChipLogError(BDX, "Timeout");
     VerifyOrDo(mIsAcceptReceived, SendCommandResponse(StatusEnum::kDenied));
-    Reset();
+    mTransfer.AbortTransfer(GetBdxStatusCodeFromChipError(CHIP_ERROR_TIMEOUT));
```
harimau-qirex, line 240: How does the transfer get reset after this point (and similarly in OnInternalError)? My understanding is AbortTransfer will trigger a status report to be sent and then the exchange will close since no response is expected for the status report. I don't think this class calls Reset when handling those (through OnMsgToSend or OnExchangeClosing), so would the rest of the internal state still be set?

nivi-apple: Yeah I guess you are right. I looked at BDX a long time ago. From what I remember for OTA we abort transfer and then waited sometime for the status report to go through and called Reset from what I recall. We could something similar. That being said I am not sure if the test is failing due to a genuine issue or merely not cleaning up after finishing the tests and overlaps the next test. I am looking at that now. will update the PR accordingly

PR #33790, opened 2024-06-06
Proposed changes to split tests into separate binaries for efr32 [Version 2]

feasel0:
REPLACED BY:  [34769](https://github.com/project-chip/connectedhomeip/pull/34769)

The approach here is to move the `silabs_executable` target into `chip_test_suite`.  The advantage is that we don't need to update anything if we add a new directory of tests, since it will add new efr32 binary targets each time `chip_test_suite` gets invoked.  The disadvantage is that it moves a lot of efr32-specific stuff into `chip_test_suite`, which may not be desirable.

This version creates a binary for each test source (e.g. `src/messaging/tests/TestExchange.cpp`).  It can easily be modified to create a binary for each suite (e.g. `src/messaging/tests`) just by moving the location of the silabs_executable target outside of the `foreach(_test, invoker.test_sources)` loop and making a few other minor changes.  However in the interest of allowing for very large tests or for test directories with many test sources in them, I decided to split it per test source.

So far I have not gotten this to build successfully, probably due to `nl_test_service` being defined in the wrong place.  Solving this issue is the next step, however I wanted to stop and get some feedback about the overall approach before going further.

### DIFF
--- a/build/chip/chip_test_suite.gni
+++ b/build/chip/chip_test_suite.gni
@@ -12,13 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-#++++ DEBUG: These flags control whether we're creating a binary for each test suite or each individual test source.  This is a temporary solution while debugging.
-enable_efr32_build_per_suite = false  #++++ TODO: Figure out how to enable things based on `chip_device_platform == "efr32"` which is not currently in scope.
 enable_efr32_build_per_testsource = true  #++++ TODO: Figure out how to enable things based on `chip_device_platform == "efr32"` which is not currently in scope.
 
 import("//build_overrides/build.gni")
 import("//build_overrides/chip.gni")
-if (enable_efr32_build_per_suite || enable_efr32_build_per_testsource) {
+if (enable_efr32_build_per_testsource) {
   import("//build_overrides/efr32_sdk.gni")  # Defines efr32_sdk_build_root
   import("//build_overrides/nlunit_test.gni")  # Defines nlunit_test_root
 }
@@ -26,7 +24,7 @@ if (enable_efr32_build_per_suite || enable_efr32_build_per_testsource) {
 import("${chip_root}/build/chip/chip_test.gni")
 import("${chip_root}/build/chip/tests.gni")
 import("${dir_pw_unit_test}/test.gni")
-if (enable_efr32_build_per_suite || enable_efr32_build_per_testsource) {
+if (enable_efr32_build_per_testsource) {
   import("${efr32_sdk_build_root}/efr32_sdk.gni")
   import("${efr32_sdk_build_root}/silabs_executable.gni")
   #import("$dir_pw_protobuf_compiler/proto.gni")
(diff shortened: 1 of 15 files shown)


build/chip/chip_test_suite.gni
```diff
@@ -12,12 +12,28 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+enable_efr32_build_per_testsource = true  #++++ TODO: Figure out how to enable things based on `chip_device_platform == "efr32"` which is not currently in scope.
```
feasel0, line 15: Having this flag here is just a temporary solution, since I didn't yet figure out how to access the value of `chip_device_platform` in this scope in order to test if we're building for efr32.

build/chip/chip_test_suite.gni
```diff
@@ -12,12 +12,28 @@
...
+if (enable_efr32_build_per_testsource) {
+  import("${efr32_sdk_build_root}/efr32_sdk.gni")
+  import("${efr32_sdk_build_root}/silabs_executable.gni")
+  #import("$dir_pw_protobuf_compiler/proto.gni")
+
+  #++++ TODO: Figure out how to get access to these variables from src/test_driver/efr32/BUILD.gn so we don't have to set them explicitly here.
+  examples_plat_dir = "${chip_root}/examples/platform/silabs/efr32"
+  examples_common_plat_dir = "${chip_root}/examples/platform/silabs"
+  #efr32_project_dir = "${chip_root}/src/test_driver/efr32"
+}
```
feasel0, line 36: A bunch of stuff needs importing into here since it is used by `silabs_executable`.  Eventually I'll find a cleaner way to do this.

third_party/silabs/silabs_executable.gni
```diff
@@ -68,7 +68,7 @@ template("silabs_executable") {
   ]
   copy(flashing_runtime_target) {
     sources = flashing_script_inputs
-    outputs = [ "${root_out_dir}/{{source_file_part}}" ]
+    outputs = [ "${root_out_dir}/${target_name}--{{source_file_part}}" ]
```
feasel0, line 71: The "copy" action only needs to be done once per build, however now that `silabs_executable` is being called many times it ends up trying to copy these same python files repeatedly which causes an error.  I could not find a clean way of making it only copy once, so as a temporary fix I changed the names of the files to be something unique.  However ultimately this is not what we want, so I'll need to circle back and fix this properly.

build/chip/chip_test_suite.gni
```diff
@@ -111,6 +127,82 @@ template("chip_test_suite") {
...
+            ldflags = [
+              "-T" + rebase_path(ldscript, root_build_dir),
+              "-Wl,--no-warn-rwx-segment",
+            ]
+
+            output_dir = root_out_dir
+          }
+        }
+
         pw_test(_test_name) {
```
feasel0, line 206: The PW version of `chip_test_suite` template makes calls to `pw_test` (part of pigweed) for each test source instead of calling `chip_test` (part of our codebase).  Currently `chip_test` is only being used by `chip_test_suite_using_nltest`, so I presume it has not been updated to work with PW unit testing.  If `chip_test` were updated, we could make it so that `chip_test_suite` calls `chip_test` (instead of `pw_test`), and then I could probably move all of the `silabs_executable` stuff and the efr32 stuff into `chip_test` rather than `chip_test_suite`.

PR #33797, opened 2024-06-07
Generate LockUserChange events when clearing a credential.

bzbarsky-apple:
We were passing false for sendUserChangeEvent at all callsites of clearCredential(), which is not correct for the "clear a single credential" case.  The other cases generate their own LockUserChange events.

### DIFF
--- a/src/app/tests/suites/DL_UsersAndCredentials.yaml
+++ b/src/app/tests/suites/DL_UsersAndCredentials.yaml
@@ -2051,6 +2051,23 @@ tests:
               - name: "NextUserIndex"
                 value: 2
 
+    - label: "Make sure a LockUserChange event was generated"
+      command: "readEvent"
+      event: "LockUserChange"
+      # I wish there were a way to not hardcode this 25, but it's experimentally
+      # determined: doing a read without an eventNumber filter here shows 24
+      # LockUserChange events before this removal.
+      eventNumber: 25
+      response:
+          value:
+              {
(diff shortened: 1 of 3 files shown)


src/app/tests/suites/DL_UsersAndCredentials.yaml
```diff
@@ -2051,6 +2051,22 @@ tests:
               - name: "NextUserIndex"
                 value: 2
 
+    - label: "Make sure a LockUserChange event was generated"
+      command: "readEvent"
+      event: "LockUserChange"
+      # I wish there were a way to not hardcode this 25, but it's experimentally
+      # determined: doing a read without an eventNumber filter here shows 24
+      # LockUserChange events before this removal.
```
andy31415, line 2059: Could we add a TODO/issue here? Maybe we could have some `reset event count` somewhere to make this a lot more predictable. 24/25 feels very much arbitrary.

andy31415 (approved):

src/app/tests/suites/DL_UsersAndCredentials.yaml
bzbarsky-apple: I can file an issue, but fundamentally the issue is that the Door Lock app does not implement any non-standard clusters that would let us do that..

I filed https://github.com/project-chip/connectedhomeip/issues/33806

bauerschwan (approved):

woody-apple (approved):

PR #33801, opened 2024-06-07
Auto-commissioner: support secondary network interface commissioning

DejinChen:
Ref to SEPC PR [Clarify Commissioner behavior to support multiple networks](https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/9083), added supports for secondary network interface commissioning.
#### Change overview
- CHIP-TOOL pairing command: pairing ble-wifi-thread.
- AutoCommissioner and CHIPDeviceController: If primary network (on endpoint 0) failed, disable it (by writing InterfaceEnabled attribute, require related PRs) and try to connect secondary network.

#### Related PRs
- added Wi-Fi interface enabled support (#33859)
- OpenThread: added Thread interface enabled support (#33860)

#### Test
- Tested CHIP-TOOL pairing ble-wifi-thread command on ESP32-C6, which both Wi-Fi driver and Thread driver are enabled.

### DIFF
--- a/examples/chip-tool/commands/pairing/Commands.h
+++ b/examples/chip-tool/commands/pairing/Commands.h
@@ -165,6 +165,14 @@ public:
     {}
 };
 
+class PairBleWiFiOrThread : public PairingCommand
+{
+public:
+    PairBleWiFiOrThread(CredentialIssuerCommands * credsIssuerConfig) :
+        PairingCommand("ble-wifi-thread", PairingMode::Ble, PairingNetworkType::WiFiOrThread, credsIssuerConfig)
+    {}
+};
+
 class PairSoftAP : public PairingCommand
 {
 public:
@@ -233,6 +241,7 @@ void registerCommandsPairing(Commands & commands, CredentialIssuerCommands * cre
         make_unique<PairCodeThread>(credsIssuerConfig),
         make_unique<PairBleWiFi>(credsIssuerConfig),
         make_unique<PairBleThread>(credsIssuerConfig),
+        make_unique<PairBleWiFiOrThread>(credsIssuerConfig),
         make_unique<PairSoftAP>(credsIssuerConfig),
         make_unique<PairAlreadyDiscovered>(credsIssuerConfig),
         make_unique<PairAlreadyDiscoveredByIndex>(credsIssuerConfig),
--- a/examples/chip-tool/commands/pairing/Commands.h
+++ b/examples/chip-tool/commands/pairing/Commands.h
@@ -69,6 +69,14 @@ public:
     {}
 };
 
+class PairCodeWiFiThread : public PairingCommand
+{
+public:
+    PairCodeWiFiThread(CredentialIssuerCommands * credsIssuerConfig) :
+        PairingCommand("code-wifi-thread", PairingMode::Code, PairingNetworkType::WiFiOrThread, credsIssuerConfig)
+    {}
+};
+
 class PairOnNetwork : public PairingCommand
 {
 public:
@@ -165,14 +173,6 @@ public:
     {}
 };
 
-class PairBleWiFiOrThread : public PairingCommand
-{
-public:
-    PairBleWiFiOrThread(CredentialIssuerCommands * credsIssuerConfig) :
-        PairingCommand("ble-wifi-thread", PairingMode::Ble, PairingNetworkType::WiFiOrThread, credsIssuerConfig)
-    {}
-};
-
 class PairSoftAP : public PairingCommand
 {
 public:
@@ -239,9 +239,9 @@ void registerCommandsPairing(Commands & commands, CredentialIssuerCommands * cre
         make_unique<PairCodePase>(credsIssuerConfig),
         make_unique<PairCodeWifi>(credsIssuerConfig),
         make_unique<PairCodeThread>(credsIssuerConfig),
+        make_unique<PairCodeWiFiThread>(credsIssuerConfig),
         make_unique<PairBleWiFi>(credsIssuerConfig),
         make_unique<PairBleThread>(credsIssuerConfig),
-        make_unique<PairBleWiFiOrThread>(credsIssuerConfig),
         make_unique<PairSoftAP>(credsIssuerConfig),
         make_unique<PairAlreadyDiscovered>(credsIssuerConfig),
         make_unique<PairAlreadyDiscoveredByIndex>(credsIssuerConfig),
--- a/src/controller/AutoCommissioner.cpp
+++ b/src/controller/AutoCommissioner.cpp
@@ -298,6 +298,19 @@ CommissioningStage AutoCommissioner::GetNextCommissioningStage(CommissioningStag
 
 CommissioningStage AutoCommissioner::GetNextCommissioningStageNetworkSetup(CommissioningStage currentStage, CHIP_ERROR & lastErr)
 {
+    if (IsSecondaryNetworkSupported())
+    {
+        if (IsTriedSecondaryNetwork())
+        {
+            // Try secondary network interface.
+            return mDeviceCommissioningInfo.network.wifi.endpoint == kRootEndpointId ? CommissioningStage::kThreadNetworkSetup
+                                                                                     : CommissioningStage::kWiFiNetworkSetup;
+        }
+        // Try primary network interface
+        return mDeviceCommissioningInfo.network.wifi.endpoint == kRootEndpointId ? CommissioningStage::kWiFiNetworkSetup
+                                                                                 : CommissioningStage::kThreadNetworkSetup;
+    }
+
     if (mParams.GetWiFiCredentials().HasValue() && mDeviceCommissioningInfo.network.wifi.endpoint != kInvalidEndpointId)
     {
         return CommissioningStage::kWiFiNetworkSetup;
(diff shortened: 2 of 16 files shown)


examples/chip-tool/commands/pairing/Commands.h
```diff
@@ -165,6 +165,14 @@ class PairBleThread : public PairingCommand
     {}
 };
 
+class PairBleWiFiOrThread : public PairingCommand
```
bzbarsky-apple, line 168: The "ble" commands are basically deprecated.  People should be using the "code" commands.  So this should really add "code-wifi-thread" if we add something here.  If we strongly feel that ble-wifi-thread is also useful, we can add that too, but the "code" version is required.

src/controller/AutoCommissioner.cpp
```diff
@@ -298,6 +298,19 @@ CommissioningStage AutoCommissioner::GetNextCommissioningStage(CommissioningStag
 
 CommissioningStage AutoCommissioner::GetNextCommissioningStageNetworkSetup(CommissioningStage currentStage, CHIP_ERROR & lastErr)
 {
+    if (IsSecondaryNetworkSupported())
+    {
+        if (IsTriedSecondaryNetwork())
```
bzbarsky-apple, line 303:
```suggestion
        if (HaveTriedSecondaryNetwork())
```

src/controller/AutoCommissioner.cpp
```diff
@@ -298,6 +298,19 @@ CommissioningStage AutoCommissioner::GetNextCommissioningStage(CommissioningStag
...
+        {
+            // Try secondary network interface.
```
bzbarsky-apple, line 305: I don't get it.  The test above passed, so we _have_ tried it already....  Unless the naming is just completely off somehow?

src/controller/AutoCommissioner.cpp
```diff
@@ -729,6 +728,16 @@ CHIP_ERROR AutoCommissioner::CommissioningStepFinished(CHIP_ERROR err, Commissio
                 report.stageCompleted = CommissioningStage::kScanNetworks;
             }
         }
+
+        if (err != CHIP_NO_ERROR && IsSecondaryNetworkSupported() && !IsTriedSecondaryNetwork() &&
```
bzbarsky-apple, line 732: OK, so maybe that boolean should be called `TryingSecondaryNetwork()`?

src/controller/AutoCommissioner.h
```diff
@@ -94,6 +94,21 @@ class AutoCommissioner : public CommissioningDelegate
...
+    // for wifi and thread.
+    bool IsSecondaryNetworkSupported() const
+    {
+        return ((mParams.GetSupportsConcurrentConnection().ValueOr(false) && mParams.GetWiFiCredentials().HasValue() &&
+                 mDeviceCommissioningInfo.network.wifi.endpoint != kInvalidEndpointId) &&
+                mParams.GetThreadOperationalDataset().HasValue() &&
+                mDeviceCommissioningInfo.network.thread.endpoint != kInvalidEndpointId);
+    }
+
+    void SetTrySecondaryNetwork() { mAttemptedSecondaryNetwork = true; }
```
bzbarsky-apple, line 108: OK, so where does mAttemptedSecondaryNetwork (needs to be named better) get set to false again?

Because it sure looks like the first commissioning will work right, but the second will fail to configure networks properly.

src/controller/CHIPDeviceController.cpp
```diff
@@ -2740,6 +2747,18 @@ DeviceCommissioner::SendCommissioningCommand(DeviceProxy * device, const Request
...
 
+template <typename AttrType>
+CHIP_ERROR DeviceCommissioner::SendCommissioningWriteRequest(DeviceProxy * device, EndpointId endpoint, ClusterId cluster,
+                                                             AttributeId attribute, const AttrType & requestData,
+                                                             WriteResponseSuccessCallback successCb,
+                                                             WriteResponseFailureCallback failureCb)
+{
+    auto onSuccessCb = [this, successCb](const app::ConcreteAttributePath & aPath) { successCb(this); };
+    auto onFailureCb = [this, failureCb](const app::ConcreteAttributePath * aPath, CHIP_ERROR aError) { failureCb(this, aError); };
+    return WriteAttribute(device->GetSecureSession().Value(), endpoint, cluster, attribute, requestData, onSuccessCb, onFailureCb,
```
bzbarsky-apple, line 2758: This write needs to be cancelable, so we don't get problems if someone stops the commissioning and starts a new one while this is in flight, no?

src/controller/CHIPDeviceController.cpp
```diff
@@ -2740,6 +2747,18 @@ DeviceCommissioner::SendCommissioningCommand(DeviceProxy * device, const Request
...
+                          NullOptional, nullptr, NullOptional);
```
bzbarsky-apple, line 2759: Those last three args are pretty inscrutable from here and could use some `/* argName = */` comments on them.

src/controller/CHIPDeviceController.cpp
```diff
@@ -3424,6 +3443,26 @@ void DeviceCommissioner::PerformCommissioningStep(DeviceProxy * proxy, Commissio
         );
     }
     break;
+    case CommissioningStage::kPrimaryOperationalNetworkFailed: {
+        // nothing to do. This stage indicates that the primary operation network failed and the network interface should be
```
bzbarsky-apple, line 3447:
```suggestion
        // nothing to do. This stage indicates that the primary operational network failed and the network interface should be
```

src/controller/CHIPDeviceController.cpp
```diff
@@ -3424,6 +3443,26 @@ void DeviceCommissioner::PerformCommissioningStep(DeviceProxy * proxy, Commissio
...
+        // disabled later.
+        break;
+    }
+    case CommissioningStage::kDisablePrimaryNetworkInterface: {
+        NetworkCommissioning::Attributes::InterfaceEnabled::TypeInfo::Type request = false;
+        CHIP_ERROR err                                                             = SendCommissioningWriteRequest(
+            proxy, endpoint, NetworkCommissioning::Attributes::InterfaceEnabled::TypeInfo::GetClusterId(),
+            NetworkCommissioning::Attributes::InterfaceEnabled::TypeInfo::GetAttributeId(), request,
```
bzbarsky-apple, line 3455: Why not just write `NetworkCommissioning::Id` and `NetworkCommissioning::Attributes::InterfaceEnabled::Id`?  Would be shorter.  If you were templating on the typeinfo it would make sense to use the getters on it, but you're not.

src/controller/CommissioningDelegate.cpp
```diff
@@ -136,6 +136,12 @@ const char * StageToString(CommissioningStage stage)
     case kNeedsNetworkCreds:
         return "NeedsNetworkCreds";
 
+    case kPrimaryOperationalNetworkFailed:
+        return "kPrimaryOperationalNetworkFailed";
```
bzbarsky-apple, line 140:
```suggestion
        return "PrimaryOperationalNetworkFailed";
```

src/controller/CommissioningDelegate.cpp
```diff
@@ -136,6 +136,12 @@ const char * StageToString(CommissioningStage stage)
...
+
+    case kDisablePrimaryNetworkInterface:
+        return "kDisablePrimaryNetworkInterface";
```
bzbarsky-apple, line 143:
```suggestion
        return "DisablePrimaryNetworkInterface";
```

src/platform/ESP32/NetworkCommissioningDriver.cpp
```diff
@@ -147,6 +147,11 @@ CHIP_ERROR ESPWiFiDriver::CommitConfiguration()
 CHIP_ERROR ESPWiFiDriver::RevertConfiguration()
 {
     mStagingNetwork = mSavedNetwork;
+    if (!GetEnabled())
```
bzbarsky-apple, line 150: Please split out the server-side changes into a separate PR from the controller/client changes.

Possibly separate PRs for ESP32 and OpenThread, since they need to be reviewed by different people.

bzbarsky-apple (changes requested):
Changes requested largely for the non-cancelable write bits; the other issues need fixing too, but that one is critical.

src/controller/CHIPDeviceController.cpp
DejinChen: Write operation here is used for writing `interfaceEnabled` attribute of commissionee to false. The commissionee will revert network to the previous state and reset  `interfaceEnabled` attribute to true, when fail-safe timer is expired or commissionee has been rebooted with the fail-safe armed. 
BUT I am not clear that how to make the write operation cancelable. Any specific suggestions?

examples/chip-tool/commands/pairing/Commands.h
DejinChen: Added `code-wifi-thread` command.

src/controller/AutoCommissioner.cpp
DejinChen: Done.

src/controller/AutoCommissioner.h
DejinChen: Added reset for the flag.

src/controller/CHIPDeviceController.cpp
DejinChen: Done.

src/platform/ESP32/NetworkCommissioningDriver.cpp
DejinChen: Ok, will add PRs for server-side changes.

src/controller/CHIPDeviceController.cpp
DejinChen: Done.

src/controller/CHIPDeviceController.cpp
bzbarsky-apple: See how command invokes work in CHIPDeviceController.

Basically: if commissioning is stopped, we need to ensure that callbacks from this write do not reach the CHIPDeviceController after that point.

DejinChen: Added write request cancel function.

bzbarsky-apple (approved):
Thank you!

jmartinez-silabs (approved):

PR #33809, opened 2024-06-07
Fix OperationalSessionSetup notifying success with inactive sessions.

bzbarsky-apple:
When a success callback marks the session defunct for some reason, the other succcess callbacks should probably not be called.

### DIFF
--- a/src/app/OperationalSessionSetup.cpp
+++ b/src/app/OperationalSessionSetup.cpp
@@ -420,7 +420,7 @@ void OperationalSessionSetup::NotifyConnectionCallbacks(Cancelable & failureRead
             // fact have an active session for them, and if they try to put the
             // session in a holder that will fail, and then trying to use the
             // holder as if it has a session will crash.  But we should also try
-            // to make sure we don't call failure callbacks the correspond to
+            // to make sure we don't call failure callbacks that correspond to
             // the success callbacks we already called, since their context
             // might be dead.
             //
--- a/src/app/OperationalSessionSetup.cpp
+++ b/src/app/OperationalSessionSetup.cpp
@@ -396,9 +396,81 @@ void OperationalSessionSetup::NotifyConnectionCallbacks(Cancelable & failureRead
                                                         System::Clock::Milliseconds16 requestedBusyDelay)
 {
     //
-    // If we encountered no error, go ahead and call all success callbacks. Otherwise,
-    // call the failure callbacks.
+    // Go ahead and call our callbacks.  We want to walk through all the lists,
+    // because we need to Cancel() each callback no matter what.  Which
+    // callbacks we then call depends on whether we have encountered an error.
     //
+    while (successReady.mNext != &successReady)
+    {
+        // We expect that we only have callbacks if we are not performing just address update.
+        VerifyOrDie(!performingAddressUpdate);
+        Callback::Callback<OnDeviceConnected> * cb = Callback::Callback<OnDeviceConnected>::FromCancelable(successReady.mNext);
+
+        cb->Cancel();
+        if (error == CHIP_NO_ERROR)
+        {
+            VerifyOrDie(exchangeMgr);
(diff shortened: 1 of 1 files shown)


yunhanw-google (approved):

src/app/OperationalSessionSetup.cpp
```diff
@@ -396,9 +396,81 @@ void OperationalSessionSetup::NotifyConnectionCallbacks(Cancelable & failureRead
...
+            // case, and the session is active.
+            VerifyOrDie(optionalSessionHandle.Value()->AsSecureSession()->IsActiveSession());
+            cb->mCall(cb->mContext, *exchangeMgr, optionalSessionHandle.Value());
+
+            // That call might have made the session inactive.  If it did, then
+            // we should not call any more success callbacks, since we do not in
+            // fact have an active session for them, and if they try to put the
+            // session in a holder that will fail, and then trying to use the
+            // holder as if it has a session will crash.  But we should also try
+            // to make sure we don't call failure callbacks the correspond to
```
jtung-apple, line 423:
```suggestion
            // to make sure we don't call failure callbacks that correspond to
```

jtung-apple (approved):

src/app/OperationalSessionSetup.cpp
```diff
@@ -396,9 +396,81 @@ void OperationalSessionSetup::NotifyConnectionCallbacks(Cancelable & failureRead
...
+            // kinds of error callbacks.  But what we should probably do is just
+            // remove the first callback that has the same context as the
+            // callback we just called.  And if there is more than one that
+            // matches, clearly the API consumer was not expecting to delete
+            // itself when called, so we're OK with only dropping the first one.
+            //
+            // This would all be simpler if we just had a single callback per
+            // API consumer that communicated success-or-failure in a single
+            // call...
+            for (auto * failureCancelable = failureReady.mNext; failureCancelable != &failureReady;
```
jtung-apple, line 440: I'm wondering if `Connect()` could just wrap `mContext` in some `mContextHolder`, so the `onConnection` and `onFailure` are strongly associated. Then your logic here would find the exact failure callback every time.

jtung-apple: This would guard against API consumer that uses the same mContext but do not consistently specify a non-null `onFailure`

bzbarsky-apple: The problem is that it would need to allocate to do that, right?   And we don't want this code to be allocating....

jtung-apple: I think the question I'd like to ask is, would we expect a potential API consumer to do this:
1. Call Connect with only `onConnection`
2. Call it again with both `onConnection` and `onFailure` with same context
3. When the first `onConnection` gets called it kills the session
4. Now the `onFailure` would get removed, unintentionally

If we think this is likely then we should probably have some logic to guard against this situation. Otherwise I'm fine with proceeding with this PR.

bzbarsky-apple: None of our internal callers do that.  But general SDK API consumers could do whatever...

That said, calling twice with the same context would be pretty odd: how do you tell the callbacks apart?  And if you don't need to, why are you calling it twice?

ksperling-apple: Ideally we would have a list of callback holders, where each holder contains the success and failure callbacks that belong together... I'm not sure we can really rely on the context being the same for the success and failure callbacks, or even being non-null (these may well be true for current SDK usages, but it seems like a fraught assumption to make in general.)

ksperling-apple: I've got a solution that doesn't rely on the context values, I'll put a PR up later.

bzbarsky-apple:
Closing in favor of https://github.com/project-chip/connectedhomeip/pull/33822; that one is much better.

PR #33811, opened 2024-06-07
[ICD] Add ClientType support to the ICDManager and ICD Management cluster

mkardous-silabs:
#### Description
PR adds support for the ClientType in the ICDManager and the ICD Management cluster.

- Adds the ClientType to the RegisterClient command & the RegisteredClients attributes
- Adds the ClientType validation tothe ICDManager when checking for ICD Check-In messages
- Updates the ICDM 3.1 test plan to match test plan update
- Update to unit tests for the ClientType

Spec PR : 

- https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/8871
- https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/9315.
- https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/9322

Test Plan PR to adapt ICDM 3.1 test case : 

- https://github.com/CHIP-Specifications/chip-test-plans/pull/4200

#### Drive by changes :
PairingCommand used the wrong node-id for the ICDClientInfo storage. Modified the command sending to use to use the CheckInNodeID.

#### Note to reviewers
PR seems very large but isn't in reality. Since the XML was updates, the generated files updates is present in this PR as well.

#### Tests
- Unit tests
- Integration tests
- Manual tests to validate no Check-In message is sent out


src/app/clusters/icd-management-server/icd-management-server.cpp
```diff
@@ -252,6 +253,7 @@ Status ICDManagementServer::RegisterClient(CommandHandler * commandObj, const Co
     FabricIndex fabricIndex            = commandObj->GetAccessingFabricIndex();
     NodeId nodeId                      = commandData.checkInNodeID;
     uint64_t monitoredSubject          = commandData.monitoredSubject;
+    ClientTypeEnum clientType          = commandData.clientType;
```
bzbarsky-apple, line 256: Something here needs to check that this is a valid ClientTypeEnum value, no?

src/app/icd/server/ICDCheckInSender.cpp
```diff
@@ -27,10 +27,14 @@ namespace app {
 
 using namespace Protocols::SecureChannel;
 
-ICDCheckInSender::ICDCheckInSender(Messaging::ExchangeManager * exchangeManager)
+ICDCheckInSender::ICDCheckInSender(Messaging::ExchangeManager * exchangeManager, AddressResolve::Resolver * addressResolver)
```
bzbarsky-apple, line 30: These changes for the address resolver seem unrelated to ClientType.  Shouldn't they be in a separate PR?

src/app/icd/server/ICDManager.cpp
```diff
@@ -187,6 +192,13 @@ void ICDManager::SendCheckInMsgs()
                 continue;
             }
 
+            if (entry.clientType == ClientTypeEnum::kEphemeral)
+            {
+                // If the registered client is ephemeral, do not send a Check-In message
+                // continue to next entry
```
bzbarsky-apple, line 198: I see no support for this behavior in the spec.  Spec says to send check-ins to all registered clients, regardless of type, no?

src/app/tests/suites/TestIcdManagementCluster.yaml
```diff
@@ -208,6 +210,8 @@ tests:
                 value: 1001
               - name: "Key"
                 value: "\x10\x11\x12\x13\x14\x15\x16\x17\x18\x19\x1a\x1b\x1c\x1d\x1e\x1f\xff"
+              - name: "ClientType"
+                value: 0
```
bzbarsky-apple, line 214: Please use symbolic enum names in the YAML.  Much more readable.  So `value: ClientType.Permanent` or `value: ClientType.Ephemeral`.

src/app/zap-templates/zcl/data-model/chip/icd-management-cluster.xml
```diff
@@ -39,17 +39,24 @@ limitations under the License.
         <field name="AppDefinedButton" mask="0x10000"/>
     </bitmap>
 
+    <enum name="ClientTypeEnum" type="enum8">
```
bzbarsky-apple, line 42: Spec does not say this is an enum8.  It says nothing about the sizing.  Needs to be fixed.

src/app/icd/server/ICDManager.cpp
mkardous-silabs: A sentence was added to say that an ICD should not send Check-In messages to Ephemeral clients.
https://github.com/CHIP-Specifications/connectedhomeip-spec/blob/b3652909bdf595c7634cea1da0f1e58f837ea35f/src/data_model/ICDBehavior.adoc?plain=1#L223

src/app/zap-templates/zcl/data-model/chip/icd-management-cluster.xml
mkardous-silabs: Agreed, opened a spec issue for this last week 
https://github.com/CHIP-Specifications/connectedhomeip-spec/issues/9281

src/app/tests/suites/TestIcdManagementCluster.yaml
mkardous-silabs: done

src/app/icd/server/ICDCheckInSender.cpp
mkardous-silabs: done

examples/chip-tool/commands/clusters/ClusterCommand.h
```diff
@@ -68,8 +69,9 @@ class ClusterCommand : public InteractionModelCommands, public ModelCommand, pub
                            const chip::app::Clusters::IcdManagement::Commands::RegisterClient::Type & value)
     {
         ReturnErrorOnFailure(InteractionModelCommands::SendCommand(device, endpointId, clusterId, commandId, value));
-        mScopedNodeId     = chip::ScopedNodeId(value.checkInNodeID, device->GetSecureSession().Value()->GetFabricIndex());
+        mScopedNodeId     = chip::ScopedNodeId(device->GetDeviceId(), device->GetSecureSession().Value()->GetFabricIndex());
```
yunhanw-google, line 72: we need to use checkin nodeId instead of peer nodeId, since we may register different nodeId for LIT.

examples/chip-tool/commands/clusters/ClusterCommand.h
```diff
@@ -148,6 +150,7 @@ class ClusterCommand : public InteractionModelCommands, public ModelCommand, pub
                 clientInfo.peer_node         = mScopedNodeId;
                 clientInfo.monitored_subject = mMonitoredSubject;
                 clientInfo.start_icd_counter = value.ICDCounter;
+                clientInfo.client_type       = mClientType;
```
yunhanw-google, line 153: We should not store client type in icd clientInfo, since application knows the client type, and we don't receive any check-in message for ephemeral, i cannot figure out when and how to use clientInfo for which purpose..

yunhanw-google (changes requested):

src/app/icd/server/ICDManager.cpp
yunhanw-google: I think we should still allow to send check-in message when manually wake-up device so that controller, particularly mobile controller, has the way to know the LIT becomes active, and trigger the customized strategy if controller implements it, for example, send several stayActive command for x seconds, for example, user uses mobile controller to interact with one sensor in tough place, they may wanna interact with this lit for longer period instead of manually wake-up back and forth..

yunhanw-google: @bzbarsky-apple @mkardous-silabs @turon

examples/chip-tool/commands/clusters/ClusterCommand.h
mkardous-silabs: In that case, what is stored during the commissioning registration is incorrect - the ICDs NodeId was being stored.

I'll revert this changes and fix the PairingCommand bits.

examples/chip-tool/commands/clusters/ClusterCommand.h
mkardous-silabs: I am not sure i agree with the statement.
Not persisting the ClientType forces an all or nothing scenario. A client will be only be able to use one type. If it reboots, the only way it knows the type it used at registration if it is not persisted is by always using the same type.

The registration ClientType affects the assumptions a client can do ie. is it garanteed a check-in message will be sent or not and as such we should make sure this information is always available.

> i cannot figure out when and how to use clientInfo for which purpose.

If you do not believe it is necessary, nothing prevents you from not using the information but we shouldn't assume that others might not find a benefit of having this information persisted. It is only a uint8 so the cost is pretty minimal and can enable several use-case we might have missed.

src/app/icd/server/ICDManager.cpp
mkardous-silabs: How do you differentiate between a user using his device for normal operations vs. triggering the user active mode trigger do interact from the mobile only controller?

examples/chip-tool/commands/clusters/ClusterCommand.h
yunhanw-google: yes, please fix commissioning registration and allow to set the other nodeId, thanks

mkardous-silabs: done

src/app/zap-templates/zcl/data-model/chip/icd-management-cluster.xml
mkardous-silabs: https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/9322

src/app/clusters/icd-management-server/icd-management-server.cpp
mkardous-silabs: done

src/app/icd/client/DefaultICDClientStorage.cpp
```diff
@@ -274,6 +274,10 @@ CHIP_ERROR DefaultICDClientStorage::Load(FabricIndex fabricIndex, std::vector<IC
         memcpy(clientInfo.hmac_key_handle.AsMutable<Crypto::Symmetric128BitsKeyByteArray>(), hmacBuf.data(),
                sizeof(Crypto::Symmetric128BitsKeyByteArray));
 
+        // ClientType
+        ReturnErrorOnFailure(reader.Next(TLV::ContextTag(ClientInfoTag::kClientType)));
+        ReturnErrorOnFailure(reader.Get(clientInfo.client_type));
```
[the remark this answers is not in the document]
mkardous-silabs: Since the feauture is still provisionnal, i would say yes. Once LIT is officially out, it won't be.

examples/chip-tool/commands/clusters/ClusterCommand.h
yunhanw-google: Ack, thanks

src/app/icd/client/DefaultICDClientStorage.h
```diff
@@ -130,6 +130,7 @@ class DefaultICDClientStorage : public ICDClientStorage
         kMonitoredSubject = 5,
         kAesKeyHandle     = 6,
         kHmacKeyHandle    = 7,
+        kClientType       = 8,
```
yunhanw-google, line 133: please also update MaxICDClientInfoSize in https://github.com/project-chip/connectedhomeip/blob/master/src/app/icd/client/DefaultICDClientStorage.h#L159

src/app/clusters/icd-management-server/icd-management-server.cpp
```diff
@@ -252,10 +253,14 @@ Status ICDManagementServer::RegisterClient(CommandHandler * commandObj, const Co
...
     ByteSpan key                       = commandData.key;
     Optional<ByteSpan> verificationKey = commandData.verificationKey;
     bool isClientAdmin                 = false;
 
+    // Check if ClientType is valid
+    VerifyOrReturnError(clientType < ClientTypeEnum::kUnknownEnumValue, InteractionModel::Status::ConstraintError);
```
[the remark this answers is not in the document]
mkardous-silabs: Done - added a negative test also for the `kUnknownEnumValue`

examples/chip-tool/commands/pairing/PairingCommand.h
```diff
@@ -71,6 +71,8 @@ class PairingCommand : public CHIPCommand,
                     "The check-in node id for the ICD, default: node id of the commissioner.");
         AddArgument("icd-monitored-subject", 0, UINT64_MAX, &mICDMonitoredSubject,
                     "The monitored subject of the ICD, default: The node id used for icd-check-in-nodeid.");
+        AddArgument("icd-client-type", 0, 1, &mICDClientType,
+                    "The ClientType of the client regsitering, default: Permanent client - 0");
```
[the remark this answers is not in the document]
mkardous-silabs: done

examples/fabric-admin/commands/pairing/PairingCommand.h
```diff
@@ -70,6 +70,8 @@ class PairingCommand : public CHIPCommand,
                     "The check-in node id for the ICD, default: node id of the commissioner.");
         AddArgument("icd-monitored-subject", 0, UINT64_MAX, &mICDMonitoredSubject,
                     "The monitored subject of the ICD, default: The node id used for icd-check-in-nodeid.");
+        AddArgument("icd-client-type", 0, 1, &mICDClientType,
+                    "The ClientType of the client regsitering, default: Permanent client - 0");
```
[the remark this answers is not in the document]
mkardous-silabs: done

src/app/icd/client/DefaultICDClientStorage.h
mkardous-silabs: done

src/app/tests/suites/TestIcdManagementCluster.yaml
```diff
@@ -221,6 +225,23 @@ tests:
                 value: 1001
               - name: "Key"
                 value: "\x10\x11\x12\x13\x14\x15\x16\x17\x18\x19\x1a\x1b\x1c\x1d\x1e\x1f"
+              - name: "ClientType"
+                value: ClientTypeEnum.UnknownEnumValue
```
bzbarsky-apple, line 229: You could also consider sending some other out-of-range value if the YAML will let you (like 3 or something).  That should also result in ConstraintError.

bzbarsky-apple (approved):

yunhanw-google (approved):

jmartinez-silabs (approved):

src/app/tests/suites/TestIcdManagementCluster.yaml
mkardous-silabs: done

mkardous-silabs: @bzbarsky-apple When adding the test case with a value > UnknownEnumValue, yaml parsing fails since it checks the valid values of the enum. Couldn't add the test case with the yaml test.

This will need to go in the python test - will do in a follow up.

PR #33814, opened 2024-06-08
Enable ReadHandler to support large payloads

pidarped:
When sending reports, if the session established with the peer supports large payloads, the ReadHandler would allocate a large buffer to, potentially, fit more cluster data.

### DIFF
--- a/src/app/ReadHandler.cpp
+++ b/src/app/ReadHandler.cpp
@@ -921,14 +921,12 @@ void ReadHandler::ClearStateFlag(ReadHandlerFlags aFlag)
 
 size_t ReadHandler::GetReportBufferMaxSize()
 {
-    size_t maxBufSize                  = kMaxSecureSduLengthBytes;
     Transport::SecureSession * session = GetSession();
     if (session && session->AllowsLargePayload())
     {
-        maxBufSize = kMaxLargeSecureSduLengthBytes;
+        return kMaxLargeSecureSduLengthBytes;
     }
-
-    return maxBufSize;
+    return kMaxSecureSduLengthBytes;
 }
 
 } // namespace app
--- a/src/app/ReadHandler.cpp
+++ b/src/app/ReadHandler.cpp
@@ -919,5 +919,17 @@ void ReadHandler::ClearStateFlag(ReadHandlerFlags aFlag)
     SetStateFlag(aFlag, false);
 }
 
+size_t ReadHandler::GetReportBufferMaxSize()
+{
+    size_t maxBufSize                  = chip::app::kMaxSecureSduLengthBytes;
+    Transport::SecureSession * session = GetSession();
+    if (session && session->AllowsLargePayload())
+    {
+        maxBufSize = chip::app::kMaxLargeSecureSduLengthBytes;
+    }
+
+    return maxBufSize;
+}
+
 } // namespace app
 } // namespace chip
--- a/src/app/ReadHandler.cpp
+++ b/src/app/ReadHandler.cpp
@@ -921,11 +921,11 @@ void ReadHandler::ClearStateFlag(ReadHandlerFlags aFlag)
 
 size_t ReadHandler::GetReportBufferMaxSize()
 {
-    size_t maxBufSize                  = chip::app::kMaxSecureSduLengthBytes;
+    size_t maxBufSize                  = kMaxSecureSduLengthBytes;
     Transport::SecureSession * session = GetSession();
     if (session && session->AllowsLargePayload())
     {
-        maxBufSize = chip::app::kMaxLargeSecureSduLengthBytes;
+        maxBufSize = kMaxLargeSecureSduLengthBytes;
     }
 
     return maxBufSize;
--- a/src/app/ReadHandler.h
+++ b/src/app/ReadHandler.h
@@ -335,9 +335,10 @@ private:
 
     /*
      * Get the appropriate size of a packet buffer to allocate for encoding a Report message.
-     * Depending on the underlying session, which may or may not support large
-     * payloads, a buffer with the corresponding max size would be allocated.
+     * This size might depend on the underlying session used by the ReadHandler.
      *
+     * The size returned here is the size not including the various prepended headers
+     * (what System::PacketBuffer calls the "available size").
      */
     size_t GetReportBufferMaxSize();
 
--- a/src/app/ReadHandler.h
+++ b/src/app/ReadHandler.h
@@ -333,6 +333,14 @@ private:
      */
     CHIP_ERROR SendReportData(System::PacketBufferHandle && aPayload, bool aMoreChunks);
 
+    /*
+     * Get the appropriate size of a packet buffer to allocate for encoding a Report message.
+     * Depending on the underlying session, which may or may not support large
+     * payloads, a buffer with the corresponding max size would be allocated.
+     *
+     */
+    size_t GetReportBufferMaxSize();
+
     /**
      *  Returns whether this ReadHandler represents a subscription that was created by the other side of the provided exchange.
      */
--- a/src/app/reporting/Engine.cpp
+++ b/src/app/reporting/Engine.cpp
@@ -492,10 +492,11 @@ CHIP_ERROR Engine::BuildAndSendSingleReportData(ReadHandler * apReadHandler)
     CHIP_ERROR err = CHIP_NO_ERROR;
     chip::System::PacketBufferTLVWriter reportDataWriter;
     ReportDataMessage::Builder reportDataBuilder;
-    chip::System::PacketBufferHandle bufHandle = System::PacketBufferHandle::New(chip::app::kMaxSecureSduLengthBytes);
+    chip::System::PacketBufferHandle bufHandle = nullptr;
     uint16_t reservedSize                      = 0;
     bool hasMoreChunks                         = false;
     bool needCloseReadHandler                  = false;
+    size_t maxSduSize                          = 0;
 
     // Reserved size for the MoreChunks boolean flag, which takes up 1 byte for the control tag and 1 byte for the context tag.
     const uint32_t kReservedSizeForMoreChunksFlag = 1 + 1;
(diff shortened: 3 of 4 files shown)


src/app/ReadHandler.h
```diff
@@ -491,6 +491,9 @@ class ReadHandler : public Messaging::ExchangeDelegate
     /// @param aFlag Flag to clear
     void ClearStateFlag(ReadHandlerFlags aFlag);
 
+    void UseLargePayloadBuffer() { mUseLargePayloadBuffer = true; }
+    bool ShouldUseLargePayloadBuffer() { return mUseLargePayloadBuffer; }
```
andy31415, line 495: I wonder if instead of pushing a bool down the stack if maybe ReadHandler should instead have the ability to return correctly sized data instead.

Do we expect to need `ShouldUseLargePayloadBuffer` more than just creating a buffer?
Otherwise it may be that we could split logic a bit more by having `ReadHandler::AllocatePacketForReply` that returns a correctly sized buffer instead and localizes the size decisions.

src/app/reporting/Engine.cpp
```diff
@@ -512,11 +513,18 @@ CHIP_ERROR Engine::BuildAndSendSingleReportData(ReadHandler * apReadHandler)
 
     VerifyOrExit(apReadHandler != nullptr, err = CHIP_ERROR_INVALID_ARGUMENT);
     VerifyOrExit(apReadHandler->GetSession() != nullptr, err = CHIP_ERROR_INCORRECT_STATE);
+
+    if (apReadHandler->ShouldUseLargePayloadBuffer())
```
andy31415, line 517: How about:

```cpp
const size_t maxSduSize = apReadHandler->ShouldUseLargePayloadBuffer()
    ? chip::app::kMaxLargeSecureSduLengthBytes
    : chip::app::kMaxSecureSduLengthBytes;
```

To make the 2 constants clearer by having them close together (and also making the value constant)? also see note above: could we make ReadHandler have this logic instead of passing down a bool around?

src/app/reporting/Engine.cpp
```diff
@@ -492,7 +492,8 @@ CHIP_ERROR Engine::BuildAndSendSingleReportData(ReadHandler * apReadHandler)
     CHIP_ERROR err = CHIP_NO_ERROR;
     chip::System::PacketBufferTLVWriter reportDataWriter;
     ReportDataMessage::Builder reportDataBuilder;
-    chip::System::PacketBufferHandle bufHandle = System::PacketBufferHandle::New(chip::app::kMaxSecureSduLengthBytes);
+    size_t maxSduSize                          = chip::app::kMaxSecureSduLengthBytes;
+    chip::System::PacketBufferHandle bufHandle = nullptr;
```
andy31415, line 496: please declare this when initialized - having all variables at the top of the function is too much C-style API and hinders readability (we want the narrowest scope for variable as possible)

src/app/InteractionModelEngine.cpp
```diff
@@ -853,6 +853,11 @@ Protocols::InteractionModel::Status InteractionModelEngine::OnReadInitialRequest
         return Status::ResourceExhausted;
     }
 
+    if (apExchangeContext->GetSessionHandle()->AllowsLargePayload())
```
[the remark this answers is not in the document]
bzbarsky-apple: Yes, why are we doing this out-of-band instead of the ReadHandler getting this information itself?

src/app/ReadHandler.h
```diff
@@ -572,6 +575,8 @@ class ReadHandler : public Messaging::ExchangeDelegate
 
     // TODO (#27675): Merge all observers into one and that one will dispatch the callbacks to the right place.
     Observer * mObserver = nullptr;
+
+    bool mUseLargePayloadBuffer = false;
```
bzbarsky-apple, line 579: 1. Is this the right place to put this boolean, in terms of packing?
2. Are we ever going to be in a situation where we need to allocate the buffer but don't have an exchange or a session to query this state from?  In other words, do we need a member at all?

Note the RAM increase from this PR on constrained systems...  Ideally there would be no RAM increase.

src/app/InteractionModelEngine.cpp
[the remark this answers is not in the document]
pidarped: I was working with the model of storing the flag when the request came in and then using that whenever the response was being sent. I was using that model for the Command invocations.  
However, I noticed later that it keeps the SessionHandle as a member and that can be used to work inband during sending the response. I will change it to that.

src/app/ReadHandler.h
pidarped: These are now removed. The size inference is being done in-band within ReadHandler and returned by ReadHandler::GetReportBufferMaxSize()

src/app/reporting/Engine.cpp
pidarped: There is a VerifyOrExit(..) on the readHandler and so I cannot declare this after that where it is being assigned from a ReadHandler function.

src/app/ReadHandler.h
pidarped: boolean removed. The sessionHandle stored as a member in ReadHandler is used to fetch the correct size in ReadHandler.

src/app/reporting/Engine.cpp
pidarped: ReadHandler now provides a function to infer this internally from the member session handle.

src/app/ReadHandler.cpp
```diff
@@ -919,5 +919,17 @@ void ReadHandler::ClearStateFlag(ReadHandlerFlags aFlag)
     SetStateFlag(aFlag, false);
 }
 
+size_t ReadHandler::GetReportBufferMaxSize()
+{
+    size_t maxBufSize                  = chip::app::kMaxSecureSduLengthBytes;
```
bzbarsky-apple, line 924: This is all inside chip::app, no?
```suggestion
    size_t maxBufSize                  = kMaxSecureSduLengthBytes;
```

src/app/ReadHandler.cpp
```diff
@@ -919,5 +919,17 @@ void ReadHandler::ClearStateFlag(ReadHandlerFlags aFlag)
...
+    Transport::SecureSession * session = GetSession();
+    if (session && session->AllowsLargePayload())
+    {
+        maxBufSize = chip::app::kMaxLargeSecureSduLengthBytes;
```
bzbarsky-apple, line 928:
```suggestion
        maxBufSize = kMaxLargeSecureSduLengthBytes;
```

src/app/ReadHandler.h
```diff
@@ -333,6 +333,14 @@ class ReadHandler : public Messaging::ExchangeDelegate
      */
     CHIP_ERROR SendReportData(System::PacketBufferHandle && aPayload, bool aMoreChunks);
 
+    /*
+     * Get the appropriate size of a packet buffer to allocate for encoding a Report message.
+     * Depending on the underlying session, which may or may not support large
+     * payloads, a buffer with the corresponding max size would be allocated.
+     *
```
bzbarsky-apple, line 340:
```suggestion
     * Get the appropriate size of a packet buffer to allocate for encoding a Report message.
     * This size might depend on the underlying session used by the ReadHandler.
     *
     * The size returned here is the size not including the various prepended headers
     * (what System::PacketBuffer calls the "available size").
```

src/app/reporting/Engine.cpp
```diff
@@ -492,10 +492,11 @@ CHIP_ERROR Engine::BuildAndSendSingleReportData(ReadHandler * apReadHandler)
     CHIP_ERROR err = CHIP_NO_ERROR;
     chip::System::PacketBufferTLVWriter reportDataWriter;
     ReportDataMessage::Builder reportDataBuilder;
-    chip::System::PacketBufferHandle bufHandle = System::PacketBufferHandle::New(chip::app::kMaxSecureSduLengthBytes);
+    chip::System::PacketBufferHandle bufHandle = nullptr;
     uint16_t reservedSize                      = 0;
     bool hasMoreChunks                         = false;
     bool needCloseReadHandler                  = false;
+    size_t maxSduSize                          = 0;
```
bzbarsky-apple, line 499: What is an "Sdu"

This should be something like `reportBufferMaxSize` or something.

src/app/reporting/Engine.cpp
```diff
@@ -512,11 +513,17 @@ CHIP_ERROR Engine::BuildAndSendSingleReportData(ReadHandler * apReadHandler)
 
     VerifyOrExit(apReadHandler != nullptr, err = CHIP_ERROR_INVALID_ARGUMENT);
     VerifyOrExit(apReadHandler->GetSession() != nullptr, err = CHIP_ERROR_INCORRECT_STATE);
+
+    // Depending on whether the session supports large payload or not, the
+    // appropriate  max size would be returned for the Report buffer.
```
bzbarsky-apple, line 518: Why do we care about that here?  We ask the ReadHandler for the size to use, it gives us a size, we use it.

On the other hand, the comment on line 535 (after your changes; 528 before your changes) needs fixing to probably talk about fitting in the size the ReadHandler returned or something.

src/app/reporting/Engine.cpp
pidarped: Will rename. I went with how the constant was defined in StatusResponse.h. I guess SDU stands for Service Data Unit, as is the canonical defintion, similar to PDU etc. Not sure.
This definition in StatusResponse.h also seems redundant and is essentially the same as one of the constants in MessageHeader.h. I felt that we probably need to consolidate the definitions into a single place(at least not leave them strewn around in various locations) but did not want to do that in this PR.
StatusResponse.h does not seem like the ideal place for this definition too, isn't it?

src/app/reporting/Engine.cpp
pidarped: Felt that I needed to add a comment above this call but I guess these details are unnecessary here. Will fix the other comment as well.

bzbarsky-apple (approved):

src/app/ReadHandler.cpp
```diff
@@ -919,5 +919,17 @@ void ReadHandler::ClearStateFlag(ReadHandlerFlags aFlag)
...
+size_t ReadHandler::GetReportBufferMaxSize()
+{
+    size_t maxBufSize                  = kMaxSecureSduLengthBytes;
+    Transport::SecureSession * session = GetSession();
+    if (session && session->AllowsLargePayload())
+    {
+        maxBufSize = kMaxLargeSecureSduLengthBytes;
+    }
+
+    return maxBufSize;
```
andy31415, line 931:
```suggestion
    Transport::SecureSession * session = GetSession();
    if (session && session->AllowsLargePayload())
    {
        return kMaxLargeSecureSduLengthBytes;
    }
    return kMaxSecureSduLengthBytes;
```

This is to reduce the usage and scope of a local variable. Otherwise when I see the final return, I need to go up the function to read the initial assignment.

andy31415 (approved):

PR #33822, opened 2024-06-10
Fix OperationalSessionSetup notifying success with inactive sessions

ksperling-apple:
When a success callback marks the session defunct for some reason, other succcess callbacks should not be called. Implement a GroupedCallbackList to make this logic possible.

This specific solution is based on the assumption that we don't want to change the `OperationalSessionSetup` API which takes success and two variants of failure callbacks as separate, client-provided `Callback` objects, with all callbacks being optional to provide. We also don't want to introduce additional dynamic allocation within `OperationalSessionSetup` e.g. to allocate a struct holding the related callbacks.

The GroupedCallbackList class makes use of the existing prev/next pointers within the client-allocated `Callback` objects to capture the grouping relationship between them.

### DIFF
--- a/src/lib/core/GroupedCallbackList.h
+++ b/src/lib/core/GroupedCallbackList.h
@@ -1,0 +1,246 @@
+/*
+ *
+ *    Copyright (c) 2024 Project CHIP Authors
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+#pragma once
+
+#include <lib/core/CHIPCallback.h>
+#include <lib/support/CodeUtils.h>
+
+#include <utility>
+
+namespace chip {
+namespace Callback {
+
+namespace detail {
+// Internal helper functions
+template <size_t Index>
+void TaggedDequeueGroup(Cancelable * cancelable);
+void EnqueueWithGroup(Cancelable * cancelable, Cancelable *& group, Cancelable * anchor, void (*cancelFn)(Cancelable *));
+void LinkGroup(Cancelable * prev, Cancelable * next);
+inline Cancelable * ClearCancelable(Cancelable * cancelable);
+} // namespace detail
+
+/**
+ * A GroupedCallbackList manages a list of Callback objects (see CHIPCallback.h).
+ * The state of the list is maintained using the prev/next pointers of each Callback.
+ * Unlike a normal linked list where entries are managed individually, this class
+ * manages a number of related callbacks as a group, with the callback function types
+ * given as template parameters.
+ *
+ * For example, a `GroupedCallbackList<SuccessFn, FailureFn>` manages groups of a
+ * `Callback<SuccessFn>` and a `Callback<FailureFn>`.
+ *
+ * Groups of callbacks are enqueued and dequeued (or cancelled) as a unit.
+ * Within a group each callback is optional (i.e. can be null), however attempting
+ * to enqueue a group where all callbacks are null has no effect.
+ */
+template <typename... T>
+class GroupedCallbackList : protected Cancelable
+{
+public:
+    GroupedCallbackList() = default;
+    ~GroupedCallbackList() { Clear(); }
+
+    GroupedCallbackList(GroupedCallbackList const &)             = delete;
+    GroupedCallbackList & operator=(GroupedCallbackList const &) = delete;
+
+    bool IsEmpty() { return mNext == this; }
+
+    /**
+     * Enqueues the specified group of callbacks, any of which may be null.
+     */
+    void Enqueue(Callback<T> *... callback) { Enqueue(std::index_sequence_for<T...>{}, callback...); }
+
+    /**
+     * If the list is non-empty, populates the reference arguments with the first
+     * group of callbacks and returns true. Returns false if the list is empty.
+     */
+    bool Peek(Callback<T> *&... callback) const { return Peek(std::index_sequence_for<T...>{}, callback...); }
+
+    /**
+     * Like Peek(), but additionally removes the first group of callbacks from the list.
+     */
+    bool Take(Callback<T> *&... callback)
+    {
+        VerifyOrReturnValue(Peek(callback...), false);
+        mNext->Cancel();
+        return true;
+    }
+
+    /**
+     * Moves all elements of the source list into this list, leaving the source list empty.
+     */
+    void EnqueueTakeAll(GroupedCallbackList & source)
+    {
+        VerifyOrReturn(!source.IsEmpty() && this != &source);
+        detail::LinkGroup(mPrev, source.mNext);
+        source.mPrev->mNext = this;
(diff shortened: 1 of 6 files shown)


src/lib/core/GroupedCallbackList.h
```diff
@@ -0,0 +1,213 @@
...
+    //              |== Grouped Nodes ==|
+    //
+    //   +---+      +---+   +---+   +---+      +---+
+    //   |   |----->|   |-->|   |-->|   |----->|   |
+    //   |   |      |   |   |   |   |   |      |   |
+    //   |   |<-----|   |  -|   |  -|   |<-----|   |
+    //   +---+  \   +---+ / +---+ / +---+      +---+
+    //           \_______/       /
+    //            \_____________/
+    //
```
bzbarsky-apple, line 103: So in this picture, what happens when the grouped nodes are canceled?  I guess the cancel func this is using will remove the whole group if any member of the group gets canceled?

ksperling-apple: Correct, groups are enqueued and dequeued / cancelled as a unit. The cancel func is `DequeueGroup`.

src/lib/core/GroupedCallbackList.h
```diff
@@ -0,0 +1,227 @@
...
+inline Cancelable * ClearCancelable(Cancelable * cancelable);
+} // namespace detail
+
+/**
+ * A GroupedCallbackList is similar to a CallbackDeque, but manages groups of
+ * related callbacks (possibly of different types) in a single list.
+ *
+ * Groups of callbacks are enqueued and dequeued (or cancelled) as a unit.
+ * Within a group each callback is optional (i.e. can be null), however attempting
+ * to enqueuea group where all callbacks are null has no effect.
```
bzbarsky-apple, line 42:
```suggestion
 * to enqueue a group where all callbacks are null has no effect.
```

src/lib/core/GroupedCallbackList.h
```diff
@@ -0,0 +1,227 @@
...
+                }
+            }(),
+            ...);
+        return true;
+    }
+};
+
+namespace detail {
+
+inline void EnqueueWithGroup(Cancelable * cancelable, Cancelable *& group, Cancelable * anchor, void (*cancelFn)(Cancelable *))
```
bzbarsky-apple, line 169: This could use some documentation explaining what it's doing.  Same for LinkGroup and ClearCancelable.

src/lib/core/GroupedCallbackList.h
```diff
@@ -0,0 +1,227 @@
...
+
+    Cancelable * groupPrev = cancelable->mPrev;
+    do
+    {
+        cancelable->mPrev = prev;
+        cancelable        = cancelable->mNext;
+    } while (cancelable->mPrev == groupPrev);
+}
+
+inline Cancelable * ClearCancelable(Cancelable * cancelable)
```
bzbarsky-apple, line 198: Should this be called DequeueCancelable?  I guess it's not adjusting any pointers to `cancelable` from its next/prev... so maybe not.

src/lib/core/GroupedCallbackList.h
```diff
@@ -0,0 +1,227 @@
...
+    {
+        next = ClearCancelable(next);
+    } while (next->mPrev == prev);
+    LinkGroup(prev, next);
+}
+
+template <size_t Index>
+void TaggedDequeueGroup(Cancelable * cancelable)
+{
+    (void) Index; // not used, we only care that instantiations have unique addresses
```
bzbarsky-apple, line 220: Hmm.  Will they have unique addresses even in the face of identical code folding or whatnot?   I guess since we are taking the addresses of the functions the as-if rule should prevent them from being folded?

bzbarsky-apple (approved):

src/lib/core/GroupedCallbackList.h
ksperling-apple: Yeah it's specifically not messing with the pointed-to objects

src/lib/core/GroupedCallbackList.h
ksperling-apple: I believe taking account of whether addresses get taken is considered "safe" ICF, so I think the expectation is that compilers do this correctly. We do have the unit tests that will fail if it doesn't work.

src/lib/core/GroupedCallbackList.h
ksperling-apple: Done

src/lib/core/GroupedCallbackList.h
```diff
@@ -0,0 +1,231 @@
...
+namespace detail {
+template <size_t Index>
+void TaggedDequeueGroup(Cancelable * cancelable);
+void EnqueueWithGroup(Cancelable * cancelable, Cancelable *& group, Cancelable * anchor, void (*cancelFn)(Cancelable *));
+void LinkGroup(Cancelable * prev, Cancelable * next);
+inline Cancelable * ClearCancelable(Cancelable * cancelable);
+} // namespace detail
+
+/**
+ * A GroupedCallbackList is similar to a CallbackDeque, but manages groups of
```
andy31415, line 37: When reading documentation here, I do not actually yet know what a CallbackDequeue is. Please document it fully on why we use it and how.

src/lib/core/GroupedCallbackList.h
```diff
@@ -0,0 +1,231 @@
...
+
+#include <utility>
+
+namespace chip {
+namespace Callback {
+
+namespace detail {
+template <size_t Index>
+void TaggedDequeueGroup(Cancelable * cancelable);
+void EnqueueWithGroup(Cancelable * cancelable, Cancelable *& group, Cancelable * anchor, void (*cancelFn)(Cancelable *));
```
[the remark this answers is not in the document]
ksperling-apple: They are documented down near their implementation.

src/lib/core/GroupedCallbackList.h
ksperling-apple: `CallbackDeque` is the list class defined directly in `CHIPCallback.h` that this class is effectively replacing for certain use cases. I can remove the reference if it is more confusing than useful.

ksperling-apple: I've reworded things to avoid referring to `CallbackDeque` and updated the long comment block and ASCII diagram to explain the anchor element.

src/lib/core/GroupedCallbackList.h
```diff
@@ -0,0 +1,246 @@
...
+ *
+ * For example, a `GroupedCallbackList<SuccessFn, FailureFn>` manages groups of a
+ * `Callback<SuccessFn>` and a `Callback<FailureFn>`.
+ *
+ * Groups of callbacks are enqueued and dequeued (or cancelled) as a unit.
+ * Within a group each callback is optional (i.e. can be null), however attempting
+ * to enqueue a group where all callbacks are null has no effect.
+ */
+template <typename... T>
+class GroupedCallbackList : protected Cancelable
```
andy31415, line 52: Could you explain the reason of having this vs some `IntrusiveList<S>` where S in this case would be a struct containing 3 callbacks for our use case?

I see it tries hard to make sure list can be sparse, however given that when it is not sparse we pay the cost of 2 extra pointers (prev/next) for every entry, I am unsure if that is that much worth it.

andy31415 (approved):

src/lib/core/GroupedCallbackList.h
andy31415: I think it would be useful to describe that GroupedCallbackList makes use of the prev/next pointers already existing in Callback, hence it has very limited extra overhead to a regular callback (just itself as the anchor).

That would explain why use this instead of some struct.

ksperling-apple: This is really to avoid changing the existing OperationalSessionSetup API

PR #33827, opened 2024-06-10
Add SupportedCluster list to ContentAppPlatform

mthiesc:
[Problem]
The ContentAppPlatform does not know which clusters each ContentApp supports.

Currently the ContentApp is queried to get the passcode as long as the AccountLoginDelegate is present whether or not the installed ContentApp has declared support for AccountLogin cluster in its manifest.

[Solution]
Extend the native ContentApp with a SupportedCluster list. The list is initialized from the Android AppPlatformService whenever an installed ContentApp is discovered and added to the native AppPlatform.

This list is used to check if AccountLogin cluster is supported before querying the ContentApp for the passcode.

[Test]
The feature is tested end-to-end using a native Linux casting-app and and an Android platform-app and content-app. The content-app static_matter_clusters raw asset was manipulated to verify that the clusters are parsed correctly and that the passcode is only retrievable when the AccountLogin cluster is declared.

### DIFF
--- a/examples/tv-app/android/java/AppPlatform-JNI.cpp
+++ b/examples/tv-app/android/java/AppPlatform-JNI.cpp
@@ -18,12 +18,12 @@
 
 #include "AppImpl.h"
 
+#include <app/app-platform/ContentApp.h>
 #include <jni.h>
 #include <lib/core/CHIPError.h>
 #include <lib/support/CHIPJNIError.h>
 #include <lib/support/JniReferences.h>
 #include <lib/support/JniTypeWrappers.h>
-#include <app/app-platform/ContentApp.h>
 
 using namespace chip;
 using namespace chip::app;
@@ -36,7 +36,7 @@ using namespace chip::Credentials;
  */
 
 // Forward declaration
-std::vector<ContentApp::SupportedCluster> convert_to_cpp(JNIEnv* env, jobject supportedClusters);
+std::vector<ContentApp::SupportedCluster> convert_to_cpp(JNIEnv * env, jobject supportedClusters);
 
 #define JNI_METHOD(RETURN, METHOD_NAME)                                                                                            \
     extern "C" JNIEXPORT RETURN JNICALL Java_com_matter_tv_server_tvapp_AppPlatform_##METHOD_NAME
@@ -48,8 +48,8 @@ JNI_METHOD(void, nativeInit)(JNIEnv *, jobject app, jobject contentAppEndpointMa
 }
 
 JNI_METHOD(jint, addContentApp)
-(JNIEnv *, jobject, jstring vendorName, jint vendorId, jstring appName, jint productId, jstring appVersion, jobject supportedClusters,
- jobject manager)
+(JNIEnv *, jobject, jstring vendorName, jint vendorId, jstring appName, jint productId, jstring appVersion,
+ jobject supportedClusters, jobject manager)
 {
     chip::DeviceLayer::StackLock lock;
     JNIEnv * env = JniReferences::GetInstance().GetEnvForCurrentThread();
@@ -63,8 +63,8 @@ JNI_METHOD(jint, addContentApp)
 }
 
 JNI_METHOD(jint, addContentAppAtEndpoint)
-(JNIEnv *, jobject, jstring vendorName, jint vendorId, jstring appName, jint productId, jstring appVersion, jobject supportedClusters, jint endpointId,
- jobject manager)
+(JNIEnv *, jobject, jstring vendorName, jint vendorId, jstring appName, jint productId, jstring appVersion,
+ jobject supportedClusters, jint endpointId, jobject manager)
 {
     chip::DeviceLayer::StackLock lock;
     JNIEnv * env = JniReferences::GetInstance().GetEnvForCurrentThread();
@@ -72,8 +72,9 @@ JNI_METHOD(jint, addContentAppAtEndpoint)
     JniUtfString vName(env, vendorName);
     JniUtfString aName(env, appName);
     JniUtfString aVersion(env, appVersion);
-    EndpointId epId = AddContentApp(vName.c_str(), static_cast<uint16_t>(vendorId), aName.c_str(), static_cast<uint16_t>(productId),
-                                    aVersion.c_str(), convert_to_cpp(env, supportedClusters), static_cast<EndpointId>(endpointId), manager);
+    EndpointId epId =
+        AddContentApp(vName.c_str(), static_cast<uint16_t>(vendorId), aName.c_str(), static_cast<uint16_t>(productId),
+                      aVersion.c_str(), convert_to_cpp(env, supportedClusters), static_cast<EndpointId>(endpointId), manager);
     return static_cast<uint16_t>(epId);
 }
 
@@ -99,12 +100,15 @@ JNI_METHOD(void, addSelfVendorAsAdmin)
     AddSelfVendorAsAdmin();
 }
 
-std::vector<uint32_t> consume_and_convert_to_cpp(JNIEnv* env, jobject intArrayObject) {
+std::vector<uint32_t> consume_and_convert_to_cpp(JNIEnv * env, jobject intArrayObject)
+{
     std::vector<uint32_t> uintVector;
-    if (intArrayObject != nullptr) {
-        jsize length = env->GetArrayLength(static_cast<jintArray>(intArrayObject));
-        jint* elements = env->GetIntArrayElements(static_cast<jintArray>(intArrayObject), nullptr);
-        if (elements != nullptr) {
+    if (intArrayObject != nullptr)
+    {
+        jsize length    = env->GetArrayLength(static_cast<jintArray>(intArrayObject));
+        jint * elements = env->GetIntArrayElements(static_cast<jintArray>(intArrayObject), nullptr);
+        if (elements != nullptr)
+        {
             // OBS: Implicit type ambiguation from int32_t to uint32_t
             uintVector.assign(elements, elements + length);
             env->ReleaseIntArrayElements(static_cast<jintArray>(intArrayObject), elements, JNI_ABORT);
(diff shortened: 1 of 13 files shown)


src/app/app-platform/ContentApp.h
```diff
@@ -94,11 +96,30 @@ class ContentAppClientCommandSender
 class DLL_EXPORT ContentApp
 {
 public:
+
+    struct SupportedCluster {
+        int clusterIdentifier { -1 };
```
bzbarsky-apple, line 101: Is there a reason this is not using the ClusterId type and the relevant invalid value?

Because `int` cannot express all cluster ids....

src/app/app-platform/ContentApp.h
```diff
@@ -94,11 +96,30 @@ class ContentAppClientCommandSender
...
+        int features { -1 };
```
bzbarsky-apple, line 102: Again, how is this supposed to deal with a featuremap that actually uses bit 31?

src/app/app-platform/ContentApp.h
```diff
@@ -94,11 +96,30 @@ class ContentAppClientCommandSender
...
+        std::vector<int> optionalCommandIdentifiers;
+        std::vector<int> optionalAttributesIdentifiers;
```
bzbarsky-apple, line 104: Neither attribute nor command IDs can all be expressed in `int`.  Please use the right types.

src/app/app-platform/ContentApp.cpp
```diff
@@ -94,6 +94,15 @@ void ContentApp::SendAppObserverCommand(chip::Controller::DeviceCommissioner * c
     ChipLogProgress(Controller, "Completed send of AppObserver command");
 }
 
+bool ContentApp::HasSupportedCluster(chip::ClusterId clusterId) const {
+    for (const auto& supportedCluster : mSupportedClusters) {
+        if (static_cast<int>(clusterId) == supportedCluster.clusterIdentifier) {
```
bzbarsky-apple, line 99: The fact that you had to static_cast here should have been a hint that things are not right.

bzbarsky-apple (changes requested):

src/app/app-platform/ContentApp.h
mthiesc: That is an interesting question. The original reason for this choice was to map the types exactly after the SupportedCluster defined here https://github.com/project-chip/connectedhomeip/blob/master/examples/tv-app/android/App/common-api/src/main/aidl/com/matter/tv/app/api/SupportedCluster.aidl which are converted to this C++ type to make the type conversions one-to-one. But I see that the CluserId is defined as uint32. The discrepency in the existing code is likely because Java does not have a separate type to represent unsigned integers. My understanding is then that all the Java ints *should be* interpreted as unsigned and not signed values, and that they can safely/correctly be cast to uint32_t if I change this type.

Thanks for bringing this up, will make this change.

src/app/app-platform/ContentApp.h
mthiesc: Is there a type alias for the features map as well similar to ClusterId/AttributeId/CommandId or is it just uint32_t?

src/app/app-platform/ContentApp.h
mthiesc: Will do!

src/app/app-platform/ContentApp.h
mthiesc: Updated!

src/app/app-platform/ContentApp.cpp
mthiesc: As mentioned above the date originates as int32 in Java, which we have to compare to uint32 values in C++. So either way we have to do (implicit or explicit) type ambiguations one way or the other to correctly do the comparison. I updated the code using the uint32 types in the C++ SupportedCluster struct and the result is that we do several type ambiguations before storing the int32_t/jint as uint32_t instead of when comparing values here. 

The previous approach mentioned here requires an explicit static_cast to compile. The new approach does not, as it silently converts. In general I think it's good to be (forced to be) explicit about type casts/ambiguations.

src/app/app-platform/ContentApp.h
```diff
@@ -94,11 +96,30 @@ class ContentAppClientCommandSender
...
 {
 public:
+    struct SupportedCluster
+    {
+        chip::ClusterId clusterIdentifier{ chip::kInvalidClusterId };
+        uint32_t features{ 0 };
+        std::vector<chip::CommandId> optionalCommandIdentifiers;
+        std::vector<chip::AttributeId> optionalAttributesIdentifiers;
+
+        SupportedCluster(chip::ClusterId clusterId, uint32_t feats, const std::vector<chip::CommandId> & commandIds,
```
lazarkov, line 106: Can you please rename feats into features? and for the struct variable above, you can use mFeatures?

src/app/app-platform/ContentAppPlatform.cpp
```diff
@@ -528,6 +528,12 @@ uint32_t ContentAppPlatform::GetPasscodeFromContentApp(uint16_t vendorId, uint16
         return 0;
     }
```
lazarkov, line 530: Should you add the same update for the above on the line 504?

examples/tv-app/android/java/AppPlatform-JNI.cpp
```diff
@@ -93,3 +99,86 @@ JNI_METHOD(void, addSelfVendorAsAdmin)
...
+    {
+        return {};
+    }
+
+    // Find Set Iterator Object
+    jobject iteratorObject = env->CallObjectMethod(supportedClustersObject, iteratorMethod);
+    if (iteratorObject == nullptr)
+    {
+        return {};
+    }
```
lazarkov, line 162: Since all of the logic for this method that happened before is validation, maybe that can be explicit as a comment or you can create another method and name it validate input

src/app/app-platform/ContentApp.h
```diff
@@ -122,6 +143,7 @@ class DLL_EXPORT ContentApp
 
 protected:
     EndpointId mEndpointId = 0;
+    std::vector<SupportedCluster> mSupportedClusters;
```
chrisdecenzo, line 146: We should probably make the default be something that makes sense. Or make changes in AppImpl.h since its passing an empty list for the dummy content apps.  

For example, there are mandatory clusters on a content app - KeypadInput, AppLauncher, AppBasic. Then there are the clusters which the code assumes are present (see https://github.com/project-chip/connectedhomeip/blob/bd0422b25ac329323b0ba93808fe7084ec8f2c0a/examples/tv-app/tv-common/src/AppTv.cpp#L315)

src/app/app-platform/ContentApp.h
mthiesc: I renamed feats -> features, no problem using the same name in the end!

src/app/app-platform/ContentAppPlatform.cpp
mthiesc: Added!

examples/tv-app/android/java/AppPlatform-JNI.cpp
mthiesc: It's not just validation but set up to get all the pointers to the classes/methods/fields/objects from Java that are needed. It's not simple/convenient to write a function for this as all these pointers would have to be returned. There are already comments which describe what is going on in each step, so will keep this as is.

src/app/app-platform/ContentApp.h
mthiesc: Interesting. I have some follow up questions:

* Is this only really relevant for the Linux/native TvApp? For the Android TvApp I would assume we (only) want to populate the ContentApp clusters list from the Manifest (?). If so I don't think it would be the best to add defaults to this mSupportedClusters list (as it might conflict/overlap with the manifest declarations on Android. Or is it not an expectation for the manifest to declare all clusters, just what is extra beyond what is deemed default?
* Is the list in [AppImpl.h](https://github.com/project-chip/connectedhomeip/blob/e985c267b22cbb6d207b9c145ddd77a57fb7730c/examples/tv-app/linux/AppImpl.h#L137) only used for Linux testing?
* Is the static/default cluster/attribute list declared in AppTv.cpp that you linked available to the native ContentAppPlatform in some way? How/where is it used?

lazarkov (approved):

src/app/app-platform/ContentApp.h
mthiesc: After talking to @lazarkov about this I have added a list of default supported cluster identifiers to the mContentApp list initializer in AppImpl (for Android) and AppTv (for Linux). These lists are only used for testing to add sensible defaults and pass the new AccountLogin check.

src/app/app-platform/ContentApp.h
```diff
@@ -94,11 +95,32 @@ class ContentAppClientCommandSender
 class DLL_EXPORT ContentApp
 {
 public:
+    struct SupportedCluster
+    {
+        chip::ClusterId mClusterIdentifier{ chip::kInvalidClusterId };
```
bzbarsky-apple, line 100: Note: this is all inside `namespace chip`, so all these `chip::` are probably just noise.

src/app/app-platform/ContentAppPlatform.cpp
```diff
@@ -502,6 +502,12 @@ bool ContentAppPlatform::HasTargetContentApp(uint16_t vendorId, uint16_t product
         return true;
     }
 
+    if (!app->HasSupportedCluster(chip::app::Clusters::AccountLogin::Id))
```
bzbarsky-apple, line 505: This file has `using namespace chip::app::Clusters;` so this can just be:
```suggestion
    if (!app->HasSupportedCluster(AccountLogin::Id))
```

src/app/app-platform/ContentAppPlatform.cpp
```diff
@@ -528,6 +534,12 @@ uint32_t ContentAppPlatform::GetPasscodeFromContentApp(uint16_t vendorId, uint16
         return 0;
     }
 
+    if (!app->HasSupportedCluster(chip::app::Clusters::AccountLogin::Id))
```
bzbarsky-apple, line 537:
```suggestion
    if (!app->HasSupportedCluster(AccountLogin::Id))
```

bzbarsky-apple (approved):
Approving, but note that I did not review the java reflection bits; someone familiar with those needs to do that.

src/app/app-platform/ContentAppPlatform.cpp
mthiesc: Updated!

src/app/app-platform/ContentAppPlatform.cpp
mthiesc: Updated!

src/app/app-platform/ContentApp.h
mthiesc: Updated!

lazarkov (approved):

PR #33832, opened 2024-06-10
Demand Response Load Control.4

steveholmgren:
Updating initial submission after reworking pull request from cluster_and_device_type_dev/how_to_add_new_dts_and_clusters.html.  Believe this is a better pull request reflecting the current DRLC spec.  

Also - Looks like too many changes are related to some kind of pretty print operation on the .kt sources that need to be reverted


bzbarsky-apple (changes requested):
Please split this into separate PRs that:

1. Update the XML (this needs careful review against the spec, etc).
2. Actually add the implementation of the cluster.

not least because chances are you want somewhat different people reviewing those bits.

steveholmgren:
Based on comments on the pull request - there are significant changes splitting this into two PRs so closing this one and will come back with dual PRs per requests

PR #33841, opened 2024-06-11
[OnOff] reset onTime on off with LVL cluster

lecndav:
If an endpoint has the OnOff and the LevelControl cluster enabled, the onTime attribute is not reset in every case. See 1.5.7.1.1 and 1.5.7.4.3.

This PR resets the onTime attribute if the LevelControl cluster is enabled and a command triggers the EP to turn off.

It also fixes some tests in TC-OO-2.3.

### DIFF
--- a/src/app/clusters/on-off-server/on-off-server.cpp
+++ b/src/app/clusters/on-off-server/on-off-server.cpp
@@ -452,12 +452,12 @@ Status OnOffServer::setOnOffValue(chip::EndpointId endpoint, chip::CommandId com
                 ChipLogProgress(Zcl, "ERR: writing on/off %x", to_underlying(status));
                 return status;
             }
+        }
 
-            if (SupportsLightingApplications(endpoint))
-            {
-                ChipLogProgress(Zcl, "Off completed. reset OnTime to  0");
-                Attributes::OnTime::Set(endpoint, 0); // Reset onTime
-            }
+        if (SupportsLightingApplications(endpoint))
+        {
+            ChipLogProgress(Zcl, "Off completed. reset OnTime to  0");
+            Attributes::OnTime::Set(endpoint, 0); // Reset onTime
         }
     }
 


src/app/clusters/on-off-server/on-off-server.cpp
```diff
@@ -452,12 +452,12 @@ Status OnOffServer::setOnOffValue(chip::EndpointId endpoint, chip::CommandId com
...
+        }
 
-            if (SupportsLightingApplications(endpoint))
-            {
-                ChipLogProgress(Zcl, "Off completed. reset OnTime to  0");
-                Attributes::OnTime::Set(endpoint, 0); // Reset onTime
-            }
+        if (SupportsLightingApplications(endpoint))
+        {
+            ChipLogProgress(Zcl, "Off completed. reset OnTime to  0");
```
andy31415, line 459: If you move this outside the "else { // off" branch, the comment probably needs to update as well.

andy31415: Never mind - was mis-counting brackets. This method is not friendly to review :/

andy31415 (approved):

src/app/clusters/on-off-server/on-off-server.cpp
```diff
@@ -452,12 +452,12 @@ Status OnOffServer::setOnOffValue(chip::EndpointId endpoint, chip::CommandId com
...
+            Attributes::OnTime::Set(endpoint, 0); // Reset onTime
```
bzbarsky-apple, line 460: Should this happen immediately, or after the level effect bits finish?

@Rob-Houtepen-Signify @jmartinez-silabs

jmeg-sfy (approved):

src/app/clusters/on-off-server/on-off-server.cpp
jmartinez-silabs: @lecndav @bzbarsky-apple I think in the case where the level control fade out to off effect is used, the on time attribute should be resetted once the food out as completed and the onoff state is finally set to false.

andy31415:
NO updates since 2024. PR considered stale. closing.

PR #33845, opened 2024-06-11
Device type DM XML parser

cecille:
Test: Please see attached test.


src/python_testing/TestSpecParsingDeviceType.py
```diff
@@ -0,0 +1,107 @@
+#
+#    Copyright (c) 2023 Project CHIP Authors
```
cecille, line 2:
```suggestion
#    Copyright (c) 2024 Project CHIP Authors
```

src/python_testing/spec_parsing_support.py
```diff
@@ -128,6 +165,34 @@ class CommandType(Enum):
               0x0072: 'ACFREMON',
               0x0405: 'RH'}
 
+CLUSTER_NAME_FIXES = {0x0036: 'WiFi Network Diagnostics', 0x042a: 'PM25 Concentration Measurement', 0x0006: 'On/Off'}
+DEVICE_TYPE_NAME_FIXES = {0x010b: 'Dimmable Plug-In Unit', 0x010a: 'On/Off Plug-in Unit'}
+
+
+def get_location_from_element(element: ElementTree.Element, cluster_id: int):
+    # Conformance is missing, so let's record the problem and treat it as optional for lack of a better choice
```
cecille, line 173:
```suggestion
```

src/python_testing/TestSpecParsingDeviceType.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def test_device_parsing_clusters(self):
+        device_type_id = 0xBBEF
+        revision = 2
+        classification_class = "simple"
+        classification_scope = "endpoint"
+        clusters = {0x0003: "Identify", 0x0004: "Groups"}
+        # Conformance support tests the different types of conformance for clusters, so here we just want to ensure that we're correctly parsing the XML into python
+        # adds the same attributes and features to every cluster. This is fine for testing.
+        template = Template('<deviceType xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="types types.xsd devicetype devicetype.xsd" id="{{ device_type_id }}" name="Test Device Type" revision="{{ revision }}">\n'
```
andy31415, line 42: Consider multiline strings ... if you really want no indent you could use `textwrap.dedent`. Would save extra '\n' items.

src/python_testing/TestSpecParsingDeviceType.py
```diff
@@ -0,0 +1,107 @@
...
+                              classification_scope=classification_scope, clusters=clusters)
+        et = ElementTree.fromstring(xml)
+        device_type, problems = parse_single_device_type(et)
+        asserts.assert_equal(len(problems), 0, "Unexpected problems parsing device type conformance")
+        asserts.assert_equal(len(device_type.keys()), 1, "Unexpected number of device types returned")
+        asserts.assert_true(device_type_id in device_type.keys(), "device type id not found in returned data")
+        asserts.assert_equal(device_type[device_type_id].revision, revision, "Unexpected revision")
+        asserts.assert_equal(len(device_type[device_type_id].server_clusters), len(clusters), "Unexpected number of clusters")
+
+        # Bad device type ID
```
andy31415, line 84: how about having each of these as individual tests rather than a single test. We could set up the template in the constructor or some setUp call.

src/python_testing/matter_testing_support.py
```diff
@@ -417,8 +417,23 @@ class CustomCommissioningParameters:
 
 
 @dataclass
-class AttributePathLocation:
+class ProblemLocation:
```
andy31415, line 420: A dataclass without members feels odd?

How about defining a type instead of 

ProblemLocation = typing.Union[ClusterPathLocation, DeviceTypeLocation]

since the only use of this seems to be typings.

src/python_testing/matter_testing_support.py
```diff
@@ -436,55 +451,52 @@ def as_string(self, mapper: ClusterMapper):
...
 @dataclass
-class FeaturePathLocation:
-    endpoint_id: int
-    cluster_id: int
-    feature_code: str
+class DeviceTypePathLocation(ProblemLocation):
+    device_type_id: int
+    cluster_id: Optional[int]
+
+    def __init__(self, device_type_id: int, cluster_id: Optional[int] = None):
```
andy31415, line 490: init should not be needed for dataclasses. I guess this may be due to using a base class? I am suggesting not using a base class in this instance.

andy31415: You can set the default for cluster_id in the type `i.e. cluster_id: Optional[int] = None` and then this constructor should be auto-created for you.

src/python_testing/matter_testing_support.py
```diff
@@ -436,55 +451,52 @@ def as_string(self, mapper: ClusterMapper):
...
+        self.device_type_id = device_type_id
+        self.cluster_id = cluster_id
 
     def __str__(self):
-        return (f'\n        Endpoint: {self.endpoint_id},'
-                f'\n        Cluster:  {cluster_id_str(self.cluster_id)},'
-                f'\n        Feature:  {self.feature_code}')
+        msg = f'\n       DeviceType: {self.device_type_id}'
```
andy31415, line 495: existing code, however I would note that str usuaully does not start with \n. It may be better to have some methods like `Format` if we want custom formatting rather than trying to use str or repr.

src/python_testing/TestSpecParsingDeviceType.py
cecille: changed.

src/python_testing/matter_testing_support.py
cecille: yeah, I think I originally had the endpoint in there and then changed my mind. union is fine too - changed.

src/python_testing/matter_testing_support.py
cecille: Going to leave this for now - if we want to muck about with string formatting, that's fine, but in a separate PR. There's a lot of existing code here.

src/python_testing/matter_testing_support.py
cecille: fixed.

src/python_testing/matter_testing_support.py
cecille: @andy31415 - LMK if that's ok. I think before we push up as a package we'll just need a bit of general cleanup here and should handle that in one editorial PR.

jmartinez-silabs (approved):

andy31415 (approved):

PR #33846, opened 2024-06-11
Read base path for config files from TMPDIR env var

jpm-canonical:
Fixes #30790

Change hard coded paths for `chip_factory.ini`, `chip_config.ini` and `chip_counters.ini` to rather use existing `TMPDIR` environment variable.

Before this change these files are stored under `/tmp/*`. After the change they are stored under `$TMPDIR/*`, with a fallback to `/tmp/*` if the variable is not set.

## Rationale

Chip Tool might be used by different users on a system. If the paths are set at build time, multiple users will use the same storage file. This causes a permission error. The first user that runs Chip Tool will be the owner of the files, and the only user that can read and write the files. Subsequent users will see Chip Tool failing with a permission error on these files.

By allowing to override the path for storage file with an environment variable, one can set different paths for different users. This allows isolating different users' storage files, and prevents permission errors on these files.


src/platform/Linux/PosixConfig.cpp
```diff
@@ -453,6 +453,19 @@ bool PosixConfig::ConfigValueExists(Key key)
...
+    const char * dir = getenv("TMPDIR");
+    if (dir == nullptr)
+    {
+        dir = "/tmp";
+    }
+    std::string storageDir = dir;
+
+    return storageDir + "/" + defaultFileName;
+}
+
```
locnnil, line 468: About this function, I have two suggestion approaches:

Either:
```suggestion
std::string PosixConfig::GetFilePath(const std::string &defaultFileName)
{
    // Match what GetFilename in ExamplePersistentStorage.cpp does.
    const char * dir = std::getenv("TMPDIR");
    if (dir == nullptr)
    {
        dir = "/tmp";
    }
    std::string storageDir = dir;

    return storageDir + "/" + defaultFileName;
}

```
- Just use a constant reference for the input parameters.
- And use `std::getenv` to enforce more [idiomatic C++ style](https://en.cppreference.com/w/cpp/utility/program/getenv).

Or

```suggestion
std::string PosixConfig::GetFilePath(const std::string &defaultFileName)
{
    // Match what GetFilename in ExamplePersistentStorage.cpp does.
    const char *dir = std::getenv("TMPDIR");
    if (dir == nullptr)
    {
        dir = "/tmp";
    }

    std::filesystem::path storageDir(dir);
    std::filesystem::path filePath = storageDir / defaultFileName;

    return filePath.string();
}

```
- All in the first suggestion with the addition of using the [C++ filesystem library](https://en.cppreference.com/w/cpp/filesystem), since this repository [uses C++17](https://github.com/project-chip/connectedhomeip/blob/bd0422b25ac329323b0ba93808fe7084ec8f2c0a/.vscode/c_cpp_properties.json#L20), for [all platforms](https://github.com/search?q=repo%3Aproject-chip%2Fconnectedhomeip%20%2B%2B17&type=code). 

**ps**: Remember to change the function signature on `src/platform/Linux/PosixConfig.h` and if you pick the second, don't forget to add the [filesystem header](https://en.cppreference.com/w/cpp/header/filesystem).

locnnil (commented):
I leave an optional suggestion for you!

andy31415:
PR seems stale. Closing.

Issue #30453, opened 2023-11-13
Invoke Batch Commands Feature for 1.3


tehampson:
The following feature is currently planned to be delivered in the following PRs:
- [ ] Add new SessionParameters add to spec for 1.3
- [ ] Add required attributes to BasicInformationCluster add to spec for 1.3
- [ ] Update `CommandDataIB` to allow Building and Parsing multiple commands
- [ ] Update `CommandSender` to support sending multiple commands
- [ ] Update `CommandHandler` to support recieving and processing multiple commands

PR #33849, opened 2024-06-11
Rollback InvokeRequestMessage when AddResponseData fails

tehampson:
Problem: The size of an InvokeRequestMessage, today, is constrained by UDP packet size. When trying to batch many commands using AddRequestData, the call may fail when it runs out of space, causing the InvokeRequestMessage packet to be in a state that cannot be sent to a device.

Solution: Use rollback mechanism within AddRequestData so that the InvokeRequestMessage can be left in a previously known good state should we fail in the middle of adding an InvokeRequest.


src/app/tests/TestCommandInteraction.cpp
```diff
@@ -107,30 +107,33 @@ namespace app {
 
 CommandHandler::Handle asyncCommandHandle;
 
-struct ForcedSizeBuffer
+class ForcedSizeBuffer : public app::DataModel::EncodableToTLV
```
tehampson, line 110: I'll either break the the test component of the PR out to a standalone follow up PR, or wait for #33772 to land first

src/app/CommandSender.cpp
```diff
@@ -128,6 +128,30 @@ CHIP_ERROR CommandSender::SendCommandRequestInternal(const SessionHandle & sessi
     return SendInvokeRequest();
 }
 
+void CommandSender::CreateBackupForRequestRollback(RollbackData & aRollbackData)
+{
+    VerifyOrReturn(mBufferAllocated);
+    VerifyOrReturn(mState == State::Idle || mState == State::AddedCommand);
+    VerifyOrReturn(mInvokeRequestBuilder.GetInvokeRequests().GetError() == CHIP_NO_ERROR);
+    VerifyOrReturn(mInvokeRequestBuilder.GetError() == CHIP_NO_ERROR);
+    mInvokeRequestBuilder.Checkpoint(aRollbackData.backupWriter);
```
andy31415, line 137: This sets several members on aRollbackData. How about having a `RollbackData::Snapshot` or similar that does everything (including setting rollbackIsValid) and then this code would be:

```cpp
aRollbackData.Snapshot(mInvokeRequestBuilder, mState);
```

src/app/tests/TestCommandInteraction.cpp
```diff
@@ -109,30 +109,33 @@ enum class ForcedSizeBufferLengthHint
...
         TLV::TLVType outerContainerType;
         ReturnErrorOnFailure(aWriter.StartContainer(aTag, TLV::kTLVType_Structure, outerContainerType));
-        ReturnErrorOnFailure(app::DataModel::Encode(aWriter, TLV::ContextTag(1), ByteSpan(mBuffer.Get(), mBuffer.AllocatedSize())));
+        ReturnErrorOnFailure(
+            app::DataModel::Encode(aWriter, TLV::ContextTag(1), ByteSpan(mBuffer_.Get(), mBuffer_.AllocatedSize())));
         return aWriter.EndContainer(outerContainerType);
     }
+
+private:
+    chip::Platform::ScopedMemoryBufferWithSize<uint8_t> mBuffer_;
```
andy31415, line 138: is there a reason for the final `_` ? We have `m` prefix for members in chip.

andy31415 (approved):

src/app/CommandSender.cpp
```diff
@@ -657,5 +688,25 @@ void CommandSender::MoveToState(const State aTargetState)
     ChipLogDetail(DataManagement, "ICR moving to [%10.10s]", GetStateStr());
 }
 
+void CommandSender::RollbackData::Checkpoint(CommandSender & aCommandSender)
+{
+    // InvokeRequestMessage::Builder& aInvokeRequestBuilder, const State& aState) {
```
andy31415, line 693: delete this comment?

src/app/CommandSender.cpp
```diff
@@ -128,6 +128,23 @@ CHIP_ERROR CommandSender::SendCommandRequestInternal(const SessionHandle & sessi
     return SendInvokeRequest();
 }
 
+void CommandSender::CreateBackupForRequestRollback(RollbackData & aRollbackData)
+{
+    VerifyOrReturn(mBufferAllocated);
+    VerifyOrReturn(mState == State::Idle || mState == State::AddedCommand);
+    VerifyOrReturn(mInvokeRequestBuilder.GetInvokeRequests().GetError() == CHIP_NO_ERROR);
+    VerifyOrReturn(mInvokeRequestBuilder.GetError() == CHIP_NO_ERROR);
```
andy31415, line 136: since RollbackData checkpoint is passed in *this, I wonder if all these checks should be there? Then we seem to not even need a CreateBackupForRequestRollback as a separate entry.

Otherwise maybe we need to define why some are here vs others inside Checkpoint.

src/app/CommandSender.cpp
```diff
@@ -540,13 +540,27 @@ CHIP_ERROR CommandSender::FinishCommand(FinishCommandParameters & aFinishCommand
...
+    VerifyOrExit(writer != nullptr, err = CHIP_ERROR_INCORRECT_STATE);
+    SuccessOrExit(err = aEncodable.EncodeTo(*writer, TLV::ContextTag(CommandDataIB::Tag::kFields)));
+    {
+        FinishCommandParameters finishCommandParams(aAddRequestDataParams);
+        SuccessOrExit(err = FinishCommand(finishCommandParams));
+    }
+exit:
+    if (err != CHIP_NO_ERROR)
+    {
+        LogErrorOnFailure(rollbackData.Rollback(*this));
```
bzbarsky-apple, line 561: Can we make ~RollbackData roll back automatically unless it's been committed?  And then we don't need the `exit` bits here....

src/app/CommandSender.cpp
```diff
@@ -657,5 +671,30 @@ void CommandSender::MoveToState(const State aTargetState)
...
+    mRollbackIsValid = true;
+}
+
+CHIP_ERROR CommandSender::RollbackData::Rollback(CommandSender & aCommandSender)
+{
+    VerifyOrReturnError(mRollbackIsValid, CHIP_ERROR_INCORRECT_STATE);
+    VerifyOrReturnError(aCommandSender.mState == State::AddingCommand, CHIP_ERROR_INCORRECT_STATE);
+    ChipLogDetail(DataManagement, "Rolling back response");
+    // TODO(#30453): Rollback of mInvokeRequestBuilder should handle resetting
+    // InvokeResponses.
```
bzbarsky-apple, line 691: What does InvokeResponses have to do with anything here?

src/app/CommandSender.h
```diff
@@ -511,6 +517,34 @@ class CommandSender final : public Messaging::ExchangeDelegate
...
+         */
+        void Checkpoint(CommandSender & aCommandSender);
+        /**
+         * Rolls back CommandSender's buffer containing InvokeRequestMessage to a previously
+         * saved state. Must have previously called Checkpoint in a known good state.
+         *
+         * @param [in] aCommandSender reference to CommandSender.
+         */
+        CHIP_ERROR Rollback(CommandSender & aCommandSender);
+        bool RollbackIsValid() { return mRollbackIsValid; }
```
bzbarsky-apple, line 540: Where is this used?

tehampson: Not used anymore, removed

src/app/CommandSender.cpp
tehampson: Updated to InvokeRequest

src/app/CommandSender.cpp
tehampson: I like it. Thanks! Updated let me know if this is more to your liking

src/app/CommandSender.cpp
```diff
@@ -657,5 +662,34 @@ void CommandSender::MoveToState(const State aTargetState)
...
+    mRollbackInDestructor = true;
+}
+
+CommandSender::RollbackInvokeRequest::~RollbackInvokeRequest()
+{
+    VerifyOrReturn(mRollbackInDestructor);
+    VerifyOrReturn(mCommandSender.mState == State::AddingCommand);
+    ChipLogDetail(DataManagement, "Rolling back response");
+    // TODO(#30453): Rollback of mInvokeRequestBuilder should handle resetting
+    // InvokeRequest.
```
bzbarsky-apple, line 682:
```suggestion
    // InvokeRequests.
```

src/app/CommandSender.cpp
```diff
@@ -657,5 +662,34 @@ void CommandSender::MoveToState(const State aTargetState)
...
+    mBackupState          = mCommandSender.mState;
+    mRollbackInDestructor = true;
+}
+
+CommandSender::RollbackInvokeRequest::~RollbackInvokeRequest()
+{
+    VerifyOrReturn(mRollbackInDestructor);
+    VerifyOrReturn(mCommandSender.mState == State::AddingCommand);
+    ChipLogDetail(DataManagement, "Rolling back response");
+    // TODO(#30453): Rollback of mInvokeRequestBuilder should handle resetting
```
bzbarsky-apple, line 681: Is that the right issue number?  That issue does not seem to track this problem anywhere....

bzbarsky-apple (approved):

src/app/CommandSender.cpp
tehampson: Essentially #30453 is a catch all for issues related to the feature including things discovered while working on the feature.

Before marking #30453 as complete, I will search for all code reference and either:
* Resolve them appropriately.
* Create a new issue that better describes that specific problem and change the reference issue number in the code.

PR #33854, opened 2024-06-11
Choice conformance: Add choice conformance markers

cecille:
Adds the ability to set choice conformance markers on optional conformances. 

Test: Included in PR.


src/python_testing/TestConformanceSupport.py
```diff
@@ -750,6 +751,207 @@ def test_device_type_conformance(self):
...
+        # - NOT expressions (!O.a)
+        # - internal expressions [AB.a], [attr1.a], [cmd1.a]
+        # - provisional P.a
+        # - disallowed X.a
+        # - deprecated D.a
+
+        choices = [('a+', 'choice="a" more="true"', True), ('a', 'choice="a"', False)]
+        for suffix, xml_attrs, more in choices:
+            def check_good_choice(xml: str, conformance_str: str) -> Conformance:
+                msg = 'Bad choice conformance string'
```
andy31415, line 780: Suggest placing the msg inline to the assert_equal

src/python_testing/conformance_support.py
```diff
@@ -50,13 +50,50 @@ def __str__(self):
...
+    choice: str
+    more: bool
+
+    def __str__(self):
+        more_str = '+' if self.more else ''
+        return '.' + self.choice + more_str
+
+
+def parse_choice(element: ElementTree.Element) -> Optional[Choice]:
+    choice = element.get('choice', '')
```
andy31415, line 73: nit: if you do a `if not choice: return None` you can rely that choice is valid in the lines below (saves some conditions)`.

is `more` a valid field if choice is not specified?

src/python_testing/conformance_support.py
```diff
@@ -50,13 +50,50 @@ def __str__(self):
...
+
+
 class ConformanceDecision(Enum):
     MANDATORY = auto()
     OPTIONAL = auto()
     NOT_APPLICABLE = auto()
     DISALLOWED = auto()
     PROVISIONAL = auto()
 
+    def set_choice(self, choice: Choice):
```
andy31415, line 87: Is this right? Enums are singletons, so if you do:

ConformanceDecision.MANDATORY.set_choice(123) then all of ConformanceDecision.MANDATORY have choice of 123.

I was going to suggest:

```py
def __init__(self, *a, **k):
   super().__init__(*a, **k)
   self.choice = None
```

to avoid the AttributeError below, however overall this feels off because of singletons. I believe you may need a dataclass with a Conformance member and a choice member. Enum does not seem made for having extra members.

src/python_testing/conformance_support.py
```diff
@@ -78,55 +115,65 @@ def is_disallowed(conformance: Callable):
     return conformance(0, [], []) == ConformanceDecision.DISALLOWED
 
 
-class zigbee:
+@dataclass
+class Conformance(Callable):
+    def __call__(self, feature_map: uint, attribute_list: list[uint], all_command_list: list[uint]) -> ConformanceDecision:
```
andy31415, line 120: a doccomment about what the arguments mean would be helpful.

src/python_testing/conformance_support.py
```diff
@@ -78,55 +115,65 @@ def is_disallowed(conformance: Callable):
...
 
 
-class optional:
+class optional(Conformance):
+    def __init__(self, choice: Optional[Choice] = None):
+        self.choice = choice
+
     def __call__(self, feature_map: uint, attribute_list: list[uint], all_command_list: list[uint]) -> ConformanceDecision:
-        return ConformanceDecision.OPTIONAL
+        return ConformanceDecision.OPTIONAL.set_choice(self.choice)
```
andy31415, line 146: I am afraid this changes globals. I think we get:

```py
a = ConformanceDecision.OPTIONAL.set_choice(1)
b = ConformanceDecision.OPTIONAL.set_choice(2)

print(a.choice) # I bleieve this prints out 2, which is probably not expected
```

andy31415 (dismissed):
I believe enums should be immutable. Using enums to contain a global variable seems off.

src/python_testing/conformance_support.py
```diff
@@ -58,14 +87,27 @@ class ConformanceDecision(Enum):
     PROVISIONAL = auto()
 
 
+@dataclass
+class ConformanceDecisionWithChoice:
+    decision: ConformanceDecision
+    choice: Optional[Choice] = None
+
+    def __eq__(self, other):
+        if isinstance(other, self.__class__):
```
andy31415, line 96: A comment saying "this supports comparison to ConformanceDecision without considering the choice" seems maybe useful as this seems to be the important point.

Maybe also `@dataclass(eq=false)` to make it clear we do not want `__eq__` autogenerated by dataclass.

andy31415: This also is probably tricky because `a == b` does not imply `b == a` ... which seems kind of dangerous.

How about removing `__eq__` and implementing a `SameDecision(self, other)` to make the ordering requirement explicit?

src/python_testing/conformance_support.py
```diff
@@ -250,66 +315,70 @@ def __call__(self, feature_map: uint, attribute_list: list[uint], all_command_li
             raise ConformanceException('NOT operation on optional or disallowed item')
         # Features in device types degrade to optional so a not operation here is still optional because we don't have any way to verify the features since they're not exposed anywhere
         elif decision == ConformanceDecision.OPTIONAL:
-            return ConformanceDecision.OPTIONAL
+            return decision
```
andy31415, line 318: comment about return types seems to apply throughout the PR.

src/python_testing/conformance_support.py
cecille: Removed this and just compared directly - you're right, it's more maintainable etc.

src/python_testing/conformance_support.py
cecille: as above - changed this to be more explicit by checking against the data member directly and removing the equality operator.

jmartinez-silabs (approved):

CKnoerzer (approved):

andy31415 (approved):

PR #33856, opened 2024-06-11
[Darwin] MTRDeviceController getSessionForNode: to use subscription pool as needed

jtung-apple:
This change makes MTRDeviceController's `-getSessionForNode:completion:` also participate in the subscription pool when there is an MTRDevice associated with the nodeID, and is also known to use Thread.

With this change, MTRBaseDevice and MTRBaseClusters reads/subscriptions for devices using Thread would also get the CASE session establishment gated by the subscription pool, and help avoid congesting the Thread network when a controller starts and tries to get status for all its Thread devices.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -1061,6 +1061,12 @@ static NSString * const sLastInitialSubscribeLatencyKey = @"lastInitialSubscribe
     [self _changeState:MTRDeviceStateUnreachable];
 }
 
+- (BOOL)deviceUsesThread
+{
+    std::lock_guard lock(_lock);
+    return [self _deviceUsesThread];
+}
+
 // This method is used for signaling whether to use the subscription pool. This functions as
 // a heuristic for whether to throttle subscriptions to the device via a pool of subscriptions.
 // If products appear that have both Thread and Wifi enabled but are primarily on wifi, this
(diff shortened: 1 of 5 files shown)


src/darwin/Framework/CHIPTests/MTRPerControllerStorageTests.m
```diff
@@ -2269,6 +2269,7 @@ - (void)doTestSubscriptionPoolWithSize:(NSInteger)subscriptionPoolSize deviceOnb
     XCTAssertNotNil(factory);
 
     __auto_type queue = dispatch_get_main_queue();
+    //    __auto_type queue = dispatch_queue_create("test.subscription.pool", DISPATCH_QUEUE_SERIAL_WITH_AUTORELEASE_POOL);
```
woody-apple, line 2272:
```suggestion
```

src/darwin/Framework/CHIP/MTRDevice_Internal.h
```diff
@@ -117,6 +117,9 @@ MTR_TESTABLE
 
 - (void)setStorageBehaviorConfiguration:(MTRDeviceStorageBehaviorConfiguration *)storageBehaviorConfiguration;
 
+// Returns whether this MTRDevice uses Thread for communication
+- (BOOL)deviceUsesThread;
```
woody-apple, line 121:
```suggestion
- (BOOL)_deviceUsesThread;
```

woody-apple, line 121: We should probably have a public or private bitmask for what type of transports it supports...

woody-apple: (More a note, I might take a stab at this)

jtung-apple: Right, this is currently just used internally to signal when to pool these attempts. I think things can be named in a more generic way..... Naming is a hard problem and we definitely need to give it more thought.

woody-apple: It's fine for now :)

woody-apple (approved):

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -2055,194 +2061,195 @@ - (void)_setupSubscriptionWithReason:(NSString *)reason
         });
     }
 
+    // Call directlyGetSessionForNode because the subscription setup already goes through the subscription pool queue
```
bzbarsky-apple, line 2064: So I was figuring we would stop putting the subscription setup in there and just call getSessionForNode and have _that_ get queued.... why are we not doing that?

jtung-apple: Subscriptions on controller startup can cause a thundering herd issue with many Thread devices. If we only use the N-pool for getSessionForNode, then as soon as session is established, another device subscription can move forward, and we can get into a situation where many subscriptions and their priming reports are in the air at the same time.

PR #33860, opened 2024-06-12
OpenThread: added Thread interface enabled support

DejinChen:
### Change overview

Added support for enabling/disabling the OpenThread network interface.
- Set the interface to true or false, will connect or disconnect Thread.
- If the interface is set to false, Thread will not connect automatically during reboot.

Related PR: Auto-commissioner: support secondary network interface commissioning (#33801)

### DIFF
--- a/src/platform/OpenThread/GenericNetworkCommissioningThreadDriver.cpp
+++ b/src/platform/OpenThread/GenericNetworkCommissioningThreadDriver.cpp
@@ -54,6 +54,8 @@ CHIP_ERROR GenericThreadDriver::Init(Internal::BaseDriver::NetworkStatusChangeCa
     // must be restored on the boot. If there's no backup, the below function is a no-op.
     RevertConfiguration();
 
+    CheckInterfaceEnabled();
+
     return CHIP_NO_ERROR;
 }
 
@@ -204,10 +206,9 @@ CHIP_ERROR GenericThreadDriver::SetEnabled(bool enabled)
 
     ReturnErrorOnFailure(PersistedStorage::KeyValueStoreMgr().Put(kInterfaceEnabled, &enabled, sizeof(enabled)));
 
-    if ((!enabled && DeviceLayer::ThreadStackMgrImpl().IsThreadEnabled()) ||
-        (enabled && DeviceLayer::ThreadStackMgrImpl().IsThreadProvisioned()))
+    if ((!enabled && ThreadStackMgrImpl().IsThreadEnabled()) || (enabled && ThreadStackMgrImpl().IsThreadProvisioned()))
     {
-        ReturnErrorOnFailure(DeviceLayer::ThreadStackMgrImpl().SetThreadEnabled(enabled));
+        ReturnErrorOnFailure(ThreadStackMgrImpl().SetThreadEnabled(enabled));
     }
     return CHIP_NO_ERROR;
 }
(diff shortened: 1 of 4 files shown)


src/platform/OpenThread/GenericThreadStackManagerImpl_OpenThread.hpp
```diff
@@ -1131,8 +1134,12 @@ CHIP_ERROR GenericThreadStackManagerImpl_OpenThread<ImplClass>::DoInit(otInstanc
     memset(&mSrpClient, 0, sizeof(mSrpClient));
 #endif // CHIP_DEVICE_CONFIG_ENABLE_THREAD_SRP_CLIENT
 
+#ifndef _NO_NETWORK_COMMISSIONING_DRIVER_
+    InterfaceEnabled = sGenericThreadDriver.GetEnabled();
```
Damian-Nordic, line 1138: It introduces a cyclic dependency. Normally, it should be Thread driver talking to `ThreadStackManager` and not the other way around.

But overall, why does this have to involve the thread driver and persistent storage. Wouldn't it be enough to provide some `CHIP_DEVICE_CONFIG_ENABLE_IF_AUTOSTART` configuration which you could disable in your app and only enable the Thread interface if needed?

DejinChen: The cyclic dependency is indeed a problem. But Thread interface may be enabled and disabled during program running, the compiling configuration may not fit the requires. 

For example, if a device is commissioned successfully through Thread, it should auto-start Thread network. But if commissioning failed, we want to disable Thread interface and try other network interface like Wi-Fi. In this case, the Thread should not auto-start.

Damian-Nordic: The compiling configuration for disabling the autostart should allow you to start or stop the interface based on your application's requirements. On top of that you can implement persisting the interface state and either call `SetThreadEnable()` right after `InitThreadStack()` or not. Wouldn't that be convenient for you?

My point is that this PR seems to implement a very specific use case (e.g. the interface is automatically brought up if `ConnectNetwork` is called) and I wonder if we can keep the common code as generic as possible.

Damian-Nordic (approved):

src/platform/OpenThread/GenericThreadStackManagerImpl_OpenThread.hpp
Damian-Nordic: Thanks, great that you got rid of the cyclic dependency. I'm not sure if I understand precisely how `InterfaceEnabled` should actually work but the logic makes sense to me.

src/platform/OpenThread/GenericNetworkCommissioningThreadDriver.cpp
```diff
@@ -95,6 +97,12 @@ CHIP_ERROR GenericThreadDriver::RevertConfiguration()
     // since the fail-safe was armed, so return with no error.
     ReturnErrorCodeIf(error == CHIP_ERROR_PERSISTED_STORAGE_VALUE_NOT_FOUND, CHIP_NO_ERROR);
 
+    if (!GetEnabled())
+    {
+        // If backup is found, set InterfaceEnabled to default value (true).
```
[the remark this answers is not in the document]
DejinChen: The SPEC says:
 "It is undefined what happens if InterfaceEnabled is written to false on the same interface as that
which is used to write the value. In that case, it is possible that the Administrator would have to
await expiry of the fail-safe, and associated recovery of network configuration to prior safe values,
before being able to communicate with the node again"
So we set the InterfaceEnabled attribute to the default value when reverting network.

jmartinez-silabs (approved):

PR #33876, opened 2024-06-12
[NXP] Fix TC-I-2.2 test failure

Martin-NXP:
Fix TC-I-2.2 test failure (step2b step6b: The response value should be lower or equal to the constraint). This was due to an identify cluster initialization issue, it was initialized twice for endpoint 1 due to a typo error


andy31415 (dismissed):
`Fix identify cluster issue on endpoint 0` does not seem to explain the change.

Please update the title and summary to explain what the problem is and what the fix was. Assume people read through the changelog text to figure out deltas.

andy31415 (approved):

bzbarsky-apple (approved):

PR #33877, opened 2024-06-12
implement 'ReadAttribute' inside the `codegen-data-model` IM/DM split

andy31415:
This implements ReadAttribute to be compatible with existing `ember-compatibility-functions` and adds corresponding unit tests.

Given that we use AttributeValueEncoder, I removed the `ReadState` type and parameter from the reads (the encoder currently contains the state).

image

## Implementation notes

This is intended as a drop-in replacement for ember-compatibility-functions.cpp (which would be removed once this is fully integrated and working). As a result, some of the logic is duplicated (e.g. FindAttributeMetadata logic) and simlar handling exists for error reporting for path expansion.


src/app/codegen-data-model/CodegenDataModel_Read.cpp
```diff
@@ -0,0 +1,356 @@
...
+
+// Fetch the source for the given attribute path: either a cluster (for global ones) or attribute
+// path.
+//
+// if returning a CHIP_ERROR, it will NEVER be CHIP_NO_ERROR.
+std::variant<const EmberAfCluster *,           // global attribute, data from a cluster
+             const EmberAfAttributeMetadata *, // a specific attribute stored by ember
+             CHIP_ERROR                        // error, this will NEVER be CHIP_NO_ERROR
+             >
+FindAttributeMetadata(const ConcreteAttributePath & aPath)
```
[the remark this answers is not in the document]
andy31415: I aim to provide a drop-in replacement for ember-compatibility-functions and eventually that source file should be removed.

As a result there will be high overlap potentially, however at the same time I purposfully do not try to centralize/reuse as that other file should be removed.

src/app/codegen-data-model/CodegenDataModel_Read.cpp
```diff
@@ -0,0 +1,356 @@
...
+    {
+        return std::nullopt;
+    }
+
+    CHIP_ERROR err = aai->Read(path, encoder);
+
+    // explict translate UnsupportedRead to Access denied. This is to allow callers to determine a
+    // translation for this: usually wildcard subscriptions MAY just ignore these where as direct reads
+    // MUST translate them to UnsupportedAccess
+    ReturnErrorCodeIf(err == CHIP_IM_GLOBAL_STATUS(UnsupportedRead), CHIP_ERROR_ACCESS_DENIED);
```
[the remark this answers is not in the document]
andy31415: This was because of ember-compatibiltiy treating this as access denied:

https://github.com/project-chip/connectedhomeip/blob/master/src/app/util/ember-compatibility-functions.cpp#L204

```cpp
    if (err == CHIP_IM_GLOBAL_STATUS(UnsupportedRead) && aPath.mExpanded)
    {
        //
        // Set this to true to ensure our caller will return immediately without proceeding further.
        //
        *aTriedEncode = true;
        return CHIP_NO_ERROR;
    }
```

Unfortunately the comment does not exactly explain why an unsupported read while path-expanding is supposed to be a "success/continue" and the only current expansion logic that I have seen was that "access denied is not an error for path expansion" hence my change.

It may also be that this is for "write-only" attribute skipping, but that seems also to translate reasonably as access denied.

The change is mainly as a "do whatever ember compatibility does".

src/app/codegen-data-model/CodegenDataModel_Read.cpp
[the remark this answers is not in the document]
andy31415: Updated the commit comment with this.

src/app/codegen-data-model/CodegenDataModel_Read.cpp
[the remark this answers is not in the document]
andy31415: Alternative would be to allow `UnsupportedRead` to pass through and document that as a "potentially skippable for wildcard" as well, however I was unsure.

Thinking about it, I imagine we have certification tests that ask for 'unsupported read needs to be returned' so it seems I should pass it through. I will update the code.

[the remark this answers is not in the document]
andy31415: Code updated.

src/app/codegen-data-model/CodegenDataModel_Read.cpp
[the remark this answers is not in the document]
cecille: actually, ignore this - it's addressed in your comments on top.

src/app/codegen-data-model/CodegenDataModel_Read.cpp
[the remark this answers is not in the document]
cecille: It's actually reasonably difficult to test unsupported read, unfortunately - we don't have any in the spec, I think we have one as a part of unit testing, but I've only ever encountered them for real in MEI clusters.

src/app/codegen-data-model/CodegenDataModel_Read.cpp
```diff
@@ -0,0 +1,350 @@
...
+             CHIP_ERROR                        // error, this will NEVER be CHIP_NO_ERROR
+             >
+FindAttributeMetadata(const ConcreteAttributePath & aPath)
+{
+    for (auto & attr : GlobalAttributesNotInMetadata)
+    {
+        if (attr == aPath.mAttributeId)
+        {
+            const EmberAfCluster * cluster = emberAfFindServerCluster(aPath.mEndpointId, aPath.mClusterId);
+            ReturnErrorCodeIf(cluster == nullptr, CHIP_IM_GLOBAL_STATUS(UnsupportedAttribute));
```
bzbarsky-apple, line 66: This logic does not match the logic in the existing `ReadSingleClusterData`, and the result is that it's returning the wrong error code.

The error code for a supported global attribute will _never_ be UnsupportedAttribute, per spec.

I strongly suggest copying the "have a helper UnsupportedAttributeStatus function" bits from ember-compatibility-functions.cpp.

src/app/codegen-data-model/CodegenDataModel_Read.cpp
```diff
@@ -0,0 +1,350 @@
...
+
+    CHIP_ERROR err = aai->Read(path, encoder);
+
+    if (err != CHIP_NO_ERROR)
+    {
+        return std::make_optional(err);
+    }
+
+    // If the encoder tried to encode, then a value should have been written.
+    //   - if encode, assueme DONE (i.e. FINAL CHIP_NO_ERROR)
```
bzbarsky-apple, line 119: Spelling.

src/app/codegen-data-model/CodegenDataModel_Read.cpp
```diff
@@ -0,0 +1,350 @@
...
+/// separated-out ReadAttribute implementation (given existing complexity)
+///
+/// Generally will:
+///    - validate ACL (only for non-internal requests)
+///    - Try to read attribute via the AttributeAccessInterface
+///    - Try to read the value from ember RAM storage
+CHIP_ERROR CodegenDataModel::ReadAttribute(const InteractionModel::ReadAttributeRequest & request, AttributeValueEncoder & encoder)
+{
+    ChipLogDetail(DataManagement,
+                  "Reading attribute: Cluster=" ChipLogFormatMEI " Endpoint=%x AttributeId=" ChipLogFormatMEI " (expanded=%d)",
```
bzbarsky-apple, line 286:
```suggestion
                  "Reading attribute: Cluster=" ChipLogFormatMEI " Endpoint=0x%x AttributeId=" ChipLogFormatMEI " (expanded=%d)",
```

Or just %d; usually endpoint numbers are not hex.  I know you just copied this from ReadSingleClusterData...

src/app/codegen-data-model/CodegenDataModel_Read.cpp
```diff
@@ -0,0 +1,350 @@
...
+                  ChipLogValueMEI(request.path.mClusterId), request.path.mEndpointId, ChipLogValueMEI(request.path.mAttributeId),
+                  request.path.mExpanded);
+
+    // ACL check for non-internal requests
+    if (!request.operationFlags.Has(InteractionModel::OperationFlags::kInternal))
```
bzbarsky-apple, line 291: So what is an "internal request" and why would those be going through the TLV interfaces?  Especially because the underlying AttributeAccessInterface logic does not have a concept of "internal request"...

src/app/codegen-data-model/CodegenDataModel_Read.cpp
```diff
@@ -0,0 +1,350 @@
...
+    {
+        ReturnErrorCodeIf(!request.subjectDescriptor.has_value(), CHIP_ERROR_INVALID_ARGUMENT);
+
+        Access::RequestPath requestPath{ .cluster = request.path.mClusterId, .endpoint = request.path.mEndpointId };
+        ReturnErrorOnFailure(Access::GetAccessControl().Check(*request.subjectDescriptor, requestPath,
```
bzbarsky-apple, line 296: Spec requires _very_ particular behavior here depending on whether path was expanded or not.  That needs to be handled here, because caller will not be able to tell reliably that this was a failed ACL check as opposed to some other failure.  And no, sniffing the exact value of the CHIP_ERROR is not reliable.

Alternately, we need a way to communicate to the caller that this failed because of the ACL check....

src/app/codegen-data-model/CodegenDataModel_Read.cpp
```diff
@@ -0,0 +1,350 @@
...
+    {
+        Compatibility::GlobalAttributeReader aai(*cluster);
+        aai_result = TryReadViaAccessInterface(request.path, &aai, encoder);
+    }
+    else
+    {
+        aai_result = TryReadViaAccessInterface(
+            request.path, GetAttributeAccessOverride(request.path.mEndpointId, request.path.mClusterId), encoder);
+    }
+    ReturnErrorCodeIf(aai_result.has_value(), *aai_result);
```
bzbarsky-apple, line 321: This seems broken; if AAI returns UnsupportedRead and the path is expanded, this needs to be doing something quite different from what it does, no?

src/app/codegen-data-model/CodegenDataModel_Read.cpp
```diff
@@ -0,0 +1,350 @@
...
+
+    if (!std::holds_alternative<const EmberAfAttributeMetadata *>(metadata))
+    {
+        // if we only got a cluster, this was for a global attribute. We cannot read ember attributes
+        // at this point, so give up (although GlobalAttributeReader should have returned something here).
+        // Return a permanent failure...
+        return CHIP_IM_GLOBAL_STATUS(UnsupportedAttribute);
```
bzbarsky-apple, line 328: Is there a reason this is not a VerifyOrDie? That said, you already hit the relevant VerifyOrDieWithMsg in GlobalAttributeReader::Read, no?  This line is unreachable...

src/app/codegen-data-model/tests/TestAttributeReportIBsEncoding.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+#include <app/MessageDef/AttributePathIB.h>
+#include <app/MessageDef/AttributeReportIB.h>
+
+using namespace chip::app;
+
+namespace chip {
+namespace Test {
+
+CHIP_ERROR DecodedAttributeData::DecodeFrom(const chip::app::AttributeDataIB::Parser & parser)
```
bzbarsky-apple, line 27: Why the extra prefixes?

src/app/codegen-data-model/tests/TestAttributeReportIBsEncoding.cpp
```diff
@@ -0,0 +1,130 @@
...
+}
+
+CHIP_ERROR DecodeAttributeReportIBs(ByteSpan data, std::vector<DecodedAttributeData> & decoded_items)
+{
+    // Espected data format:
+    //   CONTAINER (anonymous)
+    //     0x01 => Array (i.e. report data ib)
+    //       ReportIB*
+    //
+    // Overally this is VERY hard to process ...
```
bzbarsky-apple, line 46: "Overally"?

src/app/codegen-data-model/tests/TestAttributeReportIBsEncoding.cpp
bzbarsky-apple, line 1: I have not checked details here, not least because I can't tell what this test is trying to test.

src/app/codegen-data-model/tests/TestCodegenModelViaMocks.cpp
bzbarsky-apple, line 1: Given the bugs in the implementation, I assume there are bugs in this test too that prevent those being caught....  That said, I did not carefully review this test.

src/app/data-model-interface/DataModel.h
```diff
@@ -55,22 +55,24 @@ class DataModel : public DataModelMetadataTree
...
 
-    /// List reading has specific handling logic:
-    ///   `state` contains in/out data about the current list reading. MUST start with kInvalidListIndex on first call
-    ///
     /// Return codes:
-    ///   CHIP_ERROR_MORE_LIST_DATA_AVAILABLE (NOTE: new error defined for this purpose)
-    ///      - partial data written to the destination
-    ///      - destination will contain AT LEAST one valid list entry fully serialized
-    ///      - destination will be fully valid (it will be rolled back on partial list writes)
+    ///   CHIP_ERROR_ACCESS_DENIED:
```
bzbarsky-apple, line 59: Sniffing specific error values like this is not reliable; something deep in the guts of some platform impl might return CHIP_ERROR_ACCESS_DENIED for some random reason, ReturnErrorOnFailure propagates it right out, and we have incorrect behavior.

src/app/data-model-interface/DataModel.h
```diff
@@ -55,22 +55,24 @@ class DataModel : public DataModelMetadataTree
...
+    ///      - May be ignored if reads use path expansion (e.g. to skip inaccessible attributes
+    ///        during subscription requests). This code MUST be used for ACL failures and only for ACL failures.
+    ///   CHIP_IM_GLOBAL_STATUS(UnsupportedRead):
+    ///      - May be ignored if reads use path expansion (e.g. to skip reading write-only attributes
+    ///        during subscription requests).
```
bzbarsky-apple, line 64: This one is a bit better, perhaps, since it's sort of existing API surface for AttributeAccessInterface, but still not very reliable.....

bzbarsky-apple (changes requested):
Changes requested because of spec-noncompliance bits.

src/app/codegen-data-model/CodegenDataModel_Read.cpp
andy31415: The background here is that I am looking to abstract ember away from code. However ember has the concept of internal/external via emberAfReadOrWriteAttribute which is external and emberAfWriteAttribute which is marked as `internal` implicitly from the comments that state:

```
...
* This function will not check to see if the attribute is                                                                                                                                                                                  * writable since the read only / writable characteristic                                                                                                                                                                                  * of an attribute only pertains to external devices writing                                                                                                                                                                               * over the air. Because this function is being called locally 
...
```

I am unsure how to express that better. The way it seems intended (at least in our access cluster functions) is that internal code is allowed to both bypass ACL (timer code for example would not create subject descriptors) and bypass read/write flags (you can write read-only attributes, typically allowing factory data to be read in at boot time, like "write" of a serial number).

src/app/codegen-data-model/CodegenDataModel_Read.cpp
andy31415: Would it be acceptable to have a dedicated error for these checks?

using path.mExpanded is strange and I would claim wrong: it is not obvious that mExpanded means `ACL check is not done` unless one reads comments carefully and if you need to read comments carefully, this means it is significantly harder to maintain.

The user  that actually decides if ACL failures or unsupported read is acceptable seems to be the caller (subscription with wildcard or not), so we should either:
   - get this call to Read tell us if this failed ACL (and caller decides how to handle the error) - this is the path I took
   - pass in to the caller if you want read be ok with ACL failure via some argument. I have some complains about how ember-compatibility-functions does this today:
       - mExpanded is not a reasonably named flag for this "ignore acl checks" logic
       - returning CHIP_NO_ERROR but not actually filling the read value is odd. I would expect "if you do not get a read error, the read buffer is filled with appropriate data". Given that null and empty values can be valid, just returning CHIP_NO_ERROR does not seem reasonable.

Is there an approach that you would suggest? I would like one that addresses my concerns with returning CHIP_NO_ERROR.

src/app/codegen-data-model/CodegenDataModel_Read.cpp
andy31415: That is what ember-compatibility-functions.cpp does ... and I claim it is not properly maintainable. Here instead I will return `unsupportedread` and the caller can decide what to do there. Thecaller knows if it needs expansion or not.

src/app/data-model-interface/DataModel.h
andy31415: I could return a dedicated error (but that would map to access denied ... so not that great either) or return a variant of CHIP_ERROR or some different dedicated type to force mapping ... but that would just mean CHIP_ERROR or "special error" and just makes the interface harder.

I need a way to tell the caller what happend. If something deep in the guts of the platform says ACCESS_DENIED why is it not correct to treat it just like any other access denied? This would allow AAI to also say access defined and be handled correctly.

src/app/codegen-data-model/CodegenDataModel_Read.cpp
andy31415: https://github.com/project-chip/connectedhomeip/blob/master/src/app/util/ember-compatibility-functions.cpp#L305 seems to do this 

Overall I tried hard to keep compatibility with ember-compatibility-functions except that I found that its handling and usage of mExpanded to need updates.

src/app/codegen-data-model/CodegenDataModel_Read.cpp
bzbarsky-apple: The right answer for things like that is that clusters should expose APIs to set attributes on them as needed, instead of having a generic API that you are allowed to use to just poke the internals of clusters.

That will also avoid the need to stick the attribute values being written through the pinhole of TLV in the process.

In practice, the _IM_ never calls that API.  It's only the application that does, right?

src/app/codegen-data-model/CodegenDataModel_Read.cpp
bzbarsky-apple: UnsupportedAttributeStatus never returns UnsupportedAttribute for a global attribute, right?

andy31415: Will re-read the logic to double-check.

src/app/codegen-data-model/CodegenDataModel_Read.cpp
bzbarsky-apple: > Would it be acceptable to have a dedicated error for these checks?

You mean a dedicated CHIP_ERROR value?  I would really rather not.  People use CHIP_ERROR values willy-nilly based on what they sound like.

If we are trying to communicate a specific state, we should communicate that state.  If we need a richer return type than CHIP_ERROR, we should do that.

> using path.mExpanded is strange and I would claim wrong

Just to be clear, the only reason mExpanded exists is to allow this check to be done, yes?  It affects more than just ACL checks, by the way, which is why it's named like it is.

> unless one reads comments carefully

Or actually reads the spec.

> The user that actually decides if ACL failures or unsupported read is acceptable seems to be the caller
> (subscription with wildcard or not)

It's not just subscriptions.  It's reads too.

I think a fundamental issue here seems to be that you are reading the ember-compatibility-functions code, trying to guess what it's trying to do, and then trying to reimplement that, instead of actually implementing the spec.  Reverse-engineering the spec from the code is much harder than reading the spec, and hence more error-prone.

> get this call to Read tell us if this failed ACL (and caller decides how to handle the error) - this is the path I took

Again, it's not just ACL checks.

> mExpanded is not a reasonably named flag for this "ignore acl checks" logic

That's because it affects things other than ACL checks.  The relevant spec bits say:

> if the path is a concrete path:
> (acl checks, existence checks, and the like here)
>Else perform Request Path Expansion and process each expanded existent path as follows:
>
>   If the path indicates attribute data that is not readable, then the path SHALL be discarded.
>
>    Else if reading from the attribute in the path requires a privilege that is not granted to access the cluster in the path, then the path SHALL be discarded.
>
>each remaining error free existent path is processed as follows: (actually do reads here)

That's the actual logic that needs to be implemented.  The way it's handled in ember-compatibility-functions is a specific way to implement it that it pushes the "this was not a concrete path" boolean down into the the read and does all the checks inside the read.  This means it re-does existence checks for things we got from wildcard expansion, which is a bit silly, but keeps the codepaths down.  But we could adopt whatever implementation strategy we want here.

src/app/codegen-data-model/CodegenDataModel_Read.cpp
andy31415: Ok, what threw me off I believe is that the UnsupportedAttributeStatus could return (for arbitrary path) a UnsupportedAttribute, however for globals we know that emberAfFindServerCluster is null so it will never return it.

Revised code, hopefully this is both readable and correct.

src/app/codegen-data-model/CodegenDataModel_Read.cpp
bzbarsky-apple: I agree that this reliance on magic error code values is bad for maintainability.  Past that, see discussion above.

src/app/data-model-interface/DataModel.h
bzbarsky-apple: >  If something deep in the guts of the platform says ACCESS_DENIED why is it not correct to treat it just like any other access denied? 

Because absolutely nothing guarantees that CHIP_ERROR_ACCESS_DENIED matches the "reading from the attribute in the path requires a privilege that is not granted to access the cluster in the path" concept from the spec.  Maybe it does, maybe it does not.  The spec doesn't say "ignore random denials" it says to ignore _very specific things_.

Concretely, just looking at our code various platform time implementations return CHIP_ERROR_ACCESS_DENIED from `ClockImpl::SetClock_RealTime`, say.  How do you guarantee that nothing in AAI somewhere tries to do that?  You can't.

src/app/codegen-data-model/CodegenDataModel_Read.cpp
andy31415: Right, I need some ember compatibility layer that does not depend on ember linkage, so I need to in some way provide a way to say "bypass these rules" because we have code requiring this currently (specifically all Accessors, unsure if others call ember directly, they probably do not because we discourage that).

This is a flag essentially saying "accessor call" (or internal app call). It is also true that IM never calls this, yet I need cluster implementations to be able to call it this way. Either I have a flag (like here) or find some other way, but that was unclear (having a custom subject seemed wrong, especially coupled with a requirement to bypass read/write flags).

This could be done better by not having these functions check ACL or r/w flags at all since metadata provides that, however then I cannot make this a drop-in replacement for ember immediately and I would need to:
  - replace ember-compatibility
  - move ACL and read/write flag check out
  - implement accessors because now it is possible
 
 I can probably still do the above, but the internal flag provides me with a stop-gap where things can be worked on in paralle.

src/app/codegen-data-model/CodegenDataModel_Read.cpp
andy31415: This method does not technically have visibility in GlobalAttributeReader::Read ... it needs to work as a stand-alone. This seemed appropriate but I can do VerifyOrDie as well.

src/app/codegen-data-model/tests/TestAttributeReportIBsEncoding.cpp
andy31415: I think my brain tried to type `Generally` and `Overall` at the same time... amazing.
Updated.

src/app/codegen-data-model/CodegenDataModel_Read.cpp
bzbarsky-apple: OK.  I think we should perhaps document that this is temporary until other things get fixed to be sane?  Again, here should be no direct writing of attributes from random places.  Your setup actually makes that part worse, since right now you can't write things that are handled by AAI, but your API nominally allows that....  So we need something to make sure people don't start abusing it and then we have to keep it forever.

src/app/data-model-interface/DataModel.h
andy31415: Updated code to keep mExpanded logic and notes about this being temporary updates and that code will change.

src/app/codegen-data-model/tests/TestAttributeReportIBsEncoding.cpp
andy31415: This is a test support file, because encoding/decoding is a pain. I am not sure how to convey that or if it would be practical to.

src/app/codegen-data-model/CodegenDataModel_Read.cpp
andy31415: Updated code to keep logic from ember-compatibility-functions and added unit tests for this.

src/app/codegen-data-model/CodegenDataModel_Read.cpp
andy31415: Documentation updated and usage of mExpanded vs error codes changed.
It seems only memory usage failures are relevant now (and this is what would prevent us currently from returning just a status code instead of a CHIP_ERROR.

src/app/codegen-data-model/tests/TestAttributeReportIBsEncoding.cpp
bzbarsky-apple, line 27:
```suggestion
CHIP_ERROR DecodedAttributeData::DecodeFrom(const AttributeDataIB::Parser & parser)
```

src/app/codegen-data-model/tests/TestAttributeReportIBsEncoding.cpp
bzbarsky-apple, line 1: So generally, files starting with Test contain actual tests, right?  Would it make sense to rename this to be clear it's not a test entrypoint?

bzbarsky-apple (approved):

src/app/codegen-data-model/tests/TestAttributeReportIBsEncoding.cpp
andy31415: Yes, I will rename to something more suitable (will have to figure out what).

andy31415: Renamed to not have Test prefix.

src/app/codegen-data-model/CodegenDataModel_Read.cpp
```diff
@@ -0,0 +1,370 @@
...
+/// Given a ByteSpan containing data from ember, interpret it
+/// as a span of type OUT (i.e. ByteSpan or CharSpan) given a ENCODING
+/// where ENCODING is Short or Long pascal strings.
+template <class OUT, class ENCODING>
+std::optional<OUT> ExtractEmberString(ByteSpan data)
+{
+    typename ENCODING::LengthType len;
+
+    VerifyOrDie(sizeof(len) <= data.size());
+    memcpy(&len, data.data(), sizeof(len));
```
tcarmelveilleux, line 163: This does not account for endianness, but IIRC the endianness is native for the 16 bit lengths. Suggest commenting on data format assumptjons for Pascal strings, especially since direct native bytes are copied into scalars.

src/app/codegen-data-model/CodegenDataModel_Read.cpp
```diff
@@ -0,0 +1,370 @@
...
+    record.endpoint                            = request.path.mEndpointId;
+    record.clusterId                           = request.path.mClusterId;
+    record.attributeId                         = request.path.mAttributeId;
+    Protocols::InteractionModel::Status status = emAfReadOrWriteAttribute(
+        &record, &attributeMetadata, gEmberAttributeIOBufferSpan.data(), static_cast<uint16_t>(gEmberAttributeIOBufferSpan.size()),
+        /* write = */ false);
+
+    if (status != Protocols::InteractionModel::Status::Success)
+    {
+        return ChipError(ChipError::SdkPart::kIMGlobalStatus, to_underlying(status), __FILE__, __LINE__);
```
tcarmelveilleux, line 363: This leaks file/line into binary. Suggest avoiding at least __FILE__

src/app/codegen-data-model/tests/AttributeReportIBEncodeDecode.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+    AttributePathIB::Parser pathParser;
+    ReturnErrorOnFailure(parser.GetPath(&pathParser));
+    ReturnErrorOnFailure(pathParser.GetConcreteAttributePath(attributePath, AttributePathIB::ValidateIdRanges::kNo));
+    ReturnErrorOnFailure(parser.GetData(&dataReader));
+
+    return CHIP_NO_ERROR;
+}
+
+CHIP_ERROR DecodeAttributeReportIBs(ByteSpan data, std::vector<DecodedAttributeData> & decoded_items)
```
tcarmelveilleux, line 39: What's the lifetime of the vector? Do those hang around a lot? This is one of the first instances of std::vector in core code.

tcarmelveilleux: Also, as a follow-up, suggest doing 2 pass read: first pass to count elements, then do a reserve on the vector, then actually process. This will reduce the amount of heap used transiently.

src/app/codegen-data-model/tests/AttributeReportIBEncodeDecode.h
```diff
@@ -0,0 +1,68 @@
...
+    CHIP_ERROR StartEncoding(app::AttributeReportIBs::Builder & builder);
+    CHIP_ERROR FinishEncoding(app::AttributeReportIBs::Builder & builder);
+
+    /// Decode the embedded attribute report IBs.
+    /// The TLVReaders inside data have a lifetime tied to the current object (its readers point
+    /// inside the current object)
+    CHIP_ERROR Decode(std::vector<DecodedAttributeData> & decoded_items);
+
+private:
+    uint8_t mTlvDataBuffer[1024];
```
tcarmelveilleux, line 61: Why this fixed buffer? This is very large. Does it impose a limitation on TCP max attribute size?

src/app/codegen-data-model/tests/EmberReadWriteOverride.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+#include <app/util/attribute-storage.h>
+
+using chip::Protocols::InteractionModel::Status;
+
+namespace {
+
+constexpr size_t kMaxTesetIoSize = 128;
+
+uint8_t gEmberIoBuffer[kMaxTesetIoSize];
```
tcarmelveilleux, line 27: Typo on "test"

src/app/codegen-data-model/tests/InteractionModelTemporaryOverrides.cpp
```diff
@@ -0,0 +1,83 @@
...
+#include <app/CommandHandler.h>
+#include <app/ConcreteAttributePath.h>
+#include <app/ConcreteCommandPath.h>
+#include <app/WriteHandler.h>
+#include <app/util/attribute-storage.h>
+
+using chip::Protocols::InteractionModel::Status;
+
+// TODO: most of the functions here are part of EmberCompatibilityFunctions and is NOT decoupled
+//       from IM current, but it SHOULD be
```
tcarmelveilleux, line 28: Add an issue number for the TODO

tcarmelveilleux (approved):

src/app/codegen-data-model/CodegenDataModel_Read.cpp
andy31415: Will add comments. this is indeed "ember internal format" which uses native byte order.

src/app/codegen-data-model/CodegenDataModel_Read.cpp
andy31415: Updated to use macros.

src/app/codegen-data-model/tests/AttributeReportIBEncodeDecode.cpp
andy31415: This is specificaly for unit tests, so I did not try hard to optimize it. I would do so only if we actually encounter issues (which is unlikely because when these ever run on devices, they would be individually separated otherwise they cannot really run with others due to ember overrides).

src/app/codegen-data-model/CodegenDataModel_Read.cpp
andy31415: Macros fail with `chip::Protocols::InteractionModel::Status status` is not const`

because our CHIP_SDK_ERROR are constexpr. I will have to see about a workaround.

andy31415: I added a separate macro that depends on error source config flags for this situation.

PR #33886, opened 2024-06-12
Generate provisional attribute marker in matter files

andy31415:
The IDL parser already supports this (and it supports this across ALL members and fields ... much more so than ZAP support).

Since ZAP supports this for attributes, make use of it.


examples/air-purifier-app/air-purifier-common/air-purifier-app.matter
```diff
@@ -567,9 +567,9 @@ cluster NetworkCommissioning = 49 {
   readonly attribute access(read: administer) nullable NetworkCommissioningStatusEnum lastNetworkingStatus = 5;
   readonly attribute access(read: administer) nullable octet_string<32> lastNetworkID = 6;
   readonly attribute access(read: administer) nullable int32s lastConnectErrorValue = 7;
-  readonly attribute optional WiFiBandEnum supportedWiFiBands[] = 8;
-  readonly attribute optional ThreadCapabilitiesBitmap supportedThreadFeatures = 9;
-  readonly attribute optional int16u threadVersion = 10;
+  provisional readonly attribute optional WiFiBandEnum supportedWiFiBands[] = 8;
```
tcarmelveilleux, line 570: I think these 3 are no longer provisional

tcarmelveilleux (approved):

examples/air-purifier-app/air-purifier-common/air-purifier-app.matter
andy31415: I think we should update this as a separate PR.

saurabhst (approved):

rbultman (approved):

jmartinez-silabs (approved):

PR #33008, opened 2024-04-16
Update test_metadata.py to use temporary files for test cases, for easier extensibility

vatsalghelani-csa:
Work for Issue #31317   

Follow up to PR #32752 
PR #4 - Modified the test script to make it more cleaner, and creating the temporary file required on the fly.

Files Worked:
- scripts/tests/py/test_metadata.py

PR #33888, opened 2024-06-12
Automated implementation of initially 1 python test to use Metadata script as an intermediate argument holder

vatsalghelani-csa:
Work for Issue #31317 

Follow up to PR #33008 

PR #5 - Automated implementation of initially running 1 python test to use Metadata script as an intermediate argument holder. Modified one of the python test script to now use the newer approach of extracting the arguments required through the new created metadata script and use it for execution.

Files Worked:
- .github/workflows/tests.yaml
- scripts/tests/py/metadata.py
- scripts/tests/run_python_test.py
- src/python_testing/TC_SC_3_6.py
- scripts/tests/py/test_metadata.py


scripts/tests/run_python_test.py
```diff
@@ -89,7 +90,34 @@ def DumpProgramOutputToQueue(thread_list: typing.List[threading.Thread], tag: st
...
+                script_args=script_args,
+                factoryreset=factoryreset,
+                factoryreset_app_only=factoryreset_app_only,
+                script_gdb=script_gdb,
+                quiet=quiet
+            )
+        ]
+
+    for run in runs:
+        print(f"Executing run: {run.py_script_path}")
```
andy31415, line 114: for multiple runs this will print out the same run several times.
Don't we have some discriminator somewhere? In your changed script this would be `run1`.

Ideally this should say something like `Executing TCS_SC_3_6.py run1` or just `Executing run1`

.github/workflows/tests.yaml
```diff
@@ -475,6 +475,16 @@ jobs:
                       build \
                       --copy-artifacts-to objdir-clone \
                    "
+            - name: Generate an argument environment file
+              run: |
+                  touch /tmp/test_env.yaml
```
andy31415, line 480:
```suggestion
                  rm -f /tmp/test_env.yaml
                  touch /tmp/test_env.yaml
```

andy31415: Thinking we should ensure file exists and is clear. Either rm -f first or echo empty to clear it.

scripts/tests/run_python_test.py
vatsalghelani-csa: Fixed.

vatsalghelani-csa: Fixed.

.github/workflows/tests.yaml
vatsalghelani-csa: Added and Fixed.

andy31415 (approved):

PR #33889, opened 2024-06-12
Changes to CommandSender and CommandHandler for supporting commands with large payloads

pidarped:


src/app/CommandHandlerExchangeInterface.h
```diff
@@ -102,6 +102,18 @@ class CommandHandlerExchangeInterface
...
+     * @brief Gets the maximum size of a packet buffer to encode a Command
+     * message. This size depends on the underlying session used by the
+     * exchange.
+     *
+     * The size returned here is the size not including the prepended headers.
+     *
+     * Called by CommandHandler when allocating buffer for encoding the Command
+     * response.
+     */
+    virtual size_t GetCommandMaxBufferSize() = 0;
```
bzbarsky-apple, line 116:
```suggestion
    virtual size_t GetResponseMaxBufferSize() = 0;
```

src/app/CommandHandlerExchangeInterface.h
```diff
@@ -102,6 +102,18 @@ class CommandHandlerExchangeInterface
      * Called by CommandHandler to relay this information to the requester.
      */
     virtual void ResponseDropped() = 0;
+
+    /**
+     * @brief Gets the maximum size of a packet buffer to encode a Command
+     * message. This size depends on the underlying session used by the
```
bzbarsky-apple, line 108: This is encoding a command response message, no?

src/app/CommandHandlerImpl.cpp
```diff
@@ -58,14 +58,17 @@ CommandHandlerImpl::~CommandHandlerImpl()
 
 CHIP_ERROR CommandHandlerImpl::AllocateBuffer()
 {
+    System::PacketBufferHandle commandPacket = nullptr;
```
bzbarsky-apple, line 61: This should be declared where first used, not up here.

src/app/CommandResponseSender.cpp
```diff
@@ -206,6 +206,10 @@ void CommandResponseSender::OnInvokeCommandRequest(Messaging::ExchangeContext *
     VerifyOrDieWithMsg(ec != nullptr, DataManagement, "Incoming exchange context should not be null");
     VerifyOrDieWithMsg(mState == State::ReadyForInvokeResponses, DataManagement, "state should be ReadyForInvokeResponses");
 
+    if (ec->GetSessionHandle()->AllowsLargePayload())
+    {
+        mUseLargePayloadBuffer = true;
```
bzbarsky-apple, line 211: This member seems to be write-only.  Why is it here?

src/app/CommandResponseSender.cpp
```diff
@@ -226,6 +230,21 @@ void CommandResponseSender::OnInvokeCommandRequest(Messaging::ExchangeContext *
     }
 }
 
+size_t CommandResponseSender::GetCommandMaxBufferSize()
+{
+    VerifyOrExit(mExchangeCtx && mExchangeCtx->HasSessionHandle(),
```
bzbarsky-apple, line 235: This would probably more readable as an actual if/log/return without the weird exit bits.

src/app/CommandSender.cpp
```diff
@@ -55,18 +55,19 @@ CHIP_ERROR GetRef(ParserT aParser, Optional<uint16_t> & aRef, bool commandRefReq
 } // namespace
 
 CommandSender::CommandSender(Callback * apCallback, Messaging::ExchangeManager * apExchangeMgr, bool aIsTimedRequest,
-                             bool aSuppressResponse) :
+                             bool aSuppressResponse, bool aRequireLargePayload) :
```
bzbarsky-apple, line 58: Require or Allow?  This seems more like an allow....

src/app/CommandHandlerExchangeInterface.h
pidarped: Updated name.

src/app/CommandHandlerExchangeInterface.h
pidarped: Yes. modified comment.

src/app/CommandHandlerImpl.cpp
pidarped: Done.

src/app/CommandResponseSender.cpp
pidarped: This was residual from the previous logic of using an out-of-band mechanism. Should not have been here.

src/app/CommandResponseSender.cpp
pidarped: Ok. Changed to an if-return.

src/app/CommandSender.cpp
pidarped: Changed to Allow

yufengwangca (approved):

src/app/CommandHandlerImpl.cpp
```diff
@@ -65,7 +65,10 @@ CHIP_ERROR CommandHandlerImpl::AllocateBuffer()
     {
         mCommandMessageWriter.Reset();
 
-        System::PacketBufferHandle commandPacket = System::PacketBufferHandle::New(chip::app::kMaxSecureSduLengthBytes);
+        System::PacketBufferHandle commandPacket = nullptr;
+        size_t commandBufferMaxSize              = mpResponder->GetCommandResponseMaxBufferSize();
+        commandPacket                            = System::PacketBufferHandle::New(commandBufferMaxSize);
```
bzbarsky-apple, line 70:
```suggestion
        const size_t commandBufferMaxSize  = mpResponder->GetCommandResponseMaxBufferSize();
        auto commandPacket.                           = System::PacketBufferHandle::New(commandBufferMaxSize);
```

Why the existence on declaring things before they are needed?

bzbarsky-apple (approved):

src/app/CommandHandlerImpl.cpp
pidarped: Fixed.

PR #33895, opened 2024-06-13
[NXP][platform] Fixing RAM buffer alignment issue on NXP platforms

chapongatien:
Issue description: 

An issue may happen on NXP platforms supporting Matter with the RAM storage module + littlefs due to a RAM buffer alignment issue preventing the device to load matter file system data in RAM.

Fix proposed: 

As the NXP mflash_drv_read function mandates the alignment of dst read buffer to 4 bytes, make sure to always align read buffers to 4 bytes.

### DIFF
--- a/src/platform/nxp/common/NXPConfig.cpp
+++ b/src/platform/nxp/common/NXPConfig.cpp
@@ -84,14 +84,14 @@ typedef struct
 {
     uint16_t chipConfigRamBufferLen;
     uint16_t padding;
-    uint8_t chipConfigRamBuffer[CONFIG_CHIP_NVM_RAMBUFFER_SIZE_KEY_INT];
+    uint8_t chipConfigRamBuffer[CONFIG_CHIP_NVM_RAMBUFFER_SIZE_KEY_INT] __attribute__((aligned(4)));
 } ChipConfigRamStructKeyInt;
 
 typedef struct
 {
     uint16_t chipConfigRamBufferLen;
     uint16_t padding;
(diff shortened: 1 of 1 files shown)


Damian-Nordic (approved):

bzbarsky-apple (approved):

src/platform/nxp/common/NXPConfig.cpp
```diff
@@ -84,14 +84,14 @@ typedef struct
...
-    uint8_t chipConfigRamBuffer[CONFIG_CHIP_NVM_RAMBUFFER_SIZE_KEY_INT];
+    uint8_t chipConfigRamBuffer[CONFIG_CHIP_NVM_RAMBUFFER_SIZE_KEY_INT] __attribute__((aligned(4)));
 } ChipConfigRamStructKeyInt;
 
 typedef struct
 {
     uint16_t chipConfigRamBufferLen;
     uint16_t padding;
-    uint8_t chipConfigRamBuffer[CONFIG_CHIP_NVM_RAMBUFFER_SIZE_KEY_STRING];
+    uint8_t chipConfigRamBuffer[CONFIG_CHIP_NVM_RAMBUFFER_SIZE_KEY_STRING] __attribute__((aligned(4)));
```
bzbarsky-apple, line 94: That said, the relevant places that rely on the alignment should be static_asserting that alignof(whatever) is correct.

chapongatien: Yes agree, I added an additional check to valid it.

bzbarsky-apple (approved):

PR #33896, opened 2024-06-13
Remove unnecessary error log from CurrentFabricRemover

agners:
Currently, when the CurrentFabricRemover is successful it logs the following error with err log level:

Remove Current Fabric Result : src/controller/CurrentFabricRemover.cpp:133: Success

Get rid of the message if successful, it is confusing.

### DIFF
--- a/src/controller/CurrentFabricRemover.cpp
+++ b/src/controller/CurrentFabricRemover.cpp
@@ -146,7 +146,9 @@ void CurrentFabricRemover::OnCommandFailure(void * context, CHIP_ERROR err)
 void CurrentFabricRemover::FinishRemoveCurrentFabric(void * context, CHIP_ERROR err)
 {
     if (err != CHIP_NO_ERROR)
+    {
         ChipLogError(Controller, "Remove Current Fabric Failed : %" CHIP_ERROR_FORMAT, err.Format());
+    }
     auto * self     = static_cast<CurrentFabricRemover *>(context);
     self->mNextStep = Step::kAcceptRemoveFabricStart;
     if (self->mCurrentFabricRemoveCallback != nullptr)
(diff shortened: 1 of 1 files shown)


src/controller/CurrentFabricRemover.cpp
```diff
@@ -145,7 +145,8 @@ void CurrentFabricRemover::OnCommandFailure(void * context, CHIP_ERROR err)
 
 void CurrentFabricRemover::FinishRemoveCurrentFabric(void * context, CHIP_ERROR err)
 {
-    ChipLogError(Controller, "Remove Current Fabric Result : %" CHIP_ERROR_FORMAT, err.Format());
+    if (err != CHIP_NO_ERROR)
```
Damian-Nordic, line 148: Please use `{}` for the if-block.

Damian-Nordic (approved):

src/controller/CurrentFabricRemover.cpp
agners: Addressed!

arkq (approved):

src/controller/CurrentFabricRemover.cpp
```diff
@@ -145,7 +145,10 @@ void CurrentFabricRemover::OnCommandFailure(void * context, CHIP_ERROR err)
...
```
bzbarsky-apple, line 148: @agners In the success case we should have left a progress-level log saying it completed successfully...  Please put that back.

agners: Ok, #33914 adds a progress-log level on success.

bzbarsky-apple: Thank you!

PR #33897, opened 2024-06-13
Enable additional adv data on Linux as a reference platform

arkq:
### Problem

Linux is a reference platform for verification purposes, so all features should be able to compile and functional on that platform.

### Changes

- enable additional advertising data on lighting-app
- use `g_memdup2()`, since `g_memdup()` is deprecated since glib 2.68

### Testing

CI will verify potential build breaks.

### DIFF
--- a/examples/chip-tool/args.gni
+++ b/examples/chip-tool/args.gni
@@ -29,6 +29,7 @@ matter_log_json_payload_hex = true
 matter_log_json_payload_decode_full = true
 
 chip_enable_additional_data_advertising = true
+chip_enable_rotating_device_id = true
 
 # make chip-tool very strict by default
 chip_tlv_validate_char_string_on_read = true
--- a/examples/chip-tool/args.gni
+++ b/examples/chip-tool/args.gni
@@ -28,6 +28,8 @@ matter_enable_tracing_support = true
 matter_log_json_payload_hex = true
 matter_log_json_payload_decode_full = true
 
+chip_enable_additional_data_advertising = true
+
 # make chip-tool very strict by default
 chip_tlv_validate_char_string_on_read = true
 chip_tlv_validate_char_string_on_write = true
--- a/examples/chip-tool/args.gni
+++ b/examples/chip-tool/args.gni
@@ -28,9 +28,6 @@ matter_enable_tracing_support = true
 matter_log_json_payload_hex = true
 matter_log_json_payload_decode_full = true
 
-chip_enable_additional_data_advertising = true
-chip_enable_rotating_device_id = true
-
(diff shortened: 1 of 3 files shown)


Damian-Nordic (approved):

andy31415:
@arkq we have a compiler error:

```
INFO    ../../examples/chip-tool/third_party/connectedhomeip/src/platform/Linux/bluez/BluezEndpoint.cpp:420:59: error: implicit conversion loses integer precision: 'size_t' (aka 'unsigned long') to 'guint' (aka 'unsigned int') [-Werror,-Wshorten-64-to-32]
INFO      420 |     data = g_memdup2(bufferHandle->Start(), bufferHandle->DataLength());
```

examples/chip-tool/args.gni
```diff
@@ -28,6 +28,9 @@ matter_enable_tracing_support = true
 matter_log_json_payload_hex = true
 matter_log_json_payload_decode_full = true
 
+chip_enable_additional_data_advertising = true
+chip_enable_rotating_device_id = true
```
bzbarsky-apple, line 32: How does this make sense for chip-tool at all?

arkq: I've enabled it for the chip-tool since it's a controller. I thought that it might make sense to enable rotating device ID in the server part and the client part, so CI could verify potential build breaks. To be honest I'm not sure whether there is any command in chip-tool which can read this ID, but if there is none, it could be added, so the functionality could be tested.

bzbarsky-apple: I think my point is that this compile flag was supposed to control whether a commissionable device has a rotating device id.  chip-tool is not a commissionable device....

I guess someone started using this flag in `AdditionalDataPayloadGenerator::generateAdditionalDataPayload`, which is not about commissionables.  So now what this flag means is anyone's guess.

arkq: > I think my point is that this compile flag was supposed to control whether a commissionable device has a rotating device id. chip-tool is not a commissionable device....

OK, I see. So, maybe I could fix the usage of that flag as well. So, one more question. Since this flag is only for commissionable devices, than the support for reading such ID can/shall be in chip-tool regardless of any flag?

And as for the chip-tool. Maybe my wording was not right. I do not want to enable generation of rotating ID in the chip-tool, but only a support for reading it and printing to user (in logs or via dedicated command).

arkq: I've read the code more carefully and indeed you are right. These options do not make any sense for chip-tool. Many thanks for pointing that out!

andy31415 (approved):

PR #33898, opened 2024-06-13
[NXP] rw61x openthread and BT updates

Martin-NXP:
* Expose API to get or set reset in idle and add call to otPlatSetResetFunction to set reset function in OT to application function
*  Defining OPENTHREAD_CONFIG_PLATFORM_RADIO_COEX_ENABLE as 0 by default. OPENTHREAD_CONFIG_PLATFORM_RADIO_COEX_ENABLE was defined as 1 in the past as a temporary workaround due to some ot files dependences, no longer needed, so redefining OPENTHREAD_CONFIG_PLATFORM_RADIO_COEX_ENABLE as 0 to avoid having unuseful commands within the otcli/matter-cli
* Enabling otcli ping command in the matter-cli
* Change the BLE name for TC-DD-2.1

### DIFF
--- a/examples/platform/nxp/rt/rw61x/app/project_include/openthread/OpenThreadConfig.h
+++ b/examples/platform/nxp/rt/rw61x/app/project_include/openthread/OpenThreadConfig.h
@@ -65,12 +65,9 @@
 #define OPENTHREAD_CONFIG_SRP_CLIENT_ENABLE 1
 #define OPENTHREAD_CONFIG_ECDSA_ENABLE 1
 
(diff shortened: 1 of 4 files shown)


andy31415:
@Martin-NXP did you update the submodules intentionally? If so please set the label and explain the change in the PR summary.

We added failsafes for updating submodules by mistake via git commit -a, and CI complains.

andy31415 (approved):

andy31415:
Assuming third-party changes are intentional as I see individual files updated rather than a submodule.

PR #33902, opened 2024-06-13
Demand Response Load Control - xml

steveholmgren:
XML PR with drlc-cluster.xml that it is believed matches the spec


andy31415:
@steveholmgren it looks like commit history has different identities and CLA is unhappy. Could you fix?

Apart from that, XML does not seem valid and probably needs some fixes. You may want to use alchemy to generate the xml (@hasty has a tool that can generate XMLs out of spec text).

src/app/zap-templates/zcl/data-model/chip/drlc-cluster.xml
```diff
@@ -87,9 +72,6 @@ limitations under the License.
     <item name="Level3" value="0x04"/>
```
andy31415, line 72: These bitmap values are suspect: are they supposed to be a bitmap or are they enum incremental values?

src/app/zap-templates/zcl/data-model/chip/drlc-cluster.xml
```diff
@@ -1,42 +1,27 @@
...
 See the License for the specific language governing permissions and
 limitations under the License.
 -->
 <configurator>
-  <domain name="Energy Management"/>
-
+  <domain name="Energy"/>
 
-  <bitmap name="DeviceClassBitmap" type="bitmap32">
+  <bitmap name="Feature" type="bitmap32">
```
andy31415, line 17: Updated format uses features rather than a bitmap named feature.

See https://github.com/project-chip/connectedhomeip/blob/master/src/app/zap-templates/zcl/data-model/chip/onoff-cluster.xml#L58

src/app/zap-templates/zcl/data-model/chip/drlc-cluster.xml
steveholmgren: they are supposed to be enums - <enum name="CrititcalityLevelEnum" type="enum8">
                                                            <item name="Unknown" value="0x00"/>
                                                            <item name="Received" value="0x01"/>
                                                          </enum>
inside and enum tag.  Do I have the sub tags labeled wrong as "item name" and should be something else like "enum name".  Ill look into generating the xml with Hasty's app.  Thanks

src/app/zap-templates/zcl/data-model/chip/drlc-cluster.xml
steveholmgren: Device class was removed from the latest spec and most recent XML - let me make sure I have the latest xml in the PR.

bzbarsky-apple (changes requested):

src/app/zap-templates/zcl/data-model/chip/drlc-cluster.xml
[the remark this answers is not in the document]
steveholmgren: Working to update drlc-cluster.xml using Hasty's alchemy and might have pushed older xml.  Alchemy looks good, but having issues with output so will fix, but need to get alchemy execution and output corrected first and then will update here.  Apologies for the firedrill on this one

andy31415:
Closing stale PR.

PR #33904, opened 2024-06-13
Add new AddStatus overloads using ClusterStatusCode

### LINKED ISSUE #31120: Cluster-specific codes cannot be passed to CommandHandler::AddStatus
There is an error-prone method to pass cluster codes to CommandHandler status when responding to a command. This prevents building code that has a single return type, but can return both Status and cluster-specific code.

This causes code like this:

```
    if (status.HasValue())
    {
        ChipLogError(Zcl, "Failed to open commissioning window. Cluster status 0x%02x", to_underlying(status.Value()));
        commandObj->AddClusterSpecificFailure(commandPath, to_underlying(status.Value()));
    }
    else
    {
        if (globalStatus != InteractionModel::Status::Success)
        {
            ChipLogError(Zcl, "Failed to open commissioning window. Global status " ChipLogFormatIMStatus,
                         ChipLogValueIMStatus(globalStatus));
        }
        commandObj->AddStatus(commandPath, globalStatus);
    }
```

rather than just different values of `status` set (with or without cluster code), and then `commandObj->AddStatus(commandPath, status)`.

tcarmelveilleux:
- CommandHandler and WriteHandler did not have a way to cleanly just `AddStatus` with a cluster-specific status code (to eventually harmonize handling methods to always return just a ClusterStatusCode).

This PR:

- Introduces an `AddStatus` overload for `ClusterStatusCode` to CommandHandler and WriteHandler.
- Removes the unimplemented `WriteClient::Shutdown` method.
- Adds notes to the `AddClusterSpecificSuccess` and `AddClusterSpecificFailure`.
- Removes implicit conversion from `ClusterStatusCode` to `Status`
  - Was never used and was error-prone/lossy

Fixes #31120

Testing done:
- Updated necessary unit tests.
- Added unit tests for cluster specific statuses on writes.
- Integration tests still pass.

### DIFF
--- a/src/app/CommandHandlerImpl.cpp
+++ b/src/app/CommandHandlerImpl.cpp
@@ -30,8 +30,8 @@
 #include <lib/support/TypeTraits.h>
 #include <messaging/ExchangeContext.h>
 #include <platform/LockTracker.h>
-#include <protocols/secure_channel/Constants.h>
 #include <protocols/interaction_model/StatusCode.h>
+#include <protocols/secure_channel/Constants.h>
 
 namespace chip {
 namespace app {
@@ -593,8 +593,8 @@ CHIP_ERROR CommandHandlerImpl::AddStatusInternal(const ConcreteCommandPath & aCo
     return TryAddingResponse([&]() -> CHIP_ERROR { return TryAddStatusInternal(aCommandPath, aStatus); });
 }
 
-void CommandHandlerImpl::AddStatus(const ConcreteCommandPath & aCommandPath, const Protocols::InteractionModel::ClusterStatusCode& status,
-                                   const char * context)
+void CommandHandlerImpl::AddStatus(const ConcreteCommandPath & aCommandPath,
+                                   const Protocols::InteractionModel::ClusterStatusCode & status, const char * context)
 {
 
     CHIP_ERROR error = FallibleAddStatus(aCommandPath, status, context);
@@ -611,7 +611,8 @@ void CommandHandlerImpl::AddStatus(const ConcreteCommandPath & aCommandPath, con
     }
 }
 
-CHIP_ERROR CommandHandlerImpl::FallibleAddStatus(const ConcreteCommandPath & path, const Protocols::InteractionModel::ClusterStatusCode& status,
+CHIP_ERROR CommandHandlerImpl::FallibleAddStatus(const ConcreteCommandPath & path,
+                                                 const Protocols::InteractionModel::ClusterStatusCode & status,
                                                  const char * context)
 {
     if (!status.IsSuccess())
@@ -629,10 +630,11 @@ CHIP_ERROR CommandHandlerImpl::FallibleAddStatus(const ConcreteCommandPath & pat
 
     if (status.HasClusterSpecificCode())
     {
-      return AddStatusInternal(path, StatusIB{(status.IsSuccess() ? Status::Success : Status::Failure), status.GetClusterSpecificCode().Value()});
+        return AddStatusInternal(
+            path, StatusIB{ (status.IsSuccess() ? Status::Success : Status::Failure), status.GetClusterSpecificCode().Value() });
     }
 
-    return AddStatusInternal(path, StatusIB{status.GetStatus()});
+    return AddStatusInternal(path, StatusIB{ status.GetStatus() });
 }
 
 CHIP_ERROR CommandHandlerImpl::PrepareInvokeResponseCommand(const ConcreteCommandPath & aResponseCommandPath,
--- a/src/app/CommandHandlerImpl.cpp
+++ b/src/app/CommandHandlerImpl.cpp
@@ -31,6 +31,7 @@
 #include <messaging/ExchangeContext.h>
 #include <platform/LockTracker.h>
 #include <protocols/secure_channel/Constants.h>
+#include <protocols/interaction_model/StatusCode.h>
 
 namespace chip {
 namespace app {
@@ -592,11 +593,11 @@ CHIP_ERROR CommandHandlerImpl::AddStatusInternal(const ConcreteCommandPath & aCo
     return TryAddingResponse([&]() -> CHIP_ERROR { return TryAddStatusInternal(aCommandPath, aStatus); });
 }
 
-void CommandHandlerImpl::AddStatus(const ConcreteCommandPath & aCommandPath, const Protocols::InteractionModel::Status aStatus,
+void CommandHandlerImpl::AddStatus(const ConcreteCommandPath & aCommandPath, const Protocols::InteractionModel::ClusterStatusCode& status,
                                    const char * context)
 {
 
-    CHIP_ERROR error = FallibleAddStatus(aCommandPath, aStatus, context);
+    CHIP_ERROR error = FallibleAddStatus(aCommandPath, status, context);
 
     if (error != CHIP_NO_ERROR)
     {
@@ -610,10 +611,10 @@ void CommandHandlerImpl::AddStatus(const ConcreteCommandPath & aCommandPath, con
     }
 }
 
-CHIP_ERROR CommandHandlerImpl::FallibleAddStatus(const ConcreteCommandPath & path, const Protocols::InteractionModel::Status status,
+CHIP_ERROR CommandHandlerImpl::FallibleAddStatus(const ConcreteCommandPath & path, const Protocols::InteractionModel::ClusterStatusCode& status,
                                                  const char * context)
 {
-    if (status != Status::Success)
+    if (!status.IsSuccess())
     {
         if (context == nullptr)
         {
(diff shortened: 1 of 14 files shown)


src/app/CommandHandlerImpl.cpp
```diff
@@ -623,20 +625,16 @@ CHIP_ERROR CommandHandlerImpl::FallibleAddStatus(const ConcreteCommandPath & pat
         ChipLogError(DataManagement,
                      "Endpoint=%u Cluster=" ChipLogFormatMEI " Command=" ChipLogFormatMEI " status " ChipLogFormatIMStatus " (%s)",
                      path.mEndpointId, ChipLogValueMEI(path.mClusterId), ChipLogValueMEI(path.mCommandId),
-                     ChipLogValueIMStatus(status), context);
+                     ChipLogValueIMStatus(status.GetStatus()), context);
```
andy31415, line 628: on failure, now that this is a cluster specific status code, should we also log the cluster status?

src/app/CommandHandlerImpl.cpp
```diff
@@ -623,20 +625,16 @@ CHIP_ERROR CommandHandlerImpl::FallibleAddStatus(const ConcreteCommandPath & pat
...
-}
-
-CHIP_ERROR CommandHandlerImpl::AddClusterSpecificSuccess(const ConcreteCommandPath & aCommandPath, ClusterStatus aClusterStatus)
-{
-    return AddStatusInternal(aCommandPath, StatusIB(Status::Success, aClusterStatus));
-}
+    if (status.HasClusterSpecificCode())
+    {
+        return AddStatusInternal(
+            path, StatusIB{ (status.IsSuccess() ? Status::Success : Status::Failure), status.GetClusterSpecificCode().Value() });
```
andy31415, line 634: This relies on the fact that if cluster specific code exists, then status can only be success/failure.

Are the other states invalid? I.e. can't I have `Status::InvalidAction, 1234` as a `status + cluster_status` pair?
Some comments explaining why this is correct maybe referencing spec items would be helpful. Otherwise I would expect to not lose information and have `StatusIB{ status.GetStatuis(), status.GetClusterSpecificCode().Value())`

when (if?) we convert to std::optional compiler may not ralize that HasClusterSpecifiCode is directly checking for value existence, so maybe:


```cpp
chip::Optional<ClusterStatus> clusterStatus = status.GetClusterSpecificCode();
if (clusterStatus.HasValue()) {
    return AddStatusInternal(
         path, 
         StatusIB{ (status.IsSuccess() ? Status::Success : Status::Failure), 
                    clusterStatus.Value() 
                  });
}
```

Is clearer on intent.

src/app/CommandResponseHelper.h
```diff
@@ -40,9 +41,22 @@ class CommandResponseHelper
         return CHIP_NO_ERROR;
     };
 
+    // WARNING: Do not use for basic response-payload-free success, as SUCCESS status
```
andy31415, line 44: I am not sure I completely follow this warning. Could you expand with some examples of legal and illegal (and why) usages?

src/app/CommandResponseHelper.h
```diff
@@ -40,9 +41,22 @@ class CommandResponseHelper
...
+    // code should be handled without a cluster-specific status.
     CHIP_ERROR Success(ClusterStatus aClusterStatus)
     {
-        CHIP_ERROR err = mCommandHandler->AddClusterSpecificSuccess(mCommandPath, aClusterStatus);
+        CHIP_ERROR err = mCommandHandler->FallibleAddStatus(
+            mCommandPath, Protocols::InteractionModel::ClusterStatusCode::ClusterSpecificSuccess(aClusterStatus));
+        if (err == CHIP_NO_ERROR)
+        {
+            mSentResponse = true;
+        }
```
andy31415, line 53: I wonder why this was not `mSentResponse = (err == CHIP_NO_ERROR);`? Does it assume mSentResponse may have been true already and we do not want to change that?

Do you happen to know the existing logic reason?

src/app/WriteHandler.cpp
```diff
@@ -616,22 +616,29 @@ Status WriteHandler::ProcessWriteRequest(System::PacketBufferHandle && aPayload,
...
 
-CHIP_ERROR WriteHandler::AddStatus(const ConcreteDataAttributePath & aPath, const Protocols::InteractionModel::Status aStatus)
+CHIP_ERROR WriteHandler::AddStatus(const ConcreteDataAttributePath & aPath,
+                                   const Protocols::InteractionModel::ClusterStatusCode & aStatus)
 {
-    return AddStatus(aPath, StatusIB(aStatus));
+    if (aStatus.HasClusterSpecificCode())
+    {
+        return AddStatusInternal(
+            aPath, StatusIB{ (aStatus.IsSuccess() ? Status::Success : Status::Failure), aStatus.GetClusterSpecificCode().Value() });
```
andy31415, line 625: same comment here about why this would be legal. Since used in two places, maybe it would make sense to re-use code? We seem to generally want to get an StatusIB from a ClusterStatusCode.

src/controller/tests/data_model/TestWrite.cpp
```diff
@@ -43,15 +44,20 @@ constexpr EndpointId kTestEndpointId       = 1;
...
     kSendAttributeSuccess,
     kSendAttributeError,
     kSendMultipleSuccess,
     kSendMultipleErrors,
+    kSendClusterSpecificSuccess,
+    kSendClusterSpecificFailure,
 };
 
-ResponseDirective responseDirective;
+ResponseDirective gResponseDirective;
```
andy31415, line 60: This global is not really reset so I wonder if this makes tests have sideffects.

Could we initialize it to something explicitly (guessing it gets success now because BSS is 0 by default) and use https://github.com/project-chip/connectedhomeip/blob/master/src/lib/support/Scoped.h#L32 ?

I also realize this is not the scope of your PR really... so if you feel this is better later/separate that is fine as well.

andy31415 (approved):

src/app/CommandHandlerImpl.cpp
tcarmelveilleux: We could

src/controller/tests/data_model/TestWrite.cpp
tcarmelveilleux: :)

src/controller/tests/data_model/TestCommands.cpp
```diff
@@ -489,7 +492,10 @@ TEST_F(TestCommands, TestSuccessNoDataResponseWithClusterStatus)
 
     // Passing of stack variables by reference is only safe because of synchronous completion of the interaction. Otherwise, it's
     // not safe to do so.
-    auto onFailureCb = [&onFailureWasCalled](CHIP_ERROR aError) { onFailureWasCalled = true; };
+    auto onFailureCb = [&onFailureWasCalled](CHIP_ERROR aError) {
+        printf("   ----> %" CHIP_ERROR_FORMAT "\n", aError.Format());
```
bzbarsky-apple, line 496: Why is this printf bit here?

More generally, why were any changes needed in this file?

src/protocols/interaction_model/StatusCode.h
```diff
@@ -99,14 +97,19 @@ class ClusterStatusCode
      *             (e.g. chip::app::Clusters::AdministratorCommissioning::CommissioningWindowStatusEnum::kWindowNotOpen)
      * @return a ClusterStatusCode instance properly configured.
      */
-    template <typename T>
+    template <typename T, std::enable_if_t<std::is_enum<T>::value, bool> = true>
```
bzbarsky-apple, line 100:
```suggestion
    template <typename T, typename std::enable_if_t<std::is_enum<T>::value, bool> = true>
```

src/protocols/interaction_model/StatusCode.h
```diff
@@ -116,14 +119,19 @@ class ClusterStatusCode
      *             (e.g. chip::app::Clusters::AdministratorCommissioning::CommissioningWindowStatusEnum::kBasicWindowOpen)
      * @return a ClusterStatusCode instance properly configured.
      */
-    template <typename T>
+    template <typename T, std::enable_if_t<std::is_enum<T>::value, bool> = true>
```
bzbarsky-apple, line 122:
```suggestion
    template <typename T, typename std::enable_if_t<std::is_enum<T>::value, bool> = true>
```

bzbarsky-apple (approved):

src/controller/tests/data_model/TestCommands.cpp
tcarmelveilleux: Missed it when I was debugging something else related to it.

src/app/WriteHandler.cpp
tcarmelveilleux: Done

src/controller/tests/data_model/TestWrite.cpp
tcarmelveilleux: Done

src/app/CommandHandlerImpl.cpp
tcarmelveilleux: Done

PR #33905, opened 2024-06-13
[Python] Make Commissioning APIs more pythonic and consistent

agners:
This commit makes the commissioning APIs more pythonic and consistent by not returning PyChipError but simply raising ChipStackError exceptions instead.

### DIFF
--- a/src/controller/python/chip/ChipDeviceCtrl.py
+++ b/src/controller/python/chip/ChipDeviceCtrl.py
@@ -230,11 +230,14 @@ class CommissionableNode(discovery.CommissionableNode):
     def SetDeviceController(self, devCtrl: 'ChipDeviceController'):
         self._devCtrl = devCtrl
 
-    def Commission(self, nodeId: int, setupPinCode: int) -> PyChipError:
+    def Commission(self, nodeId: int, setupPinCode: int) -> int:
         ''' Commission the device using the device controller discovered this device.
 
         nodeId: The nodeId commissioned to the device
         setupPinCode: The setup pin code of the device
+
+        Returns:
+            - Effective Node ID of the device (as defined by the assigned NOC)
         '''
         return self._devCtrl.CommissionOnNetwork(
             nodeId, setupPinCode, filterType=discovery.FilterType.INSTANCE_NAME, filter=self.instanceName)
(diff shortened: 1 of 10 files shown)


src/controller/python/chip/ChipDeviceCtrl.py
```diff
@@ -545,11 +548,7 @@ def ConnectBLE(self, discriminator, setupPinCode, nodeid) -> PyChipError:
                     self.devCtrl, discriminator, setupPinCode, nodeid)
             ).raise_on_error()
 
-            # TODO: Change return None. Only returning on success is not useful.
-            # but that is what the previous implementation did.
-            res = self._commissioning_complete_future.result()
-            res.raise_on_error()
-            return res
+            return self._commissioning_complete_future.result()
```
arkq, line 551: Function marked as return None, but your are returning something here.

src/controller/python/chip/ChipDeviceCtrl.py
```diff
@@ -2026,9 +2026,11 @@ def CommissionOnNetwork(self, nodeId: int, setupPinCode: int,
         finally:
             self._commissioning_complete_future = None
 
-    def CommissionWithCode(self, setupPayload: str, nodeid: int, discoveryType: DiscoveryType = DiscoveryType.DISCOVERY_ALL) -> PyChipError:
+    def CommissionWithCode(self, setupPayload: str, nodeid: int, discoveryType: DiscoveryType = DiscoveryType.DISCOVERY_ALL) -> None:
```
arkq, line 2029: This function also returns something.

arkq (approved):

src/controller/python/chip/ChipDeviceCtrl.py
agners: I do set the result of `_commissioning_complete_future` to `None`, so this technically should be fine. I was considering both directions, I felt this will make it easier to change the return value.

But it has the risk of accidental return value change. And given you would not expect a return statement here, it is probably better to get rid of it :sweat_smile:

src/controller/python/chip/ChipDeviceCtrl.py
```diff
@@ -533,7 +536,7 @@ def IsConnected(self):
                 self.devCtrl)
         )
 
-    def ConnectBLE(self, discriminator, setupPinCode, nodeid) -> PyChipError:
+    def ConnectBLE(self, discriminator, setupPinCode, nodeid) -> None:
```
tcarmelveilleux, line 539: Watch-out, this will conflict with https://github.com/project-chip/connectedhomeip/pull/33933

tcarmelveilleux (approved):

src/controller/python/chip/ChipDeviceCtrl.py
agners: Thanks for the heads up, will rebase!

agners:
FYI, I've rebased this on top of master to get the changes from #33954. This makes the implementation of `src/python_testing/TC_CGEN_2_4.py` cleaner (which requires the PyChipError instance to get insights in the exact error).

PR #33909, opened 2024-06-13
[Fabric Bridge] Overwrite the default implementation of Administrator Commissioning C…

### LINKED ISSUE #33764: [Feature] Overwrite the default implementation of Administrator Commissioning Cluster for bridged devices
### Feature description

Each Bridged Node contains Administrator Commissioning Cluster, we need to overwrite the default implementation ([https://github.com/project-chip/connectedhomeip/blob/master/src/app/clusters/admin[…]tor-commissioning-server/administrator-commissioning-server.cpp](https://github.com/project-chip/connectedhomeip/blob/master/src/app/clusters/administrator-commissioning-server/administrator-commissioning-server.cpp#L143))  in the Bridged Node endpoint, such as we it receive OCW command, instead of default operations, it needs to notify the Fabric Admin to open the commissioning window

### Platform

other

### Platform Version(s)

_No response_

### Anything else?

_No response_

yufengwangca:
Each Bridged Node contains Administrator Commissioning Cluster, we need to overwrite the default implementation ([https://github.com/project-chip/connectedhomeip/blob/master/src/app/clusters/admin[…]tor-commissioning-server/administrator-commissioning-server.cpp](https://github.com/project-chip/connectedhomeip/blob/master/src/app/clusters/administrator-commissioning-server/administrator-commissioning-server.cpp#L143)) in the Bridged Node endpoint, such as we it receive OCW command, instead of default operations, it needs to notify the Fabric Admin to open the commissioning window

Fixes #33764

### DIFF
--- a/examples/fabric-bridge-app/linux/main.cpp
+++ b/examples/fabric-bridge-app/linux/main.cpp
@@ -130,6 +130,8 @@ void AdministratorCommissioningCommandHandler::InvokeCommand(HandlerContext & ha
 
 #if defined(PW_RPC_FABRIC_BRIDGE_SERVICE) && PW_RPC_FABRIC_BRIDGE_SERVICE
     Device * device = DeviceMgr().GetDevice(endpointId);
+
+    // TODO: issues:#33784, need to make OpenCommissioningWindow synchronous
     if (device != nullptr && OpenCommissioningWindow(device->GetNodeId()) == CHIP_NO_ERROR)
     {
         ChipLogProgress(NotSpecified, "Commissioning window is now open");
@@ -139,6 +141,9 @@ void AdministratorCommissioningCommandHandler::InvokeCommand(HandlerContext & ha
         status = Status::Failure;
         ChipLogProgress(NotSpecified, "Commissioning window is failed to open");
     }
+#else
+    status = Status::Failure;
+    ChipLogProgress(NotSpecified, "Commissioning window failed to open: PW_RPC_FABRIC_BRIDGE_SERVICE not defined");
 #endif // defined(PW_RPC_FABRIC_BRIDGE_SERVICE) && PW_RPC_FABRIC_BRIDGE_SERVICE
 
     handlerContext.mCommandHandler.AddStatus(handlerContext.mRequestPath, status);
--- a/examples/fabric-bridge-app/linux/main.cpp
+++ b/examples/fabric-bridge-app/linux/main.cpp
@@ -36,6 +36,10 @@
 
 using namespace chip;
 
+using namespace chip::app;
+using namespace chip::app::Clusters;
+using namespace chip::app::Clusters::AdministratorCommissioning;
+
 #define ZCL_DESCRIPTOR_CLUSTER_REVISION (1u)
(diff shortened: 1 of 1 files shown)


examples/fabric-bridge-app/linux/main.cpp
```diff
@@ -97,10 +101,57 @@ void AttemptRpcClientConnect(System::Layer * systemLayer, void * appState)
...
+    if (device != nullptr && OpenCommissioningWindow(device->GetNodeId()) == CHIP_NO_ERROR)
+    {
+        ChipLogProgress(NotSpecified, "Commissioning window is now open");
+    }
+    else
+    {
+        status = Status::Failure;
+        ChipLogProgress(NotSpecified, "Commissioning window is failed to open");
+    }
+#endif // defined(PW_RPC_FABRIC_BRIDGE_SERVICE) && PW_RPC_FABRIC_BRIDGE_SERVICE
```
tehampson, line 142: nit, Should there be an `#else` here that sets status to a failure command

examples/fabric-bridge-app/linux/main.cpp
```diff
@@ -97,10 +101,57 @@ void AttemptRpcClientConnect(System::Layer * systemLayer, void * appState)
...
+    }
+
+    handlerContext.SetCommandHandled();
+    Status status = Status::Success;
+
+#if defined(PW_RPC_FABRIC_BRIDGE_SERVICE) && PW_RPC_FABRIC_BRIDGE_SERVICE
+    Device * device = DeviceMgr().GetDevice(endpointId);
+    if (device != nullptr && OpenCommissioningWindow(device->GetNodeId()) == CHIP_NO_ERROR)
+    {
+        ChipLogProgress(NotSpecified, "Commissioning window is now open");
```
tehampson, line 135: nit, is the commissioning window now open or has the command been sent to open the window?

My understanding is you can call OpenCommissioningWindow with an invalid node id and get a success here because the RPC is sent successfully, but it doesn't imply that a window is actually opened

yufengwangca: Good catch, if you do not set mDevices[index] back to nullptr when returning -1 due to an error, it can leave the mDevices array in an inconsistent state, where mDevices[index] points to a device that wasn't properly added.

This issue also exist in Bridge Example App Where I copy from, https://github.com/project-chip/connectedhomeip/issues/33946

examples/fabric-bridge-app/linux/main.cpp
yufengwangca: Yes, it would be a good idea to add an #else block to handle the case where PW_RPC_FABRIC_BRIDGE_SERVICE is not defined

examples/fabric-bridge-app/linux/main.cpp
yufengwangca: Added comments, we need to make OpenCommissioningWindow synchronous to keep the following logic correct. But I need to integrate this API to CHIP https://pigweed.dev/pw_rpc/#client-synchronous-call-wrappers.

If not feasible, we need to explore other options.

yunhanw-google (approved):

jmartinez-silabs (approved):

PR #33915, opened 2024-06-13
[Python] Drop unnecessary null termination

agners:
The ctypes data type `c_char_p` takes care of null-terminating the byte array provided to it. The additional null termination doesn't hurt in practice, but it's unnecessary.

### DIFF
--- a/src/controller/python/chip/ChipDeviceCtrl.py
+++ b/src/controller/python/chip/ChipDeviceCtrl.py
@@ -732,7 +732,7 @@ class ChipDeviceControllerBase():
 
         self._ChipStack.Call(
             lambda: self._dmLib.pychip_DeviceController_DiscoverCommissionableNodes(
-                self.devCtrl, int(filterType), str(filter).encode("utf-8") + b"\x00")).raise_on_error()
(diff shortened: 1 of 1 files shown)


agners:
I've verified this to be the case by checking what is present at memory address `strlen()`, +1/+2. Also this [StackOverflow comment](https://stackoverflow.com/questions/23852311/different-behaviour-of-ctypes-c-char-p) mentions that the implicitly called [z_set](https://hg.python.org/cpython/file/c0e311e010fc/Modules/_ctypes/cfield.c#l1326) function calls [PyBytes_AsString](https://docs.python.org/3/c-api/bytes.html#c.PyBytes_AsString) which takes care of null termination, from the `PyBytes_AsString` docs:

> Return a pointer to the contents of o. The pointer refers to the internal buffer of o, which consists of len(o) + 1 bytes. The last byte in the buffer is always null, regardless of whether there are any other null bytes.

arkq (approved):

Damian-Nordic (approved):

tcarmelveilleux (approved):

PR #33921, opened 2024-06-14
reset StartUpCurrentLevel to ensure test repetitions

lecndav:
[TC-LVL-2.2] The added test step resets the StartUpCurrentLevel to another value then the one used in 6b.

Test step 6c checks if writing to the attribute StartUpCurrentLevel was successful by comparing its initial value with the value (5) written in step 6b. If the initial value and the value written in step 6b differ, the step is considered successful. However, if the test case is executed back-to-back, StartUpCurrentLevel remains unchanged between test runs, causing test step 6c to fail.

### DIFF
--- a/src/app/tests/suites/certification/Test_TC_LVL_2_2.yaml
+++ b/src/app/tests/suites/certification/Test_TC_LVL_2_2.yaml
@@ -39,6 +39,9 @@ config:
     StartUpCurrentLevelConfigValue:
         type: int8u
         defaultValue: 5
+    StartUpCurrentLevelConfigResetValue:
+        type: int8u
+        defaultValue: 100
 
 tests:
     - label:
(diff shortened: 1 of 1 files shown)


src/app/tests/suites/certification/Test_TC_LVL_2_2.yaml
```diff
@@ -230,6 +233,16 @@ tests:
               type: int8u
               notValue: StartUpCurrentLevelValue
 
+    #This is a reset step that is needed to reset the value of the attribute to another value then the one used in 6b in order to ensure successful test repetitions.
```
bzbarsky-apple, line 236: This is not great, since it introduces a divergence from the test plan in this certification test.

I know we have a step like that below already, but can we either get the test plan fixed to match the test or fix the way we run this test to match the test plan?

andy31415:
Closing stale PR.

PR #33926, opened 2024-06-14
[NXP] Update platform code for RW61x Zephyr

yeaissa:
Use common Zephyr wifi driver
OTA: wait until Matter app is built before building MCUBOOT Use subscriptions interval to compute OTA reboot delay to avoid not sending events before restart Increase NVS lookup cache size to reduce NVS operation time execution Align the default software version with FreeRTOS
Update ota software update doc

### DIFF
--- a/examples/platform/nxp/zephyr/factory_data/source/AppFactoryDataExample.cpp
+++ b/examples/platform/nxp/zephyr/factory_data/source/AppFactoryDataExample.cpp
@@ -1,0 +1,101 @@
+/*
+ *
+ *    Copyright (c) 2024 Project CHIP Authors
+ *    All rights reserved.
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+#include "AppFactoryData.h"
+
+#include <credentials/DeviceAttestationCredsProvider.h>
+#include <platform/CommissionableDataProvider.h>
+#include <platform/DeviceInstanceInfoProvider.h>
+
+#if CONFIG_CHIP_PLAT_LOAD_REAL_FACTORY_DATA
+#include "FactoryDataProvider.h"
+/*
+ * Test key used to encrypt factory data before storing it to the flash.
+ * The software key should be used only during development stage.
+ * For production usage, it is recommended to use the OTP key which needs to be fused in the RT1060 SW_GP2.
+ */
+static const uint8_t aes128TestKey[]
+    __attribute__((aligned)) = { 0x2b, 0x7e, 0x15, 0x16, 0x28, 0xae, 0xd2, 0xa6, 0xab, 0xf7, 0x15, 0x88, 0x09, 0xcf, 0x4f, 0x3c };
+#else
+#include <credentials/examples/DeviceAttestationCredsExample.h>
+#endif
+
+#if CONFIG_CHIP_FACTORY_DATA
+#include <platform/nxp/common/factory_data/FactoryDataProvider.h>
+#else
+#include <platform/nxp/zephyr/DeviceInstanceInfoProviderImpl.h>
+#endif
+
+#if CONFIG_CHIP_FACTORY_DATA && CONFIG_CHIP_ENCRYPTED_FACTORY_DATA
+#ifdef CONFIG_CHIP_ENCRYPTED_FACTORY_DATA_AES128_KEY
+
+#define KEY CONFIG_CHIP_ENCRYPTED_FACTORY_DATA_AES128_KEY
+#define HEXTONIBBLE(c) (*(c) >= 'A' ? (*(c) - 'A') + 10 : (*(c) - '0'))
+#define HEXTOBYTE(c) (HEXTONIBBLE(c) * 16 + HEXTONIBBLE(c + 1))
+#define AES128_KEY_ARRAY                                                                                                           \
+    HEXTOBYTE(KEY + 0), HEXTOBYTE(KEY + 2), HEXTOBYTE(KEY + 4), HEXTOBYTE(KEY + 6), HEXTOBYTE(KEY + 8), HEXTOBYTE(KEY + 10),       \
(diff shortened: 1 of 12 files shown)


examples/platform/nxp/zephyr/factory_data/source/AppFactoryDataExample.cpp
```diff
@@ -0,0 +1,101 @@
...
+    ReturnErrorOnFailure(FactoryDataPrvdImpl().Init());
+    SetDeviceInstanceInfoProvider(&FactoryDataPrvd());
+    SetDeviceAttestationCredentialsProvider(&FactoryDataPrvd());
+    SetCommissionableDataProvider(&FactoryDataPrvd());
+#else
+    SetDeviceInstanceInfoProvider(&DeviceInstanceInfoProviderMgrImpl());
+    SetDeviceAttestationCredentialsProvider(Examples::GetExampleDACProvider());
+#endif /* CONFIG_CHIP_FACTORY_DATA */
+    return CHIP_NO_ERROR;
+}
```
andy31415, line 101: please add newlines at the end of files

src/platform/nxp/common/OTAImageProcessorImpl.cpp
```diff
@@ -22,6 +22,8 @@
 
 #include "OTAImageProcessorImpl.h"
 
+#define MIN_REBOOT_DELAY 200
```
andy31415, line 25: Add units to names to this like `_MS` or `_SEC`. Can this be a C++-style constexpr instead of a macro?

src/platform/nxp/common/OTAImageProcessorImpl.h
```diff
@@ -77,6 +78,8 @@ class OTAImageProcessorImpl : public OTAImageProcessorInterface
 
     /* Buffer used for transaction storage */
     uint8_t mPostedOperationsStorage[NB_PENDING_TRANSACTIONS * TRANSACTION_SZ];
+
+    uint16_t delayBeforeReboot = 0;
```
andy31415, line 82: add units to variables. Also members use `m` in CHIP as a prefix. This should probably be `mDelayBeforeRebootMs` or something like that.

src/platform/nxp/zephyr/ota/OTAImageProcessorImpl.cpp
```diff
@@ -96,10 +97,16 @@ CHIP_ERROR OTAImageProcessorImpl::Apply()
...
+        PlatformMgr().HandleServerShuttingDown();
+        /*
+         * Restart the device in order to apply the update image.
+         * This should be done with a delay so the device has enough time to send
+         * the state-transition event when applying the update.
+         */
+        ChipLogProgress(SoftwareUpdate, "Restarting device, will reboot in %d seconds ...", delayBeforeReboot);
         return SystemLayer().StartTimer(
-            System::Clock::Milliseconds32(CHIP_DEVICE_CONFIG_OTA_REQUESTOR_REBOOT_DELAY_MS),
+            System::Clock::Milliseconds32(delayBeforeReboot * 1000 + MIN_REBOOT_DELAY),
```
andy31415, line 108: what is the unit of measurement for delayBeforeReboot? At first I thought MS, but the clock is already Milliseconds32 ... is this thousentofams ? that seems odd.

src/platform/nxp/zephyr/ota/OTAImageProcessorImpl.h
```diff
@@ -40,11 +40,13 @@ class OTAImageProcessorImpl : public OTAImageProcessorInterface
     CHIP_ERROR ProcessBlock(ByteSpan & aBlock) override;
     bool IsFirstImageRun() override;
     CHIP_ERROR ConfirmCurrentImage() override;
+    void SetRebootDelay(uint16_t rebootDelay);
 
 private:
     CHIP_ERROR PrepareDownloadImpl();
     CHIP_ERROR ProcessHeader(ByteSpan & aBlock);
 
+    uint16_t delayBeforeReboot  = 0;
```
andy31415, line 49: units of measurement please

src/platform/nxp/zephyr/ota/OTAImageProcessorImpl.h
```diff
@@ -40,11 +40,13 @@ class OTAImageProcessorImpl : public OTAImageProcessorInterface
     CHIP_ERROR ProcessBlock(ByteSpan & aBlock) override;
     bool IsFirstImageRun() override;
     CHIP_ERROR ConfirmCurrentImage() override;
+    void SetRebootDelay(uint16_t rebootDelay);
```
andy31415, line 43: Methods should also have unit of measurement in them like `SetRebootDelayMs` or use a typesafe unit type like Millisecond32 or similar.

andy31415 (changes requested):
Time variables and methods should specify units or should use unit-safe types.

src/platform/nxp/zephyr/ota/OTAImageProcessorImpl.cpp
andy31415: Usage of MIN_REBOOT_DELAY is also suspect - it seems to be a delta, otherwise I would expect some std::min call not a plus.

yeaissa: @andy31415, The unit for `delayBeforeReboot` is seconds. I have added the unit to the variable name.
MIN_REBOOT_DELAY is used here as a delta, I have updated the name to `deltaRebootDelayMs`.
Sorry for the confusion.

yeaissa:
> Time variables and methods should specify units or should use unit-safe types.

Updated. Thanks for the review @andy31415

src/platform/nxp/common/OTAImageProcessorImpl.cpp
```diff
@@ -279,8 +281,10 @@ void OTAImageProcessorImpl::HandleApply(intptr_t context)
      * This should be done with a delay so the device has enough time to send
      * the state-transition event when applying the update.
      */
-    ChipLogProgress(SoftwareUpdate, "Restarting device in 5 seconds ...");
-    DeviceLayer::SystemLayer().StartTimer(System::Clock::Milliseconds32(5 * 1000), HandleRestart, nullptr);
+    ChipLogProgress(SoftwareUpdate, "Restarting device, will reboot in %d seconds ...", imageProcessor->mDelayBeforeRebootSec);
```
andy31415, line 284: Technically we add deltaRebootDelayMs to the delay ... do we just rely it is very small? If so small that insignificant, why do we have it?

andy31415 (approved):

src/platform/nxp/common/OTAImageProcessorImpl.cpp
yeaissa: When I did this modification, I thought having a small delta is better to make sure the report is sent, I tested with it and it worked fine. I prefer keeping it.

yeaissa:
IoT SDK CI failure is not related to this PR.

PR #33932, opened 2024-06-14
Moving to NSMapTable to avoid dangling MTRDevices

woody-apple:
We have a situation where MTRDevices that clients no longer care about are being held on by us. I'd like to move us to a more weak ownership here, as well as expose some explicit removal.

Testing this approach.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDeviceController.mm
+++ b/src/darwin/Framework/CHIP/MTRDeviceController.mm
@@ -121,7 +121,7 @@ using namespace chip::Tracing::DarwinFramework;
     MTRP256KeypairBridge _operationalKeypairBridge;
     MTRDeviceAttestationDelegateBridge * _deviceAttestationDelegateBridge;
     MTRDeviceControllerFactory * _factory;
-    NSMutableDictionary * _nodeIDToDeviceMap;
+    NSMapTable * _nodeIDToDeviceMap;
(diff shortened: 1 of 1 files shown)


src/darwin/Framework/CHIP/MTRDeviceController.mm
```diff
@@ -121,7 +121,7 @@ @implementation MTRDeviceController {
     MTRP256KeypairBridge _operationalKeypairBridge;
     MTRDeviceAttestationDelegateBridge * _deviceAttestationDelegateBridge;
     MTRDeviceControllerFactory * _factory;
-    NSMutableDictionary * _nodeIDToDeviceMap;
+    NSMapTable * _nodeIDToDeviceMap;
```
bzbarsky-apple, line 124: Won't this potentially break some API consumers that get a device, set a delegate on it, then don't worry about the device anymore?   I guess at this point our API surface perhaps doesn't lead one to think this should work (though it does)?

woody-apple: I'm worried about that too, however we need to not own these thing, as they need to clean up automatically. Usually this is done with ref counting, or an NSMapTable.

woody-apple: Been talking with the home team about this, as they need to hold on to a reference

PR #33933, opened 2024-06-14
Fix python Wi-Fi/Thread setup with manual code

### LINKED ISSUE #26907: [python] ChipDeviceController BLE API do not support discovery by short discriminator
When only a manual pairing code is available, only the short discriminator (upper 4 bits of 12-bit discriminator) are available for discovery.

None of the Python API for commissioning over BLE support a version where the short discriminator is given.

Suggest:

- Adapting the ConnectBLE to have some way of expression short discriminator (e.g. by setting high bit of discriminator)
- Adapting CommissionThread/CommissionWifi to have `shortDiscriminator` keyword argument defaulting to None that causes disregarding the long `discriminator` argument, and uses short discriminator instead.

tcarmelveilleux:
- Plumbing was missing to pass down the short discriminator
- Passing `--manual-code 1234-567-8901` which only has short discriminator, would always fail to find device over BLE

Fixes #26907

This PR:

- Adds plumbing to detect short discriminator in Python controller
- Improves code-based setup in CHIPDeviceController to honor the SetupDiscriminator value, including whether short/long.

Testing done:
- Ran `python3 src/python_testing/TC_SC_3_6.py --commissioning-method ble-wifi --wifi-ssid MySsid --wifi-passphrase Secret123 --manual-code 2168-374-4904 --storage-path kvs1`
  - Before fix, discriminator always mismatched.
  - After fix, commissioning succeeds.
- Unit tests and other integration tests still pass

### DIFF
--- a/src/lib/support/SetupDiscriminator.h
+++ b/src/lib/support/SetupDiscriminator.h
@@ -32,7 +32,7 @@ namespace chip {
 class SetupDiscriminator
 {
 public:
-    constexpr SetupDiscriminator() : mDiscriminator(0xFFFFu), mIsShortDiscriminator(false) {}
+    constexpr SetupDiscriminator() : mDiscriminator(0), mIsShortDiscriminator(false) {}
 
     // See section 5.1.2. QR Code in the Matter specification
     static constexpr int kLongBits = 12;
--- a/src/lib/support/SetupDiscriminator.h
+++ b/src/lib/support/SetupDiscriminator.h
@@ -32,7 +32,7 @@ namespace chip {
 class SetupDiscriminator
 {
 public:
-    constexpr SetupDiscriminator() : mDiscriminator(0), mIsShortDiscriminator(0) {}
+    constexpr SetupDiscriminator() : mDiscriminator(0xFFFFu), mIsShortDiscriminator(false) {}
 
     // See section 5.1.2. QR Code in the Matter specification
     static constexpr int kLongBits = 12;
@@ -104,8 +104,8 @@ private:
     // discriminator).
     static_assert(kLongBits == 12, "Unexpected field length");
     static_assert(kShortBits <= kLongBits, "Unexpected field length");
-    uint16_t mDiscriminator : 12;
-    uint16_t mIsShortDiscriminator : 1;
+    uint16_t mDiscriminator;
+    bool mIsShortDiscriminator;
 };
 
 } // namespace chip
--- a/src/lib/support/SetupDiscriminator.h
+++ b/src/lib/support/SetupDiscriminator.h
@@ -23,9 +23,9 @@
 
 #pragma once
 
-#include <lib/support/CodeUtils.h>
-
 #include <cstdint>
+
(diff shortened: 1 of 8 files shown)


src/protocols/secure_channel/RendezvousParameters.h
```diff
@@ -59,11 +60,55 @@ class RendezvousParameters
...
+        return mSetupDiscriminator.GetLongValue();
+    }
+
+    SetupDiscriminator GetSetupDiscriminator() const
+    {
+        if (!mHasDiscriminator)
+        {
+            ChipLogError(Discovery,
+                         "Get RendezvousParameters::GetSetupDiscriminator() called without discriminator in params (inconsistent). "
+                         "Returning last!");
```
bzbarsky-apple, line 96: I'm not really sure what this "Returning last" bit means...

src/protocols/secure_channel/RendezvousParameters.h
```diff
@@ -128,9 +173,10 @@ class RendezvousParameters
     }
 
 private:
-    Transport::PeerAddress mPeerAddress;  ///< the peer node address
-    uint32_t mSetupPINCode  = 0;          ///< the target peripheral setup PIN Code
-    uint16_t mDiscriminator = UINT16_MAX; ///< the target peripheral discriminator
+    Transport::PeerAddress mPeerAddress; ///< the peer node address
+    uint32_t mSetupPINCode = 0;          ///< the target peripheral setup PIN Code
+    bool mHasDiscriminator = false;
+    SetupDiscriminator mSetupDiscriminator;
```
bzbarsky-apple, line 179: Any value in using `std::optional<SetupDiscriminator>' here?

src/lib/support/SetupDiscriminator.h
bzbarsky-apple, line 1: Why the changes to this file?  They're sane if we don't care about packing SetupDiscriminator into 16 bits (which maybe we don't)....

If we do make these changes, do we still need the `kLongBits == 12` and whatnot static assert?  It was there to make it safe to store the value in a 12-bit bitfield.

bzbarsky-apple (approved):

src/lib/support/SetupDiscriminator.h
tcarmelveilleux: I wanted to avoid yak-shaving the public API backwards compatibility more than I need too

src/protocols/secure_channel/RendezvousParameters.h
tcarmelveilleux: Fixed

src/protocols/secure_channel/RendezvousParameters.h
tcarmelveilleux: Done

src/lib/support/SetupDiscriminator.h
tcarmelveilleux: Also, `kLongBits` is used other places.

yufengwangca (approved):

PR #33934, opened 2024-06-14
[Metrics] Added additional metrics for CASE and Subscription

anush-apple:
- CASE session tracking
- Subscription Setup
- MRP Retries

### DIFF
--- a/src/app/ReadClient.cpp
+++ b/src/app/ReadClient.cpp
@@ -33,6 +33,7 @@
 #include <messaging/ReliableMessageMgr.h>
 #include <messaging/ReliableMessageProtocolConfig.h>
 #include <platform/LockTracker.h>
+#include <tracing/metric_event.h>
 
 #include <app-common/zap-generated/cluster-objects.h>
 #include <app-common/zap-generated/ids/Attributes.h>
@@ -186,6 +187,11 @@ void ReadClient::Close(CHIP_ERROR aError, bool allowResubscription)
     }
     else
     {
+        if (IsAwaitingInitialReport() || IsAwaitingSubscribeResponse())
+        {
+            MATTER_LOG_METRIC_END(Tracing::kMetricDeviceSubscriptionSetup, aError);
+        }
+
         ClearActiveSubscriptionState();
         if (aError != CHIP_NO_ERROR)
         {
@@ -491,6 +497,7 @@ CHIP_ERROR ReadClient::OnMessageReceived(Messaging::ExchangeContext * apExchange
         ChipLogProgress(DataManagement, "SubscribeResponse is received");
         VerifyOrExit(apExchangeContext == mExchange.Get(), err = CHIP_ERROR_INCORRECT_STATE);
         err = ProcessSubscribeResponse(std::move(aPayload));
+        MATTER_LOG_METRIC_END(Tracing::kMetricDeviceSubscriptionSetup, err);
     }
     else if (aPayloadHeader.HasMessageType(Protocols::InteractionModel::MsgType::StatusResponse))
     {
@@ -684,6 +691,7 @@ void ReadClient::OnResponseTimeout(Messaging::ExchangeContext * apExchangeContex
 {
     ChipLogError(DataManagement, "Time out! failed to receive report data from Exchange: " ChipLogFormatExchange,
                  ChipLogValueExchange(apExchangeContext));
+
     Close(CHIP_ERROR_TIMEOUT);
 }
 
@@ -1096,11 +1104,18 @@ CHIP_ERROR ReadClient::SendSubscribeRequest(const ReadPrepareParams & aReadPrepa
     VerifyOrReturnError(aReadPrepareParams.mMinIntervalFloorSeconds <= aReadPrepareParams.mMaxIntervalCeilingSeconds,
                         CHIP_ERROR_INVALID_ARGUMENT);
 
-    return SendSubscribeRequestImpl(aReadPrepareParams);
+    auto err =  SendSubscribeRequestImpl(aReadPrepareParams);
+    if (CHIP_NO_ERROR != err)
+    {
+        MATTER_LOG_METRIC_END(Tracing::kMetricDeviceSubscriptionSetup, err);
+    }
+    return err;
 }
 
 CHIP_ERROR ReadClient::SendSubscribeRequestImpl(const ReadPrepareParams & aReadPrepareParams)
 {
+    MATTER_LOG_METRIC_BEGIN(Tracing::kMetricDeviceSubscriptionSetup);
+
     VerifyOrReturnError(ClientState::Idle == mState, CHIP_ERROR_INCORRECT_STATE);
 
     if (&aReadPrepareParams != &mReadPrepareParams)
--- a/src/app/ReadClient.cpp
+++ b/src/app/ReadClient.cpp
@@ -1104,7 +1104,7 @@ CHIP_ERROR ReadClient::SendSubscribeRequest(const ReadPrepareParams & aReadPrepa
     VerifyOrReturnError(aReadPrepareParams.mMinIntervalFloorSeconds <= aReadPrepareParams.mMaxIntervalCeilingSeconds,
                         CHIP_ERROR_INVALID_ARGUMENT);
 
-    auto err =  SendSubscribeRequestImpl(aReadPrepareParams);
+    auto err = SendSubscribeRequestImpl(aReadPrepareParams);
     if (CHIP_NO_ERROR != err)
     {
         MATTER_LOG_METRIC_END(Tracing::kMetricDeviceSubscriptionSetup, err);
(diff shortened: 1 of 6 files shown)


src/app/ReadClient.cpp
```diff
@@ -186,6 +187,11 @@ void ReadClient::Close(CHIP_ERROR aError, bool allowResubscription)
     }
     else
     {
+        if (IsAwaitingInitialReport() || IsAwaitingSubscribeResponse())
+        {
+            MATTER_LOG_METRIC_END(Tracing::kMetricDeviceSubscriptionSetup, aError);
```
tcarmelveilleux, line 192: Same event recorded in two places under different conditions may dilute the understanding of the underlying reason.

src/protocols/secure_channel/CASESession.cpp
```diff
@@ -1294,10 +1327,17 @@ CHIP_ERROR CASESession::HandleSigma2Resume(System::PacketBufferHandle && msg)
 CHIP_ERROR CASESession::HandleSigma2_and_SendSigma3(System::PacketBufferHandle && msg)
 {
     MATTER_TRACE_SCOPE("HandleSigma2_and_SendSigma3", "CASESession");
-    ReturnErrorOnFailure(HandleSigma2(std::move(msg)));
-    ReturnErrorOnFailure(SendSigma3a());
+    auto err = HandleSigma2(std::move(msg));
```
tcarmelveilleux, line 1330: `CHIP_ERROR err = .....`

src/protocols/secure_channel/CASESession.cpp
```diff
@@ -1294,10 +1327,17 @@ CHIP_ERROR CASESession::HandleSigma2Resume(System::PacketBufferHandle && msg)
...
+    MATTER_LOG_METRIC_END(kMetricDeviceCASESessionSigma1, err);
```
tcarmelveilleux, line 1331: Curuous why this is `Sigma1` related event rather than Sigma2

tcarmelveilleux (approved):

src/app/ReadClient.cpp
anush-apple: I was trying to capture all cases when subscription would fail. Close seems liked the funnel point. Would we not want to capture if closed due to initial report or waiting for a response?

src/protocols/secure_channel/CASESession.cpp
anush-apple: Will fix

src/protocols/secure_channel/CASESession.cpp
anush-apple: This is marking the end of Sigma1 from the initiator side once it receives the Sigma2 from the responder.

src/app/ReadClient.cpp
```diff
@@ -491,6 +497,7 @@ CHIP_ERROR ReadClient::OnMessageReceived(Messaging::ExchangeContext * apExchange
         ChipLogProgress(DataManagement, "SubscribeResponse is received");
         VerifyOrExit(apExchangeContext == mExchange.Get(), err = CHIP_ERROR_INCORRECT_STATE);
         err = ProcessSubscribeResponse(std::move(aPayload));
+        MATTER_LOG_METRIC_END(Tracing::kMetricDeviceSubscriptionSetup, err);
```
bzbarsky-apple, line 500: If something inside ProcessSubscribeResponse() before it calls `MoveToState` fails, then we will do the MATTER_LOG_METRIC_END here _and_ in the resulting Close() call, right?

I wonder whether it would be better to put ` MATTER_LOG_METRIC_END(Tracing::kMetricDeviceSubscriptionSetup, CHIP_NO_ERROR);` in either MoveToState when moving to the `SubscriptionActive` state or at the one place where we move to that state.  That way we unambiguously do the "ended with error" in Close() and "ended with success" on the codepath that does that.

src/app/ReadClient.cpp
```diff
@@ -1096,11 +1104,18 @@ CHIP_ERROR ReadClient::SendSubscribeRequest(const ReadPrepareParams & aReadPrepa
...
+    if (CHIP_NO_ERROR != err)
+    {
+        MATTER_LOG_METRIC_END(Tracing::kMetricDeviceSubscriptionSetup, err);
+    }
+    return err;
 }
 
 CHIP_ERROR ReadClient::SendSubscribeRequestImpl(const ReadPrepareParams & aReadPrepareParams)
 {
+    MATTER_LOG_METRIC_BEGIN(Tracing::kMetricDeviceSubscriptionSetup);
```
bzbarsky-apple, line 1117: I assume we don't care that the unit-test-only `SendSubscribeRequestWithoutValidation` calls `SendSubscribeRequestImpl` and then does not do the "end on failure" behavior that `SendSubscribeRequest` was modified to do?  If that's the case, might be worth documenting.  Or we could try to just fix both of the callsites to have identical behavior.

It would be nice, perhaps, if we had a macro for "do start here, do end if err is failure when we are returning" of some sort....  But that's out of scope for this PR.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -115,7 +115,7 @@ - (void)_deviceInternalStateChanged:(MTRDevice *)device;
 using namespace chip;
 using namespace chip::app;
 using namespace chip::Protocols::InteractionModel;
-
+using namespace chip::Tracing::DarwinFramework;
 typedef void (^FirstReportHandler)(void);
```
bzbarsky-apple, line 119:
```suggestion

typedef void (^FirstReportHandler)(void);
```

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -1007,6 +1007,8 @@ - (void)_handleSubscriptionEstablished
 {
     os_unfair_lock_lock(&self->_lock);
 
+    MATTER_LOG_METRIC_END(kMetricMTRDeviceSubscriptionSetup, CHIP_NO_ERROR);
```
bzbarsky-apple, line 1010: So this will get called on internal re-subscribes from the ReadClient, so we can get a MATTER_LOG_METRIC_END here without a corresponding START.  Is that OK?  If not, consider conditioning this on whether `HadSubscriptionEstablishedOnce()` is true (see code a bit below this).

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -1055,6 +1057,8 @@ - (void)_handleSubscriptionError:(NSError *)error
 {
     std::lock_guard lock(_lock);
 
+    MATTER_LOG_METRIC_END(kMetricMTRDeviceSubscriptionSetup, [MTRError errorToCHIPErrorCode:error]);
```
bzbarsky-apple, line 1060: So this function is called in three cases:

1. Getting a CASE session for the initial subscription fails.
2. Kicking off the subscription on the CASE session we got fails.
3. We failed to kick off a resubscribe for some reason.

Item 3 does not seem relevant to us, at first glance.  Maybe this should be inside `_handleSubscriptionReset`, though that can also get called if everything here is being torn down.....

On the other hand, how do we want to handle the case of "got a CASE session, but first attempt to set up a subscription failed, so we will try again"?  Right now we will keep waiting for it and eventually when it's established log a "success" end metric.  Is that what we want?

Also: we seem to be treating subscription reattempts that come from MTRDevice (i.e. "before we get a CASE session") and the ones that come from inside ReadClient (i.e. "after we get a CASE session") differently: the former log begin/end on every attempt, while the latter logs begin on first attempt start and then end on success, which might be a while later).  This asymmetry is a bit strange to me.

woody-apple (approved):
Approving pending comments

src/app/ReadClient.cpp
anush-apple: The backend needs to handle multiple end or begin calls. The usual way is to capture the first of these events.

src/darwin/Framework/CHIP/MTRMetricKeys.h
```diff
@@ -87,6 +88,9 @@ constexpr Tracing::MetricKey kMetricBLEDevicesRemoved = "dwnfw_ble_devices_remov
 // Unexpected C quality attribute update outside of priming
 constexpr Tracing::MetricKey kMetricUnexpectedCQualityUpdate = "dwnpm_bad_c_attr_update";
 
+// Setup from darwin MTRDevice for initial subscrption to a device
```
bzbarsky-apple, line 91:
```suggestion
// Setup from darwin MTRDevice for initial subscription to a device
```

src/app/ReadClient.cpp
anush-apple: The handler of the _BEGIN and _END  usually needs to account for multiple invocations. Usually, the first instance of the event is captured and the remaining ones are ignored. Given that, this should be fine since we are capturing the failure only if the subscription setup fails.

PR #33935, opened 2024-06-14
Update DA-1.4/1.8 helper script

cecille:
I added the PID to the test_vector, so we should pull it from there. There's a new fallback that uses a different PID.

### DIFF
--- a/credentials/development/gen_commissioner_dut_test_plan_table.py
+++ b/credentials/development/gen_commissioner_dut_test_plan_table.py
@@ -86,8 +86,7 @@ def main():
         with open(path, 'r') as f:
             j = json.loads(f.read())
             success_expected = j['is_success_case'].lower() == 'true'
-            pid = 177 if 'fallback_encoding' in p else 32768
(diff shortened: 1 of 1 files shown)


credentials/development/gen_commissioner_dut_test_plan_table.py
```diff
@@ -86,8 +86,7 @@ def main():
         with open(path, 'r') as f:
             j = json.loads(f.read())
             success_expected = j['is_success_case'].lower() == 'true'
-            pid = 177 if 'fallback_encoding' in p else 32768
-            desc = TestInfo(desc=j['description'], dir=p, pid=pid)
+            desc = TestInfo(desc=j['description'], dir=p, pid=int(j['basic_info_pid']))
```
tcarmelveilleux, line 89: What if it's not there? Suggest having a pre-check for presence in `j` with a log in case it's missing

tcarmelveilleux (approved):

jmeg-sfy (approved):

PR #33937, opened 2024-06-14
Add option to ChipTool to allow session setup for large payloads.

### LINKED ISSUE #29697: Update ChipTool with an option to set up a large payload session with peer
### Feature description

ChipTool needs to be updated with an option so that it can appropriately set up a session that supports large payload transfers when interacting with devices that support it. When set to true, this option would make ChipTool try to set up CASE over TCP to allow the application to then pack more bytes in each message over that session.

### Platform

core

### Platform Version(s)

_No response_

### Anything else?

_No response_

pidarped:
By default, assumed false.
When set to 1, it will trigger an attempt to set up a session over TCP to support large payload transfers.
E.g., ./chip-tool onoff read on-off 15 2 --allow-large-payload 1

Fixes #29697

### DIFF
--- a/examples/chip-tool/commands/clusters/ModelCommand.cpp
+++ b/examples/chip-tool/commands/clusters/ModelCommand.cpp
@@ -44,8 +44,11 @@ CHIP_ERROR ModelCommand::RunCommand()
         return SendCommand(commissioneeDeviceProxy, mEndPointId);
     }
 
+    // Check whether the session needs to allow large payload support.
+    TransportPayloadCapability transportPayloadCapability =
+        AllowLargePayload() ? TransportPayloadCapability::kLargePayload : TransportPayloadCapability::kMRPPayload;
     return CurrentCommissioner().GetConnectedDevice(mDestinationId, &mOnDeviceConnectedCallback,
-                                                    &mOnDeviceConnectionFailureCallback);
+                                                    &mOnDeviceConnectionFailureCallback, transportPayloadCapability);
 }
 
 void ModelCommand::OnDeviceConnectedFn(void * context, chip::Messaging::ExchangeManager & exchangeMgr,
@@ -133,4 +136,9 @@ bool ModelCommand::IsPeerLIT()
 {
     CheckPeerICDType();
     return mIsPeerLIT.ValueOr(false);
+}
+
+bool ModelCommand::AllowLargePayload()
+{
+    return mAllowLargePayload.ValueOr(false);
 }
--- a/examples/chip-tool/commands/clusters/ModelCommand.h
+++ b/examples/chip-tool/commands/clusters/ModelCommand.h
@@ -57,6 +57,9 @@ public:
             "Whether to treat the peer as a LIT ICD. false: Always no, true: Always yes, (not set): Yes if the peer is registered "
             "to this controller.");
         AddArgument("timeout", 0, UINT16_MAX, &mTimeout);
+        AddArgument("allow-large-payload", 0, 1, &mAllowLargePayload,
+                    "If true, indicates that the session should allow large application payloads (which requires a TCP connection)."
+                    "Defaults to false, which uses a UDP+MRP session.");
     }
 
     /////////// CHIPCommand Interface /////////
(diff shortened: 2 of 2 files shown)


examples/chip-tool/commands/clusters/ModelCommand.cpp
```diff
@@ -134,3 +137,15 @@ bool ModelCommand::IsPeerLIT()
     CheckPeerICDType();
     return mIsPeerLIT.ValueOr(false);
 }
+
+bool ModelCommand::AllowLargePayload()
+{
+    if (mAllowLargePayload.HasValue())
+    {
+        return (mAllowLargePayload.Value());
+    }
```
yufengwangca, line 146: nit: return mAllowLargePayload.HasValue() ? mAllowLargePayload.Value() : false;

examples/chip-tool/commands/clusters/ModelCommand.h
```diff
@@ -82,9 +86,12 @@ class ModelCommand : public CHIPCommand
     chip::NodeId mDestinationId;
     std::vector<chip::EndpointId> mEndPointId;
     chip::Optional<bool> mIsPeerLIT;
+    chip::Optional<bool> mAllowLargePayload;
```
yufengwangca, line 89: since AllowLargePayload always has a value, and return false when !mAllowLargePayload.HasValue(), why do we need to make this Optional?

examples/chip-tool/commands/clusters/ModelCommand.cpp
```diff
@@ -44,8 +44,11 @@ CHIP_ERROR ModelCommand::RunCommand()
         return SendCommand(commissioneeDeviceProxy, mEndPointId);
     }
 
+    // Check if the session needs to allow large payload support.
```
bzbarsky-apple, line 47:
```suggestion
    // Check whether the session needs to allow large payload support.
```

examples/chip-tool/commands/clusters/ModelCommand.cpp
```diff
@@ -134,3 +137,15 @@ bool ModelCommand::IsPeerLIT()
...
+bool ModelCommand::AllowLargePayload()
+{
+    if (mAllowLargePayload.HasValue())
+    {
+        return (mAllowLargePayload.Value());
+    }
+    else
+    {
+        return false;
+    }
```
bzbarsky-apple, line 150:
```suggestion
    return mAllowLargePayload.ValueOr(false);
```

examples/chip-tool/commands/clusters/ModelCommand.cpp
bzbarsky-apple: No, please use `ValueOr()`.

examples/chip-tool/commands/clusters/ModelCommand.h
```diff
@@ -57,6 +57,10 @@ class ModelCommand : public CHIPCommand
             "Whether to treat the peer as a LIT ICD. false: Always no, true: Always yes, (not set): Yes if the peer is registered "
             "to this controller.");
         AddArgument("timeout", 0, UINT16_MAX, &mTimeout);
+        AddArgument(
+            "allow-large-payload", 0, 1, &mAllowLargePayload,
+            "Indicates if the session should allow large application payloads. When true, the controller tries to establish a "
+            "secure session over TCP. Otherwise session is established over MRP. Assumed to be false when not specified.");
```
bzbarsky-apple, line 63: I
```suggestion
            "If true, indicates that the session should allow large application payloads (which requires a TCP connection).  Defaults to false, which uses a UDP+MRP session.");
```

examples/chip-tool/commands/clusters/ModelCommand.h
bzbarsky-apple: It's Optional so that it doesn't have to be provided on the command line.

bzbarsky-apple (approved):
In the commit message, please use `--allow-large-payload true`, not `--allow-large-payload 1`.

examples/chip-tool/commands/clusters/ModelCommand.h
pidarped: That's right. the command option is not mandatory.

examples/chip-tool/commands/clusters/ModelCommand.cpp
yufengwangca: return mAllowLargePayload.ValueOr(false);

examples/chip-tool/commands/clusters/ModelCommand.h
yufengwangca: I see, 'allow-large-payload'  is optional, that is why you use chip::Optional<bool> instead of bool

yufengwangca (approved):

PR #33943, opened 2024-06-17
Removing NL Unit Infrastructure

Alami-Amine:
- since nl_unit_tests are now deprecated and all tests converted to pw_unit_test, `nl_unit` dependencies will be removed from the stack.

- pw_unit conversion was first discussed in: https://github.com/project-chip/connectedhomeip/pull/29342

### In This PR

1. removing `nl_unit` infrastructure from:
   1. ESP32
   2. Nordic
   3. OpenIoTSDK
   4. mbed
   5. EFR32 : partially done, since in #33714  other aspects related to `nl_unit_tests` are being removed.

2. all `#include` related to nl_unit_tests
3. cleaning `nl_unit` references in GN build files 
4. deleting `nl_unit` related files and scripts such as `scripts/gen_test_driver.py`


src/test_driver/mbed/unit_tests/main/main.cpp
```diff
@@ -63,8 +62,12 @@ int main()
 
     ChipLogProgress(NotSpecified, "Mbed unit-tests application run");
 
-    int status = RunRegisteredUnitTests();
-    ChipLogProgress(NotSpecified, "CHIP test status: %d", status);
+    int status = 0;
```
andy31415, line 65: status should probably not be 0, and instead we could log an error and then return non-zero.

andy31415 (approved):

src/test_driver/mbed/unit_tests/main/main.cpp
```diff
@@ -63,8 +62,14 @@ int main()
 
     ChipLogProgress(NotSpecified, "Mbed unit-tests application run");
 
-    int status = RunRegisteredUnitTests();
-    ChipLogProgress(NotSpecified, "CHIP test status: %d", status);
+    // Migration to pw_unit_tests was not done for mbed since the unit-testing FW used to run out of flash and was disabled from
```
tehampson, line 65: Is there an github issues to track to re-enable this after we split tests to multiple smaller binaries?

cecille: Can you ping me the number on slack too? Then I can add it to the project tracker.

andy31415: The issue is that MBED has no maintainer anymore for a long time now, so we have no platform support for issues. 

We could work on it, however without a maintainer for a platform, us taking over running and configuring tests on a platform seeems rough. Unsure if it should be on an official project tracker as a result.

andy31415: Issue link should probably also be added to https://github.com/project-chip/connectedhomeip/blob/master/.github/workflows/examples-mbed.yaml which is where this started (we disabled mbed tests because of flash)

arkq (approved):

src/test_driver/mbed/unit_tests/main/main.cpp
Alami-Amine: Issue created and added to the comments (of the yaml as well)

PR #33944, opened 2024-06-17
Update cirque pip flags to allow breaking system packages

andy31415:
Ubuntu 24.04 update requires this and it seems cirque always pulls the latest build (rather than being fixed).
We should probably force cirque to use specific versions instead, however for now this seems to be the quick fix.

### DIFF
--- a/scripts/tests/cirque_tests.sh
+++ b/scripts/tests/cirque_tests.sh
@@ -114,8 +114,6 @@ function cirquetest_bootstrap() {
     set -ex
 
     cd "$REPO_DIR"/third_party/cirque/repo
-    pip3 uninstall -y setuptools --break-system-packages
-    pip3 install --break-system-packages setuptools==65.7.0
     pip3 install --break-system-packages pycodestyle==2.5.0 wheel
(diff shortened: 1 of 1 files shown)


scripts/tests/cirque_tests.sh
```diff
@@ -114,9 +114,7 @@ function cirquetest_bootstrap() {
     set -ex
 
     cd "$REPO_DIR"/third_party/cirque/repo
-    pip3 uninstall -y setuptools
-    pip3 install setuptools==65.7.0
-    pip3 install pycodestyle==2.5.0 wheel
+    pip3 install --break-system-packages pycodestyle==2.5.0 wheel
```
tcarmelveilleux, line 117: Why do we not install a virtualenv instead? Why "break system packages" ?

andy31415: Because this is what cirque original script did. Updating this to be sane (virtualenv and likely to stop pulling 'latest' all the time) should be done as a maintenance pass rather than a hotfix.

PR #33945, opened 2024-06-17
Add XML and generated files of the Energy Calendar cluster

Michail-Antropov:
Please check `data_model/clusters/EnergyCalendar.xml` and `src/app/zap-templates/zcl/data-model/chip/energy-calendar-cluster.xml` only.
Other - generated files


src/app/common/templates/config-data.yaml
```diff
@@ -36,6 +36,7 @@ CommandHandlerInterfaceOnlyClusters:
     - Microwave Oven Control
     - Energy EVSE
     - Energy EVSE Mode
+    - Energy Calendar
```
bzbarsky-apple, line 39: This is a claim about implementation.  It does not belong in this PR.

src/app/zap-templates/zcl/data-model/chip/energy-calendar-cluster.xml
```diff
@@ -0,0 +1,126 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<configurator>
+  <domain name="Energy Management"/>
+
+  <enum name="PeakPeriodSeverityEnum" type="enum8">
```
bzbarsky-apple, line 17: For future reference: putting the XML in the same order as the spec helps review a _lot_.

src/app/zap-templates/zcl/data-model/chip/energy-calendar-cluster.xml
```diff
@@ -0,0 +1,126 @@
...
+    <field name="Sunday" mask="0x01"/>
+    <field name="Monday" mask="0x02"/>
+    <field name="Tuesday" mask="0x04"/>
+    <field name="Wednesday" mask="0x08"/>
+    <field name="Thursday" mask="0x10"/>
+    <field name="Friday" mask="0x20"/>
+    <field name="Saturday" mask="0x40"/>
+  </bitmap>
+
+  <struct name="Date">
```
bzbarsky-apple, line 48: This struct definition isn't in the spec for this cluster.

The spec seems to be using the "date" Data Model type, which is currently not really supported in the ZAP XML, I suspect.  chip-types.xml has it, but whether that actually leads to any sane behavior, who knows.

What needs to happen here is figuring out how "date" should actually be implemented.  That _can_ happen as a followup if we just need to unblock some work here, but it's a must-fix before this can ship.

src/app/zap-templates/zcl/data-model/chip/energy-calendar-cluster.xml
```diff
@@ -0,0 +1,126 @@
...
+    <cluster code="0x009A"/>
+    <item fieldId="0" name="Year" type="int16u" default="0" isNullable="true" optional="true"/>
+    <item fieldId="1" name="Month" type="int8u" min="1" max="12" default="1" isNullable="true" optional="true"/>
+    <item fieldId="2" name="Day" type="int8u" min="1" max="31" default="1" isNullable="true" optional="true"/>
+    <item fieldId="3" name="DayOfWeek" type="int8u" min="1" max="7" isNullable="true" optional="true"/>
+  </struct>
+
+  <struct name="CalendarPeriod">
```
bzbarsky-apple, line 56: This name does not match the spec.

src/app/zap-templates/zcl/data-model/chip/energy-calendar-cluster.xml
```diff
@@ -0,0 +1,126 @@
...
+    <item fieldId="2" name="Transitions" type="TransitionStruct" array="true" min="1"/>
+    <item fieldId="3" name="CalendarID" type="int32u" default="0" optional="true"/>
+  </struct>
+
+  <struct name="PeakPeriodStruct">
+    <cluster code="0x009A"/>
+    <item fieldId="0" name="Severity" type="PeakPeriodSeverityEnum"/>
+    <item fieldId="1" name="PeakPeriod" type="int16u"/>
+    <item fieldId="2" name="StartTime" type="epoch_s"/>
+    <item fieldId="3" name="EndTime" type="epoch_s"/>
```
bzbarsky-apple, line 75: These are both nullable in the spec.

I am stopping here; this XML clearly does not match the spec, and the different ordering makes it very hard to review.  Please fix the ordering and make the XML actually match the spec.

bzbarsky-apple (changes requested):

src/app/common/templates/config-data.yaml
Michail-Antropov: Fix

src/app/zap-templates/zcl/data-model/chip/energy-calendar-cluster.xml
Michail-Antropov: Fixed by using alchemy tool to generate XML.

src/app/zap-templates/zcl/data-model/chip/energy-calendar-cluster.xml
Michail-Antropov: Fixed by using alchemy tool to generate XML.

src/app/zap-templates/zcl/data-model/chip/energy-calendar-cluster.xml
Michail-Antropov: Fixed by using alchemy tool to generate XML.

Michail-Antropov: Perhaps the specification (*.adoc) is changing faster than I can recreate the source and its derivative files.
Can you give me access to the `connectedhomeip-spec` repository?
In addition, verification tasks are actively changing on github itself.
For example, from the last one: now in *.kt files in the description of functions there is a comma after the last parameter.
Was:
```
class GeneralDiagnosticsClusterNetworkFaultChangeEvent(
 val current: List<UInt>,
 val previous: List<UInt>
) {
```
Became:
```
class GeneralDiagnosticsClusterNetworkFaultChangeEvent(
 val current: List<UInt>,
 val previous: List<UInt>,
) {
```

At the initial stage, this was due to the fact that I created XML manually, not knowing about the alchemy tool, but since a week this is no longer relevant.

cecille: What are you using as a reference if you don't have access to the spec repo?

bzbarsky-apple: I'd like to echo @cecille's question.

I don't have a way to give you access to the spec repo, but if you need that you should talk to Chris LaPre at the CSA.

> In addition, verification tasks are actively changing on github itself.

Yes, various other changes to the codegen can be happening...  That's normal.

To make this converge faster, please order the XML in the same order as the spec.  Alchemy will try to preserve your existing XML as much as it can if you run it with some XML already in place, so it will not fix issues like the ordering one for you unless you run it from a blank slate....

src/app/zap-templates/zcl/data-model/chip/energy-calendar-cluster.xml
Michail-Antropov: @bzbarsky-apple I set "date" type and get uint32.
It produce some question:
- field can be null? is eqval: day, month =0, year = 0xff;
- byte order: 0xYYMMDDWW or 0xWWDDMYY?

It field use for a day in the Special Days attribute. Maybe change to epoch-s and get midnight value?

bzbarsky-apple: Again, the `date` type is not implemented in the SDK.  It needs to be encoded/decoded as a struct on the wire, per spec, not as uint32....  The bits in chip-types.xml for it are clearly wrong.

Michail-Antropov: Add DateStruct type

robszewczyk:
Not reflective of current SDK text.  suggest closing.

andy31415:
Closing stale PR.

PR #33949, opened 2024-06-17
make sure test event trigger delegate is not null before using it

fessehaeve:
If an app is not setting the test event trigger delegate, while CHIP_CONFIG_ENABLE_ICD_SERVER is enabled, it would result in accessing a nonexistent memory.

### DIFF
--- a/src/app/server/Server.cpp
+++ b/src/app/server/Server.cpp
@@ -359,7 +359,10 @@ CHIP_ERROR Server::Init(const ServerInitParams & initParams)
                      chip::app::InteractionModelEngine::GetInstance());
 
     // Register Test Event Trigger Handler
-    mTestEventTriggerDelegate->AddHandler(&mICDManager);
+    if (mTestEventTriggerDelegate != nullptr)
+    {
+        mTestEventTriggerDelegate->AddHandler(&mICDManager);
+    }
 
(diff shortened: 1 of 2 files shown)


tcarmelveilleux (approved):

src/app/server/Server.cpp
```diff
@@ -359,7 +359,10 @@ CHIP_ERROR Server::Init(const ServerInitParams & initParams)
                      chip::app::InteractionModelEngine::GetInstance());
 
     // Register Test Event Trigger Handler
-    mTestEventTriggerDelegate->AddHandler(&mICDManager);
+    if (mTestEventTriggerDelegate != nullptr)
```
tcarmelveilleux, line 362: Please add a log warning that's it's null before these lines, in the earlier parts of Server::Init.

mkardous-silabs:
I'm not sure i agree with this... Test Event Trigger is required by those apps.
I would prefer a VerifyOrDieWithMsg to have the benefit of forcing the trigger to be present and informing of the why when crashing.

fessehaeve:
@mkardous-silabs It could be said that the smoke-co-alarm app could indeed need a test trigger but I don't see a reason not to give the user an option not to not have it. Test triggers are needed only for dev and cert, and not intended to have in shipping products, obviously.

bauerschwan (approved):

Issue #33936, opened 2024-06-14
[CERT-TEST-FAILURE] Test_TC_IDM_5_2.yaml test cases 3 and 4 should expect TIMEOUT?


Apollon77:
### Feature Area

Other

### Test Case

Test_TC_IDM_5_2

### Reproduction steps

The test cases 3 and 4 are frommmy perspective testing a Invoke/Write interaction with an expired TimedInteraction before.

https://github.com/project-chip/connectedhomeip/blob/a12a0c070e2ad6e0347e617ace1e8098b4ccd8d3/src/app/tests/suites/certification/Test_TC_IDM_5_2.yaml#L90

When I check specs (1.1+) it tells e.g.

> Incoming Write Request Action
> If this action is part of a Timed Write transaction, and the Timeout has expired from the preced­ ing Timed Request action, then a Status Response action with the TIMEOUT Status Code SHALL be submitted to the message layer and this interaction SHALL terminate.

and for read:
> If this action is part of a Timed Invoke transaction, and the Timeout has expired from the pre­ ceding Timed Request action, then a Status Response action with the TIMEOUT Status Code SHALL be submitted to the message layer and this interaction SHALL terminate.

This test here expects UnsupportedAccess as error.

Do I miss anything or is the test defined wrong or the specticification?

### Bug prevalence

came back as error from a current 1.1 certification

### GitHub hash of the SDK that was being used

master

### Platform

core

cecille:
This appears to be quite deliberate in the code - see https://github.com/project-chip/connectedhomeip/blob/master/src/app/TimedHandler.cpp#L69.

@bzbarsky-apple - I know this was three years ago, but do you recall what part of the spec was being referenced here? Do we have a spec conflict?

cecille:
Found it - please see https://github.com/CHIP-Specifications/connectedhomeip-spec/commit/0bd0e4f937ebaf84d6ebe7dc22e79c5af605bc2a. Error was changed, it would appear that the SDK and tests were not updated.

bzbarsky-apple:
Yeah, code was written 3 years ago, spec stealth-changed 2 years ago....  :(

cjandhyala:
@Apollon77 closing the issue given SDK and test plan issues are closed. Feel free to reopen if you still think this issue needs to be addressed.

PR #33952, opened 2024-06-17
TC-IDM-5.2: Fix status code in test and SDK

cecille:
Please see:
https://github.com/project-chip/connectedhomeip/issues/33936

### DIFF
--- a/src/app/TimedHandler.cpp
+++ b/src/app/TimedHandler.cpp
@@ -66,10 +66,9 @@ CHIP_ERROR TimedHandler::OnMessageReceived(Messaging::ExchangeContext * aExchang
                       ChipLogValueX64(now.count()), this, ChipLogValueExchange(aExchangeContext));
         if (now > mTimeLimit)
         {
-            // Time is up.  Spec says to send UNSUPPORTED_ACCESS.
             ChipLogError(DataManagement, "Timeout expired: handler %p exchange " ChipLogFormatExchange, this,
                          ChipLogValueExchange(aExchangeContext));
-            return StatusResponse::Send(Status::UnsupportedAccess, aExchangeContext, /* aExpectResponse = */ false);
+            return StatusResponse::Send(Status::Timeout, aExchangeContext, /* aExpectResponse = */ false);
         }
 
         if (aPayloadHeader.HasMessageType(MsgType::InvokeCommandRequest))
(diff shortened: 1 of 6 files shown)


src/app/TimedHandler.cpp
```diff
@@ -66,10 +66,9 @@ CHIP_ERROR TimedHandler::OnMessageReceived(Messaging::ExchangeContext * aExchang
                       ChipLogValueX64(now.count()), this, ChipLogValueExchange(aExchangeContext));
         if (now > mTimeLimit)
         {
-            // Time is up.  Spec says to send UNSUPPORTED_ACCESS.
             ChipLogError(DataManagement, "Timeout expired: handler %p exchange " ChipLogFormatExchange, this,
                          ChipLogValueExchange(aExchangeContext));
-            return StatusResponse::Send(Status::UnsupportedAccess, aExchangeContext, /* aExpectResponse = */ false);
+            return StatusResponse::Send(Status::Timeout, aExchangeContext, /* aExpectResponse = */ false);
```
bzbarsky-apple, line 71: Looking at https://github.com/CHIP-Specifications/connectedhomeip-spec/commit/0bd0e4f937ebaf84d6ebe7dc22e79c5af605bc2a and the current spec text we also need to fix `WriteHandler::ProcessWriteRequest`.  The `mIsTimedRequest != aIsTimedWrite` bit needs to return TimedRequestMismatch, right?  It looks like this was changed for invoke but not write, probably because we don't have any actual timed writable attributes to test on....

cecille: If it's ok, I'd like to address this in a followup. We can do a non-cert test of a writable timed attribute using the unit testing cluster, but I don't believe I will have a change to add attributes and tests etc. before the TE2 deadline. I can make an issue if the write is not a blocker for this PR.

cecille: As it turns out, we have a timed write attribute in the unit testing cluster and a test, so may as well do this now.

cecille: Actually, it turns out we DO have a unit test for this, so I'm more comfortable just fixing this here. Opened an issue to add to cert as that's going to require a LOT more plumbing. Issue here: https://github.com/CHIP-Specifications/chip-test-plans/issues/4256

bzbarsky-apple (approved):
Thank you!

tcarmelveilleux (approved):

andy31415 (approved):

tehampson (approved):

PR #33953, opened 2024-06-17
[Fix] Proposed resolution of issue #230: remove PICS from timesync tests

j-ororke:
Possible fix for issue 230: remove PICS from timesync tests:
- Removed if statement checks with PICS from timesync tests 2_1 and 2_2
- Added self.wait_for_user_input() to allow user input for timesync test 2_1 if using TSC, NTPC, NTPS, or TZ features on DUT.


j-ororke:
Updated method for this fix after being provided some further information from Cecille on Tuesday:
- Removed using the self.wait_for_user_input() to reusing established code from TestCommisioningTimeSync.py to gather supported features from DUT to use in place of if statements in timesync 2_1 test module.

src/python_testing/TC_TIMESYNC_2_1.py
```diff
@@ -29,45 +29,42 @@ async def read_ts_attribute_expect_success(self, endpoint, attribute):
         cluster = Clusters.Objects.TimeSynchronization
         return await self.read_single_attribute_check_success(endpoint=endpoint, cluster=cluster, attribute=attribute)
 
-    def pics_TC_TIMESYNC_2_1(self) -> list[str]:
-        return ["TIMESYNC.S"]
-
     @async_test_body
     async def test_TC_TIMESYNC_2_1(self):
-
         endpoint = self.user_params.get("endpoint", 0)
```
cecille, line 34: Actually, while you're in here, can you remove this endpoint check? The spec got changed after the test, so this cluster is now mandated to be on EP0 and only EP0, so this is just 0 always. ie, here just do endpoint=0

src/python_testing/TC_TIMESYNC_2_2.py
```diff
@@ -29,9 +29,6 @@ async def read_ts_attribute_expect_success(self, endpoint, attribute):
         cluster = Clusters.Objects.TimeSynchronization
         return await self.read_single_attribute_check_success(endpoint=endpoint, cluster=cluster, attribute=attribute)
 
-    def pics_TC_TIMESYNC_2_2(self) -> list[str]:
```
cecille, line 32: as above - leave this in for now.

src/python_testing/TC_TIMESYNC_2_1.py
```diff
@@ -29,45 +29,42 @@ async def read_ts_attribute_expect_success(self, endpoint, attribute):
         cluster = Clusters.Objects.TimeSynchronization
         return await self.read_single_attribute_check_success(endpoint=endpoint, cluster=cluster, attribute=attribute)
 
-    def pics_TC_TIMESYNC_2_1(self) -> list[str]:
```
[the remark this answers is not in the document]
j-ororke: Thank you, apologies I had not realized previously that this was needed after we removed the PICS from the test. 
I added this back in with the most recent changes.

src/python_testing/TC_TIMESYNC_2_1.py
j-ororke: Thank you, I have updated endpoint variable to be statically set to 0 in this test module.

src/python_testing/TC_TIMESYNC_2_2.py
j-ororke: Thank you, my apologies, in the most recent changes I have restored this part of the code.

src/python_testing/TC_TIMESYNC_2_1.py
```diff
@@ -34,40 +34,51 @@ def pics_TC_TIMESYNC_2_1(self) -> list[str]:
...
+        self.supports_ntps = bool(features & Clusters.TimeSynchronization.Bitmaps.Feature.kNTPServer)
+        self.supports_trusted_time_source = bool(features & Clusters.TimeSynchronization.Bitmaps.Feature.kTimeSyncClient)
+        timesource_attr_id = Clusters.Objects.TimeSynchronization.Attributes.TimeSource.attribute_id
+
+        # Reading attributes from DUT
+        timesync_attr_list = Clusters.Objects.TimeSynchronization.Attributes.AttributeList
+        read_request = await self.default_controller.ReadAttribute(self.dut_node_id, [timesync_attr_list])
+        attributes_list = read_request[0]
+
+        # Extracting gathered attributes dictionary values into a list containing only supported attribute ids
```
cecille, line 53: I'm not totally sure what this is doing - the attribute list should already be a list that only has the supported attribute ids. It's not guaranteed to be ordered.

You should just be able to read the attribute using the above function and check whether the time source is present by doing
if Clusters.TimeSynchronization.Attributes.TimeSource.attribute_id in attribute_list:

src/python_testing/TC_TIMESYNC_2_1.py
```diff
@@ -34,40 +34,51 @@ def pics_TC_TIMESYNC_2_1(self) -> list[str]:
...
+
+        self.supports_time_zone = bool(features & Clusters.TimeSynchronization.Bitmaps.Feature.kTimeZone)
+        self.supports_ntpc = bool(features & Clusters.TimeSynchronization.Bitmaps.Feature.kNTPClient)
+        self.supports_ntps = bool(features & Clusters.TimeSynchronization.Bitmaps.Feature.kNTPServer)
+        self.supports_trusted_time_source = bool(features & Clusters.TimeSynchronization.Bitmaps.Feature.kTimeSyncClient)
+        timesource_attr_id = Clusters.Objects.TimeSynchronization.Attributes.TimeSource.attribute_id
+
+        # Reading attributes from DUT
+        timesync_attr_list = Clusters.Objects.TimeSynchronization.Attributes.AttributeList
+        read_request = await self.default_controller.ReadAttribute(self.dut_node_id, [timesync_attr_list])
```
cecille, line 50: There's a function in matter_testing_support called "read_single_attribute_check_success" (or...something like that), which is what you'll want to use here. It will extract the attribute info from the endpoint

src/python_testing/TC_TIMESYNC_2_2.py
```diff
@@ -78,10 +78,9 @@ async def test_TC_TIMESYNC_2_2(self):
         compare_time(received=utc_dut, utc=th_utc, tolerance=tolerance)
 
         self.print_step(5, "Read time source")
-        if self.check_pics("TIMESYNC.S.A0002"):
-            source = await self.read_ts_attribute_expect_success(endpoint=endpoint, attribute=attributes.TimeSource)
-            if utc_dut_initial is NullValue:
-                asserts.assert_equal(source, Clusters.Objects.TimeSynchronization.Enums.TimeSourceEnum.kAdmin)
+        source = await self.read_ts_attribute_expect_success(endpoint=endpoint, attribute=attributes.TimeSource)
```
cecille, line 81: you'll need to do the same thing as above, where you first check if this attribute is supported, otherwise this is changing the test logic.

src/python_testing/TC_TIMESYNC_2_1.py
j-ororke: Thank you,
Using the above method had provided back an embedded dictionary, so had to parse the dictionary to get the attributes list out of it.
However, I have updated this section of the code in TC_TIMESYNC_2_1 test module to now use the get_single_attribute_check_success function from matter_testing_support module which provided the attributes list back in a list, so code is much cleaner now.

src/python_testing/TC_TIMESYNC_2_1.py
j-ororke: Thank you, updated to using read_single_attribute_check_success() from matter_testing_support works very nicely it appears and made this section of the code much cleaner now.

src/python_testing/TC_TIMESYNC_2_2.py
j-ororke: I have updated this section of the code in TC_TIMESYNC_2_2 to match with TC_TIMESYNC_2_1 test module now.

src/python_testing/TC_TIMESYNC_2_1.py
```diff
@@ -41,40 +41,47 @@ def pics_TC_TIMESYNC_2_1(self) -> list[str]:
...
-        endpoint = self.user_params.get("endpoint", 0)
+        features = await self.read_single_attribute(dev_ctrl=self.default_controller, node_id=self.dut_node_id,
+                                                    endpoint=endpoint, attribute=Clusters.TimeSynchronization.Attributes.FeatureMap)
+
+        self.supports_time_zone = bool(features & Clusters.TimeSynchronization.Bitmaps.Feature.kTimeZone)
+        self.supports_ntpc = bool(features & Clusters.TimeSynchronization.Bitmaps.Feature.kNTPClient)
+        self.supports_ntps = bool(features & Clusters.TimeSynchronization.Bitmaps.Feature.kNTPServer)
+        self.supports_trusted_time_source = bool(features & Clusters.TimeSynchronization.Bitmaps.Feature.kTimeSyncClient)
+
+        time_cluster = Clusters.Objects.TimeSynchronization
```
cecille, line 54:
```suggestion
        time_cluster = Clusters.TimeSynchronization
```

cecille: This isn't wrong, but we moved these back into the Clusters section in the init (no sure why...whatever), but if we keep this consistent, it's easier for folks to know what's going on

src/python_testing/TC_TIMESYNC_2_1.py
```diff
@@ -41,40 +41,47 @@ def pics_TC_TIMESYNC_2_1(self) -> list[str]:
...
+        timesync_attr_list = Clusters.Objects.TimeSynchronization.Attributes.AttributeList
```
cecille, line 55:
```suggestion
        timesync_attr_list = time_cluster.Attributes.AttributeList
```

cecille: might as well use this given that you already have the class ref in a var

src/python_testing/TC_TIMESYNC_2_1.py
```diff
@@ -41,40 +41,47 @@ def pics_TC_TIMESYNC_2_1(self) -> list[str]:
...
+        attribute_list = await self.read_single_attribute_check_success(endpoint=endpoint, cluster=time_cluster, attribute=timesync_attr_list)
+        timesource_attr_id = Clusters.Objects.TimeSynchronization.Attributes.TimeSource.attribute_id
```
cecille, line 57:
```suggestion
        timesource_attr_id = time_cluster.Attributes.TimeSource.attribute_id
```

src/python_testing/TC_TIMESYNC_2_2.py
```diff
@@ -46,6 +46,9 @@ async def test_TC_TIMESYNC_2_2(self):
         endpoint = 0
 
         time_cluster = Clusters.Objects.TimeSynchronization
+        timesync_attr_list = Clusters.Objects.TimeSynchronization.Attributes.AttributeList
```
cecille, line 49: changes as above

cecille (commented):
I'm good to approve this with the last nits addressed.

src/python_testing/TC_TIMESYNC_2_2.py
j-ororke: Resolved using your above mentioned suggestions in TC_TIMESYNC_2_1 test module

cecille (approved):

src/python_testing/TC_TIMESYNC_2_1.py
```diff
@@ -121,28 +124,23 @@ async def test_TC_TIMESYNC_2_1(self):
                 last_valid_until = dst.validUntil
                 if dst.validUntil is NullValue or dst.validUntil is None:
                     asserts.assert_equal(dst, dst_dut[-1], "DSTOffset list must have Null ValidUntil at the end")
-        elif self.check_pics("TIMESYNC.S.F00"):
-            asserts.assert_true(False, "DSTOffset is mandatory if the TZ (TIMESYNC.S.F00) feature is supported")
```
aronrosenberg, line 125: Do we still need this check for the DSTOffset attribute existence since dst_dut could be an empty list or does the `await self.read_ts_attribute_expect_success(`  call fail if the attribute isn't found?

aronrosenberg: For example, the test below for Read UTCTime has a `if utc_dut is NullValue:` which then fails the test.

aronrosenberg (approved):

src/python_testing/TC_TIMESYNC_2_1.py
j-ororke: Hi @aronrosenberg ,
Thank you for your review and comments on this PR, please find my response below!

The elif statement you have commented was removed since if the self.supports_time_zone is not true in this instance then test step 7 would be skipped which appears to be what the elif statement commented on was checking in this instance, as time zone is a required feature for DSTOffset attribute to be supported on the device.

This is a good question and the test appears should fail if the DSTOffset atttribute was not found in the async await function that you have mentioned ` dst_dut = await self.read_ts_attribute_expect_success(endpoint=endpoint, attribute=attributes.DSTOffset)` as it is expecting it to be a supported attribute since time zone is a supported feature on the device to get to that point. await The function should display below output in the terminal at the point of failure if the attribute was not found:
AttributeError: type object 'Attributes' has no attribute 'DSTOffset'

aronrosenberg: Makes sense, you can resolve this.

j-ororke: Thank you!

PR #33954, opened 2024-06-17
[Python] Store original PyChipError in ChipStackException

agners:
This change stores the original PyChipError in ChipStackException so that details of the original error code can still be retrieved. This is useful to get access to the properties returning processed information about the original error code. It also preserves the line and code file which can be helpful.

This also makes sure to use PyChipError return value where PyChipErrors are actually returned. This then also allows us to use the typical .raise_on_error() pattern. Previously, in several places just `c_uint_32` was used, which worked because that was the first element of the `PyChipErrors` struct. Using the correct type `PyChipErrors` is cleaner.

A drive-by find here was that ctypes functions argtypes did not get set correctly by the `NativeLibraryHandleMethodArguments` helper. Adjusting this caused some additional fixes.

### DIFF
--- a/src/controller/python/chip/setup_payload/setup_payload.py
+++ b/src/controller/python/chip/setup_payload/setup_payload.py
@@ -14,11 +14,10 @@
 #    limitations under the License.
 #
 
-from ctypes import CFUNCTYPE, c_char_p, c_int32, c_uint8, c_uint16, c_uint32
+from ctypes import CFUNCTYPE, c_char_p, c_uint8, c_uint16, c_uint32
 from typing import Optional
 
-from chip.exceptions import ChipStackError
-from chip.native import GetLibraryHandle, NativeLibraryHandleMethodArguments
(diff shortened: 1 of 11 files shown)


tehampson (approved):

src/controller/python/chip/setup_payload/setup_payload.py
```diff
@@ -46,34 +45,23 @@ def AddVendorAttribute(tag, value):
...
     def ParseQrCode(self, qrCode: str):
         self.Clear()
-        err = self.chipLib.pychip_SetupPayload_ParseQrCode(qrCode.upper().encode(),
-                                                           self.attribute_visitor,
-                                                           self.vendor_attribute_visitor)
-
-        if err != 0:
-            raise ChipStackError(err)
-
-        return self
```
arkq, line 56: The `ParseManualPairingCode` still returns `self`, so maybe this one should too.

arkq (approved):

PR #33963, opened 2024-06-18
Fix GRPKEY cluster in PICSGenerator

tcarmelveilleux:
- Group Key Management cluster was incorrectly mapped, leading to errors.

This PR:

- Fixes the mapping of GroupKeyManagement cluster

Testing done:

- Ran PICSGenerator against my product after changes, got no error where there previously was one.


src/tools/PICS-generator/PICSGenerator.py
```diff
@@ -53,6 +53,9 @@ def GenerateDevicePicsXmlFiles(clusterName, clusterPicsCode, featurePicsList, at
     elif "Group Key Management" == clusterName:
         clusterName = "Group Communication"
```
ReneJosefsen, line 55:
```suggestion
```

If you want we can just remove this one then. The names changes from time to time.

ReneJosefsen (approved):

src/tools/PICS-generator/PICSGenerator.py
tcarmelveilleux: This was tested, so I am leaving it :)

ReneJosefsen: No problem, we can always clean up later.

src/tools/PICS-generator/PICSGenerator.py
```diff
@@ -134,7 +138,8 @@ def GenerateDevicePicsXmlFiles(clusterName, clusterPicsCode, featurePicsList, at
     # Attributes PICS
     # TODO: Only check if list is not empty
     # console.print(attributePicsList)
-    serverAttributesNode = root.find("./clusterSide[@type='Server']/attributes")
+    serverAttributesNode = root.find(
```
andy31415, line 141: please change your autoformat plugin settings - this PR seems to need to be 2-3 line change, instead it got to 43/20 ... restyle.io should be the only restyler.

tcarmelveilleux: This is from restyler

ReneJosefsen: I wondered about this as well. but I figured we updated the restyler since I did the original version of this?

andy31415 (approved):

PR #33975, opened 2024-06-18
[Android] Support ICD LIT Device using commissionDevice

joonhaengHeo:
When Android App try to commission using `commissionDevice` method, App can't set ICD LIT config.
For this reason, I add new method for setting ICD LIT.


src/controller/java/src/chip/devicecontroller/ChipDeviceController.java
```diff
@@ -387,7 +387,29 @@ public void commissionDevice(long deviceId, @Nullable NetworkCredentials network
...
+   *
+   * @param deviceId the ID of the node to be commissioned
+   * @param csrNonce a nonce to be used for the CSR request
+   * @param networkCredentials the credentials (Wi-Fi or Thread) to be provisioned
+   * @param icdRegistrationInfo the informations for ICD registration. For detailed information
+   *     {@link ICDRegistrationInfo}. If this value is null when commissioning an ICD device, {@link
+   *     CompletionListener.onICDRegistrationInfoRequired} is called to request the
+   *     ICDRegistrationInfo value.
+   */
+  public void commissionDevice(
```
andy31415, line 406: adding more and more parameters to a method generally does not scale well. Would it be possible to maybe have some builder to set parameters instead? So that I can write:

```
commissonDevice(
   CommissionParameters.Builder()
       .setCsrNonce(...)
       .setNetworkCredentials(...)
       .setIcdRegistrationInfo(...)
)
```

and then you can omit what you do not need.

yunhanw-google (approved):

src/controller/java/src/chip/devicecontroller/ChipDeviceController.java
joonhaengHeo: I agree. There are many modifications, but there are many similar methods in ChipDeviceController. Therefore, I thought it would be better to change them all together, so I changed them all. Please review.

PR #33976, opened 2024-06-18
Pass largepayload flag into CommandSender in IM test app.

pidarped:
Use the session information from DeviceProxy to construct CommandSender for large payload if session supports it.

### DIFF
--- a/src/app/tests/suites/commands/interaction_model/InteractionModel.h
+++ b/src/app/tests/suites/commands/interaction_model/InteractionModel.h
@@ -252,7 +252,9 @@ protected:
             chip::app::CommandPathParams commandPath = { endpointId, clusterId, commandId,
                                                          (chip::app::CommandPathFlags::kEndpointIdValid) };
             auto commandSender                       = std::make_unique<chip::app::CommandSender>(
-                mCallback, device->GetExchangeManager(), mTimedInteractionTimeoutMs.HasValue(), mSuppressResponse.ValueOr(false));
+                mCallback, device->GetExchangeManager(), mTimedInteractionTimeoutMs.HasValue(), mSuppressResponse.ValueOr(false),
+                device->GetSecureSession().Value()->AllowsLargePayload());
+
             VerifyOrReturnError(commandSender != nullptr, CHIP_ERROR_NO_MEMORY);
 
             chip::app::CommandSender::AddRequestDataParameters addRequestDataParams(mTimedInteractionTimeoutMs);


src/app/tests/suites/commands/interaction_model/InteractionModel.h
```diff
@@ -251,8 +251,20 @@ class InteractionModelCommands
...
+                                                                           mTimedInteractionTimeoutMs.HasValue(),
+                                                                           mSuppressResponse.ValueOr(false),
+                                                                           /* allowLargePayload = */ true);
+            }
+            else
+            {
+                commandSender = std::make_unique<chip::app::CommandSender>(mCallback, device->GetExchangeManager(),
+                                                                           mTimedInteractionTimeoutMs.HasValue(),
+                                                                           mSuppressResponse.ValueOr(false));
+            }
```
andy31415, line 267:
```suggestion
            std::unique_ptr<chip::app::CommandSender> commandSender =
                std::make_unique<chip::app::CommandSender>(
                     mCallback, 
                     device->GetExchangeManager(),
                     mTimedInteractionTimeoutMs.HasValue(),
                     mSuppressResponse.ValueOr(false),
                     device->GetSecureSession().Value()->AllowsLargePayload());

```

andy31415: For what it's worth, the constructor of commandSender taking in 3 separate bool values is terrible. This should have become a params object instead so we can have defaults and make it extensible. `CommandSender(callback, exchangeManager, true,false,true)` is not readable and even if you add comments you have to ensure order and comments match.

src/app/tests/suites/commands/interaction_model/InteractionModel.h
```diff
@@ -249,10 +249,12 @@ class InteractionModelCommands
         while (repeat--)
         {
 
-            chip::app::CommandPathParams commandPath = { endpointId, clusterId, commandId,
-                                                         (chip::app::CommandPathFlags::kEndpointIdValid) };
-            auto commandSender                       = std::make_unique<chip::app::CommandSender>(
-                mCallback, device->GetExchangeManager(), mTimedInteractionTimeoutMs.HasValue(), mSuppressResponse.ValueOr(false));
+            chip::app::CommandPathParams commandPath                = { endpointId, clusterId, commandId,
+                                                                        (chip::app::CommandPathFlags::kEndpointIdValid) };
+            std::unique_ptr<chip::app::CommandSender> commandSender = std::make_unique<chip::app::CommandSender>(
```
bzbarsky-apple, line 254: Why is this not using `auto`?  This is the canonical case for `auto`: the entire type is already spelled out on the right-hand side of the assignment.

pidarped: Yes, this can be an auto. It was under an if else block before with the declaration before that, but now after bringing the AllowLargePayload() call inline it is back to a one-liner and should be an auto. Will change it.

yufengwangca (approved):

bzbarsky-apple (approved):

PR #33977, opened 2024-06-18
Add --wifipaf commission in chip-tool and apps of Linux platform

crlonxp:
* Need to enable the function by adding "chip_device_config_enable_wifipaf=true"
	* chip-tool: Add wifi-paf option for pairing. 
	*     Example: $ chip-tool pairing wifipaf 1 ap_ssid ap_pwd [setup-pin-code] [discriminator]
	* apps: Add --wifipaf. 
	*     Example: $ chip-all-clusters-app --wifi --wifipaf Note: Need the usd functions of wpa_supplicant to run this option

### DIFF
--- a/examples/chip-tool/commands/pairing/Commands.h
+++ b/examples/chip-tool/commands/pairing/Commands.h
@@ -181,6 +181,16 @@ public:
     {}
 };
 
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+class PairWiFiPAF : public PairingCommand
+{
+public:
+    PairWiFiPAF(CredentialIssuerCommands * credsIssuerConfig) :
+        PairingCommand("wifi-paf", PairingMode::WiFiPAF, PairingNetworkType::WiFi, credsIssuerConfig)
+    {}
+};
+#endif
+
 class PairAlreadyDiscovered : public PairingCommand
 {
 public:
@@ -243,6 +253,9 @@ void registerCommandsPairing(Commands & commands, CredentialIssuerCommands * cre
         make_unique<PairBleWiFi>(credsIssuerConfig),
         make_unique<PairBleThread>(credsIssuerConfig),
         make_unique<PairSoftAP>(credsIssuerConfig),
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+        make_unique<PairWiFiPAF>(credsIssuerConfig),
+#endif
         make_unique<PairAlreadyDiscovered>(credsIssuerConfig),
         make_unique<PairAlreadyDiscoveredByIndex>(credsIssuerConfig),
         make_unique<PairAlreadyDiscoveredByIndexWithWiFi>(credsIssuerConfig),
--- a/examples/chip-tool/commands/pairing/Commands.h
+++ b/examples/chip-tool/commands/pairing/Commands.h
@@ -186,7 +186,7 @@ class PairWiFiPAF : public PairingCommand
 {
 public:
     PairWiFiPAF(CredentialIssuerCommands * credsIssuerConfig) :
-        PairingCommand("wifi-paf", PairingMode::WiFiPAF, PairingNetworkType::WiFi, credsIssuerConfig)
+        PairingCommand("wifipaf-wifi", PairingMode::WiFiPAF, PairingNetworkType::WiFi, credsIssuerConfig)
     {}
 };
 #endif
--- a/examples/chip-tool/commands/payload/SetupPayloadParseCommand.cpp
+++ b/examples/chip-tool/commands/payload/SetupPayloadParseCommand.cpp
@@ -108,6 +108,14 @@ CHIP_ERROR SetupPayloadParseCommand::Print(chip::SetupPayload payload)
                     }
                     humanFlags.Add("On IP network");
                 }
+                if (payload.rendezvousInformation.Value().Has(RendezvousInformationFlag::kWiFiPAF))
+                {
+                    if (!humanFlags.Empty())
+                    {
+                        humanFlags.Add(", ");
+                    }
+                    humanFlags.Add("WIFIPAF");
+                }
             }
             else
             {
--- a/examples/chip-tool/commands/payload/SetupPayloadParseCommand.cpp
+++ b/examples/chip-tool/commands/payload/SetupPayloadParseCommand.cpp
@@ -114,7 +114,7 @@ CHIP_ERROR SetupPayloadParseCommand::Print(chip::SetupPayload payload)
                     {
                         humanFlags.Add(", ");
                     }
-                    humanFlags.Add("WIFIPAF");
+                    humanFlags.Add("Wi-Fi PAF");
                 }
             }
             else
--- a/examples/platform/linux/AppMain.cpp
+++ b/examples/platform/linux/AppMain.cpp
@@ -366,14 +366,15 @@ int ChipLinuxAppInit(int argc, char * const argv[], OptionSet * customOptions,
     CHIP_ERROR err = CHIP_NO_ERROR;
 #if CONFIG_NETWORK_LAYER_BLE
     RendezvousInformationFlags rendezvousFlags = RendezvousInformationFlag::kBLE;
-#elif CHIP_DEVICE_CONFIG_ENABLE_WPA && CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
-    RendezvousInformationFlags rendezvousFlags = RendezvousInformationFlag::kWiFiPAF;
 #else  // CONFIG_NETWORK_LAYER_BLE
     RendezvousInformationFlag rendezvousFlags = RendezvousInformationFlag::kOnNetwork;
 #endif // CONFIG_NETWORK_LAYER_BLE
 
 #ifdef CONFIG_RENDEZVOUS_MODE
     rendezvousFlags = static_cast<RendezvousInformationFlags>(CONFIG_RENDEZVOUS_MODE);
+#endif
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+    rendezvousFlags.Set(RendezvousInformationFlag::kWiFiPAF);
 #endif
 
     err = Platform::MemoryInit();
--- a/examples/platform/linux/AppMain.cpp
+++ b/examples/platform/linux/AppMain.cpp
@@ -483,7 +483,7 @@ int ChipLinuxAppInit(int argc, char * const argv[], OptionSet * customOptions,
             ChipLogProgress(NotSpecified, "Wi-Fi Management started");
             DeviceLayer::ConnectivityManager::WiFiPAFAdvertiseParam args;
             args.enable  = LinuxDeviceOptions::GetInstance().mWiFiPAF;
-            args.ExtCmds = LinuxDeviceOptions::GetInstance().mWiFiPafExtCmds;
+            args.ExtCmds = LinuxDeviceOptions::GetInstance().mWiFiPAFExtCmds;
             DeviceLayer::ConnectivityMgr().SetWiFiPAFAdvertisingEnabled(args);
         }
     }
--- a/examples/platform/linux/AppMain.cpp
+++ b/examples/platform/linux/AppMain.cpp
@@ -366,6 +366,8 @@ int ChipLinuxAppInit(int argc, char * const argv[], OptionSet * customOptions,
     CHIP_ERROR err = CHIP_NO_ERROR;
 #if CONFIG_NETWORK_LAYER_BLE
     RendezvousInformationFlags rendezvousFlags = RendezvousInformationFlag::kBLE;
+#elif CHIP_DEVICE_CONFIG_ENABLE_WPA && CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+    RendezvousInformationFlags rendezvousFlags = RendezvousInformationFlag::kWiFiPAF;
 #else  // CONFIG_NETWORK_LAYER_BLE
     RendezvousInformationFlag rendezvousFlags = RendezvousInformationFlag::kOnNetwork;
 #endif // CONFIG_NETWORK_LAYER_BLE
@@ -471,6 +473,17 @@ int ChipLinuxAppInit(int argc, char * const argv[], OptionSet * customOptions,
         }
     }
 #endif // CHIP_DEVICE_CONFIG_ENABLE_WPA
+#if CHIP_DEVICE_CONFIG_ENABLE_WPA && CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+    ChipLogProgress(NotSpecified, "WiFi-PAF: initialzing");
+    if (LinuxDeviceOptions::GetInstance().mWiFi)
+    {
+        if (EnsureWiFiIsStarted())
+        {
+            ChipLogProgress(NotSpecified, "Wi-Fi Management started");
+            DeviceLayer::ConnectivityMgr().SetWiFiPAFAdvertisingEnabled(LinuxDeviceOptions::GetInstance().mWiFiPAF);
+        }
+    }
+#endif
 
 #if CHIP_ENABLE_OPENTHREAD
     if (LinuxDeviceOptions::GetInstance().mThread)
--- a/examples/platform/linux/AppMain.cpp
+++ b/examples/platform/linux/AppMain.cpp
@@ -481,7 +481,10 @@ int ChipLinuxAppInit(int argc, char * const argv[], OptionSet * customOptions,
         if (EnsureWiFiIsStarted())
         {
             ChipLogProgress(NotSpecified, "Wi-Fi Management started");
-            DeviceLayer::ConnectivityMgr().SetWiFiPAFAdvertisingEnabled(LinuxDeviceOptions::GetInstance().mWiFiPAF);
+            DeviceLayer::ConnectivityManager::WiFiPAFAdvertiseParam args;
+            args.enable = LinuxDeviceOptions::GetInstance().mWiFiPAF;
+            args.ExtCmds = LinuxDeviceOptions::GetInstance().mWiFiPafExtCmds;
+            DeviceLayer::ConnectivityMgr().SetWiFiPAFAdvertisingEnabled(args);
         }
     }
 #endif
--- a/examples/platform/linux/AppMain.cpp
+++ b/examples/platform/linux/AppMain.cpp
@@ -367,7 +367,7 @@ int ChipLinuxAppInit(int argc, char * const argv[], OptionSet * customOptions,
 #if CONFIG_NETWORK_LAYER_BLE
     RendezvousInformationFlags rendezvousFlags = RendezvousInformationFlag::kBLE;
 #else  // CONFIG_NETWORK_LAYER_BLE
-    RendezvousInformationFlag rendezvousFlags = RendezvousInformationFlag::kOnNetwork;
+    RendezvousInformationFlags rendezvousFlags = RendezvousInformationFlag::kOnNetwork;
 #endif // CONFIG_NETWORK_LAYER_BLE
 
 #ifdef CONFIG_RENDEZVOUS_MODE
--- a/examples/platform/linux/AppMain.cpp
+++ b/examples/platform/linux/AppMain.cpp
@@ -482,7 +482,7 @@ int ChipLinuxAppInit(int argc, char * const argv[], OptionSet * customOptions,
         {
             ChipLogProgress(NotSpecified, "Wi-Fi Management started");
             DeviceLayer::ConnectivityManager::WiFiPAFAdvertiseParam args;
-            args.enable = LinuxDeviceOptions::GetInstance().mWiFiPAF;
+            args.enable  = LinuxDeviceOptions::GetInstance().mWiFiPAF;
             args.ExtCmds = LinuxDeviceOptions::GetInstance().mWiFiPafExtCmds;
             DeviceLayer::ConnectivityMgr().SetWiFiPAFAdvertisingEnabled(args);
         }
--- a/examples/platform/linux/Options.cpp
+++ b/examples/platform/linux/Options.cpp
@@ -121,7 +121,7 @@ OptionDef sDeviceOptionDefs[] = {
     { "wifi", kNoArgument, kDeviceOption_WiFi },
     { "wifi-supports-5g", kNoArgument, kDeviceOption_WiFiSupports5g },
 #if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
-    { "wifipaf", kNoArgument, kDeviceOption_WiFi_PAF },
+    { "wifipaf", kArgumentRequired, kDeviceOption_WiFi_PAF },
 #endif // CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
 #endif // CHIP_DEVICE_CONFIG_ENABLE_WPA
 #if CHIP_ENABLE_OPENTHREAD
@@ -197,9 +197,10 @@ const char * sDeviceOptionHelp =
 #endif // CHIP_DEVICE_CONFIG_ENABLE_WIFI
 #if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
     "\n"
-    "  --wifipaf\n"
+    "  --wifipaf freq_list=<freq_1>,<freq_2>... \n"
     "       Enable Wi-Fi PAF via wpa_supplicant.\n"
-#endif // CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+    "       Give an empty string if not setting freq_list: \"\"\n"
+#endif // CHIP_DEVICE_CONFIG_ENABLE_WIFIPAFs
 #if CHIP_ENABLE_OPENTHREAD
     "\n"
     "  --thread\n"
@@ -603,6 +604,7 @@ bool HandleOption(const char * aProgram, OptionSet * aOptions, int aIdentifier, 
 #if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
     case kDeviceOption_WiFi_PAF: {
         LinuxDeviceOptions::GetInstance().mWiFiPAF = true;
+        LinuxDeviceOptions::GetInstance().mWiFiPafExtCmds = aValue;
         break;
     }
 #endif
--- a/examples/platform/linux/Options.cpp
+++ b/examples/platform/linux/Options.cpp
@@ -603,7 +603,7 @@ bool HandleOption(const char * aProgram, OptionSet * aOptions, int aIdentifier, 
 #endif
 #if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
     case kDeviceOption_WiFi_PAF: {
-        LinuxDeviceOptions::GetInstance().mWiFiPAF = true;
+        LinuxDeviceOptions::GetInstance().mWiFiPAF        = true;
         LinuxDeviceOptions::GetInstance().mWiFiPafExtCmds = aValue;
         break;
     }
--- a/examples/platform/linux/Options.cpp
+++ b/examples/platform/linux/Options.cpp
@@ -107,7 +107,7 @@ enum
     kDeviceOption_FaultInjection,
 #endif
 #if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
-    kDeviceOption_WiFi_PAF = 0x1028,
+    kDeviceOption_WiFi_PAF,
 #endif
 };
 
--- a/examples/platform/linux/Options.cpp
+++ b/examples/platform/linux/Options.cpp
@@ -106,6 +106,9 @@ enum
 #if CHIP_WITH_NLFAULTINJECTION
     kDeviceOption_FaultInjection,
 #endif
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+    kDeviceOption_WiFi_PAF = 0x1028,
+#endif
 };
 
 constexpr unsigned kAppUsageLength = 64;
@@ -117,6 +120,9 @@ OptionDef sDeviceOptionDefs[] = {
 #if CHIP_DEVICE_CONFIG_ENABLE_WIFI
     { "wifi", kNoArgument, kDeviceOption_WiFi },
     { "wifi-supports-5g", kNoArgument, kDeviceOption_WiFiSupports5g },
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+    { "wifi-paf", kNoArgument, kDeviceOption_WiFi_PAF },
+#endif // CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
 #endif // CHIP_DEVICE_CONFIG_ENABLE_WPA
 #if CHIP_ENABLE_OPENTHREAD
     { "thread", kNoArgument, kDeviceOption_Thread },
@@ -189,6 +195,11 @@ const char * sDeviceOptionHelp =
     "  --wifi-supports-5g\n"
     "       Indicate that local Wi-Fi hardware should report 5GHz support.\n"
 #endif // CHIP_DEVICE_CONFIG_ENABLE_WIFI
+#if (CHIP_DEVICE_CONFIG_ENABLE_WIFI && CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF)
+    "\n"
+    "  --wifi-paf\n"
+    "       Enable Wi-Fi PAF via wpa_supplicant.\n"
+#endif // CHIP_DEVICE_CONFIG_ENABLE_WIFI && CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
 #if CHIP_ENABLE_OPENTHREAD
     "\n"
     "  --thread\n"
@@ -586,6 +597,12 @@ bool HandleOption(const char * aProgram, OptionSet * aOptions, int aIdentifier, 
             PrintArgError("%s: Invalid fault injection specification\n", aProgram);
             retval = false;
         }
+        break;
+    }
+#endif
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+    case kDeviceOption_WiFi_PAF: {
+        LinuxDeviceOptions::GetInstance().mWiFiPAF = true;
         break;
     }
 #endif
--- a/examples/platform/linux/Options.cpp
+++ b/examples/platform/linux/Options.cpp
@@ -604,7 +604,7 @@ bool HandleOption(const char * aProgram, OptionSet * aOptions, int aIdentifier, 
 #if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
     case kDeviceOption_WiFi_PAF: {
         LinuxDeviceOptions::GetInstance().mWiFiPAF        = true;
-        LinuxDeviceOptions::GetInstance().mWiFiPafExtCmds = aValue;
+        LinuxDeviceOptions::GetInstance().mWiFiPAFExtCmds = aValue;
         break;
     }
 #endif
--- a/examples/platform/linux/Options.cpp
+++ b/examples/platform/linux/Options.cpp
@@ -121,7 +121,7 @@ OptionDef sDeviceOptionDefs[] = {
     { "wifi", kNoArgument, kDeviceOption_WiFi },
     { "wifi-supports-5g", kNoArgument, kDeviceOption_WiFiSupports5g },
 #if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
-    { "wifi-paf", kNoArgument, kDeviceOption_WiFi_PAF },
+    { "wifipaf", kNoArgument, kDeviceOption_WiFi_PAF },
 #endif // CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
 #endif // CHIP_DEVICE_CONFIG_ENABLE_WPA
 #if CHIP_ENABLE_OPENTHREAD
@@ -195,11 +195,11 @@ const char * sDeviceOptionHelp =
     "  --wifi-supports-5g\n"
     "       Indicate that local Wi-Fi hardware should report 5GHz support.\n"
 #endif // CHIP_DEVICE_CONFIG_ENABLE_WIFI
-#if (CHIP_DEVICE_CONFIG_ENABLE_WIFI && CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF)
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
     "\n"
-    "  --wifi-paf\n"
+    "  --wifipaf\n"
     "       Enable Wi-Fi PAF via wpa_supplicant.\n"
-#endif // CHIP_DEVICE_CONFIG_ENABLE_WIFI && CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+#endif // CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
 #if CHIP_ENABLE_OPENTHREAD
     "\n"
     "  --thread\n"
--- a/examples/platform/linux/Options.h
+++ b/examples/platform/linux/Options.h
@@ -50,7 +50,7 @@ struct LinuxDeviceOptions
     bool mWiFi                 = false;
     bool mThread               = false;
 #if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
-    bool mWiFiPAF              = false;
+    bool mWiFiPAF = false;
 #endif
 #if CHIP_DEVICE_CONFIG_ENABLE_BOTH_COMMISSIONER_AND_COMMISSIONEE || CHIP_DEVICE_ENABLE_PORT_PARAMS
     uint16_t securedDevicePort         = CHIP_PORT;
--- a/examples/platform/linux/Options.h
+++ b/examples/platform/linux/Options.h
@@ -49,6 +49,9 @@ struct LinuxDeviceOptions
     bool wifiSupports5g        = false;
     bool mWiFi                 = false;
     bool mThread               = false;
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+    bool mWiFiPAF              = false;
+#endif
 #if CHIP_DEVICE_CONFIG_ENABLE_BOTH_COMMISSIONER_AND_COMMISSIONEE || CHIP_DEVICE_ENABLE_PORT_PARAMS
     uint16_t securedDevicePort         = CHIP_PORT;
     uint16_t unsecuredCommissionerPort = CHIP_UDC_PORT;
--- a/examples/platform/linux/Options.h
+++ b/examples/platform/linux/Options.h
@@ -50,7 +50,7 @@ struct LinuxDeviceOptions
     bool mWiFi                 = false;
     bool mThread               = false;
 #if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
-    bool mWiFiPAF = false;
+    bool mWiFiPAF                = false;
     const char * mWiFiPafExtCmds = nullptr;
 #endif
 #if CHIP_DEVICE_CONFIG_ENABLE_BOTH_COMMISSIONER_AND_COMMISSIONEE || CHIP_DEVICE_ENABLE_PORT_PARAMS
--- a/examples/platform/linux/Options.h
+++ b/examples/platform/linux/Options.h
@@ -51,6 +51,7 @@ struct LinuxDeviceOptions
     bool mThread               = false;
 #if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
     bool mWiFiPAF = false;
+    const char * mWiFiPafExtCmds = nullptr;
 #endif
 #if CHIP_DEVICE_CONFIG_ENABLE_BOTH_COMMISSIONER_AND_COMMISSIONEE || CHIP_DEVICE_ENABLE_PORT_PARAMS
     uint16_t securedDevicePort         = CHIP_PORT;
--- a/examples/platform/linux/Options.h
+++ b/examples/platform/linux/Options.h
@@ -51,7 +51,7 @@ struct LinuxDeviceOptions
     bool mThread               = false;
 #if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
     bool mWiFiPAF                = false;
-    const char * mWiFiPafExtCmds = nullptr;
+    const char * mWiFiPAFExtCmds = nullptr;
 #endif
 #if CHIP_DEVICE_CONFIG_ENABLE_BOTH_COMMISSIONER_AND_COMMISSIONEE || CHIP_DEVICE_ENABLE_PORT_PARAMS
     uint16_t securedDevicePort         = CHIP_PORT;
--- a/src/controller/AutoCommissioner.cpp
+++ b/src/controller/AutoCommissioner.cpp
@@ -591,7 +591,8 @@ CHIP_ERROR AutoCommissioner::StartCommissioning(DeviceCommissioner * commissione
     mCommissioner            = commissioner;
     mCommissioneeDeviceProxy = proxy;
 
-    auto transportType = mCommissioneeDeviceProxy->GetSecureSession().Value()->AsSecureSession()->GetPeerAddress().GetTransportType();
+    auto transportType =
+        mCommissioneeDeviceProxy->GetSecureSession().Value()->AsSecureSession()->GetPeerAddress().GetTransportType();
     mNeedsNetworkSetup = (transportType == Transport::Type::kBle);
 #if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
     mNeedsNetworkSetup = mNeedsNetworkSetup || (transportType == Transport::Type::kWiFiPAF);
--- a/src/controller/AutoCommissioner.cpp
+++ b/src/controller/AutoCommissioner.cpp
@@ -596,8 +596,9 @@ CHIP_ERROR AutoCommissioner::StartCommissioning(DeviceCommissioner * commissione
     CHIP_ERROR err               = CHIP_NO_ERROR;
     CommissioningStage nextStage = GetNextCommissioningStage(CommissioningStage::kSecurePairing, err);
 #if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
-    mNeedsNetworkSetup = mNeedsNetworkSetup || (mCommissioneeDeviceProxy->GetSecureSession().Value()->AsSecureSession()->GetPeerAddress().GetTransportType() ==
-            Transport::Type::kWiFiPAF);
+    mNeedsNetworkSetup = mNeedsNetworkSetup ||
+        (mCommissioneeDeviceProxy->GetSecureSession().Value()->AsSecureSession()->GetPeerAddress().GetTransportType() ==
+         Transport::Type::kWiFiPAF);
 #endif
     mCommissioner->PerformCommissioningStep(mCommissioneeDeviceProxy, nextStage, mParams, this, GetEndpoint(nextStage),
                                             GetCommandTimeout(mCommissioneeDeviceProxy, nextStage));
--- a/src/controller/AutoCommissioner.cpp
+++ b/src/controller/AutoCommissioner.cpp
@@ -592,10 +592,10 @@ CHIP_ERROR AutoCommissioner::StartCommissioning(DeviceCommissioner * commissione
     mCommissioneeDeviceProxy = proxy;
     mNeedsNetworkSetup =
         (mCommissioneeDeviceProxy->GetSecureSession().Value()->AsSecureSession()->GetPeerAddress().GetTransportType() ==
-        Transport::Type::kBle)
+         Transport::Type::kBle)
 #if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
         || (mCommissioneeDeviceProxy->GetSecureSession().Value()->AsSecureSession()->GetPeerAddress().GetTransportType() ==
-        Transport::Type::kWiFiPAF)
+            Transport::Type::kWiFiPAF)
 #endif
         ;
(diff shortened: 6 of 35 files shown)


src/transport/raw/PeerAddress.h
```diff
@@ -167,6 +170,11 @@ class PeerAddress
 #endif
                 snprintf(buf, bufSize, "TCP:[%s%s]:%d", ip_addr, interface, mPort);
             break;
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
```
bzbarsky-apple, line 173: This should probably not be conditional.

src/transport/raw/PeerAddress.h
```diff
@@ -53,6 +53,9 @@ enum class Type : uint8_t
     kUdp,
     kBle,
     kTcp,
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
```
bzbarsky-apple, line 56: This should not be conditional.

src/transport/raw/PeerAddress.h
```diff
@@ -208,6 +216,9 @@ class PeerAddress
     {
         return TCP(addr).SetPort(port).SetInterface(interface);
     }
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
```
bzbarsky-apple, line 219: Should not be conditional.

src/transport/raw/WiFiPAF.h
```diff
@@ -0,0 +1,116 @@
...
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+/**
+ *    @file
+ *      This file defines the CHIP Connection object that maintains a Wi-Fi PAF connection.
```
bzbarsky-apple, line 21: Matter, please, not CHIP.

bzbarsky-apple (changes requested):

examples/platform/linux/Options.cpp
```diff
@@ -106,6 +106,9 @@ enum
 #if CHIP_WITH_NLFAULTINJECTION
     kDeviceOption_FaultInjection,
 #endif
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+    kDeviceOption_WiFi_PAF = 0x1028,
```
ksperling-apple, line 110:
```suggestion
    kDeviceOption_WiFi_PAF,
```
Don't assign option values manually

examples/platform/linux/Options.cpp
```diff
@@ -189,6 +195,11 @@ const char * sDeviceOptionHelp =
     "  --wifi-supports-5g\n"
     "       Indicate that local Wi-Fi hardware should report 5GHz support.\n"
 #endif // CHIP_DEVICE_CONFIG_ENABLE_WIFI
+#if (CHIP_DEVICE_CONFIG_ENABLE_WIFI && CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF)
```
ksperling-apple, line 198: The other ifdefs all seem to be `#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF`.

The conditions should be consistent between the enum definition, the help text, and the code handling the option.

src/controller/AutoCommissioner.cpp
```diff
@@ -586,8 +586,13 @@ CHIP_ERROR AutoCommissioner::StartCommissioning(DeviceCommissioner * commissione
     mCommissioner            = commissioner;
     mCommissioneeDeviceProxy = proxy;
     mNeedsNetworkSetup =
-        mCommissioneeDeviceProxy->GetSecureSession().Value()->AsSecureSession()->GetPeerAddress().GetTransportType() ==
-        Transport::Type::kBle;
+        (mCommissioneeDeviceProxy->GetSecureSession().Value()->AsSecureSession()->GetPeerAddress().GetTransportType() ==
```
ksperling-apple, line 589: Please factor out the `(mCommissioneeDeviceProxy->GetSecureSession().Value()->AsSecureSession()->GetPeerAddress().GetTransportType()` expression into a local

src/controller/CHIPDeviceController.cpp
```diff
@@ -803,6 +812,37 @@ CHIP_ERROR DeviceCommissioner::EstablishPASEConnection(NodeId remoteDeviceId, Re
...
+                    {
+                        break;
+                    }
+                    usleep(kWiFiStartCheckTimeUsec);
+                }
+            }
+            if (!DeviceLayer::ConnectivityMgrImpl().IsWiFiManagementStarted())
+            {
+                ChipLogError(NotSpecified, "Wi-Fi Management taking too long to start - device configuration will be reset.");
+            }
```
ksperling-apple, line 839: This doesn't feel like the kind of code that belongs in this method, especially with the retries and busy waiting.

src/controller/CHIPDeviceController.cpp
```diff
@@ -871,6 +911,44 @@ void DeviceCommissioner::OnDiscoveredDeviceOverBleError(void * appState, CHIP_ER
...
+        LogErrorOnFailure(self->EstablishPASEConnection(remoteId, params));
+    }
+}
+
+void DeviceCommissioner::OnWiFiPAFSubscribeError(void * appState, CHIP_ERROR err)
+{
+    auto self   = static_cast<DeviceCommissioner *>(appState);
+    auto device = self->mDeviceInPASEEstablishment;
+
+    ChipLogProgress(Controller, "WiFi-PAF: Subscription Error!");
```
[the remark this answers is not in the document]
ksperling-apple: `err` should also be included in the log

src/include/platform/ConnectivityManager.h
```diff
@@ -171,6 +174,16 @@ class ConnectivityManager
     bool IsWiFiStationProvisioned();
     void ClearWiFiStationProvision();
     CHIP_ERROR GetAndLogWiFiStatsCounters();
+#if CHIP_DEVICE_CONFIG_ENABLE_WPA && CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+    CHIP_ERROR SetWiFiPAFAdvertisingEnabled(bool val);
+
+    typedef void (*OnConnectionCompleteFunct)(void * appState);
+    typedef void (*OnConnectionErrorFunct)(void * appState, CHIP_ERROR err);
+    CHIP_ERROR WiFiPAFConnect(void * appState, OnConnectionCompleteFunct onSuccess, OnConnectionErrorFunct onError);
```
ksperling-apple, line 182: Does this API assume there can only be a single PAF connection ongoing at a time? We shouldn't be repeating the mistakes made with the BLE implementation.

src/transport/raw/WiFiPAF.cpp
```diff
@@ -0,0 +1,96 @@
...
+    VerifyOrReturnError(mState == State::kNotReady, CHIP_ERROR_INCORRECT_STATE);
+    DeviceLayer::ConnectivityMgr().SetWiFiPAF(this);
+    mState = State::kInitialized;
+
+    return CHIP_NO_ERROR;
+}
+
+CHIP_ERROR WiFiPAFBase::SendMessage(const Transport::PeerAddress & address, System::PacketBufferHandle && msgBuf)
+{
+    ChipLogProgress(NotSpecified, "=====> WiFiPAFBase::SendMessage()");
```
ksperling-apple, line 62: Logging method entry and exit at Progress level and without any further details doesn't seem very useful.

src/controller/SetUpCodePairer.cpp
```diff
@@ -243,6 +254,19 @@ CHIP_ERROR SetUpCodePairer::StopConnectOverSoftAP()
     return CHIP_NO_ERROR;
 }
 
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+CHIP_ERROR SetUpCodePairer::StartDiscoverOverWiFiPAF(SetupPayload & payload)
+{
+    return CHIP_ERROR_UNSUPPORTED_CHIP_FEATURE;
```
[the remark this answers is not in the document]
ksperling-apple: We should consider refactoring the interfaces so that there is a generic DiscoveryTransport API that is implemented by both BLE and PAF that the SetupCodePairer and other parts of the stack can use. It seems rather sad to simply duplicate all the code instead.

examples/chip-tool/commands/pairing/Commands.h
```diff
@@ -173,6 +173,16 @@ class PairSoftAP : public PairingCommand
     {}
 };
 
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+class PairWiFiPAF : public PairingCommand
+{
+public:
+    PairWiFiPAF(CredentialIssuerCommands * credsIssuerConfig) :
+        PairingCommand("wifi-paf", PairingMode::WiFiPAF, PairingNetworkType::WiFi, credsIssuerConfig)
```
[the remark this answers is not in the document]
crlonxp: PAF is a feature of Wi-Fi spec, which is to send the public action frame to advertise it's existence and exchanges the messages. The Wi-Fi only device can advertise itself from this. I will prefer using "wifipaf-wifi" if it's ok.
Could you let me know  which document should I reference for the rule of naming?
"code-wifi" is different since BLE is required

examples/platform/linux/AppMain.cpp
```diff
@@ -360,6 +360,8 @@ int ChipLinuxAppInit(int argc, char * const argv[], OptionSet * customOptions,
     CHIP_ERROR err = CHIP_NO_ERROR;
 #if CONFIG_NETWORK_LAYER_BLE
     RendezvousInformationFlags rendezvousFlags = RendezvousInformationFlag::kBLE;
+#elif CHIP_DEVICE_CONFIG_ENABLE_WPA && CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+    RendezvousInformationFlags rendezvousFlags = RendezvousInformationFlag::kWiFiPAF;
```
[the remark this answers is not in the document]
crlonxp: I will change it.

src/controller/CHIPDeviceController.cpp
```diff
@@ -803,6 +812,37 @@ CHIP_ERROR DeviceCommissioner::EstablishPASEConnection(NodeId remoteDeviceId, Re
...
+                    }
+                    usleep(kWiFiStartCheckTimeUsec);
+                }
+            }
+            if (!DeviceLayer::ConnectivityMgrImpl().IsWiFiManagementStarted())
+            {
+                ChipLogError(NotSpecified, "Wi-Fi Management taking too long to start - device configuration will be reset.");
+            }
+            mRendezvousParametersForDeviceDiscoveredOverWiFiPAF = params;
+            DeviceLayer::ConnectivityMgr().WiFiPAFConnect(this, OnWiFiPAFSubscribeComplete, OnWiFiPAFSubscribeError);
```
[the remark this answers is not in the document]
crlonxp: This function will use dbus to register the callback function to wpa_supplicant. If the controller is shutdown or destroyed, dbus will know and not reply back if the condition occurs.

src/controller/CHIPDeviceController.cpp
```diff
@@ -871,6 +911,44 @@ void DeviceCommissioner::OnDiscoveredDeviceOverBleError(void * appState, CHIP_ER
...
+
+    ChipLogProgress(Controller, "WiFi-PAF: Subscription Completed!");
+    if (nullptr != device && device->GetDeviceTransportType() == Transport::Type::kWiFiPAF)
+    {
+        auto remoteId = device->GetDeviceId();
+        auto params   = self->mRendezvousParametersForDeviceDiscoveredOverWiFiPAF;
+
+        self->mRendezvousParametersForDeviceDiscoveredOverWiFiPAF = RendezvousParameters();
+        self->ReleaseCommissioneeDevice(device);
+        DeviceLayer::ConnectivityMgr().GetWiFiPAF()->SetWiFiPAFState(Transport::WiFiPAFBase::State::kConnected);
```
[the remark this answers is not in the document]
crlonxp: I will fix it

src/controller/CHIPDeviceController.cpp
```diff
@@ -871,6 +911,44 @@ void DeviceCommissioner::OnDiscoveredDeviceOverBleError(void * appState, CHIP_ER
...
+        chip::DeviceLayer::StackLock stackLock;
```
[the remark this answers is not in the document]
crlonxp: I will fix it

src/controller/CHIPDeviceController.cpp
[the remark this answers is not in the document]
crlonxp: I will fix it

src/controller/CHIPDeviceController.cpp
```diff
@@ -871,6 +911,44 @@ void DeviceCommissioner::OnDiscoveredDeviceOverBleError(void * appState, CHIP_ER
 }
 #endif // CONFIG_NETWORK_LAYER_BLE
 
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+void DeviceCommissioner::OnWiFiPAFSubscribeComplete(void * appState)
+{
+    auto self   = static_cast<DeviceCommissioner *>(appState);
+    auto device = self->mDeviceInPASEEstablishment;
+
+    ChipLogProgress(Controller, "WiFi-PAF: Subscription Completed!");
```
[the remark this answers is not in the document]
crlonxp: I will fix it

src/controller/SetUpCodePairer.cpp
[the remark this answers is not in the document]
crlonxp: StartDiscoverOverWiFiPAF() has not been implemented and will do it soon. 
As for refactoring the interface, should it be another task which won't be done in here for now?

src/controller/SetUpCodePairer.h
```diff
@@ -150,6 +154,9 @@ class DLL_EXPORT SetUpCodePairer : public DevicePairingDelegate
         kBLETransport = 0,
         kIPTransport,
         kSoftAPTransport,
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+        kWiFiPAF,
```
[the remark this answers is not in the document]
crlonxp: I will fix it

src/include/platform/CHIPDeviceEvent.h
```diff
@@ -276,7 +276,10 @@ enum InternalEventTypes
      * This event should populate CHIPoBLEConnectionError structure.
      */
     kCHIPoBLEConnectionError,
-    kCHIPoBLENotifyConfirm
+    kCHIPoBLENotifyConfirm,
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+    kCHIPoWiFiPAFWriteReceived,
```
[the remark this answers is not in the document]
crlonxp: I will fix it

src/include/platform/ConnectivityManager.h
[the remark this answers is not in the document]
crlonxp: The "connection" is valid only in PASE period. The connection will drop and switch to the original network connection after it's ready. These callback functions are registered through the dbus interface to wpa_supplicant that it will know if appState is going away and not report up.

src/platform/Linux/ConnectivityManagerImpl.cpp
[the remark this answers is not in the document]
crlonxp: ok, thank you

src/setup_payload/SetupPayload.cpp
```diff
@@ -62,7 +62,12 @@ bool PayloadContents::isValidQRCodePayload(ValidationMode mode) const
     if (mode == ValidationMode::kProduce)
     {
         chip::RendezvousInformationFlags valid(RendezvousInformationFlag::kBLE, RendezvousInformationFlag::kOnNetwork,
-                                               RendezvousInformationFlag::kSoftAP);
+                                               RendezvousInformationFlag::kSoftAP
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
```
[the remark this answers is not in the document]
crlonxp: If this feature is on, but not in the flag, shouldn't we say it's invalid? Could you suggest me the right way to go? Thank you

src/setup_payload/SetupPayload.h
```diff
@@ -101,6 +101,9 @@ enum class RendezvousInformationFlag : uint8_t
     kSoftAP    = 1 << 0, ///< Device supports Wi-Fi softAP
     kBLE       = 1 << 1, ///< Device supports BLE
     kOnNetwork = 1 << 2, ///< Device supports Setup on network
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+    kWiFiPAF = 1 << 3, ///< Device supports Wi-Fi Public Action Frame for discovery
```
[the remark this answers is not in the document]
crlonxp: I will fix it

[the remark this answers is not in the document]
crlonxp: I will add the code in SetupPayloadParseCommand.cpp. 
Could you suggest me more information about adding tests in src/setup_payload/tests? Thank you

src/transport/raw/PeerAddress.h
crlonxp: I will fix it

src/transport/raw/PeerAddress.h
crlonxp: I will fix it

src/transport/raw/PeerAddress.h
crlonxp: I will fix it

src/transport/raw/WiFiPAF.h
crlonxp: I will fix it

examples/platform/linux/Options.cpp
crlonxp: I will fix it

src/controller/AutoCommissioner.cpp
crlonxp: I will fix it

src/controller/CHIPDeviceController.cpp
crlonxp: I will move it to the initialization function

src/include/platform/ConnectivityManager.h
crlonxp: Yes, I assume there is a single PAF at a time for now. We will check if more PAF concurrent connection are available.
BTW, could you let me know what's the mistake of the BLE implementation? Thank you

src/controller/SetUpCodePairer.cpp
```diff
@@ -129,6 +129,17 @@ CHIP_ERROR SetUpCodePairer::Connect(SetupPayload & payload)
             }
             VerifyOrReturnError(searchOverAll || CHIP_NO_ERROR == err || CHIP_ERROR_UNSUPPORTED_CHIP_FEATURE == err, err);
         }
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
```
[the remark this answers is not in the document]
crlonxp: Is it correct that we should just remove "#if" to make it unconditional?

src/platform/Linux/ConnectivityManagerImpl.cpp
```diff
@@ -797,6 +810,78 @@ bool ConnectivityManagerImpl::IsWiFiManagementStarted()
...
+    if (DeviceLayer::GetDeviceInstanceInfoProvider()->GetProductId(PafPublish_ssi.ProductId) != CHIP_NO_ERROR)
+    {
+        PafPublish_ssi.ProductId = 0;
+    }
+    if (DeviceLayer::GetDeviceInstanceInfoProvider()->GetVendorId(PafPublish_ssi.VendorId) != CHIP_NO_ERROR)
+    {
+        PafPublish_ssi.VendorId = 0;
+    }
+    strcat(args, NAN_PUBLISH_SSI_TAG);
+    ret = chip::Encoding::BytesToUppercaseHexString((uint8_t *) &PafPublish_ssi, sizeof(PafPublish_ssi), &args[strlen(args)],
```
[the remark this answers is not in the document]
crlonxp: I will fix it

src/transport/raw/WiFiPAF.cpp
ksperling-apple: Please don't mark comments as resolved until you have actually made the code changes.

src/include/platform/ConnectivityManager.h
ksperling-apple: The mistake in the BLE implementation was baking the assumption that we can only be connected to a single device over BLE at a time into the internal APIs. We will need the PAF APIs to cleanly support discovering multiple commissionable devices at the same time, and send and receive follow-up packets to/from one or more of them.

src/setup_payload/SetupPayload.cpp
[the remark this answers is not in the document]
ksperling-apple: Whether a SetupCode is simply determined by the Matter spec, independent of any particular device.
You can simply remove the `#if` around the inclusion of `RendezvousInformationFlag::kWiFiPAF`.

src/transport/raw/PeerAddress.h
```diff
@@ -208,6 +216,9 @@ class PeerAddress
...
+    static PeerAddress WiFiPAF() { return PeerAddress(Type::kWiFiPAF); }
```
ksperling-apple, line 216: A PeerAddress of type WiFiPAF should actually carry the relevant details like the peer MAC address, or at least some sort of identifier to distinguish connections at the lower level.

src/transport/raw/WiFiPAF.cpp
crlonxp: I double checked the code and it has been removed in the commit 391dd07d0b.
Do I miss anything?

src/include/platform/ConnectivityManager.h
crlonxp: May we support one single device at a time for now?

We think more tasks are required to support multiple sessions concurrently:
1. We think the CHIPDeviceController may need to change to support multiple PASEConnection concurrently
2. The chip-tool parameters should be changed to pair with multiple nodes concurrently
3. Maybe the manual_pairing_code and qrcode need to be changed as well.
Are these ready that we can test if supporting multiple sessions concurrently? If not, may we add it in the todo tasks in the next PR? If yes, could you suggest more information that we can reference? Thank you.

src/setup_payload/SetupPayload.cpp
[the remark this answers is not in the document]
crlonxp: Thank you for the suggestion. I will remove it.

src/transport/raw/PeerAddress.h
crlonxp: I will add the identifier

src/include/platform/ConnectivityManager.h
bzbarsky-apple: CHIPDeviceController would definitely need to change.

chip-tool might keep supporting only one thing at a time.

But the APIs for PAF that CHIPDeviceController uses should be done in such a way that they can handle more than one thing concurrently, ideally.

src/setup_payload/SetupPayload.h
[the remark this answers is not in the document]
bzbarsky-apple: @crlonxp Well, it needs test coverage for the new RendezvousInformationFlag value, no?

examples/chip-tool/commands/pairing/Commands.h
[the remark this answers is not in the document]
bzbarsky-apple: code-wifi should scan on all network technologies (PAF, DNS-SD, BLE) and then commission onto wifi.

"wifipaf-wifi" would be OK in general, but code-wifi needs to work with PAF as well.

>  Could you let me know which document should I reference for the rule of naming?

There is no document, just the existing naming bits.

src/controller/CHIPDeviceController.cpp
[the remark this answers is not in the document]
bzbarsky-apple: How will dbus know that?  I see nothing in this PR notifying it.

More to the point, dbus is not the only backend for PAF commissioning (obviously; there is no dbus on an iPhone, or Android phone), so the API needs to be designed in a way that allows the controller to notify things when it goes away, no?

Please do not resolve comments when they are not actually resolved.

src/controller/SetUpCodePairer.cpp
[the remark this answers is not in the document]
bzbarsky-apple: > StartDiscoverOverWiFiPAF() has not been implemented and will do it soon.

Just to be clear: having this work is a pre-requisite for this landing, because "code-wifi" needs to work for PAF, no?

src/include/platform/ConnectivityManager.h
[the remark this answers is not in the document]
bzbarsky-apple: How will it know the appState is going away?  There is no way to let it know that I can see.

examples/chip-tool/commands/pairing/Commands.h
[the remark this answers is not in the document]
crlonxp: The WiFiPAF has been added to code-wifi that we can use this method to pair with the device. Please review the latest patch

src/controller/SetUpCodePairer.cpp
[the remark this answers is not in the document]
crlonxp: The latest commit has added the implementation that code-wifi can work by using wifipaf to do pairing now. Please review.

src/setup_payload/SetupPayload.h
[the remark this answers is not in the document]
crlonxp: Yes, more test coverage should be required for the new feature. But I have no idea what tests should be added for now. 
May we leave this to the specification discussion in the group?

examples/chip-tool/commands/payload/SetupPayloadParseCommand.cpp
```diff
@@ -108,6 +108,14 @@ CHIP_ERROR SetupPayloadParseCommand::Print(chip::SetupPayload payload)
                     }
                     humanFlags.Add("On IP network");
                 }
+                if (payload.rendezvousInformation.Value().Has(RendezvousInformationFlag::kWiFiPAF))
+                {
+                    if (!humanFlags.Empty())
+                    {
+                        humanFlags.Add(", ");
+                    }
+                    humanFlags.Add("WIFIPAF");
```
bzbarsky-apple, line 117: Shouldn't this be more like:
```suggestion
                    humanFlags.Add("Wi-Fi PAF");
```

src/controller/AutoCommissioner.cpp
```diff
@@ -586,10 +586,15 @@ CHIP_ERROR AutoCommissioner::StartCommissioning(DeviceCommissioner * commissione
...
+         Transport::Type::kBle);
     CHIP_ERROR err               = CHIP_NO_ERROR;
     CommissioningStage nextStage = GetNextCommissioningStage(CommissioningStage::kSecurePairing, err);
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+    mNeedsNetworkSetup = mNeedsNetworkSetup ||
+        (mCommissioneeDeviceProxy->GetSecureSession().Value()->AsSecureSession()->GetPeerAddress().GetTransportType() ==
```
bzbarsky-apple, line 595: This still has the issue @ksperling-apple pointed out, but also looks wrong in the sense that if GetNextCommissioningStage examines mNeedsNetworkSetup it will get the wrong value, no?

src/controller/SetUpCodePairer.cpp
```diff
@@ -243,6 +252,32 @@ CHIP_ERROR SetUpCodePairer::StopConnectOverSoftAP()
...
+    }
+    return err;
+#else
+    return CHIP_ERROR_UNSUPPORTED_CHIP_FEATURE;
+#endif // CONFIG_NETWORK_LAYER_BLE
+}
+
+CHIP_ERROR SetUpCodePairer::StopConnectOverWiFiPAF()
+{
+    mWaitingForDiscovery[kWiFiPAFTransport] = false;
```
bzbarsky-apple, line 277: We handed a pointer to ourselves to `WiFiPAFConnect()`.  We might be about to be destroyed.  This is a security bug.

src/controller/SetUpCodePairer.cpp
```diff
@@ -335,6 +370,36 @@ void SetUpCodePairer::OnBLEDiscoveryError(CHIP_ERROR err)
...
 
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+void SetUpCodePairer::OnDiscoveredDeviceOverWifiPAF()
+{
+    ChipLogProgress(Controller, "Discovered device to be commissioned over WiFiPAF, RemoteId: %lu", mRemoteId);
+
+    mWaitingForDiscovery[kWiFiPAFTransport] = false;
+    auto param                              = SetUpCodePairerParameters();
+    param.SetPeerAddress(Transport::PeerAddress(Transport::Type::kWiFiPAF, mRemoteId));
+    mDiscoveredParameters.emplace_front(param);
```
bzbarsky-apple, line 381: This should document why it's "jumping the queue" to go in the front.

src/controller/SetUpCodePairer.cpp
```diff
@@ -335,6 +370,36 @@ void SetUpCodePairer::OnBLEDiscoveryError(CHIP_ERROR err)
...
+    ConnectToDiscoveredDevice();
+}
+
+void SetUpCodePairer::OnWifiPAFDiscoveryError(CHIP_ERROR err)
+{
+    ChipLogError(Controller, "Commissioning discovery over WiFiPAF failed: %" CHIP_ERROR_FORMAT, err.Format());
+    mWaitingForDiscovery[kWiFiPAFTransport] = false;
+    LogErrorOnFailure(err);
```
bzbarsky-apple, line 389: We already logged the error.  Why is this LogErrorOnFailure here?

examples/chip-tool/commands/pairing/Commands.h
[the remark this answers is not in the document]
bzbarsky-apple: Thank you.

src/controller/CHIPDeviceController.cpp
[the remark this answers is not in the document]
bzbarsky-apple: My point is that the API surface being provided here has no way to avoid use-after-free.

Again: the controller is being destroyed.  How does it tell the connectivity manager to not try to call into it?

src/controller/CHIPDeviceController.cpp
```diff
@@ -803,6 +812,24 @@ CHIP_ERROR DeviceCommissioner::EstablishPASEConnection(NodeId remoteDeviceId, Re
...
+        if (DeviceLayer::ConnectivityMgr().GetWiFiPAF()->GetWiFiPAFState() != Transport::WiFiPAFBase::State::kConnected)
+        {
+            ChipLogProgress(Controller, "WiFi-PAF: Subscribing the NAN-USD devices");
+            if (!DeviceLayer::ConnectivityMgrImpl().IsWiFiManagementStarted())
+            {
+                ChipLogError(Controller, "Wi-Fi Management should have be started now.");
+                ExitNow(CHIP_ERROR_INTERNAL);
+            }
+            mRendezvousParametersForDeviceDiscoveredOverWiFiPAF = params;
+            DeviceLayer::ConnectivityMgr().WiFiPAFConnect(params.GetSetupDiscriminator().value(), this, OnWiFiPAFSubscribeComplete,
```
bzbarsky-apple, line 828: So just to reiterate: That needs to be cancelable when `this` is destroyed.  We need to not have exploitable security bugs here.

src/include/platform/CHIPDeviceConfig.h
```diff
@@ -425,6 +425,15 @@
 #define CHIP_DEVICE_CONFIG_WIFI_STATION_IF_NAME "wlan0"
 #endif
 
+/**
+ * CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+ *
+ * Enable support for the WiFiPAF commissioning function
```
bzbarsky-apple, line 431:
```suggestion
 * Enable support for commissioning using Wi-Fi Public Action Frame as the transport.
```

src/include/platform/ConnectivityManager.h
```diff
@@ -171,6 +174,17 @@ class ConnectivityManager
     bool IsWiFiStationProvisioned();
     void ClearWiFiStationProvision();
     CHIP_ERROR GetAndLogWiFiStatsCounters();
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+    CHIP_ERROR SetWiFiPAFAdvertisingEnabled(bool val);
+
+    typedef void (*OnConnectionCompleteFunct)(void * appState);
+    typedef void (*OnConnectionErrorFunct)(void * appState, CHIP_ERROR err);
+    CHIP_ERROR WiFiPAFConnect(const SetupDiscriminator & connDiscriminator, void * appState, OnConnectionCompleteFunct onSuccess,
```
bzbarsky-apple, line 182: Again, this needs to have documentation and functionality for avoiding exploitable security bugs when appState is destroyed.

src/setup_payload/SetupPayload.h
bzbarsky-apple, line 1: Just to reiterate: needs tests.

Basic ones like parsing and serialization working correctly with this new discovery flag set.

src/transport/raw/PeerAddress.h
```diff
@@ -167,6 +171,9 @@ class PeerAddress
 #endif
                 snprintf(buf, bufSize, "TCP:[%s%s]:%d", ip_addr, interface, mPort);
             break;
+        case Type::kWiFiPAF:
+            snprintf(buf, bufSize, "WiFiPAF");
```
bzbarsky-apple, line 175:
```suggestion
            snprintf(buf, bufSize, "Wi-Fi PAF");
```

src/controller/SetUpCodePairer.cpp
crlonxp: Yes, it's more reasonable to go to the end. I will change it.

src/controller/AutoCommissioner.cpp
crlonxp: Yes, it's a problem. I will change it

src/controller/SetUpCodePairer.cpp
crlonxp: I will remove the redundant one.

src/setup_payload/SetupPayload.h
crlonxp: We have tried the cases of https://github.com/CHIP-Specifications/chip-test-plans/pull/4257/ by using code-wifi and the device can be paired

src/include/platform/ConnectivityManager.h
crlonxp: We just pushed a commit which uses weak_ptr to check if the pointer is valid or not.

src/controller/CHIPDeviceController.cpp
crlonxp: We just pushed a commit which uses weak_ptr to check if the pointer is valid or not.

src/controller/SetUpCodePairer.cpp
crlonxp: We just pushed a commit which uses weak_ptr to check if the pointer is valid or not while calling WiFiPAFConnect(). Will this still be a problem if having it?

src/include/platform/ConnectivityManager.h
crlonxp: So, may we keep using it for now?

src/controller/CHIPDeviceController.cpp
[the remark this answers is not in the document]
crlonxp: We just pushed a commit which uses weak_ptr to check if the pointer is valid or not.

src/include/platform/ConnectivityManager.h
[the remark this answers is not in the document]
crlonxp: We just pushed a commit which uses weak_ptr to check if the pointer is valid or not.

src/controller/CHIPDeviceController.cpp
```diff
@@ -90,6 +93,10 @@ using namespace chip::Encoding;
 #if CHIP_DEVICE_CONFIG_ENABLE_COMMISSIONER_DISCOVERY
 using namespace chip::Protocols::UserDirectedCommissioning;
 #endif // CHIP_DEVICE_CONFIG_ENABLE_COMMISSIONER_DISCOVERY
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+static std::shared_ptr<DeviceCommissioner> DevCommPtr;
```
bzbarsky-apple, line 97: I don't understand how this is supposed to work.  Per C++ spec:
```
If the object pointed to by ptr is already owned, the function generally results in undefined behavior. 
```
and that's the scenario here, no?  There are non-shared-ptr bit that will happily delete that object out from under you.

src/controller/CHIPDeviceController.cpp
```diff
@@ -804,6 +817,25 @@ CHIP_ERROR DeviceCommissioner::EstablishPASEConnection(NodeId remoteDeviceId, Re
...
+        {
+            ChipLogProgress(Controller, "WiFi-PAF: Subscribing the NAN-USD devices");
+            if (!DeviceLayer::ConnectivityMgrImpl().IsWiFiManagementStarted())
+            {
+                ChipLogError(Controller, "Wi-Fi Management should have be started now.");
+                ExitNow(CHIP_ERROR_INTERNAL);
+            }
+            DevCommPtr.reset(this, custom_del);
+            mRendezvousParametersForDeviceDiscoveredOverWiFiPAF = params;
+            DeviceLayer::ConnectivityMgr().WiFiPAFConnect(params.GetSetupDiscriminator().value(), (void *) (&(DevCommPtr)),
```
bzbarsky-apple, line 834: OK, so this passes a pointer to the `shared_ptr` as the void*, right?

src/controller/CHIPDeviceController.cpp
```diff
@@ -872,6 +904,59 @@ void DeviceCommissioner::OnDiscoveredDeviceOverBleError(void * appState, CHIP_ER
 }
 #endif // CONFIG_NETWORK_LAYER_BLE
 
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+void DeviceCommissioner::OnWiFiPAFSubscribeComplete(void * appState)
+{
+    std::weak_ptr<DeviceCommissioner> * caller     = (std::weak_ptr<DeviceCommissioner> *) (appState);
```
bzbarsky-apple, line 910: But this is casting the pointer we passed (to shared_ptr) to a weak_ptr?

I don't understand how this works.

src/controller/SetUpCodePairer.cpp
```diff
@@ -243,6 +253,39 @@ CHIP_ERROR SetUpCodePairer::StopConnectOverSoftAP()
     return CHIP_NO_ERROR;
 }
 
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+static std::shared_ptr<SetUpCodePairer> SetUpCodeParserPtr;
```
bzbarsky-apple, line 257: Again, I don't think this works.  SetUpCodePairer lifetime is not managed via the shared ptr....

src/setup_payload/SetupPayload.h
bzbarsky-apple, line 1: Just to repeat for the Nth time: this needs tests.  Not manual things that got run.  Just fix the file in the `tests` directory under here to test these bits.

bzbarsky-apple (changes requested):

src/setup_payload/SetupPayload.h
crlonxp: I think you are saying to add the tests in setup_payload/tests/TestQRCode.cpp, right?
I add more tests in the last commit. Please review. Thank you

examples/platform/linux/Options.h
```diff
@@ -50,7 +50,8 @@ struct LinuxDeviceOptions
     bool mWiFi                 = false;
     bool mThread               = false;
 #if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
-    bool mWiFiPAF = false;
+    bool mWiFiPAF                = false;
+    const char * mWiFiPafExtCmds = nullptr;
```
bzbarsky-apple, line 54: Why the inconsistent capitalization?

src/setup_payload/tests/TestQRCode.cpp
```diff
@@ -61,9 +64,25 @@ TEST(TestQRCode, TestRendezvousFlags)
...
+        RendezvousInformationFlags(RendezvousInformationFlag::kWiFiPAF, RendezvousInformationFlag::kOnNetwork));
+    EXPECT_TRUE(CheckWriteRead(inPayload));
+
     inPayload.rendezvousInformation.SetValue(RendezvousInformationFlags(
         RendezvousInformationFlag::kBLE, RendezvousInformationFlag::kSoftAP, RendezvousInformationFlag::kOnNetwork));
     EXPECT_TRUE(CheckWriteRead(inPayload));
+
+    inPayload.rendezvousInformation.SetValue(RendezvousInformationFlags(
+        RendezvousInformationFlag::kWiFiPAF, RendezvousInformationFlag::kSoftAP, RendezvousInformationFlag::kOnNetwork));
+
```
bzbarsky-apple, line 77: Shouldn't there be a ` EXPECT_TRUE(CheckWriteRead(inPayload))` here?  Otherwise, the next line will just reset things and this is not tested...

src/setup_payload/tests/TestQRCode.cpp
```diff
@@ -61,9 +64,25 @@ TEST(TestQRCode, TestRendezvousFlags)
...
+    inPayload.rendezvousInformation.SetValue(RendezvousInformationFlags(
+        RendezvousInformationFlag::kWiFiPAF, RendezvousInformationFlag::kBLE, RendezvousInformationFlag::kOnNetwork));
+
```
bzbarsky-apple, line 80: Again, should have ` EXPECT_TRUE(CheckWriteRead(inPayload))`.

src/setup_payload/SetupPayload.h
bzbarsky-apple: Yes, that is much better, thank you, except for the missing checks.

src/setup_payload/tests/TestQRCode.cpp
crlonxp: I will fix it. Thank you

examples/platform/linux/Options.h
crlonxp: I will fix it. Thank you

src/controller/CHIPDeviceController.cpp
```diff
@@ -905,19 +908,73 @@ void DeviceCommissioner::OnDiscoveredDeviceOverBleError(void * appState, CHIP_ER
...
+        {
+            ObjVector.push_back(appObj);
+            IsObjValid = true;
+        }
+    }
+    break;
+    case ObjChkAction::Check: {
+        for (auto lt = ObjVector.begin(); lt != ObjVector.end(); lt++)
+        {
+            if (*lt == appObj)
```
bzbarsky-apple, line 937: OK, so if an object got deallocated and another one got allocated at the same memory address, this check will pass, yes?  So we could be getting a callback for totally the wrong thing (or on totally the wrong object, depending on how you want to think about it).

src/controller/SetUpCodePairer.cpp
```diff
@@ -253,22 +260,16 @@ CHIP_ERROR SetUpCodePairer::StopConnectOverSoftAP()
...
 #if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
     ChipLogProgress(Controller, "Starting commissioning discovery over WiFiPAF");
-
     VerifyOrReturnError(mCommissioner != nullptr, CHIP_ERROR_INCORRECT_STATE);
     mWaitingForDiscovery[kWiFiPAFTransport] = true;
 
-    SetUpCodeParserPtr.reset(this, custom_del);
-    CHIP_ERROR err = DeviceLayer::ConnectivityMgr().WiFiPAFConnect(payload.discriminator, (void *) (&(SetUpCodeParserPtr)),
-                                                                   OnWiFiPAFSubscribeComplete, OnWiFiPAFSubscribeError);
+    DeviceCommissioner::SetChkObjValid((void *) this, DeviceCommissioner::ObjChkAction::Set, nullptr);
```
bzbarsky-apple, line 270: And because we are mixing objects of different types in the same set, we could have even bigger issues, where if a DeviceCommissioner gets allocated in the memory that a SetupCodePairer was in, or vice versa, we could end up with object type confusion bugs (read: exploitable security bug).

bzbarsky-apple (dismissed):

src/controller/SetUpCodePairer.cpp
crlonxp: Yes, will change it.

src/controller/CHIPDeviceController.cpp
crlonxp: Yes, I will change it

woody-apple (approved):

src/include/platform/ConnectivityManager.h
ksperling-apple: Please file an issue to track this follow-up work.

woody-apple:
Fast tracking this, this was caught in limbo, I don't know why. It appears to not be NXP only, but the generic support needed. Rebasing to see if this builds.

src/controller/CHIPDeviceController.cpp
```diff
@@ -466,6 +469,13 @@ DeviceCommissioner::DeviceCommissioner() :
     mDeviceNOCChainCallback(OnDeviceNOCChainGeneration, this), mSetUpCodePairer(this)
 {}
 
+DeviceCommissioner::~DeviceCommissioner()
+{
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+    DeviceLayer::ConnectivityMgr().WiFiPAFCancelConnect();
```
bzbarsky-apple, line 475: How does this work with multiple connects from different commissioner instances that might be outstanding at the same time?

src/controller/CHIPDeviceController.cpp
```diff
@@ -804,6 +820,24 @@ CHIP_ERROR DeviceCommissioner::EstablishPASEConnection(NodeId remoteDeviceId, Re
...
+#endif
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+    if (params.GetPeerAddress().GetTransportType() == Transport::Type::kWiFiPAF)
+    {
+        if (DeviceLayer::ConnectivityMgr().GetWiFiPAF()->GetWiFiPAFState() != Transport::WiFiPAFBase::State::kConnected)
+        {
+            ChipLogProgress(Controller, "WiFi-PAF: Subscribing the NAN-USD devices");
+            if (!DeviceLayer::ConnectivityMgrImpl().IsWiFiManagementStarted())
+            {
+                ChipLogError(Controller, "Wi-Fi Management should have be started now.");
```
bzbarsky-apple, line 832: Typos.  Please fix the typos throughout.

src/controller/CHIPDeviceController.cpp
```diff
@@ -804,6 +820,24 @@ CHIP_ERROR DeviceCommissioner::EstablishPASEConnection(NodeId remoteDeviceId, Re
...
+                ExitNow(CHIP_ERROR_INTERNAL);
+            }
+            mRendezvousParametersForDeviceDiscoveredOverWiFiPAF = params;
+            DeviceLayer::ConnectivityMgr().WiFiPAFConnect(params.GetSetupDiscriminator().value(), (void *) this,
```
bzbarsky-apple, line 836: Please do not use C-style casts.

src/controller/CHIPDeviceController.cpp
```diff
@@ -872,6 +906,43 @@ void DeviceCommissioner::OnDiscoveredDeviceOverBleError(void * appState, CHIP_ER
 }
 #endif // CONFIG_NETWORK_LAYER_BLE
 
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+void DeviceCommissioner::OnWiFiPAFSubscribeComplete(void * appState)
+{
+    auto self   = (DeviceCommissioner *) appState;
```
bzbarsky-apple, line 912: No C-style casts.

src/controller/CHIPDeviceController.cpp
bzbarsky-apple, line 475: Cancellation like this should happen in shutdown, not in the destructor.

src/controller/SetUpCodePairer.cpp
```diff
@@ -59,6 +60,13 @@ CHIP_ERROR GetPayload(const char * setUpCode, SetupPayload & payload)
 }
 } // namespace
 
+SetUpCodePairer::~SetUpCodePairer()
+{
+#if CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF
+    DeviceLayer::ConnectivityMgr().WiFiPAFCancelConnect();
```
bzbarsky-apple, line 66: This should happen when stopping (PAF) discovery, not in the destructor.

src/controller/SetUpCodePairer.cpp
```diff
@@ -243,6 +260,31 @@ CHIP_ERROR SetUpCodePairer::StopConnectOverSoftAP()
...
+    }
+    return err;
+#else
+    return CHIP_ERROR_UNSUPPORTED_CHIP_FEATURE;
+#endif // CONFIG_NETWORK_LAYER_BLE
+}
+
+CHIP_ERROR SetUpCodePairer::StopConnectOverWiFiPAF()
+{
+    mWaitingForDiscovery[kWiFiPAFTransport] = false;
```
bzbarsky-apple, line 284: Here.  Here is where the cancel needs to happen.

src/controller/SetUpCodePairer.cpp
```diff
@@ -335,6 +377,37 @@ void SetUpCodePairer::OnBLEDiscoveryError(CHIP_ERROR err)
...
+
+void SetUpCodePairer::OnWifiPAFDiscoveryError(CHIP_ERROR err)
+{
+    ChipLogError(Controller, "Commissioning discovery over WiFiPAF failed: %" CHIP_ERROR_FORMAT, err.Format());
+    mWaitingForDiscovery[kWiFiPAFTransport] = false;
+}
+
+void SetUpCodePairer::OnWiFiPAFSubscribeComplete(void * appState)
+{
+    auto self = (SetUpCodePairer *) appState;
```
bzbarsky-apple, line 400: No C-style casts.

src/controller/SetUpCodePairer.cpp
```diff
@@ -335,6 +377,37 @@ void SetUpCodePairer::OnBLEDiscoveryError(CHIP_ERROR err)
...
+    self->OnDiscoveredDeviceOverWifiPAF();
+}
+
+void SetUpCodePairer::OnWiFiPAFSubscribeError(void * appState, CHIP_ERROR err)
+{
+    auto self = (SetUpCodePairer *) appState;
```
bzbarsky-apple, line 406: No C-style casts.

bzbarsky-apple (dismissed):

woody-apple:
Discussed with @bzbarsky-apple , this meets the current bar for FC, but agree this needs significant follow up work.

src/controller/CHIPDeviceController.cpp
crlonxp: Yes, I will fix it

src/controller/CHIPDeviceController.cpp
crlonxp: I will change it

src/controller/CHIPDeviceController.cpp
crlonxp: I will fix it

src/controller/SetUpCodePairer.cpp
crlonxp: I will add it

src/controller/CHIPDeviceController.cpp
woody-apple: @crlonxp We'll allow this PR to merge, but please fast follow these fixes.

woody-apple: https://github.com/project-chip/connectedhomeip/issues/34612

src/controller/CHIPDeviceController.cpp
woody-apple: https://github.com/project-chip/connectedhomeip/issues/34612

src/controller/CHIPDeviceController.cpp
woody-apple: https://github.com/project-chip/connectedhomeip/issues/34612

src/controller/CHIPDeviceController.cpp
woody-apple: https://github.com/project-chip/connectedhomeip/issues/34612

src/controller/CHIPDeviceController.cpp
woody-apple: https://github.com/project-chip/connectedhomeip/issues/34612

src/controller/SetUpCodePairer.cpp
woody-apple: https://github.com/project-chip/connectedhomeip/issues/34612

src/controller/SetUpCodePairer.cpp
crlonxp: I will fix it

woody-apple:
Fast tracking, given this builds, and is ready for FC.

jczhang777:
I modified the src/platform/device.gni file. I changed chip_device_config_enable_wifipaf = false to chip_device_config_enable_wifipaf = true, and then compiled the chip tool. During the compilation, I encountered an error. The error log is as follows:

ERROR at //third_party/connectedhomeip/src/include/platform/ConnectivityManager.h:39:11: Include not allowed.
#include <transport/raw/WiFiPAF.h>
          ^----------------------
  //third_party/connectedhomeip/src/transport/raw:raw

Is there an issue with the way I made the modification?

PR #33980, opened 2024-06-18
[nrfconnect] Fixed Python requirements for factory data generation

markaj-nordic:
Recently the SetupPayload module has started utilizing `concrete` package. 
As a result, it has became an indirect dependency of implementation of the nrfconnect factory data generation.
Forward the `setup_payload/python/requirements.txt` to the nrfconnect specific requirements as a fix.


scripts/setup/requirements.nrfconnect.txt
```diff
@@ -1,6 +1,7 @@
+-r ../../src/setup_payload/python/requirements.txt
```
andy31415, line 1: I think we have to re-think this: our bootstrap logic in CI creates caches of environments and to determine if it is up to date or not, it does a hash of `scripts/setup` and `third_party/pigweed`: https://github.com/project-chip/connectedhomeip/blob/master/.github/actions/bootstrap-cache/action.yaml#L14

This means that if we do this redirection, we would not detect that python environment changed between cache runs if requirements.txt changes from src.

As ugly as this may be, it may be better to copy these over instead, maybe with a comment (in both files?) that these have to be in sync.

Damian-Nordic: or move `setup_payload/python/requirements.txt` to `scripts/setup/requirements.setuppayload.txt` and include then include that in `requirements.nrfconnect.txt`?

markaj-nordic: Ok, I will align the changes. Thanks.

Damian-Nordic (approved):

LuDuda (approved):

andy31415 (approved):

andy31415:
@markaj-nordic need to also change `.github/workflows/build.yaml` since builds fail with `ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'src/setup_payload/python/requirements.txt'`

PR #33983, opened 2024-06-18
Apply ColorTempPhysicalMaxMireds when ColorTemperatureMaximumMireds f…

### LINKED ISSUE #33903: [BUG] Incorrect handling of MoveColorTemperature and StepColorTemperature when ColorTemperatureMaximum is 0
### Reproduction steps

We have found the issue regarding MoveColorTemperature and StepColorTemperature. If the light receives the above command and ColorTemperatureMaximum is 0, the light will not take any action. The light should adjust its color temperature value according to the spec description.
```
ColorTemperatureMinimumMireds Field
If the ColorTemperatureMinimumMireds field is set to 0, ColorTempPhysicalMinMireds SHALL 
be used as the lower bound for the ColorTemperatureMireds attribute.

ColorTemperatureMaximumMireds Field
If the ColorTemperatureMaximum Mireds field is set to 0, ColorTempPhysicalMaxMireds 
SHALL be used as the upper bound for the ColorTemperatureMireds attribute.
```
But the SDK currently does not handle ColorTemperatureMaximum is 0.
https://github.com/project-chip/connectedhomeip/blob/master/src/app/clusters/color-control-server/color-control-server.cpp#L2700
https://github.com/project-chip/connectedhomeip/blob/master/src/app/clusters/color-control-server/color-control-server.cpp#L2811

### Bug prevalence

100%

d83caf7fefeaa91ebc549863552050311abe46ae

### Platform

core

### Platform Version(s)

_No response_

### Anything else?

_No response_

jmartinez-silabs:
…ield is 0 for `MoveColorTemperature` and `StepColorTemperature` commands.

Add test steps to TC_CC_6.2 and TC_CC_6.3 to validate the behaviour (tied with test plan pr [4235](https://github.com/CHIP-Specifications/chip-test-plans/pull/4235))

fixes #33903

### DIFF
--- a/src/app/clusters/color-control-server/color-control-server.cpp
+++ b/src/app/clusters/color-control-server/color-control-server.cpp
@@ -2697,7 +2697,7 @@ bool ColorControlServer::moveColorTempCommand(app::CommandHandler * commandObj, 
     {
         colorTemperatureMinimum = tempPhysicalMin;
     }
-    if (colorTemperatureMaximum > tempPhysicalMax)
+    if ((colorTemperatureMaximum == 0) || (colorTemperatureMaximum > tempPhysicalMax))
     {
         colorTemperatureMaximum = tempPhysicalMax;
     }
@@ -2808,7 +2808,7 @@ bool ColorControlServer::stepColorTempCommand(app::CommandHandler * commandObj, 
     {
         colorTemperatureMinimum = tempPhysicalMin;
     }
-    if (colorTemperatureMaximum > tempPhysicalMax)
+    if ((colorTemperatureMaximum == 0) || (colorTemperatureMaximum > tempPhysicalMax))
     {
         colorTemperatureMaximum = tempPhysicalMax;
     }
--- a/src/app/clusters/color-control-server/color-control-server.cpp
+++ b/src/app/clusters/color-control-server/color-control-server.cpp
@@ -2693,10 +2693,15 @@ bool ColorControlServer::moveColorTempCommand(app::CommandHandler * commandObj, 
         return true;
     }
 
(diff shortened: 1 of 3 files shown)


yufengwangca (approved):

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -2697,7 +2697,7 @@ bool ColorControlServer::moveColorTempCommand(app::CommandHandler * commandObj,
     {
         colorTemperatureMinimum = tempPhysicalMin;
     }
-    if (colorTemperatureMaximum > tempPhysicalMax)
+    if ((colorTemperatureMaximum == 0) || (colorTemperatureMaximum > tempPhysicalMax))
```
bzbarsky-apple, line 2700: The cited issue talks about the min temperature too, no?  Why is this PR only changing the maximum?

jmartinez-silabs: The min temperature is already covered by the existing if condition  as tempPhysicalMin cannot be lower than 0 anyway
```
if (colorTemperatureMinimum < tempPhysicalMin)
{
        colorTemperatureMinimum = tempPhysicalMin;
}
```

The issue only applied to `colorTemperatureMaximum` since the check was greater than only ence not catching the 0 condition.

bzbarsky-apple: It might be worth noting in a code comment that the spec requirement around `colorTemperatureMinimum == 0` is handled by the fact that `tempPhysicalMin >= 0`.

bzbarsky-apple (approved):

PR #33987, opened 2024-06-18
NXP folder architecture improvement

chapongatien:
**Description**:

Improving NXP matter folder architecture by moving NXP SDK support files in a dedicated submodule. 
The aim of this folder architecture rework is to reduce the number of NXP SDK files that are currently stored in the connectedhome repo by moving them in a dedicated “nxp_matter_support” repo that would be located in “third_party/nxp/nxp_matter_support”.

**Main changes summary:**

-	K32w0, k32w1 and rw61x files that are highly linked to NXP SDK have been moved to the nxp_matter_support submodule.
-	GN build files have been updated to reference new paths
-	Adding a new script called update_nxp_sdk.py allowing to ease initialization of recommended github NXP SDK(s) for platform that support it (k32w0 documentation has been updated) .


build_overrides/nxp_sdk.gni
```diff
@@ -14,5 +14,9 @@
 
 declare_args() {
   # Root directory for NXP SDKs.
-  nxp_sdk_build_root = "//third_party/nxp"
+  nxp_sdk_matter_support_root = "//third_party/nxp/nxp_matter_support"
+}
+
+declare_args() {
```
andy31415, line 20: could we merge this in a single declare args? saves some vertical space.

examples/build_overrides/nxp_sdk.gni
```diff
@@ -14,5 +14,10 @@
 
 declare_args() {
   # Root directory for NXP SDKs.
-  nxp_sdk_build_root = "//third_party/connectedhomeip/third_party/nxp"
+  nxp_sdk_matter_support_root =
+      "//third_party/connectedhomeip/third_party/nxp/nxp_matter_support"
+}
+
+declare_args() {
```
andy31415, line 21: same here: could we merge into one declare args?

build_overrides/nxp_sdk.gni
chapongatien: @andy31415  I got a GN error which recommend to separate them in 2 declare_args (as below):

2024-06-20 09:40:44 INFO    ERROR at //build_overrides/nxp_sdk.gni:19:27: Reading a variable defined in the same declare_args() call.
2024-06-20 09:40:44 INFO    
2024-06-20 09:40:44 INFO    If you need to set the value of one arg based on another, put
2024-06-20 09:40:44 INFO    them in two separate declare_args() calls, one after the other.

Do you have other recommendation to fix this error ?

scripts/setup/nxp/update_nxp_sdk.py
```diff
@@ -0,0 +1,96 @@
...
+all_platform_sdk_list = list(map(lambda plat: plat["plat_name"], ALL_PLATFORM_SDK))
+
+
+def clean_sdk(sdk_target_location):
+    print("SDK will be cleaned all local modification(s) will be lost")
+    # Cleaning all local modifications
+    git_clean_command = "git reset --hard && git clean -xdf"
+    command = ['west', 'forall', '-c', git_clean_command, '-a']
+    subprocess.run(command, cwd=sdk_target_location, check=True)
+    # Retrying an update
```
andy31415, line 39: why is this a "retry"?  It would be nice for the code to explain why things are done - I can see it does an update_platform_sdk_version, but not as clear why.

What does `False` mean?

andy31415 (approved):

scripts/setup/nxp/update_nxp_sdk.py
chapongatien: I agree, it is not very clear and the location of the retry is not correct. I re-worked this part my latest commit and moved this "retry" mechanism in a better place.  Please let me know if you are ok.

scripts/setup/nxp/update_nxp_sdk.py
```diff
@@ -0,0 +1,98 @@
...
+import argparse
+import os
+import subprocess
+import sys
+
+CHIP_ROOT = os.path.abspath(os.path.join(os.path.dirname(__file__), '../../..'))
+
+# Array listing all SDK and default targeted path for NXP SDK storage
+ALL_PLATFORM_SDK = [
+    {'plat_name': 'k32w0', 'plat_sdk_manifest_path': 'third_party/nxp/nxp_matter_support/github_sdk/k32w0'},
```
andy31415, line 27: Consider replacing stringly-typed data with something like `@dataclass` or similar. 

a `[p.name for p in ALL_PLATFORM_SDK]` is more readable than the existing `list(map(lambda p: p['plat_name'], ALL_PLATFORM_SDK))`

scripts/setup/nxp/update_nxp_sdk.py
```diff
@@ -0,0 +1,98 @@
...
+        "--update-only", help="Update NXP SDK to the correct version. Would fail if the SDK does not exist", action="store_true")
+    parser.add_argument('--platform', nargs='+', choices=all_platform_sdk_list, default=all_platform_sdk_list,
+                        help='Allows to select which SDK for a particular NXP platform to initialize')
+    parser.add_argument('--force', action='store_true',
+                        help='Force SDK initialization, hard clean will be done in case of failure - WARNING -- All local SDK modification(s) will be lost')
+
+    args = parser.parse_args()
+
+    for current_plat in args.platform:
+        plat_path = list(map(lambda res: res['plat_sdk_manifest_path'], filter(
```
andy31415, line 89: this should probably read like:

```py
plat_path = [p.manifest_path for p in ALL_PLATFORM_SDK if p.name == current_plat][0]
```

- use dataclasses
- using `[]` seems more readable than list/map/lambda

scripts/setup/nxp/update_nxp_sdk.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def init_platform_sdk_version(sdk_target_location, force):
+    print("Init SDK in: " + sdk_target_location)
+    if not force and os.path.exists(os.path.join(sdk_target_location, '.west')):
+        print("Error SDK is already initialized, use --force to force init")
+        print("WARNING -- All local SDK modification(s) will be lost")
+        sys.exit(1)
+    command = ['rm', '-rf', '.west']
+    subprocess.run(command, cwd=sdk_target_location, check=True)
```
andy31415, line 48:
```suggestion
    west_path = os.path.join(sdk_target_location, '.west')
    if  os.path.exists(west_path):
        if not force:
            print("Error SDK is already initialized, use --force to force init")
            print("WARNING -- All local SDK modification(s) will be lost")
            sys.exit(1)
        command = ['rm', '-rf', west_path]
        subprocess.run(command, check=True)
```

- seems to make it clearer and avoid running the remove if path does not exist.
- makes it clear you are deleting the exact path that is being checked for
- try to avoid having subprocess depend on cwd if possible

scripts/setup/nxp/update_nxp_sdk.py
```diff
@@ -0,0 +1,98 @@
...
+        subprocess.run(command, cwd=sdk_target_location, check=True)
+    except (RuntimeError, subprocess.CalledProcessError) as exception:
+        if force:
+            # In case of force update option and in case of update failure:
+            # 1. try to clean all local modications if any
+            # 2. Retry the west update command. It should be successfull now as all local modifications have been cleaned
+            clean_sdk_local_changes(sdk_target_location)
+            subprocess.run(command, cwd=sdk_target_location, check=True)
+        else:
+            print(exception)
```
andy31415, line 71: consider using python logging throughout. Then you can use `logging.exception` to report the exception with a full stack trace and info. otherwise probably traceback would be useful.

scripts/setup/nxp/update_nxp_sdk.py
chapongatien: I did an update in the last commit to use dataclass. Thanks for your recommendation.

scripts/setup/nxp/update_nxp_sdk.py
chapongatien: Yes agree, I switched to python logging in the last commit.

PR #33989, opened 2024-06-18
[Python] Convert async API functions to python asyncio

agners:
Convert API functions which are waiting for a callback to Python co-routines (async functions). This allows to use this API functions without blocking the asyncio event loop, which is crucial for a responsive asyncio application.

The change takes care of locking the individual functions so that independent asyncio tasks can't interfere with each other: E.g. if two asyncio tasks try to commission a device, the second asyncio task trying to commission too will wait for the first asyncio task to complete commission first.

### DIFF
--- a/scripts/tests/chiptest/yamltest_with_chip_repl_tester.py
+++ b/scripts/tests/chiptest/yamltest_with_chip_repl_tester.py
@@ -153,6 +153,4 @@ async def main(setup_code, yaml_path, node_id, pics_file):
 
 
 if __name__ == '__main__':
-    loop = asyncio.get_event_loop()
-    loop.run_until_complete(main())
-    loop.close()
+    asyncio.run(main())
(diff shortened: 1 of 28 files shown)


src/controller/python/chip/ChipDeviceCtrl.py
```diff
@@ -226,6 +226,41 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
+class CallbackContext:
```
andy31415, line 229: some documentations on purpose and usage would be nice both for this one and the CommissioningContext below.

andy31415 (approved):

scripts/tests/chiptest/yamltest_with_chip_repl_tester.py
```diff
@@ -153,4 +153,6 @@ def _StackShutDown():
 
 
 if __name__ == '__main__':
-    main()
+    loop = asyncio.get_event_loop()
```
arkq, line 156: Why not simple `asyncio.run(main())` ?

arkq (approved):

PR #33990, opened 2024-06-18
[MWOCTRL] Add error checks MWOCTRL test scripts

rbultman:
This PR updates the MWOCTRL per updates to the TP.  It also moves the 2.5 test case from YAML to python.

Ref TP PR: https://github.com/CHIP-Specifications/chip-test-plans/pull/4095


andy31415:
Closing stale PR.

PR #34000, opened 2024-06-18
Run zap_convert_all on the tree.

bzbarsky-apple:
We have had XML changes that did not get reflected into .zap files, unfortunately, including marking things as no longer provisional.

Fixing that by running zap_convert_all.py discovered a problem in the Thread Border Router Management cluster XML: it was using the wrong define value.  And apparently the .zap files involved had been hand-edited to not match the XML to allow code to compile, but regenerating them obviously "broke" that.  The fix is to use the right value in the XML.


kiel-apple (approved):
rubberstamp (generated code)

jamesharrow (approved):

andy31415 (approved):

PR #34001, opened 2024-06-18
[Python] Reset chip error in test commissioner

agners:
Make sure to reset the chip error in test commissioner on reset. This avoid spurious errors. The Python side reads the error as soon as mTestCommissionerUsed is set, which happens unconditionally. Hence clearing the error is necessary.


agners:
I've noticed this when updating tests for #33905. With #33905 the `TC_CGEN_2_4.py` test explicitly catches/expects the spurious error (since with #33905 a exception is raised, it requires handling).

Now with the spurious error gone, we can essentially revert that exception handling again.

arkq (approved):

andy31415 (approved):

PR #34004, opened 2024-06-19
Add a String Echo Request and Echo Response Command to the Unit testing cluster

pidarped:
The Command Request takes a string as argument that is echoed back in the Command response.
The string can be large to enable Command testing for large payloads over a TCP-based session.

Example ChipTool invocation:
./chip-tool unittesting string-echo-request "aaaaaaaaaaaaa" <node-id> <endpoint-id>

./chip-tool unittesting string-echo-request "aaaaaaa...aaaaaa" <node-id> <endpoint-id> --allow-large-payload true


src/app/zap-templates/zcl/data-model/chip/test-cluster.xml
```diff
@@ -472,6 +472,16 @@ limitations under the License.
       <arg name="arg1" type="int8u"/>
     </command>
 
+    <command source="client" code="0x18" name="StringEchoRequest"
+             response="StringEchoResponse" optional="true">
+      <description>
+        Command that takes an argument which is an octet string.  The response echoes
+        the string back. If the string is large then it would require a session that
```
yufengwangca, line 479: the command code should be consecutive, move this definition after TestSecondBatchHelperRequest (code="0x17")

yufengwangca (approved):

yufengwangca:
If we can split the generate portion as a separate commit, it will help the review a lot.

bzbarsky-apple (approved):

yunhanw-google (approved):

src/app/zap-templates/zcl/data-model/chip/test-cluster.xml
pidarped: Done.

PR #34025, opened 2024-06-20
Update builds to docker version 65

s07641069:
Change overview

- Update builds to latest docker version

https://github.com/project-chip/connectedhomeip/pull/33637


andy31415:
@s07641069 I think we have to update this gradually ... we switched docker to 24.04 and still need to fix some individual failures.

Damian-Nordic (approved):

bzbarsky-apple (approved):

s07641069:
> @s07641069 I think we have to update this gradually ... we switched docker to 24.04 and still need to fix some individual failures.

We can use this PR as parent to check and switch all build into latest version as soon as all vendors will apply their fixes.
Will be good to mention all vendor responsible persons which builds fails with ubuntu 24.04.

s07641069:
https://github.com/project-chip/connectedhomeip/pull/33637

jlatusek:
Hi, I've created a similar PR #33979. I had a problem with the current versions of the python packages so I've updated them. As far as I know, EFR32 still needs updates in the SDK to make it work on images newer than 54.

kkasperczyk-no:
I think that updating docker version to 65 will result in pulling in the new nRF Connect SDK 2.7.0 version of nrfconnect platform. This change requires a big number of alignments in nrfconnect platform and examples. I pushed a PR that addresses it and updates all workflows to use docker version 65: https://github.com/project-chip/connectedhomeip/pull/34403. @s07641069 do you think we could agree to close this PR and pull docker version update in my PR?

kkasperczyk-no:
> I think that updating docker version to 65 will result in pulling in the new nRF Connect SDK 2.7.0 version of nrfconnect platform. This change requires a big number of alignments in nrfconnect platform and examples. I pushed a PR that addresses it and updates all workflows to use docker version 65: #34403. @s07641069 do you think we could agree to close this PR and pull docker version update in my PR?

After having a talk with @andy31415 it seems it will be much more reasonable to update only docker version on nrf platform in my PR, so I will remove updating docker image for other platforms.

PR #34036, opened 2024-06-20
Updated AppContext, LoopbackMessagingContext, and UDPMessagingContext to serve as PW test fixture classes.

feasel0:
AppContext, LoopbackMessagingContext, and UDPMessagingContext have member functions that resemble the setup/teardown functions of pigweed (SetUp, TearDown, SetUpTestSuite, TearDownTestSuite), however we are not currently using these classes as actual pigweed test fixture classes.  Instead each unit test is creating a test fixture class that contains a _pointer_ to AppContext (or LoopbackMessagingContext, or UDPMessagingContext), and then it dynamically creates the object when the test runs, and deletes it when it's finished.  This leads to a lot of redundant code, since each unit test has to do this.

This PR modifies LoopbackMessagingContext and UDPMessagingContext so that they inherit from `::testing::Test`.  Since AppContext inherits from LoopbackMessagingContext, it too can be used as a pigweed test fixture class.

This PR also modifies all the unit tests that are dynamically creating AppContext, LoopbackMessagingContext, or UDPMessagingContext objects so that they instead inherit from the class directly.  This leads to a significant reduction of code, and in many cases the unit test doesn't even need to define a fixture class at all (it can just do `using TestSomething = AppContext;`, or whichever context class it's using).

To address the issue of the memory footprint, this PR also modifies [MessagingContext](https://github.com/feasel0/connectedhomeip/blob/d43b087f809263601e0a86c176276dda59d64ec7/src/messaging/tests/MessagingContext.h#L212) and [LoopbackMessagingContext](https://github.com/feasel0/connectedhomeip/blob/d43b087f809263601e0a86c176276dda59d64ec7/src/messaging/tests/MessagingContext.h#L267) so that they store their member variables on the heap, since this was a concern that may have been preventing us from adopting this method earlier.

Here is a summary of what these classes do.
* `::testing::Test` is what each pigweed unit test (with a fixture) must inherit.  The goal is to move this inheritance somewhere else in order to simplify the unit tests.
* MessagingContext is a test helper class that creates Alice, Bob, Charlie, David which can be used for testing messaging.
* LoopbackMessagingContext and UDPMessagingContext both inherit from MessagingContext.  They also add a transport manager (either LoopbackTransportManager or UDPTransportManager) and expose some of its functions.
* AppContext inherits from LoopbackMessagingContext, and adds some [additional setup/teardown logic](https://github.com/project-chip/connectedhomeip/blob/d43b087f809263601e0a86c176276dda59d64ec7/src/app/tests/AppTestContext.cpp#L41) that some unit tests need.

Right now we have 25 test sources that use AppContext (in `src/app/tests` and `src/controller/tests`), we have 8 test sources that use LoopbackMessagingContext (in `src/messaging/tests` and `src/protocols/secure_channel/tests`), and we have 1 test source that uses UDPMessagingContext (in `src/messaging/tests`).


jlatusek (changes requested):

src/messaging/tests/MessagingContext.h
```diff
@@ -92,16 +93,12 @@ class MessagingContext : public PlatformMemoryUser
...
     static constexpr System::Clock::Timeout kResponsiveActiveRetransTimeout = System::Clock::Milliseconds32(10);
 
-    MessagingContext() :
-        mInitialized(false), mAliceAddress(Transport::PeerAddress::UDP(GetAddress(), CHIP_PORT + 1)),
-        mBobAddress(LoopbackTransport::LoopbackPeer(mAliceAddress))
-    {}
-    // TODO Replace VerifyOrDie with Pigweed assert after transition app/tests to Pigweed.
-    // TODO Currently src/app/icd/server/tests is using MessagingConetext as dependency.
-    ~MessagingContext() { VerifyOrDie(mInitialized == false); }
+    MessagingContext() : mpData(new MessagingContextData()) {}
```
andy31415, line 96: since you new/delete automatically, would it make sense to use a std::unique_ptr so that we do not need to track down lifetime?

src/messaging/tests/MessagingContext.h
```diff
@@ -92,16 +93,12 @@ class MessagingContext : public PlatformMemoryUser
...
+    // TODO Currently src/app/icd/server/tests is using MessagingContext as dependency.
```
andy31415, line 97: Please explain the todo: what is to be done? The TODO seems to only be a statement of "this is what is going on". Please rephrase it to describe what needs to be done and ideally reference a github issue describing the work (that someone can pick up and then later mark as fixed).

src/messaging/tests/MessagingContext.h
andy31415: Why is mpData a pointer rather than just a member if it is allocated in new and freed up in the destructor? Is it stack space saving (use heap for large objects instead of stack)? If yes, this should be in a comment somewhere.

src/messaging/tests/MessagingContext.h
```diff
@@ -166,135 +163,155 @@ class MessagingContext : public PlatformMemoryUser
...
+    struct MessagingContextData
+    {
+        MessagingContextData() :
+            mInitialized(false), mAliceAddress(Transport::PeerAddress::UDP(GetAddress(), CHIP_PORT + 1)),
+            mBobAddress(LoopbackTransport::LoopbackPeer(mAliceAddress))
+        {}
+        ~MessagingContextData() { EXPECT_FALSE(mInitialized); }
+
+        bool mInitializeNodes = true;
+        bool mInitialized;
```
andy31415, line 187: you initialize a lot of other things as `= ...`, however this one you initialize in the constructor. Why?

How about ` = false` on this one and also initialize all the pointers below to nullptr to have some initial state?

andy31415 (approved):
Approving assuming remaining comments will be addressed.

src/messaging/tests/MessagingContext.h
feasel0: The comment was from a previous developer.  I left it in the code since I didn't really understand the point of it, but that's probably a reason to remove it rather than leaving it.  Deleted.

src/messaging/tests/MessagingContext.h
feasel0: Done.  Again, this was just my reluctance to change how the previous author had done things.  But that's not a good excuse.

src/messaging/tests/MessagingContext.h
feasel0: I replaced this with std:unique_ptr.

Yup, the reason for making mpData a pointer is to put it on the heap and save stack space.  I have added a comment to explain this.

jlatusek (approved):

PR #34037, opened 2024-06-20
Update Matter Casting automation test script to use the simplified Linux tv-casting-app build command in the workflow yaml file.

shaoltan-amazon:
**Problem**
The current [Matter Casting automation test script](https://github.com/project-chip/connectedhomeip/blob/master/scripts/tests/run_tv_casting_test.py) is configured for the old [Linux tv-casting-app build as defined in the workflow yaml file](https://github.com/project-chip/connectedhomeip/blob/9d56786ed093473fdfb66d770a2e2338ab579d3e/.github/workflows/examples-linux-tv-casting-app.yaml#L64). We need to update the yaml file to make the test script work for the new simplified build of the Linux tv-casting-app. Additionally, all the strings used in the `commissionee_generated_passcode_test` sequence are currently from the core SDK. We need to substitute these strings with their equivalents from the tv-casting library or tv-casting-app, where feasible. This way, the test script would be less prone to failure in the event of string format changes in the core SDK. Lastly, improvements could be made to the `send_input_cmd_to_subprocess` function along with the way we are removing cached entries. 

**Solution**
- Added `chip_casting_simplified` support to `build_examples.py` by changing `host.py` to support this flag and add a variant that is conditional to `targets.py`.
- Updated the Linux tv-casting-app build command from `scripts/examples/gn_build_example.sh examples/tv-casting-app/linux/ out/tv-casting-app` to `scripts/examples/gn_build_example.sh examples/tv-casting-app/linux/ out/tv-casting-app chip_casting_simplified=true`. 
- Updated the core SDK strings from the test sequence steps to use strings from the tv-casting library or tv-casting-app where possible. 
- Simplified the `send_input_cmd_to_subprocess` function by removing unnecessary code. 
- Changed to use `os.remove` instead of `os.system` as the former is more aligned with Python best practices. 

**Testing**
Tested locally to ensure the test script works as expected.


.github/workflows/examples-linux-tv-casting-app.yaml
```diff
@@ -61,7 +61,7 @@ jobs:
             - name: Build Linux tv-casting-app
               run: |
                   ./scripts/run_in_build_env.sh \
-                    "scripts/examples/gn_build_example.sh examples/tv-casting-app/linux/ out/tv-casting-app"
+                    "scripts/examples/gn_build_example.sh examples/tv-casting-app/linux/ out/tv-casting-app chip_casting_simplified=true"
```
andy31415, line 64: please add chip_casting_simplified support to build_examples.py as well (change host.py to support this flag, add a variant that is conditional to targets.py)

scripts/tests/run_tv_casting_test.py
```diff
@@ -63,6 +64,21 @@ def __exit__(self, exception_type, exception_value, traceback):
...
+    """Remove any cached files that match the provided pattern."""
+
+    cached_files = glob.glob(cached_file_pattern)  # Returns a list of paths that match the pattern.
+
+    for cached_file in cached_files:
+        try:
+            os.remove(cached_file)
+        except OSError as e:
+            logging.error(f'Failed to remove cached file `{cached_file}` with error: `{e.strerror}`.')
+            return False
```
andy31415, line 77: the flow seems to be: call this bool function, which does a printout of the exception and the caller exits(1).

Why not directly throw the exception and not make this return a bool? If you want to handle the exception you can catch it, however turning an exception into a bool and then using it to sys exit with error feels like just extra complexity.

If you want to add file name context, you can re-throw the exception with more context, however it should still be an exception.

sharadb-amazon (approved):

scripts/tests/run_tv_casting_test.py
shaoltan-amazon: Thank you for the feedback, I have addressed this in the latest commit.

.github/workflows/examples-linux-tv-casting-app.yaml
shaoltan-amazon: I have added this in the latest commit. Since I am not too familiar with this, please re-open this thread if I have missed anything. I have tested it by running `./scripts/build/build_examples.py --target darwin-arm64-tv-casting-app-chip-casting-simplified build`

scripts/build/builders/host.py
```diff
@@ -428,6 +429,10 @@ def __init__(self, root, runner, app: HostApp, board=HostBoard.NATIVE,
             else:
                 self.extra_gn_options.append('chip_enable_dnssd_tests=false')
 
+        if chip_casting_simplified is not None:
+            if chip_casting_simplified:
+                self.extra_gn_options.append('chip_casting_simplified=true')
```
andy31415, line 434:
```suggestion
        if chip_casting_simplified is not None:
               self.extra_gn_options.append(f'chip_casting_simplified={str(chip_casting_simplified).lower()}')
```

since otherwise you should use true/false. If you want to support none, make sue of all of None/true/false

shaoltan-amazon: Noted. Made the change in the latest commit.

scripts/build/builders/host.py
```diff
@@ -313,7 +313,8 @@ def __init__(self, root, runner, app: HostApp, board=HostBoard.NATIVE,
                  use_coverage=False, use_dmalloc=False, minmdns_address_policy=None,
                  minmdns_high_verbosity=False, imgui_ui=False, crypto_library: HostCryptoLibrary = None,
                  enable_test_event_triggers=None,
-                 enable_dnssd_tests: Optional[bool] = None
+                 enable_dnssd_tests: Optional[bool] = None,
+                 chip_casting_simplified=None
```
andy31415, line 317:
```suggestion
                 chip_casting_simplified: Optional[bool] = None
```

shaoltan-amazon: Good catch! Made the change in the latest commit.

chrisdecenzo (approved):

andy31415 (approved):

PR #34042, opened 2024-06-21
Second set of test scripts to have automated implementation of python tests to use Metadata script as an intermediate argument holder

vatsalghelani-csa:
Work for Issue #31317 

Follow up to PR #34032 

PR #7 - Second set of automated implementation of python tests to use Metadata script as an intermediate argument holder. Modified second set of python test script to now use the newer approach of extracting the arguments required through the new created metadata script and use it for execution.
Files Worked:
- .github/workflows/tests.yaml
- src/python_testing/*
  - 16 python test files


src/python_testing/TestConformanceSupport.py
```diff
@@ -15,6 +15,9 @@
 #    limitations under the License.
 #
 
+# test-runner-runs: run1
+# test-runner-run/run1/script-args: --trace-to json:${TRACE_TEST_JSON}.json --trace-to perfetto:${TRACE_TEST_PERFETTO}.perfetto
```
andy31415, line 19: This one is odd: why do we even run it as a test runner if it has no "app" to test at all? it seems like a self-test.

I believe we should separate these out somehow. Please add an issue for this and we can talk to @cecille about this.

vatsalghelani-csa: Fixed.

src/python_testing/TestMatterTestingSupport.py
```diff
@@ -15,6 +15,9 @@
 #    limitations under the License.
 #
 
+# test-runner-runs: run1
+# test-runner-run/run1/script-args: --trace-to json:${TRACE_TEST_JSON}.json --trace-to perfetto:${TRACE_TEST_PERFETTO}.perfetto
+
```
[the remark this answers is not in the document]
vatsalghelani-csa: Fixed.

andy31415 (approved):

PR #34044, opened 2024-06-24
[ESP32] Remove SupportedModes implementation of ModeSelect from nvs and use all-clusters-apps common implementation

jadhavrohit924:
- Remove SupportedModes implementation of ModeSelect from nvs and use all-clusters-apps common implementation.
- Tested esp32/all-clusters-app


jmartinez-silabs (approved):

dhrishi:
Can you try any other example (lighting-app) as well?

dhrishi (approved):

andy31415:
@jadhavrohit924 This PR does not seem to explain WHY the implementation was removed.

I imagine common code is better than specific code, however it is still useful to actually explain changes. For example I do not understand why we got #34191 as a revert now. The history of this seems to be just "tested and it works fine" and it does not explain the intent behind the change (and revert is not explained either)

jadhavrohit924:
@andy31415 Sorry for the inconvenience. Actually, we thought of using common implementation and planned to remove this fixed attributes from the factory partition because of while OTA, we cannot update these attributes as factory partition can’t be change in OTA.

But later we thought of supporting this fixed attributes in the factory partition for backward compatibility along with SDKs common implementation which has static data.

PR #34054, opened 2024-06-24
[CI] macos python update

jlatusek:
## Problem

Currently used version of python for dawin platform is 3.9. In PR #33979 related to update ubuntu based image to 24.04 (python 3.12), there is a problem with updating pip packages. Updating pip packages is required due to the removal of some deprecated APIs in Python 3.12 - some packages cannot be installed without them being updated. 

## Changes

1. I've updated version of python on macos runner and the one which is setted up by CIPD. Unforunetley for CIPD I can't find newer version than 3.11
2. I've added a script which will check currenlty avaliable version of python to not hadcode one in gn files.


andy31415 (approved):

arkq (approved):

.github/workflows/build.yaml
```diff
@@ -372,6 +372,7 @@ jobs:
     build_darwin:
         name: Build on Darwin (clang, python_lib, simulated)
         runs-on: macos-13
+        python-version: '3.11'
```
bzbarsky-apple, line 375: What about non-CI builds?  Python 3.9 is the system python on MacOS.  If our builds actually require a newer python, presumably our build documentation needs to be updated to make this clear; otherwise people following it will not be able to actually build Matter.

So more precisely:

1. We need build documentation updates.
2. We need testing to make sure that following the updated steps actually works (e.g. building Matter against homemebrew-installed Python 3.11 on Mac did not work last time I tested it, but maybe something has changed since then.). I'm happy to do that testing once there are updated steps in this PR.

bzbarsky-apple (changes requested):

.github/workflows/build.yaml
jlatusek: I've added information about installation of python@3.11 using brew. I tested it locally with this version of python and it seems to work, but I'm not daily user of macos, I would be gratefull if you can check that out.

docs/guides/BUILDING.md
```diff
@@ -107,7 +107,21 @@ sudo apt-get install libsdl2-dev
...
+
+```sh
+    brew install python@3.11
+    python3 -m pip install --upgrade setuptools
+    python3 -m pip install --upgrade pip
+    cd /usr/local/bin
+    ln -sf python3.11 python3
+    ln -sf python3 python
+    ln -sf pip3.11 pip3
+    ln -sf pip3 pip
```
bzbarsky-apple, line 121: This doesn't quite make sense.  Why do we need to link things into /usr/local/bin but we already ran python3 before that and seemed to assume it's the "right" python3?

Also, is this step only needed to do pycontroller?  Or will other builds and/or bootstrap also fail without this newer Python version?

.github/workflows/build.yaml
bzbarsky-apple: OK, so I ran the following tests:

1. Try to bootstrap (on an ARM Macbook Pro) the existing SDK without this PR against "Python 3.11.4" which I have installed from homebrew (note, this might not be the most recent `python@3.11`; if that needs testing I can try to test it).  This fails as follows:
   ```
   Installing pip requirements for all...
     error: subprocess-exited-with-error
     
     × python setup.py bdist_wheel did not run successfully.
     │ exit code: 1
     ╰─> [27 lines of output]
         running bdist_wheel
         running build
         running build_py
         creating build
         creating build/lib.macosx-13-arm64-cpython-311
         creating build/lib.macosx-13-arm64-cpython-311/bitarray
         copying bitarray/util.py -> build/lib.macosx-13-arm64-cpython-311/bitarray
         copying bitarray/__init__.py -> build/lib.macosx-13-arm64-cpython-311/bitarray
         copying bitarray/test_bitarray.py -> build/lib.macosx-13-arm64-cpython-311/bitarray
         copying bitarray/test_util.py -> build/lib.macosx-13-arm64-cpython-311/bitarray
         copying bitarray/bitarray.h -> build/lib.macosx-13-arm64-cpython-311/bitarray
         copying bitarray/pythoncapi_compat.h -> build/lib.macosx-13-arm64-cpython-311/bitarray
         copying bitarray/test_data.pickle -> build/lib.macosx-13-arm64-cpython-311/bitarray
         copying bitarray/py.typed -> build/lib.macosx-13-arm64-cpython-311/bitarray
         copying bitarray/__init__.pyi -> build/lib.macosx-13-arm64-cpython-311/bitarray
         copying bitarray/util.pyi -> build/lib.macosx-13-arm64-cpython-311/bitarray
         running build_ext
         building 'bitarray._bitarray' extension
         creating build/temp.macosx-13-arm64-cpython-311
         creating build/temp.macosx-13-arm64-cpython-311/bitarray
         clang -Wsign-compare -Wunreachable-code -fno-common -dynamic -DNDEBUG -g -fwrapv -O3 -Wall -I/private/tmp/connectedhomeip/.environment/pigweed-venv/include -I/opt/homebrew/opt/python@3.11/Frameworks/Python.framework/Versions/3.11/include/python3.11 -c bitarray/_bitarray.c -o build/temp.macosx-13-arm64-cpython-311/bitarray/_bitarray.o
         In file included from bitarray/_bitarray.c:12:
         /opt/homebrew/opt/python@3.11/Frameworks/Python.framework/Versions/3.11/include/python3.11/Python.h:23:12: fatal error: 'stdlib.h' file not found
            23 | #  include <stdlib.h>
               |            ^~~~~~~~~~
         1 error generated.
         error: command '/private/tmp/connectedhomeip/.environment/cipd/packages/pigweed/bin/clang' failed with exit code 1
         [end of output]
     
     note: This error originates from a subprocess, and is likely not a problem with pip.
     ERROR: Failed building wheel for bitarray
   ERROR: Could not build wheels for bitarray, which is required to install pyproject.toml-based projects
   ```
2. Apply this PR, try to bootstrap against system Python 3.9.6.  This seems to succeed.
3. Apply this PR, try to bootstrap against "Python 3.11.4" which I have installed from homebrew, as in step 1.  This fails in the same way as step 1.

So some questions:

* When you tested and things worked for you, were you using an Intel-CPU Mac, or an ARM one?  The above compile issue is a known problem, on ARM only.
* If you were using an ARM one, what version of python do you have from Homebrew?

Note that I did not run the `pip install` step from the .md in the PR, but I don't think those would matter for the above issue....

bzbarsky-apple: OK, I tried `brew install python@3.11`, but it seems to have installed "Python 3.12.4".  That fails in the same way as above during bootstrap, trying to compile things and failing on missing `stdlib.h`.

jlatusek: I'm using an Intel-CPU Mac. I wasn't aware of a problems on arm base one.

bzbarsky-apple: OK.  So given that after this PR I can still bootstrap fine with Python 3.9... do we really need to change any of the build instructions or CI bits?  Where is Python 3.11 really needed?  Or do we have other changes that we have not made yet but plan to that will in fact require Python 3.11 or later?

jlatusek: Latest version of pigweed requires python 3.10

bzbarsky-apple: So is this blocking a pigweed update?  Would it be possible to create a PR that has both this change and the relevant pigweed changes, so I could test what happens in that situation?

In either case, we need to figure out how this is supposed to work if bootstrap does not work on ARM Darwin with Python 3.11....  but maybe the newer pigweed would fix that.

docs/guides/BUILDING.md
jlatusek: In #33979 is an update of all CI, pigweed and python packages

.github/workflows/build.yaml
bzbarsky-apple: Thank you.

So if I apply the changes in #33979 and try my tests above:

* Bootstrapping with homebrew "Python 3.12.4" seems to work on an ARM Mac (great, it looks like something got fixed on the pigweed side here in the new pigweed?)
* Bootstrapping with the system "Python 3.9.6" fails with syntax errors on `module: str | None,`; this is presumably expected?

So in terms of landing sequencing, it sounds like we want to update the CI bits, but make the build documentation changes in the same PR that updates pigweed to a version that is not broken on ARM when using the newer Python?

I'm still investigating what the impact is of requiring a non-system Python for pigweed is on Mac in general, but hope to have some answers on this soon.

bzbarsky-apple: Looks like as long as you're not trying to run our bootstrap script nothing cares about the Python version, so we're good on that front at least, @jlatusek .

docs/guides/BUILDING.md
jlatusek: I inserted these commands in the wrong order. My goal was to change the default version of python used in the shell to python3.11. Now I think that doing it this way will change the settings not only for our project but for the whole system. How should I pass information to use a specific version of Python during bootstrapping?

.github/workflows/build.yaml
jlatusek: Ok, as I've checked on a Ubuntu-based machine, we need to upgrade python, pip packages and the pigweed submodule all at once to get everything working.

docs/guides/BUILDING.md
bzbarsky-apple: I don't know how to tell bootstrap to use a specific Python, short of messing with PATH....

Note that installing via brew will likely end up using that Python for all sorts of things, since the brew stuff will come first in PATH.

.github/workflows/build.yaml
jlatusek: @bzbarsky-apple on an ARM based mac does it only work with python3.12 or also with 3.11? I was using python 3.11 because it is the latest version available in CIPD. As I found out, the CIPD version is only installed on the Intel base platform. Do we still need this instance of python installed by CIPD? Maybe we can use brew python everywhere?

The main problem with updating Python to 3.12 is that we have to do it on all platforms at once. The problem is different versions of packages that need to be installed.

bzbarsky-apple: > on an ARM based mac does it only work with python3.12 or also with 3.11? 

I would guess both, but I don't have an obvious way to test; when I tried `brew install python@3.11` it installed 3.12....

> Maybe we can use brew python everywhere?

You mean across both intel and ARM Mac?  It would be good to have them have the same flow, yes.

docs/guides/BUILDING.md
Alami-Amine: Hello @bzbarsky-apple, @jlatusek 
are we blocked here? any change we can continue with the upgrade of the stack to python 3.11 ?

bzbarsky-apple: It needs to be done in sync with the upgrade to pigweed, because the version of pigweed we are using right now cannot bootstrap on an ARM Mac with python 3.11; see conversation above.

PR #34070, opened 2024-06-25
Unify stdio logging

### LINKED ISSUE #24701: [Tooling] No timestamps in cross-compiled chip-tool for arm64

When cross-compiling chip-tool for arm64 from x64 host with command:
`./scripts/build/build_examples.py --target linux-arm64-chip-tool-ipv6only-clang build`
the wrong logging implementation is selected and there are no logging timestamps.

### Platform

raspi

### Platform Version(s)

_No response_

### Type

CI tested

Binary is produced with:
./scripts/build/build_examples.py --target linux-arm64-chip-tool-ipv6only-clang build

### Anything else?

_No response_

mbknust:
Darwin now uses the same stdio logging implementation as other platforms.
Logs are colorized and contain timestamps.
On linux and darwin, pid and tid are also logged.

Fixes #24701


ksperling-apple (approved):

src/platform/logging/impl/stdio/Logging.cpp
```diff
@@ -4,17 +4,64 @@
...
+        break;
+    case kLogCategory_Detail:
+        printf("\033[0;34m");
+        break;
+    }
+
+#if defined(__APPLE__) || defined(__gnu_linux__)
+    timespec ts;
+    timespec_get(&ts, TIME_UTC);
+    printf("[%lld.%06ld] ", static_cast<long long>(ts.tv_sec), static_cast<long>(ts.tv_nsec / 1000));
```
bzbarsky-apple, line 46: For the Apple case this is changing the log format from "ms" to "s.us", right?  That might be OK, but we'll need to make sure there's no tooling broken by that....

src/platform/logging/impl/stdio/Logging.cpp
```diff
@@ -4,17 +4,64 @@
...
+#if defined(__APPLE__)
+    uint64_t ktid;
+    pthread_threadid_np(nullptr, &ktid);
+    printf("[%lld:%lld] ", static_cast<long long>(getpid()), static_cast<long long>(ktid));
+#elif defined(__gnu_linux__) && !defined(__NuttX__)
+    // TODO: change to getpid() and gettid() after glib upgrade
+    printf("[%lld:%lld] ", static_cast<long long>(syscall(SYS_getpid)), static_cast<long long>(syscall(SYS_gettid)));
+#endif
+
     printf("CHIP:%s: ", module);
```
bzbarsky-apple, line 58: This is also changing the formatting in the Apple case.  Again, probably fine, but we'll need to watch out for issues in log processing tooling.

Why do we want to include "CHIP" in here?

src/platform/logging/impl/stdio/Logging.cpp
```diff
@@ -4,17 +4,64 @@
...
 
 namespace chip {
 namespace Logging {
 namespace Platform {
 
+// TODO: investigate whether pw_chrono or pw_log could be used here
 void LogV(const char * module, uint8_t category, const char * msg, va_list v)
 {
+#if defined(_POSIX_THREAD_SAFE_FUNCTIONS)
+    flockfile(stdout);
```
bzbarsky-apple, line 27: Worth documenting that this is done because we print our log in pieces and we don't want other prints ending up in the middle of the log string, right?

src/platform/logging/impl/stdio/Logging.cpp
ksperling-apple: Hm yeah do we really need that granularity of timestamps? It's probably just line noise in practice

src/platform/logging/impl/stdio/Logging.cpp
mbknust: Right, I've added a comment about that.

src/platform/logging/impl/stdio/Logging.cpp
mbknust: Ok, I changed it to seconds * 1000 + milliseconds, same as it was before

src/platform/logging/impl/stdio/Logging.cpp
mbknust: I had based my format string on the linux logger:

https://github.com/project-chip/connectedhomeip/blob/58536ba64da75f0a63c569baa29966008ffdf306/src/platform/Linux/Logging.cpp#L51-L52

But you're right, it's better to stay consistent with darwin. So, I changed it; now a log line looks like this:
```
[1719389130094] [1459444:1459446] [IN] TransportMgr initialized
```

arkq: > But you're right, it's better to stay consistent with darwin.

But now it's inconsistent with Linux :D It's not possible to keep backward compatibility in case of unification.

I agree that the `CHIP` prefix should go away. It could make sense in case of system logger where logs from many applications are redirected to a single file, but it's not the case for STDIO I think. But I'm not sure about the time format. The dot between seconds and the rest is really useful for humans. Also, millisecond precision will not show time changes in many cases (microsecond precision does).

bzbarsky-apple: I agree that unless we have more ifdefs one or the other log format will change.  I'm just saying we should watch out for issues getting filed because someone's tools broke, so we can tell them what the story is.

Regarding time format, in practice, I suspect that seconds.milliseconds is the most useful time format here, in terms of readability, if we posit that it's humans reading the output.  It probably does not matter in most cases how much time has passed between two log lines if it's < 1ms.

arkq (approved):

PR #34073, opened 2024-06-25
[Silabs]Update Submodules, API, gn files to use latest silabs sdks

jmartinez-silabs:
- Replace gecko sdk by simplicity sdk which is our new SDK
- Bump wifi sdk to 3.3.0
- Update matter support for new sdk support files and libs
- Update pulled files and configs in gn files structure
- Update some API that changed
- Use silabs new memory manager instead of freeRTOS heap4
- Use new silabs docker image in ci that utilise latest sdks

Tested various app combination builds, commissioning and control on EFR32MG24
CI will also test a series of combinations


third_party/silabs/efr32_sdk.gni
```diff
@@ -99,6 +96,9 @@ declare_args() {
   sl_ot_platform_abstraction =
       "${efr32_sdk_root}/protocol/openthread/platform-abstraction"
 
+  # Enable Synchronized Sleepy End Device
+  enable_synchronized_sed = use_silabs_thread_lib
```
mkardous-silabs, line 100: Why is this change necessary?

mkardous-silabs (approved):

third_party/silabs/efr32_sdk.gni
jmartinez-silabs: fixed

yufengwangca (approved):

PR #34086, opened 2024-06-26
[Silabs] Enabling sleepy for 917soc with power manager

chirag-silabs:
Enabling power manager sleep with the wifi sdk v3.3.0

917 SoC sleepy code was present in the examples/ which is not moved to the wifi sdk tickless mode. 
Cleanup of some of the sleep related stuff as well

Tested locally with BRD4338A


jmartinez-silabs (approved):

third_party/silabs/efr32_sdk.gni
```diff
@@ -480,6 +481,10 @@ template("efr32_sdk") {
           # TODO: Remove this flag, once the communication is fixed
           "WIFI_DEBUG_ENABLED=1",
         ]
+
+        # This is kept due to the warning on the LWIP when on demand timer is added
+        # TODO: remove this flag once the warning is fixed in SiSDK MATTER-3946
+        cflags += [ "-Wno-error" ]
```
jmartinez-silabs, line 487: Is there not a more targetted flag for the error you need to ignore temporarily? No-error seems way to general and will hide other major issues

jmartinez-silabs (changes requested):

third_party/silabs/efr32_sdk.gni
chirag-silabs: Yeah, modified to cflags_c and added the required warning of implicit function declaration, since it cannot be used in the cpp

jmartinez-silabs (approved):

andy31415 (approved):

PR #34087, opened 2024-06-26
[Silabs] Add support for factory and commissionable data provisioning with the use…

jmartinez-silabs:
… of libraries

- Source files in `examples/platform/silabs/provision` are more flexible and can be modified by users. Mostly ProvisionStorageCustom.cpp
- provisioning Libs are taken from our matter_support submodule
- All headers to link the libs are in `src/platform/silabs/provision`

**On-host tool to execute the factory provisioning is not yet available on csa - comming soon TM**


examples/platform/silabs/BaseApplication.cpp
```diff
@@ -24,20 +24,21 @@
 #include "AppConfig.h"
 #include "AppEvent.h"
 #include "AppTask.h"
-
 #include <app/server/Server.h>
 
+#define APP_ACTION_BUTTON 1
```
mkardous-silabs, line 29: We should move this to the platformAbstraction to avoid re-defining each time the GetButtonState is necessary.

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -24,20 +24,21 @@
...
 #ifdef QR_CODE_ENABLED
 #include "qrcodegen.h"
 #endif // QR_CODE_ENABLED
 #endif // DISPLAY_ENABLED
 
-#include "SilabsDeviceDataProvider.h"
 #if CHIP_CONFIG_ENABLE_ICD_SERVER == 1
 #include <app/icd/server/ICDNotifier.h> // nogncheck
 #endif
+#include <ProvisionManager.h>
```
mkardous-silabs, line 41:
```suggestion
#include <platform/silabs/ProvisionManager.h>
```
Just to be inline wiht the include structure

examples/platform/silabs/SiWx917/BUILD.gn
```diff
@@ -114,46 +113,9 @@ source_set("siwx917-matter-shell") {
...
-    "${chip_root}/src/credentials",
-    "${chip_root}/src/lib/support",
-    "${chip_root}/src/platform:platform_base",
-    "${chip_root}/src/setup_payload",
-  ]
-
-  public_configs = [ ":siwx917-common-config" ]
-
-  if (sl_enable_test_event_trigger) {
-    public_configs += [ ":test-event-trigger-config" ]
```
mkardous-silabs, line 151: This public config is still necessary. It should be moved to same spot as the headers.

examples/platform/silabs/SiWx917/BUILD.gn
```diff
@@ -114,46 +113,9 @@ source_set("siwx917-matter-shell") {
...
-  }
-}
-
 config("siwx917-common-config") {
   defines = [ "OTA_PERIODIC_TIMEOUT=${ota_periodic_query_timeout_sec}" ]
+  libs = [ "${sdk_support_root}/matter/provision/lib/libProvision_si917.a" ]
```
mkardous-silabs, line 118: I suppose the family variable doesn't match the si917 name? Should we name the lib si91x?

examples/platform/silabs/efr32/BUILD.gn
```diff
@@ -246,13 +212,15 @@ config("silabs-wifi-config") {
...
-  deps = []
+  deps = [ "${chip_root}/src/platform/silabs/provision:provision-headers" ]
   defines = []
   public_deps = []
   public_configs = [
     ":efr32-common-config",
     "${efr32_sdk_build_root}:silabs_config",
     ":chip_examples_project_config",
+
+    "${chip_root}/src/platform/silabs/provision:provision-config",
```
mkardous-silabs, line 223: we don't need this on the 917 side?
I don't think we needs this since it is a public config of the headers.

examples/platform/silabs/SiWx917/BUILD.gn
jmartinez-silabs: It is in the source set of `test-event-trigger` as a public config. it is not neeeded else where

examples/platform/silabs/SiWx917/BUILD.gn
jmartinez-silabs: Only 1 family currently support on matter 917. The family name is currently set to `siwx917-common` which I don't like. I will change the set family name to si917 in a subsequent pr and it will match with the lib name. If other wifi family are added in the future we will see what name it will have/how we support them

examples/platform/silabs/efr32/BUILD.gn
jmartinez-silabs: maybe, ill try it

examples/platform/silabs/BaseApplication.cpp
jmartinez-silabs: ok will add const in the SilabsPlatform class in a subsequent pr

examples/platform/silabs/BaseApplication.cpp
jmartinez-silabs: Not sure why, but it doesn't work. With the path the include throws an error. I reverted the change.

mkardous-silabs: That is really weird... In theory it should?

mkardous-silabs (approved):

bzbarsky-apple (approved):
Rubber-stamping platform-specific change with platform owner review.

PR #34088, opened 2024-06-26
ZAP regen on master

andy31415:
Fixes merge error that broke the zap check.


jmartinez-silabs (approved):

jmartinez-silabs:
some submodule sha changes. revert those

PR #34094, opened 2024-06-27
[Fix] Created possible fix for issue #229: Remove PICS from python PWRTL tests

j-ororke:
- Removed PICS if statements from python TC_PWRTL_2_1 test module
- Used read_single_attribute_check_success() to get attribute ID's for available and active endpoints then compared attribute ID's with attribute list to verify DUT's supports those attributes.
- Updated if statement checks as possible that empty lists might also be equivalent to NullValue when checking active and available endpoints on DUT's


src/python_testing/TC_PWRTL_2_1.py
```diff
@@ -40,36 +40,43 @@ async def test_TC_PWRTL_2_1(self):
 
         attributes = Clusters.PowerTopology.Attributes
 
-        endpoint = self.user_params.get("endpoint", 1)
+        endpoint = 1
```
cecille, line 43: Please leave as-is - testers need to be able to override this on the command line.

src/python_testing/TC_PWRTL_2_1.py
```diff
@@ -40,36 +40,43 @@ async def test_TC_PWRTL_2_1(self):
...
-        if available_endpoints == NullValue:
-            logging.info("AvailableEndpoints is null")
-        else:
-            logging.info("AvailableEndpoints: %s" % (available_endpoints))
+        if avail_endpoints_attr_id in attribute_list:
+            available_endpoints = await self.read_single_attribute_check_success(endpoint=endpoint, cluster=Clusters.Objects.PowerTopology, attribute=attributes.AvailableEndpoints)
 
-            asserts.assert_less_equal(len(available_endpoints), 21,
-                                      "AvailableEndpoints length %d must be less than 21!" % len(available_endpoints))
+            if available_endpoints == NullValue or available_endpoints == []:
```
cecille, line 57: This is a strange thing you've inherited here - AvailableEndpoints isn't nullable so the null check is pointless.

src/python_testing/TC_PWRTL_2_1.py
```diff
@@ -40,36 +40,43 @@ async def test_TC_PWRTL_2_1(self):
...
+                logging.info("AvailableEndpoints is null or an empty list")
+            else:
+                logging.info("AvailableEndpoints: %s" % (available_endpoints))
+                asserts.assert_less_equal(len(available_endpoints), 21,
+                                          "AvailableEndpoints length %d must be less than 21!" % len(available_endpoints))
```
cecille, line 62: I know this isn't your code, or test plan, but doesn't it feel like there's a step missing here? Shouldn't there be a check that the listed points actually exist? Do you know if that check is done elsewhere for this cluster?

src/python_testing/TC_PWRTL_2_1.py
```diff
@@ -40,36 +40,43 @@ async def test_TC_PWRTL_2_1(self):
...
+
+            if available_endpoints == NullValue or available_endpoints == []:
+                asserts.assert_true(active_endpoints == NullValue or active_endpoints == [],
+                                    "ActiveEndpoints should be null when AvailableEndpoints is null: %s" % active_endpoints)
 
-        if available_endpoints == NullValue:
-            asserts.assert_true(active_endpoints == NullValue,
-                                "ActiveEndpoints should be null when AvailableEndpoints is null: %s" % active_endpoints)
+        else:
+            logging.info('Skipping test as active endpoints attribute ID not in attribute list on DUT')
```
cecille, line 78: can you change this to say that it's the test step being skipped?

j-ororke: Hi Cecille,
I have updated to show that test step 3 has been skipped

src/python_testing/TC_PWRTL_2_1.py
j-ororke: Hi Cecille,
I also noticed this, I am not sure why there was a null check here when it seems it could only return an empty list if no AvailableEndpoints were found for some reason. 
I updated this to only if check for empty list currently.

src/python_testing/TC_PWRTL_2_1.py
j-ororke: Hi Cecille,
I have returned this to using the prior method for setting the endpoint variable.

src/python_testing/TC_PWRTL_2_1.py
j-ororke: Hi Cecille,
I also feel like there might be something missing here, but my understanding and knowledge here is limited. As this appears to be the only power topology clusters test currently, I don't believe there is another place that checks if available endpoints exist's. I think the closest thing might be the if check above for if AvailableEndpoints is an empty list. Should an additional check be added here to make sure that endpoints exist?

src/python_testing/TC_PWRTL_2_1.py
```diff
@@ -46,30 +46,45 @@ async def test_TC_PWRTL_2_1(self):
...
+            if available_endpoints == []:
+                logging.info("AvailableEndpoints is an empty list")
+            else:
+                logging.info("AvailableEndpoints: %s" % (available_endpoints))
+                asserts.assert_less_equal(len(available_endpoints), 20,
+                                          "AvailableEndpoints length %d must be less than 21!" % len(available_endpoints))
+
+        else:
+            logging.info('Skipping test as available endpoints attribute ID not in attribute list on DUT')
             return
```
cecille, line 70:
```suggestion
```

src/python_testing/TC_PWRTL_2_1.py
cecille: kk, opened an issue here - https://github.com/project-chip/matter-test-scripts/issues/369. I assigned it to you. I will need a minor test plan change. basically just adding a sentence in the expected column saying that all the endpoints are endpoints that exist on the device. And then that can just be added here. Not a high priority, but if you want to knock out a quick one, feel free to have a go at it.

src/python_testing/TC_PWRTL_2_1.py
```diff
@@ -46,30 +46,45 @@ async def test_TC_PWRTL_2_1(self):
...
+
+            if available_endpoints == []:
+                logging.info("AvailableEndpoints is an empty list")
+            else:
+                logging.info("AvailableEndpoints: %s" % (available_endpoints))
+                asserts.assert_less_equal(len(available_endpoints), 20,
+                                          "AvailableEndpoints length %d must be less than 21!" % len(available_endpoints))
+
+        else:
+            logging.info('Skipping test as available endpoints attribute ID not in attribute list on DUT')
```
cecille, line 69:
```suggestion
            self.mark_current_step_skipped()
```
The more you do of these, the more I'm thinking it's worthwhile to pop these into a guard function, similar to pics_guard, but for features and attributes etc. There are now "has_feature" and "has_attribute" functions, it seems like this might be neater. Not a blocker here, but maybe give it some thought about whether that would make this stuff easier to write for other test authors going forward. We could force the tests to pull a full wildcard at the start of the tests (or at least a wildcard of the global attributes). We're kind of going there with the pics reduction stuff anyway.

src/python_testing/TC_PWRTL_2_1.py
```diff
@@ -46,30 +46,45 @@ async def test_TC_PWRTL_2_1(self):
...
-        self.print_step(3, "Read ActiveEndpoints attribute")
-        active_endpoints = await self.read_single_attribute_check_success(endpoint=endpoint, cluster=Clusters.Objects.PowerTopology,  attribute=attributes.ActiveEndpoints)
-        logging.info("ActiveEndpoints: %s" % (active_endpoints))
-        asserts.assert_less_equal(len(active_endpoints), 20,
-                                  "ActiveEndpoints length %d must be less than 21!" % len(active_endpoints))
-        # Verify that ActiveEndpoints is a subset of AvailableEndpoints
-        asserts.assert_true(set(active_endpoints).issubset(set(available_endpoints)),
-                            "ActiveEndpoints should be a subset of AvailableEndpoints")
+        else:
+            logging.info('Skipping test step 3 as active endpoints attribute ID not in attribute list on DUT')
```
cecille, line 86:
```suggestion
            self.mark_current_step_skipped()
```

src/python_testing/TC_PWRTL_2_1.py
```diff
@@ -46,30 +46,45 @@ async def test_TC_PWRTL_2_1(self):
...
+            return
```
cecille, line 87:
```suggestion
```

cecille (approved):

src/python_testing/TC_PWRTL_2_1.py
cecille: feel free to resolve this comment, btw...nothing to do on this PR, we can continue on the issue.

src/python_testing/TC_PWRTL_2_1.py
cecille: https://github.com/project-chip/matter-test-scripts/issues/370 - started an issue to discuss this further. We can continue discussing on there and you can resolve this comment once it's done.

j-ororke: Hi Cecille,
Thank you, I have used the suggestions you have provided in this test, the Opstate tests, and the ICDM 2 1 test regarding creating the attributes_guard function in the matter_testing_support helper module, and have used the has_attributes function for helping with the attribute checks in the OpstateCommon module in PR [#34290](https://github.com/project-chip/connectedhomeip/pull/34290), please let me know if this will work!
Thank you!

PR #34098, opened 2024-06-27
[Silabs] Adding disconnect on factory reset and OTA rename to multi ota

chirag-silabs:
Disconnecting the wifi, upon the factory reset so that the client sends a deauth, as a successfully disconnection

Renaming the `ota_image_tool.py` to `ota_multi_image_tool.py`


jmartinez-silabs (approved):

selissia (changes requested):
The README files should not be updated. The main script is not being renamed.

selissia (approved):

yufengwangca (approved):

PR #34108, opened 2024-06-27
Logging cleanup

mbknust:
Currently, the `src/platform/logging` directory contains a handful of files:
```
src/platform/logging/
├── BUILD.gn
├── impl
│   ├── android
│   │   └── Logging.cpp
│   └── stdio
│       └── Logging.cpp
└── LogV.h
```

Each of them can be put in a more suitable place, which makes it possible to get rid of the `src/platform/logging` directory entirely.

### Changes

1. Move `src/platform/logging/LogV.h` to `src/include/platform/`

2. ~~Move `src/platform/logging/impl/android/Logging.cpp` to `src/platform/android/`~~ (this became a separate PR #34192)

3. Move and rename  `src/platform/logging/impl/stdio/Logging.cpp` to `src/platform/StdioLoggingImpl.cpp`

4. Merge content of `src/platform/logging/BUILD.gn` into `src/platform/BUILD.gn`
   - `logging:stdio` became `platform:stdio-logging`
   - `logging:default` became `platform:default-logging`


examples/common/m5stack-tft/repo
arkq, line 1: Why this is modified?

mbknust: It was a mistake.

arkq (approved):

andy31415 (changes requested):
Could you explain in the PR summary why this change is desirable? The description says that we can remove `logging` directory, however from a code organization perspective a directory named `logging` would give targets of `logging:<name>` and this change moves it to `<name>-logging` so it seems like a noop but with moving thigs out of a dedicated folder. A folder named "logging" is obviously related to logging. A folder named "platform" does not seem so.

I would like to understand how this improves maintainability and structure. Naming-wise, `platform:default-logging` seems slightly worse than `platform/logging:default` (we could even remove the default and name it logging, so then you could use `platform/logging` which could not be done if we move everything in platform.

examples/chip-tool/BUILD.gn
```diff
@@ -92,7 +92,7 @@ static_library("chip-tool-utils") {
     sources += [ "commands/interactive/InteractiveCommands.cpp" ]
     deps += [
       "${chip_root}/examples/common/websocket-server",
-      "${chip_root}/src/platform/logging:headers",
+      "${chip_root}/src/platform:logging-header",
```
andy31415, line 95: This flattens the "logging" into "platform". We currently have this in `src/app` and it is not good for maintainabilit - we end up getting a lot of separate libraries in one large build files, without them being related. We seem to start getting this into "core" as well as we figured out we have to fix deps. Flattening things out into "platform" seems  undesirable.

arkq:
> I would like to understand how this improves maintainability and structure.

Currently we have "platform/Linux", "platform/ESP32", "platform/Darwin" and so on. The first impression when looking at the directory structure is that every subdirectory is for a dedicated platform. There is also a "platform/fake" which somehow fits here. And then one discovers "platform/logging". Definitely there is no platform named "logging". So, the reasoning behind getting the logging out of the platform directory was to keep only real (except the "fake") platforms here. IMHO, that should improve readability of that directory level. However, I agree that this might make a mess on the upper level...

On the other hand the "platform" directory level contains a lot of files related to platform which could be grouped into logical blocks (modules). Putting them into dedicated subdirectories will mix real platforms with platform modules...

Currently the "platform/logging" contains logger implementation for STDIO (common for all platforms in case of forced STDIO logging), and implementation for Android. I think, that the Android one should be moved to the "platform/Android" anyway (it will follow pattern for other platforms where platform-specific implementation is kept in "platform/<name>" subdirectory. After  this move, we will end up with only one file in the "platform/logging" - the common STDIO implementation. So, we thought that there is no reason to keep a dedicated directory for only one file in it. Basically that's the story behind this PR :)

arkq:
Closing, since other approach has been already merged into master.

PR #34110, opened 2024-06-27
[ESP32]: Fix ble scanning for esp32

jadhavrohit924:
Fix ble scanning for esp32

### DIFF
--- a/src/platform/ESP32/nimble/BLEManagerImpl.cpp
+++ b/src/platform/ESP32/nimble/BLEManagerImpl.cpp
@@ -659,7 +659,11 @@ bool BLEManagerImpl::SendWriteRequest(BLE_CONNECTION_OBJECT conId, const ChipBle
         return false;
     }
 
-    assert(pBuf->DataLength() <= UINT16_MAX);
+    if (pBuf->DataLength() > UINT16_MAX)
+    {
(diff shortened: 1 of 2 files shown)


src/platform/ESP32/nimble/BLEManagerImpl.cpp
```diff
@@ -659,7 +659,9 @@ bool BLEManagerImpl::SendWriteRequest(BLE_CONNECTION_OBJECT conId, const ChipBle
         return false;
     }
 
-    rc = ble_gattc_write_flat(conId, chr->chr.val_handle, pBuf->Start(), pBuf->DataLength(), OnWriteComplete, this);
+    assert(pBuf->DataLength() <= UINT16_MAX);
```
arkq, line 662: Maybe you should return an error code here instead of aborting?

arkq (approved):

bzbarsky-apple (approved):

src/platform/ESP32/nimble/BLEManagerImpl.cpp
jadhavrohit924: Done

wqx6 (approved):

PR #34117, opened 2024-06-27
Demand Response Load Control xml

steveholmgren:
Demand Response Load Control XML output from Alchemy ZAP run verified against spec


src/app/zap-templates/zcl/data-model/chip/drlc-cluster.xml
```diff
@@ -144,37 +118,34 @@ limitations under the License.
...
     <cluster code="0x0096"/>
-    <item fieldId="0" name="ProgramID" type="octet_string" length="16"/>
-    <item fieldId="1" name="Name" type="long_char_string" length="32"/>
-    <item fieldId="2" name="EnrollmentGroup" type="int8u" isNullable="true" min="1" max="0xFF"/>
-    <item fieldId="3" name="RandomStartMinutes" type="int8u" isNullable="true" min="0x00" max="0x3C"/>
-    <item fieldId="4" name="RandomDurationMinutes" type="int8u" isNullable="true" min="0x00" max="0x3C"/>
+    <item fieldId="0" name="ProgramID" type="octet_string" length="16" default="0" minLength="16"/>
+    <item fieldId="1" name="Name" type="char_string" length="32"/>
+    <item fieldId="2" name="RandomStartMinutes" type="int8u" isNullable="true" max="60"/>
+    <item fieldId="3" name="RandomDurationMinutes" type="int8u" isNullable="true" max="0x3C"/>
```
bzbarsky-apple, line 137: It's interesting that alchemy output 60 for one of those and 0x3C for the other.  Maybe check with @hasty what's going on there?

src/app/zap-templates/zcl/data-model/chip/drlc-cluster.xml
```diff
@@ -198,73 +169,65 @@ limitations under the License.
...
+      <field id="3" name="Criticality" type="CriticalityLevelEnum" min="0x00" max="0x06"/>
+      <field id="4" name="Control" type="EventControlBitmap" default="0x0000" min="0x0000" max="0x0001"/>
+      <field id="5" name="TemperatureControl" type="TemperatureControlStruct" isNullable="true" optional="true"/>
+      <field id="6" name="AverageLoadControl" type="AverageLoadControlStruct" isNullable="true" optional="true"/>
+      <field id="7" name="DutyCycleControl" type="DutyCycleControlStruct" isNullable="true" optional="true"/>
+      <field id="8" name="PowerSavingsControl" type="PowerSavingsControlStruct" isNullable="true" optional="true"/>
+      <field id="9" name="HeatingSourceControl" type="HeatingSourceControlStruct" isNullable="true" optional="true"/>
       <!-- Spec has an invalid field ID that cannot be used, so not enabling the Signature field for now.  See
            https://github.com/CHIP-Specifications/connectedhomeip-spec/issues/8515 -->
       <!-- <field id="0xFF" name="Signature" type="octet_string" length="48" isNullable="true"/> -->
```
bzbarsky-apple, line 229:
```suggestion
```
That field has been removed from the spec.

bzbarsky-apple (approved):

src/app/zap-templates/zcl/data-model/chip/drlc-cluster.xml
steveholmgren: Will remove it.

src/app/zap-templates/zcl/data-model/chip/drlc-cluster.xml
steveholmgren: Not sure about this - will ask look into it see if I can explain - otherwise ask Hasty - Thanks for the review

hasty: This is a side effect of alchemy trying to be gentle with existing XML, and leaving formatting alone when the values are equivalent, since people used random strategies for hex vs int. 

In this case, field #&#8203;2 was removed, so field #&#8203;3 became field #&#8203;2, field #&#8203;4 became field #&#8203;3, etc. When it attempted to update field #&#8203;2 to match what used to be field #&#8203;3, the max value of "0xFF" is not equivalent to "60", so it re-rendered the "max" attribute, which defaults to using base 10. But when it got to field #&#8203;3, which used to be field #&#8203;4, it already had a max attribute set to "0x3C", which is equivalent to "60", so it left it alone.

If you update it to "60", alchemy will leave it alone on future runs. Whenever the ZAP XML as a whole is updated to have predicatable formatting for numbers, we could get rid of this gentleness.

andy31415:
Closing stale PR.

PR #34121, opened 2024-06-28
[Fabric-Sync] Add RPC method RemoveSynchronizedDevice

### LINKED ISSUE #33973: [Feature]  Add RPC method RemoveSynchronizedDevice
### Feature description

...
Add RPC method RemoveSynchronizedDevice to notify the Fabric Bridge when a device is removed

### Platform

other

### Platform Version(s)

_No response_

### Anything else?

_No response_

yufengwangca:
1. Add RPC method RemoveSynchronizedDevice to notify the Fabric Bridge when a device is removed
2. Enable RPC in CI

Fixes #33973

### DIFF
--- a/examples/fabric-admin/rpc/RpcClient.cpp
+++ b/examples/fabric-admin/rpc/RpcClient.cpp
@@ -65,7 +65,7 @@ void OnRemoveDeviceResponseCompleted(const pw_protobuf_Empty & response, pw::Sta
     }
     else
     {
-        ChipLogProgress(NotSpecified, "RemoveSynchronizedDevice RPC call failed with status: %d\n", status.code());
+        ChipLogProgress(NotSpecified, "RemoveSynchronizedDevice RPC call failed with status: %d", status.code());
     }
 }
 
--- a/examples/fabric-admin/rpc/RpcClient.cpp
+++ b/examples/fabric-admin/rpc/RpcClient.cpp
@@ -90,11 +90,14 @@ CHIP_ERROR AddSynchronizedDevice(chip::NodeId nodeId)
     chip_rpc_SynchronizedDevice device;
     device.node_id = nodeId;
 
-    // The RPC will remain active as long as `addSynchronizedDeviceCall` is alive.
+    // By assigning the returned call to the global 'addSynchronizedDeviceCall', the RPC
+    // call is kept alive until it completes. When a response is received, it
+    // will be logged by the handler function and the call will complete.
     addSynchronizedDeviceCall = fabricBridgeClient.AddSynchronizedDevice(device, OnAddDeviceResponseCompleted);
 
     if (!addSynchronizedDeviceCall.active())
     {
+        // The RPC call was not sent. This could occur due to, for example, an invalid channel ID. Handle if necessary.
         return CHIP_ERROR_INTERNAL;
     }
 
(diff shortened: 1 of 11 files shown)


examples/fabric-admin/rpc/RpcClient.cpp
```diff
@@ -55,6 +56,19 @@ void OnAddDeviceResponseCompleted(const pw_protobuf_Empty & response, pw::Status
...
+// Callback function to be called when the RPC response is received
+void OnRemoveDeviceResponseCompleted(const pw_protobuf_Empty & response, pw::Status status)
+{
+    if (status.ok())
+    {
+        ChipLogProgress(NotSpecified, "RemoveSynchronizedDevice RPC call succeeded!");
+    }
+    else
+    {
+        ChipLogProgress(NotSpecified, "RemoveSynchronizedDevice RPC call failed with status: %d\n", status.code());
```
andy31415, line 68:
```suggestion
        ChipLogProgress(NotSpecified, "RemoveSynchronizedDevice RPC call failed with status: %d", status.code());
```

examples/fabric-admin/rpc/RpcClient.cpp
```diff
@@ -86,3 +100,27 @@ CHIP_ERROR AddSynchronizedDevice(chip::NodeId nodeId)
...
+    if (removeSynchronizedDeviceCall.active())
+    {
+        ChipLogError(NotSpecified, "Remove Synchronized Device operation is in progress\n");
+        return CHIP_ERROR_BUSY;
+    }
+
+    chip_rpc_SynchronizedDevice device;
+    device.node_id = nodeId;
+
+    // The RPC will remain active as long as `removeSynchronizedDeviceCall` is alive.
```
andy31415, line 117: I do not understand this comment. Could you explain why the active check below is correct? This somehow feels like a race condition however not really sure and I would hope that comments would explain this.

The way I read the code is that we issue an RPC (RemoveSynchronizedDevice) and then we validate that RPC is still on-going immediately after issuing it. Why is that correct? What if we get a context switch and the RPC is really really fast? Is the RPC guaranteed to not be finished (e.g. it also runs under the chip stack so will not progress while other code is running)?

examples/fabric-bridge-app/linux/RpcServer.cpp
```diff
@@ -64,6 +65,21 @@ pw::Status FabricBridge::AddSynchronizedDevice(const chip_rpc_SynchronizedDevice
     return pw::OkStatus();
 }
 
+pw::Status FabricBridge::RemoveSynchronizedDevice(const chip_rpc_SynchronizedDevice & request, pw_protobuf_Empty & response)
+{
+    NodeId nodeId = request.node_id;
+    ChipLogProgress(NotSpecified, "Received RemoveSynchronizedDevice: " ChipLogFormatX64, ChipLogValueX64(nodeId));
+
+    int result = DeviceMgr().RemoveDeviceByNodeId(nodeId);
+    if (result == -1)
```
andy31415, line 74:
```suggestion
    int removed_idx = DeviceMgr().RemoveDeviceByNodeId(nodeId);
    if (removed_idx < 0)
```

just "result" is not as clear about it's meaning and why -1 is special.

examples/fabric-admin/rpc/RpcClient.cpp
yufengwangca: Updated the comments:

    // By assigning the returned call to the global 'removeSynchronizedDeviceCall', the RPC
    // call is kept alive until it completes. When a response is received, it
    // will be logged by the handler function and the call will complete.
    removeSynchronizedDeviceCall = fabricBridgeClient.RemoveSynchronizedDevice(device, OnRemoveDeviceResponseCompleted);

    if (!removeSynchronizedDeviceCall.active())
    {
        // The RPC call was not sent. This could occur due to, for example, an invalid channel ID. Handle if necessary.
        return CHIP_ERROR_INTERNAL;
    }

andy31415 (approved):

chulspro (approved):

PR #32389, opened 2024-03-01
[ESP32] Enable extended ble announcement for esp32 platform

sayondeep:
Extended BLE advertisement was recently introduced in this [PR](https://github.com/project-chip/connectedhomeip/pull/31532).
This PR enables extended BLE advertisement for ESP32 platform.
The extended BLE advertisement increases the advertisement duration to up to 48 hr before closing the commissioning window.

**Tests Performed:**

- Tested successful commissioning at all three phases (Fast advertisement, Slow advertisement and Extended advertisement).

PR #34133, opened 2024-06-28
[ESP32] Fix the bluedroid build issue on ESP32.

shripad621git:
**Problem:**
- Build with Bluedroid was failing on ESP32 examples.

**Change Overview**
- Fixed the build issue with bluedroid.
- Added extended ble advertising support.
- Added a TODO for  [Bluedroid Shutdown flow](https://github.com/project-chip/connectedhomeip/issues/36919). 

#### Testing
- Tested the lighting app esp32 with bluedroid flow for esp32 and esp32c3.

### DIFF
--- a/src/platform/ESP32/bluedroid/BLEManagerImpl.cpp
+++ b/src/platform/ESP32/bluedroid/BLEManagerImpl.cpp
@@ -147,7 +147,6 @@ const uint16_t CHIPoBLEGATTAttrCount = sizeof(CHIPoBLEGATTAttrs) / sizeof(CHIPoB
 ChipDeviceScanner & mDeviceScanner = Internal::ChipDeviceScanner::GetInstance();
 #endif
 BLEManagerImpl BLEManagerImpl::sInstance;
-constexpr System::Clock::Timeout BLEManagerImpl::kFastAdvertiseTimeout;
 #ifdef CONFIG_ENABLE_ESP32_BLE_CONTROLLER
 static esp_gattc_char_elem_t * char_elem_result   = NULL;
 static esp_gattc_descr_elem_t * descr_elem_result = NULL;
@@ -228,6 +227,23 @@ exit:
     return err;
 }
 
+void BLEManagerImpl::_Shutdown()
+{
+    CancelBleAdvTimeoutTimer();
+
+    BleLayer::Shutdown();
+    mServiceMode = ConnectivityManager::kCHIPoBLEServiceMode_Disabled;
+
+    // selectively setting kGATTServiceStarted flag, in order to notify the state machine to stop the CHIPoBLE gatt service
+    mFlags.ClearAll().Set(Flags::kGATTServiceStarted);
+
+#ifdef CONFIG_ENABLE_ESP32_BLE_CONTROLLER
+    OnChipBleConnectReceived = nullptr;
+#endif // CONFIG_ENABLE_ESP32_BLE_CONTROLLER
+
+    PlatformMgr().ScheduleWork(DriveBLEState, 0);
+}
+
 CHIP_ERROR BLEManagerImpl::_SetAdvertisingEnabled(bool val)
 {
     CHIP_ERROR err = CHIP_NO_ERROR;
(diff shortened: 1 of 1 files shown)


shubhamdp (approved):

src/platform/ESP32/bluedroid/BLEManagerImpl.cpp
```diff
@@ -1158,6 +1218,27 @@ CHIP_ERROR BLEManagerImpl::ConfigureAdvertisingData(void)
         ExitNow();
     }
 
+#ifdef CHIP_DEVICE_CONFIG_BLE_EXT_ADVERTISING
```
bzbarsky-apple, line 1221: Typically, CHIP_DEVICE_CONFIG_* bits are always defined and just have a boolean value that gets tested.  That said, I don't see this define in the tree so far; is this PR introducing it?  This part does not make sense to me.

src/platform/ESP32/bluedroid/BLEManagerImpl.cpp
```diff
@@ -907,6 +930,25 @@ CHIP_ERROR BLEManagerImpl::MapBLEError(int bleErr)
     }
 }
 
+void BLEManagerImpl::CancelBleAdvTimeoutTimer(void)
+{
+    if (SystemLayer().IsTimerActive(BleAdvTimeoutHandler, nullptr))
```
bzbarsky-apple, line 935: Why do you need the IsTimerActive bit?   CancelTimer can be called even if the timer is not running and will just do nothing...

src/platform/ESP32/bluedroid/BLEManagerImpl.cpp
```diff
@@ -907,6 +930,25 @@ CHIP_ERROR BLEManagerImpl::MapBLEError(int bleErr)
...
+    {
+        SystemLayer().CancelTimer(BleAdvTimeoutHandler, nullptr);
+    }
+}
+
+void BLEManagerImpl::StartBleAdvTimeoutTimer(uint32_t aTimeoutInMs)
+{
+    CancelBleAdvTimeoutTimer();
```
bzbarsky-apple, line 943: Note that StartTimer is documented to cancel existing matching timers, so you probably don't need this part here.  But either way.

src/platform/ESP32/bluedroid/BLEManagerImpl.cpp
shripad621git: @bzbarsky-apple, these are build fixes as the extended advertising part was not implemented for bluedroid. The option was introduced in https://github.com/project-chip/connectedhomeip/pull/32389.
The bits of the build fixes are added in this PR and tested for extended advertisement.

bzbarsky-apple: > The option was introduced in https://github.com/project-chip/connectedhomeip/pull/32389.

Yes, but it does not exist on tip, if you actually look:

```
% git grep CHIP_DEVICE_CONFIG_BLE_EXT_ADVERTISING | grep define
src/include/platform/CHIPDeviceConfig.h:#define CHIP_DEVICE_CONFIG_BLE_EXT_ADVERTISING_INTERVAL_CHANGE_TIME_MS (15 * 60 * 1000)
src/include/platform/CHIPDeviceConfig.h:#define CHIP_DEVICE_CONFIG_BLE_EXT_ADVERTISING_INTERVAL_MIN 2056
src/include/platform/CHIPDeviceConfig.h:#define CHIP_DEVICE_CONFIG_BLE_EXT_ADVERTISING_INTERVAL_MAX 2056
```

The define added in https://github.com/project-chip/connectedhomeip/pull/32389 is not there.  I recall that it got renamed, and this name needs to match the actual name.

But also, in https://github.com/project-chip/connectedhomeip/pull/32389 the define is unconditional, so this ifdef check makes no sense.

shripad621git: I have changed the `CHIP_DEVICE_CONFIG_BLE_EXT_ADVERTISING` to renamed `CHIP_DEVICE_CONFIG_EXT_ADVERTISING`.

bzbarsky-apple (approved):

PR #31162, opened 2023-12-21
Optimize out no-op writes in attribute-table.

bzbarsky-apple:
This way we don't mark things dirty if they have not changed.

Does not optimize out string values that are unchanged, because we have no good way to tell how much space we need to read the old value.

Fixes https://github.com/project-chip/connectedhomeip/issues/29136

Still needs tests, so draft.

PR #34139, opened 2024-06-28
Implement the codegen-data-model Write support

andy31415:
This implements the write support via ember/attribute-access-interface. Generally the equivalent that today exists in ember-compatibility-functions.

image


src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,346 @@
...
+    {
+        return std::nullopt;
+    }
+
+    CHIP_ERROR err = aai->Write(path, decoder);
+
+    // explict translate UnsupportedRead to Access denied. This is to allow callers to determine a
+    // translation for this: usually wildcard subscriptions MAY just ignore these where as direct reads
+    // MUST translate them to UnsupportedAccess
+    ReturnErrorCodeIf(err == CHIP_IM_GLOBAL_STATUS(UnsupportedWrite), CHIP_ERROR_ACCESS_DENIED);
```
andy31415, line 65: For reviewers: this logic was remainder from previous code. I will update this to not return UnsupportedWrite but rather maintain the logic for expanded paths (and update comments)

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,341 @@
...
+/// Attempts to write via an attribute access interface (AAI)
+///
+/// If it returns a CHIP_ERROR, then this is a FINAL result (i.e. either failure or success)
+///
+/// If it returns std::nullopt, then there is no AAI to handle the given path
+/// and processing should figure out the value otherwise (generally from other ember data)
+std::optional<CHIP_ERROR> TryWriteViaAccessInterface(const ConcreteAttributePath & path, AttributeAccessInterface * aai,
+                                                     AttributeValueDecoder & decoder)
+{
+
```
tcarmelveilleux, line 52:
```suggestion
```

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,341 @@
...
+    CHIP_ERROR err = aai->Write(path, decoder);
+
+    if (err != CHIP_NO_ERROR)
+    {
+        return std::make_optional(err);
+    }
+
+    // If the decoder tried to decode, then a value should have been read for processing.
+    //   - if decode, assueme DONE (i.e. FINAL CHIP_NO_ERROR)
```
tcarmelveilleux, line 67:
```suggestion
    //   - if decoding was done, assume DONE (i.e. final CHIP_NO_ERROR)
```

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,341 @@
...
+    //   - if no encode, say that processing must continue
```
tcarmelveilleux, line 68:
```suggestion
    //   -  otherwise, if no decoding done, return that processing must continue via nullopt
```

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,341 @@
...
+    {
+        typename DataModel::Nullable<T> nullableWorkingValue;
+        ReturnErrorOnFailure(decoder.Decode(nullableWorkingValue));
+
+        if (nullableWorkingValue.IsNull())
+        {
+            VerifyOrReturnError(out.size() >= sizeof(typename ENCODING::LengthType), CHIP_ERROR_BUFFER_TOO_SMALL);
+
+            typename ENCODING::LengthType nullLength = ENCODING::kNullLength;
+            memcpy(out.data(), &nullLength, sizeof(nullLength));
```
tcarmelveilleux, line 111: State that the encoding is using native byte order.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,341 @@
...
+        workingValue = nullableWorkingValue.Value();
+    }
+    else
+    {
+        ReturnErrorOnFailure(decoder.Decode(workingValue));
+    }
+
+    typename ENCODING::LengthType len = static_cast<typename ENCODING::LengthType>(workingValue.size());
+    VerifyOrReturnError(out.size() >= sizeof(len) + len, CHIP_ERROR_BUFFER_TOO_SMALL);
+
```
tcarmelveilleux, line 125: Suggest:

```
    uint8_t * cursor = out.data();
    memcpy(cursor, &len, sizeof(len));
    cursor += sizeof(len);
    memcpy(cursor, workingValue.data(), workingValue.size());
```

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,341 @@
...
+        else
+        {
+            VerifyOrReturnError(Traits::CanRepresentValue(isNullable, *workingValue), CHIP_ERROR_INVALID_ARGUMENT);
+            Traits::WorkingToStorage(*workingValue, storageValue);
+        }
+
+        VerifyOrReturnError(out.size() >= sizeof(storageValue), CHIP_ERROR_INVALID_ARGUMENT);
+
+        const uint8_t * data = Traits::ToAttributeStoreRepresentation(storageValue);
+        memcpy(out.data(), data, sizeof(storageValue));
```
tcarmelveilleux, line 158: Indicate byte ordering assumptions for this native encoding. Is it really native on all platforms?

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,341 @@
...
+
+/// Encodes a numeric data value of type T from the given ember-encoded buffer `data`.
+///
+/// isNullable defines if the value of NULL is allowed to be encoded.
+template <typename T>
+CHIP_ERROR DecodeIntoEmberBuffer(AttributeValueDecoder & decoder, bool isNullable, MutableByteSpan out)
+{
+    if (isNullable)
+    {
+        using Traits = NumericAttributeTraits<T>;
```
tcarmelveilleux, line 139: Suggest hoisting the following lines to the top of the scope, outside the if:

```
using Traits = NumericAttributeTraits<T>;
typename Traits::StorageType storageValue;
```

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,341 @@
...
+        ReturnErrorOnFailure(decoder.Decode(workingValue));
+
+        typename Traits::StorageType storageValue;
+        Traits::WorkingToStorage(workingValue, storageValue);
+
+        VerifyOrReturnError(out.size() >= sizeof(storageValue), CHIP_ERROR_INVALID_ARGUMENT);
+
+        // This guards against trying to encode something that overlaps nullable, for example
+        // Nullable<uint8_t>(0xFF) is not representable because 0xFF is the encoding of NULL in ember
+        VerifyOrReturnError(Traits::CanRepresentValue(isNullable, workingValue), CHIP_ERROR_INVALID_ARGUMENT);
```
tcarmelveilleux, line 174: - IsNullable is always false here
- It is always valid to store 0xFF in a non-nullable field. This line should be removed.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,341 @@
...
+        return DecodeStringLikeIntoEmberBuffer<CharSpan, ShortPascalString>(decoder, isNullable, out);
+    case ZCL_LONG_CHAR_STRING_ATTRIBUTE_TYPE:
+        return DecodeStringLikeIntoEmberBuffer<CharSpan, LongPascalString>(decoder, isNullable, out);
+    case ZCL_OCTET_STRING_ATTRIBUTE_TYPE: // Octet string
+        return DecodeStringLikeIntoEmberBuffer<ByteSpan, ShortPascalString>(decoder, isNullable, out);
+    case ZCL_LONG_OCTET_STRING_ATTRIBUTE_TYPE:
+        return DecodeStringLikeIntoEmberBuffer<ByteSpan, LongPascalString>(decoder, isNullable, out);
+    default:
+        ChipLogError(DataManagement, "Attribute type 0x%x not handled", static_cast<int>(metadata->attributeType));
+        return CHIP_IM_GLOBAL_STATUS(UnsupportedWrite);
```
tcarmelveilleux, line 243: This is the wrong value --> `UnsupportedWrite` can only be used when a write is attempted to a read-only value. Please revisit every usage you make of UnsupportedWrite, and replace with ConstraintError, most likely.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,341 @@
...
+    }
+    return CHIP_IM_GLOBAL_STATUS(UnsupportedWrite);
```
tcarmelveilleux, line 245: This is unreachable.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,341 @@
...
+}
+
+} // namespace
+
+CHIP_ERROR CodegenDataModel::WriteAttribute(const InteractionModel::WriteAttributeRequest & request,
+                                            AttributeValueDecoder & decoder)
+{
+    ChipLogDetail(DataManagement,
+                  "Reading attribute: Cluster=" ChipLogFormatMEI " Endpoint=%x AttributeId=" ChipLogFormatMEI " (expanded=%d)",
```
tcarmelveilleux, line 254: This says "Reading" attribute but it's a write.

tcarmelveilleux, line 254: Also, probably a good time to make this log more compact and in better order:

`EP 0x%x/Cluster 0x..../Attr 0x.....`

The `expanded` is not useful and should be removed I would say.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,341 @@
...
+                  ChipLogValueMEI(request.path.mClusterId), request.path.mEndpointId, ChipLogValueMEI(request.path.mAttributeId),
+                  request.path.mExpanded);
+
+    // ACL check for non-internal requests
+    if (!request.operationFlags.Has(InteractionModel::OperationFlags::kInternal))
+    {
+        ReturnErrorCodeIf(!request.subjectDescriptor.has_value(), CHIP_ERROR_INVALID_ARGUMENT);
```
tcarmelveilleux, line 260: This should return IM global status of UnsupportedAccess.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,341 @@
...
+
+        Access::RequestPath requestPath{ .cluster = request.path.mClusterId, .endpoint = request.path.mEndpointId };
+        ReturnErrorOnFailure(Access::GetAccessControl().Check(*request.subjectDescriptor, requestPath,
```
tcarmelveilleux, line 263: This returns CHIP_ERROR_ACCESS_DENIED, but then that means we rely on that to be converted somewhere deep into UnsupportedAccess IM error. Suggest checking that == CHIP_NO_ERROR and explicitly return IM error UnsupportedAccess otherwise. This will reduce reliance on weird legacy conversions deep in the stack.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,341 @@
...
+                                                              RequiredPrivilege::ForWriteAttribute(request.path)));
+    }
+
+    auto metadata = Ember::FindAttributeMetadata(request.path);
+
+    // Explicit failure in finding a suitable metadata
```
tcarmelveilleux, line 269: I don't understand this sentence.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,341 @@
...
+    if (const CHIP_ERROR * err = std::get_if<CHIP_ERROR>(&metadata))
+    {
+        VerifyOrDie(*err != CHIP_NO_ERROR);
+        return *err;
+    }
+
+    // All the global attributes that we do not have metadata for are
+    // read-only (i.e. cannot write atribute_list/event_list/accepted_cmds/generated_cmds)
+    //
+    // so if no metadata available, we wil lreturn an error
```
tcarmelveilleux, line 279:
```suggestion
    // All the global attributes that we do not have metadata for are
    // read-only (i.e. cannot write attribute_list/accepted_cmds/generated_cmds, etc),
    // so if no metadata is available, we will return an error.
```

Note that this is a strong assumption. When we want to support code-backed clusters that report which attributes they support, lack of metadata may arise from EmberMetadata. Suggest adding a TODO comment with new issue  # for getting a non-ember-metadata source of truth for the presence of an attribute.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,344 @@
...
+    const EmberAfAttributeMetadata ** attributeMetadata = std::get_if<const EmberAfAttributeMetadata *>(&metadata);
+    if (attributeMetadata == nullptr)
+    {
+        return CHIP_IM_GLOBAL_STATUS(UnsupportedWrite);
+    }
+
+    if ((*attributeMetadata)->IsReadOnly() && !request.operationFlags.Has(InteractionModel::OperationFlags::kInternal))
+    {
```
tcarmelveilleux, line 291:
```suggestion
    bool isReadOnly = (attributeMetadata == nullptr);
    if (isReadOnly)
    {
        return CHIP_IM_GLOBAL_STATUS(UnsupportedWrite);
    }
    
    isReadOnly = (*attributeMetadata)->IsReadOnly();
    if (isReadOnly && !request.operationFlags.Has(InteractionModel::OperationFlags::kInternal))
    {
```

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,344 @@
...
+        // Internal is allowed to try to bypass read-only updates, however otherwise we deny read-only
+        // updates
+        return CHIP_IM_GLOBAL_STATUS(UnsupportedWrite);
+    }
+
+    if ((*attributeMetadata)->MustUseTimedWrite() && !request.writeFlags.Has(InteractionModel::WriteFlags::kTimed))
```
tcarmelveilleux, line 297: This prevents "internal writing" of attributes requiring timed write. 

Suggest having both the read-only and timed-write error-handling paths in a `if(!request.operationFlags.Has(InteractionModel::OperationFlags::kInternal)` outer block to reduce the inner ifs, and at the same time, clarify that these checks are skipped when "force writing" internally.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
andy31415: Ember compatibility functions do a memcpy:

https://github.com/project-chip/connectedhomeip/blob/master/src/app/util/ember-compatibility-functions.cpp#L511

And then ember-strings.cpp uses LittleEndian: https://github.com/project-chip/connectedhomeip/blob/master/src/app/util/ember-strings.cpp#L34

We probably never noticed because we are not really compiling for big endian :(

I am unsure what to do here. Assuming LittleEndian is ok for now because that is what the ember code is likely using at some point. It also seems that ember-compatibility-functions is broken then.

andy31415: Will use LittleEndian since that seems explicit. But this does mean my code is different from ember-compatibility-functions (but probably more correct).

src/app/codegen-data-model/CodegenDataModel_Write.cpp
andy31415: Updated. This is attribute store functions, so they take care of endianess..

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,344 @@
...
+    {
+        return CHIP_IM_GLOBAL_STATUS(NeedsTimedInteraction);
```
tcarmelveilleux, line 301: `NeedsTimedInteraction` should be renamed `TimedRequestMismatch`

src/app/codegen-data-model/CodegenDataModel_Write.cpp
andy31415: This is generic for types that are non-string-like. Specific example is the odd-sized integers. In that case say a 32-bit integer is used as "working value" however the "storage value" could be a 3-byte value.

In that case, CanRepresentValue(0x1234abcd) will return false even if non-nullable because we can only represent 0 ... 0xFFFFFF.

andy31415: I will update comments - I think I got a copy and paste.

For isNullable it is true that on each branch I know what it is (true or false), however I figured booleans are hard to read so I would rather have `CanRepresentValue(isNullable, workingValue)` then `CanRepresentValue(false /* isNullable */, workingValue)` since I then have the same functionality, but longer code and more likely to mis-use by copy & paste.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
tcarmelveilleux: Based on offline convo: need to make explicit the 3 UnsupportedEndpoint/Cluster/Attribute error cases.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,346 @@
...
+    }
+
+    if (request.path.mDataVersion.HasValue())
+    {
+        std::optional<InteractionModel::ClusterInfo> clusterInfo = GetClusterInfo(request.path);
+        if (!clusterInfo.has_value())
+        {
+            ChipLogError(DataManagement, "Unable to get cluster info for Endpoint %x, Cluster " ChipLogFormatMEI,
```
tcarmelveilleux, line 312: How can this happen? By now, a critical invariant is that we need to have all cluster data available. I think it's safer not to crash here, but this is not a DataVersionMismatch, this is an internal error.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,349 @@
...
+        if (request.path.mDataVersion.Value() != clusterInfo->dataVersion)
+        {
+            ChipLogError(DataManagement, "Write Version mismatch for Endpoint %x, Cluster " ChipLogFormatMEI,
+                         request.path.mEndpointId, ChipLogValueMEI(request.path.mClusterId));
+            return CHIP_IM_GLOBAL_STATUS(DataVersionMismatch);
+        }
+    }
+
+    std::optional<CHIP_ERROR> aai_result = TryWriteViaAccessInterface(
+        request.path, GetAttributeAccessOverride(request.path.mEndpointId, request.path.mClusterId), decoder);
```
tcarmelveilleux, line 326: Suggest splitting the GetAttributeAccessOverride in a different variable.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,349 @@
...
+    ReturnErrorCodeIf(aai_result.has_value(), *aai_result);
```
tcarmelveilleux, line 329: More readable as:

```
    if (aai_result.has_value())
    {
        return *aai_result;
    }
```

src/app/codegen-data-model/CodegenDataModel_Write.cpp
tcarmelveilleux: b

src/app/codegen-data-model/CodegenDataModel_Write.cpp
andy31415: Good point on the expanded bit: this was logged before because read has special handling for expanded paths, however that makes no sense for writes. Updated log and corresponding code.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,351 @@
...
+    record.clusterId   = request.path.mClusterId;
+    record.attributeId = request.path.mAttributeId;
+
+    Protocols::InteractionModel::Status status = emAfReadOrWriteAttribute(
+        &record, attributeMetadata, gEmberAttributeIOBufferSpan.data(), static_cast<uint16_t>(gEmberAttributeIOBufferSpan.size()),
+        /* write = */ true);
+
+    if (status != Protocols::InteractionModel::Status::Success)
+    {
+        return ChipError(ChipError::SdkPart::kIMGlobalStatus, to_underlying(status), __FILE__, __LINE__);
```
tcarmelveilleux, line 344: Isn't there a way to avoid stating the __FILE__/__LINE__ explicitly here?

src/app/codegen-data-model/EmberMetadata.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+std::variant<const EmberAfCluster *,           // global attribute, data from a cluster
+             const EmberAfAttributeMetadata *, // a specific attribute stored by ember
+             CHIP_ERROR                        // error, this will NEVER be CHIP_NO_ERROR
+             >
+FindAttributeMetadata(const ConcreteAttributePath & aPath)
+{
+    for (auto & attr : GlobalAttributesNotInMetadata)
+    {
+
```
tcarmelveilleux, line 35:
```suggestion
```

src/app/codegen-data-model/tests/TestCodegenModelViaMocks.cpp
```diff
@@ -341,6 +353,10 @@ const MockNodeConfig gTestNodeConfig({
             MOCK_ATTRIBUTE_CONFIG_NULLABLE(ZCL_IPV6ADR_ATTRIBUTE_TYPE),
             MOCK_ATTRIBUTE_CONFIG_NULLABLE(ZCL_IPV6PRE_ATTRIBUTE_TYPE),
             MOCK_ATTRIBUTE_CONFIG_NULLABLE(ZCL_HWADR_ATTRIBUTE_TYPE),
+
+            // Special case handling
+            MockAttributeConfig(kAttributeIdReadOnly, ZCL_INT32S_ATTRIBUTE_TYPE, 0),
```
tcarmelveilleux, line 358: Comment about this file --> this is not a Mock, it's a fake. Don't need to change it in this PR, but I think it's worth calling it a fake later.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
andy31415: I am working here specifically only for ember (i.e. codeggen) support, so there will never be a code-backed support here.

The implementation of code-backed bits will have to be separate, likely after global attribute handling (and ACL) is done inside IM based on metadata rather than the code here.

src/app/codegen-data-model/tests/TestCodegenModelViaMocks.cpp
```diff
@@ -1667,3 +1799,340 @@ TEST(TestCodegenModelViaMocks, ReadGlobalAttributeAttributeList)
...
+TEST(TestCodegenModelViaMocks, EmberAttributeWriteBasicTypes)
+{
+    TestEmberScalarTypeWrite<uint8_t, ZCL_INT8U_ATTRIBUTE_TYPE>(0x12);
+    TestEmberScalarTypeWrite<uint16_t, ZCL_ENUM16_ATTRIBUTE_TYPE>(0x1234);
+    TestEmberScalarTypeWrite<OddSizedInteger<3, false>, ZCL_INT24U_ATTRIBUTE_TYPE>(0x1234AB);
+    TestEmberScalarTypeWrite<uint32_t, ZCL_INT32U_ATTRIBUTE_TYPE>(0x1234ABCD);
+    TestEmberScalarTypeWrite<OddSizedInteger<5, false>, ZCL_INT40U_ATTRIBUTE_TYPE>(0x1122334455);
+    TestEmberScalarTypeWrite<OddSizedInteger<6, false>, ZCL_INT48U_ATTRIBUTE_TYPE>(0xAABB11223344);
+    TestEmberScalarTypeWrite<OddSizedInteger<7, false>, ZCL_INT56U_ATTRIBUTE_TYPE>(0xAABB11223344);
+    TestEmberScalarTypeWrite<uint64_t, ZCL_INT64U_ATTRIBUTE_TYPE>(0x1122334455667788L);
```
tcarmelveilleux, line 1824:
```suggestion
    TestEmberScalarTypeWrite<uint64_t, ZCL_INT64U_ATTRIBUTE_TYPE>(0x1122334455667788ULL);
```

src/app/codegen-data-model/tests/TestCodegenModelViaMocks.cpp
```diff
@@ -1667,3 +1799,340 @@ TEST(TestCodegenModelViaMocks, ReadGlobalAttributeAttributeList)
...
+
+    TestEmberScalarTypeWrite<int8_t, ZCL_INT8S_ATTRIBUTE_TYPE>(-10);
+    TestEmberScalarTypeWrite<int16_t, ZCL_INT16S_ATTRIBUTE_TYPE>(-123);
+    TestEmberScalarTypeWrite<OddSizedInteger<3, true>, ZCL_INT24S_ATTRIBUTE_TYPE>(-1234);
+    TestEmberScalarTypeWrite<int32_t, ZCL_INT32S_ATTRIBUTE_TYPE>(-12345);
+    TestEmberScalarTypeWrite<OddSizedInteger<5, true>, ZCL_INT40S_ATTRIBUTE_TYPE>(-123456);
+    TestEmberScalarTypeWrite<OddSizedInteger<6, true>, ZCL_INT48S_ATTRIBUTE_TYPE>(-1234567);
+    TestEmberScalarTypeWrite<OddSizedInteger<7, true>, ZCL_INT56S_ATTRIBUTE_TYPE>(-12345678);
+    TestEmberScalarTypeWrite<int64_t, ZCL_INT64S_ATTRIBUTE_TYPE>(-123456789);
```
tcarmelveilleux, line 1833: Spec has a list of the following lowest valid value for nullable, for all widths:

```
// 8 bits
-127

// 16 bits
-32767

// 24 bits
-8388607

// 32 bits
-2147483647

// 40 bits
-549755813887

// 48 bits
-140737488355327

// 56 bits
-36028797018963967

// 64 bits
-9223372036854775807
```

Suggest you use those values to cover width properly.

src/app/codegen-data-model/tests/TestCodegenModelViaMocks.cpp
```diff
@@ -1667,3 +1799,340 @@ TEST(TestCodegenModelViaMocks, ReadGlobalAttributeAttributeList)
...
+    TestWriteRequest test(kAdminSubjectDescriptor,
+                          ConcreteAttributePath(kMockEndpoint3, MockClusterId(4),
+                                                MOCK_ATTRIBUTE_ID_FOR_NULLABLE_TYPE(ZCL_LONG_CHAR_STRING_ATTRIBUTE_TYPE)));
+    AttributeValueDecoder decoder =
+        test.DecoderFor<chip::app::DataModel::Nullable<CharSpan>>(chip::app::DataModel::Nullable<CharSpan>());
+
+    ASSERT_EQ(model.WriteAttribute(test.request, decoder), CHIP_NO_ERROR);
+    chip::ByteSpan writtenData = GetEmberBuffer();
+    ASSERT_EQ(writtenData[0], 0xFF);
+    ASSERT_EQ(writtenData[0], 0xFF);
```
tcarmelveilleux, line 1942: Cut and paste error on the assert, missing index 1?

src/app/codegen-data-model/tests/TestCodegenModelViaMocks.cpp
```diff
@@ -1667,3 +1799,340 @@ TEST(TestCodegenModelViaMocks, ReadGlobalAttributeAttributeList)
...
+    AttributeValueDecoder decoder =
+        test.DecoderFor<chip::app::DataModel::Nullable<CharSpan>>(chip::app::DataModel::MakeNullable("text"_span));
+
+    ASSERT_EQ(model.WriteAttribute(test.request, decoder), CHIP_NO_ERROR);
+    chip::ByteSpan writtenData = GetEmberBuffer();
+
+    uint16_t len;
+    memcpy(&len, writtenData.data(), 2);
+    EXPECT_EQ(len, 4);
+    chip::CharSpan asCharSpan(reinterpret_cast<const char *>(writtenData.data() + 2), 4);
```
tcarmelveilleux, line 1922: Watch-out for endianness in all tests...

src/app/codegen-data-model/CodegenDataModel_Write.cpp
andy31415: This should probably be something separate - I use existing list from https://github.com/project-chip/connectedhomeip/blob/master/src/protocols/interaction_model/StatusCodeList.h#L70

src/app/codegen-data-model/CodegenDataModel_Write.cpp
andy31415: Updated to CHIP_ERROR_NOT_FOUND (even though realistically this is terrible ... but also likely not possible in the existing code). I carried it over from:

https://github.com/project-chip/connectedhomeip/blob/master/src/app/util/ember-compatibility-functions.cpp#L168 which returns CHIP_ERROR_NOT_FOUND

src/app/codegen-data-model/CodegenDataModel_Write.cpp
andy31415: Updated to hide it away via a error macro that was added in previous PRs. it still exists, but it is hidden away.

src/app/codegen-data-model/EmberMetadata.cpp
```diff
@@ -0,0 +1,75 @@
...
+        if (attr == aPath.mAttributeId)
```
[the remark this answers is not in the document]
andy31415: This was carried over  from FindAttributeMetadata from ember-compatibility-functions.cpp

https://github.com/project-chip/connectedhomeip/blob/master/src/app/util/ember-compatibility-functions.cpp#L256

My worry was that the logic may be intentional: we want to return cluster FIRST even if the attribute somehow magically exists, because global attributes need to be processed via the cluster. I am afraid that switching order may introduce sideffects that were not there before, so I prefer to keep the logic as is..

src/app/codegen-data-model/tests/TestCodegenModelViaMocks.cpp
andy31415: I inherited the naming from existing tests and the name "mock" propagated throughout.

src/app/codegen-data-model/tests/TestCodegenModelViaMocks.cpp
```diff
@@ -1667,3 +1799,340 @@ TEST(TestCodegenModelViaMocks, ReadGlobalAttributeAttributeList)
...
+
+    AttributeValueDecoder decoder = test.DecoderFor(testValue);
+    ASSERT_EQ(model.WriteAttribute(test.request, decoder), CHIP_NO_ERROR);
+
+    EXPECT_EQ(aai->GetData().a, 112);
+    EXPECT_TRUE(aai->GetData().e.data_equal("aai_write_test"_span));
+
+    // AAI does not prevent read/write of regular attributes
+    TestEmberScalarTypeWrite<uint32_t, ZCL_INT32U_ATTRIBUTE_TYPE>(1234);
+    TestEmberScalarNullWrite<int64_t, ZCL_INT64S_ATTRIBUTE_TYPE>();
```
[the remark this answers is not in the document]
andy31415: Updated the comment a bit more.

src/app/codegen-data-model/tests/TestCodegenModelViaMocks.cpp
andy31415: Added tests for lowest write.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,367 @@
...
+    ReturnErrorCodeIf(aai_result.has_value(), *aai_result);
+
+    ReturnErrorOnFailure(DecodeValueIntoEmberBuffer(decoder, *attributeMetadata, gEmberAttributeIOBufferSpan));
+
+    EmberAfAttributeSearchRecord record;
+    record.endpoint    = request.path.mEndpointId;
+    record.clusterId   = request.path.mClusterId;
+    record.attributeId = request.path.mAttributeId;
+
+    Protocols::InteractionModel::Status status = emAfReadOrWriteAttribute(
```
bzbarsky-apple, line 354: What happened to the "compare the size of the data we ended up with to the max size in the metadata" check?

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,367 @@
...
+        &record, attributeMetadata, gEmberAttributeIOBufferSpan.data(), static_cast<uint16_t>(gEmberAttributeIOBufferSpan.size()),
```
bzbarsky-apple, line 355: This looks wrong; `gEmberAttributeIOBufferSpan.size()` might be bigger than the size allocated in the attribute store for this attribute, since it's sized to the _max_ attribute size.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
bzbarsky-apple, line 354: This internal low-level API is skipping all the min/max checks that are supposed to be done.  What's doing those checks instead?

This is also skipping the optimization to not report if the value is not changing, skipping all the pre/post ember callbacks, skipping ember handling of NVM attributes, and all of this code is skipping the `MatterReportingAttributeChangeCallback` calls that might need to happen, right?  What's supposed to handle all of this?

bzbarsky-apple (changes requested):

jmartinez-silabs (approved):

src/app/codegen-data-model/CodegenDataModel_Read.cpp
```diff
@@ -282,7 +238,7 @@ CHIP_ERROR EncodeEmberValue(ByteSpan data, const EmberAfAttributeMetadata * meta
         return EncodeStringLike<ByteSpan, LongPascalString>(data, isNullable, encoder);
     default:
         ChipLogError(DataManagement, "Attribute type 0x%x not handled", static_cast<int>(metadata->attributeType));
-        return CHIP_IM_GLOBAL_STATUS(UnsupportedRead);
+        return CHIP_IM_GLOBAL_STATUS(ConstraintError);
```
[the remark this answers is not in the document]
tcarmelveilleux: UnsupportedRead was certainly not correct. Perhaps Failure is what it should be...

src/app/codegen-data-model/EmberMetadata.cpp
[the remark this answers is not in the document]
tcarmelveilleux: > If we wanted to optimize more, we could have an IsGlobalAttribute() check on the id before we start doing the "not in metadata" check or something....

That is an easy compromise to avoid re-ordering and could be done in this PR.

src/app/codegen-data-model/CodegenDataModel_Read.cpp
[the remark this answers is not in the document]
bzbarsky-apple: Failure would be fine.  But note that UnsupportedRead matches the current ember-compatibility-functions behavior, so we should change both places, probably.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,367 @@
...
+        return DecodeStringLikeIntoEmberBuffer<CharSpan, ShortPascalString>(decoder, isNullable, out);
+    case ZCL_LONG_CHAR_STRING_ATTRIBUTE_TYPE:
+        return DecodeStringLikeIntoEmberBuffer<CharSpan, LongPascalString>(decoder, isNullable, out);
+    case ZCL_OCTET_STRING_ATTRIBUTE_TYPE: // Octet string
+        return DecodeStringLikeIntoEmberBuffer<ByteSpan, ShortPascalString>(decoder, isNullable, out);
+    case ZCL_LONG_OCTET_STRING_ATTRIBUTE_TYPE:
+        return DecodeStringLikeIntoEmberBuffer<ByteSpan, LongPascalString>(decoder, isNullable, out);
+    default:
+        ChipLogError(DataManagement, "Attribute type 0x%x not handled", static_cast<int>(metadata->attributeType));
+        return CHIP_IM_GLOBAL_STATUS(ConstraintError);
```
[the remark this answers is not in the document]
tcarmelveilleux: `Failure` is better I think

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,367 @@
...
+
+    AttributeAccessInterface * aai       = GetAttributeAccessOverride(request.path.mEndpointId, request.path.mClusterId);
+    std::optional<CHIP_ERROR> aai_result = TryWriteViaAccessInterface(request.path, aai, decoder);
+
+    if (aai_result.has_value())
+    {
+        return *aai_result;
+    }
+
+    ReturnErrorCodeIf(aai_result.has_value(), *aai_result);
```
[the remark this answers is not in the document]
tcarmelveilleux: This is because there was a suggested snippet in a previous comment but it looks like the original code was retained rather than removed. 

Please remove this dupe :)

src/app/codegen-data-model/CodegenDataModel_Read.cpp
```diff
@@ -138,13 +89,22 @@ struct ShortPascalString
...
 /// Metadata of what a ember/pascal LONG string means (prepended by a u16 length)
 struct LongPascalString
 {
     using LengthType                        = uint16_t;
     static constexpr LengthType kNullLength = 0xFFFF;
+
+    static LengthType GetLength(const uint8_t * buffer)
+    {
+        // NOTE: we do NOT use emberAfLongStringLength because that will result in 0 length
+        //       for null strings
```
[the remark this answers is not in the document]
andy31415: Because I started first and then noticed the error. Updated.

src/app/codegen-data-model/tests/TestCodegenModelViaMocks.cpp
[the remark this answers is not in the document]
andy31415: Acknowledged. For what it's worth it passes and covers >95%, so likely decent (but extremely convoluted ... this is a lot of things to test and lots of attribute variations)

src/app/codegen-data-model/CodegenDataModel_Write.cpp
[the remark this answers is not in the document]
andy31415: Removed the duplication.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,367 @@
...
+static_assert(sizeof(LongPascalString::LengthType) == 2);
+
+/// Encode the value stored in 'decoder' into an ember format string 'out'
+///
+/// The value encoded will be of type T (e.g. CharSpan or ByteSpan) and it will be encoded
+/// via the given ENCODING (i.e. ShortPascalString or LongPascalString)
+///
+/// isNullable defines if the value of NULL is allowed to be encoded.
+template <typename T, class ENCODING>
+CHIP_ERROR DecodeStringLikeIntoEmberBuffer(AttributeValueDecoder decoder, bool isNullable, MutableByteSpan out)
```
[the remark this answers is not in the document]
andy31415: Updated the comments.

src/app/codegen-data-model/CodegenDataModel_Read.cpp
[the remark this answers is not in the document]
andy31415: Switched to Failure (and also updated ember-compatibility-functions then ... realistically we should never hit this piece of code).

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,367 @@
...
+/// isNullable defines if the value of NULL is allowed to be encoded.
+template <typename T>
+CHIP_ERROR DecodeIntoEmberBuffer(AttributeValueDecoder & decoder, bool isNullable, MutableByteSpan out)
+{
+    using Traits = NumericAttributeTraits<T>;
+    typename Traits::StorageType storageValue;
+
+    if (isNullable)
+    {
+        DataModel::Nullable<typename Traits::WorkingType> workingValue;
```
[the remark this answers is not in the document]
andy31415: I need `workingValue` even after  the null check, so that I can do the range check and working2storage call.

I am unsure if there is a clean way to make this shorter/clearer.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,367 @@
...
+        ReturnErrorOnFailure(decoder.Decode(workingValue));
+
+        if (workingValue.IsNull())
+        {
+            Traits::SetNull(storageValue);
+        }
+        else
+        {
+            // This guards against trying to encode something that overlaps nullable, for example
```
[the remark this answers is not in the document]
andy31415: Updated the comment. I had quite a bit of copy&paste on these generic functions.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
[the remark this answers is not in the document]
andy31415: Updated to `Failure`

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,367 @@
...
+    VerifyOrReturnError(attributeMetadata != nullptr, CHIP_IM_GLOBAL_STATUS(UnsupportedWrite));
+
+    if (request.path.mDataVersion.HasValue())
+    {
+        std::optional<InteractionModel::ClusterInfo> clusterInfo = GetClusterInfo(request.path);
+        if (!clusterInfo.has_value())
+        {
+            ChipLogError(DataManagement, "Unable to get cluster info for Endpoint %x, Cluster " ChipLogFormatMEI,
+                         request.path.mEndpointId, ChipLogValueMEI(request.path.mClusterId));
+            return CHIP_ERROR_NOT_FOUND;
```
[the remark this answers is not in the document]
andy31415: Updated.

src/app/codegen-data-model/EmberMetadata.cpp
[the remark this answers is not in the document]
andy31415: Added a IsGlobalAttribute check before the loop.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
andy31415: Updated to use the emAfWriteAttributeExternal for non-internal requests, which will also do size checks based on metadata. This preserves existing logic, however is probably non-optimal (we keep searching for metadata several times).

src/app/codegen-data-model/CodegenDataModel_Write.cpp
andy31415: From what I can see, typeSensitiveMemCopy at https://github.com/project-chip/connectedhomeip/blob/master/src/app/util/attribute-storage.cpp#L486

sets the bufferSize to am-size if write is true.

This seems to be inline with the documentation of emAfRedorWriteAttribute at https://github.com/project-chip/connectedhomeip/blob/master/src/app/util/attribute-storage.cpp#L568

```
// When writing attributes, readLength is ignored.  For non-string attributes,
// this function assumes the source buffer is the same size as the attribute
// type.  For strings, the function will copy as many bytes as will fit in the
// attribute.  This means the resulting string may be truncated.  The length
// byte(s) in the resulting string will reflect any truncated.
```

I will update comments here.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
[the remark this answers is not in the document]
bzbarsky-apple: I guess I was envisioning a "working to storage" helper that is fallible that handles the range check and the WorkingToStorage call.

But maybe that's not worth it.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,381 @@
...
+    if (aai_result.has_value())
+    {
+        return *aai_result;
+    }
+
+    ReturnErrorOnFailure(DecodeValueIntoEmberBuffer(decoder, *attributeMetadata, gEmberAttributeIOBufferSpan));
+
+    Protocols::InteractionModel::Status status;
+
+    if (!request.operationFlags.Has(InteractionModel::OperationFlags::kInternal))
```
bzbarsky-apple, line 349: Either this test or the code below is backwards.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,381 @@
...
+        // Specifically `When writing attributes, readLength is ignored.`.
+        // The actual length is used by unit tests to copy over data without needing "get attribute size" logic
+        // since metadata attribute sizes in our mocks is not generally correct.
+        status = emAfReadOrWriteAttribute(&record, attributeMetadata, /* buffer = */ gEmberAttributeIOBufferSpan.data(),
+                                          /* readLength = */ static_cast<uint16_t>(gEmberAttributeIOBufferSpan.size()),
+                                          /* write = */ true);
+    }
+    else
+    {
+        status = emAfWriteAttributeExternal(request.path.mEndpointId, request.path.mClusterId, request.path.mAttributeId,
```
bzbarsky-apple, line 368: Again, what happened to the "is the data we are writing larger than attributeMetadata->size" check?  emAfWriteAttributeExternal can't do that check, because it does not have the right information (e.g. the size of the data we are writing).  This will silently truncate too-long strings instead of correctly erroring out, as far as I can tell.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,381 @@
...
+        record.attributeId = request.path.mAttributeId;
+        // NOTE: gEmberAttributeIOBufferSpah is likely much larger than the attribute itself. Ember is supposed
+        //       to be able to pick up the "valid size" out of this
+        //
+        // Specifically `When writing attributes, readLength is ignored.`.
+        // The actual length is used by unit tests to copy over data without needing "get attribute size" logic
+        // since metadata attribute sizes in our mocks is not generally correct.
+        status = emAfReadOrWriteAttribute(&record, attributeMetadata, /* buffer = */ gEmberAttributeIOBufferSpan.data(),
+                                          /* readLength = */ static_cast<uint16_t>(gEmberAttributeIOBufferSpan.size()),
+                                          /* write = */ true);
```
bzbarsky-apple, line 364: OK, so this "internal" write is, again:

1. Skipping notifying that the value changed, for subscription reporting.  Is that expected?
2. Skipping pre/post write callbacks (maybe OK).
3. Skipping NVM handling (probably not ok).

bzbarsky-apple (changes requested):

src/app/codegen-data-model/CodegenDataModel_Write.cpp
andy31415: My goal is to have the ability to have the Accessors.cpp equivalent. Updated to use emberAfWriteAttribute since that seems used there.

I did not even check at the beginning as this seemed awfully asymmetric - the read uses `emAfReadOrWriteAttribute` in compatibility functions, where as the writes seem to have different wrappers. Accessors.cpp seems to use `emberAfReadAttribute` for reading, however that seems to be just a thin wrapper around emAfReadOrWriteAttribute + an extra log for resource exhausted.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,381 @@
...
+                         request.path.mEndpointId, ChipLogValueMEI(request.path.mClusterId));
+            return CHIP_IM_GLOBAL_STATUS(DataVersionMismatch);
+        }
+    }
+
+    AttributeAccessInterface * aai       = GetAttributeAccessOverride(request.path.mEndpointId, request.path.mClusterId);
+    std::optional<CHIP_ERROR> aai_result = TryWriteViaAccessInterface(request.path, aai, decoder);
+    if (aai_result.has_value())
+    {
+        return *aai_result;
```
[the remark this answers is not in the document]
andy31415: Added the call. Here I am going into unused territory: I would expect  IME to eventually be responsible to call the correct 'mark dirty' calls since it knows if a "success write" ocurred or not.

I added mark-dirty (and callbacks) calls inside our logic and updated unit tests, however I believe this also may be in flux a bit over time.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
andy31415: Added that one back with a unit test.

This was really painful as our unit test fixtures did not really set sizes for attributes, however I do want the external paths in unit tests because of ACL and writability tests.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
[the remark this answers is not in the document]
andy31415: I believe this implementation is esentially it: it does conversion to storage with range checks. If there are ideas to make the code cleaner, I would prefer as a separate update.

src/app/codegen-data-model/CodegenDataModel_Read.cpp
```diff
@@ -90,7 +89,12 @@ struct ShortPascalString
     using LengthType                        = uint8_t;
     static constexpr LengthType kNullLength = 0xFF;
 
-    static LengthType GetLength(const uint8_t * buffer) { return emberAfStringLength(buffer); }
+    static LengthType GetLength(const uint8_t * buffer)
+    {
+        // NOTE: we do NOT use emberAfLongStringLength because that will result in 0 length
+        //       for null strings
+        return *buffer;
```
tcarmelveilleux, line 96: Suggest `return static_cast<LengthType>(buffer[0]);`, it's much clearer as to what is being done.

andy31415: This is specific to exactly 1 byte in the class. You are converting a 7-char `*buffer` into a 34-char that is equivalent. I don't see the point.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -321,45 +332,68 @@ CHIP_ERROR CodegenDataModel::WriteAttribute(const InteractionModel::WriteAttribu
...
 
-    EmberAfAttributeSearchRecord record;
-    record.endpoint    = request.path.mEndpointId;
-    record.clusterId   = request.path.mClusterId;
-    record.attributeId = request.path.mAttributeId;
+    if (request.operationFlags.Has(InteractionModel::OperationFlags::kInternal))
+    {
+        // Internal requests use the non-External interface that has less enforcement
+        // than the external version (e.g. does not check/enforce writable settings, does not
+        // validate atribute types) - see attribute-table.h documentation for details.
```
tcarmelveilleux, line 372: nit: "attribute"

src/app/codegen-data-model/CodegenDataModel_Read.cpp
andy31415: used buffer[0] since I converted things to ByteSpan. I do not see why we need static_cast.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -321,45 +332,68 @@ CHIP_ERROR CodegenDataModel::WriteAttribute(const InteractionModel::WriteAttribu
...
+        status = emberAfWriteAttribute(request.path.mEndpointId, request.path.mClusterId, request.path.mAttributeId,
+                                       gEmberAttributeIOBufferSpan.data(), (*attributeMetadata)->attributeType);
+    }
+    else
+    {
+        if (dataBuffer.size() > (*attributeMetadata)->size)
+        {
+            ChipLogDetail(Zcl, "Data to write exceedes the attribute size claimed.");
```
tcarmelveilleux, line 380: nit: "exceeds".

Also, why is it OK to do this check on non-internal, but expect internal clients to do the right thing? When is it that `dataBuffer.size() > (*attributeMetadata)->size)` ? What about if `<` instead of `==`. The underlying conversion assumptions seem swallowed by the evil of `emAfWriteAttributeExternal` complex internal encoding rules. In other words, by looking at this code, it's really not clear why we would ever care to pre-check the dataBuffer.size() rather than let `emAfWriteAttributeExternal` tell us of errors if its assumptions are broken.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -321,45 +332,68 @@ CHIP_ERROR CodegenDataModel::WriteAttribute(const InteractionModel::WriteAttribu
...
+                                            gEmberAttributeIOBufferSpan.data(), (*attributeMetadata)->attributeType);
+    }
 
     if (status != Protocols::InteractionModel::Status::Success)
     {
         return CHIP_ERROR_IM_GLOBAL_STATUS_VALUE(status);
     }
 
+    // TODO: this may need more refinement:
+    //       - should internal requests be able to decide if something is marked dirty or not?
```
tcarmelveilleux, line 394: The answer is 100% yes, and this is critical to achieve the Q quality.

See https://github.com/project-chip/connectedhomeip/pull/31162

src/app/util/mock/MockNodeConfig.cpp
```diff
@@ -26,6 +26,62 @@
...
+        return kDefaultStringSize;
+    case ZCL_ARRAY_ATTRIBUTE_TYPE:
+    case ZCL_STRUCT_ATTRIBUTE_TYPE:
+        // These items are using AAI and are acceptable for tests
+        return 0;
+    default:
+        // this type of attribute is not supported for tests, we cannot guess its size
+        //
+        // see attribute-type.h for a list of attributes
+        ChipLogError(Test, "Warning: size for attribute type 0x%x is not set", type) return 0;
```
tcarmelveilleux, line 80: If this is used, could it lead to errors? If so, instead of logging, why not verify or die?

src/app/codegen-data-model/CodegenDataModel_Write.cpp
andy31415: This is copied over from https://github.com/project-chip/connectedhomeip/blob/master/src/app/util/ember-compatibility-functions.cpp#L798 (including the typo). I fixed the typo here, however I would prefer not to try to expand the scope of the PR with altering other existing logic. It is very large as is.

The ">" would be when you receive a string that is larger than what ember allocated in it's attribute store. so Say a char_string<4> would have 5 bytes allocated for ember (length + 4 chars, no null terminator). As a result, writhing "a" or "ab" is ok and the size is shorter. writing "abcd" fits exactly (the == case) and "abcdef" will result in the error here.

I guess TLDR: I am maintaining functionality of ember-compatibility-functions.cpp here, so all stays reasonably similar unless I have strong reasons to do it differently. I cannot actually reference "ember-compatibiltiy-functions.cpp" in comments because I expect that file will eventually be deleted if we integrate this code (as it does the same thing)

src/app/codegen-data-model/CodegenDataModel_Write.cpp
andy31415: yes, however we currently do not have this capability in ember (we do have non-dirty sets in accessors, so we have to make that work, however reality is we should just support the Q quality).

The point of this todo was that this will need more work, however in the current infrastructure is it not possible, until we integrate this code and can change both sides of the code (accessors and this).

src/app/util/mock/MockNodeConfig.cpp
andy31415: I had VerifyOrDie first, however we do have a copy&paste of a lot of attribute types (all attribute types) in our unit test code, so for now I let it log so that tests pass without crashing on static init.

Making this VerifyOrDie requires me to do more yak shaving across tests that use mocks.

src/app/util/mock/MockNodeConfig.h
```diff
@@ -31,6 +31,12 @@ namespace Test {
 
 namespace internal {
 
+constexpr uint16_t kDefaultStringSize = 16; // note: this is INCLUDING the length byte(s)
+
+// Determine an appropriate size for the given type.
+// NOTE: this is for test only, not all types are included
+uint16_t SizeForType(EmberAfAttributeType type);
```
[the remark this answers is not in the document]
andy31415: The structure type in ember is uint16_t. https://github.com/project-chip/connectedhomeip/blob/master/src/app/util/attribute-metadata.h#L148

making this size_t would require an extra cast when setting and some comments on why that is the right thing to do. u16 seems cleaner.

tcarmelveilleux (approved):

src/app/codegen-data-model/CodegenDataModel_Read.cpp
```diff
@@ -138,13 +88,30 @@ struct ShortPascalString
 {
     using LengthType                        = uint8_t;
     static constexpr LengthType kNullLength = 0xFF;
+
+    static size_t GetLength(ByteSpan buffer)
+    {
+        VerifyOrDie(buffer.size() >= 1);
+        // NOTE: we do NOT use emberAfLongStringLength because that will result in 0 length
+        //       for null sizes (i.e. 0xFF is translated to 0 and we do not want that here)
```
bzbarsky-apple, line 96: We don't use emberAfLongStringLength because it ... would just be the wrong thing to use for a _short_ string as here.

Please fix this comment.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,401 @@
...
+    {
+        // Internal requests use the non-External interface that has less enforcement
+        // than the external version (e.g. does not check/enforce writable settings, does not
+        // validate attribute types) - see attribute-table.h documentation for details.
+        status = emberAfWriteAttribute(request.path.mEndpointId, request.path.mClusterId, request.path.mAttributeId,
+                                       gEmberAttributeIOBufferSpan.data(), (*attributeMetadata)->attributeType);
+    }
+    else
+    {
+        if (dataBuffer.size() > (*attributeMetadata)->size)
```
bzbarsky-apple, line 378: So for "internal" writes we're OK with silent truncation of too-long strings?  Does something document that behavior?

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,401 @@
...
+
+    if (status != Protocols::InteractionModel::Status::Success)
+    {
+        return CHIP_ERROR_IM_GLOBAL_STATUS_VALUE(status);
+    }
+
+    // TODO: this may need more refinement:
+    //       - should internal requests be able to decide if something is marked dirty or not?
+    //       - changes-omitted paths should not be marked dirty (ember is not aware of these)
+    CurrentContext().dataModelChangeListener->MarkDirty(request.path);
```
bzbarsky-apple, line 396: This looks wrong.  `emberAfWriteAttribute` will not mark things dirty if the value is not actually changing.  At least in some cases.

Followup, not for this PR, since the MarkDirty() stuff is currently all a no-op, I assume.  But this needs some thought about what the API will _actually_ look like, and the comment should be very clear that this _definitely_ needs "refinement".

src/app/codegen-data-model/CodegenDataModel_Write.cpp
```diff
@@ -0,0 +1,401 @@
...
+
+    Protocols::InteractionModel::Status status;
+
+    if (request.operationFlags.Has(InteractionModel::OperationFlags::kInternal))
+    {
+        // Internal requests use the non-External interface that has less enforcement
+        // than the external version (e.g. does not check/enforce writable settings, does not
+        // validate attribute types) - see attribute-table.h documentation for details.
+        status = emberAfWriteAttribute(request.path.mEndpointId, request.path.mClusterId, request.path.mAttributeId,
+                                       gEmberAttributeIOBufferSpan.data(), (*attributeMetadata)->attributeType);
```
bzbarsky-apple, line 374: Also, shouldn't we be using `dataBuffer.data()` in some of these places?  The use of `gEmberAttributeIOBufferSpan` when `dataBuffer` is what we decoded the value into is a little weird, even though they do happen to point to the same data....

src/app/codegen-data-model/CodegenDataModel_Write.cpp
andy31415: This is what Accessors.h does today

src/app/codegen-data-model/CodegenDataModel_Write.cpp
andy31415: Re-worded comment and added details about what to consider regarding MatterReportingAttributeChangeCallback.

It seems a bit annoying that AAI always calls it (regardless of content) where as ember has knowledge of changed or not. This needs more iteration to make it consistent.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
andy31415: Yes, forgot to properly update.

src/app/codegen-data-model/CodegenDataModel_Write.cpp
bzbarsky-apple: @andy31415 Yes, we probably need to change the AAI API to allow the impl to say "actually, nothing changed" for impls that want to optimize.

bzbarsky-apple (approved):

PR #34142, opened 2024-06-29
Minor fixes in TCP Disconnect.

pidarped:
Use the member PeerAddress in the ActiveConnection object to compare with the passed PeerAddress instead of calling GetPeerInfo() from TCPEndPoint.

Log the connection closure in the CloseConnectionInternal function which is also called by CloseActiveConnections() from the TCPBase destructor.

### DIFF
--- a/src/transport/raw/TCP.cpp
+++ b/src/transport/raw/TCP.cpp
@@ -398,6 +398,10 @@ void TCPBase::CloseConnectionInternal(ActiveTCPConnectionState * connection, CHI
 
     if (connection->mConnectionState != TCPState::kClosed && connection->mEndPoint)
     {
+        char addrStr[Transport::PeerAddress::kMaxToStringSize];
+        connection->mPeerAddr.ToString(addrStr);
+        ChipLogProgress(Inet, "Closing connection with peer %s.", addrStr);
+
         if (err == CHIP_NO_ERROR)
         {
             connection->mEndPoint->Close();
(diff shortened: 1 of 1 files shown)


yufengwangca (approved):

src/transport/raw/TCP.cpp
```diff
@@ -616,36 +620,17 @@ CHIP_ERROR TCPBase::TCPConnect(const PeerAddress & address, Transport::AppTCPCon
...
     {
         if (mActiveConnections[i].IsConnected())
         {
-            Inet::IPAddress ipAddress;
-            uint16_t port;
-            Inet::InterfaceId interfaceId;
+            const Inet::IPAddress & ipAddress = mActiveConnections[i].mPeerAddr.GetIPAddress();
+            uint16_t port                     = mActiveConnections[i].mPeerAddr.GetPort();
 
-            err = mActiveConnections[i].mEndPoint->GetPeerInfo(&ipAddress, &port);
```
bzbarsky-apple, line 629: The commit message (and PR description) should explain why this change is being made.  Is the peer info in `mEndpoint` the same as `mPeerAddr`?  If they are, this is just code simplification, I guess? If they are not, why are we using one instead of the other?

This all needs to be in the commit message.

src/transport/raw/TCP.cpp
```diff
@@ -616,36 +620,17 @@ CHIP_ERROR TCPBase::TCPConnect(const PeerAddress & address, Transport::AppTCPCon
...
-                return;
-            }
-
-            err = mActiveConnections[i].mEndPoint->GetInterfaceId(&interfaceId);
-            if (err != CHIP_NO_ERROR)
-            {
-                ChipLogError(Inet, "TCPDisconnect: GetInterfaceId error: %" CHIP_ERROR_FORMAT, err.Format());
-                return;
-            }
             // if (address == PeerAddress::TCP(ipAddress, port, interfaceId))
```
bzbarsky-apple, line 631: This comment does not make sense: it does not match the code.

What we are actually doing here is ignoring the address type (TCP vs UDP) and ignoring the interface ID, right?  As in, this should be documenting why we are not in fact using `PeerAddress::operator==`.

bzbarsky-apple (changes requested):

src/transport/raw/TCP.cpp
pidarped: In a sense, this is a simplification. The IPAddress and Port in mEndPoint is the same as that in mPeerAddr inside the corresponding ActiveConnection object which is readily available. Will add to the comments.

src/transport/raw/TCP.cpp
pidarped: Needed to remove this. It was a lingering codeline from before. 
Yes, we are ignoring the InterfaceID because the IPAddress and Port are the minimal sufficient set of parameters for the connection. Furthermore, the interface information may not be there because it was not supplied during connection establishment, and so might pose an issue during comparison. Can add more comments on this to clarify.

bzbarsky-apple (approved):

PR #34154, opened 2024-07-02
Disentangle data model tests by moving fixtures into their own file

ksperling-apple:
The data model tests all had linker dependencies on each other, which meant that running any of Test{Read,Write,Commands} would run the tests from all of them, because the global fixture function from that test was required.

This change moves all those fixture functions into DataModelFixtures.cpp. Also disambiguate the enum types and control variable names involved.

### DIFF
--- a/src/app/util/mock/Constants.h
+++ b/src/app/util/mock/Constants.h
@@ -51,11 +51,22 @@ constexpr EventId MockEventId(const uint16_t & id)
     return (0xFFF1'0000 | id);
 }
 
+// Cluster Revision value returned by mock clusters
 extern const uint16_t mockClusterRevision;
+
+// Feature Map value returned by mock clusters
 extern const uint32_t mockFeatureMap;
+
+// Scalar value returned by mock clusters for MockAttributeId(1)
 extern const bool mockAttribute1;
+
+// Scalar value returned by mock clusters for MockAttributeId(2)
 extern const int16_t mockAttribute2;
+
+// Scalar value returned by mock clusters for MockAttributeId(3)
 extern const uint64_t mockAttribute3;
+
+// MockAttributeId(4) returns a list of octstr with this value
 extern const uint8_t mockAttribute4[256];
 
 } // namespace Test
--- a/src/app/util/mock/Constants.h
+++ b/src/app/util/mock/Constants.h
@@ -50,5 +50,13 @@ constexpr EventId MockEventId(const uint16_t & id)
 {
     return (0xFFF1'0000 | id);
 }
+
+extern const uint16_t mockClusterRevision;
+extern const uint32_t mockFeatureMap;
+extern const bool mockAttribute1;
+extern const int16_t mockAttribute2;
+extern const uint64_t mockAttribute3;
+extern const uint8_t mockAttribute4[256];
+
 } // namespace Test
 } // namespace chip
--- a/src/controller/tests/data_model/DataModelFixtures.h
+++ b/src/controller/tests/data_model/DataModelFixtures.h
@@ -16,11 +16,8 @@
  *    limitations under the License.
  */
 
-/**
- *  @file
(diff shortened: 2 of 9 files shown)


src/app/util/mock/Constants.h
```diff
@@ -50,5 +50,13 @@ constexpr EventId MockEventId(const uint16_t & id)
 {
     return (0xFFF1'0000 | id);
 }
+
+extern const uint16_t mockClusterRevision;
```
andy31415, line 54: These do not seem like constants. Having these in `Constants.h` seems odd.

src/controller/tests/data_model/DataModelFixtures.h
```diff
@@ -0,0 +1,98 @@
...
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+/**
+ *  @file
```
andy31415, line 20: please remove @file comments

src/controller/tests/data_model/DataModelFixtures.h
```diff
@@ -0,0 +1,98 @@
...
+enum class WriteResponseDirective
+{
+    kSendAttributeSuccess,
+    kSendAttributeError,
+    kSendMultipleSuccess,
+    kSendMultipleErrors,
+    kSendClusterSpecificSuccess,
+    kSendClusterSpecificFailure,
+};
+extern WriteResponseDirective gWriteResponseDirective;
```
andy31415, line 80: these seem to control how tests behave - changing this may have a sideffect for other tests.

Should we maybe use https://github.com/project-chip/connectedhomeip/blob/master/src/lib/support/Scoped.h to wrap these and force scoped changes only within tests?

src/app/util/mock/Constants.h
ksperling-apple: They are constant values returned by mock cluster attributes. I think it makes more sense to declare them as constants in the header that’s to repeat the literal values in the mock code and the test expectations — at least this way it’s easier to understand why the values match / the tests pass.

src/controller/tests/data_model/DataModelFixtures.h
ksperling-apple: I’m just trying to follow the existing convention… what’s the problem with file comments?

src/controller/tests/data_model/DataModelFixtures.h
ksperling-apple: These globals existed before but were only changeable from on of the test files each (though potentially affecting the other tests). I’ll check if all modifications fit with the scoped pattern.

bzbarsky-apple (approved):
Approving, pending @andy31415's comments being addressed.

src/app/util/mock/Constants.h
ksperling-apple: Added some comments to them.

src/controller/tests/data_model/DataModelFixtures.h
ksperling-apple: Replaced with a normal comment.

src/controller/tests/data_model/DataModelFixtures.h
ksperling-apple: I've made these use `ScopedChangeOnly` now. I added the ability to assign to a `ScopedChange` because some tests made further changes within the tests not directly scoped to a block.

I think this is in keeping with the `ScopedChange` pattern since the key thing is that the value is restored at the end of the scope; what happens within the scope is not important.

andy31415 (approved):

PR #34156, opened 2024-07-02
Drop obsolete comment about heap based secure session allocation

agners:
The current implementation doesn't support heap based allocation for secure sessions. The comment is misleading, so lets remove it.

### DIFF
--- a/src/lib/core/CHIPConfig.h
+++ b/src/lib/core/CHIPConfig.h
@@ -759,11 +759,6 @@
  *  - 1 reserved slot for CASEServer as a responder.
  *  - 1 reserved slot for PASE.
  *
- *  NOTE: On heap-based platforms, there is no pre-allocation of the pool.
- *  Due to the use of an LRU-scheme to manage sessions, the actual active
- *  size of the pool will grow up to the value of this define,
- *  after which, it will remain at or around this size indefinitely.
- *
  */
 #ifndef CHIP_CONFIG_SECURE_SESSION_POOL_SIZE
 #define CHIP_CONFIG_SECURE_SESSION_POOL_SIZE (CHIP_CONFIG_MAX_FABRICS * 3 + 2)


agners:
At least from what I can tell there is no heap based allocation for secure sessions right now, or do I miss something? :thinking: 

At least it does seem that the limit of 50 devices apply for a Linux built controller configuration.

@bzbarsky-apple in talks with others you seem to be the person to know?

tcarmelveilleux (approved):

arkq (approved):

src/lib/core/CHIPConfig.h
```diff
@@ -759,11 +759,6 @@
  *  - 1 reserved slot for CASEServer as a responder.
  *  - 1 reserved slot for PASE.
  *
- *  NOTE: On heap-based platforms, there is no pre-allocation of the pool.
- *  Due to the use of an LRU-scheme to manage sessions, the actual active
- *  size of the pool will grow up to the value of this define,
- *  after which, it will remain at or around this size indefinitely.
```
bzbarsky-apple, line 765: I don't understand why this comment is being removed.  CHIP_CONFIG_SECURE_SESSION_POOL_SIZE is used like so:

```
ObjectPool<SecureSession, CHIP_CONFIG_SECURE_SESSION_POOL_SIZE> mEntries;
```

Then we have:

```
#if CHIP_SYSTEM_CONFIG_POOL_USE_HEAP
...
class ObjectPool<T, N, ObjectPoolMem::kHeap> : public HeapObjectPool<T>
```

so the comment in fact correctly describes the behavior as far as I can tell.

bzbarsky-apple (changes requested):

src/lib/core/CHIPConfig.h
bzbarsky-apple: And the point is, in platforms that use heap pools:

1. The pool will start with no memory allocated.
2. It will then allocate up to CHIP_CONFIG_SECURE_SESSION_POOL_SIZE entries.
3. After that, entries will start getting evicted, not because of logic in the pool itself (HeapObjectPool ignores the N), but because of logic in the consumer of the pool.

On platforms that don't use heap pools, the pool starts off with CHIP_CONFIG_SECURE_SESSION_POOL_SIZE entries in RAM in a contiguous buffer, and then uses them to emplace sessions. Once the pool fills up we start evicting (as opposed to erroring out with "pool full", which is what the pure pool logic would do).

agners: Hm, I see, yeah I think I misread "grow up to the value of this define". I probably interpreted too much into "there is no pre-allocation" and assumed it would grow indefinitely. Sorry about that.

PR #34159, opened 2024-07-02
[NXP] Add some common platform changes

marius-alex-tache:
Changes:
-  Add StoreSoftwareUpdateCompleted API with empty implementation
- Add ICD flags in common config header
- Create CHIP_DEVICE_CONFIG_USE_ZEPHYR_BLE flag for BLE Zephry usage
- Enclose CHIPDevicePlatformEvent header sections under CHIP_DEVICE_CONFIG_USE_ZEPHYR_BLE
- Enclose lwip specific headers under CHIP_SYSTEM_CONFIG_USE_LWIP
- Add temporary flag CHIP_DEVICE_CONFIG_INIT_OT_PLAT_ALARM to condition otPlatAlarmInit call. Some platforms are doing this separately and two calls will break things.
- Add CHIP_DEVICE_CONFIG_PROCESS_BLE_IN_THREAD flag to override ProcessThreadActivity by extending its default implementation. Some platforms opt to do the BLE processing inside the Thread task, to avoid creating a separate BLE app task.
- Add support of boot reason.


andy31415 (commented):
Where are the "legacy" files used?

Adding new files without linking in source code feels off, especially if they are "legacy" (i.e. probably should not be used anymore)?

Why do we do this? If this is to preserve old copies, why is git history not sufficient for that?

andy31415 (changes requested):
I do not understand the "legacy" folder. The comment of `Add some files in legacy folder.` in the PR summary does not help me understand them either.

marius-alex-tache:
Hi, @andy31415. The folder is called legacy because these files are mainly used by k32w0 (which is based on a legacy SDK). I should have specified that these files will be used by some upcoming changes. I am waiting for @chapongatien's PR related to NXP matter support submodule to be merged. Once that is done, I can clean the k32w0 further and sync what we have internally. So this is just an intermediate step and the files will eventually be used. If it is clunky, I can remove this commit for now and just keep the changes to the common NXP part.

marius-alex-tache:
I've removed the commits. Will add them in the future.

andy31415 (approved):

doru91 (approved):

PR #21074, opened 2022-07-21
Ignore StartupMode when rebooting due to OTA.

bzbarsky-apple:
Fixes https://github.com/project-chip/connectedhomeip/issues/19272

This will only work on platforms that actually set boot reasons
properly (i.e. probably not Mac/Linux).

#### Problem
See #19272.

#### Change overview
Change StartupMode handling to ignore the attribute when rebooting for OTA.  This is extracted from https://github.com/project-chip/connectedhomeip/pull/20440

#### Testing
1. Made sure that normal restart of all-clusters-app on Mac picks up the StartupMode.
2. Hacked up the Darwin version of `DiagnosticDataProviderImpl::GetBootReason` to return `kSoftwareUpdateCompleted` and verified that in this case the pre-restart CurrentMode value sticks around instead of StartupMode overwriting it.

PR #34160, opened 2024-07-02
OnOffServer: ignore StartUpOnOff value after reboot due to OTA

j0tunn:
From the specification:
> 1.5.6.6. StartUpOnOff Attribute
> This attribute SHALL define the desired startup behavior of a device when it is supplied with power and this state SHALL be reflected in the OnOff attribute. If the value is null, the OnOff attribute is set to its previous value. Otherwise, the behavior is defined in the table defining StartUpOnOf­fEnum.
> **_This behavior does not apply to reboots associated with OTA. After an OTA restart, the OnOff attribute SHALL return to its value prior to the restart_**.

Related PR:  https://github.com/project-chip/connectedhomeip/pull/21074

### DIFF
--- a/src/app/clusters/on-off-server/on-off-server.cpp
+++ b/src/app/clusters/on-off-server/on-off-server.cpp
@@ -98,7 +98,8 @@ bool IsKnownEnumValue(EnumType value)
     return (EnsureKnownEnumValue(value) != EnumType::kUnknownEnumValue);
 }
 
-BootReasonType GetBootReason() {
+BootReasonType GetBootReason()
+{
     BootReasonType bootReason = BootReasonType::kUnspecified;
 
     CHIP_ERROR error = DeviceLayer::GetDiagnosticDataProvider().GetBootReason(bootReason);
@@ -561,8 +562,9 @@ Status OnOffServer::getOnOffValueForStartUp(chip::EndpointId endpoint, bool & on
     }
 
     bool currentOnOff = false;
-    status = Attributes::OnOff::Get(endpoint, &currentOnOff);
-    if (status != Status::Success) {
+    status            = Attributes::OnOff::Get(endpoint, &currentOnOff);
+    if (status != Status::Success)
+    {
         return status;
     }
 
--- a/src/app/clusters/on-off-server/on-off-server.cpp
+++ b/src/app/clusters/on-off-server/on-off-server.cpp
@@ -45,12 +45,15 @@
 #endif                                                               // MATTER_DM_PLUGIN_MODE_BASE
 
 #include <platform/CHIPDeviceLayer.h>
+#include <platform/DiagnosticDataProvider.h>
 #include <platform/PlatformManager.h>
 
 using namespace chip;
 using namespace chip::app::Clusters;
 using namespace chip::app::Clusters::OnOff;
 using chip::Protocols::InteractionModel::Status;
+
+using BootReasonType = GeneralDiagnostics::BootReasonEnum;
 
 namespace {
 
@@ -93,6 +96,21 @@ template <typename EnumType>
 bool IsKnownEnumValue(EnumType value)
 {
     return (EnsureKnownEnumValue(value) != EnumType::kUnknownEnumValue);
+}
+
+BootReasonType GetBootReason() {
+    BootReasonType bootReason = BootReasonType::kUnspecified;
+
+    CHIP_ERROR error = DeviceLayer::GetDiagnosticDataProvider().GetBootReason(bootReason);
+    if (error != CHIP_NO_ERROR)
+    {
+        ChipLogError(Zcl, "Unable to retrieve boot reason: %" CHIP_ERROR_FORMAT, error.Format());
+        bootReason = BootReasonType::kUnspecified;
+    }
+
+    ChipLogProgress(Zcl, "Boot reason: %u", to_underlying(bootReason));
+
+    return bootReason;
 }
 
 } // namespace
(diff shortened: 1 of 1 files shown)


src/app/clusters/on-off-server/on-off-server.cpp
```diff
@@ -500,7 +518,7 @@ void OnOffServer::initOnOffServer(chip::EndpointId endpoint)
 
         bool onOffValueForStartUp = false;
         Status status             = getOnOffValueForStartUp(endpoint, onOffValueForStartUp);
-        if (status == Status::Success)
+        if (status == Status::Success && GetBootReason() != BootReasonType::kSoftwareUpdateCompleted)
```
bzbarsky-apple, line 521: So if StartUpOnOff is true but OnOff is false and we do an OTA restart, what happens with the OnMode/CurrentMode bits below, and is it the right behavior?

Separately, what happens if OTA happens and then 3 days later a new endpoint that has an On/Off cluster is added (which will call this code, I believe)?  That said, that handling might have been odd already, since arguably in that case doing StartUpOnOff and OnMode is weird.  So a followup issue on this is probably best.

j0tunn: > So if StartUpOnOff is true but OnOff is false and we do an OTA restart, what happens with the OnMode/CurrentMode bits below, and is it the right behavior?

Good point! I've tried to minimize changes, but looks like it should be fixed too. I've updated the PR

> Separately, what happens if OTA happens and then 3 days later a new endpoint that has an On/Off cluster is added (which will call this code, I believe)? That said, that handling might have been odd already, since arguably in that case doing StartUpOnOff and OnMode is weird. So a followup issue on this is probably best.

For now the behaviour will be "leave as is". So onOff value will be set to its previous value. And because we have no previous value, it will be the default value.

src/app/clusters/on-off-server/on-off-server.cpp
```diff
@@ -541,35 +560,41 @@ Status OnOffServer::getOnOffValueForStartUp(chip::EndpointId endpoint, bool & on
...
+        onOffValueForStartUp = false; // Off
+        break;
+    case OnOff::StartUpOnOffEnum::kOn:
+        onOffValueForStartUp = true; // On
+        break;
+    case OnOff::StartUpOnOffEnum::kToggle:
+        onOffValueForStartUp = !currentOnOff;
+        break;
+    default:
+        // All other values 0x03- 0xFE are reserved - no action.
```
bzbarsky-apple, line 593: "No action" would mean `onOffValueForStartUp = currentOnOff`, no?

src/app/clusters/on-off-server/on-off-server.cpp
bzbarsky-apple: > For now the behaviour will be "leave as is". So onOff value will be set to its previous value.

No, it'll be set to the StartUpOnOff value, since the boot reason is "OTA".... That's the same as the behavior before your PR, though.  Again, this needs a followup issue.

j0tunn: Looks like we are talking about different things. I agree that this needs a followup issue, I'll create it. Just lets clarify.

**Before this PR**:
In case of any boot reason:
- `newOnOff = f(oldOnOff, StartUpOnOff)`
- if `newOnOff == true`: `CurrentMode = OnMode`

**After this PR**:
In case of OTA boot reason:
- `newOnOff = oldOnOff` (thats what I call "leave as is")
- if `newOnOff == true`: `CurrentMode = OnMode`

In case of any other boot reason:
- `newOnOff = f(oldOnOff, StartUpOnOff)`
- if `newOnOff == true`: `CurrentMode = OnMode`

So after this PR the behaviour in case of scenario "if OTA happens and then 3 days later a new endpoint that has an On/Off cluster is added" will be changed from `newOnOff = f(oldOnOff, StartUpOnOff)` to `newOnOff = oldOnOff` which in such case will be `newOnOff = defaultOnOff`.
The problem is that the spec doesn't  specify such behaviour. All we know is "StartUpOnOff Attribute
defines the desired startup behavior of a device when it is **supplied with power**" and probably shouldn't affect the behaviour of adding new endpoint to powered device.

Also could you please give an example of the device where such case is possible?

src/app/clusters/on-off-server/on-off-server.cpp
j0tunn: fixed

src/app/clusters/on-off-server/on-off-server.cpp
```diff
@@ -541,35 +560,42 @@ Status OnOffServer::getOnOffValueForStartUp(chip::EndpointId endpoint, bool & on
 {
     app::DataModel::Nullable<OnOff::StartUpOnOffEnum> startUpOnOff;
     Status status = Attributes::StartUpOnOff::Get(endpoint, startUpOnOff);
-    if (status == Status::Success)
+    if (status != Status::Success)
```
jmartinez-silabs, line 563: Nit VerifyOrReturnError

src/app/clusters/on-off-server/on-off-server.cpp
```diff
@@ -541,35 +560,42 @@ Status OnOffServer::getOnOffValueForStartUp(chip::EndpointId endpoint, bool & on
...
-    return status;
+
+    bool currentOnOff = false;
+    status            = Attributes::OnOff::Get(endpoint, &currentOnOff);
+    if (status != Status::Success)
+    {
+        return status;
+    }
+
+    if (startUpOnOff.IsNull() || GetBootReason() == BootReasonType::kSoftwareUpdateCompleted)
```
jmartinez-silabs, line 575: Reading the description of the spec you shared and your assumption that OTA doesn't apply to this condition then probably all reboots that aren't due brownouts should be added to this same logic

j0tunn: Probably should. But the spec describes only behaviour of OTA reboots.
That's not my assumption, that's all the quote from the spec.

src/app/clusters/on-off-server/on-off-server.cpp
bzbarsky-apple: > Also could you please give an example of the device where such case is possible?

A bridge, for starters.  But there are other possibilities too (e.g. a device that can have new hardware added to it when it's powered on).

bzbarsky-apple (approved):

src/app/clusters/on-off-server/on-off-server.cpp
j0tunn: > So a followup issue on this is probably best

https://github.com/project-chip/connectedhomeip/issues/34452

bauerschwan (approved):

PR #34162, opened 2024-07-02
AccountLogin Login/Logout command support

mthiesc:
[Problem]
* ContentAppPlatform should send AccountLogin::Login command after successfull commissioning if user was shown setupPIN prompt.
* Handling Login/Logout commands is currently not implemented in the Android.

[Solution]
* Call AccountLoginManager::HandleLogin if commissioning succeeded successfully with setupPIN prompt flow.
* Implement HandleLogin (and HandleLogout) for Android AccountLoginManager using the ContentAppCommandDelegate.

[Testing]
Tested using TV Android App and TV Linux App.


src/controller/CommissionerDiscoveryController.cpp
```diff
@@ -455,6 +460,10 @@ void CommissionerDiscoveryController::InternalHandleContentAppPasscodeResponse()
                     cd, Transport::PeerAddress::UDP(client->GetPeerAddress().GetIPAddress(), client->GetCdPort()));
                 return;
             }
+            // Store commissioner passcode as setup PIN (string/charspan).
+            // If this PIN is not empty it signals that user prompt was shown to enter PIN and AccountLogin::Login command has to be sent.
+            mCommissionerSetupPin = CharSpan::fromCharString(std::to_string(passcode));
```
chrisdecenzo, line 465: we already have local variable mPasscode so we should not need to capture this a second time (see calls to SetPasscode).

src/controller/CommissionerDiscoveryController.cpp
```diff
@@ -440,6 +442,9 @@ void CommissionerDiscoveryController::InternalHandleContentAppPasscodeResponse()
             }
             CharSpan rotatingIdSpan(rotatingIdBuffer, 2 * rotatingIdLength);
 
+            // Store rotating ID as tempAccountIdentifier to use in AccountLogin::Login payload later.
+            mTempAccountIdentifier = rotatingIdSpan;
```
chrisdecenzo, line 446: I suggest grabbing this from the client at the time when you need it rather than here since this only gets called when we get the passcode back from the content app. 

The platform should invoke the Login command after commissioning, regardless of whether the content app provided the passcode or not.

chrisdecenzo (commented):
please see comments and suggestions

src/controller/CommissionerDiscoveryController.cpp
bzbarsky-apple, line 446: This is use-after-return, no?

src/controller/CommissionerDiscoveryController.cpp
```diff
@@ -28,6 +28,7 @@
 #include <controller/CommissionerDiscoveryController.h>
 #include <platform/CHIPDeviceLayer.h>
 #include <setup_payload/AdditionalDataPayloadGenerator.h>
+#include <app/app-platform/ContentAppPlatform.h>
```
bzbarsky-apple, line 31: This is failing to compile.

src/controller/CommissionerDiscoveryController.cpp
bzbarsky-apple, line 465: I don't understand how this compiles either.  That's causing the `std::string` here to be converted to a `const char *` and how?

But again, this looks like a use-after-return.

src/controller/CommissionerDiscoveryController.h
```diff
@@ -38,6 +38,7 @@
 
 #if CHIP_DEVICE_CONFIG_ENABLE_COMMISSIONER_DISCOVERY
 
+using chip::CharSpan;
```
bzbarsky-apple, line 41: Please don't put `using` at toplevel in headers.  I know the code here was already doing that for some types.  It shouldn't be.

bzbarsky-apple (changes requested):

examples/tv-app/android/java/ContentAppCommandDelegate.cpp
```diff
@@ -153,6 +147,7 @@ Status ContentAppCommandDelegate::InvokeCommand(EndpointId epId, ClusterId clust
             return chip::Protocols::InteractionModel::Status::Failure;
         }
 
+        // Q: It would seem that this is the successfull return case (?)
```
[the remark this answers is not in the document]
mthiesc: Status 0 indeed means success, but I don't think this resolves my question because there does not seem to be a path where a 0 Status is returned:

`value[FAILURE_STATUS_KEY]` is only being returned on line 145, which is only being parsed if `value[FAILURE_KEY]` is non-empty - indicating that the Status value also will be non-zero because there is an error response. The Failure status should be one of the failure responses formatted [here](https://github.com/project-chip/connectedhomeip/blob/master/examples/tv-app/android/App/platform-app/src/main/java/com/matter/tv/server/service/ContentAppAgentService.java#L151).

This code looks wrong to me because InvokeCommand returns a Status but the command response is only parsed (to read a status field) if there in the failure case. The returned Status will always be a failure case, but only 'valid' if there was an actual failure case - which the caller can then never really know. In `AccountLoginManager::GetSetupPIN` the returned Status is ignored, and the response is parsed explicitly with a call to `mCommandDelegate->FormatGetSetupPINResponse` to parse the response object. But this is only valid to do if InvokeCommand returns successfully - which it seems like it will never do. 

It seems like the Status returned from InvokeCommand on line [156](https://github.com/project-chip/connectedhomeip/blob/master/examples/tv-app/android/java/ContentAppCommandDelegate.cpp#L156) in the current main branch could/should be Success, to indicate that the command returned a non-failure response, which can then be parsed by the caller. But it returns Failure status (UnSupportedEndpoint) instead.

I'm hesistant to start changing things here before I can get confirmed if there is something I'm missing or if indeed this code is ambiguous.

examples/tv-app/android/java/ContentAppCommandDelegate.cpp
```diff
@@ -107,6 +100,7 @@ void ContentAppCommandDelegate::InvokeCommand(CommandHandlerInterface::HandlerCo
 Status ContentAppCommandDelegate::InvokeCommand(EndpointId epId, ClusterId clusterId, CommandId commandId, std::string payload,
                                                 bool & commandHandled, Json::Value & value)
 {
+    // Q1: It seems like this InvokeCommand *never* returns success status.
```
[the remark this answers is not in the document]
mthiesc: Please see my comment above to help resolve this issue.

src/controller/CommissionerDiscoveryController.cpp
mthiesc: But is this the same passcode? This case is when the mPasscode == 0 and this is storing the new passcode returned from GetCommissionerPasscode. Or are you saying it's OK to reuse mPasscode to store this new value?

mthiesc: Yes, I think the better approach here is to just store the uint passcode, and convert on demand when necessary, to avoid user-after-return with the spans.

src/controller/CommissionerDiscoveryController.h
mthiesc: Yup, did it to align with the existing code, but would rather avoid it. As long as it's not a problem to be inconsistent with the current style, I will change it.

src/controller/CommissionerDiscoveryController.cpp
mthiesc: Yes, good catch. Will refactor in line with chrisdecenzo comments, but will keep this in mind when dealing with spans. Thanks.

src/controller/CommissionerDiscoveryController.cpp
```diff
@@ -606,6 +615,23 @@ void CommissionerDiscoveryController::CommissioningSucceeded(uint16_t vendorId,
     mVendorId  = vendorId;
     mProductId = productId;
     mNodeId    = nodeId;
+
+    // Send AccountLogin::Login command if user was prompted to enter setup PIN manually.
+    // Q: Is this the correct place to have this logic?
+    // Q: Is there an easier way call AccountLoginDelegate from here?
```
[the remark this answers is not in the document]
mthiesc: Interesting. To confirm my understanding, are you suggesting to add the call to `app->AddClientNode(subjectNodeId)` below the line for `app->AddClientNode(subjectNodeId)` and just make sure `ManageClientAccess` includes `tempAccountId` and `passcode` in the parameter list? Which would be seeded from [here](https://github.com/project-chip/connectedhomeip/blob/master/src/controller/CommissionerDiscoveryController.cpp#L612) in `CommissioningSucceeded`?

src/controller/CommissionerDiscoveryController.cpp
mthiesc: Are you saying we want to send the Login command on successfull commissioning in all cases - regardless if the setup PIN prompt was shown or not?

src/controller/CommissionerDiscoveryController.cpp
tcarmelveilleux: "avoiding use after return" is a mandatory/critical thing, indeed...

src/controller/CommissionerDiscoveryController.h
tcarmelveilleux: top-level `using` should practically never be used in public headers. If there are other usages, they should be removed/replaced rather than copied.

mthiesc: Thanks, good to know. I will remove all top-level `using` uses in this file to be consistent.

src/controller/CommissionerDiscoveryController.cpp
```diff
@@ -606,6 +615,23 @@ void CommissionerDiscoveryController::CommissioningSucceeded(uint16_t vendorId,
     mVendorId  = vendorId;
     mProductId = productId;
     mNodeId    = nodeId;
+
+    // Send AccountLogin::Login command if user was prompted to enter setup PIN manually.
+    // Q: Is this the correct place to have this logic?
```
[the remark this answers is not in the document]
mthiesc: Changed to use `ContentAppPlatform::MangeClientAccess` as you mentioned below.

src/controller/CommissionerDiscoveryController.cpp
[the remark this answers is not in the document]
mthiesc: Updated the code as described here. Please take a look to see if this is aligned with what you had in mind.

src/controller/CommissionerDiscoveryController.cpp
mthiesc: Removed this code in the refactoring. But please see updated question [here](https://github.com/project-chip/connectedhomeip/pull/34162/files#diff-828770158d3226d53556b9b513b02d618a1a493bedca2c8301a4a3ae3f90764cR639) after refactoring to clarify commissionee vs commissioner passcode.

src/controller/CommissionerDiscoveryController.cpp
mthiesc: Please see updated question regarding this [here](https://github.com/project-chip/connectedhomeip/pull/34162/files#diff-0619e2ba03ec92e4032b69b6a7c5704aee9b14df0da72d9664d8e3b8879ffb13R766)

mthiesc: Code removed, fixed.

src/controller/CommissionerDiscoveryController.cpp
mthiesc: Code removed, fixed.

src/controller/CommissionerDiscoveryController.h
mthiesc: Code removed, fixed.

src/app/app-platform/ContentAppPlatform.cpp
```diff
@@ -755,6 +756,19 @@ CHIP_ERROR ContentAppPlatform::ManageClientAccess(Messaging::ExchangeManager & e
...
                 app->AddClientNode(subjectNodeId);
+
+                auto tempAccountId = CharSpan::fromCharString(rotatingId.c_str());
+                auto setupPIN = CharSpan::fromCharString(std::to_string(passcode).c_str());
+                auto nodeId = MakeOptional(subjectNodeId); // Q: Is subjectNodeId correct NodeId to use?
+
+                // send login command to content app
+                if (!tempAccountId.empty() && !setupPIN.empty()) {
+                    // Q: Should we only do this if (commissioner) passcode is not 0, e.g when user was shown PIN prompt?
+                    auto status = app->GetAccountLoginDelegate()->HandleLogin(tempAccountId, setupPIN, nodeId);
```
mthiesc, line 767: @chrisdecenzo Would be very helpful to understand what we want to do here.

src/controller/CommissionerDiscoveryController.cpp
```diff
@@ -606,10 +627,23 @@ void CommissionerDiscoveryController::CommissioningSucceeded(uint16_t vendorId,
...
+    auto rotatingId = std::string{};
+    auto passcode = uint32_t{ 0 };
+
+    auto client = GetUDCClientState();
+    if (client != nullptr)
+    {
+        // Get rotating ID and cached (commissioner?) passcode to handle AccountLogin
+        rotatingId = getRotatingIdString(client);
+        // Q: Should we use mPasscode, client->GetCommissionerPasscode, or client->GetCachedCommissionerPasscode?
+        passcode = std::to_string(client->GetCachedCommissionerPasscode());
```
mthiesc, line 640: @chrisdecenzo Would be very helpful to understand how to correctly use cached passcodes here.

examples/tv-app/android/java/ContentAppCommandDelegate.cpp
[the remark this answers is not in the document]
chrisdecenzo: It looks to me like the only caller of this method is AccountLoginManager::GetSetupPin and it ignores the return value, so you may be right that it never returns success. As a result, its safe for you to make changes so that it returns a sane value.

src/app/app-platform/ContentAppPlatform.cpp
chrisdecenzo: I was thinking that we would add tempAccountId and setupPIN as arguments to the AddClientNode method and then move the HandleLogin call to within AddClientNode. That method is keeping track of all of the clients of the content app so each time one is added, the HandleLogin should be called (and each time one is removed, the HandleLogout would be called).

src/controller/CommissionerDiscoveryController.cpp
chrisdecenzo: I don't think this include is needed anymore

src/controller/CommissionerDiscoveryController.cpp
chrisdecenzo: mPasscode is the value used for commissioning so no need to call client->GetCachedCommissionerPasscode(). In fact, sometimes the commissioner passcode will be wrong since the user manually entered the passcode.

Simply pass mPasscode in to CommissioningCompleted

src/controller/CommissionerDiscoveryController.cpp
```diff
@@ -606,10 +627,23 @@ void CommissionerDiscoveryController::CommissioningSucceeded(uint16_t vendorId,
...
     mNodeId    = nodeId;
+
+    auto rotatingId = std::string{};
+    auto passcode = uint32_t{ 0 };
+
+    auto client = GetUDCClientState();
+    if (client != nullptr)
+    {
+        // Get rotating ID and cached (commissioner?) passcode to handle AccountLogin
+        rotatingId = getRotatingIdString(client);
```
chrisdecenzo, line 638: Since we always need rotating id at the end of commissioning now, feel free to cache it locally in the CommissionerDiscoveryController object (adjacent to mPasscode), so that you can cleanup some duplicate code in this file.

The first time we enter this flow is OnUserDirectedCommissioningRequest and so that's probably the best place to cache it.

src/app/app-platform/ContentAppPlatform.cpp
```diff
@@ -755,6 +756,19 @@ CHIP_ERROR ContentAppPlatform::ManageClientAccess(Messaging::ExchangeManager & e
             {
                 // notify content app about this nodeId
                 app->AddClientNode(subjectNodeId);
+
+                auto tempAccountId = CharSpan::fromCharString(rotatingId.c_str());
+                auto setupPIN = CharSpan::fromCharString(std::to_string(passcode).c_str());
```
bzbarsky-apple, line 761: This is use-after-free right here.

src/controller/CommissionerDiscoveryController.cpp
```diff
@@ -28,9 +28,37 @@
...
+#include <app/app-platform/ContentAppPlatform.h>
 
 #if CHIP_DEVICE_CONFIG_ENABLE_COMMISSIONER_DISCOVERY
 
+namespace {
+
+bool fillRotatingIdBuffer(UDCClientState* client, char[] buffer) {
+    auto rotatingIdLength = client->GetRotatingIdLength();
+    auto hexBufferByteCount = 2 * rotatingIdLength;
+    if (sizeof(buffer) < hexBufferByteCount) {
```
bzbarsky-apple, line 40: This code does not do the right thing (and I would expect it to fail to compile with -Werror).

bzbarsky-apple (dismissed):

src/controller/CommissionerDiscoveryController.cpp
```diff
@@ -606,10 +610,15 @@ void CommissionerDiscoveryController::CommissioningSucceeded(uint16_t vendorId,
...
+
     if (mPostCommissioningListener != nullptr)
     {
         ChipLogDetail(Controller, "CommissionerDiscoveryController calling listener");
-        mPostCommissioningListener->CommissioningCompleted(vendorId, productId, nodeId, exchangeMgr, sessionHandle);
+        // Qs:
+        // * Is it right to use mPasscode here?
+        // * Would it not be zero if HandleContentAppPasscodeResponse returns zero passcode?
+        // * Which client/app does the passcode refer to? Same question for Rotating ID.
+        mPostCommissioningListener->CommissioningCompleted(vendorId, productId, nodeId, GetRotatingIdSpan(), mPasscode, exchangeMgr, sessionHandle);
```
mthiesc, line 621: Open questions.

examples/tv-app/android/java/ContentAppCommandDelegate.cpp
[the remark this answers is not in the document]
mthiesc: Fixed in latest commit.

src/controller/CommissionerDiscoveryController.cpp
mthiesc: Fixed.

src/app/app-platform/ContentAppPlatform.cpp
mthiesc: I ended up not adding the logic to AddClientNode itself, just to keep that method as is. But would be simple to refactor if that is preferred.

src/controller/CommissionerDiscoveryController.cpp
mthiesc: Updated to use cached mPasscode. See open questions about it, however.

src/app/app-platform/ContentAppPlatform.cpp
mthiesc: The lifetimes should be valid here as far as I can tell, but either way this logic is removed in latest update.

src/controller/CommissionerDiscoveryController.cpp
mthiesc: Code removed in latest update.

examples/tv-app/android/java/ContentAppCommandDelegate.cpp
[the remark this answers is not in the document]
mthiesc: Fixed.

src/controller/CommissionerDiscoveryController.cpp
mthiesc: Updated to cache rotating ID in OnUserDirectedCommissioningRequest and use that value when calling HandleLogin, as suggested. Kept the other calculations of rotating ID as is just to be safe, in case the cache might be stale (?).

src/app/app-platform/ContentApp.h
```diff
@@ -135,7 +135,8 @@ class DLL_EXPORT ContentApp
                                                             uint16_t maxReadLength);
     Protocols::InteractionModel::Status HandleWriteAttribute(ClusterId clusterId, AttributeId attributeId, uint8_t * buffer);
 
-    void AddClientNode(NodeId clientNodeId);
+    // returns true only if new node is added
```
chrisdecenzo, line 138:
```suggestion
    // returns true only if new node is added. If node was added previously, then false is returned.
```

lazarkov (approved):

src/controller/CommissionerDiscoveryController.h
```diff
@@ -443,6 +445,11 @@ class CommissionerDiscoveryController : public chip::Protocols::UserDirectedComm
      */
     UDCClientState * GetUDCClientState();
 
+    /**
+     * Get the rotating ID string CharSpan view
+     */
+    inline chip::CharSpan GetRotatingIdSpan() { return chip::CharSpan(mRotatingId.data(), mRotatingId.size()); }
```
bzbarsky-apple, line 451: This seems like a pretty unsafe API, in the sense that the caller has no idea how long the span remains valid for....

lazarkov: Can you please suggest a change that would make the API safer?

_The logic relies on the fact that every time commissioning starts the value mRotatingId will be filled in and unless commissioning succeeds or fails it won't be overridden._

bzbarsky-apple: You could have an API that has a MutableCharSpan outparam and copies the data, for example.

mthiesc: I don't agree this API is unsafe or needs to be changed: The span is created on demand from the string and has a valid lifetime in the scope it is created and used - same as any span you might create. It is up to the caller to not (ab)use the span in an unsafe way by for example storing the reference out of scope. From the perspective of the caller I think this is predictable and as expected.

I don't think copying the buffer to a MutableCharSpan outparam is a better option. If the caller needs to persist the value outside the current scope the caller should have the responsibility to do this by copying to another buffer. There is no need to overcomplicate this API which just returns a span to the existing buffer. There might be an argument that you might want both APIs to exist, for convenience, but to support the current usage there is only a need to have the current API which constructs a span to the underlying string buffer.

bzbarsky-apple: >  and has a valid lifetime in the scope it is created and used

The lifetime is only valid until `mRotatingId` changes, no?  What guarantees that it won't change while the caller is working with the span, e.g. due to a side-effect from something else the caller does?  This API does not clearly define the lifetime guarantees anywhere that I can see.

mthiesc: I think there are two concerns here:

i) Is the API safe in the sense that the reference returned will be valid and there is no use after free from caller side.
ii) Is the API safe in the sense that the (referenced) value returned is 'valid' and referring to the current node being commissioned.

My understanding/assumption was the original question is regarding i), and which was addressed above.

Regarding ii) I think this is a reasonable concern, and addressed previously by @lazarkov:
> The logic relies on the fact that every time commissioning starts the value mRotatingId will be filled in and unless commissioning succeeds or fails it won't be overridden

The overall architecture of commissioning relies on this assumption, and that commissioning several nodes happens in a subsequent first-in/first-out manner, where the rotating ID cannot change during the commissioning flow. If this assumption does not hold true, I think there will have to be a substantial refactoring to solve for this and part of a larger discussion. Please shout out if you think this might actually be a problem!

If not, as long as the span returned is used responsibly within the calling scope, and not saved for some arbitrary time later, this usage is safe. And to be safe here, regarding both concerns, in the PostCommissioningListener(s) the incoming rotating ID span is cached as a string copy, and this API just returned a span to that cached value.

bzbarsky-apple: Again: if the API and its consumer rely on some invariants, those invariants _need to be documented in the API_.

So if the value is valid until commissioning succeeds or fails, that needs to be documented.  Because nothing right now tells the API consumer "don't fail commissioning while working with this span".

mthiesc: I updated the code to remove the class methods - there was never a reason for having a publicly callable API in the first place as this was just a convenience method used internally. Now the span is just created on the stack from the string member value.

chrisdecenzo (approved):

hawk248 (approved):

PR #34168, opened 2024-07-03
Darwin: Prohibit static initializers in Matter.framework

ksperling-apple:
Globals should either be "constinit" (i.e. use a constrexpr constructor) and trivially destructible, or use the `Global<>` / `AtomicGlobal<>` helpers.

### DIFF
--- a/src/messaging/ReliableMessageProtocolConfig.cpp
+++ b/src/messaging/ReliableMessageProtocolConfig.cpp
@@ -67,7 +67,7 @@ namespace {
 
 // This is not a static member of ReliableMessageProtocolConfig because the free
 // function GetLocalMRPConfig() needs access to it.
-// Use std::optional to avoid a static initializers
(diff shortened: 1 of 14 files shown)


src/messaging/ReliableMessageProtocolConfig.cpp
```diff
@@ -62,14 +67,15 @@ namespace {
 
 // This is not a static member of ReliableMessageProtocolConfig because the free
 // function GetLocalMRPConfig() needs access to it.
-Optional<ReliableMessageProtocolConfig> sDynamicLocalMPRConfig;
+// Use std::optional to avoid a static initializers
```
bzbarsky-apple, line 70:
```suggestion
// Use std::optional to avoid a static initializer
```

bzbarsky-apple (approved):

jmartinez-silabs (approved):

PR #34175, opened 2024-07-03
Code refactoring and scope optimization for some clusters.

shripad621git:
**Change Overview**
- The PR includes code refactoring and scope optimization for the below clusters:
   - temperature-control cluster
   - thermostat cluster
   - laundry-washers-control cluster
   - door-lock cluster
   - power-source-cluster
   
  **Testing** 
  - Tested the changes with esp32 all-clusters-app .
  - Tested commission and control and chip-tool commands specific to these 
     clusters,

### DIFF
--- a/src/app/clusters/door-lock-server/door-lock-server.cpp
+++ b/src/app/clusters/door-lock-server/door-lock-server.cpp
@@ -84,8 +84,6 @@ void SetDefaultDelegate(EndpointId endpoint, Delegate * delegate)
 } // namespace app
 } // namespace chip
 
-DoorLockServer DoorLockServer::instance;
-
 class DoorLockClusterFabricDelegate : public chip::FabricTable::Delegate
 {
     void OnFabricRemoved(const FabricTable & fabricTable, FabricIndex fabricIndex) override
@@ -101,7 +99,6 @@ class DoorLockClusterFabricDelegate : public chip::FabricTable::Delegate
         }
     }
 };
-static DoorLockClusterFabricDelegate gFabricDelegate;
 
 /**********************************************************
  * DoorLockServer public methods
@@ -109,7 +106,8 @@ static DoorLockClusterFabricDelegate gFabricDelegate;
 
 DoorLockServer & DoorLockServer::Instance()
 {
-    return instance;
+    static DoorLockServer sInstance;
+    return sInstance;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


src/app/clusters/door-lock-server/door-lock-server.cpp
```diff
@@ -101,15 +99,15 @@ class DoorLockClusterFabricDelegate : public chip::FabricTable::Delegate
...
-static DoorLockClusterFabricDelegate gFabricDelegate;
 
 /**********************************************************
  * DoorLockServer public methods
  *********************************************************/
 
 DoorLockServer & DoorLockServer::Instance()
 {
-    return instance;
+    static DoorLockServer sInstance;
```
Damian-Nordic, line 109: Why is this better than a global? It introduces a slight overhead due to lazy initialization of static local variables and C++-imposed thread safety of the initialization, which may be justified but what's really the goal of this change?

tcarmelveilleux (changes requested):
- [ ] All the changes would be better done by just moving the global to the anonymous namespace. Strongly recommend against just moving to statics inside the methods as it adds little value for methods which have no choice but to use a global (in the way they are done today).

There is no justification of which problem is being solved in this PR, as well...

shripad621git:
> * [ ]  All the changes would be better done by just moving the global to the anonymous namespace. Strongly recommend against just moving to statics inside the methods as it adds little value for methods which have no choice but to use a global (in the way they are done today).
> 
> There is no justification of which problem is being solved in this PR, as well...

@tcarmelveilleux , @Damian-Nordic  The change has been made to leverage memory optimizations for  the implementations supporting dynamic data model.In case of such implementations, linking against a static library includes the unused stuff in the binary drastically increasing the binary size. due to these global objects in the clusters. Reducing their scope to local does not affect the functionality in terms of operations involving these clusters except for the lazy initialization part and also helps to reduce on memory in case of dynamic data model implementations.

bzbarsky-apple:
> and also helps to reduce on memory in case of dynamic data model implementations

I'm not clear on exactly how this works.  The memory for those statics still needs to be pre-allocated, since they're not going to do heap allocation when the relevant function is called, right?  What am I missing?

shripad621git:
> > and also helps to reduce on memory in case of dynamic data model implementations
> 
> I'm not clear on exactly how this works. The memory for those statics still needs to be pre-allocated, since they're not going to do heap allocation when the relevant function is called, right? What am I missing?

In case of dynamic data model implementations, all the clusters need to be present in the build setup for creating or trying out any device type. In case of linking against a static library, some portion of the code of the clusters not related to the particular device type make into the binary due to these static global objects in the src/app/clusters.
It saved  around 5kB of binary size in case the scope of these global objects was reduced for the case when these clusters were unused.



> > and also helps to reduce on memory in case of dynamic data model implementations
> 
> I'm not clear on exactly how this works. The memory for those statics still needs to be pre-allocated, since they're not going to do heap allocation when the relevant function is called, right? What am I missing?

@bzbarsky-apple , In dynamic data model implementations, all clusters must be included in the build setup to create or test any device type. When linking against a static library, some code from unrelated clusters for a specific device type can be included in the binary due to static global objects in the src/app/clusters.
 
By reducing the scope of these global objects for unused clusters, a significant amount of flash can be saved.

PR #34181, opened 2024-07-03
[Fix] Possible fix for matter-test-scripts issue #228: Remove PICS from RVC tests

j-ororke:
Following changes were applied and used:
- Removed PICS from python RVC Clean and RVC Run tests
-- Replaced with commands and attributes gathered from DUT
- Added checking command line arguments for "--app-pid" to enable self.is_ci variable in RVC test scripts in place of PICS "PICS_SDK_CI_ONLY" check.

#### Testing


src/python_testing/TC_RVCCLEANM_2_1.py
```diff
@@ -84,19 +84,40 @@ async def test_TC_RVCCLEANM_2_1(self):
         self.endpoint = self.matter_test_config.endpoint
         self.mode_ok = self.matter_test_config.global_test_params['PIXIT.RVCCLEANM.MODE_CHANGE_OK']
         self.mode_fail = self.matter_test_config.global_test_params['PIXIT.RVCCLEANM.MODE_CHANGE_FAIL']
-        self.is_ci = self.check_pics("PICS_SDK_CI_ONLY")
-        if self.is_ci:
-            app_pid = self.matter_test_config.app_pid
-            if app_pid == 0:
-                asserts.fail("The --app-pid flag must be set when PICS_SDK_CI_ONLY is set")
+        app_pid = self.matter_test_config.app_pid
```
cecille, line 87: You can just leave this CI check in for now. The question of how to determine whether an app is running in the CI is fairly open, but this PICS code is kind of meta. The goal for these issues is to remove the PICS codes for things that are represented on the device already (clusters, features, attributes, commands etc). This one is fine.

src/python_testing/TC_RVCCLEANM_2_1.py
```diff
@@ -84,19 +84,40 @@ async def test_TC_RVCCLEANM_2_1(self):
...
+        if app_pid != 0:
```
cecille, line 88: I know this isn't your code, but I'm also not sure this is correct. If the flag isn't set, won't app_pid be None?

j-ororke: Hi Cecille,
So when I run this in the Raspberry pi Dev container it was returning a 0 for app_pid if --app-pid is not provided in the CLI arguments. 
In the matter_testing_support module it appears to set the possible variables to 0 if they are to be gathered by argparse and not found in the provided command line arguments:
- config.app_pid = 0 if args.app_pid is None else args.app_pid

src/python_testing/TC_RVCCLEANM_1_2.py
```diff
@@ -57,11 +57,16 @@ def pics_TC_RVCCLEANM_1_2(self) -> list[str]:
...
+        attributes = RVCClean_cluster.Attributes
+        RVCClean_attr_list = attributes.AttributeList
+        attribute_list = await self.read_single_attribute_check_success(endpoint=self.endpoint, cluster=RVCClean_cluster, attribute=RVCClean_attr_list)
+        supported_modes_attr_id = attributes.SupportedModes.attribute_id
+        current_mode_attr_id = attributes.CurrentMode.attribute_id
 
         self.print_step(1, "Commissioning, already done")
 
-        if self.check_pics("RVCCLEANM.S.A0000"):
+        if supported_modes_attr_id in attribute_list:
```
cecille, line 69: Q for the RVC folks - this is mandatory, right? Can we just remove this check?

src/python_testing/TC_RVCCLEANM_1_2.py
```diff
@@ -124,7 +129,7 @@ async def test_TC_RVCCLEANM_1_2(self):
             asserts.assert_true(has_vacuum_or_mop_mode_tag,
                                 "At least one ModeOptionsStruct entry must include either the Vacuum or Mop mode tag")
 
-        if self.check_pics("RVCCLEANM.S.A0001"):
+        if current_mode_attr_id in attribute_list:
```
cecille, line 132: ditto here

src/python_testing/TC_RVCCLEANM_2_1.py
```diff
@@ -87,16 +87,39 @@ async def test_TC_RVCCLEANM_2_1(self):
...
+        # Gathering Accepted and Generated Commands and associated ids
+        commands = RVCClean_cluster.Commands
+        RVCClean_accptcmd_list = attributes.AcceptedCommandList
+        RVCClean_gencmd_list = attributes.GeneratedCommandList
+        accepted_cmd_list = await self.read_single_attribute_check_success(endpoint=self.endpoint, cluster=RVCClean_cluster, attribute=RVCClean_accptcmd_list)
+        generated_cmd_list = await self.read_single_attribute_check_success(endpoint=self.endpoint, cluster=RVCClean_cluster, attribute=RVCClean_gencmd_list)
+        chg_mode_cmd_id = commands.ChangeToMode.command_id
+        chg_rsp_cmd_id = commands.ChangeToModeResponse.command_id
+
+        if supported_modes_attr_id not in attribute_list:
```
cecille, line 112: same q here - these are all mandatory, would it make sense to just remove these given that we have conformance checks now?

cecille (commented):
General Q for the RVC folks - do you think we need these checks at all, or should we just remove the checks entirely?

In 1.4 we're moving to a system where you need to pass conformance and PICS checks before running any other tests, so in theory you won't see non-conformant devices come through to here. Or would you prefer to leave these in as an additional check?

src/python_testing/TC_RVCCLEANM_1_2.py
hicklin: Yes it is. Same of `A0001`. Should this check be made somewhere else, or is it dealt by the data model?

src/python_testing/TC_RVCCLEANM_2_1.py
```diff
@@ -92,16 +92,39 @@ async def test_TC_RVCCLEANM_2_1(self):
...
-                asserts.fail("The --app-pid flag must be set when PICS_SDK_CI_ONLY is set")
-            self.app_pipe = self.app_pipe + str(app_pid)
-
-        asserts.assert_true(self.check_pics("RVCCLEANM.S.A0000"), "RVCCLEANM.S.A0000 must be supported")
-        asserts.assert_true(self.check_pics("RVCCLEANM.S.A0001"), "RVCCLEANM.S.A0001 must be supported")
-        asserts.assert_true(self.check_pics("RVCCLEANM.S.C00.Rsp"), "RVCCLEANM.S.C00.Rsp must be supported")
-        asserts.assert_true(self.check_pics("RVCCLEANM.S.C01.Tx"), "RVCCLEANM.S.C01.Tx must be supported")
-
-        attributes = Clusters.RvcCleanMode.Attributes
+            if app_pid != 0:
```
hicklin, line 95: I like this. Should this be indented one less together with like 94 and line 93 be removed?

Same comment in clean 2.2 and run 2.1 and 2.2.

src/python_testing/TC_RVCCLEANM_2_1.py
hicklin: If there are other ways of checking this, yes.

src/python_testing/TC_RVCCLEANM_2_2.py
```diff
@@ -94,13 +94,23 @@ async def test_TC_RVCCLEANM_2_2(self):
...
+        RVCRun_attr_list = attributes.AttributeList
+        attribute_list = await self.read_single_attribute_check_success(endpoint=self.endpoint, cluster=RVCRun_cluster, attribute=RVCRun_attr_list)
+        supported_modes_attr_id = attributes.SupportedModes.attribute_id
+        current_mode_attr_id = attributes.CurrentMode.attribute_id
+
+        if supported_modes_attr_id not in attribute_list:
+            asserts.fail("Supported modes needs to be supported attribute")
+
+        if current_mode_attr_id not in attribute_list:
+            asserts.fail("Current mode needs to be supported attribute")
```
hicklin, line 113: I guess these can be removed as well now. Same in the following scripts.

hicklin (changes requested):

src/python_testing/TC_RVCCLEANM_2_2.py
j-ororke: Thank you, I have removed these conditional checks for current mode and supported mode attributes from the RVC Clean and Run test modules.

src/python_testing/TC_RVCCLEANM_2_1.py
j-ororke: I believe you are correct, I have unindented lines 95 and 94 as the self.is_ci condition check on line 93 does not appear to be needed after running some tests without it. I have done the same in clean 2.2 and run 2.1 and 2.2 test modules as well.

cecille (commented):
I think the CI checks still need to be put back, but this looks otherwise correct to me. 

@hicklin - would you mind taking another look?

src/python_testing/TC_RVCCLEANM_2_1.py
j-ororke: Sorry, I have re-added the if self.is_ci checks as they appear to be needed for these tests

src/python_testing/TC_RVCCLEANM_2_1.py
```diff
@@ -96,14 +96,24 @@ async def test_TC_RVCCLEANM_2_1(self):
             app_pid = self.matter_test_config.app_pid
             if app_pid == 0:
                 asserts.fail("The --app-pid flag must be set when PICS_SDK_CI_ONLY is set")
-            self.app_pipe = self.app_pipe + str(app_pid)
```
hicklin, line 99: I think that this is needed as `write_to_app_pipe` is used later without specifying the `app_pipe_name`. See [here](https://github.com/project-chip/connectedhomeip/blob/f212d601738728d58e88ada3bc7e65b84cdfe7f0/src/python_testing/matter_testing_infrastructure/chip/testing/matter_testing.py#L1028).

src/python_testing/TC_RVCCLEANM_2_2.py
```diff
@@ -107,15 +107,11 @@ async def test_TC_RVCCLEANM_2_2(self):
         self.directmodech_bit_mask = Clusters.RvcCleanMode.Bitmaps.Feature.kDirectModeChange
         self.endpoint = self.matter_test_config.endpoint
         self.is_ci = self.check_pics("PICS_SDK_CI_ONLY")
+        app_pid = self.matter_test_config.app_pid
         if self.is_ci:
             app_pid = self.matter_test_config.app_pid
             if app_pid == 0:
-                asserts.fail("The --app-pid flag must be set when PICS_SDK_CI_ONLY is set.c")
-            self.app_pipe = self.app_pipe + str(app_pid)
```
hicklin, line 114: I think that this is needed as `write_to_app_pipe` is used later without specifying the `app_pipe_name`.

src/python_testing/TC_RVCRUNM_2_1.py
```diff
@@ -90,15 +90,25 @@ async def test_TC_RVCRUNM_2_1(self):
         if self.is_ci:
             app_pid = self.matter_test_config.app_pid
             if app_pid == 0:
-                asserts.fail("The --app-pid flag must be set when PICS_SDK_CI_ONLY is set.c")
-            self.app_pipe = self.app_pipe + str(app_pid)
```
hicklin, line 94: this may be needed.

src/python_testing/TC_RVCRUNM_2_2.py
```diff
@@ -135,21 +135,26 @@ async def test_TC_RVCRUNM_2_2(self):
...
         self.endpoint = self.matter_test_config.endpoint
-        self.is_ci = self.check_pics("PICS_SDK_CI_ONLY")
         self.mode_a = self.matter_test_config.global_test_params['PIXIT.RVCRUNM.MODE_A']
         self.mode_b = self.matter_test_config.global_test_params['PIXIT.RVCRUNM.MODE_B']
+        self.is_ci = self.check_pics("PICS_SDK_CI_ONLY")
         if self.is_ci:
             app_pid = self.matter_test_config.app_pid
             if app_pid == 0:
-                asserts.fail("The --app-pid flag must be set when PICS_SDK_CI_ONLY is set.c")
-            self.app_pipe = self.app_pipe + str(app_pid)
```
hicklin, line 145: This may be needed

hicklin (changes requested):
Looks good. I think there is a small mistake in removing the setting of  `self.app_pipe`.

src/python_testing/TC_RVCCLEANM_2_1.py
j-ororke: Thank you @hicklin,
You are correct that was definitely needed in there and not sure how I had deleted it. 
However, I have returned it to the code now.

src/python_testing/TC_RVCCLEANM_2_2.py
j-ororke: Thank you, you are indeed correct that was definitely needed in there and not sure how I had deleted it.
However, I have returned it to the code now.

src/python_testing/TC_RVCRUNM_2_1.py
j-ororke: Thank you, you are correct that was definitely needed in there. I have returned it to its rightful place here now.

src/python_testing/TC_RVCRUNM_2_2.py
j-ororke: Thank you, you are correct that was definitely needed in there. I have returned it to its rightful place here now.

PR #34182, opened 2024-07-03
[Darwin] MTRDevice in-memory cache needs to be resilient to data store returning nil

jtung-apple:
In the case that the underlying data store returns nil for a cluster path that MTRDevice has previously persisted data, MTRDevice's in-memory cache would be in a bad state, and can only recover by reestablishing subscription to the device, to reload attributes.

Working on local testing as well as better unit testing.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -1235,6 +1235,12 @@ static NSString * const sLastInitialSubscribeLatencyKey = @"lastInitialSubscribe
 - (void)_handleSubscriptionError:(NSError *)error
 {
     std::lock_guard lock(_lock);
+    [self _doHandleSubscriptionError:error];
+}
+
+- (void)_doHandleSubscriptionError:(NSError *)error
+{
+    os_unfair_lock_assert_owner(&_lock);
 
     [self _changeInternalState:MTRInternalDeviceStateUnsubscribed];
     _unreportedEvents = nil;
@@ -1400,6 +1406,12 @@ static NSString * const sLastInitialSubscribeLatencyKey = @"lastInitialSubscribe
 - (void)_handleSubscriptionReset:(NSNumber * _Nullable)retryDelay
 {
     std::lock_guard lock(_lock);
+    [self _doHandleSubscriptionReset:retryDelay];
+}
+
+- (void)_doHandleSubscriptionReset:(NSNumber * _Nullable)retryDelay
+{
+    os_unfair_lock_assert_owner(&_lock);
 
     // If we are here, then either we failed to establish initial CASE, or we
     // failed to send the initial SubscribeRequest message, or our ReadClient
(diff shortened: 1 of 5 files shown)


src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -2122,6 +2134,9 @@ - (nullable MTRDeviceClusterData *)_clusterDataForPath:(MTRClusterPath *)cluster
     MTRDeviceClusterData * data = [_deviceController.controllerDataStore getStoredClusterDataForNodeID:_nodeID endpointID:clusterPath.endpoint clusterID:clusterPath.cluster];
     if (data != nil) {
         [_persistedClusterData setObject:data forKey:clusterPath];
+    } else {
+        // If clusterPath is in _persistedClusters and the data store returns nil for it, then the in-memory cache is now not dependable, and subscription should be reset and reestablished to reload cache from device
+        [self _resetSubscriptionWithReasonString:[NSString stringWithFormat:@"Data store has no data for cluster %@", clusterPath]];
```
bzbarsky-apple, line 2139: Here's a question.  Say we do that.  Then we start getting data reports from the server. That will call `_getAttributesToReportWithReportedValues:`, which will do `_cachedAttributeValueForPath`, which will land right back in this code trying to get the existing value, and nuke the subscription again, etc, no?

Actually, it's worse: just trying to start the subscription will call `_getCachedDataVersions` which will already do bad things, I expect.

At the very least, we need to clear the relevant path from `_persistedClusters` here.  But we might also need to clear other things; needs some careful thought about which ones (clusterDataToPersist?  Storage?  Others?)

jtung-apple: I have changed this part to reconcile `_persistedClusters` with what's in storage when this condition is detected, so the subscription reset doesn't get triggered repeatedly from the same storage data loss.

woody-apple (approved):

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -2100,6 +2112,22 @@ - (void)unitTestClearClusterData
...
+
+    NSMutableSet * clusterPathsToRemove = [NSMutableSet set];
+    for (MTRClusterPath * clusterPath in _persistedClusters) {
+        MTRDeviceClusterData * data = [_deviceController.controllerDataStore getStoredClusterDataForNodeID:_nodeID endpointID:clusterPath.endpoint clusterID:clusterPath.cluster];
+        if (!data) {
+            [clusterPathsToRemove addObject:clusterPath];
+        }
+    }
+
+    MTR_LOG_ERROR("%@ ", self);
```
woody-apple, line 2127: Is this safe, also what is this logging?

jtung-apple: Meant to log the difference between in-memory cache and storage - I've fixed it now. (And also the incomplete log was safe, even if unclear)

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -916,6 +916,9 @@ - (void)invalidate
         // the subscrption after this point.
         std::lock_guard lock(self->_lock);
         self->_currentReadClient = nullptr;
```
woody-apple, line 918: Does this have the same problem as below?

jtung-apple: Yes this is an existing issue that invalidate doesn't actually delete the SubscriptionCallback object.

jtung-apple: And ReadClient is held by SubscriptionCallback, and the callback object destructor nils / deletes the ReadClient.

PR #34183, opened 2024-07-03
Remove no-longer-used MTRDevice logic for truncating data version lists

bzbarsky-apple:
After https://github.com/project-chip/connectedhomeip/pull/34111, ReadClient handles this logic itself, so the attempted truncation in MTRDevice was now dead code.

REVIEW NOTE: A whitespace-ignoring diff might be simpler to review.

### DIFF
--- a/src/app/ReadClient.cpp
+++ b/src/app/ReadClient.cpp
@@ -403,7 +403,11 @@ CHIP_ERROR ReadClient::BuildDataVersionFilterList(DataVersionFilterIBs::Builder 
                                                   const Span<DataVersionFilter> & aDataVersionFilters,
                                                   bool & aEncodedDataVersionList)
 {
-    ChipLogProgress(DataManagement, "Attempting to encode %lu data version filters", static_cast<unsigned long>(aDataVersionFilters.size()));
+#if CHIP_PROGRESS_LOGGING
+    size_t encodedFilterCount    = 0;
+    size_t irrelevantFilterCount = 0;
+    size_t skippedFilterCount    = 0;
+#endif
     for (auto & filter : aDataVersionFilters)
     {
         VerifyOrReturnError(filter.IsValidDataVersionFilter(), CHIP_ERROR_INVALID_ARGUMENT);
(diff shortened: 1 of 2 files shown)


jtung-apple (approved):

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -2523,8 +2502,6 @@ - (void)_setupSubscriptionWithReason:(NSString *)reason
                                return;
                            }
 
-                           MTR_LOG("%@ Subscribe with data version list size %lu, reduced by %lu", self, static_cast<unsigned long>(dataVersions.count), static_cast<unsigned long>(dataVersionFilterListSizeReduction));
```
jtung-apple, line 2526: Actually, could we keep this log, but remove the `dataVersionFilterListSizeReduction` bits? This is the only place that MTRDevice logs when it starts a subscription.

bzbarsky-apple: Yes, can do.

bzbarsky-apple: Done.

src/app/ReadClient.cpp
```diff
@@ -403,6 +403,7 @@ CHIP_ERROR ReadClient::BuildDataVersionFilterList(DataVersionFilterIBs::Builder
                                                   const Span<DataVersionFilter> & aDataVersionFilters,
                                                   bool & aEncodedDataVersionList)
 {
+    ChipLogProgress(DataManagement, "Attempting to encode %lu data version filters", static_cast<unsigned long>(aDataVersionFilters.size()));
```
ksperling-apple, line 406: We could move this to the end so in either case we get one log statement saying how many we encoded

bzbarsky-apple: The "we didn't encode everything" log statement logs how many were chopped off.  Not that this is not trivially derivable from "how many we encoded" and "how many we were asked to encode", since we skip encoding the ones that don't match our request paths.....

bzbarsky-apple: Changed to one log statement.

saurabhst (approved):

PR #34187, opened 2024-07-04
Make RPC call synchronous in Fabric Sync

### LINKED ISSUE #33784: [Feature] Make RPC call synchronous in Fabric Admin and Fabric Bridge
### Feature description

...
see if we can integrate this API to CHIP https://pigweed.dev/pw_rpc/#client-synchronous-call-wrappers

### Platform

other

### Platform Version(s)

_No response_

### Anything else?

_No response_

yufengwangca:
We need to make all RPC call synchronous in Fabric Sync to grantee the correct sequence of Fabric Sync.

I have explored all possible options to make an RPC call synchronous but found that busy-waiting is the most reliable option so far.

A better option would be to use the pw client synchronous call wrappers, but they are quite expensive and not available in the current nanopb integration in the CHIP project.

Fixes #33784

### DIFF
--- a/examples/fabric-admin/rpc/RpcClient.cpp
+++ b/examples/fabric-admin/rpc/RpcClient.cpp
@@ -36,12 +36,35 @@ using namespace chip;
 namespace {
 
 // Constants
+constexpr uint32_t kRpcTimeoutMs     = 1000;
 constexpr uint32_t kDefaultChannelId = 1;
 
 // Fabric Bridge Client
 rpc::pw_rpc::nanopb::FabricBridge::Client fabricBridgeClient(rpc::client::GetDefaultRpcClient(), kDefaultChannelId);
 pw::rpc::NanopbUnaryReceiver<::pw_protobuf_Empty> addSynchronizedDeviceCall;
 pw::rpc::NanopbUnaryReceiver<::pw_protobuf_Empty> removeSynchronizedDeviceCall;
+
+template <typename CallType>
+CHIP_ERROR WaitForResponse(CallType & call)
+{
+    // Wait for the response or timeout
+    uint32_t elapsedTimeMs     = 0;
+    const uint32_t sleepTimeMs = 100;
+
+    while (call.active() && elapsedTimeMs < kRpcTimeoutMs)
+    {
+        usleep(sleepTimeMs * 1000);
+        elapsedTimeMs += sleepTimeMs;
+    }
+
+    if (elapsedTimeMs >= kRpcTimeoutMs)
+    {
+        fprintf(stderr, "RPC Response timed out!");
+        return CHIP_ERROR_TIMEOUT;
+    }
+
+    return CHIP_NO_ERROR;
+}
 
 // Callback function to be called when the RPC response is received
 void OnAddDeviceResponseCompleted(const pw_protobuf_Empty & response, pw::Status status)
@@ -101,7 +124,7 @@ CHIP_ERROR AddSynchronizedDevice(chip::NodeId nodeId)
         return CHIP_ERROR_INTERNAL;
     }
 
-    return CHIP_NO_ERROR;
+    return WaitForResponse(addSynchronizedDeviceCall);
 }
 
 CHIP_ERROR RemoveSynchronizedDevice(chip::NodeId nodeId)
@@ -128,5 +151,5 @@ CHIP_ERROR RemoveSynchronizedDevice(chip::NodeId nodeId)
         return CHIP_ERROR_INTERNAL;
     }
 
-    return CHIP_NO_ERROR;
+    return WaitForResponse(removeSynchronizedDeviceCall);
 }
--- a/examples/fabric-admin/rpc/RpcClient.cpp
+++ b/examples/fabric-admin/rpc/RpcClient.cpp
@@ -19,9 +19,11 @@
 #include "RpcClient.h"
 #include "RpcClientProcessor.h"
 
+#include <chrono>
+#include <condition_variable>
+#include <mutex>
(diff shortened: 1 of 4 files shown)


tcarmelveilleux (changes requested):
- Busy waiting on shared data is guaranteed to be racy
- Instead, replace the RPC operations to be "starts" of operations which can be polled by a status RPC by the caller, for example. Alternatively, consider using event loop posting design with a separate worker to have a way to broker synchronization.

examples/fabric-admin/rpc/RpcClient.cpp
```diff
@@ -101,7 +124,7 @@ CHIP_ERROR AddSynchronizedDevice(chip::NodeId nodeId)
         return CHIP_ERROR_INTERNAL;
     }
 
-    return CHIP_NO_ERROR;
+    return WaitForResponse(addSynchronizedDeviceCall);
```
andy31415, line 127: busy-looping stops the CHIP main progress loop - this seems off here.
Why don't we use teh on_response/on_rpc_error to advance a state machine instead?

https://pigweed.dev/pw_rpc/nanopb/#id1

andy31415:
> My requirement is very simple, make the following function return only after the the operation completes, it is better to halt all following operations to make Fabric Sync in sequence.
> ...
> Could you elaborate more how to achieve this in another way? OnOpenCommissioningWindowCompleted and OpenCommissioningWindow are both called from the same thread, so conditional_varible does not work here.

Please describe what operations need to be done in what order.
Generally if you need to move around specific states (e.g. initiate something after something else) you should create a state machine that has transitions at specific events (e.g. rpc response received/ rpc error etc). If you build such a state machine, you will also have the benefit of it being testable code.

examples/fabric-admin/rpc/RpcClient.cpp
yufengwangca: We already did, OnAddDeviceResponseCompleted is the on_response and on_rpc_error is nullptr. Yes, it stop the main loop here to prevent from maintaining a more complicated state machine

yufengwangca:
> > My requirement is very simple, make the following function return only after the the operation completes, it is better to halt all following operations to make Fabric Sync in sequence.
> > ...
> > Could you elaborate more how to achieve this in another way? OnOpenCommissioningWindowCompleted and OpenCommissioningWindow are both called from the same thread, so conditional_varible does not work here.
> 
> Please describe what operations need to be done in what order. Generally if you need to move around specific states (e.g. initiate something after something else) you should create a state machine that has transitions at specific events (e.g. rpc response received/ rpc error etc). If you build such a state machine, you will also have the benefit of it being testable code.

I want to make all RPC call synchronous, so that I don't have to maintain a state machine that has transitions at specific events. 

Updated the PR, the WaitForResponse function uses a condition variable to wait for the response or timeout. The response completion callbacks (OnAddDeviceResponseCompleted and OnRemoveDeviceResponseCompleted) set the response flags and notify the waiting thread. This avoids busy waiting and makes the code more efficient and responsive.

tcarmelveilleux (approved):

examples/fabric-admin/rpc/RpcClient.cpp
```diff
@@ -36,16 +38,45 @@ using namespace chip;
...
 pw::rpc::NanopbUnaryReceiver<::pw_protobuf_Empty> addSynchronizedDeviceCall;
 pw::rpc::NanopbUnaryReceiver<::pw_protobuf_Empty> removeSynchronizedDeviceCall;
 
+std::mutex responseMutex;
+std::condition_variable responseCv;
+bool responseReceived    = false;
+CHIP_ERROR responseError = CHIP_NO_ERROR;
+
+template <typename CallType>
+CHIP_ERROR WaitForResponse(CallType & call)
```
tehampson, line 55: What is `CallType & call` used for?

examples/fabric-bridge-app/linux/RpcClient.cpp
```diff
@@ -36,15 +38,44 @@ using namespace chip;
...
 rpc::pw_rpc::nanopb::FabricAdmin::Client fabricAdminClient(rpc::client::GetDefaultRpcClient(), kDefaultChannelId);
 pw::rpc::NanopbUnaryReceiver<::chip_rpc_OperationStatus> openCommissioningWindowCall;
 
+std::mutex responseMutex;
+std::condition_variable responseCv;
+bool responseReceived    = false;
+CHIP_ERROR responseError = CHIP_NO_ERROR;
+
+template <typename CallType>
+CHIP_ERROR WaitForResponse(CallType & call)
```
tehampson, line 54: same, what is `CallType & call` used for?

yufengwangca: The CallType & call in the WaitForResponse function template is a reference to an instance of a call type that is used to perform an RPC call. 

In this specific context, CallType is the type of the object that represents the ongoing RPC call. The call parameter is used within the function to wait for the RPC response, using the condition variable responseCv to block the calling thread until the response is received or a timeout occurs.

examples/fabric-admin/rpc/RpcClient.cpp
yufengwangca: The CallType & call in the WaitForResponse function template is a reference to an instance of a call type that is used to perform an RPC call.

In this specific context, CallType is the type of the object that represents the ongoing RPC call. The call parameter is used within the function to wait for the RPC response, using the condition variable responseCv to block the calling thread until the response is received or a timeout occurs.

joonhaengHeo (approved):

PR #34190, opened 2024-07-04
 Darwin: Ensure we tell the OTA delegate when a transfer ends

ksperling-apple:
Ensure we call `handleBDXTransferSessionEndForNodeID:...` if we have called
`handleBDXTransferSessionBeginForNodeID:...`, even if the session is reset via an
error code path.

### DIFF
--- a/src/darwin/Framework/CHIP/MTROTAProviderDelegateBridge.mm
+++ b/src/darwin/Framework/CHIP/MTROTAProviderDelegateBridge.mm
@@ -211,7 +211,7 @@ private:
             // If the send was successful for a status report, since we are not expecting a response the exchange context is
             // already closed. We need to null out the reference to avoid having a dangling pointer.
             mExchangeCtx = nullptr;
-            ResetState(CHIP_ERROR_INTERNAL); // TODO: is this an error scenario?
+            ResetState(CHIP_ERROR_INTERNAL);
         }
         return err;
     }
--- a/src/darwin/Framework/CHIP/MTROTAProviderDelegateBridge.mm
+++ b/src/darwin/Framework/CHIP/MTROTAProviderDelegateBridge.mm
@@ -311,7 +311,7 @@ private:
         } else if (event.EventType != TransferSession::OutputEventType::kAckEOFReceived) {
             error = CHIP_ERROR_INTERNAL;
         }
-        
+
         ResetState(error); // will notify the delegate
         return CHIP_NO_ERROR;
     }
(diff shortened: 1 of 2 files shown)


src/darwin/Framework/CHIP/MTROTAProviderDelegateBridge.mm
```diff
@@ -188,12 +206,12 @@ CHIP_ERROR OnMessageToSend(TransferSession::OutputEvent & event)
...
             mExchangeCtx = nullptr;
-            ResetState();
-        } else if (event.msgTypeData.HasMessageType(Protocols::SecureChannel::MsgType::StatusReport)) {
+            ResetState(err);
+        } else if (msgTypeData.HasMessageType(Protocols::SecureChannel::MsgType::StatusReport)) {
             // If the send was successful for a status report, since we are not expecting a response the exchange context is
             // already closed. We need to null out the reference to avoid having a dangling pointer.
             mExchangeCtx = nullptr;
-            ResetState();
+            ResetState(CHIP_ERROR_INTERNAL); // TODO: is this an error scenario?
```
bzbarsky-apple, line 214: This is a "we decided to send an error status for some reason" scenario.  In theory, we could have extracted the exact status we sent from the event.MsgData (before we moved it into SendMessage), but not sure it's worth it.

src/darwin/Framework/CHIPTests/MTROTAProviderTests.m
```diff
@@ -833,23 +835,32 @@ - (void)test003_ReceiveQueryImageRequestWhileHandlingBDX_RespondImplicitBusy
                                         softwareVersion:kUpdatedSoftwareVersion_5
                                   softwareVersionString:kUpdatedSoftwareVersionString_5
                                              completion:completion];
-        [queryExpectation1 fulfill];
```
bzbarsky-apple, line 836: Why the ordering change here?  If the new order is important, should probably have a comment explaining that it is and why.

bzbarsky-apple (approved):

src/darwin/Framework/CHIPTests/MTROTAProviderTests.m
ksperling-apple: Well, the expectations refer to receiving delegate callbacks, so it's easier to understand / be sure that the expectation order we assert later via those expectations is correct if fulfilling the expectation is actually the first thing we do in the block. Further below we had

```
[sOTAProviderDelegate respondErrorWithCompletion:outerCompletion];
[queryExpectation2 fulfill];
```

where in response to the query we cancel the existing transfer. Depending how exactly the callbacks are dispatched this could have the effect of the TransferEnd callback being called before we fulfill that expectation.
(With the current way we dispatch them it would probably have been fine, but it's just harder to understand.)

I moved the fulfills to the start of the callbacks as a separate commit first and ensured the tests still pass that way before making any of the other changes.

ksperling-apple:

PR #34191, opened 2024-07-04
Revert "[ESP32] Remove SupportedModes implementation of ModeSelect from nvs and use all-clusters-apps common implementation"

jadhavrohit924:
Reverts project-chip/connectedhomeip#34044

- Reason for revert

Actually, we thought of using common implementation and planned to remove this fixed attributes from the factory partition because of while OTA, we cannot update these attributes as factory partition can’t be change in OTA. in #34044

But later we thought of supporting this fixed attributes in the factory partition for backward compatibility along with SDKs common implementation which has static data.


andy31415 (dismissed):
@jadhavrohit924 please add the reason for the revert in the summary of the PR.

We should try to make PRs clear in what they are doing and why.

jadhavrohit924:
@andy31415 Sorry, for the inconvenience. I have updated the PR description.

shubhamdp (approved):

dhrishi (approved):

jmartinez-silabs (approved):

PR #34194, opened 2024-07-04
Update CI images to latest version

### LINKED ISSUE #35121: [BUG] Clang-Tidy error - bugprone-pointer-arithmetic-on-polymorphic-object in Eventmanagement
### Reproduction steps

New version of clang-tidy avaliable with latest pigweed version, found new types of errors:
- `bugprone-pointer-arithmetic-on-polymorphic-object` - error:

```
2024-08-19 08:28:29 INFO TIDY ../../src/app/EventManagement.cpp: /mnt/dane/projects/matter/connectedhomeip-second/out/sanitizers/../../src/app/EventManagement.cpp:107:51: error: pointer arithmetic on polymorphic object of type 'CircularEventBuffer' can result in undefined behavior if the dynamic type differs from the pointer type [bugprone-pointer-arithmetic-on-polymorphic-object,-warnings-as-errors]`  
    107 | next = (bufferIndex < aNumBuffers - 1) ? &apCircularEventBuffer[bufferIndex + 1] : nullptr;
            | ^~~~~~~~~~~~~~~~~~~~~
/mnt/dane/projects/matter/connectedhomeip-second/out/sanitizers/../../src/app/EventManagement.cpp:109:20: error: pointer arithmetic on polymorphic object of type 'CircularEventBuffer' can result in undefined behavior if the dynamic type differs from the pointer type [bugprone-pointer-arithmetic-on-polymorphic-object,-warnings-as-errors]
    109 | current = &apCircularEventBuffer[bufferIndex];
```
- `bugprone-suspicious-stringview-data-usage` - error: [clang-tidy.log](https://github.com/user-attachments/files/16688414/clang-tidy.log)

### Bug prevalence

always

ff6863ed127fb1a8a6f134ed15b2ed98ed75e418

### Platform

core

### Platform Version(s)

_No response_

### Anything else?

_No response_

jlatusek:
## Problem

There are few images in CI which are outdated. Update of them bring up few additional issues:
1. Script `runner.py` stopped to report success during intentionally failing tests. 
2. Changes in pigweed api related to `WriteLine` function which break compilation
3. Latest version of clang-tidy has regression which causes to report `-c` compilator argument as unused. Fix were merged into clang-tidy master but it will take some time to be available in `CIPD`
4. Clang-tidy is not correctly interpret `CHIP_ERROR_POSIX(errno);` from `GetSocket` method. It will report an error that function related to sockets are called with `-1` argument. 
5. silabs sdk provided its own version of python which should be used by internal tools. 
6. Zap generated files are formated in different way. 
7. After updates of Tizen images it started failing during preparing of docker images on gh worker

## Changes

I updated the outdated CI images to one based on Ubuntu 24.04. Unfortunately, this required some code changes and updating of python packages (old version of packages required by some embedded targets do not want to be instaled on python 3.12) and pigweed submodule simultaneously.

1. Method `_run` were throwing an error, due to bug in python `3.9.5`. As [documentations](https://docs.python.org/3/tutorial/errors.html#defining-clean-up-actions) states: `If the finally clause executes a break, continue or return statement, exceptions are not re-raised.` but the code to work properly expect the `_run` method to throw an error. I've fixed it by removing unnecessary try block. 
2. Change function to expect object instead of reference to object
3. I changed script calling clang-tidy to remove `-c` from command line arguments passed to it, after update of clang-tidy in pigweed submodule this change can be removed. 
4. I manually disable linting for problematic lines. I wasn't able to find better solution
5. I added to `PATH` python from silabs SDK to use it during calling of internal tools
6. Regenerate files from zap templates
7. Add user argument for tizen qemu image. There is a bug in gh actions described [here](https://github.com/actions/checkout/issues/956), which prevent from running container with different user than root
8. Update mac-os to use python3.12 installed by `brew` - cipd do not have python3.12 avaliable right now
9. Update pigweed submodule
10. Fix some clang-tidy issues described here: #35121


arkq:
Done in different PRs, e.g.:

- https://github.com/project-chip/connectedhomeip/pull/35644
- https://github.com/project-chip/connectedhomeip/pull/35802

PR #34195, opened 2024-07-04
Add Fabric Synchronization Guide

yufengwangca:
Add Fabric Synchronization Guide


samadDotDev (approved):

docs/guides/fabric_synchronization_guide.md
```diff
@@ -0,0 +1,148 @@
+# Fabric Synchronization Guide
+
+-   [Fabric Synchronization Guide](#fabric-synchronization-guide)
+    -   [Fabric Sync Example Applications](#fabric-sync-example-applications)
+    -   [Run Fabric Sync Demo on RP4](#run-fabric-sync-demo-on-rp4)
+
+## Fabric Sync Example Applications
+
+Fabric-Admin and Fabric-Bridge applications are provided to demonstrate Fabric
+Synchronization feature.. You can find them in the examples.
```
andy31415, line 10:
```suggestion
Synchronization feature. You can find them in the examples.
```

examples/fabric-admin/scripts/run_fabric_sink_rp4.sh
andy31415, line 1: why do we have separate scripts for rp4 vs x86? we seem to just differ by path, however for that you could have path arguments and some defaults if unset depending on the platform. 

Having separate scripts makes it significantly harder to maintain as a lot of code is copied & pasted

examples/fabric-admin/scripts/run_fabric_source_x86.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+# Kill fabric-admin if it is running
```
andy31415, line 3: why don't we call `stop_fabric_source.sh` instead of having the same code inside?

examples/fabric-admin/scripts/run_fabric_source_x86.sh
```diff
@@ -0,0 +1,26 @@
...
+    kill -9 "$fabric_bridge_app_pid"
+    echo "Killed fabric-bridge-app with PID $fabric_bridge_app_pid"
+fi
+
+# Remove /tmp/chip_* files and directories
+sudo rm -rf /tmp/chip_*
+echo "Removed /tmp/chip_* files and directories"
+
+# Start fabric-bridge-app and redirect its output to /dev/null
+out/debug/standalone/fabric-bridge-app >/dev/null 2>&1 &
```
andy31415, line 22: why do we send logs to dev/null? That seems to make it hard to debug. How about a logfile instead?

docs/guides/fabric_synchronization_guide.md
```diff
@@ -0,0 +1,148 @@
+# Fabric Synchronization Guide
```
tcarmelveilleux, line 1: This document should explain much more what is the topology/goal of the example, with context.

yufengwangca: Added

docs/guides/fabric_synchronization_guide.md
```diff
@@ -0,0 +1,164 @@
...
+
+## Fabric Sync Example Applications
+
+Fabric-Admin and Fabric-Bridge example applications are provided to demonstrate
+Fabric Synchronization feature. You can find them in the examples.
+
+![matter_fabric_synchronization](images/matter_fabric_synchronization.png)
+
+Fabric-Admin example app implements the Fabric Administrator role and
+communicate with the Fabric-Bridge-App on the other side, facilitating the
```
andy31415, line 15:
```suggestion
communicates with the Fabric-Bridge-App on the other side, facilitating the
```

examples/fabric-admin/scripts/run_fabric_sink.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+# Default paths
+FABRIC_ADMIN_PATH="./fabric-admin"
+FABRIC_ADMIN_LOG="/tmp/fabric_admin.log"
+
+# Detect platform and set default paths
+if [[ $(uname -m) == "x86_64" ]]; then
+    FABRIC_ADMIN_PATH="out/debug/standalone/fabric-admin"
```
andy31415, line 9: could we use some file finding logic? For example I compile with build_examples.py so the path is never `debug/standalone`. I would suggest some form of iteration of predefined paths and the first one wins. like:

```sh
find_fabric_admin() {
   local choices=(
       "./fabric-admin"
       "out/debug/standalone/fabric-admin"
       "out/linux-x64-fabric-admin/fabric-admin"
       "out/darwin-arm64-fabric-admin/fabric-admin"
   )
   
   for path in "${choices[@]}"; do
       if [[ -e "$path" ]]; then
          echo "$path"
          return 0
      fi
   done
  return 1
}

FABRIC_ADMIN_PATH=$(find_fabric_admin)
if [[ $? -ne 0 ]]; then
   >&2 echo "Could not find the fabric-admin binary"
   exit 1
fi
echo "PATH IS: $FABRIC_ADMIN_PATH"
```
Potentially allow an argument to specify the path to use.

examples/fabric-admin/scripts/run_fabric_sink.sh
```diff
@@ -0,0 +1,41 @@
...
+fi
+
+# Check if --verbose option is specified
+VERBOSE=false
```
andy31415, line 13: consider supporting arguments for paths. Then this becomes:

```sh
while true; do
    case "$1" in
        -v | --verbose ) VERBOSE=true; shift ;;
        --fabric-admin ) FABRIC_ADMIN_PATH="$2"; shift 2 ;;
        *) break;;
    esac
done

# If fabric-admin is empty, then try to find it.

```

examples/fabric-admin/scripts/stop_fabric_source.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+FABRIC_ADMIN_PATH="./fabric-admin"
```
andy31415, line 3: are these paths still correct if the paths above use `out/debug` or other paths?

Should we grep for `/fabric-bridge-app` without the prefix?

examples/fabric-admin/scripts/run_fabric_source.sh
```diff
@@ -0,0 +1,44 @@
...
+for arg in "$@"; do
+    if [ "$arg" == "--verbose" ]; then
+        VERBOSE=true
+        break
+    fi
+done
+
+# Stop any running instances and clean up
+
+if [[ "$PLATFORM" == "x86_64" ]]; then
```
andy31415, line 27: can we get the path to the current script and use that as a relative location? It seems we have these 3 scripts together always.

examples/fabric-admin/scripts/stop_fabric_source.sh
yufengwangca: it is no more valid if we use find binary

andy31415: The issue I have is that this is misleading easily because of the assumption of `.` being the current directory. It is not and you regex on `./fabric-admin` as `<ANY_CHAR>/fabric_admin` so this matches even `some/dir/here/fabric-admin` (because `.` matches `e` in the example).

I am suggesting to just do `/fabric-admin` and `/fabric-bridge-app` (since I am unclear why we want the . there .... is there any benefit of having a any-char-grep in there?)

andy31415 (approved):

saurabhst (approved):

PR #34196, opened 2024-07-04
Move Nullable from src/app/data-model to lib/support

tcarmelveilleux:
- Nullable.h and associated files (attribute-storage-null-handling.h, NullObject.h) were in src/app and src/app-util, in ways that led to several interesting circular dependencies when attempting to write portable code relying on the Matter complex chip::app::Nullable abstraction.
- This PR moves the portable nullable support to src/lib/support to break some possible dependency cycles between data model and underlying deps.
- Added an "include trap" for previous users of Nullable.h to help facilitate the move.

Testing done:

- Moved unit tests still work, and validated they are correctly run.


src/lib/support/Nullable.h
```diff
@@ -0,0 +1,132 @@
...
+#pragma once
+
+#include <lib/support/MatterScalarNullHandling.h>
+#include <optional>
+#include <type_traits>
+#include <utility>
+
+namespace chip {
+namespace app {
+namespace DataModel {
```
andy31415, line 28: This is the first time we introduce a Datamodel (and app:: as well) into lib/support.

I am unsure how we should approach things here ... I feel uneasy about having app:: and others into lib/support however changing that introduces more blast radius. Could we set up aliases instead in some way?

PR #34198, opened 2024-07-04
Split the targets for nullable support

tcarmelveilleux:
- Fixes a prior circular dependency on nullable scalar traits
- Allows portable code to only depend on the nullable support without needing to also include any code-generated artifacts


src/app/util/BUILD.gn
```diff
@@ -15,22 +15,35 @@
 import("//build_overrides/chip.gni")
 import("${chip_root}/src/app/common_flags.gni")
 
+source_set("null-utils") {
```
andy31415, line 18: Trying to avoid filler words of utils/helpers/core/support as they are vague.

When I look at `null` only, it is not obvious why "odd sized integers" is related to null. Can we rename?

I propose one of `primitives` or `basic-types`

src/app/data-model/BUILD.gn
```diff
@@ -13,6 +13,19 @@
 # limitations under the License.
 import("//build_overrides/chip.gni")
 
+source_set("nullable-support") {
```
andy31415, line 16: I think just `nullable` is enough. Remove the filler word.

andy31415 (approved):

bzbarsky-apple (approved):

PR #34206, opened 2024-07-05
[NXP] Add docker image for Zephyr examples

yeaissa:
Adding docker image for Zephyr examples


andy31415 (approved):

integrations/docker/images/stage-2/chip-build-nxp-zephyr/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+    xz-utils \
+    && : # last line
+
+WORKDIR /opt/nxp-zephyr
+RUN set -x \
+    && wget https://github.com/zephyrproject-rtos/sdk-ng/releases/download/v0.16.5/zephyr-sdk-0.16.5_linux-x86_64_minimal.tar.xz \
+    && tar xvf zephyr-sdk-0.16.5_linux-x86_64_minimal.tar.xz \
+    && rm -rf zephyr-sdk-0.16.5_linux-x86_64_minimal.tar.xz \
+    && zephyr-sdk-0.16.5/setup.sh -t arm-zephyr-eabi \
+    && pip3 install -U --no-cache-dir west \
```
andy31415, line 17: @yeaissa found this after I tried to compile this image: this does not work on ubuntu 24.04 (which images are now based on) because it by default prevents system installs.

andy31415: Fixed in #34211

We may want to use virtualenvs though (and we should probably find a way to test these PRs in advance).

PR #33799, opened 2024-06-07
Allows opening a commissioning window on a bridged node 

yufengwangca:
AdministratorCommissioningCluster is not always on endpoint 0, to open Commissioning Window on bridged node, we need to support sending an OpenCommissioningWindow command with the given parameters to the given endpoint of the given node.

Fixes #33798

PR #34209, opened 2024-07-05
Add OCW verifier, pass params between fabric bridge & admin

### LINKED ISSUE #33763: [Feature] Using ECM instead of BCM for Fabric Sync
### Feature description

...
BCM is optional in the product, we need to switch to use ECM to sync the Matter devices

We need to handle OCW parameters in fabric bridge app,  update RPC and add verifier to CommissioningWindowOpener.

### Platform

other

### Platform Version(s)

_No response_

### Anything else?

_No response_

samadDotDev:
This introduces an optional PAKE passcode verifier arg in the commissioning window opener to support the pass-through of this parameter to the device via admin from the OCW request received on the bridge (`fabric-bridge -RPC-> fabric-admin --> device`). The RPC contract update between fabric admin and fabric bridge app to pass the OCW params and use of the new function arg is done in a separate commit.

Note that the [OCW command](https://github.com/CHIP-Specifications/connectedhomeip-spec/blob/master/src/service_device_management/AdminCommissioningCluster.adoc#81-opencommissioningwindow-ocw-command) which is being proxied here doesn't contain the setup pin code to compute the verifier with included iterations count.

Fixes #33763

### DIFF
--- a/examples/chip-tool/commands/pairing/OpenCommissioningWindowCommand.cpp
+++ b/examples/chip-tool/commands/pairing/OpenCommissioningWindowCommand.cpp
@@ -39,8 +39,9 @@ CHIP_ERROR OpenCommissioningWindowCommand::RunCommand()
                                                           .SetTimeout(mCommissioningWindowTimeout)
                                                           .SetIteration(mIteration)
                                                           .SetDiscriminator(mDiscriminator)
-                                                          .SetReadVIDPIDAttributes(true),
-                                                      &mOnOpenCommissioningWindowCallback, ignored);
+                                                          .SetReadVIDPIDAttributes(true)
+                                                          .SetCallback(&mOnOpenCommissioningWindowCallback),
+                                                      ignored);
     }
 
     ChipLogError(chipTool, "Unknown commissioning window option: %d", to_underlying(mCommissioningWindowOption));
--- a/examples/chip-tool/commands/pairing/OpenCommissioningWindowCommand.cpp
+++ b/examples/chip-tool/commands/pairing/OpenCommissioningWindowCommand.cpp
@@ -34,10 +34,17 @@ CHIP_ERROR OpenCommissioningWindowCommand::RunCommand()
     if (mCommissioningWindowOption == Controller::CommissioningWindowOpener::CommissioningWindowOption::kTokenWithRandomPIN)
     {
         SetupPayload ignored;
-        return mWindowOpener->OpenCommissioningWindow(mNodeId, System::Clock::Seconds16(mCommissioningWindowTimeout), mIteration,
-                                                      mDiscriminator, NullOptional, NullOptional, NullOptional,
-                                                      &mOnOpenCommissioningWindowCallback, ignored,
-                                                      /* readVIDPIDAttributes */ true);
+        Controller::CommissioningWindowOpener::CommissioningWindowPasscodeParams params = {
+            .common               = { .deviceId      = mNodeId,
+                                      .timeout       = System::Clock::Seconds16(mCommissioningWindowTimeout),
+                                      .iteration     = mIteration,
+                                      .discriminator = mDiscriminator },
+            .setupPIN             = NullOptional,
+            .salt                 = NullOptional,
+            .readVIDPIDAttributes = true,
+            .callback             = &mOnOpenCommissioningWindowCallback,
+        };
+        return mWindowOpener->OpenCommissioningWindow(params, ignored);
     }
 
     ChipLogError(chipTool, "Unknown commissioning window option: %d", to_underlying(mCommissioningWindowOption));
--- a/examples/chip-tool/commands/pairing/OpenCommissioningWindowCommand.cpp
+++ b/examples/chip-tool/commands/pairing/OpenCommissioningWindowCommand.cpp
@@ -34,17 +34,13 @@ CHIP_ERROR OpenCommissioningWindowCommand::RunCommand()
     if (mCommissioningWindowOption == Controller::CommissioningWindowOpener::CommissioningWindowOption::kTokenWithRandomPIN)
     {
         SetupPayload ignored;
-        Controller::CommissioningWindowOpener::CommissioningWindowPasscodeParams params = {
-            .common               = { .deviceId      = mNodeId,
-                                      .timeout       = System::Clock::Seconds16(mCommissioningWindowTimeout),
-                                      .iteration     = mIteration,
-                                      .discriminator = mDiscriminator },
-            .setupPIN             = NullOptional,
-            .salt                 = NullOptional,
-            .readVIDPIDAttributes = true,
-            .callback             = &mOnOpenCommissioningWindowCallback,
-        };
-        return mWindowOpener->OpenCommissioningWindow(params, ignored);
+        return mWindowOpener->OpenCommissioningWindow(Controller::CommissioningWindowPasscodeParams()
+                                                          .SetNodeId(mNodeId)
+                                                          .SetTimeout(mCommissioningWindowTimeout)
+                                                          .SetIteration(mIteration)
+                                                          .SetDiscriminator(mDiscriminator)
+                                                          .SetReadVIDPIDAttributes(true),
+                                                      &mOnOpenCommissioningWindowCallback, ignored);
     }
 
     ChipLogError(chipTool, "Unknown commissioning window option: %d", to_underlying(mCommissioningWindowOption));
--- a/examples/chip-tool/commands/pairing/OpenCommissioningWindowCommand.cpp
+++ b/examples/chip-tool/commands/pairing/OpenCommissioningWindowCommand.cpp
@@ -35,7 +35,7 @@ CHIP_ERROR OpenCommissioningWindowCommand::RunCommand()
     {
         SetupPayload ignored;
         return mWindowOpener->OpenCommissioningWindow(mNodeId, System::Clock::Seconds16(mCommissioningWindowTimeout), mIteration,
-                                                      mDiscriminator, NullOptional, NullOptional,
+                                                      mDiscriminator, NullOptional, NullOptional, NullOptional,
                                                       &mOnOpenCommissioningWindowCallback, ignored,
                                                       /* readVIDPIDAttributes */ true);
     }
--- a/examples/common/pigweed/protos/fabric_admin_service.proto
+++ b/examples/common/pigweed/protos/fabric_admin_service.proto
@@ -5,8 +5,13 @@ import 'pw_protobuf_protos/common.proto';
 package chip.rpc;
 
 // Define the message for a synchronized end device with necessary fields
-message DeviceInfo {
+message DeviceCommissioningWindowInfo {
   uint64 node_id = 1;
+  uint32 commissioning_timeout = 2;
+  uint32 discriminator = 3;
+  uint32 iterations = 4;
+  bytes salt = 5;
+  bytes verifier = 6;
 }
 
 // Define the response message to convey the status of the operation
@@ -15,6 +20,5 @@ message OperationStatus {
 }
 
 service FabricAdmin {
-  rpc OpenCommissioningWindow(DeviceInfo) returns (OperationStatus){}
+  rpc OpenCommissioningWindow(DeviceCommissioningWindowInfo) returns (OperationStatus){}
 }
-
--- a/examples/fabric-admin/commands/pairing/OpenCommissioningWindowCommand.h
+++ b/examples/fabric-admin/commands/pairing/OpenCommissioningWindowCommand.h
@@ -48,7 +48,9 @@ public:
                     &mIteration, "Number of PBKDF iterations to use to derive the verifier.  Ignored if 'option' is 0.");
         AddArgument("discriminator", 0, 4096, &mDiscriminator, "Discriminator to use for advertising.  Ignored if 'option' is 0.");
         AddArgument("timeout", 0, UINT16_MAX, &mTimeout, "Time, in seconds, before this command is considered to have timed out.");
-        AddArgument("salt", &mSalt, "Salt payload encoded in hexadecimal. Random salt will be generated if absent");
+        AddArgument("salt", &mSalt,
+                    "Salt payload encoded in hexadecimal. Random salt will be generated if absent. "
+                    "This needs to be present if verifier is provided, corresponding to salt used for generating verifier");
         AddArgument("verifier", &mVerifier,
                     "PAKE Passcode verifier encoded in hexadecimal format. Will be generated from random setup pin and other "
                     "params if absent");
--- a/examples/fabric-admin/commands/pairing/OpenCommissioningWindowCommand.h
+++ b/examples/fabric-admin/commands/pairing/OpenCommissioningWindowCommand.h
@@ -47,6 +47,10 @@ public:
                     &mIteration, "Number of PBKDF iterations to use to derive the verifier.  Ignored if 'option' is 0.");
         AddArgument("discriminator", 0, 4096, &mDiscriminator, "Discriminator to use for advertising.  Ignored if 'option' is 0.");
         AddArgument("timeout", 0, UINT16_MAX, &mTimeout, "Time, in seconds, before this command is considered to have timed out.");
+        AddArgument("salt", &mSalt, "Salt payload encoded in hexadecimal. Random salt will be generated if absent");
+        AddArgument("verifier", &mVerifier,
+                    "PAKE Passcode verifier encoded in hexadecimal format. Will be generated from random setup pin and other "
+                    "params if absent");
     }
 
     void RegisterDelegate(CommissioningWindowDelegate * delegate) { mDelegate = delegate; }
@@ -69,6 +73,8 @@ private:
     uint16_t mDiscriminator;
 
     chip::Optional<uint16_t> mTimeout;
+    chip::Optional<chip::ByteSpan> mSalt;
+    chip::Optional<chip::ByteSpan> mVerifier;
 
     chip::Platform::UniquePtr<chip::Controller::CommissioningWindowOpener> mWindowOpener;
 
--- a/examples/fabric-admin/commands/pairing/OpenCommissioningWindowCommand.h
+++ b/examples/fabric-admin/commands/pairing/OpenCommissioningWindowCommand.h
@@ -35,6 +35,7 @@ public:
     OpenCommissioningWindowCommand(CredentialIssuerCommands * credIssuerCommands) :
         CHIPCommand("open-commissioning-window", credIssuerCommands),
         mOnOpenCommissioningWindowCallback(OnOpenCommissioningWindowResponse, this),
+        mOnOpenCommissioningWindowVerifierCallback(OnOpenCommissioningWindowVerifierResponse, this),
         mOnOpenBasicCommissioningWindowCallback(OnOpenBasicCommissioningWindowResponse, this)
     {
         AddArgument("node-id", 0, UINT64_MAX, &mNodeId, "Node to send command to.");
@@ -79,8 +80,10 @@ private:
     chip::Platform::UniquePtr<chip::Controller::CommissioningWindowOpener> mWindowOpener;
 
     static void OnOpenCommissioningWindowResponse(void * context, NodeId deviceId, CHIP_ERROR status, chip::SetupPayload payload);
+    static void OnOpenCommissioningWindowVerifierResponse(void * context, NodeId deviceId, CHIP_ERROR status);
     static void OnOpenBasicCommissioningWindowResponse(void * context, NodeId deviceId, CHIP_ERROR status);
 
     chip::Callback::Callback<chip::Controller::OnOpenCommissioningWindow> mOnOpenCommissioningWindowCallback;
+    chip::Callback::Callback<chip::Controller::OnOpenCommissioningWindowWithVerifier> mOnOpenCommissioningWindowVerifierCallback;
     chip::Callback::Callback<chip::Controller::OnOpenBasicCommissioningWindow> mOnOpenBasicCommissioningWindowCallback;
 };
--- a/examples/fabric-bridge-app/linux/RpcClient.cpp
+++ b/examples/fabric-bridge-app/linux/RpcClient.cpp
@@ -95,12 +95,36 @@ CHIP_ERROR InitRpcClient(uint16_t rpcServerPort)
     return rpc::client::StartPacketProcessing();
 }
 
-CHIP_ERROR OpenCommissioningWindow(NodeId nodeId)
+CHIP_ERROR OpenCommissioningWindow(NodeId nodeId, uint16_t commissioningTimeout, uint16_t discriminator, uint32_t iterations,
+                                   chip::Optional<chip::ByteSpan> salt, chip::Optional<chip::ByteSpan> verifier)
 {
     ChipLogProgress(NotSpecified, "OpenCommissioningWindow with Node Id 0x:" ChipLogFormatX64, ChipLogValueX64(nodeId));
 
-    chip_rpc_DeviceInfo device;
-    device.node_id = nodeId;
+    chip_rpc_DeviceCommissioningWindowInfo device;
+    device.node_id               = nodeId;
+    device.commissioning_timeout = commissioningTimeout;
+    device.discriminator         = discriminator;
+    device.iterations            = iterations;
+
+    if (salt.HasValue())
+    {
+        if (salt.Value().size() > sizeof(device.salt.bytes))
+        {
+            return CHIP_ERROR_INTERNAL;
+        }
+        memcpy(device.salt.bytes, salt.Value().data(), salt.Value().size());
+        device.salt.size = static_cast<size_t>(salt.Value().size());
+    }
+
+    if (verifier.HasValue())
+    {
+        if (verifier.Value().size() > sizeof(device.verifier.bytes))
+        {
+            return CHIP_ERROR_INTERNAL;
+        }
+        memcpy(device.verifier.bytes, verifier.Value().data(), verifier.Value().size());
+        device.verifier.size = static_cast<size_t>(verifier.Value().size());
+    }
 
     // The RPC call is kept alive until it completes. When a response is received, it will be logged by the handler
     // function and the call will complete.
--- a/examples/fabric-bridge-app/linux/RpcClient.cpp
+++ b/examples/fabric-bridge-app/linux/RpcClient.cpp
@@ -95,35 +95,28 @@ CHIP_ERROR InitRpcClient(uint16_t rpcServerPort)
     return rpc::client::StartPacketProcessing();
 }
 
-CHIP_ERROR OpenCommissioningWindow(NodeId nodeId, uint16_t commissioningTimeout, uint16_t discriminator, uint32_t iterations,
-                                   chip::Optional<chip::ByteSpan> salt, chip::Optional<chip::ByteSpan> verifier)
+CHIP_ERROR OpenCommissioningWindow(CommissioningWindowParams params)
 {
-    ChipLogProgress(NotSpecified, "OpenCommissioningWindow with Node Id 0x:" ChipLogFormatX64, ChipLogValueX64(nodeId));
+    ChipLogProgress(NotSpecified, "OpenCommissioningWindow with Node Id 0x:" ChipLogFormatX64, ChipLogValueX64(params.nodeId));
 
     chip_rpc_DeviceCommissioningWindowInfo device;
-    device.node_id               = nodeId;
-    device.commissioning_timeout = commissioningTimeout;
-    device.discriminator         = discriminator;
-    device.iterations            = iterations;
+    device.node_id               = params.nodeId;
+    device.commissioning_timeout = params.commissioningTimeout;
+    device.discriminator         = params.discriminator;
+    device.iterations            = params.iterations;
 
-    if (salt.HasValue())
+    if (params.salt.HasValue())
     {
-        if (salt.Value().size() > sizeof(device.salt.bytes))
-        {
-            return CHIP_ERROR_INTERNAL;
-        }
-        memcpy(device.salt.bytes, salt.Value().data(), salt.Value().size());
-        device.salt.size = static_cast<size_t>(salt.Value().size());
+        VerifyOrReturnError(params.salt.Value().size() <= sizeof(device.salt.bytes), CHIP_ERROR_BUFFER_TOO_SMALL);
+        memcpy(device.salt.bytes, params.salt.Value().data(), params.salt.Value().size());
+        device.salt.size = static_cast<size_t>(params.salt.Value().size());
     }
 
-    if (verifier.HasValue())
+    if (params.verifier.HasValue())
     {
-        if (verifier.Value().size() > sizeof(device.verifier.bytes))
-        {
-            return CHIP_ERROR_INTERNAL;
-        }
-        memcpy(device.verifier.bytes, verifier.Value().data(), verifier.Value().size());
-        device.verifier.size = static_cast<size_t>(verifier.Value().size());
+        VerifyOrReturnError(params.verifier.Value().size() <= sizeof(device.verifier.bytes), CHIP_ERROR_BUFFER_TOO_SMALL);
+        memcpy(device.verifier.bytes, params.verifier.Value().data(), params.verifier.Value().size());
+        device.verifier.size = static_cast<size_t>(params.verifier.Value().size());
     }
 
     // The RPC call is kept alive until it completes. When a response is received, it will be logged by the handler
--- a/examples/fabric-bridge-app/linux/RpcClient.cpp
+++ b/examples/fabric-bridge-app/linux/RpcClient.cpp
@@ -95,29 +95,9 @@ CHIP_ERROR InitRpcClient(uint16_t rpcServerPort)
     return rpc::client::StartPacketProcessing();
 }
 
-CHIP_ERROR OpenCommissioningWindow(CommissioningWindowParams params)
+CHIP_ERROR OpenCommissioningWindow(chip_rpc_DeviceCommissioningWindowInfo device)
 {
-    ChipLogProgress(NotSpecified, "OpenCommissioningWindow with Node Id 0x:" ChipLogFormatX64, ChipLogValueX64(params.nodeId));
-
-    chip_rpc_DeviceCommissioningWindowInfo device;
-    device.node_id               = params.nodeId;
-    device.commissioning_timeout = params.commissioningTimeout;
-    device.discriminator         = params.discriminator;
-    device.iterations            = params.iterations;
-
-    if (params.salt.HasValue())
-    {
-        VerifyOrReturnError(params.salt.Value().size() <= sizeof(device.salt.bytes), CHIP_ERROR_BUFFER_TOO_SMALL);
-        memcpy(device.salt.bytes, params.salt.Value().data(), params.salt.Value().size());
-        device.salt.size = static_cast<size_t>(params.salt.Value().size());
-    }
-
-    if (params.verifier.HasValue())
-    {
-        VerifyOrReturnError(params.verifier.Value().size() <= sizeof(device.verifier.bytes), CHIP_ERROR_BUFFER_TOO_SMALL);
-        memcpy(device.verifier.bytes, params.verifier.Value().data(), params.verifier.Value().size());
-        device.verifier.size = static_cast<size_t>(params.verifier.Value().size());
-    }
+    ChipLogProgress(NotSpecified, "OpenCommissioningWindow with Node Id 0x" ChipLogFormatX64, ChipLogValueX64(device.node_id));
 
     // The RPC call is kept alive until it completes. When a response is received, it will be logged by the handler
     // function and the call will complete.
@@ -130,4 +110,36 @@ CHIP_ERROR OpenCommissioningWindow(CommissioningWindowParams params)
     }
 
     return WaitForResponse(call);
+}
+
+CHIP_ERROR
+OpenCommissioningWindow(chip::Controller::CommissioningWindowPasscodeParams params)
+{
+    chip_rpc_DeviceCommissioningWindowInfo device;
+    device.node_id               = params.GetNodeId();
+    device.commissioning_timeout = params.GetTimeout().count();
+    device.discriminator         = params.GetDiscriminator();
+    device.iterations            = params.GetIteration();
+
+    return OpenCommissioningWindow(device);
+}
+
+CHIP_ERROR
+OpenCommissioningWindow(chip::Controller::CommissioningWindowVerifierParams params)
+{
+    chip_rpc_DeviceCommissioningWindowInfo device;
+    device.node_id               = params.GetNodeId();
+    device.commissioning_timeout = params.GetTimeout().count();
+    device.discriminator         = params.GetDiscriminator();
+    device.iterations            = params.GetIteration();
+
+    VerifyOrReturnError(params.GetSalt().size() <= sizeof(device.salt.bytes), CHIP_ERROR_BUFFER_TOO_SMALL);
+    memcpy(device.salt.bytes, params.GetSalt().data(), params.GetSalt().size());
+    device.salt.size = static_cast<size_t>(params.GetSalt().size());
+
+    VerifyOrReturnError(params.GetVerifier().size() <= sizeof(device.verifier.bytes), CHIP_ERROR_BUFFER_TOO_SMALL);
+    memcpy(device.verifier.bytes, params.GetVerifier().data(), params.GetVerifier().size());
+    device.verifier.size = static_cast<size_t>(params.GetVerifier().size());
+
+    return OpenCommissioningWindow(device);
 }
--- a/examples/fabric-bridge-app/linux/include/RpcClient.h
+++ b/examples/fabric-bridge-app/linux/include/RpcClient.h
@@ -41,4 +41,5 @@ CHIP_ERROR InitRpcClient(uint16_t rpcServerPort);
  * - CHIP_ERROR_BUSY: Another commissioning window is currently in progress.
  * - CHIP_ERROR_INTERNAL: An internal error occurred.
  */
-CHIP_ERROR OpenCommissioningWindow(chip::NodeId nodeId);
+CHIP_ERROR OpenCommissioningWindow(chip::NodeId nodeId, uint16_t commissioningTimeout, uint16_t discriminator, uint32_t iterations,
+                                   chip::Optional<chip::ByteSpan> salt, chip::Optional<chip::ByteSpan> verifier);
--- a/examples/fabric-bridge-app/linux/include/RpcClient.h
+++ b/examples/fabric-bridge-app/linux/include/RpcClient.h
@@ -32,14 +32,24 @@ constexpr uint16_t kFabricAdminServerPort = 33001;
  */
 CHIP_ERROR InitRpcClient(uint16_t rpcServerPort);
 
+struct CommissioningWindowParams
+{
+    chip::NodeId nodeId;
+    uint16_t commissioningTimeout;
+    uint16_t discriminator;
+    uint32_t iterations;
+    chip::Optional<chip::ByteSpan> salt     = chip::NullOptional;
+    chip::Optional<chip::ByteSpan> verifier = chip::NullOptional;
+};
+
 /**
  * Opens a commissioning window for a specified node.
  *
- * @param nodeId The identifier of the node for which the commissioning window should be opened.
+ * @param params    Params for opening the commissioning window on given node.
  * @return CHIP_ERROR An error code indicating the success or failure of the operation.
  * - CHIP_NO_ERROR: The RPC command was successfully processed.
  * - CHIP_ERROR_BUSY: Another commissioning window is currently in progress.
  * - CHIP_ERROR_INTERNAL: An internal error occurred.
  */
-CHIP_ERROR OpenCommissioningWindow(chip::NodeId nodeId, uint16_t commissioningTimeout, uint16_t discriminator, uint32_t iterations,
-                                   chip::Optional<chip::ByteSpan> salt, chip::Optional<chip::ByteSpan> verifier);
+CHIP_ERROR
+OpenCommissioningWindow(CommissioningWindowParams params);
--- a/examples/fabric-bridge-app/linux/include/RpcClient.h
+++ b/examples/fabric-bridge-app/linux/include/RpcClient.h
@@ -18,6 +18,7 @@
 
 #pragma once
 
+#include <controller/CommissioningWindowParams.h>
 #include <platform/CHIPDeviceLayer.h>
 
 constexpr uint16_t kFabricAdminServerPort = 33001;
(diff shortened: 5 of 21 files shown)


src/controller/CommissioningWindowOpener.h
```diff
@@ -104,8 +106,8 @@ class CommissioningWindowOpener
      */
     CHIP_ERROR OpenCommissioningWindow(NodeId deviceId, System::Clock::Seconds16 timeout, uint32_t iteration,
                                        uint16_t discriminator, Optional<uint32_t> setupPIN, Optional<ByteSpan> salt,
-                                       Callback::Callback<OnOpenCommissioningWindow> * callback, SetupPayload & payload,
-                                       bool readVIDPIDAttributes = false);
+                                       Optional<ByteSpan> verifier, Callback::Callback<OnOpenCommissioningWindow> * callback,
```
andy31415, line 109: we had 9 parameters to a method. Now we have 10 ... why is this ok to keep going this route? This is not easy to read and maintain. For example

```
 mWindowOpener->OpenCommissioningWindow(mNodeId, System::Clock::Seconds16(mCommissioningWindowTimeout), mIteration,
                                                      mDiscriminator, NullOptional, NullOptional, NullOptional,
                                                      &mOnOpenCommissioningWindowCallback, ignored,
                                                      /* readVIDPIDAttributes */ true);
```

there `NullOptional` is repeated 3 times without any indication of what they mean. Let's start making code cleaner: please create a `params` structure that uses a builder pattern and then you will not need to change every single call site.

Even more, if we are willing to break ABI compatibility (we seem to want to since we are changing this call anyway) we may as well make the params builder the only call that someone can make.

examples/fabric-bridge-app/linux/RpcClient.cpp
```diff
@@ -73,8 +74,31 @@ CHIP_ERROR OpenCommissioningWindow(NodeId nodeId)
...
+    device.node_id               = nodeId;
+    device.commissioning_timeout = commissioningTimeout;
+    device.discriminator         = discriminator;
+    device.iterations            = iterations;
+
+    if (salt.HasValue())
+    {
+        if (salt.Value().size() > sizeof(device.salt.bytes))
+        {
+            return CHIP_ERROR_INTERNAL;
```
andy31415, line 87: insufficient memory or space seems like a better error code than internal.

use CHIP_ERROR_NO_MEMORY or CHIP_ERROR_BUFFER_TOO_SMALL or similar. (I like buffer to small for now)

examples/fabric-bridge-app/linux/include/RpcClient.h
```diff
@@ -41,4 +41,5 @@ CHIP_ERROR InitRpcClient(uint16_t rpcServerPort);
  * - CHIP_ERROR_BUSY: Another commissioning window is currently in progress.
  * - CHIP_ERROR_INTERNAL: An internal error occurred.
  */
-CHIP_ERROR OpenCommissioningWindow(chip::NodeId nodeId);
+CHIP_ERROR OpenCommissioningWindow(chip::NodeId nodeId, uint16_t commissioningTimeout, uint16_t discriminator, uint32_t iterations,
```
andy31415, line 44: This is a method with 6 arguments. I say after 3-4 arguments we should re-think how they are passed in.

examples/fabric-bridge-app/linux/main.cpp
```diff
@@ -74,7 +74,7 @@ void BridgePollingThread()
 #if defined(PW_RPC_FABRIC_BRIDGE_SERVICE) && PW_RPC_FABRIC_BRIDGE_SERVICE
             else if (ch == 'o')
             {
-                CHIP_ERROR err = OpenCommissioningWindow(0x1234);
+                CHIP_ERROR err = OpenCommissioningWindow(0x1234, 300, 3840, 1000, chip::NullOptional, chip::NullOptional);
```
andy31415, line 77: it is totally unreadable what 300 and 3840 and 1000 or the various NullOptional actually mean. Use param builders.

andy31415 (dismissed):
Requesting changes on making arguments to functions readable and argument count reasonable.

examples/fabric-bridge-app/linux/main.cpp
```diff
@@ -129,26 +129,45 @@ void AdministratorCommissioningCommandHandler::InvokeCommand(HandlerContext & ha
...
+    chip::Optional<chip::ByteSpan> salt;
+    chip::Optional<chip::ByteSpan> verifier;
+    Device * device = nullptr;
+    Status status   = Status::Failure;
+
+    Commands::OpenCommissioningWindow::DecodableType commandData;
+    CHIP_ERROR tlvError = DataModel::Decode(handlerContext.mPayload, commandData);
+    SuccessOrExit(tlvError);
+
+    commissioningTimeout = commandData.commissioningTimeout;
```
andy31415, line 145: why do you separate declaration of variables from assignment? I assume because of successOrExit... can we avoid that pattern?

Do we need separate variables? why not use commandData.discriminator and such directly and not have separate variables?

samadDotDev:
@andy31415 I intentionally held off on doing the argument refactor / params class introduction since that's already done by @yufengwangca in https://github.com/project-chip/connectedhomeip/pull/33799. However, since that PR is blocked due to other reasons and this is requested here, I can pull in those changes here if that's okay. 

Additionally, I can go either way with breaking/maintaining ABI compatibility. Most of the callers of this function are currently passing around `NullOptional` and unnamed args so it may be useful for everyone to just use the new and only function with params class.

src/controller/CommissioningWindowOpener.h
bzbarsky-apple, line 109: This makes no sense.  There is never a situation in which you would both pass in the PIN _and_ pass in the verifier, so why do we have a method that lets you pass both?

It sounds like we need to have two separate methods, and the impl gets the verifier (provided or generated) and then calls a common impl method.

bzbarsky-apple (changes requested):

src/controller/CommissioningWindowOpener.h
samadDotDev: With the params refactor, here's the new shape of the two functions I'm thinking of, let me know your feedback

```cpp

    struct CommissioningWindowParams
    {
        // The device Id of the node.
        NodeId deviceId;
        // The duration for which the commissioning window should remain open.
        System::Clock::Seconds16 timeout;
        // The PAKE iteration count associated with the PAKE Passcode ID and ephemeral PAKE passcode verifier to be used for this
        // commissioning.
        uint32_t iteration;
        // The long discriminator for the DNS-SD advertisement.
        uint16_t discriminator;
        // The salt to use, or NullOptional to use a randomly-generated one. If provided, must be at least
        // kSpake2p_Min_PBKDF_Salt_Length bytes and at most kSpake2p_Max_PBKDF_Salt_Length bytes in length.
        Optional<ByteSpan> salt = NullOptional;
        // Should the API internally read VID and PID from the device while opening the commissioning window.  If this argument is
        // `true`, the API will read VID and PID from the device and include them in the setup payload passed to the callback.
        bool readVIDPIDAttributes = false;
        Callback::Callback<OnOpenCommissioningWindow> *
            callback; // The function to be called on success or failure of opening the commissioning window.
    };

    /**
     * @brief
     *   Try to look up the device attached to our controller with the given
     *   node id and ask it to re-enter commissioning mode with a PASE verifier
     *   derived from the given information and the given discriminator. The
     *   device will exit commissioning mode after a successful commissioning,
     *   or after the given `timeout` time.
     *
     * @param[in] params        The common parameters required to open the commissioning window.
     * @param[in] setupPIN      The setup PIN to use, or NullOptional to use a randomly-generated one.
     * @param[out] payload      The setup payload, not including the VID/PID bits,
     *                          even if those were asked for, that is generated
     *                          based on the passed-in information.  The payload
     *                          provided to the callback function, unlike this
     *                          out parameter, will include the VID/PID bits if
     *                          readVIDPIDAttributes is true.
     */
    CHIP_ERROR OpenCommissioningWindow(const CommissioningWindowParams & params, Optional<uint32_t> setupPIN,
                                       SetupPayload & payload);

    /**
     * @brief
     *   Try to look up the device attached to our controller with the given
     *   node id and ask it to re-enter commissioning mode with a PASE verifier
     *   derived from the given information and the given discriminator. The
     *   device will exit commissioning mode after a successful commissioning,
     *   or after the given `timeout` time.
     *
     * @param[in] params        The common parameters required to open the commissioning window.
     * @param[in] verifier      The PAKE passcode verifier to use, or NullOptional to generate verifier based on randomly generated
     *                          PIN and other parameters.
     * @param[out] payload      The setup payload, not including the VID/PID bits,
     *                          even if those were asked for, that is generated
     *                          based on the passed-in information.  The payload
     *                          provided to the callback function, unlike this
     *                          out parameter, will include the VID/PID bits if
     *                          readVIDPIDAttributes is true.
     */
    CHIP_ERROR OpenCommissioningWindow(const CommissioningWindowParams & params, Optional<ByteSpan> verifier,
                                       SetupPayload & payload);
```

bzbarsky-apple: I think there are a few issues.  First, optional for salt (which means "use a random one" if not provided) does not make sense with a verifier, because the salt needs to match the verifier.  

I think there are two ways of presenting the "things verifier will be derived from":

1. Iteration count, optional passcode (mis-named PIN in this API), optional salt.
2. Iteration count, salt, verifier that corresponds to the provided iteration count and salt and some not-provided passcode.

In terms of modeling this in the API, seems like the cleanest thing are two separate object classes to encapsulate those two separate pieces of information, right?  Whether then we do some sort of static dispatch via overloads or have them inherit from a class that has a pure virtual method that outputs the iteration count, salt, and verifier is something to think about.

Second issue: there is no way to create a SetupPayload from a verifier: the whole point of a SetupPayload is to know the passcode so you can do PASE.  So the API version that takes a verifier can't have a SetupPayload outparam.

Third issue: The version of the API with a verifier can't pass a SetupPayload to the callback function, for exactly the same reasons.

samadDotDev: Thanks for detailing these. Indeed I came across the second issue during the refactor and that led to digging in for the third one which seems to be even more problematic.

For the first one, I'm leaning towards just taking the salt out of common params class, and making it part of first API (with passcode) as optional, and second one (with verifier) as required.

Second: the setup payload doesn't need to be an outparam for the second API, and the current consumer of it (fabric-admin) just ignores that anyway so we can safely remove this.

Third: This is where I'm torn between either introducing two different callback functions or just making the setup payload optional in the existing one. Essentially either
```cpp
typedef void (*OnOpenCommissioningWindowWithPasscode)(void * context, NodeId deviceId, CHIP_ERROR status, SetupPayload payload);
typedef void (*OnOpenCommissioningWindowWithVerifier)(void * context, NodeId deviceId, CHIP_ERROR status);
typedef void (*OnOpenBasicCommissioningWindow)(void * context, NodeId deviceId, CHIP_ERROR status);
```
(we could also keep the default `OnOpenCommissioningWindowWithPasscode` as `OnOpenCommissioningWindow` to keep the diff small)

or
```cpp
typedef void (*OnOpenCommissioningWindow)(void * context, NodeId deviceId, CHIP_ERROR status, Optional<SetupPayload payload>);
typedef void (*OnOpenBasicCommissioningWindow)(void * context, NodeId deviceId, CHIP_ERROR status);
```

Callback function can then also be taken out of common params depending on the solution opted for (3)

bzbarsky-apple: At that point, it seems like we might just want to have two different params classes (which perhaps have a common superclass for the shared bits if desired).  Having different callback function types for the two cases makes the most sense to me.

Keeping the existing API working on top of the new one seems pretty easy and would reduce the amount of code churn SDK clients have to have.

examples/fabric-bridge-app/linux/RpcClient.cpp
samadDotDev: Addressed in https://github.com/project-chip/connectedhomeip/pull/34209/commits/d29e4fc464816478013110707686ad8d75a5d806

src/controller/CommissioningWindowOpener.h
samadDotDev: Added a params class and brought back the stable API for external consumers in https://github.com/project-chip/connectedhomeip/pull/34209/commits/5d4ce206b24d3a3a4201e6d1cdc816ca6e7f8b15

samadDotDev: Addressed these changes in https://github.com/project-chip/connectedhomeip/pull/34209/commits/5d4ce206b24d3a3a4201e6d1cdc816ca6e7f8b15, LMK if there's anything else missing or if it should be done differently.

examples/fabric-bridge-app/linux/main.cpp
samadDotDev: Fixed in https://github.com/project-chip/connectedhomeip/pull/34209/commits/5d4ce206b24d3a3a4201e6d1cdc816ca6e7f8b15

examples/fabric-bridge-app/linux/main.cpp
samadDotDev: Introduced separate param builder for example in https://github.com/project-chip/connectedhomeip/pull/34209/commits/d29e4fc464816478013110707686ad8d75a5d806

examples/fabric-bridge-app/linux/include/RpcClient.h
samadDotDev: Introduced param builder for example in https://github.com/project-chip/connectedhomeip/pull/34209/commits/d29e4fc464816478013110707686ad8d75a5d806

examples/fabric-bridge-app/linux/include/RpcClient.h
```diff
@@ -32,13 +32,24 @@ constexpr uint16_t kFabricAdminServerPort = 33001;
  */
 CHIP_ERROR InitRpcClient(uint16_t rpcServerPort);
 
+struct CommissioningWindowParams
```
andy31415, line 35: please make this a builder pattern. It should read code like: 

```cpp
OCW(
CommisioningWindowParams()
   .SetNodeId(1234)
   .SetDiscriminator(123)
   .SetIterations(100)
   )
```

- [ ] make sure all members have initializers (if you do not know or they are required, set to 0 just in case)
- [ ] make this fluent builders to make it readable and not require pre-declaration
- [ ] It seems that @bzbarsky-apple mentioned some things are not supposed to work together. Add tests for that in the setters (probably VerifyOrDie is fine to avoid mis-use ... but at a minimum class coments describing use should exist here)

src/controller/CommissioningWindowOpener.h
```diff
@@ -107,6 +163,38 @@ class CommissioningWindowOpener
...
+     * @param[in] params        The parameters required to open an enhanced commissioning window
+     *                          with the provided or generated passcode.
+     * @param[out] payload      The setup payload, not including the VID/PID bits,
+     *                          even if those were asked for, that is generated
+     *                          based on the passed-in information.  The payload
+     *                          provided to the callback function, unlike this
+     *                          out parameter, will include the VID/PID bits if
+     *                          readVIDPIDAttributes is true.
+     */
+    CHIP_ERROR OpenCommissioningWindow(const CommissioningWindowPasscodeParams & params, SetupPayload & payload);
```
tcarmelveilleux, line 183: Why have a passcode version? The verifier version beloew is sufficient.

src/controller/CommissioningWindowOpener.h
```diff
@@ -107,6 +163,38 @@ class CommissioningWindowOpener
...
+     *   Try to look up the device attached to our controller with the given
+     *   node id and ask it to re-enter commissioning mode with a PASE verifier
+     *   derived from the given information and the given discriminator. The
+     *   device will exit commissioning mode after a successful commissioning,
+     *   or after the given `timeout` time.
+     *
+     * @param[in] params    The parameters required to open an enhanced commissioning window
+     *                      with the provided PAKE passcode verifier.
+     */
+    CHIP_ERROR OpenCommissioningWindow(const CommissioningWindowVerifierParams & params);
```
tcarmelveilleux, line 196: Please add tests that cover the new methods.

examples/fabric-bridge-app/linux/include/RpcClient.h
samadDotDev: Added fluent builder pattern, initialized some of the member variables, and added some commentary in https://github.com/project-chip/connectedhomeip/pull/34209/commits/16bf72cd708ca52afceff75c669f65e99068489b, LMK if we need to go further in this route.

src/controller/CommissioningWindowOpener.h
```diff
@@ -47,6 +48,61 @@ class CommissioningWindowOpener
         mDeviceConnectionFailure(&OnDeviceConnectionFailureCallback, this)
     {}
 
+    struct CommissioningWindowCommonParams
+    {
+        // The device Id of the node.
```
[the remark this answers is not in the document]
samadDotDev: Remnants from previous args, updated in the new parameters class here: https://github.com/project-chip/connectedhomeip/pull/34209/commits/16bf72cd708ca52afceff75c669f65e99068489b

src/controller/CommissioningWindowOpener.h
samadDotDev: This is to move off the previous function with growing number of args (for the passcode version) as requested in https://github.com/project-chip/connectedhomeip/pull/34209#discussion_r1667046245. The existing API is there only to reduce the external SDK client churn (could potentially mark it as _deprecated_ in some form, or note for consumers to move off that?)

src/controller/CommissioningWindowOpener.cpp
```diff
@@ -124,6 +136,37 @@ CHIP_ERROR CommissioningWindowOpener::OpenCommissioningWindow(NodeId deviceId, S
...
+    mSetupPayload.discriminator.SetLongValue(params.GetDiscriminator());
+    mSetupPayload.rendezvousInformation.SetValue(RendezvousInformationFlag::kOnNetwork);
+
+    mNextStep = Step::kOpenCommissioningWindow;
+
+    return mController->GetConnectedDevice(mNodeId, &mDeviceConnected, &mDeviceConnectionFailure);
+}
+
 CHIP_ERROR CommissioningWindowOpener::OpenCommissioningWindowInternal(Messaging::ExchangeManager & exchangeMgr,
                                                                       const SessionHandle & sessionHandle)
```
yufengwangca, line 171: All operations are still targeted to RootEndpoint, how the bridge route OCW command to the bridged device after receiving it?

samadDotDev: In this context, there is no concept of directing at a specific endpoint once the bridge has mapped the original command (or any requests in general) from the received endpoint to the respective node being bridged. The aforementioned code is being used by controller part of fabric-admin app, which is forwarding a command to its connected device referenced by node id at the root endpoint.

Here's how it's working in the example fabric-admin and fabric-bridge apps (mostly for posterity and others reference since I realize you're already aware of this):


A. Fabric admin commissions a device with nodeId 2

```
[1720416217.588] [2568037:2568040] [CTL] Commissioning complete for node ID 0x0000000000000002: success
New device with Node ID: 0x2 has been successfully added.
[1720416217.588] [2568037:2568040] [-] AddSynchronizedDevice
[1720416217.658] [2568037:2568042] [-] AddSynchronizedDevice RPC call succeeded!
```

B. Fabric bridge adds the device to its parts list (maps nodeId 2 to endpoint 3)
```
[1720416217.632] [2568020:2568022] [-] Received AddSynchronizedDevice: 0000000000000002
[1720416217.658] [2568020:2568022] [-] Added device with nodeId=0x0000000000000002 to dynamic endpoint 3 (index=1)
```


C. Fabric bridge receives the OCW request directed at endpoint 3
```
[1720416217.745] [2568020:2568020] [DMG] InvokeRequestMessage =
[1720416217.745] [2568020:2568020] [DMG] {
[1720416217.745] [2568020:2568020] [DMG] 	suppressResponse = false,
[1720416217.745] [2568020:2568020] [DMG] 	timedRequest = true,
[1720416217.745] [2568020:2568020] [DMG] 	InvokeRequests =
[1720416217.745] [2568020:2568020] [DMG] 	[
[1720416217.745] [2568020:2568020] [DMG] 		CommandDataIB =
[1720416217.745] [2568020:2568020] [DMG] 		{
[1720416217.745] [2568020:2568020] [DMG] 			CommandPathIB =
[1720416217.745] [2568020:2568020] [DMG] 			{
[1720416217.745] [2568020:2568020] [DMG] 				EndpointId = 0x3,
[1720416217.745] [2568020:2568020] [DMG] 				ClusterId = 0x3c,
[1720416217.745] [2568020:2568020] [DMG] 				CommandId = 0x0,
[1720416217.745] [2568020:2568020] [DMG] 			},
[1720416217.745] [2568020:2568020] [DMG]
[1720416217.745] [2568020:2568020] [DMG] 			CommandFields =
[1720416217.745] [2568020:2568020] [DMG] 			{
[1720416217.745] [2568020:2568020] [DMG] 				0x0 = 900 (unsigned),
[1720416217.745] [2568020:2568020] [DMG] 				0x1 = [
[1720416217.745] [2568020:2568020] [DMG] 						0x73, 0xff, 0x12, 0xc4, 0x33, 0x84, 0x08, 0xf2, 0xff, 0xa0, 0x25, 0x0c, 0x15, 0x8a, 0xc4, 0x98, 0xd6, 0x9b, 0xb5, 0x56, 0x59, 0x7e, 0x25, 0x3c, 0x58, 0xc9, 0x87, 0xbc, 0x8c, 0x0b, 0x0c, 0xaf, 0x04, 0x09, 0x96, 0x35, 0x75, 0x2b, 0xac, 0xc1, 0x17, 0xaf, 0xfc, 0x81, 0xdc, 0x9e, 0x4f, 0xa4, 0x0b, 0xca, 0xc1, 0x64, 0xb7, 0xa4, 0xb0, 0x9e, 0xa5, 0x2b, 0xfc, 0x40, 0x44, 0x24, 0xc8, 0x09, 0x45, 0x56, 0x84, 0x76, 0xe1, 0x53, 0x35, 0x47, 0x6d, 0xb2, 0xb9, 0x36, 0x59, 0x4f, 0xd1, 0xbd, 0xad, 0x73, 0x29, 0x9f, 0xfd, 0xb7, 0x41, 0x74, 0x6e, 0x6a, 0x0a, 0x49, 0x9f, 0xf1, 0x5c, 0x3c, 0x97,
[1720416217.745] [2568020:2568020] [DMG] 				] (97 bytes)
[1720416217.745] [2568020:2568020] [DMG] 				0x2 = 995 (unsigned),
[1720416217.745] [2568020:2568020] [DMG] 				0x3 = 10000 (unsigned),
[1720416217.745] [2568020:2568020] [DMG] 				0x4 = [
[1720416217.745] [2568020:2568020] [DMG] 						0xd8, 0x07, 0x0d, 0xbf, 0x38, 0xaf, 0x85, 0xef, 0x1e, 0x3b, 0x30, 0x2d, 0xfd, 0x68, 0x77, 0x54, 0xd3, 0xcf, 0xe3, 0x05, 0x7c, 0x13, 0x09, 0x87, 0xa6, 0x0f, 0x51, 0xcb, 0x43, 0x9e, 0xaf, 0x62,
[1720416217.745] [2568020:2568020] [DMG] 				] (32 bytes)
[1720416217.745] [2568020:2568020] [DMG] 			},
[1720416217.745] [2568020:2568020] [DMG] 		},
[1720416217.745] [2568020:2568020] [DMG]
[1720416217.745] [2568020:2568020] [DMG] 	],
[1720416217.745] [2568020:2568020] [DMG]
[1720416217.745] [2568020:2568020] [DMG] 	InteractionModelRevision = 11
[1720416217.745] [2568020:2568020] [DMG] },
...
[1720416217.745] [2568020:2568020] [-] Received command to open commissioning window on Endpoint: 3
```

D. Fabric bridge looks up its mapping for bridged devices (endpointId 3 -> nodeId 2) and generates the RPC call to fabric-admin app for that node:
```
[1720416217.745] [2568020:2568020] [-] OpenCommissioningWindow with Node Id 0x:0000000000000002
```

E. Fabric admin receives the RPC call and pushes it as a command (2 is the node id for device on fabric-admin controller)
```
[1720416217.788] [2568037:2568038] [-] Received OpenCommissioningWindow request: 0x2
[1720416217.788] [2568037:2568038] [-] PushCommand: pairing open-commissioning-window 2 0 1 900 10000 995 --salt hex:d8070dbf38af85ef1e3b302dfd687754d3cfe3057c130987a60f51cb439eaf62 --verifier hex:73ff12c4338408f2ffa0250c158ac498d69bb556597e253c58c987bc8c0b0caf04099635752bacc117affc81dc9e4fa40bcac164b7a4b09ea52bfc404424c80945568476e15335476db2b936594fd1bdad73299ffdb741746e6a0a499ff15c3c97
[1720416217.788] [2568037:2568037] [-] Command: pairing open-commissioning-window 2 0 1 900 10000 995 --salt hex:d8070dbf38af85ef1e3b302dfd687754d3cfe3057c130987a60f51cb439eaf62 --verifier hex:73ff12c4338408f2ffa0250c158ac498d69bb556597e253c58c987bc8c0b0caf04099635752bacc117affc81dc9e4fa40bcac164b7a4b09ea52bfc404424c80945568476e15335476db2b936594fd1bdad73299ffdb741746e6a0a499ff15c3c97
```

F. Fabric Admin pushes the command to the respective device using the `CommissioningWindowOpener` in question (notice the endpoint is 0 as expected for the node itself):
```
[1720416217.792] [2568037:2568040] [EM] <<< [E:15423i S:63557 M:145490627 (Ack:195235134)] (S) Msg TX to 1:0000000000000002 [F6C8] [UDP:[fe80::f184:6119:bc23:f7f0%tailscale0]:5549] --- Type 0001:08 (IM:InvokeCommandRequest)
[1720416217.792] [2568037:2568040] [DMG] ICR moving to [AwaitingRe]
[1720416217.868] [2568037:2568040] [EM] >>> [E:15423i S:63557 M:195235135 (Ack:145490627)] (S) Msg RX from 1:0000000000000002 [F6C8] --- Type 0001:09 (IM:InvokeCommandResponse)
[1720416217.868] [2568037:2568040] [EM] Found matching exchange: 15423i, Delegate: 0x706a6c0182e8
[1720416217.868] [2568037:2568040] [EM] Rxd Ack; Removing MessageCounter:145490627 from Retrans Table on exchange 15423i
[1720416217.868] [2568037:2568040] [DMG] ICR moving to [ResponseRe]
[1720416217.868] [2568037:2568040] [DMG] InvokeResponseMessage =
[1720416217.868] [2568037:2568040] [DMG] {
[1720416217.869] [2568037:2568040] [DMG] 	suppressResponse = false,
[1720416217.869] [2568037:2568040] [DMG] 	InvokeResponseIBs =
[1720416217.869] [2568037:2568040] [DMG] 	[
[1720416217.869] [2568037:2568040] [DMG] 		InvokeResponseIB =
[1720416217.869] [2568037:2568040] [DMG] 		{
[1720416217.869] [2568037:2568040] [DMG] 			CommandStatusIB =
[1720416217.869] [2568037:2568040] [DMG] 			{
[1720416217.869] [2568037:2568040] [DMG] 				CommandPathIB =
[1720416217.869] [2568037:2568040] [DMG] 				{
[1720416217.869] [2568037:2568040] [DMG] 					EndpointId = 0x0,
[1720416217.869] [2568037:2568040] [DMG] 					ClusterId = 0x3c,
[1720416217.869] [2568037:2568040] [DMG] 					CommandId = 0x0,
[1720416217.869] [2568037:2568040] [DMG] 				},
[1720416217.869] [2568037:2568040] [DMG]
[1720416217.869] [2568037:2568040] [DMG] 				StatusIB =
[1720416217.869] [2568037:2568040] [DMG] 				{
[1720416217.869] [2568037:2568040] [DMG] 					status = 0x00 (SUCCESS),
[1720416217.869] [2568037:2568040] [DMG] 				},
[1720416217.869] [2568037:2568040] [DMG]
[1720416217.869] [2568037:2568040] [DMG] 			},
[1720416217.869] [2568037:2568040] [DMG]
[1720416217.869] [2568037:2568040] [DMG] 		},
[1720416217.869] [2568037:2568040] [DMG]
[1720416217.869] [2568037:2568040] [DMG] 	],
[1720416217.869] [2568037:2568040] [DMG]
[1720416217.869] [2568037:2568040] [DMG] 	InteractionModelRevision = 11
[1720416217.869] [2568037:2568040] [DMG] },
[1720416217.869] [2568037:2568040] [DMG] Received Command Response Status for Endpoint=0 Cluster=0x0000_003C Command=0x0000_0000 Status=0x0
[1720416217.869] [2568037:2568040] [CTL] Successfully opened pairing window on the device
```

G. The end device gets the command on root endpoint
```
[1720416217.805] [2568097:2568097] [EM] >>> [E:15423r S:42682 M:145490627 (Ack:195235134)] (S) Msg RX from 1:000000000001B669 [F6C8] --- Type 0001:08 (IM:InvokeCommandRequest)
[1720416217.805] [2568097:2568097] [EM] Found matching exchange: 15423r, Delegate: 0x5b1a4e4c7020
[1720416217.805] [2568097:2568097] [EM] Rxd Ack; Removing MessageCounter:195235134 from Retrans Table on exchange 15423r
[1720416217.805] [2568097:2568097] [DMG] Timed following action arrived at 0x0000000011AB93C5: handler 0x5b1a4e4c7020 exchange 15423r
[1720416217.805] [2568097:2568097] [DMG] Handing timed invoke to IM engine: handler 0x5b1a4e4c7020 exchange 15423r
[1720416217.805] [2568097:2568097] [DMG] InvokeRequestMessage =
[1720416217.805] [2568097:2568097] [DMG] {
[1720416217.805] [2568097:2568097] [DMG] 	suppressResponse = false,
[1720416217.805] [2568097:2568097] [DMG] 	timedRequest = true,
[1720416217.805] [2568097:2568097] [DMG] 	InvokeRequests =
[1720416217.805] [2568097:2568097] [DMG] 	[
[1720416217.805] [2568097:2568097] [DMG] 		CommandDataIB =
[1720416217.805] [2568097:2568097] [DMG] 		{
[1720416217.805] [2568097:2568097] [DMG] 			CommandPathIB =
[1720416217.805] [2568097:2568097] [DMG] 			{
[1720416217.805] [2568097:2568097] [DMG] 				EndpointId = 0x0,
[1720416217.805] [2568097:2568097] [DMG] 				ClusterId = 0x3c,
[1720416217.805] [2568097:2568097] [DMG] 				CommandId = 0x0,
[1720416217.805] [2568097:2568097] [DMG] 			},
[1720416217.805] [2568097:2568097] [DMG]
[1720416217.805] [2568097:2568097] [DMG] 			CommandFields =
[1720416217.805] [2568097:2568097] [DMG] 			{
[1720416217.805] [2568097:2568097] [DMG] 				0x0 = 900 (unsigned),
[1720416217.805] [2568097:2568097] [DMG] 				0x1 = [
[1720416217.805] [2568097:2568097] [DMG] 					0x73, 0xff, 0x12, 0xc4, 0x33, 0x84, 0x08, 0xf2, 0xff, 0xa0, 0x25, 0x0c, 0x15, 0x8a, 0xc4, 0x98, 0xd6, 0x9b, 0xb5, 0x56, 0x59, 0x7e, 0x25, 0x3c, 0x58, 0xc9, 0x87, 0xbc, 0x8c, 0x0b, 0x0c, 0xaf, 0x04, 0x09, 0x96, 0x35, 0x75, 0x2b, 0xac, 0xc1, 0x17, 0xaf, 0xfc, 0x81, 0xdc, 0x9e, 0x4f, 0xa4, 0x0b, 0xca, 0xc1, 0x64, 0xb7, 0xa4, 0xb0, 0x9e, 0xa5, 0x2b, 0xfc, 0x40, 0x44, 0x24, 0xc8, 0x09, 0x45, 0x56, 0x84, 0x76, 0xe1, 0x53, 0x35, 0x47, 0x6d, 0xb2, 0xb9, 0x36, 0x59, 0x4f, 0xd1, 0xbd, 0xad, 0x73, 0x29, 0x9f, 0xfd, 0xb7, 0x41, 0x74, 0x6e, 0x6a, 0x0a, 0x49, 0x9f, 0xf1, 0x5c, 0x3c, 0x97,
[1720416217.805] [2568097:2568097] [DMG] 				] (97 bytes)
[1720416217.805] [2568097:2568097] [DMG] 				0x2 = 995 (unsigned),
[1720416217.805] [2568097:2568097] [DMG] 				0x3 = 10000 (unsigned),
[1720416217.805] [2568097:2568097] [DMG] 				0x4 = [
[1720416217.805] [2568097:2568097] [DMG] 					0xd8, 0x07, 0x0d, 0xbf, 0x38, 0xaf, 0x85, 0xef, 0x1e, 0x3b, 0x30, 0x2d, 0xfd, 0x68, 0x77, 0x54, 0xd3, 0xcf, 0xe3, 0x05, 0x7c, 0x13, 0x09, 0x87, 0xa6, 0x0f, 0x51, 0xcb, 0x43, 0x9e, 0xaf, 0x62,
[1720416217.805] [2568097:2568097] [DMG] 				] (32 bytes)
[1720416217.805] [2568097:2568097] [DMG] 			},
[1720416217.805] [2568097:2568097] [DMG] 		},
[1720416217.805] [2568097:2568097] [DMG]
[1720416217.805] [2568097:2568097] [DMG] 	],
[1720416217.805] [2568097:2568097] [DMG]
[1720416217.805] [2568097:2568097] [DMG] 	InteractionModelRevision = 11
[1720416217.805] [2568097:2568097] [DMG] },
[1720416217.805] [2568097:2568097] [DMG] AccessControl: checking f=1 a=c s=0x000000000001B669 t= c=0x0000_003C e=0 p=a
[1720416217.805] [2568097:2568097] [DMG] AccessControl: allowed
[1720416217.805] [2568097:2568097] [DMG] Received command for Endpoint=0 Cluster=0x0000_003C Command=0x0000_0000
[1720416217.805] [2568097:2568097] [ZCL] Received command to open commissioning window
```

Complete logs:
[light-app.log](https://github.com/user-attachments/files/16146080/light-app.log)
[fabric-bridge.log](https://github.com/user-attachments/files/16146081/fabric-bridge.log)
[fabric-admin.log](https://github.com/user-attachments/files/16146082/fabric-admin.log)

src/controller/CommissioningWindowOpener.h
samadDotDev: Added some tests for the new methods mostly focused on the args: https://github.com/project-chip/connectedhomeip/pull/34209/commits/5a9ae85462fa9aa2455f0111117ef14ce9950cc4

src/controller/CommissioningWindowOpener.cpp
yufengwangca: Yes, in your step C. Fabric bridge receives the OCW request directed at endpoint 3

I was not able to understand how 'OCW request directed at endpoint 3' is generated at the soruce Fabric Admin if we don't modify OpenCommissioningWindowInternal to specify the target endpoint

samadDotDev: Yeah, the `OpenCommissioningWindowInternal` isn't really invoked at that step by source fabric-admin.

Essentially, step C is getting the endpoint id [here](https://github.com/samadDotDev/connectedhomeip/blob/b154a1b651f627afd514d4a476201ae62d3bdf7f/examples/fabric-bridge-app/linux/main.cpp#L126), then for step D, fabric bridge looks up the device using endpoint id [here](https://github.com/samadDotDev/connectedhomeip/blob/b154a1b651f627afd514d4a476201ae62d3bdf7f/examples/fabric-bridge-app/linux/main.cpp#L147), which is then used to provide the node id to fabric admin over RPC as propagated from [here](https://github.com/samadDotDev/connectedhomeip/blob/b154a1b651f627afd514d4a476201ae62d3bdf7f/examples/fabric-bridge-app/linux/main.cpp#L152).

Eventually, at step F, fabric-admin calls `CommissioningWindowOpener::OpenCommissioningWindow()` [here](https://github.com/samadDotDev/connectedhomeip/blob/b154a1b651f627afd514d4a476201ae62d3bdf7f/examples/fabric-admin/commands/pairing/OpenCommissioningWindowCommand.cpp#L39) (which leads to invoking `OpenCommissioningWindowInternal` at the next step driven by the state machine there) and that point the endpoint 3 isn't important since that context only exists in the bridge, and the controller is already aware of the device (by its node id reference) on which this OCW command is being invoked on the device's root.

Let me know if you have any concerns or if there is anything I should be addressing as part of this PR (or in a follow-up one)?

yufengwangca: Discussed offline, we will continue to use https://github.com/project-chip/connectedhomeip/pull/33799 to track adding endpoint id parameter to the params. 

Scope to the current PR to the ECM params change.

yufengwangca (approved):

examples/chip-tool/commands/pairing/OpenCommissioningWindowCommand.cpp
```diff
@@ -34,10 +34,13 @@ CHIP_ERROR OpenCommissioningWindowCommand::RunCommand()
...
-                                                      mDiscriminator, NullOptional, NullOptional,
-                                                      &mOnOpenCommissioningWindowCallback, ignored,
-                                                      /* readVIDPIDAttributes */ true);
+        return mWindowOpener->OpenCommissioningWindow(Controller::CommissioningWindowPasscodeParams()
+                                                          .SetNodeId(mNodeId)
+                                                          .SetTimeout(mCommissioningWindowTimeout)
+                                                          .SetIteration(mIteration)
+                                                          .SetDiscriminator(mDiscriminator)
+                                                          .SetReadVIDPIDAttributes(true),
+                                                      &mOnOpenCommissioningWindowCallback, ignored);
```
andy31415, line 43: why are the callbacks and setuppayload outside the params structure?

src/controller/CommissioningWindowOpener.h
```diff
@@ -107,6 +108,45 @@ class CommissioningWindowOpener
...
+     *                          generated from provided parameters.
+     * @param[out] payload      The setup payload, not including the VID/PID bits,
+     *                          even if those were asked for, that is generated
+     *                          based on the passed-in information.  The payload
+     *                          provided to the callback function, unlike this
+     *                          out parameter, will include the VID/PID bits if
+     *                          readVIDPIDAttributes is true.
+     */
+    CHIP_ERROR OpenCommissioningWindow(const CommissioningWindowPasscodeParams & params,
+                                       Callback::Callback<OnOpenCommissioningWindow> * callback, SetupPayload & payload);
```
andy31415, line 132: callback should be part of `params`

examples/chip-tool/commands/pairing/OpenCommissioningWindowCommand.cpp
samadDotDev: It is mostly to make the params more generic so places other than this API like the[ RPC functions](https://github.com/project-chip/connectedhomeip/blob/b154a1b651f627afd514d4a476201ae62d3bdf7f/examples/fabric-bridge-app/linux/include/RpcClient.h#L58) which are passing these around internally shouldn't need yet another struct or having to define an ignored out param reference. Let me know if you have strong opinions on making it all part of the same class.

andy31415: why dow e allow RPC to dictate src/controller/CommissioningWindowOpener.h ?

RPC is not the main API in our SDK and it is an optional app-specific extension. src/controller/CommissioningWindowOpener.h is a SDK API, I would prefer to make that one clean to use.

samadDotDev: Moved the callback to params class in https://github.com/project-chip/connectedhomeip/pull/34209/commits/8862cacfbc9519b71496b65ff923ddbd06cb31dc

src/controller/CommissioningWindowOpener.h
samadDotDev: Moved the callback to params in https://github.com/project-chip/connectedhomeip/pull/34209/commits/8862cacfbc9519b71496b65ff923ddbd06cb31dc

examples/common/pigweed/protos/fabric_admin_service.proto
bzbarsky-apple, line 1: I did not review the pigweed bits, or any of the other rpc stuff, and only glanced at the fabric-admin parts.

examples/fabric-admin/commands/pairing/OpenCommissioningWindowCommand.h
```diff
@@ -47,6 +48,10 @@ class OpenCommissioningWindowCommand : public CHIPCommand
                     &mIteration, "Number of PBKDF iterations to use to derive the verifier.  Ignored if 'option' is 0.");
         AddArgument("discriminator", 0, 4096, &mDiscriminator, "Discriminator to use for advertising.  Ignored if 'option' is 0.");
         AddArgument("timeout", 0, UINT16_MAX, &mTimeout, "Time, in seconds, before this command is considered to have timed out.");
+        AddArgument("salt", &mSalt, "Salt payload encoded in hexadecimal. Random salt will be generated if absent");
```
bzbarsky-apple, line 51: This doesn't match the actual code, which requires the salt to be specified in some cases.  I strongly suggest having the docs not be misleading.

src/controller/CHIPDeviceController.h
```diff
@@ -167,6 +167,16 @@ struct CommissionerInitParams : public ControllerInitParams
     Credentials::DeviceAttestationVerifier * deviceAttestationVerifier = nullptr;
 };
 
+// Interface class for DeviceController methods that need to be mocked
+class IDeviceController
```
bzbarsky-apple, line 171: Why is this here?  Nothing ever uses this type except for DeviceController inheriting from it, right?

This needs a better explanation of what this thing is for, at which point we can figure out a reasonable name for it.

src/controller/CommissioningWindowOpener.cpp
```diff
@@ -124,6 +136,36 @@ CHIP_ERROR CommissioningWindowOpener::OpenCommissioningWindow(NodeId deviceId, S
...
+    mBasicCommissioningWindowCallback    = nullptr;
+    mCommissioningWindowCallback         = nullptr;
+    mNodeId                              = params.GetNodeId();
+    mCommissioningWindowTimeout          = params.GetTimeout();
+    mPBKDFIterations                     = params.GetIteration();
+    mCommissioningWindowOption           = CommissioningWindowOption::kTokenWithProvidedPIN;
+    mSetupPayload                        = SetupPayload();
+    mSetupPayload.version                = 0;
+    mSetupPayload.discriminator.SetLongValue(params.GetDiscriminator());
+    mSetupPayload.rendezvousInformation.SetValue(RendezvousInformationFlag::kOnNetwork);
```
bzbarsky-apple, line 162: This part is a bit confusing, because in fact the only part of mSetupPayload used in this mode is `mSetupPayload.discriminator`.  So why are we setting its `rendezvousInformation` at all?

Even better would be if we just stored the discriminator outside the setup payload, used that in `OpenCommissioningWindowInternal`, and did not touch the setup payload at all in this codepath.

src/controller/CommissioningWindowParams.h
```diff
@@ -0,0 +1,185 @@
...
+    {
+        mDiscriminator = discriminator;
+        return static_cast<Derived &>(*this);
+    }
+
+private:
+    NodeId mNodeId                    = 0;
+    System::Clock::Seconds16 mTimeout = System::Clock::Seconds16(300);
+    uint32_t mIteration               = 1000;
+    uint16_t mDiscriminator           = 0;
```
bzbarsky-apple, line 81: A problem with the builder pattern thing is that it's not clear which things are required and which are optional....  For some cases (the timeout/iteration) we are now making them optional and defaulting. But the node id (which should be `kUndefinedNodeId`) we are just silently allowing invalid values in the callee.  And for discriminator we are just defaulting a valid value without any sort of "you have to have set this" sentinel.

What we should do if we do a builder is that anything that is not being defaulted sanely should have a sentinel value at init, and then when one tries to use the value we should fail out if the sentinels are still set.

src/controller/CommissioningWindowParams.h
```diff
@@ -0,0 +1,185 @@
...
+};
+
+class CommissioningWindowPasscodeParams : public CommissioningWindowCommonParams<CommissioningWindowPasscodeParams>
+{
+public:
+    CommissioningWindowPasscodeParams() = default;
+
+    bool HasSetupPIN() const { return mSetupPIN.HasValue(); }
+    // Get the value of salt if present.
```
bzbarsky-apple, line 90: This comment makes no sense.

src/controller/CommissioningWindowParams.h
```diff
@@ -0,0 +1,185 @@
...
+    // Returns 0 if absent, make sure to check HasSetupPIN() if a valid value is required.
```
bzbarsky-apple, line 91: Why not just go ahead and crash on misuse by calling `Value()`, not `ValueOr()`?

src/controller/CommissioningWindowParams.h
```diff
@@ -0,0 +1,185 @@
...
+        mReadVIDPIDAttributes = readVIDPIDAttributes;
+        return *this;
+    }
+
+    Callback::Callback<OnOpenCommissioningWindow> * GetCallback() const { return mCallback; }
+    // The function to be called on success or failure of opening the commissioning window.
+    // This will include the SetupPayload generated from provided parameters.
+    CommissioningWindowPasscodeParams & SetCallback(Callback::Callback<OnOpenCommissioningWindow> * callback)
+    {
+        mCallback = callback;
```
bzbarsky-apple, line 135: See, so here: we need to ensure that this got set, in the thing that uses these params, right?

src/controller/CommissioningWindowParams.h
```diff
@@ -0,0 +1,185 @@
...
+    // commissioning window. This will NOT include the SetupPayload.
+    CommissioningWindowVerifierParams & SetCallback(Callback::Callback<OnOpenCommissioningWindowWithVerifier> * callback)
+    {
+        mCallback = callback;
+        return *this;
+    }
+
+private:
+    ByteSpan mVerifier;
+    ByteSpan mSalt;
```
bzbarsky-apple, line 180: This needs to have sentinels so we properly fail out if these never get initialized.

examples/fabric-admin/commands/pairing/OpenCommissioningWindowCommand.h
samadDotDev: Added a comment that the corresponding salt needs to be present when the verifier is provided: https://github.com/project-chip/connectedhomeip/pull/34209/commits/da28dd643e7419092f7a073ab2011fda251f2504

src/controller/CHIPDeviceController.h
samadDotDev: My bad, I introduced it for mocks in `TestCommissioningWindowOpener.cpp` and ended up overriding the `DeviceController`'s function instead. Just removed this abstract class and added virtual keyword to the function of interest (the one that needs to be overridden for tests): https://github.com/project-chip/connectedhomeip/pull/34209/commits/da28dd643e7419092f7a073ab2011fda251f2504

src/controller/CommissioningWindowOpener.cpp
samadDotDev: Good idea! Tracking discriminator out of setup payload now and using that for the internal function. The value will still be copied for setup payload for the paths that propagate it via out param or callback (passcode one): https://github.com/project-chip/connectedhomeip/pull/34209/commits/da28dd643e7419092f7a073ab2011fda251f2504

src/controller/CommissioningWindowParams.h
samadDotDev: Totally agree, other than some helpful defaults (for iteration/timeout), I have initialized members to either sentinel value (`kUndefinedNodeId`) or made them `Optional` type when the sentinel is either not as clear or if the entire range is valid: https://github.com/project-chip/connectedhomeip/pull/34209/commits/da28dd643e7419092f7a073ab2011fda251f2504

src/controller/CommissioningWindowParams.h
samadDotDev: Copy pasta, fixed in https://github.com/project-chip/connectedhomeip/pull/34209/commits/da28dd643e7419092f7a073ab2011fda251f2504

src/controller/CommissioningWindowParams.h
samadDotDev: 💥 https://github.com/project-chip/connectedhomeip/pull/34209/commits/da28dd643e7419092f7a073ab2011fda251f2504

src/controller/CommissioningWindowParams.h
samadDotDev: Params class can be re-used without callback from the consumer so `nullptr` can be a valid default here. 

Also, this has been existing behavior on the current API if the caller isn't interested in knowing the result for any reason - a nullptr callback can be supplied which isn't validated during request acceptance and ofcourse doesn't get notified by `OnOpenCommissioningWindowSuccess/OnOpenCommissioningWindowFailure` when it's done.

src/controller/CommissioningWindowParams.h
samadDotDev: Changed these bytespan members type to optional and made things go kaboom in accessors with `Value()`: https://github.com/project-chip/connectedhomeip/pull/34209/commits/da28dd643e7419092f7a073ab2011fda251f2504

src/controller/CHIPDeviceController.h
bzbarsky-apple: Ah, this makes sense.  OK, great.

src/controller/CommissioningWindowParams.h
bzbarsky-apple: Ah, OK.  If we support no callback, great.

src/controller/CommissioningWindowParams.h
```diff
@@ -0,0 +1,189 @@
...
+    uint32_t GetIteration() const { return mIteration; }
+    Derived & SetIteration(uint32_t iteration)
+    {
+        mIteration = iteration;
+        return static_cast<Derived &>(*this);
+    }
+
+    // The long discriminator for the DNS-SD advertisement.
+    uint16_t GetDiscriminator() const { return mDiscriminator.Value(); }
+    bool HasDiscriminator() const { return mDiscriminator.HasValue(); }
```
bzbarsky-apple, line 75: Why do we need HasDiscriminator? Does not seem to be used anywhere....

bzbarsky-apple (approved):

src/controller/CommissioningWindowParams.h
samadDotDev: Added the use in https://github.com/project-chip/connectedhomeip/pull/34209/commits/19c0ffef9361764d6881cd9bc5de06317785dfbf, alongwith `HasNodeId()` - returning invalid args error

bzbarsky-apple: Makes sense.

PR #33761, opened 2024-06-05
Fix some random Darwin test failures due to timing dependency.

bzbarsky-apple:
Some Darwin tests failed randomly if a subscription took longer than 10s to establish.  Add test-only conditionals around the relevant code so we suppress that behavior in tests.

PR #34210, opened 2024-07-05
Fix more random Darwin test failures due to timing dependency.

bzbarsky-apple:
The fix in https://github.com/project-chip/connectedhomeip/pull/33761 only fixed the Objective-C version of the test.  The Swift version would still fail if it took longer than 10s to set up a subscription.


jtung-apple (approved):

kiel-apple (approved):

bzbarsky-apple:

andy31415 (approved):

PR #34212, opened 2024-07-05
Clear data from MTRDevice's ClusterStateCache when elements are removed.

bzbarsky-apple:
The idea is for that cache to store DataVersions and cluster data sizes, so on resubscribe we send the right data along.  But ClusterStateCache itself never removes anything, so if we detect that an endpoint or cluster or attribute has been removed, we need to explicitly remove the relevant bits from the cache.

Not doing that can lead to incorrect (so less efficient) DataVersion sorting for the case when an attribute was removed, of to unnecessary DataVersions being included in a subscribe (and potentially preventing useful ones from being included) when a cluster or endpoint was removed.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRBaseSubscriptionCallback.h
+++ b/src/darwin/Framework/CHIP/MTRBaseSubscriptionCallback.h
@@ -93,6 +93,11 @@ public:
 
     chip::app::BufferedReadCallback & GetBufferedCallback() { return mBufferedReadAdapter; }
 
+    // Accessor for the cluster state cache, so that stale data can be cleared
+    // from it as needed.  The returned pointer must not be held long-term,
+    // because it can dangle if this callback object is destroyed.
+    chip::app::ClusterStateCache * GetClusterStateCache() { return mClusterStateCache.get(); }
+
     // We need to exist to get a ReadClient, so can't take this as a constructor argument.
     void AdoptReadClient(std::unique_ptr<chip::app::ReadClient> aReadClient) { mReadClient = std::move(aReadClient); }
     void AdoptClusterStateCache(std::unique_ptr<chip::app::ClusterStateCache> aClusterStateCache)
--- a/src/darwin/Framework/CHIP/MTRBaseSubscriptionCallback.h
+++ b/src/darwin/Framework/CHIP/MTRBaseSubscriptionCallback.h
@@ -22,6 +22,7 @@
 #include <app/BufferedReadCallback.h>
 #include <app/ClusterStateCache.h>
 #include <app/ConcreteAttributePath.h>
+#include <app/ConcreteClusterPath.h>
 #include <app/EventHeader.h>
 #include <app/MessageDef/StatusIB.h>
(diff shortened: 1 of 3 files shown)


src/darwin/Framework/CHIP/MTRBaseSubscriptionCallback.h
```diff
@@ -93,6 +93,11 @@ class MTRBaseSubscriptionCallback : public chip::app::ClusterStateCache::Callbac
 
     chip::app::BufferedReadCallback & GetBufferedCallback() { return mBufferedReadAdapter; }
 
+    // Accessor for the cluster state cache, so that stale data can be cleared
+    // from it as needed.  The returned pointer must not be held long-term,
+    // because it can dangle if this callback object is destroyed.
+    chip::app::ClusterStateCache * GetClusterStateCache() { return mClusterStateCache.get(); }
```
andy31415, line 99: Providing access to the raw pointer seems to require more documentation about lifetime and provides a lot of usage. Any reason we did this instead of providing some `Clear` methods instead?

bzbarsky-apple: I didn't feel it was worth the extra method definitions, and more importantly I could not figure out what to name the methods...  But I'm open to doing whatever here.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -3429,6 +3429,13 @@ - (void)_pruneEndpointsIn:(MTRDeviceDataValueDictionary)previousPartsListValue
         }
         [self _removeClusters:clusterPathsToRemove doRemoveFromDataStore:NO];
         [self.deviceController.controllerDataStore clearStoredClusterDataForNodeID:self.nodeID endpointID:endpoint];
+
+        if (_currentSubscriptionCallback) {
+            auto clusterStateCache = _currentSubscriptionCallback->GetClusterStateCache();
+            if (clusterStateCache) {
+                clusterStateCache->ClearAttributes(static_cast<EndpointId>(endpoint.unsignedLongLongValue));
```
jtung-apple, line 3436: Is it safe to mutate clusterStateCache while not being on the Matter queue?

bzbarsky-apple: No, good catch!  For some reason I thought this was on the Matter queue.

src/darwin/Framework/CHIP/MTRBaseSubscriptionCallback.h
bzbarsky-apple: Added Clear methods instead.  @andy31415 Please take a look when you get the chance?

src/darwin/Framework/CHIP/MTRDevice.mm
bzbarsky-apple: Fixed.

jtung-apple (approved):

yufengwangca (approved):

PR #34231, opened 2024-07-08
python_testing: generate controller's nodeid and shutdown subscription

doublemis1:
* implementation of generate_random_node_id with exculdes values, the method add possibilty to run multiple test cases within one test session (the controllers' nodeid is the same for new controller in different test cases)
* ACE-1.2 script update: shutdown scubscriptions

### DIFF
--- a/src/python_testing/matter_testing_support.py
+++ b/src/python_testing/matter_testing_support.py
@@ -1506,6 +1506,7 @@ def generate_random_nodeid(excluded_nodeid: typing.Optional[typing.Set] = set())
         return generate_random_nodeid(excluded_nodeid)
     return nodeid
 
+
 def async_test_body(body):
     """Decorator required to be applied whenever a `test_*` method is `async def`.
 
@@ -1710,7 +1711,7 @@ def run_tests_no_exit(test_class: MatterBaseTest, matter_test_config: MatterTest
 
             try:
                 runner.run()
-                ok = runner.results.is_all_pass and ok      
+                ok = runner.results.is_all_pass and ok
             except TimeoutError:
                 ok = False
             except signals.TestAbortAll:
--- a/src/python_testing/matter_testing_support.py
+++ b/src/python_testing/matter_testing_support.py
@@ -1499,6 +1499,13 @@ def parse_matter_test_args(argv: Optional[List[str]] = None) -> MatterTestConfig
     return convert_args_to_matter_config(parser.parse_known_args(argv)[0])
 
 
+def generate_random_nodeid(excluded_nodeid: typing.Optional[typing.Set] = set()) -> int:
+    "Generate random complainat nodeid, excluding a set of provided nodeids."
+    nodeid = random.randint(1, 0xFFFF_FFEF_FFFF_FFFF)
+    if nodeid in excluded_nodeid:
+        return generate_random_nodeid(excluded_nodeid)
+    return nodeid
+
 def async_test_body(body):
     """Decorator required to be applied whenever a `test_*` method is `async def`.
 
@@ -1703,7 +1710,7 @@ def run_tests_no_exit(test_class: MatterBaseTest, matter_test_config: MatterTest
 
             try:
                 runner.run()
-                ok = runner.results.is_all_pass and ok
+                ok = runner.results.is_all_pass and ok      
             except TimeoutError:
                 ok = False
             except signals.TestAbortAll:
--- a/src/python_testing/TC_ACE_1_2.py
+++ b/src/python_testing/TC_ACE_1_2.py
@@ -30,7 +30,7 @@ from chip.clusters import ClusterObjects as ClusterObjects
 from chip.clusters.Attribute import EventReadResult, SubscriptionTransaction, TypedAttributePath
 from chip.exceptions import ChipStackError
 from chip.interaction_model import Status
-from matter_testing_support import MatterBaseTest, async_test_body, default_matter_test_main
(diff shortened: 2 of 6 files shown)


src/python_testing/matter_testing_support.py
```diff
@@ -1492,19 +1492,35 @@ def parse_matter_test_args(argv: Optional[List[str]] = None) -> MatterTestConfig
 
     return convert_args_to_matter_config(parser.parse_known_args(argv)[0])
 
+def async_to_sync(awaitable):
+    "Call sync, the async function."
+    # Usage of new implementation of ChipDeviceController cause the necessary to not use asyncio.run,
+    # because the initialization of `asyncio.Lock` use the `asyncio.get_event_loop()` on init.
+    # Usage of `asyncio.run()` and `asyncio.get_event_loop()` are exclusive.
```
tcarmelveilleux, line 1499: It's not 100% clear why this change is needed.

src/python_testing/matter_testing_support.py
```diff
@@ -1492,19 +1492,35 @@ def parse_matter_test_args(argv: Optional[List[str]] = None) -> MatterTestConfig
...
+    loop = asyncio.get_event_loop()
+    return loop.run_until_complete(awaitable)
+
+def generate_random_nodeid(excluded_nodeid: typing.Optional[typing.List] = None) -> int:
+    if excluded_nodeid:
+        while True:
+            node_id=random.randint(0, 2^64-1)
```
tcarmelveilleux, line 1506: Node ID 0 is not allowed (it means "unspecified").

There are some forbidden node IDs in the 0xFFFF_FFFx_xxxx_xxxx.

Also `2^64` is incorrect Python syntax and will lead to a very small number (66). Correct syntax is `2**64`.

```suggestion
            node_id=random.randint(1, 0xFFFF_FFEF_FFFF_FFFF)
```

The above suggestion is spec compliant.

src/python_testing/matter_testing_support.py
```diff
@@ -1492,19 +1492,35 @@ def parse_matter_test_args(argv: Optional[List[str]] = None) -> MatterTestConfig
...
+            if node_id not in excluded_nodeid:
+                return node_id
+    else:
+        return random.randint(0, 2^64-1)
```
tcarmelveilleux, line 1510: Refactor loop not to need a repeated random.randint by just generating until "not invalid".

src/python_testing/matter_testing_support.py
```diff
@@ -1492,19 +1492,35 @@ def parse_matter_test_args(argv: Optional[List[str]] = None) -> MatterTestConfig
 
     return convert_args_to_matter_config(parser.parse_known_args(argv)[0])
 
+def async_to_sync(awaitable):
+    "Call sync, the async function."
+    # Usage of new implementation of ChipDeviceController cause the necessary to not use asyncio.run,
+    # because the initialization of `asyncio.Lock` use the `asyncio.get_event_loop()` on init.
+    # Usage of `asyncio.run()` and `asyncio.get_event_loop()` are exclusive.
+    loop = asyncio.get_event_loop()
```
andy31415, line 1500: https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.get_event_loop says "deprecated since version 3.12" and documentation recommends to use alternatives.

I also am not very clear on what is going on based on the above comment and worry about the deprecation and warnings in the python documentation. Would it be correct to use get_running_loop for example?

Or is this code really needing to handle both coroutine vs non-coroutine use cases?

src/python_testing/matter_testing_support.py
```diff
@@ -1492,19 +1492,35 @@ def parse_matter_test_args(argv: Optional[List[str]] = None) -> MatterTestConfig
...
+    return loop.run_until_complete(awaitable)
+
+def generate_random_nodeid(excluded_nodeid: typing.Optional[typing.List] = None) -> int:
```
andy31415, line 1503: List is not ideal, maybe Set would be better.

Why not set the default to an empty set, and then you do not need to handle None and everything can go through the same `while True` branch.

src/python_testing/matter_testing_support.py
doublemis1: The usage of `asyncio.run()` and  `asyncio.get_event_loop()` are mutually exclusive. 
In python 3.9.18 creation of asyncio.Lock call  `asyncio.get_event_loop()`:
https://github.com/python/cpython/blob/3.9/Lib/asyncio/locks.py#L81

Example of reproduction:
```
import asyncio

async def hello():
    print("hello")

lock_1 = asyncio.Lock()
asyncio.run(hello())
lock_2 = asyncio.Lock()
```

doublemis1: I think that I will fix that issue on my side, while using the scripts.
Since it should be an issue in the Python 3.9 and older version.

src/python_testing/matter_testing_support.py
doublemis1: Thanks for comment.
your suggestion has been applied

src/python_testing/matter_testing_support.py
doublemis1: Done, used as recursive function

src/python_testing/matter_testing_support.py
doublemis1: Removed change from the PR, the issue is visible on specific version of Python (Python 3.9.x and older), mentioned here https://github.com/project-chip/connectedhomeip/pull/34231#discussion_r1668837629.

PR #34232, opened 2024-07-08
Do not force STDIO output for Linux example apps

arkq:
### Problem

Example apps should be designed as real applications. Hence, they should integrate with platform logger instead of logging to stdio.

### Changes

- remove forced logging type from examples/platform, so logging can be selected per-app if required
- update cirque test script, so it will work with stdio and platform logging

### Testing

CI will verify potential build breaks.

### DIFF
--- a/config/nuttx/chip-gn/BUILD.gn
+++ b/config/nuttx/chip-gn/BUILD.gn
@@ -44,6 +44,7 @@ static_library("nuttx") {
     public_deps += [
       "${chip_root}/examples/lighting-app/lighting-common",
       "${chip_root}/examples/lighting-app/lighting-common:lighting-manager",
+      "${chip_root}/src/platform/logging:default",
     ]
   }
 
--- a/examples/chip-tool/BUILD.gn
+++ b/examples/chip-tool/BUILD.gn
@@ -109,7 +109,7 @@ static_library("chip-tool-utils") {
     "${chip_root}/src/app/tests/suites/commands/interaction_model",
     "${chip_root}/src/controller/data_model",
     "${chip_root}/src/credentials:file_attestation_trust_store",
-    "${chip_root}/src/lib",
+    "${chip_root}/src/lib:without-logging",
     "${chip_root}/src/lib/core:types",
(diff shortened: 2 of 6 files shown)


andy31415 (approved):

yufengwangca (approved):

kiel-apple (approved):

bzbarsky-apple (changes requested):
Test failures are:
```
    raise Exception('Timeout while waiting for %s' % waitForString)
```

because the tests are waiting for some stdio output from the example apps that never comes.

I expect this is actually quite common in workflows involving the example apps (e.g. it's very common for people to find the relevant pairing code in the stdio output).  And certainly we have a bunch of tests that log things from those apps to files in CI and expect those logs to be available later if the test fails.

So I think we need to:

1. Decide whether we actually want to change the default behavior, given the significant pain this will impose on people who currently use the example apps.
2. Decide whether this should be an across-the-board change or whether some commonly-used example apps where this could be a particular problem should default to stdio (for now, or in general).
3. Fix all our build documentation to describe the options for logging and how to select the one you want (no matter what the default is).
4. If we change the default behavior for some apps, ensure that all existing testing that uses those apps is audited to see whether it's OK with the behavior change, and if not update the relevant build args.

But long story short: as-is this will break a _ton_ of people's workflows.  A bunch of in-tree ones (some of them leading to CI failures), but also lots of out-of-tree ones.

arkq:
> because the tests are waiting for some stdio output from the example apps that never comes.

Yes, I'm aware of that. Now, I'm trying to figure how much work would it be to add support for changing logging to stdio for CI (and potentially other out of tree cases) while still allowing to build example apps with proper platform logging. The case is, that I would like to use the example app on a real device without mich hassle. For Linux apps it's not an issue since the platform logging happens to be stdout as well. But, Darwin and Tizen (Tizen is my real case) by default use dedicated logger. However, the example apps are both based on Linux example platform code, so that's why I'd like the platform common code not to force stdio logging.

I'll convert this PR to draft for now.

PR #34233, opened 2024-07-08
Docs: Move platform docs into their own directory and separate into subfolders

cecille:
The handbook is planning to link over to the SDK docs repo for SDK content, but requested a bit of cleanup prior to that. Per discussion in https://docs.google.com/document/d/1stNF20dL0aQzR1E5PR_ghicjEEp7eP-NHioPwYax4kQ/edit, guides directory was one of the places that needed some work.

This PR moves all the platform-specific guides into a separate directory, with sub-directories for each platform. Also adjusts the links.

Note - this is not the complete set of cleanups, but probably the largest.


docs/index.md
```diff
@@ -10,10 +10,12 @@ PROJECT_FLOW
 VSCODE_DEVELOPMENT
 api/index
 ci-cd/index
+development_controllers/index
 discussion/index
 getting_started/index
```
Vyrastas, line 15: I feel like this should be at the top?  Before api/index

docs/index.md
```diff
@@ -10,10 +10,12 @@ PROJECT_FLOW
...
 cluster_and_device_type_dev/index
```
Vyrastas, line 16: Does it make more sense to have the (zap|spec)_clusters files in here?  Or maybe more towards the top rather than at the bottom?

I feel like that's a page a lot of people will want to visit.

docs/guides/index.md
```diff
@@ -9,58 +9,17 @@ and features.
...
 
 *
-esp32/README
-nxp/README
-ti/ti_matter_overview
 ```
 
 ## Build Guides
 
 -   [Building](./BUILDING.md)
```
Vyrastas, line 16: Should we move this to Getting Started?

In general, this "Guides" grouping doesn't seem that useful now with most of the other content moved out of it.  I would maybe just remove it and put the content in the other groupings as needed.

Vyrastas (commented):
Based on our chat, we could remove most of the content from /docs/style/style_guide.md, some of it is inaccurate (the folder tree especially) and we can add a link to the Google Developers style guide.

But that could be done in a follow up PR too

docs/index.md
cecille: Possibly, but they need a re-vamp anyway. That's a bit orthogonal to this change through. Can do in a follow up.

docs/guides/index.md
cecille: yeah, the eventual goal is to get rid of the "Guides" directory, but this PR is JUST moving the platform stuff out. I want to be able to have the platform folks take a look and make sure they're OK with this without adding in a bunch of other changes that don't affect them.

Vyrastas (approved):

docs/index.md
cecille: discussed offline - follow up ok.

docs/guides/index.md
cecille: discussed offline - handling in follow up

docs/index.md
cecille: discussed offline - handling in follow up.

woody-apple (approved):

docs/platforms/darwin/index.md
```diff
@@ -0,0 +1,11 @@
+# Darwin
```
woody-apple, line 1: Thanks again for doing this, let me add a symlink in here, so the existing paths that public websites are using, can redirect to the new location

cecille: Moving darwin back directly into the guides directory to avoid blocking this PR. Note that the handbook won't be linking the guides directory since it's a bit of a dumping ground, so if you want the darwin info to appear in the handbook you'll need to move this into a more sensible location with whatever simlinks you need.

arkq (approved):

PR #34262, opened 2024-07-09
Fix handle cleanup in CommandHandler.

bzbarsky-apple:
When a CommandHandler went away, it invalidated its handles, but did not remove them from the list.  This would cause it to fail a fatal assert when the non-empty list was destroyed.

The fix is to clear the list once we have invalidated all the handles, since we no longer need to track them once they have been invalidated.


kiel-apple (approved):

woody-apple (approved):

woody-apple:
Note: This is a crash reported from the wild

woody-apple:
Fast tracking quality fix

PR #34263, opened 2024-07-09
Reduce codesize for door locks a bit more.

bzbarsky-apple:
There is no need for SupportsAliroProvisioning checks in Aliro command handlers. If we got into that code, that means that we recognized the command ID on the relevant endpoint's Door Lock cluster, and if we passed cert that means that we had the right feature map bit set too.

### DIFF
--- a/src/app/clusters/door-lock-server/door-lock-server.cpp
+++ b/src/app/clusters/door-lock-server/door-lock-server.cpp
@@ -3990,14 +3990,6 @@ void DoorLockServer::clearAliroReaderConfigCommandHandler(CommandHandler * comma
     EndpointId endpointID = commandPath.mEndpointId;
     ChipLogProgress(Zcl, "[ClearAliroReaderConfig] Incoming command [endpointId=%d]", endpointID);
 
-    // If Aliro Provisioning feature is not supported, return UNSUPPORTED_COMMAND.
-    if (!SupportsAliroProvisioning(endpointID))
-    {
-        ChipLogProgress(Zcl, "[ClearAliroReaderConfig] Aliro Provisioning is not supported [endpointId=%d]", endpointID);
-        commandObj->AddStatus(commandPath, Status::UnsupportedCommand);
-        return;
-    }
-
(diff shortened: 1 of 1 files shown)


nivi-apple (approved):

src/app/clusters/door-lock-server/door-lock-server.cpp
```diff
@@ -3990,14 +3990,6 @@ void DoorLockServer::clearAliroReaderConfigCommandHandler(CommandHandler * comma
     EndpointId endpointID = commandPath.mEndpointId;
     ChipLogProgress(Zcl, "[ClearAliroReaderConfig] Incoming command [endpointId=%d]", endpointID);
 
-    // If Aliro Provisioning feature is not supported, return UNSUPPORTED_COMMAND.
-    if (!SupportsAliroProvisioning(endpointID))
```
nivi-apple, line 3994: We need to remove [https://github.com/project-chip/connectedhomeip/blob/9d3d8d21c249410070d3275fe5900[…]984e19db9d/src/app/clusters/door-lock-server/door-lock-server.h](https://github.com/project-chip/connectedhomeip/blob/9d3d8d21c249410070d3275fe59003984e19db9d/src/app/clusters/door-lock-server/door-lock-server.h#L231) and [https://github.com/project-chip/connectedhomeip/blob/9d3d8d21c249410070d3275fe5900[…]984e19db9d/src/app/clusters/door-lock-server/door-lock-server.h](https://github.com/project-chip/connectedhomeip/blob/9d3d8d21c249410070d3275fe59003984e19db9d/src/app/clusters/door-lock-server/door-lock-server.h#L241)

bzbarsky-apple: No, those are still used.

woody-apple:
Fast tracking test

woody-apple:
My mistake, I thought this was in the example app, this should not have been fast tracked.

woody-apple:
Tagging @andy31415 Who reviewed the prior part of this yesterday.

PR #34265, opened 2024-07-09
Update .gitignore - adds py environment folder

raul-marquez-csa:
### Update .gitignore - adds py environment folder

All the files and folders in the py environment folder are appearing as new changes to commit. This repository doesn't include that folder, so I added the py folder to ignore it.

This command creates the py environment folder:
`./scripts/build_python.sh -i py`

Explanation is here:
https://github.com/project-chip/connectedhomeip/blob/bf4edc4f0eaf949de3e39082ec4b8e64122cfd67/docs/testing/python.md?plain=1#L517-L523

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -85,3 +85,5 @@ examples/*/esp32/dependencies.lock
 # jupyter temporary files
 .ipynb_checkpoints
 
+# py environment folder
+py


.gitignore
```diff
@@ -85,3 +85,5 @@ examples/*/esp32/dependencies.lock
 # jupyter temporary files
 .ipynb_checkpoints
 
+# py environment folder
+py
```
tcarmelveilleux, line 89:
```suggestion
py
__pycache__
```

tcarmelveilleux (approved):

.gitignore
raul-marquez-csa: That's interesting, __pycache__ was already in the file, add it again?

woody-apple (approved):

andy31415 (changes requested):
What is a `py` environment folder? What build process creates that?

I do not have such a thing.

andy31415:
> What is a `py` environment folder? What build process creates that?
> 
> I do not have such a thing.

Specifcally:
  - name does not feel special (no `__` prefix nor `.` even for hidden)
  - name is actually common, like `scripts/tests/py/` and `src/test_driver/efr32/py/`
  
We have to figure out what build command creates it and dirties the source tree. Please add this information to the PR summary so we can determine if the command is buggy or if we really need to ignore (and add details to gitignore).

Just ignoring things that dirty our source tree is not ok (we should not accept dirtying of the source tree)

raul-marquez-csa:
> > What is a `py` environment folder? What build process creates that?
> > I do not have such a thing.
> 
> Specifcally:
> 
> * name does not feel special (no `__` prefix nor `.` even for hidden)
> * name is actually common, like `scripts/tests/py/` and `src/test_driver/efr32/py/`
> 
> We have to figure out what build command creates it and dirties the source tree. Please add this information to the PR summary so we can determine if the command is buggy or if we really need to ignore (and add details to gitignore).
> 
> Just ignoring things that dirty our source tree is not ok (we should not accept dirtying of the source tree)

This command creates it
`./scripts/build_python.sh -i py`

Explanation is here:
https://github.com/project-chip/connectedhomeip/blob/bf4edc4f0eaf949de3e39082ec4b8e64122cfd67/docs/testing/python.md?plain=1#L517-L523

raul-marquez-csa:
I can update it to this
`/py`
This will only target the py folder in the root folder, not subfolders

bzbarsky-apple:
Shouldn't builds go into `out` or something like that which already gets gitignored?

andy31415:
> > > What is a `py` environment folder? What build process creates that?
> > > I do not have such a thing.
> > 
> > 
> > Specifcally:
> > 
> > * name does not feel special (no `__` prefix nor `.` even for hidden)
> > * name is actually common, like `scripts/tests/py/` and `src/test_driver/efr32/py/`
> > 
> > We have to figure out what build command creates it and dirties the source tree. Please add this information to the PR summary so we can determine if the command is buggy or if we really need to ignore (and add details to gitignore).
> > Just ignoring things that dirty our source tree is not ok (we should not accept dirtying of the source tree)
> 
> This command creates it `./scripts/build_python.sh -i py`
> 
> Explanation is here:
> 
> https://github.com/project-chip/connectedhomeip/blob/bf4edc4f0eaf949de3e39082ec4b8e64122cfd67/docs/testing/python.md?plain=1#L517-L523

We should fix the documentation to place it in `out/py`.

andy31415:
Closing old PR: we have in the mean time updated our documentation in a different PR (#35901)

PR #34266, opened 2024-07-09
Introduce a building block usable for all Q attributes

tcarmelveilleux:
- Q quality requires marking attributes as dirty for the purposes of reporting only when certain conditions arise.
- This PR introduces a building block attribute value wrapper compatible with any nullable or non-nullable numerical attribute, which allows applying the necessary policies, and all complex policies that currently exist in the Matter spec (e.g. any CountdownTime, CurrentLevel, etc).

Testing done:

- Added unit tests. Integration in existing clusters will follow in a different PR.

### DIFF
--- a/src/app/cluster-building-blocks/QuieterReporting.h
+++ b/src/app/cluster-building-blocks/QuieterReporting.h
@@ -35,36 +35,37 @@ enum class QuieterReportingPolicyEnum
     kMarkDirtyOnIncrement        = (1u << 2),
 };
 
-using QuieterReportingPolicyFlags = ::chip::BitFlags<QuieterReportingPolicyEnum>;
+using QuieterReportingPolicyFlags = BitFlags<QuieterReportingPolicyEnum>;
 
 namespace detail {
 
-using Timestamp = chip::System::Clock::Milliseconds64;
+using Timestamp = System::Clock::Milliseconds64;
 template <typename T>
-using Nullable = chip::app::DataModel::Nullable<T>;
+using Nullable = DataModel::Nullable<T>;
 
 /**
  * This class helps track reporting state of an attribute to properly keep track of whether
- * it needs to be marked as dirty or not for purposes of reporting using Q quality.
+ * it needs to be marked as dirty or not for purposes of reporting using
+ * "7.7.9 Quieter Reporting Quality" (Q quality)
  *
  * The class can be configured via `SetPolicy` to have some/all of the common reasons
  * for reporting (e.g. increment only, decrement only, change to/from zero).
  *
  * Changes of null to non-null or non-null to null are always considered dirty.
  *
- * It is possible to force mark the attribute as dirty (see `ForceMarkAsDirty`) such as
+ * It is possible to force mark the attribute as dirty (see `ForceDirty`) such as
  * for conditions like "When there is any increase or decrease in the estimated time
  * remaining that was due to progressing insight of the server's control logic".
  *
- * Usage is simple:
+ * Usage:
  *
  * - Call `SetValue()` with the new value and current monotonic system timestamp
  *   - There is an overload with a `SufficientChangePredicate` which will apply externally
  *     provided checks between old/new value and time last marked dirty to allow for complex
  *     rules like marking dirty less than once per second, etc.
- * - *Maybe* call `ForceMarkAsDirty()` in some choice situations (e.g. know for sure it's
+ * - *Maybe* call `ForceDirty()` in some choice situations (e.g. know for sure it's
  *    an important update, like at the edge of an operational state change).
- * - Call `GetThenResetDirtyState()`. If it returns true, mark the attribute dirty, with the
+ * - Call `WasJustMarkedDirty()`. If it returns true, mark the attribute dirty, with the
  *   method most suitable at the call site (e.g. `MatterReportingAttributeChangeCallback` call
  *   or similar methods).
  *
@@ -84,7 +85,7 @@ template <typename T>
 class QuieterReportingAttribute
 {
 public:
-    explicit QuieterReportingAttribute(const chip::app::DataModel::Nullable<T> & initialValue) :
+    explicit QuieterReportingAttribute(const Nullable<T> & initialValue) :
         mValue(initialValue), mLastDirtyValue(initialValue)
     {}
 
@@ -99,7 +100,7 @@ public:
      * @return a functor usable for the `changedPredicate` arg of `SetValue()`
      */
     static SufficientChangePredicate
-    GetPredicateForSufficientTimeSinceLastDirty(chip::System::Clock::Milliseconds64 minimumDurationMillis)
+    GetPredicateForSufficientTimeSinceLastDirty(System::Clock::Milliseconds64 minimumDurationMillis)
     {
         return [minimumDurationMillis](Timestamp previousDirtyTime, Timestamp now, const Nullable<T> & oldDirtyValue,
                                        const Nullable<T> & newValue) -> bool {
@@ -107,15 +108,16 @@ public:
         };
     }
 
-    chip::app::DataModel::Nullable<T> value() const { return mValue; }
-    QuieterReportingPolicyFlags policy() const { return mPolicyFlags; }
+    Nullable<T> value() const { return mValue; }
+    QuieterReportingPolicyFlags & policy() { return mPolicyFlags; }
+    const QuieterReportingPolicyFlags & policy() const { return mPolicyFlags; }
 
     void SetPolicy(QuieterReportingPolicyFlags policyFlags) { mPolicyFlags = policyFlags; }
 
     /**
      * When this returns true, attribute should be marked for reporting. Auto-resets to false after call.
      */
-    bool GetThenResetDirtyState()
+    bool WasJustMarkedDirty()
     {
         bool wasDirty = mIsDirty;
         mIsDirty      = false;
@@ -125,9 +127,9 @@ public:
     /**
      * Force marking this attribute as dirty, with the `now` timestamp given as the reference point.
      *
-     * WARNING: Only call `ForceMarkAsDirty` after a `SetValue` call.
+     * WARNING: Only call `ForceDirty` after a `SetValue` call.
      */
-    void ForceMarkAsDirty(Timestamp now)
+    void ForceDirty(Timestamp now)
     {
         mIsDirty                  = true;
         mLastDirtyTimestampMillis = now;
--- a/src/app/cluster-building-blocks/QuieterReporting.h
+++ b/src/app/cluster-building-blocks/QuieterReporting.h
@@ -19,6 +19,7 @@
 #pragma once
 
 #include <functional>
+#include <type_traits>
 #include <stdbool.h>
 
 #include <app/data-model/Nullable.h>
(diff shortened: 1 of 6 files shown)


src/app/cluster-building-blocks/QuieterReporting.h
```diff
@@ -0,0 +1,194 @@
...
+ *
+ * @tparam T - the type of underlying numerical value that will be held by the class.
+ */
+template <typename T>
+class QuieterReportingAttribute
+{
+  public:
+    explicit QuieterReportingAttribute(const chip::app::DataModel::Nullable<T>& initialValue) : mValue(initialValue), mLastDirtyValue(initialValue) {}
+
+    using SufficientChangePredicate = std::function<bool(Timestamp /* previousDirtyTime */, Timestamp /* now */, const Nullable<T> & /* previousDirtyValue */, const Nullable<T> & /* newValue */)>;
```
andy31415: How about passing a struct here instead of 4 arguments? I am thinking that would get us:
   - smaller signature (4 arguments start being large)
   - nicer backwards compatibility in the unlikely case we want to pass in more things here
   - enforced naming (otherwise both callers and implementors may have a much easier time to mis-order same-type of values ... we have 2 identical timestamps and nullable<t> so switching order is very easy)

We need some documentation: what does previousDirtyTime mean? It seems that if the previous value was not dirty we default-init the milliseconds there as {}. Is that 0 or current time? Would a optional not be more correct here?

src/app/cluster-building-blocks/QuieterReporting.h
```diff
@@ -0,0 +1,194 @@
...
+ *
+ * - Call `SetValue()` with the new value and current monotonic system timestamp
+ *   - There is an overload with a `SufficientChangePredicate` which will apply externally
+ *     provided checks between old/new value and time last marked dirty to allow for complex
+ *     rules like marking dirty less than once per second, etc.
+ * - *Maybe* call `ForceMarkAsDirty()` in some choice situations (e.g. know for sure it's
+ *    an important update, like at the edge of an operational state change).
+ * - Call `GetThenResetDirtyState()`. If it returns true, mark the attribute dirty, with the
+ *   method most suitable at the call site (e.g. `MatterReportingAttributeChangeCallback` call
+ *   or similar methods).
```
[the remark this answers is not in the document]
andy31415: Are we always going to only support Nullable values and never non-nullable quieter reporting?

src/app/cluster-building-blocks/QuieterReporting.h
```diff
@@ -0,0 +1,194 @@
...
+    {
+      return [minimumDurationMillis](Timestamp previousDirtyTime, Timestamp now, const Nullable<T> &oldDirtyValue, const Nullable<T> &newValue) -> bool {
+        return (oldDirtyValue != newValue) && ((now - previousDirtyTime) >= minimumDurationMillis);
+      };
+    }
+
+    chip::app::DataModel::Nullable<T> value() const { return mValue; }
+    QuieterReportingPolicyFlags policy() const { return mPolicyFlags; }
+
+    void SetPolicy(QuieterReportingPolicyFlags policyFlags) { mPolicyFlags = policyFlags; }
```
andy31415, line 113: Did we consider having:

```cpp
const QuieterReportingPolicyFlags &policy() const { return mPolicyFlags; }
QuieterReportingPolicyFlags &policy() { return mPolicyFlags; }
```

instead and then we can do things like:

```cpp
foo.policy().Set(QuieterReportingPolicyEnum::kMarkDirtyOnDecrement);

// or use clear or even operator=
```

instead of needing

```cpp
foo.SetPolicy()QuieterReportingPolicyFlags(QuieterReportingPolicyEnum::kMarkDirtyOnDecrement));
```

src/app/cluster-building-blocks/QuieterReporting.h
```diff
@@ -0,0 +1,194 @@
...
+
+    /**
+     * When this returns true, attribute should be marked for reporting. Auto-resets to false after call.
+     */
+    bool GetThenResetDirtyState()
```
andy31415, line 118: Methods with compund statements (and/then/...) feel off. Why don't we have `IsDirty` and `ClearDirty`? I can replace `foo = bar.GetThenResetDirtyState` with `foo = bar.IsDirty(); bar.ClearDirty()`... slightly more verbose however significant less cognitive load reading.

src/app/cluster-building-blocks/QuieterReporting.h
```diff
@@ -0,0 +1,194 @@
...
+        mIsDirty = false;
+        return wasDirty;
+    }
+
+    /**
+     * Force marking this attribute as dirty, with the `now` timestamp given as the reference point.
+     *
+     * WARNING: Only call `ForceMarkAsDirty` after a `SetValue` call.
+     */
+    void ForceMarkAsDirty(Timestamp now)
```
andy31415, line 130:
```suggestion
    void ForceDirty(Timestamp now)
```

src/app/cluster-building-blocks/QuieterReporting.h
```diff
@@ -0,0 +1,194 @@
...
+     */
+    void SetValue(const chip::app::DataModel::Nullable<T>& newValue, Timestamp now)
+    {
+        SetValue(newValue, now, [](Timestamp, Timestamp, const Nullable<T> &, const Nullable<T> &) -> bool { return false; });
+    }
+
+  protected:
+    chip::app::DataModel::Nullable<T> mValue;
+    chip::app::DataModel::Nullable<T> mLastDirtyValue;
+    bool mIsDirty = false;
```
andy31415, line 194: A blurb on what each of these contain and relation between values (e.g. mLastDirtyTimestampMillis being updated whenever mLastDirtyValue is updated) would help understanding how members are used, and mLastDirtyValue lags behind mValue and is updated whenever mDirty transitions.

andy31415: Use of mLastDirtyValue seems to be `mFirstDirtyValue` as it is used in SetValue: once something is dirty, we set the mIsDirty flag and never update the dirty value ever again. Maybe just naming `dirtyValue` would be better (as the forcing dirty does indeed force the value to be the last).

src/app/cluster-building-blocks/QuieterReporting.h
andy31415: It also allows IsDirty in unit tests (or other code) without resetting, such as asserts or other logic.

tcarmelveilleux: Resetting is a key part of the API and critical to avoid mistakes. Staying sorry because of a missed ClearDirty can lead to performance issues and regression of the while point of the thing.

src/app/cluster-building-blocks/QuieterReporting.h
tcarmelveilleux: Will rework.

src/app/cluster-building-blocks/QuieterReporting.h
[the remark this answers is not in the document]
tcarmelveilleux: > Are we always going to only support Nullable values and never non-nullable quieter reporting?

The nullable works with both.

src/app/cluster-building-blocks/QuieterReporting.h
tcarmelveilleux: Will rework

src/app/cluster-building-blocks/QuieterReporting.h
tcarmelveilleux: Done

src/app/cluster-building-blocks/QuieterReporting.h
tcarmelveilleux: Renamed to `WasJustMarkedDirty` which is more accurate.

andy31415 (changes requested):
Changes requested on sideffects api: a getter that always resets feels off.

src/app/cluster-building-blocks/QuieterReporting.h
andy31415: `WasJust` seems to imply some temporality that the class does not have. We just have dirty and a timestamp.

This is a naming + concept issue. Leaning towards looking at atomics compareexchange/fetchset or a `Clear` returning the previous value.

src/app/cluster-building-blocks/QuieterReporting.h
```diff
@@ -0,0 +1,249 @@
...
+     *
+     * Internal recording will be done about last dirty value and last dirty timestamp based on the policies having applied.
+     *
+     * @param newValue - new value to set for the attribute
+     * @param now - system monotonic timestamp at the time of the call
+     * @param changedPredicate - functor to possibly override dirty state
+     */
+    void SetValue(const chip::app::DataModel::Nullable<T> & newValue, Timestamp now, SufficientChangePredicate changedPredicate)
+    {
+        bool isChangeOfNull       = newValue.IsNull() ^ mValue.IsNull();
```
andy31415, line 187: any chance we could have a:

```cpp
if (mIsDirty) 
{
    mValue = newValue;
    return;
}
```

here to bypass all the computations below if already dirty? Then we can rely on mIsDirty being false below.

Or are we required to call changedPredicate on every newValue even if dirty?

andy31415 (approved):

jmartinez-silabs (approved):

woody-apple (approved):

src/app/cluster-building-blocks/QuieterReporting.h
```diff
@@ -0,0 +1,241 @@
...
+     *
+     * Internal recording will be done about last dirty value and last dirty timestamp based on the policies having applied.
+     *
+     * @param newValue - new value to set for the attribute
+     * @param now - system monotonic timestamp at the time of the call
+     * @param changedPredicate - functor to possibly override dirty state
+     * @return AttributeDirtyState::kMustReport if attribute must be marked dirty right away, or
+     * AttributeDirtyState::kNoReportNeeded otherwise.
+     */
+    AttributeDirtyState SetValue(const chip::app::DataModel::Nullable<T> & newValue, Timestamp now,
```
bzbarsky-apple, line 175: This is inside namespace chip and namespace app; why the extra prefixes?

src/app/cluster-building-blocks/QuieterReporting.h
```diff
@@ -0,0 +1,241 @@
...
+    using SufficientChangePredicate = std::function<bool(const SufficientChangePredicateCandidate &)>;
+
+    /**
+     * @brief Factory to generate a functor for "attribute was last reported" at least `minimumDurationMillis` ago.
+     *
+     * @param minimumDurationMillis - number of millis needed since last marked as dirty before we mark dirty again.
+     * @return a functor usable for the `changedPredicate` arg of `SetValue()`
+     */
+    static SufficientChangePredicate
+    GetPredicateForSufficientTimeSinceLastDirty(System::Clock::Milliseconds64 minimumDurationMillis)
```
bzbarsky-apple, line 137: Do we really need "millis" in both the type and the arg name?

src/app/cluster-building-blocks/QuieterReporting.h
```diff
@@ -0,0 +1,241 @@
...
+     * @param newValue - new value to set for the attribute
+     * @param now - system monotonic timestamp at the time of the call
+     * @param changedPredicate - functor to possibly override dirty state
+     * @return AttributeDirtyState::kMustReport if attribute must be marked dirty right away, or
+     * AttributeDirtyState::kNoReportNeeded otherwise.
+     */
+    AttributeDirtyState SetValue(const chip::app::DataModel::Nullable<T> & newValue, Timestamp now,
+                                 SufficientChangePredicate changedPredicate)
+    {
+        bool isChangeOfNull       = newValue.IsNull() ^ mValue.IsNull();
```
bzbarsky-apple, line 178: This would be more readable as `newValue.IsNull() != mValue.IsNull()`, I think...

src/app/cluster-building-blocks/QuieterReporting.h
```diff
@@ -0,0 +1,241 @@
...
+        bool areBothValuesNonNull = !newValue.IsNull() && !mValue.IsNull();
+
+        bool changeToFromZero = areBothValuesNonNull && (*newValue == 0 || *mValue == 0);
```
bzbarsky-apple, line 181: If both values are 0, this looks like it will incorrectly trigger reporting, no?

Maybe up front in this function we should just `AttributeDirtyState::kNoReportNeeded` if the values are equal?

src/app/cluster-building-blocks/QuieterReporting.h
```diff
@@ -0,0 +1,241 @@
...
+     * AttributeDirtyState::kNoReportNeeded otherwise.
+     */
+    AttributeDirtyState SetValue(const chip::app::DataModel::Nullable<T> & newValue, Timestamp now)
+    {
+        return SetValue(newValue, now, [](const SufficientChangePredicateCandidate &) -> bool { return false; });
+    }
+
+protected:
+    // Current value of the attribute.
+    chip::app::DataModel::Nullable<T> mValue;
```
bzbarsky-apple, line 227: Again, extra prefixes.

src/app/cluster-building-blocks/QuieterReporting.h
```diff
@@ -0,0 +1,241 @@
...
+     * @brief Factory to generate a functor that forces reportable now.
+     * @return a functor usable for the `changedPredicate` arg of `SetValue()`
+     */
+    static SufficientChangePredicate GetForceReportablePredicate()
+    {
+        return [](const SufficientChangePredicateCandidate & candidate) -> bool { return true; };
+    }
+
+    Nullable<T> value() const { return mValue; }
+    QuieterReportingPolicyFlags & policy() { return mPolicyFlags; }
```
bzbarsky-apple, line 155: Do we really want policy to be randomly mutable?  I would expect it to be a constructor argument and immutable...

src/app/cluster-building-blocks/QuieterReporting.h
```diff
@@ -0,0 +1,241 @@
...
+
+protected:
+    // Current value of the attribute.
+    chip::app::DataModel::Nullable<T> mValue;
+    // Last value that was marked as dirty (to use in comparisons for change, e.g. by SufficientChangePredicate).
+    chip::app::DataModel::Nullable<T> mLastDirtyValue;
+    // Enabled internal change detection policies.
+    QuieterReportingPolicyFlags mPolicyFlags{ 0 };
+    // Timestamp associated with the last time the attribute was marked dirty (to use in comparisons for change).
+    chip::System::Clock::Milliseconds64 mLastDirtyTimestampMillis{};
```
bzbarsky-apple, line 233: We seem to be inconsistently using Milliseconds64 some places for timestamps, and Timestamp in others.  Why not just Timestamp everywhere?

src/app/cluster-building-blocks/QuieterReporting.h
tcarmelveilleux: I much prefer units in argument names no matter what, when a unit is involved. The fact it's "trivial" here should not override the idea.

src/app/cluster-building-blocks/QuieterReporting.h
tcarmelveilleux: Fixed

src/app/cluster-building-blocks/QuieterReporting.h
tcarmelveilleux: I want the change predicate to be able to fully override, so that we can get. Fixed the 0 -> 0 case.

src/app/cluster-building-blocks/QuieterReporting.h
tcarmelveilleux: done

src/app/cluster-building-blocks/QuieterReporting.h
tcarmelveilleux: I prefer being able to update the policy, and I don't think it's a big deal. Will leave like this.

src/app/cluster-building-blocks/QuieterReporting.h
tcarmelveilleux: Fixed.

PR #34269, opened 2024-07-09
[Darwin] MTRDeviceController getSessionForNode should always use live MTRDevice object

jtung-apple:
In the case that the MTRDeviceController API client doesn't retain MTRDevice objects for known node IDs, the next `getSessionForNode` call may not know if the device uses Thread, and doesn't make use of the CASE/subscription pool mechanism to behave well on the network.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -539,6 +539,9 @@ static NSString * const sLastInitialSubscribeLatencyKey = @"lastInitialSubscribe
 - (void)dealloc
 {
     [[NSNotificationCenter defaultCenter] removeObserver:_systemTimeChangeObserverToken];
+
+    // TODO: retain cycle and clean up https://github.com/project-chip/connectedhomeip/issues/34267
+    MTR_LOG("%@ dealloc called", self);
 }
 
 - (NSString *)description
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -541,7 +541,7 @@ static NSString * const sLastInitialSubscribeLatencyKey = @"lastInitialSubscribe
     [[NSNotificationCenter defaultCenter] removeObserver:_systemTimeChangeObserverToken];
 
     // TODO: retain cycle and clean up https://github.com/project-chip/connectedhomeip/issues/34267
-    MTR_LOG("%@ dealloc called", self);
+    MTR_LOG("MTRDevice dealloc: %p", self);
 }
 
 - (NSString *)description
--- a/src/darwin/Framework/CHIP/MTRDeviceController.mm
+++ b/src/darwin/Framework/CHIP/MTRDeviceController.mm
@@ -605,11 +605,20 @@ using namespace chip::Tracing::DarwinFramework;
             MTR_LOG("Loaded attribute values for %lu nodes from storage for controller uuid %@", static_cast<unsigned long>(clusterDataByNode.count), self->_uniqueIdentifier);
 
             std::lock_guard lock(self->_deviceMapLock);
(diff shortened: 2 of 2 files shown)


src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -539,6 +539,9 @@ - (instancetype)initWithNodeID:(NSNumber *)nodeID controller:(MTRDeviceControlle
 - (void)dealloc
 {
     [[NSNotificationCenter defaultCenter] removeObserver:_systemTimeChangeObserverToken];
+
+    // TODO: retain cycle and clean up https://github.com/project-chip/connectedhomeip/issues/34267
+    MTR_LOG("%@ dealloc called", self);
```
woody-apple, line 544: can't log self in dealloc FYI

woody-apple: (pointer is OK)

src/darwin/Framework/CHIP/MTRDeviceController.mm
```diff
@@ -605,11 +605,20 @@ - (BOOL)startup:(MTRDeviceControllerStartupParamsInternal *)startupParams
...
                 MTRDevice * device = [self _setupDeviceForNodeID:nodeID prefetchedClusterData:clusterData];
                 MTR_LOG("Loaded %lu cluster data from storage for %@", static_cast<unsigned long>(clusterData.count), device);
+
+                [deviceList addObject:device];
             }
+
+#define kSecondsToWaitBeforeAPIClientRetainsMTRDevice 60
+            // Keep the devices retained for a while, in case API client doesn't immedieately retain them
+            dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t) (kSecondsToWaitBeforeAPIClientRetainsMTRDevice * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
+                MTR_LOG("MTRDeviceController: un-retain devices loaded at startup %lu", static_cast<unsigned long>(deviceList.count));
```
woody-apple, line 620: Do we need to do anything if the MTRDeviceController is shutdown?

src/darwin/Framework/CHIP/MTRDevice.mm
woody-apple, line 544:
```suggestion
    MTR_LOG("MTRDevice dealloc: %p", self);
```

src/darwin/Framework/CHIP/MTRDeviceController.mm
jtung-apple: We don't need to do anything as no behavior is performed on the MTRDevice objects themselves. And controller shutdown would call -invalidate on the device objects.

src/darwin/Framework/CHIP/MTRDeviceController.mm
```diff
@@ -605,11 +605,20 @@ - (BOOL)startup:(MTRDeviceControllerStartupParamsInternal *)startupParams
...
             for (NSNumber * nodeID in clusterDataByNode) {
                 NSDictionary * clusterData = clusterDataByNode[nodeID];
                 MTRDevice * device = [self _setupDeviceForNodeID:nodeID prefetchedClusterData:clusterData];
                 MTR_LOG("Loaded %lu cluster data from storage for %@", static_cast<unsigned long>(clusterData.count), device);
+
+                [deviceList addObject:device];
             }
+
+#define kSecondsToWaitBeforeAPIClientRetainsMTRDevice 60
+            // Keep the devices retained for a while, in case API client doesn't immedieately retain them
```
bzbarsky-apple, line 618:
```suggestion
            // Keep the devices retained for a while, in case API client doesn't immediately retain them.
            //
            // Note that this is just an optimization to avoid throwing the information away and immediately
            // re-reading it from storage.
```

bzbarsky-apple (approved):

PR #34272, opened 2024-07-10
[Silabs] [917SoC] Fixing sleepy with non LCD build fix

chirag-silabs:
917SoC Sleepy needs sleeptimer and lcd also needs sleeptimer so merging them together

Fixes the build with icd enabled and lcd disabled case

- removing the nvm direct write from the KVS since tickless is implemented now

### DIFF
--- a/third_party/silabs/SiWx917_sdk.gni
+++ b/third_party/silabs/SiWx917_sdk.gni
@@ -291,6 +291,14 @@ template("siwx917_sdk") {
     if (invoker.enable_dic) {
       _include_dirs += [ "${chip_root}/third_party/silabs/mqtt/stack" ]
     }
+
+    if (chip_enable_icd_server || !disable_lcd) {
+      defines += [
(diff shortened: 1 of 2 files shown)


third_party/silabs/SiWx917_sdk.gni
```diff
@@ -312,6 +310,12 @@ template("siwx917_sdk") {
       ]
     }
 
+    if (chip_enable_icd_server || !disable_lcd) {
```
mkardous-silabs, line 313: small nit - can this go before the lcd block of defines?

third_party/silabs/SiWx917_sdk.gni
```diff
@@ -793,6 +794,14 @@ template("siwx917_sdk") {
...
```
mkardous-silabs, line 797: same comment

mkardous-silabs (approved):

yufengwangca (approved):

PR #34273, opened 2024-07-10
move AttemptToGetTimeFromTrustedNode out of TIME_SYNC_ENABLE_TSC_FEATURE blocks

mykrupp:
… since it is called out of TIME_SYNC_ENABLE_TSC_FEATURE blocks

AttemptToGetTimeFromTrustedNode() is called outside of TIME_SYNC_ENABLE_TSC_FEATURE blocks, so it should be declared outside of a TIME_SYNC_ENABLE_TSC_FEATURE. Otherwise, if the time sync client feature is turned off (time_sync_enable_tsc_feature = 0), we get a build error - error: 'AttemptToGetTimeFromTrustedNode' was not declared in this scope

### DIFF
--- a/src/app/clusters/time-synchronization-server/time-synchronization-server.h
+++ b/src/app/clusters/time-synchronization-server/time-synchronization-server.h
@@ -124,8 +124,9 @@ public:
     void OnAttributeData(const ConcreteDataAttributePath & aPath, TLV::TLVReader * apData, const StatusIB & aStatus) override;
     void OnDone(ReadClient * apReadClient) override;
 
-    CHIP_ERROR AttemptToGetTimeFromTrustedNode();
 #endif
+
+    CHIP_ERROR AttemptToGetTimeFromTrustedNode();
 
     // Platform event handler functions
     void OnPlatformEventFn(const DeviceLayer::ChipDeviceEvent & event);


src/app/clusters/time-synchronization-server/time-synchronization-server.h
```diff
@@ -124,9 +124,10 @@ class TimeSynchronizationServer : public FabricTable::Delegate
     void OnAttributeData(const ConcreteDataAttributePath & aPath, TLV::TLVReader * apData, const StatusIB & aStatus) override;
     void OnDone(ReadClient * apReadClient) override;
 
-    CHIP_ERROR AttemptToGetTimeFromTrustedNode();
 #endif
 
+    CHIP_ERROR AttemptToGetTimeFromTrustedNode();
```
mkardous-silabs, line 129: This seems to be a time sync specific function. Should we instead #ifdef out were it used insteand?

mykrupp: That is another option

cecille: This is fine - the internals of the function are ifdef'd.

cecille: ie, 6 of one...

cecille (approved):

jmartinez-silabs (approved):

PR #34276, opened 2024-07-10
Impl basic information home location attribute

hicklin:
This PR Fixes #33568.

This PR depends on PR #34208 that adds the Device Location attribute to the Basic Information cluster XML.

This is still work in progress as we try to figure out the best way to store and retrieve the Device Location information.


src/include/platform/internal/GenericConfigurationManagerImpl.ipp
```diff
@@ -586,6 +586,57 @@ CHIP_ERROR GenericConfigurationManagerImpl<ConfigClass>::SetFailSafeArmed(bool v
...
+    ReturnErrorOnFailure(tlvReader.Next(TLV::AnonymousTag()));
+    ReturnErrorOnFailure(loc->Decode(tlvReader));
+
+    if (loc.IsNull())
+    {
+        location.SetNull();
+        return CHIP_NO_ERROR;
+    }
+
+    memcpy((void *) location.Value().locationName.data(), loc.Value().locationName.data(), loc.Value().locationName.size());
```
hicklin, line 616: @bzbarsky-apple Is there a simpler way of achieving this? I was expecting the `Decode` method to copy the contents of the `tlvReader` to the `location` argument. Maybe there is a different `Decode` method that does this?

bzbarsky-apple: Here's the thing.  `DeviceLocatioType` [sic] is an alias for `DeviceLocation::TypeInfo::Type`, which is `Nullable<Structs::HomeLocationStruct::Type>`.  And that's defined as:

```
struct Type
{
public:
    chip::CharSpan locationName;
```

so `location.Value().locationName` is a `const char *`.

The code above is hiding that "oh, I am copying to a pointer that points to const data, which is generally not OK" with that C-style cast (which I bet got added because the compiler complained about passing a const pointer to a non-const `void*` arg!), but there is absolutely no way to tell from the code in this function that this memcpy business is safe, much less telling it from inside the `Decode` method or anything like that.

The normal way I would approach this is:

1. Define a struct type inheriting from `DeviceLocatioType` [sic] which has storage for the data.
2. Define a copy assignment operator on that which takes `DeviceLocatioType` [sic] and copies things correctly.
3. Pass in a reference to this new type.
4. Decode into a `DeviceLocatioType` on the stack and then assign into the outparam.

bzbarsky-apple (changes requested):
That "write to const memory area" thing absolutely needs to be fixed.

And this PR probably needs to be broken up into separate "update XML" and "add implementation" PRs.

src/include/platform/internal/GenericConfigurationManagerImpl.ipp
```diff
@@ -586,6 +586,59 @@ CHIP_ERROR GenericConfigurationManagerImpl<ConfigClass>::SetFailSafeArmed(bool v
...
+
+    char locationNameBuffer[1] = { 0 };
+    DeviceLocatioType loc({
+        .locationName = CharSpan(locationNameBuffer, 0),
+    });
+
+    ReturnErrorOnFailure(tlvReader.Next(TLV::AnonymousTag()));
+    ReturnErrorOnFailure(loc->Decode(tlvReader));
+
+    // This would not be needed if the Decode methed proprly decodes a null value.
```
hicklin, line 615: @bzbarsky-apple related to the slack message, I cannot find a way to decode a null `DeviceLocatioType` from the encoded bytes. This is the solution that I have came up with so far.

andy31415:
Closing stale PR.

PR #34284, opened 2024-07-10
make zap_regen_all auto-trigger a buildenv bootstrap/activate when run if not already activated

andy31415:
We seem to always need to run this script in a bootstrapped environment. Make it auto-do so.


bzbarsky-apple (approved):
Thank you!

tcarmelveilleux (approved):

PR #34288, opened 2024-07-10
Update AttributePathExpandIterator to be able to use the IM/DM DataModel split

andy31415:
### Changes

- Split AttributePathExpandIterator.h/cpp into 3:
   - `Ember` representing the old code identically
   - `DataModel` that uses the data model interface exclusively
   - `Checked` that runs both Ember and DataModel in parallel and asserts that the outputs are identical
- Usage of the 3 is controlled by a `chip_use_data_model_interface` that has enabled/disabled/check. Linux defaults to check (slower runs, however will exercise all unit tests and integration tests to run the interfaces in parallel) all others default to disabled to have minimal deltas
- Ensured we start having a `DataModel` available. In particular `InteractionModelEngine` provides a datamodel that defaults to the codegendatamodel (i.e. ember) if not set

InteractionModel namespace is duplicated with `chip::Protocols::InteractionModel` so some namespaces/using logic had to be updated.

image

Low coverage for `Checked` is because the error reporting lines are never reached (all tests pass).


andy31415:
Converted to draft:

- [ ] need to fix compile errors
- [ ] need to ensure that if data model is not in use, we do not link in the codegen data model (so we do not incur the flash size). Hopefully easy to figure out as without we seem to get link errors anyway.

src/app/AttributePathExpandIterator-Checked.cpp
```diff
@@ -0,0 +1,91 @@
...
+void AttributePathExpandIteratorChecked::CheckOutputsIdentical(const char * msg)
+{
+    ConcreteAttributePath dmPath;
+    ConcreteAttributePath emPath;
+
+    bool dmResult = mDataModelIterator.Get(dmPath);
+    bool emResult = mEmberIterator.Get(emPath);
+
+    // NOTE: extra logic because mExpanded is NOT considered in operator== (ugly...)
+    if ((dmResult == emResult) && (dmPath == emPath) && (dmPath.mExpanded == emPath.mExpanded))
```
bzbarsky-apple, line 73: If both results are false, aren't the paths possibly garbage, so they could test different and whatnot?

src/app/AttributePathExpandIterator-Checked.cpp
```diff
@@ -0,0 +1,91 @@
...
+    {
+        // outputs are identical. All is good
+        return;
+    }
+
+    ChipLogProgress(Test, "Different paths in DM vs EMBER (%d and %d) in %s", dmResult, emResult, msg);
+    ChipLogProgress(Test, "   DM PATH:    0x%X/0x%X/0x%X (%s)", static_cast<int>(dmPath.mEndpointId),
+                    static_cast<int>(dmPath.mClusterId), static_cast<int>(dmPath.mAttributeId),
```
bzbarsky-apple, line 81: Casting unsigned 32-bit values to signed 32-bit for logging is rather odd.  Especially because the log format is then hex....

Is there a reason we are not using existing format/value bits for MEIs here?

src/app/AttributePathExpandIterator-DataModel.cpp
```diff
@@ -0,0 +1,243 @@
...
+        }
+
+        return std::nullopt;
+    }
+
+    // advance the existing attribute id if it can be advanced
+    VerifyOrReturnValue(mpAttributePath->mValue.HasWildcardAttributeId(), std::nullopt);
+
+    // NOTE: this switch matches the order in GlobalAttributesNotInMetadata so that order-dependent
+    //       checks match
```
bzbarsky-apple, line 84: How about implementing this as a loop over `GlobalAttributesNotInMetadata` so we don't have multiple sources of truth?  The whole point of GlobalAttributesNotInMetadata was to have one source of truth for that set.

src/app/AttributePathExpandIterator-DataModel.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+    // If this is a null iterator, or the attribute id of current cluster info is not a wildcard attribute id, then this function
+    // will do nothing, since we won't be expanding the wildcard attribute ids under a cluster.
+    VerifyOrReturn(mpAttributePath != nullptr && mpAttributePath->mValue.HasWildcardAttributeId());
+
+    // Reset to the first attribute of the current cluster.
+    // It is expected that the output path is already positioned on a proper cluster
+    //
+    AttributeEntry entry = mDataModel->FirstAttribute(mOutputPath);
+    VerifyOrDie(entry.IsValid()); // TODO: is this sane? do we require validity here?
```
bzbarsky-apple, line 165: I don't understand this block of code.  Is this just trying to reset the data model's optimization cache bits?  Why?

src/app/AttributePathExpandIterator-DataModel.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+    VerifyOrReturnValue(mpAttributePath->mValue.HasWildcardEndpointId(), std::nullopt);
+
+    EndpointId id = mDataModel->NextEndpoint(mOutputPath.mEndpointId);
+    return (id != kInvalidEndpointId) ? std::make_optional(id) : std::nullopt;
+}
+
+void AttributePathExpandIteratorDataModel::ResetCurrentCluster()
+{
+    ChipLogError(Test, "Reset current cluster data model");
```
bzbarsky-apple, line 155: Why is this a test log?

src/app/clusters/administrator-commissioning-server/administrator-commissioning-server.cpp
bzbarsky-apple, line 1: The changes to this file are unrelated to AttributePathExpandIterator.

src/app/clusters/icd-management-server/icd-management-server.cpp
bzbarsky-apple, line 1: The changes to this file are unrelated to AttributePathExpandIterator.

src/app/tests/TestCommandInteraction.cpp
bzbarsky-apple, line 1: Changes to this file seem unrelated to AttributePathExpandIterator.

src/controller/tests/data_model/DataModelFixtures.cpp
bzbarsky-apple, line 1: Again, the changes seem to be unrelated.

If the issue is that now we are pulling in some headers that define a different InteractionModel namespace than Protocols::InteractionModel so that things are now ambiguous.....  Ideally either we make the mass-changes in a separate PR or we make the naming different so this is not a problem.

src/controller/tests/data_model/TestCommands.cpp
bzbarsky-apple, line 1: As above.

bzbarsky-apple (changes requested):

src/app/clusters/administrator-commissioning-server/administrator-commissioning-server.cpp
andy31415: They are a sideffect of header inclusion unfortunately: InteractionModel is now visible as a namespace because we have both `chip::Protocols::InteractionModel` and `chip::InteractionModel` and things do not compile without these changes.

Are you asking for a separate PR for these?

src/app/AttributePathExpandIterator-Checked.cpp
andy31415: I was unsure here - the code does reset the paths to a known value, it looks like this in AttributePathExpandIterator:

```
    // Reset to default, invalid value.
    mOutputPath = ConcreteReadAttributePath();
    return false;
```

And the updated code matches that. So you are correct that technically if the values are false, the path content should not matter. However at the same time I am making a stronger statement here: the values match even if returning false.

I can change it if you think it is important, however for now I think this has a stronger coverage.

src/app/AttributePathExpandIterator-Checked.cpp
andy31415: This is expected code that is never reached, so I did not update it past basic formatting. Updated now.

src/app/AttributePathExpandIterator-DataModel.cpp
andy31415: Updated.

src/app/AttributePathExpandIterator-DataModel.cpp
andy31415: Debug remnants. Removed.

src/app/AttributePathExpandIterator-DataModel.cpp
andy31415: Removed this double-check as our "next" implementation has all the logic existing regarding selecting the first valid cluster or making `Get` return false.

src/app/clusters/administrator-commissioning-server/administrator-commissioning-server.cpp
andy31415: Updated in #34376

src/app/AttributePathExpandIterator-Checked.cpp
bzbarsky-apple: This is just making assumptions not guaranteed by the API.  So it could start randomly failing....  I'm OK with that if you are.

andy31415: I expect we will delete the `-Check` interface at some point since I assume we will want to switch to the `data model interface` fully at some point in time. Because of that, I prefer the existing strict bug-by-bug compatibility.

andy31415: Updated to not check patch equivalence on false return values. This seems easier to review and if I had to change code anyway (either comment or this, may as well update it).

src/app/AttributePathExpandIterator-DataModel.cpp
```diff
@@ -152,18 +158,11 @@ std::optional<ClusterId> AttributePathExpandIteratorDataModel::NextEndpointId()
...
     // will do nothing, since we won't be expanding the wildcard attribute ids under a cluster.
     VerifyOrReturn(mpAttributePath != nullptr && mpAttributePath->mValue.HasWildcardAttributeId());
 
-    // Reset to the first attribute of the current cluster.
-    // It is expected that the output path is already positioned on a proper cluster
-    //
-    AttributeEntry entry = mDataModel->FirstAttribute(mOutputPath);
-    VerifyOrDie(entry.IsValid()); // TODO: is this sane? do we require validity here?
-
+    // Proceed path expansion to ask for the first attribute of the current cluster
```
bzbarsky-apple, line 165:
```suggestion
    // Reset path expansion to ask for the first attribute of the current cluster
```

bzbarsky-apple (approved):

jmartinez-silabs (approved):

src/app/AttributePathExpandIterator-DataModel.cpp
```diff
@@ -0,0 +1,242 @@
...
+        }
+
+        return std::nullopt;
+    }
+
+    // advance the existing attribute id if it can be advanced
+    VerifyOrReturnValue(mpAttributePath->mValue.HasWildcardAttributeId(), std::nullopt);
+
+    // Ensure (including ordering) that GlobalAttributesNotInMetadata is reported as needed
+    for (unsigned i = 0; i < ArraySize(GlobalAttributesNotInMetadata); i++)
```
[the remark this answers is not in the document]
andy31415: I am unsure if it makes a difference. I would prefer not to touch it because that preserves CI runs so far and also I believe `int/unsigned` tells the compiler to just use whatever is most efficient. We likely do not specifically need a size for this loop.

PR #34289, opened 2024-07-10
Fix DNS-SD TCP advertisement to use a bitmap for Client and Server.

pidarped:
The spec currently advertises TCP support independently for TCP client and server in a bitmap form. The reception logic supports this advertisement model but the advertising side was not using this mode. 
This PR updates the advertisement sending side to support the same.

### DIFF
--- a/src/app/server/Dnssd.cpp
+++ b/src/app/server/Dnssd.cpp
@@ -213,6 +213,9 @@ CHIP_ERROR DnssdServer::AdvertiseOperational()
         AddICDKeyToAdvertisement(advertiseParameters);
 #endif
 
+#if INET_CONFIG_ENABLE_TCP_ENDPOINT
+        advertiseParameters.SetTCPSupportModes(chip::Dnssd::TCPModeAdvertise::kTCPClientServer);
+#endif
         auto & mdnsAdvertiser = chip::Dnssd::ServiceAdvertiser::Instance();
 
         ChipLogProgress(Discovery, "Advertise operational node " ChipLogFormatX64 "-" ChipLogFormatX64,
--- a/src/app/server/Dnssd.h
+++ b/src/app/server/Dnssd.h
@@ -96,6 +96,7 @@ public:
 
     void SetICDManager(ICDManager * manager) { mICDManager = manager; };
 #endif
+
     /// Start operational advertising
     CHIP_ERROR AdvertiseOperational();
 
--- a/src/lib/dnssd/Advertiser.h
+++ b/src/lib/dnssd/Advertiser.h
@@ -55,6 +55,14 @@ enum class ICDModeAdvertise : uint8_t
     kLIT,  // The ICD is currently operating as a LIT. ICD=1 in DNS-SD key/value pairs.
 };
 
+enum class TCPModeAdvertise : uint16_t
+{
+    kNone            = 0,                         // The device does not support TCP.
+    kTCPClient       = 1 << 1,                    // The device supports the TCP client.
+    kTCPServer       = 1 << 2,                    // The device supports the TCP server.
+    kTCPClientServer = (kTCPClient | kTCPServer), // The device supports both the TCP client and server.
+};
+
 template <class Derived>
 class BaseAdvertisingParams
 {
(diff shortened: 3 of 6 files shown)


yufengwangca (approved):

src/lib/dnssd/Advertiser.h
```diff
@@ -55,6 +55,14 @@ enum class ICDModeAdvertise : uint8_t
     kLIT,  // The ICD is currently operating as a LIT. ICD=1 in DNS-SD key/value pairs.
 };
 
+enum class TCPModeAdvertise : uint8_t
+{
+    kNone            = 0,                         // The device does not support TCP.
+    kTCPClient       = 1 << 1,                    // The device supports the TCP client.
+    kTCPServer       = 1 << 2,                    // The device supports the TCP server.
+    kTCPClientServer = (kTCPClient | kTCPServer), // The device supports both the TCP client and server.
```
andy31415, line 63: why not using a BitFlags for storage and have the enums for client/server only?

I am concerned about constants  for all combinations - looks neat when 2 options exist, however once we go to 3 or more options these custom constants will explode.

src/lib/dnssd/Advertiser.h
```diff
@@ -124,7 +131,7 @@ class BaseAdvertisingParams
     uint8_t mMacStorage[kMaxMacSize] = {};
     size_t mMacLength                = 0;
     std::optional<ReliableMessageProtocolConfig> mLocalMRPConfig;
-    std::optional<bool> mTcpSupported;
+    TCPModeAdvertise mTcpSupportModes  = TCPModeAdvertise::kNone;
```
andy31415, line 134: the std::optional was there to explicity have support for "no T option" vs saying "T=0" for example. 

This update makes it so that "T=0" is never exposed anymore. Is this the correct path to lose that capability?

andy31415:
@pidarped please update the PR summary. The title is insufficient: "fix the DNSSD to use a bitmap" ... what does it get fixed? what was the problem, what is the change containing and how was it tested? can we have some form of unit testing?

We should know what to expect in the PR content by reading the PR summary.

src/app/server/Dnssd.h
```diff
@@ -96,6 +96,12 @@ class DLL_EXPORT DnssdServer : public ICDStateObserver
 
     void SetICDManager(ICDManager * manager) { mICDManager = manager; };
 #endif
+
+#if INET_CONFIG_ENABLE_TCP_ENDPOINT
+    template <class AdvertisingParams>
+    void AddTCPSupportToAdvertisement(AdvertisingParams & advParams);
```
andy31415, line 102: Why is this a template? why does this get wrapped instead of saying `advParams.SetTCPSupportModes(chip::Dnssd::TCPModeAdvertise::kTCPClientServer);`? It seems we are replacing one line of code with calling a template that implements one line of code. Why the extra indirection and complexity?

src/lib/dnssd/Discovery_ImplPlatform.cpp
```diff
@@ -240,7 +240,8 @@ CHIP_ERROR CopyTxtRecord(TxtFieldKey key, char * buffer, size_t bufferLen, const
     switch (key)
     {
     case TxtFieldKey::kTcpSupported:
-        return CopyTextRecordValue(buffer, bufferLen, params.GetTcpSupported());
+        VerifyOrReturnError(params.GetTCPSupportModes() != TCPModeAdvertise::kNone, CHIP_ERROR_UNINITIALIZED);
+        return CopyTextRecordValue(buffer, bufferLen, static_cast<uint16_t>(params.GetTCPSupportModes()));
```
andy31415, line 244: why uint16_t? why not to_underlying usage?

src/app/server/Dnssd.cpp
```diff
@@ -213,6 +221,9 @@ CHIP_ERROR DnssdServer::AdvertiseOperational()
         AddICDKeyToAdvertisement(advertiseParameters);
 #endif
 
+#if INET_CONFIG_ENABLE_TCP_ENDPOINT
+        AddTCPSupportToAdvertisement(advertiseParameters);
```
bzbarsky-apple, line 225: I don't get it.  If we just do this in the one spot, why do we need the complexity with the template and whatnot?  Or for that matter why do we need the separate function if it's a one-liner?

If you just copied what ICD does, note that ICD is calling AddICDKeyToAdvertisement from multiple places with different parameter types, and the logic for ICD is in fact complicated enough to warrant a separate function.

src/app/server/Dnssd.cpp
```diff
@@ -213,6 +221,9 @@ CHIP_ERROR DnssdServer::AdvertiseOperational()
         AddICDKeyToAdvertisement(advertiseParameters);
 #endif
 
+#if INET_CONFIG_ENABLE_TCP_ENDPOINT
```
bzbarsky-apple, line 224: Is it in facttrue that if INET_CONFIG_ENABLE_TCP_ENDPOINT is set we are in fact listening for TCP connections?  It used to very much not be true, but maybe that's changed now?

bzbarsky-apple (changes requested):

src/app/server/Dnssd.h
pidarped: Removed the function and replaced with a direct call to SetTCPSupportModes.

src/lib/dnssd/Discovery_ImplPlatform.cpp
pidarped: The CopyTextRecordValue() is an overloaded function but it did not have one for uint8_t. Converted the enum to uint16_t and changed it to to_underlying() over here.

bzbarsky-apple (approved):

PR #34290, opened 2024-07-10
[Fix] Possible fix for matter-test-scripts issue #227: Remove PICS from OPSTATE tests

j-ororke:
- Removed PICS checks in TC_OpstateCommon.py module and replaced with attribute and command checks from connected endpoint during tests.


src/python_testing/TC_RVCOPSTATE_2_1.py
```diff
@@ -165,63 +174,56 @@ async def test_TC_RVCOPSTATE_2_1(self):
...
                 if self.is_ci:
                     self.write_to_app_pipe('{"Name": "ErrorEvent", "Error": "UnableToStartOrResume"}')
                 else:
-                    self.wait_for_user_input(prompt_msg=f"{test_step}, and press Enter when done.\n")
+                    self.wait_for_user_input(prompt_msg="press Enter when done.\n")
                 await self.read_and_validate_opstate(step="6h", expected_state=Clusters.OperationalState.Enums.OperationalStateEnum.kError)
             if self.check_pics("RVCOPSTATE.S.M.ST_SEEKING_CHARGER"):
-                test_step = "Manually put the device in the seeking charger state"
-                self.print_step("6i", test_step)
+                self.print_step("6i", "Manually put the device in the seeking charger state")
```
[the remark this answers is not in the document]
j-ororke: My apologies,
I have corrected this in the code here now, this issue should be resolved.

src/python_testing/TC_OpstateCommon.py
```diff
@@ -213,9 +213,25 @@ def STEPS_TC_OPSTATE_BASE_1_1(self) -> list[TestStep]:
 
     async def TEST_TC_OPSTATE_BASE_1_1(self, endpoint=1, cluster_revision=1, feature_map=0):
         cluster = self.test_info.cluster
+
+        # Gathering Available Attributes and associated ids
         attributes = cluster.Attributes
-        events = cluster.Events
+        OPSTATE_attr_list = attributes.AttributeList
```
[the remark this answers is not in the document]
j-ororke: Hi Cecille,
I have resolved the upper-cased variable names in the latest changes, the variable names should all now be lower case.

src/python_testing/TC_OpstateCommon.py
```diff
@@ -1014,9 +1071,29 @@ def STEPS_TC_OPSTATE_BASE_2_5(self) -> list[TestStep]:
...
+        generated_cmd_list = await self.read_single_attribute_check_success(endpoint=endpoint, cluster=cluster, attribute=OPSTATE_gencmd_list)
+        pause_cmd_id = commands.Pause.command_id
+        start_cmd_id = commands.Start.command_id
+        stop_cmd_id = commands.Stop.command_id
+        resume_cmd_id = commands.Resume.command_id
+        ocr_cmd_id = commands.OperationalCommandResponse.command_id
+
+        # Gathering Available Event and associated id
         events = cluster.Events
+        Oprtcmplt_event = bool(events.OperationCompletion.event_id)
```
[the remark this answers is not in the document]
j-ororke: Hi Cecille,
I have removed these variables and if checks, and replaced them with the prior PICS checks that were in the code before, so this should now be resolved and back to the way it was previously.

src/python_testing/TC_OpstateCommon.py
```diff
@@ -1014,9 +1071,29 @@ def STEPS_TC_OPSTATE_BASE_2_5(self) -> list[TestStep]:
 
     async def TEST_TC_OPSTATE_BASE_2_5(self, endpoint=1):
         cluster = self.test_info.cluster
+
+        # Gathering Available Attributes and associated ids
         attributes = cluster.Attributes
+        OPSTATE_attr_list = attributes.AttributeList
+        attribute_list = await self.read_single_attribute_check_success(endpoint=endpoint, cluster=cluster, attribute=OPSTATE_attr_list)
+        countdown_time_attr_id = attributes.CountdownTime.attribute_id
+        oprtnlstate_attr_id = attributes.OperationalState.attribute_id
```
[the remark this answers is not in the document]
j-ororke: Hi Cecille,
I removed the operational state attribute id variables and changed the if checks for operational state attribute ids in attribute list to not checking for those any longer.

src/python_testing/TC_RVCOPSTATE_2_1.py
```diff
@@ -89,15 +89,24 @@ async def test_TC_RVCOPSTATE_2_1(self):
                 asserts.fail("The --app-pid flag must be set when PICS_SDK_CI_ONLY is set")
             self.app_pipe = self.app_pipe + str(app_pid)
 
+        cluster = Clusters.RvcOperationalState
         attributes = Clusters.RvcOperationalState.Attributes
+        RVCOpstate_attr_list = attributes.AttributeList
```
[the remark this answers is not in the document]
j-ororke: Hi Cecille,
I have created 2 new functions in the TC_OpstateCommon module to create dictionaries containing the attributes and commands for opstate clusters. 
Please let me know if this works.

src/python_testing/TC_OpstateCommon.py
```diff
@@ -117,6 +117,43 @@ def write_to_app_pipe(self, command):
         with open(self.app_pipe, "w") as app_pipe:
             app_pipe.write(command + "\n")
 
+    async def get_attributes(self, cluster, endpoint: int):
```
cecille, line 120: Can you remove this? A dict of string -> int is more fragile that what's already code generated. You can get rid of all this code and just do ex. 

start_cmd_id = commands.StartCommand.id.

In fact, you don't even need to pull those into separate attributes - you can just use them from the codegen directly.

j-ororke: Hi Cecille, 
I have removed the variable functions created in OpstateCommon module and followed the method you mentioned above and called the variable names directly in the if checks now in tests contained in OpstateCommon and RVCOPSTATE_2_1 test module. This should reduce the amount of clutter for variable setting that was in the tests previously.

src/python_testing/TC_RVCOPSTATE_2_1.py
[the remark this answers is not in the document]
j-ororke: Thank you Cecille,
I have replaced the prior variables and removed the variable functions and now just calling the codegen'd names directly in the if checks in the above 2 test modules (TC_RVCOPSTATE_2_1 and TC_OpstateCommon.py) that should hopefully help to remove some of the clutter that was being noticed due to the prior variable setting and make it a bit cleaner.

src/python_testing/TC_OpstateCommon.py
```diff
@@ -263,7 +267,7 @@ async def TEST_TC_OPSTATE_BASE_1_1(self, endpoint=1, cluster_revision=1, feature
                 events.OperationalError.event_id,
             ]
 
-            if self.check_pics(f"{self.test_info.pics_code}.S.E01"):
+            if self.pics_guard(self.check_pics(f"{self.test_info.pics_code}.S.E01")):
```
cecille, line 270: This one will end up going away with a quick test plan change - event list has been officially removed from the spec, so this step will never get run. 

I opened a (very short) issue - this step can be deleted from the test plan and script. I'd say do it in a separate PR to keep the changes focused. Can do it after submitting this one.

https://github.com/project-chip/matter-test-scripts/issues/371

cecille: Feel free to resolve this once you've read it, we can continue on the issue

src/python_testing/TC_OpstateCommon.py
```diff
@@ -369,7 +374,7 @@ async def TEST_TC_OPSTATE_BASE_2_1(self, endpoint=1):
 
         # STEP 4: TH reads from the DUT the CountdownTime attribute
         self.step(4)
-        if self.pics_guard(self.check_pics(f"{self.test_info.pics_code}.S.A0002")):
+        if attributes.CountdownTime.attribute_id in attribute_list:
```
cecille, line 377: One of the things that's getting removed here is that pics_guard automatically marks the step as skipped if the condition doesn't hold. This change removes that.

You had another PR were I commented that we should framework some of these attribute checks now that we have some functions in place to do it. https://github.com/project-chip/matter-test-scripts/issues/371. This might be a good PR to start this on, where the pics guard is directly swapped for something like attribute_guard (not totally sold on the naming, so if you can come up with a batter name, please do)

j-ororke: Hi Cecille,
I have created a function in MatterBaseTest class named "attributes_guard" as you had mentioned using the check_attributes function in the matter_testing_scripts helper module.
This should work I hope, please see the update to the issue https://github.com/project-chip/matter-test-scripts/issues/370 you have filed and let me know if this will help resolve the above comment.
Thank you!

src/python_testing/TC_RVCOPSTATE_2_1.py
```diff
@@ -94,15 +94,16 @@ async def test_TC_RVCOPSTATE_2_1(self):
                 asserts.fail("The --app-pid flag must be set when PICS_SDK_CI_ONLY is set")
```
cecille, line 94: Can't comment up there, but having the endpoint be set is actually pretty important for this test because endpoint == None defaults to EP0, which is not what we want here. Can you add an assertion below that line that endpoint is actually set?

src/python_testing/matter_testing_infrastructure/chip/testing/matter_testing.py
```diff
@@ -1415,6 +1415,24 @@ def pics_guard(self, pics_condition: bool):
...
+
+              self.step("1")
+              if self.attribute_guard(condition1_needs_to_be_true_to_execute):
+                  # do the test for step 1
+
+              self.step("2")
+              if self.attribute_guard(condition2_needs_to_be_false_to_skip_step):
+                  # skip step 2 if condition not met
+           """
+        attr_condition = await asyncio.wait_for(should_run_test_on_endpoint(self, has_attribute(attribute_condition)), timeout=60)
```
cecille, line 1431: The should_run_test_on_endpoint function is going to do a wildcard read on every call.

Given that we're moving now towards doing a lot of checks based on the device conformance, I think it would make sense to move that wildcard read up to the setup_class function and have it stored as self.wildcard and use that. Then you can just call has_attribute directly here (and swap over the should_run... to use the stored wildcard)

cecille: Actually, we can remove some duplication with the basic composition tests if you do the read then save that as self.endpoints / self.endpoints_tlv then swap over from using wildcard.

j-ororke: Hi Cecille,
I have updated the above method for the attributes_guard function as discussed during our call this afternoon, thank you

src/python_testing/TC_RVCOPSTATE_2_1.py
j-ororke: Hi Cecille, 
I have added an additional check in to make sure the endpoint was not None or 0, please let me know if this will resolve the issue noticed and mentioned in your above comment!

src/python_testing/matter_testing_infrastructure/chip/testing/matter_testing.py
```diff
@@ -1094,6 +1094,10 @@ def setup_class(self):
         self.current_step_index = 0
         self.step_start_time = datetime.now(timezone.utc)
         self.step_skipped = False
+        self.global_wildcard = asyncio.wait_for(self.default_controller.Read(self.dut_node_id, [(Clusters.Descriptor), Attribute.AttributePath(None, None, GlobalAttributeIds.ATTRIBUTE_LIST_ID), Attribute.AttributePath(
+            None, None, GlobalAttributeIds.FEATURE_MAP_ID), Attribute.AttributePath(None, None, GlobalAttributeIds.ACCEPTED_COMMAND_LIST_ID)]), timeout=60)
+        # self.stored_global_wildcard stores value of self.global_wildcard after first async call, appears needed in order to not timeout during commissioning
+        self.stored_global_wildcard = None
```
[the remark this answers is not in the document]
j-ororke: Hi Cecille,
I have sent a slack message regarding the issue noticed here, please let me know if ok to continue with current method given that the asyncio.run() seems to cause timeout during commissioning, appears not to be able to find endpoint before timeout.

src/python_testing/matter_testing_infrastructure/chip/testing/matter_testing.py
```diff
@@ -1451,6 +1455,26 @@ def pics_guard(self, pics_condition: bool):
             self.mark_current_step_skipped()
         return pics_condition
 
+    async def attributes_guard(self, endpoint: int, attribute: str):
```
cecille, line 1458:
```suggestion
    async def attribute_guard(self, endpoint: int, attribute: ClusterObjects.ClusterAttributeDescriptor):
```
WDYT about wrapping the async part into a asyncio.run so that the caller doesn't need to await this? I think it might be less error prone for authors.

j-ororke: I have added the above code and think that is much better than it was previously. 👍

j-ororke: I think the problem of implementing the asyncio.run() presviously in place of the await here is that I found it was inside an event loop, so if we ran it from the attribute_guard(), then it complained that we can't do that since asyncio.run() has to be run outside of an active event loop. I had also tried with replacing the await in the test itself, but it also complained there as well.
However, recollecting what you mentioned during our teams call on tuesday, you had mentioned to do the assignment of the self.global_wildcard var after the self.is_commissioning was done, thus I have now moved the assignment of the self.global_wildcard var to the setup_test() instead of the setup_class() in matter testing support module, then put it in the if statement for if not self.is_commissioning and self.runner_hook, this allows us to do the assignment of self.global_wildcard at the start of the test runs and allows us to do it in a asyncio.run().
Please let me know if this will be acceptable.

j-ororke: My apologies, while moving the self.global_wildcard var to the setup_test() in matter_testing support module worked on some tests, it unfortunately did not work on some others, you had mentioned this was a concern during the call as well, but I had not noticed this during initial testing. 
I have reverted the code back to a prior version in order to get it to work on all the tests.

src/python_testing/test_metadata.yaml
```diff
@@ -97,6 +97,7 @@ slow_tests:
     - { name: TC_PS_2_3.py, duration: 30 seconds }
     - { name: TC_RR_1_1.py, duration: 25 seconds }
     - { name: TC_SWTCH.py, duration: 1 minute }
+    - { name: TC_TestAttrAvail.py, duration: 30 seconds }
```
cecille, line 100: Does this test actually take that long?

cecille (approved):

src/python_testing/test_metadata.yaml
j-ororke: This takes a bit of time, maybe not entirely 30 seconds, but due to the commissioning for runs 1 and 2 it does take some time to complete.

PR #34293, opened 2024-07-11
[occupancy-sensing]Updated occupancy sensing cluster SDK & Sample app(all-clusters-app) implementation to Rev 5 (Matter 1.4)

OliverFan1:
- Updated occupancy sensing cluster SDK implementation based on XML changes(PR#34163) for Occupancy Sensing rev 5 

- Updated occupancy sensing cluster sample app(all-clusters-app) implementation to Rev 5 in the zap and matter file of all-clusters-app

- Remove "occupancy-sensor-server" from "ClientDirectories" in zap_cluster_list.json to resolve a building issue found from rebuilding the chip-tool


src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.h
```diff
@@ -17,8 +17,76 @@
...
+
+namespace chip {
+namespace app {
+namespace Clusters {
+namespace OccupancySensing {
+
+/**
+ * Interface to help manage the Hold Time Limits of the Occupancy Sensing Cluster.
+ */
+class HoldTimeLimitsManager : public AttributeAccessInterface
```
tcarmelveilleux, line 36: The HoldTimeLimitsManager should not have `AttributeAccessInterface`, but should be business-logic only.

There needs to be a separate class for the OccupancySensing AttributeAccessInterface.

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.h
```diff
@@ -17,8 +17,76 @@
...
+{
+public:
+    static constexpr size_t kOccupancySensingServerMaxEndpointCount =
+        MATTER_DM_OCCUPANCY_SENSING_CLUSTER_SERVER_ENDPOINT_COUNT + CHIP_DEVICE_CONFIG_DYNAMIC_ENDPOINT_COUNT;
```
tcarmelveilleux, line 40: Do not use this pattern for memory allocation of endpoints as it over-allocates.

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.h
```diff
@@ -17,8 +17,76 @@
...
+
+    private:
+        /// @brief Returns the index of the HoldTimeLimits associated to an endpoint
+        /// @param[in] endpoint target endpoint
+        /// @param[out] endpointIndex index of the corresponding HoldTimeLimits for an endpoint
+        /// @return CHIP_NO_ERROR or CHIP_ERROR_NOT_FOUND, CHIP_ERROR_INVALID_ARGUMENT if invalid endpoint
+        CHIP_ERROR FindHoldTimeLimitsIndex(EndpointId endpoint, size_t & endpointIndex);
+
+        Structs::HoldTimeLimitsStruct::Type mHoldTimeLimitsStructs[kOccupancySensingServerMaxEndpointCount];
+    };
```
tcarmelveilleux, line 58: This should be an interface

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.h
```diff
@@ -17,8 +17,76 @@
...
+    HoldTimeLimits mHoldTimeLimits;
+
+    // Instance
+    static HoldTimeLimitsManager mInstance;
+};
+
+inline bool HasFeature(EndpointId ep, Feature feature)
+{
+    uint32_t map;
+    bool success = (Attributes::FeatureMap::Get(ep, &map) == Protocols::InteractionModel::Status::Success);
```
tcarmelveilleux, line 87: Do not use global attribute getter for the features. Make the specific example have to initialize an instance with a `chip::BitFlags<>` of the technology features, from the initialization code, rather than using this round-about way of reading features which makes the ZAP file the source of truth in a way that is harder to understand than code, and which prevents one example from supporting multiple technologies.

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.cpp
```diff
@@ -16,14 +16,144 @@
...
+}
+
+HoldTimeLimitsManager HoldTimeLimitsManager::mInstance;
+
+HoldTimeLimitsManager & HoldTimeLimitsManager::Instance()
+{
+    return mInstance;
+}
+
+CHIP_ERROR HoldTimeLimitsManager::Init()
```
tcarmelveilleux, line 86: The overall cluster implementation needs tests before the PR can land.

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.cpp
```diff
@@ -16,14 +16,144 @@
...
+{
+    // Prevents re-initializing
+    VerifyOrReturnError(!mIsInitialized, CHIP_ERROR_INCORRECT_STATE);
+
+    for (size_t i = 0; i <= kOccupancySensingServerMaxEndpointCount; i++)
+    {
+        if (emberAfContainsServer(EndpointId(i), OccupancySensing::Id))
```
tcarmelveilleux, line 93: Don't use `emberAfContainsServer` to set the capabilities, as this, again, prevents code-based configuration of the cluster.

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.cpp
```diff
@@ -16,14 +16,144 @@
...
+        {
+            Structs::HoldTimeLimitsStruct::Type holdTimeLimitsInit;
+
+            // Set up some sane initial values for hold time limits structures
+            holdTimeLimitsInit.holdTimeMin     = 1;
+            holdTimeLimitsInit.holdTimeMax     = 300;
+            holdTimeLimitsInit.holdTimeDefault = 10;
```
tcarmelveilleux, line 100: These should always come from the client and not from a "default" value since there are no good default values that apply to all products.

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.cpp
```diff
@@ -16,14 +16,144 @@
...
+    mIsInitialized = true;
+    return CHIP_NO_ERROR;
+}
+
+// AttributeAccessInterface
+CHIP_ERROR HoldTimeLimitsManager::Read(const ConcreteReadAttributePath & aPath, AttributeValueEncoder & aEncoder)
+{
+    switch (aPath.mAttributeId)
+    {
+    case Attributes::HoldTimeLimits::Id: {
```
tcarmelveilleux, line 117: Please make the following attributes also available by AttributeAccessInterface:

- HoldTime
- FeatureMap
- Occupancy
- OccupancySensorType
- OccupancySensorTypeBitmap

This will allow testing the cluster business logic and will also prevent mismatch between code-backed implementation of the feature and the ZAP file.

tcarmelveilleux (changes requested):
- [x] Avoid mixing code-backed and Attribute-store-backed attributes
- [ ] Introduce tests in the same PR as the feature
- [ ] Avoid global singletons to maintain state
- [ ] Initialize the cluster from the application
- [ ] Maintain state dynamically per cluster rather than using constants like `CHIP_DEVICE_CONFIG_DYNAMIC_ENDPOINT_COUNT`
- [ ] Report features and revision based on code, not on ZAP-configured values (so that we can properly configure runtime instances)

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.cpp
bzbarsky-apple, line 93: This is not going to work right for, say, a bridge that needs to add these endpoints dynamically.

One possible place to initialize the entry for an enpoint id is when this cluster is actually initialized on that endpoint.   See `emberAfOccupancySensingClusterServerInitCallback` below.

That's for initialization the SDK itself can do.  For initialization the application does, they would need to implement `emberAfOccupancySensorInitCallback` (note no "Server" in the name), assuming you stick to the ember-callbacks method of initializing things.

bzbarsky-apple, line 93: `EndpointId(i)` is not right.  You have no idea what the right endpoint ID is for a particular index outside the fixed endpoints.

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.cpp
```diff
@@ -16,14 +16,144 @@
...
+
+        Structs::HoldTimeLimitsStruct::Type * holdTimeLimitsStruct =
+            HoldTimeLimitsManager::Instance().mHoldTimeLimits.GetHoldTimeLimitsStruct(aPath.mEndpointId);
+        Structs::HoldTimeLimitsStruct::Type res;
```
bzbarsky-apple, line 121: Why do we need an extra copy?  Why can't we just `aEncoder.Encode(*holdTimeLimitsStruct)`?

Separately, how confident are we that `holdTimeLimitsStruct` is never null here?

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.cpp
```diff
@@ -16,14 +16,144 @@
...
+} // namespace Clusters
+} // namespace app
+} // namespace chip
 
 using namespace chip;
+using namespace chip::app;
+using namespace chip::app::Clusters;
 using namespace chip::app::Clusters::OccupancySensing;
 
+void chip::app::Clusters::OccupancySensing::emberAfPluginOccupancyClusterServerPostInitCallback(EndpointId endpoint) {}
```
bzbarsky-apple, line 155: Why is this here?  I know this is pre-existing, but it's just dead code that should be removed, no?

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.h
```diff
@@ -1,6 +1,6 @@
 /*
  *
- *    Copyright (c) 2020 Project CHIP Authors
+ *    Copyright (c) 2024 Project CHIP Authors
```
bzbarsky-apple, line 3:
```suggestion
 *    Copyright (c) 2020-2024 Project CHIP Authors
```

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.cpp
OliverFan1: Yes, you're right, dead code removed.

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.h
OliverFan1: Changed in new code, please check

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.h
OliverFan1: Please check the new code, I refer to Electrical Energy Measurement Cluster's implementation for this, if I'm in the wrong way, help suggest a reference, thanks~

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.h
OliverFan1: Changed in new code, please check

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.h
OliverFan1: Changed in new code, please check

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.cpp
OliverFan1: I'm not fully understand "Introduce tests in the same PR as the feature", I find recently there is a new test plan for Occupancy Sensing Rev5, https://github.com/CHIP-Specifications/chip-test-plans/pull/4291, do you mean this SDK implementation need to pass the new test plan?

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.cpp
OliverFan1: Changed in new code, please check

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.cpp
OliverFan1: Changed in new code, do the initialization in application code, please check

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.cpp
OliverFan1: HoldTime and FeatureMap has been added in new code, please check if it is the right way to do it, if it is, I'll add the rest in

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.cpp
```diff
@@ -82,11 +212,16 @@ void halOccupancyStateChangedCallback(EndpointId endpoint, HalOccupancyState occ
...
 
-void emberAfPluginOccupancyClusterServerPostInitCallback(EndpointId endpoint) {}
-
 HalOccupancySensorType __attribute__((weak)) halOccupancyGetSensorType(EndpointId endpoint)
 {
     return HAL_OCCUPANCY_SENSOR_TYPE_PIR;
 }
 
-void MatterOccupancySensingPluginServerInitCallback() {}
+void MatterOccupancySensingPluginServerInitCallback()
```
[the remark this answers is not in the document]
OliverFan1: Changed in new code, please check

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.cpp
OliverFan1: Changed in new code, do the initialization in application code, please check

OliverFan1: Changed in new code, do the initialization in application code, got rid of this EndpointId(i), please check

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.cpp
OliverFan1: Added check on null pointer, and got rid of the extra copy, please check

examples/all-clusters-app/all-clusters-common/src/occupancy-sensing-stub.cpp
```diff
@@ -0,0 +1,57 @@
...
+using namespace chip::app::Clusters::OccupancySensing;
+using namespace chip::app::Clusters::OccupancySensing::Structs;
+using namespace chip::DeviceLayer;
+
+using chip::Protocols::InteractionModel::Status;
+
+static std::unique_ptr<OccupancySensingAttrAccess>
+    gAttrAccess[MATTER_DM_OCCUPANCY_SENSING_CLUSTER_SERVER_ENDPOINT_COUNT + CHIP_DEVICE_CONFIG_DYNAMIC_ENDPOINT_COUNT];
+
+void emberAfOccupancySensingClusterInitCallback(chip::EndpointId endpointId)
```
bzbarsky-apple, line 34:
```suggestion
void emberAfOccupancySensingClusterInitCallback(EndpointId endpointId)
```

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.cpp
```diff
@@ -16,12 +16,156 @@
...
+
+namespace chip {
+namespace app {
+namespace Clusters {
+namespace OccupancySensing {
+
+Structs::HoldTimeLimitsStruct::Type
+    mHoldTimeLimitsStructs[MATTER_DM_OCCUPANCY_SENSING_CLUSTER_SERVER_ENDPOINT_COUNT + CHIP_DEVICE_CONFIG_DYNAMIC_ENDPOINT_COUNT];
+
+uint16_t mHoldTime[MATTER_DM_OCCUPANCY_SENSING_CLUSTER_SERVER_ENDPOINT_COUNT + CHIP_DEVICE_CONFIG_DYNAMIC_ENDPOINT_COUNT];
```
bzbarsky-apple, line 38: Should these be in an anonymous namespace so they're not global variables that anything can touch from anywhere?

And the "m" prefix is for class members, which these are not.  "s" would be better if these are file statics (as they should be).

OliverFan1: Thanks, yes, it would be better to include them in an anonymous namespace, code added, names changed as well

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.cpp
```diff
@@ -16,12 +16,158 @@
...
+CHIP_ERROR OccupancySensingAttrAccess::Read(const ConcreteReadAttributePath & aPath, AttributeValueEncoder & aEncoder)
+{
+    VerifyOrDie(aPath.mClusterId == app::Clusters::OccupancySensing::Id);
+
+    switch (aPath.mAttributeId)
+    {
+    case Attributes::FeatureMap::Id:
+        ReturnErrorOnFailure(aEncoder.Encode(mFeature));
+        break;
+    case Attributes::HoldTime::Id: {
```
bzbarsky-apple, line 62: Should add FeatureMap and HoldTime to the list of attributes handled via AttributeAccessInterface in `src/app/zap-templates/zcl/zcl-with-test-extensions.json` and `src/app/zap-templates/zcl/zcl.json`

bzbarsky-apple (approved):

tcarmelveilleux (approved):

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.cpp
tcarmelveilleux: For occupancy, we can leave it using the global accessors as it's how they are all currently written.

examples/all-clusters-app/all-clusters-common/src/occupancy-sensing-stub.cpp
```diff
@@ -0,0 +1,57 @@
...
+using namespace chip::app::Clusters::OccupancySensing;
+using namespace chip::app::Clusters::OccupancySensing::Structs;
+using namespace chip::DeviceLayer;
+
+using chip::Protocols::InteractionModel::Status;
+
+static std::unique_ptr<OccupancySensingAttrAccess>
+    gAttrAccess[MATTER_DM_OCCUPANCY_SENSING_CLUSTER_SERVER_ENDPOINT_COUNT + CHIP_DEVICE_CONFIG_DYNAMIC_ENDPOINT_COUNT];
+
+void emberAfOccupancySensingClusterInitCallback(EndpointId endpointId)
```
tcarmelveilleux, line 34: In a follow-up PR, please move this out of ClusterInitCallback and move it into the application code. This works for now.

examples/all-clusters-app/all-clusters-common/src/occupancy-sensing-stub.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+using namespace chip;
+using namespace chip::app::Clusters;
+using namespace chip::app::Clusters::OccupancySensing;
+using namespace chip::app::Clusters::OccupancySensing::Structs;
+using namespace chip::DeviceLayer;
+
+using chip::Protocols::InteractionModel::Status;
+
+static std::unique_ptr<OccupancySensingAttrAccess>
```
tcarmelveilleux, line 31: This is a generic name. Please call it gOccupancySensorClusterInstances and put it in anonymous namespace.

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.h
```diff
@@ -17,15 +17,48 @@
...
- *
- * @param endpoint Endpoint that is being initialized  Ver.: always
- */
-void emberAfPluginOccupancyClusterServerPostInitCallback(chip::EndpointId endpoint);
+namespace chip {
+namespace app {
+namespace Clusters {
+namespace OccupancySensing {
+
+class OccupancySensingAttrAccess : public AttributeAccessInterface
```
tcarmelveilleux, line 33: Call it `Instance` to follow the model we have right now in other places. This will be `chip::app::Clusters::OccupancySensing::Instance`.

src/app/clusters/occupancy-sensor-server/occupancy-sensor-server.cpp
```diff
@@ -16,12 +16,158 @@
...
+namespace chip {
+namespace app {
+namespace Clusters {
+namespace OccupancySensing {
+
+namespace {
+Structs::HoldTimeLimitsStruct::Type
+    sHoldTimeLimitsStructs[MATTER_DM_OCCUPANCY_SENSING_CLUSTER_SERVER_ENDPOINT_COUNT + CHIP_DEVICE_CONFIG_DYNAMIC_ENDPOINT_COUNT];
+
+uint16_t sHoldTime[MATTER_DM_OCCUPANCY_SENSING_CLUSTER_SERVER_ENDPOINT_COUNT + CHIP_DEVICE_CONFIG_DYNAMIC_ENDPOINT_COUNT];
```
tcarmelveilleux, line 39: These should be members of the class, not globals. Can be fixed as a follow-up.

tcarmelveilleux (approved):

OliverFan1:
Can anybody help to merge, if it's approved? Thank you

PR #34306, opened 2024-07-12
Support raw ByteSpan values correctly in SafeAttributePersistenceProvider

ksperling-apple:
The `DefaultAttributePersistenceProvider` implementation of `SafeReadValue()` had the undocumented behavior of enforcing that the value exactly filled the provided buffer. This does not make sense for an API that purports to store general ByteSpan values. (Note that `SafeReadValue()` was previously protected in `SafeAttributePersistenceProvider`, but public in the `DefaultAttributePersistenceProvider` implementation.)

Instead move this validation into `ReadScalarValue()` in `SafeAttributePersistenceProvider`, and separate `InternalReadValue()` into one method that does the only the actual read, and a second one that performs the additional validation for reading typed ember attributes (strings with length prefixes etc). Use the former one from `SafeReadValue()`.

Also fix possible out of bounds access when reading ember strings by ensuring we have read the length byte(s) before attempting to interpret them.

### DIFF
--- a/src/app/SafeAttributePersistenceProvider.h
+++ b/src/app/SafeAttributePersistenceProvider.h
@@ -40,8 +40,8 @@ public:
 
     // The following API provides helper functions to simplify the access of commonly used types.
     // The API may not be complete.
-    // Currently implemented write and read types are: uint8_t, uint16_t, uint32_t, unit64_t and
-    // their nullable varieties, and bool.
+    // Currently implemented write and read types are: bool, uint8_t, uint16_t, uint32_t, unit64_t and
+    // their nullable varieties, as well as ByteSpans.
 
     /**
      * Write an attribute value of type intX, uintX or bool to non-volatile memory.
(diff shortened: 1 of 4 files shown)


src/app/SafeAttributePersistenceProvider.h
```diff
@@ -50,7 +50,7 @@ class SafeAttributePersistenceProvider
      * @param [in] aValue the data to write.
      */
     template <typename T, std::enable_if_t<std::is_integral<T>::value, bool> = true>
-    CHIP_ERROR WriteScalarValue(const ConcreteAttributePath & aPath, T & aValue)
+    CHIP_ERROR WriteScalarValue(const ConcreteAttributePath & aPath, const T & aValue)
```
bzbarsky-apple, line 53: Is there a reason this is pass-by-reference at all, not pass-by-value?

bzbarsky-apple (approved):

src/app/SafeAttributePersistenceProvider.h
ksperling-apple: Not that I can see. Changed it to by value.

yufengwangca (approved):

PR #34316, opened 2024-07-12
Reduce static footprint

nk-shelly:
These changes reduce memory usage in our firmware by about 30K when matter is disabled.
This is important to us, as matter is optional and not the only stack we support, making large static and global variables too expensive.


src/app/MessageDef/MessageDefHelper.cpp
```diff
@@ -37,26 +37,32 @@ namespace app {
 // this is used to run in signle thread for IM message debug purpose
 namespace {
 uint32_t gPrettyPrintingDepthLevel = 0;
-char gLineBuffer[CHIP_CONFIG_LOG_MESSAGE_MAX_SIZE];
-size_t gCurLineBufferSize = 0;
+size_t gCurLineBufferSize          = 0;
+
+std::vector<char> & GetLineBuffer()
+{
+    static std::vector<char> sLineBuffer(CHIP_CONFIG_LOG_MESSAGE_MAX_SIZE);
```
Damian-Nordic, line 44: Please don't use STL containers in the code that can be run on constrained devices.

Damian-Nordic (changes requested):
This change completely ignores constrained devices for which using the heap for global objects both adds unnecessary overhead and makes it more difficult to estimate the actual memory usage.
Why not to use the existing `Global<>` helper if you prefer to customize the global object handling for your platform?

rojer:
> This change completely ignores constrained devices for which using the heap for global objects both adds unnecessary overhead and makes it more difficult to estimate the actual memory usage. 

Hi Damian, Matter SDK uses dynamic allocation extensively, so arguably this battle has already been lost. We are very much aware of constrained environments - ours is very tight indeed, and the reason we are doing this is because in our case Matter can be enabled and disabled at runtime, and we don't want to be paying the static memory price when it's disabled.

rojer:
this change saves ~30K or RAM when Matter is disabled (not initialized), which is a significant amount of memory in our environment (ESP32-C3).

Damian-Nordic:
> Hi Damian, Matter SDK uses dynamic allocation extensively, so arguably this battle has already been lost. We are very much aware of constrained environments - ours is very tight indeed, and the reason we are doing this is because in our case Matter can be enabled and disabled at runtime, and we don't want to be paying the static memory price when it's disabled.

Well, our tests show that currently devices remain stable with less than 10kB of heap and the heap seems to be mainly used for temporary objects, which is great because when used for temporary objects, the heap can actually reduce the total memory usage. On the other hand, using the heap for lazy initialization of objects that are never destroyed may cause the heap fragmentation and hence larger total memory usage.

That's why I suggest to hide the global object initialization under a configurable abstraction, such as `Global<>`. Then we can both win without stepping on each others' toes :).

src/app/MessageDef/MessageDefHelper.cpp
nk-shelly: Changed to unique_ptr.

nk-shelly:
> That's why I suggest to hide the global object initialization under a configurable abstraction, such as `Global<>`.

Could you clarify that suggestion, please? Global<> doesn't seem to handle anything array-like at the moment. Also, we've changed it to allocate from heap in this PR too.

Damian-Nordic:
> Could you clarify that suggestion, please? Global<> doesn't seem to handle anything array-like at the moment. Also, we've changed it to allocate from heap in this PR too.

I mean that e.g. instead of turning:
```
static BindingTable sInstance;
```
into a static `unique_ptr`, you can turn it into:
```
static Global<BindingTable> sInstance;
```
and configure `Global` to use lazy initialization.

I'm not against your changes in `Global<>` (this is up to @ksperling-apple), but only against turning all globals into heap-allocated objects.

nk-shelly:
> I'm not against your changes in `Global<>` (this is up to @ksperling-apple), but only against turning all globals into heap-allocated objects unconditionally. Speaking of arrays, `Global<int[20]> a` built OK to me. Can you clarify what doesn't work for you?

Thanks Damian, I'm not sure it will work properly even if it compiles. Actually, I'm pretty sure the original create and destroy functions didn't - they would neither construct nor destruct each element of the array. It doesn't seem such use was intended.

Also, another thing Global<> doesn't support is non-default constructed variables and I remember at least one such case here: https://github.com/project-chip/connectedhomeip/pull/34316/files#diff-971ae85e36e17285500f7fe2d75688390cdf62f52383edf011b064f510d9cb21R773

Is a compile-time option to switch between static and heap allocations what you actually want? I'm not sure Global<> is the best place for it or that LAZY_INIT is the best option. Maybe adding a new option would work better (and not necessarily be restricted to Global<>)?

Damian-Nordic:
> Thanks Damian, I'm not sure it will work properly even if it compiles. Actually, I'm pretty sure the original create and destroy functions didn't - they would neither construct nor destruct each element of the array. It doesn't seem such use was intended.
> 
> Also, another thing Global<> doesn't support is non-default constructed variables and I remember at least one such case here: https://github.com/project-chip/connectedhomeip/pull/34316/files#diff-971ae85e36e17285500f7fe2d75688390cdf62f52383edf011b064f510d9cb21R773
> 
> Is a compile-time option to switch between static and heap allocations what you actually want? I'm not sure Global<> is the best place for it or that LAZY_INIT is the best option. Maybe adding a new option would work better (and not necessarily be restricted to Global<>)?

Ok, I won't argue about the original methods though `new (mStorage) T();` should work OK for `T = MyClass[10]`. The thing is it surely can be fixed to support arrays. It's true that non-default constructed variables are an issue but we have options to work around them:
- Replace the constructor with `Init()` method.
- `CommonCaseDeviceServerInitParams` is just the default, and an application doesn't have to use it. You can just construct `ServerInitParams` from scratch (or we can do it if your approach is preferred by others :)).

I personally like the idea of adding a new option for `Global<>` but it would be good to get feedback from @ksperling-apple as he added the original one.

bzbarsky-apple (changes requested):

bzbarsky-apple:
And yes, I suspect adding more configurability options for Global and using it more is the way to go.

bauerschwan:
I do fully agree to @bzbarsky-apple that these changes (going from static to dynamic) introduce high risks for most of the use cases. Instead, this kind of optimization should be handled by compile-time configuration.
Obviously, this would mean that the original intention of this PR (better memory usage when Matter is disabled) would not be possible.

rojer:
ok, thanks for feedback everyone.
a compile-time option to make static allocations dynamic, with default to keep things static, would also do the jobs.
`#define CHIP_ENABLE_DYNAMIC_STATICS` that makes all `Global<>`s lazy or something.
is that something that would be acceptable?

bzbarsky-apple:
I would call it CHIP_CONFIG_DYNAMIC_GLOBALS or something, but yes, that seems reasonable.

nk-shelly:
> This change completely ignores constrained devices for which using the heap for global objects both adds unnecessary overhead and makes it more difficult to estimate the actual memory usage. Why not to use the existing `Global<>` helper if you prefer to customize the global object handling for your platform?

Done - added a CHIP_CONFIG_DYNAMIC_GLOBALS option (disabled by default).

src/app/MessageDef/MessageDefHelper.cpp
```diff
@@ -37,26 +37,32 @@ namespace app {
 // this is used to run in signle thread for IM message debug purpose
 namespace {
 uint32_t gPrettyPrintingDepthLevel = 0;
-char gLineBuffer[CHIP_CONFIG_LOG_MESSAGE_MAX_SIZE];
-size_t gCurLineBufferSize = 0;
+size_t gCurLineBufferSize          = 0;
+
+char * GetLineBuffer()
+{
+    static auto sLineBuffer = std::make_unique<char[]>(CHIP_CONFIG_LOG_MESSAGE_MAX_SIZE);
```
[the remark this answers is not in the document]
nk-shelly: It is now opt-in (CHIP_CONFIG_DYNAMIC_GLOBALS option).

src/app/server/Server.cpp
```diff
@@ -248,20 +237,27 @@ CHIP_ERROR Server::Init(const ServerInitParams & initParams)
     chip::Dnssd::Resolver::Instance().Init(DeviceLayer::UDPEndPointManager());
 
 #if CHIP_CONFIG_ENABLE_SERVER_IM_EVENT
+#define CHIP_NUM_EVENT_LOGGING_BUFFERS 3
+    static auto sInfoEventBuffer      = std::make_unique<uint8_t[]>(CHIP_DEVICE_CONFIG_EVENT_LOGGING_INFO_BUFFER_SIZE);
```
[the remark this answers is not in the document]
nk-shelly: Done.

Damian-Nordic (commented):
Looking good overall. A couple of style / compliance suggestions.

src/lib/core/Global.h
```diff
@@ -97,6 +111,31 @@ class Global
...
 
 private:
+#if CHIP_CONFIG_DYNAMIC_GLOBALS
+    T & _get() { return *reinterpret_cast<T *>(create(static_cast<T *>(nullptr))); }
+    template <class U, std::size_t N>
+    static U * create(U (*value)[N])
+    {
+#if CHIP_CONFIG_GLOBALS_NO_DESTRUCT
+        static U * ptr = new U[N];
+        return ptr;
```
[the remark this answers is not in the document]
nk-shelly: There is a slight setback with this:

Platform::MemoryAlloc, MemoryFree & cie will VerifyOrDieWithoutLogging if used after chip::Platform::MemoryShutdown() or before chip::Platform::MemoryInit().

IIUC, Darwin is the only platform that has CHIP_CONFIG_GLOBALS_LAZY_INIT enabled, in which case __cxa_atexit((f), (p), &__dso_handle) is used to call the destructor of the Global<T>.

However, chip::Platform::MemoryShutdown() happens inside main, before any of the Global<T>s are destroyed, so... if T or any member of T needs to chip::Platform::MemoryFree() something from it's destructor it will crash.

@bzbarsky-apple, is my understanding correct? Any ideas how to mitigate this? I've now moved a bunch of large classes into Global<T>s and I'm starting to wonder if that was good idea in the first place...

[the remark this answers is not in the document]
Damian-Nordic: oh, right. Using `Platform::New` isn't then a good idea in the case of globals. In such a case, I'm fine with the current code, so please only address the comment about `static` local variables.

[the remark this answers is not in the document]
nk-shelly: @Damian-Nordic my concern is it might be already broken. In case any member of a member of the classes that I converted to Global<T>s has a destructor that use Platform::Memory*() that would lead to a crash (at least on Darwin).

Even if I change the destroy functions to not use the platform memory allocator, nothing is preventing destructors from using them.

Well, now that I think of it some more, destructors of global variables would have caused that issue even without my changes, so the conversion might be safe in this case.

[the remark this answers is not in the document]
Damian-Nordic: Right, I guess the idea behind `Platform::Memory*` APIs was to use it to allocate objects that are explicitly released using methods like `Shutdown()`. If this is the case, then `Global<>` is simply not a valid use case scenario for `Platform::Memory` so my suggestion was wrong.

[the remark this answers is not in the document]
nk-shelly: Thanks @Damian-Nordic!
I've removed the static local and changed the Linux platform config to use CHIP_CONFIG_DYNAMIC_GLOBALS to test the change. I'm not sure how to test this (now opt-in) option unless a platform is actually using it. If that's not OK, any suggestions are welcome.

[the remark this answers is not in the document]
bzbarsky-apple: Not sure whether you need something from me here, but using Platform::Memory for something with "static" lifetime indeed does not work.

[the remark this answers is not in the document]
nk-shelly: Thanks for confirming @bzbarsky-apple!

src/lib/core/Global.h
```diff
@@ -85,6 +85,16 @@ class Global
     /// NOT thread-safe, external synchronization is required.
     T & get() { return _get(); }
     T * operator->() { return &_get(); }
+    template <class U = std::remove_extent_t<T>>
+    U & operator[](std::size_t i)
```
bzbarsky-apple, line 89: So this is only needed for the `&sLoggingBuffer[0]` thing, right?  You could just as easily replace that with `sLoggingBuffer.get()` or `&sLoggingBuffer.get()[0]` if desired and not add the extra complexity here.

bzbarsky-apple (commented):
I only looked at Global.h for now, in case changes there affect how the other bits work, but I tried this PR on Mac with `#define CHIP_CONFIG_DYNAMIC_GLOBALS 1` and it fails unit tests with ASAN failures....  So that's worth looking into, perhaps.  But maybe after we settle on how Global.h should work.

src/lib/core/Global.h
```diff
@@ -99,23 +109,57 @@ class Global
 private:
     // Zero-initialize everything. We should technically leave mStorage uninitialized,
     // but that can sometimes cause clang to be unable to constant-initialize the object.
+#if CHIP_CONFIG_DYNAMIC_GLOBALS
+    unsigned char * mStorage = nullptr;
```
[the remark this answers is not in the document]
nk-shelly: static void destroy(void * value) { static_cast<T *>(value)->~T(); }
wouldn't work when T is an array, that's why I added the _destroy_array loop:
error: object expression of non-scalar type cannot be used in a pseudo-destructor expression
Arrays need to be supported in both cases.

src/lib/core/Global.h
```diff
@@ -85,6 +85,16 @@ class Global
     /// NOT thread-safe, external synchronization is required.
     T & get() { return _get(); }
     T * operator->() { return &_get(); }
+    template <class U = std::remove_extent_t<T>>
```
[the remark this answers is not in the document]
nk-shelly: Removed the whole function (also size()).

src/lib/core/Global.h
```diff
@@ -99,23 +109,57 @@ class Global
...
     alignas(T) unsigned char mStorage[sizeof(T)] = {};
+#endif // CHIP_CONFIG_DYNAMIC_GLOBALS
     OnceStrategy mOnce;
 
     T & _get()
     {
-        T * value = reinterpret_cast<T *>(mStorage);
-        mOnce.call(&create, value);
-        return *value;
+#if CHIP_CONFIG_DYNAMIC_GLOBALS
```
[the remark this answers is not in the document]
nk-shelly: Done

src/lib/core/Global.h
nk-shelly: Removed

src/lib/core/Global.h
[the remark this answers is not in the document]
nk-shelly: Replaced the #else part with:
```
    template <typename U = T, std::enable_if_t<std::is_array_v<U>, bool> = true>
    static void destroy(void * value)
    {
        using V = std::remove_extent_t<T>;
        for (std::size_t i = 0; i != std::extent_v<T>; ++i)
            (static_cast<V *>(value) + i)->~V();
    }

    template <typename U = T, std::enable_if_t<!std::is_array_v<U>, bool> = true>
    static void destroy(void * value)
    {
        static_cast<T *>(value)->~T();
    }
```

src/platform/Linux/CHIPPlatformConfig.h
```diff
@@ -27,6 +27,12 @@
 
 #define CHIP_CONFIG_ABORT() abort()
 
+extern "C" int __cxa_atexit(void (*func)(void *), void * arg, void * d);
+#define CHIP_CXA_ATEXIT(f, p) __cxa_atexit((f), (p), nullptr)
+
+#define CHIP_CONFIG_DYNAMICALLY_ALLOCATED_GLOBALS 1
```
Damian-Nordic, line 33: Is this just for CI or you're going to merge this? If the latter, I'm not sure it's a good idea to set this unconditionally for Linux. Perhaps, adding a GN flag to enable this or at least defining these conditionally using `#ifndef` would help Linux device makers to select the desired configuration.

Damian-Nordic (approved):
Thanks for taking the `Global<>` path :)

src/platform/Linux/CHIPPlatformConfig.h
nk-shelly: OK, I'll revert this for now. GN is new to me, so I would appreciate any ideas about how I could go about enabling this just for CI.

nk-shelly:
> I only looked at Global.h for now, in case changes there affect how the other bits work, but I tried this PR on Mac with `#define CHIP_CONFIG_DYNAMIC_GLOBALS 1` and it fails unit tests with ASAN failures.... So that's worth looking into, perhaps. But maybe after we settle on how Global.h should work.

How can I reproduce this? `ninja -C out/host check` seems to be passing on my Mac. Is ASAN enabled by default and how should I run the tests with it if not?

src/app/server/Server.cpp
```diff
@@ -253,19 +254,19 @@ CHIP_ERROR Server::Init(const ServerInitParams & initParams)
...
+    err = sGlobalEventIdCounter->Init(mDeviceStorage, DefaultStorageKeyAllocator::IMEventNumber(),
+                                      CHIP_DEVICE_CONFIG_EVENT_ID_COUNTER_EPOCH);
     SuccessOrExit(err);
 
     {
         ::chip::app::LogStorageResources logStorageResources[] = {
-            { &sDebugEventBuffer[0], sizeof(sDebugEventBuffer), ::chip::app::PriorityLevel::Debug },
-            { &sInfoEventBuffer[0], sizeof(sInfoEventBuffer), ::chip::app::PriorityLevel::Info },
-            { &sCritEventBuffer[0], sizeof(sCritEventBuffer), ::chip::app::PriorityLevel::Critical }
+            { sDebugEventBuffer.get(), CHIP_DEVICE_CONFIG_EVENT_LOGGING_DEBUG_BUFFER_SIZE, ::chip::app::PriorityLevel::Debug },
```
bzbarsky-apple, line 263: This can easily get out of sync with the definition of sDebugEventBuffer (and similar in MessageDefHelper).  The `size()` bit on Global was a good idea; it was just `operator[]` that I didn't think we needed...

src/lib/core/CHIPConfig.h
bzbarsky-apple, line 1: I am still getting ASAN failures on Mac with this PR if I set

```
#define CHIP_CONFIG_DYNAMICALLY_ALLOCATED_GLOBALS 1
```
and then do:
```
./gn_build.sh is_asan=true enable_host_clang_build=false enable_host_gcc_mbedtls_crypto_tests=false enable_host_clang_boringssl_crypto_tests=false chip_enable_dnssd_tests=false
```
I prodded at this, and the issue is that at least PlatformManager has the following places where it does not properly initialize members, but that was being covered up by 0-init of static things:

1. In the Darwin `PlatformManagerImpl`, `mRunLoopSem` should be getting initialized to nullptr and isn't.  This was the proximate cause of the test failure.
2. In `GenericPlatformManagerImpl`, `mAppEventHandlerList` should be getting initialized to nullptr and isn't.  I'm surprised anything real actually passed with that problem....  `mMsgLayerWasActive` is also not initialized in the constructor, but looks like it gets initialized in `_InitChipStack`, so that seems OK.

The question is: do any of the other things that this behavior change applies to have these sorts of problems?  Could probably use an audit.

src/app/reporting/ReportSchedulerImpl.h
```diff
@@ -62,7 +62,8 @@ class ReportSchedulerImpl : public ReportScheduler
 public:
     using Timeout = System::Clock::Timeout;
 
-    ReportSchedulerImpl(TimerDelegate * aTimerDelegate);
+    ReportSchedulerImpl() = default;
+    ReportSchedulerImpl(TimerDelegate * aTimerDelegate) : ReportScheduler(aTimerDelegate) {}
```
lpbeliveau-silabs, line 66: Why remove the nullptr check for this constructor?

src/app/server/Server.h
```diff
@@ -244,30 +245,31 @@ struct CommonCaseDeviceServerInitParams : public ServerInitParams
...
+            this->opCertStore = &sPersistentStorageOpCertStore.get();
         }
 
         // Injection of report scheduler WILL lead to two schedulers being allocated. As recommended above, this should only be used
         // for IN-TREE examples. If a default scheduler is desired, the basic ServerInitParams should be used by the application and
         // CommonCaseDeviceServerInitParams should not be allocated.
         if (this->reportScheduler == nullptr)
         {
-            reportScheduler = &sReportScheduler;
+            sReportScheduler->Init(&sTimerDelegate.get());
```
lpbeliveau-silabs, line 257: This will not get called on injection of a report scheduler, we can either keep the nullptr check in the constructor that takes a TimerDelegate as a param, or have an `EnsureInit` that takes no parameter and that gets called in an  `else` condition here.

src/app/server/Server.h
```diff
@@ -276,44 +278,44 @@ struct CommonCaseDeviceServerInitParams : public ServerInitParams
...
     }
 
 private:
-    static KvsPersistentStorageDelegate sKvsPersistenStorageDelegate;
-    static PersistentStorageOperationalKeystore sPersistentStorageOperationalKeystore;
-    static Credentials::PersistentStorageOpCertStore sPersistentStorageOpCertStore;
-    static Credentials::GroupDataProviderImpl sGroupDataProvider;
-    static chip::app::DefaultTimerDelegate sTimerDelegate;
-    static app::reporting::ReportSchedulerImpl sReportScheduler;
+    static Global<KvsPersistentStorageDelegate> sKvsPersistenStorageDelegate;
```
lpbeliveau-silabs, line 302: The `CommonCaseDeviceServerInitParams` is meant to be for in-tree examples only, if the aim was to reduce stack usage in a specific platform app, couldn't implementing a platform specific `ServerInitParams` achieve the same result without modifying the generic server? Or does the way the resources are spread out in the Server.cpp / Server.h prevent that?

nk-shelly: The aim is to reduce the memory used by global variables (not stack), especially ones that are not used. I'm not sure under what conditions the compiler/linker is allowed to optimise them away or what needs to be done to ensure that it does. At least some of those were certainly taking up space in our firmware and AFAICT we're not using CommonCaseDeviceServerInitParams explicitly. Our platform is esp32.

src/app/server/Server.h
nk-shelly: I'm not sure I get your point. Why initialise that global at all, especially when it's not going to be used? The whole point of this PR is to allow us to not even allocate the memory for it.

src/app/server/Server.h
lpbeliveau-silabs: If these objects were taking space even without using `CommonCaseDeviceServerInitParams` there is indeed something going wrong with the optimisation. Might be worth opening an issue in the future. otherwise LGTM.

src/app/server/Server.h
lpbeliveau-silabs: `this->reportScheduler == nullptr` only checks if there were no report Scheduler injected. If this check fails, that means the reportScheduler was injected somewhere else, so why not check that it is initialized?

something like:
`reportScheduler.EnsureInit()` to make sure that whatever is injected has been properly generated?

src/app/reporting/ReportSchedulerImpl.h
lpbeliveau-silabs: How and where would the Init method get called is the constructor using a TimerDelegate is used?

So far, it doesn't seem it happens. Also, if the constructor accepts a TimerDelegate, how is the Init method supposed to be used?

src/app/reporting/ReportScheduler.h
```diff
@@ -189,10 +189,17 @@ class ReportScheduler : public ReadHandler::Observer, public ICDStateObserver
         BitFlags<ReadHandlerNodeFlags> mFlags;
     };
 
+    ReportScheduler() : mTimerDelegate(nullptr) {}
     ReportScheduler(TimerDelegate * aTimerDelegate) : mTimerDelegate(aTimerDelegate) {}
 
     virtual ~ReportScheduler() = default;
 
+    void Init(TimerDelegate * aTimerDelegate)
```
lpbeliveau-silabs, line 197: How is this method supposed to be used with this constructor:
`ReportScheduler(TimerDelegate * aTimerDelegate) : mTimerDelegate(aTimerDelegate) {}`?

It would overwrite the TimerDelegate previously used, which defeats the purpose of having a constructor that accepts a TimerDelegate.

nk-shelly: Yes, right, Global<T> doesn't do anything except default constructors, so I added this Init method instead. I didn't delete the constructor with the pointer as I suspect it's already used by people. The intention was you can use the default constructor and then Init() later, or you can use the constructor with a pointer, in which case you don't need to Init().

src/app/server/Server.h
nk-shelly: Doesn't the same logic apply to all other things that can be injected? My impression was you need to Init() them before injecting (in this case you can alternatively construct it with a pointer). I don't see EnsureInit() used for any of the other ones...

src/app/reporting/ReportSchedulerImpl.h
nk-shelly: The Init() method is only used right above your other comment (https://github.com/project-chip/connectedhomeip/pull/34316/#discussion_r1696911888) and only for the sReportScheduler in CommonCaseDeviceServerInitParams, which is always default constructed.

src/app/server/Server.cpp
```diff
@@ -90,15 +91,15 @@ class DeviceTypeResolver : public chip::Access::AccessControl::DeviceTypeResolve
...
+Global<Server> Server::sServer;
 
 #if CHIP_CONFIG_ENABLE_SERVER_IM_EVENT
 #define CHIP_NUM_EVENT_LOGGING_BUFFERS 3
-static uint8_t sInfoEventBuffer[CHIP_DEVICE_CONFIG_EVENT_LOGGING_INFO_BUFFER_SIZE];
-static uint8_t sDebugEventBuffer[CHIP_DEVICE_CONFIG_EVENT_LOGGING_DEBUG_BUFFER_SIZE];
-static uint8_t sCritEventBuffer[CHIP_DEVICE_CONFIG_EVENT_LOGGING_CRIT_BUFFER_SIZE];
-static ::chip::PersistedCounter<chip::EventNumber> sGlobalEventIdCounter;
-static ::chip::app::CircularEventBuffer sLoggingBuffer[CHIP_NUM_EVENT_LOGGING_BUFFERS];
+static Global<uint8_t[CHIP_DEVICE_CONFIG_EVENT_LOGGING_INFO_BUFFER_SIZE]> sInfoEventBuffer;
```
ksperling-apple, line 98: As above, maybe these should be grouped into a struct constructed by a single `Global` ?

src/app/server/Server.cpp
ksperling-apple: Having the arrays inside a struct also means `sizeof` continues to work on the member as expected.

src/app/server/Server.cpp
```diff
@@ -762,23 +763,22 @@ void Server::ResumeSubscriptions()
...
 Credentials::IgnoreCertificateValidityPeriodPolicy Server::sDefaultCertValidityPolicy;
 
-KvsPersistentStorageDelegate CommonCaseDeviceServerInitParams::sKvsPersistenStorageDelegate;
-PersistentStorageOperationalKeystore CommonCaseDeviceServerInitParams::sPersistentStorageOperationalKeystore;
-Credentials::PersistentStorageOpCertStore CommonCaseDeviceServerInitParams::sPersistentStorageOpCertStore;
-Credentials::GroupDataProviderImpl CommonCaseDeviceServerInitParams::sGroupDataProvider;
-app::DefaultTimerDelegate CommonCaseDeviceServerInitParams::sTimerDelegate;
-app::reporting::ReportSchedulerImpl
-    CommonCaseDeviceServerInitParams::sReportScheduler(&CommonCaseDeviceServerInitParams::sTimerDelegate);
+Global<KvsPersistentStorageDelegate> CommonCaseDeviceServerInitParams::sKvsPersistenStorageDelegate;
```
ksperling-apple, line 766: Instead of all these separate globals, you can create a `DefaultCommonCaseDeviceServerInitParams` class inside `Server.cpp` (could be  a sub-class of `CommonCaseDeviceServerInitParams` I think) with an appropriate default constructor, and a single `Global` constructing that. Then reference that global in `InitializeStaticResourcesBeforeServerInit` (which should probably move into the .cpp file) instead of those individual globals.

This also avoids having to add an `Init` method to `ReportScheduler`.

(A simpler existing example of this pattern is `struct ControllerAccessControl`.)

src/app/server/Server.h
ksperling-apple: As noted above, I think we should avoid changing how ReportScheduler is initialized.

src/include/platform/internal/GenericConnectivityManagerImpl_UDP.ipp
```diff
@@ -33,8 +34,8 @@ namespace Internal {
 template <class ImplClass>
 chip::Inet::EndPointManager<Inet::UDPEndPoint> & GenericConnectivityManagerImpl_UDP<ImplClass>::_UDPEndPointManager()
 {
-    static chip::Inet::UDPEndPointManagerImpl sUDPEndPointManagerImpl;
-    return sUDPEndPointManagerImpl;
+    static Global<chip::Inet::UDPEndPointManagerImpl> sUDPEndPointManagerImpl;
```
ksperling-apple, line 37: Please move the Global out of the method, otherwise we end up with the compiler-generated lazy init + whatever Global does.

src/lib/core/CHIPConfig.h
```diff
@@ -187,6 +187,18 @@
...
 
+/**
+ *  @def CHIP_CONFIG_DYNAMICALLY_ALLOCATED_GLOBALS
+ *
+ *  @brief
+ *    Whether to use dynamic allocation for chip::Global objects.
+ *
+ *    The default is to use static allocation.
+ */
+#ifndef CHIP_CONFIG_DYNAMICALLY_ALLOCATED_GLOBALS
```
ksperling-apple, line 198: Can we call rename this to `CHIP_CONFIG_GLOBALS_DYNAMIC`, or something that starts with `CHIP_CONFIG_GLOBALS_` anyway?

This also semantically implies `CHIP_CONFIG_GLOBALS_LAZY_INIT` and/or conflicts with that setting if we take `CHIP_CONFIG_GLOBALS_LAZY_INIT` to mean "static but lazy". Maybe the simplest thing is to just check if `CHIP_CONFIG_GLOBALS_DYNAMIC` and `CHIP_CONFIG_GLOBALS_LAZY_INIT` are both 1 and `#error` if so.

src/lib/core/Global.h
```diff
@@ -97,25 +97,70 @@ class Global
     ~Global()          = default;
 
 private:
+#if CHIP_CONFIG_DYNAMICALLY_ALLOCATED_GLOBALS
+    std::remove_extent_t<T> * mStorage = nullptr;
```
ksperling-apple, line 101: There don't seem to be any strong use cases for `Global` with arrays, so I would prefer to just `static_assert` here that T isn't an array and have the code be simpler.

ksperling-apple (commented):
I think overall this is looking really good, just a few comments.

src/app/reporting/ReportScheduler.h
ksperling-apple: I think the better pattern is to have the Global be of a sub-class or wrapper struct with the default constructor we want.

src/app/MessageDef/MessageDefHelper.cpp
```diff
@@ -37,26 +38,27 @@ namespace app {
 // this is used to run in signle thread for IM message debug purpose
 namespace {
 uint32_t gPrettyPrintingDepthLevel = 0;
-char gLineBuffer[CHIP_CONFIG_LOG_MESSAGE_MAX_SIZE];
+Global<char[CHIP_CONFIG_LOG_MESSAGE_MAX_SIZE]> gLineBuffer;
 size_t gCurLineBufferSize = 0;
```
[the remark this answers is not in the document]
ksperling-apple: This pattern also avoids the extra overhead for the extra `Global`s in the lazy / dynamic cases.

src/app/reporting/ReportSchedulerImpl.h
lpbeliveau-silabs: Which is why the nullptr check should remain in this constructor right? Otherwise the other places in the code using something else than the default constructor just lost their nullptr check.

src/app/server/Server.cpp
ksperling-apple: Thinking about this more, it may be worthwhile to look at some of those individual default implementation objects and see how large they are, to determine if it makes sense to keep them as separate Globals. However the pattern of wrapping or sub-classing the ReportScheduler should still be used to avoid having to change the initialization.

andy31415:
Closing stale PR.

PR #34321, opened 2024-07-12
Fixing crashes when comparing objects to nil

woody-apple:
Nil checks are good :)

### DIFF
--- a/src/darwin/Framework/CHIP/MTRBaseDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRBaseDevice.mm
@@ -283,7 +283,7 @@ public:
         ErrorCallback errorCallback, MTRDeviceResubscriptionScheduledHandler _Nullable resubscriptionScheduledHandler,
         MTRSubscriptionEstablishedHandler _Nullable subscriptionEstablishedHandler, OnDoneHandler _Nullable onDoneHandler)
         : MTRBaseSubscriptionCallback(attributeReportCallback, eventReportCallback, errorCallback, resubscriptionScheduledHandler,
-              subscriptionEstablishedHandler, onDoneHandler)
+            subscriptionEstablishedHandler, onDoneHandler)
     {
     }
 
--- a/src/darwin/Framework/CHIP/MTRBaseDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRBaseDevice.mm
@@ -283,7 +283,7 @@ public:
         ErrorCallback errorCallback, MTRDeviceResubscriptionScheduledHandler _Nullable resubscriptionScheduledHandler,
         MTRSubscriptionEstablishedHandler _Nullable subscriptionEstablishedHandler, OnDoneHandler _Nullable onDoneHandler)
         : MTRBaseSubscriptionCallback(attributeReportCallback, eventReportCallback, errorCallback, resubscriptionScheduledHandler,
-            subscriptionEstablishedHandler, onDoneHandler)
+              subscriptionEstablishedHandler, onDoneHandler)
     {
     }
 
@@ -2290,8 +2290,12 @@ MTREventPriority MTREventPriorityForValidPriorityLevel(chip::app::PriorityLevel 
 
 - (BOOL)isEqualToAttributeRequestPath:(MTRAttributeRequestPath *)path
 {
-    return [_endpoint isEqualToNumber:path.endpoint] && [_cluster isEqualToNumber:path.cluster] &&
-        [_attribute isEqualToNumber:path.attribute];
+    if (!path)
+        return NO;
+
+    return (path.endpoint && [_endpoint isEqualToNumber:path.endpoint])
+        && (path.cluster && [_cluster isEqualToNumber:path.cluster])
+        && (path.attribute && [_attribute isEqualToNumber:path.attribute]);
 }
 
 - (BOOL)isEqual:(id)object
@@ -2362,8 +2366,12 @@ MTREventPriority MTREventPriorityForValidPriorityLevel(chip::app::PriorityLevel 
 
 - (BOOL)isEqualToEventRequestPath:(MTREventRequestPath *)path
 {
-    return
-        [_endpoint isEqualToNumber:path.endpoint] && [_cluster isEqualToNumber:path.cluster] && [_event isEqualToNumber:path.event];
+    if (!path)
+        return NO;
+
+    return (path.endpoint && [_endpoint isEqualToNumber:path.endpoint])
+        && (path.cluster && [_cluster isEqualToNumber:path.cluster])
+        && (path.event && [_event isEqualToNumber:path.event]);
 }
 
 - (BOOL)isEqual:(id)object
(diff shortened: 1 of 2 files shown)


kiel-apple (approved):
weak opinions: 
- `foo != nil &&` is more legible than `foo &&`
- macros could make the combined condition more readable if we don't expect the debugger to land on these lines very frequently

src/darwin/Framework/CHIP/MTRBaseDevice.mm
```diff
@@ -2290,8 +2290,12 @@ + (MTRAttributeRequestPath *)requestPathWithEndpointID:(NSNumber * _Nullable)end
 
 - (BOOL)isEqualToAttributeRequestPath:(MTRAttributeRequestPath *)path
 {
-    return [_endpoint isEqualToNumber:path.endpoint] && [_cluster isEqualToNumber:path.cluster] &&
-        [_attribute isEqualToNumber:path.attribute];
+    if (!path)
+        return NO;
```
bzbarsky-apple, line 2294: I don't get it. This is only called internally, and only when the arg (`path` in this case) had `[object class] == [self class]` test true, which it totally would not if `object` (aka `path`) is nil.

So how could we have nil here?

src/darwin/Framework/CHIP/MTRBaseDevice.mm
```diff
@@ -2290,8 +2290,12 @@ + (MTRAttributeRequestPath *)requestPathWithEndpointID:(NSNumber * _Nullable)end
...
+
+    return (path.endpoint && [_endpoint isEqualToNumber:path.endpoint])
+        && (path.cluster && [_cluster isEqualToNumber:path.cluster])
+        && (path.attribute && [_attribute isEqualToNumber:path.attribute]);
```
bzbarsky-apple, line 2298: These are just wrong.  The old code is wrong, and the new code is wrong....  If all three fields on both `path` and `self` are nil, the correct return value is `YES`, isn't it?

src/darwin/Framework/CHIP/MTRBaseDevice.mm
```diff
@@ -2362,8 +2366,12 @@ + (MTREventRequestPath *)requestPathWithEndpointID:(NSNumber * _Nullable)endpoin
...
 - (BOOL)isEqualToEventRequestPath:(MTREventRequestPath *)path
 {
-    return
-        [_endpoint isEqualToNumber:path.endpoint] && [_cluster isEqualToNumber:path.cluster] && [_event isEqualToNumber:path.event];
+    if (!path)
+        return NO;
+
+    return (path.endpoint && [_endpoint isEqualToNumber:path.endpoint])
+        && (path.cluster && [_cluster isEqualToNumber:path.cluster])
+        && (path.event && [_event isEqualToNumber:path.event]);
```
bzbarsky-apple, line 2374: Same comments as for attribute request paths.

src/darwin/Framework/CHIP/MTRBaseDevice.mm
```diff
@@ -2432,7 +2440,11 @@ ConcreteClusterPath path(static_cast<chip::EndpointId>([endpointID unsignedShort
 
 - (BOOL)isEqualToClusterPath:(MTRClusterPath *)clusterPath
 {
-    return [_endpoint isEqualToNumber:clusterPath.endpoint] && [_cluster isEqualToNumber:clusterPath.cluster];
+    if (!clusterPath)
+        return NO;
```
bzbarsky-apple, line 2444: Again, how can this be nil here?

src/darwin/Framework/CHIP/MTRBaseDevice.mm
```diff
@@ -2432,7 +2440,11 @@ ConcreteClusterPath path(static_cast<chip::EndpointId>([endpointID unsignedShort
...
+
+    return (clusterPath.endpoint && [_endpoint isEqualToNumber:clusterPath.endpoint])
+        && (clusterPath.cluster && [_cluster isEqualToNumber:clusterPath.cluster]);
```
bzbarsky-apple, line 2447: Here, the API claims none of these things are nullable.  How would they be nil?  And any use of these things if they are nil would blow up in Swift, I expect.

If we want to protect against this, we should be failing init with nil to start with.  But again, those things are not supposed to be nullable even there!

src/darwin/Framework/CHIP/MTRBaseDevice.mm
```diff
@@ -2520,7 +2532,10 @@ ConcreteDataAttributePath path(static_cast<chip::EndpointId>([endpointID unsigne
 
 - (BOOL)isEqualToAttributePath:(MTRAttributePath *)attributePath
 {
-    return [self isEqualToClusterPath:attributePath] && [_attribute isEqualToNumber:attributePath.attribute];
+    if (!attributePath)
+        return NO;
+
+    return [self isEqualToClusterPath:attributePath] && attributePath.attribute && [_attribute isEqualToNumber:attributePath.attribute];
```
bzbarsky-apple, line 2538: Same questions as for ConcreteClusterPath.

src/darwin/Framework/CHIP/MTRBaseDevice.mm
```diff
@@ -2613,7 +2628,10 @@ ConcreteEventPath path(static_cast<chip::EndpointId>([endpointID unsignedShortVa
 
 - (BOOL)isEqualToEventPath:(MTREventPath *)eventPath
 {
-    return [self isEqualToClusterPath:eventPath] && [_event isEqualToNumber:eventPath.event];
+    if (!eventPath)
+        return NO;
+
+    return [self isEqualToClusterPath:eventPath] && eventPath.event && [_event isEqualToNumber:eventPath.event];
```
bzbarsky-apple, line 2634: Same questions as for ConcreteClusterPath.

src/darwin/Framework/CHIP/MTRBaseDevice.mm
```diff
@@ -2703,7 +2721,10 @@ ConcreteCommandPath path(static_cast<chip::EndpointId>([endpointID unsignedShort
 
 - (BOOL)isEqualToCommandPath:(MTRCommandPath *)commandPath
 {
-    return [self isEqualToClusterPath:commandPath] && [_command isEqualToNumber:commandPath.command];
+    if (!commandPath)
+        return NO;
+
+    return [self isEqualToClusterPath:commandPath] && commandPath.command && [_command isEqualToNumber:commandPath.command];
```
bzbarsky-apple, line 2727: Same questions as for ConcreteClusterPath.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -336,7 +336,11 @@ - (id)copyWithZone:(NSZone *)zone
 
 - (BOOL)isEqualToClusterData:(MTRDeviceClusterData *)otherClusterData
 {
-    return [_dataVersion isEqual:otherClusterData.dataVersion] && [_attributes isEqual:otherClusterData.attributes];
+    if (!otherClusterData)
+        return NO;
```
bzbarsky-apple, line 340: Again, how can that happen?

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -336,7 +336,11 @@ - (id)copyWithZone:(NSZone *)zone
...
+
+    return (otherClusterData.dataVersion && [_dataVersion isEqual:otherClusterData.dataVersion])
+        && (otherClusterData.attributes && [_attributes isEqual:otherClusterData.attributes]);
```
bzbarsky-apple, line 343: This is again wrong: dataVersion is allowed to be nil, and if it's nil we should be doing the right checks.

On the other hand, attributes is not allowed to be nil, and we should not be worrying about it being nil.  We should fix `initWithCoder:` to guarantee it's not nil, just like `initWithDataVersion:attributes:` does.

PR #34322, opened 2024-07-12
Fix metadata_parser python module to make it wheel buildable.

vatsalghelani-csa:
Work for Issue #31317 

Follow up to PR #34277 

PR 11 - This fixes the metadata python module wheel build. Earlier to this fix we had the issue where if "${chip_root}/scripts/tests/py:metadata_parser.wheel" was added as a dependency and scripts/build_python.sh use to fail with the error: Multiple top-level modules discovered in a flat-layout: ['test_metadata', 'metadata'].
Followed pigweed documentation: https://pigweed.dev/python_build.html#building-python-wheels


BUILD.gn
```diff
@@ -106,6 +106,7 @@ if (current_toolchain != "${dir_pw_toolchain}/default:default") {
     if (chip_enable_python_modules) {
       deps = [
         "${chip_root}/scripts:matter_yamltests_distribution.wheel",
+        "${chip_root}/scripts/tests/py:metadata_parser.wheel",
```
cecille, line 109: question, possibly for @andreilitvin - why do we have a single script that's a separate wheel file?

vatsalghelani-csa: @andy31415

scripts/tests/run_python_test.py
```diff
@@ -32,7 +32,7 @@
 import click
 import coloredlogs
 from colorama import Fore, Style
-from py.metadata import Metadata, MetadataReader
```
andy31415, line 35: we have to get rid of the py prefix.

since we are building the metadata as a module properly now, please update build_python to also install this module and then something like `metadata_parser.metadata` should be a fully valid module.

Since we assume this will work in a general python environment, please also name it something like `matter_testing_support.metadata_parser` or `chip_testing_support.metadata_parser` or similar. That would make it distinct and unique enough to work well with others.


I would like to read code like:

```
from matter_testing_infrastructure.metadata_parser import Metadata, MetadataReader
```

BUILD.gn
andy31415: I think the intent should be to pull in all testing support/infra files separately, so I imagine we want eventually something that contains matter_testing_support and global_attribute_ids and conformance_support and such (and we really should remove the `support` smurf suffix from them when we move).

For this I proposed to have a different name for things as a result: no more `py` in the module name (but maybe it should be in a folder name since pigweed seems to do it often (like https://github.com/google/pigweed/tree/main/pw_bloat or https://github.com/google/pigweed/tree/main/pw_watch) ... then again they prepare for different folder structures.

Overall I would like to be able to `import matter_testing_infrastructure.global_attribute_ids` instead of the current `import global_attribute_ids`.

scripts/tests/run_python_test.py
andy31415: naming wise: I am trying to avoid support/helper/core/base/util and similar filler words, however realistically this is actually a grab-all package ... so maybe some form will be appropriate.

For now I like in order:
   - `matter_integration_tests`
   - `matter_testing_infrastructure`
   - `matter_testing`
   - `matter_test_support`
   - `matter_tests`
   
Happy with any reasonable name though.

andy31415 (approved):

BUILD.gn
cecille: So is the plan to move the python testing framework files into the scripts directory? 

I'd much prefer to leave them where they are (maybe in a sub directory, but at least close). Developers know where to find these files. If you want the metadata parser to be part of the testing wheel, it should be moved to be with the testing stuff and not the other way around.

But honestly, I don't think this really feels like the same thing - the metadata parser is just a CI script, whereas the testing stuff feels like a cohesive module.

andy31415: we can place them in testing as well. They are indeed separated enough purpose-wise, however it also seems we want to minimize the number of wheels overall. So maybe separate packages in the same wheel?

andy31415 (approved):

BUILD.gn
andy31415: marking resolved: the root path was now moved.

PR #34327, opened 2024-07-14
[Fabric-Bridge] Refactor platform independent files out of linux folder

yufengwangca:
Refactor platform independent files out of linux folder


examples/fabric-bridge-app/linux/BUILD.gn
```diff
@@ -33,10 +33,11 @@ if (bridge_enable_pw_rpc) {
 executable("fabric-bridge-app") {
   sources = [
     "${chip_root}/examples/fabric-bridge-app/fabric-bridge-common/include/CHIPProjectAppConfig.h",
-    "Device.cpp",
-    "DeviceManager.cpp",
-    "include/Device.h",
-    "include/DeviceManager.h",
+    "${chip_root}/examples/fabric-bridge-app/fabric-bridge-common/include/Device.h",
```
joonhaengHeo, line 36: Wouldn't it be better to create a BUILD.gn file for "fabric-bridge-common" similar to the code below?
https://github.com/project-chip/connectedhomeip/blob/master/examples/lighting-app/lighting-common/BUILD.gn

yufengwangca: Acked

joonhaengHeo (approved):

examples/fabric-bridge-app/fabric-bridge-common/src/ZCLCallbacks.cpp
```diff
@@ -0,0 +1,92 @@
...
+    {
+        using namespace app::Clusters::BridgedDeviceBasicInformation::Attributes;
+        ChipLogProgress(NotSpecified, "HandleReadBridgedDeviceBasicAttribute: attrId=%d, maxReadLength=%d", attributeId,
+                        maxReadLength);
+
+        if ((attributeId == Reachable::Id) && (maxReadLength == 1))
+        {
+            *buffer = dev->IsReachable() ? 1 : 0;
+        }
+        else if ((attributeId == NodeLabel::Id) && (maxReadLength == 32))
```
tehampson, line 51: Why is this `maxReadLength == 1` and not `maxReadLength >= 1`

examples/fabric-bridge-app/fabric-bridge-common/src/ZCLCallbacks.cpp
```diff
@@ -0,0 +1,92 @@
...
+                                                                          uint8_t * buffer)
+{
+    uint16_t endpointIndex                  = emberAfGetDynamicIndexFromEndpoint(endpoint);
+    Protocols::InteractionModel::Status ret = Protocols::InteractionModel::Status::Failure;
+
+    Device * dev = DeviceMgr().GetDevice(endpointIndex);
+    if (dev != nullptr && dev->IsReachable())
+    {
+        ChipLogProgress(NotSpecified, "emberAfExternalAttributeWriteCallback: ep=%d, clusterId=%d", endpoint, clusterId);
+        ret = Protocols::InteractionModel::Status::Success;
```
tehampson, line 88: How is this a success? You are not writing anything here? Is this even a writeable attribure, should we give an access error here?

I think at this point you are always expected to be done so you can return the value here instead of setting ret otherwise later on you might end up with different unexpected execution after this

tehampson (approved):

examples/fabric-bridge-app/fabric-bridge-common/src/ZCLCallbacks.cpp
yufengwangca: because 'Reachable' type is bool, it must be always 1 byte 

        <attribute side="server" code="17" define="REACHABLE"                 type="boolean"

examples/fabric-bridge-app/fabric-bridge-common/src/ZCLCallbacks.cpp
yufengwangca: Some endpoints of the bridge represent the bridged device. Therefore, all reads or writes to the attributes must be handled within the emberAfExternalAttributeWriteCallback and emberAfExternalAttributeReadCallback functions declared herein. This fits the typical model of a bridge, as a bridge typically maintains its own state database representing the devices connected to it.

If the bridge receives a write operation targeting one of its bridged devices and the device is reachable, we should intercept this operation here. Otherwise, the write operation will go to the bridge itself at that endpoint.

Currently, we just print out a message and do nothing. After all FS clusters have landed, we might need to forward the write operation to the bridged device.

PR #34331, opened 2024-07-15
ThreadOperationalDataset: various bug fixes

ksperling-apple:
- Ensure TLVs read have the correct length
- Default construct as empty (mLength == 0)
- Change MakeRoom() size argument to size_t to avoid chance of truncation
- Check for null in SetNetworkName
- Use Encoding::BigEndian instead of hand-rolled math where possible
- Use ReturnErrorOnFailure / VerifyOrReturnError consistently
- Make tests independent from each other (non-static dataset member)
- Add more tests

### DIFF
--- a/src/lib/support/ThreadOperationalDataset.cpp
+++ b/src/lib/support/ThreadOperationalDataset.cpp
@@ -15,10 +15,12 @@
  *    limitations under the License.
  */
 
-#include <assert.h>
-#include <string.h>
-
 #include <lib/support/ThreadOperationalDataset.h>
+
+#include <lib/core/CHIPEncoding.h>
+
+#include <cassert>
+#include <cstring>
 
 namespace chip {
 namespace Thread {
--- a/src/lib/support/ThreadOperationalDataset.cpp
+++ b/src/lib/support/ThreadOperationalDataset.cpp
@@ -70,7 +70,7 @@ public:
         mLength = aLength;
     }
 
-    const void * GetValue() const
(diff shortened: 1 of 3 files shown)


jmartinez-silabs (approved):

tcarmelveilleux (approved):

src/lib/support/ThreadOperationalDataset.cpp
```diff
@@ -247,26 +210,19 @@ CHIP_ERROR OperationalDataset::SetActiveTimestamp(uint64_t aActiveTimestamp)
...
-        aChannel              = static_cast<uint16_t>((value[1] << 8) | value[2]);
-        return CHIP_NO_ERROR;
-    }
-
-    return CHIP_ERROR_TLV_TAG_NOT_FOUND;
+    VerifyOrReturnError(tlv != nullptr, CHIP_ERROR_TLV_TAG_NOT_FOUND);
+    VerifyOrReturnError(tlv->GetLength() == 3, CHIP_ERROR_INVALID_TLV_ELEMENT);
+    // Note: The channel page (byte 0) is not returned
+    const uint8_t * value = tlv->GetValue();
+    aChannel              = static_cast<uint16_t>((value[1] << 8) | value[2]);
```
bzbarsky-apple, line 217: Could just BigEndian::Get16 starting at the right byte?

ksperling-apple: ah yes, I'll include that change in my other PR.

PR #34335, opened 2024-07-15
[Silabs] MDNS and matter packet filter on WiFi Devices

chirag-silabs:
Adding the filter on the MDNS and matter packet filter for the wifi device when device is sleepy.
Tested on SiWx917 SoC

### DIFF
--- a/src/platform/silabs/ConnectivityManagerImpl_WIFI.cpp
+++ b/src/platform/silabs/ConnectivityManagerImpl_WIFI.cpp
@@ -40,8 +40,8 @@
 #endif
 
 #include "CHIPDevicePlatformConfig.h"
-#include <platform/silabs/wifi/WifiInterfaceAbstraction.h>
 #include "EndpointQueueFilter.h"
+#include <platform/silabs/wifi/WifiInterfaceAbstraction.h>
 
 using namespace ::chip;
 using namespace ::chip::Inet;
--- a/src/platform/silabs/ConnectivityManagerImpl_WIFI.cpp
+++ b/src/platform/silabs/ConnectivityManagerImpl_WIFI.cpp
@@ -53,7 +53,6 @@ namespace DeviceLayer {
 
 ConnectivityManagerImpl ConnectivityManagerImpl::sInstance;
 
-
 CHIP_ERROR ConnectivityManagerImpl::_Init()
 {
     CHIP_ERROR err;
--- a/src/platform/silabs/ConnectivityManagerImpl_WIFI.cpp
+++ b/src/platform/silabs/ConnectivityManagerImpl_WIFI.cpp
@@ -431,7 +431,13 @@ void ConnectivityManagerImpl::UpdateInternetConnectivityState(void)
         {
             sl_wfx_mac_address_t macaddr;
             wfx_get_wifi_mac_addr(SL_WFX_STA_INTERFACE, &macaddr);
-            if (mEndpointQueueFilter.SetHostName(ByteSpan(macaddr.octet)) == CHIP_NO_ERROR)
+            char macaddrString[13]; // 12 characters + null terminator
+            sprintf(macaddrString,"%02X%02X%02X%02X%02X%02X", macaddr.octet[0], macaddr.octet[1], macaddr.octet[2], macaddr.octet[3], macaddr.octet[4], macaddr.octet[5]);
+            EndpointQueueFilterConfig config;
+            config.allowedQueuedPackets = 20; // Set the desired value
+
+            mEndpointQueueFilter.SetConfig(config);
+            if (mEndpointQueueFilter.SetHostName(chip::CharSpan(macaddrString)) == CHIP_NO_ERROR)
             {
                 chip::Inet::UDPEndPointImpl::SetQueueFilter(&mEndpointQueueFilter);
             } else {
--- a/src/platform/silabs/ConnectivityManagerImpl_WIFI.cpp
+++ b/src/platform/silabs/ConnectivityManagerImpl_WIFI.cpp
@@ -428,12 +428,14 @@ void ConnectivityManagerImpl::UpdateInternetConnectivityState(void)
         event.InternetConnectivityChange.IPv6      = GetConnectivityChange(hadIPv6Conn, haveIPv6Conn);
         event.InternetConnectivityChange.ipAddress = addr;
 
+#if CHIP_CONFIG_ENABLE_ICD_SERVER
         {
             sl_wfx_mac_address_t macaddr;
             wfx_get_wifi_mac_addr(SL_WFX_STA_INTERFACE, &macaddr);
             mEndpointQueueFilter.SetMacAddr(ByteSpan(macaddr.octet));
             chip::Inet::UDPEndPointImpl::SetQueueFilter(&mEndpointQueueFilter);
         }
+#endif // CHIP_CONFIG_ENABLE_ICD_SERVER
 
         (void) PlatformMgr().PostEvent(&event);
 
--- a/src/platform/silabs/ConnectivityManagerImpl_WIFI.cpp
+++ b/src/platform/silabs/ConnectivityManagerImpl_WIFI.cpp
@@ -438,9 +438,13 @@ void ConnectivityManagerImpl::UpdateInternetConnectivityState(void)
             config.allowedQueuedPackets = 20; // Set the desired value
 
             mEndpointQueueFilter.SetConfig(config);
-            if (mEndpointQueueFilter.SetHostName(chip::CharSpan(macaddrString)) == CHIP_NO_ERROR)
+
+            // Convert macaddrString to Span<const unsigned char>
+            Span<const unsigned char> macaddrSpan(reinterpret_cast<const unsigned char *>(macaddrString), sizeof(macaddrString));
+
+            if (mEndpointQueueFilter.SetHostName(macaddrSpan) == CHIP_NO_ERROR)
             {
-                chip::Inet::UDPEndPointImpl::SetQueueFilter(&mEndpointQueueFilter);
+                UDPEndPointImpl::SetQueueFilter(&mEndpointQueueFilter);
             }
             else
             {
@@ -448,8 +452,6 @@ void ConnectivityManagerImpl::UpdateInternetConnectivityState(void)
             }
         }
 #endif // CHIP_CONFIG_ENABLE_ICD_SERVER
-
-        (void) PlatformMgr().PostEvent(&event);
 
         if (haveIPv4Conn != hadIPv4Conn)
         {
--- a/src/platform/silabs/ConnectivityManagerImpl_WIFI.cpp
+++ b/src/platform/silabs/ConnectivityManagerImpl_WIFI.cpp
@@ -431,8 +431,12 @@ void ConnectivityManagerImpl::UpdateInternetConnectivityState(void)
         {
             sl_wfx_mac_address_t macaddr;
             wfx_get_wifi_mac_addr(SL_WFX_STA_INTERFACE, &macaddr);
-            mEndpointQueueFilter.SetMacAddr(ByteSpan(macaddr.octet));
-            chip::Inet::UDPEndPointImpl::SetQueueFilter(&mEndpointQueueFilter);
+            if (mEndpointQueueFilter.SetHostName(ByteSpan(macaddr.octet)) == CHIP_NO_ERROR)
+            {
+                chip::Inet::UDPEndPointImpl::SetQueueFilter(&mEndpointQueueFilter);
+            } else {
+                ChipLogError(DeviceLayer, "Failed to set host name filter");
+            } 
         }
 #endif // CHIP_CONFIG_ENABLE_ICD_SERVER
 
--- a/src/platform/silabs/ConnectivityManagerImpl_WIFI.cpp
+++ b/src/platform/silabs/ConnectivityManagerImpl_WIFI.cpp
@@ -432,7 +432,8 @@ void ConnectivityManagerImpl::UpdateInternetConnectivityState(void)
             sl_wfx_mac_address_t macaddr;
             wfx_get_wifi_mac_addr(SL_WFX_STA_INTERFACE, &macaddr);
             char macaddrString[13]; // 12 characters + null terminator
-            sprintf(macaddrString,"%02X%02X%02X%02X%02X%02X", macaddr.octet[0], macaddr.octet[1], macaddr.octet[2], macaddr.octet[3], macaddr.octet[4], macaddr.octet[5]);
+            sprintf(macaddrString, "%02X%02X%02X%02X%02X%02X", macaddr.octet[0], macaddr.octet[1], macaddr.octet[2],
+                    macaddr.octet[3], macaddr.octet[4], macaddr.octet[5]);
             EndpointQueueFilterConfig config;
             config.allowedQueuedPackets = 20; // Set the desired value
 
@@ -440,9 +441,11 @@ void ConnectivityManagerImpl::UpdateInternetConnectivityState(void)
             if (mEndpointQueueFilter.SetHostName(chip::CharSpan(macaddrString)) == CHIP_NO_ERROR)
             {
                 chip::Inet::UDPEndPointImpl::SetQueueFilter(&mEndpointQueueFilter);
-            } else {
+            }
+            else
+            {
                 ChipLogError(DeviceLayer, "Failed to set host name filter");
-            } 
+            }
         }
 #endif // CHIP_CONFIG_ENABLE_ICD_SERVER
 
--- a/src/platform/silabs/ConnectivityManagerImpl_WIFI.cpp
+++ b/src/platform/silabs/ConnectivityManagerImpl_WIFI.cpp
@@ -41,6 +41,7 @@
 
 #include "CHIPDevicePlatformConfig.h"
 #include <platform/silabs/wifi/WifiInterfaceAbstraction.h>
+#include "EndpointQueueFilter.h"
 
 using namespace ::chip;
 using namespace ::chip::Inet;
@@ -51,6 +52,7 @@ namespace chip {
 namespace DeviceLayer {
 
 ConnectivityManagerImpl ConnectivityManagerImpl::sInstance;
+
 
 CHIP_ERROR ConnectivityManagerImpl::_Init()
 {
@@ -425,6 +427,15 @@ void ConnectivityManagerImpl::UpdateInternetConnectivityState(void)
         event.InternetConnectivityChange.IPv4      = GetConnectivityChange(hadIPv4Conn, haveIPv4Conn);
         event.InternetConnectivityChange.IPv6      = GetConnectivityChange(hadIPv6Conn, haveIPv6Conn);
         event.InternetConnectivityChange.ipAddress = addr;
+
+        {
+            sl_wfx_mac_address_t macaddr;
+            wfx_get_wifi_mac_addr(SL_WFX_STA_INTERFACE, &macaddr);
+            mEndpointQueueFilter.SetMacAddr(ByteSpan(macaddr.octet));
+            chip::Inet::UDPEndPointImpl::SetQueueFilter(&mEndpointQueueFilter);
+        }
+
+        (void) PlatformMgr().PostEvent(&event);
 
         if (haveIPv4Conn != hadIPv4Conn)
         {
--- a/src/platform/silabs/EndpointQueueFilter.cpp
+++ b/src/platform/silabs/EndpointQueueFilter.cpp
@@ -1,0 +1,143 @@
+#include "EndpointQueueFilter.h"
+#include <support/CodeUtils.h>
+#include <support/logging/CHIPLogging.h>
+#include <string.h>
+
+namespace chip {
+namespace Inet {
+
+using FilterOutcome     = EndpointQueueFilter::FilterOutcome;
+using namespace chip::Logging;
+
+namespace {
+
+bool IsValidMdnsHostName(const chip::CharSpan & hostName)
+{
+    for (size_t i = 0; i < hostName.size(); ++i)
+    {
+        char ch_data = *(hostName.data() + i);
+        if (!((ch_data >= '0' && ch_data <= '9') || (ch_data >= 'A' && ch_data <= 'F')))
+        {
+            return false;
+        }
+    }
+    return true;
+}
+
+bool PayloadContains(const chip::System::PacketBufferHandle & payload, const chip::ByteSpan & byteSpan)
+{
+    if (payload->HasChainedBuffer() || payload->TotalLength() < byteSpan.size())
+    {
+        return false;
+    }
+    for (size_t i = 0; i <= payload->TotalLength() - byteSpan.size(); ++i)
+    {
+        if (memcmp(payload->Start() + i, byteSpan.data(), byteSpan.size()) == 0)
+        {
+            return true;
+        }
+    }
+    return false;
+}
+
+bool PayloadContainsCaseInsensitive(const chip::System::PacketBufferHandle & payload, const ByteSpan & name)
+{
+    ReturnErrorCodeIf(name.size() < HostNameFilter::kHostNameLengthMax, false);
+
+    uint8_t lower_case[HostNameFilter::kHostNameLengthMax];
+    memcpy(lower_case, name.data(), name.size());
+    for (size_t i = 0; i < sizeof(lower_case); ++i)
+    {
+        if (lower_case[i] <= 'F' && lower_case[i] >= 'A')
+        {
+            lower_case[i] = static_cast<uint8_t>('a' + lower_case[i] - 'A');
+        }
+    }
+    return PayloadContains(payload, name) || PayloadContains(payload, ByteSpan(lower_case));
+}
+
+} // namespace
+
+
+FilterOutcome MdnsBroadcastFilter::Filter(const void * endpoint, const IPPacketInfo & pktInfo,
+                                    const chip::System::PacketBufferHandle & pktPayload)
+{
+    if (pktInfo.DestPort == kMdnsPort)
+    {
+#if INET_CONFIG_ENABLE_IPV4
+        ip_addr_t mdnsIPv4BroadcastAddr = IPADDR4_INIT_BYTES(224, 0, 0, 251);
+        if (pktInfo.DestAddress == chip::Inet::IPAddress(mdnsIPv4BroadcastAddr))
+        {
+            return FilterOutcome::kAllowPacket;
+        }
+#endif
+        ip_addr_t mdnsIPv6BroadcastAddr = IPADDR6_INIT_HOST(0xFF020000, 0, 0, 0xFB);
+        if (pktInfo.DestAddress == chip::Inet::IPAddress(mdnsIPv6BroadcastAddr))
+        {
+            return FilterOutcome::kAllowPacket;
+        }
+    }
+    return FilterOutcome::kDropPacket;
+}
+
+
+FilterOutcome HostNameFilter::Filter(const void * endpoint, const IPPacketInfo & pktInfo,
+                                    const chip::System::PacketBufferHandle & pktPayload)
+{
+    // Drop the mDNS packets which don't contain 'matter' or '<device-hostname>'.
+    const uint8_t matterBytes[] = { 'm', 'a', 't', 't', 'e', 'r' };
+    if (PayloadContains(pktPayload, ByteSpan(matterBytes)) || PayloadContainsCaseInsensitive(pktPayload, ByteSpan(mHostName)))
+    {
+        return FilterOutcome::kAllowPacket;
+    }
+    return FilterOutcome::kDropPacket;
+}
+
+
+CHIP_ERROR HostNameFilter::SetHostName(const chip::CharSpan & name)
+{
+    ReturnErrorCodeIf(name.size() != sizeof(mHostName), CHIP_ERROR_INVALID_ARGUMENT);
+    ReturnErrorCodeIf(!IsValidMdnsHostName(name), CHIP_ERROR_INVALID_ARGUMENT);
+    memcpy(mHostName, name.data(), name.size());
+    return CHIP_NO_ERROR;
+}
+
+
+CHIP_ERROR HostNameFilter::SetMacAddr(const chip::ByteSpan & mac_addr)
+{
+    ReturnErrorCodeIf(mac_addr.size() < 6, CHIP_ERROR_INVALID_ARGUMENT);
+    const uint8_t *p = mac_addr.data();
+    snprintf((char*)mHostName, sizeof(mHostName), "%02x%02x%02x%02x%02x%02x", p[0], p[1], p[2], p[3], p[4], p[5]);
+    return CHIP_NO_ERROR;
+}
+
+
+namespace Silabs {
+
+EndpointQueueFilter::EndpointQueueFilter() :
+    mTooManyFilter(kDefaultAllowedQueuedPackets) {}
+
+EndpointQueueFilter::EndpointQueueFilter(size_t maxAllowedQueuedPackets) :
+    mTooManyFilter(maxAllowedQueuedPackets) {}
+
+
+FilterOutcome EndpointQueueFilter::FilterBeforeEnqueue(const void * endpoint, const IPPacketInfo & pktInfo,
+                                  const chip::System::PacketBufferHandle & pktPayload)
+{
+    ReturnErrorCodeIf(FilterOutcome::kDropPacket == mTooManyFilter.FilterBeforeEnqueue(endpoint, pktInfo, pktPayload), FilterOutcome::kDropPacket);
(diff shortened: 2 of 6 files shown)


src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,141 @@
...
+        {
+            return true;
+        }
+    }
+    return false;
+}
+
+bool PayloadContainsCaseInsensitive(const chip::System::PacketBufferHandle & payload, const ByteSpan & name)
+{
+    ReturnErrorCodeIf(name.size() < HostNameFilter::kHostNameLengthMax, false);
```
tcarmelveilleux, line 45: This condition is wrong and will prevent finding `name` if it's 48 bits instead of 64. Please revisit this

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,141 @@
...
+    uint8_t lower_case[HostNameFilter::kHostNameLengthMax];
+    memcpy(lower_case, name.data(), name.size());
+    for (size_t i = 0; i < sizeof(lower_case); ++i)
+    {
+        if (lower_case[i] <= 'F' && lower_case[i] >= 'A')
+        {
+            lower_case[i] = static_cast<uint8_t>('a' + lower_case[i] - 'A');
+        }
+    }
+    return PayloadContains(payload, name) || PayloadContains(payload, ByteSpan(lower_case));
```
tcarmelveilleux, line 55: This is incorrect --> it will fail to deal with mixed case. Suggest having a single loop of non-case-sensitive search, rather than these odd copies

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+} // namespace
+
+FilterOutcome MdnsBroadcastFilter::Filter(const void * endpoint, const IPPacketInfo & pktInfo,
+                                          const chip::System::PacketBufferHandle & pktPayload)
+{
+    if (pktInfo.DestPort == kMdnsPort)
+    {
+#if INET_CONFIG_ENABLE_IPV4
+        ip_addr_t mdnsIPv4BroadcastAddr = IPADDR4_INIT_BYTES(224, 0, 0, 251);
```
tcarmelveilleux, line 66: Don't use ip_addr_t directly. Use `IPAddress` constructors.

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,141 @@
...
+    if (PayloadContains(pktPayload, ByteSpan(matterBytes)) || PayloadContainsCaseInsensitive(pktPayload, ByteSpan(mHostName)))
+    {
+        return FilterOutcome::kAllowPacket;
+    }
+    return FilterOutcome::kDropPacket;
+}
+
+CHIP_ERROR HostNameFilter::SetHostName(const chip::CharSpan & name)
+{
+    ReturnErrorCodeIf(name.size() != sizeof(mHostName), CHIP_ERROR_INVALID_ARGUMENT);
```
tcarmelveilleux, line 95: Suggest avoiding equality here if there could be slightly different length. See my other comment.

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,141 @@
...
+    ReturnErrorCodeIf(!IsValidMdnsHostName(name), CHIP_ERROR_INVALID_ARGUMENT);
+    memcpy(mHostName, name.data(), name.size());
+    return CHIP_NO_ERROR;
+}
+
+CHIP_ERROR HostNameFilter::SetMacAddr(const chip::ByteSpan & mac_addr)
+{
+    ReturnErrorCodeIf(mac_addr.size() < 6, CHIP_ERROR_INVALID_ARGUMENT);
```
tcarmelveilleux, line 103: If you expect 6, then check for == 6.

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,141 @@
...
+    const uint8_t * p = mac_addr.data();
+    snprintf((char *) mHostName, sizeof(mHostName), "%02x%02x%02x%02x%02x%02x", p[0], p[1], p[2], p[3], p[4], p[5]);
```
tcarmelveilleux, line 105: Please comment that you are setting to lowercase hex because you are doing a lower-case-normalized check whenever mHostName is  looked-at.

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,141 @@
...
+                        FilterOutcome::kDropPacket);
+    if (FilterOutcome::kAllowPacket == mMdnsFilter.Filter(endpoint, pktInfo, pktPayload))
+    {
+        return FilterOutcome::kAllowPacket;
+    }
+    if (FilterOutcome::kAllowPacket == mHostNameFilter.Filter(endpoint, pktInfo, pktPayload))
+    {
+        return FilterOutcome::kAllowPacket;
+    }
+    return FilterOutcome::kAllowPacket;
```
tcarmelveilleux, line 128: This will never drop anything.

The exact order of checks has to be:

```
    if (FilterOutcome::kAllowPacket != mMdnsFilter.Filter(endpoint, pktInfo, pktPayload))
    {
        return FilterOutcome::kDropPacket;
    }
    if (FilterOutcome::kAllowPacket != mHostNameFilter.Filter(endpoint, pktInfo, pktPayload))
    {
        return FilterOutcome::kDropPacket;
    }
    return FilterOutcome::kAllowPacket;
```

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,141 @@
...
+}
+
+FilterOutcome EndpointQueueFilter::FilterAfterDequeue(const void * endpoint, const IPPacketInfo & pktInfo,
+                                                      const chip::System::PacketBufferHandle & pktPayload)
+{
+    ReturnErrorCodeIf(FilterOutcome::kDropPacket == mTooManyFilter.FilterAfterDequeue(endpoint, pktInfo, pktPayload),
```
tcarmelveilleux, line 134: This can be replaced by a `return  mTooManyFilter.FilterAfterDequeue(..............)`

src/platform/silabs/EndpointQueueFilter.h
```diff
@@ -0,0 +1,80 @@
...
+private:
+    uint8_t mHostName[kHostNameLengthMax] = { 0 };
+};
+
+namespace SilabsEndpointQueueFilter {
+
+class EndpointQueueFilter : public chip::Inet::EndpointQueueFilter
+{
+public:
+    static constexpr size_t kDefaultAllowedQueuedPackets = 10;
```
tcarmelveilleux, line 59: This is not a default, it's a forever value! Please make this configurable. 10 is quite small.

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+namespace chip {
+namespace Inet {
+
+using FilterOutcome = EndpointQueueFilter::FilterOutcome;
+using namespace chip::Logging;
+
+namespace {
+
+bool IsValidMdnsHostName(const chip::CharSpan & hostName)
```
bzbarsky-apple, line 14:
```suggestion
bool IsValidMdnsHostName(const CharSpan & hostName)
```

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,141 @@
...
+{
+    for (size_t i = 0; i < hostName.size(); ++i)
+    {
+        char ch_data = *(hostName.data() + i);
```
bzbarsky-apple, line 18:
```suggestion
    for (auto ch_data : hostName)
    {
```

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,141 @@
...
+        if (!((ch_data >= '0' && ch_data <= '9') || (ch_data >= 'A' && ch_data <= 'F')))
+        {
+            return false;
+        }
+    }
+    return true;
+}
+
+bool PayloadContains(const chip::System::PacketBufferHandle & payload, const chip::ByteSpan & byteSpan)
```
bzbarsky-apple, line 27:
```suggestion
bool PayloadContains(const System::PacketBufferHandle & payload, const ByteSpan & byteSpan)
```

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,141 @@
...
+    {
+        if (memcmp(payload->Start() + i, byteSpan.data(), byteSpan.size()) == 0)
+        {
+            return true;
+        }
+    }
+    return false;
+}
+
+bool PayloadContainsCaseInsensitive(const chip::System::PacketBufferHandle & payload, const ByteSpan & name)
```
bzbarsky-apple, line 43: What does "case insensitive" even mean for _bytes_?   This must be getting used on some fairly narrow range of input values, so:

1. It should probably take a CharsSpan, not a ByteSpan.
2. It should document its assumptions (e.g. that the input is an ASCII string?).

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,141 @@
...
+            lower_case[i] = static_cast<uint8_t>('a' + lower_case[i] - 'A');
+        }
+    }
+    return PayloadContains(payload, name) || PayloadContains(payload, ByteSpan(lower_case));
+}
+
+} // namespace
+
+FilterOutcome MdnsBroadcastFilter::Filter(const void * endpoint, const IPPacketInfo & pktInfo,
+                                          const chip::System::PacketBufferHandle & pktPayload)
```
bzbarsky-apple, line 61:
```suggestion
                                          const System::PacketBufferHandle & pktPayload)
```

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,141 @@
...
+        if (pktInfo.DestAddress == chip::Inet::IPAddress(mdnsIPv6BroadcastAddr))
+        {
+            return FilterOutcome::kAllowPacket;
+        }
+    }
+    return FilterOutcome::kDropPacket;
+}
+
+FilterOutcome HostNameFilter::Filter(const void * endpoint, const IPPacketInfo & pktInfo,
+                                     const chip::System::PacketBufferHandle & pktPayload)
```
bzbarsky-apple, line 82:
```suggestion
                                     const System::PacketBufferHandle & pktPayload)
```

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,141 @@
...
+    // Drop the mDNS packets which don't contain 'matter' or '<device-hostname>'.
+    const uint8_t matterBytes[] = { 'm', 'a', 't', 't', 'e', 'r' };
+    if (PayloadContains(pktPayload, ByteSpan(matterBytes)) || PayloadContainsCaseInsensitive(pktPayload, ByteSpan(mHostName)))
+    {
+        return FilterOutcome::kAllowPacket;
+    }
+    return FilterOutcome::kDropPacket;
+}
+
+CHIP_ERROR HostNameFilter::SetHostName(const chip::CharSpan & name)
```
bzbarsky-apple, line 93:
```suggestion
CHIP_ERROR HostNameFilter::SetHostName(const CharSpan & name)
```

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,141 @@
...
+{
+    ReturnErrorCodeIf(name.size() != sizeof(mHostName), CHIP_ERROR_INVALID_ARGUMENT);
+    ReturnErrorCodeIf(!IsValidMdnsHostName(name), CHIP_ERROR_INVALID_ARGUMENT);
```
bzbarsky-apple, line 96: This enforces uppercase.  Is that expected?

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,141 @@
...
+    memcpy(mHostName, name.data(), name.size());
+    return CHIP_NO_ERROR;
+}
+
+CHIP_ERROR HostNameFilter::SetMacAddr(const chip::ByteSpan & mac_addr)
```
bzbarsky-apple, line 101:
```suggestion
CHIP_ERROR HostNameFilter::SetMacAddr(const ByteSpan & mac_addr)
```

src/platform/silabs/EndpointQueueFilter.cpp
bzbarsky-apple: But actual hostnames are stored in mHostName as uppercase, not lowercase, right?

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,141 @@
...
+}
+
+namespace SilabsEndpointQueueFilter {
+
+EndpointQueueFilter::EndpointQueueFilter() : mTooManyFilter(kDefaultAllowedQueuedPackets) {}
+
+EndpointQueueFilter::EndpointQueueFilter(size_t maxAllowedQueuedPackets) : mTooManyFilter(maxAllowedQueuedPackets) {}
+
+FilterOutcome EndpointQueueFilter::FilterBeforeEnqueue(const void * endpoint, const IPPacketInfo & pktInfo,
+                                                       const chip::System::PacketBufferHandle & pktPayload)
```
bzbarsky-apple, line 116:
```suggestion
                                                       const System::PacketBufferHandle & pktPayload)
```

src/platform/silabs/EndpointQueueFilter.h
```diff
@@ -0,0 +1,80 @@
...
+#include <inet/BasicPacketFilters.h>
+
+namespace chip {
+namespace Inet {
+
+struct EndpointFilter
+{
+    virtual ~EndpointFilter()                                                                              = default;
+    virtual EndpointQueueFilter::FilterOutcome Filter(const void * endpoint, const IPPacketInfo & pktInfo,
+                                                      const chip::System::PacketBufferHandle & pktPayload) = 0;
```
bzbarsky-apple, line 29:
```suggestion
                                                      const System::PacketBufferHandle & pktPayload) = 0;
```

src/platform/silabs/EndpointQueueFilter.h
```diff
@@ -0,0 +1,80 @@
...
+};
+
+struct MdnsBroadcastFilter : EndpointFilter
+{
+    static constexpr size_t kMdnsPort = 5353;
+
+    EndpointQueueFilter::FilterOutcome Filter(const void * endpoint, const IPPacketInfo & pktInfo,
+                                              const chip::System::PacketBufferHandle & pktPayload) override;
```
bzbarsky-apple, line 37:
```suggestion
                                              const System::PacketBufferHandle & pktPayload) override;
```

src/platform/silabs/EndpointQueueFilter.h
```diff
@@ -0,0 +1,80 @@
...
+};
+
+struct HostNameFilter : EndpointFilter
+{
+    static constexpr size_t kHostNameLengthMax = 13;
+
+    EndpointQueueFilter::FilterOutcome Filter(const void * endpoint, const IPPacketInfo & pktInfo,
+                                              const chip::System::PacketBufferHandle & pktPayload) override;
```
bzbarsky-apple, line 45:
```suggestion
                                              const System::PacketBufferHandle & pktPayload) override;
```

src/platform/silabs/EndpointQueueFilter.h
```diff
@@ -0,0 +1,80 @@
...
+
+    CHIP_ERROR SetHostName(const chip::CharSpan & name);
+    CHIP_ERROR SetMacAddr(const chip::ByteSpan & addr);
```
bzbarsky-apple, line 48:
```suggestion
    CHIP_ERROR SetHostName(const CharSpan & name);
    CHIP_ERROR SetMacAddr(const ByteSpan & addr);
```

src/platform/silabs/EndpointQueueFilter.h
```diff
@@ -0,0 +1,80 @@
...
+
+private:
+    uint8_t mHostName[kHostNameLengthMax] = { 0 };
+};
+
+namespace SilabsEndpointQueueFilter {
+
+class EndpointQueueFilter : public chip::Inet::EndpointQueueFilter
```
bzbarsky-apple, line 56:
```suggestion
class EndpointQueueFilter : public Inet::EndpointQueueFilter
```

src/platform/silabs/EndpointQueueFilter.h
```diff
@@ -0,0 +1,80 @@
...
+{
+public:
+    static constexpr size_t kDefaultAllowedQueuedPackets = 10;
+
+    EndpointQueueFilter();
+    EndpointQueueFilter(size_t maxAllowedQueuedPackets);
+
+    FilterOutcome FilterBeforeEnqueue(const void * endpoint, const IPPacketInfo & pktInfo,
+                                      const chip::System::PacketBufferHandle & pktPayload) override;
```
bzbarsky-apple, line 65:
```suggestion
                                      const System::PacketBufferHandle & pktPayload) override;
```

src/platform/silabs/EndpointQueueFilter.h
```diff
@@ -0,0 +1,80 @@
...
+
+    FilterOutcome FilterAfterDequeue(const void * endpoint, const IPPacketInfo & pktInfo,
+                                     const chip::System::PacketBufferHandle & pktPayload);
```
bzbarsky-apple, line 68:
```suggestion
                                     const System::PacketBufferHandle & pktPayload);
```

src/platform/silabs/EndpointQueueFilter.h
```diff
@@ -0,0 +1,80 @@
...
+
+    CHIP_ERROR SetMacAddr(const chip::ByteSpan & addr) { return mHostNameFilter.SetMacAddr(addr); }
```
bzbarsky-apple, line 70:
```suggestion
    CHIP_ERROR SetMacAddr(const ByteSpan & addr) { return mHostNameFilter.SetMacAddr(addr); }
```

src/platform/silabs/EndpointQueueFilter.cpp
tcarmelveilleux: Yeah, so this is what is concerning me is the lack of normalization.

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,141 @@
...
+    return true;
+}
+
+bool PayloadContains(const chip::System::PacketBufferHandle & payload, const chip::ByteSpan & byteSpan)
+{
+    if (payload->HasChainedBuffer() || payload->TotalLength() < byteSpan.size())
+    {
+        return false;
+    }
+    for (size_t i = 0; i <= payload->TotalLength() - byteSpan.size(); ++i)
```
andy31415, line 33: Any chance this is the same:

```cpp
std::basic_string_view<uint8_t> needle(byteSpan.data(), byteSpan.size());
std::basic_string_view<uint8_t> haystack(payload->Start(), payload->TotalLength);

return haystack.find(needle) != std::npos;
```

(note: not tested code, however figured maybe shorter/clearer).

src/platform/silabs/EndpointQueueFilter.cpp
chirag-silabs: removed as per discussion

src/platform/silabs/EndpointQueueFilter.cpp
chirag-silabs: Having only 2 search now, "matter" and "hostname"

src/platform/silabs/EndpointQueueFilter.cpp
chirag-silabs: done

src/platform/silabs/EndpointQueueFilter.cpp
chirag-silabs: Changed to save in the uppercase

src/platform/silabs/EndpointQueueFilter.cpp
chirag-silabs: refactored the logic, thanks!

src/platform/silabs/EndpointQueueFilter.h
```diff
@@ -0,0 +1,80 @@
...
+{
+    static constexpr size_t kMdnsPort = 5353;
+
+    EndpointQueueFilter::FilterOutcome Filter(const void * endpoint, const IPPacketInfo & pktInfo,
+                                              const chip::System::PacketBufferHandle & pktPayload) override;
+};
+
+struct HostNameFilter : EndpointFilter
+{
+    static constexpr size_t kHostNameLengthMax = 13;
```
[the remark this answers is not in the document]
chirag-silabs: Done

src/platform/silabs/EndpointQueueFilter.cpp
chirag-silabs: Done

src/platform/silabs/EndpointQueueFilter.cpp
chirag-silabs: Thanks for the suggestion, I have verified it. It works. Thanks!

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,108 @@
...
+#include <string_view>
+
+namespace chip {
+namespace Inet {
+
+using FilterOutcome = EndpointQueueFilter::FilterOutcome;
+
+namespace {
+
+bool IsValidMdnsHostName(uint8_t * hostName)
```
tcarmelveilleux, line 14: using a CharSpan. This will ensure sizing is valid and no assumptions can be broken later.

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,108 @@
...
+    return true;
+}
+
+bool IsMdnsBroadcastPacket(const void * endpoint, const IPPacketInfo & pktInfo, const System::PacketBufferHandle & pktPayload)
+{
+    // if the packet is not a broadcast packet to mDNS port, drop it.
+    VerifyOrReturnValue(pktInfo.DestPort == 5353, false);
+#if INET_CONFIG_ENABLE_IPV4
+    ip_addr_t mdnsIPv4BroadcastAddr = IPADDR4_INIT_BYTES(224, 0, 0, 251);
+    VerifyOrReturnValue(pktInfo.DestAddress == Inet::IPAddress(mdnsIPv4BroadcastAddr), false);
```
tcarmelveilleux, line 33: If IPV4 is enabled, this will always not check if the packet is IPv6, and immeidately return false.

Please layer the changes to avoid VerifyOrReturnValue. This is not a `verify` case.

This should be:

```
if (pktInfo.DestAddress == Inet::IPAddress(mdnsIPv4BroadcastAddr)
{
    return true;
}
```

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,108 @@
...
+#endif
+    ip_addr_t mdnsIPv6BroadcastAddr = IPADDR6_INIT_HOST(0xFF020000, 0, 0, 0xFB);
+    VerifyOrReturnValue(pktInfo.DestAddress == Inet::IPAddress(mdnsIPv6BroadcastAddr), false);
+    return true;
```
tcarmelveilleux, line 37: Default return must be false.

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,108 @@
...
+        return false;
+    }
+
+    if (payload->HasChainedBuffer() || payload->TotalLength() < pattern.size())
+    {
+        return false;
+    }
+
+    std::basic_string_view<uint8_t> payloadView(payload->Start(), payload->TotalLength());
+    std::basic_string_view<uint8_t> patternView(pattern.data(), pattern.size());
```
tcarmelveilleux, line 53: Don't use std::bassic_string_view --> we have CharSpan for this and you use anything to make it case insensitve.

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,108 @@
...
+    return payloadView.find(patternView) != std::basic_string_view<uint8_t>::npos;
```
tcarmelveilleux, line 54: This search is case sensitive, so it will not work to find case insensitively

src/platform/silabs/EndpointQueueFilter.cpp
chirag-silabs: Modified to use CharSpan

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,108 @@
...
+        char ch_data = hostName[i];
+        if (!((ch_data >= '0' && ch_data <= '9') || (ch_data >= 'A' && ch_data <= 'F')))
+        {
+            return false;
+        }
+    }
+    return true;
+}
+
+bool IsMdnsBroadcastPacket(const void * endpoint, const IPPacketInfo & pktInfo, const System::PacketBufferHandle & pktPayload)
```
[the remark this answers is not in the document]
chirag-silabs: done

src/platform/silabs/EndpointQueueFilter.cpp
chirag-silabs: done

src/platform/silabs/EndpointQueueFilter.cpp
chirag-silabs: Modified the default to false

src/platform/silabs/EndpointQueueFilter.cpp
chirag-silabs: Modified the case insensitive search

src/platform/silabs/EndpointQueueFilter.cpp
chirag-silabs: done

src/platform/silabs/ConnectivityManagerImpl_WIFI.cpp
```diff
@@ -426,6 +427,30 @@ void ConnectivityManagerImpl::UpdateInternetConnectivityState(void)
...
+            char macaddrString[13]; // 12 characters + null terminator
+            sprintf(macaddrString, "%02X%02X%02X%02X%02X%02X", macaddr.octet[0], macaddr.octet[1], macaddr.octet[2],
+                    macaddr.octet[3], macaddr.octet[4], macaddr.octet[5]);
+            EndpointQueueFilterConfig config;
+            config.allowedQueuedPackets = 20; // Set the desired value
+
+            mEndpointQueueFilter.SetConfig(config);
+            if (mEndpointQueueFilter.SetHostName(chip::CharSpan(macaddrString)) == CHIP_NO_ERROR)
+            {
+                chip::Inet::UDPEndPointImpl::SetQueueFilter(&mEndpointQueueFilter);
```
bzbarsky-apple, line 443: It's a little confusing why some things like `EndpointQueueFilterConfig` don't get a chip:: prefix while other things do....  None of them need it, given the `using namespace` above.

src/platform/silabs/ConnectivityManagerImpl_WIFI.cpp
```diff
@@ -426,6 +427,30 @@ void ConnectivityManagerImpl::UpdateInternetConnectivityState(void)
...
+            }
+            else
+            {
+                ChipLogError(DeviceLayer, "Failed to set host name filter");
+            }
+        }
+#endif // CHIP_CONFIG_ENABLE_ICD_SERVER
+
+        (void) PlatformMgr().PostEvent(&event);
```
bzbarsky-apple, line 452: Why are we double-posting the event now?

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+using FilterOutcome = EndpointQueueFilter::FilterOutcome;
+
+namespace {
+
+bool IsValidMdnsHostName(const CharSpan & hostName)
+{
+    for (size_t i = 0; i < hostName.size(); ++i)
+    {
+        char ch_data = *(hostName.data());
```
bzbarsky-apple, line 18: Is there a reason not to do:
```suggestion
    for (char ch_data : hostName)
    {
```
?

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,118 @@
...
+    {
+        return false;
+    }
+
+    if (payload->HasChainedBuffer() || payload->TotalLength() < pattern.size())
+    {
+        return false;
+    }
+
+    CharSpan payloadView(reinterpret_cast<const char *>(payload->Start()), payload->TotalLength());
```
bzbarsky-apple, line 58: Please consider using `Uint8::to_char` to make it clear what you are doing.

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,118 @@
...
+    CharSpan patternView(reinterpret_cast<const char *>(pattern.data()), pattern.size());
+
+    auto toLower = [](char c) { return std::tolower(static_cast<unsigned char>(c)); };
+
+    auto it = std::search(payloadView.data(), payloadView.data() + payloadView.size(), patternView.data(),
+                          patternView.data() + patternView.size(), [&](char a, char b) { return toLower(a) == toLower(b); });
```
bzbarsky-apple, line 64: So I have to ask: if we are going to just be passing these to something that wants unsigned char, is there a reason we are not just using `Span<unsigned char>` instead of `CharSpan`?  And for that matter, having pattern be a `Span<unsigned char>` to start with?

src/platform/silabs/ConnectivityManagerImpl_WIFI.cpp
chirag-silabs: done

src/platform/silabs/ConnectivityManagerImpl_WIFI.cpp
chirag-silabs: remove the double post

src/platform/silabs/EndpointQueueFilter.cpp
chirag-silabs: notice the span interation is not working, will have a check once

src/platform/silabs/EndpointQueueFilter.cpp
chirag-silabs: done

src/platform/silabs/EndpointQueueFilter.cpp
chirag-silabs: changed the CharSpan to Span<unsigned char>

src/platform/silabs/EndpointQueueFilter.cpp
bzbarsky-apple: I don't understand this comment...

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    if (payload->HasChainedBuffer() || payload->TotalLength() < pattern.size())
+    {
+        return false;
+    }
+
+    Span<const unsigned char> payloadView(payload->Start(), payload->TotalLength());
+
+    auto toLower = [](unsigned char c) { return std::tolower(static_cast<unsigned char>(c)); };
```
bzbarsky-apple, line 64: Why is this static_cast needed?

src/platform/silabs/EndpointQueueFilter.cpp
```diff
@@ -0,0 +1,121 @@
...
+
+} // namespace
+
+FilterOutcome HostNameFilter::Filter(const void * endpoint, const IPPacketInfo & pktInfo,
+                                     const System::PacketBufferHandle & pktPayload)
+{
+    // Drop the mDNS packets which don't contain 'matter' or '<device-hostname>'.
+    const unsigned char matterBytes[] = { 'm', 'a', 't', 't', 'e', 'r' };
+    if (PayloadContainsCaseInsensitive(pktPayload, Span<const unsigned char>(matterBytes, sizeof(matterBytes))) ||
+        PayloadContainsCaseInsensitive(pktPayload, Span<const unsigned char>(mHostName, sizeof(mHostName))))
```
bzbarsky-apple, line 80: Why do you need the `sizeof()` bits here?  The constructor at https://github.com/project-chip/connectedhomeip/blob/61e288a1968f359a57ebd40b60197976ebec46f7/src/lib/support/Span.h#L64-L68 should do the right thing with arrays of `const unsigned char`.

PR #34342, opened 2024-07-15
Created automated test TC_ADMIN_1_9:

j-ororke:
Follows test steps in test plan PR #[4267](https://github.com/CHIP-Specifications/chip-test-plans/pull/4267):
- TH1 able to open commissioning window on DUT
- TH2 Fails to connect to endpoint for 20 iterations with correct payload but incorrect passcode 
- TH2 Fails to connect to endpoint due to commissioning window being closed after the 20th failed attempt 
- TH1 able to open commissioning window on the endpoint again 
- TH1 able to revoke commissioning window on the endpoint


src/python_testing/TC_CADMIN_1_9.py
```diff
@@ -0,0 +1,113 @@
+#
+#    Copyright (c) 2022 Project CHIP Authors
```
cecille, line 2:
```suggestion
#    Copyright (c) 2024 Project CHIP Authors
```

src/python_testing/TC_CADMIN_1_9.py
```diff
@@ -0,0 +1,113 @@
...
+                logging.info("-----------------Current Iteration {}-------------------------".format(cycle+1))
+                setup_code = self.generate_unique_random_value(params.setupPinCode)
+                errcode = self.th2.CommissionOnNetwork(
+                    nodeId=self.dut_node_id, setupPinCode=setup_code,
+                    filterType=ChipDeviceCtrl.DiscoveryFilterType.LONG_DISCRIMINATOR, filter=self.discriminator)
+                logging.info('Commissioning complete done. Successful? {}, errorcode = {}'.format(errcode.is_success, errcode))
+                asserts.assert_false(errcode.is_success, 'Commissioning complete did not error as expected')
+                asserts.assert_true(errcode.sdk_code == expectedErrCode,
+                                    'Unexpected error code returned from CommissioningComplete')
+                sleep(1)
```
cecille, line 68: Do you actually need this sleep for the test to work? If so, can drop the error logs into the slack channel?

src/python_testing/TC_CADMIN_1_9.py
```diff
@@ -0,0 +1,113 @@
...
+        elif expectedErrCode == 50:
+            logging.info("-----------------Attempting connection expecting timeout-------------------------")
+            errcode = self.th2.CommissionOnNetwork(
+                nodeId=self.dut_node_id, setupPinCode=params.setupPinCode,
+                filterType=ChipDeviceCtrl.DiscoveryFilterType.LONG_DISCRIMINATOR, filter=self.discriminator)
+            logging.info('Commissioning complete done. Successful? {}, errorcode = {}'.format(errcode.is_success, errcode))
+            asserts.assert_false(errcode.is_success, 'Commissioning complete did not error as expected')
+            asserts.assert_true(errcode.sdk_code == expectedErrCode, 'Unexpected error code returned from CommissioningComplete')
+
+    def TC_CADMIN_1_9(self) -> list[str]:
```
cecille, line 79:
```suggestion
    def pics_TC_CADMIN_1_9(self) -> list[str]:
```

src/python_testing/TC_CADMIN_1_9.py
```diff
@@ -0,0 +1,113 @@
...
+        return ["CADMIN.S"]
+
+    @async_test_body
+    async def test_TC_ADMIN_1_9(self):
```
cecille, line 83:
```suggestion
    async def test_TC_CADMIN_1_9(self):
```

src/python_testing/TC_CADMIN_1_9.py
```diff
@@ -0,0 +1,113 @@
...
+        self.print_step(1, "Commissioning, already done")
```
cecille, line 84: Could you please move these into a separate steps_ method? That will let the TH follow along as it goes. See https://project-chip.github.io/connectedhomeip-doc/testing/python.html#test-harness-helpers for an explanation, https://github.com/project-chip/connectedhomeip/blob/master/src/python_testing/TC_TMP_2_1.py for a good example.

src/python_testing/TC_CADMIN_1_9.py
```diff
@@ -0,0 +1,113 @@
...
+from matter_testing_support import MatterBaseTest, async_test_body, default_matter_test_main
+from mobly import asserts
+
+
+class TC_CADMIN_1_9(MatterBaseTest):
+
+    def OpenCommissioningWindow(self) -> CommissioningParameters:
+        try:
+            params = self.th1.OpenCommissioningWindow(
+                nodeid=self.dut_node_id, timeout=900, iteration=10000, discriminator=self.discriminator, option=1)
```
cecille, line 41: oop, sorry, just one more thing - I added a step to the test plan just before it went in to change this from 900 to the max failsafe value. TP is merged now, see step 2. https://github.com/CHIP-Specifications/chip-test-plans/blob/master/src/multiplefabrics.adoc#tc-cadmin-1-9-device-exit-commissioning-mode-after-20-failed-commission-attempts-ecm-dut-commissionee

src/python_testing/TC_CADMIN_1_9.py
```diff
@@ -0,0 +1,113 @@
+#
```
cecille, line 1: You'll also need to add this to tests.yaml (assuming you get this in before Vatsal implements file globbing)

src/python_testing/TC_CADMIN_1_9.py
j-ororke: Hi Cecille,
So, I had found during testing that if I did not add the sleep time duration after the 20th iteration that the 21st iteration appeared to fail similarly to how it does for iterations 1-20, then it would timeout on the 22nd iteration. So, my thought was adding the sleep duration of 1 second allows the endpoint enough time to close the commissioning window before another attempt can be made to it. 
This also means that the 1 second sleep time does not need to be made for every iteration, so I will move this sleep time out of the for loop where it currently is.
Please let me know if this is acceptable.

j-ororke: After discussing this with you, I will create a bug for the issue found above and place a comment in the test code for why the sleep is there.
Will also move the sleep into the main section of the test code and out of this function.

j-ororke: Created issue #34383 regarding failure noticed when using correct passcode during test step 4.

src/python_testing/TC_CADMIN_1_9.py
j-ororke: Hi Cecille,
I have updated the duration for the commissioning window here now mentioned in the test plan for test step 2.

src/python_testing/TC_CADMIN_1_9.py
j-ororke: Hi Cecille,
I have updated tests.yaml and added this test into the repl_tests_linux section.

src/python_testing/TC_CADMIN_1_9.py
j-ororke: Hi Cecille,
I have updated the method for the test steps following the example you provided in TC_TMP_2_1.

cecille (approved):

raju-apple (approved):

KishokG:
@j-ororke Could you please put the PR to remove TC-CADMIN-1.9 YAML script. Here is the issue for the same: https://github.com/project-chip/matter-test-scripts/issues/363

PR #34354, opened 2024-07-16
[Python] Avoid RuntimeException if APIs with future raise an error

agners:
Currently, when calling an API which uses a future causes an error (e.g. CommissionWithCode with an invalid code), then the API call already returns an error. In this case the call `raise_on_error()` on the returned PyChipError object make sure that an exception is raised. However, this also causes the `CallbackContext` context manager to exit.

At this point the future is initialized but never completed, which triggers the previously introduced sanity check in `CallbackContext`: `RuntimeError("CallbackContext future not completed")`.

Remove the RuntimeError as existing the context manager early without completing the future is a use case (when the call setting up the callback raises an exception).

Instead, just cancel the future in the context manager if it hasn't been complete yet, in case someone has a reference to it and expects it to complete.

Also, since most API calls return PyChipError, this changes `CallAsync()` to raise an exception by default instead of returning a PyChipError object. If the PyChipError object is required or an API returns something else, the CallAsyncWithResult() method can be used.

### DIFF
--- a/src/controller/python/chip/ChipDeviceCtrl.py
+++ b/src/controller/python/chip/ChipDeviceCtrl.py
@@ -249,7 +249,10 @@ class CallbackContext:
 
     async def __aexit__(self, exc_type, exc_value, traceback):
         if not self._future.done():
-            raise RuntimeError("CallbackContext future not completed")
+            # In case the initial call (which sets up for the callback) fails,
+            # the future will never be used actually. So just cancel it here
+            # for completeness, in case somebody is expecting it to be completed.
+            self._future.cancel()
         self._future = None
         self._lock.release()
 
@@ -603,11 +606,10 @@ class ChipDeviceControllerBase():
 
         async with self._commissioning_context as ctx:
             self._enablePairingCompleteCallback(True)
-            res = await self._ChipStack.CallAsync(
+            await self._ChipStack.CallAsync(
                 lambda: self._dmLib.pychip_DeviceController_ConnectBLE(
                     self.devCtrl, discriminator, isShortDiscriminator, setupPinCode, nodeid)
             )
-            res.raise_on_error()
 
             return await asyncio.futures.wrap_future(ctx.future)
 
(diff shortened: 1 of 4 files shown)


src/controller/python/chip/ChipDeviceCtrl.py
```diff
@@ -603,23 +603,30 @@ async def ConnectBLE(self, discriminator: int, setupPinCode: int, nodeid: int, i
...
-                lambda: self._dmLib.pychip_DeviceController_ConnectBLE(
-                    self.devCtrl, discriminator, isShortDiscriminator, setupPinCode, nodeid)
-            )
-            res.raise_on_error()
+            try:
+                await self._ChipStack.CallAsync(
+                    lambda: self._dmLib.pychip_DeviceController_ConnectBLE(
+                        self.devCtrl, discriminator, isShortDiscriminator, setupPinCode, nodeid)
+                )
+            except:
```
arkq, line 611: Can you narrow the scope of this catch (in particular not to catch `KeyboardInterrupt`)?
In other places as well.

arkq (approved):

src/controller/python/chip/ChipDeviceCtrl.py
agners: It is not catching any exception in that sense, as we re-raise immediately.

If we would narrow this down to `KeyboardInterrupt`, We'd end up with a `RuntimeError` from the CommissionContext.

Maybe it would be better to just get rid of the RuntimeError there actually :thinking:

agners: Yeah thinking a bit more about it, I think this is the better way:

```patch
     async def __aexit__(self, exc_type, exc_value, traceback):
         if not self._future.done():
-            raise RuntimeError("CallbackContext future not completed")
+            # In case the initial call (which sets up for the callback) fails,
+            # the future will never be used actually. So just cancel it here
+            # for completeness, in case somebody is expecting it to be completed.
+            self._future.cancel()
         self._future = None
         self._lock.release()
```

Then we don't need this `try .. except` everywhere.

agners: Changed, also updated the PR description.

cecille (approved):

Issue #34304, opened 2024-07-11
Implement ActionSwitch feature in all-clusters-ap


tcarmelveilleux:
- Add new named pipe commands to generate sequence of events properly for AS
- Add action switch feature to matter file

PR #34355, opened 2024-07-16
Add an Action Switch endpoint to all-clusters-app

tcarmelveilleux:
- Action Switch is a new feature for a TCR
- Need to have the ZAP files updated before adding the logic to reduce review burden.
- Re-ran zap_convert_all.py (some of the changes are due to that)

Issue #34304

Testing done:

- No regressions of integration tests


bzbarsky-apple (approved):

ksperling-apple:
At least one of the Darwin test failures is due to this
```
/Users/runner/work/connectedhomeip/connectedhomeip/src/darwin/Framework/CHIPTests/MTRDeviceTests.m:2032:
error: -[MTRDeviceTests test020_ReadMultipleAttributes] : ((endpoint == 0 || endpoint == 1 || endpoint == 2) is true) failed
```

All up there are 40 failing test cases, not sure if they all make similar assumptions about what endpoints there are in all-clusters. @bzbarsky-apple ?

bzbarsky-apple:
>  not sure if they all make similar assumptions about what endpoints there are in all-clusters

I don't think they all do.  `MTRDeviceTests test020_ReadMultipleAttributes` and `MTRDeviceTests test021_ReadMultipleWildcardPathsIncludeUnsupportedAttribute` clearly do, but stuff after that seems to be failing because those failed tests fail to clean up properly somehow and then the server is in a bad state.

So I'd fix those two and see what things look like.

ReneJosefsen (approved):

PR #34358, opened 2024-07-16
Fixing wildcard comparisons to correct

woody-apple:
Also updating our class checks to allow for subclassing

### DIFF
--- a/src/darwin/Framework/CHIP/MTRBaseDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRBaseDevice.mm
@@ -283,7 +283,7 @@ public:
         ErrorCallback errorCallback, MTRDeviceResubscriptionScheduledHandler _Nullable resubscriptionScheduledHandler,
         MTRSubscriptionEstablishedHandler _Nullable subscriptionEstablishedHandler, OnDoneHandler _Nullable onDoneHandler)
         : MTRBaseSubscriptionCallback(attributeReportCallback, eventReportCallback, errorCallback, resubscriptionScheduledHandler,
-              subscriptionEstablishedHandler, onDoneHandler)
+            subscriptionEstablishedHandler, onDoneHandler)
     {
     }
 
--- a/src/darwin/Framework/CHIP/MTRBaseDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRBaseDevice.mm
@@ -2306,7 +2306,7 @@ MTREventPriority MTREventPriorityForValidPriorityLevel(chip::app::PriorityLevel 
         return YES;
     }
 
-    return [super isEqual:path];
+    return NO;
 }
 
 - (BOOL)isEqual:(id)object
@@ -2315,7 +2315,7 @@ MTREventPriority MTREventPriorityForValidPriorityLevel(chip::app::PriorityLevel 
         return [self isEqualToAttributeRequestPath:object];
     }
 
-    return [super isEqual:object];
+    return NO;
 }
 
 - (NSUInteger)hash
@@ -2394,7 +2394,7 @@ MTREventPriority MTREventPriorityForValidPriorityLevel(chip::app::PriorityLevel 
         return YES;
     }
 
-    return [super isEqual:path];
+    return NO;
 }
 
 - (BOOL)isEqual:(id)object
@@ -2403,7 +2403,7 @@ MTREventPriority MTREventPriorityForValidPriorityLevel(chip::app::PriorityLevel 
         return [self isEqualToEventRequestPath:object];
     }
 
-    return [super isEqual:object];
+    return NO;
 }
 
 - (NSUInteger)hash
@@ -2482,7 +2482,7 @@ MTREventPriority MTREventPriorityForValidPriorityLevel(chip::app::PriorityLevel 
         return [self isEqualToClusterPath:object];
     }
 
-    return [super isEqual:object];
+    return NO;
 }
 
 - (NSUInteger)hash
@@ -2578,7 +2578,7 @@ static NSString * const sClusterKey = @"clusterKey";
         return [self isEqualToAttributePath:object];
     }
 
-    return [super isEqual:object];
+    return NO;
 }
 
 - (NSUInteger)hash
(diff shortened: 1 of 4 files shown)


src/darwin/Framework/CHIP/MTRBaseDevice.mm
```diff
@@ -2290,20 +2290,32 @@ + (MTRAttributeRequestPath *)requestPathWithEndpointID:(NSNumber * _Nullable)end
...
+    if ((_endpoint == path.endpoint || (path.endpoint && [_endpoint isEqualToNumber:path.endpoint]))
+        && (_cluster == path.cluster || (path.cluster && [_cluster isEqualToNumber:path.cluster]))
+        && (_attribute == path.attribute || (path.attribute && [_attribute isEqualToNumber:path.attribute]))) {
+        return YES;
+    }
 
-    return (path.endpoint && [_endpoint isEqualToNumber:path.endpoint])
-        && (path.cluster && [_cluster isEqualToNumber:path.cluster])
-        && (path.attribute && [_attribute isEqualToNumber:path.attribute]);
+    return [super isEqual:path];
```
bzbarsky-apple, line 2309: Shouldn't this just return NO?  The other thing is an instance of MTRAttributeRequestPath but the members don't match us.  Who cares what `super` has to say?

src/darwin/Framework/CHIP/MTRBaseDevice.mm
```diff
@@ -2290,20 +2290,32 @@ + (MTRAttributeRequestPath *)requestPathWithEndpointID:(NSNumber * _Nullable)end
...
 - (BOOL)isEqual:(id)object
 {
-    if ([object class] != [self class]) {
-        return NO;
+    if ([[object class] isKindOfClass:[MTRAttributeRequestPath class]]) {
+        return [self isEqualToAttributeRequestPath:object];
     }
-    return [self isEqualToAttributeRequestPath:object];
+
+    return [super isEqual:object];
```
bzbarsky-apple, line 2318: Ok.  So if the expectation is that everyone terminates in `[super isEqual:object]` for things that are not `isKindOfClass:[OurClassName class]`, then let's walk through this for a second.  Let's say we have a subclass, `A` of `MTRAttributeRequestPath` and it implements this logic.  `A` has some extra state compared to `MTRAttributeRequestPath` and its `isEqual` methods checks that state when given an instances of `A`.  So the method on `A` looks like this:

```
- (BOOL)isEqual:(id)object
{
    if ([[object class] isKindOfClass:[A class]]) {
        return [self someComparisonMethodThatChecksOurState:object];
    }

    return [super isEqual:path];
}
```

right?  Now say you have `a1` and `a2` that are different instances of `A`, but have the same `MTRAttributeRequestPath` bits, which are also the same as some object `path` which is a non-subclassed `MTRAttributeRequestPath`.  In that case `[a1 isEqual:a2]` will be false, but `[a1 isEqual:path]` is true (because `A` will forward that to the super-class) and `[path isEqual:a2]` is true (because `a2` will test true for `isKindOfClass`), so `isEqual` is not transitive (but does look like it's reflexive and symmetric; the symmetry comes from the super-call at the end).

This leads to weird behavior where, for example, depending on the order in which you add the three objects to an NSSet you get different resulting sets.  And if you use `a1` as a dictionary key for a write to the dictionary, then you can retrieve that entry using `a1` or `path` as a key, but not `a2`.  But if you use `path` as the key for the write, then you can retrieve using any of the three objects as a key.

Basically, the behavior gets _really_ confusing if you do this, as far as I can tell.  Using these objects in sets, as dictionary keys, etc, will sometimes "work" and sometimes "not work" depending on the exact order things happen in...

src/darwin/Framework/CHIP/MTRBaseDevice.mm
woody-apple: That is correct, as we discussed, this is a redundant thanks to the self == path above.

src/darwin/Framework/CHIP/MTRBaseDevice.mm
woody-apple: Discussed offline

woody-apple: Resolving,  but keeping in draft until I resolve one item re: transitive. Confirming with implementations.

src/darwin/Framework/CHIP/MTRBaseDevice.mm
```diff
@@ -2290,20 +2290,32 @@ + (MTRAttributeRequestPath *)requestPathWithEndpointID:(NSNumber * _Nullable)end
...
-        && (path.cluster && [_cluster isEqualToNumber:path.cluster])
-        && (path.attribute && [_attribute isEqualToNumber:path.attribute]);
+    return NO;
 }
 
 - (BOOL)isEqual:(id)object
 {
-    if ([object class] != [self class]) {
-        return NO;
+    if ([[object class] isKindOfClass:[MTRAttributeRequestPath class]]) {
```
ksperling-apple, line 2314: This is just wrong, it leads to incorrectly behaving isEqual when a sub-class is added (and if there are no sub-classes it behaves the same either way).

(Also, implementing isEqual correctly in a way that allows a sub-class instance with additional state to compare equal to a super class instance is way more complicated than just using `isKindOfClass:`. See e.g. http://www.angelikalanger.com/Articles/JavaSolutions/SecretsOfEquals/Equals.html and http://www.angelikalanger.com/Articles/JavaSolutions/SecretsOfEquals/Equals-2.html for a details discussion)

src/darwin/Framework/CHIP/MTRBaseDevice.mm
```diff
@@ -2290,20 +2290,32 @@ + (MTRAttributeRequestPath *)requestPathWithEndpointID:(NSNumber * _Nullable)end
...
         return NO;
+    }
+
+    if (path == self) {
+        return YES;
+    }
+
+    // Note: Nil is a wildcard, and empty endpoint/cluster/attribute values imply nil as well.
+    // Pass iff we have objects that are pointer equivalent (eg: nil == nil || object_a == object_a), or they have values and the same logical values
+    if ((_endpoint == path.endpoint || (path.endpoint && [_endpoint isEqualToNumber:path.endpoint]))
```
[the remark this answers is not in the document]
woody-apple: Thanks, didn't know about this.

src/darwin/Framework/CHIP/MTRBaseDevice.mm
woody-apple: This is not correct, but I'm going to remove this part of the PR, and we'll move on.

PR #34362, opened 2024-07-17
Fixing nil crashes

woody-apple:
Reported nil crashes affecting customers, fixing patterns across Matter.framework.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -133,7 +133,8 @@
         "random": "cpp",
         "thread": "cpp",
         "variant": "cpp",
-        "any": "cpp"
+        "any": "cpp",
+        "future": "cpp"
     },
     // Configure paths or glob patterns to exclude from file watching.
     "files.watcherExclude": {
--- a/src/darwin/Framework/CHIP/MTRBaseDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRBaseDevice.mm
@@ -16,7 +16,6 @@
  */
 #import <Matter/Matter.h>
 
-#import "MTRUtilities.h"
 #import "MTRAttributeTLVValueDecoder_Internal.h"
 #import "MTRBaseDevice_Internal.h"
 #import "MTRBaseSubscriptionCallback.h"
@@ -31,6 +30,7 @@
 #import "MTRLogging_Internal.h"
 #import "MTRMetricKeys.h"
 #import "MTRSetupPayload_Internal.h"
+#import "MTRUtilities.h"
 #import "NSDataSpanConversion.h"
 #import "NSStringSpanConversion.h"
 #import "zap-generated/MTRCommandPayloads_Internal.h"
--- a/src/darwin/Framework/CHIP/MTRBaseDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRBaseDevice.mm
@@ -14,8 +14,7 @@
  *    See the License for the specific language governing permissions and
  *    limitations under the License.
  */
-#import <Matter/MTRClusterConstants.h>
-#import <Matter/MTRDefines.h>
+#import <Matter/Matter.h>
 
 #import "MTRAttributeTLVValueDecoder_Internal.h"
 #import "MTRBaseDevice_Internal.h"
@@ -35,27 +34,27 @@
 #import "NSStringSpanConversion.h"
 #import "zap-generated/MTRCommandPayloads_Internal.h"
 
-#include "app/ConcreteAttributePath.h"
-#include "app/ConcreteCommandPath.h"
-#include "app/ConcreteEventPath.h"
-#include "app/StatusResponse.h"
-#include "lib/core/CHIPError.h"
-#include "lib/core/DataModelTypes.h"
+#import "app/ConcreteAttributePath.h"
+#import "app/ConcreteCommandPath.h"
+#import "app/ConcreteEventPath.h"
+#import "app/StatusResponse.h"
+#import "lib/core/CHIPError.h"
+#import "lib/core/DataModelTypes.h"
 
-#include <app/AttributePathParams.h>
-#include <app/BufferedReadCallback.h>
-#include <app/ClusterStateCache.h>
-#include <app/InteractionModelEngine.h>
-#include <app/ReadClient.h>
-#include <app/data-model/List.h>
-#include <controller/CommissioningWindowOpener.h>
-#include <controller/ReadInteraction.h>
-#include <controller/WriteInteraction.h>
-#include <crypto/CHIPCryptoPAL.h>
-#include <setup_payload/SetupPayload.h>
-#include <system/SystemClock.h>
+#import <app/AttributePathParams.h>
+#import <app/BufferedReadCallback.h>
+#import <app/ClusterStateCache.h>
+#import <app/InteractionModelEngine.h>
+#import <app/ReadClient.h>
+#import <app/data-model/List.h>
+#import <controller/CommissioningWindowOpener.h>
+#import <controller/ReadInteraction.h>
+#import <controller/WriteInteraction.h>
+#import <crypto/CHIPCryptoPAL.h>
+#import <setup_payload/SetupPayload.h>
+#import <system/SystemClock.h>
 
-#include <memory>
+#import <memory>
 
 using namespace chip;
 using namespace chip::app;
@@ -2290,8 +2289,9 @@ MTREventPriority MTREventPriorityForValidPriorityLevel(chip::app::PriorityLevel 
 
 - (BOOL)isEqualToAttributeRequestPath:(MTRAttributeRequestPath *)path
 {
-    return [_endpoint isEqualToNumber:path.endpoint] && [_cluster isEqualToNumber:path.cluster] &&
-        [_attribute isEqualToNumber:path.attribute];
+    return MTREqualObjects(_endpoint, path.endpoint)
+        && MTREqualObjects(_cluster, path.cluster)
+        && MTREqualObjects(_attribute, path.attribute);
 }
 
 - (BOOL)isEqual:(id)object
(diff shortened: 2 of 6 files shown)


.vscode/settings.json
```diff
@@ -133,7 +133,8 @@
         "random": "cpp",
         "thread": "cpp",
         "variant": "cpp",
-        "any": "cpp"
+        "any": "cpp",
+        "future": "cpp"
```
woody-apple, line 137: This is what happens if you open any new file and save it in VSCode, so doing it for folks here

src/darwin/Framework/CHIP/MTRUtilities.mm
```diff
@@ -19,5 +19,5 @@
 
 BOOL MTREqualObjects(id<NSObject> _Nullable a, id<NSObject> _Nullable b)
 {
-    return (a == nil) ? (b == nil) : [a isEqual:b];
+    return (a == nil) ? (b == nil) : (b == nil ? NO : [a isEqual:b]);
```
woody-apple, line 22: This is allowing a check to isEqual: nil, which isn't correct.

src/darwin/Framework/CHIP/MTRUtilities.mm
```diff
@@ -19,5 +19,16 @@
 
 BOOL MTREqualObjects(id<NSObject> _Nullable a, id<NSObject> _Nullable b)
 {
-    return (a == nil) ? (b == nil) : [a isEqual:b];
+    // Check if A is nil, and return based on B being nil (or not)
```
woody-apple, line 22: Making this readable, and fixing where we can compare to nil

anush-apple (approved):

src/darwin/Framework/Matter.xcodeproj/project.pbxproj
```diff
@@ -109,7 +109,7 @@
 		3D0C484B29DA4FA0006D811F /* MTRErrorTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 3D0C484A29DA4FA0006D811F /* MTRErrorTests.m */; };
 		3D3928D72BBCEA3D00CDEBB2 /* MTRAvailabilityTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 3D3928D62BBCEA3D00CDEBB2 /* MTRAvailabilityTests.m */; settings = {COMPILER_FLAGS = "-UMTR_NO_AVAILABILITY -Wno-unguarded-availability-new"; }; };
 		3D4733AF2BDF1B80003DC19B /* MTRSetupPayloadTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 3D4733AE2BDF1B80003DC19B /* MTRSetupPayloadTests.m */; };
-		3D4733B32BE2D1DA003DC19B /* MTRUtilities.h in Headers */ = {isa = PBXBuildFile; fileRef = 3D4733B22BE2D1CF003DC19B /* MTRUtilities.h */; };
+		3D4733B32BE2D1DA003DC19B /* MTRUtilities.h in Headers */ = {isa = PBXBuildFile; fileRef = 3D4733B22BE2D1CF003DC19B /* MTRUtilities.h */; settings = {ATTRIBUTES = (Public, ); }; };
```
woody-apple, line 112: Working on this, our imports are broken. Need to move to xcconfigs, but... hopefully not in this PR to fix the crash.

src/darwin/Framework/CHIP/MTRBaseDevice.mm
```diff
@@ -14,8 +14,7 @@
  *    See the License for the specific language governing permissions and
  *    limitations under the License.
  */
-#import <Matter/MTRClusterConstants.h>
-#import <Matter/MTRDefines.h>
+#import <Matter/Matter.h>
```
woody-apple, line 17: This is also not the correct import, this should be a precompiled header for the project, but this'll patch it over for now.

src/darwin/Framework/CHIP/MTRBaseDevice.mm
```diff
@@ -35,27 +34,27 @@
...
-#include <app/data-model/List.h>
-#include <controller/CommissioningWindowOpener.h>
-#include <controller/ReadInteraction.h>
-#include <controller/WriteInteraction.h>
-#include <crypto/CHIPCryptoPAL.h>
-#include <setup_payload/SetupPayload.h>
-#include <system/SystemClock.h>
-
-#include <memory>
+#import "app/ConcreteAttributePath.h"
```
woody-apple, line 37: These should all be imports, I'll be moving these to xc-configs in a follow on PR.

src/darwin/Framework/CHIP/Matter.h
```diff
@@ -68,5 +68,6 @@
 #import <Matter/MTRStorage.h>
 #import <Matter/MTRStructsObjc.h>
 #import <Matter/MTRThreadOperationalDataset.h>
+#import <Matter/MTRUtilities.h>
```
woody-apple, line 71: We were missing an import, but we were extern-ing it. Adding it. I'll fix this file and others with a mass xcconfig fix.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -41,17 +41,17 @@
 #import "MTRUnfairLock.h"
 #import "zap-generated/MTRCommandPayloads_Internal.h"
 
-#include "lib/core/CHIPError.h"
-#include "lib/core/DataModelTypes.h"
-#include <app/ConcreteAttributePath.h>
-#include <lib/support/FibonacciUtils.h>
+#import "lib/core/CHIPError.h"
```
woody-apple, line 44: These should all be imports, I'll be moving these to xc-configs in a follow on PR.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -15,7 +15,7 @@
  *    limitations under the License.
  */
 
-#import <Matter/MTRDefines.h>
+#import <Matter/Matter.h>
```
woody-apple, line 18: This is the wrong way to import things, I'll fix this in a follow-on PR with general xcconfig and PCH (precompiled headers). This is a quick general patch.

src/darwin/Framework/CHIP/MTRUtilities.mm
bzbarsky-apple, line 22: So here's what confuses me.  The new logic here is equivalent to the old logic, so that's fine as far as it goes.  But why did we need to change this at all?  isEqual is documented as:

```
object

    The object to be compared to the receiver. May be nil, in which case this method returns NO.
```

so if `b` is nil and we reached the `[a isEqual:b]` stuff we would return NO as desired.  The new code is more explicit about it, I guess?

bzbarsky-apple (commented):
So the real fix here is making sure we don't pass nil to `isEqualToNumber:`, right?  The rest is fixing other correctness issues and/or other issues altogether.

Would be good for the commit message to make this clear....

src/darwin/Framework/CHIP/MTRBaseDevice.mm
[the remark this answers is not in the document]
woody-apple: It's still incorrect., I'll clean these up as a touch all the files. I'm cleaning up headers as I touch files, I encourage us all to do the same.

src/darwin/Framework/CHIP/MTRUtilities.mm
woody-apple: Yes it is, and readable.

woody-apple: The old code allowed you to pass nil to [a isEqual: b];

src/darwin/Framework/Matter.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
woody-apple: I think you'll find it is:

```

/**
 * Calls `-[NSObject isEqual:]` but returns YES when comparing nil to nil.
 */
MTR_EXTERN BOOL MTREqualObjects(id<NSObject> _Nullable a, id<NSObject> _Nullable b);
```

That said, if we don't want to make it public, that's fine too.

woody-apple:
> So the real fix here is making sure we don't pass nil to `isEqualToNumber:`, right? The rest is fixing other correctness issues and/or other issues altogether.

The fix here is to fix a crasher, where our comparators compare nil, which crashes.

When seeing this, I fixed this same anti-pattern in files that use it. I'm also cleaning up imports, since the files did not import the utilities function, so builds failed. I moved our includes to be imports, as is correct in the files affected. I mentioned I'll have a followup to fix all of these.

> Would be good for the commit message to make this clear....

I don't think commit messages require review, as this is a squashed commit anyway from my attempts to fix builds as I discovered import breakage (well, lack of imports).

src/darwin/Framework/Matter.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
woody-apple: Will move this to _Internal for now, given we don't need to fix the examples (which have the same issue)

andy31415 (approved):

src/darwin/Framework/Matter.xcodeproj/project.pbxproj
bzbarsky-apple, line 1: Not sure why there's a diff here, but it's not actually changing anything, right?

bzbarsky-apple (approved):

src/darwin/Framework/Matter.xcodeproj/project.pbxproj
woody-apple: No idea why, but re-saving xcode did this on a newer version

PR #34370, opened 2024-07-17
[Python] Only auto re-subscribe after initial subscription

agners:
The subscription logic waits for the first successful subscription before the Read() call is being returned (the future is awaited which is only released on handleSubscriptionEstablished). If the first subscription attempt fails (e.g. because the CASE session doesn't establish) the Read() never returns, not with an error but also not with a subscription transaction. And since the Python side has no access to the SubscriptionTransaction object at this point yet, there is also no way to stop this subscription attempt.

With this change, we only resubscribe if the initial subscription was successful. This changes semantics slightly, but really allows the caller to decide if it wants to continue try to establish the subscription.

### DIFF
--- a/src/controller/python/chip/clusters/attribute.cpp
+++ b/src/controller/python/chip/clusters/attribute.cpp
@@ -145,18 +145,20 @@ public:
 
     void OnSubscriptionEstablished(SubscriptionId aSubscriptionId) override
     {
+        // Only enable auto resubscribe if the subscription is established successfully.
+        mAutoResubscribeNeeded = mAutoResubscribe;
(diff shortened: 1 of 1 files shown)


cecille (approved):

cecille:
Stefan - if you have some time, would you mind adding that description into the function docstring around the autoResubscribe parameter?

agners:
> Stefan - if you have some time, would you mind adding that description into the function docstring around the autoResubscribe parameter?

Yeah good point. I've created a new PR with some other missing parameters: https://github.com/project-chip/connectedhomeip/pull/34380.

tehampson (approved):

arkq (approved):

PR #34371, opened 2024-07-17
External sdk support

rasmus-zyax:
Added a way to depend on an external SDK to allow an external platforms to provide dependencies to the matter libraries the same way as platforms currently included within the source tree

### DIFF
--- a/src/system/BUILD.gn
+++ b/src/system/BUILD.gn
@@ -43,6 +43,8 @@ declare_args() {
   #
   # see https://clang.llvm.org/docs/ThreadSafetyAnalysis.html
   chip_enable_thread_safety_checks = is_clang
+
+  chip_external_sdk_target = ""
 }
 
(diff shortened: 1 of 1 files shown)


src/system/BUILD.gn
```diff
@@ -43,6 +43,8 @@ declare_args() {
   #
   # see https://clang.llvm.org/docs/ThreadSafetyAnalysis.html
   chip_enable_thread_safety_checks = is_clang
+
+  chip_external_sdk_target = ""
```
andy31415, line 47: please add documentation on how this is indended to be used.

Why don't we default this to the `${efr32_sdk_build_root}:efr32_sdk"` or other values depending on the platform and then let others override this? This seems to be more consistent than adding a completely separate flag.

andy31415: I would propose a `chip_platform_sdk_target` path that gets initialized like the existing deps, but allows overriding.

PR #34375, opened 2024-07-17
Implements the Commissioner Control Cluster in Matter SDK

### LINKED ISSUE #33222: [Feature] Implements the Commissioner Control Cluster in Matter SDK
### Feature description

...
Implements https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/9089

### Platform

all

### Platform Version(s)

_No response_

### Anything else?

_No response_

yufengwangca:
Implements https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/9089

This PR contains the implementation of the Commissioner Control Cluster server itself. The actual usage of this cluster across multiple example applications (Fabric Admin and Fabric Bridge), will be addressed in a follow-up PR.

Fixes #33222

### DIFF
--- a/examples/fabric-bridge-app/fabric-bridge-common/fabric-bridge-app.matter
+++ b/examples/fabric-bridge-app/fabric-bridge-common/fabric-bridge-app.matter
@@ -1334,6 +1334,57 @@ cluster GroupKeyManagement = 63 {
   fabric command access(invoke: administer) KeySetReadAllIndices(): KeySetReadAllIndicesResponse = 4;
 }
 
+/** Supports the ability for clients to request the commissioning of themselves or other nodes onto a fabric which the cluster server can commission onto. */
+provisional cluster CommissionerControl = 1873 {
+  revision 1;
+
+  bitmap SupportedDeviceCategoryBitmap : bitmap32 {
+    kFabricSynchronization = 0x1;
+  }
+
+  fabric_sensitive info event access(read: manage) CommissioningRequestResult = 0 {
+    int64u requestId = 0;
+    node_id clientNodeId = 1;
+    enum8 statusCode = 2;
+    fabric_idx fabricIndex = 254;
+  }
+
+  readonly attribute access(read: manage) SupportedDeviceCategoryBitmap supportedDeviceCategories = 0;
+  readonly attribute command_id generatedCommandList[] = 65528;
+  readonly attribute command_id acceptedCommandList[] = 65529;
+  readonly attribute event_id eventList[] = 65530;
+  readonly attribute attrib_id attributeList[] = 65531;
+  readonly attribute bitmap32 featureMap = 65532;
+  readonly attribute int16u clusterRevision = 65533;
+
+  request struct RequestCommissioningApprovalRequest {
+    int64u requestId = 0;
+    vendor_id vendorId = 1;
+    int16u productId = 2;
+    optional char_string<64> label = 3;
+  }
+
+  request struct CommissionNodeRequest {
+    int64u requestId = 0;
+    int16u responseTimeoutSeconds = 1;
+    optional octet_string ipAddress = 2;
+    optional int16u port = 3;
+  }
+
+  response struct ReverseOpenCommissioningWindow = 2 {
+    int16u commissioningTimeout = 0;
+    octet_string PAKEPasscodeVerifier = 1;
+    int16u discriminator = 2;
+    int32u iterations = 3;
+    octet_string<32> salt = 4;
+  }
+
+  /** This command is sent by a client to request approval for a future CommissionNode call. */
+  command access(invoke: manage) RequestCommissioningApproval(RequestCommissioningApprovalRequest): DefaultSuccess = 0;
+  /** This command is sent by a client to request that the server begins commissioning a previously approved request. */
+  command access(invoke: manage) CommissionNode(CommissionNodeRequest): ReverseOpenCommissioningWindow = 1;
+}
+
 endpoint 0 {
   device type ma_rootdevice = 22, version 1;
 
@@ -1646,6 +1697,21 @@ endpoint 0 {
     handle command KeySetRemove;
     handle command KeySetReadAllIndices;
     handle command KeySetReadAllIndicesResponse;
+  }
+
+  server cluster CommissionerControl {
+    emits event CommissioningRequestResult;
+    ram      attribute supportedDeviceCategories default = 0;
+    callback attribute generatedCommandList;
+    callback attribute acceptedCommandList;
+    callback attribute eventList;
+    callback attribute attributeList;
+    ram      attribute featureMap default = 0;
+    ram      attribute clusterRevision default = 1;
+
+    handle command RequestCommissioningApproval;
+    handle command CommissionNode;
+    handle command ReverseOpenCommissioningWindow;
   }
 }
 endpoint 1 {
--- a/examples/fabric-bridge-app/fabric-bridge-common/src/ZCLCallbacks.cpp
+++ b/examples/fabric-bridge-app/fabric-bridge-common/src/ZCLCallbacks.cpp
@@ -20,7 +20,6 @@
 #include <app-common/zap-generated/cluster-objects.h>
 #include <app-common/zap-generated/ids/Attributes.h>
 #include <app-common/zap-generated/ids/Clusters.h>
-#include <app/clusters/commissioner-control-server/commissioner-control-server.h>
 #include <lib/support/ZclString.h>
 
 using namespace ::chip;
@@ -90,14 +89,4 @@ Protocols::InteractionModel::Status emberAfExternalAttributeWriteCallback(Endpoi
     }
 
     return ret;
-}
-
-void MatterCommissionerControlPluginServerInitCallback()
-{
-    ChipLogProgress(Zcl, "MatterCommissionerControlPluginServerInitCallback");
-
-    BitMask<CommissionerControl::SupportedDeviceCategoryBitmap> supportedDeviceCategories;
-    supportedDeviceCategories.SetField(CommissionerControl::SupportedDeviceCategoryBitmap::kFabricSynchronization, 1);
-    CommissionerControl::CommissionerControlServer::Instance().SetSupportedDeviceCategoriesValue(kRootEndpointId,
-                                                                                                 supportedDeviceCategories);
 }
--- a/examples/fabric-bridge-app/fabric-bridge-common/src/ZCLCallbacks.cpp
+++ b/examples/fabric-bridge-app/fabric-bridge-common/src/ZCLCallbacks.cpp
@@ -20,6 +20,7 @@
 #include <app-common/zap-generated/cluster-objects.h>
 #include <app-common/zap-generated/ids/Attributes.h>
 #include <app-common/zap-generated/ids/Clusters.h>
+#include <app/clusters/commissioner-control-server/commissioner-control-server.h>
 #include <lib/support/ZclString.h>
 
 using namespace ::chip;
@@ -89,4 +90,14 @@ Protocols::InteractionModel::Status emberAfExternalAttributeWriteCallback(Endpoi
     }
 
     return ret;
+}
+
+void MatterCommissionerControlPluginServerInitCallback()
+{
+    ChipLogProgress(Zcl, "MatterCommissionerControlPluginServerInitCallback");
+
+    BitMask<CommissionerControl::SupportedDeviceCategoryBitmap> supportedDeviceCategories;
+    supportedDeviceCategories.SetField(CommissionerControl::SupportedDeviceCategoryBitmap::kFabricSynchronization, 1);
+    CommissionerControl::CommissionerControlServer::Instance().SetSupportedDeviceCategoriesValue(kRootEndpointId,
+                                                                                                 supportedDeviceCategories);
 }
--- a/src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
+++ b/src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
@@ -38,14 +38,6 @@ using chip::Protocols::InteractionModel::Status;
 
 namespace {
 
-NodeId getNodeId(const app::CommandHandler * commandObj)
-{
-    VerifyOrDie(commandObj);
-    auto descriptor = commandObj->GetSubjectDescriptor();
-
-    return descriptor.subject;
-}
-
 void AddReverseOpenCommissioningWindowResponse(CommandHandler * commandObj, const ConcreteCommandPath & path,
                                                const Clusters::CommissionerControl::CommissioningWindowParams & params)
 {
@@ -128,7 +120,8 @@ CommissionerControlServer::SetSupportedDeviceCategoriesValue(EndpointId endpoint
     return status;
 }
 
-CHIP_ERROR CommissionerControlServer::GenerateCommissioningRequestResultEvent(const Events::CommissioningRequestResult::Type & result)
+CHIP_ERROR
+CommissionerControlServer::GenerateCommissioningRequestResultEvent(const Events::CommissioningRequestResult::Type & result)
 {
     EventNumber eventNumber;
     CHIP_ERROR error = LogEvent(result, kRootEndpointId, eventNumber);
@@ -154,12 +147,16 @@ bool emberAfCommissionerControlClusterRequestCommissioningApprovalCallback(
 
     ChipLogProgress(Zcl, "Received command to request commissioning approval");
 
-    auto sourceNodeId = getNodeId(commandObj);
+    auto descriptor   = commandObj->GetSubjectDescriptor();
+    auto sourceNodeId = descriptor.subject;
     auto fabricIndex  = commandObj->GetAccessingFabricIndex();
     auto requestId    = commandData.requestId;
     auto vendorId     = commandData.vendorId;
     auto productId    = commandData.productId;
-    auto & label      = commandData.label;
+
+    // The label assigned from commandData need to be stored in CommissionerControl::Delegate which ensure that the backing buffer
+    // of it has a valid lifespan during fabric sync setup process.
+    auto & label = commandData.label;
 
     // Create a CommissioningApprovalRequest struct and populate it with the command data
     Clusters::CommissionerControl::CommissioningApprovalRequest request = { .requestId    = requestId,
@@ -197,7 +194,8 @@ bool emberAfCommissionerControlClusterCommissionNodeCallback(
 
     ChipLogProgress(Zcl, "Received command to commission node");
 
-    auto sourceNodeId = getNodeId(commandObj);
+    auto descriptor   = commandObj->GetSubjectDescriptor();
+    auto sourceNodeId = descriptor.subject;
     auto requestId    = commandData.requestId;
 
     auto info       = std::make_unique<Clusters::CommissionerControl::CommissionNodeInfo>();
@@ -210,7 +208,9 @@ bool emberAfCommissionerControlClusterCommissionNodeCallback(
     VerifyOrExit(sourceNodeId != kUndefinedNodeId, err = CHIP_ERROR_WRONG_NODE_ID);
     VerifyOrExit(delegate != nullptr, err = CHIP_ERROR_INCORRECT_STATE);
 
-    // Handle commissioning approval request
+    // Handle commissioning approval request, the ipAddress assigned from commandData need to be stored in
+    // CommissionerControl::Delegate which ensure that the backing buffer of ipAddress has a valid lifespan until the deferred task
+    // is executed.
     err = delegate->ValidateCommissionNodeCommand(sourceNodeId, requestId, info->params);
     SuccessOrExit(err == CHIP_NO_ERROR);
 
@@ -228,4 +228,9 @@ exit:
     }
 
     return true;
+}
+
+void MatterCommissionerControlPluginServerInitCallback()
+{
+    ChipLogProgress(Zcl, "Initiating Commissioner Control cluster.");
 }
--- a/src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
+++ b/src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
@@ -1,0 +1,251 @@
+/*
+ *
+ *    Copyright (c) 2024 Project CHIP Authors
+ *    All rights reserved.
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+#include "commissioner-control-server.h"
+
+#include <protocols/interaction_model/StatusCode.h>
+
+#include <app-common/zap-generated/attributes/Accessors.h>
+#include <app-common/zap-generated/cluster-enums.h>
+#include <app-common/zap-generated/cluster-objects.h>
+#include <app/AttributeAccessInterface.h>
+#include <app/AttributeAccessInterfaceRegistry.h>
+#include <app/CommandHandler.h>
+#include <app/EventLogging.h>
+#include <app/reporting/reporting.h>
+#include <platform/PlatformManager.h>
+
+#include <memory> // For std::shared_ptr
+
+using namespace chip;
+using namespace chip::app;
+
+using chip::Protocols::InteractionModel::Status;
+
+namespace {
+
+NodeId getNodeId(const app::CommandHandler * commandObj)
+{
+    if (nullptr == commandObj || nullptr == commandObj->GetExchangeContext())
+    {
+        ChipLogError(Zcl, "Cannot access ExchangeContext of Command Object for Node ID");
+        return kUndefinedNodeId;
+    }
+
+    if (!commandObj->GetExchangeContext()->HasSessionHandle())
+    {
+        ChipLogError(Zcl, "Cannot access session of Command Object for Node ID");
+        return kUndefinedNodeId;
+    }
+
+    auto descriptor = commandObj->GetExchangeContext()->GetSessionHandle()->GetSubjectDescriptor();
+    if (descriptor.authMode != Access::AuthMode::kCase)
+    {
+        ChipLogError(Zcl, "Cannot get Node ID from non-CASE session of Command Object");
+        return kUndefinedNodeId;
+    }
+
+    return descriptor.subject;
+}
+
+void AddReverseOpenCommissioningWindowResponse(CommandHandler * commandObj, const ConcreteCommandPath & path,
+                                               const Clusters::CommissionerControl::CommissioningWindowParams & params)
+{
+    Clusters::CommissionerControl::Commands::ReverseOpenCommissioningWindow::Type response;
+    response.commissioningTimeout = params.commissioningTimeout;
+    response.discriminator        = params.discriminator;
+    response.iterations           = params.iterations;
+    response.PAKEPasscodeVerifier = params.PAKEPasscodeVerifier;
+    response.salt                 = params.salt;
+
+    commandObj->AddResponse(path, response);
+}
+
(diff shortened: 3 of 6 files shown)


src/app/clusters/commissioner-control-server/commissioner-control-server.h
```diff
@@ -0,0 +1,99 @@
...
+    uint16_t discriminator;
+    ByteSpan PAKEPasscodeVerifier;
+    ByteSpan salt;
+};
+
+struct CommissionNodeInfo
+{
+    CommissioningWindowParams params;
+    Optional<ByteSpan> ipAddress;
+    Optional<uint16_t> port;
```
tcarmelveilleux, line 50: When do we ever commission specifically by IP address? How is this going to be used?

src/app/clusters/commissioner-control-server/commissioner-control-server.h
```diff
@@ -0,0 +1,99 @@
...
+namespace CommissionerControl {
+
+struct CommissioningApprovalRequest
+{
+    uint64_t requestId;
+    VendorId vendorId;
+    uint16_t productId;
+    NodeId clientNodeId;
+    FabricIndex fabricIndex;
+    Optional<chip::CharSpan> label;
```
tcarmelveilleux, line 34: WHere is the ownership of this data? Suggest making the struct own a char array large enough to be free-standing and owning for the label span to point to.

yufengwangca: The ownership of this data is the application, who implement the delegate interface 

    virtual CHIP_ERROR HandleCommissioningApprovalRequest(const CommissioningApprovalRequest & request) = 0;
    
 Because this data is only used in the application, not within the Commissioner Control Cluster server

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
```diff
@@ -0,0 +1,251 @@
...
+        ChipLogError(Zcl, "Cannot access session of Command Object for Node ID");
+        return kUndefinedNodeId;
+    }
+
+    auto descriptor = commandObj->GetExchangeContext()->GetSessionHandle()->GetSubjectDescriptor();
+    if (descriptor.authMode != Access::AuthMode::kCase)
+    {
+        ChipLogError(Zcl, "Cannot get Node ID from non-CASE session of Command Object");
+        return kUndefinedNodeId;
+    }
```
tehampson, line 60: Why not just do:

if (commandObj == nullptr || commandObj->GetSubjectDescriptor())

Or is your concern that you will crash on various VerifyOrDie in there

src/app/clusters/commissioner-control-server/commissioner-control-server.h
yufengwangca: If the client provides the ipAddress and port, the application can optionally choose to use 'pairing already-discovered' instead of 'pairing code' to reverse pair the device. But this is up to the application, this just provide the flexibility.

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
yufengwangca: I copy this logic from the existing code, it looks we need to make sure we have a valid secure session before we extra the node id from that.

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
```diff
@@ -0,0 +1,251 @@
...
+#include <app-common/zap-generated/cluster-enums.h>
+#include <app-common/zap-generated/cluster-objects.h>
+#include <app/AttributeAccessInterface.h>
+#include <app/AttributeAccessInterfaceRegistry.h>
+#include <app/CommandHandler.h>
+#include <app/EventLogging.h>
+#include <app/reporting/reporting.h>
+#include <platform/PlatformManager.h>
+
+#include <memory> // For std::shared_ptr
```
tehampson, line 32: Do you actually use shared_ptr?

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
tehampson: Yes but you should always have a secure session when this is being called. If you don't something has gone seriously wrong

tehampson: actually I think you can reduce this code further and do:

```
VerifyOrDie(commandObj);
auto descriptor = || commandObj->GetSubjectDescriptor();
```

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+    Clusters::CommissionerControl::Delegate * delegate =
+        Clusters::CommissionerControl::CommissionerControlServer::Instance().GetDelegate();
+
+    VerifyOrExit(sourceNodeId != kUndefinedNodeId, err = CHIP_ERROR_WRONG_NODE_ID);
+    VerifyOrExit(delegate != nullptr, err = CHIP_ERROR_INCORRECT_STATE);
+
+    // Handle commissioning approval request
+    err = delegate->ValidateCommissionNodeCommand(sourceNodeId, requestId, info->params);
+    VerifyOrExit(err == CHIP_NO_ERROR, err = CHIP_ERROR_INCORRECT_STATE);
```
tehampson, line 232: nit, what is the purpose of setting the error code to `err = CHIP_ERROR_INCORRECT_STATE`? I don't see it used 
```suggestion
    VerifyOrExit(err == CHIP_NO_ERROR);
```

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
tehampson: Oh it is for unique_ptr

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
```diff
@@ -0,0 +1,251 @@
...
+    return true;
+
+exit:
+    if (err != CHIP_NO_ERROR)
+    {
+        ChipLogError(Zcl, "emberAfCommissionerControlClusterCommissionNodeCallback error: %s", err.AsString());
+        status = Status::Failure;
+    }
+
+    commandObj->AddStatus(commandPath, status);
```
tehampson, line 249: You have already added a response in `AddReverseOpenCommissioningWindowResponse` in some instance here.

I think you only want to call AddStatus in the failure case (if statement above)

tehampson: Oh...

I see what happened you have return true before the exit here... Um I thought the exit para-dime coding practice was that all code needed to end up in the exit even the success case.

Right now only error code ends up in the exit section. Right?

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
```diff
@@ -0,0 +1,251 @@
...
+    info->ipAddress = commandData.ipAddress;
+    info->port      = commandData.port;
+
+    Clusters::CommissionerControl::Delegate * delegate =
+        Clusters::CommissionerControl::CommissionerControlServer::Instance().GetDelegate();
+
+    VerifyOrExit(sourceNodeId != kUndefinedNodeId, err = CHIP_ERROR_WRONG_NODE_ID);
+    VerifyOrExit(delegate != nullptr, err = CHIP_ERROR_INCORRECT_STATE);
+
+    // Handle commissioning approval request
```
tehampson, line 230: Should some common validation be done here to help the application such as 
```
The server SHALL return FAILURE if the CommissionNode command is not sent from the same NodeId as the RequestCommissioningApproval

File: src/app/clusters/commissioner-control-server/commissioner-control-server.cpp [hunk 8]
```diff
@@ -0,0 +1,251 @@
...
+    err = delegate->ValidateCommissionNodeCommand(sourceNodeId, requestId, info->params);
+    VerifyOrExit(err == CHIP_NO_ERROR, err = CHIP_ERROR_INCORRECT_STATE);
+
+    // Add the response for the commissioning window
+    AddReverseOpenCommissioningWindowResponse(commandObj, commandPath, info->params);
```
tehampson, line 235: Is the result ready at this point. I don't think it is you are sending a response before you have the result because you are scheduling the work. I think you need to do to respond to this asyncronously

yufengwangca: The client only send the  CommissionNode command after it received CommissioningRequestResult event with StatusCode of SUCCESS
  
```
    // Add the response for the commissioning window
    AddReverseOpenCommissioningWindowResponse(commandObj, commandPath, info->params);
```
    
This adds the ReverseOpenCommissioningWindow response synchronously, but the bridge won't send this response until return from the calling function.

To make sure the bridge send ReverseCommissionNode after ReverseOpenCommissioningWindow response, we need to use ScheduleWork to schedule this task after this function returns.

examples/fabric-bridge-app/fabric-bridge-common/src/ZCLCallbacks.cpp
```diff
@@ -90,3 +91,13 @@ Protocols::InteractionModel::Status emberAfExternalAttributeWriteCallback(Endpoi
 
     return ret;
 }
+
+void MatterCommissionerControlPluginServerInitCallback()
```
bzbarsky-apple, line 95: This callback is meant to be implemented by the SDK itself, to do whatever it needs to do for non-endpoint-specific cluster backend init, not applications.

src/app/clusters/commissioner-control-server/commissioner-control-server.h
bzbarsky-apple, line 34:
```suggestion
    Optional<CharSpan> label;
```

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
```diff
@@ -0,0 +1,231 @@
...
+    if ((status = Attributes::SupportedDeviceCategories::Set(endpoint, supportedDeviceCategories)) != Status::Success)
+    {
+        ChipLogProgress(Zcl, "CommissionerControl: writing supportedDeviceCategories, err:0x%x", to_underlying(status));
+        return status;
+    }
+
+    return status;
+}
+
+CHIP_ERROR CommissionerControlServer::EmitCommissioningRequestResultEvent(const Events::CommissioningRequestResult::Type & result)
```
bzbarsky-apple, line 131:
```suggestion
CHIP_ERROR CommissionerControlServer::GenerateCommissioningRequestResultEvent(const Events::CommissioningRequestResult::Type & result)
```

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
```diff
@@ -0,0 +1,231 @@
...
+bool emberAfCommissionerControlClusterRequestCommissioningApprovalCallback(
+    app::CommandHandler * commandObj, const app::ConcreteCommandPath & commandPath,
+    const Clusters::CommissionerControl::Commands::RequestCommissioningApproval::DecodableType & commandData)
+{
+    CHIP_ERROR err = CHIP_NO_ERROR;
+    Status status  = Status::Success;
+
+    ChipLogProgress(Zcl, "Received command to request commissioning approval");
+
+    auto sourceNodeId = getNodeId(commandObj);
```
bzbarsky-apple, line 157: OK, so this can be a group node ID or some such, right?  Or the PASE made-up ID thing?

The spec requires the event to include "the NodeId of the client which sent the RequestCommissioningApproval command", but that might not be known (e.g. if the command was groupcast).

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
```diff
@@ -0,0 +1,231 @@
...
+{
+    CHIP_ERROR err = CHIP_NO_ERROR;
+
+    ChipLogProgress(Zcl, "Received command to commission node");
+
+    auto sourceNodeId = getNodeId(commandObj);
+    auto requestId    = commandData.requestId;
+
+    auto info       = std::make_unique<Clusters::CommissionerControl::CommissionNodeInfo>();
+    info->ipAddress = commandData.ipAddress;
```
bzbarsky-apple, line 204: Once this function returns, ipAddress is a dangling pointer, no?

In general, the memory management around this stuff needs to be very clearly documented.  Right now it's not.

bzbarsky-apple (changes requested):

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
yufengwangca: Move this logic into the command handler itself, the Commissioner Control cluster server can only receive RequestCommissioningApproval and CommissionNode command over a CASE session from uni-cast node ID

bzbarsky-apple: > the Commissioner Control cluster server can only receive RequestCommissioningApproval and CommissionNode command over a CASE session from uni-cast node ID

I don't see the spec saying anything like that.  Where do you see that?

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
yufengwangca: The ipAddress assigned from commandData need to be stored in CommissionerControl::Delegate which ensure that the backing buffer of ipAddress has a valid lifespan until the deferred task is executed.

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
yufengwangca: Add the check to return kUndefinedNodeId if get the NodeID from a group or PASE session, in the following validation of CommissionNode command, it should fail, since it won't match the NodeId as the RequestCommissioningApproval.

bzbarsky-apple: Where does the spec say that this is how it should work?

yufengwangca: In the section of "12.5.7.4. Enabling Fabric Synchronization Between Two Ecosystems"
"After commissioning is complete, the Commissionee Fabric Synchronizing Administrator will now
be accessible on the Fabric administrated by the Commissioner.
Reverse Commissioning
After the Commissioner has finished commissioning the Commissionee, the Commissioner SHOULD
initiate Reverse Commissioning using the Commissioner Control Cluster."

This add the explicit requirement to the commands
https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/10105

But I don't think the PR should be blocked on this additional clarification since we already mandate that the Commissioner SHOULD only initiate Reverse Commissioning using the Commissioner Control Cluster after the Commissioner has finished commissioning the Commissionee.

bzbarsky-apple: I don't see where any ofthe quoted text says anything about CASE or not.

bzbarsky-apple: Anyway, https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/10129 now defines the behavior clearly, and we should implement that behavior here in the SDK.

yufengwangca: The spec PR just merged, update the SDK PR, the function getNodeId already verifies the session type and returns kUndefinedNodeId if the session is not CASE. If getNodeId returns kUndefinedNodeId, we set the status to UNSUPPORTED_ACCESS.

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
```diff
@@ -0,0 +1,277 @@
...
+        return kUndefinedNodeId;
+    }
+
+    if (!commandObj->GetExchangeContext()->HasSessionHandle())
+    {
+        ChipLogError(Zcl, "Cannot access session of Command Object for Node ID");
+        return kUndefinedNodeId;
+    }
+
+    auto descriptor = commandObj->GetExchangeContext()->GetSessionHandle()->GetSubjectDescriptor();
```
bzbarsky-apple, line 55: This is a really complicated way of writing:
```suggestion
    auto descriptor = commandObj->GetSubjectDescriptor();
```
Note that commandObj can't be null here either.

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+    // Return kUndefinedNodeId if get the NodeID from a group or PASE session.
```
bzbarsky-apple, line 57:
```suggestion
    // Return kUndefinedNodeId if this is not a CASE session.
```

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
```diff
@@ -0,0 +1,277 @@
...
+#include <memory>
+
+using namespace chip;
+using namespace chip::app;
+
+using chip::Protocols::InteractionModel::Status;
+
+namespace {
+
+NodeId getNodeId(const app::CommandHandler * commandObj)
```
bzbarsky-apple, line 41:
```suggestion
NodeId GetNodeId(const CommandHandler * commandObj)
```

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+    Clusters::CommissionerControl::Delegate * delegate =
+        Clusters::CommissionerControl::CommissionerControlServer::Instance().GetDelegate();
+
+    if (delegate != nullptr)
+    {
+        CHIP_ERROR err = delegate->ReverseCommissionNode(info->params, info->ipAddress, info->port);
+        if (err != CHIP_NO_ERROR)
+        {
+            ChipLogError(Zcl, "ReverseCommissionNode error: %s", err.AsString());
```
bzbarsky-apple, line 92:
```suggestion
            ChipLogError(Zcl, "ReverseCommissionNode error: %" CHIP_ERROR_FORMAT, err.Format());
```

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
```diff
@@ -0,0 +1,277 @@
...
+    return mInstance;
+}
+
+CHIP_ERROR CommissionerControlServer::Init(Delegate & delegate)
+{
+    mDelegate = &delegate;
+    return CHIP_NO_ERROR;
+}
+
+Protocols::InteractionModel::Status CommissionerControlServer::GetSupportedDeviceCategoriesValue(
```
bzbarsky-apple, line 123:
```suggestion
Status CommissionerControlServer::GetSupportedDeviceCategoriesValue(
```

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
```diff
@@ -0,0 +1,277 @@
...
+{
+    Status status = Attributes::SupportedDeviceCategories::Get(endpoint, supportedDeviceCategories);
+    if (status != Status::Success)
+    {
+        ChipLogProgress(Zcl, "CommissionerControl: reading  supportedDeviceCategories, err:0x%x", to_underlying(status));
+    }
+    return status;
+}
+
+Protocols::InteractionModel::Status
```
bzbarsky-apple, line 134:
```suggestion
Status
```

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
```diff
@@ -0,0 +1,277 @@
...
+    Status status  = Status::Success;
+
+    ChipLogProgress(Zcl, "Received command to request commissioning approval");
+
+    auto sourceNodeId = getNodeId(commandObj);
+
+    // Check if the command is executed via a CASE session
+    if (sourceNodeId == kUndefinedNodeId)
+    {
+        ChipLogError(Zcl, "Command not executed via CASE session, failing with UNSUPPORTED_ACCESS");
```
bzbarsky-apple, line 181: We already logged an error for this in getNodeId.  Why are we logging both places?

Also, I strongly suggest saying _which_ command.

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+    VerifyOrExit(delegate != nullptr, err = CHIP_ERROR_INCORRECT_STATE);
+
+    // Handle commissioning approval request
+    err = delegate->HandleCommissioningApprovalRequest(request);
+
+exit:
+    if (err != CHIP_NO_ERROR)
+    {
+        ChipLogError(Zcl, "emberAfCommissionerControlClusterRequestCommissioningApprovalCallback error: %s", err.AsString());
```
bzbarsky-apple, line 214:
```suggestion
        ChipLogError(Zcl, "emberAfCommissionerControlClusterRequestCommissioningApprovalCallback error: %" CHIP_ERROR_FORMAT, err.Format());
```

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
```diff
@@ -0,0 +1,277 @@
...
+        status = Status::Failure;
```
bzbarsky-apple, line 215: I strongly suggest converting the actual error to a status, in case the delegate wants to communicate a specific status.:
```suggestion
        status = StatusIB(err).mStatus;
```

or so.

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
```diff
@@ -0,0 +1,277 @@
...
+    CHIP_ERROR err = CHIP_NO_ERROR;
+
+    ChipLogProgress(Zcl, "Received command to commission node");
+
+    auto sourceNodeId = getNodeId(commandObj);
+
+    // Check if the command is executed via a CASE session
+    if (sourceNodeId == kUndefinedNodeId)
+    {
+        ChipLogError(Zcl, "Command not executed via CASE session, failing with UNSUPPORTED_ACCESS");
```
bzbarsky-apple, line 235: See comments above.

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
```diff
@@ -0,0 +1,277 @@
...
+    // Add the response for the commissioning window.
+    AddReverseOpenCommissioningWindowResponse(commandObj, commandPath, commissionNodeInfo->params);
+
+    // Schedule the deferred reverse commission node task
+    DeviceLayer::PlatformMgr().ScheduleWork(RunDeferredCommissionNode, reinterpret_cast<intptr_t>(commissionNodeInfo.release()));
+
+exit:
+    if (err != CHIP_NO_ERROR)
+    {
+        ChipLogError(Zcl, "emberAfCommissionerControlClusterCommissionNodeCallback error: %s", err.AsString());
```
bzbarsky-apple, line 267:
```suggestion
        ChipLogError(Zcl, "emberAfCommissionerControlClusterCommissionNodeCallback error: %" CHIP_ERROR_FORMAT, err.Format());
```

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
```diff
@@ -0,0 +1,277 @@
...
+        commandObj->AddStatus(commandPath, Status::Failure);
+    }
+
+    return true;
+}
+
+void MatterCommissionerControlPluginServerInitCallback()
+{
+    ChipLogProgress(Zcl, "Initiating Commissioner Control cluster.");
```
bzbarsky-apple, line 276:
```suggestion
    ChipLogProgress(Zcl, "Initializing Commissioner Control cluster.");
```

src/app/clusters/commissioner-control-server/commissioner-control-server.h
```diff
@@ -0,0 +1,180 @@
...
+
+#include <app-common/zap-generated/cluster-objects.h>
+
+namespace chip {
+namespace app {
+namespace Clusters {
+namespace CommissionerControl {
+
+// Set to 64 bytes, which is more than enough for storing an IPv6 address in either binary or text form.
+static constexpr size_t kIpAddressBufferSize = 64;
```
bzbarsky-apple, line 28: I don't get it.  Spec says this field is 4 or 16 bytes.  Where is the 64 coming from?

There is no "text form" involved; it's binary form only.

src/app/clusters/commissioner-control-server/commissioner-control-server.h
```diff
@@ -0,0 +1,180 @@
...
+        if (!address.HasValue())
+        {
+            ipAddress.ClearValue();
+            return CHIP_NO_ERROR;
+        }
+
+        const ByteSpan & addressSpan = address.Value();
+        size_t addressLength         = addressSpan.size();
+
+        if (addressLength > kIpAddressBufferSize)
```
bzbarsky-apple, line 67: The correct check is that it's 4 or 16.

src/app/clusters/commissioner-control-server/commissioner-control-server.h
```diff
@@ -0,0 +1,180 @@
...
+        {
+            return CHIP_ERROR_BUFFER_TOO_SMALL;
+        }
+
+        if (addressLength == 0)
+        {
+            ipAddress.ClearValue();
+            return CHIP_NO_ERROR;
```
bzbarsky-apple, line 75: Per spec, this should be an error, no?

src/app/clusters/commissioner-control-server/commissioner-control-server.h
```diff
@@ -0,0 +1,180 @@
...
+    uint16_t commissioningTimeout;
+    uint16_t discriminator;
+    ByteSpan PAKEPasscodeVerifier;
+    ByteSpan salt;
+};
+
+struct CommissionNodeInfo
+{
+    CommissioningWindowParams params;
+    Optional<ByteSpan> ipAddress;
```
bzbarsky-apple, line 52: Strongly suggest making ipAddress protected and having a getter for a const thing, to avoid people accidentally creating security bugs.

src/app/clusters/commissioner-control-server/commissioner-control-server.cpp
```diff
@@ -0,0 +1,277 @@
...
+    AddReverseOpenCommissioningWindowResponse(commandObj, commandPath, commissionNodeInfo->params);
+
+    // Schedule the deferred reverse commission node task
+    DeviceLayer::PlatformMgr().ScheduleWork(RunDeferredCommissionNode, reinterpret_cast<intptr_t>(commissionNodeInfo.release()));
+
+exit:
+    if (err != CHIP_NO_ERROR)
+    {
+        ChipLogError(Zcl, "emberAfCommissionerControlClusterCommissionNodeCallback error: %s", err.AsString());
+        commandObj->AddStatus(commandPath, Status::Failure);
```
bzbarsky-apple, line 268: Again, why are we suppressing the actual error?

src/app/clusters/commissioner-control-server/commissioner-control-server.h
```diff
@@ -0,0 +1,180 @@
...
+
+    /**
+     * @brief Get the parameters for the commissioning window.
+     *
+     * This method is called to retrieve the parameters needed for the commissioning window.
+     *
+     * @param[out] outParams The parameters for the commissioning window.
+     * @return CHIP_ERROR indicating the success or failure of the operation.
+     */
+    virtual CHIP_ERROR GetCommissioningWindowParams(CommissioningWindowParams & outParams) = 0;
```
bzbarsky-apple, line 126: Where is this used?

tehampson: Removed

tehampson: Yufeng pinged me and mentioned section of code where it was supposed to be used but was accidentally removed. It is now added back in

bzbarsky-apple (approved):

tehampson (approved):

examples/fabric-bridge-app/fabric-bridge-common/fabric-bridge-app.matter
```diff
@@ -1647,6 +1698,21 @@ endpoint 0 {
     handle command KeySetReadAllIndices;
     handle command KeySetReadAllIndicesResponse;
   }
+
+  server cluster CommissionerControl {
```
arkq, line 1702: @tehampson Is that OK, that the CommissionerControl is on endpoint 0? In the spec, the CommissionerControl should be on Aggregator endpoint, which is endpoint 1 here.

PR #34376, opened 2024-07-17
Prefix `InteractionModel` namespace with `protocols`

andy31415:
These are naming changes in preparation for #34288 - as these are not direct code/functionality changes but rather resolving namespace conflicts.

### DIFF
--- a/src/app/tests/suites/commands/interaction_model/InteractionModel.cpp
+++ b/src/app/tests/suites/commands/interaction_model/InteractionModel.cpp
@@ -19,7 +19,20 @@
 #include "InteractionModel.h"
 
 using namespace chip;
-using namespace chip::app;
+
+using chip::app::AttributePathParams;
+using chip::app::CommandSender;
+using chip::app::ConcreteAttributePath;
+using chip::app::ConcreteCommandPath;
+using chip::app::ConcreteDataAttributePath;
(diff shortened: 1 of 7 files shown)


saurabhst (approved):

src/app/tests/suites/commands/interaction_model/InteractionModel.cpp
```diff
@@ -19,7 +19,20 @@
 #include "InteractionModel.h"
 
 using namespace chip;
-using namespace chip::app;
```
ksperling-apple, line 22: Are there conflicts when using the whole app namespace? This seems a little cumbersome

ksperling-apple (approved):

src/app/tests/suites/commands/interaction_model/InteractionModel.cpp
andy31415: I do not remember what the conflicts were ... I think there are probably ways to make this better, there is only so much time available to spend here. Maybe when/if we touch this next we can update it to cleaner code.

andy31415: It is also unfortunate that this code as a stand-alone does not need these changes, just #34288 will need them. Once that code is integrated, we can start cleaning up things a bit more.

PR #34377, opened 2024-07-17
Introduced a python script (execute_python_tests.py) that automates the execution of python test scripts within the python_testing directory

vatsalghelani-csa:
Work for Issue #31317 

Follow up to PR #34322 

PR 12 - This PR introduces a python script (execute_python_tests.py) that automates the execution of python test scripts within the python_testing directory. By utilizing glob to identify all .py files and excluding support/non-app specific tests.


src/python_testing/execute_python_tests.py
```diff
@@ -0,0 +1,92 @@
...
+
+# Define the directory to search for Python scripts
+search_directory = "src/python_testing"
+
+# Define the files and patterns to exclude
+excluded_patterns = [
+    "MinimalRepresentation.py",
+    "TC_ACL_2_2.py",
+    "TC_CNET_4_4.py",
+    "TC_DGGEN_3_2.py",
```
andy31415, line 47: why are these TC_* excluded?

We need some explanation - I believe it may have to do with test binaries not being able to support them, however it would be nice to have a comment for each (or for a group like the support/test ones) explaining the reasoning.

@cecille do you happen to have a TLDR for some of them?

src/python_testing/execute_python_tests.py
```diff
@@ -0,0 +1,92 @@
...
+
+# Get all .py files in the directory
+all_python_files = glob.glob(os.path.join(search_directory, "*.py"))
+
+# Filter out the files matching the excluded patterns
+python_files = []
+for file in all_python_files:
+    exclude = False
+    for pattern in excluded_patterns:
+        if glob.fnmatch.fnmatch(os.path.basename(file), pattern):
```
andy31415, line 82: why fnmatch? if you make the excluded_patterns a set (via `{....}`) then you may be able to do: `if os.path.basename(file) in excluded_patterns:` which is more readable.

src/python_testing/execute_python_tests.py
```diff
@@ -0,0 +1,92 @@
...
+            exclude = True
+            break
+    if not exclude:
+        python_files.append(file)
+
+# Run each script with the base command
+for script in python_files:
+    full_command = f"{base_command} --load-from-env /tmp/test_env.yaml --script {script}"
+    print(f"Running command: {full_command}")
+    subprocess.run(full_command, shell=True)
```
andy31415, line 92: did you check what happens if a test fails? I.e. do we need `check=True` here in order to fail the execution if something fails?

src/python_testing/execute_python_tests.py
```diff
@@ -0,0 +1,92 @@
...
+    for pattern in excluded_patterns:
+        if glob.fnmatch.fnmatch(os.path.basename(file), pattern):
+            exclude = True
+            break
+    if not exclude:
+        python_files.append(file)
+
+# Run each script with the base command
+for script in python_files:
+    full_command = f"{base_command} --load-from-env /tmp/test_env.yaml --script {script}"
```
andy31415, line 90: how do we know `/tmp/test_env.yaml` is the correct path? Please use arguments.

src/python_testing/execute_python_tests.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+# Load environment variables from test_env.yaml
+load_env_from_yaml('/tmp/test_env.yaml')
+
+# Define the base command to run tests
+base_command = "./scripts/tests/run_python_test.py"
+
+# Define the directory to search for Python scripts
+search_directory = "src/python_testing"
```
andy31415, line 40: use arguments with defaults.

src/python_testing/execute_python_tests.py
```diff
@@ -0,0 +1,92 @@
...
+def load_env_from_yaml(file_path):
+    with open(file_path, 'r') as file:
+        for line in file:
+            if line.strip():  # Skip empty lines
+                key, value = line.strip().split(': ', 1)
+                os.environ[key] = value
+
+
+# Load environment variables from test_env.yaml
+load_env_from_yaml('/tmp/test_env.yaml')
```
andy31415, line 34: do not hardcode paths.

src/python_testing/execute_python_tests.py
```diff
@@ -0,0 +1,92 @@
...
+
+# Function to load environment variables from a YAML file
+
+
+def load_env_from_yaml(file_path):
+    with open(file_path, 'r') as file:
+        for line in file:
+            if line.strip():  # Skip empty lines
+                key, value = line.strip().split(': ', 1)
+                os.environ[key] = value
```
andy31415, line 30: why do we do this? What is the purpose of this yaml.

andy31415 (changes requested):
Please cleanup the PR:

- [x] use arguments instead of hardcoding paths (you can have defaults for things)
- [x] do not assume `current directory`. For file paths use the directory of __file__ to figure out chip_root
- [x] explain the "move yaml into os.environ" if you really need it. That one looks odd as the previous code did not need it
- [x] ensure that this script will fail if a test execution fails
- [x] add explanations for excluded files

src/python_testing/execute_python_tests.py
vatsalghelani-csa: For why I excluded: I just excluded the list of tests from the tests.yaml block which were not the app tests or being run through that block of tests mentioned.

src/python_testing/execute_python_tests.py
vatsalghelani-csa: Fixed.

src/python_testing/execute_python_tests.py
vatsalghelani-csa: Fixed.

src/python_testing/execute_python_tests.py
vatsalghelani-csa: Fixed.

src/python_testing/execute_python_tests.py
vatsalghelani-csa: Fixed.

src/python_testing/execute_python_tests.py
vatsalghelani-csa: Fixed.

src/python_testing/execute_python_tests.py
vatsalghelani-csa: Fixed.

src/python_testing/execute_python_tests.py
```diff
@@ -0,0 +1,116 @@
...
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+import argparse
+import glob
+import os
+import subprocess
+
+# Function to load environment variables from a YAML file
```
cecille, line 23: This comment isn't quite true

vatsalghelani-csa: Fixed.

vatsalghelani-csa:
@cecille - We might either need to exclude TC_VALCC* and TC_BOOLCFG* tests or need to add the necessary arguments patterns as per the new design in those tests scripts. This tests were not part of the tests.yaml block script that is executing or were also not defined in the tests.yaml to run earlier.

Since I see the CI checks too failed in REPL Tests - Linux (Run Tests section) with the below error:
```
Exception: No valid runs were found. Make sure you add runs to your file, see https://github.com/project-chip/connectedhomeip/blob/master/docs/testing/python.md document for reference/example.
.
subprocess.CalledProcessError: Command '/__w/connectedhomeip/connectedhomeip/scripts/tests/run_python_test.py --load-from-env /tmp/test_env.yaml --script src/python_testing/TC_VALCC_3_4.py' returned non-zero exit status 1.
.
Running command: /__w/connectedhomeip/connectedhomeip/scripts/tests/run_python_test.py --load-from-env /tmp/test_env.yaml --script src/python_testing/TC_VALCC_3_4.py
Error: Process completed with exit code 1.
```

Though I see other tests following the pattern succeeded, like Executing TC_OVENOPSTATE_2_2.py, TC_ACE_1_3.py, and test_scripts/mobile-device-test.py...

Apollon77:
@cecille Please also make sure that the other parameters can be set here, like it was on "executing the runners normally" because matter.js would also like to use this but here we need to be able to exchange at minimum the "app location" of the DUT. I think we only need the "--app" one especially  currently.

In general see https://github.com/project-chip/matter.js/blob/main/.github/workflows/chip-tool-tests.yml#L337

Apollon77:
@vatsalghelani-csa I could also imagine to have an optional "config file" per test (same name with different extension or one config file with infos per test) to include special execution parameters to be used too

cecille:
The errors are pretty good about telling you what's missing. Ex.

Details=PIXIT.BOOLCFG.TEST_EVENT_TRIGGER_KEY must be included on the command line in the --hex-arg flag as PIXIT.BOOLCFG.TEST_EVENT_TRIGGER_KEY:<key>, e.g. --hex-arg PIXIT.BOOLCFG.TEST_EVENT_TRIGGER_KEY:000102030405060708090a0b0c0d0e0f, Extras=None

so...that needs to be added.

cecille:
@Apollon77 - we can consider that as a follow up. I think it'll probably end up in the metadata parser, which is not yet in its final form.

andy31415 (approved):

andy31415:
> @cecille Please also make sure that the other parameters can be set here, like it was on "executing the runners normally" because matter.js would also like to use this but here we need to be able to exchange at minimum the "app location" of the DUT. I think we only need the "--app" one especially currently.
> 
> In general see https://github.com/project-chip/matter.js/blob/main/.github/workflows/chip-tool-tests.yml#L337

the metadata loader currently uses a `yaml` file for the environment to run under to pick its parameters (in particular app locations). In this case matter.js could pick different application paths as needed.

andy31415:
> @vatsalghelani-csa I could also imagine to have an optional "config file" per test (same name with different extension or one config file with infos per test) to include special execution parameters to be used too

I would prefer to not separate data as much as possible, this is why we currently have the config in the test itself (which makes a copy & paste trivially work).

The alternative considered is having a separate config where the metadata loads from, that contains everything. Either way, I very much prefer to limit the number of files and locations changes occur. One of the challanges we are trying to fix was that `tests.yaml` essientially contained configs for things under `src/python` so anyone wanting to just run a test was unable without seeing how tests.yaml is setup (also very easy to miss updating tests.yaml and not have some tests in CI).

PR #34378, opened 2024-07-17
Depend less on precise all-clusters-app endpoints layout in Darwin tests.

### LINKED ISSUE #34374: Darwin tests expect no endpoints to ever be added to all-clusters-app
Darwin tests have many assumptions that there are endpoints 0, 1, 2. Adding a 3rd endpoint breaks a lot of tests.

Example:

```
2024-07-17T03:16:51.5527330Z /Users/runner/work/connectedhomeip/connectedhomeip/src/darwin/Framework/CHIPTests/MTRPerControllerStorageTests.m:2548: error: -[MTRPerControllerStorageTests testDataStorageUpdatesWhenRemovingEndpoints] : (([[controller.controllerDataStore _fetchEndpointIndexForNodeID:deviceID] isEqualToArray:testEndpoints]) is true) failed
```

which relates to:

```
    __block NSMutableDictionary<NSNumber *, NSArray<NSNumber *> *> * initialClusterIndex = [[NSMutableDictionary alloc] init];
    __block NSMutableArray<NSNumber *> * testEndpoints;

    delegate.onReportEnd = ^{
        XCTAssertNotNil(dataVersionForPartsList);
        XCTAssertNotNil(testClusterDataValueForPartsList);

        testEndpoints = [self getEndpointArrayFromPartsList:testDataForPartsList forDevice:device];

        // Make sure that the cluster data in the data storage is populated with cluster index and cluster data for endpoints 0, 1 and 2.
        // We do not need to check _persistedClusterData here. _persistedClusterData will be paged in from storage when needed so
        // just checking data storage should suffice here.
        dispatch_sync(self->_storageQueue, ^{
```

The test above (among others) expects exactly 3 endpoints enabled, and all of them to use NV storage.

This breaks the ability of anyone to change the composition of the all-clusters-app without cryptic failures related to low-level tests of the Darwin regression suite.

bzbarsky-apple:
* test020_ReadMultipleAttributes and test021_ReadMultipleWildcardPathsIncludeUnsupportedAttribute were assuming three endpoints numbered 0, 1, 2. The fix is to just read the PartsList from the device instead of making assumptions.
* test033_TestMTRDeviceDeviceConfigurationChanged was assuming various things about which clusters are on which endpoints, as well as assuming that the PartsList of EP0 is _not_ [1, 2, 3].  Fix things so that we track which endpoint things are actually on, and make sure we take the PartsList from the device and add an entry that's not in it already to it instead of assuming its value.
* testDataStorageUpdatesWhenRemovingEndpoints was comparing arrays of endpoints, but some of those arrays come from enumerating dictionary keys.  Changing the set of keys can change their order.  The fix is to compare sets of endpoints instead.  Also, add asserts that the endpoint we are removing (2) is actually in the list, so that if that ever stops being true the test failure will be more understandable.

### DIFF
--- a/src/darwin/Framework/CHIPTests/MTRPerControllerStorageTests.m
+++ b/src/darwin/Framework/CHIPTests/MTRPerControllerStorageTests.m
@@ -2535,7 +2535,7 @@ static const uint16_t kSubscriptionPoolBaseTimeoutInSeconds = 30;
     };
 
     __block NSMutableDictionary<NSNumber *, NSArray<NSNumber *> *> * initialClusterIndex = [[NSMutableDictionary alloc] init];
-    // Some of the places we gets endpoint lists get them from enumerating dictionary keys, which
+    // Some of the places we get endpoint lists get them from enumerating dictionary keys, which
     // means order is not guaranteed.  Make sure we compare sets of endpoints, not arrays, to
     // account for that.
     __block NSSet<NSNumber *> * testEndpoints;
@@ -2620,7 +2620,7 @@ static const uint16_t kSubscriptionPoolBaseTimeoutInSeconds = 30;
         XCTAssertNotNil(testClusterDataValueForPartsList);
         testEndpoints = [NSSet setWithArray:[self getEndpointArrayFromPartsList:testDataForPartsList forDevice:device]];
 
-        // Make sure that the cluster data is remoeved from the data storage for the endpoint we deleted, but still there for the others.
+        // Make sure that the cluster data is removed from the data storage for the endpoint we deleted, but still there for the others.
         // We do not need to check _persistedClusterData here. _persistedClusterData will be paged in from storage when needed so
         // just checking data storage should suffice here.
         dispatch_sync(self->_storageQueue, ^{
--- a/src/darwin/Framework/CHIPTests/MTRPerControllerStorageTests.m
+++ b/src/darwin/Framework/CHIPTests/MTRPerControllerStorageTests.m
@@ -2508,11 +2508,12 @@ static const uint16_t kSubscriptionPoolBaseTimeoutInSeconds = 30;
     __block NSNumber * rootEndpoint = @0;
 
     // This test will do the following -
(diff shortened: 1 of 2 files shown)


src/darwin/Framework/CHIPTests/MTRPerControllerStorageTests.m
```diff
@@ -2612,13 +2618,13 @@ - (void)testDataStorageUpdatesWhenRemovingEndpoints
 
     delegate.onReportEnd = ^{
         XCTAssertNotNil(testClusterDataValueForPartsList);
-        testEndpoints = [self getEndpointArrayFromPartsList:testDataForPartsList forDevice:device];
+        testEndpoints = [NSSet setWithArray:[self getEndpointArrayFromPartsList:testDataForPartsList forDevice:device]];
 
-        // Make sure that the cluster data in the data storage for endpoints 0 and 1 are present but not for endpoint 2.
+        // Make sure that the cluster data is remoeved from the data storage for the endpoint we deleted, but still there for the others.
```
kiel-apple, line 2623:
```suggestion
        // Make sure that the cluster data is removed from the data storage for the endpoint we deleted, but still there for the others.
```

src/darwin/Framework/CHIPTests/MTRPerControllerStorageTests.m
```diff
@@ -2534,18 +2535,21 @@ - (void)testDataStorageUpdatesWhenRemovingEndpoints
     };
 
     __block NSMutableDictionary<NSNumber *, NSArray<NSNumber *> *> * initialClusterIndex = [[NSMutableDictionary alloc] init];
-    __block NSMutableArray<NSNumber *> * testEndpoints;
+    // Some of the places we gets endpoint lists get them from enumerating dictionary keys, which
```
kiel-apple, line 2538:
```suggestion
    // Some of the places we get endpoint lists get them from enumerating dictionary keys, which
```

kiel-apple (approved):

bzbarsky-apple:

PR #34384, opened 2024-07-18
python test framework: PICS 2.0

cecille:
Adds test decorators and hooks that will let the test framework select ALL appropriate endpoints for a test and run the test against each endpoint. This means that tests will only need to be run once for the entire node, rather than against a specific endpoint.

Includes a hooks callback to indicate when no appropriate endpoints are selected as well as new decorators for per_endpoint and per_node tests. decorator demonstrated on the time sync test.


src/python_testing/matter_testing_support.py
```diff
@@ -1541,12 +1561,78 @@ def async_test_body(body):
...
+        return False
+
+def has_attribute(attribute: ClusterObjects.ClusterAttributeDescriptor) -> EndpointCheckFunction:
+    return partial(_has_attribute, attribute=attribute)
+
+async def get_accepted_endpoints_for_test(self:MatterBaseTest, accept_function: EndpointCheckFunction):
+    wildcard = await self.default_controller.Read(self.dut_node_id, [()])
+    return [e for e in wildcard.attributes.keys() if accept_function(wildcard, e)]
+
+def per_endpoint_test(accept_function):
```
andy31415, line 1607: Please add doccomments for all the newly introduced top level functions. We need to describe why they exist and how they are to be used.

src/python_testing/matter_testing_support.py
```diff
@@ -949,12 +954,11 @@ def on_fail(self, record):
...
             record is of type TestResultRecord
         '''
+        self.failed = True
         if self.runner_hook and not self.is_commissioning:
             exception = record.termination_signal.exception
             step_duration = (datetime.now(timezone.utc) - self.step_start_time) / timedelta(microseconds=1)
-            # This isn't QUITE the test duration because the commissioning is handled separately, but it's clsoe enough for now
-            # This is already given in milliseconds
-            test_duration = record.end_time - record.begin_time
+            test_duration = datetime.now(timezone.utc) - self.test_start_time
```
ccruzagralopes, line 961: When trying to run this on TH we got an error because TH expects an int for duration. Could you convert it to an int representing the time in milliseconds?

cecille: changed.

ccruzagralopes (approved):
Tested this using TH and it worked fine 😄

src/python_testing/matter_testing_support.py
cecille: added

raju-apple (approved):

src/python_testing/test_testing/TestDecorators.py
```diff
@@ -0,0 +1,291 @@
...
+import chip.clusters as Clusters
+from chip.clusters import Attribute
+from chip.clusters import ClusterObjects as ClusterObjects
+
+try:
+    from matter_testing_support import (MatterBaseTest, async_test_body, get_accepted_endpoints_for_test, has_attribute,
+                                        has_cluster, per_endpoint_test, per_node_test)
+except ImportError:
+    sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), '..')))
+    from matter_testing_support import MatterBaseTest, async_test_body, has_cluster, has_attribute, get_accepted_endpoints_for_test, per_endpoint_test, per_node_test
```
tehampson, line 40: nit, Is there a reason why you import different things when on ImportError?

cecille: okie dokie - updated.

src/python_testing/test_testing/TestDecorators.py
```diff
@@ -0,0 +1,292 @@
...
+
+        endpoints = await get_accepted_endpoints_for_test(self, has_timesync_utc)
+        asserts.assert_equal(endpoints, [], msg)
+
+    # This test should cause an assertion because it has pics_ method
+    @per_node_test
+    async def test_whole_node_with_pics(self):
+        pass
+
+    def pics_whole_node_with_pics(self):
```
tehampson, line 162: What is `pics_whole_node_with_pics` used for? I cannot seems to find where it is used

src/python_testing/test_testing/TestDecorators.py
```diff
@@ -0,0 +1,292 @@
...
+    @per_endpoint_test(has_attribute(Clusters.OnOff.Attributes.OnOff) and has_cluster(Clusters.OnOff))
+    async def test_endpoint_boolean_yes(self):
+        pass
+
+    # This test should be skipped since we have an OnOff cluster, but no Time sync
+    @per_endpoint_test(has_cluster(Clusters.OnOff) and has_cluster(Clusters.TimeSynchronization))
+    async def test_endpoint_boolean_no(self):
+        pass
+
+    # TODO: add test for assertions on whole node, and on each endpoint of an endpoint test
```
tehampson, line 213: Is this a TODO you wanted to do in this PR or as a follow up. If follow up, do you have a tracking issue that would provide additional context for a future reader should you not have time to follow up?

src/python_testing/test_testing/TestDecorators.py
```diff
@@ -0,0 +1,292 @@
+#
```
tehampson, line 1: Do we need to add this file to .github/workflows/tests.yaml so that is runs as a part of CI

tehampson (approved):

src/python_testing/test_testing/TestDecorators.py
cecille: There is an assertion on the decorator that tests with per_endpoint_test or per_node_test can't have a corresponding pics method, because pics are supplied per endpoint, and these tests require information from the entire node. 

Added a comment on both.

src/python_testing/test_testing/TestDecorators.py
cecille: yup - good call, thanks.

src/python_testing/test_testing/TestDecorators.py
cecille: nope, just forgot. Added now.

src/python_testing/test_testing/TestDecorators.py
cecille: added.

PR #34387, opened 2024-07-18
Return CHIP_ERROR from BlePlatformDelegate APIs

arkq:
### Problem

It's not possible to return error code from the platform delegate to the BLE module.

### Changes

Change platform delegate API signature from `bool Func(...)` -> `CHIP_ERROR Func(...)`

### Testing

CI will verify potential build breaks

### DIFF
--- a/src/platform/Darwin/BlePlatformDelegate.h
+++ b/src/platform/Darwin/BlePlatformDelegate.h
@@ -30,14 +30,16 @@ namespace Internal {
 class BlePlatformDelegateImpl : public Ble::BlePlatformDelegate
 {
 public:
-    virtual bool SubscribeCharacteristic(BLE_CONNECTION_OBJECT connObj, const ChipBleUUID * svcId, const ChipBleUUID * charId);
-    virtual bool UnsubscribeCharacteristic(BLE_CONNECTION_OBJECT connObj, const ChipBleUUID * svcId, const ChipBleUUID * charId);
-    virtual bool CloseConnection(BLE_CONNECTION_OBJECT connObj);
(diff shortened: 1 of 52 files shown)


src/platform/Darwin/BlePlatformDelegate.h
```diff
@@ -30,14 +30,16 @@ namespace Internal {
 class BlePlatformDelegateImpl : public Ble::BlePlatformDelegate
 {
 public:
-    virtual bool SubscribeCharacteristic(BLE_CONNECTION_OBJECT connObj, const ChipBleUUID * svcId, const ChipBleUUID * charId);
-    virtual bool UnsubscribeCharacteristic(BLE_CONNECTION_OBJECT connObj, const ChipBleUUID * svcId, const ChipBleUUID * charId);
-    virtual bool CloseConnection(BLE_CONNECTION_OBJECT connObj);
+    virtual CHIP_ERROR SubscribeCharacteristic(BLE_CONNECTION_OBJECT connObj, const ChipBleUUID * svcId,
+                                               const ChipBleUUID * charId);
+    virtual CHIP_ERROR UnsubscribeCharacteristic(BLE_CONNECTION_OBJECT connObj, const ChipBleUUID * svcId,
```
bzbarsky-apple, line 35: If these are overrides, should they be marked as "override", not "virtual"?

Maybe a followup PR, or just let me know when this lands and I can fix it at least for Darwin.

bzbarsky-apple (approved):

src/platform/Darwin/BlePlatformDelegate.h
arkq: Yes, that should be fixed. I'll try to do it in this PR so we will not forget about that.

andy31415 (approved):

bauerschwan (approved):

PR #34392, opened 2024-07-18
Update `ember-compatibility-functions` to always return error codes instead of directly encoding them as success.

andy31415:
This lets the read implementation in `Engine.cpp` do the work of encoding failures and doing rollback/restore.

Without this, the logic was duplicated in ember-compatibility-functions which results in extra complexity and flash usage.

For reference, https://github.com/project-chip/connectedhomeip/blob/master/src/app/reporting/Engine.cpp#L205 has logic for handling errors and generally it has to accept them because it has special handling for out of space.

### DIFF
--- a/src/app/util/ember-compatibility-functions.cpp
+++ b/src/app/util/ember-compatibility-functions.cpp
@@ -455,14 +455,8 @@ CHIP_ERROR ReadSingleClusterData(const SubjectDescriptor & aSubjectDescriptor, b
         uint8_t dataLength = gEmberAttributeIOBufferSpan[0];
         if (dataLength == 0xFF)
         {
-            if (isNullable)
-            {
-                ReturnErrorOnFailure(writer->PutNull(tag));
-            }
-            else
-            {
-                return CHIP_ERROR_INCORRECT_STATE;
-            }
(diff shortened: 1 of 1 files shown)


src/app/util/ember-compatibility-functions.cpp
```diff
@@ -372,214 +351,173 @@ CHIP_ERROR ReadSingleClusterData(const SubjectDescriptor & aSubjectDescriptor, b
...
-    if (status == Protocols::InteractionModel::Status::Success)
-    {
-        return SendSuccessStatus(attributeReport, attributeDataIBBuilder);
+    default:
+        ChipLogError(DataManagement, "Attribute type 0x%x not handled", static_cast<int>(attributeType));
+        return CHIP_IM_GLOBAL_STATUS(Failure);
     }
 
-    return SendFailureStatus(aPath, aAttributeReports, status, &backup);
+    // If we got this far, write succeeded. Encode the success by terminating the write
```
bzbarsky-apple, line 518:
```suggestion
    // If we got this far, placing the data to be read in the output TLVWriter succeeded.
    // Try to terminate our attribute report to signal success.
```

bzbarsky-apple (approved):

yufengwangca (approved):

PR #34393, opened 2024-07-18
[SILABS] Minor fixes for LWIP usage 

mykrupp:
Partially transition to use CMSIS OS2 api calls for lwip

### DIFF
--- a/examples/platform/silabs/efr32/rs911x/rsi_if.c
+++ b/examples/platform/silabs/efr32/rs911x/rsi_if.c
@@ -26,6 +26,7 @@
 #include "em_usart.h"
 
 #include "sl_status.h"
+#include "FreeRTOS.h"
 #include <cmsis_os2.h>
 // TODO Fix include order issue #33120
 #include "wfx_host_events.h"
--- a/examples/platform/silabs/efr32/rs911x/rsi_if.c
+++ b/examples/platform/silabs/efr32/rs911x/rsi_if.c
@@ -25,8 +25,8 @@
 #include "em_ldma.h"
 #include "em_usart.h"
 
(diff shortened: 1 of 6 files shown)


src/platform/silabs/PlatformManagerImpl.cpp
```diff
@@ -47,13 +47,13 @@ namespace DeviceLayer {
 PlatformManagerImpl PlatformManagerImpl::sInstance;
 
 #if defined(SL_MBEDTLS_USE_TINYCRYPT)
-sys_mutex_t PlatformManagerImpl::rngMutexHandle = NULL;
+sys_mutex_t PlatformManagerImpl::rngMutexHandle = nullptr;
```
jmartinez-silabs, line 50:
```suggestion
osMutexId_t PlatformManagerImpl::rngMutexHandle = nullptr;
```

examples/platform/silabs/efr32/rs911x/rsi_if.c
```diff
@@ -25,6 +25,7 @@
 #include "em_ldma.h"
 #include "em_usart.h"
 
+#include "FreeRTOS.h"
```
jmartinez-silabs, line 28: Why is freeRTOS needed here if #include <cmsis_os2.h> is used lower?

cmsis should be favoured over freeRTOS since we are transitioning to full cmsis apis

mykrupp: Good point, removed the FreeRTOS includes.

jmartinez-silabs (approved):

andy31415 (approved):

PR #34396, opened 2024-07-18
Add checks for OffWithEffect command parameters contraints

### LINKED ISSUE #34214: OnOff Cluster does not enforce EffectVariant value requirements correctly
### Reproduction steps

The type for Field 1 of Command 0x40 in the OnOff cluster should be DelayedAllOffEffectVariantEnum, but it is currently written as enum8. The range of values is inconsistent.

image

The range of DelayedAllOffEffectVariantEnum is 0-2, result in 
```
./chip-tool onoff off-with-effect 0 4   0x654325 1
```
This command could work incorrectly due to the missing constraints.

### Bug prevalence

each time

5bb5c9e23d532cea40476fc0bd1d3008522792ba

### Platform

other

### Platform Version(s)

1.3

### Type

Spec Compliance Issue, Security Issue, Core SDK Performance Improvement, Core SDK Interopability Issue

### Anything else?

_No response_

jmartinez-silabs:
fixes #34214

It was possible to pass invalid enum values to the `OffWithEffect` command from the on-off cluster.

Add checks against the passed arguments in the command handler
- handle the `effectVariant` parameter range variation depending on the `effectId` parameter.

**Note the On-Off cluster test plan must be expanded to test invalid parameter handling.**

Manually tested with chip-tool and efr32 lighting-app.
**Invalid effectVariant when effectId is DelayedAllOff**
```
./chip-tool onoff off-with-effect 0 10 $NODE_ID 1
[1721319091.737648][24180:24182] CHIP:DMG: Received Command Response Status for Endpoint=1 Cluster=0x0000_0006 Command=0x0000_0040 Status=0x87
[1721319091.737713][24180:24182] CHIP:TOO: Error: IM Error 0x00000587: General error: 0x87 (CONSTRAINT_ERROR)
```

**Valid effectVariant when effectId is DelayedAllOff**
```
./chip-tool onoff off-with-effect 0 2 $NODE_ID 1
[1721319152.105680][24187:24189] CHIP:DMG: Received Command Response Status for Endpoint=1 Cluster=0x0000_0006 Command=0x0000_0040 Status=0x0
```

**Invalid effectId**
```
./chip-tool onoff off-with-effect 2 2 $NODE_ID 1
[1721319203.173323][24194:24196] CHIP:DMG: Received Command Response Status for Endpoint=1 Cluster=0x0000_0006 Command=0x0000_0040 Status=0x87
[1721319203.173389][24194:24196] CHIP:TOO: Error: IM Error 0x00000587: General error: 0x87 (CONSTRAINT_ERROR)
```

**Invalid effectVariant when effectId is DyingLight**
```
./chip-tool  onoff off-with-effect 1 2 $NODE_ID 1
[1721319220.474435][24200:24202] CHIP:DMG: Received Command Response Status for Endpoint=1 Cluster=0x0000_0006 Command=0x0000_0040 Status=0x87
[1721319220.474514][24200:24202] CHIP:TOO: Error: IM Error 0x00000587: General error: 0x87 (CONSTRAINT_ERROR)
```

**Valid effectVariant when effectId is DyingLight**
```
./chip-tool onoff off-with-effect 1 0 $NODE_ID 1
 Received Command Response Status for Endpoint=1 Cluster=0x0000_0006 Command=0x0000_0040 Status=0x0
```

### DIFF
--- a/src/app/clusters/on-off-server/on-off-server.cpp
+++ b/src/app/clusters/on-off-server/on-off-server.cpp
@@ -621,13 +621,13 @@ bool OnOffServer::offWithEffectCommand(app::CommandHandler * commandObj, const a
         // The following check validates that effectVariant value is valid in relation to the applicable enum type.
         // DelayedAllOffEffectVariantEnum or DyingLightEffectVariantEnum
         if (effectId == EffectIdentifierEnum::kDelayedAllOff &&
-            !isKnownEnumValue(static_cast<DelayedAllOffEffectVariantEnum>(effectVariant)))
+            !IsKnownEnumValue(static_cast<DelayedAllOffEffectVariantEnum>(effectVariant)))
         {
             // The server does not support the given variant, it SHALL use the default variant.
             effectVariant = to_underlying(DelayedAllOffEffectVariantEnum::kDelayedOffFastFade);
         }
         else if (effectId == EffectIdentifierEnum::kDyingLight &&
-                 !isKnownEnumValue(static_cast<DyingLightEffectVariantEnum>(effectVariant)))
+                 !IsKnownEnumValue(static_cast<DyingLightEffectVariantEnum>(effectVariant)))
         {
             // The server does not support the given variant, it SHALL use the default variant.
             effectVariant = to_underlying(DyingLightEffectVariantEnum::kDyingLightFadeOff);
--- a/src/app/clusters/on-off-server/on-off-server.cpp
+++ b/src/app/clusters/on-off-server/on-off-server.cpp
@@ -89,6 +89,12 @@ void UpdateModeBaseCurrentModeToOnMode(EndpointId endpoint)
 
 #endif // MATTER_DM_PLUGIN_MODE_BASE
 
+template <typename EnumType>
+bool isKnownEnumValue(EnumType value)
+{
+    return (EnsureKnownEnumValue(value) == EnumType::kUnknownEnumValue) ? false : true;
+}
+
 } // namespace
 
 #ifdef MATTER_DM_PLUGIN_LEVEL_CONTROL
@@ -608,6 +614,30 @@ bool OnOffServer::offWithEffectCommand(app::CommandHandler * commandObj, const a
     auto effectVariant        = commandData.effectVariant;
     chip::EndpointId endpoint = commandPath.mEndpointId;
     Status status             = Status::Success;
+
+    if (isKnownEnumValue(effectId))
+    {
+        // Depending on effectId value, effectVariant enum type varies.
+        // The following check validates that effectVariant value is valid in relation to the applicable enum type.
+        // DelayedAllOffEffectVariantEnum or DyingLightEffectVariantEnum
+        if ((effectId == EffectIdentifierEnum::kDelayedAllOff &&
+             !isKnownEnumValue(static_cast<DelayedAllOffEffectVariantEnum>(effectVariant))) ||
+            (effectId == EffectIdentifierEnum::kDyingLight &&
+             !isKnownEnumValue(static_cast<DyingLightEffectVariantEnum>(effectVariant))))
+        {
+            status = Status::ConstraintError;
+        }
+    }
+    else
+    {
+        status = Status::ConstraintError;
+    }
+
+    if (status != Status::Success)
+    {
+        commandObj->AddStatus(commandPath, status);
+        return true;
+    }
 
     if (SupportsLightingApplications(endpoint))
     {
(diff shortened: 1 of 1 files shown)


lpbeliveau-silabs (approved):

src/app/clusters/on-off-server/on-off-server.cpp
```diff
@@ -609,6 +615,30 @@ bool OnOffServer::offWithEffectCommand(app::CommandHandler * commandObj, const a
     chip::EndpointId endpoint = commandPath.mEndpointId;
     Status status             = Status::Success;
 
+    if (isKnownEnumValue(effectId))
```
bzbarsky-apple, line 618: This could just be a direct compare against `EffectIdentifierEnum::kUnknownEnumValue`, no?  Decoding of a command's fields will check them for validity if they are known enum types, and replace invalid values with `kUnknownEnumValue`

src/app/clusters/on-off-server/on-off-server.cpp
```diff
@@ -609,6 +615,30 @@ bool OnOffServer::offWithEffectCommand(app::CommandHandler * commandObj, const a
...
+    {
+        // Depending on effectId value, effectVariant enum type varies.
+        // The following check validates that effectVariant value is valid in relation to the applicable enum type.
+        // DelayedAllOffEffectVariantEnum or DyingLightEffectVariantEnum
+        if ((effectId == EffectIdentifierEnum::kDelayedAllOff &&
+             !isKnownEnumValue(static_cast<DelayedAllOffEffectVariantEnum>(effectVariant))) ||
+            (effectId == EffectIdentifierEnum::kDyingLight &&
+             !isKnownEnumValue(static_cast<DyingLightEffectVariantEnum>(effectVariant))))
+        {
+            status = Status::ConstraintError;
```
bzbarsky-apple, line 628: Spec says:

> If the server does not support the given variant, it SHALL use the default variant.

not to error out...

jmartinez-silabs: But this isn't an unsupported variant by the server but rather it is an undefined value per spec.

src/app/clusters/on-off-server/on-off-server.cpp
jmartinez-silabs: True for this enum, but using `isKnownEnumValue` will also compare against end up comparing against `EffectIdentifierEnum::kUnknownEnumValue` and it lines up with the following checks. 

No strong opinion on this as this creates an extra function call. I can change it

src/app/clusters/on-off-server/on-off-server.cpp
bzbarsky-apple: It's an undefined value per spec _at the point when the server was created_, right?  Or it might be a newer spec-defined thing the server does not support.  Depends on what "support" means in that spec section.

I think the intent of the thing here is that if a new variant is added in the future clients that try to use it with old servers get the default variant instead of failures.  I agree it would be clearer if the spec said that an unsupported or unknown value is treated as default...

src/app/clusters/on-off-server/on-off-server.cpp
```diff
@@ -89,6 +89,12 @@ void UpdateModeBaseCurrentModeToOnMode(EndpointId endpoint)
 
 #endif // MATTER_DM_PLUGIN_MODE_BASE
 
+template <typename EnumType>
+bool isKnownEnumValue(EnumType value)
+{
+    return (EnsureKnownEnumValue(value) == EnumType::kUnknownEnumValue) ? false : true;
```
andy31415, line 95:
```suggestion
    return (EnsureKnownEnumValue(value) != EnumType::kUnknownEnumValue);
```

src/app/clusters/on-off-server/on-off-server.cpp
jmartinez-silabs: Trying to get clarity on this with Rob 
https://github.com/project-chip/connectedhomeip/issues/34214#issuecomment-2237327207

I'll change if needed.

src/app/clusters/on-off-server/on-off-server.cpp
```diff
@@ -89,6 +89,12 @@ void UpdateModeBaseCurrentModeToOnMode(EndpointId endpoint)
 
 #endif // MATTER_DM_PLUGIN_MODE_BASE
 
+template <typename EnumType>
+bool isKnownEnumValue(EnumType value)
```
bzbarsky-apple, line 93:
```suggestion
bool IsKnownEnumValue(EnumType value)
```
is how things generally get named in this code at this point, I think....

bzbarsky-apple (approved):

PR #34404, opened 2024-07-18
Update Build on Linux (fake, gcc_release, clang, simulated) image fix clang-tidy issues

jlatusek:
## Problem

On newer version of images, clang-tidy found additional problems.

## Changes

1. Fix the problem found. A log of this is attached to this PR. 
2. Update CI image to 65


src/tools/spake2p/Cmd_GenVerifier.cpp
```diff
@@ -377,18 +382,24 @@ bool Cmd_GenVerifier(int argc, char * argv[])
...
-            return false;
+            ExitNow(ret = false);
         }
 
         // If the file with PIN codes is not provided, the PIN code on next iteration will be randomly generated.
         gPinCode = GetNextPinCode();
         // On the next iteration the Salt will be randomly generated.
         gSaltDecodedLen = 0;
     }
+exit:
```
andy31415, line 393: could we use a RAII pattern instead of a gotoexit pattern?

jlatusek: Ok, I will change that

src/inet/TCPEndPointImplSockets.cpp
```diff
@@ -127,6 +127,7 @@ CHIP_ERROR TCPEndPointImplSockets::BindImpl(IPAddressType addrType, const IPAddr
 
         if (res == CHIP_NO_ERROR)
         {
+            VerifyOrReturnValue(mSocket >= 0, CHIP_ERROR_INCORRECT_STATE);
```
bzbarsky-apple, line 130: Why is this being checked here but not above in the `reuseAddr` case ?   Either mSocket has to be valid here, or we should be checking it everywhere, no?

But also, GetSocket() returned success, so I would think it has to be valid here.

src/inet/TCPEndPointImplSockets.cpp
```diff
@@ -248,6 +249,7 @@ CHIP_ERROR TCPEndPointImplSockets::ConnectImpl(const IPAddress & addr, uint16_t
         return INET_ERROR_WRONG_ADDRESS_TYPE;
     }
 
+    VerifyOrReturnError(mSocket >= 0, CHIP_ERROR_INCORRECT_STATE);
```
bzbarsky-apple, line 252: Again, we touched mSocket above.  Why is this check here?

src/inet/UDPEndPointImplSockets.cpp
```diff
@@ -185,6 +187,7 @@ CHIP_ERROR UDPEndPointImplSockets::BindImpl(IPAddressType addressType, const IPA
 #if INET_CONFIG_ENABLE_IPV4
     else if (addressType == IPAddressType::kIPv4)
     {
+        VerifyOrReturnError(mSocket >= 0, CHIP_ERROR_INVALID_ARGUMENT);
```
bzbarsky-apple, line 190: GetSocket succeeded.

src/inet/UDPEndPointImplSockets.cpp
```diff
@@ -410,6 +413,7 @@ CHIP_ERROR UDPEndPointImplSockets::SendMsgImpl(const IPPacketInfo * aPktInfo, Sy
 #endif // INET_CONFIG_UDP_SOCKET_PKTINFO
 
     // Send IP packet.
+    VerifyOrReturnError(mSocket >= 0, CHIP_ERROR_INCORRECT_STATE);
```
bzbarsky-apple, line 416: GetSocket succeeded.

src/tools/spake2p/Cmd_GenVerifier.cpp
```diff
@@ -271,34 +271,38 @@ bool HandleOption(const char * progName, OptionSet * optSet, int id, const char
 
 } // namespace
 
+// outFile is sometimes assigned to stdout, which should not be closed.
+// Unfortunately checking condition if outFile is stdout is not understand by clang-tidy tool
+// and will always give warning. NOLINTBEGIN and NOLINTEND are used to suppress the warning.
+// NOLINTBEGIN
```
bzbarsky-apple, line 277: Can we only suppress the specific lint involved?

src/tools/spake2p/Cmd_GenVerifier.cpp
```diff
@@ -308,16 +312,17 @@ bool Cmd_GenVerifier(int argc, char * argv[])
     if (gOutFileName == nullptr)
     {
         fprintf(stderr, "Please specify the output file name, or - for stdout.\n");
-        return false;
+        ExitNow(ret = false);
```
bzbarsky-apple, line 315: So why are we doing these "exit" things before the part where we really have to worry bout `outFile`?

This would be clearer if the decl of `outFile` were moved to right before the block that sets it.  Anything above that decl can return, not exit.

bzbarsky-apple (changes requested):

src/tools/spake2p/Cmd_GenVerifier.cpp
jlatusek: @andy31415  I changed implementation of reading from file to use C++ methods.

src/tools/spake2p/Cmd_GenVerifier.cpp
jlatusek: I've change implementation as @andy31415 suggest. Now I'm using C++ methods and there is no more `exit` things

bzbarsky-apple (approved):

arkq (approved):

PR #34414, opened 2024-07-19
Move command handler registered list out of InteractionModelEngine and into a separate `Registry`

andy31415:
This is done for symmetry with the AttributeAccessInterface. The intent is to decouple usage (registering such interfaces now does not rely on the InteractionModelEngine) and allow eventual separation in DataModel processing (the processing of these are specific to data handling, same as attributes, and should not be separated out in the IME).

This has the downside that now separate IME instances cannot have separate command handler lists. I expect this to still be OK for now as for AttributeAccessInterface we did not have a separation to start with.

### Changes

- created a `CommandHandlerInterfaceRegistry` with register/unregister functions
- updated `upgrade.md` with description of what functions to use to replace the old register/unregister methods
- replaced register/unregister calls throghout our cluster and example implementations

The only odd TODO is that IME used to clear all registered command handler interfaces on shutdown, which seems somewhat reasonable for a list owned by the IME, however it does seem odd for a globally managed list. I kept that clear for now, however I expect we probably want to remove that clear in the future and eventually move it to `CodegenDataModel::Shutdown` once data models are in use.


src/app/CommandHandlerInterfaceRegistry.cpp
```diff
@@ -0,0 +1,139 @@
...
+
+namespace {
+
+CommandHandlerInterface * gCommandHandlerList = nullptr;
+
+}
+
+namespace chip {
+namespace app {
+namespace CommandHandlerInterfaceRegistry {
```
bzbarsky-apple, line 28: So this makes it impossible to have separate registries, if we ever get to the point where we can run multiple stacks in the same process.

That's probably fine for now, but just something to watch out for.  Might have been better if there were an actual object, not a namespace, so if we ever get to a point where we have a better way to get this object than just calling a global function there would be fewer updates to consumers.

src/app/InteractionModelEngine.cpp
```diff
@@ -95,20 +96,9 @@ void InteractionModelEngine::Shutdown()
...
-    while (handlerIter)
-    {
-        CommandHandlerInterface * nextHandler = handlerIter->GetNext();
-        handlerIter->SetNext(nullptr);
-        handlerIter = nextHandler;
-    }
-
-    mCommandHandlerList = nullptr;
+    // TODO: individual object clears the entire command handler interface registry.
+    //       This may not be expected.
```
bzbarsky-apple, line 100: Right, because the scoping of the registry is a bit weird; see above.

src/app/util/attribute-storage.cpp
```diff
@@ -421,10 +422,7 @@ static void shutdownEndpoint(EmberAfDefinedEndpoint * definedEndpoint)
         }
     }
 
-    // Clear out any command handler overrides registered for this
-    // endpoint.
-    chip::app::InteractionModelEngine::GetInstance()->UnregisterCommandHandlers(definedEndpoint->endpoint);
-
+    chip::app::CommandHandlerInterfaceRegistry::UnregisterAllCommandHandlersForEndpoint(definedEndpoint->endpoint);
```
bzbarsky-apple, line 425:
```suggestion
    CommandHandlerInterfaceRegistry::UnregisterAllCommandHandlersForEndpoint(definedEndpoint->endpoint);
```
I know, pre-existing....

src/controller/tests/TestServerCommandDispatch.cpp
```diff
@@ -61,10 +62,10 @@ class TestClusterCommandHandler : public chip::app::CommandHandlerInterface
 public:
     TestClusterCommandHandler() : chip::app::CommandHandlerInterface(Optional<EndpointId>::Missing(), Clusters::UnitTesting::Id)
     {
-        chip::app::InteractionModelEngine::GetInstance()->RegisterCommandHandler(this);
+        chip::app::CommandHandlerInterfaceRegistry::RegisterCommandHandler(this);
```
bzbarsky-apple, line 65:
```suggestion
        CommandHandlerInterfaceRegistry::RegisterCommandHandler(this);
```

src/controller/tests/TestServerCommandDispatch.cpp
```diff
@@ -61,10 +62,10 @@ class TestClusterCommandHandler : public chip::app::CommandHandlerInterface
...
     }
 
-    ~TestClusterCommandHandler() { chip::app::InteractionModelEngine::GetInstance()->UnregisterCommandHandler(this); }
+    ~TestClusterCommandHandler() { chip::app::CommandHandlerInterfaceRegistry::UnregisterCommandHandler(this); }
```
bzbarsky-apple, line 68:
```suggestion
    ~TestClusterCommandHandler() { CommandHandlerInterfaceRegistry::UnregisterCommandHandler(this); }
```

bzbarsky-apple (approved):

src/app/CommandHandlerInterfaceRegistry.cpp
andy31415: My objective is to decouple the CommandHandlerInterfaceRegistry/list from the InteractionModelEngine and place it to be an implementation detail of the data model handling.

After that, the IME can have its own data-model and we can run things in parallel. Agree that this is a step backwards right now ... unfortunate sideffect to prepare to have full decoupling for later. Right now we are not able to parallel run properly anyway because AttributeAccessInterface is singleton-based.

src/app/CommandHandlerInterfaceRegistry.cpp
```diff
@@ -0,0 +1,139 @@
...
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+#include <app/CommandHandlerInterfaceRegistry.h>
+
+using namespace chip::app;
+
+namespace {
+
+CommandHandlerInterface * gCommandHandlerList = nullptr;
```
tehampson, line 22: Just wondering, didn't YY work towards having the old model with IME (using GetInstance()) to move away from using a global for registered commands?

andy31415: It is a step backwards here. My reason for making the step backwards was that after that, using the data model interface is a much larger step forward: the existing CHI in the InteractionModelEngine still does not cover the attribute access interface and the data model split would.

I will ask YY about this as well.

yyzhong-g: I am okay with this if the later goal is to have this list one per IMEngine. This should still work for us temporarily since we do not manage multiple test nodes yet. The previous work I did is mostly for CommandHandlers to find the correct IMEngine and reverse is not really needed at the moment but definitely something we want in the future as well.

src/app/CommandHandlerInterfaceRegistry.cpp
bzbarsky-apple: > Agree that this is a step backwards right now 

My point is that this is an entirely avoidable step backwards, no?

andy31415: Discussed offline: this should be symmetric with AAI, but AAI choice is already bad just because legacy. We will create objects with singletons as a followup on both.

src/app/InteractionModelEngine.cpp
andy31415: Will create a singleton and will remove this clear in that PR.

tehampson (approved):

PR #34415, opened 2024-07-19
fixing type hints errors on python controller code

Alami-Amine:
in preparation of releasing chip python modules on PyPi, `mypy` is being run on the code to ensure compliance with type hint and annotation standards (such as PEP 484)

### UPDATE (Changes included):
- type hints
 - added missing `await` in  `ChipDeviceCtrl.py:221`
 - removed deprecated `property` from `ChipDeviceControllerBase`
- removed  `__init__` of dataclass EventHeader
- refactored `__init__`  of dataclass TypedAttributePath to `__post_init__`


Alami-Amine:
CI is failing with the following error: 
```
ERROR   19:00:07.633 - CHIP_REPL_YAML_TESTER ERR:   File "/__w/connectedhomeip/connectedhomeip/out/venv/lib/python3.9/site-packages/chip/crypto/p256keypair.py", line 35, in <module>
ERROR   19:00:07.633 - CHIP_REPL_YAML_TESTER ERR:     def _pychip_ECDSA_sign_msg(self_: 'P256Keypair', message_buf: pointer[c_uint8], message_size: int, signature_buf: pointer[c_uint8], signature_buf_size: pointer[c_size_t]) -> bool:
ERROR   19:00:07.633 - CHIP_REPL_YAML_TESTER ERR: TypeError: 'builtin_function_or_method' object is not subscriptable
```

## Cause 
Type annotations for `ctypes` pointers are problematic, in that they have to be of the form `POINTER[...]` instead of `POINTER(...)` (according to PEP 484).
However, type annotations of the form `POINTER[...]` (i.e. ones with subscripts) are not resolved at run time, and as such a workaround has to be added to allow both forms to work.
In addition, in newer mypy versions, support for both `POINTER[...]`  nor `pointer[...]` was discontinued and `Pointer[...]` is added instead.

more info here: https://github.com/python/mypy/issues/7540 and https://stackoverflow.com/questions/71054602/python-type-hinting-ctypes-pointer-to-x-types-in-a-way-that-mypy-accepts

cecille (approved):

agners:
>     * so the question is will we transition to the newer mypy versions (>1.0) or stay with current (0.97)?

Yes, please let's update mypy. It is safe to use a newer version, even the latest version supports older Python versions (down to Python 3.7 it seems). So updating mypy can only improve checking :+1: 

> In addition, in newer mypy versions, support for both `POINTER[...]` nor `pointer[...]` was discontinued and was replaced by `_Pointer[...]`

Does that mean we can get rid of the subscriptable pointer type work around? If so, I definitly go for mypy upgrade in this PR, and drop the work around.

If we still need the work around, we can update mypy in a separate PR.

Alami-Amine:
> Does that mean we can get rid of the subscriptable pointer type work around? If so, I definitly go for mypy upgrade in this PR, and drop the work around.
> 
> If we still need the work around, we can update mypy in a separate PR.

- So I have already integrated new mypy
-è unfortunately we still need the workaround for subscriptable pointer type --> its just that the workaround with has to be using done using `_Pointer[...]` instead of `POINTER[...]`

src/controller/python/chip/ChipDeviceCtrl.py
```diff
@@ -124,21 +127,22 @@ def _DeviceAvailableCallback(closure, device, err):
 @_IssueNOCChainCallbackPythonCallbackFunct
 def _IssueNOCChainCallbackPythonCallback(devCtrl, status: PyChipError, noc: c_void_p, nocLen: int, icac: c_void_p,
                                          icacLen: int, rcac: c_void_p, rcacLen: int, ipk: c_void_p, ipkLen: int, adminSubject: int):
-    nocChain = NOCChain(None, None, None, None, 0)
+    nocChain = NOCChain(bytes(), bytes(), bytes(), bytes(), 0)
```
agners, line 130: I think I'd prefer if we can adjust `NOCChain` field type here (make the fields optional so they can be set to None).

src/controller/python/chip/ChipDeviceCtrl.py
```diff
@@ -410,77 +426,92 @@ def HandleCommissioningComplete(nodeId: int, err: PyChipError):
...
+            self._dmLib.pychip_DeviceController_SetIcdRegistrationParameters(
+                False, None)
 
             if self._dmLib.pychip_TestCommissionerUsed():
                 err = self._dmLib.pychip_GetCompletionError()
 
             if self._commissioning_context.future is None:
-                LOGGER.exception("HandleCommissioningComplete called unexpectedly")
+                LOGGER.exception(
+                    "HandleCommissioningComplete called unexpectedly")
```
agners, line 437: Not sure what reformatted this, but we use line length of 132 by default. So this (and many other lines) should not wrap.

agners (dismissed):
In general, fixing/improving type hints is very welcome, so thank you for the PR :pray: 

It is a bit hard to review currently due to the additional wrapping. Can you reformat again with line-lenght set to 132?

One thing I'd like to consider is bumping the minimal Python version at one point. It really gets interesting with Python 3.10 [PEP 604](https://peps.python.org/pep-0604/), but maybe it is a bit early for that.

src/controller/python/chip/ChipDeviceCtrl.py
Alami-Amine: that was VSCode, I was using auopep8 without the required settings: created a PR to resolve that here: #34475

woody-apple (approved):

PR #34416, opened 2024-07-19
Fix post-review comments on Q quality utils from #34266

### LINKED ISSUE #34334: Resolve post-review minor comments on Q quality building block (#34266)
@bzbarsky-apple raised some minor comments post-submit on https://github.com/project-chip/connectedhomeip/pull/34266#pullrequestreview-2173994817.

These should be resolved.

tcarmelveilleux:
- PR #34266 had post review comments See: https://github.com/project-chip/connectedhomeip/pull/34266#pullrequestreview-2173994817
- Fix 0->0 case
- Introduce `Timestamp` more places where it makes sense
- Simplify some code lines

Fixes #34334

Testing done:

- Added unit test for 0->0
- Tests still pass

### DIFF
--- a/src/app/cluster-building-blocks/QuieterReporting.h
+++ b/src/app/cluster-building-blocks/QuieterReporting.h
@@ -172,11 +172,10 @@ public:
      * @return AttributeDirtyState::kMustReport if attribute must be marked dirty right away, or
      * AttributeDirtyState::kNoReportNeeded otherwise.
      */
-    AttributeDirtyState SetValue(const DataModel::Nullable<T> & newValue, Timestamp now,
-                                 SufficientChangePredicate changedPredicate)
+    AttributeDirtyState SetValue(const DataModel::Nullable<T> & newValue, Timestamp now, SufficientChangePredicate changedPredicate)
     {
-        bool isChangeOfNull       = newValue.IsNull() != mValue.IsNull();
-        bool areBothValuesNonNull = !newValue.IsNull() && !mValue.IsNull();
+        bool isChangeOfNull         = newValue.IsNull() != mValue.IsNull();
+        bool areBothValuesNonNull   = !newValue.IsNull() && !mValue.IsNull();
         bool areBothValuesDifferent = areBothValuesNonNull && (newValue.Value() != mValue.Value());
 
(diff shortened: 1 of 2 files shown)


andy31415 (approved):

src/app/cluster-building-blocks/QuieterReporting.h
```diff
@@ -172,13 +172,13 @@ class QuieterReportingAttribute
...
      */
-    AttributeDirtyState SetValue(const chip::app::DataModel::Nullable<T> & newValue, Timestamp now,
-                                 SufficientChangePredicate changedPredicate)
+    AttributeDirtyState SetValue(const DataModel::Nullable<T> & newValue, Timestamp now, SufficientChangePredicate changedPredicate)
     {
-        bool isChangeOfNull       = newValue.IsNull() ^ mValue.IsNull();
-        bool areBothValuesNonNull = !newValue.IsNull() && !mValue.IsNull();
+        bool isChangeOfNull         = newValue.IsNull() != mValue.IsNull();
+        bool areBothValuesNonNull   = !newValue.IsNull() && !mValue.IsNull();
+        bool areBothValuesDifferent = areBothValuesNonNull && (newValue.Value() != mValue.Value());
```
bzbarsky-apple, line 179: Hmm.  I guess we don't want to just early-return if `newValue == mValue` because we want to always give `changedPredicate` a shot at marking dirty even if the value has not changed?

I think that's worth documenting here so that someone doesn't "optimize" away that capability.  We should also clearly document that change predicates need to be aware of the fact that they might be called even if the value is not actually changing.

That said, it seems odd to me to allow this.  Why would we ever want to mark things dirty if they are not actually changing?  It might help to include some of those use-cases as examples in the API documentation for change predicates or something....

tcarmelveilleux: It's basically not to lose parity with MatterReportingAttributeChangeCallback() having been called unconditionally.

tcarmelveilleux: Fixed

bzbarsky-apple (approved):

PR #30730, opened 2023-11-30
[Darwin] Adding stub API to store data

woody-apple:
Adding stub data for this client metadata. Needs to be hooked up to storage later.

PR #34418, opened 2024-07-19
[Darwin] Client Data Store (Part 1)

kiel-apple:
Implement data storage stubbed in #30730.

### TODO
- [ ] address `REVIEWERS` comments
- [ ] address `TODO` comments
- [x] remove compile-time flag for stub functionality (never shipped as public API)
- [ ] define expected behavior when `controllerDataStore` is not present


src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -4038,84 +4038,70 @@ - (MTRBaseDevice *)newBaseDevice
...
 {
+    // TODO: KMO: check this list, maybe use list from controller DS
+    // TODO: KMO: or, uplevel this out of MTRDevice to another file for use in both places
     return @[ [NSData class], [NSString class], [NSNumber class], [NSDictionary class], [NSArray class] ];
 }
 
+#pragma mark Client Metadata Storage: By Node ID
+// REVIEWERS:  does it make sense to constrain this to non-null return and return empty array if necessary?
+// REVIEWERS:  does it make sense to constrain this to NSArray<NSString *> * since it's keys? kmo 19 jul 2024 09h28
+// TODO:  should this include all keys for node and endpoints on that node?
```
kiel-apple, line 4054: @woody-apple regarding this TODO:  the docs for the `clientDataKeys` method and the endpoint-specific keys method are the same in the header, but there was a note in the endpoint-specific keys stub implementation to "enumerate this better" when returning `allKeys` - that sounds like the intention was to return just the nodeID-indexed keys here, and return the nodeID-endpointID indexed keys from the endpoint-specific method.  is that accurate?  if so, can I update the comments for these methods in the header?

woody-apple: What's here looks good to me

Issue #21233, opened 2022-07-26
Cleanly shutdown IM engine


mrjerryjohns:
#### Problem

The InteractionModelEngine is not cleanly clearing up `mpFabricTable` and `mpExchangeMgr` today on `Shutdown`. Doing so invites a world of problems in the SDK:

1. `ReadClient` continues to depend on those members being non-null in its destruction due to its call into in the IM engine in `ReadClient::CancelLivenessCheckTimer`. This is not a problem in and of itself, but it does alter the paradigm that `ReadClient` instances can continue to exist well-past IM engine shutdown, which is implied here in `InteractionModelEngine::Shutdown`:

```
    //
    // We hold weak references to ReadClient objects. The application ultimately
    // actually owns them, so it's on them to eventually shut them down and free them
    // up.
    //
    // However, we should null out their pointers back to us at the very least so that
    // at destruction time, they won't attempt to reach back here to remove themselves
    // from this list.
    //
```

We need to fix this, and consequently, fix the various tests that depend on this premise.

2. Lots of test (`TestReadInteraction` and `TestWriteInteraction`) manually Init and Shutdown the engine repeatedly in their tests. At the end of the test, the `AppContext` does a second Shutdown of the engine as well, which causes problems since Shutdown today is not meant to be called multiple times.

mrjerryjohns:
FYI @bzbarsky-apple @yunhanw-google

PR #34419, opened 2024-07-19
Implement DataModel based read support in the interaction model engine.

andy31415:
This implements reads via ember and via codegen as well as a checked-mode where reads are validated as returning the same status codes and data size (because data content is not available)

### Changes

- Implement Read-Ember/DataModel/Checked inside `src/app/reporting`. 
- `chip_use_data_model_interface` moved to `common_flags.gni` so that the flag is accessible across various BUILD.gn files.
- `InteractionModelEngine` now maintains a `DataModel` that can be set/replaced and which will be used as the basis for reads. NOTE: this does NOT yet implement init/shutdown as I have yet to build the relevant interfaces. Those interfaces are relevant for writes and invokes mainly (ability to provide current exchange and dirty paths)
- Fixed unit tests to use the new interfaces
- Removed the data version from the read input arguments as such data is not currently needed/provided (interaction model engine seems to use the data version from iteration).

image

ReadChecked coverage is low because the "there are differences" codepaths are never hit.


src/app/reporting/Read-Checked.cpp
```diff
@@ -0,0 +1,132 @@
...
+        // Note log + chipDie instead of VerifyOrDie so that breakpoints (and usage of rr)
+        // is easier to debug.
+        ChipLogError(Test, "Different return codes between ember and DM");
+        ChipLogError(Test, "  Ember error: %" CHIP_ERROR_FORMAT, errEmber.Format());
+        ChipLogError(Test, "  DM error: %" CHIP_ERROR_FORMAT, errDM.Format());
+        chipDie();
+    }
+
+    // data should be identical for most cases EXCEPT that for time-deltas (e.g. seconds since boot or similar)
+    // it may actually differ. As a result, the amount of data written in bytes MUST be the same, however if the rest of the
```
andy31415, line 88: Technicaly not always 100% accurate: if data size changes, e.g. a stamp changing from 0xFF to 0x100 the sizes of written data will also differ.

I am unsure if we can actually detect this reliably or if it is that likely. We will use checked for internal tests only (otherwise it is both slow and bloaty), so for now leaning towards leaving it as-is.

andy31415: I updated the comment and logic.

src/app/InteractionModelEngine.cpp
```diff
@@ -1793,6 +1793,13 @@ CommandHandlerInterface * InteractionModelEngine::FindCommandHandler(EndpointId
     return nullptr;
 }
 
+InteractionModel::DataModel * InteractionModelEngine::SetDataModel(InteractionModel::DataModel * model)
```
bzbarsky-apple, line 1796: This namespacing is still really weird.  DataModel does not really live under InteractionModel....

src/app/reporting/Read-Checked.cpp
```diff
@@ -0,0 +1,140 @@
...
+    // data is not the same, we just print it out as a warning for manual inspection
+    //
+    // We have no direct access to TLV buffer data (especially given backing store splits)
+    // so for now we check that data length was identical.
+    //
+    // NOTE: RetrieveClusterData is responsible for encoding StatusIB errors in case of failures
+    //       so we validate length written requirements for BOTH success and failure.
+    //
+    // NOTE: data length is NOT reliable if the data content differs in encoding lentgh. E.g. numbers changing
```
bzbarsky-apple, line 97:
```suggestion
    // NOTE: data length is NOT reliable if the data content differs in encoding length. E.g. numbers changing
```

src/app/reporting/Read-Checked.cpp
```diff
@@ -0,0 +1,140 @@
...
+        {
+            if (encoderState->AllowPartialData() != stateDm.AllowPartialData())
+            {
+                ChipLogError(Test, "Different partial data");
+                chipDie();
+            }
+            if (encoderState->CurrentEncodingListIndex() != stateDm.CurrentEncodingListIndex())
+            {
+                ChipLogError(Test, "Different partial data");
+                chipDie();
```
bzbarsky-apple, line 126: This could get hit if you have a list of time-dependent values, so in one version they fit in the packet and in the other they do not....

src/app/reporting/Read-DataModel.cpp
```diff
@@ -0,0 +1,97 @@
...
+    TLV::TLVWriter checkpoint;
+    reportBuilder.Checkpoint(checkpoint);
+
+    AttributeValueEncoder attributeValueEncoder(reportBuilder, subjectDescriptor, path, version, isFabricFiltered, encoderState);
+    CHIP_ERROR err = dataModel->ReadAttribute(readRequest, attributeValueEncoder);
+
+    if (err == CHIP_NO_ERROR)
+    {
+#if !CHIP_CONFIG_USE_EMBER_DATA_MODEL
+        DataModelCallbacks::GetInstance()->AttributeOperation(DataModelCallbacks::OperationType::Read,
```
bzbarsky-apple, line 75: I know this preserves the old logic for when this callback is called, but that logic seems off to me....  Separate PR, though.

src/app/reporting/Read-DataModel.cpp
```diff
@@ -0,0 +1,97 @@
...
+#endif // !CHIP_CONFIG_USE_EMBER_DATA_MODEL
+        return CHIP_NO_ERROR;
+    }
+
+    if (IsOutOfSpaceError(err) || err == CHIP_IM_GLOBAL_STATUS(InvalidDataType))
+    {
+        // Only update state for retryable errors. This matches logic in legacy ember compatibility layer
+        if (encoderState != nullptr)
+        {
+            *encoderState = attributeValueEncoder.GetState();
```
bzbarsky-apple, line 86: This does not make sense to me.  The "ember" code sets the state on all errors from AttributeAccessInterface (and never in other cases), right?  Why are we only doing it for some errors here?

src/app/reporting/Read-DataModel.cpp
```diff
@@ -0,0 +1,97 @@
...
+        }
+    }
+
+    ChipLogError(DataManagement, "Failed to read attribute: %" CHIP_ERROR_FORMAT, err.Format());
```
bzbarsky-apple, line 90: For a chunked read, this log is _really_ misleading.

src/app/reporting/Read-DataModel.cpp
andy31415: I did this for unit testing. Unfortunately interface-wise it makes no sense for the data provider to have information on "this is AAI or not".

I updated these to work and pass tests, will have to revert and add some descriptions. In my mind the correct thing to do would be to provide some guarantees/description like:
   - encoder state always forwarded (even if not useful)
   - encoder state is only for out of space, so it should be forwarded for out of space errors only
   
The logic of `encoder state forwarded if AAI is used, otherwise if ember is used it is not` ties our interface to implementation details (i.e. ember) and I believe that is not ok.

src/app/InteractionModelEngine.cpp
andy31415: It is bad (and forced separate PRs due to clash with protocols::InteractionModel).

Not sure what a better place is though and replacing should probably be a separate change for rename only. If you have a better name idea I can make a separate PR. Otherwise I am leaning towards defering that while the implementations are on-going.

src/app/reporting/Read-Checked.cpp
andy31415: Moved to have all chipDie to unit tests only and added comments.

andy31415:
I am unclear why casting tests fail... tried to debug locally and got #34443 and #34444.

Will keep debugging however for now it seems like some form of race or minmdns/dnssd issue could be at fault.

andy31415:
I manually tested tv app + casting locally on my machine and I do get `Secure Pairing Success` with updated code. However when I run the python test, things fail (app locks up on `controller ux ok`)

src/app/InteractionModelEngine.cpp
bzbarsky-apple: Well, we could have `chip::app::DataModel`, but that's an existing namespace.

We could call the class `chip::app::DataModel::Instance`?

src/app/reporting/Read-DataModel.cpp
bzbarsky-apple: The problem is exactly that the post read is not called on failure, so you can get pre-read a few times, then a post-read, and how do you know what that means?

More generally: what is the intent of these callbacks?  And does calling this only on success serve that intent?

src/app/reporting/Read-DataModel.cpp
bzbarsky-apple: No, my question is: why are we doing _any_ restriction on the error here?  Just pass through all errors as-is.

andy31415: Unit tests fail if I do that (because ember currently does not do it). Will run a test run to check which.

bzbarsky-apple: > because ember currently does not do it

It sure does as far as I can tell.

bzbarsky-apple: Er, note that this is about resetting the encoder state on all errors, to be clear.

andy31415: Apparently this used to fail, maybe when I updated on both failure and success or before I fixed some code paths. Updated to always set the state on error.

src/app/reporting/Read-DataModel.cpp
andy31415: Added TODO. We can clean it up later (and probably much easier given the current placement).

src/app/InteractionModelEngine.cpp
andy31415: `Instance` is not a good class name because we have a lot of singleton Instance methods. `Provider` ?

bzbarsky-apple: Provider seems fine.

Separate PR for this part, obviously.

andy31415: Will have a followup on this.

bzbarsky-apple (approved):

src/controller/tests/data_model/DataModelFixtures.cpp
```diff
@@ -464,5 +485,114 @@ Protocols::InteractionModel::Status ServerClusterCommandExists(const ConcreteCom
...
+{
+    return CHIP_ERROR_NOT_IMPLEMENTED;
+}
+
+CHIP_ERROR CustomDataModel::Invoke(const InvokeRequest & request, chip::TLV::TLVReader & input_arguments, CommandHandler * handler)
+{
+    return CHIP_ERROR_NOT_IMPLEMENTED;
+}
+
+EndpointId CustomDataModel::FirstEndpoint()
```
cecille, line 532: all this seems really similar to the test interaction-model-api stuff above - would you mind adding some comments about how this is intended to be used vs. the other one?

cecille (approved):

src/app/InteractionModelEngine.h
```diff
@@ -408,6 +408,13 @@ class InteractionModelEngine : public Messaging::UnsolicitedMessageHandler,
 
     InteractionModel::DataModel * GetDataModel() const;
 
+    // MUST NOT be used while the interaction model engine is running as interaction
```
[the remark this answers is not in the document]
andy31415: I am looking at it ... but for example `Init` claims we have a kState_NotInitialized however there is no such thing in practice (yay for refering to private details when describing return errors...).

So unsure if there is a proper way to do this. We could check if Init was called by checking some members, but comments claim that #21233 tracks that we cannot have clean shutdowns :(

Will spend some time on it, but I may not be able to find something reasonable given that IM lifetime seems odd overall.

src/app/BUILD.gn
```diff
@@ -260,6 +251,29 @@ static_library("interaction-model") {
 
   public_configs = [ "${chip_root}/src:includes" ]
 
+  if (chip_use_data_model_interface == "disabled") {
+    sources += [
+      "reporting/Read-Ember.cpp",
+      "reporting/Read-Ember.h",
+    ]
+  } else if (chip_use_data_model_interface == "check") {
```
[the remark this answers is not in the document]
andy31415: Totally not for production. I enabled it in linux for CI really, however in production we should use one or the other.

src/app/InteractionModelEngine.h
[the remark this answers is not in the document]
andy31415: For now added a check of `read handlers are empty` since read handlers are the ones that use iteration in general and should have some state.

Not ideal, but better than nothing.

src/app/BUILD.gn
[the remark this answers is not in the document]
andy31415: Added a warning in IME::Init, printed out as an error so it always shows up (hopefully)

PR #34420, opened 2024-07-19
Update Fabric Synchronization Guide to avoid confusion

yufengwangca:


saurabhst (approved):

docs/guides/fabric_synchronization_guide.md
```diff
@@ -82,11 +85,17 @@ fabricsync enable-auto-sync 1
...
+fabricsync add-bridge 1 <fabric-source-bridge-ip>
 ```
 
 ### Pair Light Example to Fabric-Source
 
+Since Fabric-Bridge also functions as a Matter server, running it alongside the
+Light Example app on the same machine would cause conflicts. Therefore, you need
+to run the Matter Light Example app on a separate physical machine from the one
+hosting Fabric-Admin and Fabric-Bridge. You can then commission the Matter Light
+Example app using Fabric-Admin on the source side.
```
samadDotDev, line 97: This can be worked around by using a different port and also not reusing the same default KVS path for multiple apps on the same machine. I have been running the following for instance:

```sh
# Bridge App with non-default discriminator/passcode and its own KVS path
./out/linux-x64-fabric-bridge-no-ble-clang-rpc/fabric-bridge-app --discriminator 3842 --passcode 20202022 --KVS /tmp/chip_kvs_fabric_bridge_app

# Lock App with different discriminator/passcode, ports and KVS
./out/linux-x64-lock-clang/chip-lock-app --discriminator 3846 --passcode 20202026 --secured-device-port 5546 --unsecured-commissioner-port 5556 --KVS /tmp/chip_kvs_lock_app

# Light App with yet another set of different discriminator/passcode, ports and KVS
./out/linux-x64-light-clang/chip-lighting-app --discriminator 3843 --passcode 20202023 --secured-device-port 5543 --unsecured-commissioner-port 5553 --KVS /tmp/chip_kvs_lighting_app
```

samadDotDev (approved):

docs/guides/fabric_synchronization_guide.md
yufengwangca: Added, thanks for the tip

arkq (approved):

joonhaengHeo (approved):

Issue #34421, opened 2024-07-19
Implement Q on operationalstate CountdownTime


tcarmelveilleux:
Actually implement reporting on CountdownTime whenever edges arise, and at most every 10s when requested by delegate.

PR #34422, opened 2024-07-19
Add Q quality to OperationalState CountdownTime

tcarmelveilleux:
Add Q quality to OperationalState CountdownTIme
- Update operational state cluster server to report countdown time at edges.
- Add a way for cluster delegate to request an update of time.

Issue #34421

Testing done:

- Existing tests still pass
- Will add cert test when the text is finalized (week of July 22)


src/app/cluster-building-blocks/QuieterReporting.h
```diff
@@ -188,13 +180,17 @@ class QuieterReportingAttribute
...
+        // Only execute predicate on value change from last marked dirty.
+        if (newValue != mLastDirtyValue)
+        {
+            SufficientChangePredicateCandidate candidate{
+                mLastDirtyTimestampMillis, // lastDirtyTimestamp
+                now,                       // nowTimestamp
+                mLastDirtyValue,           // lastDirtyValue
+                newValue                   // newValue
+            };
+            isNewlyDirty = isNewlyDirty || changedPredicate(candidate);
```
tcarmelveilleux, line 192: These changes on QuieterReporting are in #34416. When that one lands, we can merge master here.

src/app/clusters/operational-state-server/operational-state-server.cpp
```diff
@@ -177,6 +191,43 @@ void Instance::ReportOperationalStateListChange()
...
+            uint32_t lastDirtyValue = candidate.lastDirtyValue.Value();
+            uint32_t newValue = candidate.newValue.Value();
+            uint32_t kNumSecondsDeltaToReport = 10;
+            return (newValue < lastDirtyValue) && ((lastDirtyValue - newValue) > kNumSecondsDeltaToReport);
+        };
+        markDirty = (mCountdownTime.SetValue(newCountdownTime, now, predicate) == AttributeDirtyState::kMustReport);
+    }
+    else
+    {
+        auto predicate = [](const decltype(mCountdownTime)::SufficientChangePredicateCandidate &) -> bool { return true; };
```
bzbarsky-apple, line 223: So all cluster-logic updates, including for random derived commands, force reporting?

src/app/clusters/operational-state-server/operational-state-server.h
```diff
@@ -192,6 +200,18 @@ class Instance : public CommandHandlerInterface, public AttributeAccessInterface
...
+    /**
+     * Causes reporting/udpating of CountdownTime attribute from driver if sufficient changes have
+     * occurred (based on Q quality definition for operational state). Calls the Delegate::GetCountdownTime() method.
+     *
+     * @param fromDelegate true if the change notice was triggered by the delegate, false if internal to cluster logic.
+     */
+    void UpdateCountdownTime(bool fromDelegate);
+
+    /**
+     * @brief Whenever the cluster logic things time should be updated, call this.
```
bzbarsky-apple, line 212:
```suggestion
     * @brief Whenever the cluster logic thinks time should be updated, call this.
```

src/app/clusters/operational-state-server/operational-state-server.h
```diff
@@ -262,9 +283,10 @@ class Delegate
     virtual ~Delegate() = default;
 
     /**
-     * Get the countdown time.
-     * NOTE: Changes to this attribute should not be reported.
-     * From the spec: Changes to this value SHALL NOT be reported in a subscription.
+     * Get the countdown time. This will get called on many edges such as
+     * commands to change operational state, or whne the delegate deals with
```
bzbarsky-apple, line 287:
```suggestion
     * commands to change operational state, or when the delegate deals with
```

src/app/clusters/operational-state-server/operational-state-server.cpp
tcarmelveilleux: They do not force reporting, unless there was a change. Since your request in the QuieterReporting PR was to only take the `true` return of predicate to mean "report for sure on change, with no additional checks other than change". In other words, this is considered, if called from internal logic, as the internal knowledge of the cluster.

src/app/clusters/operational-state-server/operational-state-server.h
tcarmelveilleux: Done

src/app/clusters/operational-state-server/operational-state-server.h
tcarmelveilleux: Done

src/app/clusters/operational-state-server/operational-state-server.cpp
bzbarsky-apple: No, here's what I mean.  Let's say a random derived-cluster command gets executed.  Since in that case `Instance::InvokeCommand` calls `UpdateCountdownTimeFromClusterLogic()`, we will report the CountdownTime if it's changed at all from the last-reported value, no matter how much time has passed since the last report, and completely ignoring the actual `kMarkDirtyOnIncrement | kMarkDirtyOnChangeToFromZero` policy.  In particular, we will report even if all that happened is the time ticked down normally.

What the spec says is:

> Changes to this attribute merely due to the normal passage of time with no other dynamic change of device state SHALL NOT be reported.

If a derived cluster command does not involve any change of device state (e.g. it's just a data request), then we would be violating that part of the spec.

I actually kind of have similar questions about the call in `OnOperationalErrorDetected` (spec doesn't seem to say that should report, if it didn't change the CountdownTime to 0 or null?), `OnOperationCompletionDetected` (again, if the phase/state did not change in the process should this really report?) and `ReportPhaseListChange` (why would this on its own trigger a report?).

And a similar concern in `HandleStartState`.  If we are currently running, why would receiving that command (which is a no-op in that situation) trigger an immediate report of the CountdownTime (yes, assuming that's changed since the last report)?

tcarmelveilleux: Sure, I could report less. You are basically assuming ill intent from the cluster developer here. What would you recommend we do? I can take out quite a few of these, but I needed to at least have the edges of operational states report current time if changed (even if by one), as that is from important edges. In practice they won't change every second.

bzbarsky-apple: I'm not assuming ill intent at all. I am assuming the cluster developer is trying their best to get correct behavior for their derived cluster.  But they can't control reporting because we are overriding it here.

And again: even within just this cluster, the reporting is not correct in `HandleStartState` already.

> but I needed to at least have the edges of operational states report current time if changed (even if by one),

That gets handled by the fact that a change in the operational state value triggers reporting, no?

tcarmelveilleux: Left them out from commands. Fixed.

src/app/clusters/operational-state-server/operational-state-server.cpp
```diff
@@ -179,6 +195,43 @@ void Instance::ReportOperationalStateListChange()
...
+        auto predicate = [](const decltype(mCountdownTime)::SufficientChangePredicateCandidate & candidate) -> bool {
+            if (candidate.lastDirtyValue.IsNull() || candidate.newValue.IsNull())
+            {
+                return false;
+            }
+
+            uint32_t lastDirtyValue           = candidate.lastDirtyValue.Value();
+            uint32_t newValue                 = candidate.newValue.Value();
+            uint32_t kNumSecondsDeltaToReport = 10;
+            return (newValue < lastDirtyValue) && ((lastDirtyValue - newValue) > kNumSecondsDeltaToReport);
```
bzbarsky-apple, line 221: I do wonder why this is not just comparing `lastDirtyTimestamp` to `nowTimestamp`, but I guess this is to catch the timer decreasing faster than wall-clock time?  If so, worth a comment explaining that.

src/app/clusters/operational-state-server/operational-state-server.cpp
```diff
@@ -399,6 +453,8 @@ void Instance::HandleStopState(HandlerContext & ctx, const Commands::Stop::Decod
     response.commandResponseState = err;
 
     ctx.mCommandHandler.AddResponse(ctx.mRequestPath, response);
+
+    UpdateCountdownTimeFromClusterLogic();
```
bzbarsky-apple, line 457: I think this should not be here either, at least not if `opState == to_underlying(OperationalStateEnum::kStopped)` so we never call into the delegate and never change state....

And if we _do_ call into the delegate and change state, won't that trigger this reporting anyway?

bzbarsky-apple (approved):

src/app/clusters/operational-state-server/operational-state-server.cpp
tcarmelveilleux, line 457:
```suggestion
```

tcarmelveilleux: Done

cecille (approved):

src/python_testing/TC_OpstateCommon.py
```diff
@@ -354,7 +354,7 @@ async def TEST_TC_OPSTATE_BASE_2_1(self, endpoint=1):
             if phase_list is not NullValue:
                 phase_list_len = len(phase_list)
                 asserts.assert_less_equal(phase_list_len, 32,
-                                          f"PhaseList length({phase_list_len}) must be less than 32!")
+                                          f"PhaseList length({phase_list_len}) must be at most than 32 entries!")
```
rbultman, line 357:
```suggestion
                                          f"PhaseList length({phase_list_len}) must be at most 32 entries!")
```

rbultman (approved):

src/app/clusters/operational-state-server/operational-state-server.cpp
```diff
@@ -94,9 +98,11 @@ CHIP_ERROR Instance::SetOperationalState(uint8_t aOpState)
         return CHIP_ERROR_INVALID_ARGUMENT;
     }
 
+    bool runUpdateCountdownTime = false;
```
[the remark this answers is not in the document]
tcarmelveilleux: Fixed

src/python_testing/TC_OpstateCommon.py
tcarmelveilleux: Fixed

PR #34424, opened 2024-07-19
TC-IDM-10.5: Device type conformance - Add

cecille:
Initial test only looks at clusters. Remaining
- revisions
- feature conformance
- cluster elements


examples/lock-app/lock-common/lock-app.matter
```diff
@@ -2502,6 +2502,7 @@ cluster DoorLock = 257 {
 endpoint 0 {
   device type ma_rootdevice = 22, version 1;
   device type ma_powersource = 17, version 1;
+  device type ma_otarequestor = 18, version 1;
```
bzbarsky-apple, line 2505: Does this app actually implement OTA provider client and OTA requestor server in a sane way?  When I was checking it looked like nothing other than the ota-requestor app actually did any of the OTA bits, though some other apps have them enabled.

cecille: I'm not sure. Do you think it would be better to remove the cluster?

cecille: @fessehaeve - do you know? Does this app do anything real with OTA or is this cluster just legacy?

cecille: I'm going to assume legacy - remove both the device type and the cluster, test passes.

bzbarsky-apple (commented):
The lock-app changes look good to me; someone who is closer to the Python testing stuff than I am should review those changes.

examples/lock-app/lock-common/lock-app.matter
fessehaeve: I guess you are referring to the Linux app? I know that you can use the nrfconnect version with real OTA support. I don't think all platforms have real support.

fessehaeve: Btw why does it matter whether the app has real OTAU capability for this test? Isn't the main need only to have ota provider client and ota requestor server clusters? Wouldn't that make this test pass?

lpbeliveau-silabs (approved):

tcarmelveilleux (approved):

PR #34425, opened 2024-07-19
Add target endpoint to `CommissioningWindowOpener`

### LINKED ISSUE #33798: [Feature] Allows opening a commissioning window on a bridged node that is not on Endpoint 0.
### Feature description

...
FabricSync spec changes were merged,
https://github.com/CHIP-Specifications/connectedhomeip-spec/blob/master/src/multi_admin/fabric_synchronization.adoc

With the spec update, AdministratorCommissioningCluster is not always on endpoint 0, to open Commissioning Window on bridged node, we need to specify the endpointID

### Platform

all

### Platform Version(s)

_No response_

### Anything else?

_No response_

samadDotDev:
Revives https://github.com/project-chip/connectedhomeip/pull/33799 after the params API refactor.

I'm not sure if this PR needs to be strictly blocked on the issue raised in https://github.com/CHIP-Specifications/connectedhomeip-spec/issues/9354 since this is just adding the endpoint parameter for `CommissioningWindowOpener` to target when sending out ACC (`AdminCommissioningCluster`) commands instead of always sending to root as being done currently, but I'll follow the conversation there to ensure if this needs to be gated by specific feature bits.

The behavior of restricting endpoint ID to root in server handling of ACC command has been intentionally left out since that's definitely not spec-compliant as discussed [here](https://github.com/project-chip/connectedhomeip/pull/33799#discussion_r1631415487). Please correct my understanding if that's not the case or if there needs to be a different requirement for it.

For context: The 2-way Eco-to-Eco Fabric sync has a use case of a fabric-admin containing fabric-bridge as one of its connected devices. The requirement is to be able to direct the OCW command at a specific endpoint of the bridge [`(bridgeNodeId, bridgedEndpointId)`] which will then be expected to be translated and forwarded to the relevant bridged node [`(bridgedNodeId, rootOrAccEndpointId)`] by the bridge.

Fixes #33798


yufengwangca (approved):
Previous https://github.com/project-chip/connectedhomeip/pull/33799 is pending on whether to add the Endpoint parameter to the OpenCommissioningWindow API. This parameter is currently already part of the CommissioningWindowPasscodeParams. I believe we should avoid hardcoding it to kRootEndpointId in the SDK, particularly since the application can now set this Endpoint parameter via the public API.

@bzbarsky-apple

tehampson (approved):

src/controller/CommissioningWindowOpener.cpp
```diff
@@ -126,6 +126,10 @@ CHIP_ERROR CommissioningWindowOpener::OpenCommissioningWindow(const Commissionin
     mCommissioningWindowVerifierCallback = nullptr;
     mNodeId                              = params.GetNodeId();
     mCommissioningWindowTimeout          = params.GetTimeout();
+    if (params.HasEndpointId())
+    {
+        mTargetEndpointId = params.GetEndpointId();
```
bzbarsky-apple, line 131: This looks wrong.  A CommissioningWindowOpener object can be reused, and this is leaking state from the previous params, no?

The right thing to do is that either this code resets to kRootEndpointId, or the params return that if they have nothing else specified.  The latter seems nicer to me, since the params can just always have an endpoint id, defaulting to kRootEndpointId.

bzbarsky-apple (changes requested):

src/controller/CommissioningWindowOpener.cpp
samadDotDev: Great catch! The params now default to this: https://github.com/project-chip/connectedhomeip/pull/34425/commits/badd7f4e6f357476d446ea5d01910b4faa50c55b

bzbarsky-apple (approved):

PR #33968, opened 2024-06-18
network-manager-app: Add support for OpenThread BR integration on OpenWrt

ksperling-apple:
### Summary

Add a new ThreadBROpenThreadUbus implementation of the TBRM delegate that interacts with otbr-agent via ubus. Rename the existing dummy implementation to ThreadBRFake.h.

Add some generic utilities for interacting with ubus (and the underlying uloop / ubox libraries) from C++. At the moment this lives under the network-manager example directory, but could move into platform/Linux in the core SDK at a later stage. (Future work based on this is likely to include interacting with hostapd via ubus, as well as exposing some information about the matter application itself, e.g. pairing status.)

### Testing

Tested manually on OpenWrt for now, since this integration code is not easy to meaningfully test in unit tests.

PR #34433, opened 2024-07-22
Add the ability for EventLoopHandlers to participate in a Select-based event loop

ksperling-apple:
`EventLoopHandler`s can be registered with a `LayerSocketsLoop` instance to enable participation of those handlers in the processing cycle of the event loop. This makes it possible to implement adapters that allow components utilizing a third-party event loop API (e.g. ubus) to participate in the Matter event loop, instead of having to run an entirely separate event loop on another thread.

This was split out of #33968.


src/lib/support/IntrusiveList.h
```diff
@@ -414,6 +414,11 @@ class IntrusiveList : public IntrusiveListBase
         ConstIterator(IntrusiveListBase::ConstIteratorBase && base) : IntrusiveListBase::ConstIteratorBase(std::move(base)) {}
         const T * operator->() { return Hook::ToObject(mCurrent); }
         const T & operator*() { return *Hook::ToObject(mCurrent); }
+
+        ConstIterator & operator++() { return static_cast<ConstIterator &>(IntrusiveListBase::ConstIteratorBase::operator++()); }
+        ConstIterator operator++(int) { return IntrusiveListBase::ConstIteratorBase::operator++(1); }
```
bzbarsky-apple, line 419: How come this and the `--(int)` version do not need static casts on the return type?  Same thing below.

bzbarsky-apple (approved):

src/lib/support/IntrusiveList.h
ksperling-apple: They return a new object rather than `this`, and that new object is actually of the base class.
Pretty sure this ends up using the `ConstIterator(IntrusiveListBase::ConstIteratorBase && base)` move constructor.

src/system/SystemLayerImplSelect.cpp
```diff
@@ -608,6 +610,32 @@ SocketEvents LayerImplSelect::SocketEventsFromFDs(int socket, const fd_set & rea
     return res;
 }
 
+#if !CHIP_SYSTEM_CONFIG_USE_DISPATCH
+enum : intptr_t
```
andy31415, line 614: why is the state a intptr_t ? why not plain int or named enum?

andy31415: Once this decision is done, we have `auto` and `decltype` everywhere even though it is not obvious that a `GetState`-like call results in a intptr_t

andy31415 (dismissed):
Please add documentation to the PR summary for me to understand what this PR is about before reading it. the title of Add the ability for EventLoopHandlers to participate ... does not explain what event loop handlers are, why we have them.

Also overall there should be better documentation in the files and eventloophandler seems stand-alone and should be a separate header.

src/platform/Darwin/PlatformManagerImpl.h
```diff
@@ -58,8 +68,8 @@ class PlatformManagerImpl final : public PlatformManager, public Internal::Gener
 private:
     // ===== Methods that implement the PlatformManager abstract interface.
     CHIP_ERROR _InitChipStack();
-    void _Shutdown();
 
+#if CHIP_SYSTEM_CONFIG_USE_DISPATCH
```
andy31415, line 72: why was the shutdown removed? The PR description does not say anything about it.

If it needs to be removed because unused, why not a separate PR since it looks like very independent

src/system/SystemLayer.h
```diff
@@ -288,6 +290,32 @@ class LayerSockets : public Layer
     virtual SocketWatchToken InvalidSocketWatchToken() = 0;
 };
 
+class LayerSocketsLoop;
+
+/**
+ * Enables the participation of subordinate event loops in the SystemLayer event loop.
```
andy31415, line 296: Needs more documentation. What is a subordinate event loop? What are usage examples?

src/system/SystemLayer.h
```diff
@@ -288,6 +290,32 @@ class LayerSockets : public Layer
...
+ */
+class EventLoopHandler : public chip::IntrusiveListNodeBase<>
+{
+public:
+    virtual ~EventLoopHandler() {}
+
+    /**
+     * Prepares events and returns the next requested wake time.
```
andy31415, line 304: what events? wake up time to do what? Overall I need more documentation on this class.

Why is this in `SystemLayer` and not a separate header? Please do not increase the size of headers.

src/system/SystemLayer.h
```diff
@@ -288,6 +290,32 @@ class LayerSockets : public Layer
...
+    virtual Clock::Timestamp PrepareEvents(Clock::Timestamp now) { return Clock::Timestamp::max(); }
+
+    /**
+     * Handles / dispatches pending events.
+     * Every call to this method will have been preceded by a call to `PrepareEvents`.
+     */
+    virtual void HandleEvents() = 0;
+
+private:
+    friend class LayerSocketsLoop;
```
andy31415, line 315: why are you using friend classes? Friend classes create strong coupling and should be generally avoided. If we use friend, please explain in a comment. "Because others may mis-use it" is not a great reason (e.g. we probably want some form of unit tests which often want internal state access).

src/system/SystemLayer.h
```diff
@@ -298,13 +326,22 @@ class LayerSocketsLoop : public LayerSockets
...
     virtual void SetDispatchQueue(dispatch_queue_t dispatchQueue) = 0;
     virtual dispatch_queue_t GetDispatchQueue()                   = 0;
 #elif CHIP_SYSTEM_CONFIG_USE_LIBEV
     virtual void SetLibEvLoop(struct ev_loop * aLibEvLoopP) = 0;
     virtual struct ev_loop * GetLibEvLoop()                 = 0;
 #endif // CHIP_SYSTEM_CONFIG_USE_DISPATCH/LIBEV
+
+protected:
+    // Expose EventLoopHandler.mState to sub-classes
+    decltype(EventLoopHandler::mState) & LoopHandlerState(EventLoopHandler & handler) { return handler.mState; }
```
andy31415, line 344: Why doesn't the EventLoopHandler provide a `GetState() const`? Seems safe enough.

src/system/SystemLayer.h
ksperling-apple: It's explained in the comment one line below. It's so the event loop implementation can use `mState`. I can spell it out more explicitly.

src/system/SystemLayer.h
ksperling-apple: It's not const because the event loop implementations needs to write as well as read it. By using the friend class and this protected helper I'm making it only available to `LayerSocketsLoop` sub-classes, without having to list implementation classes explicitly as friends.

src/platform/Darwin/PlatformManagerImpl.h
ksperling-apple: It was removed because it simply called the inherited method, which is the same as not overriding the method at all. I don't think this warrants a separate PR.

src/system/SystemLayer.h
ksperling-apple: This is similar to how `IntrusiveListNodePrivateBase` contains state (`mPrev` and `mNext`) for use by `IntrusiveList`, except here `LayerSocketsLoop` is abstract, so the state is really used by the concrete implementation class.

ksperling-apple:
> Please add documentation to the PR summary for me to understand what this PR is about before reading it. the title of Add the ability for EventLoopHandlers to participate ... does not explain what event loop handlers are, why we have them.

(Sorry accidentally ended up editing the comment when I meant to hit quote reply)

As far as understanding the context goes, in my opinion it would have been easier to see what the purpose of an EventLoopHandler is by seeing an example of how one is used. This is why I originally included this code in #33968, but broke it out into a separate PR based on your request.

I'm adding more documentation as requested and investigating the test failure.

src/system/SystemLayer.h
ksperling-apple: This is not in a separate header because it belongs quite tightly to `LayerSocketsLoop`. I've clarified this in the added documentation comments.

src/system/SystemLayerImplSelect.cpp
ksperling-apple: It's not a named enum because it's really up to the specific implementation class what gets stored there (though at the moment there's only one implementation class). In practice there aren't going to be a large number of these handlers, and due to alignment the memory cost of an int or an enum are going to be the same anyway. Using an intptr_t gives the implementation the ability to store a pointer if necessary.

src/system/SystemLayer.h
ksperling-apple: I've updated the documentation.

jmartinez-silabs (approved):

PR #34439, opened 2024-07-22
bumping mypy python package version to 1.10.1 in venv

Alami-Amine:
- bumping mypy python package version to 1.10.1 in venv from 0.971 to 1.10.1 
- this was done because there have been many changes in the versions as well as improvements and to make sure that mypy fixing activities will be inline with latest version.

- this was discussed in #34415


cecille (approved):

andy31415:
Fast track: infra change and CI passes.

PR #34447, opened 2024-07-22
[OO] Scenes Integration Tests

lpbeliveau-silabs:
Added test for scenes integration in onoff cluster.

Note:
Some extra steps had to be added due to the integration with the level control cluster, they will be added to the test plan repo once this PR is approved.

### DIFF
--- a/src/app/tests/suites/certification/Test_TC_OO_2_7.yaml
+++ b/src/app/tests/suites/certification/Test_TC_OO_2_7.yaml
@@ -247,3 +247,23 @@ tests:
       attribute: "OnOff"
       response:
           value: 0
+
+    - label:
+          "Cleanup: TH sends a RemoveAllScenes command to DUT with the GroupID
+          field set to G1."
+      command: "RemoveAllScenes"
+      arguments:
+          values:
+              - name: "GroupID"
+                value: G1
+      response:
+          values:
+              - name: "Status"
+                value: 0x00
+              - name: "GroupID"
+                value: G1
+
+    - label: "Cleanup: TH sends a RemoveAllGroups command to DUT."
+      cluster: "Groups"
+      endpoint: endpoint
+      command: "RemoveAllGroups"
--- a/src/app/tests/suites/certification/Test_TC_OO_2_7.yaml
+++ b/src/app/tests/suites/certification/Test_TC_OO_2_7.yaml
@@ -1,0 +1,314 @@
+# Copyright (c) 2024 Project CHIP Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


src/app/tests/suites/certification/Test_TC_OO_2_7.yaml
```diff
@@ -0,0 +1,314 @@
+# Copyright (c) 2021 Project CHIP Authors
```
cecille, line 1:
```suggestion
# Copyright (c) 2024 Project CHIP Authors
```

src/app/tests/suites/certification/Test_TC_OO_2_7.yaml
```diff
@@ -0,0 +1,314 @@
...
+          values:
+              - name: "Status"
+                value: 0x00
+              - name: "GroupID"
+                value: G1
+              - name: "SceneList"
+                value: []
+
+    - label: "Step 2a: TH sends Off command to DUT"
+      PICS: OO.S && !LVL.S.C04.Rsp
```
cecille, line 126: suggest removing the OO.S here - it's already a top level PICS.

And on the next step.

src/app/tests/suites/certification/Test_TC_OO_2_7.yaml
```diff
@@ -0,0 +1,314 @@
...
+              - name: "SceneID"
+                value: 0x01
+
+    - label: "Step 3: TH sends a AddScene command to DUT with the GroupID field
+          set to G1, the SceneID field set to 0x02, the TransitionTime field set
+          to 1000 (1s) and the ExtensionFieldSets set to: '[{ ClusterID: 0x0006,
+          AttributeValueList: [{ AttributeID: 0x0000, ValueUnsigned8: 0x01 }]}]'
+          "
+      command: "AddScene"
+      ## The default handlers for level control and on/off assume that if both clusters are in the same endpoint,
```
cecille, line 172: Is that specified somewhere? If someone comes with a different interpretation, will they fail cert and is that as expected?

src/app/tests/suites/certification/Test_TC_OO_2_7.yaml
```diff
@@ -0,0 +1,314 @@
...
+      ## the levvel control cluster will use the onoff with level command and thus onoff will be turned on.
```
cecille, line 173:
```suggestion
      ## the level control cluster will use the onoff with level command and thus onoff will be turned on.
```

src/app/tests/suites/certification/Test_TC_OO_2_7.yaml
lpbeliveau-silabs: That is how it is implemented in the SDK as off now, so I do not expect a different interpretation. 

That being said it isn't specified so I added a Spec PR here:
https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/10114

We can have this as a followup, or should we leave the PR on hold?

cecille: I wouldn't put it past matter.js to be ON it with scenes, btw. But the spec clarification works. As long as they both get it, let's not hold this up on that.

cecille (approved):

jmartinez-silabs (approved):

PR #34450, opened 2024-07-22
Add TransportPayloadCapability flag for GetConnectedDevices() API in ChipDeviceCtrl.py and Script bindings

pidarped:
Add indicator flag for large payloads to GetConnectedDevices() API and bubble up the flag to the wrapper IM Python APIs.

Add python script binding methods required for upcoming TCP/LargePayload tests 
--to check if session allows large payload.
--to close the underlying TCP connection.
--to check if the session is active.

### DIFF
--- a/src/controller/python/ChipDeviceController-ScriptBinding.cpp
+++ b/src/controller/python/ChipDeviceController-ScriptBinding.cpp
@@ -213,7 +213,8 @@ PyChipError pychip_DeviceCommissioner_CloseBleConnection(chip::Controller::Devic
 const char * pychip_Stack_StatusReportToString(uint32_t profileId, uint16_t statusCode);
 
 PyChipError pychip_GetConnectedDeviceByNodeId(chip::Controller::DeviceCommissioner * devCtrl, chip::NodeId nodeId,
-                                              chip::Controller::Python::PyObject * context, DeviceAvailableFunc callback);
+                                              chip::Controller::Python::PyObject * context, DeviceAvailableFunc callback,
+                                              int transportPayloadCapability);
 PyChipError pychip_FreeOperationalDeviceProxy(chip::OperationalDeviceProxy * deviceProxy);
 PyChipError pychip_GetLocalSessionId(chip::OperationalDeviceProxy * deviceProxy, uint16_t * localSessionId);
 PyChipError pychip_GetNumSessionsToPeer(chip::OperationalDeviceProxy * deviceProxy, uint32_t * numSessions);
@@ -239,6 +240,13 @@ void pychip_Storage_ShutdownAdapter(chip::Controller::Python::StorageAdapter * s
 // ICD
 //
 void pychip_CheckInDelegate_SetOnCheckInCompleteCallback(PyChipCheckInDelegate::OnCheckInCompleteCallback * callback);
+
+//
+// LargePayload and TCP
+PyChipError pychip_SessionAllowsLargePayload(chip::OperationalDeviceProxy * deviceProxy, bool * allowsLargePayload);
+PyChipError pychip_IsSessionOverTCPConnection(chip::OperationalDeviceProxy * deviceProxy, bool * isSessionOverTCP);
+PyChipError pychip_IsActiveSession(chip::OperationalDeviceProxy * deviceProxy, bool * isActiveSession);
+PyChipError pychip_CloseTCPConnectionWithPeer(chip::OperationalDeviceProxy * deviceProxy);
 }
 
 void * pychip_Storage_InitializeStorageAdapter(chip::Controller::Python::PyObject * context,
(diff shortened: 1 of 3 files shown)


yufengwangca (approved):

src/controller/python/chip/ChipDeviceCtrl.py
```diff
@@ -906,7 +950,7 @@ async def FindOrEstablishPASESession(self, setupCode: str, nodeid: int, timeoutM
         if res.is_success:
             return DeviceProxyWrapper(returnDevice, DeviceProxyWrapper.DeviceProxyType.COMMISSIONEE, self._dmLib)
 
-    def GetConnectedDeviceSync(self, nodeid, allowPASE=True, timeoutMs: int = None):
+    def GetConnectedDeviceSync(self, nodeid, allowPASE=True, timeoutMs: int = None, payloadCapability: int = TransportPayloadCapability.MRP_PAYLOAD):
```
cecille, line 953: This is matching existing functinality, I'm guessing, but do you want the default to be MRP or both?

cecille (approved):
Looks reasonable to me. @agners - would you mind taking a look to make sure this is OK with your use cases?

src/controller/python/ChipDeviceController-ScriptBinding.cpp
```diff
@@ -805,11 +812,45 @@ struct GetDeviceCallbacks
...
+{
+    VerifyOrReturnError(deviceProxy->GetSecureSession().HasValue(), ToPyChipError(CHIP_ERROR_MISSING_SECURE_SESSION));
+    VerifyOrReturnError(allowsLargePayload != nullptr, ToPyChipError(CHIP_ERROR_INVALID_ARGUMENT));
+
+    *allowsLargePayload = deviceProxy->GetSecureSession().Value()->AsSecureSession()->AllowsLargePayload();
+
+    return ToPyChipError(CHIP_NO_ERROR);
+}
+
+PyChipError pychip_IsSessionActive(chip::OperationalDeviceProxy * deviceProxy, bool * isSessionActive)
```
agners, line 834: Hm, I'd prefer if we align naming with the SDK here:
```suggestion
PyChipError pychip_IsActiveSession(chip::OperationalDeviceProxy * deviceProxy, bool * isSessionActive)
```

And also the

src/controller/python/chip/ChipDeviceCtrl.py
```diff
@@ -371,6 +381,40 @@ def attestationChallenge(self) -> bytes:
 
         return bytes(buf)
 
+    @property
+    def sessionAllowsLargePayload(self) -> bool:
```
agners, line 385: Properties are typically capitalized:

```suggestion
    def SessionAllowsLargePayload(self) -> bool:
```

src/controller/python/chip/ChipDeviceCtrl.py
```diff
@@ -371,6 +381,40 @@ def attestationChallenge(self) -> bytes:
...
+        supportsLargePayload = ctypes.c_bool(False)
+
+        builtins.chipStack.Call(
+            lambda: self._dmLib.pychip_SessionAllowsLargePayload(self._deviceProxy, pointer(supportsLargePayload))
+        ).raise_on_error()
+
+        return supportsLargePayload.value
+
+    @property
+    def isSessionActive(self) -> bool:
```
agners, line 398: Here too, and also the naming mentioned above.
```suggestion
    def IsActiveSession(self) -> bool:
```

src/controller/python/chip/ChipDeviceCtrl.py
agners: I guess existing devices don't support other capabilities? Or what would be the implications here?

There is `MRP_OR_TCP_PAYLOAD`, which I guess would choose MRP unless TCP is supported? :thinking:

agners (changes requested):

src/controller/python/chip/ChipDeviceCtrl.py
agners: Hm, I see all properties are lower case here :see_no_evil: 

I think we should fix this, but should be a follow up PR. For this PR, you can leave as is (and the follow up will change), or already adopt the capitalize stance, either way works for me.

src/controller/python/chip/ChipDeviceCtrl.py
pidarped: So, I have kept the default to match the existing functionality(as you noted). Making the default to MRPOrTCP would also require changes within the SDK APIs. If we choose to do that(which sounds perfectly reasonable), I would do that in a subsequent PR where it will contain that change throughout the SDK and the python scripts.

src/controller/python/ChipDeviceController-ScriptBinding.cpp
pidarped: Ok. Modified the naming in all places.

src/controller/python/chip/ChipDeviceCtrl.py
pidarped: Yes, all existing ones are lowercase and that is the convention I followed. I will leave this one as-is for now.

agners (approved):

robszewczyk (approved):

tehampson (approved):

aronrosenberg (approved):

src/controller/python/chip/ChipDeviceCtrl.py
```diff
@@ -84,6 +84,16 @@
 
 _ChipDeviceController_IterateDiscoveredCommissionableNodesFunct = CFUNCTYPE(None, c_char_p, c_size_t)
 
+# Defines for the transport payload types to use to select the suitable
+# underlying transport of the session.
+# class TransportPayloadCapability(ctypes.c_int):
+
+
+class TransportPayloadCapability(ctypes.c_int):
```
ksperling-apple, line 92: Is this the direct equivalent of a C++ enum? If so please documenting that.
Other than that it would be good to use either "LARGE" or "TCP" consistently.
Also should this be a bit mask with one bit each for MRP and TCP?

PR #34451, opened 2024-07-23
Mark session for eviction on TCP Client side when disconnecting with peer.

pidarped:
The TCP server on receiving a TCPDisconnect from the client marks the corresponding secure session for eviction.
The TCP client should also mark its session for eviction when proactively closing the connection with peer.

### DIFF
--- a/src/transport/SessionManager.cpp
+++ b/src/transport/SessionManager.cpp
@@ -714,27 +714,7 @@ void SessionManager::HandleConnectionClosed(Transport::ActiveTCPConnectionState 
 {
     VerifyOrReturn(conn != nullptr);
 
-    // Mark the corresponding secure sessions as defunct
-    mSecureSessions.ForEachSession([&](auto session) {
-        if (session->IsActiveSession() && session->GetTCPConnection() == conn)
-        {
-            SessionHandle handle(*session);
-            // Notify the SessionConnection delegate of the connection
-            // closure.
-            if (mConnDelegate != nullptr)
-            {
-                mConnDelegate->OnTCPConnectionClosed(handle, conErr);
-            }
-
-            // Dis-associate the connection from session by setting it to a
-            // nullptr.
-            session->SetTCPConnection(nullptr);
-            // Mark session as defunct
-            session->MarkAsDefunct();
-        }
-
-        return Loop::Continue;
-    });
+    MarkSecureSessionOverTCPForEviction(conn, conErr);
 
     // TODO: A mechanism to mark an unauthenticated session as unusable when
     // the underlying connection is broken. Issue #32323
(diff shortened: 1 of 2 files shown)


src/transport/SessionManager.h
```diff
@@ -621,6 +621,10 @@ class DLL_EXPORT SessionManager : public TransportMgrDelegate, public FabricTabl
         return payloadHeader.HasMessageType(Protocols::SecureChannel::MsgType::MsgCounterSyncReq) ||
             payloadHeader.HasMessageType(Protocols::SecureChannel::MsgType::MsgCounterSyncRsp);
     }
+
+#if INET_CONFIG_ENABLE_TCP_ENDPOINT
+    void MarkSecureSessionDefunctForConnection(Transport::ActiveTCPConnectionState * conn, CHIP_ERROR conErr);
+#endif // INET_CONFIG_ENABLE_TCP_ENDPOINT
```
yufengwangca, line 627: nit: move this above IsControlMessage so that all non static member functions stay together

yufengwangca (approved):

src/transport/SessionManager.cpp
```diff
@@ -1336,6 +1318,33 @@ Optional<SessionHandle> SessionManager::FindSecureSessionForNode(ScopedNodeId pe
...
+            if (mConnDelegate != nullptr)
+            {
+                mConnDelegate->OnTCPConnectionClosed(handle, conErr);
+            }
+
+            // Dis-associate the connection from session by setting it to a
+            // nullptr.
+            session->SetTCPConnection(nullptr);
+            // Mark session as defunct
+            session->MarkAsDefunct();
```
bzbarsky-apple, line 1340: So here's a question: Why are we doing MarkAsDefunct(), not MarkForEviction()?  Is there a way this session can still be OK when the connection has been closed?  I wouldn't think so...

bzbarsky-apple (approved):

src/transport/SessionManager.cpp
pidarped: I think it can be marked for eviction. The original intent was to, potentially, re-establish the connection and link a defunct session if one existed with said peer. But, currently, for the sake of simplicity, when a new connection is set up, a corresponding new session is also set up. This might be better as we may not want to reuse a session that has been stale for long. But I thought that a defunct session could also be evicted, no? If not then MarkForEviction is probably the right option.

bzbarsky-apple: A defunct session can absolutely be evicted, but it might be prioritized for eviction below other sessions.  So if we know for sure our session is never going to be usable, MarkForEviction will be better in terms of making sure we don't leave this session around while evicting a potentially-usable session.

bzbarsky-apple (approved):

PR #34457, opened 2024-07-23
Make tv casting tests use multithreading.

### LINKED ISSUE #34443: run_tv_casting_test.py timeouts do not work.
The test is supposed to time out within 10 seconds by default if the required response is not received. Instead, for my machine when I timeout I currently have 31 seconds (and the tests pass in 2 seconds when they pass, so this is not some cummulative thing)

### LINKED ISSUE #34444: run_tv_casting_test.py is flaky on linux
This was on builds on a10b47d4da020aef369041967dcf8901b534ddea:

Compiled with

```sh
source scripts/activate.sh

./scripts/examples/gn_build_example.sh examples/tv-app/linux/ out/tv-app
./scripts/examples/gn_build_example.sh  examples/tv-casting-app/linux/ out/tv-casting-app 'chip_casting_simplified=true'
```

And then ran `python3 ./scripts/tests/run_tv_casting_test.py`. I periodically get this to fail (I was investigating why a PR I had was failing, which seems to make it fail consistently, however even in ToT it seems flaky). Tried repeatedly and I get 10-20% failure rate on my machine:

image

andy31415:
TV casting tests had inline stdout reading as a single process, which caused several issues:
  - output reads were blocked while waiting for another process, potentially blocking the sending side for the "other" app, essentially deadlocking the chip main thread if the logging is blocking
  - timeouts were not properly working, since `readline` is blocking
  
Fixes #34443 and fixes #34444

### Changes

- create a new class for that does stdout i/o in a separate thread and timeouts that work
- move log reporting into exception handlers, so that a separate log path and exits are not needed
- move from a central `handle_failure` to exceptions
- enable json tracing for communication in the sample apps and during test runs (to see a clearer picture of packets that are sent/received during runs)

### Tested

tv_casting test still passes (and newly passes on my pr that had failures before)

### DIFF
--- a/scripts/tests/linux/log_line_processing.py
+++ b/scripts/tests/linux/log_line_processing.py
@@ -66,15 +66,15 @@ class ProcessOutputCapture:
             while not self.done:
                 changes = out_wait.poll(0.1)
                 if changes:
-                    l = self.process.stdout.readline()
-                    f.write(l)
-                    self.output_lines.put(l)
+                    out_line = self.process.stdout.readline()
+                    f.write(out_line)
+                    self.output_lines.put(out_line)
 
                 changes = err_wait.poll(0)
                 if changes:
-                    l = self.process.stderr.readline()
-                    if l:
-                        f.write(f"!!STDERR!! : {l}")
+                    err_line = self.process.stderr.readline()
+                    if err_line:
+                        f.write(f"!!STDERR!! : {err_line}")
 
     def __enter__(self):
         self.done = False
@@ -102,8 +102,8 @@ class ProcessOutputCapture:
             # When we fail because of an exception, report the entire log content
             logging.error(f"-------- START: LOG DUMP FOR {self.command!r} -----")
             with open(self.output_path, "rt") as f:
-                for l in f.readlines():
-                    logging.error(l.strip())
+                for output_line in f.readlines():
+                    logging.error(output_line.strip())
             logging.error(f"-------- END:   LOG DUMP FOR {self.command!r} -----")
 
     def next_output_line(self, timeout_sec=None):
--- a/scripts/tests/linux/log_line_processing.py
+++ b/scripts/tests/linux/log_line_processing.py
@@ -108,7 +108,10 @@ class ProcessOutputCapture:
 
     def next_output_line(self, timeout_sec=None):
         """Fetch an item from the output queue, potentially wit h a timeout."""
-        return self.output_lines.get(timeout=timeout_sec)
+        try:
+            return self.output_lines.get(timeout=timeout_sec)
+        except Empty:
+            return None
 
     def send_to_program(self, what):
         """Sends the given string to the program.
--- a/scripts/tests/linux/log_line_processing.py
+++ b/scripts/tests/linux/log_line_processing.py
@@ -118,5 +118,5 @@ class ProcessOutputCapture:
 
         NOTE: remember to append a `\n` for terminal applications
         """
-        self.process.stdin.write(what)
+        self.process.stdin.write(input_cmd)
         self.process.stdin.flush()
--- a/scripts/tests/linux/log_line_processing.py
+++ b/scripts/tests/linux/log_line_processing.py
@@ -113,8 +113,8 @@ class ProcessOutputCapture:
         except queue.Empty:
             return None
 
-    def send_to_program(self, what):
-        """Sends the given string to the program.
+    def send_to_program(self, input_cmd):
+        """Sends the given input command string to the program.
 
         NOTE: remember to append a `\n` for terminal applications
         """
--- a/scripts/tests/linux/log_line_processing.py
+++ b/scripts/tests/linux/log_line_processing.py
@@ -30,7 +30,7 @@ class ProcessOutputCapture:
         the process stdout)
       - provides read timeouts for incoming data
 
-    Use as part of a resource mangement block like:
+    Use as part of a resource management block like:
 
     with ProcessOutputCapture("test.sh", "logs.txt") as p:
        p.send_to_program("input\n")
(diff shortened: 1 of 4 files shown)


scripts/tests/linux/log_line_processing.py
```diff
@@ -0,0 +1,122 @@
...
+    """
+    Captures stdout from a process and redirects such stdout to a given file.
+
+    The capture serves several purposes as opposed to just reading stdout:
+      - as data is received, it will get written to a separate file
+      - data is accumulated in memory for later processing (does not starve/block
+        the process stdout)
+      - provides read timeouts for incoming data
+
+    Use as part of a resource mangement block like:
```
shaoltan-amazon, line 33: Typo: management

scripts/tests/linux/log_line_processing.py
```diff
@@ -0,0 +1,119 @@
...
+        out_wait.register(self.process.stdout, select.POLLIN)
+
+        err_wait = select.poll()
+        err_wait.register(self.process.stderr, select.POLLIN)
+
+        with open(self.output_path, "wt") as f:
+            while not self.done:
+                changes = out_wait.poll(0.1)
+                if changes:
+                    out_line = self.process.stdout.readline()
```
cecille, line 69: why is it possible for there to be no err_line, but there's always an out_line?

scripts/tests/linux/log_line_processing.py
```diff
@@ -0,0 +1,119 @@
...
+                    f.write(out_line)
+                    self.output_lines.put(out_line)
+
+                changes = err_wait.poll(0)
+                if changes:
+                    err_line = self.process.stderr.readline()
+                    if err_line:
+                        f.write(f"!!STDERR!! : {err_line}")
```
cecille, line 77: does stderr not go to output_lines?

scripts/tests/linux/log_line_processing.py
```diff
@@ -0,0 +1,119 @@
...
+        self.done = False
+
+    def _io_thread(self):
+        """Reads process lines and writes them to an output file.
+
+        It also sends the output lines to `self.output_lines` for later
+        reading
+        """
+        out_wait = select.poll()
+        out_wait.register(self.process.stdout, select.POLLIN)
```
cecille, line 60: what in the input poll used for here?

scripts/tests/run_tv_casting_test.py
```diff
@@ -35,243 +36,268 @@
...
+                )
                 for line in msg_block:
-                    logging.info(f'{test_sequence_name} - {line}')
+                    logging.info(f"{test_sequence_name} - {line}")
 
-                return True
+                # successful completion
+                return
+
+    raise TestStepException("Unespected exit", test_sequence_name, test_sequence_step)
```
shaoltan-amazon, line 176:
```suggestion
    raise TestStepException("Unexpected exit", test_sequence_name, test_sequence_step)
```

scripts/tests/linux/log_line_processing.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def next_output_line(self, timeout_sec=None):
+        """Fetch an item from the output queue, potentially wit h a timeout."""
+        try:
+            return self.output_lines.get(timeout=timeout_sec)
+        except queue.Empty:
+            return None
+
+    def send_to_program(self, what):
+        """Sends the given string to the program.
```
shaoltan-amazon, line 117: Nit:
```suggestion
    def send_to_program(self, input_cmd):
        """Sends the given input command string to the program.
```

scripts/tests/linux/log_line_processing.py
```diff
@@ -0,0 +1,122 @@
...
+
+        NOTE: remember to append a `\n` for terminal applications
+        """
+        self.process.stdin.write(what)
```
shaoltan-amazon, line 121: Nit:
```suggestion
        self.process.stdin.write(input_cmd)
```

scripts/tests/linux/log_line_processing.py
```diff
@@ -0,0 +1,122 @@
...
+        if exception_value:
+            # When we fail because of an exception, report the entire log content
+            logging.error(f"-------- START: LOG DUMP FOR {self.command!r} -----")
+            with open(self.output_path, "rt") as f:
+                for output_line in f.readlines():
+                    logging.error(output_line.strip())
+            logging.error(f"-------- END:   LOG DUMP FOR {self.command!r} -----")
+
+    def next_output_line(self, timeout_sec=None):
+        """Fetch an item from the output queue, potentially wit h a timeout."""
```
shaoltan-amazon, line 110:
```suggestion
        """Fetch an item from the output queue, potentially with a timeout."""
```

scripts/tests/linux/log_line_processing.py
shaoltan-amazon:
```suggestion
    Use as part of a resource management block like:
```

scripts/tests/linux/log_line_processing.py
andy31415: Process exit results in a '' error line. I thought this is ok for output lines, but because I prefix `STDERR!!` in front of errors, I figured empty error lines are annoying (and I did get them while testing) hence this change.

cecille: ah, ok, gotcha.

andy31415: Removed the logic though ... I think it is probably fine to see a empty stderr at the end.

scripts/tests/linux/log_line_processing.py
cecille: answered as part of the other question below. It's ok to write empty because there's nothing prepended.

cecille (approved):

shaoltan-amazon (approved):

chrisdecenzo (approved):

PR #34459, opened 2024-07-23
Add Ecosystem Information Cluster Server implementation

### LINKED ISSUE #34553: [Feature] Implements the Ecosystem Information Server in Matter SDK
### Feature description

Add Ecosystem Information Cluster Server implementation to src/app/clusters/ for provisional fabric sync

This registers AttributeAccessInterface so that we can override the read attributes. This instance was developed in a way so that ownership of the underlying types are embedded into the Device and Location structs themselves and that we would only create instance of those structs if the data is valid.

To be completed in follow up PRs:
Proper fabric scoping of elements in the device and location lists.
Testing:
With local changes to examples/fabric-bridge-app/ confirmed was able to read attributes from EcosystemInformationCluster

### Platform

other

### Platform Version(s)

_No response_

### Anything else?

_No response_

tehampson:
This adds Ecosystem Information Cluster Server implementation to `src/app/clusters/` for provisional fabric sync

This registers AttributeAccessInterface so that we can override the read attributes. This instance was developed in a way so that ownership of the underlying types are embedded into the Device and Location structs themselves and that we would only create instance of those structs if the data is valid.

#### To be completed in follow up PRs:
* Proper fabric scoping of elements in the device and location lists.

#### Testing:
With local changes to `examples/fabric-bridge-app/` confirmed was able to read attributes from EcosystemInformationCluster


src/app/clusters/ecosystem-information-server/ecosystem-information-server.cpp
```diff
@@ -0,0 +1,356 @@
+/*
```
tehampson, line 1: Blocking comment on https://github.com/project-chip/connectedhomeip/pull/34291/ being merged first

src/app/clusters/ecosystem-information-server/ecosystem-information-server.h
```diff
@@ -0,0 +1,211 @@
...
+        uint64_t mDeviceNameLastEditEpochUs = 0;
+        EndpointId mBridgedEndpoint         = kInvalidEndpointId;
+        EndpointId mOriginalEndpoint        = kInvalidEndpointId;
+        std::vector<Structs::DeviceTypeStruct::Type> mDeviceTypes;
+        std::vector<std::string> mUniqueLocationIds;
+        uint64_t mUniqueLocationIdsLastEditEpochUs = 0;
+        bool mIsAlreadyBuilt                       = false;
+    };
+
+    EcosystemDeviceStruct(const EcosystemDeviceStruct &) = delete;
```
andy31415, line 69: why is this not allowed? I see no self-referencing things at a first glance. Could we add a comment?

Should we delete `operator=` as well if really not allowed?

src/app/clusters/ecosystem-information-server/ecosystem-information-server.h
```diff
@@ -0,0 +1,211 @@
...
+
+    CHIP_ERROR Encode(const AttributeValueEncoder::ListEncodeHelper & aEncoder, const FabricIndex & aFabricIndex);
+
+private:
+    // Constructor is intentionally private. This is to ensure that it is only constructed with
+    // values that conform to the spec.
+    explicit EcosystemDeviceStruct(std::string aDeviceName, uint64_t aDeviceNameLastEditEpochUs, EndpointId aBridgedEndpoint,
```
andy31415, line 76: If this is built using a builder, would it make sense to use `&&` for the larger types (string, vectors)?

src/app/clusters/ecosystem-information-server/ecosystem-information-server.cpp
```diff
@@ -0,0 +1,368 @@
...
+std::unique_ptr<EcosystemDeviceStruct> EcosystemDeviceStruct::Builder::Build()
+{
+    bool deviceNameIsInvalid            = mDeviceName.size() > kDeviceNameMaxSize;
+    bool originalEndpointIsInvalid      = mOriginalEndpoint == kInvalidEndpointId;
+    bool deviceTypesListIsInvalid       = mDeviceTypes.empty();
+    bool uniqueLocationIdsListIsInvalid = mUniqueLocationIds.size() > kUniqueLocationIdsListMaxSize;
+    if (mIsAlreadyBuilt || deviceNameIsInvalid || originalEndpointIsInvalid || deviceTypesListIsInvalid ||
+        uniqueLocationIdsListIsInvalid)
+    {
+        return nullptr;
```
andy31415, line 141: here we lose the reason of why building failed. Since this is likely infrequently hit, would it make sense to log some error text whenever we return nullptr?

Also why do we do all boolean setting and only then an if? The first invalid should already fail like:

```
VerifyOrReturnValue(mDeviceName.size() <= kDeviceNameMaxSize, nullptr, ChipLogError(Zcl, "Device name too large"));
VerifyOrReturnValue(mOriginalEndpoint != kInvalidEndpointId;, nullptr, ChipLogError(Zcl, "Invalid original endpoint"));

....
```

src/app/clusters/ecosystem-information-server/ecosystem-information-server.cpp
```diff
@@ -0,0 +1,368 @@
...
+    for (auto & locationId : mUniqueLocationIds)
+    {
+        if (locationId.size() > kUniqueLocationIdMaxSize)
+        {
+            return nullptr;
+        }
+    }
+
+    // std::make_unique does not have access to private constructor we workaround with using new
+    std::unique_ptr<EcosystemDeviceStruct> ret{ new EcosystemDeviceStruct(
```
andy31415, line 153: consider using make_unique? Or does that fail because of private constructor?

src/app/clusters/ecosystem-information-server/ecosystem-information-server.cpp
```diff
@@ -0,0 +1,368 @@
...
+namespace Clusters {
+namespace EcosystemInformation {
+namespace {
+
+constexpr size_t kDeviceNameMaxSize            = 64;
+constexpr size_t kUniqueLocationIdMaxSize      = 64;
+constexpr size_t kUniqueLocationIdsListMaxSize = 64;
+constexpr size_t kHomeLocationNameMaxSize      = 128;
+
+constexpr size_t kDeviceDirectoryMaxSize   = 256;
```
andy31415, line 34: are there any spec-required reasons for this maximum count?

If these are large devices that support STL, would it make sense to bump this up a lot of make them a configurable option?

src/app/clusters/ecosystem-information-server/ecosystem-information-server.cpp
```diff
@@ -0,0 +1,368 @@
...
+    return mInstance;
+}
+
+CHIP_ERROR EcosystemInformationServer::AddDeviceInfo(EndpointId aEndpoint, std::unique_ptr<EcosystemDeviceStruct> aDevice)
+{
+    VerifyOrReturnError(aDevice, CHIP_ERROR_INVALID_ARGUMENT);
+    VerifyOrReturnError((aEndpoint != kRootEndpointId && aEndpoint != kInvalidEndpointId), CHIP_ERROR_INVALID_ARGUMENT);
+
+    auto & deviceInfo = mDevicesMap[aEndpoint];
+    VerifyOrReturnError((deviceInfo.mDeviceDirectory.size() >= kDeviceDirectoryMaxSize), CHIP_ERROR_NO_MEMORY);
```
andy31415, line 263: this verify seems reversed ... should we verify that the size is smaller than the maximum size and otherwise return an error?

Same below.

src/app/clusters/ecosystem-information-server/ecosystem-information-server.cpp
```diff
@@ -0,0 +1,368 @@
...
+    auto & deviceInfo = mDevicesMap[aEndpoint];
+    deviceInfo.mRemovedOn.SetValue(aEpochUs);
+    return CHIP_NO_ERROR;
+}
+
+CHIP_ERROR EcosystemInformationServer::EncodeRemovedOnAttribute(EndpointId aEndpoint, AttributeValueEncoder & aEncoder)
+{
+    if (mDevicesMap.find(aEndpoint) == mDevicesMap.end())
+    {
+        return CHIP_IM_GLOBAL_STATUS(UnsupportedCluster);
```
andy31415, line 294: UnsupportedEndpoint?

or is the unsupported cluster correct? This may warrant a comment there since I read this as "search by endpoint and if not found return ..."

src/app/clusters/ecosystem-information-server/ecosystem-information-server.cpp
```diff
@@ -0,0 +1,368 @@
...
+    }
+
+    auto & deviceInfo = mDevicesMap[aEndpoint];
```
andy31415, line 297: could we save and use the result of `::find` to avoid the double map lookup?

src/app/clusters/ecosystem-information-server/ecosystem-information-server.cpp
andy31415: I guess if mDevicesMap does not include a endpoint this means the cluster implementation does not exist on that endpoint so we always assume "endpoint is valid, but cluster may not be active there"?

src/app/clusters/ecosystem-information-server/ecosystem-information-server.h
tehampson: 100%, thanks!

src/app/clusters/ecosystem-information-server/ecosystem-information-server.cpp
tehampson: Thanks!

src/app/clusters/ecosystem-information-server/ecosystem-information-server.cpp
tehampson: Fails because of private constructor. There is a comment above mentioning this

src/app/clusters/ecosystem-information-server/ecosystem-information-server.cpp
tehampson: Yes this is a spec requirement. I don't know why there is a limit simply implementing constrains as defined.

src/app/clusters/ecosystem-information-server/ecosystem-information-server.cpp
tehampson: Good catch.

This comment is also now left open until I confirm with my local setup that my latest updates and check work as expected

src/app/clusters/ecosystem-information-server/ecosystem-information-server.cpp
tehampson: That is exactly the reason.

src/app/clusters/ecosystem-information-server/ecosystem-information-server.h
tehampson: Removed

src/app/clusters/ecosystem-information-server/ecosystem-information-server.cpp
tehampson: Confirmed I was able to add and read values from Device and Location directory

andy31415 (approved):

src/app/clusters/ecosystem-information-server/ecosystem-information-server.cpp
yufengwangca: https://github.com/project-chip/connectedhomeip/pull/34291 landed

yufengwangca (approved):

src/app/clusters/ecosystem-information-server/ecosystem-information-server.cpp
```diff
@@ -0,0 +1,375 @@
...
+    for (auto & id : mUniqueLocationIds)
+    {
+        locationIds.push_back(CharSpan(id.c_str(), id.size()));
+    }
+    deviceStruct.uniqueLocationIDs = DataModel::List<CharSpan>(locationIds.data(), locationIds.size());
+
+    deviceStruct.uniqueLocationIDsLastEdit = mUniqueLocationIdsLastEditEpochUs;
+
+    // TODO(#33223) this is a hack, use mFabricIndex when it exists. Additionally checking fabric
+    // index matches should happen at the top of this method to prevent building the encodable
```
bzbarsky-apple, line 178: Then you have to make assumptions here about which fields are sensitive and which are not, which can diverge from the actual spec/XML....

src/app/clusters/ecosystem-information-server/ecosystem-information-server.cpp
```diff
@@ -0,0 +1,375 @@
...
+        return EcosystemInformationServer::Instance().EncodeDeviceDirectoryAttribute(aPath.mEndpointId, aEncoder);
+    case Attributes::LocationDirectory ::Id:
+        return EcosystemInformationServer::Instance().EncodeLocationStructAttribute(aPath.mEndpointId, aEncoder);
+    default:
+        break;
+    }
+    return CHIP_NO_ERROR;
+}
+
+Structs::HomeLocationStruct::Type GetEncodableLocationDescriptorStruct(const LocationDescriptorStruct & aLocationDescriptor)
```
bzbarsky-apple, line 63: This is a bit unsafe in that the caller's LocationDescriptorStruct could die while our return value is referencing it, but I don't really know a good way to address that in C++...

One thing you _could_ try to do to avoid temporaries being passed in is to have an overload that takes an rvalue ref and either fails to compile (would be easier if it were a deleted member method) or at least at runtime dies so that if something calls it we will notice.

bzbarsky-apple (approved):

src/app/clusters/ecosystem-information-server/ecosystem-information-server.cpp
tehampson: Yeah this was the best I could come up with. An earlier version had a failed direct encoder that was very ugly

I meant to add documentation to this to talk about the lifetime considerations (I will add comment before merging). I also chose to have this be in the anonymous namespace in the cpp file as it would would limit where and how it was used.

>One thing you could try to do to avoid temporaries being passed in is to have an overload that takes an rvalue ref and either fails to compile (would be easier if it were a deleted member method) or at least at runtime dies so that if something calls it we will notice.

Not sure I understand

src/app/clusters/ecosystem-information-server/ecosystem-information-server.cpp
tehampson: I will update the comment.

Initially when I wrote that comment I assumed all read were fabric filtered read. I have since learned that we always have to encode

src/app/clusters/ecosystem-information-server/ecosystem-information-server.cpp
bzbarsky-apple: No, I am referring to overloading `GetEncodableLocationDescriptorStruct` with a version that takes a `LocationDescriptorStruct &&` which either causes compilation to fail when called or at least causes a ChipDie when called.

tehampson: Ahh.... understood. If this were accessible outside of this cpp file that might be worth it, but since this is used in the file with a warning label above it, and it is used properly here I am going to say it is not worth it.

I added a TODO comment so I might try the method of having this be a member method and deleting version that takes the `LocationDescriptorStruct &&`

PR #34462, opened 2024-07-23
Make AttributeAccessInterface and CommandHandlerInterface registries consistent as objects with singletons

andy31415:
This makes both AAI and CHI consistent: objects with an `Instance` method and UpperCamelCase method names.

This is a followup on #34414


src/app/InteractionModelEngine.cpp
```diff
@@ -99,11 +99,6 @@ CHIP_ERROR InteractionModelEngine::Init(Messaging::ExchangeManager * apExchangeM
 void InteractionModelEngine::Shutdown()
 {
     mpExchangeMgr->GetSessionManager()->SystemLayer()->CancelTimer(ResumeSubscriptionsTimerCallback, this);
-
-    // TODO: individual object clears the entire command handler interface registry.
-    //       This may not be expected.
-    CommandHandlerInterfaceRegistry::UnregisterAllHandlers();
```
bzbarsky-apple, line 105: Well, hold on.  We do in fact need to UnregisterAllHandlers() somewhere on shutdown.  Where is that happening?

src/app/clusters/resource-monitoring-server/resource-monitoring-server.h
```diff
@@ -74,7 +74,8 @@ class Instance : public CommandHandlerInterface, public AttributeAccessInterface
      * @die                                     If the endpoint and cluster ID have not been enabled in zap.
      * @return CHIP_ERROR_INVALID_ARGUMENT      If the CommandHandler or Attribute Handler could not be registered.
      * @return CHIP_ERROR_INCORRECT_STATE       If the CommandHandler was already registered
-     * @return CHIP_ERROR_INCORRECT_STATE       If the registerAttributeAccessOverride fails.
+     * @return CHIP_ERROR_INCORRECT_STATE       If the
+     * AttributeAccessInterfaceRegistry::Instance().RegisterAttributeAccessOverride fails.
```
bzbarsky-apple, line 78: Weird indentation here... should have some spaces, no?

src/app/AttributeAccessInterfaceRegistry.h
```diff
@@ -16,39 +16,50 @@
...
+     * endpoint it's registered for is disabled (or until shutdown if it's
+     * registered for all endpoints) or until it is explicitly unregistered.
+     * Registration will fail if there is an already-registered override for the
+     * same set of attributes.
+     *
+     * @return false if there is an existing override that the new one would
+     *               conflict with.  In this case the override is not registered.
+     * @return true if registration was successful.
+     */
+    bool RegisterAttributeAccessOverride(chip::app::AttributeAccessInterface * attrOverride);
```
bzbarsky-apple, line 39: I know you kept the existing naming, but that naming was for a context-less free function.  In the context of this class, should this be perhaps called `Register`?  Or `RegisterAttributeAccess` or something?

bzbarsky-apple, line 39:
```suggestion
    bool RegisterAttributeAccessOverride(AttributeAccessInterface * attrOverride);
```

src/app/AttributeAccessInterfaceRegistry.h
```diff
@@ -16,39 +16,50 @@
...
 
-/**
- * Unregister all attribute access interfaces that match this given endpoint.
- */
-void unregisterAllAttributeAccessOverridesForEndpoint(EmberAfDefinedEndpoint * definedEndpoint);
+    /**
+     * Unregister an attribute access override (for example if the object
+     * implementing AttributeAccessInterface is being destroyed).
+     */
+    void UnregisterAttributeAccessOverride(chip::app::AttributeAccessInterface * attrOverride);
```
bzbarsky-apple, line 45:
```suggestion
    void UnregisterAttributeAccessOverride(AttributeAccessInterface * attrOverride);
```

bzbarsky-apple, line 45: Again, perhaps just `Unregister`?

src/app/AttributeAccessInterfaceRegistry.cpp
```diff
@@ -60,43 +58,28 @@ void UnregisterMatchingAttributeAccessInterfaces(F shouldUnregister)
...
+AttributeAccessInterfaceRegistry & AttributeAccessInterfaceRegistry::Instance()
 {
-    UnregisterMatchingAttributeAccessInterfaces(
-        [endpoint = definedEndpoint->endpoint](AttributeAccessInterface * entry) { return entry->MatchesEndpoint(endpoint); });
+    static AttributeAccessInterfaceRegistry instance;
+    return instance;
 }
 
-bool registerAttributeAccessOverride(AttributeAccessInterface * attrOverride)
+void AttributeAccessInterfaceRegistry::UnregisterAllForEndpoint(EmberAfDefinedEndpoint * definedEndpoint)
```
bzbarsky-apple, line 70: If we are changing this anyway, the argument here should be `EndpointId`.  It was convenient to have it be `EmberAfDefinedEndpoint` back when this all lived in ember stuff, but we don't need that dependency now, right?

Followup is fine for this part.

src/app/AttributeAccessInterfaceRegistry.h
andy31415: Updated. Had it lowercase first and `register` was a reserved word.

src/app/AttributeAccessInterfaceRegistry.h
andy31415: Yes, Started renaming and it started lowercase with `register` being a reserved word. Updated.

src/app/AttributeAccessInterfaceRegistry.cpp
```diff
@@ -60,43 +58,28 @@ void UnregisterMatchingAttributeAccessInterfaces(F shouldUnregister)
...
 {
-    gAttributeAccessInterfaceCache.Invalidate();
```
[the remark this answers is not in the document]
andy31415: I think methods were moved by copy & paster. The invalidate in unregister and register still exist.

src/app/InteractionModelEngine.cpp
andy31415: We are symmetric with AAI: there is no UnregisterAll for AAI, so i figured we would not use it for CHI right now.

We do have a UnregisterAllForEndpoint in endpoint shutdowns.

src/app/AttributeAccessInterfaceRegistry.cpp
```diff
@@ -106,21 +89,44 @@ app::AttributeAccessInterface * GetAttributeAccessOverride(EndpointId endpointId
...
 
         // Did not find AAI registered: mark as definitely not using.
-        gAttributeAccessInterfaceCache.MarkUnused(endpointId, clusterId);
+        mAttributeAccessInterfaceCache.MarkUnused(endpointId, clusterId);
     }
 
     return nullptr;
 }
 
+void AttributeAccessInterfaceRegistry::UnregisterAttributeAccessOverride(AttributeAccessInterface * attrOverride)
```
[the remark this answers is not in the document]
andy31415: Placed them however the copy and paste took them. Updated ordering.

src/app/InteractionModelEngine.cpp
andy31415: I can keep the unregister, however because there is no ownership of this anymore, it is odd. If we run multiple IME in paralle, having any IME shutdown clear the handlers for all the others IME seems bad.

We do not do this for AAI.

bzbarsky-apple: When we run multiple IME in parallel, each one will have a separate registry is the point.

For AAI, we should have a "remove all on shutdown" as well if we want to allow any sort of non-static registration, but AAI was initially created with static-only things in mind.

andy31415: I think the question is if this PR is acceptable now or should be deferred until we have a usable DataModel (or DataModel::Provider as the new name) inside the IME.

The reason I am splitting this is to have advance work to decouple IME from all implementations (so they use a provider instead of the IME). My concern is that given that CodeGen metadata is used to list available things, including Generated/Accepted commands, this is still centrally managed and code generated today to some extent.


Review feedback request on path forward:
   - keep the odd clear in here (even if ownership is odd)
   - remove the clear claiming 'things are centrally managed anyway, and since IME does not own the registry, it should not clear it on shutdown'
   
I do not care as much since I imagine medium-term we will have a real data model split anyway which should make things sane (and codegen data model being a singleton can absolutely do the clear of AAI/CHI as it wants).

bzbarsky-apple: I think this PR is fine in its current general form, but I think we should keep the odd clear in here for now, and later on once we have separate registries for separate stack instances the clear would move into stack instance shutdown.

andy31415: Added clear back.

src/app/AttributeAccessInterfaceRegistry.cpp
```diff
@@ -60,43 +58,50 @@ void UnregisterMatchingAttributeAccessInterfaces(F shouldUnregister)
...
-    UnregisterMatchingAttributeAccessInterfaces([attrOverride](AttributeAccessInterface * entry) { return entry == attrOverride; });
+    mAttributeAccessInterfaceCache.Invalidate();
+    UnregisterMatchingAttributeAccessInterfaces([attrOverride](AttributeAccessInterface * entry) { return entry == attrOverride; },
+                                                mAttributeAccessOverrides);
 }
 
-void unregisterAllAttributeAccessOverridesForEndpoint(EmberAfDefinedEndpoint * definedEndpoint)
+void AttributeAccessInterfaceRegistry::UnregisterAllForEndpoint(EndpointId endpointId)
 {
     UnregisterMatchingAttributeAccessInterfaces(
```
bzbarsky-apple, line 79: OK, I see.  This just never invalidated the cache.  But it needs to, no?

Followup is OK for this, since it's pre-existing.

bzbarsky-apple (approved):

src/app/AttributeAccessInterfaceRegistry.cpp
andy31415: Added it to this PR ... may as well fix things.
I think there was a chance we never noticed because we unregister only on EP disabling, at which point the AAI is not usable anyway (and future enabling re-register things that invalidate the cache).

scripts/py_matter_idl/matter_idl/test_idl_generator.py
```diff
@@ -55,8 +55,8 @@ def ReadMatterIdl(repo_path: str) -> str:
         return stream.read()
 
 
-def ParseMatterIdl(repo_path: str, skip_meta: bool) -> Idl:
-    return CreateParser(skip_meta=skip_meta).parse(ReadMatterIdl(repo_path))
+def ParseMatterIdl(repo_path: str, skip_meta: bool, merge_globals: bool) -> Idl:
```
ksperling-apple, line 58: Is this change related to the title of this PR?

ksperling-apple (commented):
If we're changing these APIs anyway, we should consider merging AttributeAccessInterface and CommandHandlerInterface. In most cases cluster servers implement both interfaces, so this would reduce the amount of code required to register a cluster server, as well as memory used by cluster server instances and the registries themselves.

andy31415:
> If we're changing these APIs anyway, we should consider merging AttributeAccessInterface and CommandHandlerInterface. In most cases cluster servers implement both interfaces, so this would reduce the amount of code required to register a cluster server, as well as memory used by cluster server instances and the registries themselves.

I plan to make this change, however for backwards compatibility I am placing after the DM::Provider interface split and usage.

scripts/py_matter_idl/matter_idl/test_idl_generator.py
andy31415: No, I think it was a master merge. Updated with a master merge.

tehampson (approved):

src/app/AttributeAccessInterfaceRegistry.h
```diff
@@ -16,39 +16,49 @@
...
-AttributeAccessInterface * GetAttributeAccessOverride(EndpointId aEndpointId, ClusterId aClusterId);
+    /**
+     *  Get the registered attribute access override. nullptr when attribute access override is not found.
+     */
+    AttributeAccessInterface * Get(EndpointId aEndpointId, ClusterId aClusterId);
+
+    static AttributeAccessInterfaceRegistry & Instance();
+
+private:
+    AttributeAccessInterface * mAttributeAccessOverrides = nullptr;
```
cecille, line 59: Is this always intended to be a singleton? If so, would it make sense to hide the constructor in private?

cecille (approved):

src/app/AttributeAccessInterfaceRegistry.h
andy31415: Used as a singleton in the app, however I am unsure if we ever need testing. I have been having a hard time working with code that went too far into protecting from the user (usually with many private things and having friend only access) so I generally default to public for testability unless there are very strong reasons to do otherwise.

Here it probably does not matter as much as we could `clear all` in tests, however I don't see too much of a harm if we allow constructors. Overally globals are not ideal. In the future I would rather expect we want CodegenDataModelProvider to own these if possible (not possible today because all cluster implementations are coupled and assume a global).

PR #34464, opened 2024-07-23
[Color Control] Scenes Integration test scripts

lpbeliveau-silabs:
Added Test_TC_10_1.yaml to test scenes integration for the color control cluster.

Note:
This test might have to be semi-automated due to the high variance of possible feature combinations in CC, it is currently fully automated for a use case where the DUT uses ALL the different color modes, if there are other common configurations, we could also automate them.

I am open to suggestions as to how to move to fully automation. (Removing the View Scene Command is also a possibility).


src/app/tests/suites/certification/Test_TC_CC_10_1.yaml
```diff
@@ -0,0 +1,1041 @@
+# Copyright (c) 2021 Project CHIP Authors
```
cecille, line 1:
```suggestion
# Copyright (c) 2024 Project CHIP Authors
```

src/app/tests/suites/certification/Test_TC_CC_10_1.yaml
```diff
@@ -0,0 +1,1041 @@
...
+              - name: "ColorY"
+                value: 19660
+              - name: "TransitionTime"
+                value: 0
+              - name: "OptionsMask"
+                value: 1
+              - name: "OptionsOverride"
+                value: 1
+
+    - label: "Wait 100ms"
```
cecille, line 216: why are these waits required?

src/app/tests/suites/certification/Test_TC_CC_10_1.yaml
```diff
@@ -0,0 +1,1041 @@
...
+                                        },
+                                        {
+                                            AttributeID: 0x4001,
+                                            ValueUnsigned8: 0x03,
+                                        },
+                                    ],
+                            },
+                        ]
+
+      # Note : There are too many possible combinations of the extension field sets depending on the supported features, so if they are not all supported, this step will be manual
```
cecille, line 455: not going to block FC on this, but this is fairly key to the test, and it's manual, which means it's error prone. can you assess parts of the extension field sets per feature? ie, is this a separable problem?

lpbeliveau-silabs: Initially tried doing this per feature using the `contains` check with yaml, but when specifying contain for a list of extension field set, it expects the exact specific set to be inside the list, i.e.

```
                   contains:
                        [
                            { ClusterID: 0x0300, AttributeValueList: [ { AttributeID: 0x0003, ValueUnsigned16: 32768, },{ AttributeID: 0x0004,ValueUnsigned16: 19660, },],},
                        ]
```
Would look explicitely for an extensionfield set containing only:
                            `{ ClusterID: 0x0300, AttributeValueList: [ { AttributeID: 0x0003, ValueUnsigned16: 32768, },{ AttributeID: 0x0004,ValueUnsigned16: 19660, },],},`
                            And fail if it has other fields.

src/app/tests/suites/certification/Test_TC_CC_10_1.yaml
lpbeliveau-silabs: Those were to let time to transition since some of the move to calls were asynchronous and "instant" might not always be so "instant" turns out most of them were un-necessary, will remove.

src/app/tests/suites/certification/Test_TC_CC_10_1.yaml
lpbeliveau-silabs: The test is now converted to python and fully automated

src/python_testing/TC_CC_10_1.py
```diff
@@ -0,0 +1,484 @@
...
+        result = await self.TH1.SendCommand(self.dut_node_id, 1, Clusters.Groups.Commands.AddGroup(self.kGroup1, "Group1"))
+        asserts.assert_equal(result.status, Status.Success, "Adding Group 1 failed")
+
+        self.step("1b")
+        result = await self.TH1.SendCommand(self.dut_node_id, 1, Clusters.ScenesManagement.Commands.RemoveAllScenes(self.kGroup1))
+        asserts.assert_equal(result.status, Status.Success, "Remove All Scenes failed on status")
+        asserts.assert_equal(result.groupID, self.kGroup1, "Remove All Scenes failed on groupID")
+
+        self.step("1c")
+        result = await self.TH1.SendCommand(self.dut_node_id, 1, Clusters.ScenesManagement.Commands.GetSceneMembership(self.kGroup1))
```
cecille, line 180: Not a problem, just an FYI - many of the tests you've written have the same setup. You can put multiple tests into a single test class if you want. Test classes have functions called "setup_test" and "teardown_test" that get run before and after each test. If you end up doing another one of these in python, that could be handy.

cecille: Feel free to close this comment, btw, nothing to be done here just though it might be handy info.

src/python_testing/TC_CC_10_1.py
```diff
@@ -0,0 +1,484 @@
...
+        asserts.assert_equal(result.status, Status.Success, "Get Scene Membership failed on status")
+        asserts.assert_equal(result.groupID, self.kGroup1, "Get Scene Membership failed on groupID")
+        asserts.assert_equal(result.sceneList, [], "Get Scene Membership failed on sceneList")
+
+        self.step("1d")
+        result = await self.TH1.ReadAttribute(self.dut_node_id, [(1, attributes.ColorTempPhysicalMinMireds)])
```
cecille, line 186: you've hardcoded the endpoint here, but this test needs to work against any endoint.

You can set the endpoint for the test using the --endpoint flag on the command line. It appears as self.matter_test_config.endpoint. There's also a function in the matter_testing_support file called "read_single_attribute_check_success" that sets the endpoint from there automatically. It also returns the attribute value as a single value so you don't need to dig it out of the dict.

cecille: ditto for all the interactions below.

src/python_testing/TC_CC_10_1.py
```diff
@@ -0,0 +1,484 @@
...
+
+        self.step("4")
+        result = await self.TH1.SendCommand(self.dut_node_id, 1, Clusters.ScenesManagement.Commands.ViewScene(self.kGroup1, 0x01))
+        asserts.assert_equal(result.status, Status.Success, "View Scene failed on status")
+        asserts.assert_equal(result.groupID, self.kGroup1, "View Scene failed on groupID")
+        asserts.assert_equal(result.sceneID, 0x01, "View Scene failed on sceneID")
+        asserts.assert_equal(result.transitionTime, 0, "View Scene failed on transitionTime")
+
+        for EFS in result.extensionFieldSets:
+            if EFS.clusterID == 0x0300:
```
cecille, line 266: It's a bit hard to parse in a diff, but I think you can unnest this a bit by doing 
if EFS.clusterID != 0x0300:
  continue

src/python_testing/TC_CC_10_1.py
```diff
@@ -0,0 +1,484 @@
...
+
+                    if AV.attributeID == 0x0007 and self.pics_guard(self.check_pics("CC.S.C0a.Rsp")):
+                        asserts.assert_less_equal(AV.valueUnsigned16, ColorTempPhysicalMaxMiredsValue,
+                                                  "View Scene failed on ColorTemperatureMireds above limit")
+                        asserts.assert_greater_equal(AV.valueUnsigned16, ColorTempPhysicalMinMiredsValue,
+                                                     "View Scene failed on ColorTemperatureMireds below limit")
+
+                    if AV.attributeID == 0x4000 and self.pics_guard(self.check_pics("CC.S.C43.Rsp")):
+                        asserts.assert_less_equal(AV.valueUnsigned16, 21800, "View Scene failed on EnhancedHue above limit")
+                        asserts.assert_greater_equal(AV.valueUnsigned16, 18200, "View Scene failed on EnhancedHue below limit")
```
cecille, line 288: if you get back an attributeID that wasn't checked by any of the above conditions, is that also a failure?

src/python_testing/TC_CC_10_1.py
lpbeliveau-silabs: Noted thanks!

src/python_testing/TC_CC_10_1.py
lpbeliveau-silabs: No it isn't. The color loop attributes are expected to be in there. I just didn't add a colorloop command since an active color loop would modify the enhanced hue so it was more convenient to have color loops commands at the end of the test.

cecille (approved):

jmartinez-silabs (approved):

PR #34470, opened 2024-07-24
Zephyr should have mDNS implementation ?

QuentinCaldeira-eaton:
This PR is more like a POC, in order to know if it's the right way to go.
I'm working with the Zephyr platform to perform an OnNetwork commissioning. If I understand correctly,
- Device is already OnNetwork, so commissioning will be with DNS-SD
- Even if a device is on network, it should be commissioned to a Matter Network
- DNS-SD should publish a service like _matterc._udp or _tpc depending on the protocol. Extra parameters like vendor ID could be associated in order to filter
- OpenBasicCommissioning should publish the matter service (I guess, so if I write a platform implementation, it should be call).
- DnssdImpl is used for both commissioner and commissionnee ? I only implemented the commissioner part (the device)
- CHIP-TOOL will pair the first device corresponding to the _matterc service (with filters optional)

Is there something special to enable in the app or in the build system, because when I debug the app with the DnssdImpl on the zephyr platform, I never go into, so I can't test if it's working for now. 

If someone could tell me if it's the right way to go, it could be really helpful !
Thanks :)


src/platform/Zephyr/BUILD.gn
```diff
@@ -71,6 +74,13 @@ static_library("Zephyr") {
     }
   }
 
+  if (chip_mdns == "minimal") {
```
Damian-Nordic, line 77: This doesn't make sense. "minimal" is the implementation that is platform-agnostic and doesn't use `ChipDnssdXXX` APIs at all. If you use "minima", you don't need any Zephyr-specific bits except for multicast support. `ChipDnssdXXX` APIs are only relevant for chip_mdns="platform".

Damian-Nordic: Is your device Thread-based? In some issue you were asking about Ethernet driver.

QuentinCaldeira-eaton: My device is Ethernet-based. I only use Zephyr SDK, no ST Microelectronics things.

PR #34472, opened 2024-07-24
[ESP32] Fix compiling examples with insights

shubhamdp:
Compilation may break if esp-insights is not used through component manager.

Check if the component is present in the component library before adding it to static library.


dhrishi:
@shubhamdp Is the compilation tested with both the cases - standalone insights as well as component based?

shubhamdp:
@dhrishi Yes compilation is tested in both cases. Also, tested with esp-rainmaker.

dhrishi (approved):

bzbarsky-apple (approved):

PR #34475, opened 2024-07-24
VS Code Python Formatting 

Alami-Amine:
- Adding python formatting settings (enforcing maxline length to 132) for autopep8 (python formatter) to be inline with restyled settings (#32384)
- Adding `autopep8` and `isort` as recommended extensions when installing a new VSCode workspace (since `isort` is also part of restyled settings #24148)

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -21,7 +21,9 @@
         "vadimcn.vscode-lldb",
         "xaver.clang-format",
         "yuichinukiyama.vscode-preview-server",
-        "yzhang.markdown-all-in-one"
+        "yzhang.markdown-all-in-one",
+        "ms-python.autopep8",
+        "ms-python.isort"
     ],
     // List of extensions recommended by VS Code that should not be recommended for users of this workspace.
     "unwantedRecommendations": []
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -192,5 +192,5 @@
     "[python]": {
         "editor.defaultFormatter": "ms-python.autopep8"
     },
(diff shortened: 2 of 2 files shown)


andy31415 (approved):

marcelveldt:
Why not go straight for ruff and skip autopep8 and isort ? 
That is a modern tool that can do multiple checks and lint at once and is superfast at it.

Also 132 characters for Python code is not very common. Most projects choose 100 or even 88 characters...

Just my 2 cents as I have always wondered a bit why the python code in the sdk is not exactly pep formatted.

We can do a PR to adjust it but it will also mean a very large diff as all code will be re-formatted.
Maybe that is the reason to start a bit cautious ?

tehampson (approved):

bzbarsky-apple (approved):

Alami-Amine:
> Why not go straight for ruff and skip autopep8 and isort ? That is a modern tool that can do multiple checks and lint at once and is superfast at it.
> 
> Also 132 characters for Python code is not very common. Most projects choose 100 or even 88 characters...
> 
> Just my 2 cents as I have always wondered a bit why the python code in the sdk is not exactly pep formatted.
> 
> We can do a PR to adjust it but it will also mean a very large diff as all code will be re-formatted. Maybe that is the reason to start a bit cautious ?

- using ruff might be a good idea, but the idea of this PR is just to ensure consistency between the settings in VSCode and those used by restyled. 
- and then regarding changing the line settings, as you said, we would have many large diffs and it will pollute the project history, maybe this can be discussed and we see from there?

PR #34476, opened 2024-07-24
2nd attempt to fix random EEM and EPM CI failures

jamesharrow:
After more review of the troublesome failures - it looks like the failures tend to be in the EEM cluster when checking that 2 energy readings are different values.

The reason for this is because we compute a random "power" in mW - e.g. -3000000 +/- 20000mW (i.e. somewhere between -2980W and -3020W.

But then we convert this into energy over 5 seconds in mWh by x5 / 3600s = so our values become in the range 4138 to 4194.

Or in the case of the Load - we generate random power values in the range 1000000 +/- 20000mW every 2 seconds.
So the energy values are only in the range 544 to 566.

So sooner or later we will get 2 consecutive values which are the same (even if we skipped a few values in between).

So adding a fixed random seed _should_ help this when running under CI.

Since the previous attempt at introducing more delay hasn't helped, I have reverted the change to the test scripts to remove these delays, and this means that the test plans do not need to change - so issue https://github.com/CHIP-Specifications/chip-test-plans/issues/4284 can be closed.


andy31415 (approved):

bzbarsky-apple (approved):

andy31415:
@jamesharrow could you fix merge conflicts?

PR #34484, opened 2024-07-24
Add the missing implementation for BdxTransferSession::RejectTransfer.

harimau-qirex:
BdxTransferSession has a method declaraction for RejectTransfer, but there's no implementation for it. This adds an implementation that rejects the transfer by sending a status message and changes the session's state to kTransferDone.


src/protocols/bdx/BdxTransferSession.cpp
```diff
@@ -261,6 +261,17 @@ CHIP_ERROR TransferSession::AcceptTransfer(const TransferAcceptData & acceptData
     return CHIP_NO_ERROR;
 }
 
+CHIP_ERROR TransferSession::RejectTransfer(StatusCode reason)
+{
+    VerifyOrReturnError(mState == TransferState::kNegotiateTransferParams, CHIP_ERROR_INCORRECT_STATE);
+    VerifyOrReturnError(mPendingOutput == OutputEventType::kNone, CHIP_ERROR_INCORRECT_STATE);
+
+    PrepareStatusReport(reason);
```
nivi-apple, line 269: should we call Reset as well here. There is a comment in the API saying - Use Reset() to prepare this object for another transfer. We have not set up a BDX session yet so maybe minimal cleanup is needed. Can you please check?

andy31415 (approved):

src/protocols/bdx/BdxTransferSession.cpp
harimau-qirex: My understanding is whatever calls `RejectTransfer` is what needs to call `Reset` to reuse the session object, and `RejectTransfer` shouldn't prematurely reset the internal state instead in case the caller needs information about the rejected transfer.

nivi-apple: sorry. i misread the documentation.

nivi-apple (approved):

PR #34486, opened 2024-07-24
TC-OCC-2.1 python test script

jaehs6sam:
Created a python test script for TC-OCC-2.1 test case under occupancy sensing cluster 1.4 release test plan.


jaehs6sam:
made a mistake.  do not merge.

PR #34488, opened 2024-07-24
[LevelControl] Implemented the Q quality logic for the CurrentLevel a…

### LINKED ISSUE #33861: Add implementation for added Q quality in Level Control
https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/9064 and https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/9181

Adds the Q quality to RemainingTime, CurrentLevel and CurrentFrequency. The SDK needs to be updated to match the rules proposed by the spec on how often these values SHALL be reported. 

Note: CurrentFrequency is properly not relevant, since the Frequency feature is Provisional.

jmartinez-silabs:
…nd RemainingTime attributes

CurrentLevel attribute is reported on at most every second, but usually 4 times throughout the transition or at transition start and end or if the attribute goes from or to `Null`.

RemainingTime is reported on value increase, or when it changes from or to `0`

I also cleaned up some namespace usage in the level-control.cpp file

#fixes #33861

Tests:
 - Existing Level control cert tests.
 - Manual test by subscribing to the attributes

### DIFF
--- a/src/app/clusters/level-control/level-control.cpp
+++ b/src/app/clusters/level-control/level-control.cpp
@@ -25,6 +25,7 @@
 #include <app-common/zap-generated/cluster-objects.h>
 #include <app/CommandHandler.h>
 #include <app/ConcreteCommandPath.h>
+#include <app/cluster-building-blocks/QuieterReporting.h>
 #include <app/util/attribute-storage.h>
 #include <app/util/config.h>
 #include <app/util/util.h>
@@ -51,6 +52,7 @@
 #include <assert.h>
 
 using namespace chip;
+using namespace chip::app;
 using namespace chip::app::Clusters;
 using namespace chip::app::Clusters::LevelControl;
 using chip::Protocols::InteractionModel::Status;
@@ -98,6 +100,8 @@ typedef struct
     uint32_t transitionTimeMs;
     uint32_t elapsedTimeMs;
     CallbackScheduleState callbackSchedule;
+    QuieterReportingAttribute<uint8_t> quietCurrentLevel{ DataModel::NullNullable };
+    QuieterReportingAttribute<uint16_t> quietRemainingTime{ DataModel::MakeNullable<uint16_t>(0) };
 } EmberAfLevelControlState;
 
 static EmberAfLevelControlState stateTable[kLevelControlStateTableSize];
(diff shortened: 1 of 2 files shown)


src/app/clusters/level-control/level-control.cpp
```diff
@@ -363,18 +369,68 @@ static void reallyUpdateCoupledColorTemp(EndpointId endpoint)
...
+        // At the start or end of the movement/transition we must report
+        auto predicate = [](const decltype(state->quietCurrentLevel)::SufficientChangePredicateCandidate &) -> bool {
+            return true;
+        };
+        dirtyState = state->quietCurrentLevel.SetValue(newValue, now, predicate);
+    }
+    else
+    {
+        // During transtions, reports should be at most once per second
+        System::Clock::Milliseconds64 reportInterval =
```
tcarmelveilleux, line 406: We should consider, in a follow-up to make this longer (e.g. 5 seconds, as per one of my recommendations in spec PR). This is strictly better than today, so we can go with this for now.

tcarmelveilleux (approved):

src/app/clusters/level-control/level-control.cpp
jmartinez-silabs: Agreed, I applied what you suggested before,  for a long transition we will report at max Transition time / 4 so anything longer than 4 seconds, we would start seeing benefits.

mkardous-silabs (approved):

src/app/clusters/level-control/level-control.cpp
```diff
@@ -363,18 +369,68 @@ static void reallyUpdateCoupledColorTemp(EndpointId endpoint)
 }
 #endif // IGNORE_LEVEL_CONTROL_CLUSTER_OPTIONS && MATTER_DM_PLUGIN_COLOR_CONTROL_SERVER_TEMP
 
+/*
+ * @brief
+ * This function is used to update the current level attribute
+ * while respecting it's defined quiet reporting quality:
```
bzbarsky-apple, line 375: "its"

src/app/clusters/level-control/level-control.cpp
```diff
@@ -363,18 +369,68 @@ static void reallyUpdateCoupledColorTemp(EndpointId endpoint)
...
+ * The attribute will be reported:
+ * - At most once per second, or
+ * - At the start of the movement/transition, or
+ * - At the end of the movement/transition, or
+ * - When it changes from null to any other value and vice versa.
+ *
+ * @param endpoint: endpoint on which the currentLevel attribute must be updated.
```
bzbarsky-apple, line 382: "CurrentLevel", not "currentLevel".

src/app/clusters/level-control/level-control.cpp
```diff
@@ -363,18 +369,68 @@ static void reallyUpdateCoupledColorTemp(EndpointId endpoint)
...
+        // During transtions, reports should be at most once per second
+        System::Clock::Milliseconds64 reportInterval =
+            std::max(System::Clock::Milliseconds64(1000), System::Clock::Milliseconds64(state->transitionTimeMs / 4));
+        auto predicate = state->quietCurrentLevel.GetPredicateForSufficientTimeSinceLastDirty(reportInterval);
+        dirtyState     = state->quietCurrentLevel.SetValue(newValue, now, predicate);
+    }
+
+    if (dirtyState == AttributeDirtyState::kMustReport)
+    {
+        markDirty = MarkAttributeDirty::kIfChanged;
```
bzbarsky-apple, line 414: Are we guaranteed that we can't have `isStartOrEndOfTransition` true, the value not changing, and reporting needing to be triggered?

Also, why is markDirty declared all the way at the start of the function?

src/app/clusters/level-control/level-control.cpp
```diff
@@ -478,11 +535,20 @@ static void writeRemainingTime(EndpointId endpoint, uint16_t remainingTimeMs)
...
+        // The quiet report is determined by the previously set policies:
+        // - kMarkDirtyOnChangeToFromZero : When the value changes from 0 to any other value and vice versa, or
+        // - kMarkDirtyOnIncrement : When the value increases.
+        if (state->quietRemainingTime.SetValue(remainingTimeDs, now) == AttributeDirtyState::kMustReport)
         {
-            ChipLogProgress(Zcl, "ERR: writing remaining time %x", to_underlying(status));
+            markDirty = MarkAttributeDirty::kIfChanged;
         }
+
+        Attributes::RemainingTime::Set(endpoint, state->quietRemainingTime.value().ValueOr(0), markDirty);
```
bzbarsky-apple, line 551: Why the `ValueOr(0)`?  We just set it to a non-null value, so why is this not just `Value()`?

src/app/clusters/level-control/level-control.cpp
```diff
@@ -1255,14 +1314,13 @@ static void stopHandler(app::CommandHandler * commandObj, const app::ConcreteCom
 
     if (!shouldExecuteIfOff(endpoint, commandId, optionsMask, optionsOverride))
     {
-        status = Status::Success;
         goto send_default_response;
     }
 
     // Cancel any currently active command.
     cancelEndpointTimerCallback(endpoint);
+    SetCurrentLevelQuietReport(endpoint, state, state->quietCurrentLevel.value(), true /*isStartOrEndOfTransition*/);
```
bzbarsky-apple, line 1322: So this one is a bit complicated: if we are in fact at start of transition, but the the current value of quietCurrentLevel has not been reported, then we should in fact report here (right?), and I don't think we will given the code above.

It looks like we need a MarkAttributeDirty::kYes value, probably?

PR #34492, opened 2024-07-24
Added TC-OCC-2.3 Python test script

jaehs6sam:
Created a python test script for TC-OCC-2.3 test case under occupancy sensing cluster 1.4 release test plan.

### DIFF
--- a/src/python_testing/TC_OCC_2_3.py
+++ b/src/python_testing/TC_OCC_2_3.py
@@ -1,0 +1,106 @@
+#
+#    Copyright (c) 2023 Project CHIP Authors
+#    All rights reserved.
+#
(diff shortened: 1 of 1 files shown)


src/python_testing/TC_OCC_2_3.py
```diff
@@ -0,0 +1,106 @@
+#
+#    Copyright (c) 2023 Project CHIP Authors
```
bzbarsky-apple, line 2: Wrong year.

jaehs6sam:
Will consolidate all TC-OCC-xx test script PRs (PR 34490, PR 34491, PR 34492, PR 34523) to one PR.

PR #34496, opened 2024-07-25
Add draft test scripts for WIFINM cluster

ksperling-apple:


ksperling-apple:
I'm going to include this script in #34469 because it depends on CI script changes from that PR to work.

PR #34504, opened 2024-07-25
[nrfconnect] Updated documentation for all nrfconnect examples

kkasperczyk-no:
Updated docs for all nrfconnect examples to:
* be aligned with the most recent NCS 2.7.0 version
* mention that the examples are not production ready and redirect to the production ready NCS Matter samples

### DIFF
--- a/examples/all-clusters-app/nrfconnect/README.md
+++ b/examples/all-clusters-app/nrfconnect/README.md
@@ -1,5 +1,21 @@
 # Matter nRF Connect All Clusters Example Application
 
+> **Note:** This example is intended only to perform smoke tests of a Matter
+> solution integrated with nRF Connect SDK platform. The example quality is not
+> production ready and it may contain minor bugs or use not optimal
+> configuration. It is not recommended to use this example as a basis for
+> creating a market ready product.
+>
+> For the production ready and optimized Matter samples, see
+> [nRF Connect SDK samples](https://docs.nordicsemi.com/bundle/ncs-latest/page/nrf/samples/matter.html).
+> The Matter samples in nRF Connect SDK use various additional software
+> components and provide multiple optional features that improve the developer
+> and user experience. To read more about it, see
+> [Matter support in nRF Connect SDK](https://docs.nordicsemi.com/bundle/ncs-latest/page/nrf/protocols/matter/index.html#ug-matter)
+> page. Using Matter samples from nRF Connect SDK allows you to get a full
+> Nordic technical support via [DevZone](https://devzone.nordicsemi.com/)
+> portal.
+
 The nRF All Clusters Example Application implements various ZCL clusters
 populated on three endpoints. You can use this example as a reference for
 creating your own application.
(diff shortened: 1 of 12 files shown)


Damian-Nordic (approved):

examples/all-clusters-app/nrfconnect/README.md
```diff
@@ -248,14 +264,15 @@ Complete the following steps to build the sample:
 
 2.  Run the following command to build the example, with _build-target_ replaced
     with the build target name of the Nordic Semiconductor's kit you own, for
-    example `nrf52840dk_nrf52840`:
+    example `nrf52840dk/nrf52840`:
 
-         $ west build -b build-target
+         $ west build -b build-target --sysbuild
```
markaj-nordic, line 269: nit: Of course it won't hurt to have some redundancy here, but in general (except from some unusual cases :) ) the sysbuild is used by default. Maybe we could add some note about (it affects the development and related tooling less without --sysbuild flag).

examples/all-clusters-app/nrfconnect/README.md
```diff
@@ -270,7 +287,7 @@ following command:
 To build the example with release configuration that disables the diagnostic
 features like logs and command-line interface, run the following command:
 
-    $ west build -b build-target -- -DCONF_FILE=prj_release.conf
+    $ west build -b build-target --sysbuild -- -DCONF_FILE=prj_release.conf
```
markaj-nordic, line 290: And what about using `FILE_SUFFIX` instead of `CONF_FILE`?

markaj-nordic (approved):
Looking good. Left some nits. A general question, should we point out somewhere that child_images are still supported for now?

examples/all-clusters-app/nrfconnect/README.md
kkasperczyk-no: Unfortunately sysbuild is not used by default :) And sadly this is necessary to add if you are outside of nRF Connect SDK repo (--sysbuild is added only for dependent repos to NCS, and project-connectedhomeip is not one of them).

examples/all-clusters-app/nrfconnect/README.md
kkasperczyk-no: Good catch! Thanks!

examples/all-clusters-app/nrfconnect/README.md
```diff
@@ -1,5 +1,21 @@
 # Matter nRF Connect All Clusters Example Application
 
+> **Note:** This example is intended only to perform smoke tests of a Matter
+> solution integrated with nRF Connect SDK platform. The example quality is not
+> production ready and it may contain minor bugs or use not optimal
+> configuration. It is not recommended to use this example as a basis for
+> creating a market ready product.
```
fessehaeve, line 7: As a basis .. reads to me like as a starting point. Isn't that the whole point of this SDK - to act as a starting point?

tcarmelveilleux (approved):

examples/all-clusters-app/nrfconnect/README.md
```diff
@@ -1,5 +1,21 @@
 # Matter nRF Connect All Clusters Example Application
 
+> **Note:** This example is intended only to perform smoke tests of a Matter
```
fessehaeve, line 3: As this Note is a new thing, is it possible to add that this is introduced from Matter x.x or NCS x.x.x version?

kkasperczyk-no: I would not like to do that. It would suggest that before the Matter x.x or NCS x.x.x it was recommended to use this examples instead of the ones located in NCS. We have always recommended to use samples from NCS as a reference, while this examples were only to perform the newest features integration as soon as possible and we couldn't promise they are production ready.

Unfortunately this was a mistake from our side, that we haven't communicated it in a more clear way from the beginning. But we do see customers that struggle with multiple problems using this examples, that are not present in samples from NCS, so we would like to try to minimize the risk and improve the user experience for our platform.

examples/all-clusters-app/nrfconnect/README.md
kkasperczyk-no: Maybe this should be called a reference? Well, this actually depends how the user is going to understand the reference word :smile: What I would like to achieve is inform someone that "Hey you can use this example to play with the platform, or even create an own app, but please be careful, as this is not tested enough to call it production ready. If you want to build an own product, I recommend to use samples from NCS as a reference.". Does it sound right to you, or I should try to use other wording to not make it confusing?

PR #34515, opened 2024-07-25
Add global types (structs/enums/bitmaps) to matter file parsing

andy31415:
### Changes

- supports `struct/bitmap/enum` at a top level in matter files
- parsing by default will copy over global types into the clusters that reference them (so code generation that is not aware of global types will generate non-global type code)
- Updated idl text generation
- Updated backwards compatibility checker
- added unit tests and made the idl_parser unit test print out much nicer errors on failures


bzbarsky-apple (approved):
More or less rubber-stamping; we will see what the generated IDL looks like in the end.

scripts/py_matter_idl/matter_idl/matter_idl_parser.py
```diff
@@ -504,15 +505,25 @@ def idl(self, items):
...
+        global_bitmaps = []
+        global_enums = []
+        global_structs = []
+
         for item in items:
             if isinstance(item, Cluster):
                 clusters.append(item)
             elif isinstance(item, Endpoint):
                 endpoints.append(item)
+            elif isinstance(item, Enum):
```
cecille, line 517: just to clarify - do you know these are global because the non-global items are somewhere else? In the cluster?

scripts/py_matter_idl/matter_idl/matter_idl_parser.py
```diff
@@ -524,9 +535,84 @@ def c_comment(self, token: Token):
...
+    """
+    Maintains global type mapping from an IDL
+    """
+
+    def __init__(self, idl: Idl):
+        self.bitmap_map = {b.name: b for b in idl.global_bitmaps}
+        self.enum_map = {e.name: e for e in idl.global_enums}
+        self.struct_map = {s.name: s for s in idl.global_structs}
+
+        self.global_types = set(self.bitmap_map.keys()).union(set(self.enum_map.keys())).union(set(self.struct_map.keys()))
```
cecille, line 567: Are these names guaranteed to be unique?

cecille (approved):

scripts/py_matter_idl/matter_idl/matter_idl_parser.py
andy31415: Yes, the name of the method is the lark toke name. So when I get `idl` called, it means all the elements passed in are the list of things that were parsed at the top level (i.e. in lark we say `idl: (struct|enum|bitmap|cluster|endpoint)*` and all these are at a top level guaraneteed. The individual enum/bitmap/struct at the cluster level are passed to the `cluster` method.

scripts/py_matter_idl/matter_idl/matter_idl_parser.py
andy31415: Not by spec :( . 

Conceptually they must be because otherwise if I have `enum Foo` and `bitmap Foo` both defined, then when I have somewhere `cluster Bar { attribute Foo foo = 1 }` it is unclear how to determine what it is. Even zap I thin has `type=....` and we do not actually say what namespace to use.

I will add a check.

bzbarsky-apple: By spec, type names really do have to be unique.  Otherwise referencing types by name does not work.

PR #34520, opened 2024-07-25
API renames: `chip::app::InteractionModel::*` becomes `chip::app::DataModel` and renamed `DataModel` to `Provider`

andy31415:
the usage of `InteractionModel` felt off in code reviews and clashed with things like `Protocols::InteractionModel`.

Renamed the namespace to `DataModel` (which already exists) and renamed the `DataModel` class (and related) to `Provider`.

This change was mostly renames + sed + restyle. No functionality or logic changes should exist.


src/app/data-model-provider/ActionContext.h
```diff
@@ -20,7 +20,7 @@
 
 namespace chip {
 namespace app {
-namespace InteractionModel {
+namespace DataModel {
 
 // Context for a currently executing action
 class ActionContext
```
bzbarsky-apple, line 26: Actions are an interaction model concept, not a data model one...

src/app/data-model-provider/OperationTypes.h
```diff
@@ -26,7 +26,7 @@
 
 namespace chip {
 namespace app {
-namespace InteractionModel {
+namespace DataModel {
 
 /// Contains common flags among all interaction model operations: read/write/invoke
 enum class OperationFlags : uint32_t
```
bzbarsky-apple, line 32: Again, these seem like they should be under IM, not DM.

src/app/data-model-provider/ActionContext.h
andy31415: I can move them around. My typical expectation was that a folder generally operates under a single namespace, hence the bulk rename.

Should I just go ahead and move actioncontext and operationtypes into IM instead of DM? We would still then have the namespace conflict whenever we include these files (and they are included in interactionmodelengine, so often visible) ... however beyond that I don't have strong opinions, I just want some usable naming that is stable.

bzbarsky-apple: Ah, hmm.  How about we merge this, then think about naming for these bits, then.  The name conflicts sound annoying.

bzbarsky-apple (approved):

tcarmelveilleux (approved):

PR #34522, opened 2024-07-25
Add idea for handling Q quality

rbultman:


src/protocols/interaction_model/StatusCodeList.h
```diff
@@ -74,5 +74,7 @@ CHIP_IM_STATUS_CODE(TimedRequestMismatch  , TIMED_REQUEST_MISMATCH  , 0xc9)
 CHIP_IM_STATUS_CODE(FailsafeRequired      , FAILSAFE_REQUIRED       , 0xca)
 CHIP_IM_STATUS_CODE(InvalidInState        , INVALID_IN_STATE        , 0xcb)
 CHIP_IM_STATUS_CODE(NoCommandResponse     , NO_COMMAND_RESPONSE     , 0xcc)
+CHIP_IM_STATUS_CODE(SuppressReport        , SUPPRESS_REPORT         , 0xcd)
+CHIP_IM_STATUS_CODE(ForceReport           , FORCE_REPORT            , 0xce)
```
bzbarsky-apple, line 78: This is the list of spec-defined status codes.  We can't just randomly add things to it....

In particular, 0xCD and 0xCE are in fact defined in the 1.4 spec and just haven't been added here yet.

src/app/clusters/fan-control-server/fan-control-server.cpp
```diff
@@ -298,6 +298,17 @@ MatterFanControlClusterServerPreAttributeChangedCallback(const ConcreteAttribute
         }
         break;
     }
+    case PercentCurrent::Id: {
+        res = Status::Success;
+        // ReportRequiredByWhatever - Something that get's triggered on edges to force reporting
+        if (ReportRequiredByWhatever) {
```
bzbarsky-apple, line 304: OK, but where do we store the state to compute this value?  And at that point, it seems like it should also just store the last-reported PercentCurrent value and cluster users should call a cluster API to set the attribute instead of directly poking the attribute store....

src/app/util/attribute-table.cpp
```diff
@@ -416,9 +416,21 @@ Status emAfWriteAttribute(EndpointId endpoint, ClusterId cluster, AttributeId at
...
-        return imStatus;
+        case Protocols::InteractionModel::Status::ForceReport: {
+            markDirty = MarkAttributeDirty::kYes;
+            break;
+        }
+        case Protocols::InteractionModel::Status::SuppressReport: {
+            markDirty = MarkAttributeDirty::kNo;
+            break;
+        }
+        case Protocols::InteractionModel::Status::imStatus: {
```
rbultman, line 429:
```suggestion
        case Protocols::InteractionModel::Status:: Success: {
```

src/protocols/interaction_model/StatusCodeList.h
rbultman: Noted.  Seems like this would be useful to other clusters having Q-quality attributes and not using the AttibuteAccessInterface?

src/app/clusters/fan-control-server/fan-control-server.cpp
rbultman: Right, that would be the next step.  I had similar thoughts about local storage.  Wanted to see if the overall approach (enforce in the PreAttributeChangeCallback with support from the IM for the actual reporting or not) seemed sane.

src/app/clusters/fan-control-server/fan-control-server.cpp
```diff
@@ -298,6 +312,40 @@ MatterFanControlClusterServerPreAttributeChangedCallback(const ConcreteAttribute
...
+            {
+                // need a report at the start of a move.
+                percentCurrentQTracking[ep].moveStarted = false;
+                percentCurrentQTracking[ep].lastUpdateTime = GetTimeNowSomehow();
+                res = Status::ForceReport;
+            }
+            else
+            {
+                SomeTimeType currentTime = GetTimeNowSomehow();
+                if (currentTime - percentCurrentQTracking[ep].lastUpdateTime >= 1) {
```
mhazley, line 336: Magic number, worth defining somewhere

src/app/clusters/fan-control-server/fan-control-server.cpp
```diff
@@ -50,6 +50,15 @@ static_assert(kFanControlDelegateTableSize <= kEmberInvalidEndpointIndex, "FanCo
 
 Delegate * gDelegateTable[kFanControlDelegateTableSize] = { nullptr };
 
+typedef struct PercentCurrentQTracking {
+    bool initialized;
+    bool moveStarted;
+    DataModel::Nullable<chip::Percent> startPercent;
+    DataModel::Nullable<chip::Percent> endPercent;
+    auto lastUpdateTime;
```
mhazley, line 58: Might be worth explicitly initialising to zero in case we get a current update without a setting update

src/app/clusters/fan-control-server/fan-control-server.cpp
```diff
@@ -298,6 +312,40 @@ MatterFanControlClusterServerPreAttributeChangedCallback(const ConcreteAttribute
...
+        }
+        else
+        {
+            // default to suppressing a report
+            res = Status::SuppressReport;
+            if (percentCurrentQTracking[ep].moveStarted)
+            {
+                // need a report at the start of a move.
+                percentCurrentQTracking[ep].moveStarted = false;
+                percentCurrentQTracking[ep].lastUpdateTime = GetTimeNowSomehow();
```
bzbarsky-apple, line 330: If we're going to store this stuff out-of-band like this anyway, it's probably a good idea to use src/app/cluster-building-blocks/QuieterReporting.h bits to store a bunch of this state.

rbultman:
Closing as this is super old and no longer really valid.

PR #34525, opened 2024-07-26
Spec rename of HomeLocationStruct to LocationDescriptorStruct

tehampson:
https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/10115 changed struct name from HomeLocationStruct to LocationDescriptorStruct. This reflects the spec text change

Backwards compatibility checker is skipped because ServiceArea is a provisional cluster and is marked as such now in this PR


src/app/zap-templates/zcl/data-model/chip/ecosystem-information-cluster.xml
```diff
@@ -30,7 +30,7 @@ limitations under the License.
   <struct name="EcosystemLocationStruct" isFabricScoped="true" apiMaturity="provisional">
     <cluster code="0x0750"/>
     <item fieldId="0" name="UniqueLocationID" type="char_string" isFabricSensitive="true" length="64"/>
-    <item fieldId="1" name="LocationDescriptor" type="HomeLocationStruct" isFabricSensitive="true"/>
+    <item fieldId="1" name="LocationDescriptor" type="LocationDescriptorStruct" isFabricSensitive="true"/>
```
bzbarsky-apple, line 33: This type is called "locationdesc" in the spec in some places and "LocationDescriptorStruct" in other places... so as far as I can tell, our XML here is not in fact matching the spec, thought that's largely because the spec is completely broken....

But this needs to be addressed in the spec, after which we might need to change the XML some more depending on what the spec ends up doing.

bzbarsky-apple (approved):

yufengwangca (approved):

src/app/zap-templates/zcl/data-model/chip/ecosystem-information-cluster.xml
andy31415: `locationdesc` would be a poor name ... assuming this comment is not intended to stop this PR but rather encourage fixing the spec @cletnick

andy31415: https://github.com/CHIP-Specifications/connectedhomeip-spec/issues/10152 exists for the spec. Marking this closed.

PR #34526, opened 2024-07-26
TC-SWTCH-2.3: Implement

cecille:


src/python_testing/matter_testing_support.py
```diff
@@ -238,7 +238,7 @@ def __init__(self, expected_cluster: ClusterObjects.Cluster):
         self._q = queue.Queue()
         self._expected_cluster = expected_cluster
 
-    async def start(self, dev_ctrl, node_id: int, endpoint: int, fabric_filtered: bool = False, min_interval_sec: int = 0, max_interval_sec: int = 30) -> Any:
+    async def start(self, dev_ctrl, node_id: int, endpoint: int, fabric_filtered: bool = False, min_interval_sec: int = 0, max_interval_sec: int = 5) -> Any:
```
tcarmelveilleux, line 241: Why 5 instead of 30? Battery powered  Thread switches may not be able to do 5. Suggest keeping 30

tcarmelveilleux (approved):

src/python_testing/matter_testing_support.py
cecille: If the ceiling is 30 and we're waiting for 30 seconds, we might not get reports at all. Device can adjust the ceiling - this is just a request.

cecille, line 241:
```suggestion
    async def start(self, dev_ctrl, node_id: int, endpoint: int, fabric_filtered: bool = False, min_interval_sec: int = 0, max_interval_sec: int = 30) -> Any:
```

andy31415:
Fast tracking test-only change

PR #34534, opened 2024-07-26
Check for FabricSynchronization condition on Aggregator device

arkq:
Simple test script to verify FabricSynchronization condition on the Bridged Node device type.

Spec [bits under the `in-progress,matter-fabric-synchronization` condition]: https://github.com/CHIP-Specifications/connectedhomeip-spec/blob/master/src/device_types/BridgedNode.adoc

Currently this test script does not pass, because the fabric-bridge-app lacks some features.


src/app/tests/suites/TestFabricSyncBridgedNode.yaml
```diff
@@ -0,0 +1,58 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Test Fabric Synchronization condition on Bridged Node Device Type
+
+config:
+    nodeId: 0x12344321
```
cecille, line 18: you'll want to add a default endpoint here for the ci

src/app/tests/suites/TestFabricSyncBridgedNode.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+tests:
+    - label:
+          "Read the DeviceTypeList attribute of the Descriptor cluster and
+          check whether the device type is Aggregator."
+      endpoint: 1
```
cecille, line 24: If you put the endpoint in the test step, it's always going to run this step against that endpoint.

Suggest adding endpoint to the config at the top with the CI default. That can be overridden in the TH / on the command line. Then remove from the individual steps.

src/app/tests/suites/TestFabricSyncBridgedNode.yaml
```diff
@@ -0,0 +1,58 @@
...
+              type: list
+              contains: [
+                  0x0751, # Commissioner Control Cluster
+              ]
+
+    - label:
+          "Read the SupportedDeviceCategories attribute of the Commissioner
+          Control cluster and check whether the FabricSynchronization bit is
+          set."
+      endpoint: 0
```
cecille, line 51: Is the commissioner controller cluster mandated to be on ep0?

tehampson: Right now it is a little ambiguous if this is supposed to be on the root endpoint or aggregator endpoint. Right now example app assumed root endpoint.

This is one of those things were we are going to assume root node for the 29th and get clarification on spec and update accordingly

src/app/tests/suites/TestFabricSyncBridgedNode.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - label:
+          "Read the DeviceTypeList attribute of the Descriptor cluster and
+          check whether the device type is Aggregator."
+      endpoint: 1
+      cluster: "Descriptor"
+      command: "readAttribute"
+      attribute: "DeviceTypeList"
+      response:
+          value: [{
+              DeviceType: 14, # Aggregator
```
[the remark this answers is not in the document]
arkq: Yes, it works (or at least it seems that it works), the step does not report error.

src/app/tests/suites/TestFabricSyncBridgedNode.yaml
```diff
@@ -0,0 +1,58 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Test Fabric Synchronization condition on Bridged Node Device Type
+
+config:
+    nodeId: 0x12344321
+
```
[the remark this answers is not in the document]
arkq: OK, done.

src/app/tests/suites/TestFabricSyncBridgedNode.yaml
arkq: Done

src/app/tests/suites/TestFabricSyncBridgedNode.yaml
arkq: I've added dedictaed config for that, so once the spec will be clarified and the app fixed, one can change the  Commissioner Control cluster number

src/app/tests/suites/TestFabricSyncBridgedNode.yaml
arkq: Done

tehampson (approved):

src/app/tests/suites/TestFabricSyncBridgedNode.yaml
```diff
@@ -0,0 +1,64 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Test Fabric Synchronization condition on Bridged Node Device Type
+
+PICS:
+    - MCORE.BRIDGE.FS
```
tehampson, line 18: nit I think this is supposed to be `MCORE.FS`

tehampson:
>Currently this test script does not pass, because the fabric-bridge-app lacks some features.

I would split this up into yaml and all the CI test stuff separately so that you don't block getting the yaml test in

src/app/tests/suites/TestFabricSyncBridgedNode.yaml
jlatusek: Hi, as far as I know this has been discussed with Chris Letnick and as per him `MCORE.BRIDGE.FS` makes sense so we can retain the same since we are verifying the FS feature specifically on a Bridge node.

jlatusek:
@tehampson I have created a separate PR #34626. Unfortunately @arkq is on holiday and I'm not able to update this PR.

src/app/tests/suites/TestFabricSyncBridgedNode.yaml
tehampson: The bridge and admin are one DUT. You can't test one without the other. I would argue we should simplify the PICS and just use MCORE.FS. That is my 2cents

saurabhst: @tehampson agree It's just that we wanted to be more specific so that the PICS name itself clarifies that we are verifying FS on bridge node in particular. @cletnick thoughts ?

saurabhst (approved):

PR #34537, opened 2024-07-26
Update ACL API to support ARL use cases

tleacmcsa:
ARL needs to know what attribute/command/event is being accessed and with which interaction model action. This change brings this knowledge into the AccessControl classes Check method for further enhancement with ARL feature on top.


src/access/AccessControl.h
```diff
@@ -25,13 +25,16 @@
...
 #include <lib/support/CodeUtils.h>
+#include <protocols/interaction_model/Constants.h>
 
 // Dump function for use during development only (0 for disabled, non-zero for enabled).
 #define CHIP_ACCESS_CONTROL_DUMP_ENABLED 0
 
 namespace chip {
 namespace Access {
 
+using namespace chip::Protocols::InteractionModel;
```
andy31415, line 36: please no using namespace in headers.

src/access/AccessControl.h
```diff
@@ -629,13 +632,15 @@ class AccessControl
...
-     * requiring a privilege) should be allowed or denied.
+     * requiring a privilege) should be allowed or denied. The IM action
+     * is provided for context.
      *
      * @retval #CHIP_ERROR_ACCESS_DENIED if denied.
      * @retval other errors should also be treated as denied.
      * @retval #CHIP_NO_ERROR if allowed.
      */
-    CHIP_ERROR Check(const SubjectDescriptor & subjectDescriptor, const RequestPath & requestPath, Privilege requestPrivilege);
+    CHIP_ERROR Check(const SubjectDescriptor & subjectDescriptor, const RequestPath & requestPath, Privilege requestPrivilege,
```
andy31415, line 642: please fix the order to keep request last. It seems the intent was request data (descriptor, path) to be first, then the request last.

Also this starts growing into having too many arguments. We may decide to have a builder pattern starting now, something like `Check(Request, Privilege)` where Request is a builder with SetDescriptor, SetPath, SetMsgType).

src/access/AccessControl.cpp
```diff
@@ -171,6 +171,23 @@ char GetPrivilegeStringForLogging(Privilege privilege)
     return 'u';
 }
 
+char GetActionStringForLogging(MsgType action)
```
andy31415, line 174: please fix return type or name: char is not a string.

andy31415: I see the previous implementation also did that :( So pre-existing.... sigh. Can be a followup that probably never happens.

tleacmcsa: The intent was to log a single character, not a string.

src/access/RequestPath.h
```diff
@@ -26,8 +26,10 @@ namespace Access {
 struct RequestPath
 {
     // NOTE: eventually this will likely also contain node, for proxying
-    ClusterId cluster   = 0;
-    EndpointId endpoint = 0;
+    ClusterId cluster     = 0;
+    EndpointId endpoint   = 0;
+    uint32_t entityId     = 0; // attribute, command, or event id.  Ignored if entityWildcarded is true
```
andy31415, line 31: how about using a std::optional? then nullopt is "wildcard" otherwise an actual id.

How does one distinguish between this being a attribute, command or event ID? The description should explain this. The "or" is confusing. The IDs here are of a different namespace and are not unique.

E.g.  AccessControl::AccessControlExtensionChanged is event 1, AccessControl::extension is attribute 1,

andy31415: I think comment should be that this is based on MsgType ... however then why is this separated out? Maybe requestpath should instead be a structure of `request` that contains all information.

src/access/AccessControl.h
```diff
@@ -629,13 +632,15 @@ class AccessControl
 
     /**
      * Check whether access (by a subject descriptor, to a request path,
-     * requiring a privilege) should be allowed or denied.
+     * requiring a privilege) should be allowed or denied. The IM action
+     * is provided for context.
```
bzbarsky-apple, line 636: It's not clear that the "IM action" is the right abstraction here, or at least it would need a lot more documentation.  For example, what happens if someone passes in MsgType::ReportData (which is the obvious action for a subscription report)?

It would be better if there were a clear "operation being performed" enum (more or less matching AccessRestrictionTypeEnum in the spec, but I think the access module has its own data types defined for things like that).  Because the fact that for events you should pass SubscribeRequest but for attributes you should pass ReadRequest even if the attribute is in fact being reported due to a subscription is (1) odd and (2) completely un-guessable from this API surface.

Plausible values for that enum would be: ReadAttribute, WriteAttribute, InvokeCommand, ReadEvent.

Also, fundamentally this value here affects how the entityId is interpreted in the RequestPath, right?  It seems like fundamentally we should be keeping that information together, in that RequestPath should have an additional member indicating what entityId really means, and we should not have a separate arg here.

src/app/reporting/Engine.cpp
```diff
@@ -333,10 +333,10 @@ CHIP_ERROR Engine::CheckAccessDeniedEventPaths(TLV::TLVWriter & aWriter, bool &
             aHasEncodedData = true;
         }
 
-        Access::RequestPath requestPath{ .cluster = current->mValue.mClusterId, .endpoint = current->mValue.mEndpointId };
+        Access::RequestPath requestPath{ .cluster = current->mValue.mClusterId, .endpoint = current->mValue.mEndpointId, .entityId = current->mValue.mEventId };
         Access::Privilege requestPrivilege = RequiredPrivilege::ForReadEvent(path);
 
-        err = Access::GetAccessControl().Check(apReadHandler->GetSubjectDescriptor(), requestPath, requestPrivilege);
+        err = Access::GetAccessControl().Check(apReadHandler->GetSubjectDescriptor(), requestPath, requestPrivilege, MsgType::SubscribeRequest);
```
andy31415, line 339: Engine handles both read and subscribe requests. Can we differentiate between them at this point?

src/access/RequestPath.h
```diff
@@ -26,8 +26,10 @@ namespace Access {
...
+    bool entityWildcarded = false;
```
bzbarsky-apple, line 32: entityWildcarded seems to be entirely unused (looking at the everything-together PR at https://github.com/project-chip/connectedhomeip/pull/34485.  Why is it here?

src/app/CommandHandlerImpl.cpp
```diff
@@ -404,7 +404,8 @@ Status CommandHandlerImpl::ProcessCommandDataIB(CommandDataIB::Parser & aCommand
         Access::SubjectDescriptor subjectDescriptor = GetSubjectDescriptor();
         Access::RequestPath requestPath{ .cluster = concretePath.mClusterId, .endpoint = concretePath.mEndpointId };
```
bzbarsky-apple, line 405: Don't we need to pass in the entityId here and in other places?  Looks like a few places like EventManagement were fixed to do that, but others were not.

It's unfortunate that we are using these structs with no way to force consumers to initialize the data, so the compiler is not catching this problem for us.  :(

bzbarsky-apple: Maybe we should:

1. static_assert that `kInvalidCommandId`, `kInvalidAttributeId`, and `kInvalidEventId` are all equal.
2. Initialize `entityId` to that value they're all equal to (e.g. to `kInvalidAttributeId`).
3. Make sure that Check() fails out immediately if the request path has the invalid value.

That way at least at runtime we will hopefully check places that forget to set the entity id.

src/access/RequestPath.h
tleacmcsa: Oops, missed it in the PR breakup. Good catch.

src/access/AccessControl.h
tleacmcsa: done.

src/access/AccessControl.h
tleacmcsa: done.

src/access/RequestPath.h
tleacmcsa: I believe I have addressed these concerns.

src/app/reporting/Engine.cpp
tleacmcsa: Here we are dealing with events and along with the RequestType I just added, I think I have addressed your concern.

src/access/AccessControl.h
tleacmcsa: Done.

src/access/RequestPath.h
tleacmcsa: Fixed.

src/app/CommandHandlerImpl.cpp
tleacmcsa: I believe all Check calls have populated requestType and entityId.  The more clear requestType should make it clear what entityId represents now.  Let me know if you still feel that more is needed.

src/access/AccessControl.h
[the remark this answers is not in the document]
tleacmcsa: done.

src/access/RequestPath.h
```diff
@@ -19,15 +19,29 @@
...
 struct RequestPath
 {
     // NOTE: eventually this will likely also contain node, for proxying
-    ClusterId cluster   = 0;
-    EndpointId endpoint = 0;
+    ClusterId cluster       = 0;
+    EndpointId endpoint     = 0;
+    RequestType requestType = RequestType::kRequestTypeUnknown;
+
+    //entityId represents an attribute, command, or event ID, which is determined by the requestType. Wildcard if omitted.
```
bzbarsky-apple, line 43: OK, so what request type is used for an event read (not subscribe)?

src/access/RequestPath.h
```diff
@@ -19,15 +19,29 @@
...
+    std::optional<uint32_t> entityId;
```
bzbarsky-apple, line 44: So wildcarding is used here only for "reading or subscribing to events" and "it's a wildcard path", right?

But then how is this meant to be implemented, exactly?  We really do need to know what events are involved to do the relevant checks correctly for ARL, no?

src/access/AccessControl.cpp
```diff
@@ -316,11 +333,12 @@ CHIP_ERROR AccessControl::Check(const SubjectDescriptor & subjectDescriptor, con
...
-                        "AccessControl: checking f=%u a=%c s=0x" ChipLogFormatX64 " t=%s c=" ChipLogFormatMEI " e=%u p=%c",
+                        "AccessControl: checking f=%u a=%c s=0x" ChipLogFormatX64 " t=%s c=" ChipLogFormatMEI " e=%u p=%c r=%c",
                         subjectDescriptor.fabricIndex, GetAuthModeStringForLogging(subjectDescriptor.authMode),
                         ChipLogValueX64(subjectDescriptor.subject),
                         GetCatStringForLogging(catLogBuf, sizeof(catLogBuf), subjectDescriptor.cats),
-                        ChipLogValueMEI(requestPath.cluster), requestPath.endpoint, GetPrivilegeStringForLogging(requestPrivilege));
+                        ChipLogValueMEI(requestPath.cluster), requestPath.endpoint, GetPrivilegeStringForLogging(requestPrivilege),
+                        GetRequestTypeStringForLogging(requestPath.requestType));
     }
 #endif // CHIP_PROGRESS_LOGGING && CHIP_CONFIG_ACCESS_CONTROL_POLICY_LOGGING_VERBOSITY > 1
```
bzbarsky-apple, line 344: Doesn't this need to fail out if the RequestPath is not well-formed (i.e. request type is unknown)?

src/access/AccessControl.h
```diff
@@ -25,6 +25,7 @@
 #include <lib/core/CHIPCore.h>
 #include <lib/core/Global.h>
 #include <lib/support/CodeUtils.h>
+#include <protocols/interaction_model/Constants.h>
```
bzbarsky-apple, line 28: Why is this needed?

hawk248 (approved):

src/access/RequestPath.h
tleacmcsa: The spec only considers restrictions on events for subscribe, not read.  If that ends up being a gap in the spec, we would address here after spec changes (which are now merged).

src/access/RequestPath.h
tleacmcsa: A wildcard here would only make sense when reading events or attributes, or subscribing to them, yes.

If a request is made using a wildcard, restriction checks would fail if there is any intersection between that and what is restricted.  Meaning, a single attribute/event restriction entry would  block a wildcard request if that wildcard would have included the entry.

Make sense?  Or am I missing something?

src/access/RequestPath.h
```diff
@@ -19,15 +19,29 @@
...
 namespace chip {
 namespace Access {
 
+enum class RequestType : uint8_t
+{
+    kRequestTypeUnknown,
+    kReadRequest,
+    kWriteRequest,
+    kInvokeRequest,
+    kSubscribeEventRequest
```
tcarmelveilleux, line 33: Why is subscribe event called out here? Subscribe privileges are the same as read privileges

tleacmcsa: In the spec:

if arl_restriction.Type == AccessRestrictionTypeEnum.EventForbidden and
action.Type == Action.TypeEnum.SubscribeRequest:
# if the event id is NULL (wildcard) or matches the cluster_element
if arl_restriction.ID == NULL or arl_restriction.ID == cluster_element:
# then no granted privileges
return set()

Maybe this needs to be clarified.  What we are really after is restricting access to an event(s), regardless of if its an attempt to read or subscribe.  These enum entries capture the requested action, which could be read or subscribe.

src/access/AccessControl.cpp
tleacmcsa: Fixed.

src/access/AccessControl.h
tleacmcsa: Removed.

src/access/RequestPath.h
bzbarsky-apple: Fundamentally, "read" vs "subscribe" is not a useful distinction for access purposes.  The distinctions we care about are:

1. What does the entity id represent?  Is it an attribute, event, or command?
2. Which access type are we performing?  Really, for events and commands there is only one access type (read and invoke respectively), while for attributes there are potentially two: read and write.

src/access/RequestPath.h
bzbarsky-apple: That's not how wildcards are supposed to work with access check failures.  The way it's _supposed_ to work is that you just exclude anything you don't have access to from the set of things you actually get reports on.  If that results in an empty set, the whole subscription is denied.

But doing whole-node wildcard subscriptions is the _normal_ way controllers operate.  If you disallow those if there is "any intersection", you would break every controller out there.

src/access/RequestPath.h
bzbarsky-apple: Fundamentally: the spec is broken here.

src/access/RequestPath.h
tleacmcsa: This should be resolved.

src/access/RequestPath.h
tleacmcsa: Got it.  Wildcard subscriptions will succeed.  The check will be on the event read and at time of event delivery.

chrisdecenzo (approved):

src/access/tests/TestAccessControl.cpp
```diff
@@ -820,260 +820,260 @@ struct CheckData
 constexpr CheckData checkData1[] = {
     // Checks for implicit PASE
     { .subjectDescriptor = { .fabricIndex = 0, .authMode = AuthMode::kPase, .subject = kPaseVerifier0 },
-      .requestPath       = { .cluster = 1, .endpoint = 2 },
+      .requestPath       = { .cluster = 1, .endpoint = 2, .requestType = RequestType::kAttributeReadRequest },
```
tcarmelveilleux, line 823: Why not leave as unknown? 

Unless ARL is hooked-up and implemented (which it isn't in this PR), all the tests should still pass with the current cases having `unknown`.

I only expect future NEW tests to test ARL, AFTER setting an ARL over top of the non-ARL test data.

tleacmcsa: The updated contract to existing ACL code is that requestType is required to be set per Boris' request/comment or it will fail the Check.  I think it makes sense to validate this argument is provided correctly along with this change.

src/app/InteractionModelEngine.cpp
```diff
@@ -530,17 +536,27 @@ CHIP_ERROR InteractionModelEngine::ParseAttributePaths(const Access::SubjectDesc
...
-                      Access::Privilege aNeededPrivilege)
+#if !CHIP_CONFIG_ENABLE_EVENTLIST_ATTRIBUTE
+static bool CanAccessEvent(const Access::SubjectDescriptor & aSubjectDescriptor, const ConcreteClusterPath & aPath,
+                           Access::Privilege aNeededPrivilege)
 {
-    Access::RequestPath requestPath{ .cluster = aPath.mClusterId, .endpoint = aPath.mEndpointId };
+    Access::RequestPath requestPath{ .cluster     = aPath.mClusterId,
+                                     .endpoint    = aPath.mEndpointId,
+                                     .requestType = Access::RequestType::kEventReadOrSubscribeRequest };
+    // leave requestPath.entityId optional value unset to indicate wildcard
```
tcarmelveilleux, line 546: Please explicitly set it to a wildcard constant value. Relying on unset in some places may actually get a garbage stack value, depending how the request path is initialized.

src/access/AccessControl.cpp
tcarmelveilleux: This is no longer fiexed.

tleacmcsa: Correct.  @tcarmelveilleux requested that this validation be done behind an ARL capability check, which is coming in the main ARL PR.

tleacmcsa: Ok, another pass at this topic: a check was put in place that will only run if ARL feature is enabled.  It is not enabled as of this PR.

src/access/AccessControl.h
```diff
@@ -627,6 +627,13 @@ class AccessControl
     // Removes a listener from the listener list, if in the list.
     void RemoveEntryListener(EntryListener & listener);
 
+    /**
+     * Check whether or not Access Restriction List is supported.
+     *
+     * @retval true if Access Restriction List is supported.
+     */
+    bool IsAccessRestrictionListSupported() const;
```
tcarmelveilleux, line 635: virtual please

tcarmelveilleux (approved):

src/access/RequestPath.h
tcarmelveilleux: The fact that it's events separate from attributes is weird. Let's fix follow-up, though.

src/app/CommandHandlerImpl.cpp
tcarmelveilleux: This is resolved now.

src/access/RequestPath.h
tleacmcsa: Resolving conversation, but tracking here for post 1.4 continuation: https://github.com/project-chip/connectedhomeip/issues/34913

src/access/RequestPath.h
tleacmcsa: Resolving conversation, but tracking here for post 1.4 continuation: https://github.com/project-chip/connectedhomeip/issues/34913

src/access/AccessControl.h
tleacmcsa: Resolving conversation, but tracking here for post 1.4 continuation: https://github.com/project-chip/connectedhomeip/issues/34913

Issue #23573, opened 2022-11-10
[BUG] OTA Provider / BDX state is erroneous 


fessehaeve:
### Reproduction steps

```bash
I believe there is a bug on the OTA provider/BDX implementation. This is how to reproduce it:
Start an OTA upgrade by announcing the provider to the requestor. 
The requestor ignores the update since the version is the same/not higher. This puts the provider in a not clean state and this is the cause of the issue.
Announce the provider again to the requestor and this time the requestor proceeds to BDX. But provider isn’t in a good state and can’t reply to ReceiveInit properly. (It doesn’t send out ReceiveAccept.
```

### Bug prevalence

Always after an unsuccessful OTA upgrade.

### GitHub hash of the SDK that was being used

0b17bce8005fb155e78e243f31111cdaf2c92797

### Platform

nrf, raspi, core

### Anything else?

Please ignore the actual value of the software version you see in the logs.
OTA Requestor log

```
16:51:18.047   5855.853        6  0 I: 5299307 [DMG]Received Command Response Data, Endpoint=0 Cluster=0x0000_0029 Command=0x0000_0001
16:51:18.056   5855.862        9  0 D: 5299316 [SWU]QueryImageResponse:
16:51:18.056   5855.862        0  0 D: 5299319 [SWU]  status: 0
16:51:18.062   5855.868        6  0 D: 5299322 [SWU]  imageURI: bdx://0000000000000064//home/ubuntu/eve_6000-matter.bin
16:51:18.068   5855.874        6  0 D: 5299331 [SWU]  softwareVersion: 6000
16:51:18.075   5855.881        7  0 D: 5299335 [SWU]  softwareVersionString: x.x.0
16:51:18.075   5855.881        0  0 D: 5299339 [SWU]  updateToken: 32
16:51:18.081   5855.887        6  0 D: 5299342 [SWU]  userConsentNeeded: 0
16:51:18.081   5855.887        0  0 D: 5299346 [SWU]  metadataForRequestor: 0
16:51:18.089   5855.895        8  0 D: 5299349 [SWU]Update available from version 6000 to 6000
16:51:18.093   5855.899        4  0 D: 5299355 [DMG]ICR moving to [AwaitingDe]
16:51:18.098   5855.904        5  0 I: 5299359 [DL]OpenThread SED interval set to 2500ms
[...]

PR #34538, opened 2024-07-26
Fix: drop poll timer to avoid crash after TransferFacilitator object destruction

j0tunn:
We should drop timer on destruction or process will crash trying to access [mTransfer](https://github.com/project-chip/connectedhomeip/blob/master/src/protocols/bdx/TransferFacilitator.cpp#L77) property of destroyed object.

### DIFF
--- a/examples/ota-provider-app/esp32/main/BdxOtaSender.cpp
+++ b/examples/ota-provider-app/esp32/main/BdxOtaSender.cpp
@@ -174,7 +174,6 @@ void BdxOtaSender::HandleTransferSessionOutput(TransferSession::OutputEvent & ev
         {
             ChipLogError(BDX, "onTransferComplete Callback not set");
         }
-        mStopPolling = true; // Stop polling the TransferSession only after receiving BlockAckEOF
         Reset();
         break;
     case TransferSession::OutputEventType::kStatusReceived:
@@ -228,7 +227,7 @@ void BdxOtaSender::Reset()
 {
     mFabricIndex.ClearValue();
     mNodeId.ClearValue();
-    mTransfer.Reset();
+    ResetTransfer();
     if (mExchangeCtx != nullptr)
     {
         mExchangeCtx->Close();
--- a/src/protocols/bdx/TransferFacilitator.cpp
+++ b/src/protocols/bdx/TransferFacilitator.cpp
@@ -32,6 +32,20 @@ namespace bdx {
 constexpr System::Clock::Timeout TransferFacilitator::kDefaultPollFreq;
 constexpr System::Clock::Timeout TransferFacilitator::kImmediatePollDelay;
 
+TransferFacilitator::~TransferFacilitator()
+{
+    ResetTransfer();
+}
+
+void TransferFacilitator::ResetTransfer()
+{
+    mTransfer.Reset();
+    ChipLogProgress(BDX, "Stop polling for messages");
+
+    VerifyOrReturn(mSystemLayer != nullptr);
+    mSystemLayer->CancelTimer(PollTimerHandler, this);
+}
+
 CHIP_ERROR TransferFacilitator::OnMessageReceived(chip::Messaging::ExchangeContext * ec, const chip::PayloadHeader & payloadHeader,
                                                   chip::System::PacketBufferHandle && payload)
 {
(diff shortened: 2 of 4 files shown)


andy31415 (approved):

bzbarsky-apple (commented):
It's really broken that when we reset the transfer that does not shut off the poll....  I guess this works around that brokenness, but the real fix would be to not have that happen.

j0tunn:
> It's really broken that when we reset the transfer that does not shut off the poll.... I guess this works around that brokenness, but the real fix would be to not have that happen.

In fact ResetTransfer does shut off the polling... But it does it in graceful manner: set the `mStopPolling` flag, wait until the next poll trigger, and then shut off the polling. It works fine for the single bdx sender object, which supposed to be instantly reused.

What I do here is more like forced reset. For the scenario where we use bdx sender object for the single session and than destroy it. For the multiple parallel ota sessions.

bzbarsky-apple:
> In fact ResetTransfer does shut off the polling..

Not in a way that's useful for "this object is going away" purposes.  The fact that it polls again after you tell it to stop everything is _so_ broken....

j0tunn:
> > In fact ResetTransfer does shut off the polling..
> 
> Not in a way that's useful for "this object is going away" purposes. The fact that it polls again after you tell it to stop everything is _so_ broken....

Take a look now. I've replaced all logic based on `mStopPolling` with direct actions.

src/protocols/bdx/TransferFacilitator.cpp
```diff
@@ -71,28 +86,34 @@ void TransferFacilitator::PollTimerHandler(chip::System::Layer * systemLayer, vo
...
+    // Do the job
+    TransferFacilitator::PollTimerHandler(systemLayer, appState);
+}
+
 void TransferFacilitator::PollForOutput()
 {
     TransferSession::OutputEvent outEvent;
     mTransfer.PollOutput(outEvent, System::SystemClock().GetMonotonicTimestamp());
     HandleTransferSessionOutput(outEvent);
+}
```
bzbarsky-apple, line 105: So I feel like I am missing something here.  PollTimerHandler used to call PollForOutput, which would re-schedule the poll timer as needed.  What re-schedules that timer in the new setup?

src/protocols/bdx/TransferFacilitator.h
```diff
@@ -79,12 +84,18 @@ class TransferFacilitator : public Messaging::ExchangeDelegate, public Messaging
      * The callback for when the poll timer expires. The poll timer regulates how often the TransferSession is polled.
      */
     static void PollTimerHandler(chip::System::Layer * systemLayer, void * appState);
+    static void ImmediatePollTimerHandler(chip::System::Layer * systemLayer, void * appState);
```
bzbarsky-apple, line 87: The difference between these two handlers needs to be clearly documented.  It's not obvious from the name.

src/protocols/bdx/TransferFacilitator.cpp
j0tunn: You are right. I've just over engineered a little bit. Fixed now.

examples/ota-provider-app/esp32/main/BdxOtaSender.cpp
```diff
@@ -174,7 +174,6 @@ void BdxOtaSender::HandleTransferSessionOutput(TransferSession::OutputEvent & ev
         {
             ChipLogError(BDX, "onTransferComplete Callback not set");
         }
-        mStopPolling = true; // Stop polling the TransferSession only after receiving BlockAckEOF
         Reset();
```
nivi-apple, line 177: so we removed the mStopPolling flag being set here but how would we notify the transfer session to stop polling. I do not see a path where we call https://github.com/project-chip/connectedhomeip/pull/34538/files#diff-98b0f488c3e61ca466c5c9c383db17e6b66f4784505b8d18fd90c9d4c172f4cbR40 which stops the timer in the transfer facilitator. The Reset() here only calls Reset on the transfer object and that doesn't cancel the timer. I see that the destructor for TransferFacilitator calls the ResetTransfer API that stops the timer. so are we assuming the transfer facilitator object will be destroyed? What am I missing here?

j0tunn: `Reset()` here is the [BdxOtaSender::Reset](https://github.com/project-chip/connectedhomeip/blob/693dd386e8681984d8856fb114738f6dba2c11bd/examples/ota-provider-app/esp32/main/BdxOtaSender.cpp#L226). And it calls `ResetTransfer` inside

nivi-apple:
@j0tunn Change looks ok. What testing have you done for this? Based on the history of the stop polling boolean last time someone removed the stopPolling this issue was seen https://github.com/project-chip/connectedhomeip/issues/23573. It would be good to test the use case mentioned in the issue description.

j0tunn:
> @j0tunn Change looks ok. What testing have you done for this? Based on the history of the stop polling boolean last time someone removed the stopPolling this issue was seen #23573. It would be good to test the use case mentioned in the issue description.

This patch is in our application about two months. And it was tested by our QA team. But honestly we are using BdxOtaSender object only once. And for each update session we create new one. So we can support parallel OTA sessions.

The bug you are talking about just would be impossible after this change. This bug occurs because of the state (`mStopPolling`). I'm removing the state here at all. So `PrepareForTransfer/InitiateTransfer` starts polling, and `ResetTransfer` stops the polling immediately. Thats it. No state, no influence from the previous sessions

nivi-apple (approved):

PR #34540, opened 2024-07-26
Add some bits to exercise global structs/enums to Unit Testing cluster.

bzbarsky-apple:
* Adds things to the Unit Testing cluster XML.
* This requires those things to be enabled in all-clusters-app, all-clusters-minimal-app, and one of the chef contact sensors to pass CI.
* That requires an implementation in test-cluster-server
* At which point might as well add a YAML test to exercise it all.

skip-protocol-compatibility, because this is only changing the Unit Testing cluster (adding fields to structs there).


tehampson:
>skip-protocol-compatibility, because this is only changing the Unit Testing cluster (adding fields to structs there).

Dumb question if you are adding and not changing any attributes in testing cluster why do you need to skip backwards compatibility checker? It should pass without skipping

bzbarsky-apple:
@tehampson The checker triggers on any change to the set of members of a struct.

cecille (approved):

andy31415 (approved):

rbultman (approved):

woody-apple (approved):

PR #34563, opened 2024-07-28
Update Darwin availability annotations.

bzbarsky-apple:
Marks a few things explicitly as provisional, and enables (as provisional) the global struct and enum we have for tests.


kiel-apple (approved):
test output looks as expected!

bzbarsky-apple:

PR #34570, opened 2024-07-29
[HVAC] Use atomic writes with Presets

### LINKED ISSUE #34546: For the Thermostat presets delegate implementation we do not support deletion of presets.
This issue tracks adding support for cases where Presets have been removed as part of the write to the Presets attribute in the example app implementation

hasty:
This converts the presets implementation in #34532 to use atomic writes.


examples/thermostat/linux/BUILD.gn
```diff
@@ -20,8 +20,6 @@ executable("thermostat-app") {
     "include/low-power/LowPowerManager.cpp",
     "include/low-power/LowPowerManager.h",
     "main.cpp",
-    "thermostat-delegate-impl.cpp",
-    "thermostat-manager.cpp",
```
bzbarsky-apple, line 24: Why are we breaking thermostat-app?

src/app/clusters/thermostat-server/thermostat-delegate.h
```diff
@@ -64,6 +64,12 @@ class Delegate
      */
     virtual CHIP_ERROR GetPresetAtIndex(size_t index, PresetStructWithOwnedMembers & preset) = 0;
 
+    /**
+     * @brief Creates a new pending presets list.
```
bzbarsky-apple, line 68: I think this needs a lot more documentation.  For example, are callers allowed to call GetPendingPresetAtIndex before calling this function?  What about AppendToPendingPresetList?

In either case, this function should be down near the other pending preset functions.

And it should make it clear that it sets pending presets to current presets, at least in the documentation if not the naming.

src/app/clusters/thermostat-server/thermostat-delegate.h
```diff
@@ -121,6 +127,19 @@ class Delegate
      *
      */
     virtual void ClearPendingPresetList() = 0;
+
+    /**
+     * @brief Informs of a change in a non-atomic thermostat attribute
```
bzbarsky-apple, line 132: Separate PR for this, as requested above.

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -116,18 +116,19 @@ void TimerExpiredCallback(System::Layer * systemLayer, void * callbackContext)
...
 /**
- * @brief Schedules a timer for the given timeout in seconds.
+ * @brief Schedules a timer for the given timeout in milliseconds.
  *
  * @param[in] endpoint The endpoint to use.
- * @param[in] timeoutSeconds The timeout in seconds.
+ * @param[in] timeoutMilliseconds The timeout in milliseconds.
  */
-void ScheduleTimer(EndpointId endpoint, uint16_t timeoutSeconds)
+void ScheduleTimer(EndpointId endpoint, uint16_t timeoutMilliseconds)
```
bzbarsky-apple, line 129: How about we just have this take a System::Clock::Milliseconds16 argument to start with?

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -769,6 +758,22 @@ CHIP_ERROR ThermostatAttrAccess::Read(const ConcreteReadAttributePath & aPath, A
         Delegate * delegate = GetDelegate(aPath.mEndpointId);
         VerifyOrReturnError(delegate != nullptr, CHIP_ERROR_INCORRECT_STATE, ChipLogError(Zcl, "Delegate is null"));
 
+        if (InAtomicWrite(aPath.mEndpointId))
```
bzbarsky-apple, line 761: Doesn't this need to also check that `GetAtomicWriteScopedNodeId(aPath.mEndpointId)` matched the reader's scoped node id?    Something like:
```suggestion
        auto & desc = aEncoder.GetSubjectDescriptor();
        if (InAtomicWrite(aPath.mEndpointId) && desc.authMode == Access::AuthMode::kCase && GetAtomicWriteScopedNodeId(aPath.mEndpointId) == ScopedNodeId(desc.subject, desc.fabricIndex))
```

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -827,48 +828,31 @@ CHIP_ERROR ThermostatAttrAccess::Write(const ConcreteDataAttributePath & aPath,
 {
     VerifyOrDie(aPath.mClusterId == Thermostat::Id);
 
-    uint32_t ourFeatureMap;
-    bool localTemperatureNotExposedSupported = (FeatureMap::Get(aPath.mEndpointId, &ourFeatureMap) == imcode::Success) &&
-        ((ourFeatureMap & to_underlying(Feature::kLocalTemperatureNotExposed)) != 0);
+    EndpointId endpoint                               = aPath.mEndpointId;
+    const Access::SubjectDescriptor subjectDescriptor = aDecoder.GetSubjectDescriptor();
```
bzbarsky-apple, line 832:
```suggestion
    auto & subjectDescriptor = aDecoder.GetSubjectDescriptor();
```

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1065,7 +1079,7 @@ MatterThermostatClusterServerPreAttributeChangedCallback(const app::ConcreteAttr
             if (requested > OccupiedCoolingSetpoint - DeadBandTemp)
                 return imcode::InvalidValue;
         }
-        return imcode::Success;
+        break;
```
bzbarsky-apple, line 1082: All the changed to this function seem unrelated to presets and should be done in a separate PR, if at all.

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1287,102 +1306,122 @@ bool emberAfThermostatClusterSetActivePresetRequestCallback(
     return true;
 }
 
-bool emberAfThermostatClusterStartPresetsSchedulesEditRequestCallback(
-    chip::app::CommandHandler * commandObj, const chip::app::ConcreteCommandPath & commandPath,
-    const chip::app::Clusters::Thermostat::Commands::StartPresetsSchedulesEditRequest::DecodableType & commandData)
+bool validAtomicAttributes(const chip::app::Clusters::Thermostat::Commands::AtomicRequest::DecodableType & commandData)
```
bzbarsky-apple, line 1309:
```suggestion
bool validAtomicAttributes(Commands::AtomicRequest::DecodableType & commandData)
```

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1287,102 +1306,122 @@ bool emberAfThermostatClusterSetActivePresetRequestCallback(
...
-    ScopedNodeId sourceNodeId = GetSourceScopedNodeId(commandObj);
-
-    EndpointId endpoint = commandPath.mEndpointId;
-
-    // If the presets are editable and the scoped node id of the client sending StartPresetsSchedulesEditRequest command
-    // is not the same as the one that previously originated a StartPresetsSchedulesEditRequest command, return BUSY.
-    if (gThermostatAttrAccess.GetPresetsEditable(endpoint) &&
-        (gThermostatAttrAccess.GetOriginatorScopedNodeId(endpoint) != sourceNodeId))
+    auto attributeIdsIter = commandData.attributeRequests.begin();
+    bool requestedPresets, requestedSchedules;
```
bzbarsky-apple, line 1312: Don't those need to be initialized to false for the logic below to work right?  Right now we are going to get random behavior.

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1287,102 +1306,122 @@ bool emberAfThermostatClusterSetActivePresetRequestCallback(
...
-    commandObj->AddStatus(commandPath, imcode::Success);
     return true;
 }
 
-bool emberAfThermostatClusterCancelPresetsSchedulesEditRequestCallback(
-    chip::app::CommandHandler * commandObj, const chip::app::ConcreteCommandPath & commandPath,
-    const chip::app::Clusters::Thermostat::Commands::CancelPresetsSchedulesEditRequest::DecodableType & commandData)
+bool handleAtomicBegin(chip::app::CommandHandler * commandObj, const ScopedNodeId & sourceNodeId,
+                       const chip::app::ConcreteCommandPath & commandPath,
+                       const chip::app::Clusters::Thermostat::Commands::AtomicRequest::DecodableType & commandData)
```
bzbarsky-apple, line 1342:
```suggestion
bool handleAtomicBegin(CommandHandler * commandObj, const ScopedNodeId & sourceNodeId,
                       const ConcreteCommandPath & commandPath,
                       const Commands::AtomicRequest::DecodableType & commandData)
```

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1287,102 +1306,122 @@ bool emberAfThermostatClusterSetActivePresetRequestCallback(
...
-    gThermostatAttrAccess.SetOriginatorScopedNodeId(endpoint, sourceNodeId);
-    ScheduleTimer(endpoint, commandData.timeoutSeconds);
-    commandObj->AddStatus(commandPath, imcode::Success);
     return true;
 }
 
-bool emberAfThermostatClusterCancelPresetsSchedulesEditRequestCallback(
-    chip::app::CommandHandler * commandObj, const chip::app::ConcreteCommandPath & commandPath,
-    const chip::app::Clusters::Thermostat::Commands::CancelPresetsSchedulesEditRequest::DecodableType & commandData)
+bool handleAtomicBegin(chip::app::CommandHandler * commandObj, const ScopedNodeId & sourceNodeId,
```
bzbarsky-apple, line 1340: This function always returns true.  Why does it need a return bool at all?

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1287,102 +1306,122 @@ bool emberAfThermostatClusterSetActivePresetRequestCallback(
...
 
-    Delegate * delegate = GetDelegate(endpoint);
-
-    if (delegate == nullptr)
+    bool inAtomicWrite = gThermostatAttrAccess.InAtomicWrite(endpoint);
+    if (inAtomicWrite)
     {
-        ChipLogError(Zcl, "Delegate is null");
-        return SendResponseAndCleanUp(delegate, endpoint, commandObj, commandPath, imcode::InvalidInState);
+        chip::app::Clusters::Thermostat::Commands::AtomicResponse::Type response;
```
bzbarsky-apple, line 1363:
```suggestion
        Commands::AtomicResponse::Type response;
```

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1287,102 +1306,122 @@ bool emberAfThermostatClusterSetActivePresetRequestCallback(
...
+        Globals::Structs::AtomicAttributeStatusStruct::Type attributeStatus[] = {
+            { .attributeID = Presets::Id, .statusCode = static_cast<uint8_t>(imcode::Busy) },
+            { .attributeID = Schedules::Id, .statusCode = static_cast<uint8_t>(imcode::Busy) }
```
bzbarsky-apple, line 1366:
```suggestion
            { .attributeID = Presets::Id, .statusCode = to_underlying(imcode::Busy) },
            { .attributeID = Schedules::Id, .statusCode = to_underlying(imcode::Busy) }
```

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1287,102 +1306,122 @@ bool emberAfThermostatClusterSetActivePresetRequestCallback(
...
+        };
+
+        response.statusCode      = static_cast<uint8_t>(imcode::Failure);
```
bzbarsky-apple, line 1369:
```suggestion
        response.statusCode      = to_underlying(imcode::Failure);
```

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1287,102 +1306,122 @@ bool emberAfThermostatClusterSetActivePresetRequestCallback(
...
+        response.attributeStatus = attributeStatus;
+        commandObj->AddResponse(commandPath, response);
+        // commandObj->AddStatus(commandPath, imcode::Busy);
```
bzbarsky-apple, line 1372:
```suggestion
```

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1287,102 +1306,122 @@ bool emberAfThermostatClusterSetActivePresetRequestCallback(
...
+        return true;
     }
 
-    // Clear the timer, discard the changes and set PresetsEditable to false.
-    return SendResponseAndCleanUp(delegate, endpoint, commandObj, commandPath, imcode::Success);
+    uint16_t maxTimeout = 5000;
```
bzbarsky-apple, line 1376: Followup: this should be configurable by the application.

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1287,102 +1306,122 @@ bool emberAfThermostatClusterSetActivePresetRequestCallback(
...
+    timeout             = std::min(timeout, maxTimeout);
+
+    ScheduleTimer(endpoint, timeout);
+    gThermostatAttrAccess.SetAtomicWrite(endpoint, true);
+    gThermostatAttrAccess.SetAtomicWriteScopedNodeId(endpoint, GetSourceScopedNodeId(commandObj));
+    chip::app::Clusters::Thermostat::Commands::AtomicResponse::Type response;
```
bzbarsky-apple, line 1382:
```suggestion
    Commands::AtomicResponse::Type response;
```

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1287,102 +1306,122 @@ bool emberAfThermostatClusterSetActivePresetRequestCallback(
...
+    Globals::Structs::AtomicAttributeStatusStruct::Type attributeStatus[] = {
+        { .attributeID = Presets::Id, .statusCode = static_cast<uint8_t>(imcode::Success) },
+        { .attributeID = Schedules::Id, .statusCode = static_cast<uint8_t>(imcode::Success) }
```
bzbarsky-apple, line 1385:
```suggestion
        { .attributeID = Presets::Id, .statusCode = to_underlying(imcode::Success) },
        { .attributeID = Schedules::Id, .statusCode = to_underlying(imcode::Success) }
```

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1287,102 +1306,122 @@ bool emberAfThermostatClusterSetActivePresetRequestCallback(
...
+    };
+
+    response.statusCode      = static_cast<uint8_t>(imcode::Failure);
```
bzbarsky-apple, line 1388:
```suggestion
    response.statusCode      = to_underlying(imcode::Failure);
```

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1287,102 +1306,122 @@ bool emberAfThermostatClusterSetActivePresetRequestCallback(
...
-        commandObj->AddStatus(commandPath, imcode::UnsupportedAccess);
+        commandObj->AddStatus(commandPath, imcode::InvalidCommand);
         return true;
     }
 
-    Delegate * delegate = GetDelegate(endpoint);
-
-    if (delegate == nullptr)
+    bool inAtomicWrite = gThermostatAttrAccess.InAtomicWrite(endpoint);
+    if (inAtomicWrite)
```
bzbarsky-apple, line 1361: Per spec, if `sourceNodeId` is the thing doing the atomic write we should be returning INVALID_IN_STATE, no?

And that check comes before the "does not have a timeout" check.

How it's ordered with the validAtomicAttributes() check is not very clear, though the way the spec is laid out people would assume it comes after....

In general, since these checks return different status, the spec needs to define the order. Otherwise you might get disagreement between implementations on what happens if list of ids is invalid and the client is already associated with an atomic write state.

src/app/clusters/thermostat-server/thermostat-server.cpp
bzbarsky-apple, line 1388: So this is always failing the Begin Write?  Why?

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1287,102 +1306,122 @@ bool emberAfThermostatClusterSetActivePresetRequestCallback(
...
+    chip::app::Clusters::Thermostat::Commands::AtomicResponse::Type response;
+    Globals::Structs::AtomicAttributeStatusStruct::Type attributeStatus[] = {
+        { .attributeID = Presets::Id, .statusCode = static_cast<uint8_t>(imcode::Success) },
+        { .attributeID = Schedules::Id, .statusCode = static_cast<uint8_t>(imcode::Success) }
+    };
+
+    response.statusCode      = static_cast<uint8_t>(imcode::Failure);
+    response.attributeStatus = attributeStatus;
+    response.timeout.Emplace(timeout);
+    commandObj->AddResponse(commandPath, response);
```
bzbarsky-apple, line 1391: Things that seem to be missing here, compared to the spec: privilege checks.  Presumably that's because the fact that we managed to invoke the command at all means we have sufficient privileges to write to the attributes, but we should have a clear comment saying that, so that when someone copies/moves this code for use in another context we notice that the assumption might not hold anymore.

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1287,102 +1306,122 @@ bool emberAfThermostatClusterSetActivePresetRequestCallback(
...
+
+    return true;
 }
 
-bool emberAfThermostatClusterCommitPresetsSchedulesRequestCallback(
-    chip::app::CommandHandler * commandObj, const chip::app::ConcreteCommandPath & commandPath,
-    const chip::app::Clusters::Thermostat::Commands::CommitPresetsSchedulesRequest::DecodableType & commandData)
+bool handleAtomicCommit(chip::app::CommandHandler * commandObj, const ScopedNodeId & sourceNodeId,
+                        const chip::app::ConcreteCommandPath & commandPath,
+                        const chip::app::Clusters::Thermostat::Commands::AtomicRequest::DecodableType & commandData)
```
bzbarsky-apple, line 1398:
```suggestion
bool handleAtomicCommit(CommandHandler * commandObj, const ScopedNodeId & sourceNodeId,
                        const ConcreteCommandPath & commandPath,
                        const Commands::AtomicRequest::DecodableType & commandData)
```

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1510,7 +1548,7 @@ bool emberAfThermostatClusterCommitPresetsSchedulesRequestCallback(
             return SendResponseAndCleanUp(delegate, endpoint, commandObj, commandPath, imcode::ConstraintError);
         }
 
-        // If the preset type for the preset scenario does not support names and a name is specified, return CONSTRAINT_ERROR.
+        // If the preset type for the preset scenario does not support name and a name is specified, return CONSTRAINT_ERROR.
```
bzbarsky-apple, line 1551:
```suggestion
        // If the preset type for the preset scenario does not support names and a name is specified, return CONSTRAINT_ERROR.
```

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1439,7 +1478,6 @@ bool emberAfThermostatClusterCommitPresetsSchedulesRequestCallback(
             return SendResponseAndCleanUp(delegate, endpoint, commandObj, commandPath, imcode::InvalidInState);
```
bzbarsky-apple, line 1478: I don't think these various SendResponseAndCleanUp calls do the right thing.  Per spec, failure to commit for the reasons being checked for here is supposed to return an AtomicResponse with a list of attribute statuses, etc.

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1561,6 +1599,64 @@ bool emberAfThermostatClusterCommitPresetsSchedulesRequestCallback(
     return SendResponseAndCleanUp(delegate, endpoint, commandObj, commandPath, imcode::Success);
 }
 
+bool handleAtomicRollback(chip::app::CommandHandler * commandObj, const ScopedNodeId & sourceNodeId,
+                          const chip::app::ConcreteCommandPath & commandPath,
+                          const chip::app::Clusters::Thermostat::Commands::AtomicRequest::DecodableType & commandData)
```
bzbarsky-apple, line 1604:
```suggestion
bool handleAtomicRollback(CommandHandler * commandObj, const ScopedNodeId & sourceNodeId,
                          const ConcreteCommandPath & commandPath,
                          const Commands::AtomicRequest::DecodableType & commandData)
```

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1287,102 +1306,122 @@ bool emberAfThermostatClusterSetActivePresetRequestCallback(
...
+        }
     }
-
-    // Set presets editable to true and the scoped originator node id to the source scoped node id, and start a timer with the
-    // timeout in seconds passed in the command args. Return success.
-    gThermostatAttrAccess.SetPresetsEditable(endpoint, true);
-    gThermostatAttrAccess.SetOriginatorScopedNodeId(endpoint, sourceNodeId);
-    ScheduleTimer(endpoint, commandData.timeoutSeconds);
-    commandObj->AddStatus(commandPath, imcode::Success);
     return true;
```
bzbarsky-apple, line 1337: Needs to check `attributeIdsIter.GetStatus()`, no?

Also spec says to INVALID_COMMAND if the list is empty, but I don't see that here.

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1561,6 +1599,64 @@ bool emberAfThermostatClusterCommitPresetsSchedulesRequestCallback(
...
+    }
+    Delegate * delegate = GetDelegate(endpoint);
+
+    if (delegate == nullptr)
+    {
+        ChipLogError(Zcl, "Delegate is null");
+        commandObj->AddStatus(commandPath, imcode::InvalidInState);
+        return true;
+    }
+    return SendResponseAndCleanUp(delegate, endpoint, commandObj, commandPath, imcode::Success);
```
bzbarsky-apple, line 1632: Where is this doing the actual commit?  That needs to happen!

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1561,6 +1599,64 @@ bool emberAfThermostatClusterCommitPresetsSchedulesRequestCallback(
...
+}
+
+bool emberAfThermostatClusterAtomicRequestCallback(
+    chip::app::CommandHandler * commandObj, const chip::app::ConcreteCommandPath & commandPath,
+    const chip::app::Clusters::Thermostat::Commands::AtomicRequest::DecodableType & commandData)
+{
+    auto & requestType = commandData.requestType;
+    // auto & timeout     = commandData.timeout;
```
bzbarsky-apple, line 1640:
```suggestion
```

src/app/zap-templates/zcl/data-model/chip/thermostat-cluster.xml
```diff
@@ -424,101 +419,97 @@ limitations under the License.
...
-      <description>This command is used to clear the weekly schedule. The ClearWeeklySchedule command has no payload.</description>
+      <description>This command is used to clear the weekly schedule.</description>
       <access op="invoke" privilege="manage"/>
-    </command>  
+    </command>
+
     <command source="client" code="0x05" name="SetActiveScheduleRequest" optional="true">
-      <description>This command is used to set the active schedule.</description>
-      <arg name="ScheduleHandle" type="octet_string" length="16"/>
+      <description>Upon receipt, if the </description>
```
bzbarsky-apple, line 481: Please make sure to fix this.  Followup is OK.  But these descriptions are in fact used, and this new description is nonsense.

src/app/zap-templates/zcl/data-model/chip/thermostat-cluster.xml
```diff
@@ -424,101 +419,97 @@ limitations under the License.
...
-      <arg name="TimeoutSeconds" type="int16u" max="120"/>
-    </command>
-    <command source="client" code="0x08" name="CancelPresetsSchedulesEditRequest" optional="true">
-      <description>This command is used to cancel editing presets and schedules.</description>
-      <access op="invoke" privilege="manage"/>
-    </command>
-    <command source="client" code="0x09" name="CommitPresetsSchedulesRequest" optional="true">
-      <description>This command is used to notify the server that all edits are done and should be committed.</description>
-      <access op="invoke" privilege="manage"/>
+      <description>ID</description>
```
bzbarsky-apple, line 486: Again, description needs to be fixed.  Please check the rest of these.

src/app/zap-templates/zcl/data-model/chip/thermostat-cluster.xml
```diff
@@ -424,101 +419,97 @@ limitations under the License.
...
+      <arg id="0" name="PresetHandle" type="octet_string" length="16"/>
     </command>
 
     <!-- Server Commands/Responses -->
+    <command source="server" code="0x00" name="GetWeeklyScheduleResponse" optional="true" disableDefaultResponse="true">
+      <description>This command has the same payload format as the Set Weekly Schedule.</description>
```
bzbarsky-apple, line 492: This is a worse description than what used to be here.

examples/all-clusters-app/all-clusters-common/src/thermostat-delegate-impl.cpp
```diff
@@ -169,8 +186,19 @@ CHIP_ERROR ThermostatDelegate::AppendToPendingPresetList(const PresetStruct::Typ
...
+        {
+            // TODO: #34556 Since we support only one preset of each type, using the octet string containing the preset scenario
+            // suffices as the unique preset handle. Need to fix this to actually provide unique handles once multiple presets of
+            // each type are supported.
+            const uint8_t handle[] = { static_cast<uint8_t>(preset.presetScenario) };
+            mPendingPresets[mNextFreeIndexInPendingPresetsList].SetPresetHandle(DataModel::MakeNullable(ByteSpan(handle)));
+        }
+        else
+        {
+            mPendingPresets[mNextFreeIndexInPendingPresetsList].SetPresetHandle(preset.presetHandle);
```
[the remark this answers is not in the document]
hasty: Updated.

examples/all-clusters-app/all-clusters-common/src/thermostat-delegate-impl.cpp
```diff
@@ -134,6 +135,22 @@ CHIP_ERROR ThermostatDelegate::GetPresetAtIndex(size_t index, PresetStructWithOw
...
+    mNextFreeIndexInPendingPresetsList = 0;
+    for (uint8_t indexInPresets = 0; indexInPresets < mNextFreeIndexInPresetsList; indexInPresets++)
+    {
+        auto preset = mPresets[indexInPresets];
+        mPendingPresets[mNextFreeIndexInPendingPresetsList].SetPresetHandle(preset.GetPresetHandle());
+        mPendingPresets[mNextFreeIndexInPendingPresetsList].SetPresetScenario(preset.GetPresetScenario());
+        mPendingPresets[mNextFreeIndexInPendingPresetsList].SetName(preset.GetName());
+        mPendingPresets[mNextFreeIndexInPendingPresetsList].SetCoolingSetpoint(preset.GetCoolingSetpoint());
+        mPendingPresets[mNextFreeIndexInPendingPresetsList].SetHeatingSetpoint(preset.GetHeatingSetpoint());
+        mPendingPresets[mNextFreeIndexInPendingPresetsList].SetBuiltIn(preset.GetBuiltIn());
```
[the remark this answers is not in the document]
hasty: Renamed method, took this suggestion.

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1287,102 +1322,126 @@ bool emberAfThermostatClusterSetActivePresetRequestCallback(
...
-    commandObj->AddStatus(commandPath, imcode::Success);
-    return true;
+    // If the atomic request doesn't contain at least one of these attributes, it's invalid
+    return (requestedPresets || requestedSchedules);
 }
 
-bool emberAfThermostatClusterCancelPresetsSchedulesEditRequestCallback(
-    chip::app::CommandHandler * commandObj, const chip::app::ConcreteCommandPath & commandPath,
-    const chip::app::Clusters::Thermostat::Commands::CancelPresetsSchedulesEditRequest::DecodableType & commandData)
+bool handleAtomicBegin(chip::app::CommandHandler * commandObj, const chip::app::ConcreteCommandPath & commandPath,
```
bzbarsky-apple, line 1361:
```suggestion
bool handleAtomicBegin(CommandHandler * commandObj, const ConcreteCommandPath & commandPath,
```

bzbarsky-apple, line 1361: Again, this always returns true.  Why does it have a bool return type?

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1287,102 +1322,126 @@ bool emberAfThermostatClusterSetActivePresetRequestCallback(
...
-        commandObj->AddStatus(commandPath, imcode::UnsupportedAccess);
+        commandObj->AddStatus(commandPath, imcode::InvalidCommand);
         return true;
     }
 
-    Delegate * delegate = GetDelegate(endpoint);
+    auto timeout = commandData.timeout.Value();
 
-    if (delegate == nullptr)
+    if (!validAtomicAttributes(commandData))
```
bzbarsky-apple, line 1381: So per spec, just the way things are written, I would expect this check to come before the checks for an atomic write state and so on....  Followup to sort out the order here.

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1287,102 +1322,126 @@ bool emberAfThermostatClusterSetActivePresetRequestCallback(
...
+        commandObj->AddResponse(commandPath, response);
+        return true;
+    }
+
+    // This is a valid request to open an atomic write. Note that
+
+    uint16_t maxTimeout = 5000;
+    timeout             = std::min(timeout, maxTimeout);
+
+    ScheduleTimer(endpoint, static_cast<System::Clock::Milliseconds16>(timeout));
```
bzbarsky-apple, line 1406:
```suggestion
    ScheduleTimer(endpoint, System::Clock::Milliseconds16(timeout));
```

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1561,6 +1619,58 @@ bool emberAfThermostatClusterCommitPresetsSchedulesRequestCallback(
...
+
+    Delegate * delegate = GetDelegate(endpoint);
+
+    if (delegate == nullptr)
+    {
+        ChipLogError(Zcl, "Delegate is null");
+        commandObj->AddStatus(commandPath, imcode::InvalidInState);
+        return true;
+    }
+    return SendResponseAndCleanUp(delegate, endpoint, commandObj, commandPath, imcode::Success);
```
bzbarsky-apple, line 1646: Again, all these SendResponseAndCleanUp calls are returning the wrong sort of response, unless I am really missing something.

src/python_testing/TC_TSTAT_4_2.py
```diff
@@ -61,32 +61,36 @@ async def write_presets(self, endpoint, presets) -> Status:
         result = await self.default_controller.WriteAttribute(self.dut_node_id, [(endpoint, cluster.Attributes.Presets(presets))])
         return result[0].Status
 
-    async def send_edit_preset_request_command(self,
+    async def send_edit_atomic_request_begin_command(self,
                                                endpoint: int = None,
                                                expected_status: Status = Status.Success):
         try:
-            await self.send_single_cmd(cmd=cluster.Commands.StartPresetsSchedulesEditRequest(timeoutSeconds=180),
+            await self.send_single_cmd(cmd=cluster.Commands.AtomicRequest(requestType=0,
```
bzbarsky-apple, line 68: Followup (no need to do it in this PR): Can we use the right named enum value for that requestType?  And similar elsewhere.

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -192,13 +192,16 @@ ScopedNodeId GetSourceScopedNodeId(CommandHandler * commandObj)
 }
 
 /**
- * @brief Utility to clean up state by clearing the pending presets list, canceling the timer
- *        and setting PresetsEditable to false and clear the originator scoped node id.
+ * @brief Sends a response for the command and cleans up state by calling CleanUp()
```
bzbarsky-apple, line 195:
```suggestion
 * @brief Discards pending atomic writes and atomic state.
```

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -192,13 +192,16 @@ ScopedNodeId GetSourceScopedNodeId(CommandHandler * commandObj)
...
  *
  * @param[in] delegate The delegate to use.
  * @param[in] endpoint The endpoint to use.
+ * @param[in] commandObj The command handler to use to add the status response.
+ * @param[in] commandPath The command path.
+ * @param[in] status The status code to send as the response.
```
bzbarsky-apple, line 201:
```suggestion
```

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -883,6 +860,7 @@ CHIP_ERROR ThermostatAttrAccess::Write(const ConcreteDataAttributePath & aPath,
         // and add to the pending presets list.
         if (!aPath.IsListOperation() || aPath.mListOp == ConcreteDataAttributePath::ListOperation::ReplaceAll)
         {
+            ChipLogError(Zcl, "Replace all!");
```
bzbarsky-apple, line 863:
```suggestion
            ChipLogProgress(Zcl, "Thermostat presets: Replace all!");
```

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -909,6 +887,7 @@ CHIP_ERROR ThermostatAttrAccess::Write(const ConcreteDataAttributePath & aPath,
         // If the list operation is AppendItem, call the delegate to append the item to the list of pending presets.
         if (aPath.mListOp == ConcreteDataAttributePath::ListOperation::AppendItem)
         {
+            ChipLogError(Zcl, "Appending again!");
```
bzbarsky-apple, line 890:
```suggestion
            ChipLogProgress(Zcl, "Thermostat presets: Appending again!");
```

bzbarsky-apple (approved):

src/app/clusters/thermostat-server/thermostat-server.cpp
bzbarsky-apple: Tracked in https://github.com/project-chip/connectedhomeip/issues/34740

src/python_testing/TC_TSTAT_4_2.py
bzbarsky-apple: Filed https://github.com/project-chip/connectedhomeip/issues/34742

fessehaeve (approved):

PR #34582, opened 2024-07-29
Update to mobly-1.12.3 to avoid pip build errors

ksperling-apple:
Mobly < 1.12.3 was using `setuptools.command.test` in setup.py which breaks installs of the most recent setuptools release.

See https://github.com/pypa/setuptools/issues/4519


bzbarsky-apple (approved):

ksperling-apple:
This doesn't seem to fix the issue by itself, there are more complicated things going on during these python module builds.

PR #34593, opened 2024-07-29
[un-revert] Occupancy sensing updates for Matter 1.4

tcarmelveilleux:
PR #34293 was merged by mistake with ESP32 break. This PR brings back the code to fix and land.

- Updated occupancy sensing cluster SDK implementation based on XML changes(PR#34163) for Occupancy Sensing rev 5 

- Updated occupancy sensing cluster sample app(all-clusters-app) implementation to Rev 5 in the zap and matter file of all-clusters-app

- Remove "occupancy-sensor-server" from "ClientDirectories" in zap_cluster_list.json to resolve a building issue found from rebuilding the chip-tool


bzbarsky-apple (approved):

bzbarsky-apple:
Fast-tracking, since assuming CI passes this is just a clone of https://github.com/project-chip/connectedhomeip/pull/34293 (which had approvals), with a CI fix.

cecille (approved):

fessehaeve (approved):

examples/all-clusters-app/all-clusters-common/src/occupancy-sensing-stub.cpp
```diff
@@ -0,0 +1,57 @@
...
+using namespace chip::app::Clusters;
+using namespace chip::app::Clusters::OccupancySensing;
+using namespace chip::app::Clusters::OccupancySensing::Structs;
+using namespace chip::DeviceLayer;
+
+using chip::Protocols::InteractionModel::Status;
+
+static std::unique_ptr<OccupancySensingAttrAccess>
+    gAttrAccess[MATTER_DM_OCCUPANCY_SENSING_CLUSTER_SERVER_ENDPOINT_COUNT + CHIP_DEVICE_CONFIG_DYNAMIC_ENDPOINT_COUNT];
+
```
jadhavrohit924, line 33: @OliverFan1  This should be mapped to the EndpointId and not the array index. If I have OccupancySensingCluster in endpoints 1, 2, and 5, the code will crash on endpoint 5.

examples/all-clusters-app/all-clusters-common/src/occupancy-sensing-stub.cpp
```diff
@@ -0,0 +1,57 @@
...
+void emberAfOccupancySensingClusterInitCallback(EndpointId endpointId)
+{
+    VerifyOrDie(!gAttrAccess[endpointId]);
```
jadhavrohit924, line 36: It's crashing for ESP over here because the array is mapped by the index and not by endpointId.

examples/all-clusters-app/all-clusters-common/src/occupancy-sensing-stub.cpp
OliverFan1: This will be fixed in the follow-up PR #34658

PR #34597, opened 2024-07-29
Add a SDK part of `IM Cluster Status Success` to CHIP_ERROR

andy31415:
CHIP_ERROR currently covers global IM statuses as well as cluster-specific failures, however we did not cover cluster specific successes.

This PR adds a new part to CHIP_ERROR that encodes cluster statuses and updates StatusIB to be able to convert from them. Note that this still considers such `success` statuses as a CHIP_ERROR since the overhead of converting all tests to not check CHIP_NO_ERROR and have code inside IsSuccess would be quite expensive.

#### Background of this change:

Working on DataModel interface split, where Writes may return a CHIP_ERROR currently however that fails to encode ClusterSuccess.

I also considered making the DM provider interfaces return `ClusterStatus` however that requires more work since ReadAttribute relies on non-status errors like CHIP_NO_MEMORY to control code paths. As such, having a better return (e.g. a `std::variant<ClusterStatus, ProcessingError>` where ProcessingError would be out of space) requires more thought. Also a lot of SDK code returns CHIP_ERROR, so a CHIP_ERROR based interface seems slightly more convenient to use at this point).

### DIFF
--- a/src/lib/core/CHIPError.h
+++ b/src/lib/core/CHIPError.h
@@ -455,6 +455,12 @@ using CHIP_ERROR = ::chip::ChipError;
 //
 // value must be a compile-time constant as mandated by CHIP_SDK_ERROR.
 //
+// NOTE: CLUSTER_STATUS are generally treated as ERRORS for most logging since
+//       they do not equal CHIP_NO_ERROR and return false on IsSuccess.
+//
+//       The existence of IM_CLUSTER_STATUS_SUCCESS is for complete conversion
+//       to a StatusIB.
+//
 // value MUST be uint8_t sized
 //
 #define CHIP_IM_CLUSTER_STATUS_FAILURE(value) CHIP_SDK_ERROR(::chip::ChipError::SdkPart::kIMClusterStatusFailure, value)
--- a/src/lib/core/CHIPError.h
+++ b/src/lib/core/CHIPError.h
@@ -158,7 +158,8 @@ public:
     {}
 #else
     constexpr ChipError(SdkPart part, uint8_t code, const char * file, unsigned int line) :
-        mError(MakeInteger(part, code)) CHIP_INITIALIZE_ERROR_SOURCE(file, line, /*loc=*/nullptr){}
+        mError(MakeInteger(part, code)) CHIP_INITIALIZE_ERROR_SOURCE(file, line, /*loc=*/nullptr)
+    {}
 #endif // CHIP_CONFIG_ERROR_SOURCE && __cplusplus >= 202002L
 
     /**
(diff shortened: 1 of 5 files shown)


src/lib/core/CHIPError.h
```diff
@@ -102,14 +102,15 @@ class ChipError
...
-        kApplication     = 7, ///< Application-defined errors; see CHIP_APPLICATION_ERROR
+        kCore                   = 0, ///< SDK core errors.
+        kInet                   = 1, ///< Inet layer errors; see <inet/InetError.h>.
+        kDevice                 = 2, ///< Device layer errors; see <platform/CHIPDeviceError.h>.
+        kASN1                   = 3, ///< ASN1 errors; see <asn1/ASN1Error.h>.
+        kBLE                    = 4, ///< BLE layer errors; see <ble/BleError.h>.
+        kIMGlobalStatus         = 5, ///< Interaction Model global status code.
+        kIMClusterStatusFailure = 6, ///< Interaction Model cluster-specific status code (failure)
+        kApplication            = 7, ///< Application-defined errors; see CHIP_APPLICATION_ERROR
+        kIMClusterStatusSuccess = 8, ///< Interaction Model cluster-specific status code (success)
```
bzbarsky-apple, line 113: We sort of purposefully did not do this, because encoding a "success" status as a CHIP_ERROR is really weird and brittle.  For cases when we might have a cluster-specific success we _really_ need to be using a different return type than CHIP_ERROR, or have an out-param for the status or _something_.

andy31415: The difficulty was that conceptually Read/Write/Invoke are supposed to be StatusIB (so ClusterStatus) however for chunking we are also expecting a CHIP_ERROR to encode the out of space errors.

I am concerned that the ripple effect of that. The existing code logic is not very clean on single-return because `AddStatus` is separated from `error code return` and that allowed our types to diverge.

bzbarsky-apple: One thing we could do:

1. Add an `mOutOfSpace` member to `AttributeValueEncoder`.
2. Set this member as needed in the relevant public `Encode` methods (we can do this without too much codesize hit, maybe).
3. Have whatever created the AttributeValueEncoder check that state (like it already checks the `mTriedEncode` state) and handle it appropriately.
4. Change `AttributeAccessInterface::Read` to return `ClusterStatus`
5. Change `DataModel::Provider::ReadAttribute` to return `ClusterStatus`

That does mean not being able to use ReturnErrorOnFailure() all over in `Read`, though....

Another option is to just hang the ClusterStatus off the AttributeValueEncoder and the few consumers that want a success status can do that, and we check for it in the one chokepoint at the callsite.

andy31415:
closing for #34708

PR #34616, opened 2024-07-30
Add support for global structs/enums/bitmaps to YAML test harness.

bzbarsky-apple:


woody-apple (approved):

scripts/py_matter_yamltests/matter_yamltests/definitions.py
```diff
@@ -257,9 +261,26 @@ def __get_targets_by_cluster_name(self, cluster_name: str, target_type: _ItemTyp
...
         cluster_name = cluster_name.replace(' ', '')
-        return target_mapping[target_type].get(cluster_name)
+        global_target = global_target_mapping[target_type]
+        target = target_mapping[target_type].get(cluster_name)
+
+        if target is None:
+            return global_target
+        if global_target is None:
+            return target
+        return target | global_target
```
cecille, line 283: when would this happen?

cecille (approved):

scripts/py_matter_yamltests/matter_yamltests/definitions.py
bzbarsky-apple: Which "this"?  That a cluster has both global and non-global elements of some type?  I'd say "usually".

cecille: OH, I was thinking this was per-element, ie an element that was both global and cluster, but yeah, this makes much more sense.

PR #34617, opened 2024-07-30
Dump details about leaked ExchangeContexts before aborting

ksperling-apple:
This is implemented via a VerifyOrDieWithObject() variant of the existing VerifyOrDie() macro that calls a DumpToLog() method on the provided object if it exists (otherwise this is simply a no-op).

If CHIP_CONFIG_VERBOSE_VERIFY_OR_DIE is not enabled, VerifyOrDieWithObject() simply behaves like a plain VerifyOrDie(). DumpToLog() implementations can use ChipLogFormatRtti to log type information about an object (usually a delegate); if RTTI is disabled this simply outputs whether the object was null or not.

When we abort due to a non-empty pool we'll get logging like this (where enabled by the platform)
```
[1722316648.933] [22561:45699101] [SPT] HeapObjectPool: 1 allocated
[1722316648.933] [22561:45699101] [EM] ExchangeContext: 1234i delegate=N12_GLOBAL__N_120MockExchangeDelegateE
[1722316648.933] [22561:45699101] [SPT] VerifyOrDie failure at src/lib/support/Pool.h:405: Allocated() == 0
```

### DIFF
--- a/src/lib/support/Pool.h
+++ b/src/lib/support/Pool.h
@@ -24,6 +24,7 @@
 
 #include <lib/support/CHIPMem.h>
 #include <lib/support/CodeUtils.h>
+#include <lib/support/ObjectDump.h>
 #include <system/SystemConfig.h>
 
 #include <lib/support/Iterators.h>
@@ -263,7 +264,7 @@ class BitMapObjectPool : public internal::StaticAllocatorBitmap
 {
 public:
     BitMapObjectPool() : StaticAllocatorBitmap(mData.mMemory, mUsage, N, sizeof(T)) {}
-    ~BitMapObjectPool() { VerifyOrDie(Allocated() == 0); }
+    ~BitMapObjectPool() { VerifyOrDieWithObject(Allocated() == 0, this); }
 
     BitmapActiveObjectIterator<T> begin() { return BitmapActiveObjectIterator<T>(this, FirstActiveIndex()); }
     BitmapActiveObjectIterator<T> end() { return BitmapActiveObjectIterator<T>(this, N); }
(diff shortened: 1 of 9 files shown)


src/lib/support/Pool.h
```diff
@@ -323,6 +324,18 @@ class BitMapObjectPool : public internal::StaticAllocatorBitmap
         return ForEachActiveObjectInner(&proxy, &internal::LambdaProxy<T, Function>::ConstCall);
     }
 
+    void DumpToLog() const
+    {
+        ChipLogError(Support, "BitMapObjectPool: %lu allocated", static_cast<unsigned long>(Allocated()));
+        if constexpr (IsDumpable<T>::value)
+        {
+            ForEachActiveObjectInner(nullptr, [](void * context, const void * object) {
```
bzbarsky-apple, line 332: Any reason to not just use ForEachActiveObject and avoid the casting needed?

src/lib/support/Pool.h
```diff
@@ -570,6 +583,18 @@ class HeapObjectPool : public internal::Statistics, public HeapObjectPoolExitHan
         return mObjects.ForEachNode(&proxy, &internal::LambdaProxy<const T, Function>::ConstCall);
     }
 
+    void DumpToLog() const
+    {
+        ChipLogError(Support, "HeapObjectPool: %lu allocated", static_cast<unsigned long>(Allocated()));
+        if constexpr (IsDumpable<T>::value)
+        {
+            mObjects.ForEachNode(nullptr, [](void * context, const void * object) {
```
bzbarsky-apple, line 591: Again, ForEachActiveObject()?

src/lib/support/logging/TextOnlyLogging.h
```diff
@@ -292,6 +293,22 @@ using LogRedirectCallback_t = void (*)(const char * module, uint8_t category, co
...
+ * Logging helpers for RTTI, primarily useful when logging fatal errors in DumpToLog().
+ *
+ * Example:
+ * @code
+ * ChipLogError(Foo, "Delegate=" ChipLogFormatRtti, ChipLogValueRtti(mDelegate));
+ * @endcode
+ */
+#if __has_feature(cxx_rtti)
+#define ChipLogFormatRtti "%s"
+#define ChipLogValueRtti(ptr) ((ptr) != nullptr ? typeid(*(ptr)).name() : "0")
```
bzbarsky-apple, line 306:
```suggestion
#define ChipLogValueRtti(ptr) ((ptr) != nullptr ? typeid(*(ptr)).name() : "nullptr")
```

bzbarsky-apple (approved):

src/lib/support/Pool.h
ksperling-apple: I was wondering if that whole LambdaProxy stuff would cause extra overhead, but probably not, and it won't matter before a crash anyway.

andy31415 (approved):

PR #34621, opened 2024-07-30
[Bouffalo Lab] Update gcc toolchain and flash tool of docker image

wy-hh:
- Update Bouffalo Lab docker image
  - update flash tool: use BouffaloLabDevCube-v1.9.0 instead of bflb-iot-tool
  - update thead gcc toolchain download url.

### DIFF
--- a/.github/workflows/examples-bouffalolab.yaml
+++ b/.github/workflows/examples-bouffalolab.yaml
@@ -37,7 +37,7 @@ jobs:
     if: github.actor != 'restyled-io[bot]'
 
     container:
-      image: ghcr.io/project-chip/chip-build-bouffalolab:54
+      image: ghcr.io/project-chip/chip-build-bouffalolab:67
       volumes:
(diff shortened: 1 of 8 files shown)


.github/workflows/examples-bouffalolab.yaml
```diff
@@ -37,7 +37,7 @@ jobs:
     if: github.actor != 'restyled-io[bot]'
 
     container:
-      image: ghcr.io/project-chip/chip-build-bouffalolab:54
+      image: ghcr.io/project-chip/chip-build-bouffalolab:67
```
andy31415, line 40: This change has to be made in a separate PR from the dockerfile update: we update the dockerfile after the commit, so there ia few hours lag until they are available.

wy-hh: Updated to only contain chagnes for Docker image build.
Rest changes will be in another PR when Docker image is accessible.

woody-apple (approved):

andy31415 (approved):

PR #34628, opened 2024-07-30
Test automation for FabricSync ICD BridgedDeviceBasicInfoCluster

rochaferraz:
PR adds automation for the three tests to the Bridged Device Basic Information Cluster, focused on testing the KeepActive command and ActiveChanged event.

Test Plan: https://github.com/CHIP-Specifications/chip-test-plans/pull/4337

Clarification issue on ICD behaviours: https://github.com/CHIP-Specifications/chip-test-plans/issues/4377

### DIFF
--- a/src/python_testing/TC_BRBINFO_4_1.py
+++ b/src/python_testing/TC_BRBINFO_4_1.py
@@ -36,6 +36,7 @@ from mobly import asserts
 logger = logging.getLogger(__name__)
 _ROOT_ENDPOINT_ID = 0
 
+
 class TC_BRBINFO_4_1(MatterBaseTest):
 
     #
--- a/src/python_testing/TC_BRBINFO_4_1.py
+++ b/src/python_testing/TC_BRBINFO_4_1.py
@@ -44,6 +44,7 @@ kRootEndpointId = 0
 kMaxUserActiveModeBitmap = 0x1FFFF
 kMaxUserActiveModeTriggerInstructionByteLength = 128
 
+
 class TC_BRBINFO_4_1(MatterBaseTest):
 
     #
@@ -62,7 +63,6 @@ class TC_BRBINFO_4_1(MatterBaseTest):
     def default_timeout(self) -> int:
         return 63*60
 
-
     def desc_TC_BRBINFO_4_1(self) -> str:
         """Returns a description of this test"""
         return "[TC_BRBINFO_4_1] Verification of KeepActive Command [DUT-Server]"
@@ -80,12 +80,12 @@ class TC_BRBINFO_4_1(MatterBaseTest):
     def _ask_for_vendor_commissioniong_ux_operation(self, discriminator, setupPinCode, setupManualCode, setupQRCode):
         self.wait_for_user_input(
             prompt_msg=f"Using the DUT vendor's provided interface, commission the ICD device using the following parameters:\n"
-                        f"- discriminator: {discriminator}\n"
-                        f"- setupPinCode: {setupPinCode}\n"
-                        f"- setupQRCode: {setupQRCode}\n"
-                        f"- setupManualcode: {setupManualCode}\n"
-                        f"If using FabricSync Admin test app, you may type:\n"
-                        f">>> pairing onnetwork 111 {setupPinCode}")
+            f"- discriminator: {discriminator}\n"
+            f"- setupPinCode: {setupPinCode}\n"
+            f"- setupQRCode: {setupQRCode}\n"
+            f"- setupManualcode: {setupManualCode}\n"
+            f"If using FabricSync Admin test app, you may type:\n"
+            f">>> pairing onnetwork 111 {setupPinCode}")
 
     async def _send_keep_active_command(self, duration, endpoint_id) -> int:
         logging.info("Sending keep active command")
@@ -104,12 +104,12 @@ class TC_BRBINFO_4_1(MatterBaseTest):
         root_part_list = await self.read_single_attribute_check_success(cluster=Clusters.Descriptor, attribute=Clusters.Descriptor.Attributes.PartsList, endpoint=kRootEndpointId)
         set_of_endpoints_after_adding_device = set(root_part_list)
 
-        asserts.assert_true(set_of_endpoints_after_adding_device.issuperset(self.set_of_dut_endpoints_before_adding_device), "Expected only new endpoints to be added")
+        asserts.assert_true(set_of_endpoints_after_adding_device.issuperset(
+            self.set_of_dut_endpoints_before_adding_device), "Expected only new endpoints to be added")
         unique_endpoints_set = set_of_endpoints_after_adding_device - self.set_of_dut_endpoints_before_adding_device
         asserts.assert_equal(len(unique_endpoints_set), 1, "Expected only one new endpoint")
         newly_added_endpoint = list(unique_endpoints_set)[0]
         return newly_added_endpoint
-
 
     @async_test_body
     async def setup_class(self):
@@ -128,7 +128,7 @@ class TC_BRBINFO_4_1(MatterBaseTest):
         discriminator = 3850
         passcode = 20202021
         app_args = f'--secured-device-port {self.port} --discriminator {discriminator} --passcode {passcode} --KVS {self.kvs} ' + \
-          '--secured-device-port 5545'
+            '--secured-device-port 5545'
         cmd = f'{app} {app_args}'
 
         logging.info("Starting ICD Server App")
@@ -145,7 +145,7 @@ class TC_BRBINFO_4_1(MatterBaseTest):
         params = await self.openCommissioningWindow(dev_ctrl=self.default_controller, node_id=self.icd_nodeid)
 
         self._ask_for_vendor_commissioniong_ux_operation(params.randomDiscriminator, params.commissioningParameters.setupPinCode,
-            params.commissioningParameters.setupManualCode, params.commissioningParameters.setupQRCode)
+                                                         params.commissioningParameters.setupManualCode, params.commissioningParameters.setupQRCode)
 
     def teardown_class(self):
         logging.warning("Stopping app with SIGTERM")
@@ -177,37 +177,37 @@ class TC_BRBINFO_4_1(MatterBaseTest):
 
         # Confirms commissioning of DUT on TH as it reads its fature map
         await self._read_attribute_expect_success(
-          kRootEndpointId,
-          basic_info_cluster,
-          basic_info_attributes.FeatureMap,
-          self.dut_node_id
+            kRootEndpointId,
+            basic_info_cluster,
+            basic_info_attributes.FeatureMap,
+            self.dut_node_id
         )
 
         logging.info("Ensuring ICD is commissioned to TH")
 
         # Confirms commissioning of ICD on TH as it reads its feature map
         await self._read_attribute_expect_success(
-          kRootEndpointId,
-          basic_info_cluster,
-          basic_info_attributes.FeatureMap,
-          self.icd_nodeid
+            kRootEndpointId,
+            basic_info_cluster,
+            basic_info_attributes.FeatureMap,
+            self.icd_nodeid
         )
 
         self.step("1a")
 
         idle_mode_duration = await self._read_attribute_expect_success(
-          kRootEndpointId,
-          icdm_cluster,
-          icdm_attributes.IdleModeDuration,
-          self.icd_nodeid
+            kRootEndpointId,
+            icdm_cluster,
+            icdm_attributes.IdleModeDuration,
+            self.icd_nodeid
         )
         logging.info(f"IdleModeDuration: {idle_mode_duration}")
 
         active_mode_duration = await self._read_attribute_expect_success(
-          kRootEndpointId,
-          icdm_cluster,
-          icdm_attributes.ActiveModeDuration,
-          self.icd_nodeid
+            kRootEndpointId,
+            icdm_cluster,
+            icdm_attributes.ActiveModeDuration,
+            self.icd_nodeid
         )
         logging.info(f"ActiveModeDuration: {active_mode_duration}")
 
@@ -265,7 +265,7 @@ class TC_BRBINFO_4_1(MatterBaseTest):
         logging.info(f"Sending KeepActiveCommand({stay_active_duration})")
         self._send_keep_active_command(stay_active_duration, dynamic_endpoint_id)
 
-        ## stops (halts) the ICD server process by sending a SIGTOP signal
+        # stops (halts) the ICD server process by sending a SIGTOP signal
         self.app_process.send_signal(signal.SIGSTOP.value)
 
         if not self.is_ci:
@@ -273,7 +273,7 @@ class TC_BRBINFO_4_1(MatterBaseTest):
             self._timeout
             time.sleep(60*60)
 
-        ## resumes (continues) the ICD server process by sending a SIGCONT signal
+        # resumes (continues) the ICD server process by sending a SIGCONT signal
         self.app_process.send_signal(signal.SIGCONT.value)
 
         # wait for active changed event, expect no event will be sent
@@ -282,6 +282,7 @@ class TC_BRBINFO_4_1(MatterBaseTest):
             promised_active_duration = self.q.get(block=True, timeout=event_timeout)
         finally:
             asserts.assert_true(queue.Empty(), "ActiveChanged event received when not expected")
+
 
 if __name__ == "__main__":
     default_matter_test_main()
--- a/src/python_testing/TC_BRBINFO_4_1.py
+++ b/src/python_testing/TC_BRBINFO_4_1.py
@@ -220,7 +220,7 @@ class TC_BRBINFO_4_1(MatterBaseTest):
         logging.info("Waiting for ActiveChanged from DUT...")
         promised_active_duration = await self._wait_for_active_changed_event((idle_mode_duration + max(active_mode_duration, stay_active_duration))/1000)
 
-        asserts.assert_true(promised_active_duration >= stay_active_duration, "PromisedActiveDuration < StayActiveDuration")
+        asserts.assert_greater_equal(promised_active_duration, stay_active_duration, "PromisedActiveDuration < StayActiveDuration")
 
         self.step("2")
 
--- a/src/python_testing/TC_BRBINFO_4_1.py
+++ b/src/python_testing/TC_BRBINFO_4_1.py
@@ -18,19 +18,15 @@
 # See https://github.com/project-chip/connectedhomeip/blob/master/docs/testing/python.md#defining-the-ci-test-arguments
 # for details about the block below.
 #
-# === BEGIN CI TEST ARGUMENTS ===
-# test-runner-runs: run1
-# test-runner-run/run1/app: ${ALL_CLUSTERS_APP}
-# test-runner-run/run1/factoryreset: True
-# test-runner-run/run1/quiet: True
-# test-runner-run/run1/app-args: --discriminator 1234 --KVS kvs1 --trace-to json:${TRACE_APP}.json
-# test-runner-run/run1/script-args: --storage-path admin_storage.json --commissioning-method on-network --discriminator 1234 --passcode 20202021 --PICS src/app/tests/suites/certification/ci-pics-values --trace-to json:${TRACE_TEST_JSON}.json --trace-to perfetto:${TRACE_TEST_PERFETTO}.perfetto
-# === END CI TEST ARGUMENTS ===
+
+# This test requires a TH_SERVER application. Please specify with --string-arg th_server_app_path:<path_to_app>
+# TH_SERVER must support following arguments: --secured-device-port --discriminator --passcode --KVS --secured-device-port
+# E.g: python3 src/python_testing/TC_BRBINFO_4_1.py --commissioning-method on-network --qr-code MT:-24J042C00KA0648G00 \
+#      --string-arg th_server_app_path:out/linux-x64-lit-icd/lit-icd-app
 
 import logging
 import os
 import queue
-import pathlib
 import uuid
 import subprocess
 import time
@@ -38,21 +34,15 @@ import signal
 
 
 import chip.clusters as Clusters
-from matter_testing_support import MatterBaseTest, TestStep, async_test_body, default_matter_test_main, SimpleEventCallback
+from chip import ChipDeviceCtrl
+from matter_testing_support import (MatterBaseTest, TestStep, async_test_body, default_matter_test_main, SimpleEventCallback)
 from mobly import asserts
 
-from chip import ChipDeviceCtrl  # Needed before chip.FabricAdmin
 
 logger = logging.getLogger(__name__)
 kRootEndpointId = 0
-kICDMEndpointId = 0
-kICDBridgedEndpointId = 2
 kMaxUserActiveModeBitmap = 0x1FFFF
 kMaxUserActiveModeTriggerInstructionByteLength = 128
-
-#uat = cluster.Bitmaps.UserActiveModeTriggerBitmap
-#modes = cluster.Enums.OperatingModeEnum
-#features = cluster.Bitmaps.Feature
 
 class TC_BRBINFO_4_1(MatterBaseTest):
 
@@ -66,6 +56,12 @@ class TC_BRBINFO_4_1(MatterBaseTest):
     #
     # Test Harness Helpers
     #
+
+    # Override default timeout to support a 60 min wait
+    @property
+    def default_timeout(self) -> int:
+        return 63*60
+
 
     def desc_TC_BRBINFO_4_1(self) -> str:
         """Returns a description of this test"""
@@ -82,13 +78,6 @@ class TC_BRBINFO_4_1(MatterBaseTest):
         return steps
 
 
-    def pics_TC_BRBINFO_4_1(self) -> list[str]:
-        """ This function returns a list of PICS for this test case that must be True for the test to be run"""
-        pics = [
-            "BRBINFO.C.80",
-        ]
-        return pics
-
     def _ask_for_vendor_commissioniong_ux_operation(self, discriminator, setupPinCode):
         self.wait_for_user_input(
             prompt_msg=f"Using the DUT vendor's provided interface, commission the ICD device using the following parameters:\n"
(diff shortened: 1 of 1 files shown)


src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,287 @@
...
+                        f"If using FabricSync Admin, you may type:\n"
+                        f">>> pairing onnetwork 111 {setupPinCode}")
+
+    def _display_message_and_wait_for_user_input(self, message):
+        self.wait_for_user_input(prompt_msg=message)
+
+    async def _send_keep_active_command(self, duration):
+        logging.info("Mock sending keep active command")
+        # TODO check success
+        #keepActive = await self.default_controller.SendCommand(nodeid=self.dut_node_id, endpoint=kICDBridgedEndpointId, payload=brbinfoCluster.Commands.KeepActive(stayActiveDuration=duration))
```
tehampson, line 112: When you uncomment this line (which I am assuming you will do when you take this out of draft), I think you will need self.brbinfoCluster

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,287 @@
...
+
+    async def _wait_for_active_changed_event(self, timeout) -> int:
+        try:
+            PromisedActiveDuration = self.q.get(block=True, timeout=timeout)
+            logging.info(f"PromisedActiveDuration: {PromisedActiveDuration}")
+            return PromisedActiveDuration
+        except queue.Empty:
+            asserts.assert_true(False, "Timeout on event")
```
tehampson, line 120: nit, you can assert_fail directly
```suggestion
            asserts.assert_fail("Timeout on event")
```

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,287 @@
...
+        icdmAttributes = icdmCluster.Attributes
+        brbinfoCluster = Clusters.Objects.BridgedDeviceBasicInformation
+        brbinfoAttributes = brbinfoCluster.Attributes
+        basicinfoCluster = Clusters.Objects.BasicInformation
+        basicinfoAttributes = basicinfoCluster.Attributes
+
+        self.endpoint = 0
+
+        # Check commissioned DUT
+        self.step(1)
```
tehampson, line 177: Make sure your step lines align with spec text test steps. Like here you are saying step and you read an attribute but the test step says "Commissioning of the DUT, already done".

It is okay to have a few test steps already done you can do:
```
self.step(1)  # Done in setup
self.step(2)  # Done in setup
self.step(3)
```

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,287 @@
...
+    def desc_TC_BRBINFO_4_1(self) -> str:
+        """Returns a description of this test"""
+        return "[TC_BRBINFO_4_1] Verification of KeepActive Command [DUT-Server]"
+
+    def steps_TC_BRBINFO_4_1(self) -> list[TestStep]:
+        steps = [
+            TestStep(1, "Commissioning of the DUT, already done", is_commissioning=True),
+            TestStep(2, "Commissioning of the ICD, already done", is_commissioning=True),
+            TestStep(3, "TH reads from the ICD the IDLE_MODE_DURATION and ACTIVE_MODE_DURATION attributes"),
+            TestStep(4, "Simple KeepActive command w/ subscription. ActiveChanged event received by TH contains PromisedActiveDuration"),
```
tehampson, line 84: I see your test plan has steps like `1a`, just letting you know you can do something like:
```
TestStep("1a", "What you do in step 1a"),

....

# when using it later
self.step("1a")
```

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,287 @@
...
+        brbinfoCluster = Clusters.Objects.BridgedDeviceBasicInformation
+        brbinfoAttributes = brbinfoCluster.Attributes
+        basicinfoCluster = Clusters.Objects.BasicInformation
+        basicinfoAttributes = basicinfoCluster.Attributes
+
+        self.endpoint = 0
+
+        # Check commissioned DUT
+        self.step(1)
+        # Confirms commissioning of DUT as it reads basic info from bridge device
```
tehampson, line 178: nit, comment say bridged but you are accessing basic info.

I think you want to access the featureMap of the dynamic endpoint representing the ICD device on DUT. If that is the case let me know and I will generate the code I think you will need for that. As we will need that endpoint ahead of time and I think we can add it in the setup section

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,287 @@
...
+
+        # Subscription to ActiveChanged
+        event = brbinfoCluster.Events.ActiveChanged
+        self.q = queue.Queue()
+        urgent = 1
+        cb = SimpleEventCallback("ActiveChanged", event.cluster_id, event.event_id, self.q)
+        subscription = await self.default_controller.ReadEvent(nodeid=self.dut_node_id, events=[(kICDBridgedEndpointId, event, urgent)], reportInterval=[1, 3])
+        subscription.SetEventUpdateCallback(callback=cb)
+
+        StayActiveDuration = 1000
```
tehampson, line 224: local variable names should use lower_snake_case

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,287 @@
...
+        super().teardown_class()
+
+    #
+    # BRBINFO 4.1 Test Body
+    #
+
+    @async_test_body
+    async def test_TC_BRBINFO_4_1(self):
+
+        icdmCluster = Clusters.Objects.IcdManagement
```
tehampson, line 167: local variable should use lower_snake_case

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,287 @@
...
+        PromisedActiveDuration = await self._wait_for_active_changed_event((idle_mode_duration + max(active_mode_duration, StayActiveDuration))/1000)
+
+        # wait for active time duration
+        time.sleep(max(StayActiveDuration/1000, active_mode_duration))
+        # ICD now should be in idle mode
+
+        # sends 3x keep active commands
+        logging.info(f"Sending KeepActiveCommand({StayActiveDuration})")
+        self._send_keep_active_command(StayActiveDuration)
+        time.sleep(StayActiveDuration/1000)
```
tehampson, line 250: test spec says you should sleep for 100 ms here why is this a function of `StayActiveDuration`

src/python_testing/TC_BRBINFO_4_1.py
rochaferraz: Nice. I wrongly assumed tests needed to be integers

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,286 @@
...
+        # Test 3
+
+        StayActiveDuration = 10000
+        logging.info(f"Sending KeepActiveCommand({StayActiveDuration})")
+        self._send_keep_active_command(StayActiveDuration)
+
+        ## halt the ICD process
+        self.app_process.send_signal(signal.SIGSTOP.value)
+
+        logging.info(f"Waiting for 60 minutes")
```
tehampson, line 272: I suggest this last step of waiting 60 mins be skipped if it is CI

```
self.is_ci = self.check_pics('PICS_SDK_CI_ONLY') at the start of the test
```

then check for 
```
if not self.is_ci:
   # Run the 60 min test test step
```

src/python_testing/TC_BRBINFO_4_1.py
rochaferraz: fixed

src/python_testing/TC_BRBINFO_4_1.py
rochaferraz: fixed

src/python_testing/TC_BRBINFO_4_1.py
rochaferraz: The idea here was simply to access some cluster, like the mandatory basic info just to ensure they are commissioned to the fabric. Changed comments. PTAL

src/python_testing/TC_BRBINFO_4_1.py
rochaferraz: ack

src/python_testing/TC_BRBINFO_4_1.py
rochaferraz: ack

src/python_testing/TC_BRBINFO_4_1.py
rochaferraz: good catch

src/python_testing/TC_BRBINFO_4_1.py
rochaferraz: ack

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,278 @@
...
+            TestStep("2", "Sends 3x KeepActive commands w/ subscription. ActiveChanged event received ONCE and contains PromisedActiveDuration"),
+            TestStep("3", "KeepActive not returned after 60 minutes of offline ICD"),
+        ]
+        return steps
+
+
+    def pics_TC_BRBINFO_4_1(self) -> list[str]:
+        """ This function returns a list of PICS for this test case that must be True for the test to be run"""
+        pics = [
+            "BRBINFO.C.80",
```
cecille, line 88:
```suggestion
            "BRBINFO.S.C80.Rsp",
```

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,278 @@
...
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+# See https://github.com/project-chip/connectedhomeip/blob/master/docs/testing/python.md#defining-the-ci-test-arguments
+# for details about the block below.
+#
+# === BEGIN CI TEST ARGUMENTS ===
+# test-runner-runs: run1
+# test-runner-run/run1/app: ${ALL_CLUSTERS_APP}
```
cecille, line 23: all clusters doesn't have this cluster on any endpoint. What is this actually supposed to run against?

src/python_testing/TC_BRBINFO_4_1.py
cecille: But actually, these endpoints are dynamic, so they won't appear in the PICS file since you don't know a priori what the endpoint is. Suggest instead using the @per_endpoint_test decorator to have the test harness select the appropriate endpoints at run time.

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,278 @@
...
+        """ This function returns a list of PICS for this test case that must be True for the test to be run"""
+        pics = [
+            "BRBINFO.C.80",
+        ]
+        return pics
+
+    def _ask_for_vendor_commissioniong_ux_operation(self, discriminator, setupPinCode):
+        self.wait_for_user_input(
+            prompt_msg=f"Using the DUT vendor's provided interface, commission the ICD device using the following parameters:\n"
+                        f"- discriminator: {discriminator}\n"
```
cecille, line 95: The vendor might require the manual or QR code. I believe you have both the discriminator and the setup code hard coded, so you will know what these are if you know the VID/PID

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,278 @@
...
+                        f"- setupPinCode: {setupPinCode}\n"
+                        f"If using FabricSync Admin, you may type:\n"
+                        f">>> pairing onnetwork 111 {setupPinCode}")
+
+    def _display_message_and_wait_for_user_input(self, message):
```
cecille, line 100: Why does this function need to be wrapped?

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,278 @@
...
+        self.wait_for_user_input(prompt_msg=message)
+
+    async def _send_keep_active_command(self, duration) -> int:
+        logging.info("Mock sending keep active command")
```
cecille, line 104: As far as this test is concerned, it has no idea if this is being mocked or not. It's sending a command and receiving a response from what it thinks is the DUT, even if that's actually the mock device.

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,278 @@
...
+            logging.info(f"PromisedActiveDuration: {promised_active_duration}")
+            return promised_active_duration
+        except queue.Empty:
+            asserts.fail("Timeout on event")
+
+
+    @async_test_body
+    async def setup_class(self):
+        super().setup_class()
+        app = os.path.join(pathlib.Path(__file__).resolve().parent, '..','..','out', 'linux-x64-lit-icd', 'lit-icd-app')
```
cecille, line 120: See changes here https://github.com/project-chip/connectedhomeip/pull/34652 - per andrei's request, this should be configurable (he's totally right) But also, this should be consolidated with the other test if you're planning to use it here as well.

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,278 @@
...
+
+        logging.info("Commissioning of ICD to fabric one (TH)")
+        self.icd_nodeid = 1111
+
+        await self.default_controller.CommissionOnNetwork(nodeId=self.icd_nodeid, setupPinCode=passcode, filterType=ChipDeviceCtrl.DiscoveryFilterType.LONG_DISCRIMINATOR, filter=discriminator)
+
+        logging.info("Commissioning of ICD to fabric two (DUT)")
+        params = await self.openCommissioningWindow(dev_ctrl=self.default_controller, node_id=self.icd_nodeid)
+
+        self._ask_for_vendor_commissioniong_ux_operation(params.randomDiscriminator, params.commissioningParameters.setupPinCode)
```
cecille, line 143: so this is the vendor commissioning the app that the TH just pulled up? Do all the DUTs being tested here have to support ICD? Does it need to be a specific device type?

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,278 @@
...
+        subscription.SetEventUpdateCallback(callback=cb)
+
+        stay_active_duration = 1000
+        logging.info(f"Sending KeepActiveCommand({stay_active_duration}ms)")
+        self._send_keep_active_command(stay_active_duration)
+
+        logging.info("Waiting for ActiveChanged from DUT...")
+        promised_active_duration = await self._wait_for_active_changed_event((idle_mode_duration + max(active_mode_duration, stay_active_duration))/1000)
+
+        asserts.assert_true(promised_active_duration >= stay_active_duration, "PromisedActiveDuration < StayActiveDuration")
```
cecille, line 223:
```suggestion
        asserts.assert_greater_equal(promised_active_duration, stay_active_duration, "PromisedActiveDuration < StayActiveDuration")
```

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,278 @@
...
+
+        stay_active_duration = 1500
+        logging.info(f"Sending KeepActiveCommand({stay_active_duration}ms)")
+        self._send_keep_active_command(stay_active_duration)
+
+        logging.info("Waiting for ActiveChanged from DUT...")
+        promised_active_duration = await self._wait_for_active_changed_event((idle_mode_duration + max(active_mode_duration, stay_active_duration))/1000)
+
+        # wait for active time duration
+        time.sleep(max(stay_active_duration/1000, active_mode_duration))
```
cecille, line 235: should this be promised_active_duration?

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,278 @@
...
+        # ICD now should be in idle mode
+
+        # sends 3x keep active commands
```
cecille, line 238: Is this is a LIT device? If so, don't you need to wait for the check in to send a new command?

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,278 @@
...
+
+        stay_active_duration = 10000
+        logging.info(f"Sending KeepActiveCommand({stay_active_duration})")
+        self._send_keep_active_command(stay_active_duration)
+
+        ## halt the ICD process
+        self.app_process.send_signal(signal.SIGSTOP.value)
+
+        if not self.is_ci:
+            logging.info("Waiting for 60 minutes")
```
cecille, line 264: To do this, you'll need to adjust the test timeout. Right now it's on the order of minutes. You need WAY more than that. Actually, are you not already seeing timeouts on this test with the sleeps above?

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,278 @@
...
+            time.sleep(60*60)
+
+        ## resume the ICD
+        self.app_process.send_signal(signal.SIGCONT.value)
```
cecille, line 268: Is this signal something the ICD app recognizes? Can you add a comment here indicating what's happening on the device?

src/python_testing/TC_BRBINFO_4_1.py
tehampson: I would recommend removing this block entirely for now. When we get CI working (hopefully in roughly 2-3 weeks) we can add that in at that time.

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,278 @@
...
+import subprocess
+import time
+import signal
+
+
+import chip.clusters as Clusters
+from matter_testing_support import MatterBaseTest, TestStep, async_test_body, default_matter_test_main, SimpleEventCallback
+from mobly import asserts
+
+from chip import ChipDeviceCtrl  # Needed before chip.FabricAdmin
```
tehampson, line 44: I think you can reorder this as you are not including chip.FabricAdmin

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,278 @@
...
+logger = logging.getLogger(__name__)
+kRootEndpointId = 0
+kICDMEndpointId = 0
+kICDBridgedEndpointId = 2
+kMaxUserActiveModeBitmap = 0x1FFFF
+kMaxUserActiveModeTriggerInstructionByteLength = 128
+
+#uat = cluster.Bitmaps.UserActiveModeTriggerBitmap
+#modes = cluster.Enums.OperatingModeEnum
+#features = cluster.Bitmaps.Feature
```
tehampson, line 55: nit, remove these if they are not used

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,278 @@
...
+import chip.clusters as Clusters
+from matter_testing_support import MatterBaseTest, TestStep, async_test_body, default_matter_test_main, SimpleEventCallback
+from mobly import asserts
+
+from chip import ChipDeviceCtrl  # Needed before chip.FabricAdmin
+
+logger = logging.getLogger(__name__)
+kRootEndpointId = 0
+kICDMEndpointId = 0
+kICDBridgedEndpointId = 2
```
tehampson, line 49: This is an assumption that I don't know if we can make. I suggest doing something like this to identify the endpoint

https://github.com/project-chip/connectedhomeip/blob/51a9cedd2612ff4ad8cb2118a7767a96d82c6dcd/src/python_testing/TC_MCORE_FS_1_3.py#L118-L137

Where I commission the device you would call `_ask_for_vendor_commissioniong_ux_operation`

src/python_testing/TC_BRBINFO_4_1.py
tehampson: @cecille 
I don't think ` @per_endpoint_test` works here. The endpoint that you are targeting gets added in the middle of the test and should be identified

Would it make sense for this PICS to be on `MCORE.FS`?

cecille: Yeah, if the endpoint doesn't appear until after the test is started, then MCORE.FS is appropriate.

rochaferraz: ack

src/python_testing/TC_BRBINFO_4_1.py
rochaferraz: in any case function is not used anymore

src/python_testing/TC_BRBINFO_4_1.py
rochaferraz: ack

src/python_testing/TC_BRBINFO_4_1.py
rochaferraz: ack

src/python_testing/TC_BRBINFO_4_1.py
rochaferraz: ack

src/python_testing/TC_BRBINFO_4_1.py
rochaferraz: ack

src/python_testing/TC_BRBINFO_4_1.py
rochaferraz: from the tests I ran in terminal `kill -SIGSTOP / -SIGCONT` the process recognizes and suspends/resumes execution. It's not clear to me how to resume log access w/o using `fg` but that doesn't matter here

src/python_testing/TC_BRBINFO_4_1.py
rochaferraz: yes

src/python_testing/TC_BRBINFO_4_1.py
rochaferraz: In this case I want to send KeepActive commands to DUT precisely when they ICD is idle. My understanding is that DUT then should send a single ActiveChanged event back to TH whenever ICD becomes active. I think that was the spirit of the test proposed

src/python_testing/TC_BRBINFO_4_1.py
rochaferraz: Yes, I open the commissioning window for the ICD server and ask the DUT vendor to commission it, giving the command for convenience if the tester's DUT is using something close to the sample FabricSync bridge interface.

> Do all the DUTs being tested here have to support ICD
Yes

> Does it need to be a specific device type?
You mean the ICD?

src/python_testing/TC_BRBINFO_4_1.py
rochaferraz: done

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,286 @@
...
+    #
+
+    async def _read_attribute_expect_success(self, endpoint, cluster, attribute, node_id):
+        return await self.read_single_attribute_check_success(endpoint=endpoint, cluster=cluster, attribute=attribute, node_id=node_id)
+
+    #
+    # Test Harness Helpers
+    #
+
+    # Override default timeout to support a 60 min wait
```
tehampson, line 59: nit comment doesn't match what you are doing here, you have 63 below

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,286 @@
...
+import time
+import uuid
+
+import chip.clusters as Clusters
+from chip import ChipDeviceCtrl
+from matter_testing_support import MatterBaseTest, SimpleEventCallback, TestStep, async_test_body, default_matter_test_main
+from mobly import asserts
+
+logger = logging.getLogger(__name__)
+kRootEndpointId = 0
```
tehampson, line 41: I know there is a Precident in other files where they are doing kGlobalConstant. But this is really not the pythonic way autopep8 is `GLOBAL_CONSTANT_NAME`

There is also a prefix if an `_` if this is supposed to be internal which it is in this case.

Essentially please do: `_GLOBAL_CONSTANT_NAME`

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,286 @@
...
+            return promised_active_duration
+        except queue.Empty:
+            asserts.fail("Timeout on event ActiveChanged")
+
+    async def _get_dynamic_endpoint(self) -> int:
+        root_part_list = await self.read_single_attribute_check_success(cluster=Clusters.Descriptor, attribute=Clusters.Descriptor.Attributes.PartsList, endpoint=kRootEndpointId)
+        set_of_endpoints_after_adding_device = set(root_part_list)
+
+        asserts.assert_true(set_of_endpoints_after_adding_device.issuperset(
+            self.set_of_dut_endpoints_before_adding_device), "Expected only new endpoints to be added")
```
tehampson, line 102: nit,

recommend passing `self.set_of_dut_endpoints_before_adding_device` as a parameter instead and not adding it to self ever just having it be a local variable

tehampson: nvm I see what you did

src/python_testing/TC_BRBINFO_4_1.py
rochaferraz: ack

src/python_testing/TC_BRBINFO_4_1.py
rochaferraz: ack

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -0,0 +1,286 @@
...
+        self._send_keep_active_command(stay_active_duration, dynamic_endpoint_id)
+        time.sleep(100)
+        logging.info(f"Sending KeepActiveCommand({stay_active_duration})")
+        self._send_keep_active_command(stay_active_duration, dynamic_endpoint_id)
+        time.sleep(100)
+
+        logging.info("Waiting for ActiveChanged from DUT...")
+        promised_active_duration = await self._wait_for_active_changed_event((idle_mode_duration + max(active_mode_duration, stay_active_duration))/1000)
+
+        asserts.assert_true(self.q.qSize() == 0, "More than one event received from DUT")
```
tehampson, line 254:
```suggestion
        asserts.assert_equal(self.q.qSize(), 0, "More than one event received from DUT")
```

tehampson (approved):

src/python_testing/TC_BRBINFO_4_1.py
cecille: is it? My understanding was that it shouldn't send back multiple active changed events if it actually receives the commands. I don't think an idle device will actually process these. Can you please file a follow up issue to check this behaviour with the ICD team to ensure it is a) testing what needs to be tested and b) works with how the ICD devices work?

cecille: Follow up is OK here for TE. Please link the issue into this test plan before closing this issue. Then you can close. I'll approve this now on that condition.

cecille (approved):

samadDotDev (approved):

saurabhst (approved):

PR #34634, opened 2024-07-30
Initial Test Script for TC_MCORE_FS_1_1

cletnick:
Initial Test Script for FS Top-Level.

Test plan spec landed in https://github.com/CHIP-Specifications/chip-test-plans/pull/4328

Cases:
TC-MCORE.FS-1.1

### DIFF
--- a/src/python_testing/TC_MCORE_FS_1_1.py
+++ b/src/python_testing/TC_MCORE_FS_1_1.py
@@ -145,5 +145,6 @@ class TC_MCORE_FS_1_1(MatterBaseTest):
             asserts.assert_equal(len(th_fsa_server_fabrics) + 1, len(th_fsa_server_fabrics_new),
                                  "Unexpected number of fabrics on TH_SERVER")
 
+
 if __name__ == "__main__":
     default_matter_test_main()
--- a/src/python_testing/TC_MCORE_FS_1_1.py
+++ b/src/python_testing/TC_MCORE_FS_1_1.py
@@ -1,0 +1,146 @@
+#
+#    Copyright (c) 2024 Project CHIP Authors
+#    All rights reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+# See https://github.com/project-chip/connectedhomeip/blob/master/docs/testing/python.md#defining-the-ci-test-arguments
+# for details about the block below.
+#
+
+import ipaddress
+import logging
+import os
+import pathlib
+import random
+import signal
+import subprocess
+import time
+import uuid
+
+import chip.clusters as Clusters
+from chip import ChipDeviceCtrl
+from chip.interaction_model import InteractionModelError, Status
+from matter_testing_support import MatterBaseTest, TestStep, async_test_body, default_matter_test_main, has_cluster, per_endpoint_test
+from mobly import asserts
+
+
+class TC_MCORE_FS_1_1(MatterBaseTest):
+
+    @async_test_body
+    async def setup_class(self):
+        super().setup_class()
+        self.kvs = f'kvs_{str(uuid.uuid4())}'
+
+    def teardown_class(self):
+        os.remove(self.kvs)
+        super().teardown_class()
+
+        if self.app_process:
+            logging.warning("Stopping app with SIGTERM")
+            self.app_process.send_signal(signal.SIGTERM.value)
+            self.app_process.wait()
+
+    def steps_TC_MCORE_FS_1_1(self) -> list[TestStep]:
+        steps = [TestStep(1, "Enable Fabric Synchronization on DUT_FSA using the manufacturer specified mechanism."),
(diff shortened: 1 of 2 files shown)


src/python_testing/TC_MCORE_FS_1_1.py
```diff
@@ -0,0 +1,150 @@
...
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+# See https://github.com/project-chip/connectedhomeip/blob/master/docs/testing/python.md#defining-the-ci-test-arguments
+# for details about the block below.
```
andy31415, line 19: there is no block below ... delete this? Maybe a comment on why this is not CI?

src/python_testing/TC_MCORE_FS_1_1.py
```diff
@@ -0,0 +1,150 @@
...
+        self.TH_server_controller = new_fabric_admin.NewController(nodeId=112233, paaTrustStorePath=paa_path)
+        self.server_nodeid = 1111
+        await self.TH_server_controller.CommissionOnNetwork(nodeId=self.server_nodeid, setupPinCode=passcode, filterType=ChipDeviceCtrl.DiscoveryFilterType.LONG_DISCRIMINATOR, filter=discriminator)
+        logging.info("Commissioning TH_SERVER complete")
+
+    def teardown_class(self):
+        logging.warning("Stopping app with SIGTERM")
+        self.app_process.send_signal(signal.SIGTERM.value)
+        self.app_process.wait()
+        # TODO: Use timeout, if term doesn't work, try SIGINT
```
andy31415, line 75: is this still under development? can this be cleaned up if it just works?

src/python_testing/TC_MCORE_FS_1_1.py
```diff
@@ -0,0 +1,150 @@
...
+
+class TC_MCORE_FS_1_1(MatterBaseTest):
+
+    @async_test_body
+    async def setup_class(self):
+        super().setup_class()
+        # TODO: This needs to come from an arg and needs to be something available on the TH
+        # TODO: confirm whether we can open processes like this on the TH
+        # TODO: When bridge device is more stable switch from acll-clusters-app to bridge app.
+        app = os.path.join(pathlib.Path(__file__).resolve().parent, '..', '..', 'out',
```
andy31415, line 45: This feels off ... it hardcodes the build type. We need something better - at worst a constant somewhere else.

src/python_testing/test_testing/test_TC_MCORE_FS_1_1.py
```diff
@@ -0,0 +1,135 @@
...
+                                       EventId=Clusters.CommissionerControl.Events.CommissioningRequestResult.event_id, EventNumber=1)
+        data = Clusters.CommissionerControl.Events.CommissioningRequestResult(
+            requestId=0x1234567812345678, clientNodeId=112233, statusCode=0)
+        result = Attribute.EventReadResult(Header=header, Status=Status.Success, Data=data)
+        return [result]
+    else:
+        raise InteractionModelError(Status.Failure)
+
+
+def wildcard() -> Attribute.AsyncReadTransaction.ReadResponse:
```
andy31415, line 77: needs comments on what it does.

src/python_testing/TC_MCORE_FS_1_1.py
```diff
@@ -0,0 +1,150 @@
...
+                           'linux-x64-all-clusters-no-ble', 'chip-all-clusters-app')
+
+        self.kvs = f'kvs_{str(uuid.uuid4())}'
+        self.port = 5543
+        discriminator = random.randint(0, 4095)
+        discriminator = 3840
```
andy31415, line 51: you set it to random just above ... please do one self-review pass on this PR and clean up the flow a bit. I think this went through some iterations and does not feel like a final form.

src/python_testing/TC_MCORE_FS_1_1.py
cecille: oop, probably my fault. I had that in my test and forgot to turn it into an option. Will put up a follow up

src/python_testing/TC_MCORE_FS_1_1.py
cecille: again, just copied from my code. I'll add a cleanup PR, but it would be better to consolidate these if you're using the same setup.

It does work, but for some reason we use sigint in the test runner. Which, IMO is incorrect.

src/python_testing/TC_MCORE_FS_1_1.py
```diff
@@ -0,0 +1,150 @@
...
+import time
+import uuid
+
+import chip.clusters as Clusters
+from chip import ChipDeviceCtrl
+from matter_testing_support import MatterBaseTest, TestStep, async_test_body, default_matter_test_main
+from mobly import asserts
+
+
+class TC_MCORE_FS_1_1(MatterBaseTest):
```
cecille, line 37: General Q - what's the difference between this test and the CCTRL test? I can't see anything that's in here that didn't come from that.

tehampson: The happy path of CCTRL is identical to this test. The difference is this is a FS test, and the other is a cluster test.

I would argue this should be killed off, but for the sake of matching test plan spec for TE2 deadline I just copied it

cecille: https://github.com/CHIP-Specifications/chip-test-plans/issues/4369.

cecille (approved):

saurabhst (approved):

andy31415:
Merging directly: this script is not run by CI.

PR #34311, opened 2024-07-12
Fuzzers Implementation and build configurations

woody-apple:
• Added a base38 roundtrip Fuzzer to fuzz the encoding and decoding in base38
• Added a base38 decoder Fuzzer to fuzz the function decoding base38 strings
• Added a dependancy in TLV-Reader config to correct the build error

PR #34639, opened 2024-07-30
Add a Base38 fuzzer.

bzbarsky-apple:
This is split out from https://github.com/project-chip/connectedhomeip/pull/34311 to see which part of that PR is leading to clang-tidy failures.

### DIFF
--- a/src/setup_payload/tests/FuzzBase38Decode.cpp
+++ b/src/setup_payload/tests/FuzzBase38Decode.cpp
@@ -1,0 +1,24 @@
+#include <cstddef>
+#include <cstdint>
+#include <iostream>
+
+#include <setup_payload/Base38Decode.h>
+
+using namespace chip;
+
+/**
+ *    @file
+ *      This file describes a Fuzzer for decoding base38 encoded strings.
+ */
+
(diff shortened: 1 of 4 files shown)


src/setup_payload/tests/FuzzBase38Decode.cpp
```diff
@@ -0,0 +1,24 @@
...
+#include <cstdint>
+#include <iostream>
+
+#include <setup_payload/Base38Decode.h>
+
+using namespace chip;
+
+/**
+ *    @file
+ *      This file describes a Fuzzer for decoding base38 encoded strings.
```
andy31415, line 11: is this file comment valuable in reading code? I see the file name is `FuzzBase38Decode` and the main method creates a `base38EncodedString` and calls `chip::base38Decode`. Why are we taking up vertical space?

These file comments just end up being copied & pasted and create churn to be updated.

andy31415 (approved):

src/setup_payload/tests/FuzzBase38Decode.cpp
bzbarsky-apple: It's worth describing what exactly is being fuzzed (or not fuzzed)....

kiel-apple (approved):

PR #34649, opened 2024-07-30
Fix logging of received message sizes.

bzbarsky-apple:
For encrypted messages, we were not including the size of the MIC in the logged size, so the sizes logged on the send and received side were off by 16 from each other.


kiel-apple (approved):

andy31415 (approved):

andy31415:
Restyler seems stuck, so I will just merge it for now.

PR #34650, opened 2024-07-30
Add TC_MCORE_FS_1_3.py test implementation

tehampson:
Test plan spec landed in https://github.com/CHIP-Specifications/chip-test-plans/pull/4328


src/python_testing/TC_MCORE_FS_1_3.py
```diff
@@ -0,0 +1,158 @@
...
+        self.print_step(0, "Commissioning DUT to TH, already done")
+        self.step(1)
+        # These steps are not explicitly in step 1, but they help identify the dynamically added endpoint in step 1.
+        root_node_endpoint = 0
+        root_part_list = await self.read_single_attribute_check_success(cluster=Clusters.Descriptor, attribute=Clusters.Descriptor.Attributes.PartsList, endpoint=root_node_endpoint)
+        set_of_endpoints_before_adding_device = set(root_part_list)
+
+        discriminator, setup_pin_code = await self.create_device_for_dut_ecosystem()
+        self.wait_for_user_input(
+            prompt_msg=f"Using the DUT vendor's provided interface, commission the device using the following parameters:\n"
```
cecille, line 129: Will the DUT be able to commission from discriminator / passcode, or will it need a manual code / QR code as well?

src/python_testing/TC_MCORE_FS_1_3.py
```diff
@@ -0,0 +1,158 @@
...
+        asserts.assert_equal(len(unique_endpoints_set), 1, "Expected only one new endpoint")
+        newly_added_endpoint = list(unique_endpoints_set)[0]
+
+        th_sed_dut_unique_id = await self.read_single_attribute_check_success(cluster=Clusters.BridgedDeviceBasicInformation, attribute=Clusters.BridgedDeviceBasicInformation.Attributes.UniqueID, endpoint=newly_added_endpoint)
+        asserts.assert_true(type_matches(th_sed_dut_unique_id, str), "UniqueID should be a string")
+        asserts.assert_true(th_sed_dut_unique_id, "UniqueID should not be an empty string")
+
+        self.step(2)
+        await self.create_and_commission_device_for_th_ecosystem()
+        # TODO During setup we need to create the TH_FSA device where we would commission device created
```
cecille, line 149: Is this just a separate device that's meant to commission the DUT? Why can this not just be a separate controller on a different fabric?

src/python_testing/TC_MCORE_FS_1_3.py
tehampson: If I did need manual code / QR code, how would I get that?

src/python_testing/TC_MCORE_FS_1_3.py
tehampson: There is something missing from test plan that I will add.

Where it sat `TH_FSA`, that needs to be two device `TH` and `TH_FSA`. The reason we cannot use a different fabric here is that when the `DUT_FSA` talks to the `TH_FSA` it needs to act like a fabric sync device which chip-repl is not capable of doing at the moment

src/python_testing/TC_MCORE_FS_1_3.py
cecille: chip-tool can do it statically as long as you know the PID/VID of the device. We don't have the generator code piped through to python in a way that you could just pass in the values and allocate dynamically. You COULD, but it's work and what i thin you might actually want to do here is to check the pid/vid of the supplied app to ensure it's the one you're expecting (ie, that it'll work for this test).

Anyway, to generate from chip-tool: 
chip-tool payload generate-qrcode or chip-tool payload generate-qrcode or chip-tool generate-manualcode

Ex:
(py) cecille@cecille1:~/chip/connectedhomeip$ ./out/linux-x64-chip-tool/chip-tool payload generate-manualcode --setup-pin-code 20202021 --version 0 --vendor-id 0xFFF1 --product-id 0x8000 --discriminator 3840
[1722395593.392] [2219234:2219234] [DL] ChipLinuxStorage::Init: Using KVS config file: /tmp/chip_tool_kvs
[1722395593.392] [2219234:2219234] [TOO] Manual Code: 34970112332
(py) cecille@cecille1:~/chip/connectedhomeip$ ./out/linux-x64-chip-tool/chip-tool payload generate-qrcode --setup-pin-code 20202021 --version 0 --vendor-id 0xFFF1 --product-id 0x8000 --discriminator 3840
[1722395618.748] [2219561:2219561] [DL] ChipLinuxStorage::Init: Using KVS config file: /tmp/chip_tool_kvs
[1722395618.748] [2219561:2219561] [TOO] QR Code: MT:Y.K90-Q000KA0648G00

src/python_testing/TC_MCORE_FS_1_3.py
cecille: OK, so there needs to be an app where you can trigger the adding of a new BridgedDeviceBasicInformation endpoint?

Can you open a followup for this?

tehampson: https://github.com/CHIP-Specifications/chip-test-plans/issues/4375

src/python_testing/TC_MCORE_FS_1_3.py
tehampson: Mostly done right now.

I still need to do validate assumptions for vendor ID and product ID. This will need TH to commission the device first and open a window, but I have a meeting I need to go to right now. I will add that code in around 10:45 AM EST July 31st

src/python_testing/TC_MCORE_FS_1_3.py
```diff
@@ -0,0 +1,161 @@
...
+        # These steps are not explicitly in step 1, but they help identify the dynamically added endpoint in step 1.
+        root_node_endpoint = 0
+        root_part_list = await self.read_single_attribute_check_success(cluster=Clusters.Descriptor, attribute=Clusters.Descriptor.Attributes.PartsList, endpoint=root_node_endpoint)
+        set_of_endpoints_before_adding_device = set(root_part_list)
+
+        discriminator, passcode = await self.create_device_for_dut_ecosystem()
+        # To prevent need for runtime QR code generation test makes some assumptions
+        # about the app that it valiudates here.
+        asserts.assert_equal(discriminator, 3840, "Test implementation assumption incorrect for discriminator")
+        asserts.assert_equal(passcode, 3840, "Test implementation assumption incorrect for passcode")
```
tpmanley, line 126:
```suggestion
        asserts.assert_equal(passcode, 20202021, "Test implementation assumption incorrect for passcode")
```

tehampson: Ops... Thanks!

saurabhst (approved):

src/python_testing/TC_MCORE_FS_1_3.py
```diff
@@ -0,0 +1,181 @@
...
+        logging.info("Starting TH device for DUT ecosystem")
+        self.app_process_for_dut_eco = subprocess.Popen(cmd, bufsize=0, shell=True)
+        logging.info("Started TH device for DUT ecosystem")
+        time.sleep(3)
+
+        logging.info("Commissioning from separate fabric")
+        self.device_for_dut_eco_nodeid = 1111
+        await self.TH_server_controller.CommissionOnNetwork(nodeId=self.device_for_dut_eco_nodeid, setupPinCode=passcode, filterType=ChipDeviceCtrl.DiscoveryFilterType.LONG_DISCRIMINATOR, filter=discriminator)
+        logging.info("Commissioning device for DUT ecosystem onto TH for managing")
+        return passcode
```
samadDotDev, line 90: I think this should return the chosen discriminator in addition to the passcode for the test steps to use instead of defining that again later

src/python_testing/TC_MCORE_FS_1_3.py
```diff
@@ -0,0 +1,181 @@
...
+            logging.warning("Stopping app with SIGTERM")
+            self.app_process_for_th_eco.send_signal(signal.SIGTERM.value)
+            self.app_process_for_th_eco.wait()
+
+        os.remove(self.device_for_dut_eco_kvs)
+        if self.device_for_th_eco_kvs is not None:
+            os.remove(self.device_for_th_eco_kvs)
+        super().teardown_class()
+
+    async def create_device_for_dut_ecosystem(self):
```
samadDotDev, line 68: This is also commissioning it, right? Similar to `create_and_commission_device_for_th_ecosystem`

src/python_testing/TC_MCORE_FS_1_3.py
```diff
@@ -0,0 +1,181 @@
...
+                                                          discriminator=discriminator, option=1)
+
+        setup_qrcode = "MT:-24J0-Q000KA0648G00"
+        self.wait_for_user_input(
+            prompt_msg=f"Using the DUT vendor's provided interface, commission the device using the following parameters:\n"
+            f"- discriminator: {discriminator}\n"
+            f"- passcode: {passcode}\n"
+            f"- setupPinCode: {setup_qrcode}\n"
+            f"If using FabricSync Admin, you may type:\n"
+            f">>> pairing onnetwork 111 {setup_qrcode}")
```
samadDotDev, line 154: Just to be clear, the node id here isn't the same as `device_for_dut_eco_nodeid` but more of an example of whatever random value fabric sync admin can pick independently, right?

src/python_testing/TC_MCORE_FS_1_3.py
```diff
@@ -0,0 +1,181 @@
...
+
+        asserts.assert_true(set_of_endpoints_after_adding_device.issuperset(
+            set_of_endpoints_before_adding_device), "Expected only new endpoints to be added")
+        unique_endpoints_set = set_of_endpoints_after_adding_device - set_of_endpoints_before_adding_device
+        asserts.assert_equal(len(unique_endpoints_set), 1, "Expected only one new endpoint")
+        newly_added_endpoint = list(unique_endpoints_set)[0]
+
+        th_sed_dut_unique_id = await self.read_single_attribute_check_success(cluster=Clusters.BridgedDeviceBasicInformation, attribute=Clusters.BridgedDeviceBasicInformation.Attributes.UniqueID, endpoint=newly_added_endpoint)
+        asserts.assert_true(type_matches(th_sed_dut_unique_id, str), "UniqueID should be a string")
+        asserts.assert_true(th_sed_dut_unique_id, "UniqueID should not be an empty string")
```
samadDotDev, line 167: Isn't this read supposed to fail with `UNSUPPORTED_ATTRIBUTE` instead of returning empty string value when the unique id is absent?

src/python_testing/TC_MCORE_FS_1_3.py
tehampson: I actually was just working on that while you were reviewing :)

src/python_testing/TC_MCORE_FS_1_3.py
tehampson: Yes... they originally started off different but now they are identical I will merge the two

src/python_testing/TC_MCORE_FS_1_3.py
tehampson: The server see this as a string to encode, and can encode an empty string, we as a client might get an empty string here from a misbaving device.

I will double check in a hacked test locally, but I see no harm in having this addtional check here. If we get a `UNSUPPORTED_ATTRIBUTE` that will have an exception raised on ` self.read_single_attribute_check_success` above

src/python_testing/TC_MCORE_FS_1_3.py
tehampson: I changed this to be `<desired_node_id>`

andy31415 (approved):

src/python_testing/TC_MCORE_FS_1_3.py
tehampson: Confirm a DUT could theoretically send an empty string we should say it is not valid as there is no way that is unique

samadDotDev (approved):

tpmanley (approved):

PR #34663, opened 2024-07-31
adding parallel execution to restyle-diff

Alami-Amine:
- calling function that runs container using `xargs`, with 256 Maximum number of arguments
- add option to pull new restyled image
-  added a git fetch upstream

### DIFF
--- a/scripts/helpers/restyle-diff.sh
+++ b/scripts/helpers/restyle-diff.sh
@@ -32,85 +32,47 @@ here=${0%/*}
 
 set -e
 
+MAX_ARGS=256
+pull_image=0
+
 CHIP_ROOT=$(cd "$here/../.." && pwd)
 cd "$CHIP_ROOT"
 
-docker_run() {
-    if [ -t 0 ]; then
-        exec docker run --tty "$@"
-
-    else
-        exec docker run "$@"
-
-    fi
-}
-
 restyle-paths() {
     image=restyled/restyler:edge
-    batch_size=4
-    batch=()
-
-    for path in "$@"; do
-        batch+=("$path")
-        if [[ ${#batch[@]} -eq $batch_size ]]; then
-            for p in "${batch[@]}"; do
-                (
-                    docker_run \
-                        --env LOG_LEVEL \
-                        --env LOG_DESTINATION \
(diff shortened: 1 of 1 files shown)


scripts/helpers/restyle-diff.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 #
-# Copyright (c) 2020 Project CHIP Authors
+# Copyright (c) 2024 Project CHIP Authors
```
cecille, line 4:
```suggestion
# Copyright (c) 2020-2024 Project CHIP Authors
```

scripts/helpers/restyle-diff.sh
```diff
@@ -45,33 +46,67 @@ docker_run() {
 }
 
 restyle-paths() {
-
     image=restyled/restyler:edge
+    batch_size=4
```
andy31415, line 50: How about using `$(nproc)`? or is that too much?

Do these restylers support multiple paths in some way?

andy31415: I tried locally just passing in all files and it seemed to work. The worst may be if we somehow have too many files changed (regen?) and they do not fit in max command line arguments. We should either just pass all (first try) or pass in large batches that are configurable (like 256 or 1024 files at once or something)

tcarmelveilleux: Is there a way to get command line arguments from a file? I would think so.

BTW on linux, max command line length is ~100KiB

andy31415: we can xargs it with a large count like 256 and just assume it is ok. My `getconf ARG_MAX` on my machine is about 2MB. Or even 100 files at a time will make most cases just be ok, even the codegen ones.

Alami-Amine: ok i just added xargs, makes the script much more concise...

Alami-Amine: i made it accept 256, tested it with some old branch and it isnt crashing so looks good

andy31415 (approved):

woody-apple (approved):

PR #34683, opened 2024-07-31
cleanup-ServiceArea-LocationStructureWrapper-constructors-and-setters

### LINKED ISSUE #34519: ServiceArea: constructors with many args + assignment operator not making identical copy
Requesting changes: I would want some comments to actually be addressed instead of being argued as `do not actually need to change` (or if the argument of not changing is good, I would like to make sure I review the reply). After seeing the answers, I believe we need at least:

- [ ] use builders for arguments when the argument count starts being large (usually after around 4-5 arguments ... and always if there are same-type arguments especially bools or nullable things)
- [ ] do not use operator= to mean anything else than "exact copy"

_Originally posted by @andy31415 in https://github.com/project-chip/connectedhomeip/pull/33991#pullrequestreview-2197256892_

johnfierke:
Fixes #34519

LocationStructureWrapper:

changed operator= to make an exact copy.

deleted Set function with many parameters and replaces with Set<field> functions for each individual field.

Added a parameter to SetLocationName, SetFloorNumber, SetAreaType to allow either a simple set of the field, or allow the setter to NULL the containing structure locationInfo.locationInfo (LocationDescriptorStruct) if the resulting name is the empty string and the other fields are null.  This is the form that satisfies the requirements that are current for this implementation.
(There are specification changes in progress that may change the attributes/structures/nullable requirements.)

The constructor with parameters for ALL fields remains, as the most likely use case is creating locations from robot information when all knowledge about the location is known.  This aids in maintenance of calling code as changes, additions, or subtractions of fields will cause compiler errors that force the user code to be updated to handle the changes.

The implementation of the all parameters constructor is greatly simplified - it calls the individual setter for each field.

### DIFF
--- a/src/app/clusters/service-area-server/service-area-cluster-objects.h
+++ b/src/app/clusters/service-area-server/service-area-cluster-objects.h
@@ -38,8 +38,17 @@ struct LocationStructureWrapper : public chip::app::Clusters::ServiceArea::Struc
 {
     LocationStructureWrapper()
     {
-        Set(0, 0, CharSpan(), DataModel::Nullable<int16_t>(), DataModel::Nullable<AreaTypeTag>(),
-            DataModel::Nullable<LandmarkTag>(), DataModel::Nullable<PositionTag>(), DataModel::Nullable<FloorSurfaceTag>());
+        SetLocationID(0);
+        SetMapID(DataModel::Nullable<uint8_t>());
+        SetLandmarkTag(DataModel::Nullable<LandmarkTag>());
+        SetPositionTag(DataModel::Nullable<PositionTag>());
+        SetSurfaceTag(DataModel::Nullable<FloorSurfaceTag>());
+
+        // 'true' forces the 'canonical' form specified by the cluster requirements
+        // where the LocationDescriptorStruct is null if all it's elements are/would be empty or null.
+        SetLocationName(CharSpan(mLocationNameBuffer, 0), true);
+        SetFloorNumber(DataModel::Nullable<int16_t>(), true);
+        SetAreaType(DataModel::Nullable<AreaTypeTag>(), true);
     }
 
     /**
@@ -48,7 +57,7 @@ struct LocationStructureWrapper : public chip::app::Clusters::ServiceArea::Struc
      * @param[in] aMapId The identifier of the supported map associated with this location.
      * @param[in] aLocationName A human readable name for this location (empty string if not used).
      * @param[in] aFloorNumber The floor level of this location - use negative values for below ground.
-     * @param[in] aAreaTypeTag A common namespace Area tag - indicates an association of the location with an indoor or outdoor area
+     * @param[in] aAreaType A common namespace Area tag - indicates an association of the location with an indoor or outdoor area
      * of a home.
      * @param[in] aLandmarkTag A common namespace Landmark tag - indicates an association of the location with a home landmark.
      * @param[in] aPositionTag A common namespace Position tag - indicates the position of the location with respect to the
@@ -57,16 +66,28 @@ struct LocationStructureWrapper : public chip::app::Clusters::ServiceArea::Struc
      *
      * @note Requirements regarding what combinations of fields and values are valid are not checked by this class.
      * @note If aLocationName is larger than kLocationNameMaxSize, it will be truncated.
-     * @note If aLocationName is an empty string and aFloorNumber and aAreaTypeTag are null, locationInfo will be set to null.
+     * @note If aLocationName is an empty string and aFloorNumber and aAreaType are null, locationInfo.locationinfo will be set to null.
      */
-    LocationStructureWrapper(uint32_t aLocationId, const DataModel::Nullable<uint8_t> & aMapId, const CharSpan & aLocationName,
+    LocationStructureWrapper(uint32_t aLocationId, 
+                             const DataModel::Nullable<uint8_t> & aMapId, 
+                             const CharSpan & aLocationName,
                              const DataModel::Nullable<int16_t> & aFloorNumber,
-                             const DataModel::Nullable<AreaTypeTag> & aAreaTypeTag,
+                             const DataModel::Nullable<AreaTypeTag> & aAreaType,
                              const DataModel::Nullable<LandmarkTag> & aLandmarkTag,
                              const DataModel::Nullable<PositionTag> & aPositionTag,
                              const DataModel::Nullable<FloorSurfaceTag> & aSurfaceTag)
     {
-        Set(aLocationId, aMapId, aLocationName, aFloorNumber, aAreaTypeTag, aLandmarkTag, aPositionTag, aSurfaceTag);
+        SetLocationID(aLocationId);
+        SetMapID(aMapId);
+        SetLandmarkTag(aLandmarkTag);
+        SetPositionTag(aPositionTag);
+        SetSurfaceTag(aSurfaceTag);
+
+        // 'true' forces the 'canonical' form specified by the cluster requirements
+        // where the LocationDescriptorStruct is null if all it's elements are/would be empty or null.
+        SetLocationName(aLocationName, true);
+        SetFloorNumber(aFloorNumber, true);
+        SetAreaType(aAreaType, true );
     }
 
     /**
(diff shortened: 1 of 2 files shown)


plauric (approved):

andy31415 (approved):

src/app/clusters/service-area-server/service-area-cluster-objects.h
```diff
@@ -57,16 +66,28 @@ struct LocationStructureWrapper : public chip::app::Clusters::ServiceArea::Struc
      *
      * @note Requirements regarding what combinations of fields and values are valid are not checked by this class.
      * @note If aLocationName is larger than kLocationNameMaxSize, it will be truncated.
-     * @note If aLocationName is an empty string and aFloorNumber and aAreaTypeTag are null, locationInfo will be set to null.
+     * @note If aLocationName is an empty string and aFloorNumber and aAreaType are null, locationInfo.locationinfo will be set to null.
      */
-    LocationStructureWrapper(uint32_t aLocationId, const DataModel::Nullable<uint8_t> & aMapId, const CharSpan & aLocationName,
+    LocationStructureWrapper(uint32_t aLocationId,
```
andy31415, line 71: note that to actually fix #34519 this constructor should not exist.

andy31415: It is not readable or maintainable to have methods with more than 3-4 arguments. This has 8 .. at some point we will add the 9th..this is not practical.

src/app/clusters/service-area-server/service-area-cluster-objects.h
```diff
@@ -38,8 +38,17 @@ struct LocationStructureWrapper : public chip::app::Clusters::ServiceArea::Struc
...
-        Set(0, 0, CharSpan(), DataModel::Nullable<int16_t>(), DataModel::Nullable<AreaTypeTag>(),
-            DataModel::Nullable<LandmarkTag>(), DataModel::Nullable<PositionTag>(), DataModel::Nullable<FloorSurfaceTag>());
+        SetLocationID(0);
+        SetMapID(DataModel::Nullable<uint8_t>());
+        SetLandmarkTag(DataModel::Nullable<LandmarkTag>());
+        SetPositionTag(DataModel::Nullable<PositionTag>());
+        SetSurfaceTag(DataModel::Nullable<FloorSurfaceTag>());
+
+        // 'true' forces the 'canonical' form specified by the cluster requirements
+        // where the LocationDescriptorStruct is null if all it's elements are/would be empty or null.
```
bzbarsky-apple, line 48:
```suggestion
        // where the LocationDescriptorStruct is null if all its elements are/would be empty or null.
```

src/app/clusters/service-area-server/service-area-cluster-objects.h
```diff
@@ -83,69 +104,76 @@ struct LocationStructureWrapper : public chip::app::Clusters::ServiceArea::Struc
...
+            // Create a location info structure
             locationInfo.locationInfo.SetNonNull(Structs::LocationDescriptorStruct::Type());
-
-            locationInfo.locationInfo.Value().floorNumber = aFloorNumber;
-            locationInfo.locationInfo.Value().areaType    = aAreaType;
         }
-        else
+
+        // does the locationInfo structure need to be removed and set to NULL?
+        if ((doInfoStructCleanup) &&
```
bzbarsky-apple, line 167: Why the parentheses around a bareword boolean?

In general, these if conditions all seem way over-parenthesized.

src/app/clusters/service-area-server/service-area-cluster-objects.h
```diff
@@ -38,8 +38,17 @@ struct LocationStructureWrapper : public chip::app::Clusters::ServiceArea::Struc
 {
     LocationStructureWrapper()
     {
-        Set(0, 0, CharSpan(), DataModel::Nullable<int16_t>(), DataModel::Nullable<AreaTypeTag>(),
-            DataModel::Nullable<LandmarkTag>(), DataModel::Nullable<PositionTag>(), DataModel::Nullable<FloorSurfaceTag>());
+        SetLocationID(0);
+        SetMapID(DataModel::Nullable<uint8_t>());
+        SetLandmarkTag(DataModel::Nullable<LandmarkTag>());
```
hicklin, line 43: Note from @bzbarsky-apple is a separate PR: "While you are here, just using NullNullable here would be shorter, but maybe we want to make it clear which things we are leaving null?"

src/app/clusters/service-area-server/service-area-cluster-objects.h
johnfierke: What do you suggest for the constructor?

src/app/clusters/service-area-server/service-area-cluster-objects.h
johnfierke: Yes, my coding style is to use lot's of parenthesis in logic statements.  At best it avoids errors due to the compiler executing in the wrong order, it clarifies the intent, and yes, it does clutter things up a bit.

src/app/clusters/service-area-server/service-area-cluster-objects.h
tcarmelveilleux: Please use `NullNullable` everywhere a default null is expected.

src/app/clusters/service-area-server/service-area-cluster-objects.h
tcarmelveilleux: Builder-style is very easy to use, with a good default constructor.

Then you can have 

```
LocationStructureWrapper locationStructure;

locationStructure.SetXxx(xxx).SetYyy(yyy).
```

This can also be used in-place:

```
DoSomething(LocationStructureWrapper{}.SetXxx().SetYyy(yyy));
```

Setters must then simply return `LocationStructureWrapper &` instead of `void`.

src/app/clusters/service-area-server/service-area-cluster-objects.h
```diff
@@ -83,69 +104,76 @@ struct LocationStructureWrapper : public chip::app::Clusters::ServiceArea::Struc
      * values are deep copied.
      * @param[in] aOther The location object to copy.
      *
-     * @note If the locationName is empty string and aFloorNumber and aAreaTypeTag are null, locationInfo will be set to null.
+     * @note This will do an EXACT copy of the other object. 
      */
     LocationStructureWrapper & operator=(const LocationStructureWrapper & aOther)
```
tcarmelveilleux, line 109: Consider having an `operator==` as well that calls `IsEqual`

johnfierke: Sadly, because it took so long to get comments on the ServiceArea Cluster specifications, the implementation/code branch is running significantly behind the latest updates to the spec.

With the current xml definitions, IsEqual checks the contents of the underlying LocationStruct, which is not the same as true equality which should be what the operator== indicates. And in the cluster implementation, IsEqual is what is needed, not operator==.  So for the moment it's better not to have it, as any use of it would almost certainly be wrong.

Again, the problem is that in terms of content
locationInfo.locationInfo { null } has the same content as locationInfo.locationInfo.Value()   { "", null, null }

Once all the xml and code is updated to reflect the very latest approved specification changes, all of the structure elements in the cluster will become non-nullable.

So there won't be any need in IsEquals to check for different topologies that are equal in content but not in form.
The special setter arguments will not be needed to select how to deal with the nullable structure.
An operator== based on IsEquals with the appropriate compare enum will be correct.
IsEquals is still necessary as there are a number of use cases where the code needs to compare all of the contents EXCEPT for the locationId, but that's clear based on the compare enum.

tcarmelveilleux: OK, the point I am making is we need to be able to compare locations before/after, so that matter dirty attribute reporting can be done correctly at the right times on list changes.

johnfierke: For now IsEquals with aConfig set to 0 is the appropriate function to use.

After the updated spec changes are pulled in and there are no more nullable structure elements, the operator== based on IsEquals can be added.

src/app/clusters/service-area-server/service-area-cluster-objects.h
johnfierke: updated

src/app/clusters/service-area-server/service-area-cluster-objects.h
johnfierke: updated

hicklin:
There has been some confusion here. We were meant to halt review on this PR as there where other changes that impacted this PR that made sense to go before this PR. Please see PR #34814 instead. I have made sure to consider all the comments made in this PR so it should hopefully be an quick review.

PR #34708, opened 2024-08-01
Define an `ActionReturnStatus` for DataModel::Provider action returns

andy31415:
CHIP_ERROR is generally insufficient to provide the full IM return codes (generally a StatusIB). This is because a StatusIB can also encode `success` with a code (i.e. a clusterstatus) which is not ideally encodable into a CHIP_ERROR. I tried that in #34597 , however the (reasonable) review feedback was that it is very awkward and error prone to have a "success" code that is not CHIP_NO_ERROR.

As a result I created a ActionStatusReturn class that generally encodes all possible statuses, supporting also the full clusterstatus (which has success). 

I kept both CHIP_ERROR and ClusterStatus because for logging purposes, CHIP_ERROR has a better format on some systems (encodes originating location and a distinct code, which is much more helpful than a Status::Failure). In addition, we had to preserve the ability of  communicating "out of space, chunking can happen" which is currently done through CHIP_ERROR.


src/app/data-model-provider/ActionReturnStatus.h
```diff
@@ -0,0 +1,92 @@
...
+
+/// An ActionReturnStatus encodes the result of a read/write/invoke.
+///
+/// Generally such actions result in a StatusIB in the interaction model,
+/// which is a code (InteractionModel::Status) and may have an associated
+/// cluster-specific code.
+///
+/// However some actions specifically may return additional errors for
+/// chunking (i.e. out of space, retry or send existing partial data) hence
+/// the exitance of this class:
```
bzbarsky-apple, line 37:
```suggestion
/// However some actions specifically may return additional information for
/// chunking, hence the existence of this class:
```

src/app/data-model-provider/ActionReturnStatus.h
```diff
@@ -0,0 +1,92 @@
...
+///
+///   - encapsulates a ClusterStatusCode for an actual action result
+///   - encapsulates a underlying CHIP_ERROR for reporting purposes
+///     and to check for out-of-space specific errors
```
bzbarsky-apple, line 41:
```suggestion
///   - has a way to check for "chunking needed" status.
```

src/app/data-model-provider/ActionReturnStatus.h
```diff
@@ -0,0 +1,92 @@
...
+
+    /// Considers if the underlying error is an error or not (CHIP_NO_ERROR is the only non-erro)
+    /// or if the underlying statuscode is not an error (success and cluster specific successes
+    /// are not an error).
+    bool IsError() const { return !IsSuccess(); }
+
+    /// Checks if the underlying error is an out of space condition (i.e. something that
+    /// chunking can handle by sending partial list data).
+    ///
+    /// Generally this is when the return is based on CHIP_ERROR_NO_MEMORY or CHIP_ERROR_BUFFER_TOO_SMALL
```
bzbarsky-apple, line 75: I don't think we should document these specific status codes, because I still think we should move toward not representing "needs chunking" via magic error codes at any point other than the immediate TLVWriter API boundary, in followups to this PR.

I would consider naming this OutOfSpaceToEncodeResponse(), by the way.

src/app/data-model-provider/ActionReturnStatus.h
```diff
@@ -0,0 +1,92 @@
...
+    bool IsOutOfSpaceError() const;
+
+    // NOTE: operator== will assume statues the same between CHIP_GLOBAL_IM_ERROR and a raw cluster status
```
bzbarsky-apple, line 78:
```suggestion
    // NOTE: operator== will treat a CHIP_GLOBAL_IM_ERROR and a raw cluster status as equal if the statuses match,
```

src/app/data-model-provider/Provider.h
```diff
@@ -66,18 +68,12 @@ class Provider : public ProviderMetadataTree
     ///         > Else if reading from the attribute in the path requires a privilege that is not
     ///           granted to access the cluster in the path, then the path SHALL be discarded.
     ///
-    /// Return codes:
-    ///   CHIP_ERROR_NO_MEMORY or CHIP_ERROR_BUFFER_TOO_SMALL:
+    /// Return value notes:
+    ///   ActionReturnStatus::IsOutOfSpaceError(i.e. CHIP_ERROR_NO_MEMORY or CHIP_ERROR_BUFFER_TOO_SMALL):
```
bzbarsky-apple, line 72: Again, we should not be claiming specific error codes here, ideally.  Or at least remove that in a followup when we fix the representation.

src/app/data-model-provider/Provider.h
```diff
@@ -90,18 +86,17 @@ class Provider : public ProviderMetadataTree
...
+    ///      - `UnsupportedAccess` for ACL failures
+    ///      - `NeedsTimedInteraction` for writes that are not timed however are required to be so
+    virtual ActionReturnStatus WriteAttribute(const WriteAttributeRequest & request, AttributeValueDecoder & decoder) = 0;
 
     /// `handler` is used to send back the reply.
-    ///    - returning a value other than CHIP_NO_ERROR implies an error reply (error and data are mutually exclusive)
-    virtual CHIP_ERROR Invoke(const InvokeRequest & request, chip::TLV::TLVReader & input_arguments, CommandHandler * handler) = 0;
+    ///    - returning a value other than Success implies an error reply (error and data are mutually exclusive)
+    virtual ActionReturnStatus Invoke(const InvokeRequest & request, chip::TLV::TLVReader & input_arguments,
+                                      CommandHandler * handler) = 0;
```
bzbarsky-apple, line 99: Still need to document what it means to return an error here, right?  Does that indicate an action-level failure such that no replies were or will be encoded via the CommandHandler at all?  This is pretty unclear after these changes.

src/app/data-model-provider/ActionReturnStatus.cpp
```diff
@@ -0,0 +1,196 @@
...
+        {
+            return ClusterStatusCode(Status::Success);
+        }
+
+        if (err->IsPart(ChipError::SdkPart::kIMGlobalStatus))
+        {
+            return ClusterStatusCode(static_cast<Status>(err->GetSdkCode()));
+        }
+
+        return ClusterStatusCode(Status::Failure);
```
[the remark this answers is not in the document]
andy31415: I agree. I just did not want to pull StatusIB since the order of creation seems to be `StatusCode` returned and `StatusIB created out of it` and I am replacing StatusCode

src/app/reporting/Read-DataModel.cpp
```diff
@@ -97,11 +93,13 @@ CHIP_ERROR RetrieveClusterData(DataModel::Provider * dataModel, const Access::Su
     // Out of space errors may be chunked data, reporting those cases would be very confusing
     // as they are not fully errors. Report only others (which presumably are not recoverable
     // and will be sent to the client as well).
-    if (!IsOutOfSpaceError(err))
+    if (!status.IsOutOfSpaceError())
     {
-        ChipLogError(DataManagement, "Failed to read attribute: %" CHIP_ERROR_FORMAT, err.Format());
+        StringBuilder<128> buffer;
```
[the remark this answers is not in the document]
andy31415: I previously had a `LogError(const char *prefix)` however then I needed pretty formatting for string builders in asserts.  I can do both, but there will be duplication at least partially.

src/app/data-model-provider/ActionReturnStatus.h
andy31415: Renamed.

src/app/data-model-provider/Provider.h
andy31415: Removed details.

src/app/data-model-provider/Provider.h
```diff
@@ -90,18 +86,17 @@ class Provider : public ProviderMetadataTree
...
-    /// Return codes
-    ///   CHIP_IM_GLOBAL_STATUS(code):
-    ///       - error codes that are translatable to specific IM codes
-    ///       - in particular, the following codes are interesting/expected
-    ///         - `UnsupportedWrite` for attempts to write read-only data
-    ///         - `UnsupportedAccess` for ACL failures
-    ///         - `NeedsTimedInteraction` for writes that are not timed however are required to be so
-    virtual CHIP_ERROR WriteAttribute(const WriteAttributeRequest & request, AttributeValueDecoder & decoder) = 0;
+    /// Return notes:
+    ///    - The following codes are interesting/expected to be possible:
```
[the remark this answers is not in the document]
andy31415: There were called out before. Removed since I believe we called them out as examples for CHIP_IM_GLOBAL_ERROR but now that we have real statuses, we should not care about this as much.

src/app/data-model-provider/Provider.h
andy31415: Added documentation here to make it clear that returning anything except pure success (i.e. not CHIP_ERROR or Status::Success) will end up sending a StatusIB without any data.

This is slightly awkward because it means ClusterSpecificStatus is success and prevents data. But it seems to be what the spec demands.

src/app/data-model-provider/ActionReturnStatus.cpp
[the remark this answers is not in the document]
andy31415: Would it not be better to get a ClusterStatus from CHIP_ERROR instead? that fixes my code and allows for re-use. StatusIB is way down the chain (people don't return a statusib, they return a status or clusterstatus.

src/app/reporting/Read-DataModel.cpp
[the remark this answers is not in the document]
andy31415: Updated to support a c_str-like construct that is not thread safe (same like CHIP_ERROR does ...). However this wastes 32 bytes of BSS forever ... I am unsure if we want to go this route. maybe we should centralize the buffer for formatting things and share between CHIP_ERROR and anything else that wants to statically use some string buffer to return strings.

Does this also mean that I cannot CHIP_ERROR.Format() two separate errors in the same error log? That is not obvious....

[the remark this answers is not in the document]
bzbarsky-apple: Just to be clear, I was saying the stringbuilder is fine for now but we should figure out something better later....

> Does this also mean that I cannot CHIP_ERROR.Format() two separate errors in the same error log?

Yep.

src/app/data-model-provider/ActionReturnStatus.cpp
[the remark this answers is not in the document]
bzbarsky-apple: I think we are in violent agreement here.  The code should live in ClusterStatusCode, and we should go through that to init a StatusIB from CHIP_ERROR.

src/app/data-model-provider/Provider.h
```diff
@@ -89,19 +85,21 @@ class Provider : public ProviderMetadataTree
...
+    virtual ActionReturnStatus WriteAttribute(const WriteAttributeRequest & request, AttributeValueDecoder & decoder) = 0;
 
     /// `handler` is used to send back the reply.
-    ///    - returning a value other than CHIP_NO_ERROR implies an error reply (error and data are mutually exclusive)
-    virtual CHIP_ERROR Invoke(const InvokeRequest & request, chip::TLV::TLVReader & input_arguments, CommandHandler * handler) = 0;
+    ///    - returning a value other than Success implies an error reply (error and data are mutually exclusive)
+    ///
+    /// Returning anything other than CHIP_NO_ERROR or Status::Success (i.e. success without a return code)
+    /// means that the invoke will be considered to be returning the given status WITHOUT any data (any data
+    /// that was sent via Handler is to be rolled back/discarded).
```
bzbarsky-apple, line 95:
```suggestion
    /// that was sent via CommandHandler is to be rolled back/discarded).
```

src/app/data-model-provider/Provider.h
```diff
@@ -89,19 +85,21 @@ class Provider : public ProviderMetadataTree
...
-    virtual CHIP_ERROR WriteAttribute(const WriteAttributeRequest & request, AttributeValueDecoder & decoder) = 0;
+    virtual ActionReturnStatus WriteAttribute(const WriteAttributeRequest & request, AttributeValueDecoder & decoder) = 0;
 
     /// `handler` is used to send back the reply.
-    ///    - returning a value other than CHIP_NO_ERROR implies an error reply (error and data are mutually exclusive)
-    virtual CHIP_ERROR Invoke(const InvokeRequest & request, chip::TLV::TLVReader & input_arguments, CommandHandler * handler) = 0;
+    ///    - returning a value other than Success implies an error reply (error and data are mutually exclusive)
+    ///
+    /// Returning anything other than CHIP_NO_ERROR or Status::Success (i.e. success without a return code)
+    /// means that the invoke will be considered to be returning the given status WITHOUT any data (any data
```
bzbarsky-apple, line 94:
```suggestion
    /// means that the invoke will be considered to be returning the given path-specific status WITHOUT any data (any data
```

bzbarsky-apple (approved):

src/app/data-model-provider/tests/TestActionReturnStatus.cpp
```diff
@@ -0,0 +1,78 @@
...
+#include <lib/support/CodeUtils.h>
+
+#include <pw_unit_test/framework.h>
+
+using chip::app::DataModel::ActionReturnStatus;
+using chip::Protocols::InteractionModel::ClusterStatusCode;
+using chip::Protocols::InteractionModel::Status;
+
+TEST(TestActionReturnStatus, TestEquality)
+{
```
harimau-qirex, line 33: What should the equivalence relation be for CHIP_NO_ERROR and ClusterStatusCode::ClusterSpecificSuccess?

andy31415: Added unit tests. This also got me to figure out equivalence between pure Status::Success and CHIP_NO_ERROR (which seem like they should be equivalent) so I fixed that. Thank you for the review!

src/app/data-model-provider/tests/TestActionReturnStatus.cpp
```diff
@@ -0,0 +1,90 @@
...
+    ASSERT_EQ(ActionReturnStatus(Status::Success), Status::Success);
+    ASSERT_EQ(ActionReturnStatus(CHIP_NO_ERROR), Status::Success);
+    ASSERT_EQ(ActionReturnStatus(Status::Success), CHIP_NO_ERROR);
+
+    // status specific success is has more data, so there is no equality (i.e. an action return
+    // with specific success codes has more data than a simple success)
+    ASSERT_NE(ActionReturnStatus(ClusterStatusCode::ClusterSpecificSuccess(123)), CHIP_NO_ERROR);
+    ASSERT_NE(ActionReturnStatus(ClusterStatusCode::ClusterSpecificSuccess(123)), Status::Success);
+    ASSERT_NE(ActionReturnStatus(CHIP_NO_ERROR), ClusterStatusCode::ClusterSpecificSuccess(123));
+    ASSERT_NE(ActionReturnStatus(Status::Success), ClusterStatusCode::ClusterSpecificSuccess(123));
```
harimau-qirex, line 56: Could you add a case to compare two cluster-specific successes with different success codes?

harimau-qirex (approved):

src/app/data-model-provider/tests/TestActionReturnStatus.cpp
andy31415: Added a lot more tests ... still find it somewhat hard to pick all combinations. Hopefully I have solid coverage.

tcarmelveilleux (approved):

PR #34716, opened 2024-08-01
Draft enabling wifipaf on sample apps

gladystonfranca:
Support for wifipaf is not being  enabled on the  example apps  on the SDK Container.
SDK container is generated by using Dockerfile: 
https://github.com/project-chip/connectedhomeip/blob/master/integrations/docker/images/chip-cert-bins/Dockerfile

This PR is related / should be understood in the context of this Commit:

Add --wifipaf commission in chip-tool and apps of Linux platform (#33977). 30 Jul 2024 at 00:21
SHA: **085f57f3fc7827b45830c5976ddac16895c2fd3b**

The initial changes done here in this PR are not supposed to the  complete or  even the right solution for this issue but the entry point to get  help from who knew the right places that must be changed in order to get wifipaf available on the sample apps (SDK container)  used by Test Harness execution.

### DIFF
--- a/src/platform/device.gni
+++ b/src/platform/device.gni
@@ -25,7 +25,7 @@ declare_args() {
   chip_fake_platform = false
 
   # Include wifi-paf to commission the device or not
-  chip_device_config_enable_wifipaf = (current_os == "linux")
+  chip_device_config_enable_wifipaf = current_os == "linux"
 }
 
 if (chip_device_platform == "auto") {
--- a/src/platform/device.gni
+++ b/src/platform/device.gni
@@ -25,7 +25,7 @@ declare_args() {
   chip_fake_platform = false
 
   # Include wifi-paf to commission the device or not
-  chip_device_config_enable_wifipaf = false
+  chip_device_config_enable_wifipaf = true
 }
 
 if (chip_device_platform == "auto") {
--- a/src/platform/device.gni
+++ b/src/platform/device.gni
@@ -25,7 +25,7 @@ declare_args() {
   chip_fake_platform = false
 
   # Include wifi-paf to commission the device or not
-  chip_device_config_enable_wifipaf = true
+  chip_device_config_enable_wifipaf = (current_os == "linux")
 }
 
 if (chip_device_platform == "auto") {
--- a/examples/platform/linux/BUILD.gn
+++ b/examples/platform/linux/BUILD.gn
@@ -26,6 +26,7 @@ declare_args() {
   chip_enable_energy_reporting_trigger = false
   chip_enable_water_heater_management_trigger = false
   chip_enable_device_energy_management_trigger = false
+  chip_device_config_enable_wifipaf = true
 }
 
 config("app-main-config") {
@@ -133,6 +134,7 @@ source_set("app-main") {
     "CHIP_DEVICE_CONFIG_ENABLE_ENERGY_REPORTING_TRIGGER=${chip_enable_energy_reporting_trigger}",
     "CHIP_DEVICE_CONFIG_ENABLE_WATER_HEATER_MANAGEMENT_TRIGGER=${chip_enable_water_heater_management_trigger}",
     "CHIP_DEVICE_CONFIG_ENABLE_DEVICE_ENERGY_MANAGEMENT_TRIGGER=${chip_enable_device_energy_management_trigger}",
+    "CHIP_DEVICE_CONFIG_ENABLE_WIFIPAF=${chip_device_config_enable_wifipaf}",
   ]
 
   public_configs = [ ":app-main-config" ]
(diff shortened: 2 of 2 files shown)


src/platform/device.gni
```diff
@@ -25,7 +25,7 @@ declare_args() {
   chip_fake_platform = false
 
   # Include wifi-paf to commission the device or not
-  chip_device_config_enable_wifipaf = false
+  chip_device_config_enable_wifipaf = true
```
andy31415, line 28: Do we have a reason to have wifipaf as a default?
Otherwise we should just change what the TH builds, like I see https://github.com/project-chip/connectedhomeip/blob/master/integrations/docker/images/chip-cert-bins/Dockerfile#L177

You could add a `wifi-paf` variant to build things like `linux-x64-all-clusters-ipv6only-wifi-paf`.

To get that variant working you add the variant in https://github.com/project-chip/connectedhomeip/blob/master/scripts/build/build/targets.py#L187 and make https://github.com/project-chip/connectedhomeip/blob/master/scripts/build/builders/host.py#L331 support some form of flag that makes the gn argument take some value.

gladystonfranca: @andy31415 I don't have answer for that if it should be default or not. I just realized  support was not there. Thank you for your quick answer Andrei. I will  raise up this question and get back  here.

raju-apple: @andy31415 we are going to use the chip-cert-bins in the TH image for testing , also the way the test plan is written , it uses the same sample app for validating the commissioning flow for BLE , Wifi and on network , so I guess we would have to enable it by default ?
Let me start a conversation with other folks in slack as well

andy31415, line 28:
```suggestion
  chip_device_config_enable_wifipaf = (current_os == "linux")
```

src/platform/device.gni
```diff
@@ -25,7 +25,7 @@ declare_args() {
   chip_fake_platform = false
 
   # Include wifi-paf to commission the device or not
-  chip_device_config_enable_wifipaf = false
+  chip_device_config_enable_wifipaf = (current_os == "linux")
```
andy31415, line 28:
```suggestion
  chip_device_config_enable_wifipaf = current_os == "linux"
```

andy31415 (changes requested):
Marking as request changes until we get feedback from people familiar with PAF (maybe @ksperling-apple ?).

The question is if this is sufficient to just get PAF support on most linuxes. Do most linuxes support PAF or do we need special HW.

jczhang777:
I modified the src/platform/device.gni file. I changed chip_device_config_enable_wifipaf = false to chip_device_config_enable_wifipaf = true, and then compiled the chip tool. During the compilation, I encountered an error. The error log is as follows:

ERROR at //third_party/connectedhomeip/src/include/platform/ConnectivityManager.h:39:11: Include not allowed.
#include <transport/raw/WiFiPAF.h>
          ^----------------------
  //third_party/connectedhomeip/src/transport/raw:raw

Is there an issue with the way I made the modification?

andy31415:
> I modified the src/platform/device.gni file. I changed chip_device_config_enable_wifipaf = false to chip_device_config_enable_wifipaf = true, and then compiled the chip tool. During the compilation, I encountered an error. The error log is as follows:
> 
> ERROR at //third_party/connectedhomeip/src/include/platform/ConnectivityManager.h:39:11: Include not allowed. #include <transport/raw/WiFiPAF.h> ^---------------------- It is not in any dependency of //third_party/connectedhomeip/src/platform:platform The include file is in the target(s): //third_party/connectedhomeip/src/transport/raw:raw which should somehow be reachable.
> 
> Is there an issue with the way I made the modification?

The fix does not seem to work: there is a dependency loop here: tests/raw depends on platform and vice-versa. This is more complex to fix.

woody-apple (changes requested):
This hasn't made it into the test harness yet, and has missed enablement. Holding.

src/platform/device.gni
ksperling-apple: Given that you also need a patched wpa_supplicant version to actually have working PAF support I don't think it's helpful to enable this by default. Is the TH not able to pass build arguments to the binaries it builds?

ksperling-apple: (I also think in general it's unhelpful to treat `current_os == "linux"` as "you're building for development and want to enable every option under the sun". I've been thinking about adding some option like `matter_enable_recommended` that can be defaulted to true and then used as the default choice for other options, so that in scenarios where I want to explicitly selected what gets turned on I can just set that one option to false, without having to see what other RPC or tracing or feature xyz options have been added recently and are now defaulting to True because I happen to build on Linux.)

PR #34720, opened 2024-08-01
Fix TCP test script naming

### LINKED ISSUE #34666: [1.4][TCP_Test.py] In Python script Test function has to be updated with Test case ID's.
### Feature Area

Other

**Description:**

**TCP_Test.py** Python file containing the python script for TCP Test cases. Test functions are named with Test case name this needs to be updated with Test case ID's .

**Test plan References:**
- https://github.com/CHIP-Specifications/chip-test-plans/blob/master/src/securechannel.adoc#tcp-test-cases

image

**Python Script References:**
- https://github.com/project-chip/connectedhomeip/blob/master/src/python_testing/TCP_Tests.py

image

**Note:**
**Raised Test plan issue to update the test case IDs, as the current IDs are duplicating existing test cases**. 

https://github.com/CHIP-Specifications/chip-test-plans/issues/4371

pidarped:
Change from test names to test case Ids.

Fixes #34666

### DIFF
--- a/src/python_testing/TCP_Tests.py
+++ b/src/python_testing/TCP_Tests.py
@@ -32,8 +32,9 @@ from mobly import asserts
 
 class TCP_Tests(MatterBaseTest):
 
+    # TCP Connection Establishment
     @async_test_body
-    async def test_TCPConnectionEstablishment(self):
+    async def test_TC_SC_8_1(self):
 
         try:
             device = await self.default_controller.GetConnectedDevice(nodeid=self.dut_node_id, allowPASE=False, timeoutMs=1000,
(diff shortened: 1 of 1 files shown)


src/python_testing/TCP_Tests.py
```diff
@@ -32,8 +32,9 @@
 
 class TCP_Tests(MatterBaseTest):
 
+    # TCP Connection Establishment
     @async_test_body
-    async def test_TCPConnectionEstablishment(self):
+    async def test_TC_SC_8_1(self):
```
andy31415, line 37: Entire file should then be named TC_SC_8 at least?

pidarped: I wasn't sure about the naming conventions when I first wrote the script. There are several scripts in `src/python_testing/` with test cases named descriptively, rather than as Ids. While the Id for the test_case helps tally with the test plan, I have added a comment before each test case to describe its purpose. 
From that perspective, I would think leaving the name for the file in its descriptive form helps to easily identify which script contains the TCP tests. WDYT?

andy31415: We currently seem to have most tests named with `TC_SC...` (which I don't particularely like ... because I would find something like `Occupancy` better than OCC) and a few tests that are python tests without actual integration tests ... was wondering if that overlaps.

I guess either is fine. My tendency would have been `when in rome do what romans do` and most romans currently do `TC_....py`.

andy31415 (approved):

src/python_testing/TCP_Tests.py
andy31415: I think the question is what do we want to identify: are these TCP tests or are these `TC_SC` tests. From an external perspective, I find it very hard to say what `SC` stands for (guessing now secure connection ... but yuck). So I like your naming but caution on us being inconsistent.

cecille (approved):

aronrosenberg (approved):

PR #34735, opened 2024-08-02
[OVENOPSTATE] Add OvenOpState 2.6 test case from base

rbultman:
This PR:
- Adds the 2.6 test case to OvenOpState from the base OpState test
- Changes the existing 2.6 base test case to account for CountdownTime being Null

Note:  This requires a modification to the base test plan.  
- An issue has been filed for this: https://github.com/CHIP-Specifications/chip-test-plans/issues/4392
- A PR to fix the TP has been created: https://github.com/CHIP-Specifications/chip-test-plans/pull/4397


src/python_testing/TC_OVENOPSTATE_2_6.py
```diff
@@ -0,0 +1,62 @@
...
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+# See https://github.com/project-chip/connectedhomeip/blob/master/docs/testing/python.md#defining-the-ci-test-arguments
+# for details about the block below.
+#
+# === BEGIN CI TEST ARGUMENTS ===
+# test-runner-runs: run1
+# test-runner-run/run1/app: ${ALL_CLUSTERS_APP}
```
tcarmelveilleux, line 23: Does all cluster actually have oven?

src/python_testing/TC_OpstateCommon.py
```diff
@@ -1267,67 +1272,83 @@ async def TEST_TC_OPSTATE_BASE_2_6(self, endpoint=1):
...
             self.step(5)
-            wait_count = 0
-            while (attr_value != cluster.Enums.OperationalStateEnum.kStopped) and (wait_count < 20):
-                time.sleep(1)
-                wait_count = wait_count + 1
-                attr_value = await self.read_expect_success(
-                    endpoint=endpoint,
-                    attribute=attributes.OperationalState)
+            logging.info('Test will now collect data for 30 seconds')
+            time.sleep(30)
```
tcarmelveilleux, line 1288: In CI, this should be 10 seconds and 2 reports, or something like this, to reduce time.

src/python_testing/TC_OpstateCommon.py
```diff
@@ -1267,67 +1272,83 @@ async def TEST_TC_OPSTATE_BASE_2_6(self, endpoint=1):
...
         else:
             self.skip_step(5)
 
-        sub_handler.reset()
         self.step(6)
+        countdownTime = await self.read_expect_success(endpoint=endpoint, attribute=attributes.CountdownTime)
+        if countdownTime is not NullValue:
+            attr_value = await self.read_expect_success(
+                endpoint=endpoint,
+                attribute=attributes.OperationalState)
```
tcarmelveilleux, line 1302: Suggest pre-reading this to un-nest the if, so that you have:

`if countdownTime is not NullValue and attr_value == cluster.Enums.OperationalStateEnum.kRunning:` in one single if.

src/python_testing/TC_OVENOPSTATE_2_6.py
rbultman: Yes

src/python_testing/TC_OpstateCommon.py
rbultman: I had it that way originally, and it makes the code nicer, but it made writing the test steps harder.

PR #34746, opened 2024-08-02
Use explicit storage for ActionReturnStatus::c_str

andy31415:
Allocating some global storage for this is both not thread-safe and wastes RAM.

Use the stack for the small amount of storage (relatively ... same as 4 integers) that a return status string needs.

### Changes

- use separate storage for c_str and pass it in
- fix a "buffer reset" bug that existed in the code on buffer re-use
- Add unit tests

### DIFF
--- a/src/app/data-model-provider/ActionReturnStatus.h
+++ b/src/app/data-model-provider/ActionReturnStatus.h
@@ -48,7 +48,13 @@ public:
     struct StringStorage
     {
         // Generally size should be sufficient.
-        // len("Status<123>, Code 255") == 21 (and then 22 for null terminator. We have slack.)
+        // The longest status code from StatusCodeList is NO_UPSTREAM_SUBSCRIPTION(197)
+        // so we need space for one of:
+        //    "NO_UPSTREAM_SUBSCRIPTION(197)\0" = 30   // explicit verbose status code
+        //    "FAILURE(1), Code 255\0")                // cluster failure, verbose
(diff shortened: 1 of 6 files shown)


bzbarsky-apple (approved):
Much nicer, thank you!

src/app/data-model-provider/ActionReturnStatus.h
```diff
@@ -81,12 +96,8 @@ class ActionReturnStatus
...
     /// Get the formatted string of this status.
     ///
-    /// NOTE: this is NOT thread safe in the general case, however the safety guarantees
-    ///       are similar to chip::ErrorStr which also assumes a static buffer.
-    ///
-    /// Use this in the chip main event loop (and since that is a single thread,
-    /// there should be no races)
-    const char * c_str() const;
+    /// May use `storage` for storying the actual underlying character string.
+    const char * c_str(StringStorage & storage) const;
```
cecille, line 100: This doesn't match the normal c_str function api now, which is a bit confusing. Would it make sense to just re-name this?

note - especially since this is used in apparently only one place, I don't REALLY care too much, just figured I'd ask.

cecille (approved):

src/app/data-model-provider/ActionReturnStatus.h
andy31415: I was unsure ... it was an actual c_str() before however that had global RAM wastage implications. If you do not care as much, I would take the lazy approach and let it merge.

If you do have a different name proposal, I do not care either so willing to update it to something else.

PR #34748, opened 2024-08-02
Fix BridgedDeviceInformationCluster wildcard read on fabric-bridge-app

tehampson:
DeviceManager::GetDevice() expects the endpoint not the embr index value to properly look up the device.

This is how we GetDevice() when forwarding a command from fabric-bridge to fabric-admin, demonstrating that this is the right way it was supposed to be done
https://github.com/project-chip/connectedhomeip/blob/master/examples/fabric-bridge-app/linux/main.cpp#L143


saurabhst (approved):

samadDotDev (approved):

examples/fabric-bridge-app/fabric-bridge-common/src/ZCLCallbacks.cpp
```diff
@@ -34,10 +34,9 @@ Protocols::InteractionModel::Status emberAfExternalAttributeReadCallback(Endpoin
                                                                          const EmberAfAttributeMetadata * attributeMetadata,
                                                                          uint8_t * buffer, uint16_t maxReadLength)
 {
-    uint16_t endpointIndex  = emberAfGetDynamicIndexFromEndpoint(endpoint);
     AttributeId attributeId = attributeMetadata->attributeId;
 
-    Device * dev = DeviceMgr().GetDevice(endpointIndex);
+    Device * dev = DeviceMgr().GetDevice(endpoint);
```
cecille, line 39: Totally unrelated to this PR, but the SDK already has a device manager class, and this one doesn't look to be namespaced. You might want to namespace this one or re-name it (BridgedDeviceManager?)

cecille (approved):

examples/fabric-bridge-app/fabric-bridge-common/src/ZCLCallbacks.cpp
tehampson: I will address this in a follow up

tehampson: https://github.com/project-chip/connectedhomeip/pull/34776

yufengwangca (approved):

PR #34752, opened 2024-08-02
Remove code duplication for CHIP_ERROR to ClusterCode conversion.

andy31415:
We have a complete implementation of this in ClusterStatusCode. Remove the logic in StatusIB and use the ClusterStatusCode one instead.

Generally StatusIB is just a ClusterStatusCode ontainer, however due to direct TLV reads, it splits out its contents. I only updated the constructor to use common code.

### DIFF
--- a/src/app/MessageDef/StatusIB.cpp
+++ b/src/app/MessageDef/StatusIB.cpp
@@ -24,6 +24,7 @@
 #include "StatusIB.h"
 
 #include "MessageDefHelper.h"
+#include "protocols/interaction_model/StatusCode.h"
 
 #include <inttypes.h>
 #include <stdarg.h>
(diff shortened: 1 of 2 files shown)


woody-apple (approved):

src/app/MessageDef/StatusIB.cpp
```diff
@@ -151,27 +152,10 @@ CHIP_ERROR StatusIB::ToChipError() const
...
-        mStatus = Status::Success;
-        return;
-    }
-
-    if (aError.IsPart(ChipError::SdkPart::kIMGlobalStatus))
-    {
-        mStatus = static_cast<Status>(aError.GetSdkCode());
-        return;
-    }
+    ClusterStatusCode statusCode(aError);
```
bzbarsky-apple, line 155: Why not just have this constructor delegate to the `ClusterStatusCode` constructor in the header?

As in, this whole constructor definition in the header should be:

```
StatusIB(CHIP_ERROR aError) : StatusIB(ClusterStatusCode(aError)) {}
```

andy31415: I tried to make it a member, but when we decode TLV, we decode directly into values (and those may not (yet?) be valid as a combination of status and code). Such a change has probably slightly more ripple effects.

andy31415: Updated code to place the definition in the header and re-use the clusterstatuscode constructor.

cecille (approved):

bzbarsky-apple (approved):

PR #34753, opened 2024-08-02
Make use of DataModel::Provider in writes

andy31415:
This enables ::Write support using the datamodel interface when the data model interface is in use.

### Changes

- redirect writes to ::Write if datamodel interface is enabled
- fix cast bug that lost listOP in the codegen provider write (path was missing the `data` element and would fail writing lists)
- added explicit variants to host builds that controls what data model is used (check/enabled/disabled) for easier testing and enabling
- Writing may mark paths as dirty, which requires the data model context. starting passing in a data model context:
   - added context Startup and Shutdown in InteractionModelEngine
   - made InteractionModelEngine set the current exchange as a scoped context when processing a message via `OnMessageReceived`


andy31415:
Closing ... this one has too many commits.

Issue #33735, opened 2024-06-04
Adjust the order of checks in the access control algorithm


robszewczyk:
Per spec change https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/9024 we should modify the algorithm for access control checks to return UNSUPPORTED_ACCESS  (rather than UNSUPPORTED_ENDPOINT or UNSUPPORTED_CLUSTER) in cases where there is no endpoint/cluster defined and the certificate subject appears to confer no privilege.

bzbarsky-apple:
So basically, we should do ACL checks before existence checks, not after.

Apollon77:
When this would land also ACE 2.2 testcase and test script needs to be adjusted because it checks for the old behavior

bzbarsky-apple:
Per Slack discussion right now: what the spec says at the moment is not implementable. Spec needs to get fixed to say something sane, then we need to actually implement it.  In particular, you can't tell what privilege is needed for things that don't exist, and the spec says to check whether the needed privilege is granted before doing existence checks.

Alami-Amine:
Hello @bzbarsky-apple, I believe we can close this issue now that the spec was changed and implementation fixed:
- Spec change: https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/11812

- Implementation fixes:
  - https://github.com/project-chip/connectedhomeip/pull/39415
  - https://github.com/project-chip/connectedhomeip/pull/40173
  - https://github.com/project-chip/connectedhomeip/pull/39772
  - https://github.com/project-chip/connectedhomeip/pull/40355

PR #34754, opened 2024-08-02
Make use of DataModel::Provider in writes

andy31415:
This enables ::Write support using the datamodel interface when the data model interface is in use.

### Changes
- redirect writes to ::Write if datamodel interface is enabled
- Update write arguments to contain a `last success` path so that chunking for ACL works.
- added explicit variants to host builds that controls what data model is used (check/enabled/disabled) for easier testing and enabling
- Writing may mark paths as dirty, which requires the data model context. starting passing in a data model context:
   - added context Startup and Shutdown in InteractionModelEngine
   - made InteractionModelEngine set the current exchange as a scoped context when processing a message via OnMessageReceived
- fixed up unit tests. In particular "multiple success/failures" are now not possible anymore as a single return value is done, so that test although it passes, has little meaning anymore.
- Updated order of "read-only/existence check"  in the codegen writer to be done before "ACL" check. **THIS IS WRONG BY SPEC** however it is currently enforced by `TC_AccessChecker.py` to match existing ember_compatibility_functions.cpp


src/app/codegen-data-model-provider/CodegenDataModelProvider_Write.cpp
```diff
@@ -51,7 +51,7 @@ using Protocols::InteractionModel::Status;
 ///
 /// If it returns std::nullopt, then there is no AAI to handle the given path
 /// and processing should figure out the value otherwise (generally from other ember data)
-std::optional<CHIP_ERROR> TryWriteViaAccessInterface(const ConcreteAttributePath & path, AttributeAccessInterface * aai,
+std::optional<CHIP_ERROR> TryWriteViaAccessInterface(const ConcreteDataAttributePath & path, AttributeAccessInterface * aai,
```
andy31415, line 54: This was a quite non-obvious bug: Data contains `mListOp` however without it, there is no such thing. And amazingly enough they are cross-assignable so this was failing silently.

I did not add  an explicit unit test here because TestWriteChunking catches this big time.

src/app/InteractionModelEngine.cpp
```diff
@@ -1711,16 +1716,43 @@ DataModel::Provider * InteractionModelEngine::SetDataModelProvider(DataModel::Pr
...
 
-DataModel::Provider * InteractionModelEngine::GetDataModelProvider() const
+DataModel::Provider * InteractionModelEngine::GetDataModelProvider()
 {
 #if CHIP_CONFIG_USE_DATA_MODEL_INTERFACE
-    // TODO: this should be temporary, we should fully inject the data model
-    VerifyOrReturnValue(mDataModelProvider != nullptr, CodegenDataModelProviderInstance());
+    if (mDataModelProvider == nullptr)
+    {
+        // Generally this is expected to be thread safe as these should be called within the CHIP processing loop.
```
bzbarsky-apple, line 1753: Why are we not asserting that assumption?

src/app/InteractionModelEngine.cpp
```diff
@@ -974,6 +975,10 @@ CHIP_ERROR InteractionModelEngine::OnMessageReceived(Messaging::ExchangeContext
 
     Protocols::InteractionModel::Status status = Status::Failure;
 
+    // Ensure that DataModel::Provider have access to the currently executing
+    // exchange context.
```
bzbarsky-apple, line 979:
```suggestion
    // Ensure that DataModel::Provider have access to the exchange the message was received on.
```

src/app/WriteHandler.h
```diff
@@ -198,6 +207,10 @@ class WriteHandler : public Messaging::ExchangeDelegate
     bool mAttributeWriteSuccessful                = false;
     Optional<AttributeAccessToken> mACLCheckCache = NullOptional;
 
+#if CHIP_CONFIG_USE_DATA_MODEL_INTERFACE
+    DataModel::Provider * mDataModelProvider = nullptr;
```
bzbarsky-apple, line 211: Please watch out for RAM increases.

We should think a bit about how to reduce the set of management/provider/etc pointers this thing hold on to.

src/app/reporting/Engine.cpp
```diff
@@ -1004,7 +1005,21 @@ void Engine::ScheduleUrgentEventDeliverySync(Optional<FabricIndex> fabricIndex)
     Run();
 }
 
-}; // namespace reporting
+void Engine::MarkDirty(const ConcreteAttributePath & path)
+{
+    // NOTE: original MatterReportingAttributeChangeCallback ALSO does a
+    //       IncreaseClusterDataVersion. However that ties directly to ember so it is not
```
bzbarsky-apple, line 1011: So what increases the data versions, then?  Or is that handled by the data model provider directly?

andy31415: It was supposed to be a TODO on my side - this is why some integration tests fail I believe.

I will add this increase inside other places (generally ember is supposed to do this internally: when an attribute is changed, it should increase the version it holds and not rely externally on the version management).

andy31415: I added a change to propagate the change status from within ember, so that the codegen data model can actually match the dirty marking from ember.

src/app/WriteHandler.h
andy31415: This one is rough ... I moved the bools into a bitflags so instead of 5xbool I am now using a uint8_t, so this should save the 4 bytes for a pointer (and due to moving things around it may save more on padding...).

I agree that we may want to consider what to do with this, maybe make the IME pass in the pointer during the handle and not making the handler maintain its own pointer. However that change seems to be more complex so I would rather defer it.

src/app/InteractionModelEngine.cpp
```diff
@@ -975,8 +976,7 @@ CHIP_ERROR InteractionModelEngine::OnMessageReceived(Messaging::ExchangeContext
 
     Protocols::InteractionModel::Status status = Status::Failure;
 
-    // Ensure that DataModel::Provider have access to the currently executing
-    // exchange context.
+    // Ensure that DataModel::Provider have access to the exchange the message was received on.
```
bzbarsky-apple, line 979:
```suggestion
    // Ensure that DataModel::Provider has access to the exchange the message was received on.
```

src/app/WriteHandler.cpp
```diff
@@ -567,24 +571,33 @@ Status WriteHandler::ProcessWriteRequest(System::PacketBufferHandle && aPayload,
 #if CHIP_CONFIG_IM_PRETTY_PRINT
     writeRequestParser.PrettyPrint();
 #endif
-    err = writeRequestParser.GetSuppressResponse(&mSuppressResponse);
+    bool boolValue;
+
+    boolValue = mStateFlags.Has(StateBits::kSuppressResponse);
+    err       = writeRequestParser.GetSuppressResponse(&boolValue);
+    mStateFlags.Set(StateBits::kSuppressResponse, boolValue);
```
bzbarsky-apple, line 578: Now that we have broken up the GetSuppressResponse and the Set into separate lines of code, the Set should happen below after the SuccessOrExit(err), no?

Similar for the other bits below.

src/app/codegen-data-model-provider/CodegenDataModelProvider_Write.cpp
```diff
@@ -395,10 +404,11 @@ DataModel::ActionReturnStatus CodegenDataModelProvider::WriteAttribute(const Dat
...
     // - Internal writes may need to be able to decide if to mark things dirty or not (see AAI as well)
     // - Changes-ommited paths should not be marked dirty (ember is not aware of that flag)
-    // - This likely maps to `MatterReportingAttributeChangeCallback` HOWEVER current ember write functions
-    //   will selectively call that one depending on old attribute state (i.e. calling every time is a
-    //   change in behavior)
-    CurrentContext().dataModelChangeListener->MarkDirty(request.path);
+    if (attributeChanged == AttributeChanged::kValueChanged)
+    {
+        emberAfIncreaseClusterDataVersion(request.path);
+        CurrentContext().dataModelChangeListener->MarkDirty(request.path);
```
bzbarsky-apple, line 410: Why do we want these changes?  This still requires the ember code to support dirty-marking itself, but now we are supporting it here too, which just adds to the codesize for no good reason.  This part does not make sense to me.

src/app/reporting/Engine.cpp
```diff
@@ -1009,8 +1009,10 @@ void Engine::MarkDirty(const ConcreteAttributePath & path)
 {
     // NOTE: original MatterReportingAttributeChangeCallback ALSO does a
     //       IncreaseClusterDataVersion. However that ties directly to ember so it is not
-    //       done here. The code should be migrated to execute this.
-
+    //       done here.
+    //
+    //       Instead the DataModel::Provider is responsible to properly increase cluster
+    //       data versions.
```
bzbarsky-apple, line 1015: Fundamentally: any time things are marked as dirty for reporting that _must_ also increase cluster data versions.  Anything that decouples those things is going to be a source of bugs...

So if the intent is that the DataModel::Provider both bumps data versions and calls `Engine::MarkDirty`, that's fine; the ember version can do that via its current function it calls.

src/app/codegen-data-model-provider/CodegenDataModelProvider_Write.cpp
```diff
@@ -356,6 +357,8 @@ DataModel::ActionReturnStatus CodegenDataModelProvider::WriteAttribute(const Dat
             //       however for now this is called directly because ember code does this call
             //       inside emberAfWriteAttribute.
             MatterReportingAttributeChangeCallback(request.path);
+
+            emberAfIncreaseClusterDataVersion(request.path);
```
bzbarsky-apple, line 361: This increases the data version twice.  Why do we want that?

bzbarsky-apple (changes requested):
Meant to request changes.  The data version bits here don't make sense.

src/app/WriteHandler.cpp
andy31415: This is what I had initially, however that is not what was happening in the previous code: in the previous code, the bool setting from the `Get` was happening regardless of the return code of success or error.

So to replicate a `GetFoo(mMember)`, the exact equivalent is `boolValue = flags.Has(Member); Get(boolValue); flags.Set(Member, boolValue)`

src/app/codegen-data-model-provider/CodegenDataModelProvider_Write.cpp
andy31415: Existing ember code at write time is totally unaware of cluster versioning as far as I could tell - it lets cluster version being increased as a sideffect of change callbacks.

This is the bit that I find somewhat hard to follow on the intent: we advise clusters to call `MatterAttributeChangedCallback` which marks dirty and increases the cluster version. However cluster version seems to be a property of clusters/data-model.

Will followup over chat on this since the existing split (on why version is in emberAf ... probably because of convenience of data structure) makes things hard.

src/app/codegen-data-model-provider/CodegenDataModelProvider_Write.cpp
andy31415: We only want one call. Removed the reporting callback as the `increase version + dirty mark` should be the equivalent.

src/app/reporting/Engine.cpp
andy31415: Existing solution is circular dependency-wise: ember calls reporting changed (mark dirty) and reporting needs to change version ... which is held by ember.

I could add versioning calls to the Provider interface for specifically notifying of version changes, however this seems odd: Provider already handles versions (they are returned as part of iteration since we can filter on them) and provider calls `MarkDirty`, so it seems odd for a provider to say `mark dirty` just to get back a callback of `also increase version` which we already knew it is needed when we called MarkDirty in the first place.

src/app/WriteHandler.cpp
bzbarsky-apple: I understand that, but the previous code was sort of relying on things it would be better to not rely on to save a few lines of code, and we should not keep relying on them, since we have those extra lines of code anyway.

src/app/codegen-data-model-provider/CodegenDataModelProvider_Write.cpp
bzbarsky-apple: >  it lets cluster version being increased as a sideffect of change callbacks.

Right, cluster version increases at the place where the data mutates, whether it's due to a write or due to AttributeName::Set, or whatever.

> we advise clusters to call MatterAttributeChangedCallback which marks dirty and increases the cluster version.

We advise _whoever controls changes to the data_ to do that when the data changes.  If the cluster controls those changes, it should do that.

> cluster version seems to be a property of clusters/data-model.

What it's a property of and who is responsible for updating it are different questions.

src/app/WriteHandler.cpp
andy31415: #34992 makes this change.

src/app/codegen-data-model-provider/CodegenDataModelProvider_Write.cpp
andy31415: Updated with code changes after #35455

src/app/codegen-data-model-provider/CodegenDataModelProvider_Write.cpp
```diff
@@ -322,7 +319,45 @@ DataModel::ActionReturnStatus CodegenDataModelProvider::WriteAttribute(const Dat
...
+    // For chunking, ACL check is not re-done if the previous write was successful for the exact same
+    // path. We apply this everywhere as a shortcut, although realistically this is only for AccessControl cluster
+    if (checkAcl && request.previousSuccessPath.has_value())
+    {
+        checkAcl = ((request.path.mEndpointId != request.previousSuccessPath->mEndpointId) ||
+                    (request.path.mClusterId != request.previousSuccessPath->mClusterId) ||
+                    (request.path.mAttributeId != request.previousSuccessPath->mAttributeId));
+    }
+
+    // If not writable, do unsupported write instead of unsupported-access
```
tehampson, line 336: Is this comment still relevant? If so I am not following and it doesn't make sense to me

src/app/codegen-data-model-provider/CodegenDataModelProvider_Write.cpp
```diff
@@ -322,7 +319,45 @@ DataModel::ActionReturnStatus CodegenDataModelProvider::WriteAttribute(const Dat
...
+                                         .requestType = Access::RequestType::kAttributeWriteRequest,
+                                         .entityId    = request.path.mAttributeId };
+        CHIP_ERROR err = Access::GetAccessControl().Check(*request.subjectDescriptor, requestPath,
+                                                          RequiredPrivilege::ForWriteAttribute(request.path));
+
+        if (err != CHIP_NO_ERROR)
+        {
+            ReturnErrorCodeIf((err != CHIP_ERROR_ACCESS_DENIED) && (err != CHIP_ERROR_ACCESS_RESTRICTED_BY_ARL), err);
+
+            // TODO: when wildcard/group writes are supported, handle them to discard rather than fail with status
```
tehampson, line 353: Should this todo happen in this PR? Sending a response to a group write seems like something we shouldn't be doing

src/app/InteractionModelEngine.h
```diff
@@ -402,7 +403,10 @@ class InteractionModelEngine : public Messaging::UnsolicitedMessageHandler,
     }
 #endif
 
-    DataModel::Provider * GetDataModelProvider() const;
+    // Temporarly NOT const because the data model provider will be auto-set
+    // to codegen on first usage. This behaviour will be changed once each
+    // application must explicitly set the data model provider.
+    DataModel::Provider * GetDataModelProvider();
```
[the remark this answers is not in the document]
andy31415: No, however this is current development and I figured we would iterate on it anyway (I have todos to make data model setting be a choice of the app, so app builds its backend (e.g. ember is currently on every app and every app has its own zapfile) and then provided should be set in main. Then this can become const.

Regarding issues, we currently have 1635 open issues (and 10K closed, a lot by stale bot) so for my own development adding github issues does not seem helpful. I can add an issue if you think it is important, it is just my default to not add them if actively developing things.

src/app/codegen-data-model-provider/CodegenDataModelProvider_Write.cpp
andy31415: This was copied and pasted from src/app/util/ember-compatibility-functions.cpp which this code emulates.

src/app/codegen-data-model-provider/CodegenDataModelProvider_Write.cpp
andy31415: You are correct - this was a left over from when the bug was we need to say "unsupported write" instead of access denied. This is no handled and commented above. Removed this.

src/app/codegen-data-model-provider/CodegenDataModelProvider_Write.cpp
andy31415: I believe we currently do not handle group writes at all (since this originates from https://github.com/project-chip/connectedhomeip/blob/master/src/app/util/ember-compatibility-functions.cpp#L708 , so the best this PR can do is replicate that.

src/app/InteractionModelEngine.h
```diff
@@ -698,7 +705,23 @@ class InteractionModelEngine : public Messaging::UnsolicitedMessageHandler,
 
     SubscriptionResumptionStorage * mpSubscriptionResumptionStorage = nullptr;
 
-    DataModel::Provider * mDataModelProvider = nullptr;
+    DataModel::Provider * mDataModelProvider      = nullptr;
+    Messaging::ExchangeContext * mCurrentExchange = nullptr;
+
+    // Changes the current exchange context of a InteractionModelEngine to a given context
+    class ScopedExchangeContext
```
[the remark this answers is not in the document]
andy31415: Updated to `CurrentExchangeValueScope`

src/app/InteractionModelEngine.h
```diff
@@ -402,7 +403,10 @@ class InteractionModelEngine : public Messaging::UnsolicitedMessageHandler,
     }
 #endif
 
-    DataModel::Provider * GetDataModelProvider() const;
+    // Temporarly NOT const because the data model provider will be auto-set
```
bzbarsky-apple, line 406:
```suggestion
    // Temporarily NOT const because the data model provider will be auto-set
```

src/app/InteractionModelEngine.h
```diff
@@ -411,6 +415,9 @@ class InteractionModelEngine : public Messaging::UnsolicitedMessageHandler,
     // Returns the old data model provider value.
     DataModel::Provider * SetDataModelProvider(DataModel::Provider * model);
 
+    /* DataModel::ActionContext implementation */
+    Messaging::ExchangeContext * CurrentExchange() override { return mCurrentExchange; }
```
bzbarsky-apple, line 419: Can we make this private, please?  This API is not supposed to be used on InteractionModelEngine directly.

src/app/InteractionModelEngine.h
```diff
@@ -698,7 +705,23 @@ class InteractionModelEngine : public Messaging::UnsolicitedMessageHandler,
 
     SubscriptionResumptionStorage * mpSubscriptionResumptionStorage = nullptr;
 
-    DataModel::Provider * mDataModelProvider = nullptr;
+    DataModel::Provider * mDataModelProvider      = nullptr;
+    Messaging::ExchangeContext * mCurrentExchange = nullptr;
+
+    // Changes the current exchange context of a InteractionModelEngine to a given context
```
bzbarsky-apple, line 711: It's funny that we went to all the trouble of getting rid of this sort of global state and just passing relevant context to command handlers as part of CommandHandler, and now we are walking it back....

Put another way, maybe the ActionContext for an action should not be a global?  Followup to think about that.

src/app/WriteHandler.cpp
```diff
@@ -703,5 +720,32 @@ void WriteHandler::MoveToState(const State aTargetState)
     ChipLogDetail(DataManagement, "IM WH moving to [%s]", GetStateStr());
 }
 
+CHIP_ERROR WriteHandler::WriteClusterData(const Access::SubjectDescriptor & subject, const ConcreteDataAttributePath & path,
+                                          TLV::TLVReader & data)
```
bzbarsky-apple, line 724: Please follow file style for argument naming.

src/app/WriteHandler.h
```diff
@@ -18,11 +18,13 @@
 
 #pragma once
 
+#include "ConcreteAttributePath.h"
```
bzbarsky-apple, line 21:
```suggestion
#include <app/ConcreteAttributePath.h>
```

src/app/WriteHandler.h
```diff
@@ -69,6 +71,7 @@ class WriteHandler : public Messaging::ExchangeDelegate
      *  construction until a call to Close is made to terminate the
      *  instance.
      *
+     *  @param[in] apProvider              A Valid pointer to the model used to forward writes towards
```
bzbarsky-apple, line 74:
```suggestion
     *  @param[in] apProvider              A valid pointer to the model used to forward writes towards
```
Yes, the thing you copy/pasted has the same problem.

src/app/WriteHandler.h
```diff
@@ -182,11 +185,20 @@ class WriteHandler : public Messaging::ExchangeDelegate
...
+                                TLV::TLVReader & data);
+
     Messaging::ExchangeHolder mExchangeCtx;
     WriteResponseMessage::Builder mWriteResponseBuilder;
     Optional<ConcreteAttributePath> mProcessingAttributePath;
     Optional<AttributeAccessToken> mACLCheckCache = NullOptional;
 
+#if CHIP_CONFIG_USE_DATA_MODEL_INTERFACE
+    DataModel::Provider * mDataModelProvider = nullptr;
+    std::optional<ConcreteAttributePath> mPreviousWriteSuccess;
```
bzbarsky-apple, line 199: Please document this member, because it's type and name don't really make sense together.

Or should this be called mLastSuccessfullyWrittenPath?

src/app/codegen-data-model-provider/CodegenDataModelProvider_Write.cpp
```diff
@@ -14,6 +14,7 @@
  *    See the License for the specific language governing permissions and
  *    limitations under the License.
  */
+#include "app/util/attribute-storage.h"
```
bzbarsky-apple, line 17:
```suggestion
#include <app/util/attribute-storage.h>
```

src/app/codegen-data-model-provider/CodegenDataModelProvider_Write.cpp
```diff
@@ -273,27 +285,12 @@ DataModel::ActionReturnStatus CodegenDataModelProvider::WriteAttribute(const Dat
...
-            // TODO: when wildcard/group writes are supported, handle them to discard rather than fail with status
-            return err == CHIP_ERROR_ACCESS_DENIED ? Status::UnsupportedAccess : Status::AccessRestricted;
-        }
-    }
-
+    // TODO: ordering is to check writability/existence BEFORE ACL and this seems wrong, however
+    //       existing unit tests (TC_AcessChecker.py) validate that we get UnsupportedWrite instead of UnsupportedAccess
+    //
+    //       This should likely be fixed in spec (probably already fixed by
+    //       https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/9024)
```
bzbarsky-apple, line 292: Needs to reference https://github.com/project-chip/connectedhomeip/issues/33735 ?

src/app/codegen-data-model-provider/CodegenDataModelProvider_Write.cpp
```diff
@@ -322,7 +319,43 @@ DataModel::ActionReturnStatus CodegenDataModelProvider::WriteAttribute(const Dat
...
+    // ACL check for non-internal requests
+    bool checkAcl = !request.operationFlags.Has(DataModel::OperationFlags::kInternal);
 
+    // For chunking, ACL check is not re-done if the previous write was successful for the exact same
+    // path. We apply this everywhere as a shortcut, although realistically this is only for AccessControl cluster
+    if (checkAcl && request.previousSuccessPath.has_value())
+    {
+        checkAcl = ((request.path.mEndpointId != request.previousSuccessPath->mEndpointId) ||
+                    (request.path.mClusterId != request.previousSuccessPath->mClusterId) ||
+                    (request.path.mAttributeId != request.previousSuccessPath->mAttributeId));
```
bzbarsky-apple, line 333: Why can't you just `*request.previousSuccessPath != request.path`?  I think you can, but if you can't it should be documented why so someone doesn't come along and "simplify" the code.

src/app/codegen-data-model-provider/CodegenDataModelProvider_Write.cpp
bzbarsky-apple: We handle group writes just fine, and have CI for them.

The ember-compatibility-functions comments just need fixing, and these should not duplicate the incorrect claims.

The WriteHandler will drop the response for group writes, so this code is correct in that sense and does not need TODO comments at all.

Wildcard writes, on the other hand, are in fact not supported.

src/app/codegen-data-model-provider/CodegenDataModelProvider_Write.cpp
```diff
@@ -349,18 +382,18 @@ DataModel::ActionReturnStatus CodegenDataModelProvider::WriteAttribute(const Dat
...
         if (*aai_result == CHIP_NO_ERROR)
         {
-            // TODO: change callbacks should likely be routed through the context `MarkDirty` only
-            //       however for now this is called directly because ember code does this call
-            //       inside emberAfWriteAttribute.
-            MatterReportingAttributeChangeCallback(request.path);
-            CurrentContext().dataModelChangeListener->MarkDirty(request.path);
+            // TODO: this is awkward since it provides AAI no control over this, specifically
+            //       AAI may not want to increase versions for some attributes that are Q
+            emberAfAttributeChanged(request.path.mEndpointId, request.path.mClusterId, request.path.mAttributeId, &change_listener);
```
bzbarsky-apple, line 396: It's really odd to have any ember* thing in this code at all.  Why is it here?

src/app/data-model-provider/OperationTypes.h
```diff
@@ -70,15 +70,24 @@ struct ReadAttributeRequest : OperationRequest
...
-    kListEnd   = 0x0004, // This is the LAST list element to write
+    kTimed = 0x0001, // Write is a timed write (i.e. a Timed Request Action preceeded it)
 };
 
 struct WriteAttributeRequest : OperationRequest
 {
     ConcreteDataAttributePath path; // NOTE: this also contains LIST operation options (i.e. "data" path type)
     BitFlags<WriteFlags> writeFlags;
+
+    // The path of the previous successful write in in the same write transaction
```
bzbarsky-apple, line 81:
```suggestion
    // The path of the previous successful write in the same write transaction, if any.
```

src/app/data-model-provider/OperationTypes.h
```diff
@@ -70,15 +70,24 @@ struct ReadAttributeRequest : OperationRequest
...
+    //
+    // In particular this means that a write to this path has succeeded before (i.e. it passed required ACL checks).
+    // The intent for this is to allow short-cutting ACL checks when ACL is in progress of being updated:
+    //   - During write chunking, list writes are of the form "clear list" followed by "append item by item"
```
bzbarsky-apple, line 85:
```suggestion
    //   - During write chunking, list writes can be of the form "reset list" followed by "append item by item"
```
The initial reset might not be to empty list....

src/app/data-model-provider/OperationTypes.h
```diff
@@ -70,15 +70,24 @@ struct ReadAttributeRequest : OperationRequest
...
+    //   - When ACL is updating, such "clear list" would result in the entire ACL being deny and the "append"
```
bzbarsky-apple, line 86:
```suggestion
    //   - When ACL is updating, a reset to empty would result in the entire ACL being deny and the "append"
```

src/app/codegen-data-model-provider/CodegenDataModelProvider_Write.cpp
andy31415: Added reference.

src/app/codegen-data-model-provider/CodegenDataModelProvider_Write.cpp
andy31415: Updated to do simple compare, however I find it unintuitive:
  - request.path is a data path, so it has a mListOp and mListIndex (which will NOT be part of the compare)
  - all ConcreteAttributePath have a `mExpanded` member which is also not part of compares (which also seems ugly to me: class member that is mutable but does not participate in equality/inequality.
  
 Added a comment explaining.

src/app/codegen-data-model-provider/CodegenDataModelProvider_Write.cpp
andy31415: We need to do the same thing that ember does. We currently do a `MatterReportingAttributeChangeCallback` in https://github.com/project-chip/connectedhomeip/blob/master/src/app/util/ember-compatibility-functions.cpp#L736 however that is not something this code does because we are supposed to use the dirty marking in the context rather than the global that `MatterReportingAttributeChangeCallback` is using.

src/app/WriteHandler.h
andy31415: Renamed to mLastSuccessfullyWrittenPath

src/app/WriteHandler.cpp
andy31415: updated.

src/app/InteractionModelEngine.h
andy31415: This is because we started using global exchange manager in places. Very few places, however without it we cannot say the new code supports all the functionality of the old code.

I would prefer we would not expose this and find some other way to process things. Even comments on the context way "use sparringly, likely null in tests because it is impractical to initialize such a complex object as a mock".

src/app/InteractionModelEngine.h
andy31415: Moved the ActionContext as private inheritance. DM providers will see it because that is the best alternative to using actual globals, however I agree it is odd. This is because we have to live with code that exists today.

src/app/codegen-data-model-provider/CodegenDataModelProvider_Write.cpp
andy31415: Dropped the comment - it seems we are supposed to work with this inside WriteHandlers and it is not a responsibility of the provider to figure out what to do, it just reports the status.

src/app/InteractionModelEngine.h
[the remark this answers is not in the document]
tehampson: SGTM

src/app/codegen-data-model-provider/CodegenDataModelProvider_Write.cpp
andy31415: It turns out they are different. What I belive is going on:

- `operator==` is defined on ConcreteDataAttributePath
- ConcreteDataAttributePath has a constructor from ConcreteAttributePath (which is lastSuccessPath)
- So we compare a data path (with list index and operation) with an auto-cast (which will zero out operation and index) so compares on acl writes on lists for example fails

Will add a comment and keep the compare code.

andy31415: Updated, hopefully this is acceptable.

src/app/codegen-data-model-provider/CodegenDataModelProvider_Write.cpp
bzbarsky-apple: But the data version bump also needs to happen in the actual provider bits, not the ember data, no?

bzbarsky-apple: Ah, this is the Codegen provider, so it's tied to ember, ok.

bzbarsky-apple (approved):

yunhanw-google (approved):

PR #34755, opened 2024-08-02
Bump occupancy cluster revision to rev 5 in samples

### LINKED ISSUE #34228: [Occupancy Sensing Cluster]  Cluster Revision value has to be updated as per Spec in Occupancy Sensing Cluster XML
### Description:

 In Spec,  **Occupancy Sensing Cluster** Revision value has been updated from `4` to `5` and EndpointUniqueID attribute has been newly added. So XML has to be updated as per the spec.

**Spec Reference:**
- https://github.com/CHIP-Specifications/connectedhomeip-spec/blob/master/src/app_clusters/OccupancySensing.adoc#1-revision-history

image

**XML Reference:**
- https://github.com/project-chip/connectedhomeip/blob/master/src/app/zap-templates/zcl/data-model/chip/occupancy-sensing-cluster.xml

image

tcarmelveilleux:
- Without this, IDM tests to support latest will fail.

Fixes #34228


jaehs6sam (commented):
Looks good to me.

andy31415 (approved):

woody-apple (approved):

PR #34757, opened 2024-08-02
raise `MTRDevice` for subclassing

kiel-apple:
- [x] rename `MTRDevice` that exists to `MTRDeviceConcrete` (open to other names)
- [ ] ensure tests pass
  - [ ] `MTRSwiftDeviceTests`
  - [ ] all other tests
- [ ] create MTRDevice as interface-only base class - `MTRDeviceConcrete`'s API, but with "this method is not implemented in the base class" stubs for all methods
- [ ] review for `MTRDeviceConcrete` instances that should instead be `MTRDevice`-generic

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice_Concrete.h
+++ b/src/darwin/Framework/CHIP/MTRDevice_Concrete.h
@@ -1,6 +1,6 @@
 /**
  *
- *    Copyright (c) 2024 Project CHIP Authors
+ *    Copyright (c) 2022-2023 Project CHIP Authors
  *
  *    Licensed under the Apache License, Version 2.0 (the "License");
  *    you may not use this file except in compliance with the License.
@@ -15,15 +15,15 @@
  *    limitations under the License.
  */
 
-#import <Matter/MTRDefines.h>
+#import <Foundation/Foundation.h>
+#import <Matter/MTRDevice.h>
 
 NS_ASSUME_NONNULL_BEGIN
 
-/**
- * Calls `-[NSObject isEqual:]` but returns YES when comparing nil to nil.
- */
-MTR_EXTERN BOOL MTREqualObjects(id<NSObject> _Nullable a, id<NSObject> _Nullable b);
+@class MTRDeviceController;
 
-MTR_EXTERN NSNumber * MTRClampedNumber(NSNumber * aNumber, NSNumber * min, NSNumber * max);
+@interface MTRDevice_Concrete : MTRDevice
+
+@end
 
 NS_ASSUME_NONNULL_END
--- a/src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
@@ -31,6 +31,7 @@
 #import "MTRDeviceConnectivityMonitor.h"
 #import "MTRDeviceControllerOverXPC.h"
 #import "MTRDeviceController_Internal.h"
+#import "MTRDevice_Concrete.h"
 #import "MTRDevice_Internal.h"
 #import "MTRError_Internal.h"
 #import "MTREventTLVValueDecoder_Internal.h"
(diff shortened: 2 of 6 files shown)


src/darwin/Framework/CHIP/MTRCluster.h
```diff
@@ -63,7 +63,7 @@ MTR_AVAILABLE(ios(17.4), macos(14.4), watchos(10.4), tvos(17.4))
 /**
  * The device this cluster object is associated with.
  */
-@property (nonatomic, strong, readonly) MTRDevice * device;
+@property (nonatomic, strong, readonly) MTRDeviceConcrete * device;
```
woody-apple, line 66: This should only return a MTRDevice

woody-apple: Shouldn't know what it's working with

src/darwin/Framework/CHIP/Matter.h
```diff
@@ -38,6 +38,7 @@
 #import <Matter/MTRCommissioningParameters.h>
 #import <Matter/MTRDefines.h>
 #import <Matter/MTRDevice.h>
+#import <Matter/MTRDevice_Concrete.h>
```
woody-apple, line 41:
```suggestion
```

Not needed, shouldn't be exported :)

src/darwin/Framework/Matter.xcodeproj/project.pbxproj
```diff
@@ -302,6 +301,8 @@
 		99AECC802798A57F00B6355B /* MTRCommissioningParameters.mm in Sources */ = {isa = PBXBuildFile; fileRef = 99AECC7F2798A57E00B6355B /* MTRCommissioningParameters.mm */; };
 		99C65E10267282F1003402F6 /* MTRControllerTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 99C65E0F267282F1003402F6 /* MTRControllerTests.m */; };
 		99D466E12798936D0089A18F /* MTRCommissioningParameters.h in Headers */ = {isa = PBXBuildFile; fileRef = 99D466E02798936D0089A18F /* MTRCommissioningParameters.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		9BD33D6D2C5D81EF005A72F1 /* MTRDevice_Concrete.h in Headers */ = {isa = PBXBuildFile; fileRef = 9BD33D6B2C5D81EF005A72F1 /* MTRDevice_Concrete.h */; settings = {ATTRIBUTES = (Public, ); }; };
```
woody-apple, line 304: This should not be a publicly exported header

src/darwin/Framework/CHIP/MTRDevice.h
```diff
@@ -33,8 +33,9 @@ typedef NS_ENUM(NSUInteger, MTRDeviceState) {
 
 MTR_AVAILABLE(ios(16.1), macos(13.0), watchos(9.1), tvos(16.1))
 @interface MTRDevice : NSObject
-- (instancetype)init NS_UNAVAILABLE;
-+ (instancetype)new NS_UNAVAILABLE;
+// TODO:  implement and error? kmo 2 aug 2024 14h16
```
woody-apple, line 36: Let's revert these back to how they were

src/darwin/Framework/CHIP/MTRDevice_Concrete.h
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+#import <Foundation/Foundation.h>
+#import <Matter/MTRDevice.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class MTRDeviceController;
+
+MTR_AVAILABLE(ios(16.1), macos(13.0), watchos(9.1), tvos(16.1))
```
woody-apple, line 25: This doesn't need availability, since this is not public API

src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
```diff
@@ -366,7 +384,7 @@ - (BOOL)isEqual:(id)object
 // better behavior.
 #define MTRDEVICE_SUBSCRIPTION_LATENCY_NEW_VALUE_WEIGHT (1.0 / 3.0)
 
-@interface MTRDevice ()
+@interface MTRDevice_Concrete ()
```
woody-apple, line 387: We should duplicate the file for step 1, not rename.

That sets us up nicely for the next stage where we take all of the methods **in** MTRDevice and put warnings in them or logs, when they're called on the abstract version of the class

src/darwin/Framework/CHIP/MTRDevice.h
kiel-apple: note to self:  this was a workaround for the "looking at super" issue for that method specifically, but now it's cropping up on other methods as well.  probably related.

src/darwin/Framework/CHIP/MTRDevice_Concrete.h
```diff
@@ -0,0 +1,33 @@
...
+@interface MTRDevice_Concrete : MTRDevice
```
bzbarsky-apple, line 26: "Concrete" doesn't really mean anything. Presumably other implementations of MTRDevice will also be "concrete"...

What's special about this impl?  Really, the fact that it does things with the Matter SDK directly, right?  So maybe MTRDeviceInProcess?  MTRDeviceDirect?  Having a hard time with the naming....

kiel-apple: `Direct` crossed my mind.

src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
kiel-apple: Redid the first commit to duplicate the `.mm` file.

src/darwin/Framework/CHIP/MTRDevice_Concrete.h
woody-apple: Given this is a hidden class, we can rename it at any time, let's get this merged post tests passing, and please remove this availability flags here, since it's not public API

PR #34758, opened 2024-08-02
tv-casting-app fixed memory issues around stopDiscovery function

pgregorr-amazon:
Updated the common Linux implementation of CastingPlayerDiscovery.h/cpp and CastingPlayer.h/cpp to better manage the lifecycle of CastingPlayers. The stopDiscovery() API can now be called anytime. 

**Change Details** 

1.	Updated CastingPlayerDiscovery.h/cpp to store CastingPlayers in an internal vector to ensure that the CastingPlayer we are trying to connect to is not deleted when switching fragments and prior to calling VerifyOrEstablishConnection().
2.	Updated CastingPlayer.h/cpp  to store static memory::Weak<CastingPlayer> mTargetCastingPlayerWeak; // This is a std::weak_ptr. 
3.	Updated DiscoveryExampleFragment public void onPause() to call stopDiscovery();. This was previously deleting all the CastingPlayers held by CastingPlayerDiscovery.h/cpp std::vector<memory::Strong<CastingPlayer>> mCastingPlayers;
4.	Updated connectedhomeip/examples/tv-casting-app/android/App/app/src/main/res/layout/custom_passcode_dialog.xml so that the dialog text instructions can be visible in both dark and light mode. The text was previously white making it invisible in light mode. 

**Testing**

Verified and tested locally with the Linux, Android and iOS tv-casting-app example mobile apps, and the Linux tv-app (CastingPlayer). Able to build and commission with the example apps using both the commissionee and commissioner generated passcode flows.

### DIFF
--- a/examples/tv-casting-app/android/App/app/src/main/java/com/matter/casting/ConnectionExampleFragment.java
+++ b/examples/tv-casting-app/android/App/app/src/main/java/com/matter/casting/ConnectionExampleFragment.java
@@ -58,7 +58,8 @@ public class ConnectionExampleFragment extends Fragment {
   private TextView connectionFragmentStatusTextView;
   private Button connectionFragmentNextButton;
 
-  // Get a singleton instance of the MatterCastingPlayerDiscovery. Which can be used to call stopDiscovery() during connection.
+  // Get a singleton instance of the MatterCastingPlayerDiscovery. Which can be used to call
+  // stopDiscovery() during connection.
   static final CastingPlayerDiscovery matterCastingPlayerDiscovery =
       MatterCastingPlayerDiscovery.getInstance();
 
@@ -369,15 +370,20 @@ public class ConnectionExampleFragment extends Fragment {
   }
 
   private void stopDiscovery() {
-    Log.i(TAG, "ConnectionExampleFragment stopDiscovery() called, calling MatterCastingPlayerDiscovery.stopDiscovery()");
+    Log.i(
+        TAG,
+        "ConnectionExampleFragment stopDiscovery() called, calling MatterCastingPlayerDiscovery.stopDiscovery()");
 
     MatterError err = matterCastingPlayerDiscovery.stopDiscovery();
     if (err.hasError()) {
       Log.e(
           TAG,
-          "ConnectionExampleFragment stopDiscovery() MatterCastingPlayerDiscovery.stopDiscovery() called, err Stop: " + err);
+          "ConnectionExampleFragment stopDiscovery() MatterCastingPlayerDiscovery.stopDiscovery() called, err Stop: "
+              + err);
     } else {
-      Log.d(TAG, "ConnectionExampleFragment stopDiscovery() MatterCastingPlayerDiscovery.stopDiscovery() success");
+      Log.d(
+          TAG,
+          "ConnectionExampleFragment stopDiscovery() MatterCastingPlayerDiscovery.stopDiscovery() success");
     }
   }
 
--- a/examples/tv-casting-app/android/App/app/src/main/java/com/matter/casting/ConnectionExampleFragment.java
+++ b/examples/tv-casting-app/android/App/app/src/main/java/com/matter/casting/ConnectionExampleFragment.java
@@ -31,6 +31,8 @@ import androidx.annotation.Nullable;
 import androidx.fragment.app.Fragment;
 import com.R;
 import com.matter.casting.core.CastingPlayer;
+import com.matter.casting.core.CastingPlayerDiscovery;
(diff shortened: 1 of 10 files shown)


examples/tv-casting-app/tv-casting-common/core/CastingPlayerDiscovery.cpp
```diff
@@ -63,16 +63,59 @@ CHIP_ERROR CastingPlayerDiscovery::StartDiscovery(uint32_t deviceTypeFilter)
...
+void CastingPlayerDiscovery::ClearDisconnectedCastingPlayersInternal()
+{
+    ChipLogProgress(Discovery, "CastingPlayerDiscovery::ClearDisconnectedCastingPlayersInternal() mCastingPlayersInternal: %lu",
+                    mCastingPlayersInternal.size());
+    // Only clear the CastingPlayers in mCastingPlayersInternal with ConnectionState == CASTING_PLAYER_NOT_CONNECTED
+    ClearDisconnectedCastingPlayers(mCastingPlayersInternal);
+}
+
+void CastingPlayerDiscovery::ClearDisconnectedCastingPlayers(std::vector<std::shared_ptr<CastingPlayer>> & castingPlayers)
+{
```
sharadb-amazon, line 93: If ClearDisconnectedCastingPlayers() is only called from ClearDisconnectedCastingPlayersInternal(), can we just move its implementation to ClearDisconnectedCastingPlayersInternal()?

examples/tv-casting-app/tv-casting-common/core/CastingPlayerDiscovery.cpp
```diff
@@ -63,16 +63,59 @@ CHIP_ERROR CastingPlayerDiscovery::StartDiscovery(uint32_t deviceTypeFilter)
...
+                    mCastingPlayers.size(), mCastingPlayersInternal.size());
     VerifyOrReturnError(mState == DISCOVERY_RUNNING, CHIP_ERROR_INCORRECT_STATE);
     ReturnErrorOnFailure(mCommissionableNodeController.StopDiscovery());
 
+    // Clear mCastingPlayersInternal of disconnected CastingPlayers
+    ClearDisconnectedCastingPlayersInternal();
+    for (const auto & player : mCastingPlayers)
+    {
+        mCastingPlayersInternal.push_back(player);
+    }
```
sharadb-amazon, line 76: Why not just use the copy contructor to copy the vector?

examples/tv-casting-app/android/App/app/src/main/java/com/matter/casting/ConnectionExampleFragment.java
```diff
@@ -56,6 +58,11 @@ public class ConnectionExampleFragment extends Fragment {
   private TextView connectionFragmentStatusTextView;
   private Button connectionFragmentNextButton;
 
+  // Get a singleton instance of the MatterCastingPlayerDiscovery. Which can be used to call
+  // stopDiscovery() during connection.
+  static final CastingPlayerDiscovery matterCastingPlayerDiscovery =
+      MatterCastingPlayerDiscovery.getInstance();
```
sharadb-amazon, line 64: It is more idiomatic to directly call MatterCastingPlayerDiscovery.getInstance() inline where you want to use it given the singleton pattern employed.

pgregorr-amazon: removing the stop discovery button form the Connection page. This is no longer needed.

examples/tv-casting-app/tv-casting-common/core/CastingPlayerDiscovery.cpp
pgregorr-amazon: That works.

examples/tv-casting-app/tv-casting-common/core/CastingPlayerDiscovery.cpp
pgregorr-amazon: Since I was keeping the connected players in mCastingPlayersInternal before copying. Copying mCastingPlayers to mCastingPlayersInternal would delete the players in mCastingPlayersInternal. Thinking about this further, there is no need to keep the the connected players in mCastingPlayersInternal when stopDiscovery() is called. I'll change this to the following:
```
    // Copy mCastingPlayers to mCastingPlayersInternal
    mCastingPlayersInternal = std::vector<memory::Strong<CastingPlayer>>(mCastingPlayers);
```

sharadb-amazon (approved):

chrisdecenzo (approved):

woody-apple (approved):

PR #34765, opened 2024-08-03
Docs re-organization

cecille:
The second phase of the handbook involves linking over to the SDK docs from the handbook site. Before we do this, the SDK documentation needs a re-org and an update. This is a continuation of the re-org, following the platform updates. Remaining: Zap, tools, examples, IDs.

Done in this PR:
- remove API - this is a 4 year old design doc for a test dispatch system that was never implemented. This does not make sense to keep in the docs folder since it gives the incorrect impression that this system was implemented at some point. In reality, our testing is quite different and is decently well documented in the testing section.

- rename discussion to product consideration. This contains only the ipv6 considerations doc, which isn't a discussion. Instead, we're going to use this location to land documents related to product considerations that affect matter, but are provided outside of the SDK, ex. required network stack support (current doc), and factory considerations (coming later) for things like DACs, individually provisioned values, QR codes, config options, certification declaration inclusion, DCL.

- move designing cluster for testing and portability from testing to cluster design - it just makes more sense there.

- Creating a tips and troubleshooting section to land small guides, move in avahi troubleshooting and guide around discovery from a host computer.

- remove quick start guide - everything in here is outdated, we have a getting started guide and a readme that covers all the getting started material. Added references to that in the readme -


woody-apple (approved):

woody-apple:
Pre-approving. Thanks @cecille for cleaning this up.

docs/getting_started/index.md
```diff
@@ -14,4 +14,3 @@ The following docs are a brief introduction to SDK development.
 -   [Running your first example](./first_example.md)
 -   [SDK Basics](./SDKBasics.md)
 -   [ZAP](./zap.md)
```
Vyrastas, line 16: So I noticed this across all the different categories... the menu rendered with Sphinx can show a different order than this manual one, and the actual page title in the menu doesn't always match this link text.

Is it a Sphinx thing to require these top level pages that are just a list of links?  (I'm not that familiar with Sphinx, but in the dev docs I manage we'd never have pages like this).

docs/guides/README.md
Vyrastas, line 1: I don't think we need this page, docs/guides/index.md serves the same purpose now

docs/getting_started/index.md
cecille: I honestly don't know - this wasn't a change i made in this PR, just removing a now dead link. I'll have to defer to you and Sam for getting the organization to appear correct, I'm just trying to get the content in roughly the right spots at this point.

docs/guides/README.md
cecille: removed it.

docs/getting_started/index.md
Vyrastas: Thanks, I'll check with Sam then.

Vyrastas (approved):

ReneJosefsen (approved):

PR #34768, opened 2024-08-05
TBRM: Add missing PendingDatasetTimestamp attribute and CASE session check

### LINKED ISSUE #34751: TBRM implementation missing PendingDatasetTimestamp

wqx6:
Add the PendingDatasetTimestamp to ThreadBorderRouterManagement cluster and add CASE session check for SetDataset commands.

Fixes #34751


src/app/clusters/thread-border-router-management-server/thread-border-router-management-server.cpp
```diff
@@ -168,7 +171,8 @@ void ServerInstance::InvokeCommand(HandlerContext & ctxt)
         break;
     case Commands::SetPendingDatasetRequest::Id:
         HandleCommand<Commands::SetPendingDatasetRequest::DecodableType>(ctxt, [this](HandlerContext & ctx, const auto & req) {
-            ctx.mCommandHandler.AddStatus(ctx.mRequestPath, HandleSetPendingDatasetRequest(req));
+            Status status = HandleSetPendingDatasetRequest(IsCommandOverCASESession(ctx), req);
```
ksperling-apple, line 174: Nit: I think a nicer pattern is to just pass both the `HandlerContext` and the request to the `Handle...Request` method instead of adding additional logic to this lambda.

src/app/clusters/thread-border-router-management-server/thread-border-router-management-server.h
```diff
@@ -76,14 +76,15 @@ class ServerInstance : public CommandHandlerInterface,
...
+    Status HandleSetActiveDatasetRequest(bool isOverCASESession, CommandHandler * commandHandler,
                                          const Commands::SetActiveDatasetRequest::DecodableType & req);
-    Status HandleSetPendingDatasetRequest(const Commands::SetPendingDatasetRequest::DecodableType & req);
+    Status HandleSetPendingDatasetRequest(bool isOverCASESession, const Commands::SetPendingDatasetRequest::DecodableType & req);
     Status HandleGetDatasetRequest(bool isOverCASESession, Delegate::DatasetType type, Thread::OperationalDataset & dataset);
 
     // Attribute Read handlers
     void ReadFeatureMap(BitFlags<Feature> & feature);
     Optional<uint64_t> ReadActiveDatasetTimestamp();
+    Optional<uint64_t> ReadPendingDatasetTimestamp();
```
ksperling-apple, line 87: New code like this cluster should be using `std::optional` directly instead of `Chip::Optional`

src/app/clusters/thread-border-router-management-server/thread-border-router-management-server.cpp
```diff
@@ -264,6 +280,12 @@ CHIP_ERROR ServerInstance::Read(const ConcreteReadAttributePath & aPath, Attribu
                                                    : aEncoder.EncodeNull();
         break;
     }
+    case Attributes::PendingDatasetTimestamp::Id: {
+        Optional<uint64_t> pendingDatasetTimestamp = ReadPendingDatasetTimestamp();
+        status = pendingDatasetTimestamp.HasValue() ? aEncoder.Encode(DataModel::MakeNullable(pendingDatasetTimestamp.Value()))
```
ksperling-apple, line 285:
```suggestion
        status = pendingDatasetTimestamp.HasValue() ? aEncoder.Encode(pendingDatasetTimestamp.Value())
```
There's no need to wrap the value back into a Nullable for `Encode` since you have already checked for null

ksperling-apple (approved):
In future it would be good to push the manual changes and the effects of `zap_regen_all` as separate commits to make them easier to review.

ksperling-apple:
You'll need to update the zap file for the Network Manager example app (by opening it in ZAP and ensuring the PendingDatasetTimestamp attribute is enabled) for the builds to pass.

Also update the `Test_TC_TBRM_2_1.yaml` test to uncomment the PendingDatasetTimestamp read step at the end.

src/app/clusters/thread-border-router-management-server/thread-border-router-management-server.h
wqx6: Done.

src/app/clusters/thread-border-router-management-server/thread-border-router-management-server.cpp
wqx6: Done

src/app/clusters/thread-border-router-management-server/thread-border-router-management-server.cpp
wqx6: Done

marchemi (approved):

marchemi (approved):

examples/network-manager-app/network-manager-common/network-manager-app.zap
```diff
@@ -3420,7 +3420,7 @@
               "storageOption": "External",
               "singleton": 0,
               "bounded": 0,
-              "defaultValue": "",
+              "defaultValue": null,
```
chrisdecenzo, line 3423: are these "" -> null changes intentional?

chrisdecenzo (approved):

src/app/clusters/thread-border-router-management-server/thread-border-router-management-server.cpp
```diff
@@ -259,15 +279,28 @@ CHIP_ERROR ServerInstance::Read(const ConcreteReadAttributePath & aPath, Attribu
...
         break;
     }
     default:
         break;
     }
-    return status;
+    if (status == CHIP_ERROR_NO_MEMORY)
+    {
+        // If the status returned by Encode function is CHIP_ERROR_NO_MEMORY, map it to IM status ResourceExhausted
+        return CHIP_IM_GLOBAL_STATUS(ResourceExhausted);
```
bzbarsky-apple, line 297: This will break read/subscribe chunking.  Why are we doing this?

src/app/clusters/thread-border-router-management-server/thread-border-router-management-server.cpp
```diff
@@ -259,15 +279,28 @@ CHIP_ERROR ServerInstance::Read(const ConcreteReadAttributePath & aPath, Attribu
...
+    }
+    if (status != CHIP_NO_ERROR)
+    {
+        return CHIP_IM_GLOBAL_STATUS(Failure);
```
bzbarsky-apple, line 301: No, this is wrong.  The contract for AttributeAccessInterface right now is that any error returned from the Encode() call _must_ be propagated out as-is.  Anything else breaks chunking.

bzbarsky-apple (dismissed):

examples/network-manager-app/network-manager-common/network-manager-app.zap
wqx6: No, they are changed by zap automatically after I modified the zap file.

src/app/clusters/thread-border-router-management-server/thread-border-router-management-server.cpp
wqx6: Returned status now.

src/app/clusters/thread-border-router-management-server/thread-border-router-management-server.cpp
wqx6: Return status now

PR #34774, opened 2024-08-05
[Silabs] transition to application screen after commissioning from QR code screen

rosahay-silabs:
Adds auto transition to application screen after a successful commissioning only if application is on the QR code screen.


kiel-apple (approved):

jmartinez-silabs (approved):

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -188,6 +188,17 @@ void BaseApplicationDelegate::OnCommissioningWindowClosed()
         }
     }
 #endif // CHIP_CONFIG_ENABLE_ICD_SERVER && SLI_SI917
+    if (BaseApplication::GetProvisionStatus())
```
mkardous-silabs, line 191: Is `OnCommissioningWindowClosed` the best location for this screen change?
The API is called irrespective of if the commissioning has been succesful or not.

If we opt to keep it here, we should add the same code to transition to the QR code when the commissioning window is opened.

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -188,6 +188,17 @@ void BaseApplicationDelegate::OnCommissioningWindowClosed()
...
+    {
+#ifdef DISPLAY_ENABLED
+#ifdef QR_CODE_ENABLED
+        SilabsLCD::Screen_e screen;
+        slLCD.GetScreen(screen);
+        VerifyOrReturn(screen == SilabsLCD::Screen_e::QRCodeScreen);
```
mkardous-silabs, line 197: What happens if the screen is showing the "information" screen? Is the decision to leave it as is?

rosahay-silabs: Yes, this is only for the transition from the QR code screen to Demo UI screen. If it is information screen then it remains there.

examples/platform/silabs/BaseApplication.cpp
rosahay-silabs: `BaseApplication::GetProvisionStatus()` would be indicative of a successful provision, right? But I see a point in having the QR code screen to be tied up to commissioning window open. @jmartinez-silabs your thoughts ?

PR #34785, opened 2024-08-05
Refactoring of support test scripts to enhance matter testing infrastructure

vatsalghelani-csa:
Work for Issue #31317 

Follow up to PR #34377 

PR 15 - Implements several structural changes to enhance matter testing organization
- Directory Restructure:
   - Moved support scripts one level in the testing_support directory hierarchy to streamline the structure.
- Refactoring Modules:
   - Copied support scripts, along with their tests, into a newly created subdirectory dedicated to matter testing.
   - Removed the support suffix during the copy process for cleaner naming conventions.
- Script Updates:
   - Modified scripts to utilize the newly organized directories and modules.


src/python_testing/TestChoiceConformanceSupport.py
```diff
@@ -19,11 +19,11 @@
 import xml.etree.ElementTree as ElementTree
 
 import jinja2
-from choice_conformance_support import (evaluate_attribute_choice_conformance, evaluate_command_choice_conformance,
```
andy31415, line 22: the underlying tests should be moved in `matter_testing_infrastructure` and made run as the `tests =` for pw_python_package.

https://github.com/project-chip/connectedhomeip/blob/master/src/python_testing/matter_testing_infrastructure/BUILD.gn#L35


However if we want to do this later that is fine as well, feel free to close this comment and do as a followup. The change would be:
   - move these files and add to tests = in build.gn
   - remove them from tests.yaml (as they start running as part of `check`)
 
 At that point we can also remove them from exceptions from the globbing implementation of the runner.

src/python_testing/MinimalRepresentation.py
```diff
@@ -18,9 +18,9 @@
 from dataclasses import dataclass, field
 
 from chip.tlv import uint
-from conformance_support import ConformanceDecision
-from global_attribute_ids import GlobalAttributeIds
-from matter_testing_support import MatterBaseTest, async_test_body, default_matter_test_main
+from testing_support.conformance import ConformanceDecision
```
andy31415, line 21: `testing_support` feels a bit global. This is specific to matter/chip so how about either `matter_testing_suport` or `chip_testing_support` ? I realinze that is a larger delta, however maybe it can be done with rg and sd, like `rg 'from testing_support' --files-with-matches | xargs sd 'from testing_support' 'from matter_testing_support'` and then with a few more manual edits.

andy31415 (approved):

src/python_testing/MinimalRepresentation.py
andy31415: above was using:
   https://github.com/BurntSushi/ripgrep
   https://github.com/chmln/sd
   
I found these really useful for mass refactoring in the sdk

vatsalghelani-csa: Resolved.

cecille (changes requested):
Adding a request changes marker so this doesn't get accidentally merged during the lead up to SVE, since this is potentially disruptive.

src/python_testing/matter_testing_infrastructure/matter_testing_support/spec_parsing.py
```diff
@@ -517,18 +519,35 @@ class DataModelLevel(str, Enum):
...
 
+def _get_data_model_root() -> str:
+    """Attempts to find ${CHIP_ROOT}/data_model or equivalent."""
+
+    # Since this class is generally in a module, we have to rely on being bootstrapped or
+    # we use CWD if we cannot
+    choices = [os.getcwd()]
+
+    if 'PW_PROJECT_ROOT' in os.environ:
+        choices.insert(0, os.environ('PW_PROJECT_ROOT'))
```
andy31415, line 530:
```suggestion
        choices.insert(0, os.environ['PW_PROJECT_ROOT'])
```

vatsalghelani-csa: Resolved.

scripts/spec_xml/paths.py
```diff
@@ -0,0 +1,64 @@
...
+from enum import Enum
+
+# Define a branch enum for different versions or branches
+
+
+class Branch(Enum):
+    MASTER = "master"
+    IN_PROGRESS = "in_progress"
+    V1_3 = "v1_3"
+
```
plauric, line 27: The V1_4 entry will need to be added in the near future, so I'd create an issue for that.

plauric (approved):

scripts/spec_xml/paths.py
andy31415: Marking resolved: should probably be done as a aseparate PR.

src/python_testing/matter_testing_infrastructure/matter_testing_support/spec_parsing.py
```diff
@@ -612,25 +630,6 @@ def remove_problem(location: typing.Union[CommandPathLocation, FeaturePathLocati
             0x05: XmlAttribute(name='SupportedTemperatureLevels', datatype='list', conformance=feature(0x02, 'TL'), read_access=view, write_access=none, write_optional=False),
         }
 
-    # TODO: Need automated parsing for atomic attributes.
```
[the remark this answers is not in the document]
vatsalghelani-csa: Added it back, maybe was due to a bad merge.

src/python_testing/matter_testing_infrastructure/matter_testing_support/spec_parsing.py
```diff
@@ -777,7 +795,7 @@ def parse_single_device_type(root: ElementTree.Element) -> tuple[list[ProblemNot
 
 
 def build_xml_device_types(data_model_directory: typing.Union[PrebuiltDataModelDirectory, str] = PrebuiltDataModelDirectory.k1_4) -> tuple[dict[int, XmlDeviceType], list[ProblemNotice]]:
-    dir = _get_data_model_directory(data_model_directory, DataModelLevel.kDeviceType)
+    dir = get_data_model_directory(data_model_directory, DataModelLevel.kDeviceType)
```
cecille, line 798: can you leave this function as private?

src/python_testing/matter_testing_infrastructure/matter_testing_support/spec_parsing.py
```diff
@@ -518,19 +519,36 @@ class DataModelLevel(str, Enum):
     kDeviceType = 'device_types'
 
 
-def _get_data_model_directory(data_model_directory: typing.Union[PrebuiltDataModelDirectory, str], data_model_level: DataModelLevel) -> str:
+def _get_data_model_root() -> str:
```
cecille, line 522: I don't think this is going to work in the test harness. If bundling the module to make it portable is going to break this pathing, then we need to consider packing the data model files into the built module.

andy31415: TH will probably have to figure this out in some way. Since it passes today, I imagine TH pulls in a full checkout and has XML files. We should probably also support some additional env variable (like MATTER_DATA_MODEL_PATH) that TH can use until we can bundle XMLs into the module itself.

scripts/spec_xml/generate_spec_xml.py
```diff
@@ -24,26 +24,17 @@
...
-CURRENT_IN_PROGRESS_DEFINES = ['aliro', 'atomicwrites', 'battery-storage', 'device-location', 'e2e-jf', 'energy-calendar', 'energy-drlc',
-                               'energy-management', 'heat-pump', 'hrap-1', 'hvac', 'matter-fabric-synchronization', 'metering', 'secondary-net',
-                               'service-area-cluster', 'solar-power', 'tcp', 'water-heater', 'wifiSetup']
+# Use the get_in_progress_defines() function to fetch the in-progress defines
+CURRENT_IN_PROGRESS_DEFINES = get_in_progress_defines()
+
+# Replace hardcoded paths with dynamic paths using paths.py functions
+DEFAULT_CHIP_ROOT = get_chip_root()
+DEFAULT_OUTPUT_DIR_1_3 = get_data_model_path(Branch.V1_3)
+DEFAULT_OUTPUT_DIR_IN_PROGRESS = get_data_model_path(Branch.MASTER)
```
andy31415, line 35: in Previous code, IN_PROGRESS dir went to "in_progress" and TOT went to "master".

In the updated code, both go to Master. why?

cecille (approved):

scripts/spec_xml/paths.py
```diff
@@ -0,0 +1,109 @@
...
+
+class Branch(Enum):
+    MASTER = "master"
+    V1_3 = "v1_3"
+    V1_4 = "v1_4"
+    IN_PROGRESS = "in_progress"
+
+
+def get_chip_root():
```
arkq, line 30: HIT: How about returning `Path` object here (and in the rest of these path helpers)?

https://docs.python.org/3/library/pathlib.html

Then you will be able to semantically operate on the path object instead of using `os.path.join` or other `os.path...`. Since we are on Python >= 3.9 there are a lot of goodies in the pathlib library.

src/python_testing/matter_testing_infrastructure/chip/testing/spec_parsing.py
```diff
@@ -518,19 +518,36 @@ class DataModelLevel(str, Enum):
...
+    # we use CWD if we cannot
+    choices = [os.getcwd()]
+
+    if 'PW_PROJECT_ROOT' in os.environ:
+        choices.insert(0, os.environ['PW_PROJECT_ROOT'])
+
+    for c in choices:
+        data_model_path = os.path.join(c, 'data_model')
+        if os.path.exists(os.path.join(data_model_path, 'master', 'scraper_version')):
+            return data_model_path
```
arkq, line 534: Same as before, how about returning `Path` object.

arkq (approved):

scripts/spec_xml/paths.py
andy31415: Let's update that one as a followup due to the sizes of the PR and it being open for a very long time (i.e. harder to iterate on it)

PR #34786, opened 2024-08-05
Update tests.yaml

jaehs6sam:
Need to add TC_OCC_3_1.py and TC_OCC_3_2.py at tests.yaml


andy31415:
@jaehs6sam - this seems to fail CI (test_TC_OCC_3_1  fails at least). My understanding was that these tests require user interaction implementation or automation. Could you add those as well?

jaehs6sam:
Will close this PR and put PR together with CI modified test scripts.

PR #34795, opened 2024-08-06
[HVAC] Clear out atomic writes associated with a removed fabric

### LINKED ISSUE #34552: For presets edit requests, when a fabric is removed we need to cancel any edit request from a node in that fabric
Followup: when fabrics are removed, don't we need to clear out the corresponding open edit requests?

hasty:
Fixes #34552


src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -978,6 +979,19 @@ CHIP_ERROR ThermostatAttrAccess::AppendPendingPreset(Delegate * delegate, const
...
 
+void ThermostatAttrAccess::OnFabricRemoved(const FabricTable & fabricTable, FabricIndex fabricIndex)
+{
+    for (size_t i = 0; i < ArraySize(mAtomicWriteNodeIds); ++i)
+    {
+        auto nodeId = mAtomicWriteNodeIds[i];
+        if (nodeId.GetFabricIndex() == fabricIndex)
+        {
+            mAtomicWriteNodeIds[i] = ScopedNodeId();
+            mAtomicWriteState[i]   = false;
```
bzbarsky-apple, line 990: Followup (because it's not 100% clear what the best way to solve this is): when an endpoint is disabled, we also need to resetAtomicWrite() on that endpoint, right?  Otherwise we can end up in a bad state where the endpoint id gets reused but we still have an atomic state stored for that endpoint it.

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -112,7 +113,7 @@ void TimerExpiredCallback(System::Layer * systemLayer, void * callbackContext)
 {
     EndpointId endpoint = static_cast<EndpointId>(reinterpret_cast<uintptr_t>(callbackContext));
 
-    Delegate * delegate = GetDelegate(endpoint);
+    auto delegate = GetDelegate(endpoint);
```
andy31415, line 116: Is there a reason to change these? The delta seems ok (it is obvious that GetDelegate returns a Delegate) however this makes the diff larger as it stands.

hasty: It was going to have to change anyway, as Thermostat::Delegate collides with FabricTable::Delegate, and seem cleaner to just switch to auto.

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -978,6 +979,19 @@ CHIP_ERROR ThermostatAttrAccess::AppendPendingPreset(Delegate * delegate, const
     return delegate->AppendToPendingPresetList(preset);
 }
 
+void ThermostatAttrAccess::OnFabricRemoved(const FabricTable & fabricTable, FabricIndex fabricIndex)
+{
+    for (size_t i = 0; i < ArraySize(mAtomicWriteNodeIds); ++i)
+    {
+        auto nodeId = mAtomicWriteNodeIds[i];
+        if (nodeId.GetFabricIndex() == fabricIndex)
```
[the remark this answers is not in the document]
hasty: Changed to a struct to track inProgress, nodeId and endpoint, and now checking for inProgress first.

src/app/clusters/thermostat-server/thermostat-server.cpp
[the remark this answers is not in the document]
hasty: Changed to call resetAtomicWrite.

hasty: Added ember callback for shutdown, which calls resetAtomicWrite. Not sure how to write a test for this one.

src/app/clusters/thermostat-server/thermostat-server.cpp
bzbarsky-apple: I was assuming bareword `Delegate` was now ambiguous with the new header include... but I am not seeing that; it seems to work fine.  Why _was_ this change made?

src/app/clusters/thermostat-server/thermostat-server.cpp
bzbarsky-apple: Please make followups be followups so I don't have to keep reviewing the PR over and over again as major changes happen to it.

src/app/clusters/thermostat-server/thermostat-server.cpp
bzbarsky-apple: Why does it collide with `FabricTable::Delegate`?  Doesn't seem to...

bzbarsky-apple (approved):

src/app/clusters/thermostat-server/thermostat-server.cpp
hasty: Sorry, I missed a subtlety here; I was getting the compile error on the calls to GetDelegate from methods on ThermostatAttrAccess, but I replaced Delegate * with auto everywhere, even in this function which has no conflict as it's outside of the class. I can revert those lines if you want.

hasty: OK, reverted the lines without a collision.

hasty:
A dependent PR got merged with all these changes, so closing.

Issue #34796, opened 2024-08-06
[Darwin] MTRDevice ReadClient callbacks need to change state immediately


jtung-apple:
This issue it to track ReadClient callbacks in MTRDevice that result in state / internal state changes, and how those should be set immediately, instead of the current logic, where the state change happens after async-ing to the MTRDevice's queue.

PR #34797, opened 2024-08-06
[Darwin] MTRDevice subscription estalished handler should change state before async

jtung-apple:
The `-[MTRPerControllerStorageTests testMTRDeviceResetSubscription]` test in Darwin tests has been flaky, due to the new subscription reset mechanism introducing a race between read client callback and MTRDevice internal state changes.

This patch fixes the specific unit test failure. For the general cases, I've filed: https://github.com/project-chip/connectedhomeip/issues/34796

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -1193,7 +1193,7 @@ static NSString * const sLastInitialSubscribeLatencyKey = @"lastInitialSubscribe
 {
     os_unfair_lock_lock(&self->_lock);
 
-    // If subscription had reset since this handler was scheduled, do not ececute "established" logic below
+    // If subscription had reset since this handler was scheduled, do not execute "established" logic below
     if (!HaveSubscriptionEstablishedRightNow(_internalDeviceState)) {
         MTR_LOG("%@ _handleSubscriptionEstablished run with internal state %lu - skipping subscription establishment logic", self, static_cast<unsigned long>(_internalDeviceState));
         return;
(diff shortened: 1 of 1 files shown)


src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -1193,23 +1193,20 @@ - (void)_handleSubscriptionEstablished
...
-    [self _clearSubscriptionPoolWork];
-
-    // reset subscription attempt wait time when subscription succeeds
-    _lastSubscriptionAttemptWait = 0;
-    if (HadSubscriptionEstablishedOnce(_internalDeviceState)) {
-        [self _changeInternalState:MTRInternalDeviceStateLaterSubscriptionEstablished];
-    } else {
-        MATTER_LOG_METRIC_END(kMetricMTRDeviceInitialSubscriptionSetup, CHIP_NO_ERROR);
-        [self _changeInternalState:MTRInternalDeviceStateInitialSubscriptionEstablished];
+    // If subscription had reset since this handler was scheduled, do not ececute "established" logic below
```
kiel-apple, line 1196:
```suggestion
    // If subscription had reset since this handler was scheduled, do not execute "established" logic below
```

kiel-apple (approved):

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -2476,6 +2473,21 @@ - (void)_setupSubscriptionWithReason:(NSString *)reason
                                },
                                ^(void) {
                                    MTR_LOG("%@ got subscription established", self);
+                                   std::lock_guard lock(self->_lock);
+
+                                   // First synchronously change state
+
+                                   // reset subscription attempt wait time when subscription succeeds
```
bzbarsky-apple, line 2480: This comment should be next to the `_lastSubscriptionAttemptWait = 0;` line, right?

bzbarsky-apple (approved):

src/darwin/Framework/CHIP/MTRDevice.mm
jtung-apple: Moved in the last commit

kiel-apple:
if #34813 lands before this, please duplicate the change in `MTRDevice_Concrete.mm`

PR #34805, opened 2024-08-06
Adding CIFuzz to Actions

Alami-Amine:
- This PR adds a GitHub workflow to build and run Fuzzers on PRs.

- This workflow (CIFuzz) is based on OSS-Fuzz; it uses the same docker image for connectedhomeip, with same exact toolchain.

- I have currently set `dry-run` to True so it does not report failures if build crashes

- Time of running the fuzzer is configurable: `fuzz-seconds`

more info here: https://google.github.io/oss-fuzz/getting-started/continuous-integration/

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -17,6 +17,7 @@ jobs:
       with:
         oss-fuzz-project-name: 'connectedhomeip'
         language: c++
+        dry-run: true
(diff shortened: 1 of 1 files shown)


Alami-Amine:
after discussion with @andy31415 it was decided that it's better to only test using OSS-Fuzz, one reason being the added overhead and maintenance cost that this CI job will produce

PR #34809, opened 2024-08-06
ModeSelect: Add setter API for SupportedModesManager.

jadhavrohit924:
### Problem
- There should be a setter API for SupportedModesManager so the application can set it to its custom implementation.

### Changes
- Add setter API for SupportedModesManager.
- Use common static implementation for esp32.
- Set SupportedModesManager in the examples.

### Testing
- Tested all cluster app on linux and esp32.

### DIFF
--- a/src/app/clusters/mode-select-server/mode-select-server.cpp
+++ b/src/app/clusters/mode-select-server/mode-select-server.cpp
@@ -49,7 +49,7 @@ using BootReasonType = GeneralDiagnostics::BootReasonEnum;
 
 static InteractionModel::Status verifyModeValue(const EndpointId endpointId, const uint8_t newMode);
 
-ModeSelect::SupportedModesManager *sSupportedModesManager = nullptr;
+ModeSelect::SupportedModesManager * sSupportedModesManager = nullptr;
 
 const SupportedModesManager * ModeSelect::getSupportedModesManager()
 {
--- a/src/app/clusters/mode-select-server/mode-select-server.cpp
+++ b/src/app/clusters/mode-select-server/mode-select-server.cpp
@@ -49,7 +49,7 @@ using BootReasonType = GeneralDiagnostics::BootReasonEnum;
 
 static InteractionModel::Status verifyModeValue(const EndpointId endpointId, const uint8_t newMode);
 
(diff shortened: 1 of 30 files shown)


woody-apple (approved):

examples/all-clusters-app/all-clusters-common/src/occupancy-sensing-stub.cpp
```diff
@@ -29,7 +29,7 @@ using namespace chip::DeviceLayer;
 using chip::Protocols::InteractionModel::Status;
 
 static std::unique_ptr<OccupancySensingAttrAccess>
-    gAttrAccess[MATTER_DM_OCCUPANCY_SENSING_CLUSTER_SERVER_ENDPOINT_COUNT + CHIP_DEVICE_CONFIG_DYNAMIC_ENDPOINT_COUNT];
+    gAttrAccess[MATTER_DM_OCCUPANCY_SENSING_CLUSTER_SERVER_ENDPOINT_COUNT + CHIP_DEVICE_CONFIG_DYNAMIC_ENDPOINT_COUNT + 1];
```
shubhamdp, line 32: Why is this `+1` needed? For fixing the crash on boot for all-clusters-app/esp32?

src/app/clusters/mode-select-server/mode-select-server.cpp
```diff
@@ -49,6 +49,17 @@ using BootReasonType = GeneralDiagnostics::BootReasonEnum;
 
 static InteractionModel::Status verifyModeValue(const EndpointId endpointId, const uint8_t newMode);
 
+ModeSelect::SupportedModesManager * sSupportedModesManager = nullptr;
```
shubhamdp, line 52: Make this static or move the anonymous namespace?

examples/all-clusters-app/all-clusters-common/src/occupancy-sensing-stub.cpp
jadhavrohit924: Ughh!! This should not go in the PR. I just did this for testing, I will remove it.

shubhamdp (approved):

PR #34813, opened 2024-08-06
[Darwin] MTRDevice -> MTRDevice_Concrete

kiel-apple:
Prepare to raise with minimal code change to existing MTRDevice.


woody-apple (approved):

woody-apple:
Awesome, thanks @kiel-apple ! Fast tracking since this only affects darwin.

src/darwin/Framework/CHIP/MTRDevice_Concrete.h
```diff
@@ -0,0 +1,29 @@
...
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+#import <Foundation/Foundation.h>
+#import <Matter/MTRDevice.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class MTRDeviceController;
```
bzbarsky-apple, line 23:
```suggestion
```

src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
bzbarsky-apple, line 1: I am really hoping this was generated with "git cp" followed by some changes...  but the changeset history does not look like that?

The reason to use "git cp" is to keep history. 

Ideally the "git cp" would just happen as a PR of its own, so that the squash-on-merge does not mess up history, hopefully.  Then the other changes on top of that.

kiel-apple: that actually makes a ton of sense and would be easy-ish now that we know what changes would be needed

kiel-apple: @bzbarsky-apple can you request changes so this doesn't land?  using `git cp` seems easy and shouldn't take long now that i have a working changeset to copy changes from

woody-apple (approved):

src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
kiel-apple: #34826

PR #34817, opened 2024-08-06
[ICD] Add ICDM 5.1 Automated Test Script

mkardous-silabs:
#### Description
Adds the ICDM 5.1 Test Case automated script.
Test Plan PR: https://github.com/CHIP-Specifications/chip-test-plans/pull/4432

#### Tests
CI


jmartinez-silabs (approved):

woody-apple (approved):

yunhanw-google (approved):

src/python_testing/TC_ICDM_5_1.py
```diff
@@ -0,0 +1,206 @@
...
+    clientType: ClientTypeEnum
+
+
+logger = logging.getLogger(__name__)
+kRootEndpointId = 0
+
+client1 = Client(
+    checkInNodeID=1,
+    subjectId=1,
+    key=b"\x10\x11\x12\x13\x14\x15\x16\x17\x18\x19\x1a\x1b\x1c\x1d\x1e\x1f",
```
andy31415, line 60: I wonder if this would be both shorter and more readable:

```py
binascii.a2b_hex('101112131415161718191a1b1c1d1e1f')
```

or to avoid typos:

```py
bytes([x for x in range(0x10, 0x20)])
```

src/python_testing/TC_ICDM_5_1.py
```diff
@@ -0,0 +1,206 @@
...
+import chip.clusters as Clusters
+from chip.interaction_model import InteractionModelError, Status
+from matter_testing_support import MatterBaseTest, TestStep, async_test_body, default_matter_test_main
+from mdns_discovery import mdns_discovery
+from mobly import asserts
+
+cluster = Clusters.Objects.IcdManagement
+commands = cluster.Commands
+attributes = cluster.Attributes
+modes = cluster.Enums.OperatingModeEnum
```
andy31415, line 42:
```suggestion
Cluster = Clusters.Objects.IcdManagement
Commands = Cluster.Commands
Attributes = Cluster.Attributes
OperatingModeEnum = Cluster.Enums.OperatingModeEnum
```

`lowercase` items look too much like modules or variables. We should keep consistent naming, just shorter.

src/python_testing/TC_ICDM_5_1.py
```diff
@@ -0,0 +1,206 @@
...
+    # Class Helper functions
+    #
+
+    async def _read_icdm_attribute_expect_success(self, attribute):
+        return await self.read_single_attribute_check_success(endpoint=kRootEndpointId, cluster=cluster, attribute=attribute)
+
+    async def _send_single_icdm_command(self, command):
+        return await self.send_single_cmd(command, endpoint=kRootEndpointId)
+
+    async def _get_icd_txt_record(self):
```
andy31415, line 77:
```suggestion
    async def _get_icd_txt_record(self) -> OperatingModeEnum:
```

and make it return the enum instead of an integer. This will make errors easier to read (you will see text instead of 1 != 2).

src/python_testing/TC_ICDM_5_1.py
```diff
@@ -0,0 +1,206 @@
...
+        for service in services[mdns_discovery.MdnsServiceType.OPERATIONAL.value]:
+
+            # Generate hex for the compressedFabric and NodeId
+            serviceName = service.instance_name.split("-")
+            compressedFabric = hex(int(serviceName[0], 16))
+            nodeId = hex(int(serviceName[1], 16))
+
+            # Find the correct MdnsServiceEntry
+            if localCompressedNodeId == compressedFabric and dutNodeId == nodeId:
+                icdTxtRecord = service.txt_record['ICD']
```
andy31415, line 98:
```suggestion
                icdTxtRecord = OperatingModeEnum(int(service.txt_record['ICD']))
                if icdTxtRecord.value != int(service.txt_record['ICD'])):
                    raise AttributeError(f'Not a known ICD type: {service.txt_record["ICD"]}')
```

src/python_testing/TC_ICDM_5_1.py
```diff
@@ -0,0 +1,206 @@
...
+)
+
+
+class TC_ICDM_5_1(MatterBaseTest):
+
+    #
+    # Class Helper functions
+    #
+
+    async def _read_icdm_attribute_expect_success(self, attribute):
```
andy31415, line 71: This should return a `ClientTypeEnum` as well

andy31415: For Operating mode only ... did not notice that we use this in several places.

andy31415: Apparently this already returns typed data.

src/python_testing/TC_ICDM_5_1.py
```diff
@@ -0,0 +1,206 @@
...
+            for client in registeredClients:
+                try:
+                    await self._send_single_icdm_command(commands.UnregisterClient(checkInNodeID=client.checkInNodeID))
+                except InteractionModelError as e:
+                    asserts.assert_equal(
+                        e.status, Status.Success, "Unexpected error returned")
+
+            self.step("2a")
+            operatingMode = await self._read_icdm_attribute_expect_success(attributes.OperatingMode)
+            asserts.assert_equal(operatingMode, modes.kSit.value)
```
andy31415, line 157: operatingMode seems to be already of type modes. No need for `.value` usage here or in the other places.

andy31415 (changes requested):
Requesting changes:
  - naming updates
  - do not use `.value` for enums as that is harder to debug when things fail

src/python_testing/TC_ICDM_5_1.py
andy31415: why are we not returning immediately as soon as we found something? Why keep looping to find other results?

src/python_testing/TC_ICDM_5_1.py
```diff
@@ -0,0 +1,206 @@
...
+
+    async def _send_single_icdm_command(self, command):
+        return await self.send_single_cmd(command, endpoint=kRootEndpointId)
+
+    async def _get_icd_txt_record(self):
+        discovery = mdns_discovery.MdnsDiscovery()
+        localCompressedNodeId = hex(self.default_controller.GetCompressedFabricId())
+        dutNodeId = hex(self.dut_node_id)
+
+        services = await discovery.get_all_services(log_output=True, discovery_timeout_sec=240)
```
andy31415, line 82: from what I can tell, get_all_services discovers until the first reply is received and it defaults to a large list of services. I could not find a good way to actually control the services we discover (and we probably should have something) but a workaround seems to be:

```py
        discovery._service_types = [
                f'{self.default_controller.GetCompressedFabricId():016x}-{self.dut_node_id:016x}._matter._tcp.local.'
        ]
```

This is actually quite bad though ... we should update mdns_discovery to have proper arguments control.

andy31415 (approved):
I have made the changes manually since I understand Mathieu is OOO currently.

PR #34818, opened 2024-08-06
adding mandatory endpointComposition data to XML for ZAP to consume

### LINKED ISSUE #32941: [Feature] XML Device Types define Endpoint Composition
### Feature description

ZAP utilizes the data model XML and has recently incorporated support for Parent Endpoints.

For ZAP to store Endpoint Composition into the database, it requires the Endpoint Composition to be specified in the Device Type XML.

For instance, the XML should accurately represent the specifications regarding Endpoint Composition. Notably, the current XML lacks any mention of the Temperature Controlled Cabinet being mandatory for a Refrigerator Device Type.

Screenshot 2024-03-05 at 11 15 38 AM

Referencing the "old XML" [here](https://github.com/project-chip/connectedhomeip/blob/master/src/app/zap-templates/zcl/data-model/chip/matter-devices.xml#L2017-L2031) and the "new XML" [here](https://github.com/project-chip/connectedhomeip/blob/7bf764ab05eb4ef52b007666e822ba963f643922/data_model/device_types/Refrigerator.xml), it's evident that we need the Device Type to be defined as either "Tree" or "FullFamily" if the Device Type includes Endpoint Composition.

### Platform

all

### Platform Version(s)

_No response_

### Anything else?

_No response_

paulr34:
Fixes #32941 Adds mandatory endpointComposition data to XML for ZAP to consume

### DIFF
--- a/src/app/zap-templates/zcl/data-model/chip/matter-devices.xml
+++ b/src/app/zap-templates/zcl/data-model/chip/matter-devices.xml
@@ -2210,7 +2210,7 @@ limitations under the License.
         <class>Simple</class>
         <scope>Endpoint</scope>
         <endpointComposition>
-            <compositionType fullFamily="false", tree="true"></compositionType>
+            <compositionType fullFamily="false", tree="true"/>
             <endpoint mandatory="true" constraint="min 1">
                     <deviceType>0x0071</deviceType>
             </endpoint>
--- a/src/app/zap-templates/zcl/data-model/chip/matter-devices.xml
+++ b/src/app/zap-templates/zcl/data-model/chip/matter-devices.xml
@@ -2209,6 +2209,12 @@ limitations under the License.
         <deviceId editable="false">0x0070</deviceId>
         <class>Simple</class>
         <scope>Endpoint</scope>
+        <endpointComposition>
+            <compositionType>tree</compositionType>
+            <endpoint mandatory="true" constraint="1">
+                    <deviceType>0x0071</deviceType>
+            </endpoint>
+        </endpointComposition>
         <clusters lockOthers="true">
             <include cluster="Identify" client="false" server="false" clientLocked="true" serverLocked="false"></include>
             <include cluster="Descriptor" client="false" server="true" clientLocked="true" serverLocked="true"></include>
--- a/src/app/zap-templates/zcl/data-model/chip/matter-devices.xml
+++ b/src/app/zap-templates/zcl/data-model/chip/matter-devices.xml
@@ -2210,7 +2210,7 @@ limitations under the License.
         <class>Simple</class>
         <scope>Endpoint</scope>
         <endpointComposition>
-            <compositionType fullFamily="false", tree="true"/>
+            <compositionType>tree</compositionType>
             <endpoint mandatory="true" constraint="min 1">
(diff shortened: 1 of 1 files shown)


src/app/zap-templates/zcl/data-model/chip/matter-devices.xml
```diff
@@ -2209,6 +2209,12 @@ limitations under the License.
         <deviceId editable="false">0x0070</deviceId>
         <class>Simple</class>
         <scope>Endpoint</scope>
+        <endpointComposition>
+            <compositionType>tree</compositionType>
+            <endpoint mandatory="true" constraint="1">
```
bzbarsky-apple, line 2214: What does `constraint="1"` mean here?

bzbarsky-apple: And don't we need to at least define the "full" composition bits in the XML?

paulr34: > What does `constraint="1"` mean here?

The idea is that you need 1
I understand things could get more complicated. Perhaps I change it to "min 1" and let ZAP consume exactly the value that is in the spec?

paulr34: > And don't we need to at least define the "full" composition bits in the XML?

ZAP consumes whatever is set as <compositionType>. It did not occur to me that it needed to be defined. I guess it would make sense to do <compositionType fullFamily = false tree = true> as an alternative. Either way is fine with me.

paulr34: @bzbarsky-apple I made some changes to address your concerns

src/app/zap-templates/zcl/data-model/chip/matter-devices.xml
```diff
@@ -2209,6 +2209,12 @@ limitations under the License.
         <deviceId editable="false">0x0070</deviceId>
         <class>Simple</class>
         <scope>Endpoint</scope>
+        <endpointComposition>
+            <compositionType fullFamily="false", tree="true"/>
```
bzbarsky-apple, line 2213: This is not valid XML.

Also, seems like `<endpointComposition compositionType="type">` or `<compositionType value="type">` or `<compositionType>type</compositionType>` would all be better than having two booleans that can both be set to true (which would be very wrong).

paulr34: Okay thank you

paulr34:
@bzbarsky-apple thanks for your feedback. I changed it back to `<compositionType>type</compositionType>` so fullFamily and tree cannot both be true. 

Additionally, I set the constraint to be the full text from the spec. Of course, I could create min/max columns in the db table and change the XML from constraint = "min 1" to have a min/max. My concern is that the constraints might be more complicated than simply a minimum and a maximum. 

Also, what do you mean by this? https://github.com/project-chip/connectedhomeip/pull/34818#discussion_r1705984669

bzbarsky-apple (approved):

jmartinez-silabs (approved):

PR #34819, opened 2024-08-06
python testing: Per node endpoint fixes

cecille:
- change name of decorator to hopefully be more obvious what it does
- change to force the use of a specific endpoint so that this is easier to use in the TH

Test: Running against CI, also see TestDecorators for new tests of override


src/python_testing/matter_testing_support.py
```diff
@@ -1859,15 +1859,21 @@ def has_feature(cluster: ClusterObjects.ClusterObjectDescriptor, feature: IntFla
 
 
 async def get_accepted_endpoints_for_test(self: MatterBaseTest, accept_function: EndpointCheckFunction) -> list[uint]:
-    """ Helper function for the per_endpoint_test decorator.
+    """ Helper function for the run_for_each_matching_endpoint decorator.
 
         Returns a list of endpoints on which the test should be run given the accept_function for the test.
     """
     wildcard = await self.default_controller.Read(self.dut_node_id, [()])
```
tcarmelveilleux, line 1866: Would reading only the descriptor on each endpoint suffice? Would be a much faster read.

src/python_testing/matter_testing_support.py
```diff
@@ -1859,15 +1859,21 @@ def has_feature(cluster: ClusterObjects.ClusterObjectDescriptor, feature: IntFla
...
-    return [e for e in wildcard.attributes.keys() if accept_function(wildcard, e)]
+    matching = [e for e in wildcard.attributes.keys() if accept_function(wildcard, e)]
+    forced_endpoint = self.user_params.get('force_endpoint', None)
```
tcarmelveilleux, line 1868: Suggest logging when a forced endpoint is used.

src/python_testing/TC_LVL_2_3.py
```diff
@@ -81,7 +81,7 @@ def steps_TC_LVL_2_3(self) -> list[TestStep]:
                          "The third entry in reportedRemainingTimeValuesList is equal to 0")
                 ]
 
-    @per_endpoint_test(has_cluster(Clusters.LevelControl))
+    @run_for_each_matching_endpoint(has_cluster(Clusters.LevelControl))
```
tcarmelveilleux, line 84: This should have docs about how `--int-arg forced_endpoint:N` is how you run on only one endpoint at a time for testing.

src/python_testing/matter_testing_support.py
cecille: yeah, probably would be faster. Thanks. Will change that.

cecille: oh, no, wait, it needs to be able to determine the feature map and attribute list for each cluster as well. So I need the full wildcard.

cecille: Maybe descriptor and globals would do though

cecille: I'm going to land this change separately as a follow up PR since it requires a change to the Read function that I'd like to ensure gets proper feedback.

src/python_testing/matter_testing_support.py
cecille: added.

src/python_testing/TC_LVL_2_3.py
cecille: Added to the assertion message.

yufengwangca (approved):

tcarmelveilleux (approved):

src/python_testing/matter_testing_support.py
```diff
@@ -1896,31 +1881,69 @@ def has_feature(cluster: ClusterObjects.ClusterObjectDescriptor, feature: IntFla
...
+            finally:
+                self.matter_test_config.endpoint = old_endpoint
+        return matching_runner
+    return run_on_singleton_matching_endpoint_internal
 
 
-def per_endpoint_test(accept_function: EndpointCheckFunction):
-    """ Test decorator for a test that needs to be run once per endpoint that meets the accept_function criteria.
+def run_if_endpoint_matches(accept_function: EndpointCheckFunction):
+    """ Test decorator for a test that needs to be run only if the endpoint that meets the accept_function criteria.
```
bzbarsky-apple, line 1944:
```suggestion
    """ Test decorator for a test that needs to be run only if the endpoint meets the accept_function criteria.
```

bzbarsky-apple (approved):
I did not review TestDecorators.py carefully, and might have missed subtleties in the core guts in matter_testing_support.py...

fabiowmm (approved):
LGTM.

PR #34820, opened 2024-08-06
Fixes for the color control  quiet reporting

### LINKED ISSUE #34618: [BUG] Color control cluster will fail to report changes when it should
### Reproduction steps

See https://github.com/project-chip/connectedhomeip/pull/34544#pullrequestreview-2206729927

### Bug prevalence

Always

bef33615e9b10185264974dee7b29e1c92ad8ace

### Platform

core

### Platform Version(s)

_No response_

### Anything else?

_No response_

### LINKED ISSUE #34643: [BUG] Color control Q - Remaining time appears to only report at 0
### Reproduction steps

See TC_CC_2.2

### Bug prevalence

always

3263e403469fd66827d923484a50ddb504544112

### Platform

core

### Platform Version(s)

_No response_

### Anything else?

_No response_

### LINKED ISSUE #34646: [BUG] Q color control too many reports
### Reproduction steps

Getting more reports than expected on transitions. looks like 1 per second even after transition is complete. Or something.

### Bug prevalence

always

3263e403469fd66827d923484a50ddb504544112

### Platform

core

### Platform Version(s)

_No response_

### Anything else?

_No response_

jmartinez-silabs:
fixes #34618
fixes #34646
fixes #34643

- MarkAttributeDirty::kYes instead of kIfChanged for QuietReporting to ensure the value is reported. 
- ~~Fix Start Or End of transition Quiet report to only report if the attribute value changed since the last quiet report~~
- ~~apply same tweak for start/end of the transition to the lvl control~~
- re-enable the disabled tc_cc_2_2 steps

Tested with fully enabled TC_CC_2.2.py

### DIFF
--- a/src/app/clusters/color-control-server/color-control-server.cpp
+++ b/src/app/clusters/color-control-server/color-control-server.cpp
@@ -3120,8 +3120,8 @@ void ColorControlServer::levelControlColorTempChangeCommand(EndpointId endpoint)
  * @param quietReporter: The QuieterReportingAttribute<TYPE> object for the attribute to update.
  * @param newValue: Value to update the attribute with
  * @param isStartOrEndOfTransition: Boolean that indicatse whether the update is occurring at the start or end of a level transition
- * @return MarkAttributeDirty::kYes when the attribute must be marked dirty and be reported. MarkAttributeDirty::kNo when it
- * when it no report is needed.
+ * @return MarkAttributeDirty::kYes when the attribute must be marked dirty and be reported. MarkAttributeDirty::kNo when
+ * no report is needed.
  */
 template <typename Q, typename V>
 MarkAttributeDirty ColorControlServer::SetQuietReportAttribute(QuieterReportingAttribute<Q> & quietReporter, V newValue,
@@ -3133,8 +3133,8 @@ MarkAttributeDirty ColorControlServer::SetQuietReportAttribute(QuieterReportingA
     if (isStartOrEndOfTransition)
     {
         // At the start or end of the movement/transition we must report if the value changed
-        auto predicate = [](const typename QuieterReportingAttribute<Q>::SufficientChangePredicateCandidate & candidate) -> bool {
-            return (candidate.lastDirtyValue != candidate.newValue);
+        auto predicate = [](const typename QuieterReportingAttribute<Q>::SufficientChangePredicateCandidate &) -> bool {
+            return true;
         };
         dirtyState = quietReporter.SetValue(newValue, now, predicate);
     }
--- a/src/app/clusters/color-control-server/color-control-server.cpp
+++ b/src/app/clusters/color-control-server/color-control-server.cpp
@@ -3113,14 +3113,14 @@ void ColorControlServer::levelControlColorTempChangeCommand(EndpointId endpoint)
  * - When it changes from null to any other value and vice versa. (Implicit to the QuieterReportingAttribute class)
  *
  * The QuietReportAttribute class is updated with the new value and when the report conditions are met,
- * this function will return MarkAttributeDirty::kIfChanged.
+ * this function will return MarkAttributeDirty::kYes.
  * It is expected that the user will use this return value to trigger a reporting mechanism for the attribute with the new value
  * (Which was updated in the quietReporter)
  *
  * @param quietReporter: The QuieterReportingAttribute<TYPE> object for the attribute to update.
  * @param newValue: Value to update the attribute with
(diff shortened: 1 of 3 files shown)


src/app/clusters/level-control/level-control.cpp
```diff
@@ -545,7 +545,7 @@ static void writeRemainingTime(EndpointId endpoint, uint16_t remainingTimeMs)
             markDirty = MarkAttributeDirty::kYes;
         }
 
-        Attributes::RemainingTime::Set(endpoint, state->quietRemainingTime.value().ValueOr(0), markDirty);
+        Attributes::RemainingTime::Set(endpoint, state->quietRemainingTime.value().Value(), markDirty);
```
cecille, line 548: Is this guaranteed to be set now? Before this would default to 0 is has_value is false, but now if you don't have a value here, you'll crash.

jmartinez-silabs: Yes, It is set above line 543. 
Because `quietRemainingTime.mValue` is init as null, I had initially used ValueOr(0). But it was pointed out in an after-merge review that this will never be null because we just set it to `remainingTimeDs`

cecille (approved):

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -3113,14 +3113,14 @@ void ColorControlServer::levelControlColorTempChangeCommand(EndpointId endpoint)
...
+ * this function will return MarkAttributeDirty::kYes.
  * It is expected that the user will use this return value to trigger a reporting mechanism for the attribute with the new value
  * (Which was updated in the quietReporter)
  *
  * @param quietReporter: The QuieterReportingAttribute<TYPE> object for the attribute to update.
  * @param newValue: Value to update the attribute with
  * @param isStartOrEndOfTransition: Boolean that indicatse whether the update is occurring at the start or end of a level transition
- * @return MarkAttributeDirty::kIfChanged when the attribute must be maredk dirty and be reported. MarkAttributeDirty::kNo when it
+ * @return MarkAttributeDirty::kYes when the attribute must be marked dirty and be reported. MarkAttributeDirty::kNo when it
  * when it no report is needed.
```
bzbarsky-apple, line 3124:
```suggestion
 * @return MarkAttributeDirty::kYes when the attribute must be marked dirty and be reported. MarkAttributeDirty::kNo when
 * no report is needed.
```

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -3132,9 +3132,9 @@ MarkAttributeDirty ColorControlServer::SetQuietReportAttribute(QuieterReportingA
 
     if (isStartOrEndOfTransition)
     {
-        // At the start or end of the movement/transition we must report
-        auto predicate = [](const typename QuieterReportingAttribute<Q>::SufficientChangePredicateCandidate &) -> bool {
-            return true;
+        // At the start or end of the movement/transition we must report if the value changed
+        auto predicate = [](const typename QuieterReportingAttribute<Q>::SufficientChangePredicateCandidate & candidate) -> bool {
+            return (candidate.lastDirtyValue != candidate.newValue);
```
bzbarsky-apple, line 3137: Predicates are only called `if (newValue != mLastDirtyValue)`, so I don't think this change is needed...   As in, in this predicate `candidate.lastDirtyValue != candidate.newValue` is in fact always `true`.

src/app/clusters/level-control/level-control.cpp
```diff
@@ -391,8 +391,8 @@ static Status SetCurrentLevelQuietReport(EndpointId endpoint, EmberAfLevelContro
     if (isStartOrEndOfTransition)
     {
         // At the start or end of the movement/transition we must report
-        auto predicate = [](const decltype(state->quietCurrentLevel)::SufficientChangePredicateCandidate &) -> bool {
-            return true;
+        auto predicate = [](const decltype(state->quietCurrentLevel)::SufficientChangePredicateCandidate & candidate) -> bool {
+            return (candidate.lastDirtyValue != candidate.newValue);
```
bzbarsky-apple, line 395: Again, this change is not needed.

src/app/clusters/color-control-server/color-control-server.cpp
jmartinez-silabs: Yes, you are right, it went over my head! I guess it is not needed in the provided predicate `GetPredicateForSufficientTimeSinceLastDirty` either

bzbarsky-apple (approved):

PR #34823, opened 2024-08-06
[HVAC] Allow Thermostat delegate to supply atomic write timeout

### LINKED ISSUE #34740: [Feature] Make thermostat cluster atomic write time clamp configurable
### Feature description

After https://github.com/project-chip/connectedhomeip/pull/34570, thermostat cluster hardcodes a 5s cap on the atomic write timeout.  This should probably be configurable by the application.

### Platform

core

### Platform Version(s)

_No response_

### Anything else?

_No response_

hasty:
Fixes #34740


examples/thermostat/linux/thermostat-delegate-impl.cpp
```diff
@@ -148,6 +148,44 @@ CHIP_ERROR ThermostatDelegate::SetActivePresetHandle(const DataModel::Nullable<B
...
+        return System::Clock::Milliseconds16(0);
+    }
+    auto timeout = System::Clock::Milliseconds16(0);
+    if (requestedPresets)
+    {
+        timeout += std::chrono::milliseconds(1000);
+    }
+    if (requestedSchedules)
+    {
+        timeout += std::chrono::milliseconds(3000);
```
bzbarsky-apple, line 184: This could use some documentation explaining where these numbers are coming from and what they mean.

Also:

1. I suspect in practice on a somewhat-congested Thread network those numbers might be low.  We should put numbers here that will work ok if someone just blindly copies this whole delegate implementation.
2. It would be good to have some documentation at the top of the method explaining what this method _has_ to do vs what we are just doing as an example of things we _could_ do.

examples/thermostat/linux/thermostat-delegate-impl.cpp
```diff
@@ -148,6 +148,44 @@ CHIP_ERROR ThermostatDelegate::SetActivePresetHandle(const DataModel::Nullable<B
     return CHIP_NO_ERROR;
 }
 
+System::Clock::Milliseconds16
+ThermostatDelegate::GetAtomicWriteTimeout(DataModel::DecodableList<chip::AttributeId> attributeRequests,
```
bzbarsky-apple, line 152:
```suggestion
ThermostatDelegate::GetAtomicWriteTimeout(DataModel::DecodableList<AttributeId> attributeRequests,
```
This has `namespace chip` in scope already.

src/app/clusters/thermostat-server/thermostat-delegate.h
```diff
@@ -38,6 +38,16 @@ class Delegate
...
     virtual ~Delegate() = default;
 
+    /**
+     * @brief Get the maximum timeout for atomically writing to a set of attributes
+     *
+     * @param[in] attributeRequests The list of attributes to write to.
+     * @param[out] timeoutRequest The timeout proposed by the client.
+     * @return The maximum allowed timeout; zero if the request is invalid.
+     */
+    virtual System::Clock::Milliseconds16 GetAtomicWriteTimeout(DataModel::DecodableList<chip::AttributeId> attributeRequests,
```
bzbarsky-apple, line 48:
```suggestion
    virtual System::Clock::Milliseconds16 GetAtomicWriteTimeout(DataModel::DecodableList<AttributeId> attributeRequests,
```

bzbarsky-apple (approved):

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -116,8 +118,7 @@ void TimerExpiredCallback(System::Layer * systemLayer, void * callbackContext)
     VerifyOrReturn(delegate != nullptr, ChipLogError(Zcl, "Delegate is null. Unable to handle timer expired"));
 
     delegate->ClearPendingPresetList();
-    gThermostatAttrAccess.SetAtomicWrite(endpoint, false);
-    gThermostatAttrAccess.SetAtomicWriteScopedNodeId(endpoint, ScopedNodeId());
+    gThermostatAttrAccess.SetAtomicWrite(endpoint, ScopedNodeId(), false);
```
andy31415, line 121:
```suggestion
    gThermostatAttrAccess.SetAtomicWrite(endpoint, ScopedNodeId(), false /* inProgress */);
```

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -205,8 +206,7 @@ void resetAtomicWrite(Delegate * delegate, EndpointId endpoint)
         delegate->ClearPendingPresetList();
     }
     ClearTimer(endpoint);
-    gThermostatAttrAccess.SetAtomicWrite(endpoint, false);
-    gThermostatAttrAccess.SetAtomicWriteScopedNodeId(endpoint, ScopedNodeId());
+    gThermostatAttrAccess.SetAtomicWrite(endpoint, ScopedNodeId(), false);
```
andy31415, line 209:
```suggestion
    gThermostatAttrAccess.SetAtomicWrite(endpoint, ScopedNodeId(), false /* inProgress */);
```

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -766,7 +757,7 @@ CHIP_ERROR ThermostatAttrAccess::Read(const ConcreteReadAttributePath & aPath, A
     }
     break;
     case ActivePresetHandle::Id: {
-        Delegate * delegate = GetDelegate(aPath.mEndpointId);
+        auto delegate = GetDelegate(aPath.mEndpointId);
```
andy31415, line 760: Revert these type changes to keep the diff small?

andy31415: Or did you need to change anyway because of Thermostat:: namespace requirement?

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1412,13 +1418,17 @@ void handleAtomicBegin(CommandHandler * commandObj, const ConcreteCommandPath &
...
-    uint16_t maxTimeout = 5000;
-    timeout             = std::min(timeout, maxTimeout);
+    System::Clock::Milliseconds16 timeout =
+        delegate->GetAtomicWriteTimeout(commandData.attributeRequests, System::Clock::Milliseconds16(commandData.timeout.Value()));
 
-    ScheduleTimer(endpoint, System::Clock::Milliseconds16(timeout));
-    gThermostatAttrAccess.SetAtomicWrite(endpoint, true);
-    gThermostatAttrAccess.SetAtomicWriteScopedNodeId(endpoint, GetSourceScopedNodeId(commandObj));
-    sendAtomicResponse(commandObj, commandPath, imcode::Success, imcode::Success, imcode::Success, MakeOptional(timeout));
+    if (timeout == System::Clock::Milliseconds16(0))
```
andy31415, line 1424: This error checking seems to lose the reason for the failure. Should we maybe make this CHIP_ERROR + out param or log inside GetAtomicWriteTimeout?

andy31415: I would also consider std::optional for the timeout, since "Timeout of 0 is invalid" feels a bit magic

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -1412,13 +1418,17 @@ void handleAtomicBegin(CommandHandler * commandObj, const ConcreteCommandPath &
...
+    {
+        commandObj->AddStatus(commandPath, imcode::InvalidCommand);
+        return;
+    }
+    ScheduleTimer(endpoint, timeout);
+    gThermostatAttrAccess.SetAtomicWrite(endpoint, GetSourceScopedNodeId(commandObj), true);
```
andy31415, line 1430:
```suggestion
    gThermostatAttrAccess.SetAtomicWrite(endpoint, GetSourceScopedNodeId(commandObj), true /* inProgress */);
```

I would also consider avoiding bools for anything that are not single arguments like `SetEnabled(bool)`. A enum probably takes as much space and is much easier to read.

src/app/clusters/thermostat-server/thermostat-server.h
```diff
@@ -105,10 +97,18 @@ class ThermostatAttrAccess : public chip::app::AttributeAccessInterface
...
-    CHIP_ERROR AppendPendingPreset(Delegate * delegate, const Structs::PresetStruct::Type & preset);
+    CHIP_ERROR AppendPendingPreset(Thermostat::Delegate * delegate, const Structs::PresetStruct::Type & preset);
+
+    void OnFabricRemoved(const FabricTable & fabricTable, FabricIndex fabricIndex) override;
+
+    struct AtomicWriteState
+    {
+        bool inProgress;
+        ScopedNodeId nodeId;
+        EndpointId endpointId;
```
andy31415, line 108:
```suggestion
        bool inProgress = false;
        ScopedNodeId nodeId;
        EndpointId endpointId = kInvalidEndpointId;
```

or does this generate too much extra code? Generally I would try to initialize everything to known defaults.

andy31415 (approved):

src/app/clusters/thermostat-server/thermostat-server.h
hasty: Added initializers.

src/app/clusters/thermostat-server/thermostat-server.cpp
hasty: Switched to an enum

src/app/clusters/thermostat-server/thermostat-server.cpp
hasty: Switched to enum

src/app/clusters/thermostat-server/thermostat-server.cpp
hasty: Yeah, this is the actual namespace conflict.

src/app/clusters/thermostat-server/thermostat-server.cpp
hasty: Switched to std::optional.

src/app/clusters/thermostat-server/thermostat-server.cpp
hasty: Switched to enum.

PR #34830, opened 2024-08-07
Add chip_logging_backend option and 'none' and 'syslog' backends

ksperling-apple:
`chip_logging_backend` controls which backend is pulled in by the `src/platform/logging:default` target. The default is 'platform', retaining the current behavior.

On Darwin, remove the no-op `LoggingImpl` and make stdio the default backend when compiling tools or example apps.
Use the new 'none' backend when building Matter.framework, retaining current behavior.

Depend on `logging:default` instead of `logging:stdio` for linux:app-main examples so the backend can be selected at build time via `chip_logging_backend`.

### DIFF
--- a/src/darwin/Framework/chip_xcode_build_connector.sh
+++ b/src/darwin/Framework/chip_xcode_build_connector.sh
@@ -103,6 +103,7 @@ declare -a args=(
     'chip_enable_python_modules=false'
     'chip_device_config_enable_dynamic_mrp_config=true'
     'chip_log_message_max_size=4096' # might as well allow nice long log messages
+    'chip_logging_backend="none"'    # os_log() is integrated via CHIP_SYSTEM_CONFIG_PLATFORM_LOG
     'chip_disable_platform_kvs=true'
     'enable_fuzz_test_targets=false'
     "target_cpu=\"$target_cpu\""
--- a/src/lib/core/core.gni
+++ b/src/lib/core/core.gni
@@ -60,9 +60,9 @@ declare_args() {
 
 declare_args() {
   # Logging backend to use for targets that don't link a specific log
-  # backend (e.g. a command line utilites usually use 'stdio'). Options:
+  # backend (e.g. command line utilites usually use 'stdio'). Options:
   #   'platform' - The default log backend of the device platform
-  #   'external' - Externally LogV implementation (src/platform/logging/LogV.h)
+  #   'external' - External LogV implementation (src/platform/logging/LogV.h)
   #   'none'     - Discard all log output
   #   'stdio'    - Print to stdout
   #   'syslog'   - POSIX syslog()
--- a/src/lib/core/core.gni
+++ b/src/lib/core/core.gni
@@ -65,6 +65,7 @@ declare_args() {
   #   'external' - Externally LogV implementation (src/platform/logging/LogV.h)
   #   'none'     - Discard all log output
   #   'stdio'    - Print to stdout
+  #   'syslog'   - POSIX syslog()
   if (chip_use_external_logging) {
     chip_logging_backend = "external"
   } else {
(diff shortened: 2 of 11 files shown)


src/darwin/Framework/chip_xcode_build_connector.sh
```diff
@@ -103,6 +103,7 @@ declare -a args=(
     'chip_enable_python_modules=false'
     'chip_device_config_enable_dynamic_mrp_config=true'
     'chip_log_message_max_size=4096' # might as well allow nice long log messages
+    'chip_logging_backend="none"'    # os_log() is integrated via CHIP_SYSTEM_CONFIG_PLATFORM_LOG
```
bzbarsky-apple, line 106: So with this change, what happens with framework logs in CI?  For example when running the darwin unit tests.

Similarly, what happens with logs while running darwin-framework-tool?  Will logs from the tool itself go to stdout but not logs from the framework it links against?

src/lib/core/core.gni
```diff
@@ -54,7 +54,23 @@ declare_args() {
   # Configure chip logging to output through external logging implementation.
   # External code will need to provide implementation for CHIP log output
   # function (LogV), which is defined in "src/platform/logging/LogV.h".
+  # Same as setting chip_logging_backend = "external"
   chip_use_external_logging = false
+}
+
+declare_args() {
+  # Logging backend to use for targets that don't link a specific log
+  # backend (e.g. a command line utilites usually use 'stdio'). Options:
```
bzbarsky-apple, line 63:
```suggestion
  # backend (e.g. command line utilites usually use 'stdio'). Options:
```

src/lib/core/core.gni
```diff
@@ -54,7 +54,23 @@ declare_args() {
...
+  #   'platform' - The default log backend of the device platform
+  #   'external' - Externally LogV implementation (src/platform/logging/LogV.h)
```
bzbarsky-apple, line 65:
```suggestion
  #   'external' - External LogV implementation (src/platform/logging/LogV.h)
```

src/darwin/Framework/chip_xcode_build_connector.sh
ksperling-apple: Hmm I'll have to look into that... but my initial answer would be "whatever happens now" because the current LoggingImpl for Darwin was also a no-op.

ksperling-apple: darwin-framework-tool installs log redirect callbacks into the framework and into the copy of CHIP logging that's included directly, so that will continue to work as before.

Framework logs in unit tests go into os_log and get picked up from there by xctest as before.

bzbarsky-apple (approved):

src/platform/logging/impl/Syslog.cpp
```diff
@@ -0,0 +1,65 @@
...
+    }
+}
+} // namespace
+
+void LogV(const char * module, uint8_t category, const char * msg, va_list v)
+{
+    static std::mutex sMutex;
+    static bool sInitialized = false;
+    static char sBuffer[CHIP_CONFIG_LOG_MESSAGE_MAX_SIZE];
+    std::lock_guard guard(sMutex);
```
arkq, line 50: How about using `std::call_once` to initialize syslog?

arkq (approved):
Many thanks for that!

src/platform/logging/impl/Syslog.cpp
ksperling-apple: I would still need the lock anyway to guard the buffer unless I put it on the stack, and it's maybe too large for that (given that logging is likely to be called from deep call stacks anyway) -- at least on the unity platforms we currently use a size of 1708.

PR #34833, opened 2024-08-07
python: Add direct attribute paths to Read

cecille:
Supports one particular use case: read one or all endpoints, all clusters, specific (global) attribute. See spec 8.9.2.4. This is an allowed wildcard construct that is not currently expressible in the API.

Test: Used on wildcard read for matter_testing_support. This is
      therefore tested on any test using that decorator - switch
      and timesync.


woody-apple (approved):

woody-apple:
Fast tracking testing tools, given where we are.

Issue #34686, opened 2024-07-31
[SVE] Occupancy sensing cluster test script needs CI addition to assimilate manual test cases.


jaehs6sam:
### Reproduction steps

Python test script files, TC_OCC_3_1.py and TC_OCC_3_1.py need CI integration to assimilate an manual intervention of triggering occupancy sensor occupancy state ON and OFF.  This issue will track PR 34572 (https://github.com/project-chip/connectedhomeip/pull/34572) to fulfill CI integration needed.

### Bug prevalence

CI is not accommodated yet.

### GitHub hash of the SDK that was being used

No hash information yet.

### Platform

raspi

### Type

Test Improvement

### Testing

CI tested

### (Optional) If manually tested please explain why this is only manually tested

Intension to perform manual testing is to test occupancy ON/OFF functionality.  Thus CI to assimilate manual testing needs to be integrated for automated testing.

### Anything else?

Additionally, feature bitmap conditional statement in the script file TC_OCC_2_2.py will be further clarified and updated for better readability. Will try to triage and resolve these issues before 8/2.

jaehs6sam:
This issue was addressed by PR #34836 merging.

PR #34836, opened 2024-08-07
TC OCC 3.1 and 3.2 CI Implementation

jaehs6sam:
This PR is to implement CI for TC OCC 3.1 and 3.2 and upon approval, this should take care of issue #34686.


examples/all-clusters-app/linux/AllClustersCommandDelegate.cpp
```diff
@@ -769,6 +784,45 @@ void AllClustersAppCommandHandler::OnAirQualityChange(uint32_t aNewValue)
...
+
+    Protocols::InteractionModel::Status status = OccupancySensing::Attributes::Occupancy::Set(endpointId, occupancyValue);
+    ChipLogDetail(NotSpecified, "Set Occupancy attribute to %u", occupancyValue);
+
+    if (status != Protocols::InteractionModel::Status::Success)
+    {
+        ChipLogDetail(NotSpecified, "Invalid value/endpoint to set.");
+    }
+
+    if (1 == occupancyValue && 1 == endpointId)
```
tcarmelveilleux, line 798: Why is there a if on `1 == endpointId` here ?

examples/all-clusters-app/linux/AllClustersCommandDelegate.cpp
```diff
@@ -769,6 +784,45 @@ void AllClustersAppCommandHandler::OnAirQualityChange(uint32_t aNewValue)
...
+    {
+        uint16_t * holdTime = chip::app::Clusters::OccupancySensing::GetHoldTimeForEndpoint(endpointId);
+        if (holdTime != nullptr)
+        {
+            CHIP_ERROR err = chip::DeviceLayer::SystemLayer().StartTimer(
+                chip::System::Clock::Seconds16(*holdTime), AllClustersAppCommandHandler::OccupancyPresentTimerHandler, nullptr);
+            ChipLogDetail(NotSpecified, "Start HoldTime timer");
+            if (CHIP_NO_ERROR != err)
+            {
+                ChipLogError(NotSpecified, "Failed to start HoldTime timer.");
```
tcarmelveilleux, line 808: That has to be implemented as a building block

src/python_testing/TC_OCC_3_2.py
```diff
@@ -110,35 +128,52 @@ async def test_TC_OCC_3_2(self):
...
-        self.wait_for_user_input(
-            prompt_msg="Type any letter and press ENTER after the sensor occupancy is triggered and its occupancy state changed.")
+        # CI call to trigger on
+        if self.is_ci:
+            self.write_to_app_pipe('{"Name":"SetOccupancy", "EndpointId": 1, "Occupancy": 1}')
+            initial_dut = await self.read_occ_attribute_expect_success(attribute=attributes.Occupancy)
+            time.sleep(2)
+            # will write one more to consier priming report effect of the previous CI writing
+            self.write_to_app_pipe('{"Name":"SetOccupancy", "EndpointId": 1, "Occupancy": 0}')
+            diff_val = await self.read_occ_attribute_expect_success(attribute=attributes.Occupancy)
```
tcarmelveilleux, line 151: This makes too many low-level assumptions. The CI should only replace prompts with single individual action equivalent to the prompt. This is separated test logic so may have separate bugs

src/python_testing/TC_OCC_3_1.py
```diff
@@ -1,129 +1,160 @@
...
+        hold_time = 10  # 10 seconds for occupancy state hold time
+
+        # CI app pipe id creation
+        self.app_pipe = "/tmp/chip_all_clusters_fifo_"
+        self.is_ci = self.check_pics("PICS_SDK_CI_ONLY")
+        if self.is_ci:
+            app_pid = self.matter_test_config.app_pid
+            if app_pid == 0:
+                asserts.fail("The --app-pid flag must be set when PICS_SDK_CI_ONLY is set.c")
+            self.app_pipe = self.app_pipe + str(app_pid)
```
tcarmelveilleux, line 92: - This needs to be moved to write_to_app_pipe for now.
- Let's change `is_ci` to be an argument `--bool-arg simulate_occupancy:true` so that it's easier to test. This can be added to the command line.

src/python_testing/TC_OCC_3_1.py
```diff
@@ -1,129 +1,160 @@
...
+
+        cluster = Clusters.OccupancySensing
+        attributes = cluster.Attributes
+        attribute_list = await self.read_occ_attribute_expect_success(attribute=attributes.AttributeList)
+
+        has_hold_time = attributes.HoldTime.attribute_id in attribute_list
+
+        if has_hold_time:
+            # write 10 as a HoldTime attribute
+            await self.write_single_attribute(cluster.Attributes.HoldTime(hold_time))
```
tcarmelveilleux, line 106: Don't use 10, it may be out of bounds. Use the HoldTimeLimits.HoldTimeMin. Let's make the examples use a HoldTimeMin of 2.

I think the test is written racy.

src/python_testing/TC_OCC_3_2.py
```diff
@@ -1,204 +1,240 @@
...
+        # add Namepiped to assimilate the manual sensor untrigger here
+        self.step("3a")
+        # CI call to trigger off
+        if self.is_ci:
+            self.write_to_app_pipe('{"Name":"SetOccupancy", "EndpointId": 1, "Occupancy": 0}')
+        else:
+            self.wait_for_user_input(prompt_msg="Type any letter and press ENTER after DUT goes back to unoccupied state.")
+
+        self.step("3b")
+        if attributes.Occupancy.attribute_id in attribute_list:
```
tcarmelveilleux, line 144: This is always true, so let's remove this if.

src/python_testing/TC_OCC_3_1.py
```diff
@@ -1,129 +1,160 @@
...
+                # CI call to trigger off
+                if self.is_ci:
+                    self.write_to_app_pipe('{"Name":"SetOccupancy", "EndpointId": 1, "Occupancy": 0}')
+                else:
+                    self.wait_for_user_input(
+                        prompt_msg="Type any letter and press ENTER after the sensor occupancy is detection ready state (occupancy attribute = 0)")
+
+        # check sensor occupancy state is 0 for the next test step
+        occupancy_dut = await self.read_occ_attribute_expect_success(attribute=attributes.Occupancy)
+        asserts.assert_equal(occupancy_dut, 0, "Occupancy attribute is still 1.")
```
tcarmelveilleux, line 129: Let's revisit the test plan to test the occupancy subscription updates in 3.1, not 3.2 and keep 3.2 for timing attribute subscription reporting.

src/python_testing/TC_OCC_3_2.py
```diff
@@ -1,204 +1,240 @@
...
+        if attributes.HoldTime.attribute_id not in attribute_list:
+            logging.info("No HoldTime attribute supports. Terminate this test case")
+            self.skip_all_remaining_steps("4b")
+
+        self.step("4b")
+        initial_dut = await self.read_occ_attribute_expect_success(attribute=attributes.HoldTime)
+
+        self.step("4c")
+        # write a different a HoldTime attribute value
+        diff_val = 12
```
tcarmelveilleux, line 171: -12 may not be a valid value. Start by writing HoldTimeMin and checking via a *read* that it is OK, then reset the attribute change accumulator before writing HoldTime to HoldTimeMax.
- Then change other attributes to HoldTimeMax and validate they tracked

We are now at the following steps:

- Set HoldTime to HoldTimeMin
- Read HoldTime and check it's HoldTimeMin
- Set to HoldTime HoldTimeMax
  - Check you saw update on the applicable legacy attributes to HoldTimeMax
- For of [PIR, US, PHY] timing attributes, if supported (blocks of steps)
  - Set HoldTime to HoldTimeMin
  - Clear attribute recording history
  - Use [PIR|US|PHY]OccupiedToUnoccupiedDelay to set value HoldTimeMax
  - Check that HoldTime got updated

src/python_testing/TC_OCC_3_1.py
jaehs6sam: Used 10 for enough delay.

src/python_testing/TC_OCC_3_1.py
```diff
@@ -1,129 +1,160 @@
...
+
+        self.step(4)
+        # CI call to trigger on
+        if self.is_ci:
+            self.write_to_app_pipe('{"Name":"SetOccupancy", "EndpointId": 1, "Occupancy": 1}')
+        else:
+            # Trigger occupancy sensor to change Occupancy attribute value to 1 => TESTER ACTION on DUT
+            self.wait_for_user_input(prompt_msg="Type any letter and press ENTER after a sensor occupancy is triggered.")
```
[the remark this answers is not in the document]
jaehs6sam: I need to add time delay after self.write_to_app_pipe('{"Name":"SetOccupancy", "EndpointId": 1, "Occupancy": 1}') or 10 sec hold time is good enough.

jaehs6sam (commented):
modified 3.1 test case according to comments.

.github/workflows/tests.yaml
```diff
@@ -506,7 +506,9 @@ jobs:
               run: |
                   mkdir -p out/trace_data
                   scripts/run_in_python_env.sh out/venv './scripts/tests/run_python_test.py --load-from-env /tmp/test_env.yaml --script src/controller/python/test/test_scripts/mobile-device-test.py'
+
```
tcarmelveilleux, line 509:
```suggestion
```

.github/workflows/tests.yaml
```diff
@@ -506,7 +506,9 @@ jobs:
...
                   scripts/run_in_python_env.sh out/venv 'python3 ./src/python_testing/execute_python_tests.py --env-file /tmp/test_env.yaml --search-directory src/python_testing'
+
```
tcarmelveilleux, line 511:
```suggestion
```

src/python_testing/TC_OCC_3_1.py
jaehs6sam: I added as follows.
self.is_ci = self.matter_test_config.global_test_params['simulate_occupancy']

src/python_testing/TC_OCC_3_2.py
jaehs6sam: removed it.

src/python_testing/TC_OCC_3_1.py
jaehs6sam: Mostly so.

src/python_testing/TC_OCC_3_2.py
jaehs6sam: Both reciprocal attribute testing has been added but maybe not exactly what you inferred here.

tcarmelveilleux (approved):

examples/all-clusters-app/linux/AllClustersCommandDelegate.cpp
```diff
@@ -769,6 +803,71 @@ void AllClustersAppCommandHandler::OnAirQualityChange(uint32_t aNewValue)
...
+
+    if (status != Protocols::InteractionModel::Status::Success)
+    {
+        ChipLogDetail(NotSpecified, "Invalid value/endpoint to set.");
+        return;
+    }
+
+    EmitOccupancyChangedEvent(endpointId, newOccupancyValue);
+
+    if (1 == newOccupancyValue)
```
jmartinez-silabs, line 828: Using the enum value would be easier to follow

jmartinez-silabs (approved):

examples/all-clusters-app/linux/AllClustersCommandDelegate.cpp
jaehs6sam: @jmartinez-silabs Would this comment require some follow up action?  Otherwise, may I close the conversation?

tcarmelveilleux: Will follow-up

PR #34839, opened 2024-08-07
Remove timeSyncLock from MTRDevice.

bzbarsky-apple:
We should just use "lock" for all the device state.  As it was, we were actually accessing some state that is protected by "lock" while only holding "timeSyncLock".


woody-apple (approved):

bzbarsky-apple:
This change is wrong, looks like: it can lead to deadlocks because the time sync stuff uses public APIs that try to take the main data lock.

PR #34842, opened 2024-08-07
Make it possible to use lock-protected state in MTRDevice_Concrete's description.

bzbarsky-apple:
We need _descriptionWhileLocked for use from places where the lock is already held.


bzbarsky-apple:
Per conversation with @jtung-apple we came up with a less brittle approach: have any state `description` needs be stored separately, protected by a separate lock, and get updated while both that lock and the overall state lock are held.

PR #34851, opened 2024-08-07
Add support for more attributes in Bridged Device Basic Information cluster in FabricBridge example

andy31415:
This builds on top of #34845 to add actual functionality for extra attribute processing in the fabric bridge example.

### Changes

- update the RPC for add/remove device to include the initial values of bridged device attributes (with some fake data to start with)
- Updated RPC struct to default-initialize otherwise all strings (and numbers for that matter) are filled with garbage
- Added a unique id generation scheme if a unique ID is not available yet
- Added handlers for attribute read for all new attributes
- Use `std::unique_ptr` for memory management in the BridgedDeviceManager, which looks like it solves some potential memory leaks and makes pointer/memory ownership clear.

image


examples/common/pigweed/protos/fabric_bridge_service.proto
```diff
@@ -7,6 +7,17 @@ package chip.rpc;
...
+  string unique_id = 2;
+  string vendor_name = 3;
+  uint32 vendor_id = 4;
+  string product_name = 5;
+  uint32 product_id = 6;
+  string node_label = 7;
+  uint32 hardware_version = 8;
+  string hardware_version_string = 9;
+  uint32 software_version = 10;
+  string software_version_string = 11;
```
tehampson, line 20: Would you also be able to plumb in if the device is ans ICD device? It is okay if you cannot, but I was hoping that ultimately whatever is calling AddSynchronizedDevice would also provide this

examples/common/pigweed/protos/fabric_bridge_service.proto
```diff
@@ -7,6 +7,17 @@ package chip.rpc;
 // Define the message for a synchronized end device with necessary fields
 message SynchronizedDevice {
   uint64 node_id = 1;
+
+  string unique_id = 2;
+  string vendor_name = 3;
+  uint32 vendor_id = 4;
+  string product_name = 5;
+  uint32 product_id = 6;
```
tehampson, line 15: Is there a way to say these are optional since these are only needed when calling add, and not needed when calling remove

examples/common/pigweed/protos/fabric_bridge_service.options
```diff
@@ -0,0 +1,6 @@
+chip.rpc.SynchronizedDevice.unique_id max_size:33
```
tehampson, line 1: uniqueid can be uptto 64 bytes I thought. Some of these sizes seems to be incorrect

examples/common/pigweed/protos/fabric_bridge_service.proto
andy31415: Proto3 should auto-fill them with defaults if they are not sent over the wire. Will add optionality to them to make it clear (I think in that case we will have some has_ items).

examples/common/pigweed/protos/fabric_bridge_service.options
andy31415: Basic information cluster only has 32 in my copy of the spec:

image

examples/common/pigweed/protos/fabric_bridge_service.proto
andy31415: Updated to optional items in RPC and updated code to take into consideration the `has` values.

examples/common/pigweed/protos/fabric_bridge_service.proto
andy31415: I would like to do that in some ICD-specific followups as the code paths are likely somewhat different (I think ICD seems to affect feature map, which in turn makes a command available that is not there otherwise)

examples/common/pigweed/protos/fabric_bridge_service.options
```diff
@@ -0,0 +1,6 @@
...
+chip.rpc.SynchronizedDevice.vendor_name max_size:33
+chip.rpc.SynchronizedDevice.product_name max_size:33
+chip.rpc.SynchronizedDevice.node_label max_size:33
+chip.rpc.SynchronizedDevice.hardware_version_string max_size:33
```
tehampson, line 5: hardware string can be up to 64 same with sw version string

image

andy31415: missed those. Will correct this...z

andy31415: Fixed.

tehampson (approved):

woody-apple (approved):

PR #34854, opened 2024-08-07
Marshal the data from BasicInformationCluster to FabricBridge

andy31415:
This builds on #34851  adds the flow of:
  - Read BasicInformation Cluster
  - Record received attributes
  - Send these attributes to FabricBridge so that the initial version of the fabric bridge contains actual real data

image


examples/fabric-admin/commands/pairing/DeviceSynchronization.cpp
```diff
@@ -0,0 +1,176 @@
...
+    default:
+        ChipLogProgress(NotSpecified, "Attribute data not processed");
+        break;
+    }
+}
+
+void DeviceSynchronizer::OnReportEnd()
+{
+    // Report end is at the end of all attributes (success)
+    AddSynchronizedDevice(mCurrentDeviceData);
```
tehampson, line 126: I think this is supposed to be in an `#if defined(PW_RPC_ENABLED)`

woody-apple (approved):

examples/fabric-admin/commands/pairing/DeviceSynchronization.cpp
```diff
@@ -0,0 +1,176 @@
...
+    case Clusters::BasicInformation::Attributes::SoftwareVersion::Id:
+        mCurrentDeviceData.has_software_version = SuccessOrLog(data->Get(mCurrentDeviceData.software_version), "HardwareVersion");
+        break;
+    case Clusters::BasicInformation::Attributes::SoftwareVersionString::Id:
+        mCurrentDeviceData.has_software_version_string = SuccessOrLog(
+            data->GetString(mCurrentDeviceData.software_version_string, sizeof(mCurrentDeviceData.software_version_string)),
+            "SoftwareVersionString");
+        break;
+    default:
+        ChipLogProgress(NotSpecified, "Attribute data not processed");
```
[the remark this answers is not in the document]
andy31415: Yes ... and now thinking about it, it is spammy. Removed.

examples/fabric-admin/commands/pairing/DeviceSynchronization.cpp
andy31415: Updated ... but what is the point then to have the fabric admin without RPC? I guess we could use CLI however it seems more reasonable to require this enabling...

examples/fabric-admin/commands/pairing/DeviceSynchronization.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+void DeviceSynchronizer::OnDeviceConnectionFailure(const chip::ScopedNodeId & peerId, CHIP_ERROR error)
+{
+    ChipLogError(NotSpecified, "Device Sync failed to connect to " ChipLogFormatX64, ChipLogValueX64(peerId.GetNodeId()));
+}
+
+void DeviceSynchronizer::StartDeviceSynchronization(chip::Controller::DeviceController & controller, chip::NodeId nodeId,
+                                                    bool deviceIsIcd)
+{
+    mCurrentDeviceData            = chip_rpc_SynchronizedDevice_init_default;
```
tehampson, line 170: Does it make sense to have a bool mOngoingDeviceSync where we do VerifyOrDie(!mOngoingDeviceSync). With a comment saying this currently assumes only one ongoing device sync at a time

mOngoingDeviceSync is set to true here, and set to false in `OnReportEnd` or `OnDone`

tehampson (approved):
LGTM, just one nit comment to validate an assumption or crash

examples/fabric-admin/commands/pairing/DeviceSynchronization.cpp
andy31415: Added, set to false on ReportEnd/OnError since those seem to be the final calls.

examples/fabric-admin/commands/pairing/DeviceSynchronization.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+void DeviceSynchronizer::OnDone(chip::app::ReadClient * apReadClient)
+{
+    // Nothing to do: error reported on OnError or report ended called.
+}
+
+void DeviceSynchronizer::OnError(CHIP_ERROR error)
+{
+    ChipLogProgress(NotSpecified, "Error fetching device data: %" CHIP_ERROR_FORMAT, error.Format());
+    mDeviceSyncInProcess = false;
```
tehampson, line 138: Same thing here, we should only set this to false in OnDone

https://github.com/project-chip/connectedhomeip/blob/master/src/app/ReadClient.h#L192-L207

andy31415: Update to OnDone and also on sending request error. Kind of brittle :(

andy31415:
Merging to be able to rebase the other PR. Linux standalone compile finished and that is the only bit touching this code at the moment.

PR #34855, opened 2024-08-07
Add ICD support (featuremap and commands) to FabricAdmin/FabricBridge samples

andy31415:
Make use of the `is_icd` flag. This builds on top of #34854

### Changes:

- Declare a separate endpoint for ICD bridged devices (this style is terrible ... but unfortunately we have no control over command/attribute lists
- add `TODO` notice that we are re-using version storage, which seems wrong
- Make use of icd flag in feature maps and when deciding what endpoint to present

image

Bit 20 is the feature bit which is ugly ... but ok, 1 << 20 is indeed 1048576 ... also command is indeed 128 (also seems slightly odd ... maybe we expected this to extend/reuse basicinfo)


examples/fabric-bridge-app/fabric-bridge-common/src/BridgedDeviceManager.cpp
```diff
@@ -127,15 +129,35 @@ constexpr CommandId administratorCommissioningCommands[] = {
...
+    DECLARE_DYNAMIC_CLUSTER(AdministratorCommissioning::Id, AdministratorCommissioningAttrs, ZAP_CLUSTER_MASK(SERVER),
+                            administratorCommissioningCommands, nullptr)
+DECLARE_DYNAMIC_CLUSTER_LIST_END;
+// clang-format on
 
 // Declare Bridged Node endpoint
 DECLARE_DYNAMIC_ENDPOINT(sBridgedNodeEndpoint, bridgedNodeClusters);
+DECLARE_DYNAMIC_ENDPOINT(sIcdBridgedNodeEndpoint, icdBridgedNodeClusters);
+
+// TODO: this is a single version array, however we may have many
```
tehampson, line 159: What do you mean by this?

andy31415: Cluster version is a property of when an attribute within the cluster changes, so I imagine if we have 2 (independent) bridged device A and B ... in the current implementation the cluster version is shared, so if A updates an attribute and we increase a cluster version, the cluster version of B also changes, however that cluster did not change at all.

Unsure how to describe this in a comment, however data sharing for non-const data among bridged devices seems wrong.

woody-apple (approved):
Approving pending other review

examples/fabric-admin/commands/pairing/DeviceSynchronization.h
```diff
@@ -0,0 +1,69 @@
...
+#include <app/ReadClient.h>
+#include <controller/CHIPDeviceController.h>
+#include <lib/core/DataModelTypes.h>
+
+#include <memory>
+
+#include "fabric_bridge_service/fabric_bridge_service.pb.h"
+#include "fabric_bridge_service/fabric_bridge_service.rpc.pb.h"
+
+/// Ensures that device data is synchronized to the remove fabric bridge.
```
yufengwangca, line 29: This is used to ensures that device data is synchronized to it own fabric bridge, right?

yufengwangca (approved):

examples/fabric-admin/commands/pairing/DeviceSynchronization.h
andy31415: oops ... remove == remote.
Corrected typo.

yufengwangca: The remote fabric bridge means the fabric bridge on the other ecosystem?

I think this classed is only used to sync the info to its own fabric-bridge, right?

tehampson (approved):

PR #34863, opened 2024-08-07
Have mergify label PRs with conflicts so we can see them easily

woody-apple:
This change has been made by @woody-apple from the Mergify workflow automation editor.


woody-apple:
Note: This is auto-restyled by mergify

PR #34872, opened 2024-08-08
Checking third party deps doesn't pass after someone fixes it

woody-apple:
This is causing confusion and forcing people to close and re-open PRs because it only fails, never passes

### DIFF
--- a/.github/workflows/third-party-check.yaml
+++ b/.github/workflows/third-party-check.yaml
@@ -27,12 +27,12 @@ jobs:
         name: Check For Submodule Update Label
         runs-on: ubuntu-latest
         steps:
-            - if: ${{ contains(github.event.pull_request.labels.*.name, 'changing-submodules-on-purpose') == false }}
+            - if: ${{ !contains(github.event.pull_request.labels.*.name, 'changing-submodules-on-purpose') }}
               name: Fail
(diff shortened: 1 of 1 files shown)


andy31415:
@woody-apple what is the difference? I believe it generally works except it uses the "event label" which is only available at the time the PR is created. Does this allow a "restart" after we add the label? Maybe this is what the description should be.

woody-apple:
> @woody-apple what is the difference? I believe it generally works except it uses the "event label" which is only available at the time the PR is created. Does this allow a "restart" after we add the label? Maybe this is what the description should be.

What doesn't work, is it doesn't check for the presence of the label and pass. So it leaves the PR in a failed state, that requires the filer (or someone like I was doing last night) to open and close the PR to re-evaluate the rule.

woody-apple:
I updated the title to reflect the issue.

woody-apple:
> @woody-apple what is the difference? I believe it generally works except it uses the "event label" which is only available at the time the PR is created. Does this allow a "restart" after we add the label? Maybe this is what the description should be.

Yes, this restarts it post adding the label

Alami-Amine (approved):

andy31415 (approved):

PR #34875, opened 2024-08-08
Added DRLK_2_13 for Aliro Feature

Harshith-GRL:
Added python script for DRLK_2_13 test case in accordance with updated spec from https://github.com/sumaky/chip-test-plans/blob/DRLK2.13-Fixes/src/cluster/door_lock_Cluster.adoc


plauric (approved):

src/python_testing/TC_DRLK_2_13.py
```diff
@@ -0,0 +1,656 @@
...
+logger = logging.getLogger(__name__)
+
+cluster = Clusters.DoorLock
+
+
+class TC_DRLK_2_13(MatterBaseTest):
+
+    def steps_TC_DRLK_2_13(self) -> list[TestStep]:
+        steps = [
+            TestStep("1a", "TH reads OperationalCredentials cluster's CurrentFabricIndex save the attribute",
```
andy31415, line 48: text seems off here, maybe missing a `,` or `and` or similar

src/python_testing/TC_DRLK_2_13.py
```diff
@@ -0,0 +1,656 @@
...
+        self.fabric_idx1 = await self.read_attributes_from_dut(endpoint=self.endpoint,
+                                                               cluster=Clusters.Objects.OperationalCredentials,
+                                                               attribute=Clusters.OperationalCredentials.Attributes.CurrentFabricIndex
+                                                               )
+        if self.pics_guard(self.check_pics("DRLK.S.C1d.Rsp")):
+            self.step("1b")
+            await self.send_clear_user_cmd(user_index=int(0xFFFE))
+        if self.pics_guard(self.check_pics("DRLK.S.C26.Rsp")):
+            self.step("1c")
+            clear_all_credentials = Clusters.DoorLock.Structs.CredentialStruct(credentialIndex=int(0xFFFE))
```
andy31415, line 379: credentialType is not nullable - what will be the value? will it just be default (imagine 0, so it will be pin).
Test plan says to use null ... how does that work?

image

src/python_testing/TC_DRLK_2_13.py
```diff
@@ -0,0 +1,656 @@
...
+
+    async def aliro_attribute_verifiers(self, aliro_attribute_verify_steps: list[dict]):
+
+        for aliro_attribute_step in aliro_attribute_verify_steps:
+            step = aliro_attribute_step.get("th_step")
+            if step is not None:
+                self.step(step)
+            if self.pics_guard(
+                self.check_pics("DRLK.S.F0d") and self.check_pics("DRLK.S.A0080")) and aliro_attribute_step.get(
+                    "condition") == 1:
```
andy31415, line 171: name these constants. "condition == 1" is not clear on what it means.

src/python_testing/TC_DRLK_2_13.py
```diff
@@ -0,0 +1,656 @@
...
+                self.check_pics("DRLK.S.F0d") and self.check_pics("DRLK.S.A0081")) and aliro_attribute_step.get(
+                    "condition") == 2:
+                pass
+            elif self.pics_guard(
+                self.check_pics("DRLK.S.F0d") and self.check_pics("DRLK.S.A0084")) and aliro_attribute_step.get(
+                    "condition") == 3:
+                pass
+            else:
+                continue
+            dut_aliro_key = await self.read_attributes_from_dut(endpoint=aliro_attribute_step.get("endpoint"),
```
andy31415, line 183: this should be dataclasses instead of dict that is free-form.

src/python_testing/TC_DRLK_2_13.py
```diff
@@ -0,0 +1,656 @@
...
+            logging.exception(f"Got exception when performing SetAliroReaderConfig {e}")
+            asserts.assert_equal(e.status, expected_status, f"Unexpected error returned: {e}")
+
+    async def get_credentials_status(self, credentialIndex: int, credentialType: cluster.Enums.CredentialTypeEnum,
+                                     step, userIndex, credential_exists=True):
+        if step is not None:
+            self.step(step)
+        try:
+
+            if self.pics_guard(self.check_pics("DRLK.S.F0d") and self.check_pics("DRLK.S.C24.Rsp") and self.check_pics(
```
andy31415, line 246: the condition can be placed outside the try to reduce indent. This looks to be like:

```py
flags =  ["DRLK.S.F0d", "DRLK.S.C24.Rsp", "DRLK.S.C25.Tx"]
if not self.pics_guard(all([self.check_picks(p) for p in flags])):
    return
```

src/python_testing/TC_DRLK_2_13.py
```diff
@@ -0,0 +1,656 @@
...
+            asserts.assert_equal(e.status, Status.Success, f"Unexpected error returned: {e}")
+
+    async def set_credential_cmd(self, credential_enum: Clusters.DoorLock.Enums.CredentialTypeEnum, credentialIndex,
+                                 operationType, userIndex, credentialData, userStatus, userType, step=None):
+        if step is not None:
+            self.step(step)
+        credentials = cluster.Structs.CredentialStruct(
+            credentialType=credential_enum,
+            credentialIndex=credentialIndex)
+        if self.pics_guard(self.check_pics("DRLK.S.F0d") and self.check_pics("DRLK.S.C22.Rsp")):
```
andy31415, line 272: prefer to `if not ...: return` to reduce indent.

src/python_testing/TC_DRLK_2_13.py
```diff
@@ -0,0 +1,656 @@
...
+                                     f"DataOperationType not matching, dataOperationType={data_type_operation}")
+                break
+
+    async def clear_credentials_cmd(self, credential, step=None, expected_status: Status = Status.Success):
+        try:
+            if step is not None:
+                self.step(step)
+            if self.pics_guard(self.check_pics("DRLK.S.F0d") and self.check_pics("DRLK.S.C26.Rsp")):
+                await self.send_single_cmd(cmd=Clusters.DoorLock.Commands.ClearCredential(credential=credential),
+                                           endpoint=1,
```
andy31415, line 321: could we get a constant for endpoint id that is used everywhere? Or is endpointid 1 mandatory to be the only endpoint supported for this test (by spec)?

src/python_testing/TC_DRLK_2_13.py
```diff
@@ -0,0 +1,656 @@
...
+
+    async def send_clear_aliro_reader_config_cmd(self, expected_status: Status = Status.Success):
+        try:
+            await self.send_single_cmd(cmd=Clusters.DoorLock.Commands.ClearAliroReaderConfig(), endpoint=1,
+                                       timedRequestTimeoutMs=1000)
+            asserts.assert_equal(expected_status, Status.Success)
+        except InteractionModelError as e:
+            asserts.assert_equal(e.status, expected_status, f"Unexpected error returned: {e}")
+
+    async def send_set_aliro_reader_config_cmd(self, steps: dict, expected_status: Status = Status.Success, ):
```
andy31415, line 209: steps should be a dataclass to enforce typing and make the code clearer

src/python_testing/TC_DRLK_2_13.py
```diff
@@ -0,0 +1,656 @@
...
+        try:
+            if self.pics_guard(self.check_pics("DRLK.S.F0d") and not self.check_pics("DRLK.S.F0e") and self.check_pics(
+                    "DRLK.S.C28.Rsp")) and steps.get("condition") == 1:
```
andy31415, line 212: constants for condition should be named. Maybe the step itself could have a "check condition" that has this, so one could say "step.accept_condition(self)" for example.

src/python_testing/TC_DRLK_2_13.py
sumaky: DRLKClearCred
The test plan uses null for credential type.

src/python_testing/TC_DRLK_2_13.py
```diff
@@ -0,0 +1,656 @@
...
+            "047a4c552d753924cdf3779a3c84fec2debaa6f0b3084450878acc7ddcce7856ae57b1ebbe2561015103dd7474c2a183675378ec55f1e465ac3436bf3dd5ca54d4")
+        #  step 1 TH reads DUT Endpoint 0 OperationalCredentials cluster CurrentFabricIndex attribute
+        self.step("1a")
+        self.fabric_idx1 = await self.read_attributes_from_dut(endpoint=self.endpoint,
+                                                               cluster=Clusters.Objects.OperationalCredentials,
+                                                               attribute=Clusters.OperationalCredentials.Attributes.CurrentFabricIndex
+                                                               )
+        if self.pics_guard(self.check_pics("DRLK.S.C1d.Rsp")):
+            self.step("1b")
+            await self.send_clear_user_cmd(user_index=int(0xFFFE))
```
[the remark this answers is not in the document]
Harshith-GRL: hi @andy31415 ,  I believe i made tiny mistake here, i will call self.step() check before calling pics_gaurd function, in a scenario where the self.check_pics fails, the pics_gaurd  function will skip the step

src/python_testing/TC_DRLK_2_13.py
Harshith-GRL: text is updated with `and` to make it more readable

src/python_testing/TC_DRLK_2_13.py
```diff
@@ -0,0 +1,656 @@
...
+            attribute_value = await self.read_single_attribute_check_success(endpoint=endpoint, cluster=cluster,
+                                                                             attribute=attribute)
+            asserts.assert_equal(expected_status, Status.Success)
+            return attribute_value
+        except Exception as e:
+            logging.error(e)
+            asserts.assert_equal(expected_status, Status.Success,
+                                 f"Error reading attributes, response={attribute_value}")
+
+    async def aliro_attribute_verifiers(self, aliro_attribute_verify_steps: list[dict]):
```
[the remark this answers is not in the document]
Harshith-GRL: created a new dataclass called  `AliroAttributeVerify` code will now look like  `async def aliro_attribute_verifiers(self, aliro_attribute_verify_steps: list[AliroAttributeVerify]):`

src/python_testing/TC_DRLK_2_13.py
Harshith-GRL: if check logics are changed to be more friendly, removed  "condition == 1" and replaced with simpler logic, code will now look like this 
`th_step = aliro_attribute_step.th_step
            self.step(th_step)
            if aliro_attribute_step.attribute == Clusters.DoorLock.Attributes.AliroReaderVerificationKey:
                pics_condition = self.pics_guard(self.check_pics("DRLK.S.F0d") and self.check_pics("DRLK.S.A0080"))
            elif aliro_attribute_step.attribute == Clusters.DoorLock.Attributes.AliroReaderGroupIdentifier:
                pics_condition = self.pics_guard(self.check_pics("DRLK.S.F0d") and self.check_pics("DRLK.S.A0081"))
            elif aliro_attribute_step.attribute == Clusters.DoorLock.Attributes.AliroGroupResolvingKey:
                pics_condition = self.pics_guard(self.check_pics("DRLK.S.F0d") and self.check_pics("DRLK.S.A0084"))
            else:
                pics_condition = False

            if pics_condition is True:
                dut_aliro_key = await self.read_attributes_from_dut(endpoint=self.app_cluster_endpoint,
                                                                    cluster=aliro_attribute_step.cluster,
                                                                    attribute=aliro_attribute_step.attribute)`

src/python_testing/TC_DRLK_2_13.py
Harshith-GRL: dataclasses are used and all dict methods have been removed

src/python_testing/TC_DRLK_2_13.py
Harshith-GRL: code is updated and all dict use cases are removed

src/python_testing/TC_DRLK_2_13.py
Harshith-GRL: used similar code and execution is passing

src/python_testing/TC_DRLK_2_13.py
Harshith-GRL: started using ` self.app_cluster_endpoint = 1`  as constant value in all places

src/python_testing/TC_DRLK_2_13.py
```diff
@@ -0,0 +1,656 @@
...
+        aliro_non_evict_endpoint_credentials = cluster.Structs.CredentialStruct(credentialIndex=1,
+                                                                                credentialType=Clusters.DoorLock.Enums.CredentialTypeEnum.kAliroNonEvictableEndpointKey)
+        all_aliro_cred_list.append(aliro_non_evict_endpoint_credentials)
+        for creds in all_aliro_cred_list:
+            await self.clear_credentials_cmd(credential=creds)
+
+    def generate_unique_octstr(self, length=65):
+        octal_digits = '01234567'
+        octstr = ''.join(random.choice(octal_digits) for _ in range(length))
+        return octstr.encode()
```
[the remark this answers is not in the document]
Harshith-GRL: changed code as per suggestion and renamed the function as `generate_unique_octbytes`

src/python_testing/TC_DRLK_2_13.py
Harshith-GRL: code will now use null as per the updated spec

src/python_testing/TC_DRLK_2_13.py
Harshith-GRL: can we leave it untouched so that code has readability just like test plan spec?

src/python_testing/TC_DRLK_2_13.py
andy31415: I have updated the latest spec and re-built. The `CredentialType` is *not nullable*. If you are not setting it to any value, it will just take the default in python:

image

The test plan should be updated to say 0 or something else (and that made explicit in the test). You cannot set it to null because the value is not nullable.

src/python_testing/TC_DRLK_2_13.py
```diff
@@ -0,0 +1,663 @@
...
+             "attribute": Clusters.DoorLock.Attributes.AliroReaderVerificationKey,
+             "aliro_attribute_value": self.verificationKey},
+            {"th_step": "4", "cluster": Clusters.Objects.DoorLock,
+             "attribute": Clusters.DoorLock.Attributes.AliroReaderGroupIdentifier,
+             "aliro_attribute_value": self.groupIdentifier},
+            {"th_step": "5", "cluster": Clusters.Objects.DoorLock,
+             "attribute": Clusters.DoorLock.Attributes.AliroGroupResolvingKey,
+             "aliro_attribute_value": self.groupResolvingKey}
+        ]
+        list_of_verification_objects = [AliroAttributeVerify(**aliro_object) for aliro_object in
```
andy31415, line 407: create the objects in the list, do not convert from dict.

src/python_testing/TC_DRLK_2_13.py
```diff
@@ -0,0 +1,663 @@
...
+logger = logging.getLogger(__name__)
+
+cluster = Clusters.DoorLock
+
+
+@dataclass
+class AliroAttributeVerify:
+    th_step: str
+    attribute: Clusters.DoorLock.Attributes
+    cluster: Clusters.DoorLock
```
andy31415, line 49: Is this ever a different value than Clusters.Objects.DoorLock?

This seems not needed.

src/python_testing/TC_DRLK_2_13.py
```diff
@@ -0,0 +1,663 @@
...
+        await self.send_set_aliro_reader_config_cmd(use_group_resolving_key=True, expected_status=Status.InvalidInState)
+        #  step 7
+        self.step("7")
+        if self.check_pics("DRLK.S.C26.Rsp"):
+            await self.send_clear_aliro_reader_config_cmd()
+
+        # steps 8,9,10
+        aliro_attribute_verify_steps = [
+            {"th_step": "8", "cluster": Clusters.Objects.DoorLock,
+             "attribute": Clusters.DoorLock.Attributes.AliroReaderVerificationKey, "aliro_attribute_value": NullValue},
```
andy31415, line 425:
```suggestion
            AliroAttributeVerify(
                th_step="8", 
                attribute=Clusters.DoorLock.Attributes.AliroReaderVerificationKey, 
                attribute_value=NullValue),
```

Please make the similar change throught. Note that I dropped the cluster (redundant) and renamed the value (stutters/smurf naming)

src/python_testing/TC_DRLK_2_13.py
```diff
@@ -0,0 +1,663 @@
...
+
+cluster = Clusters.DoorLock
+
+
+@dataclass
+class AliroAttributeVerify:
+    th_step: str
+    attribute: Clusters.DoorLock.Attributes
+    cluster: Clusters.DoorLock
+    aliro_attribute_value: bytes
```
andy31415, line 50: aliro prefix is obvious ... the class is already AliroAttribute named. name it `attribute_value` or `value`

andy31415, line 50:
```suggestion
    attribute_value: bytes
```

src/python_testing/TC_DRLK_2_13.py
```diff
@@ -0,0 +1,663 @@
...
+            if aliro_attribute_step.attribute == Clusters.DoorLock.Attributes.AliroReaderVerificationKey:
+                pics_condition = self.pics_guard(self.check_pics("DRLK.S.F0d") and self.check_pics("DRLK.S.A0080"))
+            elif aliro_attribute_step.attribute == Clusters.DoorLock.Attributes.AliroReaderGroupIdentifier:
+                pics_condition = self.pics_guard(self.check_pics("DRLK.S.F0d") and self.check_pics("DRLK.S.A0081"))
+            elif aliro_attribute_step.attribute == Clusters.DoorLock.Attributes.AliroGroupResolvingKey:
+                pics_condition = self.pics_guard(self.check_pics("DRLK.S.F0d") and self.check_pics("DRLK.S.A0084"))
+            else:
+                pics_condition = False
+
+            if pics_condition is True:
```
andy31415, line 187: why not:

```suggestion
            if pics_condition:
```

src/python_testing/TC_DRLK_2_13.py
```diff
@@ -0,0 +1,663 @@
...
+        except Exception as e:
+            logging.error(e)
+            asserts.assert_equal(expected_status, Status.Success,
+                                 f"Error reading attributes, response={attribute_value}")
+
+    async def aliro_attribute_verifiers(self, aliro_attribute_verify_steps: list[AliroAttributeVerify]):
+
+        for aliro_attribute_step in aliro_attribute_verify_steps:
+            th_step = aliro_attribute_step.th_step
+            self.step(th_step)
```
andy31415, line 177: are we supposed to step even if condition is false. Unsure about UI needs here, maybe we should move this later as:

```py
if pics_condition:
   self.step(th_step)
else:
   self.skip_step(th_step)
```

src/python_testing/TC_DRLK_2_13.py
```diff
@@ -0,0 +1,663 @@
...
+        try:
+            # Checks Pics condition
+            if use_group_resolving_key is False:
+                pics_check = self.pics_guard(self.check_pics("DRLK.S.F0d") and not self.check_pics("DRLK.S.F0e") and
+                                             self.check_pics("DRLK.S.C28.Rsp"))
+
+            else:
+                pics_check = self.pics_guard(self.check_pics("DRLK.S.F0e") and self.check_pics("DRLK.S.C28.Rsp"))
+
+            if use_group_resolving_key is False and pics_check is True:
```
andy31415, line 226:
```suggestion
            if not _use_group_resolving_key and pics_check:
```

Please replace all comparisons with True/False with direct usage inside the PR

src/python_testing/TC_DRLK_2_13.py
Harshith-GRL: You're right its not needed, I have removed

src/python_testing/TC_DRLK_2_13.py
Harshith-GRL: yeah that makes sense, i will rename it.

Harshith-GRL: changes are updated

src/python_testing/TC_DRLK_2_13.py
Harshith-GRL: the self.pics_gaurd will take care of it, it will evaluate the pics condition and skip.

src/python_testing/TC_DRLK_2_13.py
Harshith-GRL: removed the `is True`

src/python_testing/TC_DRLK_2_13.py
Harshith-GRL: The PR will now have direct usage with true/false comparison

src/python_testing/TC_DRLK_2_13.py
Harshith-GRL: dict converter has been removed

src/python_testing/TC_DRLK_2_13.py
Harshith-GRL: object will be created in the list

andy31415 (approved):
Approval as rubber stamp due to timelines after brief readability reviews.

Note that some readability things should still be done like `do not use is True/False` and `if step is not None` should be `if step` or similar things.

woody-apple (approved):

kiel-apple (approved):

jmartinez-silabs (approved):

PR #34879, opened 2024-08-08
Restructured folders in chip-energy-management-app - no code change

jamesharrow:
In order to make way for WaterHeater as an example, the existing examples/energy-management-app/common was too cluttered.

This PR simply moves files into subfolders and updates the build and CMakefiles.txt so it is still compatible.


lpbeliveau-silabs:
Could we put the common files, DEM and Electrical Sensor in a common location outside of the energy-management-app? This would allow appliances app such as dishwasher, laundry machines, etc. to leverage these files and still let platform decide if they want to add the Energy Management device types or not.

jamesharrow:
> Could we put the common files, DEM and Electrical Sensor in a common location outside of the energy-management-app? This would allow appliances app such as dishwasher, laundry machines, etc. to leverage these files and still let platform decide if they want to add the Energy Management device types or not.

As discussed on slack - I think we could look at this in the wider context. The plan is to try to add other 'energy-management' examples in the future. There is a PR in the works based on this to add Water Heater support, and I would like to be able to add Solar / Battery here as examples in due course.

Once we have price information then it would be possible for a price client to use a DEM forecast to compute the cheapest / greenest time use using a simple optimiser. So the Price Server implementation (perhaps in 1.5?) would sit in this 'library'.

The operation for a EVSE, Heat pump, Water Heater, Battery etc are very flexible in how they can scheduled, whereas a white goods appliance is more table driven based on the washing cycle selected.

Ideally we would create the energy-management as a library (rather than an app) - however the convention today seems to be examples/xyz-**app**/ whereas I think we might want energy-management as a library and as you say, apps can pull in the code from these.

I think there is a risk of duplication of code in the existing way of doing things when producing a simple app, and the boiler plate code to create a simple demo app gets reproduced over an over. So a single app with different personalities (in clearly defined folders) is arguably less maintenance overhead.

The idea in the follow up PR is we have a single app with build time config to enable the relevant personality but using the common:
 - Device Energy Management cluster
 - Energy Reporting clusters (EEM / EPM)
 - Power Topology
 
 The hope is with good documentation and clear examples we can help manufacturers make energy management easy - so I'm quite keen we keep the code together for now - but perhaps we can ask if we can create an top level energy-management library.
 
 At the moment All Clusters Apps do pull code from this 'examples/energy-management/common' and if we can work out a better place to put that then happy to move it.
 
The main thing this PR does for now is to clear the way for Water Heater demonstration to be put in. But this isn't SVE or TE2 gating since we already have the All-clusters app for that.

I suspect we need a bit of a discussion on a call to work through the ideal folder structure for this.

arkq (approved):

rbultman (approved):

andy31415 (approved):

PR #34886, opened 2024-08-08
Add water heater to energy management app

### LINKED ISSUE #34966: [BUG] Example Energy Management app for EVSE etc does not set the absMinPower and adsMaxPower values in DEM
### Reproduction steps

When testing the TC_DEM_2.1 the test says check that the absMinPower and absMaxPower is as expected.

These are both unset (set to 0).

They should be set to something like 0 and 7000,000 mW

### Bug prevalence

Every time

b942f5cc7009968693652156ca28cfbbff4dc5e1

### Platform

core

### Platform Version(s)

1.4

### Anything else?

See test case https://github.com/CHIP-Specifications/chip-test-plans/blob/master/src/cluster/device_energy_management.adoc#test-procedure-1 step 5 & 6

### LINKED ISSUE #34977: [Test Failed] EWATERHRT1.1: Failure return on reading Feature-Map and Cluster Revision attribute.
### Test issue(s)

DUT returned failure when a read command was sent from the controller for feature-map and cluster-revision attributes.

### Platform

_No response_

### Anything else?

[chip_tool.txt](https://github.com/user-attachments/files/16611372/chip_tool.txt)
[device.txt](https://github.com/user-attachments/files/16611373/device.txt)

jamesharrow:
This PR aims to add a Water Heater example into the energy-management example app so that EP1 has EVSE and EP2 has WaterHeater.

Both would then share the Electrical Sensor, DEM and Power Topology clusters.

Summary of changes made in this PR
 - Disabled OnMode / StartupMode for EVSEM, DEMM and WHM Energy Management ZAP file (to align with spec)

 - Moved several WaterHeater related files from All-clusters-app to energy-management-app (updates all-clusters-build path and re-used in energy-management-app)
 - Fixed bugs around Forecast attribute not being marked for reporting when it was being updated

 - Added an application type command flag to energy-management-app (Linux)
    `--application evse` | `--application water-heater`
     - updated DEM and EVSE test scripts test runner header to include this new flag.
   
   - Added new Linux Build options for WaterHeater Test event trigger enabled by default

 - Added Water Heater support to ESP32 via KConfig settings
 - Included test event trigger support into ESP32 main.cpp based on KConfig flags
 - Added compile type support for PFR or SFR features (note that these are mutually exclusive and will fail DEM tests if both are on!) (equivalent of `--featureSet` argument in the linux chip-energy-management-app)
 
 - Updated the README.md for ESP32 and Linux

 - Fixes #34977
 - Fixes #34966


examples/energy-management-app/esp32/main/Kconfig.projbuild
```diff
@@ -185,3 +185,38 @@ depends on ENABLE_PW_RPC
...
+menu "Energy Management Example App Configuration"
+
+    config ENABLE_EXAMPLE_EVSE_DEVICE
+        bool "Enable EVSE device on EP1"
+        default "y"
+        help
+            To use Energy Management App as EVSE device and compile dependent files.
+
+    config ENABLE_EXAMPLE_WATER_HEATER_DEVICE
+        bool "Enable Water Heater device on EP2"
```
jamesharrow, line 199: Is this on EP2 or still EP1?

examples/energy-management-app/esp32/main/main.cpp
```diff
@@ -117,10 +125,18 @@ namespace Clusters {
 namespace DeviceEnergyManagement {
 
 // Keep track of the parsed featureMap option
+#if 1 // 0x7a
```
jamesharrow, line 128: Make this a config flag

examples/energy-management-app/esp32/main/main.cpp
```diff
@@ -117,10 +125,18 @@ namespace Clusters {
...
 static chip::BitMask<Feature> sFeatureMap(Feature::kPowerAdjustment, Feature::kPowerForecastReporting,
+                                          Feature::kStartTimeAdjustment, Feature::kPausable,
+                                          Feature::kForecastAdjustment, Feature::kConstraintBasedAdjustment);
+#else // 0x7c
+static chip::BitMask<Feature> sFeatureMap(Feature::kPowerAdjustment,
                                           Feature::kStateForecastReporting, Feature::kStartTimeAdjustment, Feature::kPausable,
                                           Feature::kForecastAdjustment, Feature::kConstraintBasedAdjustment);
 
+static const char * spApp = nullptr;
```
jamesharrow, line 137: Not clear why this is in this branch of the #if / #else ?

examples/energy-management-app/esp32/sdkconfig.defaults
```diff
@@ -67,3 +67,7 @@ CONFIG_LWIP_IPV6_NUM_ADDRESSES=6
 
 # Enable sntp time sync
 CONFIG_ENABLE_SNTP_TIME_SYNC=y
+
+CONFIG_ENABLE_EXAMPLE_WATER_HEATER_DEVICE=y
```
jamesharrow, line 71: Should the default be EVSE?

examples/energy-management-app/esp32/sdkconfig.defaults
```diff
@@ -67,3 +67,7 @@ CONFIG_LWIP_IPV6_NUM_ADDRESSES=6
...
+
+CONFIG_CHIP_DEVICE_CONFIG_ENABLE_WATER_HEATER_MANAGEMENT_TRIGGER=y
```
jamesharrow, line 73: Should DEM be turned on and EVSE triggers too?

examples/energy-management-app/linux/main.cpp
```diff
@@ -87,22 +98,66 @@ static uint32_t ParseNumber(const char * pString)
 void ApplicationInit()
 {
     ChipLogDetail(AppServer, "Energy Management App: ApplicationInit()");
-    EvseApplicationInit();
+    if (strcmp(spApp, kEvseApp) == 0)
```
jamesharrow, line 101: Could we make EVSE the default so if you don't specify it, then that's used?

examples/energy-management-app/linux/main.cpp
```diff
@@ -87,22 +98,66 @@ static uint32_t ParseNumber(const char * pString)
...
+static bool EnergyAppOptionHandler(const char * aProgram, chip::ArgParser::OptionSet * aOptions, int aIdentifier,
+                                   const char * aName, const char * aValue)
 {
     bool retval = true;
 
+    ChipLogError(Support, "aIdentifier 0x%04x aValue %s", aIdentifier, aValue);
     switch (aIdentifier)
     {
+    case kOptionApplication:
+        for (uint16_t idx = 0; idx < sizeof(kValidApps) / sizeof(kValidApps[0]); idx++)
```
jamesharrow, line 142: I think there is a macro for sizeof(struct)/sizeof(struct[0]) - is it ARRAY_SIZE?

examples/energy-management-app/linux/main.cpp
```diff
@@ -87,22 +98,66 @@ static uint32_t ParseNumber(const char * pString)
...
+        {
+            ChipLogError(Support, "kValidApps[%d] %s", idx, kValidApps[idx]);
+            if (strcmp(kValidApps[idx], aValue) == 0)
+            {
+                spApp = aValue;
```
jamesharrow, line 147: maybe better to assign to kValidApps[idx] since this is const and guaranteed to be there, whereas the value of aValue ptr might get destroyed after this function is called.

examples/energy-management-app/energy-management-common/energy-management-app.matter
```diff
@@ -2210,6 +2271,56 @@ cluster EnergyEvseMode = 157 {
...
+
+  struct ModeOptionStruct {
+    char_string<64> label = 0;
+    int8u mode = 1;
+    ModeTagStruct modeTags[] = 2;
+  }
+
+  readonly attribute ModeOptionStruct supportedModes[] = 0;
+  readonly attribute int8u currentMode = 1;
+  attribute optional nullable int8u startUpMode = 2;
```
jamesharrow, line 2301: I think startupMode and OnMode should be turned off now - spec change coming through which disallows these.

examples/energy-management-app/esp32/main/Kconfig.projbuild
PeterC1965: Still on EP1. I'll change the comment to EP1

examples/energy-management-app/esp32/main/main.cpp
PeterC1965: Changed with a number of #defines in the Kconfig.projbuild file

examples/energy-management-app/esp32/main/main.cpp
PeterC1965: It is obsolete as now control the app type using conditional compilation.

examples/energy-management-app/esp32/sdkconfig.defaults
PeterC1965: OK

examples/energy-management-app/linux/main.cpp
PeterC1965: Fair point :-)

examples/energy-management-app/esp32/main/Kconfig.projbuild
```diff
@@ -185,3 +185,50 @@ depends on ENABLE_PW_RPC
...
+
+    config ENABLE_EXAMPLE_WATER_HEATER_DEVICE
+        bool "Enable Water Heater device on EP1"
+        default "n"
+        help
+            To use Energy Management App as Water Heater device and compile dependent files.
+
+    config CHIP_DEVICE_CONFIG_ENABLE_DEVICE_ENERGY_MANAGEMENT_TRIGGER
+        bool "Enable DEM triggers"
+        default "y"
```
jamesharrow, line 206: Either this should be on by default and the other should be, but suspect for security we would want these OFF by default, and then in the SDKDefaults have them as set?

examples/energy-management-app/esp32/main/Kconfig.projbuild
```diff
@@ -185,3 +185,50 @@ depends on ENABLE_PW_RPC
...
+        bool "Enable Water Heater triggers"
+        default "n"
+        help
+            Enable test triggers used by the Water Heater Management tests
+
+    config DEM_SUPPORT_STATE_FORECAST_REPORTING
+        bool "State forecast reporting is supported"
+        default "n"
+        help
+            Ensure kStateForecastReporting (SFR) is set in the DeviceEnergyManagement::Feature attribute. Note cannot set SFR and PFR.
```
jamesharrow, line 226: There may be a KConfig mechanism to make these 2 mutually exclusive.
https://github.com/torvalds/linux/blob/v4.15/Documentation/kbuild/kconfig-language.txt#L104

examples/energy-management-app/esp32/main/main.cpp
```diff
@@ -117,9 +125,22 @@ namespace Clusters {
...
-                                          Feature::kStateForecastReporting, Feature::kStartTimeAdjustment, Feature::kPausable,
+                                          Feature::kStartTimeAdjustment, Feature::kPausable, Feature::kForecastAdjustment,
+                                          Feature::kConstraintBasedAdjustment);
+#elif CONFIG_DEM_SUPPORT_STATE_FORECAST_REPORTING
+static chip::BitMask<Feature> sFeatureMap(Feature::kPowerAdjustment, Feature::kStateForecastReporting,
+                                          Feature::kStartTimeAdjustment, Feature::kPausable, Feature::kForecastAdjustment,
+                                          Feature::kConstraintBasedAdjustment);
+#else
+static chip::BitMask<Feature> sFeatureMap(Feature::kPowerAdjustment, Feature::kStartTimeAdjustment, Feature::kPausable,
                                           Feature::kForecastAdjustment, Feature::kConstraintBasedAdjustment);
```
jamesharrow, line 142: I don't think you can do Pausing, StartTimeAdjustment, ForecastAdjustment or ConstraintBasedAdjustment unless you have either SFR or PFR.

So in practise that only leaves PowerAdjustment.

examples/energy-management-app/esp32/sdkconfig.defaults
jamesharrow: Not done?

examples/energy-management-app/esp32/sdkconfig.defaults
jamesharrow: Not done?

examples/energy-management-app/esp32/main/Kconfig.projbuild
jamesharrow: Note Linux has all test event triggers enabled by default so maybe align this to the same?

src/controller/python/chip/clusters/Objects.py
```diff
@@ -50958,7 +50958,7 @@ def descriptor(cls) -> ClusterObjectDescriptor:
                         ClusterObjectFieldDescriptor(Label="field2", Tag=1, Type=Globals.Enums.TestGlobalEnum),
                     ])
 
-            field1: 'Globals.Structs.TestGlobalStruct' = field(default_factory=lambda: UnitTesting.Structs.TestGlobalStruct())
+            #field1: 'Globals.Structs.TestGlobalStruct' = field(default_factory=lambda: UnitTesting.Structs.TestGlobalStruct())
```
jamesharrow, line 50961: Why has this changed?

jamesharrow:
It would be good to see if we can get EFR32 support into this PR if possible.

examples/energy-management-app/energy-management-common/energy-management-app.matter
jamesharrow: Spec already updated and no XML change needed.

examples/energy-management-app/linux/main.cpp
jamesharrow: ok

examples/energy-management-app/esp32/main/main.cpp
PeterC1965: Code updated accordingly.

examples/energy-management-app/esp32/main/Kconfig.projbuild
PeterC1965: Code updated accordingly.

examples/energy-management-app/esp32/sdkconfig.defaults
PeterC1965: Code updated accordingly.

examples/energy-management-app/esp32/sdkconfig.defaults
PeterC1965: Code updated accordingly.

src/python_testing/TC_DEM_2_2.py
```diff
@@ -23,7 +23,7 @@
 # test-runner-run/run1/app: ${ENERGY_MANAGEMENT_APP}
 # test-runner-run/run1/factoryreset: True
 # test-runner-run/run1/quiet: True
-# test-runner-run/run1/app-args: --discriminator 1234 --KVS kvs1 --trace-to json:${TRACE_APP}.json --enable-key 000102030405060708090a0b0c0d0e0f --featureSet 0x01
+# test-runner-run/run1/app-args: --discriminator 1234 --KVS kvs1 --trace-to json:${TRACE_APP}.json --enable-key 000102030405060708090a0b0c0d0e0f --featureSet 0x01 --application evse
```
lpbeliveau-silabs, line 26: Why do we need to precise the app here if the app is already mentioned three lines above:
`${ENERGY_MANAGEMENT_APP}` ?

examples/energy-management-app/linux/main.cpp
```diff
@@ -26,28 +27,36 @@ using namespace chip::app;
...
-static bool FeatureMapOptionHandler(const char * aProgram, chip::ArgParser::OptionSet * aOptions, int aIdentifier,
-                                    const char * aName, const char * aValue);
+static bool EnergyAppOptionHandler(const char * aProgram, chip::ArgParser::OptionSet * aOptions, int aIdentifier,
+                                   const char * aName, const char * aValue);
 
-constexpr uint16_t kOptionFeatureMap = 'f';
+constexpr uint16_t kOptionApplication = 0xffd0;
+constexpr uint16_t kOptionFeatureMap  = 0xffd1;
+
+constexpr const char * kEvseApp = "evse";
```
lpbeliveau-silabs, line 42: I see we are moving forward with the whole multiple apps under the same folder structure approach.
This leaves us with the choice to either 

- have every appliance app live under the energy management folder, which wouldn't make sens for matter app that don't support the device, therefore having 2 copies of each app that wants to support the energy management.
- Have different structures between EVSE + water heater vs the Apps that existed in the repo before.
- Refactor the curent folder structure to have separate folder for the common energy management files and have a separate example app for EVSE and Water heater.

I understand this is time sensitive, but we should have a followup issue somewhere to refactor this architecture, so at least we can keep in mind we should at least consider how to unify this app with the rest of the example apps in the repo.

examples/energy-management-app/linux/main.cpp
```diff
@@ -87,22 +99,56 @@ static uint32_t ParseNumber(const char * pString)
...
+        ChipLogError(Support, "Unexpected application %s", spApp);
+    }
 }
 
 void ApplicationShutdown()
 {
     ChipLogDetail(AppServer, "Energy Management App: ApplicationShutdown()");
+
     EvseApplicationShutdown();
+    FullWhmApplicationShutdown();
```
lpbeliveau-silabs, line 121: I see here we might shutdown an app that hasn't been initialized, is there no risk of undefined behavior here?

examples/energy-management-app/energy-management-common/water-heater/src/WhmMain.cpp
```diff
@@ -167,25 +167,29 @@ CHIP_ERROR WhmManufacturerShutdown()
     return CHIP_NO_ERROR;
 }
 
-void WhmApplicationInit()
+CHIP_ERROR WhmApplicationInit()
 {
-    if (WhmInit() != CHIP_NO_ERROR)
+    CHIP_ERROR err = WhmInit();
+    if (err != CHIP_NO_ERROR)
```
lpbeliveau-silabs, line 173: Any reason not to use the `ReturnErrorOnFailure` macro?

examples/energy-management-app/energy-management-common/water-heater/src/WhmMain.cpp
```diff
@@ -167,25 +167,29 @@ CHIP_ERROR WhmManufacturerShutdown()
...
     {
-        return;
+        return err;
     }
 
     /* Do this last so that the instances for other clusters can be wrapped inside */
-    if (WhmManufacturerInit() != CHIP_NO_ERROR)
+    err = WhmManufacturerInit();
+    if (err != CHIP_NO_ERROR)
```
lpbeliveau-silabs, line 180: Same here for `ReturnErrorOnFailure`

examples/energy-management-app/energy-management-common/water-heater/src/WaterHeaterMain.cpp
```diff
@@ -0,0 +1,91 @@
...
+#include <WhmMain.h>
+
+namespace chip {
+namespace app {
+namespace Clusters {
+namespace WaterHeaterManagement {
+
+void FullWhmApplicationInit()
+{
+    if (WhmApplicationInit() != CHIP_NO_ERROR)
```
lpbeliveau-silabs, line 47: Why not use `ReturnOnFailure` here? same for the steps after this one.

lpbeliveau (commented):
Minor comments for the most of it, most important point is the structure refactor to get this app in line with the rest of the example apps. A followup issue should be created.

examples/energy-management-app/linux/main.cpp
jamesharrow: Agree - I think if we can have an energy management 'library' then this would look different. 

As mentioned in a previous PR - the longer term vision is to provide different energy management examples since the use cases for a Washing Machine / Dishwasher 'forecast' is very different from a Battery / EV.

We are very much at the beginning of a journey on energy management and over time this will need to continue to evolve with the ability to use the same DEM cluster with optimisers using Time of Use tariff information, and battery and EV using the power adjustment all as library calls.

I suggest sometime after SVE we workshop this since there are counter arguments against the current approach too. I'm not sure we need a follow up since the two approaches will need to be resolved in the longer term and won't be forgotten about.

examples/energy-management-app/linux/main.cpp
jamesharrow: I think all of the functions check that they only free pointers which have been allocated. So even if you run the app in EVSE mode, it is safe to call the WaterHeaterShutdown.

examples/energy-management-app/energy-management-common/water-heater/src/WhmMain.cpp
jamesharrow: Good suggestion - changed.

examples/energy-management-app/energy-management-common/water-heater/src/WaterHeaterMain.cpp
jamesharrow: Changed

src/python_testing/TC_DEM_2_2.py
jamesharrow: The energy management app itself can run as a water heater or evse personality, so this tells the chip-energy-management-app to run with the EVSE cluster and delegates instantiated (rather than the water heater).

lpbeliveau (approved):

src/python_testing/TC_DEM_2_2.py
PeterC1965: The -application option can be either evse or water-heater. The energy management app can be run in one of two modes: EVSE or water-heater-management mode.

doru91 (approved):

hasty (approved):

examples/energy-management-app/linux/main.cpp
jamesharrow: Resolving so the PR can merge

src/python_testing/TC_DEM_2_2.py
jamesharrow: Resolving so the PR can merge

PR #34887, opened 2024-08-08
Update rvc app integrate service area

### LINKED ISSUE #33367: Update the RVC example app to include the ServiceArea cluster
Update the RVC example app to emulate an RVC device that uses the ServiceArea cluster and it's features.

hicklin:
Fixes #33367.

This PR 
- integrates the Service Area cluster with the rest of the RVC example app logic.
- updates the PICS file.
- updates the readme.


plauric (approved):

johnfierke (approved):

examples/rvc-app/rvc-common/src/rvc-service-area-delegate.cpp
```diff
@@ -77,32 +108,89 @@ CHIP_ERROR RvcServiceAreaDelegate::Init()
...
+            return false;
+        }
+
+        if (reqSize == 0)
+        {
+            return true;
+        }
+    }
+
+    // If there are less than 2 supported maps, any combination of areas is valid.
```
kiel-apple, line 135:
```suggestion
    // If there is 1 or 0 supported maps, any combination of areas is valid.
```

examples/rvc-app/rvc-common/src/rvc-service-area-delegate.cpp
```diff
@@ -77,32 +108,89 @@ CHIP_ERROR RvcServiceAreaDelegate::Init()
...
+    if (!GetInstance()->HasFeature(Feature::kMaps) || GetNumberOfSupportedMaps() <= 1)
+    {
+        return true;
+    }
+
+    // Check that all the areas are in the same map.
```
kiel-apple, line 141:
```suggestion
    // Check that all the requested areas are in the same map.
```

kiel-apple (approved):

examples/rvc-app/rvc-common/src/rvc-service-area-delegate.cpp
```diff
@@ -77,32 +108,89 @@ CHIP_ERROR RvcServiceAreaDelegate::Init()
...
+                                                   MutableCharSpan & statusText)
 {
-    // TODO IMPLEMENT
+    // if req is empty list return true.
+    {
+        size_t reqSize;
+        if (req.newAreas.ComputeSize(&reqSize) != CHIP_NO_ERROR)
+        {
+            areaStatus = SelectAreasStatus::kInvalidSet; // todo Not sure this is the correct error to use here
+            CopyCharSpanToMutableCharSpan("error computing number of selected areas"_span, statusText);
```
bzbarsky-apple, line 125: If `statusText` is too short for that string, this will leave `statusText` containing whatever happened to be on the stack up there, which will then get sent out to the client.  As in, this is a security bug that leaks stack memory on the wire.

hicklin: This is safe due to the fact that the size of all of these strings is smaller than `kMaxSizeStatusText`. I have added [this issue](https://github.com/project-chip/connectedhomeip/issues/35250) to updated the documentation wherever a `MutableCharSpan` is passed as a reference to say that if this is not edited successfully, it's size must be set to 0.

PR #34892, opened 2024-08-08
Remove pybindings from stack

Alami-Amine:
- pybindings is not used anymore and the fact that we can build it causes confusion.
- it has been replaced by the code in `src/controller/python` using python `ctypes`.


andy31415:
@Alami-Amine we could also remove `third_party/pybind11` as a submodule to speed up the submodule updates a slight bit and save some storage (also very small but at least non-zero)

Alami-Amine:
> @Alami-Amine we could also remove `third_party/pybind11` as a submodule to speed up the submodule updates a slight bit and save some storage (also very small but at least non-zero)

Done

andy31415 (approved):

saurabhst (approved):

Alami-Amine:
@mergifyio refresh

Issue #33221, opened 2024-04-29
[Feature] Integrate KeepActive command in BRBINFO cluster into fabric-bridge fabric-admin examples


yufengwangca:
### Feature description

...
Implements: https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/9088

### Platform

linux

PR #34894, opened 2024-08-08
Create new ICD Manager in fabric-admin to service KeepActive Command

tehampson:
Moves closer to completing #33221

When fabric-admin commissions the ICD device it needs to commission with the command `--icd-registration true`


examples/fabric-admin/rpc/RpcServer.cpp
```diff
@@ -34,9 +38,24 @@ using namespace ::chip;
 namespace {
 
 #if defined(PW_RPC_FABRIC_ADMIN_SERVICE) && PW_RPC_FABRIC_ADMIN_SERVICE
-class FabricAdmin final : public rpc::FabricAdmin
+class FabricAdmin final : public rpc::FabricAdmin, public IcdManager::Delegate
 {
 public:
+    void OnCheckInCompleted(const chip::app::ICDClientInfo & clientInfo) override
+    {
+        chip::NodeId nodeId = clientInfo.peer_node.GetNodeId();
```
tehampson, line 46: For some reason this `nodeId` is not the value I am expecting

tehampson: Learning that this nodeId is the id of fabric admin...

examples/fabric-admin/commands/common/StayActiveSender.cpp
```diff
@@ -0,0 +1,81 @@
...
+    {
+        ChipLogError(ICD, "Failed to send stay active command");
+    }
+}
+
+void StayActiveSender::HandleDeviceConnectionFailure(void * context, const chip::ScopedNodeId & peerId, CHIP_ERROR err)
+{
+    StayActiveSender * const _this = static_cast<StayActiveSender *>(context);
+    VerifyOrDie(_this != nullptr);
+    ChipLogError(ICD, "Failed to establish CASE for stay active command with error '%" CHIP_ERROR_FORMAT "'", err.Format());;
```
tehampson, line 80: Don't we need to free/delete `_this`, otherwise we have a memory leak?

examples/fabric-admin/commands/common/IcdManager.cpp
```diff
@@ -32,6 +34,20 @@ void IcdManager::OnCheckInComplete(const chip::app::ICDClientInfo & clientInfo)
     {
         mDelegate->OnCheckInCompleted(clientInfo);
     }
+
+    uint32_t activeDurationMs = 30000;
+    auto stayActiveSender = chip::Platform::New<StayActiveSender>(activeDurationMs, clientInfo.peer_node, chip::app::InteractionModelEngine::GetInstance());
```
tehampson, line 39: On of the issues I previously mentioned is that `clientInfo.peer_node` is the node of fabric-admin, and not the LIT device checking-in

examples/fabric-admin/commands/common/IcdManager.cpp
```diff
@@ -32,6 +34,20 @@ void IcdManager::OnCheckInComplete(const chip::app::ICDClientInfo & clientInfo)
     {
         mDelegate->OnCheckInCompleted(clientInfo);
     }
+
```
tehampson, line 37: I think all of this code needs to be in the delegate not in the IcdManager, we need to be giving a `activeDurationMs` the value we got, not just this hard-coded 30000.

examples/fabric-admin/commands/common/IcdManager.cpp
```diff
@@ -0,0 +1,45 @@
...
+ *   Unless required by applicable law or agreed to in writing, software
+ *   distributed under the License is distributed on an "AS IS" BASIS,
+ *   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ *
+ */
+
+#include "IcdManager.h"
+#include <app/InteractionModelEngine.h>
```
andy31415, line 20: what do we use interactionmodelengine for ?

examples/fabric-admin/commands/common/IcdManager.h
```diff
@@ -0,0 +1,43 @@
...
+ *   See the License for the specific language governing permissions and
+ *   limitations under the License.
+ *
+ */
+
+#pragma once
+
+#include <app/icd/client/DefaultCheckInDelegate.h>
+
+class IcdManager : public chip::app::DefaultCheckInDelegate
```
andy31415, line 23: please add a doccomment blurb on why this class exist and its purpose and expected usage.

examples/fabric-admin/commands/common/StayActiveSender.cpp
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+#include "StayActiveSender.h"
+
+#include <app-common/zap-generated/cluster-objects.h>
+#include <app/ConcreteCommandPath.h>
+#include <controller/InvokeInteraction.h>
+#include <support/CHIPMem.h>
+
+StayActiveSender::StayActiveSender(uint32_t stayActiveDuration, const chip::ScopedNodeId & peerNode,
```
andy31415, line 25:
```suggestion
StayActiveSender::StayActiveSender(uint32_t stayActiveDurationMs, const chip::ScopedNodeId & peerNode,
```

please add unit of measurement for all variables holding values with units.

examples/fabric-admin/commands/common/StayActiveSender.cpp
```diff
@@ -0,0 +1,83 @@
...
+                                   chip::app::InteractionModelEngine * engine, OnDoneCallbackType onDone) :
+    mStayActiveDuration(stayActiveDuration),
```
andy31415, line 27:
```suggestion
    mStayActiveDurationMs(stayActiveDurationMs),
```

examples/fabric-admin/commands/common/StayActiveSender.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+CHIP_ERROR StayActiveSender::SendStayActiveCommand(chip::Messaging::ExchangeManager & exchangeMgr,
+                                                   const chip::SessionHandle & sessionHandle)
+{
+    auto onSuccess = [&](const chip::app::ConcreteCommandPath & commandPath, const chip::app::StatusIB & status,
+                         const auto & dataResponse) {
+        uint32_t promisedActiveDuration = dataResponse.promisedActiveDuration;
+        ChipLogProgress(ICD, "StayActive command succeeded with promised duration %u", promisedActiveDuration);
+        mOnDone(promisedActiveDuration);
+        chip::Platform::Delete(this);
```
andy31415, line 40: Is it guaranteed that this class is created with Platform::New?
Could we ensure this by making the constructor private and have a `Create` method instead, so that both New and Delete are always together in the same file?

examples/fabric-admin/commands/common/StayActiveSender.h
```diff
@@ -0,0 +1,78 @@
...
+
+#include <stddef.h>
+
+#include <app/InteractionModelEngine.h>
+#include <lib/core/ScopedNodeId.h>
+#include <messaging/ExchangeMgr.h>
+
+/**
+ * @brief StayActiveSender contains all the data and methods needed for active period extension of an ICD client.
+ */
```
andy31415, line 28: Please update this with lifetime description and usage: this class makes assumptions on when it auto-deletes, it looks like it auto-deletes if EstablishSessionToPeer returns CHIP_NO_ERROR but is expected to be deleted by the owner if no establish is called or if establish returns an error.

examples/fabric-admin/commands/common/IcdManager.cpp
tehampson: We don't here. It has been removed

examples/fabric-admin/commands/common/StayActiveSender.cpp
tehampson: Almost. Unfortunately `Platform::New` is not able to call private function :(

```
2024-08-13 20:08:03 INFO    ../../examples/fabric-admin/third_party/connectedhomeip/src/lib/support/CHIPMem.h: In instantiation of ‘T* chip::Platform::New(Args&& ...) [with T = StayActiveSender; Args = {unsigned int&, const chip::ScopedNodeId&, chip::app::InteractionModelEngine*, std::function<void(unsigned int)>&}]’:
2024-08-13 20:08:03 INFO    ../../examples/fabric-admin/commands/common/StayActiveSender.cpp:29:66:   required from here
2024-08-13 20:08:03 INFO    ../../examples/fabric-admin/third_party/connectedhomeip/src/lib/support/CHIPMem.h:148:16: error: ‘StayActiveSender::StayActiveSender(uint32_t, const chip::ScopedNodeId&, chip::app::InteractionModelEngine*, OnDoneCallbackType)’ is private within this context
2024-08-13 20:08:03 INFO      148 |         return new (p) T(std::forward<Args>(args)...);
2024-08-13 20:08:03 INFO          |                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
2024-08-13 20:08:03 INFO    ../../examples/fabric-admin/commands/common/StayActiveSender.cpp:40:1: note: declared private here
2024-08-13 20:08:03 INFO       40 | StayActiveSender::StayActiveSender(uint32_t stayActiveDurationMs, const chip::ScopedNodeId & peerNode,
```

tehampson: Let me know if the workaround is acceptable to you. I don't have a good alternative

examples/fabric-admin/commands/common/StayActiveSender.h
tehampson: Added comment, but also lifetime is now managed in StayActiveSender

andy31415 (approved):

yufengwangca (approved):

examples/fabric-admin/commands/common/IcdManager.cpp
```diff
@@ -0,0 +1,44 @@
...
+    DefaultCheckInDelegate::OnCheckInComplete(clientInfo);
+    if (mDelegate)
+    {
+        mDelegate->OnCheckInCompleted(clientInfo);
+    }
+}
+
+void IcdManager::SetDelegate(Delegate * delegate)
+{
+    // There is an assumption that there is only ever one delegate set and it's
```
saurabhst, line 39: whats the basis of this assumption ? Could you provide a bit of background ?

tehampson: If someone wants else calls `SetDelegate` they will get this crash, if they think that more than one delegate is needed they can change the assumption to suite their needs. But at the time of writing this we have this assumption here and codify it

examples/fabric-admin/commands/common/StayActiveSender.cpp
```diff
@@ -0,0 +1,99 @@
...
+    mPeerNode(peerNode), mpImEngine(engine), mOnDone(onDone), mOnConnectedCallback(HandleDeviceConnected, this),
+    mOnConnectionFailureCallback(HandleDeviceConnectionFailure, this)
+{}
+
+CHIP_ERROR StayActiveSender::SendStayActiveCommand(chip::Messaging::ExchangeManager & exchangeMgr,
+                                                   const chip::SessionHandle & sessionHandle)
+{
+    auto onSuccess = [&](const chip::app::ConcreteCommandPath & commandPath, const chip::app::StatusIB & status,
+                         const auto & dataResponse) {
+        uint32_t promisedActiveDuration = dataResponse.promisedActiveDuration;
```
saurabhst, line 53: should `promisedActiveDuration` also not be in terms of `promisedActiveDurationMs`?

examples/fabric-admin/commands/common/StayActiveSender.h
```diff
@@ -0,0 +1,107 @@
...
+    // Ideally this would be a private constructor, unfortunately that is not possible as Platform::New does not
+    // have access to private constructors. As a workaround we have defined a private struct that can be forwarded
+    // by Platform::New that allows us to enforce that the only way this is constructed is if SendStayActiveCommand
+    // is called.
+    StayActiveSender(const ConstructorOnlyInternallyCallable & _, uint32_t stayActiveDurationMs,
+                     const chip::ScopedNodeId & peerNode, chip::app::InteractionModelEngine * engine, OnDoneCallbackType onDone);
+
+private:
+    /**
+     * @brief Sets up a CASE session to the peer for extend a client active period with the peer.
```
saurabhst, line 68:
```suggestion
     * @brief Sets up a CASE session with the peer to extend the client's active period with that peer.
```

examples/fabric-admin/commands/common/IcdManager.cpp
tehampson: I will add more to this comment to make that a little more clear that the assumption can be changed, just that we are codifying the assumption based on current simple design of IcdManager

saurabhst (approved):

PR #34906, opened 2024-08-09
[FS Example] Update the FS Example apps to support fabric sync setup process part I

### LINKED ISSUE #33224: [Feature] Add Fabric Synchronization section to multi-admin in Matter SDK.
### Feature description

https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/9106

### Platform

all

### Platform Version(s)

_No response_

### Anything else?

_No response_

yufengwangca:
Updated the Fabric-Admin and Fabric-Bridge example apps to support the Fabric-Sync setup process. 

Please note that this PR includes only the changes to the example apps. The proto update needed to complete the remaining tasks will be addressed in a separate PR for review.

Fixes #33224


examples/fabric-admin/commands/clusters/ReportCommand.cpp
```diff
@@ -46,7 +46,7 @@ void ReportCommand::OnAttributeData(const app::ConcreteDataAttributePath & path,
 
     LogErrorOnFailure(RemoteDataModelLogger::LogAttributeAsJSON(path, data));
 
-    DeviceMgr().HandleAttributeChange(path, data);
+    DeviceMgr().HandleAttributeData(path, data);
```
andy31415, line 49: This redirects all attribute values. Please see device synchronization: we can use dedicated paths & loops instead of trying to hook in the main processing loop.

examples/fabric-admin/commands/fabric-sync/FabricSyncCommand.cpp
```diff
@@ -57,10 +83,25 @@ void FabricSyncAddBridgeCommand::OnCommissioningComplete(chip::NodeId deviceId,
                         ChipLogValueX64(mBridgeNodeId));
 
         char command[kMaxCommandSize];
-        snprintf(command, sizeof(command), "descriptor subscribe parts-list %d %d %ld %d", kSubscribeMinInterval,
+
+        // Prepare and push the descriptor subscribe command
+        snprintf(command, sizeof(command), "descriptor subscribe parts-list %d %d %" PRIu64 " %d", kSubscribeMinInterval,
```
andy31415, line 88: separate objects here as well: why are we using commands via strings? This would not be integratable to some other applications.

examples/fabric-bridge-app/fabric-bridge-common/include/CommissionerControl.h
```diff
@@ -0,0 +1,57 @@
...
+    ~CommissionerControlDelegate() = default;
+
+private:
+    static constexpr size_t kLabelBufferSize = 64;
+
+    uint64_t mRequestId;
+    NodeId mClientNodeId;
+    VendorId mVendorId;
+    uint16_t mProductId;
+    char mLabelBuffer[kLabelBufferSize + 1];
```
andy31415, line 47: this is example code. We can use std::string

examples/fabric-bridge-app/fabric-bridge-common/include/CommissionerControl.h
```diff
@@ -0,0 +1,57 @@
...
+    CHIP_ERROR GetCommissioningWindowParams(CommissioningWindowParams & outParams) override;
+    CHIP_ERROR ReverseCommissionNode(const CommissioningWindowParams & params, const Optional<ByteSpan> & ipAddress,
+                                     const Optional<uint16_t> & port) override;
+
+    ~CommissionerControlDelegate() = default;
+
+private:
+    static constexpr size_t kLabelBufferSize = 64;
+
+    uint64_t mRequestId;
```
andy31415, line 43: initialize things everywhere. Use `= 0` or similar (like invalid id constants)

examples/fabric-admin/commands/fabric-sync/FabricSyncCommand.cpp
```diff
@@ -57,10 +83,25 @@ void FabricSyncAddBridgeCommand::OnCommissioningComplete(chip::NodeId deviceId,
...
                  kSubscribeMaxInterval, mBridgeNodeId, kAggragatorEndpointId);
+        PushCommand(command);
 
+        // Prepare and push the commissioner control subscribe command
+        snprintf(command, sizeof(command),
```
andy31415, line 93: even if using snprintf, you should use `StringBuilder` instead that guarantees good sizing and null termination.

andy31415 (changes requested):
I understand this is under time pressure, but we need a path to make this updatable to non-string API to have it ready for integration. Please make these changes for now:

- [x] enter a github issue to switch to ReadClient based API instead of string API
- [x] requests (even if string based) should be done in the DeviceManager, so that both request and handling are in the same class. This will also make code migration easier in the future
- [x] replace snprintf with StringBuilder

examples/fabric-admin/commands/fabric-sync/FabricSyncCommand.cpp
```diff
@@ -39,14 +39,40 @@ constexpr uint16_t kSubscribeMinInterval    = 0;
...
+
+    if (bridgeNodeId == kUndefinedNodeId)
+    {
+        // print to console
+        fprintf(stderr, "Remote Fabric Bridge is not configured yet, halting reverse commissioning.\n");
+        return;
+    }
+
+    char command[kMaxCommandSize];
+    snprintf(command, sizeof(command), "commissionercontrol read supported-device-categories %ld %d", bridgeNodeId,
```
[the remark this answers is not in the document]
yufengwangca: Yes, this is needed for product integration. 

Replacing with direct read client need a lot of refactor to guarantee each command is processed as a integrity piece and all commands are executed in the right order. We plan to do this refactor in SVE.

https://github.com/project-chip/connectedhomeip/issues/34910

examples/fabric-bridge-app/fabric-bridge-common/include/CommissionerControl.h
yufengwangca: This is used as the backing buffer for label filed (Optional<chip::CharSpan> label;) in the RequestCommissioningApproval command since the original backing buffer will be invalid after returning from emberAfCommissionerControlClusterRequestCommissioningApprovalCallback

examples/fabric-admin/commands/fabric-sync/FabricSyncCommand.cpp
yufengwangca: https://github.com/project-chip/connectedhomeip/issues/34910

examples/fabric-admin/commands/clusters/ReportCommand.cpp
yufengwangca: Followup: https://github.com/project-chip/connectedhomeip/issues/34920

andy31415 (approved):

joonhaengHeo (approved):

PR #34917, opened 2024-08-09
TC-LVL-2.3: Update step 3 to set level immediately

cecille:
Previously, the device might still have been changing when we read the current value in step 5.

### DIFF
--- a/src/python_testing/TC_LVL_2_3.py
+++ b/src/python_testing/TC_LVL_2_3.py
@@ -43,15 +43,18 @@ class TC_LVL_2_3(MatterBaseTest):
         THRead = "TH reads"
         THcommand = "TH sends the command"
         return [TestStep(1, test_plan_support.commission_if_required(), is_commissioning=True),
-                TestStep(2, f"{THRead} FeatureMap attribute."),
-                TestStep(3, f"{THRead} MaxLevel attribute and store value as maxLevel", test_plan_support.verify_success()),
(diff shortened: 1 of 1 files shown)


src/python_testing/TC_LVL_2_3.py
```diff
@@ -90,16 +93,28 @@ async def test_TC_LVL_2_3(self):
...
+
+        self.step("3b")
+        if lvl.Attributes.MinLevel.attribute_id in attributes:
+            min_level = await self.read_single_attribute_check_success(cluster=lvl, attribute=lvl.Attributes.MinLevel)
+        else:
+            min_level = 1
 
         self.step(4)
-        cmd = Clusters.LevelControl.Commands.MoveWithOnOff(moveMode=lvl.Enums.MoveModeEnum.kDown)
+        cmd = Clusters.LevelControl.Commands.MoveToLevelWithOnOff(level=min_level, transitionTime=0)
```
jmartinez-silabs, line 113: you could use 0 instead of min level. it will move to the min level of the device regardeless

jmartinez-silabs (approved):

src/python_testing/TC_LVL_2_3.py
cecille: Discussed on slack - going to leave this as is for now. It's fairly clear what's happening to someone that doesn't know how the cluster works and both work essentially the same.

andy31415 (approved):

PR #34922, opened 2024-08-10
Mitigation to unauthenticated TCP support advertisement in DNS-SD that made it susceptible to potential DoS attacks.

pidarped:
Mitigation for DoS vulnerability of TCP params advertised over DNS-SD as described in (Spec Issue: https://github.com/CHIP-Specifications/connectedhomeip-spec/issues/8692)
The mitigation is to include the same advertised info inside the CASE session parameters so that they are available over an authenticated path if a prior MRP CASE session was established.

As part of this resolution,
Add TCP support and Max receive size info within negotiated CASE Session parameters.
--Nodes publish TCP support information during CASE session(over MRP)
   negotiations.
--Add TCP param storage logic to persist received TCP support info
   to storage during CASE session establishment and retrieve it during
   future TCP session setup.
--During TCP session setup, first check if the TCP support info
   for peer is available in storage before falling back to the DNS-SD
   advertised values.

### DIFF
--- a/src/lib/support/DefaultStorageKeyAllocator.h
+++ b/src/lib/support/DefaultStorageKeyAllocator.h
@@ -256,6 +256,14 @@ public:
     // when new fabric is created, this list needs to be updated,
     // when client init DefaultICDClientStorage, this table needs to be loaded.
     static StorageKeyName ICDFabricList() { return StorageKeyName::FromConst("g/icdfl"); }
+
+    // TCP peer parameters
+    static StorageKeyName TCPPeerParams(FabricIndex fabric, NodeId nodeId)
+    {
+        return StorageKeyName::Formatted("f/%x/tcp/%08" PRIX32 "%08" PRIX32, fabric, static_cast<uint32_t>(nodeId >> 32),
+                                         static_cast<uint32_t>(nodeId));
+    }
+    static StorageKeyName TCPPeerList() { return StorageKeyName::FromConst("g/tcpl"); }
 };
 
 } // namespace chip
--- a/src/messaging/SessionParameters.h
+++ b/src/messaging/SessionParameters.h
@@ -41,10 +41,19 @@ public:
     static constexpr size_t kSizeOfInteractionModelRevision = sizeof(uint16_t);
     static constexpr size_t kSizeOfSpecificationVersion     = sizeof(uint32_t);
     static constexpr size_t kSizeOfMaxPathsPerInvoke        = sizeof(uint16_t);
+#if INET_CONFIG_ENABLE_TCP_ENDPOINT
+    static constexpr size_t kSizeOfSupportedTransports = sizeof(uint16_t);
+    static constexpr size_t kSizeOfMaxTCPMessageSize   = sizeof(uint32_t);
+#endif // INET_CONFIG_ENABLE_TCP_ENDPOINT
 
     static constexpr size_t kEstimatedTLVSize = TLV::EstimateStructOverhead(
         kSizeOfSessionIdleInterval, kSizeOfSessionActiveInterval, kSizeOfSessionActiveThreshold, kSizeOfDataModelRevision,
-        kSizeOfInteractionModelRevision, kSizeOfSpecificationVersion, kSizeOfMaxPathsPerInvoke);
+        kSizeOfInteractionModelRevision, kSizeOfSpecificationVersion, kSizeOfMaxPathsPerInvoke
+#if INET_CONFIG_ENABLE_TCP_ENDPOINT
+        ,
+        kSizeOfSupportedTransports, kSizeOfMaxTCPMessageSize
+#endif // INET_CONFIG_ENABLE_TCP_ENDPOINT
+    );
 
     // From Section 4.12.8 "Parameters and Constants" in chapter "Secure Channel".
     enum Tag : uint32_t
@@ -56,6 +65,8 @@ public:
         kInteractionModelRevision = 5,
         kSpecificationVersion     = 6,
         kMaxPathsPerInvoke        = 7,
+        kSupportedTransports      = 8,
+        kMaxTCPMessageSize        = 9,
     };
 
     const ReliableMessageProtocolConfig & GetMRPConfig() const { return mMRPConfig; }
@@ -91,6 +102,13 @@ public:
     uint16_t GetMaxPathsPerInvoke() const { return mMaxPathsPerInvoke; }
     void SetMaxPathsPerInvoke(const uint16_t maxPathsPerInvoke) { mMaxPathsPerInvoke = maxPathsPerInvoke; }
 
+#if INET_CONFIG_ENABLE_TCP_ENDPOINT
+    const Optional<uint16_t> & GetSupportedTransports() const { return mSupportedTransports; }
+    void SetSupportedTransports(const uint16_t supportedTransports) { mSupportedTransports = MakeOptional(supportedTransports); }
+
+    const Optional<uint32_t> & GetMaxTCPMessageSize() const { return mMaxTCPMessageSize; }
+    void SetMaxTCPMessageSize(const uint32_t maxTCPMessageSize) { mMaxTCPMessageSize = MakeOptional(maxTCPMessageSize); }
+#endif // INET_CONFIG_ENABLE_TCP_ENDPOINT
 private:
     ReliableMessageProtocolConfig mMRPConfig;
     // For legacy reasons if we do not get DataModelRevision it means either 16 or 17. But there isn't
@@ -104,6 +122,11 @@ private:
     Optional<uint32_t> mSpecificationVersion;
     // When maxPathsPerInvoke is not provided legacy is always 1
     uint16_t mMaxPathsPerInvoke = 1;
+
+#if INET_CONFIG_ENABLE_TCP_ENDPOINT
+    Optional<uint16_t> mSupportedTransports;
+    Optional<uint32_t> mMaxTCPMessageSize;
+#endif // INET_CONFIG_ENABLE_TCP_ENDPOINT
 };
 
 } // namespace chip
--- a/src/protocols/secure_channel/PairingSession.cpp
+++ b/src/protocols/secure_channel/PairingSession.cpp
@@ -21,6 +21,7 @@
 #include <app/SpecificationDefinedRevisions.h>
 #include <lib/core/CHIPConfig.h>
 #include <lib/core/TLVTypes.h>
+#include <lib/dnssd/Advertiser.h>
 #include <lib/support/SafeInt.h>
 #include <lib/support/TypeTraits.h>
 #include <platform/CHIPDeviceEvent.h>
@@ -142,6 +143,16 @@ CHIP_ERROR PairingSession::EncodeSessionParameters(TLV::Tag tag, const ReliableM
 
     uint16_t maxPathsPerInvoke = CHIP_CONFIG_MAX_PATHS_PER_INVOKE;
     ReturnErrorOnFailure(tlvWriter.Put(TLV::ContextTag(SessionParameters::Tag::kMaxPathsPerInvoke), maxPathsPerInvoke));
+
+#if INET_CONFIG_ENABLE_TCP_ENDPOINT
+    // Support for both TCP CLient and Server(0x06)
+    uint16_t supportedTransports = to_underlying(Dnssd::TCPModeAdvertise::kTCPClientServer);
+    ReturnErrorOnFailure(tlvWriter.Put(TLV::ContextTag(SessionParameters::Tag::kSupportedTransports), supportedTransports));
+
+    uint32_t maxTCPMessageSize = System::PacketBuffer::kLargeBufMaxSizeWithoutReserve - 4 /* Framing length field of 4 bytes */;
+    ReturnErrorOnFailure(tlvWriter.Put(TLV::ContextTag(SessionParameters::Tag::kMaxTCPMessageSize), maxTCPMessageSize));
+#endif // INET_CONFIG_ENABLE_TCP_ENDPOINT
+
     return tlvWriter.EndContainer(mrpParamsContainer);
 }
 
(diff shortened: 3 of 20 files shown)


aronrosenberg (approved):

src/lib/support/DefaultStorageKeyAllocator.h
```diff
@@ -256,6 +256,14 @@ class DefaultStorageKeyAllocator
     // when new fabric is created, this list needs to be updated,
     // when client init DefaultICDClientStorage, this table needs to be loaded.
     static StorageKeyName ICDFabricList() { return StorageKeyName::FromConst("g/icdfl"); }
+
+    // TCP peer parameters
+    static StorageKeyName TCPPeerParams(FabricIndex fabric, NodeId nodeId)
+    {
+        return StorageKeyName::Formatted("f/%x/tcp/%08" PRIX32 "%08" PRIX32, fabric, static_cast<uint32_t>(nodeId >> 32),
```
andy31415, line 263: why is the static cast assumed to be ok? Why is the node id truncation assumed ok?

src/protocols/secure_channel/PairingSession.cpp
```diff
@@ -142,6 +142,16 @@ CHIP_ERROR PairingSession::EncodeSessionParameters(TLV::Tag tag, const ReliableM
 
     uint16_t maxPathsPerInvoke = CHIP_CONFIG_MAX_PATHS_PER_INVOKE;
     ReturnErrorOnFailure(tlvWriter.Put(TLV::ContextTag(SessionParameters::Tag::kMaxPathsPerInvoke), maxPathsPerInvoke));
+
+#if INET_CONFIG_ENABLE_TCP_ENDPOINT
+    // Support for both TCP CLient and Server(0x06)
+    uint16_t supportedTransports = 0x06;
```
andy31415, line 148: please set bit flags constants and use `|`. This should read like `kTcpClient | kTcpServer` or similar. Do not hardcode magic numbers.

src/transport/raw/PeerTCPParamsStorage.cpp
```diff
@@ -0,0 +1,339 @@
...
+StorageKeyName PeerTCPParamsStorage::GetStorageKey(const ScopedNodeId & node)
+{
+    return DefaultStorageKeyAllocator::TCPPeerParams(node.GetFabricIndex(), node.GetNodeId());
+}
+
+CHIP_ERROR PeerTCPParamsStorage::FindByScopedNodeId(const ScopedNodeId & node, uint16_t & supportedTransports,
+                                                    uint32_t & maxTCPMessageSize)
+{
+    ReturnErrorOnFailure(LoadTCPParamsFromStorage(node, supportedTransports, maxTCPMessageSize));
+    return CHIP_NO_ERROR;
```
andy31415, line 45:
```suggestion
    return LoadTCPParamsFromStorage(node, supportedTransports, maxTCPMessageSize);
```

src/transport/raw/PeerTCPParamsStorage.cpp
```diff
@@ -0,0 +1,339 @@
...
+}
+
+CHIP_ERROR PeerTCPParamsStorage::SaveTrackedNodesList(const TrackedNodesList & list)
+{
+    std::array<uint8_t, MaxNodesListSize()> buf;
```
andy31415, line 50: how much is this MaxNodesListSize?

src/transport/raw/PeerTCPParamsStorage.cpp
```diff
@@ -0,0 +1,339 @@
...
+    ReturnErrorOnFailure(mTCPParamsStorage->SyncSetKeyValue(DefaultStorageKeyAllocator::TCPPeerList().KeyName(), buf.data(),
+                                                            static_cast<uint16_t>(len)));
+
+    return CHIP_NO_ERROR;
+}
+
+CHIP_ERROR PeerTCPParamsStorage::LoadTrackedNodesList(TrackedNodesList & list)
+{
+    std::array<uint8_t, MaxNodesListSize()> buf;
+    uint16_t len = static_cast<uint16_t>(buf.size());
```
andy31415, line 80:
```suggestion
    const uint16_t len = static_cast<uint16_t>(buf.size());
```

andy31415: Maybe consider creating a Span (initializing it out of an array initializes both buffer and length for it)

src/transport/raw/PeerTCPParamsStorage.cpp
```diff
@@ -0,0 +1,339 @@
...
+    CHIP_ERROR err = DeleteTCPParamsFromStorage(node);
+    if (err != CHIP_NO_ERROR && err != CHIP_ERROR_PERSISTED_STORAGE_VALUE_NOT_FOUND)
+    {
+        ChipLogError(SecureChannel,
+                     "Unable to delete TCP params info from storage for node " ChipLogFormatX64 ": %" CHIP_ERROR_FORMAT,
+                     ChipLogValueX64(node.GetNodeId()), err.Format());
+    }
+
+    bool found = false;
+    size_t i;
```
andy31415, line 179: rename i to `indexToRemove` or similar since its lifetime is longer than just a simple for loop.
Do you need `found`? it seems that if this index is `>= list.mSize` you did not find it.

src/transport/raw/PeerTCPParamsStorage.cpp
```diff
@@ -0,0 +1,339 @@
...
+    for (i = 0; i < list.mSize; i++)
+    {
+        if (list.mNodes[i] == node)
+        {
+            found = true;
+            break;
+        }
+    }
+
+    if (found)
```
andy31415, line 189: decrease indend by doing:

```cpp
if (indexToRemove >= list.mSize)
{
   ChipLog....;
   return CHIP_NO_ERROR;
}

File: src/transport/raw/PeerTCPParamsStorage.cpp [hunk 8]
```diff
@@ -0,0 +1,339 @@
...
+    {
+        // Shift elements to the left
+        for (size_t j = i; j < list.mSize - 1; j++)
```
andy31415, line 192: Is this maybe shorter? Or is it more error prone?

```cpp
std::copy(
    list.mNodes + indexToRemove + 1,
    list.mNodes + list.mSize, 
    list.mNodes + indexToRemove);
```

src/transport/raw/PeerTCPParamsStorage.cpp
```diff
@@ -0,0 +1,339 @@
...
+    return CHIP_NO_ERROR;
+}
+
+CHIP_ERROR PeerTCPParamsStorage::DeleteTCPParams(const ScopedNodeId & node)
+{
+    TrackedNodesList list;
+    ReturnErrorOnFailure(LoadTrackedNodesList(list));
+
+    CHIP_ERROR err = DeleteTCPParamsFromStorage(node);
+    if (err != CHIP_NO_ERROR && err != CHIP_ERROR_PERSISTED_STORAGE_VALUE_NOT_FOUND)
```
andy31415, line 171: none of this CHIP_ERRORS do anytying except logging. Why is that? The only possible error seems to be the LoadTrackedNodeList.

Some comments explaining that decision would help.

src/transport/raw/PeerTCPParamsStorage.cpp
andy31415: Is the stack ok? How about a ScopedMemoryBuffer so we use heap and tools will detect out of bounds if it applies? We seem to assume large systems here anyway. Large buffers on the stack seem worse than large buffers in the heap.

src/transport/raw/PeerTCPParamsStorage.cpp
```diff
@@ -0,0 +1,339 @@
...
+
+    ReturnErrorOnFailure(writer.EndContainer(arrayType));
+
+    const auto len = writer.GetLengthWritten();
+    VerifyOrReturnError(CanCastTo<uint16_t>(len), CHIP_ERROR_BUFFER_TOO_SMALL);
+
+    ReturnErrorOnFailure(mTCPParamsStorage->SyncSetKeyValue(DefaultStorageKeyAllocator::TCPPeerList().KeyName(), buf.data(),
+                                                            static_cast<uint16_t>(len)));
+
+    return CHIP_NO_ERROR;
```
andy31415, line 74:
```suggestion
    return mTCPParamsStorage->SyncSetKeyValue(DefaultStorageKeyAllocator::TCPPeerList().KeyName(), buf.data(),
                                                            static_cast<uint16_t>(len));
```

src/transport/raw/PeerTCPParamsStorage.cpp
```diff
@@ -0,0 +1,339 @@
...
+    }
+
+    ReturnErrorOnFailure(SaveTCPParamsToStorage(node, supportedTransports, maxTCPMessageSize));
+
+    list.mNodes[list.mSize++] = node;
+
+    // Save back the tracked nodes list
+    ReturnErrorOnFailure(SaveTrackedNodesList(list));
+
+    return CHIP_NO_ERROR;
```
andy31415, line 162:
```suggestion
    return SaveTrackedNodesList(list);
```

src/transport/raw/PeerTCPParamsStorage.cpp
```diff
@@ -0,0 +1,339 @@
...
+    ReturnErrorOnFailure(writer.Put(kSupportedTransportsTag, supportedTransports));
+    ReturnErrorOnFailure(writer.Put(kMaxTCPMessageSizeTag, maxTCPMessageSize));
+
+    ReturnErrorOnFailure(writer.EndContainer(outerType));
+
+    const auto len = writer.GetLengthWritten();
+    VerifyOrDie(CanCastTo<uint16_t>(len));
+
+    ReturnErrorOnFailure(mTCPParamsStorage->SyncSetKeyValue(GetStorageKey(node).KeyName(), buf.data(), static_cast<uint16_t>(len)));
+    return CHIP_NO_ERROR;
```
andy31415, line 302:
```suggestion
    return mTCPParamsStorage->SyncSetKeyValue(GetStorageKey(node).KeyName(), buf.data(), static_cast<uint16_t>(len));
```

src/transport/raw/PeerTCPParamsStorage.cpp
```diff
@@ -0,0 +1,339 @@
...
+    ReturnErrorOnFailure(reader.ExitContainer(containerType));
+    ReturnErrorOnFailure(reader.VerifyEndOfContainer());
+
+    return CHIP_NO_ERROR;
+}
+
+CHIP_ERROR PeerTCPParamsStorage::DeleteTCPParamsFromStorage(const ScopedNodeId & node)
+{
+    ReturnErrorOnFailure(mTCPParamsStorage->SyncDeleteKeyValue(GetStorageKey(node).KeyName()));
+    return CHIP_NO_ERROR;
```
andy31415, line 335:
```suggestion
    return mTCPParamsStorage->SyncDeleteKeyValue(GetStorageKey(node).KeyName());
```

src/transport/raw/PeerTCPParamsStorage.cpp
pidarped: Keeping it as a buffer and a non-const buffer size aligns with API requirements of the subsequent SyncGetKeyValue() call.

src/transport/SessionManager.h
```diff
@@ -566,6 +568,7 @@ class DLL_EXPORT SessionManager : public TransportMgrDelegate, public FabricTabl
     // Hold the TCPConnection callback context for the receiver application in the SessionManager.
     // On receipt of a connection from a peer, the SessionManager
     Transport::AppTCPConnectionCallbackCtxt * mServerTCPConnCbCtxt = nullptr;
+    Transport::PeerTCPParamsStorage mPeerTCPParamsStorage;
```
bzbarsky-apple, line 571: This should probably not be hardcoding a specific storage strategy.  The way this sort of thing is normally done is:

1. Have a delegate API that is implementation-agnostic and exposes the high-level primitives that code actually depends on.
2. Have a default implementation of that API that uses a storage delegate as desired.

That allows SDK consumers to use a different implementation of the delegate API as needed.

In particular, clients must not require new "g/whatever" keys, in general, and this implementation here very much requires that.

bzbarsky-apple (changes requested):

src/transport/raw/PeerTCPParamsStorage.cpp
bzbarsky-apple: What happens if this size is exceeded?  What happens if a client needs to speak to more than 32 different nodes?

src/messaging/SessionParameters.h
```diff
@@ -56,6 +65,8 @@ class SessionParameters
         kInteractionModelRevision = 5,
         kSpecificationVersion     = 6,
         kMaxPathsPerInvoke        = 7,
+        kSupportedTransports      = 8,
+        kMaxTCPMessageSize        = 9,
```
bzbarsky-apple, line 69: If this is the quantity called MAX_TCP_PAYLOAD_SIZE in the spec, why is the naming here different, by the way?

pidarped: I believe I had it as MaxTCPPayload at the place in the TCP layer where it was used but it got renamed to MaxTCPMessageSize based on review suggestions. I will put up a spec PR to rename it there so that they are in sync.

src/transport/SessionManager.h
pidarped: Ok. I wasn't fully aware of the dependence model for clients and thought this to be a core piece and part of the deeper transport layer and so directly implemented a concrete class. Going to separate into an Interface and Default Impl classes.

src/transport/raw/PeerTCPParamsStorage.cpp
pidarped: One of the nodes(index 0) gets replaced. This is also a compile-time configurable value in TCPConfig.h.

bzbarsky-apple: > One of the nodes(index 0) gets replaced

I guess the real question is: if it's totally ok to lose track of this info for nodes, why are we keeping track of it at all?

pidarped: Keeping track of the nodes list helps with management of the stored info.
The nodeslist has a finite size and will eventually fill up if there are more nodes that are being interacted with. But if an entry is replaced by a new one when the list is full, it can always be inserted back when a CASE negotiation happens again with that node.

bzbarsky-apple: That's not what I asked.  What I asked is: why are we doing _any_ of the work this PR does if it's totally ok to just lose the stored information?

pidarped: Losing the information is not ok, but only an outcome of limited resources. The list being finite, the node could reach a point where it would lose the info for some peer node if the latter was replaced. But at the same time, there is the underlying premise that nodes would establish CASE over MRP for other various operations(subscriptions, events, time sync, commands whatever) from time to time, and the node info would get replenished passively. We can optimize further by implementing a better cache replacement strategy(LRU, etc). Also, increasing the size of the nodes list to a larger value, say 64 or 128 might practically eliminate the replacement loss. Resource-wise, this should be ok on a device that supports TCP.

src/transport/SessionManager.h
pidarped: Created an interface class and an implementation of it. The Server and Controller sides can be initialized with an implementation of that via the Init params.

Issue #30718, opened 2023-11-29
Clarify how Aes128KeyHandle works


yunhanw-google:
This is assuming, as part of this API, that the actual key bytes are stored in the key handle in the other info, no?  Why is this a good assumption?  It seems like it would preclude secure key stores...

And in particular, doesn't whether a keyhandle is in fact a Aes128KeyByteArray depend on the crypto implementation involved?

Or do I completely misunderstand how `Aes128KeyHandle` works?

@Damian-Nordic how can this be set up to work?

Followup is fine to sort this out.

_Originally posted by @bzbarsky-apple in https://github.com/project-chip/connectedhomeip/pull/29783#discussion_r1409830647_

Damian-Nordic:
`Aes128KeyHandle` is an opaque type which can either hold `Aes128KeyByteArray` directly or it can hold just a key reference, in which case the key is stored securely in HSM or secure zone.

I agree with Boris' comment that the common code should not assume the key handle always holds a raw key. We expect that in the future there's going to be push towards better isolation of applications from security material, so it's worth designing interfaces allowing for that. 

I think that you could probably write this code in a way that copying is not necessary. Perhaps, we could just make `Aes128KeyHandle` movable to be able to use `ICDClientInfo` in `std::vector`. Btw. we probably will need another implementation of the storage as we avoid using STL dynamic containers in embedded to keep to code size reasonable.

bzbarsky-apple:
So right now ICD code assumes that `As<Aes128KeyByteArray>` works; e.g. see `src/app/icd/ICDMonitoringTable.cpp` and `src/app/icd/ICDCheckInSender.cpp`.

We need to figure out how to make those bits work without that, or to make it OK to do that, right?

In particular, unlike session keys the symmetric keys used for ICD need to be persistable somehow...

@Damian-Nordic @mkardous-silabs

Damian-Nordic:
Hmm. I think that we have the following options:
a) easy, non-secure:
Keep the ICD key as raw bytes in `ICDCheckinSender` and `ICDClientInfo`, and only import the key to the session keystore each time when a check-in message is sent (and then immediately remove it from the session keystore).
It doesn't provide a good isolation of the keys but maybe for the ICD use-case this is OK, and for sure this is more correct than assuming that `Aes128KeyHandle` contains raw bytes.

b) hard, secure:
[...]

PR #34925, opened 2024-08-12
[ICD] Implement storing of persistent keys in PSA for ICD server

maciejbaczmanski:
Description

Previously keys for ICD server were store as volatile keys in PSA, causing failures in Check-In Protocol after device reboot. This PR implements storing of persistent keys in PSA for ICD server.

* Refactor Key Attibutes classes, Implement `keyAttributesBase` and add persistent lifetime functionality.
* Implement `FindFreeKeySlotInRange` for PSA key slots
* Fix ICD CIP related build dependencies
* Remake ICDMonitoringTable::Find() not to overwrite all entry fields - If entry is not found its key handle field must not be filled with last checked entry when using PSA as it will cause key slot to be cleared by accident.
* Implement setting key persitence for ICD server

Testing
Tested on nRF52840 DK with PSA Crypto API enabled


src/crypto/PSASessionKeystore.cpp
```diff
@@ -24,62 +24,54 @@ namespace Crypto {
...
-        psa_set_key_bits(&mAttrs, CHIP_CRYPTO_SYMMETRIC_KEY_LENGTH_BYTES * 8);
+        psa_set_key_type(&mAttrs, type);
+        psa_set_key_algorithm(&mAttrs, algorithm);
+        psa_set_key_usage_flags(&mAttrs, usageFlags);
+        psa_set_key_bits(&mAttrs, bits);
+        if (to_underlying(KeyIdBase::Maximum) >= keyId && keyId >= to_underlying(KeyIdBase::Minimum))
+        {
+            psa_set_key_lifetime(&mAttrs, PSA_KEY_LIFETIME_PERSISTENT);
+            psa_set_key_id(&mAttrs, keyId);
+        }
```
mkardous-silabs, line 40: I'm not sure this is the best approach to configure the persistance of a PSA key. This seems to be "hidding" the persistance configuration behind the KeyID.

When we transitionned the ICD monitoring entries to be able to use PSA, we had established that a new API should be added to the SymmetricKeyStore (SessionKeyStore rename not done yet). See https://github.com/project-chip/connectedhomeip/issues/30718#issuecomment-1839093000

src/app/icd/server/ICDMonitoringTable.cpp
```diff
@@ -140,7 +144,20 @@ CHIP_ERROR ICDMonitoringEntry::SetKey(ByteSpan keyData)
...
+#ifdef CONFIG_CHIP_CRYPTO_PSA
+    if (persistent)
+    {
+        ReturnErrorOnFailure(Crypto::FindFreeKeySlotInRange(aesKeyHandle.AsMutable<psa_key_id_t>(),
+                                                            to_underlying(Crypto::KeyIdBase::ICDAesKeyRangeStart),
+                                                            Crypto::kMaxICDClientKeys));
+        ReturnErrorOnFailure(Crypto::FindFreeKeySlotInRange(hmacKeyHandle.AsMutable<psa_key_id_t>(),
+                                                            to_underlying(Crypto::KeyIdBase::ICDHmacKeyRangeStart),
+                                                            Crypto::kMaxICDClientKeys));
+    }
```
mkardous-silabs, line 156: This seems like it shouldn't be exposed within the ICDMonitoringTable but rather in the createKey API. 
The goal of the ICDMonitoringTable was to keep as agnostic as possible of the underlying crypto implementation.
In other words, we want to avoid as much as possible `#ifdef CONFIG_CHIP_CRYPTO_PSA` which in this case can be done by improving the Keystore APIs.

src/app/icd/server/ICDMonitoringTable.cpp
```diff
@@ -242,15 +259,21 @@ CHIP_ERROR ICDMonitoringTable::Get(uint16_t index, ICDMonitoringEntry & entry) c
...
-    while (index < this->Limit())
+    CHIP_ERROR err;
+    uint16_t index;
+    ICDMonitoringEntry tempEntry(mSymmetricKeystore);
+
+    for (index = 0; index < this->Limit(); index++)
     {
-        ReturnErrorOnFailure(this->Get(index++, entry));
-        if (id == entry.checkInNodeID)
+        SuccessOrExit(err = this->Get(index, tempEntry));
```
mkardous-silabs, line 268: I'm not sure i understand why the PSA key slots would not be clearer by this implementation compared of the previous one.

mkardous-silabs (commented):
The PSA ICD implementation seems to be adding unnecessary "optional" code paths in the ICD code that should be as much as possbile agnostic to the underlying crypto flavor.

src/app/icd/server/ICDMonitoringTable.cpp
maciejbaczmanski: here we first load only to `tempEntry`. and then we have two options:
- entry is found, all data from `tempEntry` is copied to `entry`. Then if we call `CreateKey` it will be cleared (the same behavior as before)
- entry is not found, only `entry.index` is updated upon `exit:` . This way `CreateKey` will call clearing a default handler, which is zero-initialized (and that handle is always an invalid handle for PSA). Before, `entry` would keep the key handles of last loaded entry (because of `ReturnErrorOnFailure(this->Get(index++, entry));`). So when we called key creation on that, it would call clearing an already existing slot from a different entry.

src/app/icd/server/ICDMonitoringTable.cpp
maciejbaczmanski: anwsering this and partially previous comment (https://github.com/project-chip/connectedhomeip/pull/34925#discussion_r1713639169):
I think we could create a separate API for setting the persistence. the limitation I had with that approach is the fact that i.e `Aes128KeyHandle` is used for other cryptographic operations as well. Because of that, we cannot associate it with key slot ranges as it won't be generic. I guess the solution for that would be creating ICD- specific key handles, and add overloads of `CreateKey` APIs for these specific handles that would handle key range and setting the persistence as well. The only thing that worries me is the fact that it would look good for two keys, but the solution is not really scalable- adding more PSA persistent keys in the future will result with lots of key-specific handles.

src/crypto/PSASessionKeystore.cpp
maciejbaczmanski: > I'm not sure this is the best approach to configure the persistance of a PSA key. This seems to be "hidding" the persistance configuration behind the KeyID.
> 
That is true, but all key IDs that are within `KeyIdBase::Minimum` and `KeyIdBase::Maximum` **should** be persistent. This is because that range is defined as matter-specific persistent key range. AFAIK volatile keys are created in a different range, but will check that. It is also crucial not to modify this key range as it will break backwards compatibility.

My intention of this implementation was to leave as little room for error as possible. this way we allow setting persistence only for the given range, and all keys that should be persisted are defined in one place, under `KeyIdBase` https://github.com/project-chip/connectedhomeip/pull/34925/files#diff-b7b07e88bffdc16e60f9b000afbc1eba7e2ff172b54bb9701df597ee01da77acR66 to avoid possibility of a conflict between two the same Key IDs

maciejbaczmanski:
@mkardous-silabs I've thought about your concerns and found that we could set the key persistence in `Set()`, which is called directly after `SetKey()` for entries with persistent keys. This way we could make the ICD layer agnostic to crypto (but `PersistICDKey()` methods need to be added, and because of key range, I cannot find a way to create a generic `PersistKey()` method).
I've created a quic PoC over here: https://github.com/project-chip/connectedhomeip/compare/master...maciejbaczmanski:connectedhomeip:matter_icd_psa_bck
After a small smoke test it seems like it could work. 
Could you take a look and let me know what you think about the other approach? I'll refactor it and test tomorrow.

src/crypto/PSASessionKeystore.cpp
```diff
@@ -24,62 +24,54 @@ namespace Crypto {
 
 namespace {
 
-class AesKeyAttributes
+class keyAttributesBase
```
kkasperczyk-no, line 27: This is a minor, but why does this class name start with lower case?

src/app/icd/server/ICDMonitoringTable.cpp
```diff
@@ -140,7 +144,20 @@ CHIP_ERROR ICDMonitoringEntry::SetKey(ByteSpan keyData)
     Crypto::Symmetric128BitsKeyByteArray keyMaterial;
     memcpy(keyMaterial, keyData.data(), sizeof(Crypto::Symmetric128BitsKeyByteArray));
 
-    // TODO - Add function to set PSA key lifetime
+#ifdef CONFIG_CHIP_CRYPTO_PSA
```
kkasperczyk-no, line 147: We can't use `CONFIG_CHIP_CRYPTO_PSA` here, as this is Kconfig and it does not exist for non-Zephyr based platforms. I think we could use `CHIP_CRYPTO_PSA` crypto define instead.

src/crypto/PSASessionKeystore.cpp
maciejbaczmanski: must have make it lowercase by mistake. fixed

src/app/icd/server/ICDMonitoringTable.cpp
maciejbaczmanski: Right, didn't notice `CONFIG_` prefix... but for the new approach it is not needed anymore, fortunately :)

src/app/icd/server/ICDMonitoringTable.cpp
maciejbaczmanski: Changed the approach to make ICD layer agnostic to crypto

maciejbaczmanski:
> @mkardous-silabs I've thought about your concerns and found that we could set the key persistence in `Set()`, which is called directly after `SetKey()` for entries with persistent keys. This way we could make the ICD layer agnostic to crypto (but `PersistICDKey()` methods need to be added, and because of key range, I cannot find a way to create a generic `PersistKey()` method). 

Fyi, after testing and refactoring I've decided to change the approach to described above. PR is ready for review again :)

maciejbaczmanski:
Sorry for modifying the PR so often.. This is the last change- it splits the current PSA key range into two subranges- one for mandatory keys that should be always supported, and one for optional, that depend on extra configurations.

This approach helps to prevent any mix-up between keys that are always active and those that are dependent on extra configurations. Moreover, if there's a need to activate a previously disabled range, it only necessitates the potential migration of other optional key ranges.

kkasperczyk-no (approved):

maciejbaczmanski:
@mkardous-silabs can I count on your review?

src/app/icd/server/ICDMonitoringTable.cpp
```diff
@@ -263,16 +268,26 @@ CHIP_ERROR ICDMonitoringTable::Set(uint16_t index, const ICDMonitoringEntry & en
...
+    e.clientType        = entry.clientType;
+    e.index             = index;
+    e.symmetricKeystore = entry.symmetricKeystore;
 
     memcpy(e.aesKeyHandle.AsMutable<Crypto::Symmetric128BitsKeyByteArray>(),
            entry.aesKeyHandle.As<Crypto::Symmetric128BitsKeyByteArray>(), sizeof(Crypto::Symmetric128BitsKeyByteArray));
     memcpy(e.hmacKeyHandle.AsMutable<Crypto::Symmetric128BitsKeyByteArray>(),
            entry.hmacKeyHandle.As<Crypto::Symmetric128BitsKeyByteArray>(), sizeof(Crypto::Symmetric128BitsKeyByteArray));
 
+    ReturnErrorOnFailure(e.symmetricKeystore->PersistICDKey(e.aesKeyHandle));
```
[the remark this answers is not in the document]
Damian-Nordic: > in `ICDMonitoringTable::Remove`, `entry.DeleteKey()` is called and the key is destroyed. I've checked if there are no potential key leaks and couldn't spot anything. During some smoke tests, keys have been destroyed correctly as well. Can you see any more cases when the key should be destroyed but it itsn't now?

Ah, right. After double-checking it looks correct.

> I thought about that, but for ICD entries, `index` is always shifted when entry is removed (see https://github.com/project-chip/connectedhomeip/blob/master/src/app/icd/server/ICDMonitoringTable.cpp#L288). So If I wanted to assign the key with `index`, we would need to remove that mechanism in `Remove` and add a similar one to current search for free key ID in `Set` (we would still need to find next free index). I've decided to just implement `FindFreeKeySlotInRange` as it is quite universal and could come in handy in the future if "dynamic" PSA key slot allocation would be needed somewhere else

Hmm. Even if you derived the key ID from `index`, the shifting mechanism would work correctly because during the shift `ICDMonitoringTable::Set` would be called, which would _move_ the corresponding PSA keys into the new IDs, right?

If you want to avoid moving the keys, shouldn't you exit early in `PSASessionKeystore::PersistICDKey` if the key is already persistent?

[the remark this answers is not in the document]
maciejbaczmanski: > Hmm. Even if you derived the key ID from `index`, the shifting mechanism would work correctly because during the shift `ICDMonitoringTable::Set` would be called, which would _move_ the corresponding PSA keys into the new IDs, right?
> 
> If you want to avoid moving the keys, shouldn't you exit early in `PSASessionKeystore::PersistICDKey` if the key is already persistent?

Oh. I forgot to add exiting early for already present persistent keys.. Thanks for spotting that! The idea was to minimize moving of the keys, so with that fixed, only key ID will be copied, and the only way to change key itself should be through `SetKey` . With index derivation only we would need to add yet another crypto API for key copying.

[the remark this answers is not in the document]
Damian-Nordic: > With index derivation only we would need to add yet another crypto API for key copying.

Not if `PersistICDKey` would take the index and always moved the key to the new ID, but I'm OK with the current solution if the early exiting is added :).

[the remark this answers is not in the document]
maciejbaczmanski: Added early exit in `PersistICDKey`

src/crypto/CHIPCryptoPALPSA.h
```diff
@@ -56,18 +57,95 @@ namespace Crypto {
...
+    Maximum     = DACPrivKey,
 };
 
-static_assert(to_underlying(KeyIdBase::Minimum) >= PSA_KEY_ID_USER_MIN && to_underlying(KeyIdBase::Maximum) <= PSA_KEY_ID_USER_MAX,
+static_assert(to_underlying(KeyIdBase::Minimum) >= CHIP_CONFIG_CRYPTO_PSA_KEY_ID_BASE &&
+                  to_underlying(KeyIdBase::Maximum) < CHIP_CONFIG_CRYPTO_PSA_KEY_ID_OPTIONAL,
               "PSA key ID base out of allowed range");
 
+#if CHIP_CONFIG_ENABLE_ICD_CIP
+static constexpr uint32_t kMaxICDClientKeys = CHIP_CONFIG_ICD_CLIENTS_SUPPORTED_PER_FABRIC * CHIP_CONFIG_MAX_FABRICS;
```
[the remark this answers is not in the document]
maciejbaczmanski: Ok, I've added `CHIP_CONFIG_ICD_MAX_CLIENTS` and set it to 256 by default. Could you take a look and check if I've put it in correct place?

src/lib/core/CHIPConfig.h
```diff
@@ -1667,6 +1667,22 @@ extern const char CHIP_NON_PRODUCTION_MARKER[];
...
+ * @brief
+ *   Maximum number of ICD clients. Based on this number, platforms that utilize the
+ *   PSA Crypto API should reserve key slot range.
+ *
+ * @note
+ *   For platforms that utilize the PSA Crypto API, this configuration is used to
+ *   compute the number of PSA key slots. It should remain unchanged during the device's lifetime,
+ *   as alterations may lead to issues with backwards compatibility.
+ */
+#ifndef CHIP_CONFIG_ICD_MAX_CLIENTS
```
Damian-Nordic, line 1682: nit: consider making this config specifically for the PSA crypto, such as `CHIP_CONFIG_CRYPTO_PSA_ICD_MAX_CLIENTS` as if it's made too generic, it may be mistakenly used in other contexts where it's assumed to mean the current maximum, not maximum maximum :D.

src/crypto/SessionKeystore.h
```diff
@@ -140,6 +140,22 @@ class SessionKeystore
     virtual CHIP_ERROR DeriveSessionKeys(const HkdfKeyHandle & secretKey, const ByteSpan & salt, const ByteSpan & info,
                                          Aes128KeyHandle & i2rKey, Aes128KeyHandle & r2iKey,
                                          AttestationChallenge & attestationChallenge) = 0;
+
+    /**
+     * @brief Store key in persistent PSA storage and return a key handle for an ICD Aes key.
```
Damian-Nordic, line 145: This is a generic interface and should not mention PSA. It's worth documenting that if the input is already a persistent key handle, the function is a no-op. Otherwise, if the key references a volatile key, this gets persisted and the key handle may be updated to reference the persistent key.

src/crypto/PSASessionKeystore.cpp
```diff
@@ -192,5 +189,72 @@ void PSASessionKeystore::DestroyKey(HkdfKeyHandle & key)
...
+    VerifyOrExit(psa_copy_key(previousKeyId, &attrs.Get(), &key.AsMutable<psa_key_id_t>()) == PSA_SUCCESS,
+                 err = CHIP_ERROR_INTERNAL);
+
+    psa_destroy_key(previousKeyId);
+
+exit:
+    if (err != CHIP_NO_ERROR)
+    {
+        psa_destroy_key(previousKeyId);
+        psa_destroy_key(key.As<psa_key_id_t>());
```
Damian-Nordic, line 220: This seems unnecessary as in the case of error the key doesn't have a chance to be created.

src/crypto/PSASessionKeystore.cpp
```diff
@@ -192,5 +189,72 @@ void PSASessionKeystore::DestroyKey(HkdfKeyHandle & key)
...
 
+#if CHIP_CONFIG_ENABLE_ICD_CIP
+CHIP_ERROR PSASessionKeystore::PersistICDKey(Aes128KeyHandle & key)
+{
+    CHIP_ERROR err;
+    AesKeyAttributes attrs;
+    psa_key_id_t previousKeyId = key.As<psa_key_id_t>();
+    psa_key_attributes_t previousKeyAttrs;
+
+    psa_get_key_attributes(previousKeyId, &previousKeyAttrs);
```
Damian-Nordic, line 200: For 100% correctness, you should also call `psa_reset_key_attributes` when the attributes are not needed anymore. Consider extracting this to some `IsPersistentKey` helper.

src/crypto/PSASessionKeystore.cpp
```diff
@@ -192,5 +189,72 @@ void PSASessionKeystore::DestroyKey(HkdfKeyHandle & key)
...
+    // Exit early if key is already persistent
+    if (psa_get_key_lifetime(&previousKeyAttrs) == PSA_KEY_LIFETIME_PERSISTENT)
+    {
+        ExitNow(err = CHIP_NO_ERROR);
```
Damian-Nordic, line 204: nit: in Matter it's advised to use returns than goto wherever possible, so it's more consistent with the codebase to just `return CHIP_NO_ERROR` here.

Damian-Nordic (approved):
A couple of final nits and suggestions but looks good overall.

src/lib/core/CHIPConfig.h
maciejbaczmanski: makes sense :) renamed

src/crypto/PSASessionKeystore.cpp
maciejbaczmanski: As discussed offline, I've found that we can easily modify `PersistICDKey` to accept `Symmetric128BitsKeyHandle` for both key types. I've updated the code.
Added `psa_reset_key_attributes` before returns as well

src/crypto/PSASessionKeystore.cpp
maciejbaczmanski: Changed to `return CHIP_NO_ERROR`

src/crypto/PSASessionKeystore.cpp
maciejbaczmanski: changed the approach on PersistICDKey. Destroying of keys should be fine now as well

src/app/icd/server/ICDMonitoringTable.cpp
```diff
@@ -242,15 +241,21 @@ CHIP_ERROR ICDMonitoringTable::Get(uint16_t index, ICDMonitoringEntry & entry) c
...
-    while (index < this->Limit())
+    CHIP_ERROR err;
+    uint16_t index;
+    ICDMonitoringEntry tempEntry(mSymmetricKeystore);
+
+    for (index = 0; index < this->Limit(); index++)
     {
-        ReturnErrorOnFailure(this->Get(index++, entry));
-        if (id == entry.checkInNodeID)
+        SuccessOrExit(err = this->Get(index, tempEntry));
```
jmartinez-silabs, line 250: Consider not using SuccessOrExit/ GoTo label.
You could use continue; here

jmartinez-silabs (approved):

src/app/icd/server/ICDMonitoringTable.cpp
maciejbaczmanski: Sure, thanks.
I've decided to use `break` as we want to get the first ID that has no entry assigned

maciejbaczmanski: I've tested the change on desk and everything behaves correctly, so I'm resolving the comment

jmartinez-silabs: Yes you are right break, as you want to stop the loop. Sorry!

PR #34929, opened 2024-08-12
[Silabs]Provision storage sourceset

jmartinez-silabs:
Add gn arg to select what provision/factory data storage implementation to use.

Create a source_set that deals with matching the correct storage implementation source and matching provision lib for efr32 or si917 platform


examples/platform/silabs/provision/BUILD.gn
```diff
@@ -0,0 +1,52 @@
...
+import("//build_overrides/efr32_sdk.gni")
+import("${efr32_sdk_build_root}/silabs_board.gni")
+
+if (wifi_soc) {
+  import("${efr32_sdk_build_root}/SiWx917_sdk.gni")
+} else {
+  import("${efr32_sdk_build_root}/efr32_sdk.gni")
+}
+
+source_set("provision-storage") {
```
mkardous-silabs, line 25: This ends being provision:provision which seems a bit redundant. Might be worth while to remove the `provision-` bits.

third_party/silabs/efr32_sdk.gni
```diff
@@ -82,6 +82,9 @@ declare_args() {
   chip_enable_multi_ota_requestor = false
   chip_enable_multi_ota_encryption = false
   chip_enable_ota_custom_tlv_testing = false
+
+  # Factory Provision storage
+  use_flash_storage = wifi_soc
```
mkardous-silabs, line 87:
```suggestion
  sl_use_flash_storage = wifi_soc
```

third_party/silabs/efr32_sdk.gni
```diff
@@ -469,6 +472,12 @@ template("efr32_sdk") {
       defines += [ "SL_USE_COAP_CONFIG=0" ]
     }
 
+    if (use_flash_storage) {
+      defines += [ "SL_PROVISION_FLASH=1" ]
+    } else {
+      defines += [ "SL_PROVISION_FLASH=0" ]
```
mkardous-silabs, line 478: Can this be renamed to `SL_MATTER_PROVISION_FLASH`?

src/platform/silabs/efr32/efr32-chip-mbedtls-config.h
```diff
@@ -87,6 +87,12 @@
 #define PSA_WANT_ALG_SOME_PAKE
 #endif // SL_USE_COAP_CONFIG
 
+#if SL_PROVISION_FLASH
+#define MBEDTLS_SHA256_C
+#define MBEDTLS_HMAC_DRBG_C
+#define MBEDTLS_ECDSA_DETERMINISTIC
+#endif
```
mkardous-silabs, line 94:
```suggestion
#endif // SL_PROVISION_FLASH
```

third_party/silabs/efr32_sdk.gni
mkardous-silabs, line 87: we should add `asserts` to make sure the flash/non flash makes sense with the crypto flavor.

jmartinez-silabs: Technically they could/can be independent

mkardous-silabs (approved):

saurabhst (approved):

PR #34936, opened 2024-08-12
[HVAC] Sync atomic write error order with spec

### LINKED ISSUE #34743: Need to verify that thermostat preset atomic checks follow spec order (which needs to be more clearly defined)
### Feature description

The spec is not 100% clear on the order in which checks are performed during AtomicRequest command processing.  Once that's pinned down, we should make sure the SDK matches.

### Platform

all

### Platform Version(s)

_No response_

### Anything else?

_No response_

### LINKED ISSUE #35000: Handle subscription for Preset attributes like ActivePresetHandle, etc. that are not writeable and instead set by a command
We need to call AttributeChangedCallback when ActivePresetHandle is updated via the command SetActivePresetRequest to make sure the attribute is marked dirty and reporting picks the updated value.

hasty:
Fixes #34743
Fixes #35000

This also:
1. Removes the ThermostatManager, which had been included by accident
2. Relocates the thermostat delegate code from examples/thermostat/linux to examples/thermostat/thermostat-common
3. Splits ThermostatAttrAccess into separate files for atomic operations, preset management and setpoint validation, as thermostat-server.cpp was exceeding 2000 lines


examples/thermostat/thermostat-common/include/thermostat-delegate-impl.h
```diff
@@ -26,7 +27,7 @@ namespace Clusters {
 namespace Thermostat {
 
 /**
- * The ThermostatDelegate class serves as the instance delegate for storing Presets related information and providing it to the
+ * The ThermostatManager class serves as the instance delegate for storing Presets related information and providing it to the
```
drempelg, line 30: Schedules as well?

examples/thermostat/thermostat-common/include/atomic-write-manager.h
```diff
@@ -0,0 +1,186 @@
...
+     */
+    void SetDelegate(AtomicWriteDelegate * delegate) override { mDelegate = delegate; };
+
+    /**
+     * @brief Check if there is an open atomic write on an endpoint, optionally associated with a specific attribute
+     *
+     * @param attributeId Optional attribute filter
+     * @param endpoint The endpoint to check atomic write state
+     * @return true if there is an open atomic write
+     * @return false if there is not at open atomic write
```
nivi-apple, line 55:
```suggestion
     * @return false if there is no open atomic write
```

examples/thermostat/thermostat-common/include/atomic-write-manager.h
```diff
@@ -0,0 +1,186 @@
...
+     * @brief Check if there is an open atomic write on an endpoint, associated with a given SubjectDescriptor, and optionally
+     * associated with a specific attribute
+     *
+     * @param attributeId Optional attribute filter
+     * @param subjectDescriptor The subject descriptor to check against
+     * @param endpoint The endpoint to check atomic write state
+     * @return true if there is an open atomic write
+     * @return false if there is not an open atomic write
+     */
+    bool InWrite(const std::optional<AttributeId> attributeId, const Access::SubjectDescriptor & subjectDescriptor,
```
nivi-apple, line 69: could this be a scoped node id instead of the SubjectDescriptor?

examples/thermostat/thermostat-common/include/atomic-write-manager.h
```diff
@@ -0,0 +1,186 @@
...
+                 EndpointId endpoint) override;
+
+    /**
+     * @brief Check if there is an open atomic write on an endpoint, associated with the source node for a given command invocation,
+     * and optionally associated with a specific attribute
+     *
+     * @param attributeId
+     * @param commandObj
+     * @param endpoint
+     * @return true
```
nivi-apple, line 79:
```suggestion
     * @return true if there is an open atomic write
```

examples/thermostat/thermostat-common/include/atomic-write-manager.h
```diff
@@ -0,0 +1,186 @@
...
+     * @return false
```
nivi-apple, line 80:
```suggestion
     * @return false if there is no open atomic write
```

examples/thermostat/thermostat-common/include/atomic-write-manager.h
```diff
@@ -0,0 +1,186 @@
...
+
+    /**
+     * @brief Check if there is an open atomic write on an endpoint, associated with a given SubjectDescriptor, and optionally
+     * associated with a specific attribute
+     *
+     * @param attributeId Optional attribute filter
+     * @param subjectDescriptor The subject descriptor to check against
+     * @param endpoint The endpoint to check atomic write state
+     * @return true if there is an open atomic write
+     * @return false if there is not an open atomic write
```
nivi-apple, line 67:
```suggestion
     * @return false if there is no open atomic write
```

examples/thermostat/thermostat-common/include/atomic-write-manager.h
```diff
@@ -0,0 +1,186 @@
...
+
+    /**
+     * @brief Check if there is an open atomic write on an endpoint, associated with the source node for a given command invocation,
+     * and associated with the set of attribute IDs
+     *
+     * @param attributeIds The set of attribute IDs to check against
+     * @param commandObj The command being invoked
+     * @param endpoint The endpoint for the atomic write
+     * @return true if there is an open atomic write
+     * @return false if there is not an open atomic write
```
nivi-apple, line 92:
```suggestion
     * @return false if there is no open atomic write
```

examples/thermostat/thermostat-common/include/atomic-write-manager.h
```diff
@@ -0,0 +1,186 @@
...
+     * @brief Check if there is an open atomic write on an endpoint, associated with the source node for a given command invocation,
+     * and optionally associated with a specific attribute
+     *
+     * @param attributeId
+     * @param commandObj
+     * @param endpoint
+     * @return true
+     * @return false
+     */
+    bool InWrite(const std::optional<AttributeId> attributeId, CommandHandler * commandObj, EndpointId endpoint) override;
```
nivi-apple, line 82: can we have the scoped node id as an argument instead of the commandhandler? i assume we do not need the entire commandhandler object? also how is this different from the subject descriptor API? both boil down to a scoped node id IMO. is that not correct?

examples/thermostat/thermostat-common/include/atomic-write-manager.h
```diff
@@ -0,0 +1,186 @@
...
+     * @brief Check if there is an open atomic write on an endpoint, associated with the source node for a given command invocation,
+     * and associated with the set of attribute IDs
+     *
+     * @param attributeIds The set of attribute IDs to check against
+     * @param commandObj The command being invoked
+     * @param endpoint The endpoint for the atomic write
+     * @return true if there is an open atomic write
+     * @return false if there is not an open atomic write
+     */
+    bool InWrite(const DataModel::DecodableList<AttributeId>::Iterator attributeIds, CommandHandler * commandObj,
```
nivi-apple, line 94: could we have one api with an optional list of attributeids and then the case where we need one attribute id could be a list of length 1?

nivi-apple: same applies to other sets of APIs

examples/thermostat/thermostat-common/include/atomic-write-manager.h
```diff
@@ -0,0 +1,186 @@
...
+
+namespace chip {
+namespace app {
+namespace Clusters {
+namespace Thermostat {
+
+enum AtomicWriteState
+{
+    kAtomicWriteState_Closed = 0,
+    kAtomicWriteState_Open,
```
nivi-apple, line 34: do we need to duplicate the AtomicWriteState here? how about kClosed kOpen. so we can use AtomicWriteState::kOpen or AtomicWriteState::kClosed.

examples/thermostat/thermostat-common/include/atomic-write-manager.h
hasty: No, because it needs to check the authMode to be sure that the scoped node ID is valid for atomic writes.

examples/thermostat/thermostat-common/include/atomic-write-manager.h
```diff
@@ -0,0 +1,186 @@
...
+    bool InWrite(const DataModel::DecodableList<AttributeId>::Iterator attributeIds, CommandHandler * commandObj,
+                 EndpointId endpoint) override;
+
+    /**
+     * @brief Attempt to start an atomic write
+     *
+     * @param commandObj The AtomicRequest command
+     * @param commandPath The path for the command
+     * @param commandData The payload of the command
+     * @return true if the atomic write was opened
```
nivi-apple, line 103:
```suggestion
     * @return true if the atomic write was opened successfully
```

examples/thermostat/thermostat-common/include/atomic-write-manager.h
```diff
@@ -0,0 +1,186 @@
...
+     * @return false if the atomic write was not opened
```
nivi-apple, line 104:
```suggestion
     * @return false if the request to open atomic write failed
```

examples/thermostat/thermostat-common/include/atomic-write-manager.h
```diff
@@ -0,0 +1,186 @@
...
+     */
+    bool BeginWrite(chip::app::CommandHandler * commandObj, const ConcreteCommandPath & commandPath,
+                    const Commands::AtomicRequest::DecodableType & commandData) override;
+
+    /**
+     * @brief Attempt to commit an atomic write; returns true if the server is able to commit or failed trying, false if it was
```
nivi-apple, line 110:
```suggestion
     * @brief Attempt to commit an atomic write; returns false if it was
```

examples/thermostat/thermostat-common/include/atomic-write-manager.h
```diff
@@ -0,0 +1,186 @@
...
+     * unable to find a matching atomic write
```
nivi-apple, line 111:
```suggestion
     * unable to find a matching atomic write, otherwise returns true
```

examples/thermostat/thermostat-common/include/atomic-write-manager.h
hasty: They both end up comparing a scoped node id, but they fetch that node ID in different ways. If this just took a scoped node ID, then either I'd have to expose `ScopedNodeId GetSourceScopedNodeId(CommandHandler * commandObj)`, in which case I still have an API that takes a CommandHandler, or I need to trust that anyone calling this API will get the scoped node ID in the same fashion.

examples/thermostat/thermostat-common/include/atomic-write-manager.h
```diff
@@ -0,0 +1,186 @@
...
+     *
+     * @param commandObj The AtomicRequest command
+     * @param commandPath The path for the command
+     * @param commandData The payload of the command
+     * @return true if the atomic write was committed or rolled back
```
nivi-apple, line 116:
```suggestion
     * @return true if the atomic write was either committed or rolled back
```

examples/thermostat/thermostat-common/include/atomic-write-manager.h
```diff
@@ -0,0 +1,186 @@
...
+     * @return false if the atomic write was not found
```
nivi-apple, line 117: maybe should say returns false otherwise. i am sure that false would be the default case (haven't checked the impl yet) :)

examples/thermostat/thermostat-common/include/atomic-write-manager.h
```diff
@@ -0,0 +1,186 @@
...
+     */
+    bool CommitWrite(chip::app::CommandHandler * commandObj, const ConcreteCommandPath & commandPath,
+                     const Commands::AtomicRequest::DecodableType & commandData) override;
+
+    /**
+     * @brief Attempt to roll back an atomic write; returns true if the server is able to commit or failed trying, false if it was
```
nivi-apple, line 123:
```suggestion
     * @brief Attempt to roll back an atomic write; returns true if the server is able to either commit or rollback false if it was
```

examples/thermostat/thermostat-common/include/atomic-write-manager.h
```diff
@@ -0,0 +1,186 @@
...
+     * unable to find a matching atomic write
```
nivi-apple, line 124: same comment as above. maybe better to say returns false otherwise (unless there's no other scenarios that could occur). applies to other instances in this file.

examples/thermostat/thermostat-common/include/atomic-write-manager.h
```diff
@@ -0,0 +1,186 @@
...
+     */
+    bool RollbackWrite(chip::app::CommandHandler * commandObj, const ConcreteCommandPath & commandPath,
+                       const Commands::AtomicRequest::DecodableType & commandData) override;
+
+    /**
+     * @brief Reset any atomic writes associated with the given endpoint
+     *
+     * @param endpoint
+     */
+    void ResetWrite(EndpointId endpoint) override;
```
nivi-apple, line 140:
```suggestion
    void ResetWriteForEndpoint(EndpointId endpoint) override;
```

examples/thermostat/thermostat-common/include/atomic-write-manager.h
```diff
@@ -0,0 +1,186 @@
...
+
+    /**
+     * @brief Reset any atomic writes originating from a given fabric index
+     *
+     * @param fabricIndex
+     */
+    void ResetWrite(FabricIndex fabricIndex) override;
```
nivi-apple, line 147:
```suggestion
    void ResetWriteFor Fabric(FabricIndex fabricIndex) override;
```

examples/thermostat/thermostat-common/include/thermostat-delegate-impl.h
nivi-apple: schedules are not implemented yet so maybe thats fine.

nivi-apple, line 30:
```suggestion
 * The ThermostatDelegate class serves as the instance delegate for storing Presets related information and providing it to the
```

nivi-apple: the class is still named ThermostatDelegate. unless i am missing something :)

examples/thermostat/thermostat-common/src/atomic-write-manager.cpp
```diff
@@ -0,0 +1,535 @@
...
+
+#include <platform/internal/CHIPDeviceLayerInternal.h>
+
+using namespace chip;
+using namespace chip::app;
+using namespace chip::app::Clusters::Globals::Structs;
+using namespace chip::app::Clusters::Thermostat;
+using namespace chip::app::Clusters::Thermostat::Attributes;
+
+using imcode = chip::Protocols::InteractionModel::Status;
```
nivi-apple, line 29: how about using namespace chip::Protocols::InteractionModel::Status; and the using Status everywhere. Boris had a comment on my PR that imcode was wierd :)

examples/thermostat/thermostat-common/src/atomic-write-manager.cpp
```diff
@@ -0,0 +1,535 @@
...
+        default:
+            // TODO: If this is a valid attribute on thermostat, but just isn't atomic, we shouldn't return an error code here
+            return imcode::InvalidCommand;
+        }
+    }
+    if (attributeIdsIter.GetStatus() != CHIP_NO_ERROR)
+    {
+        return imcode::InvalidCommand;
+    }
+    if (requestedPresets && requestedSchedules)
```
nivi-apple, line 99: should this be ||

examples/thermostat/thermostat-common/src/atomic-write-manager.cpp
```diff
@@ -0,0 +1,535 @@
...
+        sourceNodeId = sessionHandle->AsIncomingGroupSession()->GetPeer();
+    }
+    return sourceNodeId;
+}
+
+/**
+ * @brief Schedules a timer for the given timeout in milliseconds.
+ *
+ * @param[in] endpoint The endpoint to use.
+ * @param[in] timeoutMilliseconds The timeout in milliseconds.
```
nivi-apple, line 126:
```suggestion
 * @param[in] timeout The timeout in milliseconds.
```

examples/thermostat/thermostat-common/include/atomic-write-manager.h
hasty: Any particular reason for this? The argument makes it clear it's for an endpoint.

examples/thermostat/thermostat-common/src/atomic-write-manager.cpp
```diff
@@ -0,0 +1,535 @@
...
+{
+    for (size_t i = 0; i < ArraySize(mAtomicWriteSessions); ++i)
+    {
+
+        auto atomicWriteState = mAtomicWriteSessions[i];
+        if (atomicWriteState.state == kAtomicWriteState_Open && atomicWriteState.endpointId == endpoint)
+        {
+            if (mDelegate != nullptr)
+            {
+                mDelegate->OnRollbackWrite(endpoint, Presets::Id);
```
nivi-apple, line 153: Can we have the delegate api just take endpoint as an argument and the delegate can rollback both if applicable. Calling this twice just looks weird mDelegate->OnRollbackWrite(endpoint)

examples/thermostat/thermostat-common/src/atomic-write-manager.cpp
```diff
@@ -0,0 +1,535 @@
...
+        mAtomicWriteSessions[ep].state      = state;
+        mAtomicWriteSessions[ep].endpointId = endpoint;
+        mAtomicWriteSessions[ep].nodeId     = originatorNodeId;
+    }
+}
+
+bool ThermostatAtomicWriteManager::InWrite(const std::optional<AttributeId> attributeId, EndpointId endpoint)
+{
+    // These are the only two atomic attributes
+    if (attributeId.has_value() && attributeId.value() != Presets::Id && attributeId.value() != Schedules::Id)
```
nivi-apple, line 177:
```suggestion
    if (attributeId.has_value() && (attributeId.value() != Presets::Id || attributeId.value() != Schedules::Id))
```

examples/thermostat/thermostat-common/src/atomic-write-manager.cpp
```diff
@@ -0,0 +1,535 @@
...
+    {
+        return false;
+    }
+
+    bool inAtomicWrite = false;
```
nivi-apple, line 182: don't need this temp bool.

examples/thermostat/thermostat-common/src/atomic-write-manager.cpp
```diff
@@ -0,0 +1,535 @@
...
+    uint16_t ep =
+        emberAfGetClusterServerEndpointIndex(endpoint, Thermostat::Id, MATTER_DM_THERMOSTAT_CLUSTER_SERVER_ENDPOINT_COUNT);
+
+    if (ep < ArraySize(mAtomicWriteSessions))
+    {
+        inAtomicWrite = mAtomicWriteSessions[ep].state == kAtomicWriteState_Open;
```
nivi-apple: return (mAtomicWriteSessions[ep].state == kAtomicWriteState_Open);

examples/thermostat/thermostat-common/src/atomic-write-manager.cpp
```diff
@@ -0,0 +1,535 @@
...
+    }
+    return inAtomicWrite;
```
nivi-apple, line 190: return false;

examples/thermostat/thermostat-common/src/atomic-write-manager.cpp
```diff
@@ -0,0 +1,535 @@
...
+}
+
+bool ThermostatAtomicWriteManager::InWrite(const std::optional<AttributeId> attributeId,
```
nivi-apple, line 193: like i commented before looks like both apis with SubjectDescriptor and commandobj check the scoped node id so maybe remove both and add an api with scopednodeid as the argument

examples/thermostat/thermostat-common/src/atomic-write-manager.cpp
```diff
@@ -0,0 +1,535 @@
...
+{
+    if (!InWrite(attributeId, endpoint))
+    {
+        return false;
+    }
+    return subjectDescriptor.authMode == Access::AuthMode::kCase &&
+        GetAtomicWriteScopedNodeId(attributeId, endpoint) == ScopedNodeId(subjectDescriptor.subject, subjectDescriptor.fabricIndex);
+}
+
+bool ThermostatAtomicWriteManager::InWrite(const std::optional<AttributeId> attributeId, CommandHandler * commandObj,
```
nivi-apple, line 204: like i commented before looks like both apis with SubjectDescriptor and commandobj check the scoped node id so maybe remove both and add an api with scopednodeid as the argument

examples/thermostat/thermostat-common/src/atomic-write-manager.cpp
```diff
@@ -0,0 +1,535 @@
...
+        commandObj->AddStatus(commandPath, status);
+        return false;
+    }
+    if (attributeRequestCount == 0)
+    {
+        commandObj->AddStatus(commandPath, imcode::InvalidCommand);
+        return false;
+    }
+
+    auto timeoutRequest = System::Clock::Milliseconds16(commandData.timeout.Value());
```
nivi-apple, line 240: this seems to be used only once. maybe just use System::Clock::Milliseconds16(commandData.timeout.Value()) at line [here](https://github.com/project-chip/connectedhomeip/pull/34936/files#diff-1e54d0c08c51e7226f19f1461cb899ae549d766577aed32b612873b7f6104cf1R375) or move this line down to narrow the scope.

examples/thermostat/thermostat-common/src/atomic-write-manager.cpp
```diff
@@ -0,0 +1,535 @@
...
+            {
+                status = imcode::Failure;
+            }
+            attributeStatus.statusCode = to_underlying(statusCode);
+        }
+    }
+    if (status == imcode::Failure)
+    {
+        // Either one of the calls to OnPreCommitWrite failed, or one of the calls to OnCommitWrite failed; in the former case,
+        // discard any pending writes Do the same for the latter, knowing that the server may be in an inconsistent state
```
nivi-apple, line 416:
```suggestion
        // discard any pending writes. Do the same for the latter, knowing that the server may be in an inconsistent state
```

examples/thermostat/thermostat-common/src/atomic-write-manager.cpp
```diff
@@ -0,0 +1,535 @@
...
+    {
+        auto atomicWriteState = mAtomicWriteSessions[i];
+        if (atomicWriteState.state == kAtomicWriteState_Open && atomicWriteState.nodeId.GetFabricIndex() == fabricIndex)
+        {
+            ResetWrite(atomicWriteState.endpointId);
+        }
+    }
+}
+
+ScopedNodeId ThermostatAtomicWriteManager::GetAtomicWriteScopedNodeId(const std::optional<AttributeId> attributeId,
```
nivi-apple, line 499:
```suggestion
ScopedNodeId ThermostatAtomicWriteManager::GetAtomicWriteOriginatorScopedNodeId(const std::optional<AttributeId> attributeId,
```

examples/thermostat/thermostat-common/src/thermostat-delegate-impl.cpp
```diff
@@ -244,3 +208,23 @@ void ThermostatDelegate::ClearPendingPresetList()
 {
     mNextFreeIndexInPendingPresetsList = 0;
 }
+
+std::optional<System::Clock::Milliseconds16> ThermostatDelegate::GetWriteTimeout(AttributeId attributeId)
+{
+    switch (attributeId)
+    {
+    case Presets::Id:
+        // If the client expects to edit the presets, then we'll give it 3 seconds to do so
```
nivi-apple, line 217: isn't 3 secs too less. it might work for automation but using chip-tool might need longer (assuming this is the timeout for actually editing the presets)

nivi-apple: same for schedules

src/app/clusters/thermostat-server/atomic-write.h
```diff
@@ -0,0 +1,217 @@
...
+
+    virtual ~AtomicWriteManager() = default;
+
+    /**
+     * @brief Check if there is an open atomic write on an endpoint, optionally associated with a specific attribute
+     *
+     * @param attributeId Optional attribute filter
+     * @param endpoint The endpoint to check atomic write state
+     * @return true if there is an open atomic write
+     * @return false if there is not at open atomic write
```
nivi-apple, line 110:
```suggestion
     * @return false if there is no open atomic write
```

src/app/clusters/thermostat-server/atomic-write.h
```diff
@@ -0,0 +1,217 @@
...
+
+    /**
+     * @brief Check if there is an open atomic write on an endpoint, associated with a given SubjectDescriptor, and optionally
+     * associated with a specific attribute
+     *
+     * @param attributeId Optional attribute filter
+     * @param subjectDescriptor The subject descriptor to check against
+     * @param endpoint The endpoint to check atomic write state
+     * @return true if there is an open atomic write
+     * @return false if there is not an open atomic write
```
nivi-apple, line 122:
```suggestion
     * @return false if there is no open atomic write
```

src/app/clusters/thermostat-server/atomic-write.h
```diff
@@ -0,0 +1,217 @@
...
+     * @param commandPath The path for the command
+     * @param commandData The payload of the command
+     * @return true if the atomic write was opened
+     * @return false if the atomic write was not opened
+     */
+    virtual bool BeginWrite(chip::app::CommandHandler * commandObj, const ConcreteCommandPath & commandPath,
+                            const Commands::AtomicRequest::DecodableType & commandData) = 0;
+
+    /**
+     * @brief Attempt to commit an atomic write; returns true if the server is able to commit or failed trying, false if it was
```
nivi-apple, line 152: i have commented on these comments earlier in another file. I guess we can only have the comment in one delegate header and remove the duplicate comments elsewhere. Please see recommended changes to the comments in the other file

src/app/clusters/thermostat-server/thermostat-server-atomic.cpp
```diff
@@ -0,0 +1,96 @@
...
+namespace Thermostat {
+
+imcode ThermostatAttrAccess::OnBeginWrite(EndpointId endpoint, AttributeId attributeId)
+{
+    switch (attributeId)
+    {
+    case Presets::Id:
+        return BeginPresets(endpoint);
+    case Schedules::Id:
+        return imcode::Success;
```
nivi-apple: can we return notsupported or something similar?

nivi-apple: applies throughout. because if we return success a client would get confused why it doesn't work

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -980,6 +373,12 @@ void emberAfThermostatClusterServerInitCallback(chip::EndpointId endpoint)
     // or should this just be the responsibility of the thermostat application?
 }
 
+// IMPORTANT NOTE:
+// No Side effects are permitted in emberAfThermostatClusterServerPreAttributeChangedCallback
+// If a setpoint changes is required as a result of setpoint limit change
+// it does not happen here.  It is the responsibility of the device to adjust the setpoint(s)
+// as required in emberAfThermostatClusterServerPostAttributeChangedCallback
+// limit change validation assures that there is at least 1 setpoint that will be valid
```
nivi-apple, line 380: this line doesn't make sense combined with the previous line

examples/thermostat/thermostat-common/src/thermostat-delegate-impl.cpp
hasty: This was already bumped up per Boris' request. :)  I'll let David tell me how actual physical testing goes in terms of a valid maximum.

src/app/clusters/thermostat-server/thermostat-server.cpp
hasty: I didn't actually write this comment; it just got relocated away from this function in the initial commit of the preset code somehow

examples/thermostat/thermostat-common/src/atomic-write-manager.cpp
hasty: Leftover from when this was a loop. Will change.

examples/thermostat/thermostat-common/src/atomic-write-manager.cpp
hasty: Leftover from when this was a loop. Will change.

examples/thermostat/thermostat-common/include/atomic-write-manager.h
hasty: Rewrote to have the bool just mean whether or not the command was handled, which is more in line with how ember callbacks work.

examples/thermostat/thermostat-common/src/atomic-write-manager.cpp
hasty: As I replied before, the subject descriptor is acquired and validated differently for commands and attribute accesses; abstracting that would just move this logic elsewhere and require exporting a method with CommandHandler anyway. I'm going to stick with having this logic in one place instead of three.

examples/thermostat/thermostat-common/src/atomic-write-manager.cpp
hasty: See below

examples/thermostat/thermostat-common/src/atomic-write-manager.cpp
hasty: Will move down.

examples/thermostat/thermostat-common/src/atomic-write-manager.cpp
```diff
@@ -0,0 +1,535 @@
...
+
+        attributeStatuses.push_back({ .attributeID = attributeId, .statusCode = to_underlying(imcode::Success) });
+    }
+
+    for (auto & attributeStatus : attributeStatuses)
+    {
+        imcode statusCode = mDelegate->OnRollbackWrite(endpoint, attributeStatus.attributeID);
+        if (statusCode != imcode::Success)
+        {
+            status = imcode::Failure;
```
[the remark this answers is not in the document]
hasty: I'm not, this is the overall status code for the command.

src/app/clusters/thermostat-server/atomic-write.h
hasty: Removed the duplicate comments

examples/thermostat/thermostat-common/include/atomic-write-manager.h
nivi-apple: Yeah I agree that you get scoped node id from the two objects in different ways that doesn't justify passing the whole commandObj object or the subjectDescriptor objects if all you need is the scoped node id. Also that reduces > 1 apis to 1 consolidated API. I think it would be a good idea to do so

examples/thermostat/thermostat-common/include/atomic-write-manager.h
nivi-apple: yeah you got a point

examples/thermostat/thermostat-common/src/atomic-write-manager.cpp
nivi-apple: why would you need to export a method for CommandHandler. i do not understand this part. you have 2 utility functions that get the scoped node id from either the commandhandler object or subjectdescriptor object in the callee for these methods. and your api inwrite would look like - bool ThermostatAtomicWriteManager::InWrite(const std::optional<AttributeId> attributeId, ScopedNodeId nodeId, EndpointId endpoint) instead of 2 versions with commandobject and subjectdescriptor. Am i missing something here?

examples/thermostat/thermostat-common/src/thermostat-delegate-impl.cpp
nivi-apple: still think it's not enough based on my testing with commands but will wait for what David finds during testing

examples/thermostat/thermostat-common/src/atomic-write-manager.cpp
```diff
@@ -0,0 +1,576 @@
...
+        emberAfGetClusterServerEndpointIndex(endpoint, Thermostat::Id, MATTER_DM_THERMOSTAT_CLUSTER_SERVER_ENDPOINT_COUNT);
+
+    if (ep < ArraySize(mAtomicWriteSessions))
+    {
+        return mAtomicWriteSessions[ep].state == AtomicWriteState::Open;
+    }
+    return false;
+}
+
+bool ThermostatAtomicWriteManager::InWrite(const std::optional<AttributeId> attributeId,
```
bzbarsky-apple, line 261: Should this logic be in the _application_ as here, or in the _cluster implementation_?

Seems to me that all the logic involving subject descriptors and whatnot is not application-specific and should be handled by the cluster implementation, because otherwise we are forcing all applications to re-implement that logic.

examples/thermostat/thermostat-common/src/atomic-write-manager.cpp
hasty: Because this would mean replacing two methods with three methods:
```
bool InWrite(const std::optional<AttributeId> attributeId, const ScopedNodeId & nodeId,
                         EndpointId endpoint);
ScopedNodeId GetScopedNodeIdFromSubjectDescriptor(const Access::SubjectDescriptor & subjectDescriptor);                 
ScopedNodeId GetScopedNodeIdFromCommandHandler(const CommandHandler * commandObj);   
```
and now requiring callers to call two methods instead of one:
```
awm->InWrite(aPath.mAttributeId, awm->GetScopedNodeIdFromSubjectDescriptor(subjectDescriptor), aPath.mEndpointId)
awm->InWrite(aPath.mAttributeId, awm->GetScopedNodeIdFromCommandHandler(commandObj), aPath.mEndpointId)
```
as well as opening up the ability to work around the security check by crafting their own ScopedNodeId.

bzbarsky-apple: I think the key thing is this: finding the right ScopedNodeId is not trivial, and we should not be pushing that work off on applications if we can help it.  But I'm waiting to see what things look like with more of the logic just moved into the cluster.

src/app/clusters/thermostat-server/thermostat-server-atomic.cpp
```diff
@@ -0,0 +1,548 @@
...
+
+    if (delegate == nullptr)
+    {
+        ChipLogError(Zcl, "Delegate is null");
+        commandObj->AddStatus(commandPath, Status::InvalidInState);
+        return;
+    }
+
+    size_t attributeStatusCount = 0;
+    Platform::ScopedMemoryBuffer<AtomicAttributeStatusStruct::Type> attributeStatuses;
```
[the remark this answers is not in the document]
hasty: OK, switching.

src/app/clusters/thermostat-server/thermostat-server.h
```diff
@@ -34,25 +35,34 @@ namespace app {
...
 public:
     ThermostatAttrAccess() : AttributeAccessInterface(Optional<chip::EndpointId>::Missing(), Thermostat::Id) {}
 
     CHIP_ERROR Read(const ConcreteReadAttributePath & aPath, AttributeValueEncoder & aEncoder) override;
     CHIP_ERROR Write(const ConcreteDataAttributePath & aPath, chip::app::AttributeValueDecoder & aDecoder) override;
 
+private:
+    Protocols::InteractionModel::Status SetActivePreset(EndpointId endpoint, DataModel::Nullable<ByteSpan> presetHandle);
+    CHIP_ERROR AppendPendingPreset(Thermostat::Delegate * delegate, const Structs::PresetStruct::Type & preset);
+    Protocols::InteractionModel::Status PrecommitPresets(EndpointId endpoint);
```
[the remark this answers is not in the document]
hasty: OK, added.

src/app/clusters/thermostat-server/thermostat-server-atomic.cpp
```diff
@@ -0,0 +1,548 @@
...
+        return;
+    }
+
+    if (!gThermostatAttrAccess.InAtomicWrite(commandObj, endpoint))
+    {
+        // There's no open atomic write
+        commandObj->AddStatus(commandPath, Status::InvalidInState);
+        return;
+    }
+
```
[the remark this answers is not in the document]
hasty: Changed to track attribute IDs in AtomicWriteSession.

src/app/clusters/thermostat-server/thermostat-server-atomic.cpp
```diff
@@ -0,0 +1,548 @@
...
+    {
+        auto & attributeStatus = attributeStatuses[i];
+        auto statusCode        = Status::Success;
+        switch (attributeStatus.attributeID)
+        {
+        case Presets::Id:
+            statusCode = PrecommitPresets(endpoint);
+            if (statusCode == Status::Success)
+            {
+                CHIP_ERROR err = delegate->CommitPendingPresets();
```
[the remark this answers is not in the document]
hasty: Fixed.

src/app/clusters/thermostat-server/thermostat-server-atomic.cpp
```diff
@@ -0,0 +1,548 @@
...
+            }
+
+            break;
+        }
+    }
+
+    status = Status::Success;
+    for (size_t i = 0; i < attributeStatusCount; ++i)
+    {
+        auto & attributeStatus = attributeStatuses[i];
```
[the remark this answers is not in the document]
hasty: There's a comment to that effect in emberAfThermostatClusterAtomicRequestCallback

src/app/clusters/thermostat-server/thermostat-server-atomic.cpp
```diff
@@ -0,0 +1,548 @@
...
+        switch (attributeStatus.attributeID)
+        {
+        case Presets::Id:
+        case Schedules::Id:
+            attributeStatus.statusCode = to_underlying(Status::Success);
+            break;
+        default:
+            commandObj->AddStatus(commandPath, Status::InvalidInState);
```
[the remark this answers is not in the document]
hasty: Should no longer be necessary as InAtomicWrite checks for non-atomic attributes; removed.

src/app/clusters/thermostat-server/thermostat-server-presets.cpp
```diff
@@ -0,0 +1,568 @@
...
+using namespace chip::app;
+using namespace chip::app::Clusters;
+using namespace chip::app::Clusters::Thermostat;
+using namespace chip::app::Clusters::Thermostat::Attributes;
+using namespace chip::app::Clusters::Thermostat::Structs;
+using namespace chip::app::Clusters::Globals::Structs;
+using namespace chip::Protocols::InteractionModel;
+
+extern int16_t EnforceHeatingSetpointLimits(int16_t HeatingSetpoint, EndpointId endpoint);
+extern int16_t EnforceCoolingSetpointLimits(int16_t CoolingSetpoint, EndpointId endpoint);
```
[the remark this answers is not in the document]
hasty: Moved into Thermostat namespace; they were anonymous before, but now need to be called from multiple files.

src/app/clusters/thermostat-server/thermostat-server-presets.cpp
```diff
@@ -0,0 +1,568 @@
...
+
+/**
+ * @brief Check if a preset is valid.
+ *
+ * @param[in] preset The preset to check.
+ *
+ * @return true If the preset is valid i.e the PresetHandle (if not null) fits within size constraints and the presetScenario enum
+ *         value is valid. Otherwise, return false.
+ */
+bool IsValidPresetEntry(const PresetStruct::Type & preset)
```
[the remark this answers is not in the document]
hasty: Moved to anonymous namespace.

src/app/clusters/thermostat-server/thermostat-server-presets.cpp
```diff
@@ -0,0 +1,568 @@
...
+        // If a built in preset in the Presets attribute list is removed and not found in the pending presets list, return
+        // CONSTRAINT_ERROR.
+        if (IsBuiltIn(preset) && !found)
+        {
+            return Status::ConstraintError;
+        }
+    }
+
+    // For each preset in the presets attribute, check that the matching preset in the pending presets list does not
+    // violate any spec constraints.
```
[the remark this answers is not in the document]
hasty: Removed.

src/python_testing/TC_TSTAT_4_2.py
```diff
@@ -61,15 +61,27 @@ class TC_TSTAT_4_2(MatterBaseTest):
...
                 found_preset_status = True
+            if attrStatus.attributeID == cluster.Attributes.Schedules.attribute_id:
+                asserts.assert_equal(attrStatus.statusCode, expected_schedules_status,
+                                     "Schedules attribute should have the right status")
+                found_schedules_status = True
+        if expected_timeout is not None:
+            asserts.assert_equal(response.timeout, expected_timeout,
+                                 "Timeout should have the right value")
+        asserts.assert_true(found_preset_status, "Preset attribute should have a status")
+        asserts.assert_true(found_schedules_status, "Schedules attribute should have a status")
```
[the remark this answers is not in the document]
hasty: This was part of why I was refactoring to have a generic atomic write handler, with a virtual method for allowing a derived class in the app override which attributes have to be locked together. The code here is from our internal implementation, which requires both to be locked at the same time.

I already fixed this based on previous comments on this PR, so I'm going to leave it fixed rather than try to back it out.

src/app/clusters/thermostat-server/thermostat-server-atomic.cpp
```diff
@@ -0,0 +1,548 @@
...
+        }
+    }
+    if (requireAll)
+    {
+        if (!requestedPresets || !requestedSchedules)
+        {
+            return Status::InvalidInState;
+        }
+    }
+    else if (requestedPresets ^ requestedSchedules)
```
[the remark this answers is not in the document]
hasty: Removed entirely.

src/app/clusters/thermostat-server/thermostat-server-atomic.cpp
```diff
@@ -0,0 +1,548 @@
...
+    {
+        // Either we errored reading the list, or one of the attributes is not supported on this server
+        return Status::InvalidCommand;
+    }
+    if (attributeStatusCount == 0)
+    {
+        // List can't be empty
+        return Status::InvalidCommand;
+    }
+    if (requestedPresets ^ requestedSchedules)
```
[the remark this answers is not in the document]
hasty: Removed entirely.

src/app/clusters/thermostat-server/thermostat-server-atomic.cpp
```diff
@@ -0,0 +1,548 @@
...
+    }
+    for (size_t i = 0; i < index; ++i)
+    {
+        auto & attributeStatus = attributeStatuses[i];
+        const EmberAfAttributeMetadata * metadata =
+            emberAfLocateAttributeMetadata(endpoint, Thermostat::Id, attributeStatus.attributeID);
+
+        if (metadata == nullptr)
+        {
+            // This is not a valid attribute on the Thermostat cluster on the supplied endpoint
```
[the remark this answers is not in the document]
hasty: What's the proper way to ask ember if an attribute is supported? 

I think the spec is fine; it says: if an attribute isn't supported at all, then return invalid command for the whole thing; if it's supported but doesn't support atomic writes, then return an attribute status to that effect. 

So the only problem here is that there are supported attributes without metadata, so I need some other way of checking.

src/app/clusters/thermostat-server/thermostat-server-atomic.cpp
```diff
@@ -0,0 +1,548 @@
...
+
+extern ThermostatAttrAccess gThermostatAttrAccess;
+
+/**
+ * @brief Callback that is called when the timeout for editing the presets expires.
+ *
+ * @param[in] systemLayer The system layer.
+ * @param[in] callbackContext The context passed to the timer callback.
+ */
+void TimerExpiredCallback(System::Layer * systemLayer, void * callbackContext)
```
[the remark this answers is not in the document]
bzbarsky-apple: OK, so TimerExpiredCallback needs to be something that ThermostatAttrAccess can reference, but the other functions here can be in an anonymous namespace, no?  The ones that are referenced only in this file?

src/app/clusters/thermostat-server/thermostat-server-atomic.cpp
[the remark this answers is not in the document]
bzbarsky-apple: > What's the proper way to ask ember if an attribute is supported?

Your way is the right way _except_ for global attributes.  Those are supported on all clusters, by definition, so we don't have a bunch of in-memory metadata for them.

You're right that the spec defines a reasonable behavior here, so to implement that what we should do is:

1.  Check for metadata as you do now.
2. If no metadata, check whether the attribute ID is in `GlobalAttributesNotInMetadata` (see `src/app/GlobalAttributes.h`).  If it's in there, it's a supported attribute.  Otherwise it's not.

src/app/clusters/thermostat-server/thermostat-server-atomic.cpp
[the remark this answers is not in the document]
bzbarsky-apple: Right, but this code is what's implementing the relevant spec checks; it took me a second, when comparing it to the spec, to figure out why the check was not here.  Followup: move the comment.

src/app/clusters/thermostat-server/thermostat-server-atomic.cpp
```diff
@@ -0,0 +1,644 @@
...
+                err = delegate->CommitPendingPresets();
+                if (err != CHIP_NO_ERROR)
+                {
+                    statusCode = Status::InvalidInState;
+                }
+                break;
+            case Schedules::Id:
+                break;
+            default:
+                commandObj->AddStatus(commandPath, Status::InvalidInState);
```
bzbarsky-apple, line 543: Not actually reachable, right?

bzbarsky-apple (approved):
Approved, with the following to be addressed in followups (please NOT in this PR):

* https://github.com/project-chip/connectedhomeip/pull/34936#discussion_r1728119257
* https://github.com/project-chip/connectedhomeip/pull/34936#discussion_r1728125831
* https://github.com/project-chip/connectedhomeip/pull/34936#discussion_r1728131520
* https://github.com/project-chip/connectedhomeip/pull/34936#discussion_r1727367220

src/app/clusters/thermostat-server/thermostat-server-atomic.cpp
```diff
@@ -0,0 +1,644 @@
...
+                return;
```
bzbarsky-apple, line 544:
```suggestion
                // Not reachable, since we returned in this situation above.
                break;
```

bzbarsky-apple (approved):

jmartinez-silabs (approved):

src/app/clusters/thermostat-server/thermostat-server-atomic.cpp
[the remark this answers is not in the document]
bzbarsky-apple: Also: there is in fact no such comment in emberAfThermostatClusterAtomicRequestCallback

PR #34938, opened 2024-08-13
[Fabric-Admin] Fix 'fabricsync sync-device' command always fails

yufengwangca:
'fabricsync sync-device' command always fails on master after the OCW API has been refactored. We need to update the fabric-admin example app to adapt to the new interface.

### DIFF
--- a/examples/fabric-admin/device_manager/DeviceManager.cpp
+++ b/examples/fabric-admin/device_manager/DeviceManager.cpp
@@ -33,7 +33,6 @@ namespace {
 // Constants
 constexpr uint32_t kSetupPinCode               = 20202021;
 constexpr uint16_t kRemoteBridgePort           = 5540;
-constexpr uint16_t kDiscriminator              = 3840;
 constexpr uint16_t kWindowTimeout              = 300;
 constexpr uint16_t kIteration                  = 1000;
 constexpr uint16_t kSubscribeMinInterval       = 0;
@@ -132,11 +131,16 @@ void DeviceManager::OpenRemoteDeviceCommissioningWindow(EndpointId remoteEndpoin
     // Open the commissioning window of a device from another fabric via its fabric bridge.
     // This method constructs and sends a command to open the commissioning window for a device
     // that is part of a different fabric, accessed through a fabric bridge.
-    StringBuilder<kMaxCommandSize> commandBuilder;
+    StringBuilder<512> commandBuilder;
+
+    // Use random discriminator to have less chance of collission.
+    uint16_t discriminator = Crypto::GetRandU16() % 4097; // 4097 to include the upper limit 4096
 
     commandBuilder.Add("pairing open-commissioning-window ");
     commandBuilder.AddFormat("%lu %d %d %d %d %d", mRemoteBridgeNodeId, remoteEndpointId, kEnhancedCommissioningMethod,
-                             kWindowTimeout, kIteration, kDiscriminator);
+                             kWindowTimeout, kIteration, discriminator);
+    commandBuilder.Add(" --setup-pin 20202021");
+    commandBuilder.Add(" --salt hex:4c38d8d7708e777b5df816886167839828578ee98280ddac6e801124e8e695e7");
 
     PushCommand(commandBuilder.c_str());
 }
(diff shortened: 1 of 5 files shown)


saurabhst (approved):

samadDotDev (commented):
We don't need to supply the setup-pin / passcode as it's optional and will be [generated](https://github.com/project-chip/connectedhomeip/blob/20d6857cf5f3c67ffe24cba02f67ff1b372bab29/src/protocols/secure_channel/PASESession.cpp#L145) by commissioning window opener API when [absent](https://github.com/project-chip/connectedhomeip/blob/20d6857cf5f3c67ffe24cba02f67ff1b372bab29/src/controller/CommissioningWindowOpener.cpp#L121C84-L121C98) (and the setup payload from generated values will be returned in the callback). I just double checked that by additional debug logs when passcode parameter is `NullOptional`. 

The only thing I observed making a difference here is using a non-default discriminator in the tests I conducted. 
Here's all you need:

```diff
diff --git a/examples/fabric-admin/device_manager/DeviceManager.cpp b/examples/fabric-admin/device_manager/DeviceManager.cpp
index aea9722bc8..af267187f8 100644
--- a/examples/fabric-admin/device_manager/DeviceManager.cpp
+++ b/examples/fabric-admin/device_manager/DeviceManager.cpp
@@ -33,7 +33,6 @@ namespace {
 // Constants
 constexpr uint32_t kSetupPinCode               = 20202021;
 constexpr uint16_t kRemoteBridgePort           = 5540;
-constexpr uint16_t kDiscriminator              = 3840;
 constexpr uint16_t kWindowTimeout              = 300;
 constexpr uint16_t kIteration                  = 1000;
 constexpr uint16_t kSubscribeMinInterval       = 0;
@@ -134,9 +133,12 @@ void DeviceManager::OpenRemoteDeviceCommissioningWindow(EndpointId remoteEndpoin
     // that is part of a different fabric, accessed through a fabric bridge.
     StringBuilder<kMaxCommandSize> commandBuilder;

+    // Use random discriminator to have less chance of collission.
+    uint16_t discriminator = Crypto::GetRandU16() % 4096; // 4096 to include the upper limit 4095
+
     commandBuilder.Add("pairing open-commissioning-window ");
     commandBuilder.AddFormat("%lu %d %d %d %d %d", mRemoteBridgeNodeId, remoteEndpointId, kEnhancedCommissioningMethod,
-                             kWindowTimeout, kIteration, kDiscriminator);
+                             kWindowTimeout, kIteration, discriminator);

     PushCommand(commandBuilder.c_str());
 }
```

Test Steps:

1. e1-admin adds e2-bridge (using `fabricsync add-bridge <bridgeNodeId> <addr>` command)
2. e2-admin pairs device which becomes available on an endpoint of e2-bridge (`<bridgedEndpointId>`)
3. e1-admin syncs e2-bridge (using `fabricsync sync-device <bridgedEndpointId>` command)
4. e1-admin pairs the device using payload returned from OCW request made in [3]

Test Logs:

- [device.log](https://github.com/user-attachments/files/16594585/device.log)
- [e2-bridge.log](https://github.com/user-attachments/files/16594586/e2-bridge.log)
- [e2-admin.log](https://github.com/user-attachments/files/16594587/e2-admin.log)
- [e1-admin.log](https://github.com/user-attachments/files/16594588/e1-admin.log)
- [e1-bridge.log](https://github.com/user-attachments/files/16594589/e1-bridge.log)

examples/fabric-admin/device_manager/DeviceManager.cpp
```diff
@@ -132,11 +131,15 @@ void DeviceManager::OpenRemoteDeviceCommissioningWindow(EndpointId remoteEndpoin
     // Open the commissioning window of a device from another fabric via its fabric bridge.
     // This method constructs and sends a command to open the commissioning window for a device
     // that is part of a different fabric, accessed through a fabric bridge.
-    StringBuilder<kMaxCommandSize> commandBuilder;
+    StringBuilder<512> commandBuilder;
```
joonhaengHeo, line 134: What is "512" value mean?

examples/fabric-admin/device_manager/DeviceManager.cpp
```diff
@@ -132,11 +131,15 @@ void DeviceManager::OpenRemoteDeviceCommissioningWindow(EndpointId remoteEndpoin
...
+
+    // Use random discriminator to have less chance of collission.
+    uint16_t discriminator = Crypto::GetRandU16() % 4096; // 4096 to include the upper limit 4095
```
joonhaengHeo, line 137: Rather than using the "4096" value directly, it would be more appropriate to define the value 4096 as a const value and use that variable.

tehampson (approved):

PR #34940, opened 2024-08-13
[chip-tool] Adjust the maximum value of 'discriminator' to align with spec

yufengwangca:
https://github.com/CHIP-Specifications/connectedhomeip-spec/blob/master/src/service_device_management/AdminCommissioningCluster.adoc#81-opencommissioningwindow-ocw-command

The maximum allowed value for  the 'discriminator' parameter of OpenCommissioningWindow (OCW) Command is 4095, not 4096.


samadDotDev (approved):
It would be nice if we can add doc comments on `AddArgument` declarations in `Command.h` saying that max value is inclusive here.

arkq (approved):

pidarped (approved):

PR #34942, opened 2024-08-13
Add matter_enable_recommended meta-setting

ksperling-apple:
This setting defaults to true which leaves current behavior unchanged.

However it can be set to false to achieve a more conservative / minimal set of defaults, without having to manually disable an ever-increasing set of features. Consider using this setting as the default value for other settings that increase code size or add debugging / tracing or similar features that are not always desired. Especially settings that default to true on Linux and Mac ("because you're probably building for development") should likely use this option for their default, e.g.

```
matter_foo = matter_enable_recommended && (current_os == "linux" || current_os == "mac")
```

### DIFF
--- a/src/lib/core/core.gni
+++ b/src/lib/core/core.gni
@@ -14,8 +14,17 @@
 
 declare_args() {
   # Enable recommended settings by default. This is a meta-setting
-  # that acts as a default for various other settings. Setting it
-  # to false produces a more conservative / minimal set of defaults.
+  # that is enabled by default, and acts as a default for various
+  # other settings. Setting it to false produces a more conservative /
+  # minimal set of defaults.
+  #
+  # As additional features with their own settings are added to the SDK,
+  # consider using this meta-setting instead of a default value of 'true',
+  # especially where a different default is used based on platform (current_os):
+  # E.g. various debugging features have previously been defaulted to on
+  # for Linux and/or Mac but off for embedded platforms (on the assumption
+  # that Linux / Mac don't have resource constraints?); build settings of that
+  # nature should instead reference this meta-setting.
   matter_enable_recommended = true
 
   # Enable logging. Shorthand for chip_error_logging, etc.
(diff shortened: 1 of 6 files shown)


src/lib/core/core.gni
```diff
@@ -13,6 +13,11 @@
 # limitations under the License.
 
 declare_args() {
+  # Enable recommended settings by default. This is a meta-setting
+  # that acts as a default for various other settings. Setting it
+  # to false produces a more conservative / minimal set of defaults.
+  matter_enable_recommended = true
```
andy31415, line 19: The name is reasonably generic to be unclear about what it actually enables/disables (and this seems to be the intent: a setting that is vague on what it actually enables because in the future it seems to be intended to enable more and more).

I would maybe document the intent of guarding against new features that may increase size of the binary. We also need a clear `why` rather than just what it does. The PR summary also could use some work here: in what circumstances is this to be used? Is it just decreasing binary size or something else?

ksperling-apple: I wouldn't really look at it as "enabling more and more" in the future, but rather being a single setting that can be disabled (since it's on by default) to get a more conservative set of default options. This is particularly relevant when building for Linux (and to a lesser extent Darwin), because we seem to often use `current_os == "Linux"` to enable various "debuggy" or optional things that we don't enable by default when building for other platforms. E.g.

- `chip_use_data_model_interface` defaults to `check` on Linux but `disabled` on other platforms
- `chip_enable_transport_trace` is on by default on Linux and Darwin and disabled otherwise

Maybe another way to approach to this is to ask, what is our approach within the SDK when adding features in terms of deciding whether they should be on by default?

jmartinez-silabs (approved):

Prontondr (approved):

bzbarsky-apple (approved):

src/lib/core/core.gni
ksperling-apple: I’ve added some more comments about when this could / should be used.

PR #34945, opened 2024-08-13
[Telink] Update tlsr9528a_4m_flash.overlay

s07641069:
**Change overview**

This update rearranges the flash memory layout to optimize configuration interface related to B92 flash protection. It positions the NVS at the tail and the bootloader along with firmware slots at the head of the flash, enabling enhanced flash protection settings with an initial safeguard of the first 1M.

### DIFF
--- a/src/platform/telink/tlsr9528a_4m_flash.overlay
+++ b/src/platform/telink/tlsr9528a_4m_flash.overlay
@@ -14,31 +14,35 @@
 		};
 		slot0_partition: partition@13000 {
 			label = "image-0";
-			reg = <0x13000 0x1ec000>;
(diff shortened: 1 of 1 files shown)


andy31415:
@s07641069 any chance you could explain in the summary why things got updated? This helps reviewing history of changes. Saying "updated file" and then having the PR that changes file is not that intuitive.

s07641069:
> @s07641069 any chance you could explain in the summary why things got updated? This helps reviewing history of changes. Saying "updated file" and then having the PR that changes file is not that intuitive.

updated

kiel-apple (approved):

jmartinez-silabs (approved):

PR #34947, opened 2024-08-13
Add new common NXP dockerfile

Martin-NXP:
rw61x and k32w1 will share the same new SDK, the objective of this PR is to prepare a common docker image to be able to add new SDK support in next PRs.


integrations/docker/images/stage-2/chip-build-nxp/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+    && : # last line
+
+WORKDIR /opt/nxp/
+
+RUN set -x \
+    && git clone --branch v1.4.0-pvw1 https://github.com/NXP/nxp_matter_support.git \
+    && pip3 install --break-system-packages -U --no-cache-dir west \
+    && ./nxp_matter_support/scripts/update_nxp_sdk.py --platform common \
+    && : # last line
+
```
marius-alex-tache, line 18: Should we add a `COPY --from=build <> /opt/nxp` here to update the VSCode docker image easier? E.g. https://github.com/project-chip/connectedhomeip/blob/master/integrations/docker/images/vscode/chip-build-vscode/Dockerfile#L50

andy31415: we should make sure the vscode dockerfile can still compile NXP apps. Currently we seem to copy /opt/sdk to /opt/k32w  however I am unsure if this is correct anymore. The bits from the new docker file have to be copied over too. The intent is to allow maintainers to only use one image to compile anything.

https://github.com/project-chip/connectedhomeip/blob/master/integrations/docker/images/vscode/chip-build-vscode/Dockerfile#L50

integrations/docker/images/stage-2/chip-build-nxp/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+ENV NXP_UPDATE_SDK_SCRIPT_DOCKER=/opt/nxp/nxp_matter_support/scripts/update_nxp_sdk.py
```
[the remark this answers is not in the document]
Martin-NXP: Hello,
When this environment variable is set, we will import dynamically this python file from the nxp.py script to get the path of supported platforms SDKs

integrations/docker/images/stage-2/chip-build-nxp/Dockerfile
marius-alex-tache: The `COPY --from=k32w /opt/sdk /opt/k32w` will still be valid. We need to add here a copy statement that copies the common sdk to `/opt/nxp` for example, then update the vscode Dockerfile to add `COPY --from=nxp /opt/nxp /opt/nxp`. Please note that RW will still use an older Docker image for now, so we still need to keep `COPY --from=rw61x /opt/sdk /opt/nxp-sdk`.

Martin-NXP: Thank you Andy and Marius for your comments, I pushed the code to fix the missing COPY

andy31415 (approved):

PR #34949, opened 2024-08-13
replacing "stringcase" python package with internal implementation + generated java and kotlin code

Alami-Amine:
- the python package "stringcase" is unmaintained with the latest pypi release in 2017.
- `stringcase` is being removed  and replaced with an internal implementation.

### Changes in this PR:

1. `matter_idl`: java and kotlin generators and  `filters.py`
2. `src/controller/python/chip/yaml/runner.py`
3. generated java and kotlin code; the removed stringcase package generated wrong code in constant case conversion of variables in the jinja templates:
    - example: `COMMISSIONING_A_R_L_ATTRIBUTE_ID` instead of the correct `COMMISSIONING_ARL_ATTRIBUTE_ID`
   
4. two `golden tests` used for `scripts/py_matter_idl/matter_idl/test_generators.py` were updated.
5. unit tests for the implemented functions
6. removing stringcase from various config files

### Tested:
1. matter idl and the constcase conversion using `./scripts/codegen.py -g kotlin-class src/controller/data_model/controller-clusters.matter` 
2. Running unit test cases on the implemented functions: `python scripts/py_matter_idl/matter_idl/test_case_conversion.py -v`


cecille (commented):
Do we have a place for general internal python helpers? I can see stringcase replacement functions being useful elsewhere.

maybe....

I suppose we haven't needed them up until this point, ya?

@andreilitvin - thoughts? Is there a more central place to put this, or do you want this in the matter_idl folder?

Alami-Amine:
> Do we have a place for general internal python helpers? I can see stringcase replacement functions being useful elsewhere.
> 
> maybe....
> 
> I suppose we haven't needed them up until this point, ya?
> 
> @andreilitvin - thoughts? Is there a more central place to put this, or do you want this in the matter_idl folder?

- i thought about putting it in `src/controller/python/chip/utils` since i also used these functions in `src/controller/python/chip/yaml/runner.py` but wasn't sure.
- we also have other instances of similar "case changing" functions, in `scripts/py_matter_idl/examples/matter_idl_plugin/__init__.py`, so i can also refactor my stringcase replacement functions there too

andy31415 (approved):

andy31415:
> > Do we have a place for general internal python helpers? I can see stringcase replacement functions being useful elsewhere.
> > maybe....
> > I suppose we haven't needed them up until this point, ya?
> > @andreilitvin - thoughts? Is there a more central place to put this, or do you want this in the matter_idl folder?
> 
> * i thought about putting these functions in `src/controller/python/chip/utils` since i also used these functions in `src/controller/python/chip/yaml/runner.py` ...but wasn't sure if they should be there.
> * we also have other instances of similar "case changing" functions, in `scripts/py_matter_idl/examples/matter_idl_plugin/__init__.py`, so i can also refactor my stringcase replacement functions there too

Please do not use `utils` (or support, core, helpers, base and such names that are graball). If we do want to reuse, we could have a separate package, with some name that is focused on strings, naming or something similar.

maybe `matter_name_mangling` or `matter_string_casing` or something like that. I think we may want to iterate: have this first to get the updates on generated files in, then create a separate package if we do find reuse utility

ksperling-apple (approved):

PR #34950, opened 2024-08-13
Python test scripts:  enable remote use of named pipe

plauric:
1. Refactor the Python test scripts that were using the named pipe mechanism to send out of band commands to the app under test.
2. The named pipe mechanism was only used if the test script and the app under test were running on the same machine, such as during CI. With the changes from this PR those commands can be forwarded to a DUT using SSH.


src/python_testing/matter_testing_support.py
```diff
@@ -877,6 +878,60 @@ def get_test_desc(self, test: str) -> str:
...
+            print(f"Using DUT IP address: {dut_ip}")
+
+            dut_uname = os.getenv('LINUX_DUT_UNAME')
+            if dut_uname is None:
+                dut_uname = "root"
+                print("Using default DUT user name (root)")
+            else:
+                print(f"Using DUT user name: {dut_uname}")
+
+            app_name = os.getenv('LINUX_DUT_APPNAME')
```
hicklin, line 921: This can be removed and we can use `self.app_pipe` instead.

src/python_testing/matter_testing_support.py
```diff
@@ -877,6 +878,60 @@ def get_test_desc(self, test: str) -> str:
...
+                app_name = "rvc"
+                print("Using default DUT app name (rvc)")
+            else:
+                print(f"Using DUT app name: {app_name}")
+
+            pid = self.matter_test_config.app_pid
+            # use the dut's IP address below
+            ip = dut_ip
+            command_fixed = command.replace('\"', '\\"')
+            cmd = "echo \"%s\" | ssh %s@%s \'cat > /tmp/chip_%s_fifo_%d\'" % (command_fixed, dut_uname, ip, app_name, pid)
```
hicklin, line 932:
```suggestion
            cmd = "echo \"%s\" | ssh %s@%s \'cat > %s\'" % (command_fixed, dut_uname, ip, self.app_pipe)
```

hicklin (changes requested):

hicklin (approved):

src/python_testing/matter_testing_support.py
```diff
@@ -877,6 +878,45 @@ def get_test_desc(self, test: str) -> str:
...
+
+            dut_uname = os.getenv('LINUX_DUT_UNAME')
+            if dut_uname is None:
+                dut_uname = "root"
+                print("Using default DUT user name (root)")
+            else:
+                print(f"Using DUT user name: {dut_uname}")
+
+            command_fixed = command.replace('\"', '\\"')
+            cmd = "echo \"%s\" | ssh %s@%s \'cat > %s\'" % (command_fixed, dut_uname, dut_ip, self.app_pipe)
```
andy31415, line 917: this assumes we can send stuff passwordless right?

Please make sure there is some documentation under docs describing these settings and how it works

hicklin: There is documentation about this and how to set up an ssh key on line 893. Is that enough?

andy31415: Could we add it somewhere in https://github.com/project-chip/connectedhomeip/blob/master/docs/testing/python.md as well?

hicklin: I have added documentation on how to use this in [this commit](https://github.com/project-chip/connectedhomeip/pull/34950/commits/0c2650a6d42923dffaba59d20bfa0550a5b4dfbf).

src/python_testing/TC_RVCCLEANM_2_2.py
```diff
@@ -28,8 +28,6 @@
 # === END CI TEST ARGUMENTS ===
 
 import enum
```
hicklin, line 30:
```suggestion
import enum

```

src/python_testing/TC_RVCRUNM_2_2.py
```diff
@@ -28,8 +28,6 @@
 # === END CI TEST ARGUMENTS ===
 
 import enum
```
hicklin, line 30:
```suggestion
import enum

```

hicklin (approved):

src/python_testing/matter_testing_support.py
hicklin: See as @andy31415 is on holiday until the 30th, I'm going to resolve this comment. If the documentation is not adequate we'll raise another ticket.

docs/testing/python.md
```diff
@@ -497,6 +497,61 @@ Fabric admin for default controller:
   second_ctrl = fa.new_fabric_admin.NewController(nodeId=node_id)
 ```
 
+## Automating manual steps
+
+Some test plans have manual steps that require the tester to manually change the
+state of the DUT. To run these tests in a CI environment, specific example apps
+can be built such that these manual steps can be achieved by matter or
```
kiel-apple, line 504:
```suggestion
can be built such that these manual steps can be achieved by Matter or
```

docs/testing/python.md
```diff
@@ -497,6 +497,61 @@ Fabric admin for default controller:
...
+named-pipe commands.
+
+In the case that all the manual steps in a test script can be achieved just
+using matter commands, you can check if the `PICS_SDK_CI_ONLY` PICS is set to
```
kiel-apple, line 508:
```suggestion
using Matter commands, you can check if the `PICS_SDK_CI_ONLY` PICS is set to
```

docs/testing/python.md
```diff
@@ -497,6 +497,61 @@ Fabric admin for default controller:
...
+decide if the test script should send the required matter commands to perform
```
kiel-apple, line 509:
```suggestion
decide if the test script should send the required Matter commands to perform
```

docs/testing/python.md
```diff
@@ -497,6 +497,61 @@ Fabric admin for default controller:
...
+the manual step.
+
+```python
+self.is_ci = self.check_pics("PICS_SDK_CI_ONLY")
+```
+
+In the case that a test script requires the use of name-pipe commands to achieve
```
kiel-apple, line 516:
```suggestion
In the case that a test script requires the use of named-pipe commands to achieve
```

kiel-apple (approved):

src/python_testing/matter_testing_support.py
```diff
@@ -878,6 +879,45 @@ def get_test_desc(self, test: str) -> str:
...
+        """
+        Use the following environment variables:
+
+         - LINUX_DUT_IP 
+            * if not provided, the Matter app is assumed to run on the same machine as the test,
+              such as during CI, and the commands are sent to it using a local named pipe
+            * if provided, the commands for writing to the named pipe are forwarded to the DUT
+        - LINUX_DUT_UNAME
+            * if LINUX_DUT_IP is provided, use this for the DUT user name
+            * if LINUX_DUT_NAME is not provided, a default value is used ("root")
```
tcarmelveilleux, line 893: `root` is bad hygiene. Please use a different name or remove the `None` option

src/python_testing/matter_testing_support.py
```diff
@@ -878,6 +879,45 @@ def get_test_desc(self, test: str) -> str:
...
+            * If a remote password is needed, set up ssh keys to ensure that this script can log in to the DUT without a password:
+                 + Step 1: If you do not have a key, create one using ssh-keygen
+                 + Step 2: Authorize this key on the remote host: run ssh-copy-id user@ip once, using your password
+                 + Step 3: From now on ssh user@ip will no longer ask for your password
+        """
+        import os
+        dut_ip = os.getenv('LINUX_DUT_IP')
+
+        if dut_ip is None:
+            with open(self.app_pipe, "w") as app_pipe:
```
tcarmelveilleux, line 903: self.app_pipe is not part of the base class, so this will be a problem if called without knowing `self.app_pipe` needed to be set. It also requires duplicating the PID extraction code... Please use a path prefix argument in `write_to_app_pipe`, and please append the PID to the path from the APP PID.

src/python_testing/matter_testing_support.py
```diff
@@ -878,6 +879,45 @@ def get_test_desc(self, test: str) -> str:
...
+                app_pipe.write(command + "\n")
+            # TODO(#31239): remove the need for sleep
+            sleep(0.001)
+        else:
+            print(f"Using DUT IP address: {dut_ip}")
```
tcarmelveilleux, line 908: Please replace all `print()` with `logging.info`

src/python_testing/matter_testing_support.py
```diff
@@ -878,6 +879,45 @@ def get_test_desc(self, test: str) -> str:
         except AttributeError:
             return test
 
+    # Sends an out-of-band command to a Matter app
+    def write_to_app_pipe(self, command):
```
tcarmelveilleux, line 883: Please enforce that `command` is a dict, so that arbitrary commands do not occur.

src/python_testing/matter_testing_support.py
```diff
@@ -962,6 +963,53 @@ def get_test_desc(self, test: str) -> str:
...
+
+        if dut_ip is None:
+            with open(app_pipe_name, "w") as app_pipe:
+                app_pipe.write(command + "\n")
+            # TODO(#31239): remove the need for sleep
+            sleep(0.001)
+        else:
+            logging.info(f"Using DUT IP address: {dut_ip}")
+
+            dut_uname = os.getenv('LINUX_DUT_UNAME')
```
tcarmelveilleux, line 1004: Suggest `LINUX_DUT_USER`

src/python_testing/matter_testing_support.py
```diff
@@ -962,6 +963,53 @@ def get_test_desc(self, test: str) -> str:
         except AttributeError:
             return test
 
+    def write_to_app_pipe(self, app_pipe_name: str, cmd_d: dict):
```
tcarmelveilleux, line 966: cmd_d --> command_dict

tcarmelveilleux (approved):

src/python_testing/matter_testing_support.py
```diff
@@ -963,43 +963,51 @@ def get_test_desc(self, test: str) -> str:
...
+        - LINUX_DUT_USER
             * if LINUX_DUT_IP is provided, use this for the DUT user name
-            * if LINUX_DUT_NAME is not provided, a default value is used ("root")
             * If a remote password is needed, set up ssh keys to ensure that this script can log in to the DUT without a password:
                  + Step 1: If you do not have a key, create one using ssh-keygen
                  + Step 2: Authorize this key on the remote host: run ssh-copy-id user@ip once, using your password
                  + Step 3: From now on ssh user@ip will no longer ask for your password
         """
+
+        if not isinstance(app_pipe_name, str):
```
hicklin, line 984: Given the `:str` and `:dict` in line 966, are these `if`s required?

docs/testing/python.md
```diff
@@ -497,6 +497,61 @@ Fabric admin for default controller:
...
+
+If the name of the pipe is dynamic and based on the app's PID, the following
+snippet can be added to the start of tests that use the `write_to_app_pipe`
+method.
+
+```python
+app_pid = self.matter_test_config.app_pid
+if app_pid != 0:
+	self.is_ci = true
+	self.app_pipe = "/tmp/chip_<app name>_fifo_" + str(app_pid)
```
tcarmelveilleux, line 529: Let's automate that in a follow-up. This should not be duplicated everywhere and some tests actually auto-detect the PID

src/python_testing/matter_testing_support.py
```diff
@@ -963,43 +963,51 @@ def get_test_desc(self, test: str) -> str:
         except AttributeError:
             return test
 
-    # Sends an out-of-band command to a Matter app
-    def write_to_app_pipe(self, command):
+    def write_to_app_pipe(self, app_pipe_name: str, command_dict: dict):
```
hicklin, line 966: Why are we requesting `app_pipe_name` here. We should be able to pick this from `self.app_pip`, as is still explained in the documentation. See the example below.
```
class A():
     def __init__(self):
         self.a = "default"
     def useA(self):
         print(self.a)

class B(A):
     def setA(self, a):
         self.a = a


a = A()
a.useA() # "default"

b = B()
b.useA() # "default"

b.setA("test")
b.useA() # "test"
```

src/python_testing/matter_testing_support.py
plauric: Those are type hints (rather than "constraints"), so yes.

src/python_testing/matter_testing_support.py
hicklin: Given that this is something that is only set once at the start, it would make sense to have `app_pipe` added to the base class. This way we do not need to pass the same thing every time this method is used.

src/python_testing/matter_testing_support.py
hicklin: I have added a comment below suggesting that we should be adding `app_pipe` to the base class since this is only set once at the start. It would make the calls of this method cleaner.

Or your comment
> will be a problem if called without knowing `self.app_pipe` needed to be set.
This is explained in the doc. Admittedly this is not explained it the method's doc so it should be there as well.

> It also requires duplicating the PID extraction code
If we remove this, when someone runs the app manually, there wouldn't be a way for the user to pass a PID, no? And in that case, it cannot be detected.

docs/testing/python.md
plauric: Tracked using https://github.com/project-chip/connectedhomeip/issues/35284

src/python_testing/matter_testing_support.py
plauric: @tcarmelveilleux any objections to this approach?

tcarmelveilleux: The issue is that sometimes you need to control 2 apps, so always need to provide app pipe name. We should have helpers that auto-selects a default. Let's do that as a follow-up.

```
def write_to_app_pipe(self, command_dict: dict, app_pipe_name: Optional[str] = None):
    if app_pipe_name is None:
        app_pipe_name = self.get_default_app_pipe_name()
```

For now we would be better off having `app_pipe_name` at the end (like above signature) so that when it becomes optional, we don't have to pass None each time.

PR #34955, opened 2024-08-13
Update CCTRL 2.2 to not potentially use uninitialized variables in the class on teardown

andy31415:
Otherwise we get a teardown exception, which confuses the test harness.


tehampson (approved):
If you are feeling up for it. There are also a few other tests using th_server_app_path that could uses a similar change

cecille (approved):

andy31415:
> If you are feeling up for it. There are also a few other tests using th_server_app_path that could uses a similar change

Fixed a few more that I found.

PR #34971, opened 2024-08-13
[Android] Implement Establish PASE Connection for SetupCode

joonhaengHeo:
Implement establishPaseConnection method for Android Platform.
 -> There is an IP Address, BLE only establishPaseConnection method, but there is no establishPaseConnection method using setupCode, so a new one was added.


yufengwangca (approved):

yunhanw-google (approved):

yunhanw-google:
Could you add one test in java example app for this api?

chulspro (approved):

PR #22349, opened 2022-09-01
Prevent Subscription to attributes with no access privilege and absence

yunhanw-google:
#### Problem
Reads and Subscribes from nodes with no access still get serviced to completion (which for subs mean that a subscription is established). This ends up stealing precious resources from nodes that do have access.

Fixes: issue #18485 

#### Change overview
Prevents reads/subs from moving past the processing stage to actual allocation of resources. A StatusResponse is returned immediately.

Caveat: This doesn't address lack of access for event paths. That will come separately. currently we don't have ways to expand wildcard event path and check if path exists, suggest to figure out this in post1.0. At this moment, subscription prevention with no access privilege and absence is the best effort we can go 

#### Testing
add unit test to validate empty attribute paths and event paths in subscribe request
add unit test to validate non-exist concrete attribute path and event paths in subscribe request
add unit test to validate acl-denied attribute path and event path in subscribe request

Issue #18485, opened 2022-05-16
Reject subscription establishment if subscriber has no access


mrjerryjohns:
#### Problem

See [this](https://github.com/CHIP-Specifications/connectedhomeip-spec/issues/5219) spec issue.

 #### Proposed Solution

Iterate over every expanded path in the `SubscribeRequest` and check if any of the paths have at least `View` privilege (this is already handled by calling `RequiredPrivilege::ForReadAttribute(aPath)`). If not a single path has that privilege, the subscription should not be established.

This processing should happen BEFORE we allocate a ReadHandler to serve the subscription. This ensures we don't tear down existing valid subscriptions to serve a subscription that is not valid.

woody-apple:
Cert Blocker Review: @CamWms @bzbarsky-apple @mrjerryjohns - Can we confirm if the current spec, SDK, and test plans are currently in sync. If so, we should move this to post 1.0.

andy31415:
V1 reivew: PR existed, needs reviving. Since this is a spec issue, it will be accepted in 1.0

mrjerryjohns:
@yunhanw-google to resurrect the PR that I had open previously, and ensure the relevant changes are made to that PR to ensure it is spec compliant.

PR #34983, opened 2024-08-14
Allow subscriptions if wildcard expansion is empty.

andy31415:
In #22349 we changed logic to say "if no permissions, reject subscription" to preserve resources.

This is however problematic in case some resources will be available in the future, specifically while reviewing tests for MCORE 1.2: we subscribe to all `BridgedInformationCluster::UniqueId` but without any bridged devices, the paths are empty and expansion fails.

This PR allows subscriptions to continue if subscription finds no path matches (instead of assuming permission error). note that this is still not fully ok: even if we would have some permission failures, nothing says that in the future some things without permission failures may not appear.

Overall it is unclear if we should actually deny subscriptions when we cannot fully guarantee that they will never match a valid request. This may require some spec review and review if #18485 is actually valid.

#### Changes

- Update ParseAttributePaths/EventPaths to use a single "out" parameter since several ones (especially if bools) will be easy to mix up
- Keep track of a "was list empty" for paths, to allow subscribing for paths that are yet to be created
- Update unit tests after this change


src/app/InteractionModelEngine.cpp
```diff
@@ -462,13 +462,14 @@ Status InteractionModelEngine::OnInvokeCommandRequest(Messaging::ExchangeContext
 }
 
 CHIP_ERROR InteractionModelEngine::ParseAttributePaths(const Access::SubjectDescriptor & aSubjectDescriptor,
-                                                       AttributePathIBs::Parser & aAttributePathListParser,
+                                                       AttributePathIBs::Parser & aAttributePathListParser, bool & aPathsIsNotEmpty,
```
andy31415, line 465: Blocker comment to allow time for a careful review on logic on this PR as the underlying assumptions are reasonably deep.

Overall I wonder if we should have the "deny on no ACL" at all throughout ...

woody-apple (approved):
Approving, but needs review.

ksperling-apple:
With respect to the potential denial of service (https://github.com/CHIP-Specifications/connectedhomeip-spec/issues/5219), we could require that the client node has access to anything at all on the device (rather than specifically to something matching the wildcard)

src/app/InteractionModelEngine.cpp
```diff
@@ -796,7 +798,8 @@ Protocols::InteractionModel::Status InteractionModelEngine::OnReadInitialRequest
                 return Status::InvalidAction;
             }
 
-            if (!hasValidAttributePath && !hasValidEventPath)
+            if (!(attributePathInfo.hasValidPath || eventPathInfo.hasValidPath ||
```
ksperling-apple, line 801: If there are valid paths, then we know the client has some level of access. If all paths are empty, we can instead check if the client has access to anything at all (i.e. there is some ACL entry referencing the client). Or we could check for read access to the Descriptor cluster. Arguably wildcard expansion implicitly requires read access to the descriptor cluster anyway.

andy31415: Is descriptor cluster access sufficient to check? If so we could update to that.

Otherwise generically I can only say "this is a real client, some data could be accessible at some time in the future theoretically" so if that is the case, we should never actually deny this as we do here. But that undoes the original PR that allows auto-rejection.

ksperling-apple: Thinking about this more, maybe wildcard expansion should actually explicitly check for access to the descriptor cluster anyway? It's a little strange that we're allowing information from the Descriptor cluster to be revealed when access to it hasn't necessarily been granted. On the other hand starting to enforce this now might break existing ACLs out there. But if we're treating information from the Descriptor cluster as special in this way maybe we should make this explicit in the spec, e.g. by having read access to the descriptor cluster be implicitly granted on any endpoints the client can access.

ksperling-apple:
Actually looking at the spec again, `8.4.3.2. Incoming Read Request and Subscribe Request Action Processing` seems to say that Request Path Expansion happens during processing of the Subscribe Request. Do we need to fix the spec here to even allow / mandate the wildcards to be expanded at reporting time instead of at subscription time?

andy31415:
> Actually looking at the spec again, `8.4.3.2. Incoming Read Request and Subscribe Request Action Processing` seems to say that Request Path Expansion happens during processing of the Subscribe Request. Do we need to fix the spec here to even allow / mandate the wildcards to be expanded at reporting time instead of at subscription time?

Spec has to specify expansion at reporting time because expansion at subscription time cannot be kept in memory: many generic controllers will do a full wildcard subscribe and asking devices to keep track of every possible attribute and event path seems super painful (events are even dynamic, you do not know their paths in advance). Furthermore, asking controllers to re-subscribe on every new endpoint addition also sounds like going down a wrong path.

src/app/InteractionModelEngine.cpp
mrjerryjohns: It's been a while since I waded into the spec (would need @tcarmelveilleux or @bzbarsky-apple's input here), but wildcard expansion is a server-side operation, and doesn't require the client have descriptor access I'd think. It's reasonable I'd think for a client to not have privilege to the descriptor cluster, and upon expansion, the server realize this.

A client could have access to just say, OnOff cluster, and request a wildcard, and correctly, be returned just the OnOff cluster on all matching endpoints.

andy31415:
Closing for now: apparently this DOES have spec backing:

```
...
If this action is in response to a Subscribe Request action,
  - If both AttributeRequests and EventRequests are empty
    - a Status Response Action with the INVALID_ACTION Status Code SHALL be sent to the initiator,
...
```

It is unfortunate in my mind, however changing behavirour has probably more significant sideffects.

PR #34984, opened 2024-08-14
Testing fixes for TC_SWTCH from TE2

tcarmelveilleux:
- all-clusters-app was not generating button position changes in some cases. This was not detected in some situations since the test cases don't always test for this.
- Prompts are missing endpoint ID which makes it hard when running per-endpoint tests to know where it applies.
- Some partials could fail on decode errors, causing test errors instead of fails.

This PR:

- Adds correct generation of positions on press/release.
- Adds a way to claim endpoint tested in user prompts
- Fixes failing on decode errors in partials

Testing done:

- TC_SWTCH still passes
- Manually validated button position in multi-press test/simulation (update to TC_SWTCH needs test plan changes). Issue is in all-clusters-app for CI only. See https://github.com/CHIP-Specifications/chip-test-plans/issues/4493


Apollon77 (approved):
LGTM

scripts/py_matter_yamltests/matter_yamltests/hooks.py
```diff
@@ -221,7 +221,9 @@ async def step_manual(self):
     def show_prompt(self,
                     msg: str,
                     placeholder: Optional[str] = None,
-                    default_value: Optional[str] = None) -> None:
+                    default_value: Optional[str] = None,
+                    endpoint_id: Optional[int] = None,
```
cecille, line 225: This hooks object is owned by the test harness. In theory adding an additional optional member should be fine, but would like input from the TH team re: how they're using this.

@fabiowmm @ccruzagralopes

src/python_testing/matter_testing_support.py
```diff
@@ -1305,11 +1305,20 @@ def wait_for_user_input(self,
...
+        endpoint_id = self.user_params.get("endpoint", None)
+        if endpoint_id is None or not isinstance(endpoint_id, int):
+            endpoint_id = self.matter_test_config.endpoint
+
         if self.runner_hook:
             self.runner_hook.show_prompt(msg=prompt_msg,
                                          placeholder=prompt_msg_placeholder,
-                                         default_value=default_value)
-        logging.info("========= USER PROMPT =========")
+                                         default_value=default_value,
```
cecille, line 1319: The hooks object for the test harness won't actually do anything with endpoint currently, so you're probably better off to bake this into the prompt message.

woody-apple (approved):

scripts/py_matter_yamltests/matter_yamltests/hooks.py
fabiowmm: @rquidute please take a look here.

src/python_testing/matter_testing_support.py
rquidute: Make sense @cecille, from TH perspective we won't handle endpoint argument, so it may not be necessary to be included as argument at show prompt hook method.

scripts/py_matter_yamltests/matter_yamltests/hooks.py
rquidute: From TH perspective we won't handle endpoint argument, so it may not be necessary to be included as argument at show prompt hook method.

tehampson (approved):

PR #34990, opened 2024-08-14
[FS Example] Update the FS Example apps to support fabric sync setup process part II

### LINKED ISSUE #33224: [Feature] Add Fabric Synchronization section to multi-admin in Matter SDK.

yufengwangca:
The is the remaining part of updating the Fabric-Admin and Fabric-Bridge example apps to support the Fabric-Sync setup process.

1. Start fabric-admin/fabric-bridge on E1
`./examples/fabric-admin/scripts/run_fabric_source.sh --verbose`
2. Start fabric-admin/fabric-bridge on E2
`./examples/fabric-admin/scripts/run_fabric_source.sh --verbose`
3. Add local bridge on E1
`fabricsync add-local-bridge 1`
4. Add remote bridge on E1
`fabricsync add-bridge 2 <remote-bridge-ip>`
When remote bridge is added, the CCTRL process will be automatically triggered and the E1 bridge will be reverse paired by E2 admin at the end

Fixes https://github.com/project-chip/connectedhomeip/issues/33224


examples/fabric-bridge-app/linux/RpcClient.cpp
```diff
@@ -162,6 +162,36 @@ OpenCommissioningWindow(chip::Controller::CommissioningWindowVerifierParams para
...
+    memcpy(device.salt.bytes, params.GetSalt().data(), params.GetSalt().size());
+    device.salt.size = static_cast<size_t>(params.GetSalt().size());
+
+    VerifyOrReturnError(params.GetVerifier().size() <= sizeof(device.verifier.bytes), CHIP_ERROR_BUFFER_TOO_SMALL);
+    memcpy(device.verifier.bytes, params.GetVerifier().data(), params.GetVerifier().size());
+    device.verifier.size = static_cast<size_t>(params.GetVerifier().size());
+
+    // The RPC call is kept alive until it completes. When a response is received, it will be logged by the handler
+    // function and the call will complete.
+    auto call = fabricAdminClient.ReverseCommissionBridgeTemplate(device);
```
tehampson, line 184: `ReverseCommissionBridgeTemplate` doesn't exist anywhere else, unless I am misunderstanding something shouldn't this be `ReverseCommissionBridge`

examples/fabric-bridge-app/fabric-bridge-common/BUILD.gn
```diff
@@ -45,11 +45,9 @@ source_set("fabric-bridge-lib") {
     "include/BridgedDeviceBasicInformationImpl.h",
     "include/BridgedDeviceManager.h",
     "include/CHIPProjectAppConfig.h",
-    "include/CommissionerControl.h",
```
tehampson, line 48: It looks like you are moving these files from `examples/fabric-bridge-app/fabric-bridge-common/` to `examples/fabric-bridge-app/linux/`. But this PR add to new location and doesn't remove from previous location. That makes it harder to review the delta.

examples/common/pigweed/protos/fabric_admin_service.proto
```diff
@@ -26,5 +26,6 @@ message OperationStatus {
 
 service FabricAdmin {
   rpc OpenCommissioningWindow(DeviceCommissioningWindowInfo) returns (OperationStatus){}
+  rpc ReverseCommissionBridge(DeviceCommissioningWindowInfo) returns (pw.protobuf.Empty){}
```
tehampson, line 29: Why are we renaming this here instead of calling it what it really is CommissionNode. We are not validating the device type during commissioning process and we are blindly just commissioning this device (as we should as per the spec test of Commissioner Control Cluster)

examples/fabric-admin/device_manager/DeviceManager.h
```diff
@@ -149,6 +169,7 @@ class DeviceManager : public PairingDelegate
 
     chip::NodeId mLastUsedNodeId     = 0;
     chip::NodeId mRemoteBridgeNodeId = chip::kUndefinedNodeId;
+    chip::NodeId mLocalBridgeNodeId  = chip::kUndefinedNodeId;
```
tehampson, line 172: What is the difference between LocalBridge and RemoteBridge? We should really document this to make it easier for anyone looking at this code to understand

It seems like Remote is when we do some sort of command on fabric-admin from command line and Local is when request comes from CCTRL. Why do we need to distinguish between the two?

examples/fabric-admin/commands/fabric-sync/FabricSyncCommand.cpp
```diff
@@ -90,15 +90,15 @@ CHIP_ERROR FabricSyncAddBridgeCommand::RunCommand(NodeId remoteId)
     if (DeviceMgr().IsFabricSyncReady())
     {
         // print to console
-        fprintf(stderr, "Remote Fabric Bridge has been alread configured.");
+        fprintf(stderr, "Remote Fabric Bridge has alread been configured.");
```
tehampson, line 93: nit
```suggestion
        fprintf(stderr, "Remote Fabric Bridge has already been configured.");
```

examples/fabric-admin/device_manager/DeviceManager.cpp
```diff
@@ -410,6 +431,41 @@ void DeviceManager::HandleEventData(const chip::app::EventHeader & header, chip:
...
+                             Encoding::HexFlags::kNullTerminate);
+        ChipLogProgress(NotSpecified, "  PAKEPasscodeVerifier: %s", verifierHex);
+
+        char saltHex[Crypto::kSpake2p_Max_PBKDF_Salt_Length * 2 + 1];
+        Encoding::BytesToHex(value.salt.data(), value.salt.size(), saltHex, sizeof(saltHex), Encoding::HexFlags::kNullTerminate);
+        ChipLogProgress(NotSpecified, "  salt: %s", saltHex);
+
+        OpenDeviceCommissioningWindow(mLocalBridgeNodeId, value.commissioningTimeout, value.iterations, value.discriminator,
+                                      saltHex, verifierHex);
+    }
```
tehampson, line 466: If we are receiving a `ReverseOpenCommissioningWindow::Id` isn't the intent to call rpc to fabric-bridge so that it becomes commisionable and the other side would be commissioning our FS device?

tehampson: I kinda see what you are doing here that we commission our own bridge device, but I have to ask why are we doing that when we already have the RPC mechanism to do things like that?

examples/fabric-bridge-app/linux/CommissionerControl.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+    return CHIP_NO_ERROR;
+}
+
+CHIP_ERROR CommissionerControlDelegate::ReverseCommissionNode(const CommissioningWindowParams & params,
+                                                              const Optional<ByteSpan> & ipAddress, const Optional<uint16_t> & port)
+{
+#if defined(PW_RPC_FABRIC_BRIDGE_SERVICE) && PW_RPC_FABRIC_BRIDGE_SERVICE    
+    return ReverseCommissionBridge(Controller::CommissioningWindowVerifierParams()
+                                .SetNodeId(kUndefinedNodeId)
```
tehampson, line 134: Why are you setting the NodeId `kUndefinedNodeId`, can we add a comment as to why we are doing this here

examples/fabric-admin/device_manager/DeviceManager.cpp
tehampson: Chat offline and I understand why it needs to be done this way.

For anyone else that review this, or to copy paste into the PR description:
Essentially we have to have fabric-bridge be commissioned onto a fabric that fabric-admin is the administrator for. After it is on a bridge we can the open the commissioning window with the parameters we want

examples/fabric-bridge-app/linux/RpcClient.cpp
yufengwangca: Acked

examples/fabric-bridge-app/fabric-bridge-common/BUILD.gn
yufengwangca: CommissionerControl.h/.cpp is using RPC which make them Linux dependent, removed the old files.

examples/common/pigweed/protos/fabric_admin_service.proto
yufengwangca: CommissionNode is sent from the E1 to E2 to notify the E2 reverse commission E1 bridge. They are not sent in the same direction.

examples/fabric-admin/device_manager/DeviceManager.h
yufengwangca: In E2E, each Fabric-Admin need to pair two bridge to conduct Fabric-Sync, the LocalBridge is the bridge on the same physical device, and RemoteBridge is the bridge on the other ecosystem.

examples/fabric-bridge-app/linux/CommissionerControl.cpp
yufengwangca: Removed, we don't use Node when commission a device, we don't need to set it.

examples/fabric-admin/commands/fabric-sync/FabricSyncCommand.h
```diff
@@ -22,6 +22,10 @@
 #include <commands/pairing/OpenCommissioningWindowCommand.h>
 #include <commands/pairing/PairingCommand.h>
 
+// Constants
+constexpr uint32_t kCommissionPrepareTimeMs = 500;
+constexpr uint16_t kMaxManaulCodeLength     = 21;
```
tehampson, line 27: nit,

kMaxManualCodeLength

examples/fabric-admin/device_manager/DeviceManager.cpp
```diff
@@ -33,6 +33,7 @@ namespace {
 // Constants
 constexpr uint32_t kSetupPinCode               = 20202021;
 constexpr uint16_t kRemoteBridgePort           = 5540;
+constexpr uint16_t kLocalBridgePort            = 5540;
```
tehampson, line 36: Won't this need to be configurable based on argument if we plan to have two bridges running on the same host?

examples/fabric-admin/commands/fabric-sync/FabricSyncCommand.cpp
```diff
@@ -161,6 +157,114 @@ CHIP_ERROR FabricSyncRemoveBridgeCommand::RunCommand()
...
+    if (DeviceMgr().IsLocalBridgeReady())
+    {
+        // print to console
+        fprintf(stderr, "Local Fabric Bridge has already been configured.");
+        return CHIP_NO_ERROR;
+    }
+
+    PairingCommand * pairingCommand = static_cast<PairingCommand *>(CommandMgr().GetCommandByName("pairing", "already-discovered"));
+
+    if (pairingCommand == nullptr)
```
tehampson, line 203: Could we not just VerifyOrDie here? I am assuming we never expect this to be null

examples/fabric-admin/rpc/RpcServer.cpp
```diff
@@ -62,6 +63,43 @@ class FabricAdmin final : public rpc::FabricAdmin
...
+
+        char payloadBuffer[kMaxManaulCodeLength + 1];
+        MutableCharSpan manualCode(payloadBuffer);
+
+        CHIP_ERROR error = ManualSetupPayloadGenerator(setupPayload).payloadDecimalStringRepresentation(manualCode);
+        if (error == CHIP_NO_ERROR)
+        {
+            NodeId nodeId = DeviceMgr().GetNextAvailableNodeId();
+
+            usleep(kCommissionPrepareTimeMs * 1000);
```
tehampson, line 91: Can we have a comment on why this sleep is needed? Shouldn't the node we are commissioning be ready so I don't quite understand what we are waiting for here

examples/common/pigweed/protos/fabric_admin_service.options
```diff
@@ -1,2 +1,3 @@
 chip.rpc.DeviceCommissioningWindowInfo.verifier max_size:97  // kSpake2p_VerifierSerialized_Length
```
tehampson, line 1: I really think this PR could help other reviewers if you let people know what they are doing. This PR is functionally doing 3 different things

1. FS accepting CCTRL command (This is plumbing data from fabric-bridge to fabric-admin, and commission the node)
2. Adding command to fabric-admin commissioning it's own fabric-bridge-app (Not needed for TE2, but will be needed in SVE to improve TH scripts)
3. Adding capability for fabric-admin to send CCTRL commands where upon receiving `ReverseOpenCommissioningWindow` fabric--admin opens up the commissioning window to it's fabric-bridge app via matter invoke commands (Not needed for TE2, but will be needed in SVE to improve TH scripts)

examples/fabric-admin/rpc/RpcServer.cpp
```diff
@@ -62,6 +63,43 @@ class FabricAdmin final : public rpc::FabricAdmin
...
+
+            DeviceMgr().PairRemoteDevice(nodeId, payloadBuffer);
```
tehampson, line 93: Inside `PairRemoteDevice` you are calling `chip-tool pairing code `but you are not filtering for vendor and product id. I don't have a solution for you, but this will prevent you from passing CCTRL_2_2 because DUT is expected to fail as per my understanding of the spec

tehampson: Cecille let me know how we should be doing this, we need to first establish a PASE session and read the BasicInfo cluster and make sure all the values match. If the match we should then do what we are doing here

examples/common/pigweed/protos/fabric_admin_service.options
yufengwangca: After #35020 merged, this PR should be small enough to review

examples/fabric-admin/rpc/RpcServer.cpp
tehampson: This requires spec text fix first

https://github.com/CHIP-Specifications/connectedhomeip-spec/issues/10259

examples/common/pigweed/protos/fabric_admin_service.proto
yufengwangca: Per our offline discussion, we identified that the term "Reverse" caused confusion because it has different meanings depending on whether it's from the perspective of the sender or the executor of the CommissionNode command.

To resolve this, we agreed to use "SendCommissionNode" for the side sending the CommissionNode command and "CommissionNode" for the side executing the command.

examples/fabric-admin/rpc/RpcServer.cpp
yufengwangca: Acked

examples/fabric-admin/device_manager/DeviceManager.cpp
yufengwangca: Track in https://github.com/project-chip/connectedhomeip/issues/35030

examples/fabric-admin/rpc/RpcServer.cpp
tehampson: What happens if you don't wait. I know locally I have been able to start trying to commission a device before device is comissionable. A couple seconds later I made the device commissionable and everything was fine

yufengwangca: It will eventually go through, but based on my observations, the success rate is lower, w/o the delay.

examples/common/pigweed/protos/fabric_admin_service.proto
tehampson: Have you created a github issue so we can follow up with Chris on to make sure we got the nomenclature correct

examples/fabric-admin/device_manager/DeviceManager.cpp
```diff
@@ -284,7 +342,7 @@ void DeviceManager::HandleAttributeData(const app::ConcreteDataAttributePath & p
         if (value.Has(CommissionerControl::SupportedDeviceCategoryBitmap::kFabricSynchronization))
         {
             ChipLogProgress(NotSpecified, "Remote Fabric-Bridge supports Fabric Synchronization, start reverse commissioning.");
-            StartReverseCommissioning();
+            RequestCommissioningApproval();
```
tehampson, line 345: Outside of this PR, but why are we doing this on successfully reading the attribute and not as an independent command in fabric admin?

yufengwangca: This first check of the device categories specified in SupportedDeviceCategoryBitmap has kFabricSynchronization, 

if Yes, it call RequestCommissioningApproval() which do send an independent command RequestCommissioningApproval

examples/common/pigweed/protos/fabric_admin_service.proto
yufengwangca: Which part is still not correct? we've already removed the confusing parts related to 'Reverse' from the code. We can double-check with Chris in the next sync meeting to clarify anything further.

examples/fabric-admin/device_manager/DeviceManager.cpp
tehampson: This is automated based on reading a value. If fabric-admin does `commissionercontrol read supported-device-categories` we then trigger this entire process which seems like it is not something we want to do

yufengwangca: Why? we just auto-consent in the example app to exercise the flow, since the example app is CLI based, it is quite challenging to simulate all UI interactions with the end user.

yunhanw-google (approved):

examples/common/pigweed/protos/fabric_admin_service.proto
tehampson: I think it is all correct, but we should double check with Chris that it is all correct

examples/fabric-admin/device_manager/DeviceManager.cpp
tehampson: It give us more control. I don't see why you wouldn't just manually do `commissionercontrol request-commissioning-approval ...`

Like in CI, or in the TH we can use pipes to have more control in CI/TH as to when we do things instead of just having a bunch of things done for us just because we read an attribute at one point.

Also what happens if you read this attribute a second time you are starting this all over again

tehampson: Again this can be done in a follow up, but we should have a github issue for tracking

yufengwangca: https://github.com/project-chip/connectedhomeip/issues/35041

I believe simulating user interaction should be left for the final product integration. Including these steps in the example app will complicate the automation process. Since we don't intend to commercialize the example app, let's not prioritize this task and instead focus on completing the remaining features.

examples/common/pigweed/protos/fabric_admin_service.proto
yufengwangca: https://github.com/project-chip/connectedhomeip/issues/35042

tehampson (approved):

saurabhst (approved):

PR #34992, opened 2024-08-14
Optimize RAM utilization for WriteHandler

andy31415:
This is pulled from #34754 as a stand-alone improvement:

- WriteHandlers have a multiplicity factor as an object pool resides in InteractionModelEngine
- Replacing various booleans with a bit-based flag system as well as re-ordering members should result in some RAM savings (likely minor, but non-zero)

### DIFF
--- a/src/app/WriteHandler.cpp
+++ b/src/app/WriteHandler.cpp
@@ -16,19 +16,23 @@
  *    limitations under the License.
  */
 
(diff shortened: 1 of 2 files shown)


yunhanw-google (approved):
Thank you!

woody-apple (approved):

PR #34993, opened 2024-08-14
[Fabric-Sync] Simplify and update the fabric sync guide

yufengwangca:
1. Simplify the setup scripts and update the fabric sync guide
2. Disable the auto-sync feature to prevent cyclic syncing


docs/guides/fabric_synchronization_guide.md
```diff
@@ -114,36 +130,54 @@ Pair the Light Example with node ID 3 using its payload number:
...
-After the Light Example is successfully paired in Fabric-Source, it will be
-synced to Fabric-Sink with a new assigned node ID.
+After the device is successfully added, you will observe the following on
+Ecosystem 2 with the newly assigned Node ID:
+
+```
+>>> New device with Node ID: 0x3 has been successfully added.
+```
+
+Additionally, you should also get notified an new device is added on the
```
saurabhst, line 140:
```suggestion
Additionally, you should also get notified when a new device is added to
```

examples/fabric-admin/commands/fabric-sync/Commands.h
```diff
@@ -29,7 +29,6 @@ void registerCommandsFabricSync(Commands & commands, CredentialIssuerCommands *
         make_unique<FabricSyncAddBridgeCommand>(credsIssuerConfig),
         make_unique<FabricSyncRemoveBridgeCommand>(credsIssuerConfig),
         make_unique<FabricSyncDeviceCommand>(credsIssuerConfig),
-        make_unique<FabricAutoSyncCommand>(credsIssuerConfig),
```
tehampson, line 32: nit, why is this getting removed here?

If you do want to remove this why not also removed from `examples/fabric-admin/commands/fabric-sync/FabricSyncCommand.h`

saurabhst: > nit, why is this getting removed here?
> 
> If you do want to remove this why not also removed from `examples/fabric-admin/commands/fabric-sync/FabricSyncCommand.h`

+1 I have the same concern/comment/clarification

docs/guides/fabric_synchronization_guide.md
```diff
@@ -114,36 +130,54 @@ Pair the Light Example with node ID 3 using its payload number:
...
+Ecosystem 2 from the Ecosystem 1:
+
+```
+>>> A new devie is added on Endpoint 3.
```
tehampson, line 144: nit
```suggestion
>>> A new device is added on Endpoint 3.
```

examples/fabric-admin/commands/fabric-sync/Commands.h
yufengwangca: Auto-sync will cause cyclical syncing at this moment,  but we should be able to solve this with proper algorithm using uniqueID.

I don't see this feature in the test plan, do we want to remove this feature for good? then I will clean up all related pieces of this feature.

tehampson: We can leave the command in and mark the command as use at your own risk, but I don't understand why we are removing it

yufengwangca: Acked

saurabhst (approved):

tehampson (approved):

PR #34995, opened 2024-08-14
Add data-model enable/disable/check flags for building on the host

andy31415:
This pulls these flags from #34754 as a stand-alone change.

This adds 3 separate modifiers: `data-model-check`, `data-model-disabled` and `data-model-enabled` when building host applications.


woody-apple (approved):

woody-apple:
Fast tracking test scripts

PR #34926, opened 2024-08-12
[NXP][k32w1] Rework reference apps

marius-alex-tache:
Changelog:
- Add multiple components for the app common code base
- Remove `k32w` parent folder: `k32w/k32w1` becomes `k32w1` for all related paths
- Switch `k32w1` filesystem to NVS
- Add `nxp_factory_data` and `nxp_ota` platform gn targets
- Bump `nxp_matter_support` repo
- Other bug fixes

To do: This will require a Docker image update to reference latest SDK. https://github.com/project-chip/connectedhomeip/pull/34927

PR #35009, opened 2024-08-15
Update compatible builds to docker version 74

s07641069:
**Change overview**

- Update compatible builds to latest docker version 74

**Builds required old docker version 54 (not adopted yet for ubuntu 24.04):**

- build_linux_python_lib: name: Build on Linux (python_lib)
- build_linux_python_lighting_device: name: Build on Linux (python lighting-app)
- infineon: name: Infineon examples building
- qemu-tizen: name: Tizen

EFR32 require additional changes to use image version 74 https://github.com/project-chip/connectedhomeip/pull/35011
https://github.com/project-chip/connectedhomeip/actions/runs/10415011158/job/28844869810?pr=35009
I leave EFR32 in version 73


s07641069:
Blocked by https://github.com/project-chip/connectedhomeip/pull/34980
Will test in background which images we could update to latest by using 72 temporary, till version 73 is not published

andy31415:
> Blocked by #34980 @andy31415 I will test in background which images we could update to latest by using 72 temporary, till version 73 is not published
> 
> Builds required old docker version 54:
> 
> * build_linux_python_lib: name: Build on Linux (python_lib)
> * build_linux_python_lighting_device: name: Build on Linux (python lighting-app)
> * infineon: name: Infineon examples building
> * qemu-tizen: name: Tizen
> 
> K32W image update will be performed under #34926

I am building 73 now, however also 74 is soon following in #35011 .

s07641069:
> > Blocked by #34980 @andy31415 I will test in background which images we could update to latest by using 72 temporary, till version 73 is not published
> > Builds required old docker version 54:
> > 
> > * build_linux_python_lib: name: Build on Linux (python_lib)
> > * build_linux_python_lighting_device: name: Build on Linux (python lighting-app)
> > * infineon: name: Infineon examples building
> > * qemu-tizen: name: Tizen
> > 
> > K32W image update will be performed under #34926
> 
> I am building 73 now, however also 74 is soon following in #35011 .

No problem I can change in this PR into 74, when you will finish it
Also will be good to merge https://github.com/project-chip/connectedhomeip/pull/34926 to have almost all imges with latest docker version
See PR discribtion (change overview)

s07641069:
@andy31415
EFR32 require additional changes to use image version 74 https://github.com/project-chip/connectedhomeip/pull/35011
https://github.com/project-chip/connectedhomeip/actions/runs/10415011158/job/28844869810?pr=35009
I leave EFR32 in version 73 and add this to PR description

.github/workflows/tests.yaml
```diff
@@ -49,7 +49,7 @@ jobs:
         runs-on: ubuntu-latest
 
         container:
-            image: ghcr.io/project-chip/chip-build:54
+            image: ghcr.io/project-chip/chip-build:74
```
andy31415, line 52: This one sweems to switch from ubuntu 20.04 to 24.04 and I seem to get an error in a python test case (which makes no sense, however I am not familiar with it). I am re-running it the 3rd time now, however it failed 2 times in a row, so it may somehow be a legitimate error, potentially related to some python functionality likely (just a guess .. no idea what is going on, but python being different seems more likely than apps being different).

We may want to create an issue for this and not update it yet.

andy31415: TestPurposefulFailureExtraReportingOnToggle seems to be the unexpected passing test in CI.

andy31415: I can reproduce this issue on my local machine as well (which is a recent ubuntu-based).
I believe this test case may be in error however I am not familiar with it.

andy31415: test added as part of #31162 by @bzbarsky-apple 
it seems the intent is that we should not report attribute changes when the attribute does not change.

andy31415: Debugging locally, it seems like the parser stops completely at `Running Check for attribute report` (it does not seem to actually be able to pass that check at all, but just stops without an exception). I am unclear why.

andy31415: The test seems to timeout (get no instance.recv in  WebSocketRunner::execute), yet no timeout exception is thrown.

andy31415: what I observe is that https://github.com/project-chip/connectedhomeip/blob/master/scripts/py_matter_yamltests/matter_yamltests/runner.py#L225 finishes after 5 seconds (which is what the test calls for) and no exception is raised.

s07641069: Do I need to revert it from 74 back to 54 for now?

andy31415: I think #35045 contains a fix for this.

andy31415 (approved):

PR #35016, opened 2024-08-15
Python testing: Fix reporting on setup_class error

### LINKED ISSUE #34959: [BUG] Python test run does not call failure method from hooks when setup_class fails
### Reproduction steps

TH is not getting a failure when a python test fails the setup_class.

Running the `TC-IDM-10.2` using TH, and performing a commissioning before running the test case.
TH Execution log for TC-IDM-10.2:  [UI_Test_Run_2024_08_13_15_10_32.log](https://github.com/user-attachments/files/16603689/UI_Test_Run_2024_08_13_15_10_32.log)

### Bug prevalence

always

d0d91272068f267cf880f9d56787ca28da885673

### Platform

python

### Platform Version(s)

_No response_

### Anything else?

_No response_

cecille:
Also add error text to make the error easier to find.

Fixes: https://github.com/project-chip/connectedhomeip/issues/34959

### DIFF
--- a/src/controller/python/chip/exceptions/__init__.py
+++ b/src/controller/python/chip/exceptions/__init__.py
@@ -39,21 +39,17 @@ class ChipStackException(Exception):
 
 
 class ChipStackError(ChipStackException):
-    def __init__(self, chip_error: PyChipError, msg=None):
-        self._chip_error = chip_error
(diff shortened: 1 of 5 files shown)


woody-apple (approved):
Pre-approving so this doesn't get blocked.

PR #35017, opened 2024-08-15
Make BRBINFO_4_1 require commissioning

andy31415:
Add a `is_commissioning` step to BRBINFO 4.1

### DIFF
--- a/src/python_testing/TC_BRBINFO_4_1.py
+++ b/src/python_testing/TC_BRBINFO_4_1.py
@@ -57,7 +57,8 @@ class TC_BRBINFO_4_1(MatterBaseTest):
 
     def steps_TC_BRBINFO_4_1(self) -> list[TestStep]:
         steps = [
-            TestStep("0", "Preconditions"),
+            TestStep("0",  "DUT commissioned", is_commissioning=True),
+            TestStep("0a", "Preconditions"),
             TestStep("1a", "TH reads from the ICD the A_IDLE_MODE_DURATION, A_ACTIVE_MODE_DURATION, and ACTIVE_MODE_THRESHOLD attributes"),
             TestStep("1b", "Simple KeepActive command w/ subscription. ActiveChanged event received by TH contains PromisedActiveDuration"),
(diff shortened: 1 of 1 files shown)


src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -57,7 +57,8 @@ def desc_TC_BRBINFO_4_1(self) -> str:
 
     def steps_TC_BRBINFO_4_1(self) -> list[TestStep]:
         steps = [
-            TestStep("0", "Preconditions"),
+            TestStep("0",  "DUT commissioned", is_commissioning=True),
+            TestStep("0a", "Preconditions"),
             TestStep("1a", "TH reads from the ICD the A_IDLE_MODE_DURATION, A_ACTIVE_MODE_DURATION, and ACTIVE_MODE_THRESHOLD attributes"),
             TestStep("1b", "Simple KeepActive command w/ subscription. ActiveChanged event received by TH contains PromisedActiveDuration"),
```
tehampson, line 63: While you are touching this file can you add in `_ask_for_vendor_commissioniong_ux_operation` that fabric-admin should include `--icd-registration true`

src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -57,7 +57,8 @@ def desc_TC_BRBINFO_4_1(self) -> str:
 
     def steps_TC_BRBINFO_4_1(self) -> list[TestStep]:
         steps = [
-            TestStep("0", "Preconditions"),
+            TestStep("0",  "DUT commissioned", is_commissioning=True),
```
tehampson, line 60: I think the Precondisions step can have the `is_commissioning=True`

andy31415: I thought this would be safer/consistent. Could we keep it like this for now as it works?

saurabhst (approved):

tehampson (approved):

PR #35022, opened 2024-08-15
Add fabric scoping to ECOINFO cluster attributes

tehampson:
This adds FabricIndex to properly implement fabric-scoping to the all the DeviceDirectory entries and LocationDirectory entries.

Alternate that was considered but not implemented:
* Have FabricIndex be a set in the entry as an allowlist of FabricIndex's
   * Why approach was not taken: Tracking number of element in the directory to remain conformant to spec max become more difficult. More corner cases to consider when adding API that allows the set to be change
* Have FabricIndex be a set in the entry as an denylist of FabricIndex's
   * Add complexity where this cluster code needs to know fabric table and be aware when fabrics are added removed.
   * Complexity on when add fabric occurs and we don't have enough entry space in the directory

### DIFF
--- a/src/app/clusters/ecosystem-information-server/ecosystem-information-server.cpp
+++ b/src/app/clusters/ecosystem-information-server/ecosystem-information-server.cpp
@@ -237,7 +237,6 @@ CHIP_ERROR EcosystemLocationStruct::Encode(const AttributeValueEncoder::ListEnco
     locationStruct.locationDescriptor         = GetEncodableLocationDescriptorStruct(mLocationDescriptor);
     locationStruct.locationDescriptorLastEdit = mLocationDescriptorLastEditEpochUs;
     locationStruct.SetFabricIndex(aFabricIndex);
-
     return aEncoder.Encode(locationStruct);
 }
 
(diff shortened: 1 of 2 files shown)


src/app/clusters/ecosystem-information-server/ecosystem-information-server.cpp
```diff
@@ -128,6 +128,13 @@ EcosystemDeviceStruct::Builder & EcosystemDeviceStruct::Builder::AddUniqueLocati
     return *this;
```
tehampson, line 128: Blocking comment, waiting until after TE2 to submit this PR

src/app/clusters/ecosystem-information-server/ecosystem-information-server.h
```diff
@@ -134,15 +136,11 @@ class EcosystemLocationStruct
         mLocationDescriptor(aLocationDescriptor), mLocationDescriptorLastEditEpochUs(aLocationDescriptorLastEditEpochUs)
     {}
     // EcosystemLocationStruct is used as a value in a key-value map.
-    // Because UniqueLocationId is manditory when an entry exist, and
-    // it is unique, we use it as a key to the key-value pair and is why it is
+    // Because UniqueLocationId and FabricIndex are manditory when an entry exist,
```
andy31415, line 139:
```suggestion
    // Because UniqueLocationId and FabricIndex are mandatory when an entry exist,
```

andy31415 (approved):

yunhanw-google (approved):

saurabhst (approved):

PR #35024, opened 2024-08-15
Disable TI builds from CI

andy31415:
This is because we seem to perma-fail due to out of space recently.

To be enabled when we have space AND we update to a more recent image.

### DIFF
--- a/.github/workflows/examples-cc13xx_26xx.yaml
+++ b/.github/workflows/examples-cc13xx_26xx.yaml
@@ -15,11 +15,15 @@
 name: Build example - TI CC13XX_26XX
 
 on:
-    push:
-        branches-ignore:
-            - "dependabot/**"
-    pull_request:
-    merge_group:
+    workflow_dispatch:
+    # Temporarely disabled:
+    #    - TI CI runs out of disk space
+    #    - Image should be updated to an Ubuntu 24.04 or higher based one (move from :54 version)
+    # push:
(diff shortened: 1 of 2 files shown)


s-jain2022:
@andy31415 - Can you please help me understand - What is needed from TI to enable the CI for the CC13x4 26x4 platform again?

andy31415:
> @andy31415 - Can you please help me understand - What is needed from TI to enable the CI for the CC13x4 26x4 platform again?

@s-jain2022  Need to make CI pass. It seems like github is more strict on space available for github actions and docker images + repository checkout + build artifacts need to fit inside the given storage (I believe 14GB). TI builds were running out of space. This needs to be fixed for example by finding out how to minimize the image for the TI platform, minimize the amount of checked out data (e.g. submodules needed) or making outputs smaller (e.g. cleaning outputs if compiling many things at once or not producing large artifacts like map files).

I would suggest to try to create a PR to re-enable the CI build and see if it fails or not. if it still fails, you have to fix it somehow. Otherwise maybe we somehow trimmed dependencies in the mean time.

andy31415:
Please also make sure you use a 24.04 image (e.g. ghcr.io/project-chip/chip-build-ti:77) and NOT a 54 image. We need to use the ubuntu 24.04 versions.

s-jain2022:
@andy31415 Thanks.

PR #35026, opened 2024-08-15
Fix BRBINFO_4_1 to run further

tehampson:
Issues notices with test that prevent it from running far.

In order to assist understand time calculations suffixes were added to variable names

This PR is only to address test steps as per the current test plan spec PR.

There are other known issues with this script that will be addressed in a follow up after we have updated spec text, test plan, and test implementation

### DIFF
--- a/src/python_testing/TC_BRBINFO_4_1.py
+++ b/src/python_testing/TC_BRBINFO_4_1.py
@@ -81,11 +81,12 @@ class TC_BRBINFO_4_1(MatterBaseTest):
         keep_active = await self.default_controller.SendCommand(nodeid=self.dut_node_id, endpoint=endpoint_id, payload=Clusters.Objects.BridgedDeviceBasicInformation.Commands.KeepActive(stayActiveDuration=duration))
         return keep_active
 
-    async def _wait_for_active_changed_event(self, timeout) -> int:
+    async def _wait_for_active_changed_event(self, timeout_s) -> int:
         try:
-            promised_active_duration = self.q.get(block=True, timeout=timeout)
-            logging.info(f"PromisedActiveDuration: {promised_active_duration}")
-            return promised_active_duration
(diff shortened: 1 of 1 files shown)


src/python_testing/TC_BRBINFO_4_1.py
```diff
@@ -218,49 +219,53 @@ async def test_TC_BRBINFO_4_1(self):
...
-        time.sleep(100)
-        logging.info(f"Sending KeepActiveCommand({stay_active_duration})")
-        self._send_keep_active_command(stay_active_duration, dynamic_endpoint_id)
-        time.sleep(100)
+        logging.info(f"Step3 Sending first KeepActiveCommand({stay_active_duration_ms})")
+        await self._send_keep_active_command(stay_active_duration_ms, dynamic_endpoint_id)
+        time.sleep(0.1)
+        logging.info(f"Step3 Sending second KeepActiveCommand({stay_active_duration_ms})")
+        await self._send_keep_active_command(stay_active_duration_ms, dynamic_endpoint_id)
+        time.sleep(0.1)
```
cecille, line 254: I'm not sure what the hig sleep was before, so I have no idea if this change makes sense. Why was the sleep so long before? Is the idea to send three commands before the event can be generated? If so, this feels racy.

tehampson: That is because it is racy. I have a plan to address this (aka remove these sleeps completely) in a follow up PR in test plan spec and implementation

tehampson: Here is the first test plan change:
https://github.com/CHIP-Specifications/chip-test-plans/pull/4512

Another small tweak will be made in a follow up if https://github.com/CHIP-Specifications/connectedhomeip-spec/issues/10236 improvement lands

cecille (approved):

saurabhst (approved):

PR #35027, opened 2024-08-15
Fix test code build issues when INET_CONFIG_TCP_ENDPOINT is set to false

pidarped:
test apps echo-requestor/echo-responder and some TestInet files were failing.

### DIFF
--- a/src/messaging/tests/echo/echo_responder.cpp
+++ b/src/messaging/tests/echo/echo_responder.cpp
@@ -45,7 +45,9 @@ namespace {
 // The EchoServer object.
 chip::Protocols::Echo::EchoServer gEchoServer;
 chip::TransportMgr<chip::Transport::UDP> gUDPManager;
+#if INET_CONFIG_ENABLE_TCP_ENDPOINT
 chip::TransportMgr<chip::Transport::TCP<kMaxTcpActiveConnectionCount, kMaxTcpPendingPackets>> gTCPManager;
+#endif // INET_CONFIG_ENABLE_TCP_ENDPOINT
 chip::SessionHolder gSession;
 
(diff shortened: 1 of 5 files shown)


yufengwangca (approved):

Alami-Amine (approved):

andy31415 (approved):

src/messaging/tests/echo/echo_responder.cpp
```diff
@@ -85,13 +87,15 @@ int main(int argc, char * argv[])
...
     {
+#if INET_CONFIG_ENABLE_TCP_ENDPOINT
         err = gTCPManager.Init(chip::Transport::TcpListenParameters(chip::DeviceLayer::TCPEndPointManager())
                                    .SetAddressType(chip::Inet::IPAddressType::kIPv6));
         SuccessOrExit(err);
 
         err = gSessionManager.Init(&chip::DeviceLayer::SystemLayer(), &gTCPManager, &gMessageCounterManager, &gStorage,
                                    &gFabricTable, gSessionKeystore);
         SuccessOrExit(err);
+#endif // INET_CONFIG_ENABLE_TCP_ENDPOINT
```
aronrosenberg, line 98: Does there need to a #else here if someone tries to run the TCP test with no TCP support compiled in and have that either error out or print a warning message?

aronrosenberg (approved):

src/messaging/tests/echo/echo_responder.cpp
pidarped: Have folded in the `useTCP` flag inside INET_CONFIG_ENABLE_TCP_ENDPOINT.

Issue #33360, opened 2024-05-08
Split embedded tests into individual binaries


andy31415:
As we add more and more tests, as single large test binary that can be flashed to devices is not practical anymore. We already are excluding some tests for "not fitting in flash" and even after that we make fixes like #33347 because insufficient flash on some parts.

We should have each chip-test generate a separate flashable binanary, the same like it is built on the host.

We likely should do this after the migration of all tests to pw_unit_test or make the pw_unit_test version build separate binaries.

PR #35028, opened 2024-08-16
Generating separate binaries for each unit test for EFR32

feasel0:
Pursuant to [Issue 33360](https://github.com/project-chip/connectedhomeip/issues/33360), this PR splits up the tests for EFR32 into separate flashable binaries for each test source file.

- The binaries, along with all their related files, are output to `$root_output_dir/tests` with a filename like "matter-silabs-device_tests_TestSomething.s37".
- Updated test driver so that if it's given a directory instead of an individual binary, it will flash and run all the s37 files in that directory.
- Updated the python wheels to contain the Pigweed version of the test runner instead of the NL version.
- Removed the NL version of the test runner.
- Removed `chip_test.gni` since it is no longer used after the transition to pw unit test.
- Removed the unused/deprecated feature of having `chip_test_suite` accept the variable `tests`, which adds both a test source and a corresponding custom test driver.

Building `silabs_executable` for each test source:

- Rather than defining a `silabs_executable` target in `src/test_driver/efr32/BUILD.gn`, we set `pw_unit_test_EXECUTABLE_TARGET_TYPE` to "silabs_executable" and we let `pw_test` generate the executable target.
- `src/test_driver/efr32/BUILD.gn` defines a source_set for the test driver, which is dep'd by each `pw_test`.
- In addition to our matter tests, this also turns all the internal pigweed tests into `silabs_executable`s.  Since these do not define `output_name`, we set a default value for `output_name`.
- `silabs_executable.gni` and `flashable_executable.gni` are modified to output to `invoker.output_dir` instead of `root_out_dir` so that it's possible to specify a directory besides the root.   If `output_dir` is not defined it will default to `root_out_dir`.
- Because the efr32 test binaries are now output to the "tests" sub-directory (like what's done in the main build), `silabs_executable` needed modification to fix some conflicts that occurred if multiple `silabs_executable`s are built to the same directory:
- `silabs_firmware_utils.py` and `firmware_utils.py` can't be copied via GN's `copy()` any more since multiple tests will need to copy these files into the same output directory ("tests"), and GN won't allow that.  So the copying is now done in `silabs_firmware_utils.py` when `make_wrapper` is called during the creation of `{target}.flash.py`.
- The `{target}.flashbundle.txt` file can no longer be generated via `write_runtime_deps` since the `copy()` target is gone (and would have the wrong output name even if it wasn't).  So the creation of `{target}.flashbundle.txt` is now done in `silabs_firmware_utils.py` when `make_wrapper` is called during the creation of `{target}.flash.py`.

Changes to efr32 builder py:

- Updated the builder script to correctly build the "unit-test" example via `build_examples.py`.  This generates a flashbundle that includes all of the s37 and flash.py files (one for each unit test) and the two firmware utils scripts that support it.
- Modified the builder script so that it doesn't copy the s37 and flash.py files into the artifacts directory unless the `--enable-flashbundle` argument is given.  Previously it had been copying those files to the artifacts directory even if that argument wasn't given (the argument was not being read).

`test_sources` are no longer included as `sources` in the suite-wide common lib (except when `chip_link_tests` = false):

- When `chip_link_tests` is true we create a `pw_test` target for each test source, which includes only that one test source and then deps the suite-wide common lib (which contains the common `sources`).  Previously the common lib had also contained all of the `test_sources` (in addition to the common `sources`), but that's unnecessary.
- The one exception is the Darwin build, which seems to require all the `test_sources` be included as `sources` in the common lib.  So we add a condition that forces the inclusion when it's Darwin.


scripts/build/builders/efr32.py
```diff
@@ -259,27 +261,64 @@ def __init__(self,
...
+                os.path.join(self.output_dir, "tests", "*.flash.py"),
+                os.path.join(self.output_dir, "tests", "*.s37"),
+                os.path.join(self.output_dir, "tests", "silabs_firmware_utils.py"),
+                os.path.join(self.output_dir, "tests", "firmware_utils.py"),
+            ]
+
+            # Generate the list of files by globbing each pattern.
+            files = []
+            for pattern in patterns:
+                files += list(map(lambda x: os.path.basename(x), glob.glob(pattern)))
```
andy31415, line 280: nit: list + map is often easier to read in a `[foo(x) for x in y]` (instead of `list(map(lambda x: foo(x), y))`).

So here it may be:

```py
files.extend([os.path.basename(x) for x in glob.glob(pattern)])
```

and below it may be:

```py
for name in [os.path.basename(x) for x in glob.glob(pattern)]:
```

scripts/flashing/silabs_firmware_utils.py
```diff
@@ -169,6 +171,45 @@ def actions(self):
...
+            if args.firmware_utils is not None:
+                flashbundle_contents += "\n" + os.path.basename(args.firmware_utils)
+                shutil.copy(args.firmware_utils, output_dir)
+
+            # Create the flashbundle file.
+            try:
+                with open(args.flashbundle_file, 'w') as flashbundle_file:
+                    flashbundle_file.write(flashbundle_contents.strip())
+            except OSError as exception:
+                print(exception, sys.stderr)
```
andy31415, line 196: could you add a comment as to why we print this instead of propagating?

Would https://docs.python.org/3/library/traceback.html#traceback.print_last result in a nicer printout maybe?

andy31415 (approved):

scripts/flashing/silabs_firmware_utils.py
feasel0: Good catch.  Instead of printing here, this needs to propagate back to `maker_wrapper` so it knows to `return 1` instead of just moving on.  I also updated `make_wrapper` to use `traceback.print_last()` instead of `print(exception, sys.stderr)`.

jmartinez-silabs (approved):

PR #35038, opened 2024-08-16
[MCORE.FS 1.3] make "app not found" easier to debug

andy31415:
This removes an unused method (simpler script) and tries harder to detect application not found (much more likely since having an app without a unique ID reporting is hard).

### DIFF
--- a/src/python_testing/TC_MCORE_FS_1_3.py
+++ b/src/python_testing/TC_MCORE_FS_1_3.py
@@ -68,44 +68,31 @@ class TC_MCORE_FS_1_3(MatterBaseTest):
         super().teardown_class()
 
     async def create_device_and_commission_to_th_fabric(self, kvs, port, node_id_for_th, device_info):
-        # TODO: confirm whether we can open processes like this on the TH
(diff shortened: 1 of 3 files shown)


src/python_testing/TC_MCORE_FS_1_3.py
```diff
@@ -68,45 +68,32 @@ def teardown_class(self):
         super().teardown_class()
 
     async def create_device_and_commission_to_th_fabric(self, kvs, port, node_id_for_th, device_info):
-        # TODO: confirm whether we can open processes like this on the TH
         app = self.user_params.get("th_server_app_path", None)
```
tehampson, line 71: Would you mind also doing this to all the other test scripts that use th_server_app_path

tehampson (approved):

woody-apple (approved):

cecille (approved):

src/python_testing/TC_MCORE_FS_1_3.py
andy31415: Done.

PR #35046, opened 2024-08-16
Add Write Handler in Chef Fan Control AttributeAccessInterface

yyzhong-g:
This is a chef app fan control on change.

The fan control manager is not handling the dirty bit of FanControl::*Current when writing to FanControl::*Setting attributes. This changes is to make sure that the write happens in AttributeAccessInterface layer instead of the legacy attribute storage and mark both *Current and *Setting dirty during writes.


examples/chef/common/chef-fan-control-manager.cpp
```diff
@@ -118,17 +99,94 @@ Status ChefFanControlManager::HandleStep(StepDirectionEnum aDirection, bool aWra
...
-        return ReadPercentCurrent(aEncoder);
+    case PercentCurrent::Id: {
+        // Current percents always tracks setting immediately in our implementation.
+        return aEncoder.Encode(mPercentSetting.ValueOr(0));
+    }
+    case PercentSetting::Id: {
+        return aEncoder.Encode(mPercentSetting);
+    }
+    case SpeedCurrent::Id: {
+        // Current speed always trackssetting immediately in our implementation.
```
andy31415, line 184:
```suggestion
        // Current speed always tracks setting immediately in our implementation.
```

yunhanw-google (approved):

bzbarsky-apple (approved):

PR #35048, opened 2024-08-17
[Darwin] more XPC service tweaks

kiel-apple:
- override `_setupDeviceForNodeID:...` so we can get `MTRDevice_XPC`s from it
- more logging
- raise device map and its locks

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDeviceController.mm
+++ b/src/darwin/Framework/CHIP/MTRDeviceController.mm
@@ -120,7 +120,8 @@ using namespace chip::Tracing::DarwinFramework;
     MTRDeviceAttestationDelegateBridge * _deviceAttestationDelegateBridge;
     MTRDeviceControllerFactory * _factory;
     NSMapTable * _nodeIDToDeviceMap;
-    os_unfair_lock _deviceMapLock; // protects nodeIDToDeviceMap
+    os_unfair_lock_t _deviceMapLock; // protects nodeIDToDeviceMap
+    os_unfair_lock _underlyingDeviceMapLock;
     MTRCommissionableBrowser * _commissionableBrowser;
     MTRAttestationTrustStoreBridge * _attestationTrustStoreBridge;
 
@@ -134,14 +135,16 @@ using namespace chip::Tracing::DarwinFramework;
     MTRP256KeypairBridge _operationalKeypairBridge;
 }
 
-@synthesize deviceMapLock = _deviceMapLock;
+- (os_unfair_lock_t)deviceMapLock {
+    return &_underlyingDeviceMapLock;
+}
 
 - (instancetype)initForSubclasses
 {
     if (self = [super init]) {
         // nothing, as superclass of MTRDeviceController is NSObject
     }
-
+    _underlyingDeviceMapLock = OS_UNFAIR_LOCK_INIT;
     return self;
 }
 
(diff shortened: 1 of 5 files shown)


src/darwin/Framework/CHIP/MTRDeviceController_XPC.mm
```diff
@@ -33,6 +35,8 @@
 @interface MTRDeviceController_XPC ()
 
 @property (nonatomic, retain, readwrite) NSUUID * uniqueIdentifier;
+@property (nonatomic, readonly, nullable) NSMapTable * nodeIDToDeviceMap;
```
woody-apple, line 38: This should be exported from an internal header not re-declared here, otherwise they're a shadow

woody-apple: Put these in MTRDeviceController_Internal.h

src/darwin/Framework/CHIP/MTRDeviceController_XPC.mm
```diff
@@ -41,11 +43,13 @@ @interface MTRDeviceController_XPC ()
 @implementation MTRDeviceController_XPC
 
 @synthesize uniqueIdentifier = _uniqueIdentifier;
+@synthesize nodeIDToDeviceMap = _nodeIDToDeviceMap;
+@synthesize deviceMapLock = _deviceMapLock;
```
anush-apple, line 47: Why do you need these? They are already in the base implementation

src/darwin/Framework/CHIP/MTRDeviceController_Concrete.mm
```diff
@@ -136,6 +136,7 @@ @implementation MTRDeviceController_Concrete {
 @synthesize commissionableBrowser = _commissionableBrowser;
 @synthesize concurrentSubscriptionPool = _concurrentSubscriptionPool;
 @synthesize storageBehaviorConfiguration = _storageBehaviorConfiguration;
+@synthesize nodeIDToDeviceMap = _nodeIDToDeviceMap;
```
woody-apple, line 139: Not needed

src/darwin/Framework/CHIP/MTRDeviceController_XPC.mm
woody-apple: Yup, not needed, they'll create shadows.

src/darwin/Framework/CHIP/MTRDeviceController_XPC.mm
```diff
@@ -111,6 +115,24 @@ - (nullable instancetype)initWithParameters:(MTRDeviceControllerAbstractParamete
     return nil;
 }
 
+// If prefetchedClusterData is not provided, load attributes individually from controller data store
+- (MTRDevice *)_setupDeviceForNodeID:(NSNumber *)nodeID prefetchedClusterData:(NSDictionary<MTRClusterPath *, MTRDeviceClusterData *> *)prefetchedClusterData
```
anush-apple, line 119: Declare this in the internal header

src/darwin/Framework/CHIP/MTRDeviceController.mm
```diff
@@ -120,7 +120,8 @@ @implementation MTRDeviceController {
     MTRDeviceAttestationDelegateBridge * _deviceAttestationDelegateBridge;
     MTRDeviceControllerFactory * _factory;
     NSMapTable * _nodeIDToDeviceMap;
-    os_unfair_lock _deviceMapLock; // protects nodeIDToDeviceMap
+    os_unfair_lock_t _deviceMapLock; // protects nodeIDToDeviceMap
```
jtung-apple, line 123: This line is not needed because there is already an underlying lock. The property is just a convenience to access the address.

kiel-apple: great catch - removed.

PR #35063, opened 2024-08-19
minimal_mdns: Fix filter for Compressed Fabric Id when browsing operational nodes

wqx6:
The DNS-SD filter of CompressedFabricId type cannot work when browsing operational nodes for minimal-mdns.

Add the logic to generate qname for that filter type and add _I option for operational browse shell command to test it.


jmartinez-silabs (approved):

src/lib/shell/commands/Dns.cpp
```diff
@@ -214,13 +214,22 @@ bool ParseSubType(int argc, char ** argv, Dnssd::DiscoveryFilter & filter)
...
         return false;
     }
 
-    uint16_t code;
-    VerifyOrReturnError(ArgParser::ParseInt(subtype + 2, code), false);
-
+    uint64_t code = 0;
+    if (filterType == Dnssd::DiscoveryFilterType::kCompressedFabricId)
+    {
+        VerifyOrReturnError(ArgParser::ParseInt(subtype + 2, code, 16), false);
```
ksperling-apple, line 227: A check here that the subtype number is exactly 16 hex digits long would be good here.
Having it accidentally zero padded would probably be confusing.

ksperling-apple (approved):

src/lib/shell/commands/Dns.cpp
wqx6: Done

PR #35065, opened 2024-08-19
OpenThread: clear the previous srp host and services when connecting …

DejinChen:
To prevent duplicated srp services, clear the previous srp host and services when connecting to a new network.

### DIFF
--- a/src/platform/OpenThread/GenericNetworkCommissioningThreadDriver.cpp
+++ b/src/platform/OpenThread/GenericNetworkCommissioningThreadDriver.cpp
@@ -189,6 +189,24 @@ void GenericThreadDriver::ConnectNetwork(ByteSpan networkId, ConnectCallback * c
         status = Status::kUnknownError;
     }
 
+    if (status == Status::kSuccess && ThreadStackMgrImpl().IsThreadAttached())
+    {
+        Thread::OperationalDataset currentDataset;
+        if (ThreadStackMgrImpl().GetThreadProvision(currentDataset) == CHIP_NO_ERROR)
+        {
+            // Clear the previous srp host and services
+            if (!currentDataset.AsByteSpan().data_equal(mStagingNetwork.AsByteSpan()) &&
+                ThreadStackMgrImpl().ClearAllSrpHostAndServices() != CHIP_NO_ERROR)
+            {
+                status = Status::kUnknownError;
+            }
+        }
+        else
+        {
+            status = Status::kUnknownError;
+        }
+    }
+
     if (status == Status::kSuccess &&
         DeviceLayer::ThreadStackMgrImpl().AttachToThreadNetwork(mStagingNetwork, callback) != CHIP_NO_ERROR)
     {


bzbarsky-apple:
@Damian-Nordic

src/platform/OpenThread/GenericNetworkCommissioningThreadDriver.cpp
```diff
@@ -189,6 +189,24 @@ void GenericThreadDriver::ConnectNetwork(ByteSpan networkId, ConnectCallback * c
...
     }
 
+    if (status == Status::kSuccess && ThreadStackMgrImpl().IsThreadAttached())
+    {
+        Thread::OperationalDataset currentDataset;
+        if (ThreadStackMgrImpl().GetThreadProvision(currentDataset) == CHIP_NO_ERROR)
+        {
+            // Clear the previous srp host and services
+            if (!currentDataset.AsByteSpan().data_equal(mStagingNetwork.AsByteSpan()) &&
+                ThreadStackMgrImpl().ClearAllSrpHostAndServices() != CHIP_NO_ERROR)
```
Damian-Nordic, line 199: So this call will block the Matter thread for up to 2 seconds (waiting for the SRP deregistration) and while this might be OK while the device is being factory reset, I'm not sure if this is desired for a normal operation. `ClearAllSrpHostAndServices` could take a `bool shouldWait` parameter to be able to run this operation asynchronously and then listen to some `ChipEvent` to get notified of the completion and to continue the attach to the new network. Do you think it would make sense or you don't see an issue in blocking the thread?

Damian-Nordic: I agree that you have to wait for the clearing to have completed before switching the network but you don't have to block the thread for that. You can execute the remaining part of this function in an async callback. My concern is that when you block a thread, then tasks or events posted by external threads using `PlatformMgr().PostEvent()` or `ScheduleLambda()` are not consumed which may lead to running out of the events on constrained devices. It may lead to some unexpected behavior but I'm not sure how likely scenario this is so let me know if you accept this consequence.

Damian-Nordic: > I agree that you have to wait for the clearing to have completed before switching the network but you don't have to block the thread for that. You can execute the remaining part of this function in an async callback. My concern is that when you block a thread, then tasks or events posted by external threads using `PlatformMgr().PostEvent()` or `ScheduleLambda()` are not consumed which may lead to running out of the events on constrained devices. It may lead to some unexpected behavior but I'm not sure how likely scenario this is so let me know if you accept this risk.

DejinChen: Okay, use async callback instead.

src/platform/OpenThread/GenericNetworkCommissioningThreadDriver.cpp
```diff
@@ -189,6 +189,28 @@ void GenericThreadDriver::ConnectNetwork(ByteSpan networkId, ConnectCallback * c
...
+    if (status == Status::kSuccess && ThreadStackMgrImpl().IsThreadAttached())
+    {
+        Thread::OperationalDataset currentDataset;
+        if (ThreadStackMgrImpl().GetThreadProvision(currentDataset) == CHIP_NO_ERROR)
+        {
+            // Clear the previous srp host and services
+            if (!currentDataset.AsByteSpan().data_equal(mStagingNetwork.AsByteSpan()))
+            {
+                ThreadStackMgrImpl().ClearAllSrpHostAndServicesWithoutBlocking([this, callback]() -> CHIP_ERROR {
+                    return DeviceLayer::ThreadStackMgrImpl().AttachToThreadNetwork(mStagingNetwork, callback);
```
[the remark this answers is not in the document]
DejinChen: For point 2, i tested the change on ESP32 platform and it worked. But the `OnSrpClientNotification` does execute in the OpenThread thread. For point 3, the switching operation does require a timeout. So i'll revert the async commit.

Damian-Nordic (approved):

andy31415 (approved):

PR #35075, opened 2024-08-19
Fix SkipArea logic and removed the use of memcpy

### LINKED ISSUE #34858: ServiceArea server: use CopyCharSpanToMutableCharSpan
Recommend using CopyCharSpanToMutableCharSpan here - it already safely implements the sizing.

_Originally posted by @cecille in https://github.com/project-chip/connectedhomeip/pull/34738#discussion_r1707849404_

### LINKED ISSUE #34934: service-area-server.cpp remove check for CurrentArea == NULL when executing skip command
This is no longer a cluster requirement, the check should not be in the common SDK code:

void Instance::HandleSkipCurrentAreaCmd  change to void Instance::HandleSkipAreaCmd 

Delete the following check:

    // If the CurrentArea attribute is null, the status should be set to InvalidInMode.
    // If the Status field is not set to Success, or InvalidAreaList, the StatusText field SHALL include a vendor defined error
    // description.
    if (mCurrentArea.IsNull())
    {
        exitResponse(SkipAreaStatus::kInvalidInMode, "Current Area attribute is null"_span);
        return;
    }

hicklin:
Fixes #34934 : Removed a check in the `SkipArea` server logic that is no longer required.

Fixes #34858 : In the service-area-cluster-objects, replaced the use or `memcpy` with `CopyCharSpanToMutableCharSpan`.


cecille (approved):

rbultman (approved):

kiel-apple (approved):

src/app/clusters/service-area-server/service-area-cluster-objects.h
```diff
@@ -110,9 +110,9 @@ struct AreaStructureWrapper : public chip::app::Clusters::ServiceArea::Structs::
     {
         areaDesc.locationInfo.SetNonNull();
         // Copy the name
-        auto sizeToCopy = std::min(sizeof(mAreaNameBuffer), locationName.size());
-        memcpy(mAreaNameBuffer, locationName.data(), sizeToCopy);
-        areaDesc.locationInfo.Value().locationName = CharSpan(mAreaNameBuffer, sizeToCopy);
+        auto areaNameSpan = MutableCharSpan(mAreaNameBuffer, kAreaNameMaxSize);
+        CopyCharSpanToMutableCharSpan(locationName, areaNameSpan);
+        areaDesc.locationInfo.Value().locationName = CharSpan(areaNameSpan.data(), areaNameSpan.size());
```
bzbarsky-apple, line 115: For what it's worth, that last line could just have been:
```
areaDesc.locationInfo.Value().locationName = areaNameSpan;
```

src/app/clusters/service-area-server/service-area-cluster-objects.h
```diff
@@ -320,24 +320,10 @@ struct MapStructureWrapper : public chip::app::Clusters::ServiceArea::Structs::M
...
-        }
-        else
-        {
-            // Save full name.
-            memcpy(mMapNameBuffer, aMapName.data(), aMapName.size());
-            name = CharSpan(mMapNameBuffer, aMapName.size());
-        }
+        mapID            = aMapId;
+        auto mapNameSpan = MutableCharSpan(mMapNameBuffer, kMapNameMaxSize);
+        CopyCharSpanToMutableCharSpan(aMapName, mapNameSpan);
```
bzbarsky-apple, line 325: So this has different behavior from the old code, right?  The old code copied as much as fit.  `CopyCharSpanToMutableCharSpan` will copy nothing in that case, return error, and `mapNameSpan` will be holding random data....

I think the code above also has that problem.

hicklin: Thanks for spotting this @bzbarsky-apple. I did not check the implementation of this method. @cecille, ideally this method does not error as otherwise setting these values by simply using consecutive `.Set` methods would be cumbersome. Hence, I think we should revert to the old implementation.

PR #35076, opened 2024-08-19
Add plumbing that allows fabric-bridge endpoints to report CADMIN attr

tehampson:
This overrides the CADMIN's server cluster implementation of the AttributeAccessInterface inside the fabric-bridge in a way that allows use to redirect reads on bridged endpoint to reflect state for the bridged CADMIN cluster.

TODO in a follow up PR:
* Determine a method for ensuring we have the latest attribute state of the of the cadmin cluster of the device we are bridging


examples/fabric-bridge-app/fabric-bridge-common/src/BridgedAdministratorCommissioning.cpp
```diff
@@ -0,0 +1,86 @@
...
+#include <app/AttributeAccessInterfaceRegistry.h>
+
+using namespace chip;
+using namespace chip::app;
+using namespace chip::app::Clusters;
+using namespace chip::app::Clusters::AdministratorCommissioning;
+
+CHIP_ERROR BridgedAdministratorCommissioning::Init()
+{
+    // We expect to be initialized after all the embr pluggin cluster initialization have been called,
```
bzbarsky-apple, line 31: Please fix all the spelling issues here.   And the comment duplication.

examples/fabric-bridge-app/fabric-bridge-common/include/BridgedAdministratorCommissioning.h
```diff
@@ -0,0 +1,58 @@
...
+ * properly respond to a read attribute for bridged devices we are representing, we override the
+ * currently registered Attribute Interface such that we are first to receive any read attribute
+ * request on Administrator Commissioning Cluster, and if it is not an endpoint for a device we
+ * are a bridge for we redirect to the default cluster server implementation of Administrator
+ * Commissioning Cluster.
+ */
+class BridgedAdministratorCommissioning : public chip::app::AttributeAccessInterface
+{
+public:
+    // Register for the OperationalCredentials cluster on all endpoints.
```
yufengwangca, line 37: nit: AdministratorCommissioning

examples/fabric-bridge-app/fabric-bridge-common/include/BridgedAdministratorCommissioning.h
```diff
@@ -0,0 +1,58 @@
...
+ *    limitations under the License.
+ */
+
+#pragma once
+
+#include <app-common/zap-generated/cluster-objects.h>
+#include <app/AttributeAccessInterfaceRegistry.h>
+
+/**
+ * @brief CADMIN cluster implementation for handling attribute read of bridged device endpoints.
```
yufengwangca, line 24: We also need to handle write operation for this cluster later, right?

examples/fabric-bridge-app/fabric-bridge-common/src/BridgedAdministratorCommissioning.cpp
```diff
@@ -0,0 +1,86 @@
...
+    // including the AdministratorCommissioning. This expectation allows us to get and unregister the
+    // existing AccessAttributeInterface for AdministratorCommissioning and register ourselves so that
+    // we can first see if the ReadAttribute is for us.
+
+    // We expect initialization after all embr plugin clusters initialization. This allows us to unregister
+    // the existing AccessAttributeInterface for AdministratorCommissioning and register ourselves, ensuring
+    // we get the callback for reading attribute. If the read is not intended for a bridged device we will
+    // forward it to the original attribute interface that we are unregistering.
+    mOriginalAttributeInterface = AttributeAccessInterfaceRegistry::Instance().Get(0, AdministratorCommissioning::Id);
```
yufengwangca, line 40: use kRootEndpointId instead of 0

examples/fabric-bridge-app/fabric-bridge-common/src/BridgedAdministratorCommissioning.cpp
```diff
@@ -0,0 +1,86 @@
...
+}
+
+CHIP_ERROR BridgedAdministratorCommissioning::Read(const ConcreteReadAttributePath & aPath, AttributeValueEncoder & aEncoder)
+{
+    VerifyOrDie(aPath.mClusterId == Clusters::AdministratorCommissioning::Id);
+    EndpointId endpointId  = aPath.mEndpointId;
+    BridgedDevice * device = BridgeDeviceMgr().GetDevice(endpointId);
+
+    if (!device)
+    {
```
yufengwangca, line 54: if the device is not available on this endpoint, we just need to pass through to go to the default cluster implementation, right? why we need to call mOriginalAttributeInterface->Read(aPath, aEncoder);

tehampson: mOriginalAttributeInterface is the original default cluster implementation

examples/fabric-bridge-app/fabric-bridge-common/src/BridgedAdministratorCommissioning.cpp
tehampson: Ops the second one is my attempt after doing a self grammar and spelling check. I forgot to delete the top version

examples/fabric-bridge-app/fabric-bridge-common/include/BridgedAdministratorCommissioning.h
tehampson: write is a bypass, but I updated to say "attribute interactions" instead of read

examples/fabric-bridge-app/fabric-bridge-common/include/BridgedAdministratorCommissioning.h
```diff
@@ -0,0 +1,58 @@
...
+    BridgedAdministratorCommissioning() :
+        AttributeAccessInterface(chip::NullOptional, chip::app::Clusters::AdministratorCommissioning::Id)
+    {}
+
+    CHIP_ERROR Init();
+
+    CHIP_ERROR Read(const chip::app::ConcreteReadAttributePath & aPath, chip::app::AttributeValueEncoder & aEncoder) override;
+
+    // We do not allow writing an attributes of a bridged device endpoint. We simply redirect write request to
+    // original attribut interface.
```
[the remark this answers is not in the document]
tehampson: Hopefully better now

examples/fabric-bridge-app/fabric-bridge-common/src/BridgedAdministratorCommissioning.cpp
```diff
@@ -0,0 +1,81 @@
...
+using namespace chip::app::Clusters;
+using namespace chip::app::Clusters::AdministratorCommissioning;
+
+CHIP_ERROR BridgedAdministratorCommissioning::Init()
+{
+    // We expect initialization after all embr plugin clusters initialization. This allows us to unregister
+    // the existing AccessAttributeInterface for AdministratorCommissioning and register ourselves, ensuring
+    // we get the callback for reading attribute. If the read is not intended for a bridged device we will
+    // forward it to the original attribute interface that we are unregistering.
+    mOriginalAttributeInterface = AttributeAccessInterfaceRegistry::Instance().Get(chip::kRootEndpointId, AdministratorCommissioning::Id);
```
yufengwangca, line 35: remove chip::

yufengwangca (approved):

examples/fabric-bridge-app/fabric-bridge-common/src/BridgedAdministratorCommissioning.cpp
yufengwangca: Actually, we don't need to use mOriginalAttributeInterface to access the default implementation. From the AttributeAccessInterface comment, we just need to return success without attempting to encode any data using aEncoder

Return success and not attempt to encode any data using aEncoder.  In
     *    this case, Ember attribute access will happen for the read. This may
     *    involve reading from the attribute store or external attribute
     *    callbacks.

tehampson: embr attribute access doesn't work for this cluster

https://github.com/project-chip/connectedhomeip/blob/master/src/app/clusters/administrator-commissioning-server/administrator-commissioning-server.cpp

This cluster relies on AttributeAccessInterface and not on embr to make sure the attributes are properly read back to client

yufengwangca: If we return without touching aEncoder, it should get to the default implementation at https://github.com/project-chip/connectedhomeip/blob/master/src/app/clusters/administrator-commissioning-server/administrator-commissioning-server.cpp

Lets land this as is, I will check why it does not work as the interface described when I got a chance later

saurabhst (approved):

examples/fabric-bridge-app/fabric-bridge-common/src/BridgedAdministratorCommissioning.cpp
```diff
@@ -0,0 +1,81 @@
...
+#include <app/AttributeAccessInterfaceRegistry.h>
+
+using namespace chip;
+using namespace chip::app;
+using namespace chip::app::Clusters;
+using namespace chip::app::Clusters::AdministratorCommissioning;
+
+CHIP_ERROR BridgedAdministratorCommissioning::Init()
+{
+    // We expect initialization after all embr plugin clusters initialization. This allows us to unregister
```
bzbarsky-apple, line 31: @tehampson what is "embr"?

PR #35086, opened 2024-08-20
TC-IDM-10.2: Add check for MACL

cecille:
Test: unit tests. Note this has not been tested against the example
      app as it is still under development.

### DIFF
--- a/src/python_testing/TC_DeviceConformance.py
+++ b/src/python_testing/TC_DeviceConformance.py
@@ -50,6 +50,22 @@ class DeviceConformanceTests(BasicCompositionTests):
         self.xml_device_types, problems = build_xml_device_types()
         self.problems.extend(problems)
 
+    def _get_device_type_id(self, device_type_name: str) -> int:
+        id = [id for id, dt in self.xml_device_types.items() if dt.name.lower() == device_type_name.lower()]
+        if len(id) != 1:
+            self.fail_current_test(f"Unable to find {device_type_name} device type")
+        return id[0]
+
+    def _has_nim(self):
+        nim_id = self._get_device_type_id('network infrastructure manager')
+        for endpoint in self.endpoints_tlv.values():
+            desc = Clusters.Descriptor
+            device_types = [dt.deviceType for dt in endpoint[desc.id][desc.Attributes.DeviceTypeList.attribute_id]]
+            if nim_id in device_types:
+                # TODO: it's unclear if this needs to be present on every endpoint. Right now, this assumes one is sufficient.
+                return True
+        return False
+
     def check_conformance(self, ignore_in_progress: bool, is_ci: bool):
         problems = []
         success = True
@@ -125,6 +141,13 @@ class DeviceConformanceTests(BasicCompositionTests):
                 for f in feature_masks:
                     location = AttributePathLocation(endpoint_id=endpoint_id, cluster_id=cluster_id,
                                                      attribute_id=GlobalAttributeIds.FEATURE_MAP_ID)
+                    if cluster_id == Clusters.AccessControl.id and f == Clusters.AccessControl.Bitmaps.Feature.kManagedDevice:
+                        # Managed ACL is treated as a special case because it is only allowed if other endpoints support NIM and disallowed otherwise.
+                        if not self._has_nim():
+                            record_error(
+                                location=location, problem="MACL feature is disallowed if the network infrastructure manager device type is not present")
+                        continue
+
                     if f not in self.xml_clusters[cluster_id].features.keys():
                         record_error(location=location, problem=f'Unknown feature with mask 0x{f:02x}')
                         continue
--- a/src/python_testing/TC_DeviceConformance.py
+++ b/src/python_testing/TC_DeviceConformance.py
@@ -145,7 +145,7 @@ class DeviceConformanceTests(BasicCompositionTests):
                         # Managed ACL is treated as a special case because it is only allowed if other endpoints support NIM and disallowed otherwise.
                         if not self._has_nim():
                             record_error(
-                                location=location, problem="MACL feature is disallowed if the network infrastructure manager device type is not present")
+                                location=location, problem="MACL feature is disallowed if the Network Infrastructure Manager device type is not present")
                         continue
 
                     if f not in self.xml_clusters[cluster_id].features.keys():
--- a/src/python_testing/TestConformanceTest.py
+++ b/src/python_testing/TestConformanceTest.py
@@ -1,0 +1,131 @@
+#
+#    Copyright (c) 2024 Project CHIP Authors
+#    All rights reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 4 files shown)


ksperling-apple (approved):

src/python_testing/TC_DeviceConformance.py
```diff
@@ -50,6 +50,22 @@ async def setup_class_helper(self):
...
         self.problems.extend(problems)
 
+    def _get_device_type_id(self, device_type_name: str) -> int:
+        id = [id for id, dt in self.xml_device_types.items() if dt.name.lower() == device_type_name.lower()]
+        if len(id) != 1:
+            self.fail_current_test(f"Unable to find {device_type_name} device type")
+        return id[0]
+
+    def _has_nim(self):
+        nim_id = self._get_device_type_id('network infrastructure manager')
```
tcarmelveilleux, line 60: For `_get_device_type_id` handle case normalization? Would the call `_get_device_type_id('Network Infrastructure Manager')` also work?

src/python_testing/TestConformanceTest.py
```diff
@@ -0,0 +1,131 @@
...
+        root_endpoint[ac.id][ac.Attributes.FeatureMap.attribute_id] = ac.Bitmaps.Feature.kManagedDevice
+        root_endpoint[ac.id][ac.Attributes.Arl.attribute_id] = []
+        root_endpoint[ac.id][ac.Attributes.CommissioningARL.attribute_id] = []
+        root_endpoint[ac.id][ac.Attributes.AttributeList.attribute_id].extend([
+            ac.Attributes.Arl.attribute_id, ac.Attributes.CommissioningARL.attribute_id])
+        root_endpoint[ac.id][ac.Attributes.AcceptedCommandList.attribute_id].append(ac.Commands.ReviewFabricRestrictions.command_id)
+        root_endpoint[ac.id][ac.Attributes.GeneratedCommandList.attribute_id].append(
+            ac.Commands.ReviewFabricRestrictionsResponse.command_id)
+
+    @ async_test_body
```
tcarmelveilleux, line 100: It's odd that there is a space here between the `@` and `async_test_body`

src/python_testing/TC_DeviceConformance.py
```diff
@@ -50,6 +50,22 @@ async def setup_class_helper(self):
...
+        for endpoint in self.endpoints_tlv.values():
+            desc = Clusters.Descriptor
+            device_types = [dt.deviceType for dt in endpoint[desc.id][desc.Attributes.DeviceTypeList.attribute_id]]
+            if nim_id in device_types:
+                # TODO: it's unclear if this needs to be present on every endpoint. Right now, this assumes one is sufficient.
```
tcarmelveilleux, line 65: Presence of only 1 is OK, since it's an application device type.

tcarmelveilleux (approved):

src/python_testing/TC_DeviceConformance.py
cecille: yup

src/python_testing/TC_DeviceConformance.py
cecille: The question was around multiple endpoints. So one endpoint with NIM one with ex. lightbulb. I'm not sure how the spec text is going to land around that condition.

src/python_testing/TestConformanceTest.py
cecille, line 100:
```suggestion
    @async_test_body
```

cecille: yeah, that is odd. Fixed.

src/python_testing/TC_DeviceConformance.py
ReneJosefsen: Given ACL is node scoped and the fact that we do not want to allow MACL for other device types than NIM, I would say that we have a requirement that NIM is not allowed to have other EP's with other device types 🤔

tcarmelveilleux: Let's not throw the baby out with the bathwater. It makes no sense saying there can only be a single endpoint and it has to be NIM.

Maybe we need to update ACL TP instead and work an update on the feature conformance to add steps that it is only present if at least one endpoint is a device type that has it as O at least.

cecille: This does this currently - allowed if at least one endpoint is marked as being a NIM. It does mean you can have a weird combo device, but you're probably right that it's unlikely to happen in practice.

src/python_testing/TestConformanceTest.py
```diff
@@ -0,0 +1,131 @@
...
+        self.problems.extend(problems)
+        asserts.assert_true(success, "Unexpected failure with NIM and MACL")
+
+        # A MACL is not allowed when there is no NIM
+        self.endpoints_tlv[1] = self._create_minimal_dt(device_type_id=on_off_id)
+        success, problems = self.check_conformance(ignore_in_progress=False, is_ci=False)
+        self.problems.extend(problems)
+        asserts.assert_false(success, "Unexpected success with On/Off and MACL")
+
+        # TODO: what happens if there is a NIM and a non-NIM endpoint?
```
chrisdecenzo, line 127: That is allowed. The ARL entries would only be allowed on the clusters of the NIM endpoint

src/python_testing/TC_DeviceConformance.py
```diff
@@ -125,6 +141,13 @@ def record_warning(location, problem):
                 for f in feature_masks:
                     location = AttributePathLocation(endpoint_id=endpoint_id, cluster_id=cluster_id,
                                                      attribute_id=GlobalAttributeIds.FEATURE_MAP_ID)
+                    if cluster_id == Clusters.AccessControl.id and f == Clusters.AccessControl.Bitmaps.Feature.kManagedDevice:
+                        # Managed ACL is treated as a special case because it is only allowed if other endpoints support NIM and disallowed otherwise.
+                        if not self._has_nim():
```
chrisdecenzo, line 146: Given that we will likely add to this list in the future, shouldn't this be self._has_allowed_MACL()?

Do we have hard coding of the device library rules in the other scripts?

src/python_testing/TC_DeviceConformance.py
cecille: OK, I'm going to resolve this for now so I can get this code in. I've opened a spec issue here: https://github.com/CHIP-Specifications/connectedhomeip-spec/issues/10294

chrisdecenzo (changes requested):
Can we do this in a more generic way then hardcoding NIM? Seems like we should be checking against a list of device types that are allowed to use this feature.

src/python_testing/TestConformanceTest.py
ReneJosefsen: Where in the spec is this stated? The ACL cluster is node scopes so it applies to all the EP's on the device and I have not found any such restrictions mentioned in the spec?

PR #35088, opened 2024-08-20
Enhance chip-tool's formatting for some IDs

XToripuru:
### Issue

Chip-tool's formatting of ServerList, ClientList, AttributeList, AcceptedCommandList, GeneratedCommandList show only IDs which makes the tool more difficult to use.

Current output:
```cpp
AttributeList: 7 entries
  [1]: 0
  [2]: 1
  [3]: 65528
  [4]: 65529
  [5]: 65531
  [6]: 65532
  [7]: 65533
```

### Proposed Solution

Add simple logging functions which would additionally display the name for these cases.

### Testing

Local test suite passed while also tested with this and many other commands:
```
./chip-tool switch read attribute-list 0xF00 0
```

Output after changes:
```cpp
AttributeList: 7 entries
  [1]: 0 (NumberOfPositions)
  [2]: 1 (CurrentPosition)
  [3]: 65528 (GeneratedCommandList)
  [4]: 65529 (AcceptedCommandList)
  [5]: 65531 (AttributeList)
  [6]: 65532 (FeatureMap)
  [7]: 65533 (ClusterRevision)
```


examples/chip-tool/templates/logging/EntryToText-src.zapt
```diff
@@ -0,0 +1,61 @@
...
+{{/zcl_clusters}}
+		default: return "Unknown";
+	}
+}
+
+char const * CommandIdToText(chip::ClusterId cluster, chip::CommandId id) {
+	switch(cluster)
+	{
+{{#zcl_clusters}}
+{{#zcl_commands_source_client}}
```
bzbarsky-apple, line 44: But we're using this for both client-source and server-source commands.

That is, I think this setup (which is otherwise really nice!) produces incorrect output for GeneratedCommandList.

examples/chip-tool/templates/logging/EntryToText.zapt
```diff
@@ -0,0 +1,11 @@
...
+
+#include <app-common/zap-generated/ids/Attributes.h>
+#include <app-common/zap-generated/ids/Clusters.h>
+#include <app-common/zap-generated/ids/Commands.h>
+
+char const * ClusterIdToText(chip::ClusterId id);
+
+char const * AttributeIdToText(chip::ClusterId cluster, chip::AttributeId id);
+
+char const * CommandIdToText(chip::ClusterId cluster, chip::CommandId id);
```
bzbarsky-apple, line 11: I think we need separate AcceptedCommandIdToText and GeneratedCommandIdToText functions, probably.

XToripuru: Done.

bzbarsky-apple (approved):
Thank you!

arkq (approved):

PR #35093, opened 2024-08-20
Yaml scripts update

Harshith-GRL:
This PR has the Following Changes

1. FIX for Issues https://github.com/project-chip/matter-test-scripts/issues/333 Removed Test cases TC-SC-4.5, 4.10 as test plan was updated.
2. Fix for Issue https://github.com/project-chip/matter-test-scripts/issues/317 Test step numbers are not mentioned correctly, now they are as per Test plan.


yinyi-hu (approved):

src/app/tests/suites/certification/Test_TC_ACT_3_2.yaml
```diff
@@ -33,7 +33,7 @@ tests:
           "Step 0: Preparation: TH as server exposes an Actions server cluster
           on EP 1, with one action (supporting all possible commands) and
           corresponding ActionLists and EndpointLists attributes: ActionList:
-          contains one list element (ActionListStruct):  ActionIO = 0x1001  Name
+          contains one list element (ActionListStruct):  ActionIO = 0xA001  Name
```
cjandhyala, line 36: Can we just update test plan to use 0x1001 instead of  0xA001 , so that we don't need to update script and reference app.

bzbarsky-apple (approved):

src/app/tests/suites/certification/Test_TC_ACT_3_2.yaml
Harshith-GRL: script is reverted back to use action 0x1001

cjandhyala (approved):

yinyi-hu (approved):

raju-apple (approved):
@Harshith-GRL have SC 4-10 and SC 4-5 been removed from the Test Plan too ?

cjandhyala:
> @Harshith-GRL have SC 4-10 and SC 4-5 been removed from the Test Plan too ?

yes. they are already removed from the test plan

jmartinez-silabs (approved):

src/app/tests/suites/certification/Test_TC_ACT_3_2.yaml
Emill: Seems like the revert itself got reverted due to a bad merge?

PR #35098, opened 2024-08-20
IDM-12.1: Also dump to log

cecille:
I was a bit to optimistic about getting these files submitted separately. It would still be great to get the files, but as a backstop, I'm printing these into the logs as structured data so we can extract them if necessary from the output logs.

### DIFF
--- a/src/python_testing/TC_DeviceBasicComposition.py
+++ b/src/python_testing/TC_DeviceBasicComposition.py
@@ -775,7 +775,17 @@ class TC_DeviceBasicComposition(MatterBaseTest, BasicCompositionTests):
         software_version = self.endpoints[0][Clusters.BasicInformation][Clusters.BasicInformation.Attributes.SoftwareVersion]
         filename = f'device_dump_0x{vid:04X}_0x{pid:04X}_{software_version}.json'
         dump_device_composition_path = self.user_params.get("dump_device_composition_path", filename)
-        self.dump_wildcard(dump_device_composition_path)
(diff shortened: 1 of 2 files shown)


harimau-qirex (approved):

jmartinez-silabs (approved):

src/python_testing/TC_DeviceBasicComposition.py
```diff
@@ -775,7 +775,17 @@ def test_TC_IDM_12_1(self):
...
         filename = f'device_dump_0x{vid:04X}_0x{pid:04X}_{software_version}.json'
         dump_device_composition_path = self.user_params.get("dump_device_composition_path", filename)
-        self.dump_wildcard(dump_device_composition_path)
+        json_str, txt_str = self.dump_wildcard(dump_device_composition_path)
+
+        # Structured dump so we can pull these back out of the logs
+        def log_structured_data(start_tag: str, dump_string):
+            lines = dump_string.splitlines(keepends=True)
+            logging.info(f'{start_tag}BEGIN ({len(lines)} lines)====')
+            logging.info(f'{start_tag}{start_tag.join(lines)}')
```
tcarmelveilleux, line 784: There is double start-tag here which is a bit confusing.

I have an alternative:

```
[logging.info(f'{start_tag}{line}' for line in dump_string_splitlines()]
```

tcarmelveilleux (approved):

ksperling-apple (approved):

PR #35099, opened 2024-08-20
Add expiry to KeepActive as per spec into fabric-admin example

tehampson:
Spec states that KeepActive request should be dropped if the ICD device has not checked-in for over an hour. If a FS device wants to be aware the moment ICD wakes up it should kick the KeepActive Command at least once per hour

### DIFF
--- a/examples/fabric-admin/rpc/RpcServer.cpp
+++ b/examples/fabric-admin/rpc/RpcServer.cpp
@@ -47,6 +47,8 @@ class FabricAdmin final : public rpc::FabricAdmin, public IcdManager::Delegate
 public:
     void OnCheckInCompleted(const chip::app::ICDClientInfo & clientInfo) override
     {
+        // Needs for accessing mPendingCheckIn
+        assertChipStackLockedByCurrentThread();
         NodeId nodeId = clientInfo.peer_node.GetNodeId();
         auto it       = mPendingCheckIn.find(nodeId);
         VerifyOrReturn(it != mPendingCheckIn.end());
@@ -155,6 +157,8 @@ public:
 
     void ScheduleSendingKeepActiveOnCheckIn(chip::NodeId nodeId, uint32_t stayActiveDurationMs)
     {
+        // Needs for accessing mPendingCheckIn
+        assertChipStackLockedByCurrentThread();
 
         auto timeNow = System::SystemClock().GetMonotonicTimestamp();
         // Spec says we should expire the request 60 mins after we get it
--- a/examples/fabric-admin/rpc/RpcServer.cpp
+++ b/examples/fabric-admin/rpc/RpcServer.cpp
@@ -56,8 +56,10 @@ public:
         System::Clock::Milliseconds64 timeNowMs =
             std::chrono::duration_cast<System::Clock::Milliseconds64>(System::SystemClock().GetMonotonicMilliseconds64());
 
-        if (timeNowMs > checkInData.mRequestExpiresAtMs) {
-            ChipLogError(NotSpecified, "ICD check-in for device we have been waiting, came after KeepActive expiry. Reqeust dropped");
+        if (timeNowMs > checkInData.mRequestExpiresAtMs)
+        {
+            ChipLogError(NotSpecified,
+                         "ICD check-in for device we have been waiting, came after KeepActive expiry. Reqeust dropped");
             return;
         }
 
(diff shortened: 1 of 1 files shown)


examples/fabric-admin/rpc/RpcServer.cpp
```diff
@@ -48,19 +48,27 @@ class FabricAdmin final : public rpc::FabricAdmin, public IcdManager::Delegate
...
+        auto it             = mPendingCheckIn.find(nodeId);
+        VerifyOrReturn(it != mPendingCheckIn.end());
+
+        KeepActiveDataForCheckIn checkInData = it->second;
+        // Removed from pending map as check-in from this node has occured and we will handle the pending KeepActive
+        // request.
+        mPendingCheckIn.erase(nodeId);
+
+        auto timeNow = System::SystemClock().GetMonotonicTimestamp();
+        if (timeNow > checkInData.mRequestExpiryTimestamp)
```
cecille, line 60: Why is this kept around until the checkin? It seems likely that if the device hasn't checked in by the expiry time that it just is never going to and you'll have lots of dead entries in this map.

examples/fabric-admin/rpc/RpcServer.cpp
```diff
@@ -167,8 +187,8 @@ class FabricAdmin final : public rpc::FabricAdmin, public IcdManager::Delegate
         chip::Platform::Delete(data);
     }
 
-    // Modifications to mPendingKeepActiveTimesMs should be done on the MatterEventLoop thread
-    std::map<chip::NodeId, uint32_t> mPendingKeepActiveTimesMs;
+    // Modifications to mPendingCheckIn should be done on the MatterEventLoop thread
```
cecille, line 190: why?

examples/fabric-admin/rpc/RpcServer.cpp
yufengwangca: This design use the absolute timestamp to compare, if we want it self deleted, then we need to start a timer for each of them.

examples/fabric-admin/rpc/RpcServer.cpp
```diff
@@ -48,19 +48,27 @@ class FabricAdmin final : public rpc::FabricAdmin, public IcdManager::Delegate
...
+        KeepActiveDataForCheckIn checkInData = it->second;
+        // Removed from pending map as check-in from this node has occured and we will handle the pending KeepActive
+        // request.
+        mPendingCheckIn.erase(nodeId);
+
+        auto timeNow = System::SystemClock().GetMonotonicTimestamp();
+        if (timeNow > checkInData.mRequestExpiryTimestamp)
+        {
+            ChipLogError(NotSpecified,
+                         "ICD check-in for device we have been waiting, came after KeepActive expiry. Reqeust dropped");
```
yufengwangca, line 63: log explicitly what has been dropped

examples/fabric-admin/rpc/RpcServer.cpp
```diff
@@ -167,8 +187,8 @@ class FabricAdmin final : public rpc::FabricAdmin, public IcdManager::Delegate
         chip::Platform::Delete(data);
     }
 
-    // Modifications to mPendingKeepActiveTimesMs should be done on the MatterEventLoop thread
-    std::map<chip::NodeId, uint32_t> mPendingKeepActiveTimesMs;
+    // Modifications to mPendingCheckIn should be done on the MatterEventLoop thread
+    std::map<chip::NodeId, KeepActiveDataForCheckIn> mPendingCheckIn;
```
yufengwangca, line 191: use std::unordered_map to be more efficient.

yufengwangca, line 191: remove chip::

examples/fabric-admin/rpc/RpcServer.cpp
tehampson: You will have up to 1 entry for each node you are a bridge for. For clients this is trivial amounts of memory

ICD device can take hours to check-in, eventually we expect these to expire. If we are really concerned we can add clean up code later, but I don't think it is worth the development time.

examples/fabric-admin/rpc/RpcServer.cpp
tehampson: otherwise we would need a mutex protecting this data. Added a comment

cecille: ah, makes sense. OK, thanks.

cecille: Actually, is there an assertion for this? It's like assert check stack locked or something?

cecille (approved):

examples/fabric-admin/rpc/RpcServer.cpp
tehampson: I'll add that in the check, but I think the comment helps communicate why

yufengwangca (approved):

saurabhst (approved):

PR #35102, opened 2024-08-20
Add configurable macros for min/max address resolve lookup time

s-mcclain:
There is no way currently to configure the amount of time a device should wait at a minimum / maximum during address lookup for devices, as a NodeLookupRequest is handled internally within the Matter stack core, and does not expose any way to have the Matter stack core call `SetMinLookupTime()` nor `SetMaxLookupTime()`.

To enable a more configurable way to adjust these hardcoded times, we can expose these min/max times as macros within the CHIPConfig.h, allowing for these defaults to be overridden if desired.

### DIFF
--- a/src/lib/address_resolve/AddressResolve.h
+++ b/src/lib/address_resolve/AddressResolve.h
@@ -143,8 +143,14 @@ public:
     }
 
 private:
-    static constexpr uint32_t kMinLookupTimeMsDefault = 200;
-    static constexpr uint32_t kMaxLookupTimeMsDefault = 45000;
+    static_assert((CHIP_CONFIG_ADDRESS_RESOLVE_MIN_LOOKUP_TIME_MS) <= (CHIP_CONFIG_ADDRESS_RESOLVE_MAX_LOOKUP_TIME_MS),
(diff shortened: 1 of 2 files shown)


s-mcclain:
Tested by building chip-tool with CHIP_CONFIG_ADDRESS_RESOLVE_MIN_LOOKUP_TIME_MS=300 and CHIP_CONFIG_ADDRESS_RESOLVE_MAX_LOOKUP_TIME_MS=30000

Executed the following command (no accessory is active):
```
./chip-tool interactive start
>>> onoff toggle 5 1
...
[1724262804.062] [88356:3095624] [DIS] Checking node lookup status for BB99B250D47BB0CA-0000000000000005 after 301 ms
...
[1724263017.107] [88828:3098818] [DIS] Checking node lookup status for BB99B250D47BB0CA-0000000000000005 after 30001 ms
```

The above shows that the minimum time was correctly set to custom default of 300ms and the maximum time was correctly set custom default of 30000ms.

chrisdecenzo (approved):

jmartinez-silabs (approved):

PR #35109, opened 2024-08-21
TC-IDM-XX: try both PASE and CASE for basic comp tests.

cecille:
These tests are disruptive at cert because they could run over PASE or CASE and defaulted to PASE. This meant that they couldn't be run in sequence with other tests that require commissioning. Because the PASE setup required a qr or manual code, it also meant that these needed special parameters to run.

Now:
- can use discriminator and passcode for PASE connection
- device tries both PASE and CASE and runs over whichever works first.


jmartinez-silabs (approved):

src/controller/python/chip/ChipDeviceCtrl.py
```diff
@@ -1711,6 +1711,16 @@ def InitGroupTestingData(self):
                 self.devCtrl)
         ).raise_on_error()
 
+    def CreateManualCode(self, discriminator: int, passcode: int) -> str:
+        """ Creates a standard flow manual code from the given discriminator and passcode."""
+        # 64 bytes is WAY more than required, but let's be safe
+        size = 64
+        buf = create_string_buffer(size)
+        self._ChipStack.Call(
+            lambda: self._dmLib.pychip_CreateManualCode(discriminator, passcode, buf, size)
```
tcarmelveilleux, line 1720: You won't know about how many charactres are in teh code, since there is no output size or null termination. Suggest having an output argument for final size

src/python_testing/basic_composition_support.py
```diff
@@ -98,12 +98,15 @@ def ConvertValue(value) -> Any:
...
-    async def connect_over_pase(self, dev_ctrl):
-        asserts.assert_true(self.matter_test_config.qr_code_content == [] or self.matter_test_config.manual_code == [],
-                            "Cannot have both QR and manual code specified")
-        setupCode = self.matter_test_config.qr_code_content + self.matter_test_config.manual_code
-        asserts.assert_equal(len(setupCode), 1, "Require one of either --qr-code or --manual-code.")
-        await dev_ctrl.FindOrEstablishPASESession(setupCode[0], self.dut_node_id)
+    def get_code(self, dev_ctrl):
+        created_codes = []
+        for idx, d in enumerate(self.matter_test_config.discriminators):
+            created_codes.append(dev_ctrl.CreateManualCode(d, self.matter_test_config.setup_passcodes[idx]))
```
tcarmelveilleux, line 104:
```suggestion
        for idx, discriminator in enumerate(self.matter_test_config.discriminators):
            created_codes.append(dev_ctrl.CreateManualCode(discriminator, self.matter_test_config.setup_passcodes[idx]))
```

src/python_testing/basic_composition_support.py
```diff
@@ -98,12 +98,15 @@ def ConvertValue(value) -> Any:
...
+
+        setup_codes = self.matter_test_config.qr_code_content + self.matter_test_config.manual_code + created_codes
```
tcarmelveilleux, line 106: `manual_code` would have to be a list, no?

If a manual code is actually passed, and it's not a list, then it's possible adding list + str + list will fail.

tcarmelveilleux (approved):

src/python_testing/basic_composition_support.py
cecille: yup - it's a list. See https://github.com/project-chip/connectedhomeip/blob/master/src/python_testing/matter_testing_support.py#L640. Was changed when we did the DA test that require multiple devices. 

Manual code passing in tested in one of the CI runs.

src/controller/python/chip/ChipDeviceCtrl.py
cecille: added.

PR #35112, opened 2024-08-21
Make descriptions for MTRDevice clearly say whether it's the XPC version.

bzbarsky-apple:
This requires hoisting _nodeID and _deviceController ivars clearly into the MTRDevice superclass, so they can be accessed from subclasses.

The XPC version does not have a bunch of the state the non-XPC one does, so for now it does not try to log that

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -173,23 +173,6 @@ bool HaveSubscriptionEstablishedRightNow(MTRInternalDeviceState state)
     return state == MTRInternalDeviceStateInitialSubscriptionEstablished || state == MTRInternalDeviceStateLaterSubscriptionEstablished;
 }
 
-NSString * InternalDeviceStateString(MTRInternalDeviceState state)
-{
-    switch (state) {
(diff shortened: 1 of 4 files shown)


src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -491,6 +474,8 @@ - (instancetype)initForSubclassesWithNodeID:(NSNumber *)nodeID controller:(MTRDe
     if (self = [super init]) {
         _lock = OS_UNFAIR_LOCK_INIT;
         _delegates = [NSMutableSet set];
+        _deviceController = controller;
```
jtung-apple, line 477: This part already added in https://github.com/project-chip/connectedhomeip/pull/35110/files

jtung-apple (approved):

src/darwin/Framework/CHIP/MTRDevice.mm
bzbarsky-apple: Yes, I purposefully made this one exactly like that one, so it will merge cleanly either direction.

bzbarsky-apple:

PR #35116, opened 2024-08-21
[Fabric-Sync] Add VID/PID fields to IPC method CommissionNode

yufengwangca:
We need to pass VID/PID fields extracted from RequestCommissioningApproval to FabricAdmin as part of CommissionNode IPC.

The receiving FabricAdmin could use these values to validate against the actual VID/PID of the device to be commissioned .


arkq (approved):

tehampson (changes requested):
I don't think we are getting what we need here... I want to just take a little bit of time to confirm a few things. I think by providing the VID you are filtering out discovery by vendor ID and that is why you are able to pass the test I think if you provide an invalid PID you will fail the test again

tehampson (approved):

PR #35122, opened 2024-08-21
[Chef] BasicVideoPlayer: Fix MediaInput and MediaPlayback not reporting attribute changes due to using AttributeAccessInterface

### LINKED ISSUE #34897: [BUG] Some device clusters not reporting attribute changes (by calling `MatterReportingAttributeChangeCallback`)  when the invoked command is calling `app::AttributeAccessInterface`
### Reproduction steps

Currently there are many clusters using AccessInterface to delegate the attributes in the application layer. (e.g.  `MediaPlayback` and `MediaInput` clusters . When the commands of these clusters are called, the attributes subscriber won't be able to receive the `reportData`.  

### Reproducing steps: 

e.g.

1. running the a video player e.g. `rootnode_basicvideoplayer_0ff86e943b`

2. running 1st chip-tool to commission it and sharing it to the 2nd chip-tool 

3. After the 2nd chip-tool successfully commissioned the device, subscribe the attribute change by interactive mode
```
$ ./out/chip-tool/chip-tool interactive start

>>> mediaplayback subscribe current-state 0 100 0x11 1

```

4. Then use 1st to invoke the `Play` or `Pause`  

### Expected results

The 2nd chip-tool as the attribute subscriber should get notified 

### Actual results

The 2nd chip-tool is not notified

### Analysis 

The current `MediaPlayback` simply calls the delegate to modify the attribute data but not calling `MatterReportingAttributeChangeCallback`  [media-playback-server.cpp](https://github.com/project-chip/connectedhomeip/blob/master/src/app/clusters/media-playback-server/media-playback-server.cpp#L311-L333)
```
bool emberAfMediaPlaybackClusterPauseCallback(app::CommandHandler * command, const app::ConcreteCommandPath & commandPath,
                                              const Commands::Pause::DecodableType & commandData)
{
...
    Delegate * delegate = GetDelegate(endpoint);
    VerifyOrExit(isDelegateNull(delegate, endpoint) != true, err = CHIP_ERROR_INCORRECT_STATE);

    {
        delegate->HandlePause(responder);
    }
exit:
    if (err != CHIP_NO_ERROR)
    {
        ChipLogError(Zcl, "emberAfMediaPlaybackClusterPauseCallback error: %s", err.AsString());
        command->AddStatus(commandPath, Status::Failure);
    }

    return true;
}
```

When referring to correct implementation, it should be like the [door-lock-server.cpp](https://github.com/project-chip/connectedhomeip/blob/master/src/app/clusters/door-lock-server/door-lock-server.cpp#L4076-L4113) which calls `MatterReportingAttributeChangeCallback` after delegate successfully handles data changes.
```
void DoorLockServer::clearAliroReaderConfigCommandHandler(CommandHandler * commandObj, const ConcreteCommandPath & commandPath)
{
...

    uint8_t buffer[kAliroReaderVerificationKeySize];
    MutableByteSpan readerVerificationKey(buffer);
    CHIP_ERROR err = delegate->GetAliroReaderVerificationKey(readerVerificationKey);
...
    err = delegate->ClearAliroReaderConfig();
    if (err != CHIP_NO_ERROR)
    {
        ChipLogError(Zcl, "[SetAliroReaderConfig] Unable to set aliro reader config [endpointId=%d]", endpointID);
    }
    else
    {
        // Various attributes changed; mark them dirty.
        MatterReportingAttributeChangeCallback(endpointID, Clusters::DoorLock::Id, AliroReaderVerificationKey::Id);
        MatterReportingAttributeChangeCallback(endpointID, Clusters::DoorLock::Id, AliroReaderGroupIdentifier::Id);
        MatterReportingAttributeChangeCallback(endpointID, Clusters::DoorLock::Id, AliroGroupResolvingKey::Id);
    }
    sendClusterResponse(commandObj, commandPath, ClusterStatusCode(StatusIB(err).mStatus));
}
```

### Bug prevalence

Everytime

2b5194b

### Platform

core

### Platform Version(s)

_No response_

### Anything else?

_No response_

erwinpan1:
Fix #34897 

Calling `SelectInput` command won't trigger `reportData` in `MediaInput` cluster due to the changes made in application is not calling MatterReportingAttributeChangeCallback .

Similar issues also happens in `Play` `Pause` commands of `MediaPlayback` cluster


examples/chef/common/clusters/media-playback/MediaPlaybackManager.cpp
```diff
@@ -15,22 +15,41 @@
...
 
 using chip::CharSpan;
 using chip::app::AttributeValueEncoder;
 using chip::app::CommandResponseHelper;
+using chip::Protocols::InteractionModel::Status;
+
+MediaPlaybackManager::MediaPlaybackManager(chip::EndpointId endpoint) : mEndpoint(endpoint)
+{
+    // Sync the attributes from attribute storage
+    Status status = Attributes::CurrentState::Get(endpoint, &mCurrentState);
```
bzbarsky-apple, line 39: This is really quite odd.  Why are we storing the data in two places and keeping them in sync?  Why is the responsibility for doing that (or for triggering reporting in general) in the _application_ and not in the _cluster implementation_?

Ideally the way this would work, for simple data like this, is:

1. mCurrentState would live in the cluster implementation.
2. When the state needs to change, a setter on the cluster implementation would be called to set the state.
3. The cluster would then mark things dirty as needed.

tcarmelveilleux (approved):

examples/chef/common/clusters/media-playback/MediaPlaybackManager.cpp
erwinpan1: > Why are we storing the data in two places and keeping them in sync?
The `mCurrentState` is a variable in the RAM and could not be persistently stored, right? Considering if this is an application, you changed the variables, killed the process, and run the program again. You just reload the value from the storage instead of reset it, right?

> 1. mCurrentState would live in the cluster implementation.
I don't quite understand this. `mCurrentState` still live in the cluster. What I'm doing is just restoring the value from anywhere it stored in last change. The current example is restoring it from attribute storage.

> 2. When the state needs to change, a setter on the cluster implementation would be called to set the state.
> 3. The cluster would then mark things dirty as needed.
The cluster would then mark things dirty as needed.
 I assume the new function `MediaPlaybackManager::HandleSetCurrentState(...)` already fulfill this?

tcarmelveilleux: Let's remove mCurrentState and completely rely on the AttributeStore view if we add `Attributes::CurrentState::Get`/`Set` for attributes.

erwinpan1: > Let's remove mCurrentState and completely rely on the AttributeStore view if we add Attributes::CurrentState::Get/Set for attributes.

This seems making sense. My understanding is the Attribute access logics are delegated to the application, so the application could decide how to read/write/access the values. Reducing the variable seems can reduce potential risks of out-of-sync.

bzbarsky-apple: > The mCurrentState is a variable in the RAM and could not be persistently stored, right?

So is what you do with Attributes::Whatever::Set, by default, unless it's either marked as NVM in the ZAP config or marked external and the external attribute handler does persistence.

> My understanding is the Attribute access logics are delegated to the application

They might or might not be, depending on the ZAP configuration.

> mCurrentState still live in the cluster. 

No, it lives in the application.  The cluster is in `src/app/clusters/media-playback-server/media-playback-server.{h,cpp}`.

> I assume the new function MediaPlaybackManager::HandleSetCurrentState(...) already fulfill this?

This is part of the _application_, not part of the _cluster_.  So every application would have to reimplement this part.

Maybe that's acceptable, but it seems a little odd to me to be pushing this work off on applications when we initially could not get it right in even our own example application.

erwinpan1: > So is what you do with Attributes::Whatever::Set, by default, unless it's either marked as NVM in the ZAP config or marked external and the external attribute handler does persistence.

Yeah. I plan to modify all Chef samples to use RAM instead of NVM as much as possible in a batch. But I can modify the BasicVideoPlayer first

> They might or might not be, depending on the ZAP configuration.

You mean defining it as external in ZAP. But I mean implement the `AttributeAccessInterface` (or `AAI`) in `src/app/clusters` . e.g. [WriteSingleClusterData](https://github.com/project-chip/connectedhomeip/blob/dd8b002e7e65d3f6b1f1a95ef8a740607202ea38/src/app/util/ember-compatibility-functions.cpp#L723-L733) which is not depending on the ZAP file.  My understanding is this is delegating read/write logics to application (if application registering a AAI implementation to it), and the application has to handle storage and report data itself.  It might be a little similar to `setting Attribute access as external`.

```
    if (auto * attrOverride = AttributeAccessInterfaceRegistry::Instance().Get(aPath.mEndpointId, aPath.mClusterId))
    {
        AttributeValueDecoder valueDecoder(aReader, aSubjectDescriptor);
        ReturnErrorOnFailure(attrOverride->Write(aPath, valueDecoder));


        if (valueDecoder.TriedDecode())
        {
            MatterReportingAttributeChangeCallback(aPath);
            return apWriteHandler->AddStatus(aPath, Protocols::InteractionModel::Status::Success);
        }
    }
```

>>  mCurrentState still live in the cluster.

> No, it lives in the application. The cluster is in src/app/clusters/media-playback-server/media-playback-server.{h,cpp}.

Got it.

> Maybe that's acceptable, but it seems a little odd to me to be pushing this work off on applications when we initially could not get it right in even our own example application.

This is how the current `AAI` is designed, isn't it?

erwinpan1: @bzbarsky-apple I've modified the ZAP to put some attribute to be in NVM. [ef5a9b](https://github.com/project-chip/connectedhomeip/pull/35122/commits/ef5a9bff1c1b464319fc6e2b1a7e952f65583a74). Would you be able to  take a look ?

examples/chef/devices/rootnode_basicvideoplayer_0ff86e943b.zap
```diff
@@ -2444,7 +2444,7 @@
               "side": "server",
               "type": "boolean",
               "included": 1,
-              "storageOption": "RAM",
+              "storageOption": "NVM",
```
tcarmelveilleux, line 2447: This is a very coarse way to force a value. It makes it stateful to last value set. I recommend against doing this in ZAP and rather keeping state in the application. This at worst could cause some attributes that are not marked `N` in the spec (like `OnOff`) to be stateful across binary runs, which is likely not expected by some chef users.

erwinpan1: > This at worst could cause some attributes that are not marked N in the spec (like OnOff) to be stateful across binary runs

Only few attributes need to be kept  be during running cycle will be marked as NVM. For example, I have been asked many times that why the adjusted the level value gone when they rerun the application without factory reset?

examples/chef/common/clusters/media-input/MediaInputManager.cpp
```diff
@@ -56,11 +67,22 @@ uint8_t MediaInputManager::HandleGetCurrentInput()
...
+    {
+        ChipLogProgress(Zcl, "CurrentInput is same as new value: %u", index);
+        return true;
+    }
     for (auto const & inputData : mInputs)
     {
         if (inputData.index == index)
         {
             mCurrentInput = index;
+            // Sync the CurrentInput to attribute storage while reporting changes
```
bzbarsky-apple, line 80: So what I don't get is still: why are we storing the value in multiple places in memory and then having to sync?

tcarmelveilleux: Let's remove `mCurrentInput` management and only use Attributes::CurrentInput::Set/Get

erwinpan1: @bzbarsky-apple @tcarmelveilleux  `mCurrentInput` is removed. PTAL

examples/chef/common/clusters/media-playback/MediaPlaybackManager.cpp
erwinpan1: @bzbarsky-apple @tcarmelveilleux mPlaybackSpeed and mCurrentState are removed. PTAL

tcarmelveilleux (approved):

examples/chef/common/clusters/media-input/MediaInputManager.cpp
```diff
@@ -51,16 +53,32 @@ CHIP_ERROR MediaInputManager::HandleGetInputList(chip::app::AttributeValueEncode
 
 uint8_t MediaInputManager::HandleGetCurrentInput()
 {
-    return mCurrentInput;
+    uint8_t currentInput = 1;
+    Status status        = Attributes::CurrentInput::Get(mEndpoint, &currentInput);
+    if (Status::Success != status)
+    {
+        ChipLogError(Zcl, "Unable to save CurrentInput attribute, err:0x%x", to_underlying(status));
```
bzbarsky-apple, line 60: What is "save" doing in this message?  We are not saving anything here...

examples/chef/common/clusters/media-input/MediaInputManager.cpp
```diff
@@ -51,16 +53,32 @@ CHIP_ERROR MediaInputManager::HandleGetInputList(chip::app::AttributeValueEncode
...
+        ChipLogProgress(Zcl, "CurrentInput is same as new value: %u", index);
+        return true;
+    }
     for (auto const & inputData : mInputs)
     {
         if (inputData.index == index)
         {
-            mCurrentInput = index;
+            // Sync the CurrentInput to attribute storage while reporting changes
+            Status status = chip::app::Clusters::MediaInput::Attributes::CurrentInput::Set(mEndpoint, index);
```
bzbarsky-apple, line 77:
```suggestion
            Status status = Attributes::CurrentInput::Set(mEndpoint, index);
```

examples/chef/common/clusters/media-playback/MediaPlaybackManager.cpp
```diff
@@ -15,26 +15,34 @@
...
 
 PlaybackStateEnum MediaPlaybackManager::HandleGetCurrentState()
 {
-    return mCurrentState;
+    PlaybackStateEnum currentState = PlaybackStateEnum::kPlaying;
+
+    Status status = Attributes::CurrentState::Get(mEndpoint, &currentState);
+    if (Status::Success != status)
+    {
+        ChipLogError(Zcl, "Unable to save CurrentStage attribute, err:0x%x", to_underlying(status));
```
bzbarsky-apple, line 43: Again, why "save"?

examples/chef/common/clusters/media-playback/MediaPlaybackManager.cpp
```diff
@@ -54,7 +62,14 @@ CHIP_ERROR MediaPlaybackManager::HandleGetSampledPosition(AttributeValueEncoder
 
 float MediaPlaybackManager::HandleGetPlaybackSpeed()
 {
-    return mPlaybackSpeed;
+    float playbackSpeed = 1.0;
+
+    Status status = Attributes::PlaybackSpeed::Get(mEndpoint, &playbackSpeed);
+    if (Status::Success != status)
+    {
+        ChipLogError(Zcl, "Unable to save PlaybackSpeed attribute, err:0x%x", to_underlying(status));
```
bzbarsky-apple, line 70: Again, "save"?

examples/chef/common/clusters/media-playback/MediaPlaybackManager.cpp
```diff
@@ -99,10 +114,34 @@ CHIP_ERROR MediaPlaybackManager::HandleGetAvailableTextTracks(AttributeValueEnco
     });
 }
 
+CHIP_ERROR MediaPlaybackManager::HandleSetCurrentState(chip::app::Clusters::MediaPlayback::PlaybackStateEnum currentState)
```
bzbarsky-apple, line 117:
```suggestion
CHIP_ERROR MediaPlaybackManager::HandleSetCurrentState(PlaybackStateEnum currentState)
```

examples/chef/common/clusters/media-playback/MediaPlaybackManager.cpp
```diff
@@ -99,10 +114,34 @@ CHIP_ERROR MediaPlaybackManager::HandleGetAvailableTextTracks(AttributeValueEnco
...
+{
+    Status status = Attributes::CurrentState::Set(mEndpoint, currentState);
+
+    if (Status::Success != status)
+    {
+        ChipLogError(Zcl, "Unable to save CurrentState attribute, 0x%x", to_underlying(status));
+    }
+
+    return CHIP_NO_ERROR;
```
bzbarsky-apple, line 126: Shouldn't this in fact return `CHIP_ERROR_IM_GLOBAL_STATUS_VALUE(status)`?

examples/chef/common/clusters/media-playback/MediaPlaybackManager.cpp
```diff
@@ -99,10 +114,34 @@ CHIP_ERROR MediaPlaybackManager::HandleGetAvailableTextTracks(AttributeValueEnco
...
+CHIP_ERROR MediaPlaybackManager::HandleSetPlaybackSpeed(float playbackSpeed)
+{
+    Status status = Attributes::PlaybackSpeed::Set(mEndpoint, playbackSpeed);
+
+    if (Status::Success != status)
+    {
+        ChipLogError(Zcl, "Unable to set PlaybackSpeed attribute, 0x%x", to_underlying(status));
+    }
+
+    return CHIP_NO_ERROR;
```
bzbarsky-apple, line 138: Again, shouldn't this be `CHIP_ERROR_IM_GLOBAL_STATUS_VALUE(status)`?

examples/chef/common/clusters/media-playback/MediaPlaybackManager.cpp
erwinpan1: Fixed done. Thanks for the correction.

examples/chef/common/clusters/media-playback/MediaPlaybackManager.cpp
erwinpan1: Fixed done.

examples/chef/common/clusters/media-playback/MediaPlaybackManager.cpp
erwinpan1: Fixed done. Thanks for the correction.

examples/chef/common/clusters/media-input/MediaInputManager.cpp
erwinpan1: Fixed done. Thanks for the correction.

examples/chef/common/clusters/media-input/MediaInputManager.cpp
erwinpan1: Fixed done. Thanks for the suggestion.

examples/chef/common/clusters/media-playback/MediaPlaybackManager.cpp
erwinpan1: Fixed done. Thanks for the correction.

examples/chef/common/clusters/media-playback/MediaPlaybackManager.cpp
erwinpan1: @bzbarsky-apple   Fixed done. Thanks for the correction. Could you help to take a look again?

bzbarsky-apple (approved):

PR #35124, opened 2024-08-21
TC-DRLK-2.9

sumaky:
1. Created python file TC-DRLK-2.9.py 2. as the existing TC-DRLK-2.9.yaml file was not in sync with existing test plan as it required looping
2. removed existing yaml file form the below path src/app/tests/suites/certification/Test_TC_DRLK_2_9.yaml


src/python_testing/TC_DRLK_2_9.py
```diff
@@ -0,0 +1,566 @@
...
+                                       timedRequestTimeoutMs=1000)
+        except InteractionModelError as e:
+            logging.exception(e)
+            asserts.assert_equal(e.status, expected_status, f"Unexpected error returned: {e}")
+
+    @async_test_body
+    async def test_TC_DRLK_2_9(self):
+
+        self.Attrib = 0
+        self.numberofcredentialsupportedperuser = None
```
Harshith-GRL, line 253: Check for the usages, I am not able find any use of the values

src/python_testing/TC_DRLK_2_9.py
```diff
@@ -0,0 +1,566 @@
...
+        self.pics_send_setcredential_pin_cmd = False
+        self.pics_send_setcredential_rfid_cmd = False
+        self.pin_code = b""
+        self.pin_code1 = b""
+        self.pin_code2 = b""
+        self.inavlid_pincode = b""
+        self.rfid_tag = b""
+
+        if self.pics_guard(self.check_pics("DRLK.S.F00") and self.check_pics("DRLK.S.F08")
+                           and self.check_pics("DRLK.S.C22.Rsp") and self.check_pics("DRLK.S.C23.Tx")):
```
Harshith-GRL, line 277: self.pics_guard must be always used before using self.step do not define it will print misleading messages in terminal

src/python_testing/TC_DRLK_2_9.py
```diff
@@ -0,0 +1,566 @@
...
+                                    str(response.credentials)))
+            asserts.assert_equal(expected_status, Status.Success)
+
+        except InteractionModelError as e:
+            asserts.assert_equal(e.status, expected_status, f"Unexpected error returned: {e}")
+
+    async def get_credentials_status(self, credentialIndex: int, credentialType: drlkcluster.Enums.CredentialTypeEnum,
+                                     step, userIndex, credential_exists):
+        if step:
+            self.step(step)
```
Harshith-GRL, line 185: Remove this redundant check as most of the steps are being printed/called before this function is called

src/python_testing/TC_DRLK_2_9.py
```diff
@@ -0,0 +1,566 @@
...
+                                    str(response.userIndex)))
+            asserts.assert_true(response.userName == username,
+                                "Error when executing GetUserResponse command, credentialExists={}".format(
+                                    str(response.userName)))
+            asserts.assert_true(response.userUniqueID == useruniqueid,
+                                "Error when executing GetUserResponse command, credentialExists={}".format(
+                                    str(response.userUniqueID)))
+            asserts.assert_true(response.credentials == credentiallist,
+                                "Error when executing GetUserResponse command, credentialExists={}".format(
+                                    str(response.credentials)))
```
Harshith-GRL, line 176: replace **credentialsExists** with suitable names like UserIndex, Username etc in error messages

src/python_testing/TC_DRLK_2_9.py
```diff
@@ -0,0 +1,566 @@
...
+            response = await self.send_single_cmd(endpoint=self.app_cluster_endpoint, timedRequestTimeoutMs=1000,
+                                                  cmd=drlkcluster.Commands.GetCredentialStatus(
+                                                      credential=credentials_struct))
+            asserts.assert_true(type_matches(response, drlkcluster.Commands.GetCredentialStatusResponse),
+                                "Unexpected return type for GetCredentialStatus")
+            asserts.assert_true(response.credentialExists == credential_exists,
+                                "Error when executing GetCredentialStatus command, credentialExists={}".format(
+                                    str(response.credentialExists)))
+            if (not credential_exists):
+                asserts.assert_true(response.userIndex == NullValue,
```
Harshith-GRL, line 198: replace credentialExists with userIndex as that is the value used in format function

src/python_testing/TC_DRLK_2_9.py
```diff
@@ -0,0 +1,566 @@
...
+
+    async def read_attributes_from_dut(self, endpoint, cluster, attribute, expected_status: Status = Status.Success):
+        try:
+            attribute_value = await self.read_single_attribute_check_success(endpoint=endpoint, cluster=cluster,
+                                                                             attribute=attribute)
+            asserts.assert_equal(expected_status, Status.Success)
+            return attribute_value
+        except Exception as e:
+            logging.error(e)
+            asserts.assert_equal(expected_status, Status.Success,
```
Harshith-GRL, line 130: in the except block replace Status.success with any other forms like Status.InvalidState ot Status.constraintError etc since Status.Success will not be produced when excpet blocks are reached.

PR #35134, opened 2024-08-21
Test CI: why wouldn't this work?

cecille:


src/python_testing/TestChoiceConformanceSupport.py
```diff
@@ -14,6 +14,9 @@
 #    See the License for the specific language governing permissions and
 #    limitations under the License.
 #
+# === BEGIN CI TEST ARGUMENTS ===
+# test-runner-runs: run1
```
andy31415, line 18: This is a unit test arather than integration test. We should not use an integration test runner even if it works.

Ideally `choice_conformance_support` should be somewhere else as a python module and this test should follow it.

cecille:
I don't actually care that much how this is done, so closing this

PR #35139, opened 2024-08-21
Remove subscription establishment bits from the MTRDevice base class.

bzbarsky-apple:
_setupSubscriptionWithReason was only called from
_reattemptSubscriptionNowIfNeededWithReason.

_reattemptSubscriptionNowIfNeededWithReason was only called from:

* _handleUnsolicitedMessageFromPublisher, which is only called in callbacks set up by _setupSubscriptionWithReason.
* _triggerResubscribeWithReason, which only calls it when _reattemptingSubscription is YES.  But that's only set to YES in _doHandleSubscriptionReset, see next item.
* _doHandleSubscriptionReset, called from:
   * _handleSubscriptionReset: only called from callbacks set up by _setupSubscriptionWithReason.
   * _resetSubscriptionWithReasonString: called from unitTestResetSubscription and us detecting persisted data corruption in _clusterDataForPath.  But we don't actually have persisted data in the MTRDevice base class anymore, so that's not reachable.

So _setupSubscriptionWithReason is not reachable and can be removed.  And unitTestResetSubscription can be made a no-op on the base MTRDevice.  And we can remove the _resetSubscriptionWithReasonString call in _clusterDataForPath.

Once those are done, the following become unreachable:

* _handleSubscriptionReset
* _handleSubscriptionError
* Writing a nonzero value to _unitTestAttributesReportedSinceLastCheck
* _markDeviceAsUnreachableIfNeverSubscribed
* _handleResubscriptionNeededWithDelay
* _handleSubscriptionEstablished
* _handleUnsolicitedMessageFromPublisher
* _createDataVersionFilterListFromDictionary
* _setupConnectivityMonitoring
* _resetSubscriptionWithReasonString

Once those have been removed, the following become unreachable:

* HadSubscriptionEstablishedOnce
* _stopConnectivityMonitoring
* _doHandleSubscriptionError
* _doHandleSubscriptionReset

Once those have been removed, the following become unreachable:

* _setLastSubscriptionAttemptWait
* Setting _lastSubscriptionFailureTime to non-nil
* _clearSubscriptionPoolWork
* _scheduleSubscriptionPoolWork
* Use of _initialSubscribeStart
* _storePersistedDeviceData
* Use of _connectivityMonitor
* Use of _lastSubscriptionFailureTimeForDescription
* Any code depending on reattemptingSubscription being YES.
* Use of _lastSubscriptionAttemptWait
* Use of _lastSubscriptionAttemptWaitForDescription

Once those have been removed, _reattemptSubscriptionNowIfNeededWithReason is unreachable and can be removed.

At this point _currentReadClient and _currentSubscriptionCallback are always nil, so all code conditioned on those can be removed.  Once that's done, _triggerResubscribeWithReason is a no-op, so it and all calls to it can be removed.

Since _lastSubscriptionFailureTime is always nil, all code guarded on it not being nil can be removed, so _readThroughSkipped becomes a no-op and _lastSubscriptionFailureTime can be removed.

At this point, the following are not reachable:

* NeedToStartSubscriptionSetup
* _changeInternalState

Once those are removed, _internalDeviceStateForDescription is unused, and _internalDeviceState is always MTRInternalDeviceStateUnsubscribed.  So HaveSubscriptionEstablishedRightNow() always returns NO and can be removed.

At this point SubscriptionCallback is unused and can also be removed.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -22,7 +22,6 @@
 #import "MTRAttributeSpecifiedCheck.h"
 #import "MTRBaseClusters.h"
 #import "MTRBaseDevice_Internal.h"
-#import "MTRBaseSubscriptionCallback.h"
 #import "MTRCluster.h"
 #import "MTRClusterConstants.h"
 #import "MTRCommandTimedCheck.h"
@@ -43,16 +42,8 @@
 #import "zap-generated/MTRCommandPayloads_Internal.h"
 
 #import "lib/core/CHIPError.h"
-#import "lib/core/DataModelTypes.h"
-#import <app/ConcreteAttributePath.h>
-#import <lib/support/FibonacciUtils.h>
 
-#import <app/AttributePathParams.h>
-#import <app/BufferedReadCallback.h>
-#import <app/ClusterStateCache.h>
-#import <app/InteractionModelEngine.h>
 #import <platform/LockTracker.h>
-#import <platform/PlatformManager.h>
 
 typedef void (^MTRDeviceAttributeReportHandler)(NSArray * _Nonnull);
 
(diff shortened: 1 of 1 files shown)


jtung-apple (approved):

bzbarsky-apple:

PR #35144, opened 2024-08-22
dac_revocation: Perform cross validation against crl signer or crl signer delegator

### LINKED ISSUE #34587: Cross-validate the CRLSignerCertificate and CRLSignerDelegator per spec:
Add a TODO (and create associated issue) to cross-validate the CRLSignerCertificate and CRLSignerDelegator per spec:

```
. If a matching `RevocationSet` is found, determined if `serialNumber` is a member of the set.
    .. If the entity type whose revocation status is being verified is a PAI, then the subject PAI's issuer (a PAA) SHALL fulfill one of the following two cases, otherwise return immediately:
      ... The Subject and Subject Key of the PAI certificate's issuer matches exactly the CRLSignerCertificate's subject (i.e. the CRLSignerCertificate is a PAA).
      ... The Subject and Subject Key of the PAI certificate's issuer matches exactly the PAA which is the issuer of the CRLSignerCertificate (i.e. the CRLSignerCertificate is a CRL signer delegated by a PAA).
    .. If the entity type whose revocation status is being verified is a DAC, then:
      ... If the `CRLSignerDelegator` is present, then the subject DAC's issuer (a PAI) SHALL match the `CRLSignerDelegator` in both Subject Key and Subject, otherwise return immediately.
      ... If the `CRLSignerDelegator` is absent, then the subject DAC's issuer (a PAI) SHALL match the `CRLSignerCertificate` in both Subject Key and Subject, otherwise return immediately.
```

Those checks are not currently do, and MUST be done later once this is supported in revocation set construction.

_Originally posted by @tcarmelveilleux in https://github.com/project-chip/connectedhomeip/pull/33651#discussion_r1695256832_

shubhamdp:
#### Change Overview
- perform the cross validation of DAC/PAI with crl signer and crl signer delegator cert
- Extended and added unit tests for crl signer delegator case

#### Tests
- Added unit tests
- generated the sets for test-net and main-net dcl and used them with chip-tool.
- generated the revoked dataset and verified that commissioning fails when revoked entity is present.

Fixes #34587

### DIFF
--- a/src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
+++ b/src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
@@ -117,7 +117,7 @@ bool TestDACRevocationDelegateImpl::CrossValidateCert(const Json::Value & revoke
                                                       const std::string & issuerNameBase64Str)
 {
     std::string certPEM;
-    [[maybe_unused]] const char * certType;
+    [[maybe_unused]] std::string certType;
 
     if (revokedSet.isMember("crl_signer_delegator"))
     {
@@ -134,8 +134,8 @@ bool TestDACRevocationDelegateImpl::CrossValidateCert(const Json::Value & revoke
     std::string keyId;   // crl signer or crl signer delegator SKID
     VerifyOrReturnValue(CHIP_NO_ERROR == GetSubjectAndKeyIdFromPEMCert(certPEM, subject, keyId), false);
 
-    ChipLogDetail(NotSpecified, "%s: Subject: %s", certType, subject.c_str());
-    ChipLogDetail(NotSpecified, "%s: SKID: %s", certType, keyId.c_str());
+    ChipLogDetail(NotSpecified, "%s: Subject: %s", certType.c_str(), subject.c_str());
+    ChipLogDetail(NotSpecified, "%s: SKID: %s", certType.c_str(), keyId.c_str());
 
     return (akidHexStr == keyId && issuerNameBase64Str == subject);
 }
@@ -315,8 +315,8 @@ void TestDACRevocationDelegateImpl::CheckForRevokedDACChain(
 
     if (mDeviceAttestationRevocationSetPath.empty())
     {
-
         onCompletion->mCall(onCompletion->mContext, info, attestationError);
+        return;
     }
 
     ChipLogDetail(NotSpecified, "Checking for revoked DAC in %s", mDeviceAttestationRevocationSetPath.c_str());
--- a/src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
+++ b/src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
@@ -29,6 +29,8 @@ using namespace chip::Crypto;
 namespace chip {
 namespace Credentials {
 
+static constexpr uint32_t kMaxIssuerBase64Len = BASE64_ENCODED_LEN(kMaxCertificateDistinguishedNameLength);
+
 namespace {
 CHIP_ERROR BytesToHexStr(const ByteSpan & bytes, MutableCharSpan & outHexStr)
 {
@@ -37,6 +39,31 @@ CHIP_ERROR BytesToHexStr(const ByteSpan & bytes, MutableCharSpan & outHexStr)
     outHexStr.reduce_size(2 * bytes.size());
     return CHIP_NO_ERROR;
 }
+
+CHIP_ERROR X509_PemToDer(const std::string & pemCert, MutableByteSpan & derCert)
+{
+    std::string beginMarker = "-----BEGIN CERTIFICATE-----";
+    std::string endMarker   = "-----END CERTIFICATE-----";
+
+    std::size_t beginPos = pemCert.find(beginMarker);
+    std::size_t endPos   = pemCert.find(endMarker);
+
+    // Extract content between markers
+    std::string plainB64Str = pemCert.substr(beginPos + beginMarker.length(), endPos - (beginPos + beginMarker.length()));
+
+    // Remove all newline characters '\n' and '\r'
+    plainB64Str.erase(std::remove(plainB64Str.begin(), plainB64Str.end(), '\n'), plainB64Str.end());
+    plainB64Str.erase(std::remove(plainB64Str.begin(), plainB64Str.end(), '\r'), plainB64Str.end());
+
+    // decode b64
+    uint16_t derLen = Base64Decode(plainB64Str.c_str(), static_cast<uint16_t>(plainB64Str.size()), derCert.data());
+    VerifyOrReturnError(derLen != UINT16_MAX, CHIP_ERROR_INVALID_ARGUMENT);
+
+    derCert.reduce_size(derLen);
+
+    return CHIP_NO_ERROR;
+}
+
 } // anonymous namespace
 
 CHIP_ERROR TestDACRevocationDelegateImpl::SetDeviceAttestationRevocationSetPath(std::string_view path)
@@ -52,6 +79,59 @@ void TestDACRevocationDelegateImpl::ClearDeviceAttestationRevocationSetPath()
     mDeviceAttestationRevocationSetPath = mDeviceAttestationRevocationSetPath.substr(0, 0);
 }
 
+CHIP_ERROR TestDACRevocationDelegateImpl::GetSubjectAndKeyIdFromPEMCert(const std::string & certPEM, std::string & outSubject,
+                                                                        std::string & outKeyId)
+{
+    // buffers and spans for storing crl signer delegator OR crl signer cert info
+    char subjectBuf[kMaxIssuerBase64Len]             = { 0 };
+    char skidBuf[2 * kAuthorityKeyIdentifierLength]  = { 0 };
+    uint8_t certDerBuf[kMax_x509_Certificate_Length] = { 0 };
+
+    MutableCharSpan subject(subjectBuf);
+    MutableCharSpan keyId(skidBuf);
+    MutableByteSpan certDER(certDerBuf);
+
+    ReturnLogErrorOnFailure(X509_PemToDer(certPEM, certDER));
+    ReturnErrorOnFailure(GetSubjectNameBase64Str(certDER, subject));
+    ReturnErrorOnFailure(GetSKIDHexStr(certDER, keyId));
+
+    outSubject = std::string(subject.data(), subject.size());
+    outKeyId   = std::string(keyId.data(), keyId.size());
+
+    return CHIP_NO_ERROR;
+}
+
+// cross validate DAC/PAI cert details with crl signer OR crl signer delegator
+bool TestDACRevocationDelegateImpl::CrossValidateCert(const Json::Value & revokedSet, const CharSpan & akidHexStr,
+                                                      const CharSpan & issuerNameBase64Str)
+{
+    std::string certPEM;
+    const char * certType;
+
+    if (revokedSet.isMember("crl_signer_delegator"))
+    {
+        certPEM  = revokedSet["crl_signer_delegator"].asString();
+        certType = "CRL Signer delegator";
+    }
+    else
+    {
+        certPEM  = revokedSet["crl_signer_cert"].asString();
+        certType = "CRL Signer";
+    }
+
+    std::string subject; // crl signer or crl signer delegator subject
+    std::string keyId;   // crl signer or crl signer delegator SKID
+    VerifyOrReturnValue(CHIP_NO_ERROR == GetSubjectAndKeyIdFromPEMCert(certPEM, subject, keyId), false);
+
+    ChipLogDetail(NotSpecified, "%s Subject: %s", certType, subject.c_str());
+    ChipLogDetail(NotSpecified, "%s SKID: %s", certType, keyId.c_str());
+
+    std::string issuerSKID(akidHexStr.data(), akidHexStr.size());
+    std::string issuerName(issuerNameBase64Str.data(), issuerNameBase64Str.size());
+
+    return (issuerSKID == keyId && issuerName == subject);
+}
+
 // This method parses the below JSON Scheme
 // [
 //   {
@@ -62,11 +142,13 @@ void TestDACRevocationDelegateImpl::ClearDeviceAttestationRevocationSetPath()
 //       "serial1 bytes as base64",
 //       "serial2 bytes as base64"
 //     ]
+//     "crl_signer_cert": "<PEM incoded CRL signer certificate>",
+//     "crl_signer_delegator": <PEM incoded CRL signer delegator certificate>,
 //   }
 // ]
 //
 bool TestDACRevocationDelegateImpl::IsEntryInRevocationSet(const CharSpan & akidHexStr, const CharSpan & issuerNameBase64Str,
-                                                           const CharSpan & serialNumberHexStr)
+                                                           const CharSpan & serialNumberHexStr, bool isPAI)
 {
     std::ifstream file(mDeviceAttestationRevocationSetPath.c_str());
     if (!file.is_open())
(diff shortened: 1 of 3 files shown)


src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
```diff
@@ -31,13 +32,41 @@ namespace chip {
...
+
+    // Extract content between markers
+    std::string plainB64Str = pemCert.substr(beginPos + beginMarker.length(), endPos - (beginPos + beginMarker.length()));
+
+    // Remove all newline characters '\n' and '\r'
+    plainB64Str.erase(std::remove(plainB64Str.begin(), plainB64Str.end(), '\n'), plainB64Str.end());
+    plainB64Str.erase(std::remove(plainB64Str.begin(), plainB64Str.end(), '\r'), plainB64Str.end());
+
+    // decode b64
+    uint16_t derLen = Base64Decode(plainB64Str.c_str(), static_cast<uint16_t>(plainB64Str.size()), derCert.data());
```
bzbarsky-apple, line 62: How is this not a buffer overrun?  Where are we checking that `derCert` is big enough?

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
```diff
@@ -53,6 +82,59 @@ void TestDACRevocationDelegateImpl::ClearDeviceAttestationRevocationSetPath()
...
+
+    return CHIP_NO_ERROR;
+}
+
+// cross validate DAC/PAI cert details with crl signer OR crl signer delegator
+bool TestDACRevocationDelegateImpl::CrossValidateCert(const Json::Value & revokedSet, const CharSpan & akidHexStr,
+                                                      const CharSpan & issuerNameBase64Str)
+{
+    std::string certPEM;
+    const char * certType __attribute__((unused));
```
bzbarsky-apple, line 112: Why is `__attribute__((unused))` there?  If compilers complain when detail logging is enabled, then please use standard `[[maybe_unused]]` and not compiler extensions.

bzbarsky-apple (changes requested):
I did not review most of the logic; just stopped at the buffer overrun.

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
shubhamdp: Fixed the buffer overrun

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
shubhamdp: Replaced it with `[[maybe_unused]]`

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
```diff
@@ -30,13 +31,45 @@ namespace chip {
...
+
+    std::size_t beginPos = pemCert.find(beginMarker);
+    std::size_t endPos   = pemCert.find(endMarker);
+
+    // Extract content between markers
+    std::string plainB64Str = pemCert.substr(beginPos + beginMarker.length(), endPos - (beginPos + beginMarker.length()));
+
+    // Remove all newline characters '\n' and '\r'
+    plainB64Str.erase(std::remove(plainB64Str.begin(), plainB64Str.end(), '\n'), plainB64Str.end());
+    plainB64Str.erase(std::remove(plainB64Str.begin(), plainB64Str.end(), '\r'), plainB64Str.end());
```
tcarmelveilleux, line 58: This is missing checks to avoid crashing if the begin/end are not found. Please make this safe and add a unit test against several invalid certs.

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
```diff
@@ -52,6 +85,56 @@ void TestDACRevocationDelegateImpl::ClearDeviceAttestationRevocationSetPath()
...
+
+    return CHIP_NO_ERROR;
+}
+
+// Check if issuer and AKID matches with the crl signer OR crl signer delegator's subject and SKID
+bool TestDACRevocationDelegateImpl::CrossValidateCert(const Json::Value & revokedSet, const std::string & akidHexStr,
+                                                      const std::string & issuerNameBase64Str)
+{
+    std::string certPEM;
+    [[maybe_unused]] const char * certType;
```
tcarmelveilleux, line 115: Why `[[maybe_unused]]`. Also, if you use std::string for some parts, why not use std::string for this as well?

src/credentials/tests/TestDeviceAttestationCredentials.cpp
```diff
@@ -548,28 +556,48 @@ TEST_F(TestDeviceAttestationCredentials, TestDACRevocationDelegateImpl)
...
+        "crl_signer_cert": "-----BEGIN CERTIFICATE-----\nMIIB1DCCAXqgAwIBAgIIPmzmUJrYQM0wCgYIKoZIzj0EAwIwMDEYMBYGA1UEAwwP\nTWF0dGVyIFRlc3QgUEFBMRQwEgYKKwYBBAGConwCAQwERkZGMTAgFw0yMTA2Mjgx\nNDIzNDNaGA85OTk5MTIzMTIzNTk1OVowRjEYMBYGA1UEAwwPTWF0dGVyIFRlc3Qg\nUEFJMRQwEgYKKwYBBAGConwCAQwERkZGMTEUMBIGCisGAQQBgqJ8AgIMBDgwMDAw\nWTATBgcqhkjOPQIBBggqhkjOPQMBBwNCAASA3fEbIo8+MfY7z1eY2hRiOuu96C7z\neO6tv7GP4avOMdCO1LIGBLbMxtm1+rZOfeEMt0vgF8nsFRYFbXDyzQsio2YwZDAS\nBgNVHRMBAf8ECDAGAQH/AgEAMA4GA1UdDwEB/wQEAwIBBjAdBgNVHQ4EFgQUr0K3\nCU3r1RXsbs8zuBEVIl8yUogwHwYDVR0jBBgwFoAUav0idx9RH+y/FkGXZxDc3DGh\ncX4wCgYIKoZIzj0EAwIDSAAwRQIhAJbJyM8uAYhgBdj1vHLAe3X9mldpWsSRETET\ni+oDPOUDAiAlVJQ75X1T1sR199I+v8/CA2zSm6Y5PsfvrYcUq3GCGQ==\n-----END CERTIFICATE-----",
         "revoked_serial_numbers": ["0C694F7F866067B21234"]
     }]
     )";
     WriteTestRevokedData(jsonData, tmpJsonFile);
     revocationDelegateImpl.CheckForRevokedDACChain(info, &attestationInformationVerificationCallback);
     EXPECT_EQ(attestationResult, AttestationVerificationResult::kSuccess);
+
+    // Test DAC is revoked, and crl signer delegator is present
+    // crl_signer_cert is not a valid and just used for testing
```
tcarmelveilleux, line 589:
```suggestion
    // crl_signer_cert is not valid and just used for testing
```

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
shubhamdp: `certType` is only used in `ChipLogDetail`, and when the application is built with no detail logs, it is unused.

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
shubhamdp: Added the checks after getting marker positions, and unit tests.

src/credentials/tests/TestDeviceAttestationCredentials.cpp
shubhamdp: added this change locally along with other ones.

shubhamdp:
@tcarmelveilleux @bzbarsky-apple can you please re-review the PR.

tcarmelveilleux (approved):

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
```diff
@@ -30,13 +31,50 @@ namespace chip {
...
+    std::string endMarker   = "-----END CERTIFICATE-----";
+
+    std::size_t beginPos = pemCert.find(beginMarker);
+    VerifyOrReturnError(beginPos != std::string::npos, CHIP_ERROR_INVALID_ARGUMENT);
+
+    std::size_t endPos = pemCert.find(endMarker);
+    VerifyOrReturnError(endPos != std::string::npos, CHIP_ERROR_INVALID_ARGUMENT);
+
+    // Extract content between markers
+    std::string plainB64Str = pemCert.substr(beginPos + beginMarker.length(), endPos - (beginPos + beginMarker.length()));
```
bzbarsky-apple, line 57: What happens if `endPos < beginPos` here?

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
```diff
@@ -116,14 +213,32 @@ bool TestDACRevocationDelegateImpl::IsEntryInRevocationSet(const CharSpan & akid
     return false;
 }
 
-CHIP_ERROR TestDACRevocationDelegateImpl::GetAKIDHexStr(const ByteSpan & certDer, MutableCharSpan & outAKIDHexStr)
+CHIP_ERROR TestDACRevocationDelegateImpl::GetKeyIDHexStr(const ByteSpan & certDer, MutableCharSpan & outKeyIDHexStr, bool isAKID)
 {
-    uint8_t akidBuf[kAuthorityKeyIdentifierLength];
-    MutableByteSpan akid(akidBuf);
+    // kAuthorityKeyIdentifierLength and kSubjectKeyIdentifierLength are equal
```
bzbarsky-apple, line 218: That should be a static_assert, not a comment.

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
```diff
@@ -135,25 +250,44 @@ CHIP_ERROR TestDACRevocationDelegateImpl::GetSerialNumberHexStr(const ByteSpan &
...
+    outRDNBase64String.reduce_size(encodedLen);
     return CHIP_NO_ERROR;
 }
 
-bool TestDACRevocationDelegateImpl::IsCertificateRevoked(const ByteSpan & certDer)
+CHIP_ERROR TestDACRevocationDelegateImpl::GetIssuerNameBase64Str(const ByteSpan & certDer,
+                                                                 MutableCharSpan & outIssuerNameBase64String)
 {
-    static constexpr uint32_t maxIssuerBase64Len = BASE64_ENCODED_LEN(kMaxCertificateDistinguishedNameLength);
+    return GetRDNBase64Str(certDer, outIssuerNameBase64String, true /* isIssuer */);
```
bzbarsky-apple, line 278: Followup: instead of a boolean, we should have an enum class for the possible RDNs we might want.  It would also be self-documenting.

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
```diff
@@ -135,25 +250,44 @@ CHIP_ERROR TestDACRevocationDelegateImpl::GetSerialNumberHexStr(const ByteSpan &
...
+}
+
+CHIP_ERROR TestDACRevocationDelegateImpl::GetSubjectNameBase64Str(const ByteSpan & certDer,
+                                                                  MutableCharSpan & outSubjectNameBase64String)
+{
+    return GetRDNBase64Str(certDer, outSubjectNameBase64String, false /* isIssuer */);
+}
 
-    char issuerNameBuffer[maxIssuerBase64Len]                            = { 0 };
+// @param certDer Certificate being tested for revocation in
```
bzbarsky-apple, line 287: "in" what?  It looks like this comment got truncated?

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
```diff
@@ -30,13 +31,50 @@ namespace chip {
...
+
 CHIP_ERROR BytesToHexStr(const ByteSpan & bytes, MutableCharSpan & outHexStr)
 {
     Encoding::HexFlags flags = Encoding::HexFlags::kUppercase;
     ReturnErrorOnFailure(BytesToHex(bytes.data(), bytes.size(), outHexStr.data(), outHexStr.size(), flags));
     outHexStr.reduce_size(2 * bytes.size());
     return CHIP_NO_ERROR;
 }
+
+CHIP_ERROR X509_PemToDer(const std::string & pemCert, MutableByteSpan & derCert)
```
ksperling-apple, line 45: Do we have to handle PEM at all here? Usually we let users do the conversion via chip-cert or openssl and just have the SDK code handle DER.

shubhamdp: CRL signer certificate or CRL signer delegator certificate are stored as PEM in the DCL, [generate-revocation-set.py](https://github.com/project-chip/connectedhomeip/blob/master/credentials/generate-revocation-set.py) generates the revocation set and stores them into json file, SDK reads that json file, so for now, we are stick with that choice.

We can extract the middle part from the cert in script and store it, so only need to base64 decode to make it DER.

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
shubhamdp: added that check

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
```diff
@@ -52,6 +90,56 @@ void TestDACRevocationDelegateImpl::ClearDeviceAttestationRevocationSetPath()
     mDeviceAttestationRevocationSetPath = mDeviceAttestationRevocationSetPath.substr(0, 0);
 }
 
+CHIP_ERROR TestDACRevocationDelegateImpl::GetSubjectAndKeyIdFromPEMCert(const std::string & certPEM, std::string & outSubject,
+                                                                        std::string & outKeyId)
```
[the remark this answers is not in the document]
shubhamdp: documented the format.

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
shubhamdp: thanks, added the static_assert

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
shubhamdp: raised #36611

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
shubhamdp: "DER format" :-). Yeah, it was truncated.

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
tcarmelveilleux: It could be converted to DER in the JSON files for the test version. Let's keep that as a follow-up though.

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
```diff
@@ -135,25 +255,44 @@ CHIP_ERROR TestDACRevocationDelegateImpl::GetSerialNumberHexStr(const ByteSpan &
...
+}
+
+CHIP_ERROR TestDACRevocationDelegateImpl::GetSubjectNameBase64Str(const ByteSpan & certDer,
+                                                                  MutableCharSpan & outSubjectNameBase64String)
+{
+    return GetRDNBase64Str(certDer, outSubjectNameBase64String, false /* isIssuer */);
+}
 
-    char issuerNameBuffer[maxIssuerBase64Len]                            = { 0 };
+// @param certDer Certificate to check for revocation in DER format
```
bzbarsky-apple, line 292:
```suggestion
// @param certDer Certificate, in DER format, to check for revocation
```

bzbarsky-apple (approved):

src/credentials/attestation_verifier/TestDACRevocationDelegateImpl.cpp
shubhamdp: Raised #36625 for tracking

Issue #35030, opened 2024-08-16
[SVE] Allow configure the port number for the local and remote bridge


yufengwangca:
### Reproduction steps

[examples/fabric-admin/device_manager/DeviceManager.cpp]

namespace {
// Constants
constexpr uint32_t kSetupPinCode               = 20202021;
constexpr uint16_t kRemoteBridgePort           = 5540;
constexpr uint16_t kLocalBridgePort            = 5540;
}

local and remote bridge port number need to be configurable based on argument if we plan to have two bridges running on the same host

### Bug prevalence

always

### GitHub hash of the SDK that was being used

e8e9fc5a21d825b1b5b718dc328a11e486e2f6fa

### Platform

other

### Platform Version(s)

e8e9fc5a21d825b1b5b718dc328a11e486e2f6fa

### Type

Test Improvement

### Testing

CI tested

PR #35146, opened 2024-08-22
[Fabric-Sync] Allow to specify custom port in fabric-bridge

arkq:
### Problem

Example application for fabric-sync does not allow commissioning port customization. This is required to run more then one application on the same host.


tehampson (approved):

yufengwangca:
After we allow to specify custom port in fabric-bridge, we need to also customize the bridge port number in fabric-admin

https://github.com/project-chip/connectedhomeip/blob/master/examples/fabric-admin/device_manager/DeviceManager.cpp#L35

I assign https://github.com/project-chip/connectedhomeip/issues/35030 to you to follow-up

yufengwangca (approved):

cecille (approved):

saurabhst (approved):

PR #35149, opened 2024-08-22
Fix payload missing 0 in getting started example

wickywaka:
I was following this getting started and was copying pasting from the commands (which is never advised :P) and notice this mistake in the payload. The qr-code payload for the pairing command was missing a trailing 0 so here I will try to fix it.

I am new here so please let me know if I am missing some guidelines etc.


bzbarsky-apple (approved):
Great catch, thank you!

bzbarsky-apple:
Fast-tracking documentation fix.

PR #35150, opened 2024-08-22
Remove readAttributeWithEndpointID implementation from MTRDevice.

bzbarsky-apple:
This is implemented (differently) by the different subclasses.

Once this implementation is removed, the following become unreachable and can be removed:

* _attributeValueDictionaryForAttributePath
* _subscriptionAbleToReport
* _readThroughSkipped

At that point _subscriptionsAllowed becomes unreachable and can be removed.


jtung-apple (approved):

bzbarsky-apple:

PR #35151, opened 2024-08-22
Remove unnecessary invokeCommand overrides from MTRDevice_Concrete.

bzbarsky-apple:
The one part that is not shared with the XPC implementation is _invokeCommandWithEndpointID:....  Everything else is just generic argument massaging and forwarding that can keep living in the base MTRDevice.


jtung-apple (approved):

bzbarsky-apple:

PR #35152, opened 2024-08-22
Mark return types nullable in MTRDevice_XPC when nil can be returned.

bzbarsky-apple:


jtung-apple (approved):

bzbarsky-apple:

PR #35156, opened 2024-08-22
Update ColorControl bounds based on spec fixes

tcarmelveilleux:
### Problem

Minimum color temperature bounds in spec were too wide and included
a value for ColorTempPhysicalMinMireds that could lead to divide-by-zero
in examples (and it did). 

The spec has been fixed (https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/10202) and test plan updated to avoid the zero value (https://github.com/CHIP-Specifications/chip-test-plans/pull/4545).

### This PR

- Updates the SDK and test script to make 0 an invalid value for ColorTempPhysicalMinMireds
- Adds internal safety bound checks to prevent things from falling through the cracks even if
  attributes are written to invalid values.
- Updates the XML to make the value 0 forbidden where it should be forbidden.
- Makes all examples have reasonable Min/Max claimed physical values (2000K..6500K --> 154 to 454 mireds) as suggested  by Signify (these values can be made different, but the previous defaults were actively wrong and never correct).
  - Note that the ZAP files from SDK have been widely copied, and the exact sample values have been seen in commercially available products, so fixing the examples if the most important part.

### Testing done

- All cert tests still pass
- Validated that with a value of 0, there is no crash


examples/all-clusters-app/all-clusters-common/all-clusters-app.matter
```diff
@@ -9164,8 +9164,8 @@ endpoint 1 {
     ram      attribute colorLoopStartEnhancedHue default = 0x2300;
     ram      attribute colorLoopStoredEnhancedHue default = 0x0000;
     ram      attribute colorCapabilities default = 0x1F;
-    ram      attribute colorTempPhysicalMinMireds default = 0x0000;
-    ram      attribute colorTempPhysicalMaxMireds default = 0xFEFF;
+    ram      attribute colorTempPhysicalMinMireds default = 0x009A;
```
tcarmelveilleux, line 9167: Do not merge (or resolve this comment) until spec PR and test plan PR are fully merged.

cecille (approved):

jmartinez-silabs (approved):

bzbarsky-apple (approved):

src/app/tests/suites/certification/Test_TC_CC_6_5.yaml
```diff
@@ -75,7 +75,7 @@ tests:
       response:
           constraints:
               type: int16u
-              minValue: 0
+              minValue: 1
               maxValue: 65279
 
     - label:
```
ReneJosefsen, line 81: This one also needs to be minValue: 1, right?

ReneJosefsen (approved):

PR #35159, opened 2024-08-22
[Darwin] XPC invoke should implement bottom-most method

jtung-apple:
MTRDevice_XPC did not override the bottom-most invokeCommand method, and so MTRCluster objects would not go through the XPC path.

This patch fixes the override as well as the xpc interface / server protocol.


kiel-apple (approved):

kiel-apple:
fast-tracking platform owner change to platform-specific code

PR #35161, opened 2024-08-22
[HVAC] Handle null BuiltIn field on Preset write according to spec

hasty:


src/app/clusters/thermostat-server/thermostat-server-presets.cpp
```diff
@@ -367,37 +368,52 @@ CHIP_ERROR ThermostatAttrAccess::AppendPendingPreset(Thermostat::Delegate * dele
     }
```
bzbarsky-apple, line 368: Shouldn't we set BuiltIn to false on `preset` here if it's coming in with BuiltIn set to null and has no handle?  At least assuming this PR is implementing https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/10277

src/python_testing/TC_TSTAT_4_2.py
```diff
@@ -269,7 +269,9 @@ async def test_TC_TSTAT_4_2(self):
             await self.send_atomic_request_begin_command()
 
             # Write to the presets attribute after calling AtomicRequest command
-            await self.write_presets(endpoint=endpoint, presets=new_presets)
+            test_presets = copy.deepcopy(new_presets)
```
bzbarsky-apple, line 272: We should test both cases: null on a preset matching an existing handle, and null on a preset with a null handle.

src/app/clusters/thermostat-server/thermostat-server-presets.cpp
hasty: Done.

src/python_testing/TC_TSTAT_4_2.py
hasty: OK, added to the above test for the other case

bzbarsky-apple (approved):

bauerschwan (approved):

jmartinez-silabs (approved):

PR #35167, opened 2024-08-23
Move the constraint check for number of presets not to exceed the num…

### LINKED ISSUE #35166: Move the constraint check for the size of the presets or schedules attributes exceeding the total number supported from the atomic write commit request to the write request
Currently the constraint check for the maximum number of presets or schedules allowed is done in the atomic write commit command which is incorrect. The check should be in the write request for these attributes

Corresponding to spec issue - https://github.com/CHIP-Specifications/connectedhomeip-spec/issues/10313

nivi-apple:
…ber of presets supported from the atomic write commit command handling to the write request

Fixes: #35166

### DIFF
--- a/src/app/clusters/thermostat-server/thermostat-server-presets.cpp
+++ b/src/app/clusters/thermostat-server/thermostat-server-presets.cpp
@@ -405,8 +405,10 @@ CHIP_ERROR ThermostatAttrAccess::AppendPendingPreset(Thermostat::Delegate * dele
     // Before adding this preset to the pending presets, if the expected length of the pending presets' list
     // exceeds the total number of presets supported, return RESOURCE_EXHAUSTED. Note that the preset has not been appended yet.
 
+    uint8_t numberOfPendingPresets = CountNumberOfPendingPresets(delegate);
+
     // Increment number of pending presets by 1 to account for this preset.
-    uint8_t totalExpectedCount = CountNumberOfPendingPresets(delegate) + 1;
+    uint8_t totalExpectedCount = numberOfPendingPresets + 1;
 
     uint8_t numberOfPresetsSupported = delegate->GetNumberOfPresets();
 
--- a/src/app/clusters/thermostat-server/thermostat-server-presets.cpp
+++ b/src/app/clusters/thermostat-server/thermostat-server-presets.cpp
@@ -407,12 +407,8 @@ CHIP_ERROR ThermostatAttrAccess::AppendPendingPreset(Thermostat::Delegate * dele
 
     uint8_t numberOfPendingPresets = CountNumberOfPendingPresets(delegate);
 
-    // Increment number of pending presets by 1 to account for this preset.
-    uint8_t totalExpectedCount = numberOfPendingPresets + 1;
-
-    uint8_t numberOfPresetsSupported = delegate->GetNumberOfPresets();
-
-    if (totalExpectedCount > numberOfPresetsSupported)
+    // We will be adding one more preset, so reject if the length is already at max.
+    if (numberOfPendingPresets >= numberOfPresetsSupported)
     {
         return CHIP_IM_GLOBAL_STATUS(ResourceExhausted);
     }
--- a/src/app/clusters/thermostat-server/thermostat-server-presets.cpp
+++ b/src/app/clusters/thermostat-server/thermostat-server-presets.cpp
@@ -412,13 +412,7 @@ CHIP_ERROR ThermostatAttrAccess::AppendPendingPreset(Thermostat::Delegate * dele
 
     uint8_t numberOfPresetsSupported = delegate->GetNumberOfPresets();
 
-    if (numberOfPresetsSupported == 0)
-    {
-        ChipLogError(Zcl, "AppendPendingPreset: Failed to get NumberOfPresets");
-        return CHIP_IM_GLOBAL_STATUS(InvalidInState);
-    }
-
-    if (numberOfPresetsSupported > 0 && totalExpectedCount > numberOfPresetsSupported)
+    if (totalExpectedCount > numberOfPresetsSupported)
     {
         return CHIP_IM_GLOBAL_STATUS(ResourceExhausted);
     }
--- a/src/app/clusters/thermostat-server/thermostat-server-presets.cpp
+++ b/src/app/clusters/thermostat-server/thermostat-server-presets.cpp
@@ -402,6 +402,28 @@ CHIP_ERROR ThermostatAttrAccess::AppendPendingPreset(Thermostat::Delegate * dele
         return CHIP_IM_GLOBAL_STATUS(ConstraintError);
     }
 
+    // Before adding this preset to the pending presets, if the expected length of the pending presets' list
+    // exceeds the total number of presets supported, return RESOURCE_EXHAUSTED. Note that the preset has not been appended yet.
+
+    // Increment number of pending presets by 1 to account for this preset.
+    uint8_t totalExpectedCount = CountNumberOfPendingPresets(delegate) + 1;
+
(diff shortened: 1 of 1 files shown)


src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -933,6 +933,26 @@ CHIP_ERROR ThermostatAttrAccess::AppendPendingPreset(Thermostat::Delegate * dele
         return CHIP_IM_GLOBAL_STATUS(ConstraintError);
     }
 
+    uint8_t totalCount = CountNumberOfPendingPresets(delegate);
```
bzbarsky-apple, line 936: This conflicts with https://github.com/project-chip/connectedhomeip/pull/34936.  That needs to land, then we can merge this (much smaller) change on top.

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -933,6 +933,26 @@ CHIP_ERROR ThermostatAttrAccess::AppendPendingPreset(Thermostat::Delegate * dele
...
+
+    if (numberOfPresetsSupported == 0)
+    {
+        ChipLogError(Zcl, "AppendPendingPreset: Failed to get NumberOfPresets");
+        return CHIP_IM_GLOBAL_STATUS(InvalidInState);
+    }
+
+    // If the expected length of the presets attribute with the applied changes exceeds the total number of presets supported,
+    // return RESOURCE_EXHAUSTED. Note that the changes are not yet applied.
+    if (numberOfPresetsSupported > 0 && totalCount > numberOfPresetsSupported)
```
bzbarsky-apple, line 948: This check does not make sense.

src/app/clusters/thermostat-server/thermostat-server.cpp
nivi-apple: rewrote

src/app/clusters/thermostat-server/thermostat-server-presets.cpp
```diff
@@ -402,6 +402,28 @@ CHIP_ERROR ThermostatAttrAccess::AppendPendingPreset(Thermostat::Delegate * dele
...
+
+    uint8_t numberOfPresetsSupported = delegate->GetNumberOfPresets();
+
+    if (numberOfPresetsSupported == 0)
+    {
+        ChipLogError(Zcl, "AppendPendingPreset: Failed to get NumberOfPresets");
+        return CHIP_IM_GLOBAL_STATUS(InvalidInState);
+    }
+
+    if (numberOfPresetsSupported > 0 && totalExpectedCount > numberOfPresetsSupported)
```
hasty, line 419: Comment got lost in the rebase, apparently: this is an unsigned value, and you just checked if it was zero on 413, so the first part of this is always true.

nivi-apple: fixed.

src/app/clusters/thermostat-server/thermostat-server-presets.cpp
```diff
@@ -402,6 +402,30 @@ CHIP_ERROR ThermostatAttrAccess::AppendPendingPreset(Thermostat::Delegate * dele
...
+
+    // Increment number of pending presets by 1 to account for this preset.
+    uint8_t totalExpectedCount = numberOfPendingPresets + 1;
+
+    uint8_t numberOfPresetsSupported = delegate->GetNumberOfPresets();
+
+    if (numberOfPresetsSupported == 0)
+    {
+        ChipLogError(Zcl, "AppendPendingPreset: Failed to get NumberOfPresets");
+        return CHIP_IM_GLOBAL_STATUS(InvalidInState);
```
hasty, line 418: Shouldn't this be resource exhausted? Or just chuck this and let the next check handle it? Boris felt that if the delegate returned "0" for a timeout in atomic write, we couldn't return InvalidInState because the spec didn't account for the possibility that the delegate says something nonsensical, which is the same problem here.

nivi-apple: yeah i was thinking about that. will return resourceexhausted

nivi-apple: done.

src/app/clusters/thermostat-server/thermostat-server-presets.cpp
```diff
@@ -402,6 +402,24 @@ CHIP_ERROR ThermostatAttrAccess::AppendPendingPreset(Thermostat::Delegate * dele
...
+    // exceeds the total number of presets supported, return RESOURCE_EXHAUSTED. Note that the preset has not been appended yet.
+
+    uint8_t numberOfPendingPresets = CountNumberOfPendingPresets(delegate);
+
+    // Increment number of pending presets by 1 to account for this preset.
+    uint8_t totalExpectedCount = numberOfPendingPresets + 1;
+
+    uint8_t numberOfPresetsSupported = delegate->GetNumberOfPresets();
+
+    if (numberOfPresetsSupported == 0 || totalExpectedCount > numberOfPresetsSupported)
```
hasty, line 415: Don't really need the first part, as totalExpectedCount is at least 1, but all this is likely to get overwritten when we do scenario checking anyway. I'll approve.

hasty (approved):

src/app/clusters/thermostat-server/thermostat-server-presets.cpp
nivi-apple: done.

bzbarsky-apple (approved):

src/app/clusters/thermostat-server/thermostat-server-presets.cpp
```diff
@@ -402,6 +402,24 @@ CHIP_ERROR ThermostatAttrAccess::AppendPendingPreset(Thermostat::Delegate * dele
...
+    // exceeds the total number of presets supported, return RESOURCE_EXHAUSTED. Note that the preset has not been appended yet.
+
+    uint8_t numberOfPendingPresets = CountNumberOfPendingPresets(delegate);
+
+    // Increment number of pending presets by 1 to account for this preset.
+    uint8_t totalExpectedCount = numberOfPendingPresets + 1;
+
+    uint8_t numberOfPresetsSupported = delegate->GetNumberOfPresets();
+
+    if (totalExpectedCount > numberOfPresetsSupported)
```
bzbarsky-apple, line 415:
```suggestion
    // We will be adding one more preset, so reject if the length is already at max.
    if (numberOfPendingPresets >= numberOfPresetsSupported)
```

src/app/clusters/thermostat-server/thermostat-server-presets.cpp
```diff
@@ -402,6 +402,20 @@ CHIP_ERROR ThermostatAttrAccess::AppendPendingPreset(Thermostat::Delegate * dele
         return CHIP_IM_GLOBAL_STATUS(ConstraintError);
     }
 
+    // Before adding this preset to the pending presets, if the expected length of the pending presets' list
+    // exceeds the total number of presets supported, return RESOURCE_EXHAUSTED. Note that the preset has not been appended yet.
+
+    uint8_t numberOfPendingPresets = CountNumberOfPendingPresets(delegate);
+
+    // We will be adding one more preset, so reject if the length is already at max.
+    if (numberOfPendingPresets >= numberOfPresetsSupported)
```
bzbarsky-apple, line 411:
```suggestion
    if (numberOfPendingPresets >= delegate->GetNumberOfPresets())
```

bzbarsky-apple (approved):

ReneJosefsen (approved):

PR #35170, opened 2024-08-23
Remove writeAttributeWithEndpointID implementation from MTRDevice.

bzbarsky-apple:
This is implemented (differently) by the different subclasses.

Once this implementation is removed, removeExpectedValueForAttributePath becomes unused and can be removed.

Also removes the unused setExpectedValues declaration in MTRDevice_Internal.h and the implementations of it.


jtung-apple (approved):

bzbarsky-apple:

PR #35172, opened 2024-08-23
[NXP][rw61x][k32w1] Rework reference apps

Martin-NXP:
Changelog:

- Add multiple components for the app common code base
- Switch rw61x filesystem to NVS
- Add new rw61x revision board support
- New rw61x factory data encryption implementation
- Bump nxp_matter_support repo
- Bump ot-nxp repo
- Other bug fixes


Martin-NXP:
Hello,
Is it possible to add the submodules update label please as it is intentional ?
Thank you

chapongatien (approved):

.github/workflows/examples-nxp.yaml
```diff
@@ -150,7 +150,7 @@ jobs:
         if: github.actor != 'restyled-io[bot]'
 
         container:
-            image: ghcr.io/project-chip/chip-build-rw61x:74
```
andy31415, line 153: why the version rollback? could we use a -nxp 74 build?

I generally would not expect a version rollback in any circumstance. If we need to fix some build image, we should fix it and roll the version forward.

andy31415 (changes requested):
Request changes: do not move backward in docker image versions.

.github/workflows/examples-nxp.yaml
Martin-NXP: Hi Andy,
Thanks for the comment, I fix the docker image version

andy31415 (approved):

PR #35187, opened 2024-08-24
FS - Check VendorID and ProductID when commissioning.

cletnick:
Add a check of the Vendor ID and Product ID to match specification for requirements during reverse commissioning.


examples/fabric-admin/commands/pairing/PairingCommand.cpp
```diff
@@ -564,15 +586,46 @@ void PairingCommand::OnCurrentFabricRemove(void * context, NodeId nodeId, CHIP_E
 
 chip::Optional<uint16_t> PairingCommand::FailSafeExpiryTimeoutSecs() const
 {
-    // We don't need to set additional failsafe timeout as we don't ask the final user if he wants to continue
+    // No manual input, so do not need to extend.
     return chip::Optional<uint16_t>();
 }
 
+bool PairingCommand::ShouldWaitAfterDeviceAttestation()
```
tehampson, line 593: What is calling this?

cletnick: This is the critical point https://github.com/project-chip/connectedhomeip/blob/master/src/controller/CHIPDeviceController.cpp#L1334

examples/fabric-admin/commands/pairing/PairingCommand.cpp
```diff
@@ -564,15 +586,46 @@ void PairingCommand::OnCurrentFabricRemove(void * context, NodeId nodeId, CHIP_E
...
+                            "Detected VID: %u, Detected PID %u.",
+                            payload.vendorID, payload.productID, info.BasicInformationVendorId(), info.BasicInformationProductId());
+            deviceCommissioner->ContinueCommissioningAfterDeviceAttestation(
+                device,
+                payload.vendorID == info.BasicInformationVendorId()
+                    ? chip::Credentials::AttestationVerificationResult::kDacProductIdMismatch
+                    : chip::Credentials::AttestationVerificationResult::kDacVendorIdMismatch);
+        }
+
+        deviceCommissioner->CommissioningStageComplete(CHIP_NO_ERROR);
```
tehampson, line 625: Are we now supposed to call this because previously we returned true for `ShouldWaitAfterDeviceAttestation`

examples/fabric-admin/commands/pairing/PairingCommand.cpp
```diff
@@ -564,15 +586,46 @@ void PairingCommand::OnCurrentFabricRemove(void * context, NodeId nodeId, CHIP_E
...
-    // We don't need to set additional failsafe timeout as we don't ask the final user if he wants to continue
+    // No manual input, so do not need to extend.
     return chip::Optional<uint16_t>();
 }
 
+bool PairingCommand::ShouldWaitAfterDeviceAttestation()
+{
+    // If there is a vendor ID and product ID, request OnDeviceAttestationCompleted().
+    SetupPayload payload;
+    CHIP_ERROR err = GetPayload(mOnboardingPayload, payload);
```
tehampson, line 597: The only thing I think I would like to see as a follow up (and would help test to confirm these changes work) is that we rely on VendorID and ProductID  being populated during the CCTRL command instead of relying on parameters in GetPayload(). Mainly because if someone decided to change the payload code they won't know they have indirectly affect this code implicitly.

I think a TODO of a github issue assigned to Yufend or myself is okay as we can address that point after SVE

examples/fabric-admin/commands/pairing/PairingCommand.cpp
```diff
@@ -564,15 +586,46 @@ void PairingCommand::OnCurrentFabricRemove(void * context, NodeId nodeId, CHIP_E
...
+
 void PairingCommand::OnDeviceAttestationCompleted(chip::Controller::DeviceCommissioner * deviceCommissioner,
                                                   chip::DeviceProxy * device,
                                                   const chip::Credentials::DeviceAttestationVerifier::AttestationDeviceInfo & info,
                                                   chip::Credentials::AttestationVerificationResult attestationResult)
 {
+
+    SetupPayload payload;
+    CHIP_ERROR parse_error = GetPayload(mOnboardingPayload, payload);
+    if (parse_error == CHIP_NO_ERROR && payload.vendorID != 0 && payload.productID != 0)
```
yufengwangca, line 609: Do we need payload.vendorID != 0 && payload.productID !=  0 ?

I think we need to validate if either of them is set.

examples/fabric-admin/commands/pairing/PairingCommand.cpp
```diff
@@ -564,15 +586,46 @@ void PairingCommand::OnCurrentFabricRemove(void * context, NodeId nodeId, CHIP_E
...
+    // No manual input, so do not need to extend.
     return chip::Optional<uint16_t>();
 }
 
+bool PairingCommand::ShouldWaitAfterDeviceAttestation()
+{
+    // If there is a vendor ID and product ID, request OnDeviceAttestationCompleted().
+    SetupPayload payload;
+    CHIP_ERROR err = GetPayload(mOnboardingPayload, payload);
+    return err == CHIP_NO_ERROR && payload.vendorID != 0 && payload.productID != 0;
```
yufengwangca, line 598: Why don't we validate if the example app only sets either the vendorID or productID, but not both?

examples/fabric-admin/commands/pairing/PairingCommand.cpp
yufengwangca, line 625: You should not call deviceCommissioner->CommissioningStageComplete(CHIP_NO_ERROR); after you call ContinueCommissioningAfterDeviceAttestation

cletnick: We call this to tell deviceCommissioner that we were successful. 

ContinueCommissioningAfterDeviceAttestation() is poorly named. It will log errors even if it is told that the validation was successful. It seems specifically targeted for the failed, but platform wants to continue case.

examples/fabric-admin/commands/pairing/PairingCommand.cpp
cletnick: I don't think 0 is a valid PID or VID, so it seems like it would make more sense to require both of them?

examples/fabric-admin/commands/pairing/PairingCommand.cpp
cletnick: If anything these should be checked when RequestCommissioningApproval is received. But, I could also add a case below that fails if one is set and not the other. Thoughts?

examples/fabric-admin/commands/pairing/PairingCommand.cpp
cletnick: Thanks for catching this. Updated.

examples/fabric-admin/commands/pairing/PairingCommand.cpp
cletnick: I added the check here as well. Closing as part of this PR's scope.

yufengwangca:
We also need to enable attestation check on application side

CommissioningParameters PairingCommand::GetCommissioningParameters()
{
    auto params = CommissioningParameters();
    params.SetSkipCommissioningComplete(mSkipCommissioningComplete.ValueOr(false));

    if (mBypassAttestationVerifier.ValueOr(false))
    {
        params.SetDeviceAttestationDelegate(this);
    }
    
    .....
}

Since we always need this check in this example app, we can disable mBypassAttestationVerifier or change the default value to true.

examples/fabric-admin/commands/pairing/PairingCommand.cpp
```diff
@@ -564,15 +586,57 @@ void PairingCommand::OnCurrentFabricRemove(void * context, NodeId nodeId, CHIP_E
...
                                                   chip::Credentials::AttestationVerificationResult attestationResult)
 {
+    SetupPayload payload;
+    CHIP_ERROR parse_error = GetPayload(mOnboardingPayload, payload);
+    if (parse_error == CHIP_NO_ERROR && (payload.vendorID != 0 || payload.productID != 0))
+    {
+        if (payload.vendorID == 0 || payload.productID == 0)
+        {
+            ChipLogProgress(NotSpecified,
+                            "Failed validation of vendorID or productID must not be 0."
```
yufengwangca, line 613: This log is confusing, suggest change to:

"Failed validation: vendorID or productID must not be 0."

cletnick: Updated.

cletnick:
> We also need to enable attestation check on application side
> 
> CommissioningParameters PairingCommand::GetCommissioningParameters() { auto params = CommissioningParameters(); params.SetSkipCommissioningComplete(mSkipCommissioningComplete.ValueOr(false));
> 
> ```
> if (mBypassAttestationVerifier.ValueOr(false))
> {
>     params.SetDeviceAttestationDelegate(this);
> }
> 
> .....
> ```
> 
> }
> 
> Since we always need this check in this example app, we can disable mBypassAttestationVerifier or change the default value to true.

PTAL at the update.

examples/fabric-admin/commands/pairing/PairingCommand.cpp
yufengwangca: Could we call this function directly to end the commission process? It looks we need to call deviceCommissioner->ContinueCommissioningAfterDeviceAttestation(
            device, chip::Credentials::AttestationVerificationResult::kSuccess); in success case?

cletnick: ContinueCommissioningAfterDeviceAttestation() appears to be poorly named.  It's success case will log that attestation failed and it is commissioning anyhow. It seems it should have been named ContinueCommissioningAfterDeviceAttestationFailure(). The end result is a call to `CommissioningStageComplete(CHIP_NO_ERROR);`

The end of ContinueCommissioningAfterDeviceAttestation(): 

```
    ChipLogProgress(Controller, "Continuing commissioning after attestation failure for device ID 0x" ChipLogFormatX64,
                    ChipLogValueX64(commissioneeDevice->GetDeviceId()));

    if (attestationResult != AttestationVerificationResult::kSuccess)
    {
        ChipLogError(Controller, "Client selected error: %u for failed 'Attestation Information' for device",
                     to_underlying(attestationResult));

        CommissioningDelegate::CommissioningReport report;
        report.Set<AttestationErrorInfo>(attestationResult);
        CommissioningStageComplete(CHIP_ERROR_INTERNAL, report);
    }
    else
    {
        ChipLogProgress(Controller, "Overriding attestation failure per client and continuing commissioning");
        CommissioningStageComplete(CHIP_NO_ERROR);
    }
```

yufengwangca (approved):

examples/fabric-admin/commands/pairing/PairingCommand.cpp
yufengwangca: Per API contract by design, when OnDeviceAttestationCompleted is called (attestation failed or 
ShouldWaitAfterDeviceAttestation return true), we always call ContinueCommissioningAfterDeviceAttestation() to proceed, either in success or failure.

In success case, it is equivalent to log some messages and call CommissioningStageComplete(CHIP_NO_ERROR);

cletnick: ack. Let's conform to API spec and call ContinueCommissioningAfterDeviceAttestation(). We'll just have some noisy logs.

examples/fabric-admin/commands/pairing/PairingCommand.cpp
cletnick: Added TODO

saurabhst (approved):

PR #35194, opened 2024-08-26
[Fabric-Sync] Allow RPC ports customization in example apps

### LINKED ISSUE #35178: [BUG] Need to add support in fabric-admin and fabric-bridge to customize RPC ports
### Reproduction steps

We need to add support in fabric-admin and fabric-bridge to customize RPC ports so that we can run two sets of fabric-admin+fabric-bridge on the same host. 

### Bug prevalence

Always

### Platform

raspi

### Platform Version(s)

_No response_

### Anything else?

_No response_

arkq:
### Problem

It's not possible to run more than one set of fabric-admin+fabric-bridge on a single host.

### Changes

- optionally get RPC server ports from program arguments

### Testing

Tested locally while working on TC-MCORE-FS-1.3 update.

### DIFF
--- a/examples/fabric-admin/commands/interactive/InteractiveCommands.h
+++ b/examples/fabric-admin/commands/interactive/InteractiveCommands.h
@@ -31,8 +31,7 @@ class InteractiveCommand : public CHIPCommand
 public:
     InteractiveCommand(const char * name, Commands * commandsHandler, const char * helpText,
                        CredentialIssuerCommands * credsIssuerConfig) :
-        CHIPCommand(name, credsIssuerConfig, helpText),
-        mHandler(commandsHandler)
+        CHIPCommand(name, credsIssuerConfig, helpText), mHandler(commandsHandler)
     {
         AddArgument("advertise-operational", 0, 1, &mAdvertiseOperational,
                     "Advertise operational node over DNS-SD and accept incoming CASE sessions.");
@@ -55,7 +54,14 @@ public:
     InteractiveStartCommand(Commands * commandsHandler, CredentialIssuerCommands * credsIssuerConfig) :
         InteractiveCommand("start", commandsHandler, "Start an interactive shell that can then run other commands.",
                            credsIssuerConfig)
-    {}
+    {
+#if defined(PW_RPC_ENABLED)
+        AddArgument("fabric-bridge-server-port", 0, UINT16_MAX, &mFabricBridgeServerPort,
+                    "The fabric-bridge RPC port number to connect to (default: 33002).");
+        AddArgument("local-server-port", 0, UINT16_MAX, &mLocalServerPort,
+                    "The port number for local RPC server (default: 33001).");
+#endif
+    }
 
     /////////// CHIPCommand Interface /////////
     CHIP_ERROR RunCommand() override;
@@ -63,6 +69,11 @@ public:
 private:
     char * GetCommand(char * command);
     std::string GetHistoryFilePath() const;
+
+#if defined(PW_RPC_ENABLED)
+    chip::Optional<uint16_t> mFabricBridgeServerPort{ 33002 };
+    chip::Optional<uint16_t> mLocalServerPort{ 33001 };
+#endif
 };
 
 void PushCommand(const std::string & command);
--- a/examples/fabric-admin/commands/interactive/InteractiveCommands.h
+++ b/examples/fabric-admin/commands/interactive/InteractiveCommands.h
@@ -31,7 +31,8 @@ class InteractiveCommand : public CHIPCommand
 public:
     InteractiveCommand(const char * name, Commands * commandsHandler, const char * helpText,
                        CredentialIssuerCommands * credsIssuerConfig) :
-        CHIPCommand(name, credsIssuerConfig, helpText), mHandler(commandsHandler)
+        CHIPCommand(name, credsIssuerConfig, helpText),
+        mHandler(commandsHandler)
     {
         AddArgument("advertise-operational", 0, 1, &mAdvertiseOperational,
                     "Advertise operational node over DNS-SD and accept incoming CASE sessions.");
--- a/examples/fabric-admin/commands/interactive/InteractiveCommands.h
+++ b/examples/fabric-admin/commands/interactive/InteractiveCommands.h
@@ -24,6 +24,9 @@
 
 #include <string>
 
+constexpr uint16_t kFabricBridgeServerPort = 33002;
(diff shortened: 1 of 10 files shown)


examples/fabric-bridge-app/linux/main.cpp
```diff
@@ -59,8 +61,48 @@ constexpr uint16_t kPollIntervalMs = 100;
 constexpr uint16_t kRetryIntervalS = 3;
 #endif
 
+uint16_t gFabricAdminServerPort = 33001;
+uint16_t gLocalServerPort       = 33002;
+
 BridgedDeviceBasicInformationImpl gBridgedDeviceBasicInformationAttributes;
 
+constexpr uint16_t kOptionFabricAdminServerPortNumber = 0xFF01;
+constexpr uint16_t kOptionLocalServerPortNumber       = 0xFF02;
```
tehampson, line 70: Why is htis 0xFF01 and 0xFF02 and not 33001 and 33002

examples/fabric-bridge-app/linux/main.cpp
```diff
@@ -59,8 +61,48 @@ constexpr uint16_t kPollIntervalMs = 100;
...
+
+ArgParser::OptionDef sProgramCustomOptionDefs[] = {
+    { "fabric-admin-server-port", ArgParser::kArgumentRequired, kOptionFabricAdminServerPortNumber },
```
tehampson, line 73: If the argument is required do we need to define a default? or is there a why for this to remain optional?

tehampson: Is this just saying that if optionals are provided we need to provide both?

examples/fabric-bridge-app/linux/main.cpp
tehampson: Is this an argument identifier and not Port number?

examples/fabric-bridge-app/linux/main.cpp
arkq: This tells the parser that the argument for the option is required, i.e.:  `--option REQUIRED_ARGUMENT`

examples/fabric-admin/commands/interactive/InteractiveCommands.h
```diff
@@ -55,14 +55,26 @@ class InteractiveStartCommand : public InteractiveCommand
     InteractiveStartCommand(Commands * commandsHandler, CredentialIssuerCommands * credsIssuerConfig) :
         InteractiveCommand("start", commandsHandler, "Start an interactive shell that can then run other commands.",
                            credsIssuerConfig)
-    {}
+    {
+#if defined(PW_RPC_ENABLED)
+        AddArgument("fabric-bridge-server-port", 0, UINT16_MAX, &mFabricBridgeServerPort,
+                    "The fabric-bridge RPC port number to connect to (default: 33002).");
+        AddArgument("local-server-port", 0, UINT16_MAX, &mLocalServerPort,
```
yufengwangca, line 62: Use constant instead of hard coded number 
constexpr uint16_t kFabricBridgeServerPort = 33002;
constexpr uint16_t kFabricLocalServerPort = 33001;

examples/fabric-admin/commands/interactive/InteractiveCommands.h
```diff
@@ -55,14 +55,26 @@ class InteractiveStartCommand : public InteractiveCommand
...
 
     /////////// CHIPCommand Interface /////////
     CHIP_ERROR RunCommand() override;
 
 private:
     char * GetCommand(char * command);
     std::string GetHistoryFilePath() const;
+
+#if defined(PW_RPC_ENABLED)
+    chip::Optional<uint16_t> mFabricBridgeServerPort{ 33002 };
```
yufengwangca, line 75: Same as above

examples/fabric-admin/rpc/RpcClient.h
```diff
@@ -22,17 +22,19 @@
...
-constexpr uint16_t kFabricBridgeServerPort = 33002;
-
 /**
- * @brief Initializes the RPC client with the specified server port.
+ * @brief Sets the RPC server port to which the RPC client will connect.
  *
- * This function sets the RPC server port and starts packet processing for the RPC client.
+ * @param port The port number.
+ */
+void SetRpcClientPort(uint16_t port);
```
yufengwangca, line 30: SetRpcClientPort is confusing, since we don't have client port in RPC, we are setting the remote RPC server port to connect here

yufengwangca:
I have a general question, you customize the local and remote RPC ports from the interactive start command instead of program start args we general did in other examples. But the Start command always get called when the example starts with the default values, how you set those numbers? Via the CLI later?

examples/fabric-bridge-app/linux/RpcClient.cpp
```diff
@@ -107,9 +107,13 @@ void RpcCompletedWithEmptyResponse(const pw_protobuf_Empty & response, pw::Statu
 
 } // namespace
 
-CHIP_ERROR InitRpcClient(uint16_t rpcServerPort)
+void SetRpcClientPort(uint16_t port)
```
yufengwangca, line 110: There is no Client Port concept in RPC, we need to rename it to SetRpcRemoteServerPort

arkq:
> But the Start command always get called when the example starts with the default values, how you set those numbers? Via the CLI later?

This PR moves the RPC server/client initialization to the `interactive start` command. It's not possible to set these values after interactive session starts because by that time the RPC connection is already established.

examples/fabric-admin/commands/interactive/InteractiveCommands.h
arkq: Done

yufengwangca (approved):

examples/fabric-bridge-app/linux/main.cpp
```diff
@@ -59,8 +61,48 @@ constexpr uint16_t kPollIntervalMs = 100;
...
+    { "local-server-port", ArgParser::kArgumentRequired, kOptionLocalServerPortNumber },
+    {},
+};
+
+const char sProgramCustomOptionHelp[] = "  --fabric-admin-server-port <port>\n"
```
saurabhst, line 78: Should the README document updated with these new command args ?

saurabhst (approved):

examples/fabric-bridge-app/linux/main.cpp
arkq: Maybe as a followup. README does not cover all the options available for fabric-bridge anyway.

PR #35195, opened 2024-08-26
Add switch to disable unique_id

jlatusek:
## Problem

During tests in CI there is no way to create app without support for `UniqueId` . Such application is required to perform TC-MCORE-FS-1.3 test.

## Changes

I've addes support for option `data-model-no-unique-id` during build of apps.

## Testing 

1. Build chip-tool and lightapp

``` sh
./scripts/build/build_examples.py \
    --target linux-x64-light-data-model-no-unique-id-no-wifi \
    --target linux-x64-chip-tool \
    build
```

2. Start lightapp and chip-tool

``` sh
./out/linux-x64-chip-tool/chip-tool interactive start
./out/linux-x64-light-no-unique-id-no-wifi/chip-lighting-app
```

3. Execute commands in chip-tool
``` sh
pairing onnetwork 1 20202021
basicinformation read unique-id 1 0
```

Exptected output with UniqueId disabled:

``` log
[1724666684.106] [68952:68954] [TOO] Response Failure: IM Error 0x00000586: General error: 0x86 (UNSUPPORTED_ATTRIBUTE)
[1724666684.106] [68952:68954] [TOO] Run command failure: IM Error 0x00000586: General error: 0x86 (UNSUPPORTED_ATTRIBUTE)
```


src/app/codegen-data-model-provider/EmberMetadata.cpp
```diff
@@ -32,6 +32,13 @@ std::variant<const EmberAfCluster *,           // global attribute, data from a
              >
 FindAttributeMetadata(const ConcreteAttributePath & aPath)
 {
+#if CHIP_DISABLE_UNIQUE_ID == 1
```
arkq, line 35:
```suggestion
#if CHIP_DISABLE_UNIQUE_ID
```

scripts/build/build/targets.py
```diff
@@ -165,6 +165,7 @@ def BuildHostTarget():
     target.AppendModifier("ipv6only", enable_ipv4=False)
     target.AppendModifier("no-ble", enable_ble=False)
     target.AppendModifier("no-wifi", enable_wifi=False)
+    target.AppendModifier("no-unique-id", disable_unique_id=True)
```
arkq, line 168: Maybe more verbose option like "data-model-no-unique-id" since we already have `data-model-` options.

jlatusek: Fixed

src/app/util/ember-compatibility-functions.cpp
```diff
@@ -283,6 +283,15 @@ CHIP_ERROR ReadSingleClusterData(const SubjectDescriptor & aSubjectDescriptor, b
 
     const EmberAfCluster * attributeCluster            = nullptr;
     const EmberAfAttributeMetadata * attributeMetadata = nullptr;
+
+#if defined(CHIP_DISABLE_UNIQUE_ID) && CHIP_DISABLE_UNIQUE_ID
+    if (aPath.mEndpointId == 0 && aPath.mClusterId == chip::app::Clusters::BasicInformation::Id &&
+        aPath.mAttributeId == chip::app::Clusters::BasicInformation::Attributes::UniqueID::Id)
+    {
+        return CHIP_ERROR_IM_GLOBAL_STATUS_VALUE(UnsupportedAttributeStatus(aPath));
```
bzbarsky-apple, line 291: What about wildcard expansion and other things that look at attribute metadata?  This will lead to incorrect behavior on _write_ attempts to this attribute, for example.

The right way to do this is to have an example app configured with this attribute turned off or something, not by hacking core SDK code like this.

bzbarsky-apple, line 291: What about wildcard expansion and other things that look at attribute metadata?  This will lead to incorrect behavior on _write_ attempts to this attribute, for example.

The right way to do this is to have an example app configured with this attribute turned off or something, not by hacking core SDK code like this.

bzbarsky-apple (changes requested):

src/app/util/ember-compatibility-functions.cpp
jlatusek: Ok, I see.
Do you have any suggestions on how to do this?
I was thinking about one of the solutions:
* duplicate `examples/lighting-app/lighting-common/lighting-app.zap`,
* use the template generator to remove the `UniqueId` section from the zap file if not needed (it will be more problematic to use zap_tool)

bzbarsky-apple: Do you need to duplicate lighting-app?  Or could you just turn this attribute off in lighting-app, period?

tehampson: @jlatusek and @arkq 

I had a slack chat with Boris about this today to help describe why we need the app and the direction we should go in. His recommendation is that we create a new example app with the specific purpose of being a device that doesn't have a UniqueID based on configuration set in the zap file. So new example app called something like `no-unique-id-app`.

He mentioned (with some paraphrasing):
>The new app should:
> 1. Have the ClusterRevision set accordingly, to a revision where that's allowed.
> 2. Be protected from having that changed by the update_cluster_revision.py script (this can be a followup, probably).

tehampson:
@tehampson for my email filters to give me updates on this PR

examples/lighting-app/lighting-common/BUILD.gn
```diff
@@ -13,14 +13,19 @@
...
 import("${chip_root}/src/app/chip_data_model.gni")
 
 config("config") {
   include_dirs = [ "include" ]
 }
 
 chip_data_model("lighting-common") {
-  zap_file = "lighting-app.zap"
+  if (chip_disable_unique_id == true) {
+    zap_file = "lighting-app-data-model-no-unique-id.zap"
```
tehampson, line 25: Where is this .zap file

Also I think Boris is suggesting creating a new example examples/no-unique-id-app/

The reason why is if someone tries to build the all clusters app with the option `chip_disable_unique_id == true` that app will still have the UniqueId so the build flag is not doing what it is saying. By having a completely new example we have full control over what that app is supposed to do

jlatusek: I've added new zap and matter files without `UniqueID` in them. I thought this was enough to remove support for `UniqueID`. What additional changes are required to completely remove support for `UniqueID`? Currently the chip-tool returns information that UniqueID is not supported. 

I wanted to remove the situation with duplicate lighting apps. I thought that only differences in the zap file were required and that I could just have two versions of that file in the repo, one with uniqueID and one without.

tehampson: >I've added new zap and matter files without UniqueID in them.

Very sorry I missed the files that was added somehow in my first review 

>I wanted to remove the situation with duplicate lighting apps.

That is Boris exact suggestion, why do you want to avoid that situation? It will be very clear what the intention of the example is.

The problem currently with `chip_disable_unique_id` (as it exists in this PR) is you have an application specific build flag that doesn't make sense for any other application. With the code you have today, I would expect that if I ran `./scripts/build/build_examples.py --target linux-x64-all-clusters-data-model-no-unique-id` that the application over there should also have no unique it, but only the zapfile of lighting will accept your `chip_disable_unique_id`

jlatusek: ohhh, thanks for the clarification. In that case I will create another example application.

jlatusek: @tehampson I created another example app as you wished for. 
About `chip_disable_unique_id` it will not works for all-clusters app at all. The additional argument to gn can only be passed if light-app is compiled. I tried adding `no-unique-id` for `all-clusters` and this is what I've got:
```
 ± ./scripts/build/build_examples.py  --pw-command-launcher=ccache  --target linux-x64-all-clusters-no-wifi-data-model-no-unique-id build
INFO    Building targets: linux-x64-all-clusters-no-wifi-data-model-no-unique-id
INFO    Preparing builder 'linux-x64-all-clusters-no-wifi-data-model-no-unique-id'
INFO    Generating /mnt/dane/projects/matter/connectedhomeip-third/out/linux-x64-all-clusters-no-wifi-data-model-no-unique-id
INFO    Generating linux-x64-all-clusters-no-wifi-data-model-no-unique-id
INFO    ERROR at the command-line "--args":1:76: Build argument has no effect.
INFO    pw_command_launcher="ccache" chip_enable_wifi=false chip_disable_unique_id=true
INFO                                                                               ^---
INFO    The variable "chip_disable_unique_id" was set as a build argument
INFO    but never appeared in a declare_args() block in any buildfile.
```

arkq: >  The additional argument to gn can only be passed if light-app is compiled. I tried adding no-unique-id for all-clusters and this is what I've got:

Yes, but unfortunately the error looks like an error, not a feature :)

And duplicating the app does not necessarily mean that we have to duplicate source files. I think that on the build system level most file can be used directly from the lighting-app, only the zap needs to be explicitly specified for that new app.

bzbarsky-apple (approved):
Looks good, but we need to fix `scripts/tools/zap/update_cluster_revisions.py` to not update the ZAP file or this app, in a separate PR.

andy31415 (approved):

tehampson:
> Looks good, but we need to fix `scripts/tools/zap/update_cluster_revisions.py` to not update the ZAP file or this app, in a separate PR.

Tracking with: https://github.com/project-chip/connectedhomeip/issues/35334

PR #35206, opened 2024-08-26
darwin-framework-tool: Add support for paseonly pairing

ksperling-apple:
This includes code-paseonly and by-index-paseonly commands, as well as interacting with devices over PASE by preferring an existing MTRBaseDevice connected over PASE to creating a new CASE connection when looking up a device by node id.

PR #35209, opened 2024-08-27
Darwin: Improve CommissioningType naming

ksperling-apple:
Follow-up to #35206


ksperling-apple:
Also need to restore the semantics of WaitForCommissionee, it should only use CASE

vivien-apple (approved):

bzbarsky-apple (approved):

ksperling-apple:

PR #35224, opened 2024-08-27
Update Quiet reporting conditions for LVL and CC remaining time attribute

jmartinez-silabs:
Spec pr [10231](https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/10231) modifies the quiet reporting of the Remaining time attribute of the Level and Color control clusters.

The conditions on which the attribute must be reported are :
- When it changes from 0 to any value higher than 10, or
- When it changes, with a delta larger than 10, caused by the invoke of a command, or  
- When it changes to 0.

It also adds the Quiet report quality to the ColorTemperatureMireds attribute.
Report conditions are:
 - At most once per second, or
 - At the end of the movement/transition, or
 - When it changes from null to any other value and vice versa.

This PR reflect those changes in the SDK.

Tested on Silabs lighting-app  with cert tests TC_CC_2_2.py and TC_LVL_2_3.py

### DIFF
--- a/src/app/clusters/color-control-server/color-control-server.cpp
+++ b/src/app/clusters/color-control-server/color-control-server.cpp
@@ -3149,7 +3149,7 @@ MarkAttributeDirty ColorControlServer::SetQuietReportAttribute(QuieterReportingA
  * @brief
  * Function used to set the remaining time based on quiet reporting conditions.
  * It will update the attribute storage and report the attribute if it is determined dirty.
- * The condition on which the attribute must be reported are :
+ * The conditions on which the attribute must be reported are:
  * - When it changes from 0 to any value higher than 10, or
  * - When it changes, with a delta larger than 10, caused by the invoke of a command, or
  * - When it changes to 0.
--- a/src/app/clusters/color-control-server/color-control-server.cpp
+++ b/src/app/clusters/color-control-server/color-control-server.cpp
@@ -1040,7 +1040,7 @@ void ColorControlServer::startColorLoop(EndpointId endpoint, uint8_t startFromSt
     colorHueTransitionState->transitionTime = MAX_INT16U_VALUE;
     colorHueTransitionState->endpoint       = endpoint;
 
-    SetQuietReportRemainingTime(endpoint, MAX_INT16U_VALUE);
+    SetQuietReportRemainingTime(endpoint, MAX_INT16U_VALUE, true /* isNewTransition */);
 
     scheduleTimerCallbackMs(configureHSVEventControl(endpoint), TRANSITION_UPDATE_TIME_MS.count());
 }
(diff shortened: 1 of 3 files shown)


src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -3139,23 +3129,40 @@ MarkAttributeDirty ColorControlServer::SetQuietReportAttribute(QuieterReportingA
  * @brief
  * Function used to set the remaining time based on quiet reporting conditions.
  * It will update the attribute storage and report the attribute if it is determined dirty.
- * The condition on which the attribute must be reported are defined by the set QuieterReportingPolicyFlags
- * of the quietRemainingTime object and the implicit conditions of the QuieterReportingAttribute class
+ * The condition on which the attribute must be reported are :
```
bzbarsky-apple, line 3150:
```suggestion
 * The conditions on which the attribute must be reported are:
```

src/app/clusters/color-control-server/color-control-server.cpp
```diff
@@ -3157,23 +3147,40 @@ MarkAttributeDirty ColorControlServer::SetQuietReportAttribute(QuieterReportingA
...
+
+    auto predicate =
+        [isNewTransition, lastRemainingTime](
+            const typename QuieterReportingAttribute<uint16_t>::SufficientChangePredicateCandidate & candidate) -> bool {
+        constexpr uint16_t reportDelta = 10;
+        if (candidate.newValue.Value() == 0 || (candidate.lastDirtyValue.Value() == 0 && candidate.newValue.Value() > reportDelta))
+        {
+            return true;
+        }
+        else if (isNewTransition &&
```
bzbarsky-apple, line 3174: So if we just keep sending commands that increase the value by 2, we can make it arbitrarily large without any reporting happening?

jmartinez-silabs: I guess yes. 

> When it changes, with a delta larger than 10, caused by the invoke of a command, or
How would you express it in spec and then in implementation?

Essentially, the report of RemainingTime right now would occur only the reception of a Command which increases the current Remaining time by more than 10 and when it reaches 0.

jmartinez-silabs: Maybe it should also be if the remaining time Increases by a value greater than 10 compared to the last reported value

bzbarsky-apple: I haven't had time to think this through.  Please check with whoever wrote the spec text?

ReneJosefsen: The proposed spec text was mainly to catch any major jumps in transition time, but as pointed out it does not cover the case where a client would slowly ramp up the transition time. @bzbarsky-apple if you have any inputs for this it is much appreciated every single iteration of this I have made, have had some edge case where a report might not be transmitted. 😔

ReneJosefsen:
I was testing this change while updating the script to match the latest test plan and I had an issue with RemainingTime when after it has transitioned to use MoveToColorTemperature command, after a bit of digging, it seems like we have missed a SetQuietReportRemainingTime in the moveToColorTemp command, @jmartinez-silabs thoughts?

jmartinez-silabs:
> I was testing this change while updating the script to match the latest test plan and I had an issue with RemainingTime when after it has transitioned to use MoveToColorTemperature command, after a bit of digging, it seems like we have missed a SetQuietReportRemainingTime in the moveToColorTemp command, @jmartinez-silabs thoughts?

@ReneJosefsen it should be called in the timed event callback that updates the colortempmired attributes. Maybe it is missing. I'll confirm tomorrow/fix it if need be in this pr that I need to update anyway

src/app/clusters/color-control-server/color-control-server.cpp
tcarmelveilleux: it should be related to "10 seconds since last time the value was marked as reportable" or such, maybe.

jmartinez-silabs: In my latest commit, I set the condition for new Commands to report a the remainingTime if there is a delta > 10 between the current RemainingTime or if the new RemainingTime is > the last Reported value + 10.

tcarmelveilleux (approved):

bzbarsky-apple (approved):

PR #35231, opened 2024-08-27
[Fix] Replaced automatable PICS in SEAR tests

j-ororke:
- Replaced automatable PICS in SEAR 1.3, 1.5, and 1.6 test modules with attributes gathered from endpoint.

### Testing


ReneJosefsen:
Does this align with the Test Plan? I think this is a good addition but it needs to be aligned with the Test Plan and given the timing, this might be something we need to postpone to 1.5?

src/python_testing/TC_SEAR_1_6.py
```diff
@@ -72,7 +72,7 @@ async def read_progress(self, step):
         return progress
 
     def TC_SEAR_1_6(self) -> list[str]:
-        return ["SEAR.S", "SEAR.S.A0005", "SEAR.S.A0000", "SEAR.S.A0002", "SEAR.S.M.HAS_MANUAL_OPERATING_STATE_CONTROL"]
+        return ["SEAR.S", "SEAR.S.M.HAS_MANUAL_OPERATING_STATE_CONTROL"]
```
cecille, line 75: Actually, let's just leave this test as-is for now. The top level pics we can deal with separately and this will mess up the TH

src/python_testing/TC_SEAR_1_5.py
```diff
@@ -89,7 +89,7 @@ async def send_cmd_skip_area_expect_response(self, step, skipped_area, expected_
                              f"Command response ({ret.status}) doesn't match the expected one")
 
     def TC_SEAR_1_5(self) -> list[str]:
-        return ["SEAR.S", "SEAR.S.C02.Rsp"]
+        return ["SEAR.S"]
```
cecille, line 92: Same here - let's leave the top level PICS as-is for now and we'll deal with the top level PICS separately.

src/python_testing/TC_SEAR_1_6.py
j-ororke: Thank you,
This has been updated to leave the top level PICS as-is and should now be resolved.

src/python_testing/TC_SEAR_1_5.py
j-ororke: This has been updated to leave the top level PICS as-is.

src/python_testing/TC_SEAR_1_6.py
```diff
@@ -94,6 +94,17 @@ async def test_TC_SEAR_1_6(self):
                 asserts.fail("The --app-pid flag must be set when PICS_SDK_CI_ONLY is set")
             self.app_pipe = self.app_pipe + str(app_pid)
 
+        attributes = Clusters.ServiceArea.Attributes
+
+        if not await self.attribute_guard(endpoint=self.endpoint, attribute=attributes.SupportedAreas):
```
cecille, line 99: Given that we now have the top-level guards, this should be changed over to use those.

Also, I'm pretty sure the pics_ function in this test is incorrect - it looks like the function on line 83 is suppose to be pics_TC_SEAR_1_6, which is probably why these were added - the pics definition seems wrong.

j-ororke: I have changed the function name for the pics to be pics_TC_SEAR_1_* based on the test being run so that the pics can be found during testing if they are needed.

PR #35240, opened 2024-08-28
Generate ChipError.java for Android platform

yhoyoon:
This change introduces a new Python script, `gen_chip_error.py`, which generates the `ChipError.java` file for the Android platform. The script processes header files containing error definitions and creates corresponding Java constants and documentation comments. It is based on the existing `error_table.py` script.

Additionally, this update includes the following changes:
- Update `.github/workflows/zap_templates.yaml` to include the generation of `ChipError.java` during the CI process.
- Exclude the generated `ChipError.java` file from Restyled processing by updating `.restyled.yaml`.
- Modify `AndroidChipPlatform-JNI.cpp` to add a JNI method for converting error codes to strings.

For example, the `ChipError` enum class can be used in the following way within the `onPairingComplete` callback function:

```kt
override fun onPairingComplete(errorCode: Long) {
  val chipError = ChipError.value(errorCode)
  logger.log(Level.INFO, "onPairingComplete with error code: $chipError")
  if (chipError != ChipError.CHIP_NO_ERROR) {
    setFailure("onPairingComplete failure")
  }
}
```


joonhaengHeo:
I think enum class would be more appropriate.

joonhaengHeo:
And, this class can be used in Java controller and kotlin controller. It is necessary to review whether the current location is appropriate.

@yunhanw-google @yufengwangca 
Please check this PR.

yhoyoon:
> I think enum class would be more appropriate.

Thank you, I will change the ChipError to enum class.

> And, this class can be used in Java controller and kotlin controller. It is necessary to review whether the current location is appropriate.
> 
> @yunhanw-google @yufengwangca
> Please check this PR.

When reviewing the appropriate location, could you also consider [virtual-device-app](https://github.com/project-chip/connectedhomeip/tree/master/examples/virtual-device-app/android)? It is a commissionee app based on Android platform.

src/platform/android/java/chip/platform/ChipError.java
```diff
@@ -0,0 +1,304 @@
...
+    BLE_ERROR_INVALID_BTP_HEADER_FLAGS(1053L),
+    BLE_ERROR_INVALID_BTP_SEQUENCE_NUMBER(1054L),
+    BLE_ERROR_REASSEMBLER_INCORRECT_STATE(1055L),
+
+    /**
+     * IM Global errors
+     *
+     * These errors are defined in src/protocols/interaction_model/StatusCodeList.h
+     */
+    SUCCESS(1280L),
```
yunhanw-google, line 250: could you migrate src/controller/java/src/chip/devicecontroller/model/Status.java, update file in https://github.com/project-chip/connectedhomeip/blob/master/src/protocols/interaction_model/StatusCodeList.h#L27C28-L27C91

yunhanw-google, line 250: why it is 1280L instead of 0L?

src/platform/android/java/chip/platform/ChipError.java
```diff
@@ -0,0 +1,304 @@
...
+package chip.platform;
+
+public enum ChipError {
+
+    /**
+     * SDK Core errors
+     *
+     * These errors are defined in src/lib/core/CHIPError.h
+     */
+    CHIP_NO_ERROR(0L),
```
bzbarsky-apple, line 27: Note that these error codes are subject to change, so I am hoping this is not intended as public API...

yunhanw-google: @bzbarsky-apple  in darwin, how do we expose/express the matter error to the public?

bzbarsky-apple: See https://github.com/project-chip/connectedhomeip/blob/master/src/darwin/Framework/CHIP/MTRError.h

and there's a mapping from internal error values to public API ones.

src/platform/android/java/chip/platform/ChipError.java
yhoyoon: I understand. Since the `src/controller/java/src/chip/devicecontroller/model/Status.java` file already exists, I will modify the script to update the existing file instead of adding IM status codes to the `ChipError` class.

yhoyoon: The reason for this is that IM Status Codes are represented by a separate range in `CHIP_ERROR`. As mentioned in the review above, I will delete the IM StatusCodeList from the `ChipError.java` file and update the `src/controller/java/src/chip/devicecontroller/model/Status.java` file instead.

https://github.com/project-chip/connectedhomeip/blob/708558881bf6e5b0dac3c09489124907927b6cd2/src/app/MessageDef/StatusIB.cpp#L138-L151

src/platform/android/java/chip/platform/ChipError.java
yhoyoon: @bzbarsky-apple Thank you for sharing an example on the Darwin platform. Could you please provide a specific example to illustrate the advantage of mapping MTRErrors and CHIP_ERRORs?

For your information, the Android ChipDeviceController already exposes a CHIP_ERROR.
https://github.com/project-chip/connectedhomeip/blob/708558881bf6e5b0dac3c09489124907927b6cd2/src/controller/java/src/chip/devicecontroller/ChipDeviceController.java#L1824-L1837

src/platform/android/java/chip/platform/ChipError.java
yhoyoon: @yunhanw-google , I just pushed a change. If this was your intention, I will update the remaining files as well.

- src/controller/python/chip/interaction_model/__init__.py
- src/controller/java/src/matter/controller/model/Status.kt

src/platform/android/java/chip/platform/ChipError.java
bzbarsky-apple: All I'm saying is that CHIP_ERROR names (especially) and values (if the old value stops being used for anything in the SDK) are subject to change as needed.  If you want to expose them in public API, then you should probably have a migration plan for when those things happen.

tcarmelveilleux: @yhoyoon The recommendation is to:

- Not allocate specific IDs that match between CHIP_ERROR and the equivalent Java enum, just keep the names
- Make a factory class implementation that generates the opaque enum (i.e. no match between the C++ side and Java side based on numbers and NEVER remove an enum value) from the CHIP_ERROR. The implementation can change and be regenerated over time.

andy31415:
@yhoyoon what do we do with this PR? It is quite old now and I am unsure about the path taken here when considering backwards compatibility. Can we close it?

PR #35249, opened 2024-08-28
Fix --int-arg, --bool-arg, etc options parsing

arkq:
### Problem

Multi-argument options are non-intuitive (even our manual did not document it right: https://github.com/project-chip/connectedhomeip/blob/master/docs/testing/python.md#running-tests). Instead of that, use multi-options (which aligns with the manual) while preserving possibility to use multi-argument option style to be compatible with TH.

### Testing

CI will verify potential failures.


tcarmelveilleux (changes requested):
Please fix after SVE since the `test_parameters` config in TH is an arguments dict which means if `--string-arg a:foo b:far` shows as:

```
{
   "string-args": "a:foo b:bar"
}
```

Therefore:
```
{
   "string-arg": "a:foo",
   "string-arg": "b:bar"
}
```

would lead to only `b:bar` existing :( 


if we split, then TH can no longer support any tests that has multiple options of same type :(. This is horrendous, but it's more distruptive to change it right now than to leave it.

fabiowmm:
> Please fix after SVE since the `test_parameters` config in TH is an arguments dict which means if `--string-arg a:foo b:far` shows as:
> 
> ```
> {
>    "string-args": "a:foo b:bar"
> }
> ```
> 
> Therefore:
> 
> ```
> {
>    "string-arg": "a:foo",
>    "string-arg": "b:bar"
> }
> ```
> 
> would lead to only `b:bar` existing :(
> 
> if we split, then TH can no longer support any tests that has multiple options of same type :(. This is horrendous, but it's more distruptive to change it right now than to leave it.

This is correct. We can add the changes needed in TH to our backlog, but I defer to @raju-apple regarding the priority.

arkq:
@tcarmelveilleux I've updated this PR little bit. Now it will support both option types: multi-argument and multi-option. Also, I've added a unit test to verify that both these behaviors work as expected. What do you think about such compromise?

tcarmelveilleux (approved):

jmartinez-silabs (approved):

PR #35253, opened 2024-08-28
Updated yaml script as per the SQA issues

Saravana-kr22:
1.  Fix for https://github.com/project-chip/matter-test-scripts/issues/271

    -  DRLK-2.6, 2.7, 2.8, 2.10, 2.11 - Yaml script has been updated as per the Test-Plan and Verfication steps Document.
    
2. Fix for https://github.com/project-chip/matter-test-scripts/issues/306

    - LVL-9.1 - Yaml script has been updated as per the latest Test-plan.
    
3. Fix for https://github.com/project-chip/matter-test-scripts/issues/290

    - LVL-8.1 - Yaml script has been newly added as per the Test-plan.
    
 4. IDM- 3.2, 5.2, 6.1 -Yaml script has been updated as per the Verfication step Document.


src/app/tests/suites/certification/Test_TC_DRLK_2_6.yaml
```diff
@@ -107,40 +107,40 @@ tests:
...
       arguments:
           values:
               - name: "HolidayIndex"
-                value: 15
+                value: NumberOfHolidaySchedulesSupportedValue + 1
       response:
           values:
               - name: "HolidayIndex"
-                value: 15
+                value: NumberOfHolidaySchedulesSupportedValue + 1
```
tcarmelveilleux, line 121: Don't check the index, it's not defined how it would get returned on error, and it's not required by test plan. Only check status.

src/app/tests/suites/certification/Test_TC_DRLK_2_6.yaml
```diff
@@ -107,40 +107,40 @@ tests:
...
               - name: "Status"
                 value: 133
 
     - label:
-          "Step 6: TH sends Get Holiday Schedule Command to DUT with the
-          HolidayIndex as 10 (value is in the the range of step 1 but Holiday
-          Schedule entry not available)"
+          "Step 6: TH sends GetHolidaySchedule Command to DUT with the
+          HolidayIndex as any value that is different from 1 and value ⇐
```
tcarmelveilleux, line 127:
```suggestion
          HolidayIndex as any value that is different from 1 and value <=
```

tcarmelveilleux: The arrow is a bug in the test plan render. The source says `<=`

tcarmelveilleux (approved):

src/app/tests/suites/certification/Test_TC_DRLK_2_6.yaml
```diff
@@ -107,40 +107,38 @@ tests:
...
       command: "GetHolidaySchedule"
       arguments:
           values:
               - name: "HolidayIndex"
-                value: 15
+                value: NumberOfHolidaySchedulesSupportedValue + 1
       response:
           values:
-              - name: "HolidayIndex"
-                value: 15
```
bzbarsky-apple, line 121: So...  The spec says two things that are mutually contradictory here:

1. HolidayIndex SHALL be set to the value in the request command (even if that's out of range, as far as I can tell).
2. HolidayIndex constraint is "1 to NumberOfHolidaySchedulesSupported".

We need to get the spec fixed here, probably by fixingthat constraint to be something like "all".

In the meantime, the test plan should in fact be checking that the value that came back is `NumberOfHolidaySchedulesSupportedValue + 1`, since the spec does require that.

src/app/tests/suites/certification/Test_TC_DRLK_2_6.yaml
```diff
@@ -107,40 +107,38 @@ tests:
...
               - name: "Status"
                 value: 133
```
bzbarsky-apple, line 121:
```suggestion
                value: 0x85
```
would be more readable, since spec lists these as hex.

src/app/tests/suites/certification/Test_TC_DRLK_2_6.yaml
```diff
@@ -107,40 +107,38 @@ tests:
...
               - name: "HolidayIndex"
-                value: 10
+                value: NumberOfHolidaySchedulesSupportedValue
       response:
           values:
               - name: "HolidayIndex"
-                value: 10
+                value: NumberOfHolidaySchedulesSupportedValue
               - name: "Status"
                 value: 139
```
bzbarsky-apple, line 138:
```suggestion
                value: 0x8B
```

src/app/tests/suites/certification/Test_TC_DRLK_2_7.yaml
```diff
@@ -367,23 +343,24 @@ tests:
...
           values:
-              - name: "YearDayIndex"
-                value: 1
               - name: "UserIndex"
                 value: 1
 
-    - label: "Cleanup the created user with UserIndex 1"
-      command: "ClearUser"
-      timedInteractionTimeoutMs: 10000
+    - label: "Clear a year day schedule for the first user"
```
bzbarsky-apple, line 355: I don't see this in the test plan.  Why is it here?

src/app/tests/suites/certification/Test_TC_DRLK_2_7.yaml
```diff
@@ -367,23 +343,24 @@ tests:
...
+      command: "ClearYearDaySchedule"
       arguments:
           values:
-              - name: "UserIndex"
+              - name: "YearDayIndex"
                 value: 1
+              - name: "UserIndex"
+                value: 5
 
     - label: "Cleanup the created user with UserIndex 5"
```
bzbarsky-apple, line 365: Again, this is not in the test plan.  Why is it here?

src/app/tests/suites/certification/Test_TC_DRLK_2_6.yaml
Saravana-kr22: @bzbarsky-apple  yaml script is updated as per the suggestion.

src/app/tests/suites/certification/Test_TC_DRLK_2_6.yaml
Saravana-kr22: @bzbarsky-apple  yaml script is updated as per the suggestion.

src/app/tests/suites/certification/Test_TC_DRLK_2_7.yaml
Saravana-kr22: @bzbarsky-apple Updated the step 14 and removed this step.

src/app/tests/suites/certification/Test_TC_DRLK_2_7.yaml
Saravana-kr22: @bzbarsky-apple This step is added to cleanup the userindex 5 created in the step 7 to avoid the error. Now the test-plan is updated to clear all the users PR: https://github.com/CHIP-Specifications/chip-test-plans/pull/4571/files#diff-450e09b9acf3e7226af243b14d01c5bcc6f7a3acb5e39bda2726f22ba8dfcbdcR2165. So the step 14 is updated and removed this clean up step.

bzbarsky-apple (approved):

src/app/tests/suites/certification/Test_TC_DRLK_2_6.yaml
jrhees-cae: I think you are referring to the HolidayIndex in the GetHolidaySchedule response -- and the specific thing you're asking is: IF the GetHolidaySchedule request includes a HolidayIndex that is invalid, what shall the HolidayIndex be in the response...

I'm not sure how the SDK currently behaves, but IMO what should be done here is to change the spec for the HolidayIndex in the **response** to be **all**

bzbarsky-apple: That would be consistent with the SDK behavior, yes.  The SDK just echoes back whatever index came in, including if it's invalid.

PR #35256, opened 2024-08-28
NetworkCommissioning: Disconnect previous network when trying a new

DejinChen:
Ref to https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/10335.

#### Change overview
- If a Node receives a ConnectNetwork command, it SHALL disconnect from any operational networks (have removed the config) managed by Network Commissioning cluster. For example, NetworkCommissioning cluster on endpoint 0 has connected a Wi-Fi network during first attempt, but it failed in the following commissioning steps. When NetworkCommissioning cluster on endpoint 65534 tries to connect a Thread network, it should disconnect Wi-Fi network first.


src/app/clusters/network-commissioning/network-commissioning.h
```diff
@@ -81,6 +83,13 @@ class Instance : public CommandHandlerInterface,
     void SendNonConcurrentConnectNetworkResponse();
 #endif
 
+    // Array for recording NetworkCommissioningInstance.
+    static Instance * NetworkRecord[CHIP_MAX_NETWORK_TYPES];
```
bzbarsky-apple, line 87: Please follow naming conventions for statics; should be something like `sNetworkRecord`.

That said:

1. Why the hard cap at 3?
2. What ensures that these pointers don't dangle if network commissioning instances are dynamically added/destroyed?

It might be better to have Instances live in a linked list of some sort, and remove themselves from it on destruction, etc.

DejinChen: Use linked list instead.

src/app/clusters/network-commissioning/network-commissioning.cpp
```diff
@@ -1150,6 +1152,60 @@ void Instance::HandleQueryIdentity(HandlerContext & ctx, const Commands::QueryId
...
+    while (curr)
+    {
+        if (curr->mInstance == currentNetwork)
+        {
+            return;
+        }
+        curr = curr->next;
+    }
+
+    InstanceRecord * newInstanceRecord = reinterpret_cast<InstanceRecord *>(Platform::MemoryAlloc(sizeof(InstanceRecord)));
```
bzbarsky-apple, line 1169: Why do we need the heap alloc?  Why not just have Instance have an mNext pointer?

DejinChen: Use mNext pointer instead.

src/app/clusters/network-commissioning/network-commissioning.cpp
```diff
@@ -1150,6 +1152,58 @@ void Instance::HandleQueryIdentity(HandlerContext & ctx, const Commands::QueryId
...
+            bool haveConnectedNetwork = false;
+            EnumerateAndRelease(curr->mpBaseDriver->GetNetworks(), [&](const Network & network) {
+                if (network.connected)
+                {
+                    haveConnectedNetwork = true;
+                    return Loop::Break;
+                }
+                ++networkNo;
+                return (networkNo == maxNetworks) ? Loop::Break : Loop::Continue;
+            });
```
ksperling-apple, line 1191:
```suggestion
            bool haveConnectedNetwork = false;
            EnumerateAndRelease(curr->mpBaseDriver->GetNetworks(), [&](const Network & network) {
                if (network.connected)
                {
                    haveConnectedNetwork = true;
                    return Loop::Break;
                }
                return Loop::Continue;
            });
```

I don't think there is a need to do manual counting of networks etc here. EnumerateAndRelease will automatically stop iterating when the list is exhausted.

src/app/clusters/network-commissioning/network-commissioning.cpp
```diff
@@ -1150,6 +1152,58 @@ void Instance::HandleQueryIdentity(HandlerContext & ctx, const Commands::QueryId
...
+
+            if (!haveConnectedNetwork)
```
ksperling-apple, line 1193:
```suggestion
            // If none of the configured networks is `connected`, we may have a
            // lingering connection to a different network that we need to disconnect.
            if (!haveConnectedNetwork)
```

src/app/clusters/network-commissioning/network-commissioning.cpp
```diff
@@ -1150,6 +1152,58 @@ void Instance::HandleQueryIdentity(HandlerContext & ctx, const Commands::QueryId
 }
 #endif // CHIP_DEVICE_CONFIG_ENABLE_WIFI_PDC
 
+Instance * Instance::sInstanceRecord = nullptr;
```
ksperling-apple, line 1155: We should just use an `IntrusiveList` here instead of hand-rolling linked-list code, and also remove the cluster instance in `Shutdown`.

src/app/clusters/network-commissioning/network-commissioning.cpp
```diff
@@ -1150,6 +1152,58 @@ void Instance::HandleQueryIdentity(HandlerContext & ctx, const Commands::QueryId
...
+            });
+
+            if (!haveConnectedNetwork)
+            {
+                /* It's different to check the media driver connection state here.
+                 * Check the `Connected` flag in `NetworkInfoStruct` in stead, if not set,
+                 * exec `DisconnectFromNetwork` for driver. However, the physical media connection
+                 * may not necessarily exist in this case, such as after initialization.
+                 */
+                curr->mpWirelessDriver->DisconnectFromNetwork();
```
ksperling-apple, line 1200: Error handling?

src/include/platform/NetworkCommissioning.h
```diff
@@ -218,6 +218,11 @@ class WirelessDriver : public Internal::BaseDriver
      */
     virtual CHIP_ERROR RevertConfiguration() = 0;
 
+    /**
+     * @brief Disconnect from network, but maybe it is not connected on media driver.
+     */
```
ksperling-apple, line 223: We could guard this with `CHIP_DEVICE_CONFIG_SUPPORTS_CONCURRENT_CONNECTION`, but I think the default should be to return CHIP_ERROR_NOT_IMPLEMENTED.

src/platform/OpenThread/GenericNetworkCommissioningThreadDriver.h
```diff
@@ -106,6 +106,7 @@ class GenericThreadDriver final : public ThreadDriver
 
     CHIP_ERROR CommitConfiguration() override;
     CHIP_ERROR RevertConfiguration() override;
+    CHIP_ERROR DisconnectFromNetwork() override;
```
ksperling-apple, line 109: Nit: putting this after `ConnectNetwork` below would be a more logical grouping.

src/app/clusters/network-commissioning/network-commissioning.cpp
DejinChen: Done.

src/include/platform/NetworkCommissioning.h
DejinChen: Done.

src/platform/OpenThread/GenericNetworkCommissioningThreadDriver.h
DejinChen: Done.

src/app/clusters/network-commissioning/network-commissioning.cpp
DejinChen: `DisconnectFromNetwork` failed may not affect subsequent connections to other networks. So I'm not sure if we should handle errors here. Use `ReturnOnFailure` temporarily.

src/app/clusters/network-commissioning/network-commissioning.cpp
DejinChen: I have tried IntrusiveList, but i'm confused how to resolve `curr != currentNetwork` in `DisconnectOtherNetworks` func with IntrusiveList. The `IteratorBase` from `IntrusiveList` can not compared with the `Instance*`. Any suggestions?

ksperling-apple: I've tweaked the code to use InstrusiveList in this sub-PR: https://github.com/DejinChen/connectedhomeip/pull/1

ksperling-apple (approved):

bzbarsky-apple (approved):

chshu (approved):

PR #35258, opened 2024-08-28
Add DirectModeChange feature to RVC Run Mode and RVC Clean Mode cluster XML and all-clusters-app implementation

bzbarsky-apple:
Only the first commit here is manually written.  The rest are auto-generated.

Adding skip-protocol-compatibility, because kNoFeatures is not in fact visible on the protocol level (and should not really have been there to start with, probably, except the Python codegen required _some_ bit).

### DIFF
--- a/examples/all-clusters-app/all-clusters-common/src/rvc-modes.cpp
+++ b/examples/all-clusters-app/all-clusters-common/src/rvc-modes.cpp
@@ -16,6 +16,7 @@
  *    limitations under the License.
  */
 #include <app-common/zap-generated/attributes/Accessors.h>
+#include <lib/support/TypeTraits.h>
 #include <rvc-modes.h>
 #include <rvc-operational-state-delegate-impl.h>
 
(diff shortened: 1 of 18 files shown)


kiel-apple (approved):

kiel-apple (approved):

examples/all-clusters-app/all-clusters-common/src/rvc-modes.cpp
```diff
@@ -41,12 +42,15 @@ void RvcRunModeDelegate::HandleChangeToMode(uint8_t NewMode, ModeBase::Commands:
 {
     uint8_t currentMode = mInstance->GetCurrentMode();
 
-    // Our business logic states that we can only switch into a running mode from the idle state.
-    if (NewMode != RvcRunMode::ModeIdle && currentMode != RvcRunMode::ModeIdle)
+    if (!gRvcRunModeInstance->HasFeature(static_cast<ModeBase::Feature>(RvcRunMode::Feature::kDirectModeChange)))
```
plauric, line 45: @bzbarsky-apple are you planning to make similar updates to the rvc-app example application?

bzbarsky-apple: @plauric I would like to, but that one seemed more complicated and I need a bit more time to sort out what the right thing to do there is.

plauric (approved):

PR #35268, opened 2024-08-28
[Fabric-Admin] Fix crash when try to read UniqueId from a legacy device

yufengwangca:
The DUT crash when running TC_MCORE_FS_1_3.py

When try to read the UniqueId attribute from a legacy device which does not have the UniqueId, we should not try to decode it if the status is not success.

### DIFF
--- a/examples/fabric-admin/device_manager/DeviceSynchronization.cpp
+++ b/examples/fabric-admin/device_manager/DeviceSynchronization.cpp
@@ -72,6 +72,13 @@ void DeviceSynchronizer::OnAttributeData(const ConcreteDataAttributePath & path,
     VerifyOrDie(path.mEndpointId == kRootEndpointId);
     VerifyOrDie(path.mClusterId == Clusters::BasicInformation::Id);
 
+    CHIP_ERROR error = status.ToChipError();
+    if (CHIP_NO_ERROR != error)
+    {
+        ChipLogError(NotSpecified, "Response Failure: %" CHIP_ERROR_FORMAT, error.Format());
+        return;
+    }
+
     switch (path.mAttributeId)
     {
     case Clusters::BasicInformation::Attributes::UniqueID::Id:


tehampson (approved):

examples/fabric-admin/device_manager/DeviceSynchronization.cpp
```diff
@@ -72,6 +72,13 @@ void DeviceSynchronizer::OnAttributeData(const ConcreteDataAttributePath & path,
     VerifyOrDie(path.mEndpointId == kRootEndpointId);
     VerifyOrDie(path.mClusterId == Clusters::BasicInformation::Id);
 
+    CHIP_ERROR error = status.ToChipError();
```
tehampson, line 75: Can you help me understand something: Is this an issue for wildcard reads of the BINFO cluster? If so I would like to understand what changed, because this was working without issues a couple weeks ago even for commissioning a device that didn't have a UID.

Is this status set to an error only for the single attribute of the wild card read, or for all of the attributes

saurabhst (approved):

examples/fabric-admin/device_manager/DeviceSynchronization.cpp
yufengwangca: This is a regression after we refactor to DeviceSynchronization. 

See we have this check in:
https://github.com/project-chip/connectedhomeip/blob/master/examples/fabric-admin/commands/clusters/ReportCommand.cpp#L30

tehampson: Which refactor of `DeviceSynchronization` are you referring to? We had `DeviceSynchronization` in TE2 where we passed MCORE_1_3 (at the time of TE2 I am pretty sure Andrei commissioned TH_SERVER with no UID and was able to validate that DUT created a UID on TH_SERVERs behalf

yufengwangca: I am taking about OnAttributeData in DeviceSynchronization, do we read UID from the TH_SERVER without UID in TE2?

From the current implementation, it should 100% crash if we try to decode TLV data if status is not success.

```
void DeviceSynchronizer::OnAttributeData(const ConcreteDataAttributePath & path, TLV::TLVReader * data, const StatusIB & status)
{
    VerifyOrDie(path.mEndpointId == kRootEndpointId);
    VerifyOrDie(path.mClusterId == Clusters::BasicInformation::Id);

    switch (path.mAttributeId)
    {
    case Clusters::BasicInformation::Attributes::UniqueID::Id:
        mCurrentDeviceData.has_unique_id =
            SuccessOrLog(data->GetString(mCurrentDeviceData.unique_id, sizeof(mCurrentDeviceData.unique_id)), "UniqueId");
        break;
```

PR #35271, opened 2024-08-29
Add state transition feature

stingchang:
The Robotic Vacuum Clean is a stateful device. 
This PR is to enable the state transition for RVC Chef device. 
When device is activated, i.e. set to "running", the countdown starts.

Example countdown time is 30 seconds. When countdown ends, the device switch to "stopped".


examples/chef/common/chef-rvc-operational-state-delegate.cpp
```diff
@@ -62,10 +62,26 @@ CHIP_ERROR RvcOperationalStateDelegate::GetOperationalPhaseAtIndex(size_t index,
 
 void RvcOperationalStateDelegate::HandlePauseStateCallback(GenericOperationalError & err)
 {
```
tcarmelveilleux, line 64: Please use the `GetInstance()->UpdateCountdownTimeFromDelegate();` calls as seen in examples/all-clusters-app/all-clusters-common/src/operational-state-delegate-impl.cpp and match when they are called, otherwise there will not be proper management of countdown time.

tcarmelveilleux (approved):

examples/chef/common/chef-rvc-operational-state-delegate.cpp
stingchang: added, thank you

yunhanw-google (approved):

PR #35274, opened 2024-08-29
Split TC_MCORE_FS_1_3 to align with the Test Spec

### LINKED ISSUE #35002: [Tooling]  Integrate Fabric-Sync sync device flow into CI using python test runner.

To improve our quality gate, I'd like to include a full Fabric-Sync flow check in CI. This setup requires running four processes (Fabric-Admin and Fabric-Bridge on each sides). 

When start the apps, use --KVS with different file locations and --secured-device-port with different port numbers. If you just fire up a separate process for each, that's equivalent to having two network connected devices.

### Platform

other

### Platform Version(s)

_No response_

### Type

CI tested

_No response_

### Anything else?

The latest procedure is at 
https://github.com/project-chip/connectedhomeip/pull/34993

arkq:
### Problem

Test plan for TC_MCORE_FS_1_3 is going to be split into TC_MCORE_FS_1_3 and TC_MCORE_FS_1_4 https://github.com/CHIP-Specifications/chip-test-plans/pull/4575

Closes #35002

### Testing

This PR contains CI-ready infrastructure for TC_MCORE_FS_1_3 and TC_MCORE_FS_1_4 automation. One can run the test automatically as follows:

```sh
./scripts/build/build_examples.py --target linux-x64-light-no-ble build   # Lighting app needs to have UniqueID disabled
./scripts/build/build_examples.py --target linux-x64-fabric-admin-rpc build
./scripts/build/build_examples.py --target linux-x64-fabric-bridge-no-ble-rpc build

echo "LIGHTING_APP_NO_UNIQUE_ID: out/linux-x64-light-no-ble/chip-lighting-app" >> test-env.yaml
echo "FABRIC_ADMIN_APP: out/linux-x64-fabric-admin-rpc/fabric-admin" >> test-env.yaml
echo "FABRIC_BRIDGE_APP: out/linux-x64-fabric-bridge-no-ble-rpc/fabric-bridge-app" >> test-env.yaml

./scripts/tests/run_python_test.py --script src/python_testing/TC_MCORE_FS_1_3.py --load-from-env=test-env.yaml
./scripts/tests/run_python_test.py --script src/python_testing/TC_MCORE_FS_1_4.py --load-from-env=test-env.yaml
```


src/python_testing/TC_MCORE_FS_1_4.py
```diff
@@ -0,0 +1,483 @@
...
+        args.extend(["--discriminator", str(discriminator)])
+        args.extend(["--passcode", str(passcode)])
+        self.app = Subprocess(args, tag="SERVER")
+        self.app.start()
+
+    def stop(self):
+        self.app.stop()
+
+
+class TC_MCORE_FS_1_4(MatterBaseTest):
```
arkq, line 166: Blocked on https://github.com/CHIP-Specifications/chip-test-plans/pull/4575 landing first

src/python_testing/TC_MCORE_FS_1_4.py
```diff
@@ -0,0 +1,482 @@
...
+    def stop(self):
+        self.p.terminate()
+        self.join()
+
+
+class FabricSyncApp:
+
+    APP_PATH = "examples/fabric-admin/scripts/fabric-sync-app.py"
+    FABRIC_ADMIN_PATH = "out/linux-x64-fabric-admin-rpc/fabric-admin"
+    FABRIC_BRIDGE_PATH = "out/linux-x64-fabric-bridge-rpc/fabric-bridge-app"
```
tehampson, line 92: This will need to be provided as command line arguments for this to work on TH. These paths do not exist on TH

src/python_testing/TC_MCORE_FS_1_4.py
```diff
@@ -0,0 +1,482 @@
...
+            paaTrustStorePath=self.matter_test_config.paa_trust_store_path,
+            bridgePort=self.th_fsa_bridge_port,
+            bridgeDiscriminator=self.th_fsa_bridge_discriminator,
+            bridgePasscode=self.th_fsa_bridge_passcode,
+            vendorId=0xFFF1)
+
+        # Get the named pipe path for the DUT_FSA app input from the user params.
+        dut_fsa_stdin_pipe = self.user_params.get("dut_fsa_stdin_pipe", None)
+        if dut_fsa_stdin_pipe is not None:
+            self.dut_fsa_stdin = open(dut_fsa_stdin_pipe, "w")
```
tehampson, line 215: Does this need to be guarded by a CI check? On TH you will not be able to have this pipe to DUT as it will be running on a different host

src/python_testing/TC_MCORE_FS_1_4.py
arkq: These values are only defaults. Everything can (and should be) provided via string args

src/python_testing/TC_MCORE_FS_1_4.py
```diff
@@ -0,0 +1,482 @@
...
+    def default_timeout(self) -> int:
+        # This test has some manual steps, so we need a longer timeout.
+        return 200
+
+    def setup_class(self):
+        super().setup_class()
+
+        # Get the path to the TH_FSA (fabric-admin and fabric-bridge) app from
+        # the user params or use the default path.
+        FabricSyncApp.APP_PATH = self.user_params.get("th_fsa_app_path", FabricSyncApp.APP_PATH)
```
tehampson, line 178: I have always been under the impression that in python THING_IN_ALL_CAPS are supposed to indicate they constants. I would rather see this get assigned to lower case value.

I would rather see this passed into `FabricSyncApp` constructor, where the default if not provide is None we use the constant

examples/fabric-admin/scripts/fabric-sync-app.py
```diff
@@ -0,0 +1,251 @@
...
+async def forward_f(f_in, f_out, prefix: str):
+    """Forward f_in to f_out with a prefix attached."""
+    global BRIDGE_COMMISSIONED
+    while True:
+        line = await f_in.readline()
+        if not line:
+            break
+        if not BRIDGE_COMMISSIONED.is_set() and BRIDGE_COMMISSIONED_MSG in line:
+            BRIDGE_COMMISSIONED.set()
+        f_out.buffer.write(prefix.encode())
```
andy31415, line 40: nit: why doesn't the caller provide the encoded string in? Or could we just do the encode once before the while loop?

Could we add types for f_in/f_out? how do we know they have a `buffer` member and a flush one?

andy31415: We seem to want `https://docs.python.org/3/library/typing.html#typing.BinaryIO` since we encode bytes.

examples/fabric-admin/scripts/fabric-sync-app.py
```diff
@@ -0,0 +1,251 @@
...
+    from the pipe.
+    """
+    fd = os.open(pipe_path, os.O_RDONLY | os.O_NONBLOCK)
+    while True:
+        try:
+            data = os.read(fd, 1024)
+            if data:
+                f_out.write(data)
+            if not data:
+                await asyncio.sleep(1)
```
andy31415, line 59: Maybe consider shorter sleeps than 1 second. Maybe 0.1 or similar.

examples/fabric-admin/scripts/fabric-sync-app.py
```diff
@@ -0,0 +1,251 @@
...
+            cmd += f" --setup-pin-code {args.passcode}"
+        if args.secured_device_port is not None:
+            cmd += f" --local-port {args.secured_device_port}"
+        admin.stdin.write((cmd + "\n").encode())
+        # Wait for the bridge to be commissioned.
+        await asyncio.sleep(5)
+
+    # Open commissioning window with original setup code for the bridge,
+    # so it can be added by the TH fabric.
+    cmd = "pairing open-commissioning-window 1 0 0 600 1000 0"
```
andy31415, line 199: add in the above comment what each of the arguments mean.

examples/fabric-admin/scripts/fabric-sync-app.py
```diff
@@ -0,0 +1,251 @@
...
+    admin.stdin.write((cmd + "\n").encode())
+    # Wait some time for the bridge to open the commissioning window.
+    await asyncio.sleep(1)
```
andy31415, line 202: is there a log we can wait on? It seems like we want a pattern of creating some wait object, issuing a command and waiting with timeout. Some pseudo-python would look like:

```py
log_wait = WaitForString('Reading attribute: Cluster=0x0000_001D Endpoint=0x1 AttributeId=0x0000_0000')
admin.SendCommand('descriptor read device-type-list 1 1 --timeout 1')
log_wait.wait(timeout_sec=1)
```


unsure a bit about what interfaces  are the cleanest, however it does seem like we would want an easy way to "send command" (I should not need to add \n and encode every time) and then wait for some output.

src/python_testing/TC_MCORE_FS_1_3.py
```diff
@@ -15,15 +15,31 @@
...
+# See https://github.com/project-chip/connectedhomeip/blob/master/docs/testing/python.md#defining-the-ci-test-arguments
+# for details about the block below.
+#
+# === BEGIN CI TEST ARGUMENTS ===
+# test-runner-runs: run1
+# test-runner-run/run1/app: examples/fabric-admin/scripts/fabric-sync-app.py
+# test-runner-run/run1/app-args: --app-admin=${FABRIC_ADMIN_APP} --app-bridge=${FABRIC_BRIDGE_APP} --stdin-pipe=dut-fsa-stdin --discriminator=1234
+# test-runner-run/run1/factoryreset: True
+# test-runner-run/run1/script-args: --storage-path admin_storage.json --commissioning-method on-network --discriminator 1234 --passcode 20202021 --string-arg th_server_no_uid_app_path:${LIGHTING_APP_NO_UNIQUE_ID}
+# test-runner-run/run1/script-start-delay: 5
```
andy31415, line 31: No other scripts seem to use this script-start-delay option. Could we avoid it somehow?

Any fixed-time delay is not ideal, either it is too long because apps start fast or too short and makes tests flaky. Being able to have conditional tests is better.

src/python_testing/TC_MCORE_FS_1_3.py
```diff
@@ -32,131 +48,221 @@
 from mobly import asserts
 
 
+class Subprocess(threading.Thread):
```
andy31415, line 51: these classes seem generic. Could we place them in https://github.com/project-chip/connectedhomeip/tree/master/src/python_testing/matter_testing_infrastructure ?

src/python_testing/TC_MCORE_FS_1_3.py
tehampson: In this case we can't script-start-delay was added for this test. Every other DUT application is just a single app, but fabric sync is two applications that needs to do some level of initial configuration. The CI runner script just starts the application and assume the moment later the app is ready to go, which is not the case for fabric-bridge and fabric-admin running together

arkq: I think that we can change that as a follow up, but not today, if I may. I can add an option to search for some "needle" text in the app output in order to start test script.

src/python_testing/TC_MCORE_FS_1_3.py
```diff
@@ -32,131 +48,221 @@
...
-    async def create_device_and_commission_to_th_fabric(self, kvs, port, node_id_for_th, device_info):
-        app = self.user_params.get("th_server_app_path", None)
-        if not app:
-            asserts.fail('This test requires a TH_SERVER app. Specify app path with --string-arg th_server_app_path:<path_to_app>')
+    def steps_TC_MCORE_FS_1_3(self) -> list[TestStep]:
+        return [
+            TestStep(0, "Commission DUT if not done", is_commissioning=True),
+            TestStep(1, "TH commissions TH_SERVER_NO_UID to TH's fabric"),
+            # TestStep(2, "DUT_FSA commissions TH_SERVER_NO_UID to DUT_FSA's fabric and generates a UniqueID.",
+            #          "TH verifies a value is visible for the UniqueID from the DUT_FSA's Bridged Device Basic Information Cluster."),
```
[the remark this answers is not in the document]
arkq: Leftover after some testing...

examples/fabric-admin/scripts/fabric-sync-app.py
```diff
@@ -0,0 +1,251 @@
...
+
+    # Wait a bit for apps to start.
+    await asyncio.sleep(1)
+
+    try:
+        # Check whether the bridge is already commissioned. If it is,
+        # we will get the response, otherwise we will hit timeout.
+        cmd = "descriptor read device-type-list 1 1 --timeout 1"
+        admin.stdin.write((cmd + "\n").encode())
+        await asyncio.wait_for(BRIDGE_COMMISSIONED.wait(), timeout=1.5)
```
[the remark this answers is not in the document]
arkq: The usage of `global` was a leftover from some PoC implementation. Removed now.

src/python_testing/TC_MCORE_FS_1_4.py
arkq: Removed default

examples/fabric-admin/scripts/fabric-sync-app.py
```diff
@@ -0,0 +1,251 @@
...
+    except asyncio.TimeoutError:
+        # Commission the bridge to the admin.
+        cmd = "fabricsync add-local-bridge 1"
+        if args.passcode is not None:
+            cmd += f" --setup-pin-code {args.passcode}"
+        if args.secured_device_port is not None:
+            cmd += f" --local-port {args.secured_device_port}"
+        admin.stdin.write((cmd + "\n").encode())
+        # Wait for the bridge to be commissioned.
+        await asyncio.sleep(5)
```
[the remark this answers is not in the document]
arkq: I think that we can wait for some string or if necessary add a special log just for that purpose, I can do it as a followup.

examples/fabric-admin/scripts/fabric-sync-app.py
```diff
@@ -0,0 +1,251 @@
...
+
+
+async def run(tag, program, *args, stdin=None):
+    p = await asyncio.create_subprocess_exec(program, *args,
+                                             stdout=asyncio.subprocess.PIPE,
+                                             stderr=asyncio.subprocess.PIPE,
+                                             stdin=stdin)
+    # Add the stdout and stderr processing to the event loop.
+    asyncio.create_task(forward_f(p.stderr, sys.stderr, tag))
+    asyncio.create_task(forward_f(p.stdout, sys.stdout, tag))
```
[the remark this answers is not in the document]
arkq: I've refactored the logic a bit, so now the process is wrapped in a class to make the code more object oriented.

src/python_testing/TC_MCORE_FS_1_4.py
```diff
@@ -0,0 +1,482 @@
...
+import threading
+
+import chip.clusters as Clusters
+from chip import ChipDeviceCtrl
+from chip.interaction_model import Status
+from matter_testing_support import MatterBaseTest, TestStep, async_test_body, default_matter_test_main, type_matches
+from mobly import asserts
+
+
+class Subprocess(threading.Thread):
```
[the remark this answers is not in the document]
arkq: Will do as a followup: https://github.com/project-chip/connectedhomeip/issues/35348

src/python_testing/TC_MCORE_FS_1_3.py
arkq: Will do as a followup: https://github.com/project-chip/connectedhomeip/issues/35348

Also see this: https://github.com/project-chip/connectedhomeip/pull/35274#discussion_r1737236695

src/python_testing/TC_MCORE_FS_1_4.py
```diff
@@ -0,0 +1,482 @@
...
+
+        self.admin = Subprocess(args, stdout_cb=self._process_admin_output)
+        self.wait_for_text_text = "Connected to Fabric-Bridge"
+        self.admin.start()
+
+        # Wait for the bridge to connect to the admin.
+        self.wait_for_text()
+
+    def CommissionOnNetwork(self, nodeId, setupPinCode=None, filterType=None, filter=None):
+        self.wait_for_text_text = f"Commissioning complete for node ID 0x{nodeId:016x}: success"
```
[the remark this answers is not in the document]
arkq: Will do as a followup: https://github.com/project-chip/connectedhomeip/issues/35348

src/python_testing/TC_MCORE_FS_1_4.py
```diff
@@ -0,0 +1,482 @@
...
+        # Create a temporary storage directory for keeping KVS files.
+        self.storage = tempfile.TemporaryDirectory(prefix=self.__class__.__name__)
+        logging.info("Temporary storage directory: %s", self.storage.name)
+
+        self.th_fsa_bridge_address = "::1"
+        self.th_fsa_bridge_port = 5543
+        self.th_fsa_bridge_discriminator = random.randint(0, 4095)
+        self.th_fsa_bridge_passcode = 20202021
+
+        self.th_fsa_controller = FabricSyncApp(
```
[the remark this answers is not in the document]
arkq: I've fixed wait logic, to actually wait for commissioning window reopening which is the last step in fabric-sync-app initialization.

src/python_testing/TC_MCORE_FS_1_4.py
```diff
@@ -0,0 +1,482 @@
...
+            args.append(f"--commissioner-name={fabricName}")
+        if nodeId is not None:
+            args.append(f"--commissioner-node-id={nodeId}")
+        args.append(f"--commissioner-vendor-id={vendorId}")
+        args.append(f"--secured-device-port={bridgePort}")
+        args.append(f"--discriminator={bridgeDiscriminator}")
+        args.append(f"--passcode={bridgePasscode}")
+
+        self.admin = Subprocess(args, stdout_cb=self._process_admin_output)
+        self.wait_for_text_text = "Connected to Fabric-Bridge"
```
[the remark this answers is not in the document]
arkq: Fixed

examples/fabric-admin/scripts/fabric-sync-app.py
```diff
@@ -0,0 +1,320 @@
...
+        if args.secured_device_port is not None:
+            cmd += f" --local-port {args.secured_device_port}"
+        await admin.send(
+            cmd,
+            # Wait for the log message indicating that the bridge has been
+            # added to the fabric.
+            f"Commissioning complete for node ID {bridge_node_id:#018x}: success")
+
+    # Open commissioning window with original setup code for the bridge,
+    # so it can be added by the TH fabric.
```
tehampson, line 263: supernit, just say that so the application is commissionable. I don't think we need to mention TH fabric here since technically this can be used outside of tests

src/python_testing/TC_MCORE_FS_1_4.py
```diff
@@ -0,0 +1,485 @@
...
+        args = [fabric_sync_app_path]
+        args.append(f"--app-admin={fabric_admin_app_path}")
+        args.append(f"--app-bridge={fabric_bridge_app_path}")
+        # Override default ports, so it will be possible to run
+        # our TH_FSA alongside the DUT_FSA during CI testing.
+        args.append("--app-admin-rpc-port=44000")
+        args.append("--app-bridge-rpc-port=44001")
+        # Keep the storage directory in a temporary location.
+        args.append(f"--storage-dir={storage_dir}")
+        # FIXME: Passing custom PAA store breaks something
```
tehampson, line 117: Nit, I like to see github issues for this if we have an intention of fixing it. I am okay with you using https://github.com/project-chip/connectedhomeip/issues/35348

src/python_testing/TC_MCORE_FS_1_4.py
```diff
@@ -0,0 +1,485 @@
...
+            storage_dir=self.storage.name,
+            port=self.th_server_port,
+            discriminator=self.th_server_discriminator,
+            passcode=self.th_server_passcode)
+
+    def teardown_class(self):
+        self.th_fsa_controller.stop()
+        self.th_server.stop()
+        self.storage.cleanup()
+        super().teardown_class()
```
tehampson, line 239: nit, If we fail early enough none of these are set. Other test we set them to `none` at the very start of setup, and only clean up on each if we know it is no longer `none` representing that it is supposed to be cleaned up

src/python_testing/TC_MCORE_FS_1_4.py
```diff
@@ -0,0 +1,485 @@
...
+            TestStep(2, "TH instructs TH_FSA to commission TH_SERVER_NO_UID to TH_FSA's fabric."),
+            TestStep(3, "TH instructs TH_FSA to open up commissioning window on it's aggregator."),
+            TestStep(4, "Follow manufacturer provided instructions to have DUT_FSA commission TH_FSA's aggregator."),
+            TestStep(5, "Follow manufacturer provided instructions to enable DUT_FSA to synchronize TH_SERVER_NO_UID"
+                     " from TH_FSA onto DUT_FSA's fabric. TH to provide endpoint saved from step 2 in user prompt."),
+            TestStep(6, "DUT_FSA synchronizes TH_SERVER_NO_UID onto DUT_FSA's fabric and copies the UniqueID presented"
+                     " by TH_FSA's Bridged Device Basic Information Cluster."),
+        ]
+
+    async def commission_via_commissioner_control(self, controller_node_id: int, device_node_id: int):
```
tehampson, line 255: nit, Is this used in 1_4? If not maybe we should remove it

src/python_testing/TC_MCORE_FS_1_4.py
```diff
@@ -0,0 +1,485 @@
...
+            bridge_passcode=self.th_fsa_bridge_passcode,
+            vendor_id=0xFFF1)
+
+        # Get the named pipe path for the DUT_FSA app input from the user params.
+        dut_fsa_stdin_pipe = self.user_params.get("dut_fsa_stdin_pipe", None)
+        if dut_fsa_stdin_pipe is not None:
+            self.dut_fsa_stdin = open(dut_fsa_stdin_pipe, "w")
+
+        self.th_server_port = 5544
+        self.th_server_discriminator = random.randint(0, 4095)
```
tehampson, line 224: nit, we have a 1 in 4096 chance of a collision where we might have an issue during discovery. We could just do 

`self.th_fsa_bridge_discriminator = random.randint(0, 4094)` with a comment saying we add one for the th_server so we intentionally made max one less. Then here we add 1 to the previous result to prevent collision.

src/python_testing/TC_MCORE_FS_1_3.py
```diff
@@ -32,109 +48,218 @@
 from mobly import asserts
 
 
+class Subprocess(threading.Thread):
```
tehampson, line 51: nit, please add TODO comment above referencing #35348. Here and in FS_1_4

tehampson (approved):

src/python_testing/TC_MCORE_FS_1_4.py
arkq: Upss, missed that...

src/python_testing/TC_MCORE_FS_1_4.py
arkq: Good point!

src/python_testing/TC_MCORE_FS_1_4.py
arkq: I've just checked and it seems that my comment is no longer valid. Passing PAA store path seems to work fine now.

tehampson (approved):

yufengwangca (approved):

saurabhst (approved):

PR #35279, opened 2024-08-29
Fixed the crash due to packet buffers running out of space in CommandHandlerImpl.

shripad621git:
**Problem**
- There were intermittent crashes observed in the internal automation stress tests due to packet buffer pool empty i.e running out of packet buffers [here
](https://github.com/project-chip/connectedhomeip/blob/9bbf5b9f1de62c23bed4e8974140e3a5ddf79777/src/app/CommandHandlerImpl.cpp#L614
) 
- Although the logic mentioned says that do not crash on running out of packet buffers, the application crashes 
due to the  below mentioned reason:
- [RollbackResponse](https://github.com/project-chip/connectedhomeip/blob/c14a370535c4a3f30277761cc4ce277f84ba8949/src/app/CommandHandlerImpl.h#L297) will return `CHIP_ERROR_INCORRECT_STATE` leading to failure of the condition mentioned above leading to a crash.
[mRollbackBackupValid](https://github.com/project-chip/connectedhomeip/blob/c14a370535c4a3f30277761cc4ce277f84ba8949/src/app/CommandHandlerImpl.cpp#L794) is not set since the function [CreateBackupForResonseCallback](https://github.com/project-chip/connectedhomeip/blob/c14a370535c4a3f30277761cc4ce277f84ba8949/src/app/CommandHandlerImpl.cpp#L739) did not get executed which executes after AllocateBuffer() function in PrepareStatus() function.i,e due to packet buffer full the allocate buffer returned early.

**Change Overview:**
- Fixed the crash by returning the possible `CHIP_ERROR_NO_MEMORY` due to running out of packet buffers.


shripad621git:
@tehampson , can you please review.

src/app/CommandHandlerImpl.h
```diff
@@ -294,7 +294,13 @@ class CommandHandlerImpl : public CommandHandler
         {
             return CHIP_NO_ERROR;
         }
-        ReturnErrorOnFailure(RollbackResponse());
+        // Here the err from previous step is returned intentionally if RollBackResonse fails due to
+        // running out of packet buffers i.e due to CHIP_ERROR_NO_MEMORY in AllocateBuffer() and
+        // fails to set the mRollBackBackupValid.
```
tehampson, line 299: I don't think we should be too specific about the details and do something more along the lines of:
```suggestion
        // `err` from the previous step is returned intentionally if RollBackResponse fails.
        // RollBackResponse only fails if addResponseFunction() was unable to create a checkpoint,
        // which indicates there is nothing to rollback and that trying addResponseFunction will
        // fail for the same reason. The reason why addResponseFunction failed is what is of
        // interest to the caller.
```

tehampson (approved):

src/app/CommandHandlerImpl.h
```diff
@@ -294,7 +294,13 @@ class CommandHandlerImpl : public CommandHandler
         {
             return CHIP_NO_ERROR;
         }
-        ReturnErrorOnFailure(RollbackResponse());
+        // The error value of RollbackResponse is not important if it fails, we prioritize
+        // conveying the error generated by addResponseFunction to the
+        // caller.
```
tehampson, line 299: nit, you can have the next line moved up to here it seems odd with the alignment of the comment above to have a 3rd line.

Sorry it is hard to see the alignment stuff when writing the suggestion but it is easier to see if the review after you apply the suggestion

```suggestion
        // conveying the error generated by addResponseFunction to the caller.
```

shripad621git: Done.

Damian-Nordic (approved):
Thanks, I think I've seen this crash in the past as well.

PR #35287, opened 2024-08-29
Add force send parameter to setOnOffValue and send it on matter start

damian-kurek-wizzdev:
I sent this PR based on matter implementation for esp-matter it might work differently in other SDKs. 
This change is useful and required for an onoff device controlled via NC/NO relay. 
Since matter controls the startup parameter via "onoff write start-up-on-off" parameter. Connectedhomeip is the only place where the startup relay state can be determined. The current implementation's problem is that it only provides information about the startup relay state when it is different from the last known state before restart e.g. 
start-up-on-off is set to On when the last state of OnOff parameter is Off CHIP sends information about state change, 
start-up-on-off is set to On when the last state of OnOff parameter is On CHIP won't send information about the state of OnOff parameter

I think this makes it inconsistent and for me sending the state of OnOff parameter always at the start is the best way to go about it, so information about the startup state always comes from on/off server

### DIFF
--- a/src/app/clusters/on-off-server/on-off-server.cpp
+++ b/src/app/clusters/on-off-server/on-off-server.cpp
@@ -340,8 +340,10 @@ Status OnOffServer::getOnOffValue(chip::EndpointId endpoint, bool * currentOnOff
  * @param endpoint   Ver.: always
  * @param command   Ver.: always
  * @param initiatedByLevelChange   Ver.: always
+ * @param forceSend Send value of the On/Off even when it is the same as current On/Off value.
+ * This parameter is useful at the start when you try to determine startup state of On/Off relay that does not store state after losing power
  */
-Status OnOffServer::setOnOffValue(chip::EndpointId endpoint, chip::CommandId command, bool initiatedByLevelChange)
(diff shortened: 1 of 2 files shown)


src/app/clusters/on-off-server/on-off-server.cpp
```diff
@@ -340,8 +340,11 @@ Status OnOffServer::getOnOffValue(chip::EndpointId endpoint, bool * currentOnOff
  * @param endpoint   Ver.: always
  * @param command   Ver.: always
  * @param initiatedByLevelChange   Ver.: always
+ * @param forceSend Send value of the On/Off even when it is the same as current On/Off value.
```
bzbarsky-apple, line 343: Send what and where?

At startup, if some hardware state needs to be synced to the state of the On/Off cluster, the right sequence of actions is:

1. Start up the Matter stack, which initializes the On/Off cluster instance(s).
2. Read the relevant attributes from those cluster instances (which might involve the StartUpOnOff value, or might involve reading the persisted OnOff state from flash, or some combination thereof).
3. Sync the hardware state to the desired state as represented by the attributes.

bzbarsky-apple (changes requested):

PR #35289, opened 2024-08-29
[Silabs] Moving to CMSIS OS and fixing task priorities

chirag-silabs:
The PR have following changes:

1. Moving to cmsis os threads and timer changes
2. Removing freertos and unused code
3. Refactor wfx_rsi_host.cpp
4. Changing the priorities for siwx917 soc via the cmsis os.


examples/platform/silabs/efr32/rs911x/rsi_if.c
```diff
@@ -55,14 +55,17 @@
...
 
-/* Rsi driver Task will use as its stack */
-StackType_t driverRsiTaskStack[WFX_RSI_WLAN_TASK_SZ] = { 0 };
-
-/* Structure that will hold the TCB of the wfxRsi Task being created. */
-StaticTask_t driverRsiTaskBuffer;
-
-/* Declare a variable to hold the data associated with the created event group. */
-StaticEventGroup_t rsiDriverEventGroup;
+#define kDrvTaskSize 1792 // Example size, adjust as needed
```
jmartinez-silabs, line 58: before you were using `WFX_RSI_WLAN_TASK_SZ` define was that also 1792?

Also, you use #define but format as if it is a variable.

constexpr uint32 should be used.  
And for the comment; Users are not expected to go in the platform code source file and edit them directly. If you expect them to change the stack size of the task make it configurable in some way 

```suggestion
constexpr uint32 kDrvTaskSize = 1792;
```

examples/platform/silabs/uart.cpp
```diff
@@ -138,7 +138,13 @@ constexpr osThreadAttr_t kUartTaskAttr = { .name       = "UART",
                                            .cb_size    = osThreadCbSize,
                                            .stack_mem  = uartStack,
                                            .stack_size = kUartTaskSize,
-                                           .priority   = osPriorityRealtime };
+#if SLI_SI91X_MCU_INTERFACE
+                                           // Reducing the priority of the UART task to avoid priority inversion
```
jmartinez-silabs, line 142: priority inversion over what task? This task only depends on what is placed on its queue and then writes it to uart.

examples/platform/silabs/wfx_rsi.h
```diff
@@ -28,9 +30,8 @@
  * Interface to RSI Sapis
  */
 
-#define WFX_RSI_WLAN_TASK_SZ (1024 + 512 + 256) /* Stack for the WLAN task	 	*/
-#define WFX_RSI_TASK_SZ (1024 + 1024)           /* Stack for the WFX/RSI task		*/
-#define WFX_RSI_BUF_SZ (1024 * 10)              /* May need tweak 			*/
+#define WFX_RSI_WLAN_TASK_SZ (1024 + 512 + 256) /* Stack for the WLAN task	 	*/ // TODO: For rs9116
+#define WFX_RSI_BUF_SZ (1024 * 10) /* May need tweak 			*/                 // TODO: For rs9116
```
jmartinez-silabs, line 34: `/* May need tweak */ `May need tweak comment as been there for a while now. Is it the right value or not? can see remove the comment

examples/platform/silabs/efr32/rs911x/rsi_if.c
```diff
@@ -685,7 +702,8 @@ void ProcessEvent(WfxEvent_t inEvent)
         // is independant of IP connectivity.
         break;
     case WFX_EVT_STA_DISCONN:
-        // TODO: This event is not being posted anywhere, seems to be a dead code or we are missing something
+        // TODO: This event is not being posted anywhere, seems to be a dead code or we are missing
```
jmartinez-silabs, line 705: Todo has been there for a while. Do we need it to capture this event here or not? if not delete it.

examples/platform/silabs/wfx_rsi.h
```diff
@@ -71,13 +72,10 @@ typedef struct WfxEvent_s
 
 typedef struct wfx_rsi_s
 {
-    // TODO: Change tp WfxEventType_e once the event queue is implemented
-    EventGroupHandle_t events;
-    TaskHandle_t drv_task;
-    TaskHandle_t wlan_task;
-    TaskHandle_t init_task;
+    osThreadId_t drv_thread;
```
jmartinez-silabs, line 75: This has been pointed out before but why does the different task handle/threadID need to be kept in this struck. (which is available over multiple interface).

I feel the task id would be contained to its owner.

src/platform/silabs/efr32/BLEManagerImpl.cpp
```diff
@@ -1016,23 +1011,23 @@ void BLEManagerImpl::BleAdvTimeoutHandler(TimerHandle_t xTimer)
...
 {
-    if (xTimerIsTimerActive(sbleAdvTimeoutTimer))
+    if (osTimerIsRunning(sbleAdvTimeoutTimer))
     {
         CancelBleAdvTimeoutTimer();
     }
 
     // timer is not active, change its period to required value (== restart).
     // FreeRTOS- Block for a maximum of 100 ticks if the change period command
     // cannot immediately be sent to the timer command queue.
```
jmartinez-silabs, line 1029: This is not needed anymore. osTimerStart handles it already
```suggestion
```

src/platform/silabs/rs911x/BLEManagerImpl.cpp
```diff
@@ -1081,23 +1078,23 @@ void BLEManagerImpl::BleAdvTimeoutHandler(TimerHandle_t xTimer)
...
 {
-    if (xTimerIsTimerActive(sbleAdvTimeoutTimer))
+    if (osTimerIsRunning(sbleAdvTimeoutTimer))
     {
         CancelBleAdvTimeoutTimer();
     }
 
     // timer is not active, change its period to required value (== restart).
     // FreeRTOS- Block for a maximum of 100 ticks if the change period command
     // cannot immediately be sent to the timer command queue.
```
jmartinez-silabs, line 1096:
```suggestion
```

jmartinez-silabs (changes requested):

examples/platform/silabs/uart.cpp
chirag-silabs: noticed that the PASE session itself is not getting started if the logs are more, if LWIP logs are also enabled the issue is seen in CASE session. Therefore the UART logging priority is reduced from the BLE and LWIP task so that logging never blocks the important task

examples/platform/silabs/wfx_rsi.h
chirag-silabs: removed the comment

examples/platform/silabs/efr32/rs911x/rsi_if.c
chirag-silabs: removed the TODOs

examples/platform/silabs/wfx_rsi.h
chirag-silabs: removed the threads from the structures and added to their required file

examples/platform/silabs/efr32/rs911x/rsi_if.cpp
```diff
@@ -724,7 +749,7 @@ void ProcessEvent(WfxEvent_t inEvent)
                 if (!wfx_rsi.scan_ssid || (wfx_rsi.scan_ssid && strcmp(wfx_rsi.scan_ssid, (char *) scan->ssid) == CMP_SUCCESS))
```
jmartinez-silabs, line 749: I know you didn't touch this, but could you tweak this strcmp(wfx_rsi.scan_ssid, (char *) scan->ssid) to make sure it doesn't compare more than the smaller buffer in case those strings not null terminated.

jmartinez-silabs (approved):

examples/platform/silabs/efr32/rs911x/rsi_if.cpp
chirag-silabs: done

andy31415 (approved):

PR #35302, opened 2024-08-29
Disallow read-through on internally-created MTRDevice instances.

bzbarsky-apple:
We want to be able to instantiate an MTRDevice and readAttibute to get its cached state without ever triggering read-throughs.

The MTRDiagnosticLogsDownloader.mm changes were because it was using internal APIs to create an MTRDevice when it does not need an MTRDevice at all.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice_Internal.h
+++ b/src/darwin/Framework/CHIP/MTRDevice_Internal.h
@@ -120,7 +120,18 @@ MTR_DIRECT_MEMBERS
     // Our controller.  Declared nullable because our property is, though in
     // practice it does not look like we ever set it to nil.
     MTRDeviceController * _Nullable _deviceController;
+
+    // Whether this device has been accessed via the public deviceWithNodeID API
+    // (as opposed to just via the internal _deviceWithNodeID).
+    BOOL _accessedViaPublicAPI;
 }
+
+/**
+ * Internal way of creating an MTRDevice that does not flag the device as being
+ * visible to external API consumers.
+ */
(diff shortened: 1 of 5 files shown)


src/darwin/Framework/CHIP/MTRDevice_Internal.h
```diff
@@ -120,8 +120,19 @@ MTR_DIRECT_MEMBERS
     // Our controller.  Declared nullable because our property is, though in
     // practice it does not look like we ever set it to nil.
     MTRDeviceController * _Nullable _deviceController;
+
+    // Whether this device has been accessed via the public deviceWithNodeID API
+    // (as opposed to just via the internal _deviceWithNodeID).
+    BOOL _accessedViaPublicAPI;
```
anush-apple, line 126: Perhaps this should be renamed to: _readThoughsAllowed to match what it is trying to guard against: Allow it to be used without requiring readThroughs

bzbarsky-apple: I guess that depends on whether we will have any other behavior branching on this state in the future.

And more importantly, _readThoughsAllowed doesn't make sense on the base MTRDevice, since the control over the read-through behavior it has only applies to MTRDevice_Concrete.  It doesn't affect MTRDevice_XPC behavior at all.

anush-apple: Sounds good

anush-apple (approved):

bzbarsky-apple:

PR #35303, opened 2024-08-29
Implementation of UniqueID attribute in Bridge app

### LINKED ISSUE #34366: [1.4][TC-BRBINFO-2.1] The implementation of UniqueID attribute is required in Bridge app
### Test Case

TC-BRBINFO-2.1

#### Below mentioned attribute has to be enabled in Bridge-app as it is Mandatory in SPEC :

- UniqueID

**Spec Reference:** https://github.com/CHIP-Specifications/connectedhomeip-spec/blob/master/src/data_model/bridge-clusters.adoc#attributes

05e4c10358a18f9c9295c2dcd7cdd880502d539a

### Platform

raspi

yufengwangca:
Test Case
TC-BRBINFO-2.1

Below mentioned attribute has to be enabled in Bridge-app as it is Mandatory in SPEC :
UniqueID
Spec Reference: https://github.com/CHIP-Specifications/connectedhomeip-spec/blob/master/src/data_model/bridge-clusters.adoc#attributes

Fixes #34366

### DIFF
--- a/examples/bridge-app/linux/include/Device.h
+++ b/examples/bridge-app/linux/include/Device.h
@@ -30,7 +30,8 @@
 class Device
 {
 public:
-    static const int kDeviceNameSize = 32;
+    static const int kDeviceNameSize     = 32;
+    static const int kDeviceUniqueIdSize = 32;
 
     enum Changed_t
     {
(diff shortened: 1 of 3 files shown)


saurabhst (approved):

examples/bridge-app/linux/include/Device.h
```diff
@@ -61,7 +65,8 @@ class Device
 
 protected:
     bool mReachable;
-    char mName[kDeviceNameSize];
+    char mName[kDeviceNameSize + 1];
+    char mUniqueId[kDeviceUniqueIdSize + 1];
```
tehampson, line 69: Do we need to make sure that is it initialized with `\0`

tehampson (approved):

examples/bridge-app/linux/include/Device.h
yufengwangca: acked

PR #35313, opened 2024-08-30
OpenThread: add an api to disconnect driver

DejinChen:
Ref to https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/10335, add an API to disconnect Thread driver.

#### Required
- [ ] https://github.com/project-chip/connectedhomeip/pull/35256

### DIFF
--- a/src/platform/OpenThread/GenericNetworkCommissioningThreadDriver.cpp
+++ b/src/platform/OpenThread/GenericNetworkCommissioningThreadDriver.cpp
@@ -277,6 +277,18 @@ void GenericThreadDriver::CheckInterfaceEnabled()
 #endif
 }
 
+CHIP_ERROR GenericThreadDriver::DisconnectDriver()
+{
+    if (ThreadStackMgrImpl().IsThreadProvisioned())
+    {
+        ReturnErrorOnFailure(BackupConfiguration());
+        Thread::OperationalDataset emptyNetwork = {};
+        // Attach to an empty network will disconnect the driver.
+        ReturnErrorOnFailure(ThreadStackMgrImpl().AttachToThreadNetwork(emptyNetwork, nullptr));
+    }
+    return CHIP_NO_ERROR;
+}
+
 size_t GenericThreadDriver::ThreadNetworkIterator::Count()
 {
     return driver->mStagingNetwork.IsCommissioned() ? 1 : 0;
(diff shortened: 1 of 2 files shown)


src/platform/OpenThread/GenericNetworkCommissioningThreadDriver.cpp
```diff
@@ -277,6 +277,18 @@ void GenericThreadDriver::CheckInterfaceEnabled()
 #endif
 }
 
+CHIP_ERROR GenericThreadDriver::DisconnectDriver()
```
tcarmelveilleux, line 280: This is the wrong name. It should be called "DisconnectFromNetwork" or such.

src/platform/OpenThread/GenericNetworkCommissioningThreadDriver.cpp
```diff
@@ -277,6 +277,18 @@ void GenericThreadDriver::CheckInterfaceEnabled()
...
+{
+    if (ThreadStackMgrImpl().IsThreadProvisioned())
+    {
+        ReturnErrorOnFailure(BackupConfiguration());
```
tcarmelveilleux, line 284: Why the backup here?

src/platform/OpenThread/GenericNetworkCommissioningThreadDriver.h
```diff
@@ -106,6 +106,7 @@ class GenericThreadDriver final : public ThreadDriver
 
     CHIP_ERROR CommitConfiguration() override;
     CHIP_ERROR RevertConfiguration() override;
+    CHIP_ERROR DisconnectDriver() override;
```
ksperling-apple, line 109: `DisconnectNetwork()` would go well with `ConnectNetwork()` below.

ksperling-apple: Also, I think it would make sense to define this in `WirelessDriver`? Or how/where do you propose to implement the logic that can check if the currently active connection is no longer in the `Networks` list?

src/platform/OpenThread/GenericNetworkCommissioningThreadDriver.cpp
DejinChen: Done.

src/platform/OpenThread/GenericNetworkCommissioningThreadDriver.cpp
DejinChen: Removed.

src/platform/OpenThread/GenericNetworkCommissioningThreadDriver.h
DejinChen: Added this change to https://github.com/project-chip/connectedhomeip/pull/35256, will close this later.

PR #33745, opened 2024-06-05
[ESP32] Support Thread and Wi-Fi network revert configuration

DejinChen:
### Problem
Platform ESP32 do not support Wi-Fi reverting network nor Thread reverting network to disconnect state.

### Change overview
- Removed check for KVS value point to support Thread reverting network to disconnect state
- Added RevertConfiguration for Wi-Fi

### Testing
RevertConfiguration for Wi-Fi and Thread network on ESP32-C6.

PR #35314, opened 2024-08-30
ESP32: add an api to disconnect Wi-Fi driver

DejinChen:
Ref to https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/10335, add an API to disconnect ESP32 Wi-Fi driver.

#### Required
- [ ] https://github.com/project-chip/connectedhomeip/pull/35256
- [ ] https://github.com/project-chip/connectedhomeip/pull/33745

### DIFF
--- a/src/platform/ESP32/NetworkCommissioningDriver.cpp
+++ b/src/platform/ESP32/NetworkCommissioningDriver.cpp
@@ -236,6 +236,17 @@ CHIP_ERROR ESPWiFiDriver::ConnectWiFiNetwork(const char * ssid, uint8_t ssidLen,
     return ConnectivityMgr().SetWiFiStationMode(ConnectivityManager::kWiFiStationMode_Enabled);
 }
 
+CHIP_ERROR ESPWiFiDriver::DisconnectDriver()
+{
+    if (chip::DeviceLayer::Internal::ESP32Utils::IsStationProvisioned())
+    {
+        ReturnErrorOnFailure(BackupConfiguration());
+        // Attaching to an empty network will disconnect the network.
+        ReturnErrorOnFailure(ConnectWiFiNetwork(nullptr, 0, nullptr, 0));
+    }
+    return CHIP_NO_ERROR;
+}
+
 void ESPWiFiDriver::OnConnectWiFiNetwork()
 {
     if (mpConnectCallback)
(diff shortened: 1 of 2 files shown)


src/platform/ESP32/NetworkCommissioningDriver.cpp
```diff
@@ -236,6 +236,17 @@ CHIP_ERROR ESPWiFiDriver::ConnectWiFiNetwork(const char * ssid, uint8_t ssidLen,
     return ConnectivityMgr().SetWiFiStationMode(ConnectivityManager::kWiFiStationMode_Enabled);
 }
 
+CHIP_ERROR ESPWiFiDriver::DisconnectDriver()
```
tcarmelveilleux, line 239: This should be called `DisconnectFromNetwork`

src/platform/ESP32/NetworkCommissioningDriver.cpp
```diff
@@ -236,6 +236,17 @@ CHIP_ERROR ESPWiFiDriver::ConnectWiFiNetwork(const char * ssid, uint8_t ssidLen,
...
+{
+    if (chip::DeviceLayer::Internal::ESP32Utils::IsStationProvisioned())
+    {
+        ReturnErrorOnFailure(BackupConfiguration());
```
tcarmelveilleux, line 243: Why the backup here?

src/platform/ESP32/NetworkCommissioningDriver.cpp
tcarmelveilleux: There must be a public interface change for DisconnectFromNetwork() to be an abstract base class method, overridden by platforms. We can't have a new feature that only is implemented/implementable on a single platform.

ksperling-apple: Agree, though I think we can have a base class impl that just returns CHIP_ERROR_NOT_IMPLEMENTED since these code paths won't be relevant for single-interface devices.

src/platform/ESP32/NetworkCommissioningDriver.cpp
DejinChen: removed

src/platform/ESP32/NetworkCommissioningDriver.cpp
DejinChen: Added this change to https://github.com/project-chip/connectedhomeip/pull/35256, will close this later.

PR #35322, opened 2024-08-30
Fix GlobalAttributes check in TC_ACL_2_11

tleacmcsa:
There were a couple of problems in the test script used to validate that a DUT does not have Access Restrictions in place on global attributes.  This change properly fails the test if any cluster on any endpoint has global attribute access restricted.


src/python_testing/TC_ACL_2_11.py
```diff
@@ -118,16 +118,16 @@ async def test_TC_ACL_2_11(self):
                 restriction_type = restriction.type
                 ID1 = restriction.id
 
-                attribute = ALL_ATTRIBUTES[C1][ID1]
-                command = ALL_ACCEPTED_COMMANDS[C1][ID1]
-
                 if restriction_type == AccessControl.Enums.AccessRestrictionTypeEnum.kAttributeAccessForbidden:
+                    attribute = ALL_ATTRIBUTES[C1][ID1]
```
tcarmelveilleux, line 122: How does this handle wildcards? Was this tested against wildcards?

src/python_testing/TC_ACL_2_11.py
```diff
@@ -118,16 +118,16 @@ async def test_TC_ACL_2_11(self):
...
                     await self.read_single_attribute_expect_error(cluster=cluster, attribute=attribute, error=Status.AccessRestricted, endpoint=E1)
                 elif restriction_type == AccessControl.Enums.AccessRestrictionTypeEnum.kAttributeWriteForbidden:
+                    attribute = ALL_ATTRIBUTES[C1][ID1]
                     status = await self.write_single_attribute(attribute_value=attribute, endpoint_id=E1)
                     asserts.assert_equal(status, Status.AccessRestricted,
                                          f"Failed to verify ACCESS_RESTRICTED when writing to Attribute {ID1} Cluster {C1} Endpoint {E1}")
                 elif restriction_type == AccessControl.Enums.AccessRestrictionTypeEnum.kCommandForbidden:
+                    command = ALL_ACCEPTED_COMMANDS[C1][ID1]
                     result = await self.send_single_cmd(cmd=command, endpoint=E1)
```
tcarmelveilleux, line 131: Does this send the command if the command format is completely wrong? Do you have a log showing that the command got sent?

Also, if the element is a wildcard, you have to iterate over all known commands of the cluster, not just `ID1`, which would otherwise fail.

chrisdecenzo (approved):

src/python_testing/TC_ACL_2_11.py
```diff
@@ -160,12 +193,18 @@ async def test_TC_ACL_2_11(self):
             server_list = ret[endpoint][Clusters.Descriptor][Clusters.Descriptor.Attributes.ServerList]
             for server in server_list:
                 cluster = Clusters.ClusterObjects.ALL_CLUSTERS[server]
-                await dev_ctrl.ReadAttribute(dut_node_id, [endpoint,
-                                                           cluster.Attributes.GeneratedCommandList,
-                                                           cluster.Attributes.AcceptedCommandList,
-                                                           cluster.Attributes.AttributeList,
-                                                           cluster.Attributes.FeatureMap,
-                                                           cluster.Attributes.ClusterRevision])
+                data = await dev_ctrl.ReadAttribute(dut_node_id, [(endpoint, cluster.Attributes.GeneratedCommandList),
```
tleacmcsa, line 196: i would like to refactor this list to be in common with the global_clusters list in a subsequent change.

src/python_testing/TC_ACL_2_11.py
tleacmcsa: yes, fortunately the write attribute and invoke command operations do go over the wire and get evaluated for access before the arguments/values are validated.

src/python_testing/TC_ACL_2_11.py
```diff
@@ -113,24 +123,47 @@ async def test_TC_ACL_2_11(self):
...
                 elif restriction_type == AccessControl.Enums.AccessRestrictionTypeEnum.kAttributeWriteForbidden:
-                    status = await self.write_single_attribute(attribute_value=attribute, endpoint_id=E1)
-                    asserts.assert_equal(status, Status.AccessRestricted,
-                                         f"Failed to verify ACCESS_RESTRICTED when writing to Attribute {ID1} Cluster {C1} Endpoint {E1}")
+                    if ID1 is NullValue:
+                        for attr_id, attribute in ALL_ATTRIBUTES[C1].items():
+                            if attr_id not in self.global_attributes:
+                                status = await self.write_single_attribute(attribute_value=attribute(), endpoint_id=E1, expect_success=False)
+                                if status is not Status.UnsupportedWrite:
+                                    asserts.assert_equal(status, Status.AccessRestricted,
```
tcarmelveilleux, line 145: All writes should have failed with AccessRestricted, not UnsupportedWrite, for wildcard.

src/python_testing/TC_ACL_2_11.py
tleacmcsa: fixed.

src/python_testing/TC_ACL_2_11.py
tleacmcsa: wildcard support fixed.

src/access/AccessRestrictionProvider.cpp
```diff
@@ -197,13 +197,19 @@ CHIP_ERROR AccessRestrictionProvider::DoCheck(const std::vector<Entry> & entries
...
+                    {
+                        return CHIP_ERROR_ACCESS_RESTRICTED_BY_ARL;
+                    }
                 }
                 break;
             case Type::kAttributeWriteForbidden:
                 if (requestPath.requestType == RequestType::kAttributeWriteRequest)
                 {
-                    return CHIP_ERROR_ACCESS_RESTRICTED_BY_ARL;
+                    if (!IsGlobalAttribute(requestPath.entityId.value()))
```
chrisdecenzo, line 209: I don't believe any of the global attributes are writable, but perhaps one will be in the future

tcarmelveilleux (approved):

src/access/AccessRestrictionProvider.cpp
tcarmelveilleux: +1. I think it's fine to leave that here, but yeah, not applicable really...

tcarmelveilleux (approved):

src/python_testing/TC_ACL_2_11.py
tleacmcsa: UnsupportedWrite is returned if the attribute is not writable in the first place.

tcarmelveilleux: Access control should be applied before ACL. That's actually a bug that it returns that first

tcarmelveilleux (approved):

src/access/AccessRestrictionProvider.cpp
```diff
@@ -197,13 +197,19 @@ CHIP_ERROR AccessRestrictionProvider::DoCheck(const std::vector<Entry> & entries
...
+                    {
+                        return CHIP_ERROR_ACCESS_RESTRICTED_BY_ARL;
+                    }
                 }
                 break;
             case Type::kCommandForbidden:
```
bzbarsky-apple, line 215: Followup: Presumably this stuff should not block global commands either....

src/access/AccessRestrictionProvider.cpp
```diff
@@ -197,13 +197,19 @@ CHIP_ERROR AccessRestrictionProvider::DoCheck(const std::vector<Entry> & entries
                 if (requestPath.requestType == RequestType::kAttributeReadRequest ||
                     requestPath.requestType == RequestType::kAttributeWriteRequest)
                 {
-                    return CHIP_ERROR_ACCESS_RESTRICTED_BY_ARL;
+                    if (!IsGlobalAttribute(requestPath.entityId.value()))
```
bzbarsky-apple, line 200: Followup: `AreRestrictionsAllowed` should probably be changed to return false for the "global attribute" case (e.g. by passing in the entity id only when the requestType or restrictionType are attribute-related?)

src/python_testing/TC_ACL_2_11.py
```diff
@@ -113,24 +123,47 @@ async def test_TC_ACL_2_11(self):
...
                 restriction_type = restriction.type
                 ID1 = restriction.id
 
-                attribute = ALL_ATTRIBUTES[C1][ID1]
-                command = ALL_ACCEPTED_COMMANDS[C1][ID1]
-
                 if restriction_type == AccessControl.Enums.AccessRestrictionTypeEnum.kAttributeAccessForbidden:
-                    await self.read_single_attribute_expect_error(cluster=cluster, attribute=attribute, error=Status.AccessRestricted, endpoint=E1)
+                    # if ID1 is null, it means it is a wildcard.  We need to read all attributes on the cluster
+                    if ID1 is NullValue:
```
bzbarsky-apple, line 132: Followup: this might be clearer with a helper function that takes the cluster and attribute id and returns an iterator over the relevant attributes (one in the non-wildcard case), so you could do something like:
```
    for (attr_id, attribute) in getRelevantAttributes(C1, ID1):
        if attr_id in self.global_attributes:
            continue; # Or have getRelevantAttributes filter it out
        # Do the read_single_attribute_expect_error here
```

That way you only have the wildcard-expansion logic in one spot instead of copy/pasting it.

src/python_testing/TC_ACL_2_11.py
```diff
@@ -113,24 +123,47 @@ async def test_TC_ACL_2_11(self):
...
+                        asserts.assert_equal(status, Status.AccessRestricted,
+                                             f"Failed to verify ACCESS_RESTRICTED when writing to Attribute {ID1} Cluster {C1} Endpoint {E1}")
                 elif restriction_type == AccessControl.Enums.AccessRestrictionTypeEnum.kCommandForbidden:
-                    result = await self.send_single_cmd(cmd=command, endpoint=E1)
-                    asserts.assert_equal(result.status, Status.AccessRestricted,
-                                         f"Failed to verify ACCESS_RESTRICTED when sending command {ID1} to Cluster {C1} Endpoint {E1}")
+                    if ID1 is NullValue:
+                        for cmd_id, command in ALL_ACCEPTED_COMMANDS[C1].items():
+                            try:
+                                await self.send_single_cmd(cmd=command(), endpoint=E1, timedRequestTimeoutMs=1000)
```
bzbarsky-apple, line 156: Does this actually work right?  Don't commands need payloads?  Will this auto-generate some default payload for you?

src/python_testing/TC_ACL_2_11.py
```diff
@@ -160,12 +193,18 @@ async def test_TC_ACL_2_11(self):
...
+                                                                  (endpoint, cluster.Attributes.AcceptedCommandList),
+                                                                  (endpoint, cluster.Attributes.AttributeList),
+                                                                  (endpoint, cluster.Attributes.FeatureMap),
+                                                                  (endpoint, cluster.Attributes.ClusterRevision)])
+                for endpoint, clusters in data.items():
+                    for cluster, attributes in clusters.items():
+                        for attribute, value in attributes.items():
+                            asserts.assert_false(isinstance(value, ValueDecodeFailure) and
+                                                 isinstance(value.Reason, InteractionModelError) and
+                                                 value.Reason.status == Status.AccessRestricted,
```
bzbarsky-apple, line 206: Wouldn't it be simpler to just assert that the status is success, period?

src/python_testing/TC_ACL_2_11.py
tleacmcsa: It does work, thankfully.  Otherwise I'm not sure what I would have had to do to generically invoke each command.  It gets far enough for the access check to return AccessRestricted (what we are looking for) or another error related to the bogus payload.

src/python_testing/TC_ACL_2_11.py
tleacmcsa: When it works, the returned dictionary has valid values for each attribute of various types.  Not sure how to check for success.  The failure path for restrictions is a ValueDecodeFailure type containing an InteractionModelError which holds the AccessRestricted status.  Please let me know if you have a better method.

src/python_testing/TC_ACL_2_11.py
tleacmcsa: Yes, fortunately.  If I had to craft legit payloads for attribute writes and command invocations for each type of attribute/command, I don't know how I would pull that off.  I was relieved to find that the access checks happen before payload decoding (which makes sense).  This method does detect restricted commands.

src/python_testing/TC_ACL_2_11.py
bzbarsky-apple: I am not sure how to "check for success" in the Python stuff either, sorry, but there _has_ to be a way to check that the value is just a "server returned success"....  Please check with someone actually familiar with the Python bits?

Anyway, this can be a followup.

tcarmelveilleux: Success of a read is having a value rather than a type indicating a failure. It's runtime typed. There is no easy way to get "success" as a thing, because success is no error

bzbarsky-apple: Hmm, and there are path-specific failures other than ValueDecodeFailure and no common "these are all failures" super-class?

tcarmelveilleux: Yes, because it's the attribute cache that is abstracted

tleacmcsa: @bzbarsky-apple can we resolve this now?

bzbarsky-apple: Yes, though it might be worth a comment in the code about how there is no sane way to check for "this read succeeded", and perhaps a followup issue about adding such a way.

tleacmcsa: I added follow up to put a comment in this test file here: https://github.com/project-chip/connectedhomeip/issues/35341

I created a new issue for finding a better way to do this: https://github.com/project-chip/connectedhomeip/issues/35372

andy31415: Would success/failure not be `assert_false(isinstance(value, Error))` so that any error is bad? why do we filter it so narrowly? This requires a comment of the form: `only AccessRestricted is a failure here, because other failures such as ... or even IM errors such as ... are normal because ....`

tleacmcsa: @andy31415 I captured your comment in the follow-up issue https://github.com/project-chip/connectedhomeip/issues/35341

PR #35327, opened 2024-08-30
Fix MSR flag issues in TC-SWTCH-*

tcarmelveilleux:
- MSR is optional, but was checked as if mandatory.
- See https://github.com/CHIP-Specifications/chip-test-plans/pull/4593
- Fix initial conditions in CI


andy31415 (approved):

ReneJosefsen (approved):
A nitpick here is that some of the "If MSL is supported" has a uppercase letter after the comma. Not sure if it is worth addressing 🤷🏼

PR #35336, opened 2024-08-30
Move ACL checks before presence checks

### LINKED ISSUE #33735: Adjust the order of checks in the access control algorithm

tcarmelveilleux:
- Resolve a spec inconsistency

Testing done:

- All existing unit tests and integration tests pass

### DIFF
--- a/src/app/CommandHandlerImpl.cpp
+++ b/src/app/CommandHandlerImpl.cpp
@@ -389,17 +389,7 @@ Status CommandHandlerImpl::ProcessCommandDataIB(CommandDataIB::Parser & aCommand
     err = commandPath.GetConcreteCommandPath(concretePath);
     VerifyOrReturnError(err == CHIP_NO_ERROR, Status::InvalidAction);
 
-    {
-        Status commandExists = mpCallback->CommandExists(concretePath);
-        if (commandExists != Status::Success)
-        {
-            ChipLogDetail(DataManagement, "No command " ChipLogFormatMEI " in Cluster " ChipLogFormatMEI " on Endpoint 0x%x",
-                          ChipLogValueMEI(concretePath.mCommandId), ChipLogValueMEI(concretePath.mClusterId),
-                          concretePath.mEndpointId);
-            return FallibleAddStatus(concretePath, commandExists) != CHIP_NO_ERROR ? Status::Failure : Status::Success;
-        }
-    }
-
+    // Always check access first, before we even reply something about a supported path.
     {
         Access::SubjectDescriptor subjectDescriptor = GetSubjectDescriptor();
         Access::RequestPath requestPath{ .cluster     = concretePath.mClusterId,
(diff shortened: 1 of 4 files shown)


src/app/reporting/Engine.cpp
```diff
@@ -350,7 +338,7 @@ CHIP_ERROR Engine::CheckAccessDeniedEventPaths(TLV::TLVWriter & aWriter, bool &
         {
             ReturnErrorOnFailure(err);
         }
-        else
+
```
bzbarsky-apple, line 341: This change looks wrong to me.  This will encode status instead of the event if `err` is success, for example!  I _really_ hope this fails CI.

src/app/reporting/Engine.cpp
```diff
@@ -367,6 +355,20 @@ CHIP_ERROR Engine::CheckAccessDeniedEventPaths(TLV::TLVWriter & aWriter, bool &
                           current->mValue.mEndpointId, ChipLogValueMEI(current->mValue.mClusterId),
                           ChipLogValueMEI(current->mValue.mEventId), err == CHIP_ERROR_ACCESS_DENIED ? "ACL" : "ARL");
         }
+
+        Status status = CheckEventSupportStatus(path);
```
bzbarsky-apple, line 359: Followup: CheckEventSupportStatus should not actually exist, as far as I can tell, and certainly should not be called from here.

src/app/CommandHandlerImpl.cpp
```diff
@@ -420,6 +410,17 @@ Status CommandHandlerImpl::ProcessCommandDataIB(CommandDataIB::Parser & aCommand
         }
     }
 
+    {
+        Status commandExists = mpCallback->CommandExists(concretePath);
```
bzbarsky-apple, line 414: Followup: the spec has the fabric-scoped check before the existence check, whereas we have it after the timed invoke check.  I actually think the spec is wrong, because for commands that don't exist, how is the server supposed to know whether they are fabric-scoped?  We should fix the spec to do that check right after the existence checks, and then update this code.

src/app/CommandHandlerImpl.cpp
```diff
@@ -566,6 +559,15 @@ Status CommandHandlerImpl::ProcessGroupCommandDataIB(CommandDataIB::Parser & aCo
                 continue;
             }
         }
+
+        if (mpCallback->CommandExists(concretePath) != Status::Success)
```
tcarmelveilleux, line 563: Even though technically speaking this existence check ought to be done the before the ACL check according to spec (in a very general sense, since concurrency or reordering are not relevant if no observable side-effects of responses can arise, such as in the group case), we are doing it after the ACL check to avoid a future where allowing responses from group commands could leak existence.

src/app/reporting/Engine.cpp
tcarmelveilleux: This is indeed wrong, let me fix

src/app/reporting/Engine.cpp
bzbarsky-apple: I was wrong; things are a bit complicated here.  Resolving, but we need to get the spec to define how subscription reports get generated and what checks are done at that point....

Alami-Amine:
The work in this PR served as a foundation for the following PRs, which now replace it:
   - https://github.com/project-chip/connectedhomeip/pull/39415
   - https://github.com/project-chip/connectedhomeip/pull/40173
   - https://github.com/project-chip/connectedhomeip/pull/39772
   - https://github.com/project-chip/connectedhomeip/pull/40355

PR #35337, opened 2024-08-30
Fix Darwin CI random failures.

bzbarsky-apple:
We were starting an example app up front, then commissioning it when the test ran, but sometimes that takes longer than 15 minutes.

The fix is to start example apps as needed as parts of the pairing tests that need them.

The logging changes in MTRTestCase and MTRTestServerAppRunner are from trying to figure out why things were not being torn down right.  That's also what the added super-calls for setUp and tearDown fix.


kiel-apple (approved):
🎉

ksperling-apple (approved):

bzbarsky-apple:

PR #35338, opened 2024-08-30
Make DM provider and ember encode errors the same 

andy31415:
#### Changes

- In ember, do not optimistically encode the path to the attribute when the attribute read may still fail (realistically only happens if external callback fails)
- Add separate unit testing attributes in the unit testing cluster and make the linux all-clusters app implement support for those callbacks (NOTE: other all-clusters applications will NOT properly support those arguments as they rely on an external attribute path support)
- Add integration test to exercise this new code path.
- Removed path hardcoding from TestTimeSyncTrustedTimeSourceRunner.py and allow control of it through command line arguments


src/app/zap-templates/zcl/data-model/chip/test-cluster.xml
```diff
@@ -245,6 +245,13 @@ limitations under the License.
     <attribute side="server" code="0x4033" define="NULLABLE_GLOBAL_ENUM" type="TestGlobalEnum" isNullable="true" writable="true">nullable_global_enum</attribute>
     <attribute side="server" code="0x4034" define="NULLABLE_GLOBAL_STRUCT" type="TestGlobalStruct" isNullable="true" writable="true">nullable_global_struct</attribute>
 
+    <!-- These attributes are intended to return internal failures when read.
+         They are internal SDK-test code that are intended to only run trhrough EMBER interfaces instead of
```
bzbarsky-apple, line 249:
```suggestion
         They are internal SDK-test code that are intended to only run through EMBER interfaces instead of
```

examples/all-clusters-app/linux/main-common.cpp
```diff
@@ -336,3 +337,29 @@ void emberAfThermostatClusterInitCallback(EndpointId endpoint)
 
     SetDefaultDelegate(endpoint, &delegate);
 }
+
+using namespace chip::app;
```
bzbarsky-apple, line 341: That's already at the top of the file.

examples/all-clusters-app/linux/main-common.cpp
```diff
@@ -336,3 +337,29 @@ void emberAfThermostatClusterInitCallback(EndpointId endpoint)
...
+
+    VerifyOrReturnValue(clusterId == Clusters::UnitTesting::Id, chip::Protocols::InteractionModel::Status::Failure);
+    VerifyOrReturnValue(attributeMetadata != nullptr, chip::Protocols::InteractionModel::Status::Failure);
+
+    if (attributeMetadata->attributeId == Clusters::UnitTesting::Attributes::FailureInt32U::Id)
+    {
+        uint8_t forced_code = 0;
+        chip::Protocols::InteractionModel::Status status;
+
+        status = chip::app::Clusters::UnitTesting::Attributes::ReadFailureCode::Get(endpoint, &forced_code);
```
bzbarsky-apple, line 355:
```suggestion
        status = Clusters::UnitTesting::Attributes::ReadFailureCode::Get(endpoint, &forced_code);
```

examples/all-clusters-app/linux/main-common.cpp
```diff
@@ -336,3 +337,29 @@ void emberAfThermostatClusterInitCallback(EndpointId endpoint)
...
 }
+
+using namespace chip::app;
+chip::Protocols::InteractionModel::Status emberAfExternalAttributeReadCallback(EndpointId endpoint, ClusterId clusterId,
+                                                                               const EmberAfAttributeMetadata * attributeMetadata,
+                                                                               uint8_t * buffer, uint16_t maxReadLength)
+{
+
+    VerifyOrReturnValue(clusterId == Clusters::UnitTesting::Id, chip::Protocols::InteractionModel::Status::Failure);
+    VerifyOrReturnValue(attributeMetadata != nullptr, chip::Protocols::InteractionModel::Status::Failure);
```
bzbarsky-apple, line 348: This function could use a `using chip::Protocols::InteractionModel::Status` at the top and `Status::Failure` etc after that.

src/app/zap-templates/zcl/data-model/chip/test-cluster.xml
```diff
@@ -245,6 +245,13 @@ limitations under the License.
...
+         code-backed support (to be able to test ember-based error paths)
+      -->
+    <attribute side="server" code="0x4100" define="INT8U" type="int8u" writable="true" min="1" max="0xcb" default="1" optional="true">readFailureCode</attribute>
+    <attribute side="server" code="0x4101" define="INT32U" type="int32u" writable="true" default="0" optional="true">failureInt32U</attribute>
```
bzbarsky-apple, line 253: Please don't use ids in the 0x4... range for these things.  We've been trying to use that for nullable versions of attributes with lower-numbered ids.

Could put the new thing before the nullable attributes, or just start with 0x6 or something.

src/python_testing/TestUnitTestingErrorPath.py
bzbarsky-apple, line 1: Why is this a Python test instead of just a test in TestCluster.yaml?  This is not a test that needs Python power/complexity.

src/app/zap-templates/zcl/data-model/chip/test-cluster.xml
andy31415: My first pick was the next of 0x5000 and it turns out that the range has to be lower than 0x4FFF ...

so I must be between 0 and 0x4FFFF ... I could use 1,2 or 3 I guess.

src/python_testing/TestUnitTestingErrorPath.py
andy31415: My test for the original issue reporting was in python, this is easier to convert over. I already had the python version, did not want to try to learn how the yaml syntax works and more importantely how to run/test those locally.

src/app/zap-templates/zcl/data-model/chip/test-cluster.xml
andy31415: will switch to 0x3000 as a base.

andy31415: Updated to 0x3000/0x30001

src/python_testing/TestUnitTestingErrorPath.py
andy31415: Updated python test to actually pass locally (had missing text for asserts).

examples/all-clusters-app/linux/main-common.cpp
```diff
@@ -336,3 +337,29 @@ void emberAfThermostatClusterInitCallback(EndpointId endpoint)
 
     SetDefaultDelegate(endpoint, &delegate);
 }
+
+using chip::Protocols::InteractionModel::Status;
```
bzbarsky-apple, line 341: I meant put this _inside_ the function, so the namespace import is function-local.

If we're going to do it for "everything after this point", just put it at the top of the file with the other "global" `using`.

src/python_testing/TestUnitTestingErrorPath.py
bzbarsky-apple, line 1: Well, this needs review from someone who is not me who knows the Python test bits.  But:

1. The copyright on this is definitely not 2023.
2. I really think we should not be using Python (with its complexity and lack of readablity for our test frameworking) for things where YAML will do.  If you need help converting this to a YAML test, I am happy to help.  And YAML tests are _much_ simpler to run locally than the Python ones...

tcarmelveilleux (approved):

.github/workflows/tests.yaml
```diff
@@ -475,11 +475,18 @@ jobs:
                   mkdir objdir-clone || true
 
             - name: Build Python REPL and example apps
+              # NOTE: the data-mode-check + check-failure-die is not 100% perfect as different
+              #
```
tehampson, line 479: nit, is there a reason for this extra space?

src/python_testing/TestUnitTestingErrorPath.py
```diff
@@ -0,0 +1,102 @@
...
+class TestUnitTestingErrorPath(MatterBaseTest):
+
+    @async_test_body
+    async def test_unit_test_error_read(self):
+        endpoint_id = 1
+        attributes = Clusters.UnitTesting.Attributes
+
+        self.print_step(0, "Commissioning - already done")
+
+        self.print_step(1, "Set error to failure.")
```
tehampson, line 50: nit, It would be helpful for people reading the test to be a little more verbose in the test steps to understand the follow more quickly

For example I don't really get why you are writing to `ReadFailureCode` in step 1, but then trying to read `FailureInt32U` in step 2

It would be more helpful if step 2 explicitly stated confirming attribute not written to in step 1 remains unset

tehampson (approved):

bzbarsky-apple (approved):

.github/workflows/tests.yaml
andy31415: Inadvertent enter press. Removed.

src/python_testing/TestUnitTestingErrorPath.py
andy31415: Updated steps text to make it clear on what interactions are expected.

PR #35343, opened 2024-09-01
[ESP32] Enable TCP on ESP32 platform

sayondeep:
Enable TCP on ESP32 platform

### Testing:

Tested with camera application for esp32 platform
CI green


src/platform/ESP32/ESP32TCPEndPointImplLwIP.cpp
```diff
@@ -0,0 +1,1036 @@
...
+        // (by posting events to the thread's event queue) we can't do that here because as soon as this
+        // function returns, LwIP is free to begin calling callbacks on the new PCB. For that to work we need
+        // to have an end point associated with the PCB.
+        //
+        if (err == CHIP_NO_ERROR)
+        {
+            TCPEndPoint * connectEndPoint = nullptr;
+            {
+                //lock CHIP stack
+                chip::DeviceLayer::StackLock lock;
```
sayondeep, line 868: I had to copy TCPEndPointImplLwIP.cpp from src/inet to src/platform/ESP32 since it is required to use chip stack lock here. It is checked here src/inet/InetLayer.h:82 (https://github.com/project-chip/connectedhomeip/blob/1d7477934bbfaaaba3556d60a617c071830acf42/src/inet/InetLayer.h#L82)

pidarped: What are the differences between this file and the default inet/TCPEndPointImplLwIP.cpp? We shouldn't have to make a copy for ESP32 platform but only have one for all LwIP based platforms.
The locking should happen higher up at the application layer so lower down we have the assert checks to ensure that the lock is held.

sayondeep: Differences:
1) Added LOCK_TCPIP_CORE() and UNLOCK_TCPIP_CORE() in TCPEndPointImplLwIP::ListenImpl due to `assert failed: tcp_listen_with_backlog /IDF/components/lwip/lwip/src/core/tcp.c:834 (Required to lock TCPIP core functionality!)`

2) Added lock in TCPEndPointImplLwIP::LwIPHandleIncomingConnection as mentioned above due to 
```
E (17377) chip[DL]: Chip stack locking error at 'src/inet/InetLayer.h:82'. Code is unsafe/racy
E (17377) chip[-]: chipDie chipDie chipDie
```
I had to make a copy because it was not allowed to include <platform/PlatformManager.h>
```
ERROR at //third_party/connectedhomeip/src/inet/TCPEndPointImplLwIP.cpp:32:11: Include not allowed.
#include <platform/PlatformManager.h>
          ^-------------------------
  //third_party/connectedhomeip/src/inet:inet
```

pidarped: The TCPEndPointImplLwIP is a common piece of code for all platforms using LwIP. It is not a virtual interface that needs to be implemented by individual platforms.
For 1), The LWIP core lock can be added in TCPEndPointImplLwIP.cpp instead of this copy in ESP32, no?
For 2) have you tried acquiring the CHIP Stack lock higher up, maybe around InitChipStack()?

pidarped:
@sayondeep, could you please resolve the restyle issue so that the CI build completes?

src/platform/ESP32/ESP32TCPEndPointImplLwIP.cpp
sayondeep: This [PR](https://github.com/project-chip/connectedhomeip/pull/36962) fixed the issue.

wqx6 (approved):

shubhamdp (approved):

andy31415 (approved):

PR #35344, opened 2024-09-02
[ESP32] Add a configurable device type for commissionable advertising.

Jerry-ESP:
For matter certify test case TC-SC-4.1, step 7, if we support advertise the device type when the commissioning window opened, we should provide the DUT's primary device type. 

So add the interface in esp32 to config the correct primary device type in flash. 

And update the generate_esp32_chip_factory_bin.py  tool to specify the device when generate the factory.bin.

### DIFF
--- a/src/platform/ESP32/ConfigurationManagerImpl.cpp
+++ b/src/platform/ESP32/ConfigurationManagerImpl.cpp
@@ -271,24 +271,19 @@ CHIP_ERROR ConfigurationManagerImpl::GetLocationCapability(uint8_t & location)
 
 CHIP_ERROR ConfigurationManagerImpl::GetDeviceTypeId(uint32_t & deviceType)
 {
-#if CHIP_DEVICE_CONFIG_ENABLE_COMMISSIONABLE_DEVICE_TYPE
     uint32_t value = 0;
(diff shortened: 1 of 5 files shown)


src/platform/ESP32/ConfigurationManagerImpl.cpp
```diff
@@ -241,6 +241,28 @@ CHIP_ERROR ConfigurationManagerImpl::GetLocationCapability(uint8_t & location)
...
+    }
+    else
+    {
+        deviceType = CHIP_DEVICE_CONFIG_DEVICE_TYPE;
+    }
+
+    return err;
+#else
+    deviceType = 0;
+    return CHIP_NO_ERROR;
```
bzbarsky-apple, line 262: 0 is not a valid value; why is it being returned with CHIP_NO_ERROR?

Should this just call the superclass to get whatever behavior that has?  Or should this whole declaration/implementation be conditioned on CHIP_DEVICE_CONFIG_ENABLE_COMMISSIONABLE_DEVICE_TYPE or something?

Jerry-ESP: Update

andy31415 (approved):

PR #35349, opened 2024-09-02
Test: TC-CNET-4-23 for switching network interface

DejinChen:
Ref to https://github.com/CHIP-Specifications/chip-test-plans/pull/4598, add tests for switching network interface.

### Testing:
Test lighting-app with `SecondaryNetworkInterface` device type on esp32c6.

Run test:
```
python src/python_testing/TC_CNET_4_23.py --discriminator <discriminator> --passcode <passcode> \
               --endpoint <endpoint_value> \
               --string-arg PIXIT.CNET.WIFI_1ST_ACCESSPOINT_SSID:<wifi_ssid> \
               --string-arg PIXIT.CNET.WIFI_1ST_ACCESSPOINT_CREDENTIALS:<wifi_credentials> \
               --string-arg PIXIT.CNET.THREAD_1ST_OPERATIONALDATASET:<thread_dataset>
```
Where `<endpoint_value>` should be replaced with the actual endpoint number under test for the Network Commissioning cluster on the DUT.

### DIFF
--- a/src/python_testing/TC_CNET_4_23.py
+++ b/src/python_testing/TC_CNET_4_23.py
@@ -112,9 +112,9 @@ class TC_CNET_4_23(MatterBaseTest):
         for endpoint, value in feature_map_response.items():
             feature_dict = value.get(Clusters.Objects.NetworkCommissioning, {})
             feature_map = feature_dict.get(Clusters.Objects.NetworkCommissioning.Attributes.FeatureMap, None)
-            if feature_map == kWiFiFeature:
+            if bool(feature_map & Clusters.NetworkCommissioning.Bitmaps.Feature.kWiFiNetworkInterface):
                 wifi_endpoint = endpoint
-            elif feature_map == kThreadFeature:
+            elif bool(feature_map & Clusters.NetworkCommissioning.Bitmaps.Feature.kThreadNetworkInterface):
                 thread_endpoint = endpoint
             else:
                 continue
--- a/src/python_testing/TC_CNET_4_23.py
+++ b/src/python_testing/TC_CNET_4_23.py
@@ -1,0 +1,221 @@
+#
+#    Copyright (c) 2025 Project CHIP Authors
+#    All rights reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+import logging
+import time
+
+import chip.clusters as Clusters
+from chip.exceptions import ChipStackError
+from chip.testing.matter_testing import MatterBaseTest, TestStep, async_test_body, default_matter_test_main
+from mobly import asserts
+
+kRootEndpointId = 0
+kInvalidEndpointId = 0xFFFF
+kWiFiFeature = 1
+kThreadFeature = 2
+kExpiryLengthSeconds = 900
+kChipErrorTimeout = 0x32
+kMaxCommissioningCompleteRetryTimes = 5
+
+
+class TC_CNET_4_23(MatterBaseTest):
+    """
+    [TC-CNET-4.23] [Wi-Fi and Thread] Verification for network interface switching between Wi-Fi and Thread [DUT-Server].
+    Example Usage:
+        To run the test case, use the following command:
+        ```bash
+        python src/python_testing/TC_CNET_4_23.py --commissioning-method ble-wifi -discriminator <discriminator> -passcode <passcode> \
+               --wifi-ssid <wifi_ssid> --wifi-passphrase <wifi_credentials> --hex-arg PIXIT.CNET.THREAD_1ST_OPERATIONALDATASET:<thread_dataset>
+        ```
+        Where `<dataset_value>` should be replaced with the Thread Operational Dataset
+    """
+
+    def steps_TC_CNET_4_23(self):
+        return [TestStep("precondition", "TH is commissioned", is_commissioning=True),
+                TestStep(1, 'TH sends ArmFailSafe command to the DUT with ExpiryLengthSeconds set to 900. Verify that DUT sends ArmFailSafeResponse command to the TH'),
+                TestStep(2, 'TH reads Networks attribute on endpoint PIXIT.CNET.ENDPOINT_WIFI from the DUT. Verify that the Networks attribute list has an entry with the following fields: network_id matches the bytes in PIXIT.CNET.WIFI_1ST_ACCESSPOINT_SSID, Connected is of type bool and is TRUE'),
+                TestStep(3, 'TH finds the index of the Networks list entry with network_id for PIXIT.CNET.WIFI_1ST_ACCESSPOINT_SSID and saves it as `Userwifi_netidx`'),
+                TestStep(4, 'TH sends RemoveNetwork Command to the DUT PIXIT.CNET.ENDPOINT_WIFI endpoint with network_id field set to PIXIT.CNET.WIFI_1ST_ACCESSPOINT_SSID. Verify that DUT sends NetworkConfigResponse to command with the following fields: NetworkingStatus is Success, NetworkIndex matches previously saved `Userwifi_netidx`'),
+                TestStep(5, 'TH sends AddOrUpdateThreadNetwork command to the DUT PIXIT.CNET.ENDPOINT_THREAD endpoint with operational dataset field set to PIXIT.CNET.THREAD_1ST_OPERATIONALDATASET. Verify that DUT sends the NetworkConfigResponse command to the TH with the following fields: NetworkingStatus is Success, DebugText is of type string with max length 512 or empty'),
+                TestStep(6, 'TH reads Networks attribute from the DUT PIXIT.CNET.ENDPOINT_THREAD endpoint. Verify that the Networks attribute list has an entry with the following fields: network_id is the extended PAN ID of PIXIT.CNET.THREAD_1ST_OPERATIONALDATASET, Connected is of type bool and is FALSE'),
+                TestStep(7, 'TH sends ConnectNetwork command to the DUT PIXIT.CNET.ENDPOINT_THREAD endpoint with network_id field set to the extended PAN ID of PIXIT.CNET.THREAD_1ST_OPERATIONALDATASET'),
+                TestStep(8, 'TH sends the CommissioningComplete command to the DUT. Verify that DUT sends CommissioningCompleteResponse with the ErrorCode field set to OK'),
+                TestStep(9, 'TH reads Networks attribute from the DUT PIXIT.CNET.ENDPOINT_THREAD endpoint. Verify that the Networks attribute list has an entry with the following values: network_id is the extended PAN ID of PIXIT.CNET.THREAD_1ST_OPERATIONALDATASET, Connected is of type bool and is TRUE. TH saves the index of the Networks list entry as `Userth_netidx`'),
+                TestStep(10, 'TH sends ArmFailSafe command to the DUT with ExpiryLengthSeconds set to 900| Verify that DUT sends ArmFailSafeResponse command to the TH'),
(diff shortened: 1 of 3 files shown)


chshu (approved):

src/app/tests/suites/certification/Test_TC_CNET_4_23.yaml
```diff
@@ -0,0 +1,313 @@
...
+
+          Via the TH (chip-tool),  verify that DUT sends ArmFailSafe with timeout as 900 secs to the TH.
+
+          [1725257597.514] [1602947:1602949] [DMG] Received Command Response Data, Endpoint=0 Cluster=0x0000_0030 Command=0x0000_0001
+          [1725257597.514] [1602947:1602949] [TOO] Endpoint: 0 Cluster: 0x0000_0030 Command 0x0000_0001
+          [1725257597.514] [1602947:1602949] [TOO]   ArmFailSafeResponse: {
+          [1725257597.514] [1602947:1602949] [TOO]     errorCode: 0
+          [1725257597.514] [1602947:1602949] [TOO]     debugText:
+          [1725257597.514] [1602947:1602949] [TOO]    }
+      disabled: true
```
andy31415, line 50: manual tests are not accepted in the SDK anymore

andy31415 (changes requested):
All new SDK tests must be automated. Manual tests are not accepted in the SDK anymore

src/app/tests/suites/certification/Test_TC_CNET_4_23.yaml
DejinChen: Modified to auto test script

src/python_testing/test_metadata.yaml
```diff
@@ -19,6 +19,8 @@ not_automated:
       reason: It has no CI execution block, is not executed in CI
     - name: TC_CNET_4_22.py
       reason: It has no CI execution block, is not executed in CI
+    - name: TC_CNET_4_23.py
+      reason: It has no CI execution block, is not executed in CI
```
cecille:
```suggestion
      reason: test requires a real wifi and thread network connection, which is not available in the CI
```

src/python_testing/TC_CNET_4_23.py
```diff
@@ -0,0 +1,220 @@
...
+                TestStep(13, 'TH sends ConnectNetwork command to the DUT PIXIT.CNET.ENDPOINT_WIFI endpoint with network_id field set to PIXIT.CNET.WIFI_1ST_ACCESSPOINT_SSID'),
+                TestStep(14, 'TH sends the CommissioningComplete command to the DUT. Verify that DUT sends CommissioningCompleteResponse with the ErrorCode field set to OK'),
+                TestStep(15, 'TH reads Networks attribute from the DUT PIXIT.CNET.ENDPOINT_WIFI endpoint. Verify that the Networks attribute list has an entry with the following fields: network_id matches the bytes in PIXIT.CNET.WIFI_1ST_ACCESSPOINT_SSID, Connected is of type bool and is TRUE')]
+
+    def def_TC_CNET_4_23(self):
+        return '[TC-CNET-4.23] Verification for network interface switching between Wi-Fi and Thread [DUT-Server]'
+
+    def pics_TC_CNET_4_23(self):
+        return ["CNET.S.F00", "CNET.S.F01"]
```
cecille: The way this is marked, this test will never run. PICS are per endopoint, and you will never have an endpoint with both of these features.

You're breaking new ground here because I actually don't think we yet have any tests that test a cross-endpoint cluster dependency. Except groups, and there we know exactly where the key management cluster will be and that it will always exist.

I think here, what might make sense is to run this for every CNET cluster (CNET.S only), with a note in the test that the steps are only run if there are multiple network commissioning clusters. Right at the start of the test, the test could check for other endpoints with network commissioning clusters and just skip all the steps if one doesn't exist. In the dual-network situation, the device would need to be connected on the network represented by the network commissioning cluster on this endpoint and the test could check that (by checking the network list). Then if you run this against every endpoint, you'll see the test swap networks both ways. Or it could just swap the networks itself (which is rather the point of the test, but you do want to see this work both ways).

src/python_testing/TC_CNET_4_23.py
```diff
@@ -0,0 +1,220 @@
...
+        python src/python_testing/TC_CNET_4_23.py --commissioning-method ble-wifi -discriminator <discriminator> -passcode <passcode> \
+               --wifi-ssid <wifi_ssid> --wifi-passphrase <wifi_credentials> --hex-arg PIXIT.CNET.THREAD_1ST_OPERATIONALDATASET:<thread_dataset>
+        ```
+        Where `<dataset_value>` should be replaced with the Thread Operational Dataset
+    """
+
+    def steps_TC_CNET_4_23(self):
+        return [TestStep("precondition", "TH is commissioned", is_commissioning=True),
+                TestStep(1, 'TH sends ArmFailSafe command to the DUT with ExpiryLengthSeconds set to 900. Verify that DUT sends ArmFailSafeResponse command to the TH'),
+                TestStep(2, 'TH reads Networks attribute on endpoint PIXIT.CNET.ENDPOINT_WIFI from the DUT. Verify that the Networks attribute list has an entry with the following fields: network_id matches the bytes in PIXIT.CNET.WIFI_1ST_ACCESSPOINT_SSID, Connected is of type bool and is TRUE'),
```
cecille: For this test to work properly on all network clusters, you'll need to check for either wifi or thread depending on which network commissioning cluster you're using.

Does this also apply for ethernet?

src/python_testing/TC_CNET_4_23.py
```diff
@@ -0,0 +1,220 @@
...
+            except ChipStackError as e:  # chipstack-ok: After switching networks, it may resolve to an incorrect address
+                asserts.assert_equal(e.err, kChipErrorTimeout, "Unexpected error while trying to send CommissioningComplete")
+                logging.info(f"Will retry CommissioningComplete command in 10 seconds, attempt: {i + 1}")
+                time.sleep(10)
+        else:
+            asserts.assert_true(False, "CommissioningComplete command failed")
+
+    @async_test_body
+    async def test_TC_CNET_4_23(self):
+        asserts.assert_true('PIXIT.CNET.THREAD_1ST_OPERATIONALDATASET' in self.matter_test_config.global_test_params,
```
cecille: the thread dataset can be passed in directly just like the wifi ssid - it might make more sense to use that than the PIXIT here so you can use the same test arguments regardless of whether you're starting with the wifi endpoint or the thread endpoint.

src/python_testing/TC_CNET_4_23.py
```diff
@@ -0,0 +1,220 @@
...
+import time
+
+import chip.clusters as Clusters
+from chip.exceptions import ChipStackError
+from chip.testing.matter_testing import MatterBaseTest, TestStep, async_test_body, default_matter_test_main
+from mobly import asserts
+
+kRootEndpointId = 0
+kInvalidEndpointId = 0xFFFF
+kWiFiFeature = 1
```
cecille, line 28: These are part of the generated code - please use that.

src/python_testing/TC_CNET_4_23.py
```diff
@@ -0,0 +1,220 @@
...
+        wifi_ssid = self.get_wifi_ssid().encode("utf-8")
+        wifi_credential = self.get_credentials().encode("utf-8")
+
+        feature_map_response = await self.default_controller.ReadAttribute(self.dut_node_id, [(Clusters.NetworkCommissioning.Attributes.FeatureMap)], fabricFiltered=True)
+        wifi_endpoint = kInvalidEndpointId
+        thread_endpoint = kInvalidEndpointId
+        for endpoint, value in feature_map_response.items():
+            feature_dict = value.get(Clusters.Objects.NetworkCommissioning, {})
+            feature_map = feature_dict.get(Clusters.Objects.NetworkCommissioning.Attributes.FeatureMap, None)
+            if feature_map == kWiFiFeature:
```
cecille: be careful here - the PDC features are going to land and an equivalence test will break
```suggestion
            if bool(feature_map & Clusters.NetworkCommissioning.Bitmaps.Feature.kWiFiNetworkInterface):
```

src/python_testing/TC_CNET_4_23.py
```diff
@@ -0,0 +1,220 @@
...
+                wifi_endpoint = endpoint
+            elif feature_map == kThreadFeature:
```
cecille:
```suggestion
            elif bool(feature_map & Clusters.NetworkCommissioning.Bitmaps.Feature.kThreadNetworkInterface):
```

src/python_testing/TC_CNET_4_23.py
cecille: Actually, you do that already, so it's mostly just the PICS that's the problem

src/python_testing/TC_CNET_4_23.py
```diff
@@ -0,0 +1,220 @@
...
+        self.step(3)
+        index = 0
+        Userwifi_netidx = 0
+        verification = False
+        for network_info in networks_response:
+            if network_info.connected and network_info.networkID == wifi_ssid:
+                Userwifi_netidx = index
+                verification = True
+            index += 1
+        asserts.assert_true(verification, "There is no correct entry in Networks attribute list")
```
cecille: You can collapse this down to one line 
Though it should probably target either the wifi or the thread, depending on which endpoint you're targeting.

Userwifi_netidx = next((i for i, info in enumerate(network_response) if info.connected and info.networkID == wifi_ssid), None)
asserts.assert_true(Userwifi_netidx != None, "msg")

src/python_testing/TC_CNET_4_23.py
```diff
@@ -0,0 +1,220 @@
...
+
+    async def SendConnectNetworkWithFailure(
+            self, networkID: int, endpoint: int):
+        with asserts.assert_raises(ChipStackError) as cm:
+            cmd = Clusters.NetworkCommissioning.Commands.ConnectNetwork(networkID=networkID)
+            await self.send_single_cmd(endpoint=endpoint, cmd=cmd)
+        asserts.assert_equal(cm.exception.err, kChipErrorTimeout, "Unexpected error while trying to send ConnectNetwork command")
+        logging.info("ConnectNetwork command timed out, not assert here as it is expected during network switching")
+
+    async def SendCommissioningCompleteWithRetry(self):
```
cecille, line 121: I believe the reason you need the retries here is because you actually need to re-establish a new case session on the thread network. The retry here is waiting for the underlying session to error out and re-establish, but you can just force that to happen by expiring the session explicitly. This is done in some of the other CNET tests.

DejinChen: CommissioningCommplete is sent after ConnectNetwork, the old CASE will be expired during the response of ConnectNetwork command. When switching network from Thread to Wi-Fi, we should remove the SRP before disconnect Thread network or there will be a stale SRP, but unfortunately, we don't remove it. I add a [PR](https://github.com/project-chip/connectedhomeip/pull/40012) for this, but it is not async version. I add retry here to avoid parsing address from the stale SRP. If addresses from both SRP and Wi-Fi are parsed, it will use the one from Wi-Fi(linklocal).

src/python_testing/TC_CNET_4_23.py
DejinChen: For a multiple network device (e.g. endpoint 0:Wi-Fi, endpoint 1:Thread), 
If use PICS "CNET.S ", then
1. for endpoint 0, use "ble-wifi" commissioning method, network switching process: Wi-Fi (commissioning state) -> Thread -> Wi-Fi
2. for endpoint 1, use "ble-thread" commissioning method, network switching process: Thread (commissioning state) -> Wi-Fi -> Thread

If apply this, the test depends on the endpoint's feature, and more likely it should be divided into two tests (one test for PICS CNET.S.F00, and the other for CNET.S.F01).

Or we just use PICS "CNET.S.F00" for this test and only point 1 should be test? Wi-Fi -> Thread and Thread -> Wi-Fi are both tested in it, but only "ble-wifi" commissioning method is used.

src/python_testing/TC_CNET_4_23.py
DejinChen: Done. NOT applied for ethernet.

src/python_testing/TC_CNET_4_23.py
DejinChen: Done.

src/python_testing/TC_CNET_4_23.py
DejinChen: Have modified the process:
1. don't commissioning the DUT automatically
2. establish a PASE with the DUT over BLE
3. decide which network type on the current endpoint under test and commissioning the DUT on it
4. exec switching network process

src/python_testing/TC_CNET_4_23.py
DejinChen: It seems that [wifi info is only parsed for "ble-wifi" commissioning method](https://github.com/project-chip/connectedhomeip/blob/3593dd5f5d5fbba12afc87efe994ec82a9abde0a/src/python_testing/matter_testing_infrastructure/chip/testing/runner.py#L702), as well as Thread dataset. Use this PIXIT like [TC_CNET_4_12](https://github.com/project-chip/connectedhomeip/blob/3593dd5f5d5fbba12afc87efe994ec82a9abde0a/src/python_testing/TC_CNET_4_12.py#L42).

DejinChen:
@cecille PTAL

PR #35361, opened 2024-09-02
Update TC-CC-2_2 to match latest test plan

ReneJosefsen:
Once https://github.com/project-chip/connectedhomeip/pull/35224 lands, this PR updates the test script to align with the latest implementation and test plan.

The script is tested on a macOS light running build with the changes in PR 35224

Spec change: https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/10231

Test plan change: https://github.com/CHIP-Specifications/chip-test-plans/pull/4527


src/python_testing/TC_CC_2_2.py
```diff
@@ -48,62 +48,80 @@ def default_timeout(self) -> int:
...
 
-        def store_values(attr: str) -> str:
-            return f"TH stores the reported values of _{attr}_ in all incoming reports for _{attr}_ attribute, that contains data in _reportedCurrentHueValuesList_, over a period of 20 seconds."
+        def store_values(reportList: str, attr: str) -> str:
+            return f"TH stores the reported values of _{attr}_ in all incoming reports for _{attr}_ attribute, that contains data in _{reportList}_, over a period of 20 seconds."
 
-        def verify_entry_count(attr: str) -> str:
-            return f'TH verifies that _reportedCurrentHueValuesList_ does not contain more than 10 entries for _{attr}_'
+        def verify_entry_count(reportList: str, attr: str) -> str:
+            return f'TH verifies that _{reportList}_ does not contain more than 12 entries for _{attr}_'
```
tcarmelveilleux, line 55: All the {} will break Python string handling at runtime. Remove the `f` prefix

tcarmelveilleux (approved):

src/python_testing/TC_CC_2_2.py
ReneJosefsen: Not sureI understand, for this specifically the f was already there due to the "attr" string, I only added which list it is stored in so match TP?

yunhanw-google (approved):

src/python_testing/TC_CC_2_2.py
tcarmelveilleux: Please avoid using things like `{THCommand}` in the future, and just write it out.

tcarmelveilleux (approved):

jmartinez-silabs (approved):

src/python_testing/TC_CC_2_2.py
ReneJosefsen: Oh you are thinking about these, sure. We have likely not landed on the perfect solution for this but it is to accommodate the use of document attributes in the test plan and I think @cecille did some stuff to address this on the python side as well, but not sure. But that is not really a new thing that I added in this PR though 🤷🏻‍♂️

PR #35366, opened 2024-09-03
[Silabs] Fixing the switch ladder and removing warning flags

chirag-silabs:
Fixing code bugs
1. Fixing the switch case ladder and removed the VerifyOrReturn
2. Deleting all the fabrics on the factory reset to get rid of the multicast service.
3. Removing all the warning cflags for siwx917 soc


examples/platform/silabs/BaseApplication.cpp
```diff
@@ -802,6 +802,8 @@ void BaseApplication::ScheduleFactoryReset()
         {
             Provision::Manager::GetInstance().SetProvisionRequired(true);
         }
+        // Delete all fabrics.
+        chip::Server::GetInstance().GetFabricTable().DeleteAllFabrics();
```
lpbeliveau-silabs, line 806: Omitting this is intentional, see `examples/platform/silabs/BaseApplication.h` line 245:
`   /**
     * @brief Start the factory Reset process
     *  Almost identical to Server::ScheduleFactoryReset()
     *  but doesn't call GetFabricTable().DeleteAllFabrics(); which deletes Key per key.
     *  With our KVS platform implementation this is a lot slower than deleting the whole kvs section
     *  our silabs nvm3 driver which end up being doing in ConfigurationManagerImpl::DoFactoryReset(intptr_t arg).
     */ `

lpbeliveau-silabs (changes requested):

examples/platform/silabs/BaseApplication.cpp
lpbeliveau-silabs: Since wifi is using a different nvm3 driver, I suggest using `Server::ScheduleFactoryReset() ` inside an `#idfef` for wifi rather than modifying `BaseApplication::ScheduleFactoryReset()`

chirag-silabs: Added the wifi condition since it is needed only for wifi.Not using Server::ScheduleFactoryReset since we still need the setProvisionRequired event. 
Not adding this in the src/platform/silabs/configurationManager.cpp since it will create a circular dependency by including app in the platform

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -802,6 +802,10 @@ void BaseApplication::ScheduleFactoryReset()
         {
             Provision::Manager::GetInstance().SetProvisionRequired(true);
         }
+#if SL_WIFI
+        // Removing the matter services on factory reset
+        chip::Dnssd::ServiceAdvertiser::Instance().RemoveServices();
```
lpbeliveau-silabs, line 807: We clear our thread credentials inside of `ConfigurationManagerImpl::DoFactoryReset()`, there is even a wifi ifdefed section in there, it would make sense to put the wifi clear at the same place.

You could also create a wifi helper equivalent to  `ConfigurationManagerImpl::ClearThreadStack()` for wifi so that all the wifi factory reset is handled at the same place.

lpbeliveau-silabs: After discussion, it makes sense to keep this here as our Dnssd is entirely handled by the app server.

lpbeliveau-silabs (approved):

tehampson (approved):

PR #35395, opened 2024-09-04
[Docker Image] remove workaround for bloaty build

### LINKED ISSUE #35280: [Docker] Remove cloning of abseil from base docker image
This is a follow-up to #35277

- the cloning of abseil-cpp should be removed to keep the docker base image lightweight
- This should be done after we can build Bloaty using their own abseil-cpp submdoule instead of a system one.
- Issue opened at Bloaty: https://github.com/google/bloaty/issues/386

Alami-Amine:
fixes #35280
- Bloaty integrated their PR that bumps  use thier abseil submodule; allowing us to use it instead of a system-based copy

### DIFF
--- a/integrations/docker/images/base/chip-build/Dockerfile
+++ b/integrations/docker/images/base/chip-build/Dockerfile
@@ -123,14 +123,6 @@ RUN set -x \
     ruff \
     && : # last line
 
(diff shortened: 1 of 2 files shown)


arkq (approved):

tehampson (approved):

andy31415 (changes requested):
You must change version file for every docker file change

andy31415 (approved):

PR #35403, opened 2024-09-04
[devcontainer] Version bump and enhanced UX.

XToripuru:
### Issue

1. The current devcontainer image version is **22**, while the latest available version is **74**. Upgrading to the latest version will resolve several compatibility issues.

2. The directory of bootstrap files is hardcoded in the base image but it should be *visibly* configurable.

3. The devcontainer build process is inefficient, with a significant amount of time spent on recursively `chown`ing unnecessary files.

### Proposed Solution

1. Update the `devcontainer.json` to use the latest image version. Additionally, remove some unnecessary dependencies in `Dockerfile`.

2. Move the declaration of `PW_ENVIRONMENT_ROOT` environment variable from base image to `devcontainer.json`.

3. Eliminate certain `chown` operations, as `git config --global --add safe.directory "*"` with other minor changes provide a faster and sufficient alternative.

### Extra notes

apt installs from the `.devcontainer/Dockerfile` should be moved to the base image. This PR adds these apt installs to the base image and flags the ones in `.devcontainer/Dockerfile` to be removed in the next version when this gets merged.

### Testing

Successfully compiled a subset of examples for the affected build targets.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -24,18 +24,17 @@ ARG USER_UID=1000
 ARG USER_GID=$USER_UID
 ENV LANG en_US.utf8
 
-# TODO: remove these apt installs in the future as they will be in the base image from version 75+.
+
 RUN apt-get update \
     && apt-get install -y locales \
     && localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8 \
-    && apt-get -fy install git vim emacs sudo \
+    && apt-get -fy install vim emacs sudo \
     apt-utils dialog zsh \
-    iproute2 procps lsb-release \
+    lsb-release \
     bash-completion \
-    build-essential cmake cppcheck valgrind \
-    wget curl telnet \
+    valgrind \
     docker.io \
-    iputils-ping net-tools \
+    iputils-ping \
     && :
 
 RUN groupadd -g $USER_GID $USERNAME \
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -25,6 +25,8 @@ ARG USER_GID=$USER_UID
 ENV LANG en_US.utf8
 
 
+# These are installed for terminal/dev convenience. If more tooling for build is required, please
+# add them to chip-build (in integrations/docker/images/chip-build).
 RUN apt-get update \
     && apt-get install -y locales \
     && localedef -i en_US -c -f UTF-8 -A /usr/share/locale/locale.alias en_US.UTF-8 \
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -51,13 +51,22 @@ RUN curl https://raw.githubusercontent.com/restyled-io/restyler/master/bin/resty
 RUN mkdir -p /opt/sdk/sdks/ \
     && chown -R $USERNAME:$USERNAME \
     /opt/sdk/sdks/ `# NXP uses a patch_sdk script to change SDK files` \
-    /opt/NordicSemiconductor/nrfconnect/ `# $USERNAME needs to own west configuration to build nRF Connect examples` \
-    $IDF_PATH `# $USERNAME needs to own the esp-idf and tools for the examples to build` \
+    $ANDROID_HOME \
     $IDF_TOOLS_PATH \
-    $SYSROOT_AARCH64 `# allow read/write access to header and libraries` \
-    $ANDROID_HOME `# allow licenses to be accepted` \
-    $AMEBA_PATH `# AmebaD requires access to change build_info.h` \
-    $IMX_SDK_ROOT \
+    && find $AMEBA_PATH -name "inc_lp" -print0 | xargs -0 chown -R $USERNAME:$USERNAME \
+    && find $AMEBA_PATH -name "inc_hp" -print0 | xargs -0 chown -R $USERNAME:$USERNAME \
+    && find $AMEBA_PATH -name "project_lp" -print0 | xargs -0 chown -R $USERNAME:$USERNAME \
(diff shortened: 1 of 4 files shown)


integrations/docker/images/vscode/chip-build-vscode/Dockerfile
```diff
@@ -106,6 +106,16 @@ RUN set -x \
     telnet \
     srecord \
     openjdk-8-jdk \
+    git vim emacs sudo \
```
andy31415, line 109: There is a comment above that explained the previous choices.

We need to explain these additional ones as well. The vscode image is meant for `build` at this point but not as a full fledged development environment. 

Beyond that, any changes in Dockerfiles must be acompanied by a version bump. However overall I do not think we should have many of these items here (and if we do it seems we want them grouped/ordered and one per line for clear updates if we ever update them)

integrations/docker/images/vscode/chip-build-vscode/Dockerfile
```diff
@@ -130,7 +140,6 @@ ENV NXP_K32W0_SDK_ROOT=/opt/k32w/core
 ENV NXP_K32W1_SDK_ROOT=/opt/k32w/k32w1
 ENV NXP_SDK_ROOT=/opt/nxp-sdk/rw61x
 ENV OPENOCD_PATH=/opt/openocd/
-ENV PW_ENVIRONMENT_ROOT=/home/vscode/pigweed/env
 ENV QEMU_ESP32=/opt/espressif/qemu/qemu-system-xtensa
```
andy31415, line 143: This means that vscode images would default to `.environment` which is not desirable if run from another image (e.g. I use podman and not containers from vscode).

andy31415 (changes requested):
This probably should be split between devcontainer changtes and dockerfile changes.

I have reservations about the dockerfile changes - it makes it more towards a generic full fledged development setup, however I believe that we may want to figure out some different paths for that as everyone may have their own development environment and scripts.

integrations/docker/images/vscode/chip-build-vscode/Dockerfile
XToripuru: Makes sense. I reverted these apt installs and removed unneeded ones.

.devcontainer/devcontainer.json
```diff
@@ -14,9 +14,12 @@
     "mounts": [
         "source=/var/run/docker.sock,target=/var/run/docker.sock,type=bind"
     ],
-    "initializeCommand": "bash .devcontainer/build.sh --tag matter-dev-environment:local --version 22",
+    "initializeCommand": "bash .devcontainer/build.sh --tag matter-dev-environment:local --version 74",
     "image": "matter-dev-environment:local",
     "remoteUser": "vscode",
+    "containerEnv": {
+        "PW_ENVIRONMENT_ROOT": "/workspaces/connectedhomeip/.environment-vscode"
```
arkq, line 21: Add this to .gitignore, preferably as `/.environment*`

XToripuru: Done.

.devcontainer/Dockerfile
```diff
@@ -24,22 +24,17 @@ ARG USER_UID=1000
 ARG USER_GID=$USER_UID
 ENV LANG en_US.utf8
 
-# these are installed for terminal/dev convenience.  If more tooling for build is required, please
```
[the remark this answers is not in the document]
XToripuru: Done.

andy31415 (approved):

integrations/docker/images/vscode/chip-build-vscode/Dockerfile
```diff
@@ -144,6 +143,10 @@ ENV ZEPHYR_NXP_BASE=/opt/nxp-zephyr/zephyrproject/zephyr
 ENV ZEPHYR_NXP_SDK_INSTALL_DIR=/opt/nxp-zephyr/zephyr-sdk-0.16.5
 ENV NXP_UPDATE_SDK_SCRIPT_DOCKER=/opt/nxp/nxp_matter_support/scripts/update_nxp_sdk.py
 
+# Places bootstrap files there instead of the default one which is `.environment`.
+# NOTE: This directory is NOT persistent.
+ENV PW_ENVIRONMENT_ROOT=/home/vscode/pigweed/env
```
andy31415, line 148: how about keeping the previous placement and only adding the comment? That way I seem to be more ensured that docker will make these files identical and I can justify no version increase

andy31415: Overall it looks to me like env variables were alpha-sorted. The new place seems a bit off from that perspective.

arkq: > env variables were alpha-sorted

More or less, but in general these envs were not sorted. The case with `PW_ENVIRONMENT_ROOT` is that it does not affect the container itself, but its a sane default for a tool which is not baked into the image. The rest of the env variables if for SDKs baked into the image (`/opt/...`). There is also `PATH` variable which might get "lost" if all these envs were alpha-sorted, and I would propose to keep it separately as well.

And of course maybe we could sort all these envs now?

woody-apple (approved):

integrations/docker/images/vscode/chip-build-vscode/Dockerfile
XToripuru: These envs (in my opinion) should, and somewhat are, grouped by meaning before being sorted. `PW_ENVIRONMENT_ROOT` is not an SDK path so it might not be the best idea to mix it with SDK paths, especially considering it's importance. If a major restructure of this Dockerfile is wanted it probably should be done in other PR.

PR #35405, opened 2024-09-04
Address follow-up PR comments for fabric-admin UniqueIdGetter

### LINKED ISSUE #35382: FabricSync Address PR comments in follow up
There are a couple trivial comments that should be addressed in a follow up. They were not done in the initial PR as we needed PR to land for SVE FC deadline

- [ ] https://github.com/project-chip/connectedhomeip/pull/35377#discussion_r1742634456
- [ ] https://github.com/project-chip/connectedhomeip/pull/35377#discussion_r1742635288

tehampson:
This addresses PR comments that we allowed to be done in a follow-up for the sake of time

Fixes https://github.com/project-chip/connectedhomeip/issues/35382

### DIFF
--- a/examples/fabric-admin/device_manager/DeviceSynchronization.cpp
+++ b/examples/fabric-admin/device_manager/DeviceSynchronization.cpp
@@ -212,13 +212,13 @@ void DeviceSynchronizer::GetUniqueId()
     VerifyOrDie(mState == State::ReceivedResponse);
     VerifyOrDie(mController);
 
+    // If we have a UniqueId we can return leaving state in ReceivedResponse.
+    VerifyOrReturn(!mCurrentDeviceData.has_unique_id, ChipLogDetail(NotSpecified, "We already have UniqueId"));
+
     auto * device = DeviceMgr().FindDeviceByNode(mCurrentDeviceData.node_id);
-    if (mCurrentDeviceData.has_unique_id || !device)
-    {
-        // We already have a UniqueId, or this device is not assoicated with a remote Fabric Sync Administrator,
-        // so we can just return leaving the state in ReceivedResponse.
-        return;
-    }
+    // If there is no associated remote Fabric Sync Aggregator there is no other places for us to try
+    // getting the UniqueId fron and can return leaving the state in ReceivedResponse.
+    VerifyOrReturn(device, ChipLogDetail(NotSpecified, "No remote Fabric Sync Aggregator to get UniqueId from"));
 
     // Because device is not-null we expect IsFabricSyncReady to be true. IsFabricSyncReady indicates we have a
     // connection to the remote Fabric Sync Aggregator.
@@ -226,6 +226,7 @@ void DeviceSynchronizer::GetUniqueId()
     auto remoteBridgeNodeId               = DeviceMgr().GetRemoteBridgeNodeId();
     EndpointId remoteEndpointIdOfInterest = device->GetEndpointId();
 
+    ChipLogDetail(NotSpecified, "Attempting to get UniqueId from remote")
     CHIP_ERROR err = mUniqueIdGetter.GetUniqueId(
         [this](std::optional<CharSpan> aUniqueId) {
             if (aUniqueId.has_value())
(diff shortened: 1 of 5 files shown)


examples/fabric-admin/device_manager/DeviceSynchronization.cpp
```diff
@@ -211,14 +207,26 @@ void DeviceSynchronizer::StartDeviceSynchronization(chip::Controller::DeviceCont
...
-void DeviceSynchronizer::GetUid(EndpointId remoteEndpointIdOfInterest)
+void DeviceSynchronizer::GetUniqueId()
 {
     VerifyOrDie(mState == State::ReceivedResponse);
     VerifyOrDie(mController);
+
+    auto * device = DeviceMgr().FindDeviceByNode(mCurrentDeviceData.node_id);
+    if (mCurrentDeviceData.has_unique_id || !device)
+    {
+        // We already have a UniqueId, or this device is not assoicated with a remote Fabric Sync Administrator,
```
andy31415, line 218: for debug purposes, some logs may be helpful here, reporting if device already has a unique id or if device has no remote fabric sync admin.

andy31415 (approved):

examples/fabric-admin/device_manager/DeviceSynchronization.cpp
tehampson: Hmmm what do you think about only reporting if we have to fetch the UniqueID as the more common case is we don't need to do that

andy31415: My thinking was that I would like to be able to follow the code as I am reading logs. Log for teching unique id is fine, however I would also maybe want to differentiate between define _being nullptr vs device havin a unique_id already. If this can be done through other logs that is fine as well.

examples/fabric-admin/device_manager/DeviceSynchronization.cpp
```diff
@@ -211,32 +207,49 @@ void DeviceSynchronizer::StartDeviceSynchronization(chip::Controller::DeviceCont
...
+void DeviceSynchronizer::GetUniqueId()
 {
     VerifyOrDie(mState == State::ReceivedResponse);
     VerifyOrDie(mController);
+
+    // If we have a UniqueId we can return leaving state in ReceivedResponse.
+    VerifyOrReturn(!mCurrentDeviceData.has_unique_id, ChipLogDetail(NotSpecified, "We already have UniqueId"));
+
+    auto * device = DeviceMgr().FindDeviceByNode(mCurrentDeviceData.node_id);
+    // If there is no associated remote Fabric Sync Aggregator there is no other places for us to try
```
saurabhst, line 219:
```suggestion
    // If there is no associated remote Fabric Sync Aggregator there is no other place for us to try
```

examples/fabric-admin/device_manager/DeviceSynchronization.cpp
```diff
@@ -211,32 +207,49 @@ void DeviceSynchronizer::StartDeviceSynchronization(chip::Controller::DeviceCont
...
+    // getting the UniqueId fron and can return leaving the state in ReceivedResponse.
```
saurabhst, line 220:
```suggestion
    // getting the UniqueId from and can return leaving the state in ReceivedResponse.
```

saurabhst (approved):

examples/fabric-admin/device_manager/DeviceSynchronization.cpp
```diff
@@ -211,32 +207,49 @@ void DeviceSynchronizer::StartDeviceSynchronization(chip::Controller::DeviceCont
...
+        mUniqueIdGetter.GetUniqueId(
+            [this](std::optional<CharSpan> aUniqueId) {
+                if (aUniqueId.has_value())
+                {
+                    this->mCurrentDeviceData.has_unique_id = true;
+                    memcpy(this->mCurrentDeviceData.unique_id, aUniqueId.value().data(), aUniqueId.value().size());
+                }
+                else
+                {
+                    ChipLogError(NotSpecified, "We expected to get UniqueId from remote Fabric Sync Aggregator");
```
tehampson, line 239:
```suggestion
                    ChipLogError(NotSpecified, "We expected to get UniqueId from remote Fabric Sync Aggregator, but failed");
```

PR #35408, opened 2024-09-04
[Fuzzing] Adding all-clusters-app fuzz test to OSS-Fuzz

Alami-Amine:
This PR is identical to #34311  + a fix to the clang-tidy failures it is facing.

It involves: 
- building all-clusters-app fuzzer (based on libfuzzer).
- The goal is to have it running on OSS-Fuzz.


andy31415 (approved):

andy31415:
fast tracking this as a fixed copy of #34311 (as that PR seems to be a while without updates)

Issue #35165, opened 2024-08-22
Remaining M-ACL Issues


tleacmcsa:
These issues need to be addressed in a follow-up PR once #34932 merges:

- Return AccessRestricted when ARL denies an otherwise passing ACL check

- In TestAccessRestriction.cpp Cecille said:

for these, include #include <app-common/zap-generated/ids/Attributes.h> at the top and use
app::Clusters::NetworkCommissioning::Id etc. rather than hard coding.

Missing tests:
pase connection with a fabric index set (will happen after addnoc)
case connection before commissioning complete has been called.

It would be good to have ARLs that differ only by the fabric, to ensure that the correct fabric ARL is being applied.

- Boris pointed out lacking test coverage for AccessRestrictionProvider class

- Boris pointed out

"So the way this is currently implemented, the client can get the event before it gets the command response. This is fine, and unavoidable, but the spec should probably have some informative language making this clear to spec readers."

and "I don't see floors for these constants defined in the spec. Or ceilings. There should probably be such definitions." for CHIPConfig.h

tleacmcsa:
@bzbarsky-apple I chatted with @chrisdecenzo about one of your comments in the SDK PR that was moved to be tracked here around limits on ARL counts and their Restrictions.  Given that the ARLs are not writable and are set by the device vendor, we dont feel a limit makes sense.  Controllers would be the consumers and they shouldnt be size bound either.

bzbarsky-apple:
@tleacmcsa In general, we do in fact have caps on the size of things clients might need to read, so clients know how much memory to allocate for it.

Some clients are not memory-bound, some are....

tleacmcsa:
Is this something we should address in the spec and sdk, or just the sdk?

bzbarsky-apple:
@tleacmcsa Which exact comment are you talking about?

This is why followups should link to the actual comments involved...

tcarmelveilleux:
Also need to remove AccessRestrictionEntryChanged, and make sure ReviewFabricRestrictions has response command indicated, and update fields in FabricRestrictionReviewUpdate to be optional rather than nullable.

bzbarsky-apple:
@tleacmcsa If your "should address" question was about size caps on lists: that needed to be addressed in the spec.

PR #35411, opened 2024-09-04
Fix ACL XML for MACL and sample conformance

tcarmelveilleux:
Problem:

Issue #35165 

- AccessRestrictionEntryChanged event nees to go away since no longer in spec
- FabricRestrictionReviewUpdate event has 2 fields that are nullable but need to be optional
- ReviewFabricRestrictions is missing commands

This PR:

- Removes dead event AccessRestrictionEntryChanged
- Sets response command for ReviewFabricRestrictions
- Sets correct optionality for FabricRestrictionReviewUpdate
- Sets MGND feature in matter-network-manager-app
- Remove ResetCounts support command from EthernetNetworkDiagnostics (non conformant)

Testing done:

- IDM 10.2 and 10.4 pass against matter-network-manager-app with local updated spec XML


tleacmcsa (approved):

chrisdecenzo (approved):

cecille (approved):

src/app/zap-templates/zcl/data-model/chip/access-control-cluster.xml
```diff
@@ -156,16 +156,11 @@ limitations under the License.
...
-    </event>
-
-    <event code="0x0003" name="FabricRestrictionReviewUpdate" priority="info" side="server" isFabricSensitive="true" optional="true">
+    <event side="server" code="0x0002" name="FabricRestrictionReviewUpdate" priority="info" isFabricSensitive="true" optional="true">
       <description>The cluster SHALL send FabricRestrictionReviewUpdate events to indicate completion of a fabric restriction review.</description>
       <field id="0" name="Token" type="int64u"/>
-      <field id="1" name="Instruction" type="long_char_string" isNullable="true" length="512"/>
-      <field id="2" name="RedirectURL" type="long_char_string" isNullable="true" length="256"/>
+      <field id="1" name="Instruction" type="long_char_string" optional="true" length="512"/>
+      <field id="2" name="ARLRequestFlowUrl" type="long_char_string" optional="true" length="256"/>
```
bzbarsky-apple, line 163: This matches the spec as it stands today, but why did the spec use lower-case "rl" here?  Most URLs in the spec use "URL" with all-caps in the name...

I feel like we should align with normal spec usage and have ARLRequestFlowURL both here and in the spec.

bzbarsky-apple (approved):

src/app/zap-templates/zcl/data-model/chip/access-control-cluster.xml
tcarmelveilleux: Created #35418

PR #35413, opened 2024-09-04
MACL: Add checks that ARL / CommissioningARL are populated or not populated as required by the test

cecille:
Proposal for checking ARLs per the test plan precondition here: https://github.com/CHIP-Specifications/chip-test-plans/pull/4632

### DIFF
--- a/src/python_testing/basic_composition_support.py
+++ b/src/python_testing/basic_composition_support.py
@@ -26,10 +26,28 @@ import typing
 from pprint import pformat, pprint
 from typing import Any, Optional
 
+import chip.clusters as Clusters
 import chip.clusters.ClusterObjects
 import chip.tlv
 from chip.clusters.Attribute import ValueDecodeFailure
 from mobly import asserts
+
+
+def arls_populated(tlv_data: dict[int, Any]) -> tuple[bool, bool]:
+    """ Returns a tuple indicating if the ARL and CommissioningARL are populated.
+        Requires a wildcard read of the device TLV.
+    """
+    # ACL is always on endpoint 0
+    if 0 not in tlv_data or Clusters.AccessControl.id not in tlv_data[0]:
+        return (False, False)
+    # Both attributes are mandatory for this feature, so if one doesn't exist, neither should the other.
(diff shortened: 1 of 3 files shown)


src/python_testing/basic_composition_support.py
```diff
@@ -167,6 +185,12 @@ async def setup_class_helper(self, allow_pase: bool = True):
         logging.info("Start of actual tests")
         logging.info("###########################################################")
 
+        have_arl, have_carl = arls_populated(self.endpoints_tlv)
```
chrisdecenzo, line 188: how many tests use this method? Wondering if this will cause test cases to fail when the ARL attribute is cleared.

chrisdecenzo: We need to make sure that every test script can pass when the ARL attributes are present but empty.

cecille: This is used by TC_DeviceConformance, TC_DeviceBasicComposition and TC_AccessControl, so a lot. The function returns false when the attributes are missing or empty. See TestConformanceTest.py for a unit test showing the MACL feature on, but the lists unpopulated.

chrisdecenzo (approved):

cjandhyala (approved):

src/python_testing/basic_composition_support.py
```diff
@@ -26,12 +26,30 @@
...
 from typing import Any, Optional
 
+import chip.clusters as Clusters
 import chip.clusters.ClusterObjects
 import chip.tlv
 from chip.clusters.Attribute import ValueDecodeFailure
 from mobly import asserts
 
 
+def arls_populated(tlv_data: dict[int, Any]) -> tuple[bool, bool]:
```
andy31415, line 36: returning a dataclass with named members has_arl/has_carl (or arl/carl) would be easier to follow than two bools.

tcarmelveilleux (approved):

src/python_testing/basic_composition_support.py
cecille: I have a followup ready, but given that it's now 9:30 and we're well past the SVE deadline, and the CI takes multiple hours, I'm going to resolve this and address in a follow up PR. PR will be up in a few minutes.

PR #35416, opened 2024-09-04
[Fabric-Sync] Improve the stability of test TC_MCORE_FS_1_4

yufengwangca:
I've noticed that MCORE_FS_1_4 is still quite flaky, though the other test cases seem stable.

It turn out there is race issue and the TH paired to FS_BRIDGE instead test SERVER.

### DIFF
--- a/src/python_testing/TC_MCORE_FS_1_4.py
+++ b/src/python_testing/TC_MCORE_FS_1_4.py
@@ -48,6 +48,31 @@ from matter_testing_support import MatterBaseTest, TestStep, async_test_body, de
 from mobly import asserts
 
 
+async def wait_for_server_initialization(server_port, timeout=5):
+    """Wait until the server is ready by checking if it opens the expected port."""
+    start_time = asyncio.get_event_loop().time()
+    elapsed_time = 0
+    retry_interval = 1
+
+    logging.info(f"Waiting for server to initialize on TCP port {server_port} for up to {timeout} seconds.")
+
+    while elapsed_time < timeout:
+        try:
+            # Try connecting to the server to check if it's ready
+            reader, writer = await asyncio.open_connection('::1', server_port)
+            writer.close()
+            await writer.wait_closed()
+            logging.info(f"TH_SERVER_NO_UID is initialized and ready on port {server_port}.")
+            return
+        except (ConnectionRefusedError, OSError) as e:
+            logging.warning(f"Connection to port {server_port} failed: {e}. Retrying in {retry_interval} seconds...")
+
+        await asyncio.sleep(retry_interval)
+        elapsed_time = asyncio.get_event_loop().time() - start_time
+
+    raise TimeoutError(f"Server on port {server_port} did not initialize within {timeout} seconds. "
+                       f"Total time waited: {elapsed_time} seconds.")
+
 # TODO: Make this class more generic. Issue #35348
 class Subprocess(threading.Thread):
 
@@ -240,7 +265,7 @@ class TC_MCORE_FS_1_4(MatterBaseTest):
 
         # Wait for TH_SERVER_NO_UID get initialized.
         try:
-            asyncio.run(self.wait_for_server_initialization(self.th_server_port))
+            asyncio.run(wait_for_server_initialization(self.th_server_port))
         except TimeoutError:
             asserts.fail(f"TH_SERVER_NO_UID server failed to open port {self.th_server_port}")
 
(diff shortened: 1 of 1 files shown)


src/python_testing/TC_MCORE_FS_1_4.py
```diff
@@ -271,6 +296,10 @@ async def test_TC_MCORE_FS_1_4(self):
         self.step(1)
 
         th_server_th_node_id = 1
+
+        # Wait for TH_SERVER_NO_UID get initialized.
+        await self.wait_for_server_initialization(self.th_server_port)
```
tehampson, line 301: Would the end of `setup_class` be better after we initialized the application

tehampson: Wherever this is called, it might also be good to have a try block where the exception for TimeoutError calls `asserts.fail(f"TH_SERVER_NO_UID server failed to open port {self.th_server_port}")`

yufengwangca: But setup_class can not run await, we need to use asyncio to run the asynchronous server initialization, something like:
    
    asyncio.run(self.wait_for_server_initialization(self.th_server_port))

src/python_testing/TC_MCORE_FS_1_4.py
```diff
@@ -247,6 +253,31 @@ def teardown_class(self):
             self.storage.cleanup()
         super().teardown_class()
 
+    async def wait_for_server_initialization(self, server_port, timeout=5):
```
tehampson, line 256: nit, I don't see this using and `self....` should we just make it a stand alone method outside of the `class TC_BRBINFO_4_1`?

tehampson (approved):

andy31415 (approved):

saurabhst (approved):

src/python_testing/TC_MCORE_FS_1_4.py
```diff
@@ -48,7 +48,34 @@
 from mobly import asserts
 
 
+async def wait_for_server_initialization(server_port, timeout=5):
+    """Wait until the server is ready by checking if it opens the expected port."""
+    start_time = asyncio.get_event_loop().time()
+    elapsed_time = 0
+    retry_interval = 1
+
+    logging.info(f"Waiting for server to initialize on TCP port {server_port} for up to {timeout} seconds.")
```
bzbarsky-apple, line 57: What guarantees that TCP is even supported by that server?

yufengwangca: We use dedicated test server "lighting-app-data-mode-no-unique-id"

bzbarsky-apple: Yes, but where is it documented/guaranteed that it supports TCP?  That is, what keeps this from randomly breaking?

yufengwangca: The TCP is enabled by default. https://github.com/project-chip/connectedhomeip/blob/master/src/inet/inet.gni#L27
As long as we don't explicitly override this value in our example app, then TCP is enabled, Non-Linux platforms disable it currently, but they do it explicitly.

bzbarsky-apple: @yufengwangca Yes, but:

1. TH gets run on non-linux platforms too.
2. Absolutely nothing ensures that Linux will keep enabling TCP by default.

yufengwangca: In this case, we're checking the dedicated example app "lighting-app-data-mode-no-unique-id," which only has a Linux implementation.

I can switch to checking the specific log output from stdout instead, but I’m not convinced this is a better approach. It would be even more fragile if the log format changes.

Pradip mentioned that TCP is enabled by default, and we need to explicitly override this flag. Therefore, we can add a comment in the example app to avoid overriding it, as the example app is completely under our control.

bzbarsky-apple: > In this case, we're checking the dedicated example app "lighting-app-data-mode-no-unique-id," which only has a Linux implementation.

The "linux" bits are built on both Linux and Mac.  They should really be called "posix" or something, not "linux", but here we are.

yufengwangca: https://github.com/project-chip/connectedhomeip/pull/35547

PR #35420, opened 2024-09-04
Create a script and instructions to create an aarch64 sysroot via qemu

andy31415:
Using an actual raspberry pi is slightly harder to automate, so here is a script and instructions to create a sysroot on a linux machine (tested on a amd64 linux).

### Changes

- script using qemu/virtinstall to create a VM for aarch64 ubuntu
- script for cipd package creation and upload

### Tested

- created https://chrome-infra-packages.appspot.com/p/experimental/matter/sysroot/ubuntu-24.04-aarch64/+/version:build-2024.09.05 today
- compiled with:

```sh
echo 'experimental/matter/sysroot/ubuntu-24.04-aarch64 version:build-2024.09.05' > ensure_file.txt
cipd ensure -ensure-file ensure_file.txt -root ./
SYSROOT_AARCH64=$(pwd)/ubuntu-24.04-aarch64-sysroot ./scripts/build/build_examples.py --target linux-arm64-chip-tool-clang build
```


integrations/docker/images/stage-1/chip-build-crosscompile/start-sysroot-vm.sh
```diff
@@ -0,0 +1,211 @@
+#!/usr/bin/env bash
+
+# This script is intended to build a SYSROOT that can be used as a base to cross-compile
+# CHIP for aarch64 (i.e. that runs on a RaspberryPi) from a arm64 linux machine
```
kiel-apple, line 4: is this meant to say amd64?
```suggestion
# CHIP for aarch64 (i.e. that runs on a RaspberryPi) from a amd64 linux machine
```

integrations/docker/images/stage-1/chip-build-crosscompile/start-sysroot-vm.sh
```diff
@@ -0,0 +1,211 @@
...
+fi
+
+rm -f sysrootsrv.img
+qemu-img create -b "$CLOUD_IMAGE_FILE" -f qcow2 -F qcow2 sysrootsrv.img 10G
+
+# local user allowed to SSH in to the host directly without a password
+SSH_KEY=$(cat ~/.ssh/id_rsa.pub)
+
+# set a login password
+PASSWORD_HASH=$(mkpasswd 1234)
```
kiel-apple, line 84: i'm guessing the security story here is "ubuntu running in qemu only allows connections from `localhost`"?  is it worth noting that strategy here so folks have some peace of mind when they see a weak password here?

andy31415: We use `user` networking, so you cannot even connect to it from anywhere except through either the virtual console (which is what I added here) or through the `virsh console` which is local console only. The vm IP is not routable even from the host machine amazingly enough.

I will add a comment, however realistically this is a password so I can login from the console to debug things.

andy31415: Added text explaining password usage.

kiel-apple (approved):

tcarmelveilleux (approved):

pidarped (approved):

PR #35421, opened 2024-09-04
Fix intermittent falure in Darwin browsing test.

bzbarsky-apple:
We are running the test against Matter applications that are starting up, and during startup we do some silly unregistering/re-registering of DNS-SD advertisements.  The test was counting all the registrations, not subtracting the un-registrations, and asserting that the resulting count was not too high. But sometimes it is, if we get a "unregister/re-register" pair in there.

The fix:

1) Fixes the test to better track un-registration bits. 2) Fixes the OnBleScanRemove code to call the right delegate method. 3) Fixes OnBrowseRemove to not call the delegate if we have not notified about
   the add for the relevant result.  We could have just worked around this in
   the test, but notifying "removed" about an object with all fields set to nil
   just doesn't make much sense.

### DIFF
--- a/src/darwin/Framework/CHIPTests/MTRCommissionableBrowserTests.m
+++ b/src/darwin/Framework/CHIPTests/MTRCommissionableBrowserTests.m
@@ -36,9 +36,27 @@ static bool sHelperAppsStarted = false;
 // Singleton controller we use.
 static MTRDeviceController * sController = nil;
 
+static NSString * kInstanceNameKey = @"instanceName";
+static NSString * kVendorIDKey = @"vendorID";
+static NSString * kProductIDKey = @"productID";
+static NSString * kDiscriminatorKey = @"discriminator";
+static NSString * kCommissioningModeKey = @"commissioningMode";
+
+static NSDictionary<NSString *, id> * ResultSnapshot(MTRCommissionableBrowserResult * result)
+{
+    return @{
+        kInstanceNameKey : result.instanceName,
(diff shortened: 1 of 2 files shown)


src/darwin/Framework/CHIPTests/MTRCommissionableBrowserTests.m
```diff
@@ -36,9 +36,27 @@
...
 
+static NSString * kInstanceNameKey = @"instanceName";
+static NSString * kVendorIDKey = @"vendorID";
+static NSString * kProductIDKey = @"productID";
+static NSString * kDiscriminatorKey = @"discriminator";
+static NSString * kCommissioningModeKey = @"commissioningMode";
+
+static NSDictionary<NSString *, id> * ResultSnapshot(MTRCommissionableBrowserResult * result)
+{
+    return @{
```
ksperling-apple, line 47: Could use `dictionaryWithValuesForKeys:...` but this is fine too.

ksperling-apple (approved):

src/darwin/Framework/CHIPTests/MTRCommissionableBrowserTests.m
bzbarsky-apple: I'll stick with this, so I don't have to worry about what dictionaryWithValuesForKeys does with the unboxed boolean...

bzbarsky-apple:

kiel-apple (approved):

andy31415 (approved):

PR #35422, opened 2024-09-05
Darwin: Add delegate methods to populate initial state over new XPC

kiel-apple:
TODO:

- [x] populate NSDictionary with relevant current internal state
- [x] use description lock to find places where relevant ivars are updated so we can trigger the delegate method

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice_Concrete.h
+++ b/src/darwin/Framework/CHIP/MTRDevice_Concrete.h
@@ -24,8 +24,4 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
-// Private Internal State keys
-extern NSString * const kMTRDeviceInternalPropertyKeyVendorID;
-extern NSString * const kMTRDeviceInternalPropertyKeyProductID;
-
 NS_ASSUME_NONNULL_END
--- a/src/darwin/Framework/CHIP/MTRDevice_Concrete.h
+++ b/src/darwin/Framework/CHIP/MTRDevice_Concrete.h
@@ -24,4 +24,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
+// Private Internal State keys
+extern NSString * const kMTRDeviceInternalPropertyKeyVendorID;
+extern NSString * const kMTRDeviceInternalPropertyKeyProductID;
+
 NS_ASSUME_NONNULL_END
--- a/src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
@@ -488,7 +488,7 @@ typedef NS_ENUM(NSUInteger, MTRDeviceWorkItemDuplicateTypeID) {
     };
 }
 
-- (oneway void)notifyPrivateInternalPropertiesDelegateOfChanges
+- (void)_notifyPrivateInternalPropertiesDelegateOfChanges
 {
     if ([self.privateInternalStateDelegate respondsToSelector:@selector(device:internalStateUpdated:)]) {
         [self.privateInternalStateDelegate device:self.nodeID internalStateUpdated:[self _internalProperties]];
@@ -983,7 +983,7 @@ typedef NS_ENUM(NSUInteger, MTRDeviceWorkItemDuplicateTypeID) {
         }];
         /* END DRAGONS */
 
-        [self notifyPrivateInternalPropertiesDelegateOfChanges];
+        [self _notifyPrivateInternalPropertiesDelegateOfChanges];
     }
 }
 
@@ -1184,7 +1184,7 @@ typedef NS_ENUM(NSUInteger, MTRDeviceWorkItemDuplicateTypeID) {
         std::lock_guard lock(_descriptionLock);
         _lastSubscriptionFailureTimeForDescription = _lastSubscriptionFailureTime;
     }
-    [self notifyPrivateInternalPropertiesDelegateOfChanges];
+    [self _notifyPrivateInternalPropertiesDelegateOfChanges];
     deviceUsesThread = [self _deviceUsesThread];
 
     // If a previous resubscription failed, remove the item from the subscription pool.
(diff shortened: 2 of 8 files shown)


src/darwin/Framework/CHIP/XPC Protocol/MTRXPCClientProtocol.h
```diff
@@ -27,6 +27,10 @@ MTR_NEWLY_AVAILABLE
 - (oneway void)deviceBecameActive:(NSNumber *)nodeID;
 - (oneway void)deviceCachePrimed:(NSNumber *)nodeID;
 - (oneway void)deviceConfigurationChanged:(NSNumber *)nodeID;
+
+@optional
+// temporarily optional to avoid lockstep needs
+- (oneway void)deviceInternalStateUpdated:(NSDictionary *)dictionary;
```
woody-apple, line 33:
```suggestion
- (oneway void)device:(NSNumber *)nodeID internalStateUpdated:(NSDictionary *)dictionary;
```

src/darwin/Framework/CHIP/XPC Protocol/MTRXPCClientProtocol.h
```diff
@@ -27,6 +27,10 @@ MTR_NEWLY_AVAILABLE
 - (oneway void)deviceBecameActive:(NSNumber *)nodeID;
 - (oneway void)deviceCachePrimed:(NSNumber *)nodeID;
 - (oneway void)deviceConfigurationChanged:(NSNumber *)nodeID;
+
+@optional
```
woody-apple, line 31: No need for optional, follow on PRs will be fine

src/darwin/Framework/CHIP/XPC Protocol/MTRXPCServerProtocol.h
```diff
@@ -71,6 +71,11 @@ MTR_NEWLY_AVAILABLE
 
 - (oneway void)deviceController:(NSUUID *)controller shutdownDeviceController:(NSUUID *)controller;
 
+@optional
```
woody-apple, line 74: No need for optional, follow on PRs will be fine

woody-apple: NVM, good to be save

src/darwin/Framework/CHIP/XPC Protocol/MTRXPCClientProtocol.h
woody-apple: NVM, good to be safe

src/darwin/Framework/CHIP/XPC Protocol/MTRXPCClientProtocol.h
kiel-apple: Will change to this - something I realized later is that there's that actual individual ivar/property called `internalState`, so thought naming this `internalState` might be briefly confusing.  But with the types in the signature, I think it would draw people to look more closely and quickly understand.

src/darwin/Framework/CHIP/MTRDevice_Concrete.h
```diff
@@ -24,4 +24,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 @end
 
+// Private Internal State keys
+extern NSString * const kMTRDeviceInternalPropertyKeyVendorID;
```
woody-apple, line 28: don't extern

src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
```diff
@@ -461,7 +463,41 @@ - (NSString *)description
...
+
+- (NSDictionary *)_internalProperties {
+    id vidOrUnknown, pidOrUnknown;
+
+    {
+        std::lock_guard lock(_descriptionLock);
+
+        vidOrUnknown = _vid ?: @"Unknown";
+        pidOrUnknown = _pid ?: @"Unknown";
+        //    id networkFeatures = _allNetworkFeatures;
```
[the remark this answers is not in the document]
woody-apple: Seems fine, but ideally we'd more proactively make this when things change... but I think this is fine for now

src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
```diff
@@ -461,7 +463,41 @@ - (NSString *)description
...
+        //    id lastSubscriptionFailureTime = _lastSubscriptionFailureTimeForDescription;
+    }
+
+    return @{
+        kMTRDeviceInternalPropertyKeyVendorID : vidOrUnknown,
+        kMTRDeviceInternalPropertyKeyProductID : pidOrUnknown,
+    };
+}
+
+- (oneway void)notifyPrivateInternalPropertiesDelegateOfChanges
```
woody-apple, line 496:
```suggestion
- (void)_notifyPrivateInternalPropertiesDelegateOfChanges
```

src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
```diff
@@ -461,7 +463,41 @@ - (NSString *)description
...
+        //    id internalDeviceState = _internalDeviceStateForDescription;
+        //    id lastSubscriptionAttemptWait = _lastSubscriptionAttemptWaitForDescription;
+        //    id mostRecentReportTime = _mostRecentReportTimeForDescription;
+        //    id lastSubscriptionFailureTime = _lastSubscriptionFailureTimeForDescription;
+    }
+
+    return @{
+        kMTRDeviceInternalPropertyKeyVendorID : vidOrUnknown,
+        kMTRDeviceInternalPropertyKeyProductID : pidOrUnknown,
+    };
```
woody-apple, line 493: I can't recall what happens if vidOrUnknown is nil? I guess it doesn't matter here, but a macro that does something like:

#define OPTIONAL_ATTRIBUTE(key,value,dictionary)
if ( value ) [dictionary setObject: value for Key: key];

Tends to be something useful in general :)

kiel-apple: vidOrUnknown can only be a value - on line 481, if _vid is nil, we assign the string @"Unknown".  I copied this pattern from elsewhere in the code - I'm not sure if it's the best idea ever but it does solve the "don't try to put nil in a dictionary" problem

woody-apple: I understand, but the above is a better pattern, it's why most projects have macros like this :)

woody-apple: That said, let's get this in.

PR #35426, opened 2024-09-05
[Silabs] Enabling Uart logs for 917soc by default

chirag-silabs:
Enabling the UART logs by default for the 917 SoC board


andy31415 (approved):

scripts/examples/gn_silabs_example.sh
```diff
@@ -300,7 +300,7 @@ else
     if [ "$SILABS_BOARD" == "BRD4338A" ]; then
         echo "Compiling for 917 WiFi SOC"
         USE_WIFI=true
-        optArgs+="chip_device_platform =\"SiWx917\" is_debug=false "
+        optArgs+="chip_device_platform =\"SiWx917\" is_debug=false sl_uart_log_output=true "
```
jmartinez-silabs, line 303: `sl_uart_log_output` is declared in silabs_board.gni. Set it to true directly in that file so it doesn't have to depend on this build script...

chirag-silabs: updated the silabs_board.gni

third_party/silabs/silabs_board.gni
```diff
@@ -69,6 +69,7 @@ if (silabs_board == "BRD4338A") {
   silabs_family = "SiWx917-common"
   silabs_mcu = "SiWG917M111MGTBA"
   wifi_soc = true
+  sl_uart_log_output = true
```
jmartinez-silabs, line 72: This will make it so `sl_uart_log_output` cannot be overwritten to false by command line arg for 917. Is that what you want? If you want it to be configurable but default to true for 917 you have to do like what was done for 
`use_rps_extension` lower in this same file

chirag-silabs: updated

jmartinez-silabs (approved):

third_party/silabs/silabs_board.gni
```diff
@@ -165,6 +162,9 @@ declare_args() {
   # User can set this arg to false to skip the rps creation for WiFi SoCs.
   # e.g. for CI
   use_rps_extension = wifi_soc
+
+  # UART log forwarding enabled by default for 917 SoC
+  sl_uart_log_output = wifi_soc
```
tehampson, line 167: From an outsider's perspective what does `wifi_soc` have to do with default for 917 SoC? should this not just be set to true based on the comment? Could you not set it to false near the top in a declare_args, but set it to true for the 917?

chirag-silabs: updated the comment, in the context wifi_soc is nothing but the 917 soc both are same. Updated the comment to say
`# Default true for the wifi soc and false for EFR32 boards`

Issue #34352, opened 2024-07-16
[Feature] 


Alami-Amine:
### Feature description

feature

### Platform

all

PR #35427, opened 2024-09-05
converting all-clusters-fuzzer to fuzztest

Alami-Amine:
relates to #34352


examples/all-clusters-app/linux/fuzzing-main-pw.cpp
```diff
@@ -0,0 +1,99 @@
...
+void CleanShutdown()
+{
+    Server::GetInstance().Shutdown();
+    PlatformMgr().Shutdown();
+    // TODO: We don't Platform::MemoryShutdown because ~CASESessionManager calls
+    // Dnssd::ResolverProxy::Shutdown, which starts doing Platform::Delete.
+    // Platform::MemoryShutdown();
+}
+
+void AllClustersFuzzer(const std::vector<std::uint8_t> & Bytes)
```
bzbarsky-apple, line 48: I wonder whether it would be better to have some shared "take a bytespan and do stuff" thing that this and the existing fuzzer can share instead of just copy/pasting everything....

PR #35428, opened 2024-09-05
[Linux] Fix potential race when saving storage file

arkq:
### Problem

The `mkstemp()` call creates a unique file and returns file descriptor for opened file which should be used to write data. However, Linux implementation, instead of using file descriptor directly, it opens the temp file again by its name. This could lead to race, because someone might remove the file between these calls.

### Changes

- Create helper classes for file descriptor RAII closing, and temporary file stream class compatible with std::ostream.
- Properly sync file content before rename.
- Synchronize NuttX and WebOS implementation with Linux

### Testing

Tested locally, but CI will verify as well.


src/lib/support/TemporaryFileStream.h
```diff
@@ -0,0 +1,107 @@
...
+    TemporaryFileStream(const TemporaryFileStream &)             = delete;
+    TemporaryFileStream & operator=(const TemporaryFileStream &) = delete;
+
+    /// Open a temporary file with a given name template.
+    ///
+    /// In order to check if the file was opened successfully, use IsOpen().
+    void Open(std::string nameTemplate)
+    {
+        mFileName = std::move(nameTemplate);
+        mFd       = FileDescriptor(mkstemp(&mFileName[0]));
```
andy31415, line 82:
```suggestion
        mFd       = FileDescriptor(mkstemp(mFileName.c_str()));
```

andy31415 (approved):

src/lib/support/TemporaryFileStream.h
arkq: We need non-const pointer here. Anyway, since c++17 we can use `.data()`.

yunhanw-google (approved):

saurabhst (approved):

PR #35438, opened 2024-09-05
[Darwin] Fix for MTRDevice deadlock during subscription reset

jtung-apple:
In this code path, an internal lock of MTRDevice is not unlocked before returning, causing future deadlocks.

This bug can happen if subscription establishment happens right when subscription reset is happening.


kiel-apple (approved):

bzbarsky-apple (approved):

bzbarsky-apple:

PR #35439, opened 2024-09-05
[Silabs] Update to WiseConnect 3.3.2

rerasool:
**Change**
Updated `wifi_sdk` submodule from WiseConnect v3.3.1 to v3.3.2.

**Testing**
This WiseConnect update contains documentation changes only. To be extra safe, validated that `lighting-app` and `lock-app` compile and run successfully (QR code shows up) on BRD4338A (which uses WiseConnect codebase).

### DIFF
--- a/.github/workflows/release_artifacts.yaml
+++ b/.github/workflows/release_artifacts.yaml
@@ -64,7 +64,7 @@ jobs:
         runs-on: ubuntu-latest
 
         container:
-            image: ghcr.io/project-chip/chip-build-efr32:77
+            image: ghcr.io/project-chip/chip-build-efr32:78
         steps:
             - name: Checkout
               uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


chirag-silabs:
docker update isn't needed to merge v3.3.2 wifi sdk?

jmartinez-silabs (approved):

.github/workflows/release_artifacts.yaml
```diff
@@ -64,7 +64,7 @@ jobs:
         runs-on: ubuntu-latest
 
         container:
-            image: ghcr.io/project-chip/chip-build-efr32:77
```
andy31415, line 67: will this be required or nice to have ? Asking because cloud-build also builds efr32 so it may need updating.

andy31415: Specifically I think you should be updating all the versions of the vscode image in https://github.com/project-chip/connectedhomeip/blob/master/integrations/cloudbuild/smoke-test.yaml

rerasool: updated!

andy31415 (approved):

PR #35444, opened 2024-09-06
encourage the use of CopyCharSpanToMutableCharSpanWithTruncation

hicklin:
- Updates the service area code documentation to encourage the use of `CopyCharSpanToMutableCharSpanWithTruncation` in methods that do not provide a way propagate errors.
- Updates the rvc-app example to use `CopyCharSpanToMutableCharSpanWithTruncation`.


bzbarsky-apple (approved):
Approved, but note that truncation can produce interesting garbage at the truncation point in languages other than English where the UTF-8 encoding of characters is multi-byte....

andy31415 (approved):

PR #35448, opened 2024-09-06
[NXP] Adding pre/post-build process support for NXP RW61x platform

dinabenamar:
The aim of this Pull Request is to introduce the pre-build and post-build steps to automatize the generation of required pre/post-build files, and reduce the manual efforts for testing our applications.

Key changes : 
- Added CMake files for NXP RW61x examples, where cmake is utilized for executing the pre- and post- build tasks, and GN continues to handle the actual build of the applications.
- Added the pre- and post-build steps in the common "config/nxp/chip-cmake-freertos/CMakeLists.txt"
- Updated the NXP builder to utilize the build_examples framework for building RW61x applications with cmake (while GN build is also still supported)

### DIFF
--- a/scripts/build/builders/nxp.py
+++ b/scripts/build/builders/nxp.py
@@ -262,7 +262,7 @@ class NxpBuilder(GnBuilder):
                 flags.append('-DFILE_SUFFIX=fdata')
             else:
                 flags.append("-DCONFIG_CHIP_FACTORY_DATA=true")
-        
+
         if self.enable_ethernet:
             if self.os_env == NxpOsUsed.ZEPHYR:
                 flags.append('-DEXTRA_CONF_FILE="prj_ethernet.conf"')
--- a/scripts/build/builders/nxp.py
+++ b/scripts/build/builders/nxp.py
@@ -304,7 +304,7 @@ class NxpBuilder(GnBuilder):
             flag_board_variant = "-DCONFIG_BOARD_VARIANT=\"%s\"" % self.board_variant.BoardVariantName(self.board)
             flags.append(flag_board_variant)
 
-        build_flags = " ".join(flags) if len(flags) > 0 else ""
+        build_flags = " ".join(flags)
 
         return build_flags
 
--- a/scripts/build/builders/nxp.py
+++ b/scripts/build/builders/nxp.py
@@ -140,11 +140,12 @@ class NxpBuilder(GnBuilder):
                  enable_thread: bool = False,
                  enable_wifi: bool = False,
                  enable_ethernet: bool = False,
-                 disable_ipv4: bool = False,
                  enable_shell: bool = False,
                  enable_ota: bool = False,
                  data_model_interface: Optional[str] = None,
-                 ):
+                 enable_factory_data_build: bool = False,
+                 disable_pairing_autostart: bool = False,
+                 board_variant: str = None):
         super(NxpBuilder, self).__init__(
             root=app.BuildRoot(root, board, os_env),
             runner=runner)
@@ -160,7 +161,6 @@ class NxpBuilder(GnBuilder):
         self.enable_lit = enable_lit
         self.enable_rotating_id = enable_rotating_id
         self.has_sw_version_2 = has_sw_version_2
-        self.disable_ipv4 = disable_ipv4
         self.disable_ble = disable_ble
         self.enable_thread = enable_thread
         self.enable_wifi = enable_wifi
@@ -168,6 +168,9 @@ class NxpBuilder(GnBuilder):
         self.enable_ota = enable_ota
         self.enable_shell = enable_shell
         self.data_model_interface = data_model_interface
+        self.enable_factory_data_build = enable_factory_data_build
+        self.disable_pairing_autostart = disable_pairing_autostart
+        self.board_variant = board_variant
 
     def GnBuildArgs(self):
         args = []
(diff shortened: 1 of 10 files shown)


scripts/build/builders/nxp.py
```diff
@@ -124,16 +136,19 @@ def __init__(self,
                  disable_ble: bool = False,
                  enable_thread: bool = False,
                  enable_wifi: bool = False,
-                 disable_ipv4: bool = False,
                  enable_shell: bool = False,
-                 enable_ota: bool = False):
+                 enable_ota: bool = False,
+                 enable_factory_data_build: bool = False,
+                 disable_pairing_autostart: bool = False,
+                 board_variant = None):
```
andy31415, line 143: what is the type of board_variant? Seems `Optional[str]`? However str seems a bit too loose ... could we make it an enum?

dinabenamar: Hello @andy31415 , that's correct here the `board_variant` type is "str", I updated the nxp.py file to specify the type. I have considered converting the `board_variant` into `enum` type, but currently keeping it as string seems to be more flexible. This approach allows to pass it easily to the Cmake option / GN arg which are expecting a string value. But I am open to suggestions, if there is a better way to manage the "board_variant" string while keeping compatibility with GN / cmake args :)

andy31415: Could you explain about the flexible part? Why should we not set it as a fixed enum? If we need to support more boards we need code changes in targets.py anyway, so it seems string just makes it vague and free form.

scripts/build/builders/nxp.py
```diff
@@ -129,16 +143,19 @@ def __init__(self,
                  disable_ble: bool = False,
                  enable_thread: bool = False,
                  enable_wifi: bool = False,
-                 disable_ipv4: bool = False,
                  enable_shell: bool = False,
-                 enable_ota: bool = False):
+                 enable_ota: bool = False,
+                 enable_factory_data_build: bool = False,
+                 disable_pairing_autostart: bool = False,
+                 board_variant: str = None):
```
andy31415, line 150: this is not str given that its default is None. It should be at least `Optional[str]`. also see my previous notes about this being an enum.

Could you also please make patches incremental instead of force-push? Force-push seems to cause loss of history/placement of comments.

scripts/build/testdata/all_targets_linux_x64.txt
```diff
@@ -9,11 +9,11 @@ efr32-{brd2704b,brd4316a,brd4317a,brd4318a,brd4319a,brd4186a,brd4187a,brd2601b,b
...
+linux-{x64,arm64}-{rpc-console,all-clusters,all-clusters-minimal,chip-tool,thermostat,java-matter-controller,kotlin-matter-controller,minmdns,light,light-data-model-no-unique-id,lock,shell,ota-provider,ota-requestor,simulated-app1,simulated-app2,python-bindings,tv-app,tv-casting-app,bridge,fabric-admin,fabric-bridge,tests,chip-cert,address-resolve-tool,contact-sensor,dishwasher,microwave-oven,refrigerator,rvc,air-purifier,lit-icd,air-quality-sensor,network-manager,energy-management}[-nodeps][-nlfaultinject][-platform-mdns][-minmdns-verbose][-libnl][-same-event-loop][-no-interactive][-ipv6only][-no-ble][-no-wifi][-no-thread][-mbedtls][-boringssl][-asan][-tsan][-ubsan][-libfuzzer][-ossfuzz][-pw-fuzztest][-coverage][-dmalloc][-clang][-test][-rpc][-with-ui][-evse-test-event][-enable-dnssd-tests][-disable-dnssd-tests][-chip-casting-simplified][-data-model-check][-data-model-disabled][-data-model-enabled]
```
andy31415, line 12: why is check-failure-die gone here? is this a failed merge? Or will CI fail?

dinabenamar: I just saw this now, it must have been a wrong conflict resolution on my side when I rebased the branch. Thank you for pointing this out, I will fix it.

dinabenamar: Should be fixed now, thanks.

scripts/build/builders/nxp.py
dinabenamar: About the flexible part, I was making reference to the fact that we are passing the string value of "board_variant" to a cmake option here : https://github.com/project-chip/connectedhomeip/pull/35448/commits/c1ddce018b9bd6255553b1535aa048dd73722c86#diff-16cf3d9dbea121fce0852e138b4ba2948b5f38c9d9b74699170a4cf75cc6551cR249 , because this information is needed for the build.
But I got your point, I have made a small rework to turn it into enum, please check my last commit :)

scripts/build/builders/nxp.py
```diff
@@ -76,6 +89,20 @@ def FolderName(self, os_env):
...
+    FRDM = auto()
+
+    def BoardVariantName(self, board):
+        if board == NxpBoard.RW61X:
+            if self == NxpBoardVariant.RD:
+                return "rdrw612bga"
+            elif self == NxpBoardVariant.FRDM:
+                return "frdm"
+        else:
+            raise Exception('Unkown board variant: %r' % self)
```
andy31415, line 103:
```suggestion
        if board != NxpBoard.RW61X:
            raise Exception('Board variants only supported for RW61X')

        if self == NxpBoardVariant.RD:
            return "rdrw612bga"
        elif self == NxpBoardVariant.FRDM:
            return "frdm"
```

scripts/build/builders/nxp.py
```diff
@@ -222,26 +254,62 @@ def GnBuildArgs(self):
...
+            flags.append("-DCONFIG_CHIP_LIB_SHELL=true")
+
+        if self.disable_pairing_autostart:
+            flags.append('-DCONFIG_CHIP_ENABLE_PAIRING_AUTOSTART=false')
+
+        if self.board_variant:
+            flag_board_variant = "-DCONFIG_BOARD_VARIANT=\"%s\"" % self.board_variant.BoardVariantName(self.board)
+            flags.append(flag_board_variant)
+
+        build_flags = " ".join(flags) if len(flags) > 0 else ""
```
andy31415, line 307:
```suggestion
        build_flags = " ".join(flags or [])
```

scripts/build/builders/nxp.py
```diff
@@ -222,26 +254,62 @@ def GnBuildArgs(self):
...
+
+        return build_flags
 
     def generate(self):
+        build_flags = self.CmakeBuildFlags()
```
andy31415, line 312: delete this line? You seem to call this directly in the `format` call

andy31415: Actually seems used below ... will update comment.

scripts/build/builders/nxp.py
```diff
@@ -253,11 +321,11 @@ def generate(self):
             cmd += '\nexport ZEPHYR_BASE="$ZEPHYR_NXP_BASE"'
             cmd += '\nunset ZEPHYR_TOOLCHAIN_VARIANT'
 
-            cmd += '\nwest build -p --cmake-only -b {board_name} -d {out_folder} {example_folder}{build_args}'.format(
+            cmd += '\nwest build -p --cmake-only -b {board_name} -d {out_folder} {example_folder} {build_flags}'.format(
                 board_name=self.board.Name(self.os_env),
                 out_folder=self.output_dir,
                 example_folder=self.app.BuildRoot(self.code_root, self.board, self.os_env),
-                build_args=self.WestBuildArgs())
+                build_flags=self.CmakeBuildFlags())
```
andy31415, line 328:
```suggestion
                build_flags=build_flags)
```

andy31415 (approved):

scripts/build/builders/nxp.py
dinabenamar: Updated the PR with your suggestions, thank you !

axelnxp (approved):

doru91 (approved):

PR #35450, opened 2024-09-06
Remove LocalizationConfiguration and TimeFormatLocalization from most samples

tcarmelveilleux:
Problem:

- Most samples don't use (and will never use) localization configuration and time format localization clusters
- Since the samples are widely cut and pasted, we found a lot of products that have "dead" such clusters where they appear and can cause clients to show configuration options, but those options are never used.

This PR:
- Removes those clusters from all samples except Bridge, All Clusters, TV and Thermostat samples, which make use of them in some cases.
- Removes On/Off server from Chef On/Off Light Switch since I was in there and it's an illegal use of On/Off that caused problems in testing.


andy31415 (approved):

bzbarsky-apple (approved):

LuDuda:
@tcarmelveilleux I noticed that in ZAP any change to Endpoint 0 configuration (like adding additional device type next to Root Node) enables automatically Localization Configuration cluster even though .zap didn't have it before enabled - which was surprising for me. I guess it's due to these lines: https://github.com/nrfconnect/sdk-connectedhomeip/blob/master/src/app/zap-templates/zcl/data-model/chip/matter-devices.xml#L73C1-L76C23

Should we disable it by default? I can file a PR if it's only about above line of the code.

PR #35453, opened 2024-09-06
Sync up MTRDeviceController and MTRDeviceController_Concrete and actually start using the latter.

bzbarsky-apple:
Specific changes:

* Fix includes in MTRDeviceController_Concrete, since it uses std::optional and os_unfair_lock, and includes MTRDeviceController.h via MTRDeviceController_Concrete.h already (and has to, because of the inheritance).
* Copy the assertion counter machinery into MTRDeviceController_Concrete for now. This includes the storage of the variables that machinery relies on.
* Remove incorrect deviceMapLock `@synthesize`: this is just implemented by the superclass, no need to do anything with it in the subclass.
* Switch the places that are starting a non-XPC controller to create instances of MTRDeviceController_Concrete, not MTRDeviceController.


woody-apple (approved):

bzbarsky-apple:

PR #35455, opened 2024-09-06
Allow passing in a `Path change listener` to ember write functions

andy31415:
## Call paths description

`MatterAttributeChangeCallback` is actually an ember functionality even though it resides in `src/app/reporting.cpp` (rather than util). In practice we have this callback used by:
   - cluster implementations (to notify when code changes data especially in AAI-backed clusters)
   - `attribute-storage.cpp`, specifically for emberAfEndpointEnableDisable: once for the entire endpoint and then once per Descriptor::PartsList for the root endpoint and parent endpoints
   - `attribute-table.cpp` for emAfWriteAttribute (with various dirty detection options)

The change callback does two things:
   - increase cluster versions (this is kept by EMBER using emberAfDataVersionStorage insice metadata structues)
   - uses the reporting engine in the  InteractionModelEngine singleton to mark paths dirty
   
 ## Where things are not ideal

- MatterAttributeChangeCallback is not obviously `ember code` even though in practice it is (it uses emberAfDataVersionStorage)
- There is a singleton coupling to interactionmodelengine. This makes it impossible for DataModel::Provider to be generic/decoupled because ember hardcodes a InteractionModelEngine singleton coupling

## Changes in the PR

- place the actual implementation of MatterAttributeChangeCallback as `emberAfAttributeChanged` and `emberAfEndpointChanged` calls to make them clearly ember and use only these inside `src/app/util`. Clusters can still use MatterAttributeChangeCallback, we may need to figure out how/if we can make that one decoupled
- define a `AttributesChangedListener` that can be used by the ember callbacks to mark things dirty (instead of accessing the singleton IME object and engine)
   - default implementation is still to use the global InteractionModelEngine
   - this however allows DataModel::Provider in the future to inject its own callback
   - update ember write functions to be able to use a change path listener and also provide a default if the change path listener is not provided


src/app/zap-templates/templates/app/attributes/Accessors-src.zapt
```diff
@@ -109,7 +109,13 @@ Protocols::InteractionModel::Status Get(chip::EndpointId endpoint, {{accessorGet
       Encoding::LittleEndian::Put16(zclString, length);
     {{/if}}
     memcpy(&zclString[{{>sizingBytes}}], value.data(), value.size());
-    return emberAfWriteAttribute(endpoint, {{>clusterId}}, Id, zclString, ZCL_{{typeAsDelimitedMacro type}}_ATTRIBUTE_TYPE{{#if passMarkDirty}}, markDirty{{/if}});
+    {{#if passMarkDirty}}
+    return emberAfWriteAttribute(
+      chip::app::ConcreteAttributePath(endpoint, {{>clusterId}}, Id),
```
bzbarsky-apple, line 114:
```suggestion
      ConcreteAttributePath(endpoint, {{>clusterId}}, Id),
```

src/app/zap-templates/templates/app/attributes/Accessors-src.zapt
```diff
@@ -119,7 +125,13 @@ Protocols::InteractionModel::Status Get(chip::EndpointId endpoint, {{accessorGet
     Traits::StorageType storageValue;
     Traits::WorkingToStorage(value, storageValue);
     uint8_t * writable = Traits::ToAttributeStoreRepresentation(storageValue);
-    return emberAfWriteAttribute(endpoint, {{>clusterId}}, Id, writable, ZCL_{{typeAsDelimitedMacro type}}_ATTRIBUTE_TYPE{{#if passMarkDirty}}, markDirty{{/if}});
+    {{#if passMarkDirty}}
+    return emberAfWriteAttribute(
+      chip::app::ConcreteAttributePath(endpoint, {{>clusterId}}, Id),
```
bzbarsky-apple, line 130:
```suggestion
      ConcreteAttributePath(endpoint, {{>clusterId}}, Id),
```

src/app/zap-templates/templates/app/attributes/Accessors-src.zapt
```diff
@@ -138,13 +150,25 @@ Protocols::InteractionModel::Status Set(chip::EndpointId endpoint, {{zapTypeToEn
 {{#*inline "setNullBody"}}
     {{#if (isString type)}}
     uint8_t zclString[{{>sizingBytes}}] = { {{#if (isShortString type)}}0xFF{{else}}0xFF, 0xFF{{/if}} };
-    return emberAfWriteAttribute(endpoint, {{>clusterId}}, Id, zclString, ZCL_{{typeAsDelimitedMacro type}}_ATTRIBUTE_TYPE{{#if passMarkDirty}}, markDirty{{/if}});
+    {{#if passMarkDirty}}
+    return emberAfWriteAttribute(
+        chip::app::ConcreteAttributePath(endpoint, {{>clusterId}}, Id),
```
bzbarsky-apple, line 155:
```suggestion
        ConcreteAttributePath(endpoint, {{>clusterId}}, Id),
```

src/app/zap-templates/templates/app/attributes/Accessors-src.zapt
```diff
@@ -138,13 +150,25 @@ Protocols::InteractionModel::Status Set(chip::EndpointId endpoint, {{zapTypeToEn
...
+    {{/if}}
     {{else}}
     using Traits = NumericAttributeTraits<{{accessorTraitType type}}>;
     Traits::StorageType value;
     Traits::SetNull(value);
     uint8_t * writable = Traits::ToAttributeStoreRepresentation(value);
-    return emberAfWriteAttribute(endpoint, {{>clusterId}}, Id, writable, ZCL_{{typeAsDelimitedMacro type}}_ATTRIBUTE_TYPE{{#if passMarkDirty}}, markDirty{{/if}});
+    {{#if passMarkDirty}}
+    return emberAfWriteAttribute(
+        chip::app::ConcreteAttributePath(endpoint, {{>clusterId}}, Id),
```
bzbarsky-apple, line 167:
```suggestion
        ConcreteAttributePath(endpoint, {{>clusterId}}, Id),
```

src/app/util/attribute-table.h
```diff
@@ -17,14 +17,54 @@
...
 #include <app/util/af-types.h>
 #include <app/util/attribute-metadata.h>
 #include <lib/core/DataModelTypes.h>
 #include <protocols/interaction_model/StatusCode.h>
 
+struct EmberAfWriteDataInput
+{
+    // where the input data originates from. NOTE: at this time there is no information regarding
+    // input buffer size and it is assumed "correct" for the given data type
+    uint8_t * dataPtr;
```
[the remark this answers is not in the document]
andy31415: It should have been, however existing code takes non-const so I am concerned about blast radius:
https://github.com/project-chip/connectedhomeip/blob/master/src/app/util/attribute-table.h#L54

I think if we want to switch to const (which would be more correct) we should do it as a separate PR.

It probably also will not be as easy because ember has a `ReadOrWrite` can that can do either side, so then the pointer passed to it must be non-const because it may be a read:

https://github.com/project-chip/connectedhomeip/blob/master/src/app/util/attribute-table.cpp#L439

... in the end, it seems to me this is ugly but fixing it is rough.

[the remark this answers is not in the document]
andy31415: I added a comment about this oddity. I think fix should not be part of this PR - existing methods were already taking in non-const, so I did not change that.

src/app/util/attribute-storage.h
```diff
@@ -284,6 +284,22 @@ chip::Span<const EmberAfDeviceType> emberAfDeviceTypeListFromEndpoint(chip::Endp
...
 
+/// Returns a change listener that uses the global InteractionModelEngine
+/// instance to report dirty paths
+chip::app::AttributesChangedListener * emberAfGlobalInteractionModelAttributesChangedListener();
+
+/// Mark the given attribute as having changed:
+///   - increases the cluster data version for the given cluster
+///   - schedules reporting to consider the given attribute as dirty
+void emberAfAttributeChanged(chip::EndpointId endpoint, chip::ClusterId clusterId, chip::AttributeId attributeId,
+                             chip::app::AttributesChangedListener * listener);
```
tcarmelveilleux, line 295: How do you provide this argument? It's not clear how `listener` would be passed from the function docs.

src/app/util/attribute-storage.cpp
```diff
@@ -1462,3 +1461,48 @@ DataVersion * emberAfDataVersionStorage(const chip::app::ConcreteClusterPath & a
...
+    DataVersion * version = emberAfDataVersionStorage(ConcreteClusterPath(endpoint, clusterId));
+    if (version == nullptr)
+    {
+        ChipLogError(DataManagement, "Endpoint %x, Cluster " ChipLogFormatMEI " not found in IncreaseClusterDataVersion!", endpoint,
+                     ChipLogValueMEI(clusterId));
+    }
+    else
+    {
+        (*(version))++;
+        ChipLogDetail(DataManagement, "Endpoint %x, Cluster " ChipLogFormatMEI " update version to %" PRIx32, endpoint,
```
tcarmelveilleux, line 1498: This seems like it would be quite verbose...

src/app/util/attribute-table.h
```diff
@@ -17,14 +17,60 @@
...
 #include <app/util/attribute-metadata.h>
 #include <lib/core/DataModelTypes.h>
 #include <protocols/interaction_model/StatusCode.h>
 
+#include <cstdint>
+
+struct EmberAfWriteDataInput
+{
+    // Where the input data originates from. NOTE: at this time there is no information regarding
+    // input buffer size and it is assumed "correct" for the given data type
```
tcarmelveilleux, line 31:
```suggestion
    // Where the input data originates from. 
    //
    // NOTE: at this time there is no information regarding
    // input buffer size and it is assumed "correct" for the given data type.
```

src/app/util/attribute-table.h
```diff
@@ -17,14 +17,60 @@
...
 
+#include <app/ConcreteAttributePath.h>
 #include <app/util/af-types.h>
 #include <app/util/attribute-metadata.h>
 #include <lib/core/DataModelTypes.h>
 #include <protocols/interaction_model/StatusCode.h>
 
+#include <cstdint>
+
+struct EmberAfWriteDataInput
```
tcarmelveilleux, line 28: Suggest `EmberAfWriteParameters`

tcarmelveilleux (approved):

src/app/util/attribute-table.h
andy31415: I considered this, however `Parameters` generally sounds like the whole set of parameters, however our write functions generally have a path which I kept split as I was unclear if we actually do want a full parameters only structure here. So the logic was we have sparate "where to write" (path) and "what to write" (which is this input), although I do agree it is somewhat odd that this "what" also contains information about how to mark change and where to mark the changes callbacks.

src/app/util/attribute-storage.cpp
andy31415: This is where this lived before: https://github.com/project-chip/connectedhomeip/blob/master/src/app/reporting/reporting.cpp#L40

src/app/AttributePathParams.h
```diff
@@ -80,7 +87,7 @@ struct AttributePathParams
         mListIndex   = kInvalidListIndex;
     }
 
-    bool IsAttributePathSupersetOf(const AttributePathParams & other) const
+    [[nodiscard]] bool IsAttributePathSupersetOf(const AttributePathParams & other) const
```
tcarmelveilleux, line 90: Why are the [[nodiscard]] needed?

src/app/util/attribute-storage.h
andy31415: I updated the comment a bit, referencing the listener instead of the old `schedules reporting`.

src/app/util/attribute-storage.cpp
```diff
@@ -929,8 +929,9 @@ bool emberAfEndpointEnableDisable(EndpointId endpoint, bool enable)
         EndpointId parentEndpointId = emberAfParentEndpointFromIndex(index);
         while (parentEndpointId != kInvalidEndpointId)
         {
-            MatterReportingAttributeChangeCallback(parentEndpointId, app::Clusters::Descriptor::Id,
-                                                   app::Clusters::Descriptor::Attributes::PartsList::Id);
+            emberAfAttributeChanged(parentEndpointId, app::Clusters::Descriptor::Id,
```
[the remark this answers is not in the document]
bzbarsky-apple: Yeah, for "all" I meant "in the code we are touching anyway"...

bzbarsky-apple (approved):

PR #35462, opened 2024-09-06
Add GetNextAction API so that a caller that might want to implement a…

### LINKED ISSUE #35461: Add an API to the BDX transfer session so that a caller can get the next output event generated by the TransferSession state machine without polling.
Currently, there is a PollOutput API which requires the caller to poll frequently to get the next output event generated by the TransferSession and the caller needs to wait up to the polling interval to get the next output event which makes the BDX transfer slower. We would like to add an asynchronous transfer facilitator that can use an event driven approach to drive the BDX transfer and be able to get the next output event on demand rather than waiting on a polling based mechanism.

nivi-apple:
…n event driven mechanism to drive the BDX transfer can get the next output event from the TransferSession on demand

Currently we have the PollOutput API where the expectation is the caller needs to poll frequently for the next output event from the TransferSession which introduces a delay based on the poll interval in handling every single BDX message or transfer session event handled by the caller.

Fixes: #35461

### DIFF
--- a/src/protocols/bdx/BdxTransferSession.h
+++ b/src/protocols/bdx/BdxTransferSession.h
@@ -172,7 +172,8 @@ public:
      *   the caller will get an event of type OutputEventType::kNone.
      *
      *   It is possible that consecutive calls to this method may emit different outputs depending on the state of the
-     *   TransferSession object.  The caller is generally expected to keep calling this method until it gets an event of type OutputEventType::kNone.
+     *   TransferSession object.  The caller is generally expected to keep calling this method until it gets an event of type
+     * OutputEventType::kNone.
      *
      *   If the output event type is kMsgToSend, the caller is expected to send the message immediately on the
      *   relevant exchange.  In this case the BDX session timeout timer will start when GetNextAction is called.
--- a/src/protocols/bdx/BdxTransferSession.h
+++ b/src/protocols/bdx/BdxTransferSession.h
@@ -173,8 +173,10 @@ public:
      *   It is possible that consecutive calls to this method may emit different outputs depending on the state of the
      *   TransferSession object and so we need to call this until we get an event of type - OutputEventType::kNone
      *
-     *   Note that if the event type is kMsgToSend, the caller is expected to send the message immediately over the
-     *   exchange otherwise the caller is expected to pass the event to HandleTransferSessionOutput to handle the BDX
+     *   If the output event type is kMsgToSend, the caller is expected to send the message immediately on the
+     *   relevant exchange.  In this case the BDX session timeout timer will start when GetNextAction is called.
+     *
+     *   For all other event types, the caller is expected to pass the event to HandleTransferSessionOutput to handle the BDX
      *   message or any other TransferSession event.
      *
      *   See OutputEventType for all possible output event types.
--- a/src/protocols/bdx/BdxTransferSession.h
+++ b/src/protocols/bdx/BdxTransferSession.h
@@ -167,6 +167,24 @@ public:
 
     /**
      * @brief
+     *   Gets the pending output event from the transfer session in the event param passed in by the caller.
+     *   The output event may contain some data for the caller to act upon.
+     *
+     *   It is possible that consecutive calls to this method may emit different outputs depending on the state of the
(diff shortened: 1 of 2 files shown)


src/protocols/bdx/BdxTransferSession.h
```diff
@@ -165,6 +165,24 @@ class DLL_EXPORT TransferSession
      */
     void PollOutput(OutputEvent & event, System::Clock::Timestamp curTime);
 
+    /**
+     * @brief
+     *   Gets the pending output event from the transfer session in the event param passed in by the caller.
+     *   The output event may contain some data for the caller to act upon.
```
bzbarsky-apple, line 171: What happens if there is no pending output event?  Should be documented.

src/protocols/bdx/BdxTransferSession.h
```diff
@@ -165,6 +165,24 @@ class DLL_EXPORT TransferSession
...
+     *
+     *   It is possible that consecutive calls to this method may emit different outputs depending on the state of the
+     *   TransferSession object and so we need to call this until we get an event of type - OutputEventType::kNone
+     *
+     *   Note that if the event type is kMsgToSend, the caller is expected to send the message immediately over the
+     *   exchange otherwise the caller is expected to pass the event to HandleTransferSessionOutput to handle the BDX
+     *   message or any other TransferSession event.
```
bzbarsky-apple, line 178:
```suggestion
     *   If the output event type is kMsgToSend, the caller is expected to send the message immediately on the
     *   relevant exchange.  In this case the BDX session timeout timer will start when GetNextAction is called.
     *
     *   For all other event types, the caller is expected to pass the event to HandleTransferSessionOutput to handle the BDX
     *   message or any other TransferSession event.
```

src/protocols/bdx/BdxTransferSession.h
```diff
@@ -165,6 +165,24 @@ class DLL_EXPORT TransferSession
...
+    /**
+     * @brief
+     *   Gets the pending output event from the transfer session in the event param passed in by the caller.
+     *   The output event may contain some data for the caller to act upon.
+     *
+     *   It is possible that consecutive calls to this method may emit different outputs depending on the state of the
+     *   TransferSession object and so we need to call this until we get an event of type - OutputEventType::kNone
+     *
+     *   Note that if the event type is kMsgToSend, the caller is expected to send the message immediately over the
+     *   exchange otherwise the caller is expected to pass the event to HandleTransferSessionOutput to handle the BDX
```
bzbarsky-apple, line 177: That said, what is HandleTransferSessionOutput?  It's not related to BdxTransferSession, right?  This documentation is not making sense  to me.

nivi-apple: I guess i should remove the part about HandleTransferSessionOutput. its a virtual method that should be implemented by a subsclass of the TransferFacilitator to have the business logic to handle BDX messages or transfer session events.

src/protocols/bdx/BdxTransferSession.h
```diff
@@ -165,6 +165,24 @@ class DLL_EXPORT TransferSession
      */
     void PollOutput(OutputEvent & event, System::Clock::Timestamp curTime);
 
+    /**
+     * @brief
+     *   Gets the pending output event from the transfer session in the event param passed in by the caller.
+     *   The output event may contain some data for the caller to act upon. If there is no pending output event,
+     *   the caller will get an event of type - OutputEventType::kNone.
```
bzbarsky-apple, line 172:
```suggestion
     *   the caller will get an event of type OutputEventType::kNone.
```

src/protocols/bdx/BdxTransferSession.h
```diff
@@ -165,6 +165,24 @@ class DLL_EXPORT TransferSession
...
+     *
+     *   It is possible that consecutive calls to this method may emit different outputs depending on the state of the
+     *   TransferSession object and the caller needs to call this until we get an event of type - OutputEventType::kNone
```
bzbarsky-apple, line 175:
```suggestion
     *   TransferSession object.  The caller is generally expected to keep calling this method until it gets an event of type OutputEventType::kNone.
```

bzbarsky-apple (approved):

andy31415 (approved):

PR #35464, opened 2024-09-06
Implement MTRDevice handling in controller suspend/resume.

bzbarsky-apple:
Also blocks acquisition of CASE sessions on a suspended controller, which should ensure that new requests for MTRDevices and MTRBaseDevices associated with the controller do not hit the network.

Draft for now until https://github.com/project-chip/connectedhomeip/pull/35453 lands.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDeviceControllerParameters.h
+++ b/src/darwin/Framework/CHIP/MTRDeviceControllerParameters.h
@@ -150,6 +150,11 @@ MTR_AVAILABLE(ios(17.6), macos(14.6), watchos(10.6), tvos(17.6))
                 intermediateCertificate:(MTRCertificateDERBytes _Nullable)intermediateCertificate
                         rootCertificate:(MTRCertificateDERBytes)rootCertificate;
 
+/**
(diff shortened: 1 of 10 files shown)


src/darwin/Framework/CHIP/MTRDeviceControllerParameters.h
```diff
@@ -150,6 +150,11 @@ MTR_AVAILABLE(ios(17.6), macos(14.6), watchos(10.6), tvos(17.6))
                 intermediateCertificate:(MTRCertificateDERBytes _Nullable)intermediateCertificate
                         rootCertificate:(MTRCertificateDERBytes)rootCertificate;
 
+/**
+ * The root certificate we were initialized with.
+ */
+@property (nonatomic, copy, readonly) MTRCertificateDERBytes rootCertificate MTR_NEWLY_AVAILABLE;
```
bzbarsky-apple, line 156: I added this because it makes the unit test easier to write and it made sense to have this, but if needed I can try to do the unit test in some other way.

jtung-apple (approved):

bzbarsky-apple:

PR #35472, opened 2024-09-07
[Darwin] Darwin CI TSAN test failure fix

jtung-apple:
TSAN complained about accessing isSuspended while not holding the _suspensionLock lock. This is the change that should fix the intermittent failures.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDeviceController.mm
+++ b/src/darwin/Framework/CHIP/MTRDeviceController.mm
@@ -350,7 +350,7 @@ using namespace chip::Tracing::DarwinFramework;
 
 - (BOOL)isSuspended
 {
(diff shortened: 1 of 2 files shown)


woody-apple (approved):

woody-apple (approved):

src/darwin/Framework/CHIP/MTRDeviceController.mm
```diff
@@ -352,6 +352,7 @@ - (BOOL)isRunning
 
 - (BOOL)isSuspended
 {
+    std::lock_guard lock(_suspensionLock);
```
jtung-apple, line 355: Adding comment to avoid auto merge while I investigate other related flaky tests, or re-running successful runs.

PR #35475, opened 2024-09-08
[Darwin] Implement delegate add/remove for device controller

jtung-apple:
This PR implements add/remove semantics for MTRDeviceController delegate.

The add/remove delegates will be weak references but will keep a strong reference for delegates set from the existing setDeviceControllerDelegate API, to keep old behavior.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDefines_Internal.h
+++ b/src/darwin/Framework/CHIP/MTRDefines_Internal.h
@@ -150,3 +150,7 @@ typedef struct {} variable_hidden_by_mtr_hide;
         }                                                                                                                  \
     }
 #endif
+
+#ifndef YES_NO
+#define YES_NO(x) ((x) ? @"YES" : @"NO")
+#endif
--- a/src/darwin/Framework/CHIP/MTRDeviceController.h
+++ b/src/darwin/Framework/CHIP/MTRDeviceController.h
@@ -186,7 +186,7 @@ MTR_AVAILABLE(ios(16.1), macos(13.0), watchos(9.1), tvos(16.1))
 - (void)preWarmCommissioningSession MTR_DEPRECATED("-[MTRDeviceControllerFactory preWarmCommissioningSession]", ios(16.4, 17.6), macos(13.3, 14.6), watchos(9.4, 10.6), tvos(16.4, 17.6));
 
 /**
- * Set the Delegate for the device controller  as well as the Queue on which the Delegate callbacks will be triggered
+ * Set the Delegate for the device controller as well as the Queue on which the Delegate callbacks will be triggered
  *
  * @param[in] delegate The delegate the commissioning process should use
  *
@@ -194,6 +194,23 @@ MTR_AVAILABLE(ios(16.1), macos(13.0), watchos(9.1), tvos(16.1))
  */
 - (void)setDeviceControllerDelegate:(id<MTRDeviceControllerDelegate>)delegate
                               queue:(dispatch_queue_t)queue MTR_AVAILABLE(ios(16.4), macos(13.3), watchos(9.4), tvos(16.4));
+
+/**
+ * Adds a Delegate to the device controller as well as the Queue on which the Delegate callbacks will be triggered
+ *
+ * @param[in] delegate The delegate the commissioning process should use
+ *
+ * @param[in] queue The queue on which the callbacks will be delivered
+ */
+- (void)addDeviceControllerDelegate:(id<MTRDeviceControllerDelegate>)delegate
+                              queue:(dispatch_queue_t)queue MTR_NEWLY_AVAILABLE;
+
+/**
+ * Removes a Delegate from the device controller
+ *
+ * @param[in] delegate The delegate to be removed
+ */
+- (void)removeDeviceControllerDelegate:(id<MTRDeviceControllerDelegate>)delegate MTR_NEWLY_AVAILABLE;
 
 /**
  * Start scanning for commissionable devices.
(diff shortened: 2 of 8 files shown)


woody-apple (approved):

src/darwin/Framework/CHIP/MTRDefines_Internal.h
```diff
@@ -150,3 +150,7 @@ typedef struct {} variable_hidden_by_mtr_hide;
         }                                                                                                                  \
     }
 #endif
+
+#ifndef YES_NO
```
ksperling-apple, line 154: I think it's safer to prefix this as `MTR_YES_NO`. Also no need for the ifdef -- if the macro already has a conflicting definition a warning / error would be better than silently the wrong macro.

src/darwin/Framework/CHIP/MTRDeviceController.mm
```diff
@@ -1751,6 +1772,158 @@ + (void)forceLocalhostAdvertisingOnly
...
+        }
+
+        return _delegates.count;
+    }
+}
+
+- (void)_callDelegatesWithBlock:(void (^_Nullable)(id<MTRDeviceControllerDelegate> delegate))block logString:(const char *)logString;
+{
+    NSUInteger delegatesCalled = [self _iterateDelegateInfoWithBlock:^(MTRDeviceControllerDelegateInfo * delegateInfo) {
+        id<MTRDeviceControllerDelegate> strongDelegate = delegateInfo.delegate;
```
ksperling-apple, line 1868: A potential problem for an API client here is that callbacks can end up being called after the delegate is removed.
DelegateInfo might need some atomic flag on it that can be cleared when the delegate is removed, and then re-checked here before the actual call is made. It would also be ideal if `respondsToSelector` checks were made before dispatching.

src/darwin/Framework/CHIPTests/TestHelpers/MTRTestDeclarations.h
```diff
@@ -55,6 +58,7 @@ NS_ASSUME_NONNULL_BEGIN
 #ifdef DEBUG
 @interface MTRDeviceController (TestDebug)
 - (NSDictionary<NSNumber *, NSNumber *> *)unitTestGetDeviceAttributeCounts;
+- (NSUInteger)unitTestDelegateCount;
```
ksperling-apple, line 61: It would be great if we could move away from adding more stuff to this header, and instead use `*_Test.h` headers that are part of the public / test / internal header stack for group of .mm / .h files. The main advantage of that approach is that the test declarations are actually visible to the compiler (and other developers) when compiling the implementation.

src/darwin/Framework/CHIP/MTRDeviceController.h
```diff
@@ -195,6 +195,23 @@ MTR_AVAILABLE(ios(16.1), macos(13.0), watchos(9.1), tvos(16.1))
 - (void)setDeviceControllerDelegate:(id<MTRDeviceControllerDelegate>)delegate
                               queue:(dispatch_queue_t)queue MTR_AVAILABLE(ios(16.4), macos(13.3), watchos(9.4), tvos(16.4));
 
+/**
+ * Adds a Delegate to the device controller as well as the Queue on which the Delegate callbacks will be triggered
+ *
+ * @param[in] delegate The delegate the commissioning process should use
```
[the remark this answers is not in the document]
woody-apple: This is going to be used for suspension.

src/darwin/Framework/CHIP/MTRDeviceController.mm
```diff
@@ -567,6 +593,10 @@ - (void)cleanup
     if (_deviceControllerDelegateBridge) {
         delete _deviceControllerDelegateBridge;
         _deviceControllerDelegateBridge = nullptr;
+        @synchronized(self) {
```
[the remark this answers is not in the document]
woody-apple: He tried with locks before, and found a recursive issue. We're going to move over to `@synchronized`, so this is OK for now.

src/darwin/Framework/CHIP/MTRDefines_Internal.h
jtung-apple: Filed https://github.com/project-chip/connectedhomeip/issues/35494

src/darwin/Framework/CHIP/MTRDeviceController.h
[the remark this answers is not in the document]
jtung-apple: Tracking documentation improvement in https://github.com/project-chip/connectedhomeip/issues/35494

src/darwin/Framework/CHIP/MTRDeviceController.mm
```diff
@@ -139,6 +160,9 @@ @implementation MTRDeviceController {
     NSUInteger _keepRunningAssertionCounter;
     BOOL _shutdownPending;
     os_unfair_lock _assertionLock;
+
+    NSMutableSet<MTRDeviceControllerDelegateInfo *> * _delegates;
```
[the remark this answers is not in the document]
jtung-apple: As is you're right. I'll make a change in the follow up tracked in https://github.com/project-chip/connectedhomeip/issues/35494

src/darwin/Framework/CHIP/MTRDeviceController.mm
jtung-apple: Filed https://github.com/project-chip/connectedhomeip/issues/35495

src/darwin/Framework/CHIPTests/TestHelpers/MTRTestDeclarations.h
jtung-apple: Filed https://github.com/project-chip/connectedhomeip/issues/35496

PR #35486, opened 2024-09-09
Increase allowed time for subscription setup in test013_suspendDevices.

bzbarsky-apple:
3 seconds is often not enough time to set up a subscription in CI.  Make sure to use a longer timeout in the places where we are expecting subscription setup.


woody-apple (approved):

jtung-apple (approved):

woody-apple:
Force merging, since the jobs passed that needed, and this is causing flakes.

PR #35490, opened 2024-09-09
Ensure that all MTRDevice state/internalState changes happen on the Matter queue.

### LINKED ISSUE #34796: [Darwin] MTRDevice ReadClient callbacks need to change state immediately

bzbarsky-apple:
This avoids races where we queue blocks to different queues that both try to change the state, which were resulting in non-deterministic final state.

Fixes https://github.com/project-chip/connectedhomeip/issues/34796


woody-apple (approved):

woody-apple:
Causing flakes in CI, tagging.

woody-apple:
Merging given the impacted builds passed and this is causing flakes.

PR #35491, opened 2024-09-09
Move management of uniqueIdentifier into MTRDeviceController base class.

bzbarsky-apple:


src/darwin/Framework/CHIP/MTRDeviceController.mm
```diff
@@ -148,11 +148,14 @@ - (os_unfair_lock_t)deviceMapLock
...
 }
 
-- (instancetype)initForSubclasses:(BOOL)startSuspended
+- (instancetype)initForSubclasses:(BOOL)startSuspended uniqueIdentifier:(NSUUID *)uniqueIdentifier
 {
     if (self = [super init]) {
         // nothing, as superclass of MTRDeviceController is NSObject
     }
+
+    _uniqueIdentifier = uniqueIdentifier;
```
woody-apple, line 157:
```suggestion
    self.uniqueIdentifier = uniqueIdentifier;
```

src/darwin/Framework/CHIP/MTRDeviceController_Internal.h
```diff
@@ -69,13 +69,11 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, readonly) NSMapTable<NSNumber *, MTRDevice *> * nodeIDToDeviceMap;
 @property (readonly, assign) os_unfair_lock_t deviceMapLock;
 
-@property (readwrite, nonatomic) NSUUID * uniqueIdentifier;
```
woody-apple, line 72: Still need this to use self.

src/darwin/Framework/CHIP/MTRDeviceController.mm
```diff
@@ -148,11 +148,14 @@ - (os_unfair_lock_t)deviceMapLock
     return &_underlyingDeviceMapLock;
 }
 
-- (instancetype)initForSubclasses:(BOOL)startSuspended
+- (instancetype)initForSubclasses:(BOOL)startSuspended uniqueIdentifier:(NSUUID *)uniqueIdentifier
```
woody-apple, line 151: General comment here to hold this until I test this, this uniqueIdentifier stuff was broken by a similar change last week, and I need to verify this against cleints.

src/darwin/Framework/CHIP/MTRDeviceController_Internal.h
bzbarsky-apple: But why do we want to use self for the set?  This thing should be set once at init time, never modified again....

src/darwin/Framework/CHIP/MTRDeviceController.mm
bzbarsky-apple: Yep, can do.  Let me know.

src/darwin/Framework/CHIP/MTRDeviceController_Internal.h
woody-apple: I'd like us to move to objective C patterns, where we use properties for this, so we can fully utilize `@synchronized` and atomic properly.

bzbarsky-apple: OK.  Well, we can put the readwrite property into the `@implementation`, at least.  I'll do that.

src/darwin/Framework/CHIP/MTRDeviceController.mm
bzbarsky-apple: Done.

src/darwin/Framework/CHIP/MTRDeviceController_Internal.h
bzbarsky-apple: Moved into `@implementation`

woody-apple (approved):

PR #35498, opened 2024-09-10
[Darwin] MTRDeviceController initial device retain should not depend on matter queue

jtung-apple:
As title says, this part has nothing to do with matter queue, and is only there to hold retain count for a short period of time.


bzbarsky-apple (approved):

kiel-apple (approved):

bzbarsky-apple:

PR #35499, opened 2024-09-10
Adding isSuspended delegate for MTRDeviceController

woody-apple:
Adding this delegate

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDeviceController.mm
+++ b/src/darwin/Framework/CHIP/MTRDeviceController.mm
@@ -406,6 +406,13 @@ using namespace chip::Tracing::DarwinFramework;
         // * CASE sessions in general.
         // * Possibly try to see whether we can change our fabric entry to not advertise and restart advertising.
     }
+    
+    BOOL isSuspended = [self isSuspended];
+    [self _callDelegatesWithBlock:^(id<MTRDeviceControllerDelegate> delegate) {
+        if ([delegate respondsToSelector:@selector(controller:isSuspended:)]) {
+            [delegate controller:self isSuspended: isSuspended];
+        }
+    } logString:__PRETTY_FUNCTION__];
 }
 
 - (void)resume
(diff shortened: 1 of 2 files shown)


src/darwin/Framework/CHIP/MTRDeviceController.mm
```diff
@@ -406,6 +406,13 @@ - (void)suspend
         // * CASE sessions in general.
         // * Possibly try to see whether we can change our fabric entry to not advertise and restart advertising.
     }
+
+    BOOL isSuspended = [self isSuspended];
+    [self _callDelegatesWithBlock:^(id<MTRDeviceControllerDelegate> delegate) {
+        if ([delegate respondsToSelector:@selector(controller:isSuspended:)]) {
+            [delegate controller:self isSuspended:isSuspended];
+        }
+    } logString:__PRETTY_FUNCTION__];
```
jtung-apple, line 415: There should also be a `_callDelegatesWithBlock` call below in the resume code.

jtung-apple (approved):

src/darwin/Framework/CHIP/MTRDeviceControllerDelegate.h
```diff
@@ -106,6 +106,13 @@ MTR_AVAILABLE(ios(16.4), macos(13.3), watchos(9.4), tvos(16.4))
  */
 - (void)controller:(MTRDeviceController *)controller
     readCommissioningInfo:(MTRProductIdentity *)info MTR_AVAILABLE(ios(17.0), macos(14.0), watchos(10.0), tvos(17.0));
+
+/**
+ * Notify the delegate when the suspended state changed of the controller, after this happens
```
bzbarsky-apple, line 111: Two things:

1) "the suspended state of the controller changes".
2) "The claim that after that the controller is in the specified state does not happen to be true; it could be in the other state but we have just not gotten our async notification for that yet....

PR #35501, opened 2024-09-10
[v1.3][ESP32] Support total operational hours (#35425)

shubhamdp:
Backporting #35425 to v1.3-branch


src/test_driver/esp32/sdkconfig.defaults
```diff
@@ -31,6 +31,7 @@ CONFIG_ESP_MAIN_TASK_STACK_SIZE=32768
 
 #enable BT
 CONFIG_BT_ENABLED=y
+CONFIG_BT_NIMBLE_ENABLED=y
```
dhrishi, line 34: Why do you need to explicitly add this? Is NimBLE not enabled by default, when BT is enabled?

src/platform/ESP32/ConfigurationManagerImpl.cpp
```diff
@@ -161,18 +173,34 @@ CHIP_ERROR ConfigurationManagerImpl::Init()
...
+    if (CHIP_NO_ERROR != GetTotalOperationalHours(mTotalOperationalHours))
     {
-        err = StoreTotalOperationalHours(0);
+        err = StoreTotalOperationalHours(mTotalOperationalHours);
         SuccessOrExit(err);
     }
 
+    {
+        // Start a timer which reloads every one hour and bumps the total operational hours
+        TickType_t reloadPeriod   = (1000 * 60 * 60) / portTICK_PERIOD_MS;
```
dhrishi, line 184: Is it required to add a timer specifically for this? Is there any IDF API that gives time (ms) since bootup? If yes, can we just use that and convert to hrs when a request is made?

src/test_driver/esp32/sdkconfig.defaults
shubhamdp: Default is bluedroid, needs this to fix the CI.

shubhamdp: https://github.com/espressif/esp-idf/blob/master/components/bt/Kconfig#L12

src/platform/ESP32/ConfigurationManagerImpl.cpp
shubhamdp: Yes, it is required to add the timer for this. I think you are talking about uptime which is time since bootup. This is total operational hours, and needs to be bumped every hour, and must be persisted across reboots. There is no API for that in esp-idf.

dhrishi (approved):

PR #35507, opened 2024-09-10
Fix max timeout for open commissioning window

### LINKED ISSUE #35505: [BUG] Commissioning window timeout can be set to the value incompatible with the spec
### Reproduction steps

The `CommissioningWindorManager::MaxCommissioningTimeout` is used to determine the maximum allowed commissioning timeout used in the  `CommissioningWindowManager::OpenCommissioningWindow` method. This method was modified btw of adding BLE extended announcement feature: https://github.com/project-chip/connectedhomeip/commit/ca17912497a7eb8db22d42ddbe55f36370214d8c#diff-d6f2798a18ed8d25dbb88df7816ca0584e4d6869dc2e4618dc74ee494db7f314R64 and if `CHIP_DEVICE_CONFIG_BLE_EXT_ADVERTISING` is enabled, the returned value is equal to 48 hours.

This makes the implementation return timeout equal to 48 hours, even for the commissioned device, and the spec does not allow the window to exceed 900 s in such scenario.

It makes the CADMIN 1.21 i 1.22 certification tests fail.

### Bug prevalence

Always

83d345e18cbff6943e4925a5936a3edc72ab076b (master)

### Platform

core

### Platform Version(s)

v1.3.0

### Anything else?

_No response_

kkasperczyk-no:
Commissioning window can be opened using timeout exceeding
the maximum value of 900 s defined by the spec. This can happen
if selected transport is IP, but the device uses BLE extended
announcement feature.

Added checking if device is commissioned to be able to determine
what max timeout should be used for the particular scenario.

Renamed CHIP_DEVICE_CONFIG_BLE_EXT_ADVERTISING to the
CHIP_DEVICE_CONFIG_EXT_ADVERTISING, as config name sounds
misleading and it seems it relates only to BLE.

Fixes: #35505


src/app/server/CommissioningWindowManager.cpp
```diff
@@ -306,11 +306,13 @@ CHIP_ERROR CommissioningWindowManager::OpenBasicCommissioningWindow(Seconds32 co
...
 {
     RestoreDiscriminator();
+    bool isBle = false;
 
 #if CONFIG_NETWORK_LAYER_BLE
     // Enable BLE advertisements if commissioning window is to be opened on all supported
     // transports, and BLE is supported on the current device.
-    SetBLE(advertisementMode == chip::CommissioningWindowAdvertisement::kAllSupported);
+    isBle = advertisementMode == chip::CommissioningWindowAdvertisement::kAllSupported;
+    SetBLE(isBle);
```
bzbarsky-apple, line 315: So... "Extended Announcement" is not limited to BLE, as far as I can tell.  All it requires is that the device be "uncommissioned".

So it seems like that is what we should be implementing here: when asked to open a basic commissioning window, if we are not already commissioned, allow extended announcement duration.  In all other cases, disallow it.  No relationship to BLE.

bzbarsky-apple (changes requested):

src/app/server/CommissioningWindowManager.cpp
kkasperczyk-no: You are right, thanks for catching it. In such case the `CHIP_DEVICE_CONFIG_BLE_EXT_ADVERTISING` define sounds a little bit misleading. I will update it in this PR by removing the BLE part.

kkasperczyk-no: Another thing is that extended announcement seems to not be 100% ready for the IP transport, as in the dnssd implementation the VID and PID are always added to the commissioning data. Probably it should work similarly to BLE and update advertised data after exceeding the timeout and entering the extended part.

markaj-nordic (approved):

src/app/server/CommissioningWindowManager.cpp
bzbarsky-apple: Yes, that last bit needs to be fixed if we're going to support this at all.  Is there an issue tracking that?

bzbarsky-apple (approved):

LuDuda (approved):

src/app/server/CommissioningWindowManager.cpp
kkasperczyk-no: I've created one: https://github.com/project-chip/connectedhomeip/issues/35592

PR #35523, opened 2024-09-10
[OSS-Fuzz] Fix static library linking needed for OSS-Fuzz

Alami-Amine:
- This issue showed up after #35408
- It is related to linker errors due to missing libs

- OSS-Fuzz requires libraries to be linked statically.
- When building for OSS-Fuzz , `pkg-config.py` only outputs the direct dependencies and misses out the transitive libs that need to be explicitly linked (when linking statically).
- Fix involves  adding `--static` flag to pkg-config invocation, when we are building for OSS-Fuzz.

### Testing
- This was Tested on OSS-Fuzz Image.
- Actions Job can be seen: https://github.com/google/oss-fuzz/actions/runs/10798522931/job/29952211472?pr=12461

##### Output of `pkg-config.py` without `--static`

```python
$ python3 build/config/linux/pkg-config.py  gio-2.0
[true,
 ["/usr/include/libmount", "/usr/include/blkid", "/usr/include/glib-2.0", "/usr/lib/x86_64-linux-gnu/glib-2.0/include"], 
 [], 
 #Below is the list with libs
 ["gio-2.0", "gobject-2.0", "glib-2.0"], 
 []]
```

##### Output of `pkg-config` with `--static` 
```python
$ python3 build/config/linux/pkg-config.py --static gio-2.0
[true,
 ["/usr/include/libmount", "/usr/include/blkid", "/usr/include/glib-2.0", "/usr/lib/x86_64-linux-gnu/glib-2.0/include"], 
 [], 
  #Below is the list with libs that includes transitive libs
 ["gio-2.0", "gmodule-2.0", "z", "mount", "dl", "blkid", "selinux", "sepol", "pcre2-8", "gobject-2.0", "ffi", "glib-2.0", "m", "pcre"], 
 []]

```

### DIFF
--- a/build/config/linux/pkg-config.py
+++ b/build/config/linux/pkg-config.py
@@ -142,6 +142,8 @@ def main():
                       dest='dridriverdir')
     parser.add_option('--version-as-components', action='store_true',
                       dest='version_as_components')
+    parser.add_option('--static', action='store_true',
+                      dest='static')
     (options, args) = parser.parse_args()
 
     # Make a list of regular expressions to strip out.
@@ -203,7 +205,11 @@ def main():
         sys.stdout.write(dridriverdir.strip())
         return
 
-    cmd = [options.pkg_config, "--cflags", "--libs"] + args
+    if options.static:
+        cmd = [options.pkg_config, "--cflags", "--libs", "--static"] + args
+    else:
+        cmd = [options.pkg_config, "--cflags", "--libs"] + args
+
     if options.debug:
         sys.stderr.write('Running: %s\n' % ' '.join(cmd))
 
--- a/build/config/linux/pkg-config.py
+++ b/build/config/linux/pkg-config.py
@@ -205,10 +205,12 @@ def main():
         sys.stdout.write(dridriverdir.strip())
         return
 
+    cmd = [options.pkg_config, "--cflags", "--libs"]
+
     if options.static:
-        cmd = [options.pkg_config, "--cflags", "--libs", "--static"] + args
-    else:
-        cmd = [options.pkg_config, "--cflags", "--libs"] + args
+        cmd.append("--static")
+
+    cmd.extend(args)
 
     if options.debug:
         sys.stderr.write('Running: %s\n' % ' '.join(cmd))
--- a/build/config/linux/pkg_config.gni
+++ b/build/config/linux/pkg_config.gni
@@ -137,13 +137,16 @@ template("pkg_config") {
 
       # Link libraries statically for OSS-Fuzz fuzzer build
       if (oss_fuzz) {
-        #outputs libs needed for static linking (direct + transitive/non-direct libs), we will re-execute the script to get those libs
+        # Output libs needed for static linking (direct + transitive/non-direct libs), we will re-execute the script to get those libs
         args += [ "--static" ]
         pkgresult_static = exec_script(pkg_config_script, args, "value")
         libs = []
         ldflags = [ "-Wl,-Bstatic" ]
         foreach(lib, pkgresult_static[3]) {
-          #dl(dynamic loading) lib is not needed for linking statically and its presence triggers errors.
+          # dl(dynamic loading) lib is not needed for linking statically and its presence triggers errors.
+          # example of errors:
+          #   ld.lld: error: undefined symbol: __dlsym
+          #   >>> referenced by dlsym.o:(dlsym) in archive /lib/x86_64-linux-gnu/libdl.a
           if (lib != "dl") {
             ldflags += [ "-l$lib" ]
           }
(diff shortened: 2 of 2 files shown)


build/config/linux/pkg-config.py
```diff
@@ -203,7 +205,11 @@ def main():
         sys.stdout.write(dridriverdir.strip())
         return
 
-    cmd = [options.pkg_config, "--cflags", "--libs"] + args
+    if options.static:
+        cmd = [options.pkg_config, "--cflags", "--libs", "--static"] + args
+    else:
+        cmd = [options.pkg_config, "--cflags", "--libs"] + args
```
andy31415, line 211:
```suggestion
    cmd = [options.pkg_config, "--cflags", "--libs"]

    if options.static:
        cmd.append("--static")
        
    cmd.extend(args)
```

seems a bit less repeating and may extend more easily later if we add more flags.

build/config/linux/pkg_config.gni
```diff
@@ -137,10 +137,16 @@ template("pkg_config") {
 
       # Link libraries statically for OSS-Fuzz fuzzer build
       if (oss_fuzz) {
+        #outputs libs needed for static linking (direct + transitive/non-direct libs), we will re-execute the script to get those libs
```
andy31415, line 140:
```suggestion
        # Output libs needed for static linking (direct + transitive/non-direct libs), we will re-execute the script to get those libs
```

build/config/linux/pkg_config.gni
```diff
@@ -137,10 +137,16 @@ template("pkg_config") {
...
+        args += [ "--static" ]
+        pkgresult_static = exec_script(pkg_config_script, args, "value")
         libs = []
         ldflags = [ "-Wl,-Bstatic" ]
-        foreach(lib, pkgresult[3]) {
-          ldflags += [ "-l$lib" ]
+        foreach(lib, pkgresult_static[3]) {
+          #dl(dynamic loading) lib is not needed for linking statically and its presence triggers errors.
```
andy31415, line 146:
```suggestion
          # dl(dynamic loading) lib is not needed for linking statically and its presence triggers errors.
```

andy31415: what errors? Is there a reference link we can add for this?

andy31415 (approved):

build/config/linux/pkg_config.gni
Alami-Amine: Not Really, I saw those errors locally.
Here is an example after keeping `-ldl` in the command
```
[765/811] ld tests/fuzz-setup-payload-base38-decode
FAILED: tests/fuzz-setup-payload-base38-decode 
clang++ -O0 -fPIC -Werror -Wl,--fatal-warnings -Wl,-z,defs -fdiagnostics-color -O1 -fno-omit-frame-pointer -gline-tables-only -Wno-error=enum-constexpr-conversion -Wno-error=incompatible-function-pointer-types -Wno-error=int-conversion -Wno-error=deprecated-declarations -Wno-error=implicit-function-declaration -Wno-error=implicit-int -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=memory -fsanitize-memory-track-origins -fsanitize=fuzzer-no-link -stdlib=libc++ -fsanitize=fuzzer -fuse-ld=lld -Wl,--gc-sections -pie -O1 -fno-omit-frame-pointer -gline-tables-only -Wno-error=enum-constexpr-conversion -Wno-error=incompatible-function-pointer-types -Wno-error=int-conversion -Wno-error=deprecated-declarations -Wno-error=implicit-function-declaration -Wno-error=implicit-int -DFUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION -fsanitize=memory -fsanitize-memory-track-origins -fsanitize=fuzzer-no-link -stdlib=libc++ -fsanitize=fuzzer -Wl,-Bstatic -lssl -lcrypto -ldl -Wl,-Bdynamic @tests/fuzz-setup-payload-base38-decode.rsp -o tests/fuzz-setup-payload-base38-decode
ld.lld: error: undefined symbol: __dlsym
>>> referenced by dlsym.o:(dlsym) in archive /lib/x86_64-linux-gnu/libdl.a

ld.lld: error: undefined symbol: __dlvsym
>>> referenced by dlvsym.o:(dlvsym) in archive /lib/x86_64-linux-gnu/libdl.a
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

andy31415: I would add the text like `ld.lld: error: undefined symbol: __dlvsym` to the comment (or maybe even more line out of it) so that it is easily greppable if others hit it.

cecille (approved):

jmartinez-silabs (approved):

PR #35525, opened 2024-09-10
TC-CC-3.2: Change timing so range tolerance doesn't span wrap

cecille:
This lets us omit the manual step. Manual steps in the middle of a timed test will never work.

Also fix the tolerance calculation.

test plan: https://github.com/CHIP-Specifications/chip-test-plans/pull/4662

### DIFF
--- a/src/app/tests/suites/certification/Test_TC_CC_3_2.yaml
+++ b/src/app/tests/suites/certification/Test_TC_CC_3_2.yaml
@@ -86,58 +86,45 @@ tests:
               - name: "OptionsOverride"
                 value: 0
 
-    - label: "Wait 10s"
+    - label: "Wait 5s"
       PICS: CC.S.F00
       cluster: "DelayCommands"
       command: "WaitForMs"
       arguments:
           values:
               - name: "ms"
-                value: 10000
+                value: 5000
 
-    #Tolerance to be computed from the range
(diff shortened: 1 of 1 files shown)


leorozendaal (approved):

src/app/tests/suites/certification/Test_TC_CC_3_2.yaml
```diff
@@ -86,58 +86,45 @@ tests:
...
       command: "WaitForMs"
       arguments:
           values:
               - name: "ms"
-                value: 10000
+                value: 15000
 
-    #Tolerance to be computed from the range
-    #Start value 200, expected value 45, 200 - 45 = 155, 155 % 15 = 23 , So 45+/-23
+    # Tolerance to be computed from the range, hue wraps back to 0 the continues
```
bzbarsky-apple, line 118: What does "the continues" mean?

bzbarsky-apple (approved):

src/app/tests/suites/certification/Test_TC_CC_3_2.yaml
leorozendaal, line 118:
```suggestion
    # Tolerance to be computed from the range, hue reaches max, wraps back to 0 then continues from min value
```
@bzbarsky-apple  assume this is what is intended

cecille: addressed by leo below

andy31415 (approved):

PR #35526, opened 2024-09-10
Fix deadlock in MTRDeviceController suspend/resume.

bzbarsky-apple:
If a resume happened before the async work from a suspend finished running on the Matter queue, we could deadlock: the Matter queue would be holding the device lock and try to acquire the lock to read "suspended" from the controller, while the controller would be holding the "suspended" lock and trying to acquire the device lock as part of synchronous resume work.

The fix is to:

1) Stop trying to read the suspended state of the controller in MTRDevice_Concrete except during MTRDevice_Concrete initialization.  Instead, store the suspended state directly in MTRDevice_Concrete at init time and update it when the controller notifies about suspend/resume.

2) Fix controller suspend/resume to guarantee that we create the device list snapshot and set our suspended state in one atomic operation, so that devices are either created before that op (and then end up in the list) or after it (and then pick up the new controller suspended state.

3) Remove the locking around controller suspend/resume, because it's proving hard to reason about.  This will be replaced by something better once we figure out how threadsafety should work in controller API in general.  In the meantime, suspend/resume on a given controller should only be done on a single serial queue.

4) Use an atomic for the suspended state, so reading it can be done from any thread/queue even without locking.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDeviceController.mm
+++ b/src/darwin/Framework/CHIP/MTRDeviceController.mm
@@ -154,7 +154,11 @@ using namespace chip::Tracing::DarwinFramework;
     MTRP256KeypairBridge _signingKeypairBridge;
     MTRP256KeypairBridge _operationalKeypairBridge;
 
-    BOOL _suspended;
+    // For now, we just ensure that access to _suspended is atomic, but don't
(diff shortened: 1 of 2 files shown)


jtung-apple (approved):

bzbarsky-apple:

PR #35527, opened 2024-09-10
Add a getter to AttributeValueDecoder for raw data.

yyzhong-g:
Add a getter to AttributeValueDecoder which returns the reader from the decoder.

This is useful to get the copy of the raw data and pass it around (in multi-threaded case) instead of the decoder. Also the decoder is marked as mTriedDecoded after the call.

### DIFF
--- a/src/app/AttributeValueDecoder.h
+++ b/src/app/AttributeValueDecoder.h
@@ -55,7 +55,8 @@ public:
     /**
      * Get the reader from the decoder for raw data. Calling this is considered to be tried decoding.
      */
-    TLV::TLVReader & GetReaderForDecoding() {
+    TLV::TLVReader & GetReaderForDecoding()
+    {
         mTriedDecode = true;
         return mReader;
     }
--- a/src/app/AttributeValueDecoder.h
+++ b/src/app/AttributeValueDecoder.h
@@ -52,6 +52,14 @@ public:
         return CHIP_NO_ERROR;
     }
 
+    /**
+     * Get the reader from the decoder for raw data. Calling this is considered to be tried decoding.
+     */
+    TLV::TLVReader & GetReaderForDecoding() {
+        mTriedDecode = true;
+        return mReader;
+    }
+
     bool TriedDecode() const { return mTriedDecode; }
 
     /**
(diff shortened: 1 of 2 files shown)


src/app/AttributeValueDecoder.h
```diff
@@ -52,6 +52,15 @@ class AttributeValueDecoder
         return CHIP_NO_ERROR;
     }
 
+    /**
+     * Get the reader from the decoder for raw data. Calling this is considered to be tried decoding.
```
bzbarsky-apple, line 56: We pretty explicitly did not want to hand out the reader from the decoder, because it's too easy to misuse.

Why is this actually needed?  The "multi-threaded case" comment from the PR description sounds like "misuse"...

bzbarsky-apple (changes requested):

src/app/AttributeValueDecoder.h
yyzhong-g: If so, what do you think about returning a ByteSpan? We intended to copy the data to a different thread, not intended to modify the data.

andy31415: The use case seems to be "I want the data, but don't actually know what it contains and want an opaque blob" ... is this true?

Would you be able to implement that by creating a class for which DataModel::Decode exists and creates whatever is necessary? Decode does receive the TLV object.

yyzhong-g: Thank you. That is the use case that I need. I will try it.

bzbarsky-apple: Ah, yes, if you're going to snapshot the data then it should be fine.  This sounds like PreEncodedValue but in the other direction...

(That said, the TLV reader on the other thread which then tries to read the data will need to be handled very carefully.... and for some attributes you need a bunch of state other than the data to decode properly.)

andy31415:
@yyzhong-g was the workaround ok? Can we close this PR?

yyzhong-g:
Not needed anymore. Andrei's suggestion ```Would you be able to implement that by creating a class for which DataModel::Decode exists and creates whatever is necessary? Decode does receive the TLV object.``` is sufficient.

PR #35531, opened 2024-09-11
Remove the accessedViaPublicAPI on MTRDevice.

bzbarsky-apple:
Now that we have a concept of suspended devices, just use that to avoid wasting time on read-throughs which would fail anyway.


ksperling-apple (approved):

bzbarsky-apple:

PR #35535, opened 2024-09-11
Switch infineon builds to latest CI image

andy31415:
This switches the builds to a ubuntu-24 based one.

### DIFF
--- a/.github/workflows/examples-infineon.yaml
+++ b/.github/workflows/examples-infineon.yaml
@@ -38,6 +38,9 @@ jobs:
 
         container:
             image: ghcr.io/project-chip/chip-build-infineon:74
+            env:
+               # TODO: this should probably be part of the dockerfile itself
+               CY_TOOLS_PATHS: /opt/Tools/ModusToolbox/tools_3.2
             volumes:
                 - "/tmp/bloat_reports:/tmp/bloat_reports"
         steps:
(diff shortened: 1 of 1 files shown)


saurabhst (approved):

.github/workflows/examples-infineon.yaml
```diff
@@ -37,7 +37,10 @@ jobs:
         if: github.actor != 'restyled-io[bot]'
 
         container:
-            image: ghcr.io/project-chip/chip-build-infineon:54
+            image: ghcr.io/project-chip/chip-build-infineon:74
+            env:
+               # TODO: this should probably be part of the dockerfile itself
```
Alami-Amine, line 42: should we have an Issue for this?

Alami-Amine (approved):

kiel-apple (approved):

.github/workflows/examples-infineon.yaml
andy31415: I notified Infineon. Given that we currently have a bit over 1600 open SDK issues I am not sure if individual issues actually provide value, so I opted just for todo and slack communication for now.

I believe issues are only valuable today if they are part of a project or tracked in some way (e.g. tags). I have neither of them for this work.

Alami-Amine: sure sounds good.

PR #35537, opened 2024-09-11
Moving these to async, as the completions already work when needed

woody-apple:
These are blocking startup, moving commands to be explicity oneway void and async

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
@@ -1852,15 +1852,8 @@ typedef NS_ENUM(NSUInteger, MTRDeviceWorkItemDuplicateTypeID) {
     [self _reportAttributes:[self _getAttributesToReportWithReportedValues:attributeReport fromSubscription:isFromSubscription]];
 }
 
-#ifdef DEBUG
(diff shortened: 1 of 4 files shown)


kiel-apple (approved):

src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
```diff
@@ -1852,24 +1852,38 @@ - (void)_handleAttributeReport:(NSArray<NSDictionary<NSString *, id> *> *)attrib
...
+        dispatch_async(self.queue, ^{
+            [self _handleAttributeReport:attributeReport fromSubscription:isFromSubscription];
+            [self _handleReportEnd];
+        });
+    } errorHandler:nil];
+}
+
+- (void)_injectEventReport:(NSArray<NSDictionary<NSString *, id> *> *)eventReport
+{
+    //    [_deviceController asyncDispatchToMatterQueue:^{ // TODO: This wasn't used previously, not sure why, so keeping it here for thought, but preserving existing behavior
```
bzbarsky-apple, line 1869: Because the part that needs to happen on the Matter queue is _handleReportBegin, which the unit test event report stuff was not doing at all...

PR #35540, opened 2024-09-11
Add DataModel::Provider Invoke usage

andy31415:
### Changes

- Command handler processing has its own `AddStatus` (and variants) and may be processed asynchronously, so as a result we now have `Provider::Invoke` return a optional where missing means `status handled separately`
- Update IM to call the data model provider invoke implementation if data model provider processing is enabled
- Update unit tests


src/app/InteractionModelEngine.cpp
```diff
@@ -1717,6 +1733,7 @@ void InteractionModelEngine::DispatchCommand(CommandHandlerImpl & apCommandObj,
     }
 
     DispatchSingleClusterCommand(aCommandPath, apPayload, &apCommandObj);
+#endif
```
tehampson, line 1736: nit,
```suggestion
#endif // CHIP_CONFIG_USE_DATA_MODEL_INTERFACE
```

src/app/data-model-provider/Provider.h
```diff
@@ -98,8 +100,9 @@ class Provider : public ProviderMetadataTree
     ///    - data (as CommandDataIB) which is assumed a "response as a success"
     ///    - status (as a CommandStatusIB) which is considered a final status, usually an error however
     ///      cluster-specific success statuses also exist.
-    virtual ActionReturnStatus Invoke(const InvokeRequest & request, chip::TLV::TLVReader & input_arguments,
-                                      CommandHandler * handler) = 0;
+    ///
+    virtual std::optional<ActionReturnStatus> Invoke(const InvokeRequest & request, chip::TLV::TLVReader & input_arguments,
```
tehampson, line 104: nit, can you update documentation? Please mention that if there is an error return that the handler has not sent anything in response to the command

src/app/InteractionModelEngine.cpp
```diff
@@ -1699,6 +1703,18 @@ CHIP_ERROR InteractionModelEngine::PushFront(SingleLinkedListNode<T> *& aObjectL
...
                                              TLV::TLVReader & apPayload)
 {
+#if CHIP_CONFIG_USE_DATA_MODEL_INTERFACE
+
+    DataModel::InvokeRequest request;
+    request.path = aCommandPath;
+
+    std::optional<DataModel::ActionReturnStatus> status = GetDataModelProvider()->Invoke(request, apPayload, &apCommandObj);
+
+    if (status.has_value())
```
tehampson, line 1713: What is value() is CHIP_NO_ERROR (the equivalent that means OK for ActionReturnStatus) ?

lpbeliveau-silabs (approved):

src/app/InteractionModelEngine.cpp
andy31415: Updated documentation some more. std::nullopt is distinctly different from CHIP_NO_ERROR. Added notes both here and in the Provider.h documentation.

src/app/data-model-provider/Provider.h
```diff
@@ -88,18 +88,25 @@ class Provider : public ProviderMetadataTree
...
     /// `handler` is used to send back the reply.
+    ///    - returning `std::nullopt` means that return value was placed in handler directly.
+    ///      This includes cases where command handling and value return will be done asynchronously.
     ///    - returning a value other than Success implies an error reply (error and data are mutually exclusive)
     ///
-    /// Returning anything other than CHIP_NO_ERROR or Status::Success (i.e. success without a return code)
-    /// means that the invoke will be considered to be returning the given path-specific status WITHOUT any data (any data
-    /// that was sent via CommandHandler is to be rolled back/discarded).
+    /// Return value expectations:
+    ///   - if a response has been placed into a `handler` then std::nullopt MUST be returned. In particular
```
bzbarsky-apple, line 96:
```suggestion
    ///   - if a response has been placed into `handler` then std::nullopt MUST be returned. In particular
```

src/app/data-model-provider/Provider.h
```diff
@@ -88,18 +88,25 @@ class Provider : public ProviderMetadataTree
...
+    ///     note that CHIP_NO_ERROR is NOT the same as std::nullopt:
+    ///        > CHIP_NO_ERROR means handler had no status set and we caller to AddStatus(success)
```
bzbarsky-apple, line 98:
```suggestion
    ///        > CHIP_NO_ERROR means handler had no status set and we expect the caller to AddStatus(success)
```

src/app/data-model-provider/Provider.h
```diff
@@ -88,18 +88,25 @@ class Provider : public ProviderMetadataTree
...
+    ///     convenience to make writing Invoke calls easier.
     ///
-    /// This is because only one of the following may be encoded in a response:
-    ///    - data (as CommandDataIB) which is assumed a "response as a success"
-    ///    - status (as a CommandStatusIB) which is considered a final status, usually an error however
-    ///      cluster-specific success statuses also exist.
-    virtual ActionReturnStatus Invoke(const InvokeRequest & request, chip::TLV::TLVReader & input_arguments,
-                                      CommandHandler * handler) = 0;
+    /// Expectations if invoke returns data:
+    ///   - if a command data is to be returned, the status can only be 'Success' (it cannot be cluster
```
bzbarsky-apple, line 105: This documentation does not make sense.  If command data is to be returned, doesn't the status have to be nullopt?

andy31415: I deleted that part - it used to be there as a note that `cluster specific success` is incompatible with `data` (i.e. data has bare success implied). However this is rexplained now above about return data expectations.

tehampson (approved):

PR #35541, opened 2024-09-11
Stop persistent operational browse when all controllers are suspended.

bzbarsky-apple:


src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
```diff
@@ -1246,6 +1260,42 @@ - (PersistentStorageDelegate *)storageDelegate
...
+    auto * controllers = [self getRunningControllers];
+    for (MTRDeviceController * controller in controllers) {
+        if (!controller.suspended) {
+            return NO;
+        }
+    }
+    return YES;
+}
+
+- (void)controller:(MTRDeviceController *)controller
```
ksperling-apple, line 1276: Seems like this was added in a different PR, but this reads like the delegate is being queried whether a particular controller is suspended. Something like `controller:didChangeSuspensionState:` would be more obvious (or even two separate `controllerDidSuspend:` and `controllerDidResume:` methods).

src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
```diff
@@ -1246,6 +1260,42 @@ - (PersistentStorageDelegate *)storageDelegate
...
+       isSuspended:(BOOL)suspended
+{
+    // We use the Matter queue as our delegate queue for this notification.
+    assertChipStackLockedByCurrentThread();
+
+    if (!suspended) {
+        // As long as we have one controller that is not suspended, we should
+        // have a running operational browse. Guard against someone resuming a
```
ksperling-apple, line 1284: Shouldn't resuming a controller that's shut down fail to begin with, or at least not emit events to the delegate?

src/darwin/Framework/CHIP/MTROperationalBrowser.mm
```diff
@@ -107,7 +117,10 @@
     }
 
     ChipLogProgress(Controller, "Notifying controller factory about new operational instance: '%s'", aName);
-    [self->mDeviceControllerFactory operationalInstanceAdded:peerId];
+    MTRDeviceControllerFactory * strongFactory = self->mDeviceControllerFactory;
```
ksperling-apple, line 120: The old code did exactly the same thing, calling the message on a nil mDeviceControllerFactory is a no-op anyway.

src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
```diff
@@ -557,9 +559,18 @@ - (MTRDeviceController * _Nullable)_startDeviceController:(MTRDeviceController *
         return nil;
     }
 
-    if ([_controllers count] == 0) {
-        dispatch_sync(_chipWorkQueue, ^{
-            self->_operationalBrowser = new MTROperationalBrowser(self, self->_chipWorkQueue);
+    // Listen for changes to suspended state on this controller, so we can
+    // track whether we have any unsuspended controllers.
+    [controller addDeviceControllerDelegate:self queue:_chipWorkQueue];
```
ksperling-apple, line 564: I guess this works, but it does feel slightly odd to have the factory add itself as a generic delegate here to implement the operational browse behavior. I think it would be maybe more obvious if the MTRDeviceController itself was simply directly responsible for calling counted Start and Stop methods on the browser as appropriate when it starts/stop/suspends/resumes.

ksperling-apple (approved):

src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
bzbarsky-apple: Understood.  I'll see about changing the names in a followup; there's coupling here with some other code, unfortunately.

src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
bzbarsky-apple: Ideally yes, but right now it does not.  Once we fix it to do that, we can remove this part.

src/darwin/Framework/CHIP/MTROperationalBrowser.mm
bzbarsky-apple: Will undo this change.

src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
bzbarsky-apple: I considered having the controller notify the controller factory directly via some sort of out-of-band API, but just reusing the existing delegate API seemed simpler.

But yes, we could expose the browser to controllers and have them call methods on it directly.  Maybe that will be simpler; I'll code it up and see how it looks.

ksperling-apple: exposing the browser to controllers makes sense to me

jtung-apple (approved):

ksperling-apple (approved):

bzbarsky-apple:

src/darwin/Framework/CHIP/MTRDeviceControllerFactory.mm
bzbarsky-apple: https://github.com/project-chip/connectedhomeip/pull/35577

PR #35542, opened 2024-09-11
cleanup cmake data model build rules / remove ARG_INCLUDE_SERVER

andy31415:
The data model seems to do a lot more than just data mode as it includes more source files that are generally defaults within apps. For now make things more consistent:

### Changes

- remove (default to true) ARG_INCLUDE_SERVER as virtually every single CMakeLists.txt would set this and the places where it is not used seems to be some mbed and ameba where it likely will be without sideffects (extra files built only). 
- Separate out the ICD and cluster-object dependencies as they are separate in GN. Ideally we should separate out cmake targets/gn targets from code-generator dependent targets.

### DIFF
--- a/config/openiotsdk/cmake/chip.cmake
+++ b/config/openiotsdk/cmake/chip.cmake
@@ -74,7 +74,6 @@ function(chip_add_data_model target scope model_name)
     include(${CHIP_ROOT}/src/app/chip_data_model.cmake)
     chip_configure_data_model(${target}
         SCOPE ${scope}
-        INCLUDE_SERVER
         ZAP_FILE ${CMAKE_CURRENT_SOURCE_DIR}/../${model_name}-common/${model_name}-app.zap
     )
 endfunction()
--- a/src/app/chip_data_model.cmake
+++ b/src/app/chip_data_model.cmake
@@ -119,6 +119,27 @@ function(chip_configure_data_model APP_TARGET)
         set(APP_GEN_FILES)
     endif()
 
+    # These are:
+    #   //src/app/icd/server:notfier
+    #   //src/app/icd/server:monitoring-table
+    #   //src/app/icd/server:configuration-data
+    #
+    # TODO: ideally we would avoid duplication and would link gn-built items. In this case
+    #       it may be slightly harder because these are source_sets rather than libraries.
+    target_sources(${APP_TARGET} ${SCOPE}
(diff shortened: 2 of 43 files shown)


tehampson (approved):

src/app/chip_data_model.cmake
```diff
@@ -119,6 +117,27 @@ function(chip_configure_data_model APP_TARGET)
         set(APP_GEN_FILES)
     endif()
 
+    # These are:
+    #   //src/app/icd/server:notfier
+    #   //src/app/icd/server:monitoring-table
+    #   //src/app/icd/server:configuration-data
+    #
+    # TODO: ideally we would avoid duplication and would link gn-built items. In this case
```
Damian-Nordic, line 125: I wonder if these are already not included in libCHIP.a. `libCHIP.a` already includes `src/app` as its public_deps, and `src/app` pulls ICD bits if `chip_enable_icd_server`. But I may be wrong as there are so many dependencies between the source sets that it's hard to statically analyse all the connections ;).

Damian-Nordic (approved):

config/openiotsdk/cmake/chip.cmake
```diff
@@ -74,7 +74,6 @@ function(chip_add_data_model target scope model_name)
     include(${CHIP_ROOT}/src/app/chip_data_model.cmake)
     chip_configure_data_model(${target}
         SCOPE ${scope}
-        INCLUDE_SERVER
```
Damian-Nordic, line 77: Hmm. This argument was meant to resemble `is_server` argument from `chip_data_model.gni`. Removing it is fine as long as there are no out-of-tree controllers built with CMake (I am not aware of any but thought it's worth pointing it out).

src/app/chip_data_model.cmake
andy31415: I think we may want to start removing some bits here for sure. I just kept what was already there.

I would try to clean it up as another followup.

config/openiotsdk/cmake/chip.cmake
andy31415: Thinking that this is not used currently based on examples after several iterations.
If needed, we could revert however if the code path is never tested in any examples (even if only compilation) I have no confidence in us maintaining it.

Willing to add it back (maybe with switched defaults) if someone using cmake is requiring this.

PR #35543, opened 2024-09-11
Move StoreCurrentUpdateInfo() call from ApplyUpdate to DownloadUpdate

NicholasFay-IFX:
Update to DefaultOTARequestor.cpp where the API call to StoreCurrentUpdateInfo() is moved from ApplyUpdate() to DownloadUpdate().


andy31415 (changes requested):
@NicholasFay-IFX  Please update the summary to explain the chage: why was this change made, what was the old (and presumably incorrect) behavior and what is the new behavior and why is it better.

Currently the summary just re-states the patch.

PR #35545, opened 2024-09-11
Update Chef Dockerfile to use trixie for latest ssl and lib versions

aBozowski:
These updates are required to run the dockerized versions of chef examples.


andy31415 (approved):

andy31415:
Merging this as CI is unable to verify it.

Issue #31899, opened 2024-02-04
[1.2] Failed to resolve service '75E232F7505D0C92': 3


shuiYe-704265085:
### Reproduction steps

When I tried to discover devices through the deviceController.discoverCommissionableNodes() function, I found that only one device could be discovered at a time. When multiple devices could be discovered simultaneously, the console log displayed as follows

```
2024-02-04 15:10:28.635 23920-23920/com.akuvox.mattersdkdemo D/NsdManagerServiceBrowser: Starting service discovering for '_matterc._udp'
2024-02-04 15:10:28.642 23920-24850/com.akuvox.mattersdkdemo I/NsdManagerServiceBrowser: Started service '_matterc._udp'
2024-02-04 15:10:28.642 23920-24850/com.akuvox.mattersdkdemo I/NsdManagerServiceBrowser: Found service '93962EB963ADF462'
2024-02-04 15:10:28.642 23920-24850/com.akuvox.mattersdkdemo I/NsdManagerServiceBrowser: Found service '75E232F7505D0C92'
2024-02-04 15:10:33.641 23920-23920/com.akuvox.mattersdkdemo D/NsdManagerServiceResolver: resolve: Starting service resolution for '93962EB963ADF462' type '_matterc._udp'
2024-02-04 15:10:33.643 23920-23920/com.akuvox.mattersdkdemo D/NsdManagerServiceResolver: resolve: Starting service resolution for '75E232F7505D0C92' type '_matterc._udp'
2024-02-04 15:10:33.649 23920-24850/com.akuvox.mattersdkdemo W/NsdManagerServiceBrowser: Succeed to stop discovery service '_matterc._udp
2024-02-04 15:10:33.660 23920-24850/com.akuvox.mattersdkdemo D/NsdServiceFinderAndResolver: Service discovery started. regType: _matterc._udp
2024-02-04 15:10:33.672 23920-24850/com.akuvox.mattersdkdemo D/NsdServiceFinderAndResolver: Service discovery started. regType: _matterc._udp
2024-02-04 15:10:33.672 23920-24850/com.akuvox.mattersdkdemo D/NsdServiceFinderAndResolver: onServiceFound: found target service name: 93962EB963ADF462, type: _matterc._udp., host: null, port: 0, txtRecord: 
[...]

PR #35551, opened 2024-09-12
[Android] Fix resolve exception issue

### LINKED ISSUE #35550: [BUG][Android] Resolve fail issue due to abnormal behavior of some Android devices
### Reproduction steps

1. Try to call "getConnectedDevicePointer" method for CASE session establishment.
2. Fail connection Failed due to the reason in the log below.

CSM     : FindOrEstablishSession: PeerId = [1:534C0B956F2011EF]
CSM     : FindOrEstablishSession: No existing OperationalSessionSetup instance found
DIS     : OperationalSessionSetup[1:534C0B956F2011EF]: State change 1 --> 2
DIS     : Resolving 6E7CB9C39B4A1524:534C0B956F2011EF ...
NsdManagerServiceResolver: resolve: Starting service resolution for '6E7CB9C39B4A1524-534C0B956F2011EF' type '_matter._tcp'
NsdManagerResolverAvailState: Found NsdManager Resolver free, using it and marking it as busy
DIS     : Checking node lookup status after 201 ms
NsdServiceFinderAndResolver: Service discovery started. regType: _matter._tcp
**NsdServiceFinderAndResolver: onServiceFound: found target service name: 6E7CB9C39B4A1524-534C0B956F2011EF, type: _matter._tcp., host: null, port: 0, txtRecord: 
NsdServiceFinderAndResolver: onServiceFound: found target service name: 6E7CB9C39B4A1524-534C0B956F2011EF, type: _matter._tcp., host: null, port: 0, txtRecord:** 
NsdServiceFinderAndResolver: Discovery stopped: _matter._tcp
NsdServiceFinderAndResolver: Failed to resolve service '6E7CB9C39B4A1524-534C0B956F2011EF': 3
NsdManagerResolverAvailState: Signaling NsdManager Resolver as free
**NsdServiceFinderAndResolver: Failed to resolve service '6E7CB9C39B4A1524-534C0B956F2011EF': 3**
DIS     : Checking node lookup status after 45024 ms
E DIS     : OperationalSessionSetup[1:534C0B956F2011EF]: operational discovery failed: src/lib/address_resolve/AddressResolve_DefaultImpl.cpp:119: CHIP Error 0x00000032: Timeout

### Bug prevalence

1-2 times a week

ac289edd6050de2f1646d839f07e055eb2848c8e

### Platform

android

### Platform Version(s)

_No response_

### Anything else?

_No response_

joonhaengHeo:
Fix #35550 
If two instances of the same instanceName are called at the same time, the resolve method will be called twice. In this case, resolve will fail for similar reasons to the previously reported issue. (#31899 - Fix PR is #32459)

To prevent this, the first searched item is saved in a separate variable, and resolveService is called after discover is stopped.

### DIFF
--- a/src/platform/android/java/chip/platform/NsdServiceFinderAndResolver.java
+++ b/src/platform/android/java/chip/platform/NsdServiceFinderAndResolver.java
@@ -41,6 +41,7 @@ class NsdServiceFinderAndResolver implements NsdManager.DiscoveryListener {
   private final ChipMdnsCallback chipMdnsCallback;
   private final MulticastLock multicastLock;
   private final ScheduledFuture<?> resolveTimeoutExecutor;
+  private NsdServiceInfo discoveredServiceInfo = null;
 
   @Nullable
   private final NsdManagerServiceResolver.NsdManagerResolverAvailState nsdManagerResolverAvailState;
@@ -94,14 +95,13 @@ class NsdServiceFinderAndResolver implements NsdManager.DiscoveryListener {
 
   @Override
   public void onServiceFound(NsdServiceInfo service) {
-    if (targetServiceInfo.getServiceName().equals(service.getServiceName())) {
+    if (discoveredServiceInfo == null && targetServiceInfo.getServiceName().equals(service.getServiceName())) {
       Log.d(TAG, "onServiceFound: found target service " + service);
 
       if (stopDiscoveryRunnable.cancel(false)) {
         nsdManager.stopServiceDiscovery(this);
       }
-
-      resolveService(service, callbackHandle, contextHandle, chipMdnsCallback);
+      discoveredServiceInfo = service;
     } else {
       Log.d(TAG, "onServiceFound: found service not a target for resolution, ignoring " + service);
     }
@@ -206,7 +206,11 @@ class NsdServiceFinderAndResolver implements NsdManager.DiscoveryListener {
 
   @Override
   public void onDiscoveryStopped(String serviceType) {
-    Log.i(TAG, "Discovery stopped: " + serviceType);
+    Log.i(TAG, "Discovery stopped: " + serviceType + ", discoveredServiceInfo: " + discoveredServiceInfo);
+    if (discoveredServiceInfo != null) {
+      resolveService(discoveredServiceInfo, callbackHandle, contextHandle, chipMdnsCallback);
+      discoveredServiceInfo = null;
+    }
   }
 
   @Override
--- a/src/platform/android/java/chip/platform/NsdServiceFinderAndResolver.java
+++ b/src/platform/android/java/chip/platform/NsdServiceFinderAndResolver.java
@@ -95,7 +95,8 @@ class NsdServiceFinderAndResolver implements NsdManager.DiscoveryListener {
 
   @Override
   public void onServiceFound(NsdServiceInfo service) {
-    if (discoveredServiceInfo == null && targetServiceInfo.getServiceName().equals(service.getServiceName())) {
+    if (discoveredServiceInfo == null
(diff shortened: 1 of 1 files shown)


src/platform/android/java/chip/platform/NsdServiceFinderAndResolver.java
```diff
@@ -94,14 +95,14 @@ public void run() {
...
       Log.d(TAG, "onServiceFound: found target service " + service);
 
       if (stopDiscoveryRunnable.cancel(false)) {
         nsdManager.stopServiceDiscovery(this);
       }
-
-      resolveService(service, callbackHandle, contextHandle, chipMdnsCallback);
+      discoveredServiceInfo = service;
     } else {
       Log.d(TAG, "onServiceFound: found service not a target for resolution, ignoring " + service);
```
andy31415, line 107: this log is not correct anymore. we should report duplicates.

src/platform/android/java/chip/platform/NsdServiceFinderAndResolver.java
```diff
@@ -206,7 +207,13 @@ public void onServiceLost(NsdServiceInfo service) {
 
   @Override
   public void onDiscoveryStopped(String serviceType) {
-    Log.i(TAG, "Discovery stopped: " + serviceType);
+    Log.i(
+        TAG,
+        "Discovery stopped: " + serviceType + ", discoveredServiceInfo: " + discoveredServiceInfo);
+    if (discoveredServiceInfo != null) {
+      resolveService(discoveredServiceInfo, callbackHandle, contextHandle, chipMdnsCallback);
```
andy31415, line 214: this changes behavior: previously we would call this as soon as we would receive the entry, now we only call this on stopped. What is the lifetime here? I would generally expect long(ish) running resolve followed by a stop, so if that is the case this introduces significant latency.

Can we keep the resolve call inside the onServiceFound and just call on the first value received? or is it mandatory to have the last value found reported?

joonhaengHeo: 1. As soon as the service is found using discoverService, the stop command is called with the code below, and when the stop is completed, the corresponding method is called.
https://github.com/project-chip/connectedhomeip/blob/7a54490fc3e37a67692073922ec7ee6f15c48473/src/platform/android/java/chip/platform/NsdServiceFinderAndResolver.java#L101

-> I think resolveService works when discoverService is completely terminated I decided that it was right to do so, so I modified it as follows. Do you think it would be more appropriate to call it directly?

2. The modified code also calls only by utilizing the first Service found.

src/platform/android/java/chip/platform/NsdServiceFinderAndResolver.java
joonhaengHeo: After the discussion below is completed, we will revise it together according to the revised contents.

src/platform/android/java/chip/platform/NsdServiceFinderAndResolver.java
joonhaengHeo: @andy31415 please check my opinion.

joonhaengHeo: @andy31415 Please check about this.

andy31415: Ok. A comment would be useful explaining that only a single result is ever returned (which I am unclear if it is ideal: dnssd may be split under multiple packets and you may miss some IP addresses).

As long as this passes your tests, we can merge.

andy31415 (approved):

yunhanw-google (approved):

PR #35552, opened 2024-09-12
[Darwin] Device controller should declare conformance to MTRDeviceControllerDelegate

jtung-apple:
The proper way to declare a conformance to a protocol is at the interface or class extension. Given `MTRDeviceController_Concrete` sets itself as the delegate for the `MTRDeviceControllerDelegateBridge`, the conformance should be declared in the header.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDeviceController_Concrete.h
+++ b/src/darwin/Framework/CHIP/MTRDeviceController_Concrete.h
@@ -22,6 +22,8 @@
 #import <Matter/MTRDeviceController.h>
 #import <Matter/MTROperationalCertificateIssuer.h>
 
+#import "MTRDeviceControllerDelegate.h"
+
 @class MTRBaseDevice;
 @class MTRServerEndpoint; // Defined in MTRServerEndpoint.h, which imports MTRAccessGrant.h, which imports MTRBaseClusters.h, which imports this file, so we can't import it.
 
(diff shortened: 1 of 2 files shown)


kiel-apple (approved):

src/darwin/Framework/CHIP/MTRDeviceController_Concrete.h
```diff
@@ -36,9 +38,8 @@ typedef void (^MTRDeviceConnectionCallback)(MTRBaseDevice * _Nullable device, NS
 @class MTRCommissionableBrowserResult;
 @class MTRSetupPayload;
 @protocol MTRDevicePairingDelegate;
-@protocol MTRDeviceControllerDelegate;
 
-@interface MTRDeviceController_Concrete : MTRDeviceController
+@interface MTRDeviceController_Concrete : MTRDeviceController <MTRDeviceControllerDelegate>
```
woody-apple, line 42: I'm probably missing something, but why is the controller a delegate of itself? (This seems odd... maybe it's late)

bzbarsky-apple: Because we pass it as the MTRDeviceControllerDelegate to the SDK bridge.  The controller then just fans out the notifications to the actual delegates.

bzbarsky-apple (approved):

PR #35556, opened 2024-09-12
Fix the commandFields type on MTRDevice invoke methods.

bzbarsky-apple:
The "id" was a copy/paste error.  For the public API we can't do much about that, but internally we can sanity-check and use the right type.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -1313,6 +1313,21 @@ using namespace chip::Tracing::DarwinFramework;
     // here for now.
     // TODO: https://github.com/project-chip/connectedhomeip/issues/24563
 
+    if (![commandFields isKindOfClass:NSDictionary.class]) {
+        MTR_LOG_ERROR("%@ invokeCommandWithEndpointID passed a commandFields (%@) that is not a data-value NSDictionary object",
+            self, commandFields);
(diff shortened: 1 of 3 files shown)


jtung-apple (approved):

anush-apple (approved):

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -1313,6 +1313,21 @@ - (void)invokeCommandWithEndpointID:(NSNumber *)endpointID
     // here for now.
     // TODO: https://github.com/project-chip/connectedhomeip/issues/24563
 
+    if (![commandFields isKindOfClass:NSDictionary.class]) {
```
anush-apple, line 1316: Will this change break any clients that might have set this to nil (even though it says it not nullable)

bzbarsky-apple: Yes, it will, but I would have expected those clients to fail to compile....  That said, we could do the same thing the no-timedInvokeTimeout overload does to handle nil here.

bzbarsky-apple: Actually, no, this never worked: down lower when we try to convert the arg to TLV, we end up with an error if it's nil.  So now we're ending up with exactly the same error, but a bit earlier up the callstack.

bzbarsky-apple:

PR #35557, opened 2024-09-12
[Python] Process attribute cache updates in Python thread

agners:
Instead of processing the attribute update in the SDK thread, process them on request in the Python thread. This avoids acks being sent back too late to the device  after the last DataReport if there are many attribute updates sent at once.

Currently still the same data model and processing is done. There is certainly also room for optimization to make this more efficient.

This also removes some obsolete/unused functions and cleans up the Read function argument list.

### DIFF
--- a/src/controller/python/chip/clusters/Attribute.py
+++ b/src/controller/python/chip/clusters/Attribute.py
@@ -314,14 +314,17 @@ class AttributeCache:
     returnClusterObject: bool = False
     attributeTLVCache: Dict[int, Dict[int, Dict[int, bytes]]] = field(
         default_factory=lambda: {})
-    attributeCache: Dict[int, List[Cluster]] = field(
-        default_factory=lambda: {})
     versionList: Dict[int, Dict[int, Dict[int, int]]] = field(
+        default_factory=lambda: {})
+
+    _attributeCacheUpdateNeeded: set[AttributePath] = field(
+        default_factory=lambda: set())
+    _attributeCache: Dict[int, List[Cluster]] = field(
         default_factory=lambda: {})
 
     def UpdateTLV(self, path: AttributePath, dataVersion: int,  data: Union[bytes, ValueDecodeFailure]):
         ''' Store data in TLV since that makes it easiest to eventually convert to either the
-            cluster or attribute view representations (see below in UpdateCachedData).
+            cluster or attribute view representations (see below in GetUpdatedAttributeCache()).
         '''
         if (path.EndpointId not in self.attributeTLVCache):
             self.attributeTLVCache[path.EndpointId] = {}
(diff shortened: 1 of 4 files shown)


src/controller/python/chip/clusters/Attribute.py
```diff
@@ -745,18 +788,21 @@ def _handleReportBegin(self):
...
             for change in self._changedPathSet:
-                try:
-                    attribute_path = TypedAttributePath(Path=change)
-                except (KeyError, ValueError) as err:
-                    # path could not be resolved into a TypedAttributePath
-                    LOGGER.exception(err)
-                    continue
-                self._subscription_handler.OnAttributeChangeCb(
-                    attribute_path, self._subscription_handler)
+                if self._subscription_handler.OnAttributeChangeCb:
```
marcelveldt, line 793: Maybe change this to a guard to make the code less nested ?
So continue in this for loop if there is no 'self._subscription_handler.OnAttributeChangeCb'

src/controller/python/chip/clusters/Attribute.py
```diff
@@ -745,18 +788,21 @@ def _handleReportBegin(self):
...
+                    try:
+                        attribute_path = TypedAttributePath(Path=change)
+                    except (KeyError, ValueError) as err:
+                        # path could not be resolved into a TypedAttributePath
+                        LOGGER.exception(err)
+                        continue
+                    self._subscription_handler.OnAttributeChangeCb(
+                        attribute_path, self._subscription_handler)
+
+                if self._subscription_handler.OnRawAttributeChangeCb:
```
marcelveldt, line 803: move this part above the block above

src/controller/python/chip/clusters/Attribute.py
agners: Yeah same here, that change came also from the raw attribute access change I've dropped from this PR again. Sorry about that.

andy31415 (approved):

src/controller/python/chip/clusters/Attribute.py
```diff
@@ -1119,25 +1134,6 @@ def Read(future: Future, eventLoop, device, devCtrl,
     return res
 
 
-def ReadAttributes(future: Future, eventLoop, device, devCtrl,
```
arkq, line 1122: There are references to these functions in:
https://github.com/project-chip/connectedhomeip/blob/master/docs/guides/repl/Matter_Basic_Interactions.ipynb?short_path=41c1c78#L3507

arkq (approved):

src/controller/python/chip/clusters/Attribute.py
agners: Hm, good catch! It seems to me that those comments should reference the non-plural variant on the device controller object. I've adjusted the comments:
https://github.com/project-chip/connectedhomeip/pull/35557/commits/420963a3742a461c3d3d5d95974ffd03f367edcc

PR #35559, opened 2024-09-12
Add ability to control data-model-interface usage for cmake-based platforms

andy31415:
CMake-based platforms are diverging from each other with at least a few flavors: zephyr, pure cmake and esp-idf. As such they require some individual updates.

I changed the following platforms:
  - esp32
  - mbed
  - nrfconnect
  - telink
  - nxp

I tested all of them to compile with/without this enabled (and the flag is disabled by default) **except** NXP which does not compile on the default vscode docker image currently. I expect the fixes to be trivial if NXP fails though.

I also updated the gn-based build for QPG since this is a small target where we can test size differences and sideffects between the interface being enabled or disabled.


config/telink/chip-module/CMakeLists.txt
```diff
@@ -77,6 +77,14 @@ if (CONFIG_CHIP_OPENTHREAD_CONFIG)
     zephyr_set_openthread_config(${CHIP_OPENTHREAD_CONFIG})
 endif()
 
+# Usage of env is because `west` seems to use a hierarcical build system
+# where I could not get cmake -D variables to propagate
```
tehampson, line 81: nit, remove "I" from this comment. Maybe something like this
```suggestion
# where propagating with cmake -D variables did not work
```

config/nxp/chip-module/CMakeLists.txt
```diff
@@ -141,6 +141,14 @@ endif()
 
 matter_generate_args_tmp_file()
 
+# Usage of env is because `west` seems to use a hierarcical build system
+# where I could not get cmake -D variables to propagate
```
tehampson, line 145: same, https://github.com/project-chip/connectedhomeip/pull/35559#discussion_r1757291093

config/nrfconnect/chip-module/CMakeLists.txt
```diff
@@ -106,6 +106,14 @@ endif()
 
 get_property(CHIP_COMPILER_LAUNCHER GLOBAL PROPERTY RULE_LAUNCH_COMPILE)
 
+# Usage of env is because `west` seems to use a hierarcical build system
+# where I could not get cmake -D variables to propagate
```
tehampson, line 110: same, https://github.com/project-chip/connectedhomeip/pull/35559#discussion_r1757291093

tehampson (approved):

config/nrfconnect/chip-module/CMakeLists.txt
```diff
@@ -106,6 +106,14 @@ endif()
 
 get_property(CHIP_COMPILER_LAUNCHER GLOBAL PROPERTY RULE_LAUNCH_COMPILE)
 
+# Usage of env is because `west` seems to use a hierarcical build system
```
Damian-Nordic, line 109: note: you can pass CMake variables using west in the following way: `west build ... -- -DSOME_CMAKE_VAR=1` (all arguments following `--` are passed to CMake).

andy31415: I tried that and it did not work (although if you have a way to make it work that would be 100% better.

What I saw is that top level and mcuboot subdirectories had it in their CMakeCache.txt however the nrfconnect subdirectory had its own CMakeCache.txt hence my comment here and ugly workaround. 

I thought it is related to sysbuild which says I should namespace it: https://docs.zephyrproject.org/latest/build/sysbuild/index.html#cmake-variable-namespacing  however I tried that and it did not work for me at that point either.

axelnxp: Shouldn't we use a Kconfig to control this feature ? This is what is usually done for all zephyr based platforms.

andy31415: I need to control build logic: included source files and flags passed to GN. It did not seem like Kconfig flags are used for that (or at least I did not seem to have them visible inside cmakelists when I tried this).

andy31415: also as an aside, I did try for nrf as a namespace all of `nrf/nrfconnect/matter/chip/zephyr` and none transferred fom `-D....=...` in the cmake command line into nrfconnect/CMakeCache (which is where I need them). As a result, the existing logic with the apology comment is the only thing that I could make work.

Hoping there is a better solution however I cannot find it and also longer term I hope we can fully transition things and eventually remove it: what I have today is to allow for incremental development and testing.

axelnxp: What I meant is that usually, we map GN args with a Kconfig entry.
As an example, you could add a new Kconfig entry in `config/zephyr/Kconfig`, something like this:
```
config CHIP_DATA_MODEL_INTERFACE
    bool "Use data model interface"
    help
      something
```

then, in zephyr based platforms CMakeLists.txt, like `nxp/chip-module/CMakeLists.txt`, you can pass the argument to gn like this:
```
matter_common_gn_args(
    DEBUG           CONFIG_DEBUG
    LIB_SHELL       CONFIG_CHIP_LIB_SHELL
    LIB_TESTS       CONFIG_CHIP_BUILD_TESTS
    PROJECT_CONFIG  ${CONFIG_CHIP_PROJECT_CONFIG}
    DEVICE_INFO_EXAMPLE_PROVIDER CONFIG_CHIP_EXAMPLE_DEVICE_INFO_PROVIDER
    DATA_MODEL_INTERFACE CONFIG_CHIP_DATA_MODEL_INTERFACE
)
```
You might have to convert the boolean to your "enabled/disabled" string in `matter_common_gn_args`

I think this is preferred over using environment variables, as we can then configure the Kconfig value at application level (in the prj.conf files).

andy31415: kconfig value setting did carry over. Updated PR. Thank you!

config/zephyr/Kconfig
```diff
@@ -574,4 +574,12 @@ config CHIP_BLE_ADVERTISING_DURATION
 	  If CHIP_BLE_EXT_ADVERTISING is set to false, the maximum duration time is 15 minutes, 
 	  else the maximum duration time can be extended to 2880 minutes (48h).
 
+config USE_CHIP_DATA_MODEL_INTERFACE
+    bool "Use a DataModel::Provider interface for data access"
+    default n
```
axelnxp, line 579: `default n` is not needed here as `n` is the default value if not specified

axelnxp (approved):

PR #35578, opened 2024-09-13
fan rocksetting; thermostat dry mode

stingchang:
Two changes
1. Thermostat cluster attribute should support all modes.
Currently Dry and Sleep modes are not enabled. See [All system modes](https://github.com/project-chip/connectedhomeip/blob/7789409fe254593975c2572e7392b6d2f6ed9e57/zzz_generated/app-common/app-common/zap-generated/cluster-enums.h#L4195)

2. Update zap/matter file to enable chef device "Room Air Conditioner" support to fan RockSetting

update: remove first change until better support method is figured out.


src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -779,7 +779,7 @@ MatterThermostatClusterServerPreAttributeChangedCallback(const app::ConcreteAttr
         }
         auto RequestedSystemMode = static_cast<SystemModeEnum>(*value);
         if (ControlSequenceOfOperation > ControlSequenceOfOperationEnum::kCoolingAndHeatingWithReheat ||
-            RequestedSystemMode > SystemModeEnum::kFanOnly)
+            RequestedSystemMode > SystemModeEnum::kSleep)
```
hasty, line 782: We're actually hoping to deprecate/remove these modes, as they're better handled elsewhere and were just a carryover from Zigbee.

stingchang: Reverted.

andy31415 (approved):

PR #35583, opened 2024-09-13
Remove all un-specified draft clusters from SDK

tcarmelveilleux:
Some clusters have never existed in Matter spec since 1.0 (all removed before 1.0),
but due to early work in SDK and their existence in bits and pieces of prototypes, they had
leaked in.

The following clusters are removed by this PR:

- Binary Input (Basic) --> Replaced by Boolean State cluster
- ElectricalMeasurement --> Replaced by ElectricalEnergyMeasurement/ElectricalPowerMeasurement and the work of Energy TT
- On/Off Switch Configuration --> No replacement, was a draft that never was tested
- BarrierControl --> No replacement, never certifiable, not in spec
- ZLL Commissioning --> not Matter, Zigbee-only
- IAS Zone --> not Matter, Zigbee-only

None of these clusters have any:
- Spec
- Certification test plan (official, balloted)

None of them ever passed a SVE

Continued inclusion of these clusters only perpetuates errors and violates the point of having standardization for the clusters  in the standard range.


src/darwin/Framework/CHIP/zap-generated/MTRAttributeSpecifiedCheck.mm
tcarmelveilleux, line 1: Need shims from Apple. DO NOT MERGE until this comment is resolved.

yunhanw-google (approved):

src/darwin/Framework/CHIP/zap-generated/MTRAttributeSpecifiedCheck.mm
bzbarsky-apple: This is good to go now.

bzbarsky-apple (approved):

examples/all-clusters-minimal-app/esp32/main/CMakeLists.txt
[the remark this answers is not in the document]
tcarmelveilleux: Because updating the CMakelists caused reformat on change :(

PR #35597, opened 2024-09-16
Report mdns results from all interfaces instead of the highest priori…

vivien-apple:
…ty interface

**Problem**

The mdns discovery process reports the resolved IPs of the highest priority interface only. The other interfaces that are discovered may contains interesting addresses too.

**Changes**

- Update _src/platform/Darwin/DnssdContext.cpp_ to continue reporting results after the first discovered interface with results.

**Notes**
When connecting over **CASE** the results accumulates in _src/lib/address_resolve/AddressResolve_DefaultImpl.cpp_ so the score ranking takes effect. 
When connection over **PASE** via _src/controller/SetUpCodePairer.cpp_ the score ranking is on a per interface basis. I will issue a followup to rerank them there.

### DIFF
--- a/src/lib/dnssd/Discovery_ImplPlatform.cpp
+++ b/src/lib/dnssd/Discovery_ImplPlatform.cpp
@@ -43,7 +43,7 @@ static void HandleNodeResolve(void * context, DnssdService * result, const Span<
 {
     DiscoveryContext * discoveryContext = static_cast<DiscoveryContext *>(context);
 
-    if (error != CHIP_NO_ERROR)
+    if (error != CHIP_NO_ERROR && error != CHIP_ERROR_IN_PROGRESS)
     {
         discoveryContext->Release();
         return;
@@ -55,7 +55,13 @@ static void HandleNodeResolve(void * context, DnssdService * result, const Span<
 
     nodeData.Get<CommissionNodeData>().LogDetail();
     discoveryContext->OnNodeDiscovered(nodeData);
-    discoveryContext->Release();
+
+    // CHIP_ERROR_IN_PROGRESS indicates that more results are coming, so don't release
+    // the context yet.
+    if (error == CHIP_NO_ERROR)
+    {
+        discoveryContext->Release();
+    }
 }
 
 static void HandleNodeOperationalBrowse(void * context, DnssdService * result, CHIP_ERROR error)
--- a/src/platform/Darwin/DnssdContexts.cpp
+++ b/src/platform/Darwin/DnssdContexts.cpp
@@ -617,7 +617,6 @@ void ResolveContext::DispatchSuccess()
         auto addresses       = Span<Inet::IPAddress>(ips.data(), ips.size());
 
         ChipLogProgress(Discovery, "Mdns: Resolve success on interface %" PRIu32, interfaceKey.interfaceId);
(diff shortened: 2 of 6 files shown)


src/platform/Darwin/DnssdContexts.cpp
```diff
@@ -583,21 +603,38 @@ void ResolveContext::DispatchSuccess()
...
+                break;
+            }
+        }
+
+        if (shouldSkip)
+        {
+            continue;
+        }
+
         if (TryReportingResultsForInterfaceIndex(interface.first.interfaceId, interface.first.hostname,
```
bzbarsky-apple, line 628: So the thing is, TryReportingResultsForInterfaceIndex calls the callback with `CHIP_NO_ERROR`, no?

But the ChipDnssdResolve API caller passes in a callback that looks like this:

```
static void HandleNodeResolve(void * context, DnssdService * result, const Span<Inet::IPAddress> & addresses, CHIP_ERROR error)
...
    discoveryContext->OnNodeDiscovered(nodeData);
    discoveryContext->Release();
```

so all calls to the callback after the first one will be use-after-free, no?

vivien-apple: > ```
> static void HandleNodeResolve(void * context, DnssdService * result, const Span<Inet::IPAddress> & addresses, CHIP_ERROR error)
> ...
>     discoveryContext->OnNodeDiscovered(nodeData);
>     discoveryContext->Release();
> ```
> 
> so all calls to the callback after the first one will be use-after-free, no?

This is obviously correct. That said in practice it happens only when we are discovering commissioners or commissionable by instance names. Otherwise we used to emit a browse that does one `ChipDnssdResolve` per discovered interface (and increments the refcount of the context).

I have added to the PR a new `chip-tool` command to discover commissionable by instance name. Indeed Asan was complaining when it was used :)
I fixed up the PR to workaround the limitations of `HandleNodeResolve`. This is either this simple fix or a much larger and painful patch that updates the signature of `HandleNodeResolve` and fix all the platforms that calls this callback.

src/lib/dnssd/Discovery_ImplPlatform.cpp
```diff
@@ -43,7 +43,7 @@ static void HandleNodeResolve(void * context, DnssdService * result, const Span<
 {
     DiscoveryContext * discoveryContext = static_cast<DiscoveryContext *>(context);
 
-    if (error != CHIP_NO_ERROR)
+    if (error != CHIP_NO_ERROR && error != CHIP_ERROR_IN_PROGRESS)
```
bzbarsky-apple, line 46: As a followup, it really would be better for this to be an explicit argument to the callback.  Othewise it's too easy for someone to accidentally end up with CHIP_ERROR_IN_PROGRESS somewhere, especially since this behavior is completely undocumented, and get unexpected behavior.

bzbarsky-apple (approved):

src/lib/dnssd/Discovery_ImplPlatform.cpp
vivien-apple: [Agreed.](https://github.com/project-chip/connectedhomeip/issues/35823)

woody-apple:
Fast tracking this, given this has had 2 weeks for review, and is blocking some quality tests passing.

PR #35599, opened 2024-09-16
[NXP] Migrate docker image to NXP Zephyr 3.7 downstream release

axelnxp:
Migrate NXP Zephyr docker image to NXP Zephyr 3.7 downstream release.
This is needed before merging the platform updates for this migration.

PR #35600, opened 2024-09-16
[NXP][zephyr] Migrate to NXP Zephyr 3.7 downstream release

axelnxp:
- Kconfig/CMake updates for 3.7
- Optimize some stack sizes to reduce RAM footprint
- Optimize net configuration to reduce ram footprint
- Doc update
- Remove overlays from platform folders
- Workflows docker image updated to latest (depends on #35599)


axelnxp:
The CI cheks will fail until the docker file update PR is merged

src/crypto/PSASessionKeystore.cpp
```diff
@@ -68,12 +68,14 @@ class HkdfKeyAttributes : public KeyAttributesBase
     HkdfKeyAttributes() : KeyAttributesBase(PSA_KEY_TYPE_DERIVE, PSA_ALG_HKDF(PSA_ALG_SHA_256), PSA_KEY_USAGE_DERIVE, 0) {}
 };
 
+#if CHIP_CONFIG_ENABLE_ICD_CIP
```
axelnxp, line 71: FYI @maciejbaczmanski , following your PR #34925 , it introduced a warning (as error in our build) due to this function being defined but not used, so I put it under the same flag that the method calling it.
I hope this is ok for you.

scripts/build/builders/nxp.py
```diff
@@ -240,6 +240,9 @@ def generate(self):
             if 'ZEPHYR_NXP_BASE' not in os.environ:
                 raise Exception("ZEPHYR_NXP_BASE need to be set")
 
+            if 'ZEPHYR_TOOLCHAIN_VARIANT' in os.environ:
```
axelnxp, line 243: @andy31415 FYI, this fixes the nxp zephyr examples in the vscode image. This variable is set by nrf env, and clashes with our own, so we just unset it.

src/crypto/PSASessionKeystore.cpp
maciejbaczmanski: oh, right, I've added this helper later and must've forgotten about the guard... Thanks!

andy31415 (approved):

scripts/build/builders/nxp.py
andy31415: Note however that the intent of the build scripts is that I can do a `--dry-run` and see shell commands required to execute the build. So you are not allowed to do run python specific commands: in this case del of the environment (or in other cases file manipulation or python process execution).

Please add the logic to the `cmd` to be executed by the shell, even if ugly.

axelnxp: Understood, I changed it by adding `unset ZEPHYR_TOOLCHAIN_VARIANT` in `cmd`.

andy31415 (approved):

chapongatien (approved):

PR #35611, opened 2024-09-17
[Test] Creating CADMIN 1.19 python3 test module

j-ororke:
Creating python3 TC CADMIN 1.19 test module:
- Test module created following steps in PR: [4669](https://github.com/CHIP-Specifications/chip-test-plans/pull/4669)
- Removed yaml test script test_TC_CADMIN_1_19.yaml as no longer needed


src/python_testing/TC_CADMIN_1_19.py
```diff
@@ -0,0 +1,201 @@
...
+    def generate_unique_random_value(self, value):
+        while True:
+            random_value = random.randint(10000000, 99999999)
+            if random_value != value:
+                return random_value
+
+    async def CommissionAttempt(
+            self, setupPinCode: int, thnum: int, th: str, fail: bool):
+
+        logging.info(f"-----------------Commissioning with TH_CR{str(thnum)}-------------------------")
```
cecille, line 58:
```suggestion
        logging.info(f"-----------------Commissioning with TH_CR{thnum}-------------------------")
```

src/python_testing/TC_CADMIN_1_19.py
```diff
@@ -0,0 +1,201 @@
...
+from chip import ChipDeviceCtrl
+from chip.ChipDeviceCtrl import CommissioningParameters
+from chip.exceptions import ChipStackError
+from chip.native import PyChipError
+from matter_testing_support import MatterBaseTest, TestStep, async_test_body, default_matter_test_main
+from mobly import asserts
+
+
+class TC_CADMIN_1_19(MatterBaseTest):
+    async def OpenCommissioningWindow(self) -> CommissioningParameters:
```
cecille, line 39: can you use the openCommissionignWindow function in matter_testing instead of re-implementing?

src/python_testing/TC_CADMIN_1_19.py
```diff
@@ -0,0 +1,201 @@
...
+        for fid in range(1, max_fabrics - initial_number_of_fabrics):
+            # Make sure that current test step is 5, resets here after each loop
+            self.current_step_index = 5
+
+            self.step("5a")
+            params = await self.OpenCommissioningWindow()
+            setupPinCode = params.setupPinCode
+
+            self.step("5b")
+            fids_ca_dir['thca' + str(fid)] = self.certificate_authority_manager.NewCertificateAuthority()
```
cecille, line 148: why are you using strings as keys rather than just numbers if they're separate dicts?

src/python_testing/TC_CADMIN_1_19.py
```diff
@@ -0,0 +1,201 @@
...
+            fids_fa_dir['thfa' + str(fid)] = fids_ca_dir['thca' + str(fid)].NewFabricAdmin(vendorId=0xFFF1, fabricId=fid)
+            fids['th' + str(fid)] = fids_fa_dir['thfa' + str(fid)].NewController(nodeId=fid, useTestCommissioner=True)
+            await self.CommissionAttempt(setupPinCode, thnum=fid, th=fids['th' + str(fid)], fail=False)
+
+        self.step(6)
+        # TH reads the CommissionedFabrics attributes from the Node Operational Credentials cluster
+        current_fabrics = await self.read_single_attribute_check_success(dev_ctrl=self.th1, fabric_filtered=False, endpoint=0, cluster=OC_cluster, attribute=OC_cluster.Attributes.SupportedFabrics)
+        if current_fabrics != max_fabrics:
+            asserts.fail(f"Expected number of fabrics not correct, instead was {str(current_fabrics)}")
```
cecille, line 157:
```suggestion
            asserts.fail(f"Expected number of fabrics not correct, instead was {current_fabrics}")
```

src/python_testing/TC_CADMIN_1_19.py
```diff
@@ -0,0 +1,201 @@
...
+            self.current_step_index = 5
+
+            self.step("5a")
+            params = await self.OpenCommissioningWindow()
+            setupPinCode = params.setupPinCode
+
+            self.step("5b")
+            fids_ca_dir['thca' + str(fid)] = self.certificate_authority_manager.NewCertificateAuthority()
+            fids_fa_dir['thfa' + str(fid)] = fids_ca_dir['thca' + str(fid)].NewFabricAdmin(vendorId=0xFFF1, fabricId=fid)
+            fids['th' + str(fid)] = fids_fa_dir['thfa' + str(fid)].NewController(nodeId=fid, useTestCommissioner=True)
```
cecille, line 150: You don't need to use the test commissioner here.

j-ororke: Thank you, this has been removed from the test script

src/python_testing/TC_CADMIN_1_19.py
j-ororke: So, I had been using it to make the names a bit easier possibly to read the names, but since not needed I have changed those to using just the range number now for the dictionary key names.

src/python_testing/TC_CADMIN_1_19.py
j-ororke: I have updated the code to remove the OpenCommissioningWindow and CommissionAttempt local class functions to using the MatterBaseTest version of those here now. 
Also, added the timeout variable to allow dynamic setting of now instead of statically set.

src/python_testing/TC_CADMIN_1_19.py
```diff
@@ -0,0 +1,175 @@
...
+        fids_fa_dir[current_fabrics + 1] = fids_ca_dir[current_fabrics +
+                                                       1].NewFabricAdmin(vendorId=0xFFF1, fabricId=current_fabrics + 1)
+        try:
+            fids[current_fabrics + 1] = fids_fa_dir[current_fabrics + 1].NewController(nodeId=current_fabrics + 1)
+            await fids[current_fabrics + 1].CommissionOnNetwork(
+                nodeId=self.dut_node_id, setupPinCode=params.commissioningParameters.setupPinCode,
+                filterType=ChipDeviceCtrl.DiscoveryFilterType.LONG_DISCRIMINATOR, filter=params.randomDiscriminator)
+
+        except ChipStackError as e:
+            self.print_step("err", e.err)
```
cecille, line 148: remove print - think this might be leftover debug stuff.

cecille (approved):

src/python_testing/TC_CADMIN_1_19.py
```diff
@@ -0,0 +1,175 @@
...
+            self.step("5b")
+            fids_ca_dir[fid] = self.certificate_authority_manager.NewCertificateAuthority()
+            fids_fa_dir[fid] = fids_ca_dir[fid].NewFabricAdmin(vendorId=0xFFF1, fabricId=fid)
+            fids[fid] = fids_fa_dir[fid].NewController(nodeId=fid)
+
+            await fids[fid].CommissionOnNetwork(
+                nodeId=self.dut_node_id, setupPinCode=params.commissioningParameters.setupPinCode,
+                filterType=ChipDeviceCtrl.DiscoveryFilterType.LONG_DISCRIMINATOR, filter=params.randomDiscriminator)
+
+            self.print_step("commissioning iteration", fid)
```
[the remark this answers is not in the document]
j-ororke: Thank you, I have updated the location of that print statement to the top of the for loop

src/python_testing/TC_CADMIN_1_19.py
j-ororke: thank you, I have removed this debugging print statement.

src/python_testing/TC_CADMIN_1_19.py
```diff
@@ -0,0 +1,173 @@
...
+        fids_ca_dir = {}
+        fids_fa_dir = {}
+        fids = {}
+        self.print_step("total fabrics", max_fabrics - initial_number_of_fabrics)
+        for fid in range(1, max_fabrics - initial_number_of_fabrics):
+            self.print_step("commissioning iteration", fid)
+            # Make sure that current test step is 5, resets here after each loop
+            self.current_step_index = 5
+
+            self.step("5a")
```
tehampson, line 114: Can you now call this multiple times with the same value? Previously I recall that you could only call step on the next value and couldn't actually repeat

j-ororke: Hi @tehampson ,
So, you sort of can, it seems from my testing you just have to make the script repeat the prior step first, so when you utilize self.current_step_index and reset the value to the previous step it appears to work.

src/python_testing/TC_CADMIN_1_19.py
```diff
@@ -0,0 +1,173 @@
...
+        self.step(4)
+        OC_cluster = Clusters.OperationalCredentials
+        max_fabrics = await self.read_single_attribute_check_success(dev_ctrl=self.th1, fabric_filtered=False, endpoint=0, cluster=OC_cluster, attribute=OC_cluster.Attributes.SupportedFabrics)
+
+        self.step(5)
+        fids_ca_dir = {}
+        fids_fa_dir = {}
+        fids = {}
+        self.print_step("total fabrics", max_fabrics - initial_number_of_fabrics)
+        for fid in range(1, max_fabrics - initial_number_of_fabrics):
```
tehampson, line 109: Shouldn't range start at 0? If `max_fabrics - initial_number_of_fabrics` is 1 you will currently loop through this 0 times which seems incorrect

src/python_testing/TC_CADMIN_1_19.py
```diff
@@ -0,0 +1,173 @@
...
+        initial_number_of_fabrics = len(fabrics)
+
+        self.step(4)
+        OC_cluster = Clusters.OperationalCredentials
+        max_fabrics = await self.read_single_attribute_check_success(dev_ctrl=self.th1, fabric_filtered=False, endpoint=0, cluster=OC_cluster, attribute=OC_cluster.Attributes.SupportedFabrics)
+
+        self.step(5)
+        fids_ca_dir = {}
+        fids_fa_dir = {}
+        fids = {}
```
tehampson, line 107: Are these supposed to be dictionaries or arrays (list). You are using an index later on so it seems like they should an array starting at 0

src/python_testing/TC_CADMIN_1_19.py
```diff
@@ -0,0 +1,173 @@
...
+        current_fabrics = await self.read_single_attribute_check_success(dev_ctrl=self.th1, fabric_filtered=False, endpoint=0, cluster=OC_cluster, attribute=OC_cluster.Attributes.SupportedFabrics)
+        if current_fabrics != max_fabrics:
+            asserts.fail(f"Expected number of fabrics not correct, instead was {current_fabrics}")
+
+        self.step(7)
+        params = await self.openCommissioningWindow(dev_ctrl=self.th1, node_id=self.dut_node_id)
+
+        self.step(8)
+        # TH creates a controller on a new fabric and commissions DUT_CE using that controller
+        fids_ca_dir[current_fabrics + 1] = self.certificate_authority_manager.NewCertificateAuthority()
```
tehampson, line 137: I would set a variable to `current_fabrics + 1`, from there I would use that when accessing the list

src/python_testing/TC_CADMIN_1_19.py
```diff
@@ -0,0 +1,173 @@
...
+            fabric_index = await self.read_single_attribute_check_success(dev_ctrl=fids[thc], endpoint=0, cluster=OC_cluster, attribute=OC_cluster.Attributes.CurrentFabricIndex)
+
+            self.step("9b")
+            removeFabricCmd = Clusters.OperationalCredentials.Commands.RemoveFabric(fabric_index)
+            await self.th1.SendCommand(nodeid=self.dut_node_id, endpoint=0, payload=removeFabricCmd)
+
+        self.step(10)
+        # TH reads the CommissionedFabrics attributes from the Node Operational Credentials cluster.
+        current_fabrics = await self.read_single_attribute_check_success(dev_ctrl=self.th1, fabric_filtered=False, endpoint=0, cluster=OC_cluster, attribute=OC_cluster.Attributes.CommissionedFabrics)
+        if current_fabrics != initial_number_of_fabrics:
```
tehampson, line 168: Why not just use asserts.assert_equals(current_fabrics, initial_number_of_fabrics)

tehampson: Same comment for all the current if statements that then have an asserts.fail after that

src/python_testing/TC_CADMIN_1_19.py
```diff
@@ -0,0 +1,173 @@
...
+
+    @async_test_body
+    async def test_TC_CADMIN_1_19(self):
+        self.step(1)
+
+        # Establishing TH1 and TH2
+        self.th1 = self.default_controller
+
+        self.step(2)
+        GC_cluster = Clusters.GeneralCommissioning
```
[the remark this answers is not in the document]
j-ororke: Thank you, 
I have updated the test steps in the test module here now to match with the test plan.

src/python_testing/TC_CADMIN_1_19.py
j-ororke: You are correct, when I made a prior change this appears to have gotten changed by accident.
I have updated this to start at int 0 rather than start at int 1.

src/python_testing/TC_CADMIN_1_19.py
j-ororke: That is a great point, I have updated to setting a variable as you mentioned in your comment above, then using it throughout the rest of the test for current_fabrics + 1

src/python_testing/TC_CADMIN_1_19.py
j-ororke: This is a good idea! 
I have changed to using this asserts method in place of using the if statements.

src/python_testing/TC_CADMIN_1_19.py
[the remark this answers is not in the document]
tehampson: Lets leave this comment open until the testplan has been updated just to make sure the two are aligned before the PR merges

tehampson (approved):

raju-apple (approved):

src/python_testing/TC_CADMIN_1_19.py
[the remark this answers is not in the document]
cecille: This is now merged.

PR #35614, opened 2024-09-17
[Fabric-Admin] Refactor to use API methods instead of PushCommand (1/3)

### LINKED ISSUE #34910: [SVE] Switch from string-based cli API to the SDK API using ReadClients.
### Reproduction steps

For example

### Bug prevalence

Always

### Platform

raspi

### Platform Version(s)

_No response_

### Type

Test Improvement

### Testing

Manually tested with SDK

_No response_

### Anything else?

_No response_

yufengwangca:

This is the first iteration of the migration process."

Fixes #34910


examples/fabric-admin/device_manager/DeviceManager.cpp
```diff
@@ -30,13 +31,12 @@ using namespace chip::app::Clusters;
...
 namespace {
 
-constexpr uint16_t kWindowTimeout              = 300;
-constexpr uint16_t kIteration                  = 1000;
-constexpr uint16_t kSubscribeMinInterval       = 0;
-constexpr uint16_t kSubscribeMaxInterval       = 60;
-constexpr uint16_t kAggragatorEndpointId       = 1;
-constexpr uint16_t kMaxDiscriminatorLength     = 4095;
-constexpr uint8_t kEnhancedCommissioningMethod = 1;
+constexpr uint16_t kWindowTimeout          = 300;
```
andy31415, line 34: maybe at a later time, could we add units of measurement to these constants?
Just saw this because code changed here ...

examples/fabric-admin/device_manager/PairingManager.cpp
```diff
@@ -0,0 +1,147 @@
...
+using namespace ::chip;
+
+Platform::UniquePtr<chip::Controller::CommissioningWindowOpener> PairingManager::mWindowOpener;
+
+PairingManager::PairingManager() :
+    mOnOpenCommissioningWindowCallback(OnOpenCommissioningWindowResponse, this),
+    mOnOpenCommissioningWindowVerifierCallback(OnOpenCommissioningWindowVerifierResponse, this)
+{}
+
+CHIP_ERROR PairingManager::Init(chip::Controller::DeviceCommissioner * commissioner)
```
andy31415, line 34: Is there a reason to allow for CHIP_ERROR return when this cannot fail in any way?

examples/fabric-admin/device_manager/PairingManager.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+void PairingManager::OnOpenCommissioningWindow(intptr_t context)
+{
+    auto * params         = reinterpret_cast<CommissioningWindowParams *>(context);
+    PairingManager & self = PairingManager::Instance();
+
+    if (self.mCommissioner == nullptr)
+    {
+        ChipLogError(AppServer, "Commissioner is null, cannot open commissioning window");
+        return;
```
andy31415, line 79: all these returns leak CommissioningWindowParams.
Could we create a unique_ptr out of it to ensure deletion?

examples/fabric-admin/device_manager/PairingManager.cpp
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    if (!verifier.empty())
+    {
+        memcpy(params->verifierBuffer, verifier.data(), verifier.size());
+        params->verifier = ByteSpan(params->verifierBuffer, verifier.size());
+    }
+
+    // Schedule work on the Matter thread
+    chip::DeviceLayer::PlatformMgr().ScheduleWork(OnOpenCommissioningWindow, reinterpret_cast<intptr_t>(params));
```
andy31415, line 66: ScheduleWork returns a CHIP_ERROR. we should return that error instead of ignoring it.

Also if an error occurs on schedule work, we should delete params that was allocated

examples/fabric-admin/device_manager/PairingManager.cpp
```diff
@@ -0,0 +1,147 @@
...
+    auto * params         = reinterpret_cast<CommissioningWindowParams *>(context);
+    PairingManager & self = PairingManager::Instance();
+
+    if (self.mCommissioner == nullptr)
+    {
+        ChipLogError(AppServer, "Commissioner is null, cannot open commissioning window");
+        return;
+    }
+
+    mWindowOpener = Platform::MakeUnique<Controller::CommissioningWindowOpener>(self.mCommissioner);
```
andy31415, line 82: what is the lifetime of mWindowOpener? When does it get released? is it an error for a window opener to somehow be active while anotherone is still processing?

Some comments (maybe in the header) would help.

examples/fabric-admin/device_manager/PairingManager.h
```diff
@@ -0,0 +1,87 @@
...
+#include <crypto/CHIPCryptoPAL.h>
+
+class CommissioningWindowDelegate
+{
+public:
+    virtual void OnCommissioningWindowOpened(chip::NodeId deviceId, CHIP_ERROR err, chip::SetupPayload payload) = 0;
+    virtual ~CommissioningWindowDelegate()                                                                      = default;
+};
+
+class PairingManager
```
andy31415, line 32: please add doccomments about what this class is about and how it is supposed to be used.

examples/fabric-admin/device_manager/PairingManager.h
```diff
@@ -0,0 +1,87 @@
...
+
+    CHIP_ERROR Init(chip::Controller::DeviceCommissioner * commissioner);
+
+    CHIP_ERROR OpenCommissioningWindow(chip::NodeId nodeId, chip::EndpointId endpointId, uint16_t commissioningTimeout,
+                                       uint32_t iterations, uint16_t discriminator, const chip::ByteSpan & salt,
+                                       const chip::ByteSpan & verifier);
+
+    void RegisterOpenCommissioningWindowDelegate(CommissioningWindowDelegate * delegate)
+    {
+        ChipLogProgress(NotSpecified, "yujuan: PairingManager::RegisterOpenCommissioningWindowDelegate");
```
andy31415, line 49: `yujuan` seems like a remnant from some debugging code.

examples/fabric-admin/device_manager/PairingManager.h
```diff
@@ -0,0 +1,87 @@
...
+        mCommissioningWindowDelegate = delegate;
+    }
+
+    void UnregisterOpenCommissioningWindowDelegate() { mCommissioningWindowDelegate = nullptr; }
```
andy31415, line 53: How about just having a `SetDelegate` call that accepts null?

examples/fabric-admin/device_manager/DeviceManager.cpp
```diff
@@ -141,11 +140,15 @@ void DeviceManager::OpenRemoteDeviceCommissioningWindow(EndpointId remoteEndpoin
     uint16_t discriminator =
```
tehampson, line 140: nit, can you remove ` StringBuilder<kMaxCommandSize> commandBuilder;` from above on line 137

examples/fabric-admin/device_manager/PairingManager.cpp
tehampson: Maybe we should fail if mCommissioner is already set. If we don't we should probably VerifyOrDie because it is never expected to be called twice

examples/fabric-admin/device_manager/PairingManager.cpp
```diff
@@ -0,0 +1,147 @@
...
+    auto * params                      = new CommissioningWindowParams();
+    params->nodeId                     = nodeId;
+    params->endpointId                 = endpointId;
+    params->commissioningWindowTimeout = commissioningTimeout;
+    params->iteration                  = iterations;
+    params->discriminator              = discriminator;
+
+    if (!salt.empty())
+    {
+        memcpy(params->saltBuffer, salt.data(), salt.size());
```
tehampson, line 55: It doesn't look like you have validated that salt.size() is less than the buffer we are trying to put it into

examples/fabric-admin/device_manager/PairingManager.cpp
```diff
@@ -0,0 +1,147 @@
...
+        params->salt = ByteSpan(params->saltBuffer, salt.size());
+    }
+
+    if (!verifier.empty())
+    {
+        memcpy(params->verifierBuffer, verifier.data(), verifier.size());
```
tehampson, line 61: Same, you are not checking verifier.size()

examples/fabric-admin/device_manager/PairingManager.cpp
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    // Clean up params
+    delete params;
+}
+
+void PairingManager::OnOpenCommissioningWindowResponse(void * context, NodeId remoteId, CHIP_ERROR err, chip::SetupPayload payload)
+{
+    PairingManager * self = static_cast<PairingManager *>(context);
+    if (self->mCommissioningWindowDelegate)
```
tehampson, line 132: In PairingManager::OnOpenCommissioningWindowVerifierResponse you are checking for nullptr, but not here. Should you be checking before using it?

examples/fabric-admin/device_manager/PairingManager.cpp
```diff
@@ -0,0 +1,147 @@
...
+    OnOpenCommissioningWindowVerifierResponse(context, remoteId, err);
+}
+
+void PairingManager::OnOpenCommissioningWindowVerifierResponse(void * context, NodeId remoteId, CHIP_ERROR err)
+{
+    LogErrorOnFailure(err);
+
+    PairingManager * self = reinterpret_cast<PairingManager *>(context);
+    VerifyOrReturn(self != nullptr, ChipLogError(NotSpecified, "OnOpenCommissioningWindowCommand: context is null"));
+}
```
tehampson, line 147: Should we be doing something here, if not can you add a comment as to why we only get if the pointer is null, Otherwise I don't understand why it is important for `PairingManager::OnOpenCommissioningWindowResponse` to call this method

examples/fabric-admin/device_manager/DeviceManager.cpp
yufengwangca: only kWindowTimeout is unambiguous, changed to kWindowTimeoutS

examples/fabric-admin/device_manager/PairingManager.cpp
yufengwangca: Return a CHIP_ERROR for future initialization, but we can modify the behavior later if needed.

examples/fabric-admin/device_manager/PairingManager.h
yufengwangca: We need to add other delegate later, change to SetOpenCommissioningWindowDelegate

examples/fabric-admin/device_manager/PairingManager.cpp
```diff
@@ -0,0 +1,170 @@
...
+    auto params                        = Platform::MakeUnique<CommissioningWindowParams>();
+    params->nodeId                     = nodeId;
+    params->endpointId                 = endpointId;
+    params->commissioningWindowTimeout = commissioningTimeout;
+    params->iteration                  = iterations;
+    params->discriminator              = discriminator;
+
+    if (!salt.empty())
+    {
+        if (salt.size() > Crypto::kSpake2p_Max_PBKDF_Salt_Length)
```
andy31415, line 64: constant compare is not immediately obvious as being sizeof(params->saltBuffer) could we use sizeof instead here and below?

examples/fabric-admin/device_manager/PairingManager.h
```diff
@@ -0,0 +1,122 @@
...
+     * @param endpointId The target endpoint ID for commissioning.
+     * @param commissioningTimeout Timeout for the commissioning window in seconds.
+     * @param iterations Iterations for PBKDF calculations.
+     * @param discriminator Discriminator for commissioning.
+     * @param salt Optional salt for verifier-based commissioning.
+     * @param verifier Optional verifier for enhanced commissioning security.
+     *
+     * @return CHIP_ERROR_INCORRECT_STATE if a commissioning window is already open.
+     */
+    CHIP_ERROR OpenCommissioningWindow(chip::NodeId nodeId, chip::EndpointId endpointId, uint16_t commissioningTimeout,
```
andy31415, line 78:
```suggestion
    CHIP_ERROR OpenCommissioningWindow(chip::NodeId nodeId, chip::EndpointId endpointId, uint16_t commissioningTimeoutSec,
```

and please make all the changes subsequent to this. We need a unit of measurement for anything that is a timeout or duration, since it is not obvious for things to be seconds or milisecond or anything else. Comments are not sufficient - people will not see header comments when looking at the cpp file.

andy31415 (approved):

saurabhst (approved):

PR #35619, opened 2024-09-17
Update PICS Generator to 1.4

ReneJosefsen:
THis PR updates the PICS generator to match the 1.4 SVE PICS

It also adds a script used to check the matching of cluster names to the PICS files, this is intended to make it easier to catch naming mismatch or missing PICS files in future versions.

### DIFF
--- a/src/tools/PICS-generator/PICSGenerator.py
+++ b/src/tools/PICS-generator/PICSGenerator.py
@@ -1,5 +1,5 @@
 #
-#    Copyright (c) 2024 Project CHIP Authors
+#    Copyright (c) 2023-2024 Project CHIP Authors
 #    All rights reserved.
 #
 #    Licensed under the Apache License, Version 2.0 (the "License");
--- a/src/tools/PICS-generator/PICSGenerator.py
+++ b/src/tools/PICS-generator/PICSGenerator.py
@@ -1,5 +1,5 @@
 #
-#    Copyright (c) 2023 Project CHIP Authors
+#    Copyright (c) 2024 Project CHIP Authors
 #    All rights reserved.
 #
 #    Licensed under the Apache License, Version 2.0 (the "License");
@@ -24,6 +24,8 @@ import xml.etree.ElementTree as ET
 import chip.clusters as Clusters
 from rich.console import Console
 
+from pics_generator_support import pics_xml_file_list_loader, map_cluster_name_to_pics_xml
+
 # Add the path to python_testing folder, in order to be able to import from matter_testing_support
 sys.path.append(os.path.abspath(sys.path[0] + "/../../python_testing"))
 from matter_testing_support import MatterBaseTest, async_test_body, default_matter_test_main  # noqa: E402
@@ -40,52 +42,7 @@ def GenerateDevicePicsXmlFiles(clusterName, clusterPicsCode, featurePicsList, at
 
     console.print(f"Handling PICS for {clusterName}")
 
-    # Map clusters to common XML template if needed
-    if "ICDManagement" == clusterName:
-        picsFileName = "ICD Management"
-
-    elif "OTA Software Update Provider" in clusterName or \
-            "OTA Software Update Requestor" in clusterName:
-        picsFileName = "OTA Software Update"
-
-    elif "On/Off" == clusterName:
-        picsFileName = clusterName.replace("/", "-")
-
-    elif "GroupKeyManagement" == clusterName:
-        picsFileName = "Group Communication"
-
-    elif "Wake on LAN" == clusterName or \
-         "Low Power" == clusterName or \
-         "Keypad Input" == clusterName or \
-         "Audio Output" == clusterName or \
-         "Media Input" == clusterName or \
-         "Target Navigator" == clusterName or \
-         "Content Control" == clusterName or \
-         "Channel" == clusterName or \
-         "Media Playback" == clusterName or \
-         "Account Login" == clusterName or \
-         "Application Basic" == clusterName or \
-         "Content Launcher" == clusterName or \
-         "Content App Observer" == clusterName or \
-         "Application Launcher" == clusterName:
-
-        picsFileName = "Media Cluster"
-
-    elif "Operational Credentials" == clusterName:
-        picsFileName = "Node Operational Credentials"
-
-    # Workaround for naming colisions with current logic
-    elif "Thermostat" == clusterName:
-        picsFileName = "Thermostat Cluster"
-
-    elif "Boolean State" == clusterName:
-        picsFileName = "Boolean State Cluster"
-
-    elif "AccessControl" in clusterName:
-        picsFileName = "Access Control Cluster"
-
-    else:
-        picsFileName = clusterName
+    picsFileName = map_cluster_name_to_pics_xml(clusterName, xmlFileList)
 
     # Determine if file has already been handled and use this file
     for outputFolderFileName in os.listdir(outputPathStr):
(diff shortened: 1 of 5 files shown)


rbultman (approved):

src/tools/PICS-generator/XMLPICSValidator.py
```diff
@@ -0,0 +1,122 @@
...
+
+import argparse
+import os
+import sys
+
+from rich.console import Console
+
+# Add the path to python_testing folder, in order to be able to import from matter_testing_support
+sys.path.append(os.path.abspath(sys.path[0] + "/../../python_testing"))
+from matter_testing_support import MatterBaseTest, async_test_body, default_matter_test_main  # noqa: E402
```
andy31415, line 26: we are considering moving modules around so that we can have test packages in pypi eventually. 

Could we have CI validation that these scripts are runnable?

src/tools/PICS-generator/XMLPICSValidator.py
```diff
@@ -0,0 +1,122 @@
+#
```
andy31415, line 1: `README.md` in this folder seems to describe the content of the file. We should add information about this new script there.

src/tools/PICS-generator/XMLPICSValidator.py
```diff
@@ -0,0 +1,122 @@
...
+#    Copyright (c) 2023 Project CHIP Authors
```
andy31415, line 2:
```suggestion
#    Copyright (c) 2024 Project CHIP Authors
```

src/tools/PICS-generator/XMLPICSValidator.py
```diff
@@ -0,0 +1,122 @@
...
+
+        global xml_clusters
+        xml_clusters, problems = build_xml_clusters()
+
+        for cluster in xml_clusters:
+
+            fileName = ""
+            clusterName = xml_clusters[cluster].name
+
+            if "ICDManagement" == clusterName:
```
andy31415, line 63: for many of these `==` it seems that we could have a dictionary for mapping instead of if/else.

src/tools/PICS-generator/XMLPICSValidator.py
```diff
@@ -0,0 +1,122 @@
...
+                picsFileName = "ICD Management"
+
+            elif "OTA Software Update Provider" in clusterName or \
+                 "OTA Software Update Requestor" in clusterName:
```
andy31415, line 67: most of these have `==` however OTA and ACL seem to be `in`. What is the reason for that? what are the different names and if we use a dictionary could we just enumerate them?

src/tools/PICS-generator/XMLPICSValidator.py
```diff
@@ -0,0 +1,122 @@
...
+            elif "Boolean State" == clusterName:
+                picsFileName = "Boolean State Cluster"
+
+            elif "AccessControl" in clusterName:
+                picsFileName = "Access Control Cluster"
+
+            else:
+                picsFileName = clusterName
+
+            for file in xmlFileList:
```
andy31415, line 109: This could be extracted in a method, so that the build of the xmlFile list and the usage are not this separated. I would imagine we should have a global (since this seems to be the sharing method here ... not ideal) `XmlFileListLoader` that has an initial Load to construct the list and then a `findFileNameForPrefix` that does this loop.

src/tools/PICS-generator/XMLPICSValidator.py
```diff
@@ -0,0 +1,122 @@
...
+    print(f"{xmlTemplatePathStr}/{PICSXmlFile}")
+
+
+class XMLPICSValidator(MatterBaseTest):
+    @async_test_body
+    async def test_xml_pics_validator(self):
+
+        # Create console to print
+        global console
+        console = Console()
```
andy31415, line 53: why do we use a global just to immediately set it?

src/tools/PICS-generator/XMLPICSValidator.py
ReneJosefsen: Sure, I can do that. It was mostly intended for myself to catch these naming issues between PICS and DM cluster names, but good point. I will add some text describing what the purpose of the new script is and how to use it.

src/tools/PICS-generator/XMLPICSValidator.py
ReneJosefsen: Sure, let me try to rework this.

src/tools/PICS-generator/XMLPICSValidator.py
ReneJosefsen: I'm totally up for figuring something out here, this is basically just a "workaround" to be able to get the matter_testing_support but not having to have these tools places in the python_testing folder but instead this tools folder, which IMO made more sense 🤷🏻‍♂️

src/tools/PICS-generator/XMLPICSValidator.py
```diff
@@ -0,0 +1,122 @@
...
+            for file in xmlFileList:
+                if file.lower().startswith(picsFileName.lower()):
+                    fileName = file
+                    break
+
+            if fileName:
+                console.print(f"[blue]\"{clusterName}\" - \"{fileName}\" ✅")
+            else:
+                console.print(f"[red]Could not find matching file for \"{clusterName}\" ❌")
+                continue
```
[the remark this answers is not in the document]
ReneJosefsen: Agree, I think this got missed when I reworked it, I will get it removed. 👍🏻

src/tools/PICS-generator/XMLPICSValidator.py
ReneJosefsen: Copy paste error from the PICS Generator, this is not needed here. I will have it cleaned up.

src/tools/PICS-generator/XMLPICSValidator.py
```diff
@@ -0,0 +1,122 @@
...
+class XMLPICSValidator(MatterBaseTest):
+    @async_test_body
+    async def test_xml_pics_validator(self):
+
+        # Create console to print
+        global console
+        console = Console()
+
+        global xml_clusters
+        xml_clusters, problems = build_xml_clusters()
```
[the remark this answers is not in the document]
ReneJosefsen: Copy paste error from the PICS Generator, this is not needed here. I will have it cleaned up.

src/tools/PICS-generator/pics_generator_support.py
```diff
@@ -0,0 +1,77 @@
...
+    return pics_xml_file_list
+
+
+def map_cluster_name_to_pics_xml(cluster_name, pics_xml_file_list) -> str:
+    file_name = ""
+
+    try:
+        pics_file_name = cluster_to_pics_dict[cluster_name]
+    except KeyError:
+        pics_file_name = cluster_name
```
andy31415, line 70:
```suggestion
    pics_file_name = cluster_to_pics_dict.get(cluster_name, cluster_name)
```

andy31415 (approved):

yunhanw-google (approved):

src/tools/PICS-generator/PICSGenerator.py
```diff
@@ -1,5 +1,5 @@
 #
-#    Copyright (c) 2023 Project CHIP Authors
+#    Copyright (c) 2024 Project CHIP Authors
```
cecille, line 2:
```suggestion
#    Copyright (c) 2023-2024 Project CHIP Authors
```

cecille (approved):
(with Andrei's changes)

PR #35621, opened 2024-09-17
[Telink] Lighting app build with disabled CONFIG_PWM

BorysNykytiuk:
**Change overview**

- Add Dummy backend to the PWM Manager
- Add option to execute Lighting app with LED Manager
- Lighting app does not depend on the CONFIG_PWM now
- In case "CONFIG_PWM=n" the app will use GPIO output to control LED


s07641069 (approved):

examples/lighting-app/telink/src/AppTask.cpp
```diff
@@ -58,7 +59,9 @@ CHIP_ERROR AppTask::Init(void)
 {
     SetExampleButtonCallbacks(LightingActionEventHandler);
     InitCommonParts();
-
+#if (!CONFIG_PWM)
```
andriy-bilynskyy, line 62: It shouldn't be here. Just override Link LED method.

BorysNykytiuk: done. I've added it to AppTask class

andriy-bilynskyy (approved):

andy31415 (approved):

PR #35625, opened 2024-09-17
Sharing my local test runner script for `CI-like test app build and execution`

andy31415:
This script generally makes it convenient to `build` and `python-tests` for local execution and debugging.

image

image

### DIFF
--- a/scripts/tests/local.py
+++ b/scripts/tests/local.py
@@ -19,6 +19,21 @@ import click
 import coloredlogs
 import tabulate
 
+# We compile for the local architecture. Figure out what platform we need
+def _get_native_machine_target():
+    """
+    Returns the build prefix for applications, such as 'linux-x64'.
+    """
+    current_system_info = uname()
+    arch = current_system_info.machine
+    if arch == 'x86_64':
+        arch = 'x64'
+    elif arch == 'i386' or arch == 'i686':
+        arch = 'x86'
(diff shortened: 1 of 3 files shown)


woody-apple (approved):

tcarmelveilleux (approved):

scripts/tests/local.py
```diff
@@ -0,0 +1,771 @@
+#!/usr/bin/env python3
```
tcarmelveilleux, line 1: Please add license

scripts/tests/local.py
```diff
@@ -0,0 +1,786 @@
...
+        if not from_skip_filter.startswith("*"):
+            from_skip_filter = "*" + from_skip_filter
+        if not from_skip_filter.endswith("*"):
+            from_skip_filter = from_skip_filter + "*"
+
+    # This MUST be available or perfetto dies. This is VERY annoying to debug
+    if not os.path.exists("out/trace_data"):
+        os.mkdir("out/trace_data")
+
+    # IGNORES are taken out of `src/python_testing/execute_python_tests.py` in the SDK
```
bzbarsky-apple, line 385: Can we have a single source of truth for this?  Otherwise, these lists will diverge, which will cause problems.

andy31415: will work towards it with @vatsalghelani-csa - we probably want something loadable and then we can use it from both places (one our official runner and one being this local runner)

PR #35629, opened 2024-09-17
[libfuzzer] Fuzzing different Transport Types for all-clusters-app

Alami-Amine:
Fuzzing different Transport Types for all-clusters-app instead of just with `Transport::Type = kUndefined`

### DIFF
--- a/examples/all-clusters-app/linux/fuzzing-main.cpp
+++ b/examples/all-clusters-app/linux/fuzzing-main.cpp
@@ -73,7 +73,11 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t * aData, size_t aSize)
     // For now, just dump the data as a UDP payload into the session manager.
     // But maybe we should try to separately extract a PeerAddress and data from
     // the incoming data?
-    Transport::PeerAddress peerAddr;
+
+    // dumping payload with random transport types
+    Transport::Type fuzzedTransportType = static_cast<Transport::Type>(*aData % 5);
+    Transport::PeerAddress peerAddr(fuzzedTransportType);
+
     System::PacketBufferHandle buf =
         System::PacketBufferHandle::NewWithData(aData, aSize, /* aAdditionalSize = */ 0, /* aReservedSize = */ 0);
     if (buf.IsNull())
@@ -84,8 +88,17 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t * aData, size_t aSize)
 
     // Ignoring the return value from OnMessageReceived, because we might be
     // passing it all sorts of garbage that will cause it to fail.
-    Server::GetInstance().GetSecureSessionManager().OnMessageReceived(peerAddr, std::move(buf));
 
+    // for TCP we need to have MessageTransportContext
+    if (fuzzedTransportType == Transport::Type::kTcp)
+    {
+        Transport::MessageTransportContext msgContext;
+        Server::GetInstance().GetSecureSessionManager().OnMessageReceived(peerAddr, std::move(buf), &msgContext);
+    }
+    else
+    {
+        Server::GetInstance().GetSecureSessionManager().OnMessageReceived(peerAddr, std::move(buf));
+    }
     // Now process pending events until our sentinel is reached.
     PlatformMgr().ScheduleWork([](intptr_t) { PlatformMgr().StopEventLoopTask(); });
     PlatformMgr().RunEventLoop();
--- a/examples/all-clusters-app/linux/fuzzing-main.cpp
+++ b/examples/all-clusters-app/linux/fuzzing-main.cpp
@@ -75,7 +75,7 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t * aData, size_t aSize)
     // the incoming data?
 
     // dumping payload with fuzzed transport types
-    constexpr uint8_t numberOfTypes = static_cast<int>(Transport::Type::kLast) + 1;
+    constexpr uint8_t numberOfTypes     = static_cast<int>(Transport::Type::kLast) + 1;
     Transport::Type fuzzedTransportType = static_cast<Transport::Type>(aData[0] % numberOfTypes);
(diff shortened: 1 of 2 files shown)


examples/all-clusters-app/linux/fuzzing-main.cpp
```diff
@@ -73,7 +73,11 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t * aData, size_t aSize)
     // For now, just dump the data as a UDP payload into the session manager.
     // But maybe we should try to separately extract a PeerAddress and data from
     // the incoming data?
-    Transport::PeerAddress peerAddr;
+
+    // dumping payload with random transport types
+    Transport::Type fuzzedTransportType = static_cast<Transport::Type>(*aData % 5);
```
andy31415, line 78: how about adding a kLast constant on the Transport::Type so that `5` is nto so magical?

kiel-apple (approved):

examples/all-clusters-app/linux/fuzzing-main.cpp
bzbarsky-apple, line 78: This will always have the transport type match the first byte of the data.  Shouldn't we be using only the parts of the data starting with index 1 for the actual data, to decouple those?

examples/all-clusters-app/linux/fuzzing-main.cpp
```diff
@@ -84,8 +88,17 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t * aData, size_t aSize)
 
     // Ignoring the return value from OnMessageReceived, because we might be
     // passing it all sorts of garbage that will cause it to fail.
-    Server::GetInstance().GetSecureSessionManager().OnMessageReceived(peerAddr, std::move(buf));
 
+    // for TCP we need to have MessageTransportContext
+    if (fuzzedTransportType == Transport::Type::kTcp)
+    {
+        Transport::MessageTransportContext msgContext;
```
bzbarsky-apple, line 95: This isn't a valid MessageTransportContext (has no actual connection pointer) for TCP, so this can cause crashes that would never arise normally.

examples/all-clusters-app/linux/fuzzing-main.cpp
Alami-Amine: I added one and I name it Transport::Type::kNumofTypes since it seemed more descriptive to me, is that ok?

examples/all-clusters-app/linux/fuzzing-main.cpp
Alami-Amine: - actually I added `msgContext` and this whole `if condition` because without it `ctxt` below was null and an ASAN crash related to dereferencing the null pointer in `ctxt->conn == nullptr` showed up...I wonder if we should do something about it 
- in the code we seem to protect well against not have an actual connection pointer with `ctxt->conn == nullptr`

- what are your suggestions?

https://github.com/project-chip/connectedhomeip/blob/1d6b0a5815dbaa2a8c46bd5e27648b661b746e8c/src/transport/SessionManager.cpp#L903

examples/all-clusters-app/linux/fuzzing-main.cpp
Alami-Amine: Done

src/transport/raw/PeerAddress.h
```diff
@@ -54,6 +54,7 @@ enum class Type : uint8_t
     kBle,
     kTcp,
     kWiFiPAF,
+    kNumofTypes, // This is not an actual transport type, it just refers to the number of transport types
```
andy31415, line 57: this means kNumOfTypes will be a valid enum (compiles) but invalid handled anywhere.
How about having `kLast = kWiFiPAF` and then there will never be invalid enums in the code and you also have the count?

Alami-Amine: makes sense, done

examples/all-clusters-app/linux/fuzzing-main.cpp
andy31415: If this causes crashes that would not normally happen (or code paths that are not valid normally) we could skip TCP testing.

I guess the question is if context.conn is allowed to be null in every situation and is handled or if we rely on it being non-null in some code and if that is fine. 

@pidarped - is TCP transport fuzzable?

pidarped: The context object pointer should be *not null* for the TCP message processing path. 
The context.conn pointer is expected to be a valid pointer normally, but for the purposes of this fuzz testing can be null/garbage I believe. It is only compared with the sessionConnection pointer already stored in the secure session.

bzbarsky-apple: Sounds good.

examples/all-clusters-app/linux/fuzzing-main.cpp
```diff
@@ -73,9 +73,19 @@ extern "C" int LLVMFuzzerTestOneInput(const uint8_t * aData, size_t aSize)
...
     // But maybe we should try to separately extract a PeerAddress and data from
     // the incoming data?
-    Transport::PeerAddress peerAddr;
+
+    // dumping payload with fuzzed transport types
+    constexpr uint8_t numberOfTypes     = static_cast<int>(Transport::Type::kLast) + 1;
+    Transport::Type fuzzedTransportType = static_cast<Transport::Type>(aData[0] % numberOfTypes);
+    Transport::PeerAddress peerAddr(fuzzedTransportType);
+
+    if (aSize < 1)
```
bzbarsky-apple, line 82: This needs to be up before we do `aData[0]`, no?

Alami-Amine: changed it to `if (aSize < 2)` and displaced it above.

bzbarsky-apple (approved):

pidarped (approved):

PR #35631, opened 2024-09-17
Valve cluster: refactor to make unit-testable

cecille:
Refactor of Valve to enable unit testing of the cluster logic per the proposal in https://project-chip.github.io/connectedhomeip-doc/cluster_and_device_type_dev/unit_testing_clusters.html.

Unit tests are in TestValveConfigurationAndControl

General outline
- Server implementation receives TLV, calls into the cluster logic for all read/write/command operations. Server is only responsible for encode / decode, error translation and installing itself into the registry. Uses AttributeAccessInterface and CommandHandlerInterface
- Cluster logic 
  - handles all reads / writes and commands
  - no TLV parsing (handled by server)
  - no interaction with the ember layer (handled by the matter context)
  - no interaction with the hardware (handled by the delegate)
  - Uses a subclass for all attribute reads / writes that handles storage and attribute reporting. Data members are private to force the logic to go through the setters to properly trigger reporting and storage
- Matter context handles attribute reporting (and will handle event logging once I implement it). This can be mocked for testing. 
- Timers and clocks are injected by the test at the system layer (not the matter context) because there is a single clock for the entire system.

- App owns all the resources - delegate, matter context, logic and server and initializes them as part of the app startup
- one of each is instantiated once for each endpoint

Open questions on the general design:
- should the cluster logic return IM layer errors directly
- should the cluster logic use the codegen'd data types for the attributes
- is there a way to get rid of some of the boilerplate?
- should the cluster logic handle the global attributes or some of the global attributes?
- Should Setter calls on the cluster logic (ex SetValveFault) be piped through the delegate or should we let the app report these to the logic directly?
- One server per endpoint? Or pooled?

Open questions for the valve:
- should the cluster logic periodically reach out to the delegate to get updates on state / level for async opens? Should it allow the delegate to also report back? Both?

Remaining:
- event generation
- AutoCloseTime handling
- handling of async opens / closes
- set functions for things like valve fault

- command handling for > EP1 seems to be broken, still need to fix that.


examples/valve/linux/main.cpp
```diff
@@ -0,0 +1,168 @@
...
+};
+
+// TODO: Does this conflict with the kvs delegate that's statically allocated in the AppMain?
+// I think as long as the manager is the same, we're ok, but need to confirm.
+KvsPersistentStorageDelegate sStorage;
+
+class NonLevelValveEndpoint
+{
+public:
+    NonLevelValveEndpoint(EndpointId endpoint) :
```
jmartinez-silabs, line 64: should we inject the kvsStorage here

examples/valve/linux/main.cpp
```diff
@@ -0,0 +1,168 @@
...
+    {}
+    CHIP_ERROR Init()
+    {
+        ReturnErrorOnFailure(mLogic.Init(kConformance, kInitParams));
+        ReturnErrorOnFailure(mInterface.Init());
+        return CHIP_NO_ERROR;
+    }
+
+private:
+    const ClusterConformance kConformance = {
```
jmartinez-silabs, line 76: this ignores generated code

src/app/clusters/valve-configuration-and-control-server/valve-configuration-and-control-cluster-logic.h
```diff
@@ -0,0 +1,221 @@
...
+namespace chip {
+namespace app {
+namespace Clusters {
+namespace ValveConfigurationAndControl {
+
+struct ClusterConformance
+{
+    inline bool HasFeature(Feature feature) const { return featureMap & to_underlying(feature); }
+    uint32_t featureMap;
+    bool supportsDefaultOpenLevel;
```
jmartinez-silabs, line 40: feature map and all following bool supports should be a bitfield
maybe something like this
```
union {
    struct {
        bool supportsDefaultOpenLevel: 1;
        bool supportsValveFault: 1;
        bool supportsLevelStep: 1;
    };
    uint32_t featureMap;
};
```
I'm not sure that it is portable tho. Maybe Using bitflag and enum type would be best

src/app/clusters/valve-configuration-and-control-server/valve-configuration-and-control-matter-context.h
```diff
@@ -0,0 +1,68 @@
...
+
+namespace chip {
+namespace app {
+namespace Clusters {
+namespace ValveConfigurationAndControl {
+
+/** @brief
+ *    Interface to allow interaction with interaction model and ember layers. Can be faked for unit testing.
+ */
+class MatterContext
```
tehampson, line 34: Just an idea that there could be a Base MatterContext that is common to all clusters (like MarkDirty) That would mean what you have here would need to be a little more generic to accept EndpointId and ClusterId, but it could later allow for things like that to funnel into one central command MatterContext that is shared everywhere to reduce boiler plate

src/app/tests/TestValveConfigurationAndControl.cpp
```diff
@@ -0,0 +1,2250 @@
...
+    MockedMatterContext(EndpointId endpoint, PersistentStorageDelegate & persistentStorageDelegate) :
+        MatterContext(endpoint, persistentStorageDelegate)
+    {}
+    void MarkDirty(AttributeId id) override { mDirtyMarkedList.push_back(id); }
+    std::vector<AttributeId> GetDirtyList() { return mDirtyMarkedList; }
+    void ClearDirtyList() { mDirtyMarkedList.clear(); }
+    ~MockedMatterContext() { gSystemLayerAndClock.Clear(); }
+
+private:
+    // Won't handle double-marking an attribute, so don't do that in tests
```
tehampson, line 285: nit, not sure I follow this comment. This does supposed double-marking an attribute you would see it come up twice in this vector

src/app/clusters/valve-configuration-and-control-server/valve-configuration-and-control-cluster-logic.cpp
```diff
@@ -0,0 +1,565 @@
...
+
+CHIP_ERROR ClusterStateAttributes::SetOpenDuration(const DataModel::Nullable<ElapsedS> & openDuration)
+{
+    bool dirty          = openDuration != mState.openDuration;
+    mState.openDuration = openDuration;
+    if (dirty)
+    {
+        mMatterContext.MarkDirty(Attributes::OpenDuration::Id);
+    }
+    return CHIP_NO_ERROR;
```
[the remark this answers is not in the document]
tcarmelveilleux: +1 on that pattern.

Also, would be great to have a AttributeValue<T> class that can has a SetValue(const T& val, TheInterfaceThatHasMarkDirty *) which then we could have:

`return mOpenDuration.SetValue(openDuration, mMatterContext)`

src/app/clusters/valve-configuration-and-control-server/valve-configuration-and-control-server-disco.cpp
```diff
@@ -0,0 +1,218 @@
...
+using namespace Attributes;
+using namespace Commands;
+using namespace Protocols::InteractionModel;
+namespace {
+
+CHIP_ERROR TranslateErrorToIMStatus(CHIP_ERROR err)
+{
+    if (err == CHIP_NO_ERROR)
+    {
+        return err;
```
tcarmelveilleux, line 48: This doesn't feel like it belongs in this cluster.

andy31415: We may want to eventually use https://github.com/project-chip/connectedhomeip/blob/master/src/app/data-model-provider/ActionReturnStatus.h#L44

Beyond that though, translating a CHIP_ERROR to another CHIP_ERROR feels odd.

lboue:
Do you expect to finish soon?

cecille:
I'm going to close this PR for now - this design pattern was considered and the unit testing and readme requirements were taken into the ongoing work for code driven clusters. IMO, this proof of concept has served its purpose, and this PR will need a pretty significant re-write to work with the new cluster API for code driven clusters. Future authors can use this as a reference, but the changes here are significant enough that this warrants a new PR.

PR #35633, opened 2024-09-17
Added missing close commissioning window cmds

tleacmcsa:
The following test cases open the commissioning window in order to test an operation over PASE, but they do not close the window after the test completes.  This causes subsequent tests that want to do the same thing to fail due to the commissioning window already being open (BUSY).

TC-TBRM-2.4
TC-THNETDIR-2.3
TC-WIFINM-2.2

fixes chip-test-plan issue 4685


bzbarsky-apple (approved):

src/app/tests/suites/certification/Test_TC_TBRM_2_4.yaml
```diff
@@ -84,3 +84,9 @@ tests:
       command: GetPendingDatasetRequest
       response:
           error: UNSUPPORTED_ACCESS
+
+    - label: "Cleanup: Close the commissioning window"
```
bzbarsky-apple, line 88: These steps are present in the test plan, yes?  If not, then a test plan issue is needed to add them there.

tleacmcsa: Please see: https://github.com/CHIP-Specifications/chip-test-plans/pull/4686

cecille (approved):

ReneJosefsen (approved):

PR #35638, opened 2024-09-17
[Fabric-Admin] Update the Fabric Sync guide to explain the confusion

yufengwangca:
1. Received a question about why the local bridge only needs to be added on one side and not the other. Update the documentation to explain this.

2. Decrease the RCP connection retry interval from 5s to 3s.

### DIFF
--- a/docs/guides/fabric_synchronization_guide.md
+++ b/docs/guides/fabric_synchronization_guide.md
@@ -96,6 +96,10 @@ Run the Fabric Synchronization script:
 
 ### Fabric Sync Setup
 
+Fabric Sync Setup is the process of enabling two ecosystems that support the Fabric
+Synchronization feature to commission each other's Fabric Bridge nodes into their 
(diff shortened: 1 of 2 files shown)


tehampson (approved):

docs/guides/fabric_synchronization_guide.md
```diff
@@ -118,6 +118,11 @@ of Ecosystem 1 has successfully paired with Ecosystem 2 on Endpoint 2:
 >>> A new device is added on Endpoint 2.
 ```
 
+Note: We only need to add the local bridge to the ecosystem to trigger the
+Fabric Sync Setup process. In the example above, the Fabric Sync Setup process
```
tehampson, line 122: Does it make sense to describe what "the Fabric Sync Setup process" is because that might also no be clear.

saurabhst (approved):

PR #35644, opened 2024-09-18
Update pigweed to the latest version

andy31415:
We have not updated pigweed for about 6 months due to the dropped support of python 3.9 (and ubuntu 20.04 only had that one).

### Changes

- pigweed update
- WriteLine signature changed for pw_sys_io from `const std::string_view&` to just `std::string_view` 
- SLC generation failed and I am unsure why. Apparently prepending `/usr/bin` to PATH makes it use the right python and I do not fail with a `jinja2 ModuleNotFound` error
- pigweed brings the clang-format and since that version changed, I did a zap regen to update file formatting for checked in generated files
- Boringssl started to fail on `connect.c` on a cast of functions. Unsure why this fails only now, assuming clang is more strict, however that cast seems to be on purpose and borigssl explicitly tries to disable that warning but for MSVC only. I added the disable in the build.gn for boringssl itself.
- added dependency from PlatformConfig to ICD configuration since it looks that there are situations (at least for silabs) where ICD configuration is loaded from platform configuration.
- For OpenIOT added dependency on openiotsdk_buildconfig to the platform buildconfig
- New clang-tidy seems to complain about `-c` (and https://github.com/llvm/llvm-project/pull/101103 says that `-S` is also in the same category). Added code in our clang-tidy runner to auto-skip -c and -S arguments.
- clang-tidy fixes
  - several "bind/sentdo" socket operations that could run on -1, added NOLINTNEXTLINE for those (chages broken in prerequisite PR #35674)
  - fixed some java initialization errors (more things put to nullptr)
  - fixed java dependencies on pw_unit_test that were not declared
  - fixed some seemingly never compiled errors on java tests - questioning if we ever exercise that code ...
  - newer clang requires a template argument for `.template MemberCall(...` calls. Updated tv code to do this.
- fixed typo of `Connnection` to `Connection` (found while fixing java call signatures)


third_party/pigweed/repo
bzbarsky-apple, line 1: Please hold on merging this until I verify that this does not break bootstrap on ARM Macs (unlike previous attempts like https://github.com/project-chip/connectedhomeip/pull/34054

scripts/tools/zap/tests/outputs/all-clusters-app/app-templates/endpoint_config.h
bzbarsky-apple, line 1: Followup: we should see whether we can get the old clang-format behavior here.  The new one is pretty hard to read in practice.

bzbarsky-apple (approved):

scripts/setup/bootstrap.sh
```diff
@@ -133,10 +133,15 @@ _bootstrap_or_activate() {
...
+    local _PIGWEED_CIPD_JSON_ROOT="$_CHIP_ROOT/third_party/pigweed/repo/pw_env_setup/py/pw_env_setup/cipd_setup"
+    local _PIGWEED_CIPD_JSON="$_PIGWEED_CIPD_JSON_ROOT/pigweed.json"
+    local _PYTHON_CIPD_JSON="$_PIGWEED_CIPD_JSON_ROOT/python311.json"
     mkdir -p "$_PW_ACTUAL_ENVIRONMENT_ROOT"
     local _GENERATED_PIGWEED_CIPD_JSON="$_PW_ACTUAL_ENVIRONMENT_ROOT/pigweed.json"
-    $_CHIP_ROOT/scripts/setup/gen_pigweed_cipd_json.py -i $_PIGWEED_CIPD_JSON -o $_GENERATED_PIGWEED_CIPD_JSON
+    $_CHIP_ROOT/scripts/setup/gen_pigweed_cipd_json.py \
+        -i $_PIGWEED_CIPD_JSON                         \
+        -o $_GENERATED_PIGWEED_CIPD_JSON               \
+        -e darwin:$_PYTHON_CIPD_JSON
```
bzbarsky-apple, line 144: Actually, I need to test one more thing.  Will resolve this comment once I do.

scripts/tools/zap/tests/outputs/all-clusters-app/app-templates/endpoint_config.h
andy31415: We could try to figure it out, the best I have is maybe to use cipd to have a fixed version of clang-format pulled (but that is large - it will pull an entire copy of clang).

The issue we have is that we need the restyle to be the same across linux and mac and we currently ensure that by using whatever pigweed pulls ... but pigweed is a moving target here ...

Open to suggestions, however I have no great ones. My feeling is that it is somewhat formatted and we should maybe work on making generated code nicer if we really want readability.

src/inet/TCPEndPointImplSockets.cpp
```diff
@@ -69,9 +69,12 @@ namespace Inet {
 
 CHIP_ERROR TCPEndPointImplSockets::BindImpl(IPAddressType addrType, const IPAddress & addr, uint16_t port, bool reuseAddr)
 {
-    CHIP_ERROR res = GetSocket(addrType);
+    ReturnErrorOnFailure(GetSocket(addrType));
```
andy31415, line 72: These changes are touching internals that are sensitive, so split it to #35674

Will resolve this comment and ready to merge only after that PR as a stand-alone merges.

andy31415: I switched to NOLINT for these items, so while 35674 still is open separately, I believe the changes are now sufficiently safe to close this.

src/credentials/BUILD.gn
```diff
@@ -199,4 +200,6 @@ static_library("test_dac_revocation_delegate") {
     ":credentials",
     jsoncpp_root,
   ]
+
+  # deps = [ dir_pw_unit_test ]
```
jmartinez-silabs, line 204: is it needed and not enabled or was it part of some test?

jmartinez-silabs (approved):
Adding a check for the PW_ENVIRONMENT_ROOT variable gn_silabs_example.sh was not needed in the end?

andy31415:
> Adding a check for the PW_ENVIRONMENT_ROOT variable gn_silabs_example.sh was not needed in the end?

@jmartinez-silabs it is ok in CI because CI on the silabs image does not have a root. it would be required for people testing in the vscode (or any custom local environment that uses it).

src/credentials/BUILD.gn
andy31415: Nice find! ... I thought it was needed however in the end it appears not to be. Will remove. Thank you for spotting this!

scripts/setup/bootstrap.sh
bzbarsky-apple: OK, this looks good to go.

cecille (approved):

cecille:
Q: it looks like there's a fair number of formatting changes. They look reasonably/ better, but is that going to mess stuff up for folks when they build products with zap files that aren't in the SDK?

andy31415:
> Q: it looks like there's a fair number of formatting changes. They look reasonably/ better, but is that going to mess stuff up for folks when they build products with zap files that aren't in the SDK?

The reason for them is that we use clang-format from pigweed when formatting stuff with zap-regen (logic here was to have a clang-format that works on both linux and darwin the same ... and pigweed one is a fixed version).

I don't think others will be affected: this is strictly SDK generated files and all the rest of the files are restyle.io or not at all (whatever others are using)

PR #35649, opened 2024-09-18
[RMP] Update the logs produce by the reliable message manager

vivien-apple:
#### Problem

It is not obvious to know which values are used with **RMP**, both for the local **RMP** values and the remote values.

This PR updates the code such that:
 * Local **RMP** values sent when establishing a **PASE** or **CASE** session are logged
 * Values used when a message is transmitted or retransmitted are logged

For what it worth those logs has allowed me to find an issue (issue # to come)


woody-apple (approved):

woody-apple:
Fast tracking this, given this has had 2 weeks for review, and is blocking some quality tests passing.

PR #35650, opened 2024-09-18
[DNS-SD] Common Resolution Data such as the idle interval, active int…

vivien-apple:
…erval, etc... are ignored by the dnssd platform discovery code when a commissionable node is discovered

#### Problem

By altering the logs (#35649) used with the _ReliableMessageMgr_ I found out that https://github.com/project-chip/connectedhomeip/pull/33025 has regressed the discovery code such that the some of the values advertise by accessories are ignored.

The list of ignored values are:
 * Session Idle Interval
 * Session Active Interval
 * Session Active Threshold
 * TCP Supported
 * Long Idle Time ICD

Concretely it means the the _PBKDFParamRequest_ use to establish a **PASE** session defaults to the default **MRP** values defined in the spec, not the one advertised by the accessory.


bzbarsky-apple (approved):
Yikes, good catch!

jmartinez-silabs (approved):

tcarmelveilleux (approved):

kiel-apple:
updated to latest `master` so it will not bother with the OpenIoT check (all others passed)

PR #35653, opened 2024-09-18
Remove re-declarations of MTRDeviceController public API on MTRDeviceController_Concrete.

bzbarsky-apple:
We should just override whichever selectors we want to override here, instead of re-declaring them all.


jtung-apple (approved):

kiel-apple (approved):

bzbarsky-apple:

PR #35656, opened 2024-09-18
Fix behavior of MTRDevice readAttributePaths if it happens mid-priming.

bzbarsky-apple:
We might not know the values of AttributeList/ServerList/PartsList yet at that point.  Just go ahead and figure out which attribute values we have by walking our own data structures, not the metadata list attribute values.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
@@ -2983,11 +2983,31 @@ static BOOL AttributeHasChangesOmittedQuality(MTRAttributePath * attributePath)
     // Determine the set of what the spec calls "existent paths" that correspond
     // to the request paths.  Building the whole set in-memory is OK, because
     // we're going to need all those paths for our return value anyway.
+    //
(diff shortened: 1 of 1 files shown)


src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
```diff
@@ -2983,11 +2983,31 @@ - (void)writeAttributeWithEndpointID:(NSNumber *)endpointID
     // Determine the set of what the spec calls "existent paths" that correspond
     // to the request paths.  Building the whole set in-memory is OK, because
     // we're going to need all those paths for our return value anyway.
+    //
+    // Note that we don't use the structural attributes (PartsList, ServerList,
+    // AttributeList) to determine this set, because we might be in the middle
+    // of priming right now and not have gotten those yet.  Just use the set of
```
nivi-apple, line 2989:
```suggestion
    // of priming right now and have not gotten those yet.  Just use the set of
```

nivi-apple (approved):

bzbarsky-apple:
Fast-tracking platfom-specific fix with platform owner review.

PR #35660, opened 2024-09-18
Updates comments for the excluded python test scripts based on their current status in the CI pipeline

vatsalghelani-csa:
Work for Issue #31317  

Follow up to PR #35539  

PR 18 - This PR updates individual comments for the excluded python test scripts from glob run script based on their current status in the CI pipeline or project dependencies. The excluded tests include those tests that are not actively executed in CI, depend on unavailable example applications, or are shared utility code for other test cases.

### DIFF
--- a/src/python_testing/execute_python_tests.py
+++ b/src/python_testing/execute_python_tests.py
@@ -97,12 +97,6 @@ def main(search_directory, env_file):
         "test_plan_table_generator.py"  # Code/Test not being used or not shared code for any other tests
     }
 
-    # Explanation for excluded files:
-    # The above list of files are excluded from the tests as they are either not app-specific tests
-    # or are run through a different block of tests mentioned in tests.yaml.
-    # This is to ensure that only relevant test scripts are executed, avoiding redundancy and ensuring
-    # the correctness of the test suite.
-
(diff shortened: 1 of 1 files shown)


src/python_testing/execute_python_tests.py
```diff
@@ -55,65 +55,52 @@ def main(search_directory, env_file):
...
     }
 
-    """
-    Explanation for excluded files:
-    The above list of files are excluded from the tests as they are either not app-specific tests
-    or are run through a different block of tests mentioned in tests.yaml.
-    This is to ensure that only relevant test scripts are executed, avoiding redundancy and ensuring
-    the correctness of the test suite.
-    """
+    # Explanation for excluded files:
```
andy31415, line 99: do we still need the explanation? I would either add it on top of the variable (to comment on the variable) or delete completely as the list has now its own comments for each element.

andy31415 (approved):

src/python_testing/execute_python_tests.py
vatsalghelani-csa: I think deleting completely would avoid any confusion and avoid duplicate explanation too.

arkq (approved):

PR #35661, opened 2024-09-18
Add an ALL_EVENTS maps in symmetry to ALL_ATTRIBUTES

cecille:
Adds an ALL_EVENTS structure to match the attribute and commands.


src/controller/python/chip/clusters/ClusterObjects.py
```diff
@@ -369,6 +370,19 @@ def _cluster_object(cls) -> ClusterObject:
 
 
 class ClusterEvent(ClusterObject):
+    def __init_subclass__(cls, *args, **kwargs) -> None:
+        """Register a subclass."""
+        super().__init_subclass__(*args, **kwargs)
+        try:
+            if cls.cluster_id not in ALL_EVENTS:
```
andy31415, line 377: if we make `ALL_EVENTS = collections.defaultdict(dict)` we could always just directly do `ALL_EVENTS[cls.cluster_id][cls.event_id] = cls`.

Or do we want to be able to get the KeyNotFound error instead of a default empty dict?

cecille: Would that return an empty dict on access too? I don't want any misleading returns here - if the thing doesn't exist and you try to access it, it should fail loudly.

But I honestly just did it this way because it matches the ALL_ATTRIBUTES above.

src/controller/python/chip/clusters/ClusterObjects.py
```diff
@@ -369,6 +370,19 @@ def _cluster_object(cls) -> ClusterObject:
 
 
 class ClusterEvent(ClusterObject):
+    def __init_subclass__(cls, *args, **kwargs) -> None:
```
tcarmelveilleux, line 373: Why in `__init_subclass__` rather than in `__init__` ?

src/controller/python/chip/clusters/ClusterObjects.py
marcelveldt: I'll chime in here as I added the ALL_ATTRIBUTES and ALL_CLUSTERS but imo its more pythonic to just raise on the KeyError than secretly hide things (in this case with a default dict). Not sure if a key lookup still works (so fails if the key doesnt exist) with a defaultdict in place for example.

andy31415 (approved):

src/controller/python/chip/clusters/ClusterObjects.py
```diff
@@ -369,6 +370,19 @@ def _cluster_object(cls) -> ClusterObject:
...
+        """Register a subclass."""
+        super().__init_subclass__(*args, **kwargs)
+        try:
+            if cls.cluster_id not in ALL_EVENTS:
+                ALL_EVENTS[cls.cluster_id] = {}
+            # register this clusterattribute in the ALL_ATTRIBUTES dict for quick lookups
+            ALL_EVENTS[cls.cluster_id][cls.event_id] = cls
+        except NotImplementedError:
```
andy31415, line 381: could we have used a `hasattr` to figure out if cluster_id exists or not? Something like:

```python
if hasattr(cls, 'cluster_id') and hasattr(cls, 'event_id'):
  if cls.cluster_id not in ALL_EVENTS:
    ALL_EVENTS[cls.cluster_id] = {}
  ALL_EVENTS[cls.cluster_id][cls.event_id] = cls  
```

andy31415 (approved):

src/controller/python/chip/clusters/ClusterObjects.py
cecille: These are all codegen'd, so they're supposed to be there. I would actually prefer to fail early if the cluster_id isn't there. Maybe something like
if not hasattr(cls, 'cluster_id') or not hasattr(cls, 'event_id'):
  raise BigObviousFailureThatCodeGenIsBroken

tcarmelveilleux (approved):

src/controller/python/chip/clusters/ClusterObjects.py
cecille: went a different way and removed the try and fixed the unit tests. These should be there on all subclasses or they're not useful.

PR #35662, opened 2024-09-18
[Feature] Adding Step Endpoint Information

antonio-amjr:
Solves: https://github.com/project-chip/certification-tool/issues/367
Backend Dependency PR: https://github.com/project-chip/certification-tool-backend/pull/147 

> [!WARNING]  
> **The frontend changes and images shown here are not official and for experimental purpose only.**

## Description
Updating the Step Start hooks and the related code to include an optional endpoint information. Also, some logging were updated to show the endpoint if it exists.
This change depends on the backend's PR, mentioned above, that receives the new endpoint data and redirect it to the frontend via test UI observables so it may access and show as desired.

Please, mind this current solution will not store the endpoint information in the DB and it will not be retrieved for the UI after loading old executions. Anyway, the logs still points the endpoint on the step completion lines. 

The `TC_ACE_1_3` and `TC_BOOLCFG_2_1` tests were updated to demonstrate how to send the endpoint through the backend hooks. | update to demonstrate how to used the updated hooks to inform the endpoint associated with every step.

 
The images below show a possible usage of the new endpoint information passed to the frontend using the SDK and Backend changes:

Screenshot 2024-09-19 at 10 00 46
<p align="center">The TC ACE 1.3 example</p>

Screenshot 2024-09-19 at 10 02 19
<p align="center">The TC BOOLCFG 2.1 example</p>

### DIFF
--- a/scripts/py_matter_yamltests/matter_yamltests/hooks.py
+++ b/scripts/py_matter_yamltests/matter_yamltests/hooks.py
@@ -161,7 +161,7 @@ class TestRunnerHooks():
         request: TestStep
             The original request as defined by the test step.
         endpoint: int
-            An optional device endpoint the step will target for.
+            An optional device endpoint the step will target.
         """
         pass
 
--- a/scripts/py_matter_yamltests/matter_yamltests/hooks.py
+++ b/scripts/py_matter_yamltests/matter_yamltests/hooks.py
@@ -152,7 +152,7 @@ class TestRunnerHooks():
         """
         pass
 
-    def step_start(self, request: TestStep):
+    def step_start(self, request: TestStep, endpoint: Optional[int] = None):
         """
         This method is called when the runner starts running a step from the test.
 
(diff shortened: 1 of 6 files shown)


scripts/py_matter_yamltests/matter_yamltests/hooks.py
```diff
@@ -152,14 +152,16 @@ def step_skipped(self, name: str, expression: str):
...
+    def step_start(self, request: TestStep, endpoint: Optional[int] = None):
         """
         This method is called when the runner starts running a step from the test.
 
         Parameters
         ----------
         request: TestStep
             The original request as defined by the test step.
+        endpoint: int
+            An optional device endpoint the step will target for.
```
bzbarsky-apple, line 164:
```suggestion
            An optional device endpoint the step will target.
```

andy31415 (approved):

antonio-amjr:
@cecille This PR entered by mistake in the v1.4 but it was meant to enter only in the v1.5 in sync with the Backend's PR https://github.com/project-chip/certification-tool-backend/pull/147 to handle the feature request https://github.com/project-chip/certification-tool/issues/367.
Since I noticed the change was reverted for both master and v1.4, let me know if it's ok to create this PR again for this feature request or if it's necessary to try a different solution.

PR #35673, opened 2024-09-19
[HVAC] Thermostat deadband handling

### LINKED ISSUE #36191: [Feature] When the thermostat supports automatic mode and is in heating or cooling mode, why does it need to meet the requirement that the heating temperature is less than the cooling temperature minus the dead zone?
### Feature description

in thermostat-server.cpp
MatterThermostatClusterServerPreAttributeChangedCallback
`
if (AutoSupported)
{
  if (requested > OccupiedCoolingSetpoint - DeadBandTemp)
      return imcode::InvalidValue;
}
`
The current mode should be determined here, otherwise when the user is in heating or cooling mode, the temperature cannot be adjusted at one time; for example, the heating temperature is set to 26 degrees, and if you want to set the cooling mode, and the temperature is 22 degrees, the user must switch to heating mode, and Lower the temperature, otherwise the cooling temperature will be judged as illegal。
I want to know why Matter designed like this?

Thanks.

### Platform

all

### Platform Version(s)

_No response_

### Anything else?

_No response_

hasty:
This syncs the deadband handling behavior in thermostat-server with the spec; instead of rejecting any setpoint writes which violate the deadband, it attempts to shift the complementary setpoint to preserve the deadband, only erroring if this violates the min/max for the setpoint.

Fixes #36191


src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -220,6 +220,293 @@ int16_t EnforceCoolingSetpointLimits(int16_t CoolingSetpoint, EndpointId endpoin
...
+ *
+ * @param endpoint The endpoint on which the cooling setpoint has been changed
+ * @param currentHeatingSetpointThe current heating setpoint
+ * @param newCoolingSetpoint The newly adjusted cooling setpoint
+ * @param minHeatingSetpoint The minimum allowed cooling setpoint
+ * @param deadband The deadband to preserve
+ * @param setter A function for setting the heating setpoint
+ */
+void ensureHeatingSetpointDeadband(EndpointId endpoint, int16_t currentHeatingSetpoint, int16_t newCoolingSetpoint,
+                                   int16_t minHeatingSetpoint, int16_t deadband, setpointSetter setter)
```
tcarmelveilleux, line 441: `SetpointSetter` type should have capital `S` at the start

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -220,6 +220,293 @@ int16_t EnforceCoolingSetpointLimits(int16_t CoolingSetpoint, EndpointId endpoin
...
+ * @brief Checks to see if it's possible to adjust the cooling setpoint to preserve a given deadband
+ * if the heating setpoint is changed
+ *
+ * @param autoSupported Whether or not the thermostat supports Auto mode
+ * @param newHeatingSetpoint The desired heating setpoint
+ * @param maxCoolingSetpoint The maximum allowed cooling setpoint
+ * @param deadband The deadband to preserve
+ * @return Success if the deadband can be preserved, InvalidValue if it cannot
+ */
+Status checkCoolingSetpointDeadband(bool autoSupported, int16_t newHeatingSetpoint, int16_t maxCoolingSetpoint, int16_t deadband)
```
tcarmelveilleux, line 376: all methods/functions must follow Google-case (`CheckCoolingSetpointDeadband`), not Java case.

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -220,6 +220,293 @@ int16_t EnforceCoolingSetpointLimits(int16_t CoolingSetpoint, EndpointId endpoin
...
+    int16_t AbsMinCoolSetpointLimit;
+    int16_t AbsMaxCoolSetpointLimit;
+    int16_t MinCoolSetpointLimit;
+    int16_t MaxCoolSetpointLimit;
+    int8_t DeadBand      = 0;
+    int16_t DeadBandTemp = 0;
+    int16_t OccupiedCoolingSetpoint;
+    int16_t OccupiedHeatingSetpoint;
+    int16_t UnoccupiedCoolingSetpoint;
+    int16_t UnoccupiedHeatingSetpoint;
```
tcarmelveilleux, line 243: usual convention for struct members is argument names/variable names (e.g. `absMinHeatSetpointLimit`)

hasty: Fixed.

src/app/clusters/thermostat-server/thermostat-server.cpp
hasty: Fixed.

src/app/clusters/thermostat-server/thermostat-server.cpp
hasty: Fixed.

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -535,15 +822,15 @@ void emberAfThermostatClusterServerInitCallback(chip::EndpointId endpoint)
...
-                                                         EmberAfAttributeType attributeType, uint16_t size, uint8_t * value)
+Status MatterThermostatClusterServerPreAttributeChangedCallback(const app::ConcreteAttributePath & attributePath,
+                                                                EmberAfAttributeType attributeType, uint16_t size, uint8_t * value)
 {
     EndpointId endpoint = attributePath.mEndpointId;
     int16_t requested;
 
     // Limits will be needed for all checks
     // so we just get them all now
+    // TODO: use GetSetpointLimits to fetch this information
```
tcarmelveilleux, line 833: Can there be an issue for this TODO?

tcarmelveilleux (approved):

src/app/clusters/thermostat-server/thermostat-server.cpp
hasty: Submitted #35675

src/python_testing/TC_TSTAT_2_3.py
cjandhyala, line 1: there is not test case TSTAT_2.3 defined in the test plan, the code inside this file is for TC_TSTAT_2_2. The test steps in test script are not matching with test plan steps for 2.2 test case.

tcarmelveilleux (approved):

src/python_testing/TC_TSTAT_2_3.py
hasty: Updated to 2.3, which is in https://github.com/CHIP-Specifications/chip-test-plans/pull/4694

tcarmelveilleux: See https://github.com/CHIP-Specifications/chip-test-plans/pull/4694 for test plan

tcarmelveilleux: Let's wait for pan to merge

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -220,6 +220,293 @@ int16_t EnforceCoolingSetpointLimits(int16_t CoolingSetpoint, EndpointId endpoin
...
+
+    if (MaxHeatSetpointLimit::Get(endpoint, &setpointLimits.maxHeatSetpointLimit) != Status::Success)
+        setpointLimits.maxHeatSetpointLimit = setpointLimits.absMaxHeatSetpointLimit;
+
+    if (setpointLimits.coolSupported)
+    {
+        if (OccupiedCoolingSetpoint::Get(endpoint, &setpointLimits.occupiedCoolingSetpoint) != Status::Success)
+        {
+            ChipLogError(Zcl, "Error: Can not read Occupied Cooling Setpoint");
+            return Status::Failure;
```
bzbarsky-apple, line 307: Is there a reason we're not returning the actual status from the Get() here?  Similar in various places below.

If there is a reason (e.g. we don't want to claim UnsupportedAttribute for an attribute write to some attribute that's different from the one being actually gotten here), it would be good to document that reason.

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -220,6 +220,293 @@ int16_t EnforceCoolingSetpointLimits(int16_t CoolingSetpoint, EndpointId endpoin
...
+    }
+    if (minValidCoolingSetpoint > maxCoolingSetpoint)
+    {
+        // Adjusting the cool setpoint to preserve the deadband would violate the max cool setpoint
+        // This should have been caught in CheckCoolingSetpointDeadband, so log and exit
+        ChipLogError(Zcl, "Failed ensuring cooling setpoint deadband");
+        return;
+    }
+    // Adjust the cool setpoint to preserve deadband
+    auto status = setter(endpoint, minValidCoolingSetpoint, MarkAttributeDirty::kYes);
```
bzbarsky-apple, line 422: The kYes value should only be used if there were some previous calls with kNo but now we want to report.

And more generally: why is this using the three-arg form of the setter?  It should use the two-arg form, which would have the benefit of doing the right thing (kIfChanged) with dirty-marking.

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -220,6 +220,293 @@ int16_t EnforceCoolingSetpointLimits(int16_t CoolingSetpoint, EndpointId endpoin
...
+ * @param deadband The deadband to preserve
+ * @param setter A function for setting the heating setpoint
+ */
+void EnsureHeatingSetpointDeadband(EndpointId endpoint, int16_t currentHeatingSetpoint, int16_t newCoolingSetpoint,
+                                   int16_t minHeatingSetpoint, int16_t deadband, SetpointSetter setter)
+{
+    int16_t maxValidHeatingSetpoint = static_cast<int16_t>(newCoolingSetpoint - deadband);
+    if (currentHeatingSetpoint <= maxValidHeatingSetpoint)
+    {
+        // The current cooling setpoint doesn't violate the deadband
```
bzbarsky-apple, line 446:
```suggestion
        // The current heating setpoint doesn't violate the deadband
```

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -220,6 +220,293 @@ int16_t EnforceCoolingSetpointLimits(int16_t CoolingSetpoint, EndpointId endpoin
...
+        return;
+    }
+    if (maxValidHeatingSetpoint < minHeatingSetpoint)
+    {
+        // Adjusting the heating setpoint to preserve the deadband would violate the max cooling setpoint
```
bzbarsky-apple, line 451:
```suggestion
        // Adjusting the heating setpoint to preserve the deadband would violate the min heating setpoint
```

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -220,6 +220,293 @@ int16_t EnforceCoolingSetpointLimits(int16_t CoolingSetpoint, EndpointId endpoin
...
+    // Adjust the heating setpoint to preserve deadband
+    auto status = setter(endpoint, maxValidHeatingSetpoint, MarkAttributeDirty::kYes);
+    if (status != Status::Success)
+    {
+        ChipLogError(Zcl, "Error: EnsureHeatingSetpointDeadband failed!");
+    }
+}
+
+/**
+ * @brief For thermostats that support auto, shift endpoints to maintain the current deadband
```
bzbarsky-apple, line 465:
```suggestion
 * @brief For thermostats that support auto, shift setpoints to maintain the current deadband
```

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -220,6 +220,293 @@ int16_t EnforceCoolingSetpointLimits(int16_t CoolingSetpoint, EndpointId endpoin
...
+{
+    if (!autoSupported)
+    {
+        return Status::Success;
+    }
+    int16_t maxValidHeatingSetpoint = static_cast<int16_t>(newCoolingSetpoint - deadband);
+    if (maxValidHeatingSetpoint < minHeatingSetpoint)
+    {
+        // If we need to adjust the heating setpoint to preserve the deadband, it will go below the min heat setpoint
+        return Status::InvalidValue;
```
bzbarsky-apple, line 360: So where does the spec say to return InvalidValue in this case?  Here's what I see in the spec, for example for the OccupiedCoolingSetpoint attribute:

> If an attempt is made to set this attribute to a value greater than MaxCoolSetpointLimit or less than MinCoolSetpointLimit, a response with the status code CONSTRAINT_ERROR SHALL be returned.
>
> If this attribute is set to a value that is less than (OccupiedHeatingSetpoint + MinSetpointDeadBand), the value of OccupiedHeatingSetpoint SHALL be adjusted to (OccupiedCoolingSetpoint - MinSetpointDeadBand).

That second sentence does not seem to envision what happens when the value of OccupiedHeatingSetpoint cannot thus be adjusted.  But that first items seems to suggest that trying to do the adjustment should result in ConstraintError, no?

src/app/clusters/thermostat-server/thermostat-server.cpp
hasty: The generated code made it seem like the three-arg form was necessary to trigger a notification; switched to two-arg.

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -220,6 +220,293 @@ int16_t EnforceCoolingSetpointLimits(int16_t CoolingSetpoint, EndpointId endpoin
...
+    int16_t absMinHeatSetpointLimit;
+    int16_t absMaxHeatSetpointLimit;
+    int16_t minHeatSetpointLimit;
+    int16_t maxHeatSetpointLimit;
+    int16_t absMinCoolSetpointLimit;
+    int16_t absMaxCoolSetpointLimit;
+    int16_t minCoolSetpointLimit;
+    int16_t maxCoolSetpointLimit;
+    int8_t deadBand      = 0;
+    int16_t deadBandTemp = 0;
```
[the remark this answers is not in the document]
hasty: Switched to a single int16_t.

src/app/clusters/thermostat-server/thermostat-server.cpp
bzbarsky-apple: @hasty my point is that if it's important to not propagate the return value, that should be documented in the code, so someone does not "fix" it later.

src/app/clusters/thermostat-server/thermostat-server.cpp
bzbarsky-apple: InvalidValue is the old, deprecated name...  But OK, maybe a followup to switch it globally here?

bzbarsky-apple (approved):

src/app/clusters/thermostat-server/thermostat-server.cpp
hasty: Added comments to the lines that do this.

src/app/clusters/thermostat-server/thermostat-server.cpp
hasty: OK, will leave for a follow-up PR.

PR #35674, opened 2024-09-19
Additional socket checks for socket inet implementations

andy31415:
#35644 pulled in a new pigweed which complains that bind/sendmsg should never be passed in negative arguments.

clang-tidy is unable to validate that GetSocket() is either an error or mSocket is valid. It thinks it can return something that `IsSuccess` while leaving mSocket invalid.

As a result, I added several `NOLINTNEXTLINE` entries.

### DIFF
--- a/src/inet/TCPEndPointImplSockets.cpp
+++ b/src/inet/TCPEndPointImplSockets.cpp
@@ -69,7 +69,7 @@ namespace Inet {
 
 CHIP_ERROR TCPEndPointImplSockets::BindImpl(IPAddressType addrType, const IPAddress & addr, uint16_t port, bool reuseAddr)
 {
-  CHIP_ERROR res = GetSocket(addrType);
+    CHIP_ERROR res = GetSocket(addrType);
 
     if (res == CHIP_NO_ERROR && reuseAddr)
     {
@@ -809,7 +809,7 @@ void TCPEndPointImplSockets::HandlePendingIO(System::SocketEvents events)
 #else  // __MBED__
        // On Mbed OS, connect blocks and never returns EINPROGRESS
        // The socket option SO_ERROR is not available.
-            int osConRes     = 0;
+            int osConRes = 0;
 #endif // !__MBED__
             CHIP_ERROR conRes = CHIP_ERROR_POSIX(osConRes);
 
--- a/src/inet/TCPEndPointImplSockets.cpp
+++ b/src/inet/TCPEndPointImplSockets.cpp
@@ -162,6 +162,9 @@ CHIP_ERROR TCPEndPointImplSockets::ConnectImpl(const IPAddress & addr, uint16_t 
 
     ReturnErrorOnFailure(GetSocket(addrType));
 
+    // have to have a valid socket to send something with. Generally GetSocket ensures that
+    VerifyOrReturnValue(mSocket >= 0, CHIP_ERROR_INCORRECT_STATE);
+
     if (!intfId.IsPresent())
     {
         // The behavior when connecting to an IPv6 link-local address without specifying an outbound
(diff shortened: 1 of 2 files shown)


src/inet/UDPEndPointImplSockets.cpp
```diff
@@ -93,6 +93,8 @@ namespace {
 
 CHIP_ERROR IPv6Bind(int socket, const IPAddress & address, uint16_t port, InterfaceId interface)
 {
+    VerifyOrReturnValue(socket >= 0, CHIP_ERROR_INVALID_ARGUMENT);
```
bzbarsky-apple, line 96: Again, this is only called if GetSocket succeeds, which guarantees that socket >= 0, no?

Does it help if GetSocket() checks for < 0, instead of just checking for == -1, after doing `::socket()`?

If not, can we restructure this stuff somehow (e.g. have the socket be an `optional<unsigned>` instead of `int` so that it's crystal-clear that if we have a value it is in fact not negative?).

src/inet/TCPEndPointImplSockets.cpp
```diff
@@ -69,9 +69,12 @@ namespace Inet {
 
 CHIP_ERROR TCPEndPointImplSockets::BindImpl(IPAddressType addrType, const IPAddress & addr, uint16_t port, bool reuseAddr)
 {
-    CHIP_ERROR res = GetSocket(addrType);
+    ReturnErrorOnFailure(GetSocket(addrType));
+
+    // need to be in a connected state to be able to bind
+    VerifyOrReturnError(mSocket >= 0, CHIP_ERROR_INCORRECT_STATE);
```
bzbarsky-apple, line 75: But here's the thing: GetSocket already enforces that invariant, as it should.

andy31415: Yes, the code was always correct. The reason I have this is that clang-tidy cannot check for these guarantees. It looks like this:

image

So I can either tell clang-tidy to ignore (I can add nolint entries throughout) or I can add extra and redundant checks. Since this is a `sockets` implementation, it should be for a posix-like system so I was less worried about flash space, hence extra redundant checks seemed ok.

andy31415: And the explanation is clearly wrong: we return CHIP_ERROR_POSIX and then tidy says "what if we have IsSuccess returns true on it" which are not compatible ... but tidy cannot know because IsSuccess is some external function that in theory could change.

So the fact that this is correct is our internal knowledge. Either I do checks or NOLINT.

src/inet/UDPEndPointImplSockets.cpp
andy31415: negative check does not help: tidy knows it is exactly -1.
What clang-tidy says that "if we return a CHIP_ERROR_POSIX(errno) and IsSuccess returns true on that, then error occurs". However this is incompatible since CHIP_ERROR will not be a success, it is just clang-tidy assumes IsSuccess probably can change over time and cannot infer the CHIP_ERROR_POSIX is never a success in our usage.

andy31415: It would be fixable if GetSocket would return some `variant<CHIP_ERROR, unsigned>` and we assign that to mSocket. But that starts being odd ... we are re-writing the code a lot.

src/inet/TCPEndPointImplSockets.cpp
bzbarsky-apple: I would suggest NOLINT with comment and a followup issue to restructure this code to make the invariants clearer?

pidarped: I would suggest NOLINT as well here. But then, if clang-tidy cannot interpret ReturnErrorOnFailure() correctly, this should be a more widespread problem with it, isn't it?

andy31415: Ok, switched to NOLINT.

Regarding clang-tidy, I believe it will only be when clang can detect specific code usages that it thinks it is wrong. The tool still has limited static analyzer checking, so our code base does not fully blow up.

bzbarsky-apple (approved):

pidarped (approved):

PR #35678, opened 2024-09-19
Separate out AttributeValueEncoder/Decoder test boilerplate helper logic into separate library

andy31415:
It seems very useful to make operations that use AttributeValueEncoder and AttributeValueDecoder simple to use in tests.

This extracts out `ReadOperation` and `TestOperation` for testing `DataModel::Provider` code. This will be useful for building separate code-driven models and their unit testing logic.


tcarmelveilleux (approved):

src/app/data-model-provider/tests/BUILD.gn
```diff
@@ -30,3 +30,30 @@ chip_test_suite("tests") {
...
+source_set("encode-decode") {
+  # Handles the conversions between Raw data (usually byte buffers backed by TLV encoders
+  # and `*IBs::Builder` classes) and data used by by Providers like AttributeValueEncoder
+  # or AttributeValueDecoder.
+  #
+  # Intended to make writing tests simpler and focusing less on encoding and more on
+  # "Assume you receive an int" or "Check that the encoded value was the string 'abc'"
+  # and so on
+  sources = [
+    "ReadTesting.cpp",
```
jmartinez-silabs, line 43: nit: all test files are usually called TestXxxx.cpp

jmartinez-silabs (approved):

src/app/data-model-provider/tests/BUILD.gn
andy31415: In this particular case, these are support files, so a file `TestFoo` would include/use `ReadTesting` to test `DataModel::Provider::Read` code.

I think having a `Test` prefix would make these look like actual tests and they are not.

PR #35683, opened 2024-09-19
Move initWithFactory: and startup: declarations to MTRDeviceController_Concrete.

bzbarsky-apple:
initWithFactory: and startup: are only called from _startDeviceController:, which was only being called with MTRDeviceController_Concrete instances.

Also fixes the argument type declarations for _startDeviceController: and
initializeController: to make the types compile-time enforced.

The MTRDeviceController implementations of initWithFactory: and startup: are now obviously unreachable and can be removed.

Some static variables that are now unused have to be removed so the compiler doesn't complain about them.

Further cleanup of now-unused things will follow in separate PRs.


kiel-apple (approved):

bzbarsky-apple:

PR #35685, opened 2024-09-19
MTRDeviceControllerOverXPC should inherit from MTRDeviceController_Concrete for now.

bzbarsky-apple:
That's what it used to do, and for now we should just leave it be.  Longer-term, we need to figure out what should happen with MTRDeviceControllerOverXPC; whether it should inherit from the base MTRDeviceController, or just not exist, or what.


kiel-apple (approved):

bzbarsky-apple:

PR #35689, opened 2024-09-20
Remove unreachable shutdown code from MTRDeviceController.

bzbarsky-apple:
cleanupAfterStartup is no longer reachable and can be removed.

At that point, the only caller of controllerShuttingDown: on the factory is MTRDeviceController_Concrete, so we can go ahead and change the argument type accordingly.

At that point it becomes obvious that shutDownCppController and deinitFromFactory are only called on MTRDeviceController_Concrete instances, so those can move from MTRDeviceController_Internal to MTRDeviceController_Concrete, and the implementations in MTRDeviceController can be removed.

checkForInitError is now unused on MTRDeviceController, and can be removed. And then "cleanup" is unused and can be removed.


kiel-apple (approved):

bzbarsky-apple:

PR #35692, opened 2024-09-20
Work on TC-IDM-3-2

austina-csa:
Only first test step is implemented for now but more will be added.

This approach is done for efficiency so that errors with implementing steps can be caught quickly.


src/python_testing/TC_IDM_3_2.py
```diff
@@ -0,0 +1,182 @@
...
+
+        # Step 1
+
+        # TH selects a writeable attribute that is present on at least one endpoint.
+        # If no such attribute exists, skip this step. TH reads the attribute value and saves as `original`. TH sends a WriteRequestMessage to the DUT to write to the attribute on one endpoint.
+        #  It should set a value that is different than `original`.
+
+        # Verify that the DUT sends a WriteResponseMessage with any status except UNSUPPORTED_WRITE or DATA_VERSION_MISMATCH. If the Status is SUCCESS, verify the updated
+        #  value by sending a ReadRequestMessage for all affected paths. If the status is SUCCESS, send a WriteRequestMessage to set the value back to `original`.
+        self.user_params["use_pase_only"] = False
```
cecille, line 21: instead of overwriting this, set allow_pase to false in the setup_class_helper if you want to disallow PASE. use_pase_only was also removed recently, so this won't do anything anyway.

src/python_testing/TC_IDM_3_2.py
```diff
@@ -0,0 +1,182 @@
...
+        super().setup_class()
+        await self.setup_class_helper()
+        self.xml_clusters, self.problems = build_xml_clusters()
+        writable_attributes = []
+        for cluster_id in self.xml_clusters:
+            xml_cluster = self.xml_clusters[cluster_id]
+            attributes = xml_cluster.attributes
+            for attribute_id in attributes:
```
cecille, line 29:
```suggestion
            for attribute_id, xml_attribute in attributes.items():

```

src/python_testing/TC_IDM_3_2.py
```diff
@@ -0,0 +1,182 @@
...
+                xml_attribute = attributes[attribute_id]
```
cecille, line 30:
```suggestion

```

src/python_testing/TC_IDM_3_2.py
```diff
@@ -0,0 +1,182 @@
...
+                write_access = xml_attribute.write_access
+                if write_access:
```
cecille, line 32: can you flip this around to un-nest this a bit?

ie
if not write_access:
  continue

src/python_testing/TC_IDM_3_2.py
```diff
@@ -0,0 +1,182 @@
...
+                    if cluster_id in Clusters.ClusterObjects.ALL_ATTRIBUTES:
```
cecille, line 33:
```suggestion
                    if cluster_id in Clusters.ClusterObjects.ALL_ATTRIBUTES and attribute_id in Clusters.ClusterObjects.ALL_ATTRIBUTES[cluster_id]:

```
^ that should let you eliminate the try loop below.

src/python_testing/TC_IDM_3_2.py
```diff
@@ -0,0 +1,182 @@
...
+                        try:
+                            if hasattr(Clusters.ClusterObjects.ALL_ATTRIBUTES[cluster_id][attribute_id], 'value'):
+                                writable_attributes.append((cluster_id, attribute_id, write_access))
+                        except KeyError: # Some XML clusters aren't present in ALL_CLUSTERS or ALL_ATTRIBUTES
+                            pass
+        chosen_writable_attribute_tuple = writable_attributes[0]
```
cecille, line 39: need to ensure this attribute is present on the device first.

src/python_testing/TC_IDM_3_2.py
```diff
@@ -0,0 +1,182 @@
...
+            xml_cluster = self.xml_clusters[cluster_id]
+            attributes = xml_cluster.attributes
+            for attribute_id in attributes:
+                xml_attribute = attributes[attribute_id]
+                write_access = xml_attribute.write_access
+                if write_access:
+                    if cluster_id in Clusters.ClusterObjects.ALL_ATTRIBUTES:
+                        try:
+                            if hasattr(Clusters.ClusterObjects.ALL_ATTRIBUTES[cluster_id][attribute_id], 'value'):
+                                writable_attributes.append((cluster_id, attribute_id, write_access))
```
cecille, line 36: instead of saving the IDs here, why not just save the attribute class? it has the attribute ids and the cluster ids noted.

cecille, line 36: do you end up using the write_access for anything? The TH has admin, so if it's writable, it can be written by the TH.

src/python_testing/TC_IDM_3_2.py
```diff
@@ -0,0 +1,182 @@
...
+                        except KeyError: # Some XML clusters aren't present in ALL_CLUSTERS or ALL_ATTRIBUTES
+                            pass
+        chosen_writable_attribute_tuple = writable_attributes[0]
+        chosen_writable_attribute = Clusters.ClusterObjects.ALL_ATTRIBUTES[chosen_writable_attribute_tuple[0]][chosen_writable_attribute_tuple[1]]
+        chosen_writable_cluster = Clusters.ClusterObjects.ALL_CLUSTERS[chosen_writable_attribute_tuple[0]]
+        
+        output_1 = await self.default_controller.ReadAttribute(self.dut_node_id, [chosen_writable_attribute])
```
cecille, line 43: It might actually make your life easier to use the Read( function instead of read attribute - that function returns an object that stores the returned values by ID rather than forcing you to pass in the class object.

src/python_testing/TC_IDM_3_2.py
```diff
@@ -0,0 +1,182 @@
...
+        endpoint = next(iter(output_1))
+        await self.default_controller.WriteAttribute(self.dut_node_id, [(endpoint, chosen_writable_attribute(value=123456))])
```
cecille, line 45: This only works if you happen to hit a writable attribute that takes an int. Here, you want to look at attribute_type and pick a value based on the ClusterObjectFieldDescriptor

src/python_testing/TC_IDM_3_2.py
```diff
@@ -0,0 +1,182 @@
...
+        output_3 = await self.default_controller.ReadAttribute(self.dut_node_id, [chosen_writable_attribute])
+        
+        asserts.assert_not_equal(output_1[endpoint][chosen_writable_cluster][chosen_writable_attribute],
```
cecille, line 48: will only work if the write returned success.

src/python_testing/TC_IDM_3_2.py
```diff
@@ -0,0 +1,182 @@
...
+                                 output_3[endpoint][chosen_writable_cluster][chosen_writable_attribute],
+                                 "Output did not change")
+
```
cecille, line 51: need to add the cleanup step to return the attribute to its original value.

src/python_testing/TC_IDM_3_2.py
austina-csa: Line is now removed

src/python_testing/TC_IDM_3_2.py
austina-csa: Fixed

src/python_testing/TC_IDM_3_2.py
austina-csa: Changed to only save the attribute class

austina-csa: Nope, and this was removed

src/python_testing/TC_IDM_3_2.py
austina-csa: Added a check for this case

src/python_testing/TC_IDM_3_2.py
austina-csa: Changed to use Read()

src/python_testing/TC_IDM_3_2.py
austina-csa: Added a function (pick_writable_value) to select a value dynamically

src/python_testing/TC_IDM_3_2.py
austina-csa: Added a cleanup step by reading the initial value, saving it, and writing it back at the end of the step

PR #35697, opened 2024-09-20
Update usage of Event logic to be routable through DataModel::Provider

andy31415:
This is looking to decouple more of InteractionModelEngine from ember internals.

In this particular case, a significant number of ember calls were made for event processing (and these were not in ember-compatibility-functions.cpp)

### Changes

- added a new member for event path processing in DataModel::Provider
- Updated unit tests with a mock on this (just returns true, not actually checking clusters)
- Shared implementation via `app/ember_coupling/EventPathValidity.mixin.h`
  - one change here is to forward a `const EmberAfCluster *` which is already available to a `ClusterSupportsEvent` method instead of using the ember compatibility call of `CheckEventSupportStatus` which re-searches for endpoint and cluster (that is already done). This is to reduce coupling (no need for ember-compat) and also minor speed improvement to remove some metadata lookups.

### Testing

- Local compilation with data model enable/disable
- CI should validate at least some paths. Code paths are identical


src/app/codegen-data-model-provider/CodegenDataModelProvider_EventPathReadable.cpp
```diff
@@ -0,0 +1,169 @@
...
+#include <lib/core/DataModelTypes.h>
+
+namespace chip {
+namespace app {
+namespace {
+
+using namespace chip::Access;
+
+#if !CHIP_CONFIG_ENABLE_EVENTLIST_ATTRIBUTE
+bool CanAccessEvent(const Access::SubjectDescriptor & subjectDescriptor, const ConcreteClusterPath & path,
```
bzbarsky-apple, line 34: So this id duplicating a bunch of code from InteractionModelEngine.cpp.  Why are we doing that?  It seems to me like this code should just live here, the "checked" mode should call into the checked data model provider, and that should compare the codegen and non-codegen provider match, or something, no?

src/app/data-model-provider/Provider.h
```diff
@@ -57,6 +59,15 @@ class Provider : public ProviderMetadataTree
     // event emitting, path marking and other operations
     virtual InteractionModelContext CurrentContext() const { return mContext; }
 
+    /// Validates that the given event path is supported, where path may contain wildcards.
```
bzbarsky-apple, line 62: That's not what it validates.  It validates that expanding the event path includes at least one concrete event path that the subject descriptor has access to, in theory.

An event path can be perfectly "valid" but not accessible by the subject.

src/app/data-model-provider/Provider.h
```diff
@@ -57,6 +59,15 @@ class Provider : public ProviderMetadataTree
...
+    ///
+    /// If any wild cards exist on the given path, the implementation is expected to validate
+    /// that a valid event path exists on some wildcard expansion.
+    ///
+    /// At the very minimum this will validate that a valid endpoint/cluster can be expanded
+    /// from the input path and that the given descriptor has access to it.
+    virtual bool EventPathReadable(const EventPathParams & path, const Access::SubjectDescriptor & descriptor) = 0;
```
bzbarsky-apple, line 69: This needs a better name.  How about EventPathIncludesAccessibleConcretePath or something?

src/app/codegen-data-model-provider/CodegenDataModelProvider_EventPathReadable.cpp
andy31415: I expect to eventually delete the bits that are inside interactionmodelengine - those parts should have never contained direct ember calls. My understanding is that ember-compatibility-functions.cpp was the bridge in functionality, however ember calls then escaped because they were just available.

I cannot really reuse since I specifically want to separate out dependencies and depending on "codegen" is not separating things out. Because this couples "access" and ember calls, I cannot place it in ember (i.e. util) either.

For now I suggest keeping the duplication because I fully expect to delete the rest of the code later. Alternatively to deduplicate I would need to create some separate library only to pull it back later into codegen, which seems a pain.

src/app/data-model-provider/Provider.h
andy31415: True ... need to work more on naming and description ... will iterate on this.

src/app/codegen-data-model-provider/CodegenDataModelProvider_EventPathReadable.cpp
bzbarsky-apple: > since I specifically want to separate out dependencies and depending on "codegen" is not separating things out

I feel like I am missing something here.  For the "checked" bits, I had forgotten that they were done in a bespoke way instead of having an interface implementation that fans out to two other interface instances without depending on their implementations.  

But for this specific case where we are literally copy/pasting the code and there is nothing complicated about the API contract (you pass in path and subject, you get back a boolean), it seems like what we should do is basically always use the provider, and just set it to the codegen provider for now as needed, with the code moving into there.  The value of the "checked" mode seems very low to me in this case.

And I don't think we should cart around and maintain two copies of this code.  I know you plan to delete one of the copies, but there is no clear timeframe for that and in the meantime we have to maintain both copies.

andy31415: Discussed offline, will share via `#include` for now. I believe long term when/if we remove old code, these shared includes should be pulled back into codegen data provider.

src/app/ember_coupling/EventPathValidity.mixin.h
```diff
@@ -0,0 +1,145 @@
...
+ *    limitations under the License.
+ */
+
+// This is intended as a TOP LEVEL INCLUDE inside code that uses it
+
+namespace chip {
+namespace app {
+namespace {
+
+bool EventAccessible(const EmberAfCluster * cluster, EventId eventId)
```
bzbarsky-apple, line 25: No, this function is testing for existence, not accessibility....

Also, this is duplicating CheckEventSupportStatus.  Can we at least document that once this stops being used in multiple places we should have only one place that does _that_ work?  Better yet, again, would be to only have on implementation of this bit....

andy31415: Will rename. 

Regarding duplication, true, however in this case the reuse seems somewhat odd because `CheckEventSupportStatus` re-finds the endpoint and cluster (which is already done at the time this code is needed) so I would rather not try to "fix" that.

Overall the `CheckEventSupportStatus` seems a bit ember-tied, but maybe I will need it after all in some fashion if I am to decouple ember. I would handle that separately in future PRs.

andy31415: Renamed to `ClusterSupportsEvent` since this seems to be the intent ... not sure about "supports" .... but also not sure about better names (contains? emits ? we cannot actually be sure since this is metadata .... mayemitevent?).

In the end I am not sure this is as important naming wise given that this is an anon namespace method and not that public of an API.

andy31415: Added a comment as well regarding this being similar to CheckEventSupportStatus.
I hope we can somehow describe `CheckEventSupportStatus` in some better way at some point.

Also if event list is provisional and we end up dropping, maybe some complexity can disappear...

src/app/ember_coupling/EventPathValidity.mixin.h
```diff
@@ -0,0 +1,155 @@
...
+
+namespace chip {
+namespace app {
+namespace {
+
+/**
+ * Find out if the given EventId is reported as supported by the given cluster
+ * within its metadata. If cluster has no event metadata (i.e. no event list
+ * support is available), clusters are assumed to support any event as there is
+ * no way to actually tell.
```
tcarmelveilleux, line 29: What if I don't have a partial event list in ZAP, but I actually emit more events in code. Is that a problem?

andy31415: Current logic is that we just allow any event to be emitted, it does not matter if it is in zap list or not.

andy31415: Resolving this: current behavior is currently preserved and we actually have no validation that we only emit events in the list.

Long term it seems we may plan to migrate away from having the list in the first place as it cannot really be validated/enforced. EventList is provisional already.

tcarmelveilleux (approved):

bzbarsky-apple (approved):

PR #35700, opened 2024-09-20
Change output folder for apps

rquidute:
### What changed
The goal of this PR is to have all apps inside apps folder.
The way to access the apps will remain the same, since the apps will have a symbolic link to the apps inside apps folder.

### Related Issue
https://github.com/project-chip/certification-tool/issues/403


ccruzagralopes (approved):

antonio-amjr (approved):

integrations/docker/images/chip-cert-bins/Dockerfile
```diff
@@ -258,30 +258,35 @@ RUN apt-get update -y
...
+COPY --from=chip-build-cert-bins /root/connectedhomeip/out/chip-all-clusters-minimal-app apps/chip-all-clusters-minimal-app
+COPY --from=chip-build-cert-bins /root/connectedhomeip/out/chip-lighting-app apps/chip-lighting-app
+COPY --from=chip-build-cert-bins /root/connectedhomeip/out/chip-tv-casting-app apps/chip-tv-casting-app
+COPY --from=chip-build-cert-bins /root/connectedhomeip/out/chip-tv-app apps/chip-tv-app
+COPY --from=chip-build-cert-bins /root/connectedhomeip/out/chip-bridge-app apps/chip-bridge-app
+COPY --from=chip-build-cert-bins /root/connectedhomeip/out/thermostat-app apps/thermostat-app
+COPY --from=chip-build-cert-bins /root/connectedhomeip/out/chip-ota-provider-app apps/chip-ota-provider-app
+COPY --from=chip-build-cert-bins /root/connectedhomeip/out/chip-ota-requestor-app apps/chip-ota-requestor-app
+COPY --from=chip-build-cert-bins /root/connectedhomeip/out/chip-lock-app apps/chip-lock-app
+COPY --from=chip-build-cert-bins /root/connectedhomeip/out/chip-app1 apps/chip-app1
```
rquidute, line 276:
```suggestion
COPY --from=chip-build-cert-bins /root/connectedhomeip/out/chip-app1 apps/chip-app1
```

integrations/docker/images/chip-cert-bins/Dockerfile
```diff
@@ -258,30 +258,35 @@ RUN apt-get update -y
...
+COPY --from=chip-build-cert-bins /root/connectedhomeip/out/chip-microwave-oven-app apps/chip-microwave-oven-app
+COPY --from=chip-build-cert-bins /root/connectedhomeip/out/chip-rvc-app apps/chip-rvc-app
+COPY --from=chip-build-cert-bins /root/connectedhomeip/examples/fabric-admin/scripts/fabric-sync-app.py apps/fabric-sync-app
+COPY --from=chip-build-cert-bins /root/connectedhomeip/out/fabric-bridge-app apps/fabric-bridge-app
+COPY --from=chip-build-cert-bins /root/connectedhomeip/out/fabric-admin apps/fabric-admin
+COPY --from=chip-build-cert-bins /root/connectedhomeip/out/chip-lighting-data-model-no-unique-id-app apps/chip-lighting-data-model-no-unique-id-app
+COPY --from=chip-build-cert-bins /root/connectedhomeip/out/matter-network-manager-app apps/matter-network-manager-app
+
+# Create symbolic link to apps
+RUN ln -s apps/* .
```
andy31415, line 288: Is this temporary (if so should there be an issue tagged to it) or do we want links always? Generally I find it odd to have duplicate paths, so maybe some comments explaining either way would help .

rquidute: There is a script in TH that copies all the apps from image to host, so the initial rule was copying the app that starts with `chip-`, but recently some apps does not start with `chip-`, so we are changing this TH script in a way to add the new apps that don't start with `chip-`. So we intended to create apps folder inside image and out script should copy all apps from there. So we created the symbolic link to keep the way the user manually runs the app (at root folder).  
One possible solution, is to avoid running apps from root folder, and from now on, the user should run  apps from apps folder.

The suggested comment, in case, we are going to keep the symbolic links could be this:

> All apps are now in apps folder, so this way it will allow user to run apps from root.

What do you think @andy31415 ?

andy31415: I would place a summary of this comment above this ln logic. Thinking something like:

```
# Create symbolic links for now since this allows users to use existing configurations
# for running just `app-name` instead of `apps/app-name`
```

At the same time I think it seems more long term maintainable to have only one way of managing things and if we are changing to `apps` we should just do that and not allow 2 ways of doing things.

andy31415 (approved):

integrations/docker/images/chip-cert-bins/Dockerfile
```diff
@@ -258,30 +258,35 @@ RUN apt-get update -y
...
+COPY --from=chip-build-cert-bins /root/connectedhomeip/out/chip-energy-management-app apps/chip-energy-management-app
+COPY --from=chip-build-cert-bins /root/connectedhomeip/out/chip-microwave-oven-app apps/chip-microwave-oven-app
+COPY --from=chip-build-cert-bins /root/connectedhomeip/out/chip-rvc-app apps/chip-rvc-app
+COPY --from=chip-build-cert-bins /root/connectedhomeip/examples/fabric-admin/scripts/fabric-sync-app.py apps/fabric-sync-app
+COPY --from=chip-build-cert-bins /root/connectedhomeip/out/fabric-bridge-app apps/fabric-bridge-app
+COPY --from=chip-build-cert-bins /root/connectedhomeip/out/fabric-admin apps/fabric-admin
+COPY --from=chip-build-cert-bins /root/connectedhomeip/out/chip-lighting-data-model-no-unique-id-app apps/chip-lighting-data-model-no-unique-id-app
+COPY --from=chip-build-cert-bins /root/connectedhomeip/out/matter-network-manager-app apps/matter-network-manager-app
+
+# Create symbolic link to apps
```
rquidute, line 287:
```suggestion
# Create symbolic links for now since this allows users to use existing configurations
# for running just `app-name` instead of `apps/app-name`
```

integrations/docker/images/chip-cert-bins/Dockerfile
rquidute: Thanks @andy31415. Let's keep the comment you suggested for while, and then afterwards we remove the symbolic links creation.

Issue #35670, opened 2024-09-19
[CERT-TEST-FAILURE] [TC-IDM-10.3] Several ClusterRevision values are not in sync with the spec


bauerschwan:
### Feature Area

Other

### Test Case

TC-IDM-10.3

### Reproduction steps

When running TC-IDM-10.3 the following errors occur:
`
[MatterTest] 09-19 10:54:04.243 INFO
Problem: ProblemSeverity.ERROR
    test_name: IDM-10.3
    location:
       Endpoint: 0,
       Cluster:  29 (0x1d) Descriptor
      Attribute:65533 (0xfffd)
    problem: Revision found on cluster (2) does not match revision listed in the spec (3)
    spec_location:

[MatterTest] 09-19 10:54:04.243 INFO
Problem: ProblemSeverity.ERROR
    test_name: IDM-10.3
    location:
       Endpoint: 0,
       Cluster:  31 (0x1f) AccessControl
      Attribute:65533 (0xfffd)
    problem: Revision found on cluster (1) does not match revision listed in the spec (2)
    spec_location:

[MatterTest] 09-19 10:54:04.243 INFO
Problem: ProblemSeverity.ERROR
    test_name: IDM-10.3
    location:
       Endpoint: 0,
       Cluster:  49 (0x31) NetworkCommissioning
      Attribute:65533 (0xfffd)
    problem: Revision found on cluster (2) does not match revision listed in the spec (3)
    spec_location:
`

These clusters have hard-coded ClusterRevision in the code and are not yet updated to the spec.

### Bug prevalence

Always

### GitHub hash of the SDK that was being used

17b1a38e909e7874593bcb87c31be03a5866f1d4

### Platform

nrf

bauerschwan:
See also https://github.com/project-chip/connectedhomeip/issues/34227

cecille:
Please see https://github.com/project-chip/connectedhomeip/pull/35607 - spec changes for 0.9 landed after the SVE cut, so the data model files didn't get updated. The in-progress ones should work, you will need to pull this change because the TH doesn't have an override.

cjandhyala:
closing the issue as the PR is merged. if you think this issue is still open, pls re-open it @bauerschwan

PR #35710, opened 2024-09-21
ACL: Fix cluster revision for MACL

cecille:
Cluster revision was updated in the spec when MACL was added, SDK was not.


tcarmelveilleux (approved):

bauerschwan:
What about the other two clusters mentioned in https://github.com/project-chip/connectedhomeip/issues/35670? Are they covered by side-loading the TH files mentioned there?

yunhanw-google (approved):

tleacmcsa (approved):

jmartinez-silabs (approved):

PR #35712, opened 2024-09-22
TC-OPSTATE-2.5: Force expire session on reboot

cecille:
If the DUT has rebooted, the TH will need to establish the session again, which means expiring the current. In a real product, a fail on the case would trigger a retry, but it's not clear that we want that in the TH because it can paper over problems where the DUT is crashing. This is an intentional reboot, wihch is different.

Probably not caught in CI / QA because the CI app doesn't actually reboot.

### DIFF
--- a/src/python_testing/TC_OpstateCommon.py
+++ b/src/python_testing/TC_OpstateCommon.py
@@ -1112,10 +1112,11 @@ class TC_OPSTATE_BASE():
 
             # STEP 11: Restart DUT
             self.step(11)
-            # In CI environment, the STOP coommand (step 8) already resets the variables. Only ask for
+            # In CI environment, the STOP command (step 8) already resets the variables. Only ask for
             # reboot outside CI environment.
             if not self.is_ci:
                 self.wait_for_user_input(prompt_msg="Restart DUT. Press Enter when ready.\n")
+                self.default_controller.ExpireSessions(self.dut_node_id)
 
             # STEP 12: TH waits for {PIXIT.WAITTIME.REBOOT}
             self.step(12)


tcarmelveilleux (approved):

rbultman (approved):

beckerzito (approved):

src/python_testing/TC_OpstateCommon.py
```diff
@@ -1112,10 +1112,11 @@ async def TEST_TC_OPSTATE_BASE_2_5(self, endpoint=1):
 
             # STEP 11: Restart DUT
             self.step(11)
-            # In CI environment, the STOP coommand (step 8) already resets the variables. Only ask for
+            # In CI environment, the STOP command (step 8) already resets the variables. Only ask for
             # reboot outside CI environment.
             if not self.is_ci:
                 self.wait_for_user_input(prompt_msg="Restart DUT. Press Enter when ready.\n")
```
bzbarsky-apple, line 1118: > Probably not caught in CI / QA because the CI app doesn't actually reboot.

We should fix that.  YAML certainly has support for rebooting the server app in CI, Python should too.

cecille: Possibly. Can also do it over named pipe commands. But this app does not have support for that and it's a bit orthogonal to this change.

cecille: Started an issue here - yaml has reboot, but I don't think it has support for named pipes, which makes a full reboot complicated because we need to bring the app back up and then find the new identifier. Will think this through. Meanwhile, this is a blocker for SVE, so are you ok to resolve this and follow up on the issue?

bzbarsky-apple: Yep, I was not suggesting we should block this change on that test infrastructure work.

PR #35714, opened 2024-09-22
[Android][ICD]Fix invalid signature for ICD onRefresh

yunhanw-google:
Fix the return type signature for ICD onKeyRefreshNeeded and input parameter signature for onKeyRefreshDone 

tests: locally validated using android chip-tool for ICD refresh bits

### DIFF
--- a/src/controller/java/src/chip/devicecontroller/ICDCheckInDelegateWrapper.java
+++ b/src/controller/java/src/chip/devicecontroller/ICDCheckInDelegateWrapper.java
@@ -67,7 +67,7 @@ class ICDCheckInDelegateWrapper {
   }
 
   @SuppressWarnings("unused")
-  private void onKeyRefreshDone(Long errorCode) {
+  private void onKeyRefreshDone(long errorCode) {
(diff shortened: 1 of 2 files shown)


src/controller/java/src/chip/devicecontroller/ICDCheckInDelegateWrapper.java
```diff
@@ -67,7 +67,7 @@ private byte[] onKeyRefreshNeeded(
   }
 
   @SuppressWarnings("unused")
-  private void onKeyRefreshDone(int errorCode) {
+  private void onKeyRefreshDone(Long errorCode) {
```
joonhaengHeo, line 70: @yunhanw-google 
If signature is "(J)V", it should be "long", not "Long".

yunhanw-google: good catch, thanks

joonhaengHeo (approved):

yufengwangca (approved):

chulspro (approved):

PR #35721, opened 2024-09-23
[Silabs] Fix for thermostat app hang issue on 917SoC

arun-silabs:
**Problem** : The use of LockChipStack and UnlockChipStack in timer event handler in Sensor manager is causing a deadlock and the application is getting stuck at that point.

**Fix** : Removed the LockChipStack and UnlockChipStack calls in SensorManager and replaced it with ScheduleWork api.

**Testing** : Tested locally by commissioning with chiptool and sent read command continuously for ~2hrs.


examples/thermostat/silabs/src/SensorManager.cpp
```diff
@@ -116,11 +116,19 @@ void SensorManager::SensorTimerEventHandler(void * arg)
     if ((temperature >= (lastTemperature + kMinTemperatureDelta)) || temperature <= (lastTemperature - kMinTemperatureDelta))
     {
         lastTemperature = temperature;
-        PlatformMgr().LockChipStack();
```
bzbarsky-apple, line 119: Is something calling SensorTimerEventHandler with the platform lock already held?  What's the stack to that?

examples/thermostat/silabs/src/SensorManager.cpp
```diff
@@ -116,11 +116,19 @@ void SensorManager::SensorTimerEventHandler(void * arg)
...
         // The SensorMagager shouldn't be aware of the Endpoint ID TODO Fix this.
         // TODO Per Spec we should also apply the Offset stored in the same cluster before saving the temp
 
-        app::Clusters::Thermostat::Attributes::LocalTemperature::Set(kThermostatEndpoint, temperature);
-        PlatformMgr().UnlockChipStack();
+        chip::DeviceLayer::PlatformMgr().ScheduleWork(UpdateClusterState, reinterpret_cast<intptr_t>(data));
     }
 }
+
+void SensorManager::UpdateClusterState(intptr_t context)
```
senthilku, line 129: The function name does not look correct.  UpdateTemperatureAttributes()?

examples/thermostat/silabs/src/SensorManager.cpp
arun-silabs: ```SensorTimerEventHandler``` will be called for every ```kSensorTImerPeriodMs```. (SensorManager.cpp Line#56).

In some cases, the chip stack is waiting for an event to be triggered by timer task, but the timer task is locked on the SensorTimerEventHandler api and is waiting for chip task to give back the lock. This is causing a deadlock.To avoid this deadlock, we are not directly calling LockChipStack and we are scheduling this work so that chip task would handle it.

examples/thermostat/silabs/src/SensorManager.cpp
arun-silabs: Modified the API name.

senthilku (approved):

examples/thermostat/silabs/src/SensorManager.cpp
jmartinez-silabs: Do you know what is the the event chip task is waiting on?

> chip stack is waiting for an event to be triggered by timer task

This is now a recurring issue with the wifi platform not seen on our thread platform. which makes me think that that particular event should be handled differently.

examples/thermostat/silabs/src/SensorManager.cpp
```diff
@@ -115,12 +115,20 @@ void SensorManager::SensorTimerEventHandler(void * arg)
...
+        lastTemperature            = temperature;
+        AttributeUpdateInfo * data = chip::Platform::New<AttributeUpdateInfo>();
+        data->endPoint             = kThermostatEndpoint;
+        data->temperature          = temperature;
         // The SensorMagager shouldn't be aware of the Endpoint ID TODO Fix this.
         // TODO Per Spec we should also apply the Offset stored in the same cluster before saving the temp
 
-        app::Clusters::Thermostat::Attributes::LocalTemperature::Set(kThermostatEndpoint, temperature);
-        PlatformMgr().UnlockChipStack();
+        chip::DeviceLayer::PlatformMgr().ScheduleWork(UpdateTemperatureAttribute, reinterpret_cast<intptr_t>(data));
```
jmartinez-silabs, line 125: that can work... But it looks like we are jumping through some hoops to get the attribute update done on the chip task. Wouldn't it be easier to let the appTask of this sensor app handle what is done in SensorTimerEventHandler?

Keep the timer call back if you want and set in the app task queue this function that "calculate" the temperature with the lock in place.
This way the lock occurs in the app task context and not the timer task context

jmartinez-silabs (approved):

examples/thermostat/silabs/src/SensorManager.cpp
arun-silabs: image

Thread 4 : CHIP Task
Thread 5 : Timer Task

@bzbarsky-apple @jmartinez-silabs 
The ```RunEventLoop()``` will acquire lock on the CHIP stack (Line#193). The timer task is waiting on the CHIP task to get the lock, and the CHIP task is waiting on timer task to set the Event flags. This is the call stack of the two tasks when deadlock occurs.

bzbarsky-apple: Ah, so we have two different threads that acquire locks in different orders?  It's _extremely_ unclear that SensorTimerEventHandler is called with some other lock already held.

It's also extremely unclear that this other lock is taken when trying to use KVS from the Matter thread....

The fix makes sense given that, but this seems like a situation that will keep having deadlocks unless either lock acquisition order is uniformly enforced or the relevant functions have clear documentation regarding what locks are held when they are called.

arun-silabs: @bzbarsky-apple 
Could you please take a moment to review the PR and close the conversation if the fix looks fine..
Please let me know if you have any further questions.

jmartinez-silabs: @arun-silabs I looked through your threads stack traces and looked up the actual calls we see.

Here is what I gathered

- The timer task is Suspended waiting on the LockChipStack.
- The Event loop (chip task) is blocked because sl_si91x_driver_send_command() is waiting on an event for the command response (sl_si91x_driver_wait_for_response_packet) for the rsi_flash_write().
-  For some reason, this event never occurs. No timeout and no command response.

Now who is responsible of posting this event/response? From what I see the command  from sl_si91x_driver_send_command, is placed on a queue which is handled by the task `si91x_event_handler_thread`.

Is that task running? Is it Thread # 6 or Thread # 8 or maybe another thread that we don't see in the screenshot?
I didn't find anything that posts this command response Event on the Timer task. Do you know where that would be?

There is something preventing the command response from being posted/handled.

I'm afraid This PR, while valid and correct, will hide the real implementation issue between this sl_si91x event mechanism and that we will hit it again at a later time.

Can you find the following answer:
- What post the command response event?
- Is this event supposed to be posted by the Timer task? Where is that queued in the timer task? If so that seems problematic and will prevent ever using a LockChipTask in the timer task.
- Is the si91x_event_handler_thread running. Did I handle the command queue.

arun-silabs: ```si91x_event_handler_thread``` is running (Thread 13). 
image

```si91x_host_wait_for_async_event``` function is waiting for events to be set with a timeout of ```4294967295```. This event is set by ```osEventFlagsSet``` function in cmsis_os2.c, and the event is posted to the timer task from ISR. This confirms that event is posted to timer task and the timer task is blocked by ```SensorTimerEventHandler```.
image

Also, the bus thread (Thread7) is also in suspended state. 
image

The ```sli_si91x_raise_pkt_pending_interrupt_to_ta``` function is waiting for ```ta_events``` flag to be set, which should again be done by the timer task. Also, the timeout is given as ```osWaitForever``` , hence it is getting suspended.

All the events set from ISR will be posted by Timer task(we got this confirmation from WiFi-SDK team).

andy31415 (approved):

examples/thermostat/silabs/src/SensorManager.cpp
jmartinez-silabs: Thank you for looking into it. That explains the unknowns we had. A suggestion that could be bade to the wifi SDK team would be to use the queue from the si91x_bus_thread rather than SetEvent mechanism as that won't use the Timer task

PR #35722, opened 2024-09-23
Fixed the service area server to check if there are any selected area…

hicklin:
Fixes

**TC-SEAR-1.5 CI Step Execution**

Addressed an issue where some test steps were not running in the CI pipeline. As a result, certain bugs in the server code were not being detected. Additional commands have been added to the CI process to ensure the device is properly configured, enabling all steps to run as expected.

**Service Area Server: SkipArea Command Handling**

Fixed a bug in the server code where the SkipArea command was not returning the expected error when invoked with an empty SelectedAreas list.


src/python_testing/TC_SEAR_1_5.py
```diff
@@ -117,24 +117,28 @@ async def test_TC_SEAR_1_5(self):
             test_step = "Manually intervene to put the device in a state that prevents it from executing the SkipArea command \
                   (e.g. set CurrentArea to null or make it not operate, i.e. be in the idle state)"
```
plauric, line 118:
```suggestion
                  (e.g. set CurrentArea to null or make it not operate, i.e. be in the idle state). Ensure that SelectedArea is not empty."
```

ReneJosefsen: It looks like the current text is pulled straight from the TP, so adjusting this should also mean we should ideally tweak the TP? and if that is added , should there be a read to assert this?

ReneJosefsen (approved):

plauric (approved):

PR #35725, opened 2024-09-23
Fix error handling in Darwin DNS-SD implementation to clean up properly.

bzbarsky-apple:
We had several codepaths where we could create a service ref but then do some fallible operations before calling Add() on the context.  If those failed, they would call Finalize() on the context, which would delete the context, but _not_ DNSServiceRefDeallocate() the service ref.

The fix is to make sure that:

1) All cases where we create a service ref successfully immediately set it on
   the context.
2) All context deletion goes through MdnsContexts::Delete, so that we make sure
   to clean up the service ref if there is one.


jtung-apple (approved):

bzbarsky-apple:

PR #35727, opened 2024-09-23
Switch to macos-14 for some of the Darwin runners.

bzbarsky-apple:
macos-14 runners are ARM, and should be much faster than the macos-13 runners.

The YAML tests are not passing on macos-14 for some reason (being sorted out in https://github.com/project-chip/connectedhomeip/pull/35704), but other workflows can be switched, which is what this does.


jtung-apple (approved):

bzbarsky-apple:

PR #35728, opened 2024-09-23
Move some APIs that are only used on MTRDeviceController_Concrete to that interface.

bzbarsky-apple:
Now that MTRDeviceControllerFactory only works with MTRDeviceController_Concrete, we can move some APIs from MTRDeviceController_Internal to MTRDeviceController_Concrete.


jtung-apple (approved):

bzbarsky-apple:

PR #35729, opened 2024-09-23
Chime Cluster Definition and server code

andyg-apple:
Chime cluster definition and server code

### DIFF
--- a/src/app/clusters/chime-server/chime-server.cpp
+++ b/src/app/clusters/chime-server/chime-server.cpp
@@ -1,0 +1,169 @@
+/*
+ *
+ *    Copyright (c) 2024 Project CHIP Authors
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+/****************************************************************************'
+ * @file
+ * @brief Implementation for the Chime Server Cluster
+ ***************************************************************************/
+
+
+#include "chime-server.h"
+
+#include <app/AttributeAccessInterfaceRegistry.h>
+#include <app/CommandHandlerInterfaceRegistry.h>
+#include <app/ConcreteAttributePath.h>
+#include <app/InteractionModelEngine.h>
+#include <app/util/attribute-storage.h>
+#include <app/util/util.h>
+#include <protocols/interaction_model/StatusCode.h>
+
+using namespace chip;
+using namespace chip::app;
+using namespace chip::app::DataModel;
+using namespace chip::app::Clusters;
+using namespace chip::app::Clusters::Chime;
+using namespace chip::app::Clusters::Chime::Attributes;
+using chip::Protocols::InteractionModel::Status;
+
+
+namespace chip {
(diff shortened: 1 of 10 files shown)


andyg-apple:
Will add generated code in second commit as soon as I get some feedback on this.

src/app/clusters/chime-server/chime-server.cpp
bzbarsky-apple, line 1: Ideally, the implementation would be a separate PR from adding the XML.

src/app/clusters/chime-server/chime-server.cpp
```diff
@@ -0,0 +1,169 @@
...
+{
+    VerifyOrDie(aPath.mClusterId == Chime::Id);
+    
+    switch (aPath.mAttributeId)
+    {
+    case InstalledChimeSounds::Id:
+        return aEncoder.Encode(mDelegate.GetInstalledChimeSounds());
+
+    case ActiveChimeSoundId::Id:
+        return aEncoder.Encode(mDelegate.GetActiveChimeSoundId());
```
bzbarsky-apple, line 79: It would be better if the simple things that can be stored in the cluster were stored there, so we can mark things dirty properly on changes.

src/app/clusters/chime-server/chime-server.cpp
```diff
@@ -0,0 +1,169 @@
...
+    bool activeSoundIdChanged = !(currentSoundId == soundId);
+
+    VerifyOrReturnError(activeSoundIdChanged, CHIP_NO_ERROR);
+    VerifyOrDie(mDelegate.SetActiveChimeSoundId(soundId) == CHIP_NO_ERROR);
+    MatterReportingAttributeChangeCallback(GetEndpointId(), Chime::Id, ActiveChimeSoundId::Id);
+
+    return CHIP_NO_ERROR;
+}
+
+CHIP_ERROR ChimeServer::SetEnabled(bool enabled)
```
bzbarsky-apple, line 127: OK, but the write path is not calling this?

src/app/clusters/chime-server/chime-server.h
```diff
@@ -0,0 +1,111 @@
...
+ */
+class ChimeDelegate
+{
+public:
+    ChimeDelegate() = default;
+
+    virtual ~ChimeDelegate() = default;
+
+    // Get Attribute Methods
+    virtual DataModel::List<const Chime::Structs::ChimeSoundStruct::Type>&  GetInstalledChimeSounds() = 0;
```
bzbarsky-apple, line 77: What's the lifetime of the buffer backing that list?  How does the delegate know to not deallocate it?

Also, it it intentional that this implementation does not allow chunking the list?

src/app/zap-templates/zcl/data-model/chip/chime-cluster.xml
```diff
@@ -0,0 +1,57 @@
...
+<configurator>
+  <domain name="CHIP"/>
+
+  <struct name="ChimeSoundStruct" apiMaturity="provisional">
+    <cluster code="0x0556"/>
+    <item name="ChimeId" type="int8u"/>
+    <item name="Name" type="char_string" max="48"/>
+  </struct>
+
+  <cluster apiMaturity="provisional">
```
bzbarsky-apple, line 26: Where is the spec for this?  I don't see it in the Matter appclusters spec on "master"....

It's impossible to review this XML without the spec.

andyg-apple: it's in the Cameras branch

src/app/clusters/chime-server/chime-server.cpp
andyg-apple: Thanks, that's an error

gmarcosb:
FYI chime cluster PR went in @ https://github.com/project-chip/connectedhomeip/pull/35806

Apologies did not realize you were working on this as well or would've reached out - started a chat on slack so we can coordinate

PR #35730, opened 2024-09-23
Change some more internal MTRDeviceControllerFactory APIs to use MTRDeviceController_Concrete.

bzbarsky-apple:
These always return concrete controllers.

getRunningControllers was already claiming the right signature in the implementation, just not the header.


jtung-apple (approved):

bzbarsky-apple:

PR #35731, opened 2024-09-23
Attribute metadata attribute-type.h encapsulation

lpbeliveau-silabs:
Removed attribute-type.h from attribute-metadata and moved definitions inside of .cpp file to reduce the compilation dependencies of attribute-metadata.h

### DIFF
--- a/src/app/util/attribute-metadata.cpp
+++ b/src/app/util/attribute-metadata.cpp
@@ -18,6 +18,16 @@
 
 #include <app-common/zap-generated/attribute-type.h>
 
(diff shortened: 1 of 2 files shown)


jmartinez-silabs (approved):

andy31415 (changes requested):
Need more of a justification here: it seems this removes the need of `attribute-type` from the header, which is fair but to what end? attribute-metadata is still not stand-alone because the cpp will need the header.

I am concerned if this paves the path to include more headers without adding dependencies.

andy31415 (approved):

bzbarsky-apple (approved):

PR #35737, opened 2024-09-23
[SVE][ICD] Bump MaxICDMonitoringEntrySize to reasonable size

### LINKED ISSUE #35735: [ICD] kICDMonitoringBufferSize is too small to handle full "full width" node ID
### Reproduction steps

When using "full width" node ID, for example, 0xffff000000000000, We cannot registerClient successfully using the lit example app, it would complain the buffer is too small, 
https://github.com/project-chip/connectedhomeip/blob/17f25a1772f7ba7535d57cd4dfa43ac1fc295843/src/app/icd/server/ICDMonitoringTable.h#L37C25-L37C49

[1727120175.806454][15188:15188] CHIP:SVR: debug 63
0x08, tag[Fully Qualified (6 Bytes)]: 0xd973::0xffa3::0xff, type: Octet String (0x10), length: 0, value: hex:
[1727120175.806476][15188:15188] CHIP:SPT: debug>>>152701000000000000FFFF270201000100FDFFFFFF300310D0E9EB3B00000000
[1727120175.806486][15188:15188] CHIP:SPT: debug>>>0000000000000000300410D0E9EB3B00000000000000000000000024050018

We may do TLV::EstimateStructOverhead(sizeof(uint64_t), sizeof(uint64_t), sizeof(xxxx), etc) * 3 / 2

### Bug prevalence

always

8750c55

### Platform

core

### Platform Version(s)

1.4

### Anything else?

_No response_

yunhanw-google:
Previous MaxICDMonitoringEntrySize 60 cannot handle the "full width" node ID, for example, 0xffff000000000000 so that registerClient with the above nodeId would fail with too small buffer size error where size is 63, which is larger than 60
in this PR, we use EstimateStructOverhead *3/2 to calculate size.

we need slack of at least 50%  to make sure we can grow and then if there is a firmware upgrade/downgrade, that it doesn't brick

fixes https://github.com/project-chip/connectedhomeip/issues/35735

### DIFF
--- a/src/app/icd/server/ICDMonitoringTable.h
+++ b/src/app/icd/server/ICDMonitoringTable.h
@@ -34,11 +34,21 @@ using SymmetricKeystore = SessionKeystore;
 
 namespace chip {
 
-inline constexpr size_t kICDMonitoringBufferSize = 60;
+static constexpr size_t MaxICDMonitoringEntrySize()
+{
+    // All the fields added together
+    return TLV::EstimateStructOverhead(
+        sizeof(NodeId) /*checkInNodeID*/, 
+        sizeof(uint64_t) /*monitoredSubject*/,
+        sizeof(Crypto::Symmetric128BitsKeyByteArray) /*aes_key_handle*/,
+        sizeof(Crypto::Symmetric128BitsKeyByteArray) /*hmac_key_handle*/, 
+        sizeof(uint8_t) /*client_type*/) * 3 / 2;
+}
+
+inline constexpr size_t kICDMonitoringBufferSize = MaxICDMonitoringEntrySize();
 
 struct ICDMonitoringEntry : public PersistentData<kICDMonitoringBufferSize>
 {
-
     ICDMonitoringEntry(FabricIndex fabric = kUndefinedFabricIndex, NodeId nodeId = kUndefinedNodeId)
     {
         this->fabricIndex      = fabric;
--- a/src/app/icd/server/ICDMonitoringTable.h
+++ b/src/app/icd/server/ICDMonitoringTable.h
@@ -41,10 +41,10 @@ static constexpr size_t MaxICDMonitoringEntrySize()
                                        sizeof(Crypto::Symmetric128BitsKeyByteArray) /*aes_key_handle*/,
                                        sizeof(Crypto::Symmetric128BitsKeyByteArray) /*hmac_key_handle*/,
                                        sizeof(uint8_t) /*client_type*/) *
-         // Provide 50% extra space to make a firmware upgrade that starts storing
-         // more data followed by a downgrade work easily and reliably.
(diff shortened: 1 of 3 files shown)


mkardous-silabs (approved):

tcarmelveilleux (approved):

src/app/icd/server/ICDMonitoringTable.h
```diff
@@ -34,11 +34,20 @@ using SymmetricKeystore = SessionKeystore;
...
 
-inline constexpr size_t kICDMonitoringBufferSize = 60;
+static constexpr size_t MaxICDMonitoringEntrySize()
+{
+    // All the fields added together
+    return TLV::EstimateStructOverhead(sizeof(NodeId) /*checkInNodeID*/, sizeof(uint64_t) /*monitoredSubject*/,
+                                       sizeof(Crypto::Symmetric128BitsKeyByteArray) /*aes_key_handle*/,
+                                       sizeof(Crypto::Symmetric128BitsKeyByteArray) /*hmac_key_handle*/,
+                                       sizeof(uint8_t) /*client_type*/) *
+        3 / 2;
```
bzbarsky-apple, line 44: This 3/2 needs at the very least documentation.  Because it's not clear why it's there and why it's 3/2.

bzbarsky-apple (changes requested):

src/app/icd/server/ICDMonitoringTable.h
yunhanw-google: update the comment here, thanks

src/app/icd/server/ICDMonitoringTable.h
```diff
@@ -34,11 +34,20 @@ using SymmetricKeystore = SessionKeystore;
...
 
-inline constexpr size_t kICDMonitoringBufferSize = 60;
+static constexpr size_t MaxICDMonitoringEntrySize()
+{
+    // All the fields added together
+    return TLV::EstimateStructOverhead(sizeof(NodeId) /*checkInNodeID*/, sizeof(uint64_t) /*monitoredSubject*/,
+                                       sizeof(Crypto::Symmetric128BitsKeyByteArray) /*aes_key_handle*/,
+                                       sizeof(Crypto::Symmetric128BitsKeyByteArray) /*hmac_key_handle*/,
+                                       sizeof(uint8_t) /*client_type*/) *
+        3 / 2; // Expect to have at least 50% to make sure we can grow and then if there is a firmware upgrade/downgrade, it doesn't brick
```
bzbarsky-apple, line 44:
```suggestion
         // Provide 50% extra space to make a firmware upgrade that starts storing
         // more data followed by a downgrade work easily and reliably.
         // The 50% number is chosen fairly randomly; storage increases larger than that are
         // possible but need to be staged carefully.
        3 / 2;
```

bzbarsky-apple (approved):

PR #35740, opened 2024-09-23
[SVE] TC-MCORE-FS-1.3: Bump the DUT timeout for newly added device

### LINKED ISSUE #35739: [SVE] TC-MCORE-FS-1.3: Timeout too low for device discovery & commissioning from DUT
### Reproduction steps

In `TC-MCORE-FS-1.3`, a `TH_SERVER_NO_UID` (an example app with no unique ID) is [commissioned onto TH's fabric](https://github.com/project-chip/connectedhomeip/blob/6b99a99a762e58f0b07b6b3e4d0d4f1fbad39c22/src/python_testing/TC_MCORE_FS_1_3.py#L163), and `DUT_FSA`'s aggregator parts list is checked before the [commissioning window on TH_SERVER_NO_UID is opened using TH's fabric](https://github.com/project-chip/connectedhomeip/blob/6b99a99a762e58f0b07b6b3e4d0d4f1fbad39c22/src/python_testing/TC_MCORE_FS_1_3.py#L143) (this is in response to `CommissionerControlCluster::CommissionNode` command as part of reverse commissioning between TH and DUT). Right after that, the DUT is expected to discover and commission `TH_SERVER_NO_UID` and [just 2s after that](https://github.com/project-chip/connectedhomeip/blob/6b99a99a762e58f0b07b6b3e4d0d4f1fbad39c22/src/python_testing/TC_MCORE_FS_1_3.py#L192), the DUT's parts list is reread to determine whether the new device has been made available as a bridged node.

This timeout is pretty short because right after issuing the `CommissionerControlCluster::ReverseOpenCommissioningWindow` command, the DUT is expected to discover the newly advertised `_matterc._udp` service, resolve its address, establish PASE, install the `DUT_FSA`'s fabric, subscribe (& perform other necessary interactions during onboarding) before making it available as a newly accessible bridged node on the `DUT_FSA`'s aggregator parts list, and some of these operations can take a bit of time.

### Bug prevalence

Almost always on our DUT

56c0bd8

### Platform

other

### Platform Version(s)

_No response_

### Type

Test Improvement

### Testing

Hardware validated, Platform validated

_No response_

### Anything else?

_No response_

samadDotDev:
This timeout is pretty short because right after issuing the `CommissionerControlCluster::ReverseOpenCommissioningWindow` command, the DUT is expected to discover the newly advertised `_matterc._udp` service, resolve its address, establish PASE, install the `DUT_FSA`'s fabric, subscribe (& perform other necessary interactions during onboarding) before making it available as a newly accessible bridged node on the `DUT_FSA`'s aggregator parts list, and some of these operations can take a bit of time.

Bumping up to 30s, consistent with a few other FS test cases.

Fixes https://github.com/project-chip/connectedhomeip/issues/35739


saurabhst (approved):

src/python_testing/TC_MCORE_FS_1_3.py
```diff
@@ -189,7 +189,7 @@ async def test_TC_MCORE_FS_1_3(self):
             device_node_id=th_server_th_node_id)
 
         # Wait for the device to appear on the DUT_FSA_BRIDGE.
-        await asyncio.sleep(2)
+        await asyncio.sleep(30)
```
arkq, line 192: Please increase the timeout but only for the non-CI variant (self.is_ci).

andy31415:
@samadDotDev how did you create this PR? It runs double the CI for some reason, and I believe it may be that it uses a main repo branch instead of a fork. 

Could you confirm? Could you please switch to using forks for development?

src/python_testing/TC_MCORE_FS_1_3.py
andy31415, line 192:
```suggestion
        await asyncio.sleep(2 if self.is_pics_sdk_ci_only else 30)
```

andy31415: I commited a patch for this.

andy31415 (approved):

tehampson (approved):

PR #35741, opened 2024-09-24
Removed using namespace std from all files

### LINKED ISSUE #35713: [Build] Unnecessary `using namespace std` creates ambiguity error to C++23 standard compiler.
### Build issue(s)

When compiling both sources `connectedhomeip\src\app\clusters\refrigerator-alarm-server\refrigerator-alarm-server.cpp` and `connectedhomeip\src\app\clusters\mode-select-server\mode-select-server.cpp`, an error of ambiguity pops out by the compiler regarding the use of `to_underlying()`.

Solution: Remove the line `using namespace std;` [here](https://github.com/project-chip/connectedhomeip/blob/master/src/app/clusters/refrigerator-alarm-server/refrigerator-alarm-server.cpp#L35) and [here](https://github.com/project-chip/connectedhomeip/blob/master/src/app/clusters/mode-select-server/mode-select-server.cpp#L40).

### Platform

_No response_

### Anything else?

_No response_

jadhavrohit924:
### Problem
- Fixes #35713.

### Changes
- Removed using namespace std from all sources.

### Test
- Build all-cluster-app and tv-app for Linux.

### DIFF
--- a/src/lib/support/SafeInt.h
+++ b/src/lib/support/SafeInt.h
@@ -37,10 +37,11 @@ namespace chip {
 template <typename T, typename U, std::enable_if_t<std::is_integral<T>::value, int> = 0>
 bool CanCastTo(U arg)
 {
+    using namespace std;
     // U might be a reference to an integer type, if we're assigning from
     // something passed by reference.
-    typedef typename std::remove_reference<U>::type V; // V for "value"
-    static_assert(std::is_integral<V>::value, "Must be assigning from an integral type");
+    typedef typename remove_reference<U>::type V; // V for "value"
+    static_assert(is_integral<V>::value, "Must be assigning from an integral type");
 
     // We want to check that "arg" can fit inside T but without doing any tests
     // that are always true or always false due to the types involved, which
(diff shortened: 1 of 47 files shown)


Damian-Nordic (approved):

andy31415 (changes requested):
The description in https://github.com/project-chip/connectedhomeip/blob/master/src/lib/support/TypeTraits.h#L32 seems to say we intentionally overloaded this to have C++23 shims.

Don't we have other alternatives? i.e. for C++23 just use the existingimplementation from the standard instead of implementing our own.

This PR seems to propose to have the workaround forever instead of treating it as a temporary shim.

src/lib/support/SafeInt.h
```diff
@@ -37,11 +37,10 @@ namespace chip {
 template <typename T, typename U, std::enable_if_t<std::is_integral<T>::value, int> = 0>
 bool CanCastTo(U arg)
 {
-    using namespace std;
```
bzbarsky-apple, line 40: Probably OK, but this use was safe: it's scoped to this function, and this function does not use `to_underlying`.

src/lib/support/SafeInt.h
```diff
@@ -127,10 +126,9 @@ bool CanCastTo(U arg)
 template <typename T>
 typename std::enable_if<std::is_unsigned<T>::value, typename std::make_signed<T>::type>::type CastToSigned(T arg)
 {
-    using namespace std;
```
bzbarsky-apple, line 130: Again, this use was fine.

src/lib/support/TypeTraits.h
```diff
@@ -26,8 +26,21 @@
...
+#if __has_include(<utility>) // For C++23 and later, include <utility> if available
+#include <utility>           // Contains std::to_underlying
+#endif
+
 namespace chip {
 
+#if __cplusplus >= 202300L
+
+template <class T>
+constexpr auto to_underlying(T e)
```
bzbarsky-apple, line 38: Does `using to_underlying = std::to_underlying;` or similar not work?

bzbarsky-apple (approved):

src/lib/support/SafeInt.h
jadhavrohit924: Restored it.

src/lib/support/SafeInt.h
jadhavrohit924: Restored

src/lib/support/TypeTraits.h
jadhavrohit924: Right. Done

jadhavrohit924:
@andy31415 I have addressed the changes please take a look.

andy31415 (approved):

PR #35743, opened 2024-09-24
Updated the SEAR and RVCRUNM PICS XMLs to the later versions.

hicklin:
Updated the SEAR and RVCRUNM PICS XMLs to the later versions.


plauric (approved):

andy31415 (approved):
Approving, however also wondering about why we have a seemingly one-off with `Test Plan.xml`. 

This also seems to somewhat violate having a single source of truth for things if it needs periodic updating.

Update probably ok, but wondering what the overall global path here is: if this is a good thing to have, more examples should probably have it. If not (e.g. if we consider test plans being private a desirable thing) we should remove it from here as well.

PR #35745, opened 2024-09-24
[Tizen] Run tests from 9P-shared directory instead of ISO

XToripuru:
### Issue

Currently the way tests and runner are passed onto QEMU is by mounting an iso image containing these. The problem with this approach is that building the image is slow, and does not allow to share output files back with host which will be needed in a follow-up PR (#35786) adding code coverage for tests.

### Solution

Scalable and efficient method is to share tests directory between host and QEMU using virtfs with 9P. This approach involves copying adequate runner to `${root_out_dir}/tests` and then sharing it as `/mnt/chip` on QEMU side. This is highly configurable as the new option`tizen_qemu.py --share` takes a custom path. A service is added which performs the mounting.

### Testing

Tested locally using both interactive and non-interactive modes with and without customized shared directory.

### Extra notes
This PR contains only Dockerfile changes for the mentioned above issue and solution.

### DIFF
--- a/integrations/docker/images/stage-3/chip-build-tizen-qemu/Dockerfile
+++ b/integrations/docker/images/stage-3/chip-build-tizen-qemu/Dockerfile
@@ -151,7 +151,7 @@ RUN set -x \
     | guestfish --rw -a $TIZEN_IOT_IMAGE_ROOT -m /dev/sda upload - $SYSTEMD_UNIT_OPT_POST_MOUNT : \
     ln-sf $SYSTEMD_UNIT_OPT_POST_MOUNT $SYSTEMD_REQUIRES_LOCAL_FS \
     # Mount using 9P at /mnt/chip
-    && SYSTEMD_9P_MOUNT=$SYSTEMD_SYSTEM/mnt-chip.mount \
+    && SYSTEMD_UNIT_9P_MOUNT=$SYSTEMD_SYSTEM/mnt-chip.mount \
     && echo -n \
     "[Unit]\n" \
     "After=network.target\n" \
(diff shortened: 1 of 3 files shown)


integrations/docker/images/stage-3/chip-build-tizen-qemu/Dockerfile
```diff
@@ -147,17 +150,18 @@ RUN set -x \
...
-    && SYSTEMD_UNIT_CHIP_MOUNT=$SYSTEMD_SYSTEM/mnt-chip.mount \
+    # Mount 9P at /mnt/chip
+    && SYSTEMD_9P_MOUNT=$SYSTEMD_SYSTEM/mnt-chip.mount \
     && echo -n \
     "[Unit]\n" \
-    "ConditionPathIsMountPoint=!/mnt/chip\n" \
+    "After=network.target\n" \
     "[Mount]\n" \
-    "What=/dev/disk/by-label/CHIP\n" \
+    "What=host0\n" \
```
andy31415, line 159: we have to change version for dockerfiles for any dockerfile changes.

Are the tizen qemu setup something that can be pulled into the vscode image so that we can test in the "one large image for everyting" environment?

XToripuru: This PR is now for the Dockerfile part only.

XToripuru: Updated vscode container's env vars.

arkq: > Are the tizen qemu setup something that can be pulled into the vscode image so that we can test in the "one large image for everyting" environment?

Yes. If the vscode image will be "based" on the chip-build-tizen-qemu docker image instead of chip-build-tizen, it will be possible to build and run tests in vscode.

arkq (approved):

andy31415 (approved):

PR #35746, opened 2024-09-24
remove a call of `emberAfContainsAttribute` when using the data model provider interface

andy31415:
This continues decoupling the interaction model engine from direct ember calls.


bzbarsky-apple (approved):

src/app/codegen-data-model-provider/CodegenDataModelProvider.h
```diff
@@ -63,11 +63,26 @@ class CodegenDataModelProvider : public chip::app::DataModel::Provider
 
         /// Checks if the given command id exists in the given list
         bool Exists(const CommandId * list, CommandId toCheck);
+
+        void Reset() { mCurrentList = mCurrentHint = nullptr; }
     };
 
 public:
+    /// clears out internal caching. Expecially useful in unit tests,
```
andy31415, line 71:
```suggestion
    /// clears out internal caching. Especially useful in unit tests,
```

src/app/InteractionModelEngine.cpp
```diff
@@ -1592,9 +1592,31 @@ void InteractionModelEngine::RemoveDuplicateConcreteAttributePath(SingleLinkedLi
     while (path1 != nullptr)
     {
         bool duplicate = false;
+#if CHIP_CONFIG_USE_DATA_MODEL_INTERFACE && CHIP_CONFIG_USE_EMBER_DATA_MODEL
```
tcarmelveilleux, line 1595: Could these blocks of code with ifdefs be moved to anonymous namespace methods so that the ifdefing is not inlined?

tcarmelveilleux (approved):

src/app/InteractionModelEngine.cpp
andy31415: Updated to a method that is more self-contained.

PR #35749, opened 2024-09-24
Remove a few more calls of IME to ember-compatibility.

andy31415:
This works to move more logic into DataModel::Provider calls to decouple interactionmodel engine from ember calls.

Decoupled the following calls:
  - implemented a CommandExists logic using pure DataModel::Provider interfaces
  - Replace ember-specific `ConcreteAttributePathExists` with provider aware `IsExistingAttributePath`

### Changes

- Adjusted more unit tests to use Datamodel::provider settings
- Updated unit tests to use `mock ember data` to report a consistent data layout
   - I was NOT able to fully fix up TestReadInteraction::TestSubscribeWildcard as that one seems to basically be manually looking at report data (which is long) and that is impractical. I just reset that test to the default (which is bad, because that seems to fix the default in place ... maybe we should copy over that layout instead to make it test-specific)
- **DISABLED** chip-im-initiator/responder and the cirque InteractionModelTest.py
  - chip-im-responder hardcodes directly ember shims used by the interaction model engine without regard of consistency (e.g. all data version checks return equal). Maintaining such a thing does not seem practical and we cover a lot of IM test via regular integration (repl and yaml) tests.
  - I specifically did not make the cirque tests use "ember" data model (that would have just made it pass) because I want to be able to ensure that ember and data model provider behaves the same.
  
> [!IMPORTANT]
> chip-im-responder and chip-im-initiator were disabled. We should consider if we can fix chip-im-responder to have 
> a consistent data model or we should delete the test.


bzbarsky-apple (commented):
I did not look at the tests carefully.

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
```diff
@@ -31,6 +34,69 @@ namespace chip {
 namespace app {
 namespace {
 
+/// Handles going through callback-based enumeration of generated/accepted commands
```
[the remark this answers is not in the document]
andy31415: Absolutely - created that one to isolate the changes. Will keep this comment open until that PR merges and I can merge with master.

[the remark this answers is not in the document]
andy31415: Ready for review. Had to fix up several tests. TestReadInteraction::TestSubscribeWildcard is not maintainable by the way.... I gave up and `reset to default config` on it.

src/app/InteractionModelEngine.cpp
```diff
@@ -517,20 +517,7 @@ CHIP_ERROR InteractionModelEngine::ParseAttributePaths(const Access::SubjectDesc
...
-
-            if (
-#if CHIP_CONFIG_USE_DATA_MODEL_INTERFACE
-                GetDataModelProvider()->GetAttributeInfo(concretePath).has_value()
-
-#else
-                ConcreteAttributePathExists(concretePath)
-#endif
-            )
+            if (IsExistingAttributePath(concretePath))
```
bzbarsky-apple, line 520: Is this supposed to be checking for this being an "Existent Path" in the spec terminology?  If so, the naming should match the spec, no?  IsExistentAttributePath.

It looks like this function was added in https://github.com/project-chip/connectedhomeip/pull/35746 after I had reviewed it....  I was being confused why I wasn't recognizing it.

src/app/tests/TestCommandInteraction.cpp
bzbarsky-apple, line 1: I did not review the test bits carefully.

bzbarsky-apple (approved):

src/app/InteractionModelEngine.cpp
andy31415: I will rename. It was added after reviewer request (Tennessee) of not having ifdefs in the middle of an if condition because that is ugly to read ... I agreed.

yunhanw-google (approved):

PR #35753, opened 2024-09-24
Add support for using CommandHandlerInterface for accepted/generated command listing

andy31415:
This makes sure that `DataModel::Provider` respects the command handler interface `GeneratedCommands` and `AcceptedCommands` if we need to iterate over them.

This has some deficits:
  - list of commands is NOT sufficient for metadata (we are missing flags like  fabric scoped and timed). These are available if we return an ember-subset of values, however it is likely not generically-available
  - the iteration is `O(N^2)` becuase it converts a callback-based system into an iterator one. #35790 created for this. 

For now this seems sufficient as actual iteration should be infrequent and existing CHI is still coupled with ember metadata, however at some point we should consider speed up logic (e.g. some form of caching or converting CHI to use iterators rather than callbacks). We should also make CHI (or whatever replacement) actually support the full flag range instead of just a list iterator.

### Changes

- make use of CHI callbacks for generated and accepted commands
- add unit test


src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
```diff
@@ -31,6 +34,69 @@ namespace chip {
...
+    enum class Operation
+    {
+        FindFirst, // Find the first value in the list
+        FindExact, // Find the given value
+        FindNext   // Find the value AFTER this value
+    };
+
+    EnumeratorCommandFinder(Operation operation, CommandId target) : mOperation(operation), mTarget(target) {}
+
+    /// Callback to pass in to `Enumerate*` calls in COmmandHandlerInterface
```
tehampson, line 57: super nit,
```suggestion
    /// Callback to pass in to `Enumerate*` calls in CommandHandlerInterface
```

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
```diff
@@ -516,6 +628,44 @@ DataModel::CommandEntry CodegenDataModelProvider::NextAcceptedCommand(const Conc
...
+                    ctx->commandExists = true;
+                    return Loop::Break;
+                }
+                return Loop::Continue;
+            },
+            &context);
+
+        if (err != CHIP_ERROR_NOT_IMPLEMENTED)
+        {
+            // TODO: we know the command exists, but don't actually have info such as flags. Ideally CHI should
```
tehampson, line 659: nit what is CHI? Do you mean CHIP? Even then I don't follow this comment or TODO
```suggestion
            // TODO: we know the command exists, but don't actually have info such as flags. Ideally CHIP should
```

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
```diff
@@ -504,6 +593,29 @@ DataModel::CommandEntry CodegenDataModelProvider::FirstAcceptedCommand(const Con
...
+
+            if ((err == CHIP_NO_ERROR) && beforeId.has_value())
+            {
+                DataModel::CommandEntry result;
+                result.path = ConcreteCommandPath(before.mEndpointId, before.mClusterId, *beforeId);
+                return result;
+            }
+
+            return DataModel::CommandEntry::kInvalid;
+        }
```
tehampson, line 616: Would it make sense to just do something like this?

```suggestion
        if (err != CHIP_ERROR_NOT_IMPLEMENTED && err != CHIP_NO_ERROR)
        {
                    return DataModel::CommandEntry::kInvalid;
        }
        auto beforeId = finder.GetFound();
        if ((err == CHIP_NO_ERROR) && beforeId.has_value())
        {
            DataModel::CommandEntry result;
            result.path = ConcreteCommandPath(before.mEndpointId, before.mClusterId, *beforeId);
            return result;
        }
```

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
```diff
@@ -31,6 +34,69 @@ namespace chip {
...
+///   - FindFirst will return the first found element
+///   - FindExact finds the element with the given id
+///   - FindNext finds the element following the given id
+class EnumeratorCommandFinder
+{
+public:
+    enum class Operation
+    {
+        FindFirst, // Find the first value in the list
+        FindExact, // Find the given value
```
tehampson, line 51: Is `FindExact` planned to be used in the future or is it no longer needed?

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
```diff
@@ -31,6 +34,69 @@ namespace chip {
 namespace app {
 namespace {
 
+/// Handles going through callback-based enumeration of generated/accepted commands
+/// for CommandHandler interface based items.
```
bzbarsky-apple, line 38:
```suggestion
/// for CommandHandlerInterface based items.
```

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
bzbarsky-apple, line 57:
```suggestion
    /// Callback to pass in to `Enumerate*` calls in CommandHandlerInterface
```

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
```diff
@@ -31,6 +34,69 @@ namespace chip {
...
+
+    /// Get the element found (if any) after the `Callback` is used
+    std::optional<CommandId> GetFound() const { return mFound; }
+
+private:
+    Operation mOperation;
+    CommandId mTarget;
+    std::optional<CommandId> mFound = std::nullopt;
+
+    Loop HandlerCallback(CommandId id)
```
bzbarsky-apple, line 68: So... the existing CommandHandlerInterface API was designed to allow efficiently implementing the one thing it was used for.  Now we are using it for something quite different and the API is a bad match for this, leading to a bunch of extra code and inefficiency.

Followup: we should redesign that API to make it fit the use-case better.

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
```diff
@@ -492,6 +558,29 @@ std::optional<DataModel::AttributeInfo> CodegenDataModelProvider::GetAttributeIn
...
+        CHIP_ERROR err = interface->EnumerateAcceptedCommands(path, finder.Callback(), &finder);
+
+        if (err != CHIP_ERROR_NOT_IMPLEMENTED)
+        {
+            auto firstId = finder.GetFound();
+
+            if ((err == CHIP_NO_ERROR) && firstId.has_value())
+            {
+                DataModel::CommandEntry result;
+                result.path = ConcreteCommandPath(path.mEndpointId, path.mClusterId, *firstId);
```
bzbarsky-apple, line 576: So this is just silently not setting up state that the caller expects?

In practice the way this stuff works right now is that `EnumerateAcceptedCommands` is expected to return a _subset_ of what's enabled in the ZAP file, so that we have the right timed/etc metadata in the places that use it.  If someone hase `EnumerateAcceptedCommands` return commands that are not in the ZAP config, I guess the existing code runs and does the wrong things.   But this new code will run and do the wrong things in the case when the API is actually implemented correctly, no?

It seems to me like we should be calling `CommandEntryFrom` here, not manually constructing a CommandEntry.

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
```diff
@@ -504,6 +593,29 @@ DataModel::CommandEntry CodegenDataModelProvider::FirstAcceptedCommand(const Con
...
+        CommandHandlerInterfaceRegistry::Instance().GetCommandHandler(before.mEndpointId, before.mClusterId);
+
+    if (interface != nullptr)
+    {
+        EnumeratorCommandFinder finder(EnumeratorCommandFinder::Operation::FindNext, before.mCommandId);
+        CHIP_ERROR err = interface->EnumerateAcceptedCommands(before, finder.Callback(), &finder);
+
+        if (err != CHIP_ERROR_NOT_IMPLEMENTED)
+        {
+            auto beforeId = finder.GetFound();
```
bzbarsky-apple, line 606: `beforeId` is an odd name here.  Isn't this actually the `nextId`?

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
```diff
@@ -504,6 +593,29 @@ DataModel::CommandEntry CodegenDataModelProvider::FirstAcceptedCommand(const Con
...
+
+            if ((err == CHIP_NO_ERROR) && beforeId.has_value())
+            {
+                DataModel::CommandEntry result;
+                result.path = ConcreteCommandPath(before.mEndpointId, before.mClusterId, *beforeId);
```
bzbarsky-apple, line 611: Again, this should be using CommandEntryFrom.

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
bzbarsky-apple, line 659: See above.  If there is no ZAP metadata here the API implementation is doing it wrong.  We should be using CommandEntryFrom.

Also, this is what FindExact was added for, I guess, but we're not using it?  Why is it there, then?

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
```diff
@@ -537,6 +708,27 @@ ConcreteCommandPath CodegenDataModelProvider::FirstGeneratedCommand(const Concre
...
+        CommandHandlerInterfaceRegistry::Instance().GetCommandHandler(before.mEndpointId, before.mClusterId);
+
+    if (interface != nullptr)
+    {
+        EnumeratorCommandFinder finder(EnumeratorCommandFinder::Operation::FindNext, before.mCommandId);
+        CHIP_ERROR err = interface->EnumerateGeneratedCommands(before, finder.Callback(), &finder);
+
+        if (err != CHIP_ERROR_NOT_IMPLEMENTED)
+        {
+            auto beforeId = finder.GetFound();
```
bzbarsky-apple, line 721: Again, odd naming.

bzbarsky-apple (changes requested):

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
andy31415: I meant `CommandHandlerInterface` ... agree it is confusing. Will try to spell it out.

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
andy31415: FindExact is used to query by path - it is a query of "does this exist". I failed to update the `GetAcceptedCommandInfo` with the new class though ... will update it.

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
andy31415: I agree - this is inefficient because we want an iterator pattern when a callback pattern exists instead.

Did not try to re-make it here though as that probably requires some extra iterations. At least this is infrequently used and the inefficient path should generally not be used at all (but has to be here for completeness).

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
andy31415: Ok, I can make it that "if zap also contains this, return its info". I was not aware that we enforce always returning a subset (although that would absolutely make it easier to have correct flags)

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
andy31415: Updated GetAcceptedCommandInfo to use this.

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
andy31415: Updated to use the right constructor to forward the info.

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
andy31415: I started writing this code and then figured out a common class is better ... implemented it and did not use it. Updated now.

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
andy31415: Created https://github.com/project-chip/connectedhomeip/issues/35790 and added comment about this.

I think currently command iteration is mostly not used, it was added because we may use it for automatic Accepted/Generated commands implementation (and at that point switching the API to be callback based is maybe viable).

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
andy31415: I updated the code flow to make it shorter at call sites. Still a lot of arguments, but hoping this is easier to read.

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
bzbarsky-apple: > I was not aware that we enforce always returning a subset

We don't, but in practice all existing use cases of this do that, and if you don't you will be violating the spec in various ways...

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
bzbarsky-apple: > At least this is infrequently used

It's used all the time for wildcard subscriptions, no?

> it was added because we may use it for automatic Accepted/Generated commands implementation

Not "may", _do_.  This is how those attributes are implemented, period.

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
```diff
@@ -92,11 +108,41 @@ class EnumeratorCommandFinder
...
+
+    CHIP_ERROR err = (interface->*mCallback)(path, HandlerCallbackFn, this);
+    if (err == CHIP_ERROR_NOT_IMPLEMENTED)
+    {
+        return std::nullopt; // no data provided by the interface
+    }
+
+    if (err != CHIP_NO_ERROR)
+    {
+        // Report the error here since we lose actual error. This generally should NOT be possible as CHI usually returns
```
bzbarsky-apple, line 137:
```suggestion
        // Report the error here since we lose actual error. This generally should NOT be possible as CommandHandlerInterface usually returns
```

bzbarsky-apple (approved):

tcarmelveilleux (approved):

PR #35755, opened 2024-09-24
Update gn_silabs_example to use PW_ENVIRONMENT_ROOT's gn install if…

jmartinez-silabs:
declared else default to the expected matter bootstrapped environment location

- No change to default script usage.
- CI validates USE_SLC use case
- locally tested with PW_ENVIRONMENT_ROOT to a different pw environment

### DIFF
--- a/scripts/examples/gn_silabs_example.sh
+++ b/scripts/examples/gn_silabs_example.sh
@@ -28,14 +28,23 @@ else
     CHIP_ROOT="$MATTER_ROOT"
 fi
 
+ACTIVATE_MATTER_ENV=true
+if [[ -z "${PW_ENVIRONMENT_ROOT}" ]]; then
+    echo "Using the bootstrapped pigweed ENV in Matter root"
+    PW_PATH="$CHIP_ROOT/.environment/cipd/packages/pigweed"
+else
+    echo "Using provided $PW_ENVIRONMENT_ROOT as Pigweed ENV root"
+    PW_PATH="$PW_ENVIRONMENT_ROOT"
+    ACTIVATE_MATTER_ENV=false
+fi
+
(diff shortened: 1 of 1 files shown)


scripts/examples/gn_silabs_example.sh
```diff
@@ -28,14 +28,23 @@ else
...
+
 set -x
 env
 USE_WIFI=false
 USE_DOCKER=false
 USE_GIT_SHA_FOR_VERSION=true
 USE_SLC=false
-GN_PATH=gn
-GN_PATH_PROVIDED=false
+GN_PATH="$PW_PATH/gn"
```
andy31415, line 47: should we check if this exists?

in a vscode instance for example, PW_ENVIRONMENT_ROOT will be set however without bootstrap nothing will be populated there. So it seems here we want bootstrap to be run somewhere, but this script to not be run under a bootstrapped environment after that. Is this true?

jmartinez-silabs: In this case, there might be something else missing. I couldn't bootstrap a provided `PW_ENVIRONMENT_ROOT` from the matter root where this script is run. This PW_ENVIRONMENT_ROOT had to be pre-bootstrap.

I was under the impression the passed PW_ENVIRONMENT_ROOT was ready to use.

andy31415: PW_ENVIRONMENT_ROOT is the environment that pigweed should use at the time of running `bootstrap`. It is not guaranteed to be populated and it always starts as empty (probably not even an existing directory).

andy31415 (approved):

mkardous-silabs (approved):

PR #35756, opened 2024-09-24
MTRDevice should remember new DataVersion values even if the attribute value did not change.

bzbarsky-apple:
We used to ignore "same value but new DataVersion" because some devices would spam reports that looked like that and we were constantly hitting storage for the DataVersion updates.  But now that we have backoffs on the storage of our cluster data, we can go ahead and just note the new DataVersion and rely on that backoff to prevent hitting storage too much.

The test update verifies that:

1) Reports with same value but new DataVersion do get persisted but are subject
   to the persistence backoff settings.
2) Reports with the same value and same DataVersion do not lead to any storage
   traffic.

Review note: The test changes are best reviewed in a whitespace-insensitive diff.


woody-apple (approved):

bzbarsky-apple:

PR #35757, opened 2024-09-24
tagging rootNode as a mandatory device type in zcl.json for ZAP 

paulr34:
Adding Root Node Device Type's DeviceId to zcl.json indicating it is a "mandatoryDeviceType". 

The XML does not indicate root node being a requirement for any Matter application. That information needs to be recorded somewhere for ZAP to enable it by default.


bzbarsky-apple (approved):
Approved for now, but we should think about getting this into the XML.

jmartinez-silabs (approved):

PR #35761, opened 2024-09-24
TC-OPSTATE-2.5,6: Fix messages, re-establish subscription after reset

cecille:
Changes the prompts on OPSTATE-2.5 and 2.6 so the operator can see what to do. 

Also adds a subscription re-establishment step after the dut is reset.

### DIFF
--- a/src/python_testing/TC_OpstateCommon.py
+++ b/src/python_testing/TC_OpstateCommon.py
@@ -1123,7 +1123,7 @@ class TC_OPSTATE_BASE():
                 # Expire the session and re-establish the subscription
                 self.default_controller.ExpireSessions(self.dut_node_id)
                 if self.check_pics(f"{self.test_info.pics_code}.S.E01"):
-                    # Subscribe to Events and when they are sent push them to a queue for checking later
+                    # Subscribe to Events and when they are received push them to a queue for checking later
                     events_callback = EventSpecificChangeCallback(events.OperationCompletion)
                     await events_callback.start(self.default_controller,
                                                 self.dut_node_id,
--- a/src/python_testing/TC_OpstateCommon.py
+++ b/src/python_testing/TC_OpstateCommon.py
@@ -111,14 +111,16 @@ class TC_OPSTATE_BASE():
                     asserts.fail("The --app-pid flag must be set when PICS_SDK_CI_ONLY is set")
             self.app_pipe = self.app_pipe + str(app_pid)
 
-    def send_raw_manual_or_pipe_command(self, command: dict):
+    def send_raw_manual_or_pipe_command(self, command: dict, msg: str):
         if self.is_ci:
             self.write_to_app_pipe(command)
             time.sleep(0.1)
         else:
-            self.wait_for_user_input(prompt_msg="Press Enter when ready.\n")
+            prompt = msg if msg is not None else "Press Enter when ready."
+            prompt += '\n'
+            self.wait_for_user_input(prompt_msg=prompt)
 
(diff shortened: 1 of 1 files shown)


tcarmelveilleux (approved):

saurabhst (approved):

src/python_testing/TC_OpstateCommon.py
```diff
@@ -1116,7 +1120,14 @@ async def TEST_TC_OPSTATE_BASE_2_5(self, endpoint=1):
             # reboot outside CI environment.
             if not self.is_ci:
                 self.wait_for_user_input(prompt_msg="Restart DUT. Press Enter when ready.\n")
+                # Expire the session and re-establish the subscription
                 self.default_controller.ExpireSessions(self.dut_node_id)
+                if self.check_pics(f"{self.test_info.pics_code}.S.E01"):
+                    # Subscribe to Events and when they are sent push them to a queue for checking later
```
plauric, line 1126:
```suggestion
                    # Subscribe to Events and when they are received sent push them to a queue for checking later
```

Is this what you meant to type?

plauric, line 1126:
```suggestion
                    # Subscribe to Events and when they are received push them to a queue for checking later
```

src/python_testing/TC_OpstateCommon.py
```diff
@@ -1259,11 +1270,12 @@ async def TEST_TC_OPSTATE_BASE_2_6(self, endpoint=1):
...
         await sub_handler.start(self.default_controller, self.dut_node_id, endpoint)
 
+        self.step(3)
         if self.pics_guard(self.check_pics(f"{self.test_info.pics_code}.S.M.ST_RUNNING")):
-            self.step(3)
             self.send_manual_or_pipe_command(name="OperationalStateChange",
                                              device=self.device,
-                                             operation="Start")
+                                             operation="Start",
+                                             msg="Put the DUT into the running state")
```
plauric, line 1278:
```suggestion
                                             msg="Put DUT in running state")
```

To be consistent with the similar messages below.

plauric, line 1278: Question about the statement that follows below (i.e. time.sleep(1)): should we add a PIXIT code to control these waits between issuing a command to the DUT and reading some attribute to check the effect? For this TC and for many others. The "real DUTs", that the vendors will attempt to certify, may have wild variations for the response times.

Not something to be fixed in this PR, obviously, but if you agree, or have other ideas, would you mind creating an issue for this?

plauric (approved):

jmeg-sfy (approved):

src/python_testing/TC_OpstateCommon.py
```diff
@@ -1311,11 +1321,12 @@ async def TEST_TC_OPSTATE_BASE_2_6(self, endpoint=1):
             self.skip_step(7)
 
         sub_handler.reset()
+        self.step(8)
```
rbultman, line 1324: The else portion of this if/else block should be removed since the step was moved here.

rbultman (changes requested):

src/python_testing/TC_OpstateCommon.py
cecille: I think it's fine - the test is just attempting to verify that the device isn't over reporting on the Q attribute - 10 s should be fine to assess that.

src/python_testing/TC_OpstateCommon.py
cecille: thanks - removed. PTAL.

rbultman (approved):

PR #35765, opened 2024-09-25
setup_payload: restructure python impl to generate wheel

shubhamdp:
- Path fixups in tests which runs against chip-tool and platform scripts
- Update readme and added doc strings
- Changes for building wheel through gn build
- Added python unit tests

Tested building and installing the package locally
```
gn gen out/host
ninja -C out/host python_wheels

python3 -m pip install out/host/python/obj/src/setup_payload/python/matter-setup-payload._build_wheel/matter_setup_payload-1.0.0-py3-none-any.whl
```


src/setup_payload/python/matter_setup_payload/setup_payload.py
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 #
-# Copyright (c) 2024 Project CHIP Authors
+# Copyright (c) 2022-24 Project CHIP Authors
```
andy31415, line 3:
```suggestion
# Copyright (c) 2022-2024 Project CHIP Authors
```

andy31415 (approved):

scripts/tools/bouffalolab/factory_qrcode.py
```diff
@@ -20,13 +20,13 @@
 
 try:
     import qrcode
-    from SetupPayload import CommissioningFlow, SetupPayload
+    from matter_setup_payload import CommissioningFlow, SetupPayload
```
arkq, line 23: Would it be possible to install this package in the `chip.setup_payload` namespace? Currently we have a `setup_payload` module available here `src/controller/python/chip/setup_payload/__init__.py`. Anyway, I'm not sure what is the difference between this new package and current `from chip.setup_payload import ...`. The case is that it would be nice (IMHO) to have all Python packages created by matter SDK repo available under a common namespace, i.e.: `chip.` (since we have this namespace now).

shubhamdp: `src/controller/python/chip/setup_payload` is a wrapper on top of C++ code, and this is written in Python.

> from chip.setup_payload import .... The case is that it would be nice (IMHO)

Agreed, will restructure.

> If there is something common

Both functionally do the same thing, and yes, some utilities which only requires generating onboarding payload will not need the complete chip-core. So, this has to be split out.

PR #35770, opened 2024-09-25
Define CI tests arguments in YAML format

arkq:
### Problem

Current format for CI test arguments is hard to read/modify in case when lots of arguments needs to be passed to the test script. Unfortunately, the format does not allow line breaks...

This PR proposes YAML as a format for defining CI test arguments. It is well-known and supports line breaks! For now, the YAML parser works alongside the old format, but the intention of this PR is to remove support for old format after migrating all test to YAML.

### Changes

- add support for YAML when parsing metadata.
- convert ACE and ACL tests to show benefits of new format

### Testing

CI will verify

### DIFF
--- a/src/python_testing/matter_testing_infrastructure/chip/testing/metadata.py
+++ b/src/python_testing/matter_testing_infrastructure/chip/testing/metadata.py
@@ -12,14 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import logging
 import re
-import sys
 from dataclasses import dataclass
+from io import StringIO
 from typing import Any, Dict, List
 
 import yaml
 
 
+# TODO #35787: Remove support for non-YAML format
 def cast_to_bool(value: Any) -> bool:
     """Convert True/true/False/false strings to bool."""
     if isinstance(value, str):
@@ -81,6 +83,12 @@ class Metadata:
             self.quiet = cast_to_bool(attr_dict["quiet"])
 
 
+class NamedStringIO(StringIO):
+    def __init__(self, content, name):
+        super().__init__(content)
+        self.name = name
+
+
 def extract_runs_arg_lines(py_script_path: str) -> Dict[str, Dict[str, str]]:
     """Extract the run arguments from the CI test arguments blocks."""
 
@@ -99,6 +107,10 @@ def extract_runs_arg_lines(py_script_path: str) -> Dict[str, Dict[str, str]]:
             line = line.strip()
             line_num = line_idx + 1
 
+            # Append empty line to the line capture, so during YAML parsing
+            # line numbers will match the original file.
+            ci_args_section_lines.append("")
+
             # Detect the single CI args section, to skip the lines otherwise.
             if not done_ci_args_section and line.startswith("# === BEGIN CI TEST ARGUMENTS ==="):
                 found_ci_args_section = True
@@ -109,13 +121,14 @@ def extract_runs_arg_lines(py_script_path: str) -> Dict[str, Dict[str, str]]:
                 continue
 
             if found_ci_args_section:
-                ci_args_section_lines.append(line.lstrip("#"))
+                # Update the last line in the line capture.
+                ci_args_section_lines[-1] = " " + line.lstrip("#")
 
             runs_match = runs_def_ptrn.match(line)
             args_match = arg_def_ptrn.match(line)
 
             if not found_ci_args_section and (runs_match or args_match):
-                print(f"WARNING: {py_script_path}:{line_num}: Found CI args outside of CI TEST ARGUMENTS block!", file=sys.stderr)
+                logging.warning(f"{py_script_path}:{line_num}: Found CI args outside of CI TEST ARGUMENTS block")
                 continue
 
             if runs_match:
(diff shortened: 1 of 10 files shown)


src/python_testing/matter_testing_infrastructure/chip/testing/metadata.py
```diff
@@ -20,9 +20,11 @@
 import yaml
 
 
-def bool_from_str(value: str) -> bool:
+def cast_to_bool(value: Any) -> bool:
```
andy31415, line 23: why is this needed? would yaml not auto-convert "true/false" to a bool type rather than having everything string?

src/python_testing/matter_testing_infrastructure/chip/testing/metadata.py
```diff
@@ -119,6 +127,17 @@ def extract_runs_arg_lines(py_script_path: str) -> Dict[str, Dict[str, str]]:
...
+    if not runs_arg_lines:
+        try:
+            runs = yaml.safe_load("\n".join(ci_args_section_lines))
+            for run, args in runs.get("test-runner-runs", {}).items():
+                runs_arg_lines[run] = {}
+                runs_arg_lines[run]['run'] = run
+                runs_arg_lines[run]['py_script_path'] = py_script_path
+                runs_arg_lines[run].update(args)
+        except yaml.YAMLError as e:
+            print(f"ERROR: {py_script_path}: Failed to parse CI arguments YAML: {e}", file=sys.stderr)
```
andy31415, line 139: please use logging.error or similar. Printing out to stderr is frustrating to use because you have to do extra work to grep.

maybe printing the stack trace would also help here (via the traceback module)

src/python_testing/matter_testing_infrastructure/chip/testing/metadata.py
andy31415: ah, this is for the non-yaml case since we support both. I wonder if we should explain (and maybe eventually TODO to remove the old syntax and leave only one syntax available)

andy31415 (approved):

vatsalghelani-csa:
@arkq We need to coordinate merging here, due to SVE and moving of packages/introducing new changes to the testing infrastructure with one of my [PR #34785](https://github.com/project-chip/connectedhomeip/pull/34785) that is already awaiting merge, probably I believe we would be able to start merging next week.

arkq:
> We need to coordinate merging here, due to SVE and moving of packages/introducing new changes to the testing infrastructure with one of my https://github.com/project-chip/connectedhomeip/pull/34785 that is already awaiting merge

@vatsalghelani-csa OK. Anyway, there should be no conflicts between your PR and this one (yours does not touch metadata.py, only moves it). Also, this PR is 100% backward compatible. However, we can coordinate the follow-up PR which will remove support for non-YAML syntax. Also, I'd like to somehow go forward with https://github.com/project-chip/connectedhomeip/pull/35249 which together with this one will allow better wrapping for all `--*-args` arguments.

src/python_testing/matter_testing_infrastructure/chip/testing/metadata.py
arkq: I've added TODO item: https://github.com/project-chip/connectedhomeip/issues/35787

src/python_testing/matter_testing_infrastructure/chip/testing/metadata.py
arkq: Replaced all prints to logging.

> maybe printing the stack trace would also help here (via the traceback module)

Printing stack trace will not add any meaningful info in that case. I've modified error reporting a little bit, so right now the error will include proper line and column within the parsed test script, e.g:

```
$ ./scripts/tests/run_python_test.py --script src/python_testing/TC_ECOINFO_2_1.py --load-from-env=test-env.yaml
ERROR:root:Failed to parse CI arguments YAML: mapping values are not allowed here
  in "src/python_testing/TC_ECOINFO_2_1.py", line 24, column 10
  File ".../connectedhomeip/./scripts/tests/run_python_test.py", line 238, in <module>
    main(auto_envvar_prefix='CHIP')
  File ".../connectedhomeip/.environment/pigweed-venv/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/home/a.bokowy/connectedhomeip/.environment/pigweed-venv/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File ".../connectedhomeip/.environment/pigweed-venv/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File ".../connectedhomeip/.environment/pigweed-venv/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File ".../connectedhomeip/./scripts/tests/run_python_test.py", line 118, in main
    raise Exception(
Exception: No valid runs were found. Make sure you add runs to your file, see https://github.com/project-chip/connectedhomeip/blob/master/docs/testing/python.md document for reference/example.
```

saurabhst (approved):

PR #35772, opened 2024-09-25
Remove unused time update code from MTRDevice.

bzbarsky-apple:
_scheduleNextUpdate is only called from _updateDeviceTimeAndScheduleNextUpdate.

__updateDeviceTimeAndScheduleNextUpdate is only called from _performScheduledTimeUpdate.

_performScheduledTimeUpdate is only called from _scheduleNextUpdate.

So all three are unused in practice and can be removed.

At that point, timeUpdateScheduled, timeSyncLock, and _setTimeOnDevice become unused and can be removed.

Then _endpointsWithTimeSyncClusterServer, _setUTCTime, and _setDSTOffsets become unused and can be removed.

temporaryMetaDataCache was just unused on both MTRDevice and MTRDevice_Concrete.

_getInternalState is only used on MTRDevice_Concrete, and then _internalDeviceState is unused on MTRDevice and can be removed.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice.mm
@@ -212,23 +212,8 @@ using namespace chip::Tracing::DarwinFramework;
 @end
 
 @interface MTRDevice ()
-// protects against concurrent time updates by guarding timeUpdateScheduled flag which manages time updates scheduling,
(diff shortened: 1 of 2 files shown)


kiel-apple (approved):

bzbarsky-apple:

PR #35773, opened 2024-09-25
Add Camera clusters XML

gmarcosb:
Add camera clusters from https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/10004

Generated with (with some hand-edits to work with ZAP compile):
```
for camera_cluster in 'WebRTC_Requestor.adoc' 'WebRTC_Provider.adoc' 'Chime.adoc' 'PushAVStreamTransport.adoc' 'CameraAVStreamManagement.adoc' 'CameraAVSettingsUserLevelManagement.adoc'
do
 ./alchemy zap --attribute="in-progress"  --sdkRoot=[] --specRoot=[] camera_cluster
done
 ./scripts/tools/zap_regen_all.py
```

**Note to reviewer:** if you prefer, I'm happy to split this into 1-pull-per-cluster, I just thought it may be easier to review them together as they share structures & concepts

Note that this pull also includes some necessary dependent changes (which, I'm also happy to split out to separate pull if preferred)
1) Addition of `ThreeLevelAutoEnum`, which [although defined in the spec] did not seem to have a type spec
2) Add missing schema definitions to `zcl.xsd`


gmarcosb:
@aronrosenberg FYI as this is the zap generation for the spec pull you have

.github/workflows/tests.yaml
bzbarsky-apple, line 1: Please do split these out.  Finding several small chunks of time to review separate PRs is _much_ easier than finding a big chunk of time to review a single large PR.  Splitting things out also allows the review work to be parallelized across more people, in general...

Note that there are merge issues here with ZLL_COMMISSIONING_CLUSTER being incorrectly re-introduced and whatnot.  Those would be easier to catch in a smaller PR.

In terms of shared concepts... those are not really relevant to reviewing whether the spec was correctly translated into the XML.

By the way, whatever hand-edits you had to make should probably be turned into alchemy issues, if they were edits to alchemy output?

gmarcosb: Sounds good, will split it out!

Had some general questions here though before doing that:

1) I put these in `/draft` - is that correct, or should I just put them in `/chip`?
2) Hand edits: the camera spec is dependent on ~typedefs, e.g. `VideoStreamID: This data type is derived from uint16` - but there doesn't seem to be a way to specify that in the zap xml schema, so I think it'd actually need zap generation changes
3) structs used across clusters: should I put these in `global-structs` as I have here, have them defined in one of the clusters, or something else altogether?
4) If I don't add a `<cluster code="0x0551"/>` to those shared structures, zap generation ends up generating it in every single `.matter` file (even non-camera ones); if I add that restriction, though, it seems it causes duplicate definition (see build errors: multiple definition of ‘enum class chip::app::Clusters::detail::StreamTypeEnum’) - what's the right way of dealing with these shared structures?

Thanks Boris!

bzbarsky-apple: 1. Good question.  I think the "draft" bits got removed in #35583 so putting them in `draft` likely does not do the right thing.
2. Ah, yes, I discussed this with @andyg-apple just yesterday.  Please talk to @hasty about Alchemy handling this by desugaring the typedefs for now?
3. global-structs should be for actually global structs, which are not tied to any cluster or list of clusters at all.  There are some open questions about which, if any, of the camera structs will end up that way.  For now, I'd put them in one of the clusters, or a new camera-structs.xml or something.
4. For global things (not associated with any clusters), listing them in all .matter files is correct.  Things that are associated with clusters should work; please drop me a link on Slack to a branch or SHA that has the duplicate definition error for you, and I will take a look.

gmarcosb: 1. As far as I can tell, as long as the `/draft` path is specified in `xmlRoot` of `zcl.json`, it seems to work & generate the files - but if we're moving away from the `/draft` path (which maybe seems to be the case?) then I'll just put these in `/chip` (let me know otherwise)
2. Sounds good, I'll make the changes to alchemy, shouldn't be a problem (the files it will generate should look like my hand-edited ones, but I'll make the changes & verify in parallel)
3. Will do 👍
4. Will do 👍

gmarcosb: First smaller PR @ https://github.com/project-chip/connectedhomeip/pull/35775

bzbarsky-apple: 1. Right, but it's not specified there anymore, right?  I figure putting these in `chip` makes sense.

gmarcosb:
Closing as this has been split out to multiple PRs

PR #35780, opened 2024-09-25
Push down the deviceController ivar into MTRDevice subclasses.

bzbarsky-apple:
This allows them to store the specific controller types they are associated with, which will allow us to move APIs that only make sense for a particular type to that type.


bzbarsky-apple:
Going to take a different approach here.

PR #35781, opened 2024-09-26
Remove (non-working) implementations of various commissioning/session APIs from MTRDeviceController base class.

bzbarsky-apple:
Also removes some API implementation bits from MTRDeviceController_Concrete, since those are implemented in MTRDeviceController and delegate to APIs or internal methods that subclasses should then implement.

Documents a few more things MTRDeviceController_XPC may need to implement at some point.


kiel-apple (approved):

anush-apple (approved):

bzbarsky-apple:

PR #35782, opened 2024-09-26
Go back to macos-13 for the Darwin framework tests.

bzbarsky-apple:
The macos-14 runners are random-failing a lot.  Rolling back to the previous runner version while we investigate that.


jtung-apple (approved):

bzbarsky-apple:

PR #35786, opened 2024-09-26
[Tizen] Run tests from 9P-shared directory instead of ISO follow-up

XToripuru:
### Issue

Currently the way tests and runner are passed onto QEMU is by mounting an iso image containing these. The problem with this approach is that building the image is slow, and does not allow to share output files back with host which will be needed in a follow-up PR adding code coverage for tests.

### Solution

Scalable and efficient method is to share tests directory between host and QEMU using virtfs with 9P. This approach involves copying adequate runner to `${root_out_dir}/tests` and then sharing it as `/mnt/chip` on QEMU side. This is highly configurable as the new option`tizen_qemu.py --share` takes a custom path. A service is added which performs the mounting.

### Testing

Tested locally using both interactive and non-interactive modes with and without customized shared directory.

### Extra notes

This PR contains non-Dockerfile changes for the mentioned above issue and solution as they were merged previously (#35745).


arkq (changes requested):
Bump used docker images to 80 in all workflows (especially on Tizen, because in current form this PR disabled all Tizen QEMU tests).

src/test_driver/tizen/chip_tests/BUILD.gn
```diff
@@ -17,30 +17,26 @@ import("//build_overrides/tizen.gni")
...
   virtio_net = true
 
-  deps = [ ":chip-tests-runner" ]
-  mkisofs_outputs = get_target_outputs(":chip-tests-runner")
-  iso_image = rebase_path(mkisofs_outputs[0])
+  # Share tests directory.
+  share = "${root_out_dir}/tests"
+
+  # Copy runner script to shared directory.
+  copy("tests:runner") {
```
arkq, line 30:
```suggestion
  copy("${target_name}:runner") {
```

And the same in the integration test.

third_party/tizen/tizen_qemu.py
```diff
@@ -65,10 +65,8 @@
     help=("path to the system data image; "
           "default: $TIZEN_SDK_ROOT/iot-sysdata.img"))
 parser.add_argument(
-    '--image-iso', metavar='IMAGE',
-    help=("path to the ISO image with the runner script; the ISO image "
-          "should have 'CHIP' label and a file named 'runner.sh' at the "
-          "root directory"))
+    '--share', type=str,
+    help="directory to be shared, runner.sh inside will be run if present")
```
arkq, line 69: Maybe more verbose help, something like:

```suggestion
    help=("host directory to share with the guest; if file named 'runner.sh' "
          "is present at the root of that directory, it will be executed "
          "automatically after boot"))
```

src/test_driver/tizen/chip_tests/BUILD.gn
XToripuru: Done.

arkq (approved):

andy31415 (approved):

PR #35795, opened 2024-09-26
Remove Test*.py scripts from excluded list and enabling its execution via glob python test runner

vatsalghelani-csa:
Work for Issue #31317 

Follow up to PR #35660 

PR 19 - This PR modifies the test infrastructure by removing `Test*.py` scripts from the excluded list and ensuring they are now automatically included in test runs using the glob pattern in the Python test runner script.


andy31415 (changes requested):
Lets not do this: the runner should be an integration runner and not a path of "run arbitrary python code" just because it can.

vatsalghelani-csa:
This PR is closed for now, the work will be carried over as a follow-up task to the [PR #34785](https://github.com/project-chip/connectedhomeip/pull/34785).

PR #35802, opened 2024-09-26
Remove usage of a separate `python.json` environment.

andy31415:
We instead now pull the python according to pigweed setup (so that version is maintained and uniform).

This moves windows to also use python 3.11 (because 3.9 will not work) and removes the mac-amd64 pull of a python3.9 as that is redundant.

### Changes

- remove python3.9 checkout for windows and darwin amd64
- pull python 3.11 for windows
- Fix argument repeat bug in `gen_pigweed_cipd_json.py`
- Add some logging to `gen_pigweed_cipd_json.py` to be able to see what it does
- remove `renode` cipd pull since we do not seem to use that one

### DIFF
--- a/scripts/setup/bootstrap.sh
+++ b/scripts/setup/bootstrap.sh
@@ -141,7 +141,8 @@ _bootstrap_or_activate() {
     $_CHIP_ROOT/scripts/setup/gen_pigweed_cipd_json.py \
         -i $_PIGWEED_CIPD_JSON                         \
         -o $_GENERATED_PIGWEED_CIPD_JSON               \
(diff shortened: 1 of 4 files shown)


andy31415:
Apparently this fails because github includes a python 3.12 by default in their runners: https://github.com/actions/runner-images/blob/macos-13/20240916.98/images/macos/macos-13-Readme.md

however why that interacts oddly between python 3.9 and 3.11 is unclear to me ... we should have had installed python 3.11 and have things just work.

tcarmelveilleux (approved):

arkq (approved):

PR #35805, opened 2024-09-26
FabricSync example changed to use ScopedNodeId in some locations

### LINKED ISSUE #33221: [Feature] Integrate KeepActive command in BRBINFO cluster into fabric-bridge fabric-admin examples

tehampson:
Problem:
* In some spot within FabricAdmin we should be using ScopedNodeId instead of NodeId.
* Relying on just NodeId as a device ID between bridge and admin has the issue admin side might have multiple fabric.

Solution:
* On the fabric-admin side migrate KeepActive command to filter based on ScopedNodeId instead of NodeId
* When marshaling data between bridge and admin use a different identifier that is guaranteed to be unique.

Alternative Considered:
* Update RPC proto to pass ScopedNodeId between admin <-->bridge instead of NodeId.
  * While this would have worked, it was odd for bridge to hold onto ScopedNodeId as this information only makes sense in the context for the admin side.
  * This solution isn't easily extendable. In the future there might be more data that the fabric admin will need to keep track of for devices being enumerated on the bridge. Marshaling this data back and forth would become increasingly unsustainable

Note: Normally changing a proto should not be done. It is done in this PR as we know that the proto is for examples compiled from source and running on the same machine and not accross networks where it is more difficult to ensure everything is updated all at once.

Fixes: #33221

Test:
* CI passes
* Locally tests that BRBINFO_4_1 test still passes


examples/fabric-admin/BUILD.gn
```diff
@@ -80,6 +80,8 @@ static_library("fabric-admin-utils") {
     "commands/pairing/OpenCommissioningWindowCommand.h",
     "commands/pairing/PairingCommand.cpp",
     "commands/pairing/ToTLVCert.cpp",
+    "device_manager/BridgeAdminDeviceMapper.cpp",
```
tehampson, line 83: I am open to alternative naming suggestion. I always find naming hard

examples/fabric-admin/device_manager/BridgeAdminDeviceMapper.h
```diff
@@ -0,0 +1,58 @@
...
+
+    std::optional<uint64_t> GetHandleForBridge(const chip::ScopedNodeId & scopedNodeId);
+    std::optional<chip::ScopedNodeId> GetScopedNodeIdForAdmin(uint64_t handle);
+
+private:
+    uint64_t mNextHandle = 0;
+    // If admin side ever needs more data other than ScopedNodeId we can change
+    // mHandleToScopedNodeId value type from ScopedNodeId to AdminDeviceInfo (or something
+    // of that nature).
+    std::unordered_map<uint64_t, chip::ScopedNodeId> mHandleToScopedNodeId;
```
yufengwangca, line 56: The handle in this context seems act as an additional layer of abstraction or indirection, the NodeId is unique, and its corresponding ScopedNodeId is also unique. So why do we need an additional translation layer that requires dynamic allocation and maintenance?

examples/fabric-admin/device_manager/BridgeAdminDeviceMapper.h
```diff
@@ -0,0 +1,58 @@
...
+    std::unordered_map<chip::ScopedNodeId, uint64_t, ScopedNodeIdHasher> mScopedNodeIdToHandle;
```
yufengwangca, line 57: is ScopedNodeIdHasher same as handle?

examples/fabric-admin/device_manager/BridgeAdminDeviceMapper.h
```diff
@@ -0,0 +1,58 @@
...
+#include <unordered_map>
+
+#include <lib/core/DataModelTypes.h>
+#include <lib/core/ScopedNodeId.h>
+
+struct ScopedNodeIdHasher
+{
+    std::size_t operator()(const chip::ScopedNodeId & scopedNodeId) const
+    {
+        std::size_t h1 = std::hash<uint64_t>{}(scopedNodeId.GetFabricIndex());
```
yufengwangca, line 33: The NodeId within ScopedNodeId is 64-bit, so is it possible for different ScopedNodeId instances to map to the same hash value, even if the chances are minimized?

examples/common/pigweed/protos/fabric_admin_service.proto
```diff
@@ -6,7 +6,7 @@ package chip.rpc;
 
 // Define the message for a synchronized end device with necessary fields
 message DeviceCommissioningWindowInfo {
-  uint64 node_id = 1;
+  uint64 handle = 1;
```
yufengwangca, line 9: Why not just use ScopedNodeId instead of node_id?

examples/fabric-admin/BUILD.gn
yufengwangca: ScopedNodeIdHandleMapper?  Reflects the role of mapping between ScopedNodeId and handles. But I am not clear why can't we just use ScopedNodeId directly.

examples/fabric-admin/device_manager/BridgeAdminDeviceMapper.h
tehampson: https://en.cppreference.com/w/cpp/container/unordered_map

It is the Hash class. Because `std::hash<ScopedNodeId>` is non-existant we need to define it. I didn't want to define it in the struct of ScopedNodeId. But I can look into do that as a follow up PR where just that is being evaluated

examples/fabric-admin/device_manager/BridgeAdminDeviceMapper.h
tehampson: Collisions are fine. Overall a hash table is best case O(1), when there isn't a collision and worst case of O(n) when there are collisions all the time

https://en.wikipedia.org/wiki/Hash_table#Collision_resolution

examples/common/pigweed/protos/fabric_admin_service.proto
tehampson: This was discussed in the PR's description

image

examples/fabric-admin/device_manager/BridgeAdminDeviceMapper.h
tehampson: >The handle in this context seems act as an additional layer of abstraction or indirection

This is exactly what it is, aggregator code shouldn't care and need to marshal data that is relevant to admin. While today it one thing, in the future it can be more data.  I can move to marshaling the ScopedNodeId data instead, but as I mentioned in the PR description on why I didn't not go with that alternative approach. I am okay with changing to the alternative approach if we discuss it based on those merits.

>NodeId is unique

NodeId is not. It is unique on a per fabric basis, but it is not unique. 

>So why do we need an additional translation layer that requires dynamic allocation and maintenance?

We either need to pass the ScopedNodeId back and forth, or we need this translation layer. I am more in favor of forward thinking solutions as I don't think marshaling data to systems where that data is meaningless is not scalable only for it to pass that data back, in all other systems that I have worked with this is where you introduce a handle of some sort which is what I was trying to do here.

tehampson:
Went with alternative approach: https://github.com/project-chip/connectedhomeip/pull/35864

PR #35810, opened 2024-09-27
Add ECOINFO_2_1 and ECOINFO_2_2 to CI

### LINKED ISSUE #35482: [Workflow] Run ECOINFO tests in CI
### Workflow issues

Integrate the following FS-related tests into CI:

- [x] ECOINFO_2_1
- [x] ECOINFO_2_2

### Platform

_No response_

### Anything else?

_No response_

arkq:
Fixes https://github.com/project-chip/connectedhomeip/issues/35482

### Changes

- add ECOINFO_2_1 and ECOINFO_2_2 to CI
- make Subprocess wrapper class even more generic, so in a follow-up PR we can use it in more places
- convert TC_MCORE_FS CI arguments to YAML

### Testing

CI will verify


src/python_testing/TC_ECOINFO_2_1.py
```diff
@@ -15,20 +15,92 @@
...
+            self.th_server = AppServer(
+                th_server_app,
+                storage_dir=self.storage.name,
+                port=self.th_server_port,
+                discriminator=self.th_server_discriminator,
+                passcode=self.th_server_passcode)
+            self.th_server.start()
+
+            # Add some server to the DUT_FSA's Aggregator/Bridge.
+            self.dut_fsa_stdin.write(
```
tehampson, line 87: `self.dut_fsa_stdin` doesn't exist if dut_fsa_stdin_pipe is False

src/python_testing/TC_ECOINFO_2_1.py
```diff
@@ -15,20 +15,92 @@
...
 
+    @async_test_body
+    async def setup_class(self):
+        super().setup_class()
+
+        self.th_server = None
+        self.storage = None
+
+        # Get the path to the TH_SERVER app from the user params.
+        if th_server_app := self.user_params.get("th_server_app_path"):
```
tehampson, line 64: You should only be doing this if you are in CI, otherwise the real test should NOT be creating TH_SERVER. That is not in the test plan.

src/python_testing/matter_testing_infrastructure/chip/testing/tasks.py
```diff
@@ -13,57 +13,70 @@
...
         self.program = program
         self.args = args
         self.output_cb = output_cb
-        self.expected_output = None
+        self.f_stdout = f_stdout
+        self.f_stderr = f_stderr
+        self.output_match = None
+        self.returncode = None
+
+    def _set_output_match(self, patter: Union[str, re.Pattern]):
```
tehampson, line 68: nit,
```suggestion
    def _set_output_match(self, pattern: Union[str, re.Pattern]):
```

src/python_testing/TC_ECOINFO_2_2.py
```diff
@@ -15,32 +15,97 @@
...
+        self.storage = None
+
+        # Get the path to the TH_SERVER app from the user params.
+        if th_server_app := self.user_params.get("th_server_app_path"):
+            # Create a temporary storage directory for keeping KVS files.
+            self.storage = tempfile.TemporaryDirectory(prefix=self.__class__.__name__)
+            logging.info("Temporary storage directory: %s", self.storage.name)
+
+            # Get the named pipe path for the DUT_FSA app input from the user params.
+            if dut_fsa_stdin_pipe := self.user_params.get("dut_fsa_stdin_pipe"):
```
tehampson, line 69: If this is CI we should be making sure that a value is provided here for `dut_fsa_stdin_pipe`

src/python_testing/TC_ECOINFO_2_1.py
arkq: The intention here is that the `th_server_app_path` is going to be supplied on the CI only. However, I can add `if self.is_ci` guard here.

src/python_testing/TC_ECOINFO_2_2.py
arkq: Fixed

src/python_testing/TC_ECOINFO_2_1.py
arkq: >  because this is the same name (th_server_app_path) as the other test there is bound to be someone that accidentally uses this outside of CI with this current PR

Added `self.is_pics_sdk_ci_only` guard before running CI-related setup.

src/python_testing/TC_ECOINFO_2_1.py
arkq: Fixed

saurabhst (approved):

tehampson (approved):

PR #35817, opened 2024-09-27
[chiptests] Check the exit code of the subprocesses and raise a failure if the exit code is not 0 AND if the process was responding

vivien-apple:
#### Problem

The script at `./scripts/tests/run_test_suite.py` ignore the subprocesses return code. It could be an issue for multiple cases.

One example is when the scripts is runned using: 
```
$ LSAN_OPTIONS="detect_leaks=1 suppressions=scripts/tests/chiptest/lsan-mac-suppressions.txt" ./scripts/tests/run_test_suite.py  --target TestConfigVariables  run  --iterations 1
```

Even if there is leaks, the script will say that everything is alright.


scripts/tests/chiptest/test_definition.py
```diff
@@ -159,10 +159,15 @@ def __terminateProcess(self):
         if self.process:
             self.process.terminate()  # sends SIGTERM
             try:
-                self.process.wait(10)
+                exit_code = self.process.wait(10)
+                if exit_code:
+                    raise Exception('Subprocess failed with exit code: %d' % exit_code)
```
arkq, line 164: Why there is an exception raised in case when the `returncode !=0` but not when the process did not respond to `SIGTERM` (which also is some kind of incorrect behavior)?

vivien-apple: I agree that it is some kind of incorrect behaviour and ideally I would like to raise an exception.
This is not done in this patch for the reason that I would like to stop us from introducing new leaks and right now the leaks report is ignored - which is very bad.
So this particular patch targets this use case and not the one were the process hangs while I agree that we should ultimately raise an exception too.

vivien-apple:
Looking at the CI failures:
```
ERROR   TV_AccountLoginCluster         - FAILED in 2.05 seconds
...
Exception: Subprocess failed with exit code: -15
```

It looks like the `examples/tv-*` apps does not handle `SIGTERM` (and thought ignore the proper cleanup code path).
`SIGTERM` is disabled because those are built with the `chip_build_libshell` flag turned on.
This has the side effect of preventing `Asan` to detect leaks...

I first try to disable it for this particular type of tests but the app fails to cleanup properly when built with `Asan`.

```
[1727448871.815] [36127:5655384] [BLE] ConnectionDelegate CancelConnection
[1727448871.815] [36127:5655384] [DL] Inet Layer shutdown
[1727448871.815] [36127:5655384] [DL] BLE Layer shutdown
[1727448871.815] [36127:5655384] [DL] System Layer shutdown
[1727448871.815] [36127:5655384] [IN] SecureSession[0x6120000040c0]: Released - Type:2 LSID:1399
AddressSanitizer:DEADLYSIGNAL
=================================================================
==36127==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x000103374f5c bp 0x00016d6e5db0 sp 0x00016d6e58a0 T0)
==36127==The signal is caused by a READ memory access.
==36127==Hint: address points to the zero page.
    #0 0x103374f5c in chip::AddressResolve::Impl::Resolver::ReArmTimer() AddressResolve_DefaultImpl.cpp:398
    #1 0x103376ca8 in chip::AddressResolve::Impl::Resolver::Shutdown() AddressResolve_DefaultImpl.cpp:271
    #2 0x1032c6e38 in chip::CASESessionManager::Shutdown() CASESessionManager.cpp:34
    #3 0x103701f6c in chip::Server::Shutdown() Server.cpp:604
    #4 0x102722ab8 in ChipLinuxAppMainLoop(AppMainLoopImplementation*) AppMain.cpp:690
    #5 0x10271d538 in main main.cpp:90
    #6 0x18c054270  (<unknown module>)

==36127==Register values:
 x[0] = 0x000000016d6e5ab0   x[1] = 0x000000016d6e5e00   x[2] = 0x0000000000000000   x[3] = 0x000000016d6e5ca8  
 x[4] = 0x0000000000000018   x[5] = 0x0000000000000018   x[6] = 0x0000000103b20000   x[7] = 0x0000000000000001  
 x[8] = 0x0000007000020000   x[9] = 0x0000000000000000  x[10] = 0x0000000000000002  x[11] = 0x00000001039ca000  
x[12] = 0x000000702dafcb54  x[13] = 0x000000016d6e58a8  x[14] = 0x000000016d6e58a0  x[15] = 0x00007fffffffffff  
x[16] = 0x000000010273eaf8  x[17] = 0x0000000104f885d0  x[18] = 0x0000000000000000  x[19] = 0x000000016d6e5b00  
x[20] = 0x00000001f13f00a0  x[21] = 0x00000001f13f0050  x[22] = 0x000000016d6e6f38  x[23] = 0x000000016d6e6f38  
x[24] = 0x000000018c04e000  x[25] = 0x0000000000000000  x[26] = 0x0000000000000000  x[27] = 0x0000000000000000  
x[28] = 0x0000000000000000     fp = 0x000000016d6e5db0     lr = 0x0000000103376cac     sp = 0x000000016d6e58a0  
AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV AddressResolve_DefaultImpl.cpp:398 in chip::AddressResolve::Impl::Resolver::ReArmTimer()
```

So for the moment I will just ignore the exit code `-15` and will open a followup for fixing the tv apps.

scripts/tests/chiptest/test_definition.py
```diff
@@ -159,10 +159,20 @@ def __terminateProcess(self):
...
             self.process.terminate()  # sends SIGTERM
             try:
-                self.process.wait(10)
+                exit_code = self.process.wait(10)
+                # Ideally, we should raise an exception for all non-zero exit codes, but some
+                # applications (notably the tv-* apps) are not built with proper SIGTERM
+                # handling and will return -15 instead of 0.
+                # These apps should be built with SIGTERM handling, but currently, they crash
+                # on shutdown if it is enabled.
+                if exit_code and exit_code != -15:
```
bzbarsky-apple, line 168: Should this use `signal.SIGTERM` instead of hardcoding -15?

bzbarsky-apple (approved):

scripts/tests/chiptest/test_definition.py
```diff
@@ -159,10 +159,20 @@ def __terminateProcess(self):
         if self.process:
             self.process.terminate()  # sends SIGTERM
             try:
-                self.process.wait(10)
+                exit_code = self.process.wait(10)
+                # Ideally, we should raise an exception for all non-zero exit codes, but some
+                # applications (notably the tv-* apps) are not built with proper SIGTERM
```
andy31415, line 164: can we fix these apps instead?

bzbarsky-apple: @andy31415 We can try, but it's a pretty large amount of work best done by the people who understand those apps.  I don't think we should block this PR on it.

vivien-apple: I would also prefer if these apps were fixed, especially since they have code paths that likely aren’t exercised by other apps, making them ideal for leak checks. However, as Boris pointed out, fixing them would require significant effort and is better addressed by the developers familiar with this particular use case. This PR, on the other hand, aims to stop leaks in all the other, more commonly used, cases.

arkq (approved):

scripts/tests/chiptest/test_definition.py
vivien-apple: Spent a little bit more time. https://github.com/project-chip/connectedhomeip/pull/35843 is just moving one line of code around and fix the crashes...

vivien-apple:
I have removed the code that checks whether the return value of the subprocess is -signal.SIGTERM. Instead, I’ve opened a PR (https://github.com/project-chip/connectedhomeip/pull/35848) that should prevent the tv apps from disabling SIGTERM handling

PR #35821, opened 2024-09-27
[ESP32] Implement providers which reads unique data from the secure cert partition

shubhamdp:
- Right now secure cert partition stores only device attestation data i.e. DAC and PAI. This can be extended to store the sort of unique data that is discriminator, verifier/salt/iteration-count for passcode, and unique-id for generating rotating id.
- Update the esp_secure_cert_mgr version to 2.5.0

#### Tests
Flashed the secure cert partition which contains the above mentioned data and commissioned the ESP32

### DIFF
--- a/src/platform/ESP32/ESP32SecureCertDataProvider.cpp
+++ b/src/platform/ESP32/ESP32SecureCertDataProvider.cpp
@@ -133,8 +133,12 @@ CHIP_ERROR ESP32SecureCertDataProvider::GetSpake2pVerifier(MutableByteSpan & ver
 
 CHIP_ERROR ESP32SecureCertDataProvider::GetRotatingDeviceIdUniqueId(MutableByteSpan & uniqueIdSpan)
 {
+#if CHIP_ENABLE_ROTATING_DEVICE_ID
     ScopedTLVInfo tlvInfo(MatterTLVSubType::kRotatingDeviceIdUniqueId);
     return tlvInfo.GetValue(uniqueIdSpan);
+#else
+    return CHIP_ERROR_NOT_IMPLEMENTED;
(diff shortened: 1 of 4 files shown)


src/platform/ESP32/ESP32SecureCertDataProvider.cpp
```diff
@@ -0,0 +1,141 @@
...
+    ScopedTLVInfo tlvInfo(MatterTLVSubType::kSpake2pVerifier);
+    ReturnErrorOnFailure(tlvInfo.GetValue(verifierBuf));
+    verifierLen = verifierBuf.size();
+    return CHIP_NO_ERROR;
+}
+
+CHIP_ERROR ESP32SecureCertDataProvider::GetRotatingDeviceIdUniqueId(MutableByteSpan & uniqueIdSpan)
+{
+    ScopedTLVInfo tlvInfo(MatterTLVSubType::kRotatingDeviceIdUniqueId);
+    return tlvInfo.GetValue(uniqueIdSpan);
```
dhrishi, line 137: What if the rotating device unique id is not present in the partition? And everything else is

shubhamdp: In context of this API, it will return an error, if value is not found or if it fails any constraint check.

And the consumer of this API (most likely the ones building the commissionable data for advertising), will not start the advertising.

shubhamdp: Meanwhile added the #if checks if the feature is not enabled.

dhrishi (approved):

andy31415 (approved):

PR #35824, opened 2024-09-27
[AddressResolver] Add a check in shutdown to avoid some crashes if so…

vivien-apple:
…meone calls it multiple times

#### Problem

Some apps calls multiple times `Shutdown` on the resolver and it crashes a bit after since it tries to access a null pointer. It won't hurt much to just null check `mSystemLayer` and prevent doing the shutdown twice.

### DIFF
--- a/src/lib/address_resolve/AddressResolve_DefaultImpl.cpp
+++ b/src/lib/address_resolve/AddressResolve_DefaultImpl.cpp
@@ -248,6 +248,8 @@ CHIP_ERROR Resolver::Init(System::Layer * systemLayer)
 
 void Resolver::Shutdown()
 {
+    // mSystemLayer is set in ::Init, so if it's null that means the resolver
+    // has not been initialized or has already been shut down.
     VerifyOrReturn(mSystemLayer != nullptr);
 
     while (mActiveLookups.begin() != mActiveLookups.end())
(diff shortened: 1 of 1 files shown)


src/lib/address_resolve/AddressResolve_DefaultImpl.cpp
```diff
@@ -248,6 +248,8 @@ CHIP_ERROR Resolver::Init(System::Layer * systemLayer)
 
 void Resolver::Shutdown()
 {
+    VerifyOrReturn(mSystemLayer != nullptr);
```
andy31415, line 251:
```suggestion
    // mSystemLayer is set in ::Init, so we use this as an indicator of the resolver
    // not being yet initialised or having already been shut down.
    VerifyOrReturn(mSystemLayer != nullptr);
```

andy31415 (approved):

bzbarsky-apple (approved):

andy31415:
@vivien-apple could you update the summary to describe "what" applications do this? I can see an argument for the code to be tolerant to multiple shutdowns, however I would also have an argument that application behaviour should be well controlled and known. If they call shutdown multiple times, would they maybe call init multiple times? Should we guard the init as well or should we add logging when this out of order/extra calls are made?

PR #35829, opened 2024-09-27
Make the log download test self-contained.

bzbarsky-apple:
It should not be depending on a helper app started in a particular way; it should just start that helper app itself.

### DIFF
--- a/src/darwin/Framework/CHIPTests/MTRXPCListenerSampleTests.m
+++ b/src/darwin/Framework/CHIPTests/MTRXPCListenerSampleTests.m
@@ -419,6 +419,7 @@ static const uint16_t kPairingTimeoutInSeconds = 30;
 static const uint16_t kTimeoutInSeconds = 3;
 static const uint64_t kDeviceId = 0x12344321;
 static NSString * kOnboardingPayload = @"MT:Y.K90SO527JA0648G00";
+static NSString * _Nullable sLogContentFilePath;
 static NSString * kSimpleLogContent = @"This is a simple log\n";
 static const uint16_t kLocalPort = 5541;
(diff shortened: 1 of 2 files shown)


terafin (approved):

src/darwin/Framework/CHIPTests/MTRXPCListenerSampleTests.m
```diff
@@ -501,25 +502,38 @@ - (BOOL)unitTestShouldSkipExpectedValuesForWrite:(MTRDevice *)device
...
 {
     // Global teardown, runs once
-    if (sNeedsStackShutdown) {
-        // We don't need to worry about ResetCommissionee.  If we get here,
-        // we're running only one of our test methods (using
-        // -only-testing:MatterTests/MTROTAProviderTests/testMethodName), since
-        // we did not run test999_TearDown.
-        //        [self shutdownStack];
-    }
+    [self shutdownStack];
```
nivi-apple, line 535: should we delete the file here or we don't care and createFileAtPath will handle it if one exists?

nivi-apple (approved):

src/darwin/Framework/CHIPTests/MTRXPCListenerSampleTests.m
bzbarsky-apple: It's in a temp directory, so presumably it will be deleted whenever space is needed....

But yes, we could delete the file.  Let me add something for that.

kiel-apple (approved):

bzbarsky-apple:

woody-apple (approved):

PR #35833, opened 2024-09-27
Fix, for now, weird subclassing of MTRDevice and MTRDeviceController.

bzbarsky-apple:
Some of our API consumers are apparently relying on implementation details. Stop breaking them for now, until they can fix that.


nivi-apple (approved):

bzbarsky-apple:

PR #35834, opened 2024-09-28
Fix data type check in MTRDevice's invokeCommandWithEndpointID.

bzbarsky-apple:
We were checking two NSStrings for pointer-equality, when we should be testing for logical equality.


jtung-apple (approved):

bzbarsky-apple:
Fast-tracking platform specific fix with platform owner review.

woody-apple (approved):

PR #35841, opened 2024-09-30
Definition of the Camera AV StreamManagement Cluster.

### LINKED ISSUE #35838: Move Stream Type enum out of WebRTCProvider cluster
This enum is actually defined in AVStreamMgmt, so remove it from the WebRTC cluster when AVStreamMgmt has merged.

pidarped:
XML, json and ZAP generated files.

Spec in `cameras` branch.
Draft PR: https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/10004


src/app/zap-templates/zcl/data-model/chip/camera-av-stream-management-cluster.xml
```diff
@@ -0,0 +1,578 @@
...
+    <item fieldId="0" name="SensorWidth" type="int16u"/>
+    <item fieldId="1" name="SensorHeight" type="int16u"/>
+    <item fieldId="2" name="HDRCapable" type="boolean"/>
+    <item fieldId="3" name="MaxFPS" type="int16u"/>
+    <item fieldId="4" name="MaxHDRFPS" type="int16u"/>
+  </struct>
+
+  <struct name="VideoStreamStruct" apiMaturity="provisional">
+    <cluster code="0x0551"/>
+    <item fieldId="0" name="VideoStreamID" type="int16u"/>
```
bzbarsky-apple, line 130: Spec comment: the spec has lots of "MS" in the Default comment, which probably don't make sense.  But since in practice the XML ignores those anyway, it's not a problem for this PR.

src/app/zap-templates/zcl/data-model/chip/camera-av-stream-management-cluster.xml
```diff
@@ -0,0 +1,578 @@
...
+      </feature>
+      <feature bit="6" code="STOR" name="LocalStorage" summary="Local Storage available">
+        <optionalConform/>
+      </feature>
+    </features>
+    <server init="false" tick="false">true</server>
+    <globalAttribute code="0xFFFD" side="either" value="1"/>
+    <attribute code="0x0000" side="server" define="MAX_CONCURRENT_VIDEO_ENCODERS" type="int8u" reportable="true" optional="true">
+      <description>MaxConcurrentVideoEncoders</description>
+      <access op="read" privilege="administer"/>
```
bzbarsky-apple, line 187: Some spec questions here about whether these should be A or M, but the XML matches the spec, ok.

src/app/zap-templates/zcl/data-model/chip/camera-av-stream-management-cluster.xml
```diff
@@ -0,0 +1,578 @@
...
+      <description>ISO</description>
+      <access op="read" privilege="administer"/>
+      <access op="write" privilege="administer"/>
+    </attribute>
+    <attribute code="0x001F" side="server" define="VIEWPORT" type="ViewportStruct" optional="true">
+      <description>Viewport</description>
+      <access op="read" privilege="administer"/>
+    </attribute>
+    <attribute code="0x0020" side="server" define="SPKR_ON_OFF" type="boolean" writable="true" optional="true">
+      <description>SpkrOnOff</description>
```
bzbarsky-apple, line 323: Spec comment: here and for the other speaker attributes, should just spell out "Speaker" instead of the weird abbreviation.

src/app/zap-templates/zcl/data-model/chip/camera-av-stream-management-cluster.xml
```diff
@@ -0,0 +1,578 @@
...
+      <access op="read" privilege="administer"/>
+      <access op="write" privilege="administer"/>
+    </attribute>
+    <attribute code="0x0023" side="server" define="SPKR_MIN_LEVEL" type="int8u" max="254" writable="true" optional="true">
+      <description>SpkrMinLevel</description>
+      <access op="read" privilege="administer"/>
+      <access op="write" privilege="administer"/>
+    </attribute>
+    <attribute code="0x0024" side="server" define="MIC_ON_OFF" type="boolean" writable="true" optional="true">
+      <description>MicOnOff</description>
```
bzbarsky-apple, line 343: And should consider whether "Microphone" should be shortened.

Also "OnOff" is bad for a boolean, since it's not immediately clear whether true is on or off.

src/app/zap-templates/zcl/data-model/chip/camera-av-stream-management-cluster.xml
```diff
@@ -0,0 +1,578 @@
...
+      <access op="read" privilege="administer"/>
+      <access op="write" privilege="administer"/>
+    </attribute>
+    <attribute code="0x0027" side="server" define="MIC_MIN_LEVEL" type="int8u" max="254" writable="true" optional="true">
+      <description>MicMinLevel</description>
+      <access op="read" privilege="administer"/>
+      <access op="write" privilege="administer"/>
+    </attribute>
+    <attribute code="0x0028" side="server" define="MIC_AGC" type="boolean" writable="true" optional="true">
+      <description>MicAGC</description>
```
bzbarsky-apple, line 363: Spec comment: Consider calling this MicAGCEnabled for clarity.

src/app/zap-templates/zcl/data-model/chip/camera-av-stream-management-cluster.xml
```diff
@@ -0,0 +1,578 @@
...
+      <access op="read" privilege="administer"/>
+      <access op="write" privilege="administer"/>
+    </attribute>
+    <attribute code="0x002D" side="server" define="LOCAL_SNAPSHOT_RECORDING_ENABLED" type="boolean" writable="true" optional="true">
+      <description>LocalSnapshotRecordingEnabled</description>
+      <access op="read" privilege="administer"/>
+      <access op="write" privilege="administer"/>
+    </attribute>
+    <attribute code="0x002E" side="server" define="STATUS_LIGHT" type="boolean" writable="true" optional="true">
+      <description>StatusLight</description>
```
bzbarsky-apple, line 390: Spec comment: This really does not seem related to stream management, and seems like it would be better handled by a level control and/or On/Off cluster for the light on a separate endpoint.

src/app/zap-templates/zcl/data-model/chip/camera-av-stream-management-cluster.xml
```diff
@@ -0,0 +1,578 @@
...
+    </command>
+
+    <command code="0x05" source="client" name="VideoStreamModify" optional="true">
+      <description>This command SHALL be used to modify the resolution of a stream specified by the VideoStreamID.</description>
+      <access op="invoke" privilege="administer"/>
+      <arg id="0" name="VideoStreamID" type="int16u"/>
+      <arg id="1" name="Resolution" type="VideoResolutionStruct"/>
+    </command>
+
+    <command code="0x06" source="client" name="VideoStreamDeallocate" optional="true">
```
bzbarsky-apple, line 461: This command allows one client to deallocate a stream a different client, possibly on a different fabric, allocated, right?  Is this intentional?

src/app/zap-templates/zcl/data-model/chip/camera-av-stream-management-cluster.xml
```diff
@@ -0,0 +1,578 @@
...
+    </command>
+
+    <command code="0x0E" source="client" name="SetImageRotation" optional="true">
+      <description>The data fields for this command SHALL be as follows: </description>
+      <arg id="0" name="Angle" type="int16u" default="0" min="0" max="359"/>
+    </command>
+
+    <command code="0x0F" source="client" name="SetImageFlipHorizontal" optional="true">
+      <description>The data fields for this command SHALL be as follows: </description>
+      <arg id="0" name="OnOff" type="boolean"/>
```
bzbarsky-apple, line 520: Again, "OnOff" is a bad name for a boolean.

And yes, the On/Off cluster has it.  We should not copy it.

src/app/zap-templates/zcl/zcl-with-test-extensions.json
```diff
@@ -679,7 +680,60 @@
             "ThreadNetworks",
             "ThreadNetworkTableSize"
         ],
-        "Service Area": ["CurrentArea", "EstimatedEndTime", "FeatureMap"]
+        "Service Area": ["CurrentArea", "EstimatedEndTime", "FeatureMap"],
+        "Camera AV Stream Management": [
+            "MaxConcurrentVideoEncoders",
```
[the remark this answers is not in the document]
bzbarsky-apple: This and the comment about zap_cluster_list are the only actual non-spec issues here, by the way, and this can land once those are addressed.

src/app/zap-templates/zcl/data-model/chip/camera-av-stream-management-cluster.xml
gmarcosb, line 1: You should copy the one from https://github.com/project-chip/connectedhomeip/blob/6d83b821e6d73111f594708af790579ca90ad382/src/app/zap-templates/zcl/data-model/draft/camera-av-stream-management-cluster.xml as it contains some things that are missing here (e.g. the new pointer to the XSD with the new alchemy, description field, etc.)

pidarped: I also had some new changes here related to the definition of the VideoResolution struct. Also, we have a way to define the derived type in chip-types.xml(and a couple of places) but might need to modify the spec too. So, for now, have hardcoded to int16u.

gmarcosb: Ah I missed the spec has an updated name for the struct 👍

Copying the description field & running with the latest alchemy seems like the best approach then 👍 (I've had to do it a couple times as well after applying some of the suggested changes, e.g. `Id` -> `ID`)

src/app/zap-templates/zcl/data-model/chip/camera-av-stream-management-cluster.xml
pidarped: Expanded to Speaker

src/app/zap-templates/zcl/data-model/chip/camera-av-stream-management-cluster.xml
pidarped: Done.

src/app/zap-templates/zcl/data-model/chip/camera-av-stream-management-cluster.xml
pidarped: Changed to Enabled.

src/app/zap-templates/zcl/data-model/chip/camera-av-stream-management-cluster.xml
pidarped: Kept `Mic` but removed `OnOff`.

src/app/zap-templates/zcl/data-model/chip/camera-av-stream-management-cluster.xml
pidarped: Yes, this cluster is a global cluster across fabrics. And the streams are shared/reused. Admins are allowed to de-allocate provided the ref-count on the stream is 0.

src/app/zap-templates/zcl/data-model/chip/camera-av-stream-management-cluster.xml
pidarped: Regenerated using the latest Alchemy

src/app/zap-templates/zcl/data-model/chip/global-enums.xml
```diff
@@ -34,6 +34,18 @@ TODO: Make these structures global rather than defining them for each cluster.
     <item name="High" value="0x02"/>
     <item name="Automatic" value="0x03"/>
   </enum>
+
+  <enum name="StreamTypeEnum" type="enum8">
```
gmarcosb, line 38: This fixes https://github.com/project-chip/connectedhomeip/issues/35838

You'll need to update [webrtc-provider-cluster.xml](https://github.com/project-chip/connectedhomeip/blob/ca79d5910a50e6fb2f82748bd1540153d98601b6/src/app/zap-templates/zcl/data-model/chip/webrtc-provider-cluster.xml#L19) & remove from there

pidarped: That is also done with this PR.

andy31415 (approved):

src/app/zap-templates/zcl/data-model/chip/camera-av-stream-management-cluster.xml
```diff
@@ -0,0 +1,576 @@
...
+    </attribute>
+    <attribute code="0x0002" side="server" define="VIDEO_SENSOR_PARAMS" type="VideoSensorParamsStruct" reportable="true" optional="true">
+      <description>VideoSensorParams</description>
+      <access op="read" privilege="administer"/>
+    </attribute>
+    <attribute code="0x0003" side="server" define="NIGHT_VISION_CAPABLE" type="boolean" reportable="true" optional="true">
+      <description>NightVisionCapable</description>
+      <access op="read" privilege="administer"/>
+    </attribute>
+    <attribute code="0x0004" side="server" define="MIN_VIEW_PORT_WIDTH" type="VideoResolutionStruct" reportable="true" optional="true">
```
bzbarsky-apple, line 199: Spec issue: using VideoResolutionStruct as the type for MinViewPortWidth and MinViewPortHeight makes no sense to me....

src/app/zap-templates/zcl/data-model/chip/camera-av-stream-management-cluster.xml
```diff
@@ -0,0 +1,576 @@
...
+      <description>SpeakerMaxLevel</description>
+      <access op="read" privilege="administer"/>
+      <access op="write" privilege="administer"/>
+    </attribute>
+    <attribute code="0x0023" side="server" define="SPEAKER_MIN_LEVEL" type="int8u" max="254" writable="true" optional="true">
+      <description>SpeakerMinLevel</description>
+      <access op="read" privilege="administer"/>
+      <access op="write" privilege="administer"/>
+    </attribute>
+    <attribute code="0x0024" side="server" define="MIC_STATUS" type="boolean" writable="true" optional="true">
```
bzbarsky-apple, line 340: Spec issue: What does a boolean "MicStatus" mean?  What does true mean?  What does false mean?

This needs a better name.

src/app/zap-templates/zcl/data-model/chip/camera-av-stream-management-cluster.xml
```diff
@@ -0,0 +1,576 @@
...
+      <access op="read" privilege="administer"/>
+      <access op="write" privilege="administer"/>
+    </attribute>
+    <attribute code="0x002D" side="server" define="LOCAL_SNAPSHOT_RECORDING_ENABLED" type="boolean" writable="true" optional="true">
+      <description>LocalSnapshotRecordingEnabled</description>
+      <access op="read" privilege="administer"/>
+      <access op="write" privilege="administer"/>
+    </attribute>
+    <attribute code="0x002E" side="server" define="STATUS_LIGHT" type="boolean" writable="true" optional="true">
+      <description>StatusLight</description>
```
bzbarsky-apple, line 388: OK, if this stays, the same spec comment: What do "true" and "false" mean for an attribute called StatusLight?  Is this about whether it's on?  About whether it's _allowed_ to be on?  Something else entirely?

Needs better naming.

bzbarsky-apple (approved):
Approving, but please make sure the spec comments, both from this review and the previous one, are captured in the spec.

src/app/zap-templates/zcl/data-model/chip/camera-av-stream-management-cluster.xml
pidarped: Fixed in spec and incorporated in xml. The 2 attributes are collapsed to 1 attribute MinViewport of type VideoResolutionStruct(containing width and height).

src/app/zap-templates/zcl/data-model/chip/camera-av-stream-management-cluster.xml
pidarped: Captured comment in spec.

src/app/zap-templates/zcl/data-model/chip/camera-av-stream-management-cluster.xml
pidarped: Captured comment in spec.

src/app/zap-templates/zcl/data-model/chip/camera-av-stream-management-cluster.xml
pidarped: Updated the Access column for all fields in the spec and incorporated in the xml.

src/app/zap-templates/zcl/data-model/chip/camera-av-stream-management-cluster.xml
pidarped: MS values for these attributes have been removed in the Spec.

PR #35849, opened 2024-09-30
TC-DESC-2.2: Fix comparison when mfg label is in the list

cecille:
Unit tests included, also tested on an example app with mfg tags


src/python_testing/taglist_and_topology_test_support.py
```diff
@@ -143,12 +144,16 @@ def create_device_type_list_for_root(direct_children, endpoint_dict: dict[int, A
 
 
 def cmp_tag_list(a: Clusters.Descriptor.Structs.SemanticTagStruct, b: Clusters.Descriptor.Structs.SemanticTagStruct):
+    if type(a.mfgCode) != type(b.mfgCode):
+        return -1 if type(a.mfgCode) is Nullable else 1
```
andy31415, line 148: in various places we seem to have `is NullValue` which seems clearer than checking the type. Could we use that?

image

cecille: That does become rather longer then because I need to check if a is null and b is not, or if b is null and a is not. Whereas what I really want is to know whether I can direct compare these.

andy31415 (approved):

tcarmelveilleux (approved):

PR #35851, opened 2024-09-30
TC-VALCC-3.1: Allow immediate open of valve

cecille:
If the valve can be opened before the command is complete or before the next read is done, the test will fail because it expects the transitioning phase to happen AND it expects to be able to read the attribute in the intermediate state.

In the spec:
When the movement is complete, the device SHALL set the CurrentState attribute to the Open value.

^ this can happen before the end of the command.

The current implementation does actually set the current state to open immediately, but the test doesn't catch this because the current implementation also doesn't set the target back to NULL when the change to CurrentState happens.

Changes to the test:
- start with an close command so the device is in a known state before test start
- change the reads to a subscription

Changes to the current cluster implementation:
- add logic to set target values correctly for level and state
- push attribute changes to the current level through the update functions so they can set the targets accordingly
- remove direct set calls from the delegate

Note that these changes are being done as part of an effort to refactor this cluster entirely, so changes to the implementation will be short-lived and are implemented only to allow landing the test changes before the refactor.


src/python_testing/TC_VALCC_3_1.py
```diff
@@ -75,24 +78,26 @@ async def test_TC_VALCC_3_1(self):
...
 
-        while current_state_dut is Clusters.Objects.ValveConfigurationAndControl.Enums.ValveStateEnum.kTransitioning:
+        while current_state_dut is cluster.Enums.ValveStateEnum.kTransitioning:
+            asserts.assert_equal(target_state_dut, cluster.Enums.ValveStateEnum.kOpen, "TargetState is incorrect")
             time.sleep(1)
 
+            target_state_dut = await self.read_valcc_attribute_expect_success(endpoint=endpoint, attribute=attributes.TargetState)
             current_state_dut = await self.read_valcc_attribute_expect_success(endpoint=endpoint, attribute=attributes.CurrentState)
             asserts.assert_true(current_state_dut is not NullValue, "CurrentState is null")
+            asserts.assert_true(target_state_dut is not NullValue, "TargetState is null")
```
andy31415, line 94: should these nullability asserts also apply outside of the while? It seems like conditions should always be consistent except we have a `retry while current state is transitioning`.

andy31415: I believe the checks are also done in step 4, it just means we read it twice to get there. Will add a checkmark, up to you if this really needs update or not.

andy31415 (approved):

src/python_testing/TC_VALCC_3_1.py
cecille: I ended up having to refactor this thing a bunch more, and this code is now gone. But the code is quite different from when you last looked. Would you mind taking another peek when you get a chance?

fessehaeve:
This seems out of sync with the test plan. Is that being updated as well? 
Also, what changes are happening to the cluster? Can you please point me to where the changes are happening?

cecille:
@fessehaeve - yes, the test plan will need to be updated, I can update it from this script once this is approved. The changes to the cluster are happening in a separate branch. I should have the changes up in a PR by the SDK tomorrow, but they're orthogonal to this change - just an explanation for why I'm only doing a partial clean of the current code.

src/python_testing/TC_VALCC_3_1.py
```diff
@@ -44,12 +42,16 @@ def desc_TC_VALCC_3_1(self) -> str:
...
         steps = [
             TestStep(1, "Commissioning, already done", is_commissioning=True),
-            TestStep(2, "Send Open command"),
-            TestStep(3, "Read TargetState attribute"),
-            TestStep(4, "Read CurrentState attribute"),
-            TestStep(5, "Send Close command"),
-            TestStep(6, "Read TargetState attribute"),
-            TestStep(7, "Read CurrentState attribute"),
+            TestStep(2, "Set up a subscription to all attributes on the DUT"),
+            TestStep(3, "Send a close command to the DUT and wait until the CurrentState is closed", "DUT returns SUCCESS"),
```
ReneJosefsen, line 46: This step is used to get the valve into a well-known state before doing the test?

ReneJosefsen (approved):

src/python_testing/TC_VALCC_3_1.py
cecille: yup

src/app/clusters/valve-configuration-and-control-server/valve-configuration-and-control-server.cpp
```diff
@@ -360,9 +360,9 @@ CHIP_ERROR SetValveLevel(EndpointId ep, DataModel::Nullable<Percent> level, Data
     if (!isDelegateNull(delegate))
     {
         DataModel::Nullable<Percent> cLevel = delegate->HandleOpenValve(level);
-        if (HasFeature(ep, ValveConfigurationAndControl::Feature::kLevel))
+        if (!cLevel.IsNull())
```
fessehaeve, line 363: It is not guaranteed that the delegate will do the right thing and returning null if it doesn't support the Level feature. So keeping the feature map check seems alright to me.

src/app/clusters/valve-configuration-and-control-server/valve-configuration-and-control-server.cpp
```diff
@@ -376,14 +376,28 @@ CHIP_ERROR UpdateCurrentLevel(EndpointId ep, Percent currentLevel)
...
     {
         VerifyOrReturnError(Status::Success == CurrentLevel::Set(ep, currentLevel), CHIP_IM_GLOBAL_STATUS(ConstraintError));
-        return CHIP_NO_ERROR;
     }
-    return CHIP_IM_GLOBAL_STATUS(UnsupportedAttribute);
+    DataModel::Nullable<Percent> targetLevel = DataModel::NullNullable;
+    TargetLevel::Get(ep, targetLevel);
+    if (!targetLevel.IsNull() && currentLevel == targetLevel.Value())
+    {
+        targetLevel = DataModel::NullNullable;
```
fessehaeve, line 384: Spec currently doesn't say TargetLevel should be nulled once opening or closing is complete. It might be important not to explicitly null it, since it allows to verify the target and current level are similar for troubleshooting purposes. It could also be used, in addition to the CurrentState, to verify state of change.

fessehaeve: This change also causes VALCC_3_3 to fail.

src/app/clusters/valve-configuration-and-control-server/valve-configuration-and-control-server.cpp
```diff
@@ -376,14 +376,28 @@ CHIP_ERROR UpdateCurrentLevel(EndpointId ep, Percent currentLevel)
...
 }
 
 CHIP_ERROR UpdateCurrentState(EndpointId ep, ValveConfigurationAndControl::ValveStateEnum currentState)
 {
     VerifyOrReturnError(Status::Success == CurrentState::Set(ep, currentState), CHIP_IM_GLOBAL_STATUS(ConstraintError));
+    DataModel::Nullable<ValveStateEnum> targetState = DataModel::NullNullable;
+    TargetState::Get(ep, targetState);
+    if (currentState == targetState.ValueOr(ValveStateEnum::kUnknownEnumValue))
+    {
+        targetState = DataModel::NullNullable;
```
fessehaeve, line 398: Same comment as above.

src/app/clusters/valve-configuration-and-control-server/valve-configuration-and-control-server.cpp
ReneJosefsen: The description of the TargetState attribute does IMO indicate that it is expected to be nulled when the transition is done, this might now be when Open and Close are "done" but instead when the transition is done.

src/app/clusters/valve-configuration-and-control-server/valve-configuration-and-control-server.cpp
cecille: Added back in the feature check - should be checking both.

fessehaeve (approved):

PR #35854, opened 2024-09-30
VALCC: Remove top-level PICS

cecille:
Self-select the valve tests off the device information rather than the PICS. This makes it easier to run the tests since you can just run them all unconditionally and they will run if required.


andy31415 (approved):

arkq (approved):

cecille:
Testing at SVE shows that the pics functions are still valuable. Going to leave this as-is for now.

PR #35861, opened 2024-10-01
Add ability to report endpoint device types in the DataModel::Provider interface

andy31415:
This functionality seems used by our IM (specifically DeviceTypeResolver for access control selections.

### Changes

- Added First/Next device type iteration functions
- Implemented for codegen including using a hint for iteration caching
- Implemented device type support in the mock ember library
- Added unit test


src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
```diff
@@ -292,6 +292,21 @@ std::optional<DataModel::CommandEntry> EnumeratorCommandFinder::FindCommandEntry
     return (*id == kInvalidCommandId) ? DataModel::CommandEntry::kInvalid : CommandEntryFrom(path, *id);
 }
 
+DataModel::DeviceTypeEntry DeviceTypeEntryFromEmber(const EmberAfDeviceType & other)
```
bzbarsky-apple, line 295: We should consider just getting rid of EmberAfDeviceType, if it's got nothing that DeviceTypeEntry does not have.

Followup for that, though.

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
```diff
@@ -726,6 +741,65 @@ ConcreteCommandPath CodegenDataModelProvider::NextGeneratedCommand(const Concret
...
+std::optional<DataModel::DeviceTypeEntry> CodegenDataModelProvider::FirstDeviceType(EndpointId endpoint)
+{
+    std::optional<unsigned> endpoint_index = TryFindEndpointIndex(endpoint);
+    if (!endpoint_index.has_value())
+    {
+        return std::nullopt;
+    }
+
+    CHIP_ERROR err                            = CHIP_NO_ERROR;
+    Span<const EmberAfDeviceType> deviceTypes = emberAfDeviceTypeListFromEndpointIndex(*endpoint_index, err);
```
bzbarsky-apple, line 753: So why did we need to add `emberAfDeviceTypeListFromEndpointIndex`?  We return nullopt both on not having an index and on having an empty list... and `emberAfDeviceTypeListFromEndpoint` just returns an empty list if it's not a valid endpoint, no?

andy31415: The reason was that the codegen data provider has a cache for finding the endpoint index, so it is more efficient in that case. I was trying to avoid having the iterator do a `findEndpoint` lookup on every Next call, as it is less efficient as-is already and did not want to move it to O(n^2) on number of endpoints.

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
andy31415: Added a TODO as a followup. I would rather do this once we have only codegen data model and then we can figure out how ember can depend on the provider interfaces. Layering-wise it feels odd, since I made datamodel provider on purpose not be aware of ember. Making ember aware of datamodel provider types is not violating circularity but it may be a bit coupled.

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
andy31415: Specifically I thought the call of `emberAfIndexFromEndpoint(endpoint)` was unnecessarely expensive to be called in a Next loop.

andy31415: I placed this in FirstDeviceType for consistency. Added a comment.

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
```diff
@@ -726,6 +744,73 @@ ConcreteCommandPath CodegenDataModelProvider::NextGeneratedCommand(const Concret
...
+    if ((mDeviceTypeIterationHint < deviceTypes.size()) && IsSameDeviceTypeEntry(previous, deviceTypes[mDeviceTypeIterationHint]))
+    {
+        idx = mDeviceTypeIterationHint + 1; // target the NEXT device
+    }
+    else
+    {
+        while (idx < deviceTypes.size())
+        {
+            idx++;
+            if (IsSameDeviceTypeEntry(previous, deviceTypes[idx - 1]))
```
tcarmelveilleux, line 798: What is this used for? Same for other call above. It's not clear why we check for "same device type" on a given endpoint. You are never allowed in the first place to say the same device type ID twice (same or different version) on a given endpoint. Not sure why these guards are there, rather than at init.

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
```diff
@@ -726,6 +744,73 @@ ConcreteCommandPath CodegenDataModelProvider::NextGeneratedCommand(const Concret
...
+    if (!endpoint_index.has_value())
+    {
+        return std::nullopt;
+    }
+
+    CHIP_ERROR err                            = CHIP_NO_ERROR;
+    Span<const EmberAfDeviceType> deviceTypes = emberAfDeviceTypeListFromEndpointIndex(*endpoint_index, err);
+
+    unsigned idx = 0;
+    if ((mDeviceTypeIterationHint < deviceTypes.size()) && IsSameDeviceTypeEntry(previous, deviceTypes[mDeviceTypeIterationHint]))
```
tehampson, line 789: General question.

It is possible for there to be multiple of the same device type entry in `deviceTypes`. I don't think so, but just want to make sure that this assumption is true.

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
andy31415: It is an `operator==`. I can also make that explicit, but figured that since types are different I named it. Would `if (previous == deviceTypes[idx-1])` be more readable?

tehampson (approved):

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
andy31415: They should be unique, repeating items seems not practical.

I think as a sideffect using hints makes it progress instead of looping even if they repeat (because we always move hint+1) however that is not something we intentionally want to support.

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
```diff
@@ -726,6 +744,73 @@ ConcreteCommandPath CodegenDataModelProvider::NextGeneratedCommand(const Concret
...
+    std::optional<unsigned> endpoint_index = TryFindEndpointIndex(endpoint);
+    if (!endpoint_index.has_value())
+    {
+        return std::nullopt;
+    }
+
+    CHIP_ERROR err                            = CHIP_NO_ERROR;
+    Span<const EmberAfDeviceType> deviceTypes = emberAfDeviceTypeListFromEndpointIndex(*endpoint_index, err);
+
+    unsigned idx = 0;
```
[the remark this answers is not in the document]
andy31415: Code refactored for readability.

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
andy31415: Updated code for readability and also added operator==.

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
tehampson: Yeah lots of things start breaking if that assumption isn't true which is why I was just double checking

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
```diff
@@ -292,6 +292,56 @@ std::optional<DataModel::CommandEntry> EnumeratorCommandFinder::FindCommandEntry
     return (*id == kInvalidCommandId) ? DataModel::CommandEntry::kInvalid : CommandEntryFrom(path, *id);
 }
 
+// TODO: DeviceTypeEntry content is IDENTICAL to EmberAfDeviceType, so centralizing
```
yunhanw-google, line 295: File one ticket for this?

yunhanw-google (approved):

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
andy31415: Done: #35889

arkq (approved):

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
```diff
@@ -292,6 +292,57 @@ std::optional<DataModel::CommandEntry> EnumeratorCommandFinder::FindCommandEntry
...
+{
+    return (a.deviceTypeId == b.deviceId) && (a.deviceTypeVersion == b.deviceVersion);
+}
+
+/// Find the `index` where one of the following holds:
+///    - types[index - 1] == previous OR
+///    - index == types.size()  // i.e. not found or there is no next
+///
+/// hintWherePreviousMayBe represents a search hint where previous may exist.
+unsigned FindNextDeviceTypeIndex(Span<const EmberAfDeviceType> types, const DataModel::DeviceTypeEntry previous,
```
bzbarsky-apple, line 321: Why is this passing DeviceTypeEntry by value?

andy31415: Typo, will add a reference on monday

PR #35863, opened 2024-10-01
Make reporting::Engine able to use DataModel::Provider for cluster version checks

andy31415:
Previous logic was hard-coding an ember-compatibility-functions check for comparing versions. Updated code is able to use DataModel::Provider which compares the global cluster version.

### Changes

- added a separate `IsClusterDataVersionEqualTo` implementation in interaction model engine which is aware of data model provider, including a "check for 1:1 equivalence to ember"
- Added support to set mock model global version to a specific value
- Fixed up the TestReadInteraction unit test, which was relying on a specific version check on reads.


src/app/reporting/Read-Ember.cpp
```diff
@@ -51,6 +51,11 @@ DataModel::ActionReturnStatus RetrieveClusterData(DataModel::Provider * dataMode
     return CHIP_NO_ERROR;
 }
 
+bool IsClusterDataVersionEqualTo(DataModel::Provider * dataModel, const ConcreteClusterPath & path, DataVersion dataVersion)
+{
+    return chip::app::IsClusterDataVersionEqual(path, dataVersion);
```
bzbarsky-apple, line 56:
```suggestion
    return IsClusterDataVersionEqual(path, dataVersion);
```

bzbarsky-apple (approved):

tcarmelveilleux (approved):

tehampson (approved):

PR #35865, opened 2024-10-01
Add Attribute and Command Conformance to ZAP Template XMLs

ethanzhouyc:
- Added conformance details for attributes and commands in ZAP template XMLs from the new XML version. The added XML data is necessary for the functionality of the [Feature Page project](https://github.com/project-chip/zap/issues/1380).

- This commit only updates XMLs with initials from A to C to ease the review process. Remaining XML changes will be addressed in subsequent PRs.

- Removed the Feature bitmap in content-launch-cluster.xml and replaced it with features represented by the `<Feature>` tag. The value in original Feature bitmap is wrong: the feature masks should be `0x01, 0x02, 0x04, 0x08, 0x10`, but in the XML it was `0x01, 0x02, 0x03, 0x04, 0x05`. Consequently, after the update, the CI job would fail but the change is actually correct. Should consider correcting this bitmap checking in CI. [See job failure details](https://github.com/project-chip/connectedhomeip/actions/runs/11133255934/job/30938932982?pr=35865#step:10:16)

- Corrected the OnOff feature code in matter-devices.xml

### DIFF
--- a/src/app/zap-templates/zcl/data-model/chip/access-control-cluster.xml
+++ b/src/app/zap-templates/zcl/data-model/chip/access-control-cluster.xml
@@ -113,28 +113,57 @@ limitations under the License.
       <description>ACL</description>
       <access op="read" privilege="administer"/>
       <access op="write" privilege="administer"/>
+      <mandatoryConform/>
     </attribute>
 
     <attribute side="server" code="0x0001" define="EXTENSION" type="array" entryType="AccessControlExtensionStruct" writable="true" optional="true">
       <description>Extension</description>
       <access op="read" privilege="administer"/>
       <access op="write" privilege="administer"/>
+      <mandatoryConform>
+        <feature name="EXTS"/>
+      </mandatoryConform>
     </attribute>
 
-    <attribute side="server" code="0x0002" define="SUBJECTS_PER_ACCESS_CONTROL_ENTRY" type="int16u" min="4" default="4">SubjectsPerAccessControlEntry</attribute>
-    <attribute side="server" code="0x0003" define="TARGETS_PER_ACCESS_CONTROL_ENTRY" type="int16u" min="3" default="3">TargetsPerAccessControlEntry</attribute>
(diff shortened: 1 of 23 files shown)


andy31415:
@ethanzhouyc could you describe the following:

- were these manually added or by a robot? I would trust it more if these were script/automated added. If these require manual adding, maybe we can work with @hasty to add it to Alchemy instead
- is this supported by zap? Any links to documentation on how this will be used? This will help people looking at logs in the future

For the ZAP failure, you need to zap-regen if you change things that codegen depends on.

ethanzhouyc:
- These were manually added, but based on automated generated XMLs from [data_model/master](https://github.com/project-chip/connectedhomeip/tree/master/data_model/master) folder. Alchemy generates XML in the format used in that folder, but does not support generating the ZAP template XML format in [the folder](https://github.com/ethanzhouyc/connectedhomeip/tree/master/src/app/zap-templates/zcl/data-model/chip) I updated.

- ZAP support is not available yet, but I am working on the Feature Page project and have a pending PR that will enable this. I will follow up with documentation once it’s ready.

- To fix the ZAP failure, I will create a separate PR to correct the zap-regen

hasty:
> * These were manually added, but based on automated generated XMLs from [data_model/master](https://github.com/project-chip/connectedhomeip/tree/master/data_model/master) folder.

This is a one-line change to alchemy, if we're ok with adding this to the ZAP XML; would be a lot easier than manually copying.

hasty:
And presumably we'd want to do it for events as well?

ethanzhouyc:
@hasty
If we could automate the process, it would be great and let's do it!
Can you help adapt Alchemy to my automatic generation needs? Once it's ready, can you give instructions on how to use the new feature? Thank you!
We only focus on attribute and command conformance, and there is no need to change events.

ethanzhouyc:
> And presumably we'd want to do it for events as well?

Yes we should add conformance for events too

PR #35866, opened 2024-10-01
[DRAFT] ICD - Design document for Check-in handler

thivya-amazon:
Added a docs folder under src/app/icd to add implementation related documents and diagrams.


src/app/icd/docs/check-in_handler_design.md
```diff
@@ -0,0 +1,18 @@
+# Check-in handler overview
+
+ICD servers send check-in messages to a registered ICD client when a subscription is lost or one was never established. A subscription can be lost for several reasons, such as:
```
bzbarsky-apple, line 3:
```suggestion
ICD servers send check-in messages to a registered ICD client when a subscription is lost or one was never established. A subscription can be lost for a variety of reasons.  Some non-exhaustive examples are:
```

src/app/icd/docs/check-in_handler_design.md
```diff
@@ -0,0 +1,18 @@
...
+
+• When the ICD is powered off to change the battery.
+
+• Power or network outage causing the connection between the client and the ICD to be interrupted.
+
+• The client is unavailable for any reason
+
+The Check-In handler in an ICD client is responsible for handling unsolicited check-in messages from an ICD server. All ICD clients should add support for Check-In protocol as it is a fail safe mechanism that provides a means to re-establish a secure session with an ICD server. The heart of the module is the check-in handler that receives and parses check-in messages from ICD servers. After ensuring the validity of the message, the check-in handler passes it on to the check-in protocol to parse the message and retrieves the ICD counter value. The ICD counter is used to derive the nonce in the encryption process of the check-in protocol. Beyond that, ICD counters are also used to detect repeated reception of the same message and eliminate duplicates. Once the maximum value for an ICD counter is reached, the counter should roll-over to 0 and when that is done, it is necessary for the ICD client to re-register with the ICD server using a fresh key.
```
bzbarsky-apple, line 13:
```suggestion
The Check-In handler in an ICD client is responsible for handling unsolicited check-in messages from an ICD server. All ICD clients should add support for Check-In protocol as it is a fail safe mechanism that provides a means to re-establish a secure session with an ICD server. The heart of the module is the check-in handler that receives and parses check-in messages from ICD servers. After ensuring the validity of the message, the check-in handler passes it on to the check-in protocol to parse the message and retrieve the ICD counter value. The ICD counter is used to derive the nonce in the encryption process of the check-in protocol. Beyond that, ICD counters are also used to detect repeated reception of the same message and eliminate duplicates. Once the maximum value for an ICD counter is reached, the counter should roll-over to 0 and when that is done, it is necessary for the ICD client to re-register with the ICD server using a fresh key. 
```

bzbarsky-apple, line 13: I do wonder whether it's worth pointing to the relevant spec section here instead of trying to (poorly) duplicate it.

src/app/icd/docs/images/check-in_handler_design.png
bzbarsky-apple, line 1: What's the mechanism for ensuring this diagram is updated when the code changes?

The diagram is clearly generated by some tool; can the actual data file for the tool be checked in, so that the diagram is easy to fix if things change?

src/app/icd/docs/check-in_handler_design.md
thivya-amazon: It was just an attempt to provide a gist of what spreads across multiple paragraphs in the spec to someone who is looking at the design without any context. Didn't intend to make a poor duplicate. FYI, the PR is still a draft.

bzbarsky-apple: @thivya-amazon I appreciate that this is still a draft.   It's hard to figure out, for me, what the right tradeoff here between talking about details and just giving an overview.

What I would suggest is most useful here is to give a high-level overview of what this thing is trying to accomplish, point to the relevant parts of the specificaion for details, and discuss which parts of the spec and which invariant enforcement are handled by the SDK code vs which parts are expected to be handed by the application.

Talking about details here risks getting them wrong in ways that will confuse people.  As a concrete example, it is _not_ in fact necessary for the ICD client to register when roll-over to 0 happens.  It's necessary to do when 2^31 values have passed by, which could be way before rollover, or way after, depending on where one starts initially....

PR #35867, opened 2024-10-01
[Fabric-Admin] Refactor to use API methods instead of PushCommand (2/3)

### LINKED ISSUE #34910: [SVE] Switch from string-based cli API to the SDK API using ReadClients.

yufengwangca:

This is the second iteration of the migration process for pairing commands.

Fixes https://github.com/project-chip/connectedhomeip/issues/34910


examples/fabric-admin/device_manager/PairingManager.cpp
```diff
@@ -168,3 +247,429 @@ void PairingManager::OnOpenCommissioningWindowVerifierResponse(void * context, N
...
+        break;
+    }
+}
+
+void PairingManager::OnPairingComplete(CHIP_ERROR err)
+{
+    if (err == CHIP_NO_ERROR)
+    {
+        ChipLogProgress(NotSpecified, "Pairing Success");
+        ChipLogProgress(NotSpecified, "PASE establishment successful");
```
tehampson, line 270: Why not compile to single log line?

examples/fabric-admin/device_manager/PairingManager.cpp
```diff
@@ -168,3 +247,429 @@ void PairingManager::OnOpenCommissioningWindowVerifierResponse(void * context, N
...
+    }
+}
+
+void PairingManager::OnCommissioningComplete(NodeId nodeId, CHIP_ERROR err)
+{
+    if (err == CHIP_NO_ERROR)
+    {
+        // print to console
+        fprintf(stderr, "New device with Node ID: 0x%lx has been successfully added.\n", nodeId);
+        // CurrentCommissioner() has a lifetime that is the entire life of the application itself
```
tehampson, line 300: nit, Can you either update the comment to say `mCommissioner` or provide `&CurrentCommissioner()` to `StartDeviceSynchronization`

examples/fabric-admin/device_manager/PairingManager.cpp
```diff
@@ -168,3 +247,429 @@ void PairingManager::OnOpenCommissioningWindowVerifierResponse(void * context, N
...
+        ChipLogProgress(NotSpecified, "Pairing Deleted Failure: %s", ErrorStr(err));
+    }
+}
+
+void PairingManager::OnCommissioningComplete(NodeId nodeId, CHIP_ERROR err)
+{
+    if (err == CHIP_NO_ERROR)
+    {
+        // print to console
+        fprintf(stderr, "New device with Node ID: 0x%lx has been successfully added.\n", nodeId);
```
tehampson, line 299: nit, Please use `ChipLogFormatX64` and `ChipLogValueX64(nodeId)`

examples/fabric-admin/device_manager/PairingManager.cpp
```diff
@@ -168,3 +247,429 @@ void PairingManager::OnOpenCommissioningWindowVerifierResponse(void * context, N
...
+                ChipLogError(NotSpecified, "Failed to delete ICD entry: %s", ErrorStr(err));
+            }
+        }
+        ChipLogProgress(NotSpecified, "Device commissioning Failure: %s", ErrorStr(err));
+    }
+
+    if (mCommissioningDelegate)
+    {
+        mCommissioningDelegate->OnCommissioningComplete(nodeId, err);
+        this->SetCommissioningDelegate(nullptr);
```
tehampson, line 322: nit
```suggestion
        SetCommissioningDelegate(nullptr);
```

examples/fabric-admin/device_manager/PairingManager.cpp
```diff
@@ -168,3 +247,429 @@ void PairingManager::OnOpenCommissioningWindowVerifierResponse(void * context, N
...
+    }
+}
+
+void PairingManager::OnReadCommissioningInfo(const Controller::ReadCommissioningInfo & info)
+{
+    ChipLogProgress(AppServer, "OnReadCommissioningInfo - vendorId=0x%04X productId=0x%04X", info.basic.vendorId,
+                    info.basic.productId);
+
+    // The string in CharSpan received from the device is not null-terminated, we use std::string here for coping and
+    // appending a numm-terminator at the end of the string.
```
tehampson, line 332: nit, 
```suggestion
    // appending a null-terminator at the end of the string.
```

examples/fabric-admin/device_manager/PairingManager.cpp
```diff
@@ -168,3 +247,429 @@ void PairingManager::OnOpenCommissioningWindowVerifierResponse(void * context, N
...
+    {
+        return;
+    }
+
+    mCommissioner->RegisterDeviceDiscoveryDelegate(nullptr);
+
+    auto interfaceId = resolutionData.ipAddress[0].IsIPv6LinkLocal() ? resolutionData.interfaceId : Inet::InterfaceId::Null();
+    auto peerAddress = Transport::PeerAddress::UDP(resolutionData.ipAddress[0], port, interfaceId);
+    err              = Pair(mNodeId, peerAddress);
+    if (CHIP_NO_ERROR != err)
```
tehampson, line 422: What is this ` if (CHIP_NO_ERROR != err)` for?

examples/fabric-admin/device_manager/PairingManager.cpp
```diff
@@ -168,3 +247,429 @@ void PairingManager::OnOpenCommissioningWindowVerifierResponse(void * context, N
...
+    // No manual input, so do not need to extend.
+    return Optional<uint16_t>();
+}
+
+bool PairingManager::ShouldWaitAfterDeviceAttestation()
+{
+    // If there is a vendor ID and product ID, request OnDeviceAttestationCompleted().
+    // Currently this is added in the case that the example is performing reverse commissioning,
+    // but it would be an improvement to store that explicitly.
+    // TODO: Issue #35297 - [Fabric Sync] Improve where we get VID and PID when validating CCTRL CommissionNode command
```
tehampson, line 438: Is the reverse commissioning path going though this code path? (or will it in a future PR?)

examples/fabric-admin/device_manager/PairingManager.cpp
```diff
@@ -168,3 +247,429 @@ void PairingManager::OnOpenCommissioningWindowVerifierResponse(void * context, N
...
+                    ? Credentials::AttestationVerificationResult::kDacProductIdMismatch
+                    : Credentials::AttestationVerificationResult::kDacVendorIdMismatch);
+            return;
+        }
+
+        // NOTE: This will log errors even if the attestion was successful.
+        auto err = deviceCommissioner->ContinueCommissioningAfterDeviceAttestation(device, attestationResult);
+        if (CHIP_NO_ERROR != err)
+        {
+        }
```
tehampson, line 482: What is this `if (CHIP_NO_ERROR != err)` block for?

examples/fabric-admin/device_manager/PairingManager.cpp
```diff
@@ -168,3 +247,429 @@ void PairingManager::OnOpenCommissioningWindowVerifierResponse(void * context, N
...
+        return;
+    }
+
+    // Don't bypass attestation, continue with error.
+    auto err = deviceCommissioner->ContinueCommissioningAfterDeviceAttestation(device, attestationResult);
+    if (CHIP_NO_ERROR != err)
+    {
+    }
```
tehampson, line 490: Same, what is this `if (CHIP_NO_ERROR != err)` for?

examples/fabric-admin/device_manager/PairingManager.cpp
```diff
@@ -168,3 +247,429 @@ void PairingManager::OnOpenCommissioningWindowVerifierResponse(void * context, N
...
+
+void PairingManager::OnCommissioningComplete(NodeId nodeId, CHIP_ERROR err)
+{
+    if (err == CHIP_NO_ERROR)
+    {
+        // print to console
+        fprintf(stderr, "New device with Node ID: 0x%lx has been successfully added.\n", nodeId);
+        // CurrentCommissioner() has a lifetime that is the entire life of the application itself
+        // so it is safe to provide to StartDeviceSynchronization.
+        DeviceSynchronizer::Instance().StartDeviceSynchronization(mCommissioner, nodeId, mDeviceIsICD);
```
tehampson, line 302: Will this be called when you pair a remote bridge and local bridge? If so do you actually want to call this `StartDeviceSynchronization`?

examples/fabric-admin/device_manager/PairingManager.cpp
```diff
@@ -168,3 +247,429 @@ void PairingManager::OnOpenCommissioningWindowVerifierResponse(void * context, N
...
+    if (err != CHIP_NO_ERROR)
+    {
+        ChipLogError(NotSpecified, "Failed to pair device, error: %s", ErrorStr(err));
+    }
+}
+
+CHIP_ERROR PairingManager::UnpairDevice(NodeId nodeId)
+{
+    // Schedule work on the Matter thread
+    return DeviceLayer::PlatformMgr().ScheduleWork(OnUnpairDevice, static_cast<intptr_t>(nodeId));
```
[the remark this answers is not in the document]
yufengwangca: we can not make this assumption, using intptr_t to pass a 64-bit value like NodeId can potentially lead to loss of information,

examples/fabric-admin/device_manager/PairingManager.cpp
yufengwangca: Yes, this part is exactly same as what we added in https://github.com/project-chip/connectedhomeip/pull/35187, just porting from PairingCommand to PairingManager

examples/fabric-admin/device_manager/PairingManager.cpp
yufengwangca: This looks like a regression when we did StartDeviceSynchronization, we used to have check not to sync bridge devices.

[StartDeviceSynchronization](https://github.com/project-chip/connectedhomeip/issues/35896)

examples/fabric-admin/device_manager/PairingManager.cpp
```diff
@@ -168,3 +247,444 @@ void PairingManager::OnOpenCommissioningWindowVerifierResponse(void * context, N
...
+    }
+
+    auto params = Platform::MakeUnique<PairDeviceWithCodeParams>();
+    VerifyOrReturnError(params != nullptr, CHIP_ERROR_NO_MEMORY);
+
+    params->nodeId = nodeId;
+    Platform::CopyString(params->payloadBuffer, sizeof(params->payloadBuffer), payload);
+
+    // Schedule work on the Matter thread
+    return DeviceLayer::PlatformMgr().ScheduleWork(OnPairDeviceWithCode, reinterpret_cast<intptr_t>(params.release()));
```
arkq, line 590: How about using `ScheduleLambda()` instead of passing params via pointer and the reinterpret_casting it again (and the same thing in other places where you are doing `ScheduleWork()`?

arkq (approved):

examples/fabric-admin/device_manager/PairingManager.cpp
yufengwangca: That’s a great idea! I also noticed this API. Currently, we are using PlatformMgr().ScheduleWork throughout FabricAdmin, but I believe we should standardize on a single scheduling method. I’ve registered the following issue to update everything to use ScheduleLambda() uniformly.

https://github.com/project-chip/connectedhomeip/issues/35900

examples/fabric-admin/device_manager/PairingManager.cpp
tehampson: But with this PR we now have two spots where we have this TODO comment that we need to address. You have not removed the part from `PairingCommand`, you have only added it here to `PairingManager`.

My question is:
* Is the reverse commissioning path going though this code path?
  * If yes - > Why haven't we removed that code from `PairingCommand`
  * If no -> Why do we have this code here if this is not code that process reverse commissioning path

yufengwangca: Yes, I should remove chris's change after I move it from PairingCommand to PairingManager

tehampson (approved):
Approval is assuming you will take care of removing the VID PID check from PairingCommand as we discussed offline and there is currently an unresolved comment to prevent automerge

PR #35873, opened 2024-10-02
Remove one more direct ember coupling in reporting engine (using of CheckEventSupportStatus from ember-compatibility)

andy31415:
Make Reporting engine use datamodel::provider to determine if a event path is valid or not. I named it 'path valid' since event support status should go away in large part once we fully remove the event list attribute support from our source code.

> [!NOTE]
> #35874 is a PR to remove eventlist support completely from the server side
> (client-side we still code-generate and we may need to iterate on that). As a result, this
> PR does not handle event list attribute iteration in the implementation.


tehampson (approved):

tcarmelveilleux (approved):

src/app/reporting/Engine.cpp
```diff
@@ -23,25 +23,60 @@
...
+using Protocols::InteractionModel::Status;
+
+Status EventPathValid(DataModel::Provider * model, const ConcreteEventPath & eventPath)
+{
+#if CHIP_CONFIG_USE_DATA_MODEL_INTERFACE
+
+    if (!model->GetClusterInfo(eventPath).has_value())
+    {
+        // cluster is invalid, return an error.
+        for (EndpointId id = model->FirstEndpoint(); id != kInvalidEndpointId; id = model->NextEndpoint(id))
```
tcarmelveilleux, line 61: Consider using a method like `model->HasEndpoint(eventPath.mEndpointId)`

andy31415: Added.

arkq (approved):

PR #35874, opened 2024-10-02
Remove `event list attribute` support in the SDK

andy31415:
Specification marks it as deprecated (conformance `D`) and we have not actually supported it in previous releases.

For now remove all defines from implementing it. Client clusters still claim supported, so that may need further iteration.

### DIFF
--- a/src/app/util/ember-compatibility-functions.cpp
+++ b/src/app/util/ember-compatibility-functions.cpp
@@ -806,20 +806,8 @@ Protocols::InteractionModel::Status CheckEventSupportStatus(const ConcreteEventP
         return Status::UnsupportedCluster;
     }
 
-#if CHIP_CONFIG_ENABLE_EVENTLIST_ATTRIBUTE
-    for (size_t i = 0; i < cluster->eventCount; ++i)
-    {
-        if (cluster->eventList[i] == aPath.mEventId)
-        {
-            return Status::Success;
-        }
-    }
-
-    return Status::UnsupportedEvent;
-#else
     // No way to tell. Just claim supported.
     return Status::Success;
-#endif // CHIP_CONFIG_ENABLE_EVENTLIST_ATTRIBUTE
 }
 
 } // namespace app
--- a/config/esp32/components/chip/CMakeLists.txt
+++ b/config/esp32/components/chip/CMakeLists.txt
@@ -321,10 +321,6 @@ if (CONFIG_ENABLE_ESP_INSIGHTS_TRACE)
    target_include_directories(${COMPONENT_LIB} INTERFACE "${CHIP_ROOT}/src/tracing/esp32_trace/include")
 endif()
 
-if (CONFIG_ENABLE_MATTER_EVENT_LIST)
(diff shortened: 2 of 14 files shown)


tcarmelveilleux (approved):

tehampson (approved):

arkq (approved):

src/app/util/ember-compatibility-functions.cpp
```diff
@@ -806,20 +806,8 @@ Protocols::InteractionModel::Status CheckEventSupportStatus(const ConcreteEventP
         return Status::UnsupportedCluster;
     }
 
-#if CHIP_CONFIG_ENABLE_EVENTLIST_ATTRIBUTE
```
bzbarsky-apple, line 809: So the spec still has a non-D requirement for this check to work (which we have been violating, clearly)....  Do we need a spec issue raised here to remove that?  @tcarmelveilleux

andy31415: Could you add a link (or section number) for this here?

Code-wise I believe we did not have eventlist enabled in most cases anyway, so the change did not change behaviour (except some NXP examples). In this case probably spec should be updated to not use something marked as deprecated.

bzbarsky-apple: The relevant spec section is "8.4.3.2. Incoming Read Request and Subscribe Request Action Processing" and it says:

> * Each request path in the EventRequests field SHALL be processed as follows:
>    * If the path is a concrete path:
>...
>       * Else if the path indicates a cluster event that is not supported, an EventStatusIB SHALL be generated with the UNSUPPORTED_EVENT Status Code.

This has nothing to do with EventList _in the spec_, so the spec is not "using something marked as deprecated".  The spec just expects the server to know what events are "supported".

In terms of the SDK, we implemented that spec-required support check on top of EventList support, because the same data is needed for both.  And yes, with EventList not enabled we had a divergence from the spec but the theoretical plan for fixing that was to enable EventList sometime.

So as I see it, there are basically three options here:

1. Keep just not matching the spec here, as we have all along.
2. Figure out how to implement this spec-required check.
3. Change the spec to not do this check.

tcarmelveilleux: I think 2 can be done. We can add AAI methods to list supported events, in addition to listing supported attributes. That would allow us to determine if an event is supported. I would argue that claiming it's supported when it will never be emitted is not a big deal for any client...

andy31415: Is there actual value for "2" ? Having event list metadata requires more code and metadata and the gain seems to be "disallow subscriptions for events that can never happen". But at the same time since events are emitted by the app, you do not even have assurance that events will ever be emitted. It seems we will default to apps just "select all possible events just in case".

andy31415: I would very much prefer that after 2 years we match spec to implementation here as this was promised at some point to also happen. The alternative seems to add overhead to people to define event metadata and require extra flash (even if minimal) to maintain logic for it.

PR #35877, opened 2024-10-02
[macOS standalone apps] Add missing autoreleasepool for logging when …

vivien-apple:
…standalone apps are runned on macOS

#### Problem

When standalone apps (`chip-tool`, `all-clusters-app`, ..) are runned on macOS, there is no top level `autoreleasepool` objects can be pushed to. But the logging code uses `NSString` objects. This PR add an `autoreleasepool` for those.

### DIFF
--- a/src/platform/Darwin/Logging.mm
+++ b/src/platform/Darwin/Logging.mm
@@ -51,15 +51,36 @@ namespace Logging {
         void LogByteSpan(
             chip::Logging::LogModule moduleId, char const * moduleName, os_log_type_t type, const chip::ByteSpan & span)
         {
-            os_log_t logger = LoggerForModule(moduleId, moduleName);
-            if (os_log_type_enabled(logger, type)) {
-                auto size = span.size();
-                auto data = span.data();
-                NSMutableString * string = [[NSMutableString alloc] initWithCapacity:(size * 6)]; // 6 characters per byte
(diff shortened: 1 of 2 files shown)


src/platform/Darwin/Logging.mm
```diff
@@ -51,15 +51,36 @@ os_log_t LoggerForModule(chip::Logging::LogModule moduleID, char const * moduleN
...
+                if (os_log_type_enabled(logger, type)) {
+                    va_list v;
+                    va_start(v, msg);
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wformat-nonliteral"
+                    NSString * string = [[NSString alloc] initWithFormat:@(msg) arguments:v];
+#pragma clang diagnostic pop
+                    va_end(v);
+
+                    os_log_with_type(logger, type, "%@", string);
```
bzbarsky-apple, line 82: This will really bloat the log storage, because it will have to store the whole string being logged each time, instead of storing the format string once and then the arguments per call...

Why do we want to do this?   In the original version of this, is the issue that `os_log_with_type` internally creates things we want an autorelease pool around?

ksperling-apple (changes requested):

vivien-apple:
Replaced by #35884

PR #35880, opened 2024-10-02
Fix duplicate CI runs for both push and pull_request events

vatsalghelani-csa:
Disable CI runs for push event on any branch except `master` and `v*-branch`. This means if we want CI runs on branches not from a fork of connectedhomeip, we will need to make pull requests.


.github/workflows/build.yaml
```diff
@@ -16,6 +16,9 @@ name: Builds
 
 on:
     push:
+        branches:
+            - master
+            - 'v*-branch'
         branches-ignore:
```
andy31415, line 22: do we need an ignore at all given that we have `branches` element?

vatsalghelani-csa: Removed the `branches-ignore` element from the push event.

andy31415 (approved):

arkq (approved):

andy31415:
@vatsalghelani-csa the previous run said "EFR32" did not run. I updated to master to kick CI to see if this is a real issue or not...

.github/workflows/examples-efr32.yaml
```diff
@@ -15,9 +15,10 @@
 name: Build example - EFR32
 
 on:
-  push:
-    branches-ignore:
-      - 'dependabot/**'
+    push:
```
arkq, line 18: Maybe the problem is with inconsistent indentation? The rest of this file is indented with 2 spaces, not 4.

vatsalghelani-csa: @arkq I would agree with you on this. I noticed that too, but given the file’s long-standing state, I hesitated to make the indentation changes with this commit.

arkq: Yes, but YAML should be indented by 2 spaces (at least that's what most YAML linters do)

.github/workflows/examples-efr32.yaml
```diff
@@ -15,10 +15,9 @@
 name: Build example - EFR32
 
 on:
-    push:
-        branches:
-            - master
-            - 'v*-branch'
+  push:
+    branches-ignore:
```
arkq, line 19: I think that you can make changes here as long as you will not change the indentation (keep it as it is right now - 2 spaces).

PR #35884, opened 2024-10-02
Darwin: Avoid autorelease handling around LoggerForModule

ksperling-apple:
We don't want the LoggerForModule() implementation to call retain/autorelease on the logger, because its unnecessary, and because LoggerForModule() can be called from a C++ context without an auto-release pool in place.

To avoid this, we can have LoggerForModule() return a raw pointer, and handle the bridge cast to os_log_t on the caller side. This also avoids unnecessary code on the caller side that would otherwise attempt to rescue the presumed-autoreleased logger object from the ARP.


vivien-apple (approved):

ksperling-apple:
Fast-tracking Darwin-only fix.

PR #35892, opened 2024-10-03
Chime server

andyg-apple:
Server Code for Chime cluster and associated file changes


src/app/clusters/chime-server/chime-server.cpp
```diff
@@ -0,0 +1,185 @@
+
+/*
+ *
```
bzbarsky-apple, line 3:
```suggestion
/*
```

src/app/clusters/chime-server/chime-server.cpp
```diff
@@ -0,0 +1,185 @@
...
+    {
+    case InstalledChimeSounds::Id:
+        return aEncoder.EncodeList([this](const auto & encoder) -> CHIP_ERROR {
+            uint8_t index  = 0;
+            CHIP_ERROR err = CHIP_NO_ERROR;
+
+            // runs through all the Chime sounds available one by one
+            do
+            {
+                Chime::Structs::ChimeSoundStruct::Type chimeSound;
```
bzbarsky-apple, line 81:
```suggestion
                Structs::ChimeSoundStruct::Type chimeSound;
```

src/app/clusters/chime-server/chime-server.cpp
```diff
@@ -0,0 +1,185 @@
...
+                if ((err = this->mDelegate.GetChimeSoundByIndex(index, chimeSound)) == CHIP_NO_ERROR)
```
bzbarsky-apple, line 82: So what's the ownership of the "Name" field of that struct?  This code assumes that it's pointing into some callee-allocated buffer that the callee can then free ... when?

It would be better to have a struct inheriting from `ChimeSoundStruct::Type` that has storage for the name so the callee can just copy into that.  See what the mode base bits do for an example, or some of the thermostat bits.

src/app/clusters/chime-server/chime-server.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+            if (err == CHIP_ERROR_NOT_FOUND)
+            {
+                return CHIP_NO_ERROR;
+            }
+            return err;
+        });
+
+    case ActiveChimeID::Id:
+        return aEncoder.Encode(mDelegate.GetActiveChimeId());
```
bzbarsky-apple, line 97: So if something that's not en explicit write changes the active chime id or the enabled state, the delegate is responsible for dealing with the dirty marking?  See also https://github.com/project-chip/connectedhomeip/pull/35729#discussion_r1772149438

src/app/clusters/chime-server/chime-server.cpp
```diff
@@ -0,0 +1,185 @@
...
+        return CHIP_NO_ERROR;
+    }
+
+    default:
+        // Unknown attribute
+        return CHIP_IM_GLOBAL_STATUS(UnsupportedAttribute);
+    }
+}
+
+CHIP_ERROR ChimeServer::SetActiveChimeId(uint8_t soundId)
```
bzbarsky-apple, line 131: So this is a private method... who calls it?

src/app/clusters/chime-server/chime-server.cpp
```diff
@@ -0,0 +1,185 @@
...
+    bool activeIdChanged   = !(currentSoundId == soundId);
+
+    VerifyOrReturnError(activeIdChanged, CHIP_NO_ERROR);
+    VerifyOrDie(mDelegate.SetActiveChimeId(soundId) == CHIP_NO_ERROR);
+    MatterReportingAttributeChangeCallback(GetEndpointId(), Chime::Id, ActiveChimeID::Id);
+
+    return CHIP_NO_ERROR;
+}
+
+CHIP_ERROR ChimeServer::SetEnabled(bool enabled)
```
bzbarsky-apple, line 143: Again, this is a private method, so who is able to call it and why?

src/app/clusters/chime-server/chime-server.h
```diff
@@ -0,0 +1,112 @@
...
+namespace chip {
+namespace app {
+namespace Clusters {
+
+class ChimeDelegate;
+
+class ChimeServer : private AttributeAccessInterface, private CommandHandlerInterface
+{
+public:
+    ChimeServer(EndpointId endpointId, ChimeDelegate & delegate);
```
bzbarsky-apple, line 41: So this assumes the delegate exists before this is constructed, right?

That means this will be destroyed before the delegate, and our destructor should presumably null out the pointer to us on the delegate...

src/app/clusters/chime-server/chime-server.h
```diff
@@ -0,0 +1,112 @@
...
+    /**
+     *   Get the chime sounds.
+     *   @param index The index of the chime sound to be returned. It is assumed that chime sounds are indexable from 0 and with no
+     * gaps.
+     *   @param chimeSound A reference to receive the chime sound struct on success.
+     *   @return Returns a CHIP_NO_ERROR if there was no error and the chime sound was returned successfully,
+     *   CHIP_ERROR_NOT_FOUND if the index in beyond the list of available chime sounds.
+     */
+    virtual CHIP_ERROR GetChimeSoundByIndex(uint8_t index, Chime::Structs::ChimeSoundStruct::Type & chimeSound) = 0;
+    virtual uint8_t GetActiveChimeId()                                                                          = 0;
```
bzbarsky-apple, line 84:
```suggestion
    virtual uint8_t GetActiveChimeID()                                                                          = 0;
```
would be better to match the spec?  Same for set.

src/app/clusters/chime-server/chime-server.h
```diff
@@ -0,0 +1,112 @@
...
+    virtual CHIP_ERROR SetActiveChimeId(uint8_t chimeSoundId) = 0;
+    virtual CHIP_ERROR SetEnabled(bool enabled)               = 0;
+
+    // Commands
+    /**
+     * @brief Delegate should implement a handler to play the currently active chime sound.
+     * It should report Status::Success if successful and may
+     * return other Status codes if it fails
+     */
+    virtual Protocols::InteractionModel::Status playChimeSound() = 0;
```
bzbarsky-apple, line 97:
```suggestion
    virtual Protocols::InteractionModel::Status PlayChimeSound() = 0;
```

src/app/zap-templates/zcl/zcl-with-test-extensions.json
```diff
@@ -193,6 +193,7 @@
             "MaxPathsPerInvoke"
         ],
         "Bridged Device Basic Information": ["ProductAppearance"],
+        "Chime": ["InstalledChimeSounds"],
```
bzbarsky-apple, line 196: Not needed; lists are automatically handled.

On the other hand, you probably _do_ want to add the two non-list attributes handled via AttributeAccessInterface here.

src/app/zap-templates/zcl/zcl.json
```diff
@@ -187,6 +187,7 @@
             "MaxPathsPerInvoke"
         ],
         "Bridged Device Basic Information": ["ProductAppearance"],
+        "Chime": ["InstalledChimeSounds"],
```
bzbarsky-apple, line 190: See above.

src/app/clusters/chime-server/chime-server.cpp
andyg-apple: Redone that section in line with how mode base does it

src/app/clusters/chime-server/chime-server.cpp
andyg-apple: Redone that section in line with how mode base does it - not as efficient but safer I think

src/app/clusters/chime-server/chime-server.cpp
andyg-apple: Moved all the attributes into the server to avoid this, delegate just responsible for the chime details list, and have added comments about that.

src/app/clusters/chime-server/chime-server.cpp
andyg-apple: Error, fixed.

src/app/clusters/chime-server/chime-server.h
andyg-apple: agreed

src/app/zap-templates/zcl/zcl-with-test-extensions.json
andyg-apple: fixed

src/app/zap-templates/zcl/zcl.json
andyg-apple: fixed

src/app/clusters/chime-server/chime-server.cpp
andyg-apple: Error, fixed.

woody-apple (approved):

pidarped (approved):

PR #35897, opened 2024-10-03
Use DataModel::Provider provided command information to handle command validation

andy31415:
When data model provider interface is available, the code now can use the data model  flags for commands (required ACL, flags for needing timed invokes or being scoped) rather than direct generated-code dependencies.


src/app/CommandHandlerImpl.cpp
```diff
@@ -542,30 +505,21 @@ Status CommandHandlerImpl::ProcessGroupCommandDataIB(CommandDataIB::Parser & aCo
...
-        {
-            Access::SubjectDescriptor subjectDescriptor = GetSubjectDescriptor();
-            Access::RequestPath requestPath{ .cluster     = concretePath.mClusterId,
-                                             .endpoint    = concretePath.mEndpointId,
-                                             .requestType = Access::RequestType::kCommandInvokeRequest,
-                                             .entityId    = concretePath.mCommandId };
-            Access::Privilege requestPrivilege = RequiredPrivilege::ForInvokeCommand(concretePath);
-            err                                = Access::GetAccessControl().Check(subjectDescriptor, requestPath, requestPrivilege);
-            if (err != CHIP_NO_ERROR)
+            Status preCheckStatus = mpCallback->ValidateCommandCanBeDispatched(request);
```
andy31415, line 516: Group commands seem to not validate `fabric scoped` nor `timed invoke` flags in previous implementations.

I am unclear why. @bzbarsky-apple @tennessee-google is this a spec requirement? Wondering if I should add that logic (i.e. pass in that the command was invoked via group handling) to the validation or not .... To me it would make sense for validation to be uniform so I left it like that here.

andy31415: It seems at least for timed we have an extra check: https://github.com/project-chip/connectedhomeip/blob/master/src/app/CommandHandlerImpl.cpp#L423C9-L423C21

I will have to decouple that one too. It also seems somewhat coupled to depend on things ordering since code changes and refactors can change things.

andy31415: So this question is now specifically about fabric scoping.

tcarmelveilleux: Group commands cannot be timed-invoked. 

For fabric-scoped, I think this is a possible miss of the implementation?

src/app/InteractionModelEngine.cpp
```diff
@@ -1678,7 +1682,93 @@ void InteractionModelEngine::DispatchCommand(CommandHandlerImpl & apCommandObj,
...
+    Access::RequestPath requestPath{ .cluster     = aRequest.path.mClusterId,
+                                     .endpoint    = aRequest.path.mEndpointId,
+                                     .requestType = Access::RequestType::kCommandInvokeRequest,
+                                     .entityId    = aRequest.path.mCommandId };
+#if CHIP_CONFIG_USE_DATA_MODEL_INTERFACE
+    std::optional<DataModel::CommandInfo> commandInfo = mDataModelProvider->GetAcceptedCommandInfo(aRequest.path);
+    // This is checked by previous validations, so it should not happen
+    VerifyOrReturnValue(commandInfo.has_value(), Status::UnsupportedCommand);
+
+    Access::Privilege requestPrivilege = commandInfo->invokePrivilege;
```
tcarmelveilleux, line 1719: rename `requestPrivilege` to `minimumRequiredPrivilege`

src/app/CommandHandlerImpl.cpp
andy31415: In that case I take it that "check will not hurt" so I do not need to pass in a "invoked via group" flag around to explicitly skip some checks.

src/app/CommandHandlerImpl.cpp
```diff
@@ -542,30 +513,24 @@ Status CommandHandlerImpl::ProcessGroupCommandDataIB(CommandDataIB::Parser & aCo
...
+            if (preCheckStatus != Status::Success)
             {
                 // NOTE: an expected error is CHIP_ERROR_ACCESS_DENIED, but there could be other unexpected errors;
```
bzbarsky-apple, line 527: This comment is no longer true. For one thing, there is no "error" here at all anymore.

src/app/InteractionModelEngine.cpp
```diff
@@ -1678,7 +1682,93 @@ void InteractionModelEngine::DispatchCommand(CommandHandlerImpl & apCommandObj,
 #endif // CHIP_CONFIG_USE_DATA_MODEL_INTERFACE
 }
 
-Protocols::InteractionModel::Status InteractionModelEngine::CommandExists(const ConcreteCommandPath & aCommandPath)
+Protocols::InteractionModel::Status InteractionModelEngine::ValidateCommandCanBeDispatched(const DataModel::InvokeRequest & request)
+{
+
+    Status status = CommandCheckExists(request.path);
```
bzbarsky-apple, line 1688: Should probably be named CheckCommandExistence.

src/app/InteractionModelEngine.cpp
```diff
@@ -1678,7 +1682,93 @@ void InteractionModelEngine::DispatchCommand(CommandHandlerImpl & apCommandObj,
...
+
+    if (status != Status::Success)
+    {
+        ChipLogDetail(DataManagement, "No command " ChipLogFormatMEI " in Cluster " ChipLogFormatMEI " on Endpoint 0x%x",
```
bzbarsky-apple, line 1692: I know you just copied this, but endpoint ids should generally be decimal, not hex....

src/app/InteractionModelEngine.cpp
```diff
@@ -1678,7 +1682,93 @@ void InteractionModelEngine::DispatchCommand(CommandHandlerImpl & apCommandObj,
...
+                      ChipLogValueMEI(request.path.mCommandId), ChipLogValueMEI(request.path.mClusterId), request.path.mEndpointId);
+        return status;
+    }
+
+    status = CommandCheckACL(request);
```
bzbarsky-apple, line 1697: Should probably be called CheckCommandAccess?  Especially because it checks more than just ACL....

src/app/InteractionModelEngine.cpp
```diff
@@ -1678,7 +1682,93 @@ void InteractionModelEngine::DispatchCommand(CommandHandlerImpl & apCommandObj,
...
+    VerifyOrReturnValue(status == Status::Success, status);
+
+    return CommandCheckFlags(request);
```
bzbarsky-apple, line 1700: CheckCommandFlags.

src/app/InteractionModelEngine.cpp
```diff
@@ -1678,7 +1682,93 @@ void InteractionModelEngine::DispatchCommand(CommandHandlerImpl & apCommandObj,
...
+}
+
+Protocols::InteractionModel::Status InteractionModelEngine::CommandCheckACL(const DataModel::InvokeRequest & aRequest)
+{
+    if (!aRequest.subjectDescriptor.has_value())
```
bzbarsky-apple, line 1705: Why is that subjectDescriptor field optional at all?  That seems like very strange API.

I guess for some read/write bits right now it doesn't exit.  But for commands it absolutely always must, and I think just aborting here if it does not is fine.

src/app/InteractionModelEngine.cpp
```diff
@@ -1678,7 +1682,93 @@ void InteractionModelEngine::DispatchCommand(CommandHandlerImpl & apCommandObj,
...
+    }
+
+    Access::RequestPath requestPath{ .cluster     = aRequest.path.mClusterId,
+                                     .endpoint    = aRequest.path.mEndpointId,
+                                     .requestType = Access::RequestType::kCommandInvokeRequest,
+                                     .entityId    = aRequest.path.mCommandId };
+#if CHIP_CONFIG_USE_DATA_MODEL_INTERFACE
+    std::optional<DataModel::CommandInfo> commandInfo = mDataModelProvider->GetAcceptedCommandInfo(aRequest.path);
+    // This is checked by previous validations, so it should not happen
+    VerifyOrReturnValue(commandInfo.has_value(), Status::UnsupportedCommand);
```
bzbarsky-apple, line 1717: This is very much not going to work in a world where the ACL check has to come before the existence check (which is the world of the spec).

Can we please write this code so that it won't break when these two checks are reordered in the callsite?

Now how this is supposed to work... the spec does not in fact define it (needs a spec issue, note @tcarmelveilleux) but would think that you should default to Operate privilege for purposes of the ACL check if the command is not known.

Note that the Ember codepath in the `#else` does exactly that: if not known, uses Operate. We should not have behavior differences between the two codepaths.

src/app/InteractionModelEngine.cpp
```diff
@@ -1678,7 +1682,93 @@ void InteractionModelEngine::DispatchCommand(CommandHandlerImpl & apCommandObj,
...
+    }
+
+    return Status::Success;
+}
+
+Protocols::InteractionModel::Status InteractionModelEngine::CommandCheckFlags(const DataModel::InvokeRequest & aRequest)
+{
+#if CHIP_CONFIG_USE_DATA_MODEL_INTERFACE
+    std::optional<DataModel::CommandInfo> commandInfo = mDataModelProvider->GetAcceptedCommandInfo(aRequest.path);
+    // This is checked by previous validations, so it should not happen
```
bzbarsky-apple, line 1740: Enough "should not" that we should just abort if it does?

src/app/data-model-provider/OperationTypes.h
```diff
@@ -99,6 +100,7 @@ struct InvokeRequest : OperationRequest
 {
     ConcreteCommandPath path;
     BitFlags<InvokeFlags> invokeFlags;
+    FabricIndex accessingFabricIndex = kUndefinedFabricIndex;
```
bzbarsky-apple, line 103: I do have to ask: why do we need this if we have a subject descriptor member already?  What we should have is a getter for the accessing fabric index, which gets it from our subject descriptor, instead of having duplicated members.

src/app/InteractionModelEngine.cpp
andy31415: SubjectDescriptor is currently part of the "common" part for all requests and it is optional since we try to model the ember "internal write" (i.e. we allow writes without needing ACL if internal flags are set). As a sideffect it is optional.

I think aborting is likely similar codes size as this and not crashing the entire app seems more resilient.

src/app/InteractionModelEngine.cpp
andy31415: Sure. I think in the future I will pass this throu so I only call the Get once, however due to ember duality I duplicate some code.

src/app/InteractionModelEngine.cpp
andy31415: switched it to default to operate.

src/app/data-model-provider/OperationTypes.h
andy31415: CommandHandlerExchangeInterface has separate `GetSubjectDescriptor` and `GetAccessingFabricIndex` so they seem to be not coupled.

I could not follow and ensure that the subject descriptor fabricIndex is the `accessingFabricIndex`. I would very much prefer that, however making that change is probably a separate PR by itself. Unfortunately the methods were split and unsplitting them became harder.

We should also figure out what we want our code to look like. I seem to remember when I tried to remove a `GetAccessingFabricIndex` from https://github.com/project-chip/connectedhomeip/blob/master/src/app/AttributeValueEncoder.h#L155 (which is clearly just a getter from the descriptor) I was told that this is a spec term and I should explicitly keep the extra method. I think this resulted in the virtual `GetSubjectDescriptor` that is separate from `AccessingFabricIndex` and now are harder to clean out.

Is it guaranteed that GetAccessingFabricIndex() always is == to SubjectDescriptor.fabricIndex? If so maybe we can drop some of the virtual bits....

src/app/CommandHandlerImpl.cpp
```diff
@@ -524,9 +524,7 @@ Status CommandHandlerImpl::ProcessGroupCommandDataIB(CommandDataIB::Parser & aCo
             Status preCheckStatus = mpCallback->ValidateCommandCanBeDispatched(request);
             if (preCheckStatus != Status::Success)
             {
-                // NOTE: an expected error is CHIP_ERROR_ACCESS_DENIED, but there could be other unexpected errors;
-                // therefore, keep processing subsequent commands, and if any errors continue, those subsequent
-                // commands will likewise fail.
+                // Command failed for a specific path, keep trying the rest of the commands still
```
bzbarsky-apple, line 527:
```suggestion
                // Command failed for a specific path, but keep trying the rest of the paths.
```

src/app/InteractionModelEngine.cpp
bzbarsky-apple: OK.  It just makes it seem like there are "valid" case when it might be nullopt here, which there are not....

src/app/data-model-provider/OperationTypes.h
bzbarsky-apple: > CommandHandlerExchangeInterface has separate GetSubjectDescriptor and GetAccessingFabricIndex 

The latter is a convenience API that lets people easily do the spec operation of "get the accessing fabric index" without messing it up by trying to understand subject descriptors in contexts where they are not relevant.

They are _absolutely_ coupled, and the subject descriptor fabric index _is_ the accessing fabric index.

>  I was told that this is a spec term and I should explicitly keep the extra method

Yes, because it makes it easier to correctly implement spec algorithms that don't mention the subject descriptor but do talk about the "accessing fabric index".

> Is it guaranteed that GetAccessingFabricIndex() always is == to SubjectDescriptor.fabricIndex?

Yes.

andy31415: Updated. I will create a separate PR enforcing that `GetAccessingFabricIndex` is not a virtual call if we have a GetSubjectDescriptor

tcarmelveilleux (approved):

bzbarsky-apple (approved):

PR #35902, opened 2024-10-03
Add missing PICS to TC-SWTCH tests

### LINKED ISSUE #35358: [TC-SWTCH-##] High level PICS has to be added in the Python script.
**Description:**

- [ ] - In the following switch test case, high-level PICS need to be added according to the test plan.

1. TC-SWTCH-2.2
2. TC-SWTCH-2.3
3. TC-SWTCH-2.4
4. TC-SWTCH-2.5
5. TC-SWTCH-2.6

- [x] - TC_SWTCH_2_4 - In Test step 5  ` if MSL or AS feature is supported, TH expect receiving LongPress/LongRelease in that order` has to be updated to ` if MSL feature is supported, TH expect receiving LongPress/LongRelease in that order` as per the Test plan.

- [x] - TC_SWTCH_2_6 - Test step 9b  `TH receives MultiPressComplete event with PreviousPosition set to 1 and TotalNumberOfPressesCounted set to 2 from the DUT"` has to be updated to `TH receives MultiPressComplete event with PreviousPosition set to 1 and TotalNumberOfPressesCounted set to 1 from the DUT` as per the Test plan.

**Test plan References:**
- https://github.com/CHIP-Specifications/chip-test-plans/blob/master/src/cluster/switch.adoc

**Python Script References:**
- https://github.com/project-chip/connectedhomeip/blob/master/src/python_testing/TC_SWTCH.py

708558881bf6e5b0dac3c09489124907927b6cd2

tcarmelveilleux:
Fixes #35358

### DIFF
--- a/src/python_testing/TC_SWTCH.py
+++ b/src/python_testing/TC_SWTCH.py
@@ -300,6 +300,12 @@ class TC_SwitchTests(MatterBaseTest):
                          "Verify that the value is 0, and that a subscription report was received for that change."),
                 ]
 
+    def pics_TC_SWTCH_2_2(self) -> list[str]:
+        return [
+            "SWTCH.S",
+            "SWTCH.S.F00"
+        ]
+
     @run_if_endpoint_matches(has_feature(Clusters.Switch, Clusters.Switch.Bitmaps.Feature.kLatchingSwitch))
     async def test_TC_SWTCH_2_2(self):
         post_prompt_settle_delay_seconds = 10.0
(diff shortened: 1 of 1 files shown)


andy31415 (dismissed):

src/python_testing/TC_SWTCH.py
```diff
@@ -835,6 +859,13 @@ def steps_TC_SWTCH_2_6(self):
                          "Verify that the TH receives MultiPressComplete event with PreviousPosition set to 1 and TotalNumberOfPressesCounted set to 1 from the DUT"),
                 ]
 
+    def pics_TC_SWTCH_2_6(self) -> list[str]:
```
andy31415, line 862: apparently we double-check for run_if_endpoint_matches for these tests and those cannot have pics (we check matching conditions, not pics here)

tcarmelveilleux: That is then a bug...

cecille: no, it's not - these tests are meant to be selected unconditionally, so they have a blank PICS (ie, they're not gated on any condition). TH selects these always, then the test reports if it is applicable.

cecille: run if endpoint matches is just a slightly different mechanism for gating on the PICS.

tcarmelveilleux: @cecille said it's not needed.

PR #35912, opened 2024-10-04
Remove provisional status of Fabric Sync bits to align with the spec

yufengwangca:
Fabric Sync bits are officially landed in Matter spec ToT by default, remove provisional status of Fabric Sync bits in SDK to align with the spec


andy31415 (approved):

andy31415:
@yufengwangca could you confirm that this is not provisional in 1.4 and maybe find the PR number for the spec? 

I merged this to get the most up to date 1.4 branch, however I would also like to ensure we are matching the 1.4 spec and not later.

bzbarsky-apple:
Things are not "not provisional" until the spec is actually finalized.  Certainly not until after SVE happens and 0.9 ballot comments are handled.

PR #35918, opened 2024-10-04
TC-SWTCH - add top level PICS gating for TH

cecille:


tcarmelveilleux (approved):

cecille:
After discussion with TH team, we're going to leave this as is. The test is already self-selecting and after SVE, it is safer to leave this as a test that is always run, and skipped for non-applicable devices as was originally intended.

PR #35919, opened 2024-10-04
One more decouple from ember from src/app and ensure we can compile without `ember-compatibility-functions.cpp`

andy31415:
Final decouple step so that `ember-compatibility-functions.cpp` is not needed as part of data models.

### Changes

- Removed one more direct ember coupling from WriteHandler
- Updated GN code to not auto-compile ember-compatibility-functions.cpp in case data model is enabled (not in check mode)
- Ensure that we test that this mode compiles by making a build on arm64 build this version
- Update pw_rpc code to be able to use the data model interface for encoding attributes
- Also updated cmake to support this option, however cmake build options are more spread out so I did not use the option. Expect to still be ok because link-time we should remove unreferenced functions. The GN build should ensure no references (there is one more call in `pump-configuration-and-control-server.cpp` that we have to figure out for full support like all-clusters app, I will do that as a followup)

### DIFF
--- a/src/app/WriteHandler.cpp
+++ b/src/app/WriteHandler.cpp
@@ -80,7 +80,7 @@ void WriteHandler::Close()
     MoveToState(State::Uninitialized);
 }
 
-bool WriteHandler::IsAttributeList(const ConcreteAttributePath & path)
+bool WriteHandler::IsListAttributePath(const ConcreteAttributePath & path)
 {
 #if CHIP_CONFIG_USE_DATA_MODEL_INTERFACE
     VerifyOrReturnValue(mDataModelProvider != nullptr, false,
@@ -340,7 +340,7 @@ CHIP_ERROR WriteHandler::ProcessAttributeDataIBs(TLV::TLVReader & aAttributeData
         err = element.GetData(&dataReader);
         SuccessOrExit(err);
 
-        if (!dataAttributePath.IsListOperation() && IsAttributeList(dataAttributePath))
+        if (!dataAttributePath.IsListOperation() && IsListAttributePath(dataAttributePath))
         {
             dataAttributePath.mListOp = ConcreteDataAttributePath::ListOperation::ReplaceAll;
         }
@@ -484,7 +484,7 @@ CHIP_ERROR WriteHandler::ProcessGroupAttributeDataIBs(TLV::TLVReader & aAttribut
             {
                 needToCheckListOperation = false;
 
-                if (!dataAttributePath.IsListOperation() && IsAttributeList(dataAttributePath))
+                if (!dataAttributePath.IsListOperation() && IsListAttributePath(dataAttributePath))
                 {
                     dataAttributePath.mListOp = ConcreteDataAttributePath::ListOperation::ReplaceAll;
                 }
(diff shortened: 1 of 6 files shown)


src/app/WriteHandler.cpp
```diff
@@ -460,12 +480,11 @@ CHIP_ERROR WriteHandler::ProcessGroupAttributeDataIBs(TLV::TLVReader & aAttribut
...
-            if (attributeMetadata == nullptr)
+            if (needToCheckListOperation)
             {
-                attributeMetadata = GetAttributeMetadata(dataAttributePath);
-                bool currentAttributeIsList =
-                    (attributeMetadata != nullptr && attributeMetadata->attributeType == kListAttributeType);
-                if (!dataAttributePath.IsListOperation() && currentAttributeIsList)
+                needToCheckListOperation = false;
+
+                if (!dataAttributePath.IsListOperation() && IsListAttributePath(dataAttributePath))
```
bzbarsky-apple, line 487: This change is wrong.  If the attribute does not exist on `mapping.endpoint_id`, this will produce a garbage value for `IsListAttributePath` and never check again, so when it gets to endpoints that _do_ have the attribute the right things won't happen.

Note that the old code did not have this problem: if it go a null from GetAttributeMetadata it would avoid munging the path for this endpoint, but then obviously fail out of the write anyway, so that would not matter much, and for later endpoints it would try to get the metadata again.

src/app/WriteHandler.h
```diff
@@ -189,6 +189,9 @@ class WriteHandler : public Messaging::ExchangeDelegate
     CHIP_ERROR WriteClusterData(const Access::SubjectDescriptor & aSubject, const ConcreteDataAttributePath & aPath,
                                 TLV::TLVReader & aData);
 
+    // Checks if the given path corresponds to a list attribute
```
bzbarsky-apple, line 192: This needs some documentations about the caveats, like what this method does if it does not know the answer.  It would be worth also rephrasing the comment to something like "is known for sure to correspond to a list attribute".

bzbarsky-apple (changes requested):

src/app/WriteHandler.h
andy31415: Seems reasonable (especially regarding the comment before). Will make it return optional/ternary (basically know for sure yes/now and unknown)

src/app/WriteHandler.cpp
andy31415: We are missing test coverage :(

src/app/WriteHandler.h
```diff
@@ -189,6 +189,12 @@ class WriteHandler : public Messaging::ExchangeDelegate
     CHIP_ERROR WriteClusterData(const Access::SubjectDescriptor & aSubject, const ConcreteDataAttributePath & aPath,
                                 TLV::TLVReader & aData);
 
+    /// Checks if the given path corresponds to a list attribute
```
bzbarsky-apple, line 192:
```suggestion
    /// Checks whether the given path corresponds to a list attribute
```

bzbarsky-apple (approved):

tcarmelveilleux (approved):

PR #35616, opened 2024-09-17
Refactoring of Energy Management App

jamesharrow:
This PR moves some of the file contents to more appropriate places, so that the WaterHeater code can implement
EEM/EPM functions which were previously only in the EVSE source files.

* move DEM, PowerTopology, EEM/EPM out of EVSE and into bespoke files.

PR #35948, opened 2024-10-07
[Energy Management] Silabs Dishwasher, Electrical sensor and DEM app

lpbeliveau-silabs:
Adding a Silabs dishwasher app, supports Electrical sensor and DEM.

Relies on energy-management-app's common device-energy-management and energy-reporting files.
Slightly modified energy-management-app's DEM test event trigger and fake readings to allow them to be used by apps that are not EVSE.

Also modified energy-management-app's energy-management-mode to stop enforcing endpoint 1, and added a shutdown callback.

Can be used as POC to implement Energy management apps on top of existing device example apps.

In the long run, I suggest we ad a BUILD.gn file for device-energy-management and energy-reporting so we can just include those files as dependencies of projects that want to re-use them.


src/app/zap-templates/zcl/data-model/chip/matter-devices.xml
```diff
@@ -2685,4 +2685,17 @@ limitations under the License.
...
+        <name>Device Energy Management</name>
+        <domain>CHIP</domain>
+        <typeName>Matter Device Energy Management</typeName>
+        <class>Simple</class>
+        <profileId editable="false">0x0103</profileId>
+        <deviceId editable="false">0x050D</deviceId>
+        <clusters lockOthers="true">
+            <include cluster="Descriptor" client="false" server="true" clientLocked="true" serverLocked="true"></include>
+            <include cluster="Device Energy Management" client="false" server="true" clientLocked="true" serverLocked="true"></include>
+            <include cluster="Device Energy Management Mode" client="false" server="true" clientLocked="true" serverLocked="true"></include>
```
bzbarsky-apple, line 2698: Why is this serverLocked="True"?  This cluster is not mandatory for this device type.

lpbeliveau-silabs: Removed in latest commit.

examples/dishwasher-app/silabs/include/DeviceEnergyManager.h
```diff
@@ -0,0 +1,37 @@
+/*
```
jamesharrow, line 1: Ideally this should be a common file not Silabs specific?

examples/dishwasher-app/silabs/include/ElectricalEnergyMeasurementInstance.h
```diff
@@ -0,0 +1,75 @@
...
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+#pragma once
+
+#include "AppConfig.h"
```
jamesharrow, line 21: Again shouldn't this file be a common file?

examples/dishwasher-app/silabs/src/AppTask.cpp
```diff
@@ -0,0 +1,237 @@
...
+    err = ElectricalSensorManager::Instance().Init();
+    PlatformMgr().UnlockChipStack();
+    if (err != CHIP_NO_ERROR)
+    {
+        SILABS_LOG("ElectricalSensorMgr.Init() failed");
+        appError(err);
+    }
+
+#if SL_MATTER_TEST_EVENT_TRIGGER_ENABLED
+    // Register DEM & Electrical Sensor Test Event Trigger
```
jamesharrow, line 98: Rather than being application specific, could these be in a common Silabs platform could with #ifdefs based on Application example type - e.g. if you had a Water Heater or EVSE build then these would be enabled differently.

It depends on how silicon manufacturers want to support examples I suppose, but from a maintenance perspective from the viewpoint of someone who wants the same code to support all platforms and all applications, this approach seems to be harder to maintain.

Given you have examples/dishwasher-app/ and presumably examples/evse-app and examples water-heater-app etc then this code is going to be very similar in each one, and cut and paste each time?

I can see from the perspective of someone who only wants to make a dishwasher (and not an EVSE) then these #ifdefs are a distraction, but keeping them all in line with each other seems like more work?

examples/dishwasher-app/silabs/src/ElectricalSensorManager.cpp
```diff
@@ -0,0 +1,251 @@
...
+        EPMEndpointId, *gEPMDelegate,
+        BitMask<ElectricalPowerMeasurement::Feature, uint32_t>(
+            ElectricalPowerMeasurement::Feature::kDirectCurrent, ElectricalPowerMeasurement::Feature::kAlternatingCurrent,
+            ElectricalPowerMeasurement::Feature::kPolyphasePower, ElectricalPowerMeasurement::Feature::kHarmonics,
+            ElectricalPowerMeasurement::Feature::kPowerQuality),
+        BitMask<ElectricalPowerMeasurement::OptionalAttributes, uint32_t>(
+            ElectricalPowerMeasurement::OptionalAttributes::kOptionalAttributeRanges,
+            ElectricalPowerMeasurement::OptionalAttributes::kOptionalAttributeVoltage,
+            ElectricalPowerMeasurement::OptionalAttributes::kOptionalAttributeActiveCurrent,
+            ElectricalPowerMeasurement::OptionalAttributes::kOptionalAttributeReactiveCurrent,
```
jamesharrow, line 116: I'd be surprised if any real washing machine / dishwasher manufacturer actually supported some of these readings :) - perhaps you just need voltage, ActiveCurrent and ActivePower?

examples/dishwasher-app/silabs/src/ElectricalSensorManager.cpp
```diff
@@ -0,0 +1,251 @@
...
+    gEPMDelegate = std::make_unique<ElectricalPowerMeasurementDelegate>();
+    VerifyOrReturnError(gEPMDelegate, CHIP_ERROR_NO_MEMORY, ChipLogError(AppServer, "Failed to allocate memory for EPM Delegate"));
+
+    /* Manufacturer may optionally not support all features, commands & attributes */
+    /* Turning on all optional features and attributes for test certification purposes */
+    gEPMInstance = std::make_unique<ElectricalPowerMeasurementInstance>(
+        EPMEndpointId, *gEPMDelegate,
+        BitMask<ElectricalPowerMeasurement::Feature, uint32_t>(
+            ElectricalPowerMeasurement::Feature::kDirectCurrent, ElectricalPowerMeasurement::Feature::kAlternatingCurrent,
+            ElectricalPowerMeasurement::Feature::kPolyphasePower, ElectricalPowerMeasurement::Feature::kHarmonics,
```
jamesharrow, line 110: Would you have a PolyPhase Dishwasher/Washing machine in the real world and Harmonics and PowerQuality - seems overkill to me?

examples/dishwasher-app/silabs/src/ElectricalSensorManager.cpp
```diff
@@ -0,0 +1,251 @@
...
+    // Initialize PT (Power Topology)
+    VerifyOrReturnError(!gPTDelegate && !gPTInstance, CHIP_ERROR_INCORRECT_STATE,
+                        ChipLogError(AppServer, "PowerTopology Delegate or Instance already exist"));
+
+    gPTDelegate = std::make_unique<PowerTopologyDelegate>();
+    VerifyOrReturnError(gPTDelegate, CHIP_ERROR_NO_MEMORY, ChipLogError(AppServer, "Failed to allocate memory for PT Delegate"));
+
+    gPTInstance = std::make_unique<PowerTopologyInstance>(
+        PTEndpointId, *gPTDelegate,
+        BitMask<PowerTopology::Feature, uint32_t>(PowerTopology::Feature::kSetTopology, PowerTopology::Feature::kDynamicPowerFlow),
```
jamesharrow, line 164: Perhaps you don't need all of these features for a simple Dishwasher - you aren't generating power from the Dishwasher I hope?

examples/dishwasher-app/silabs/src/ElectricalSensorManager.cpp
```diff
@@ -0,0 +1,251 @@
...
+static std::unique_ptr<ElectricalPowerMeasurementInstance> gEPMInstance;
+static std::unique_ptr<PowerTopologyDelegate> gPTDelegate;
+static std::unique_ptr<PowerTopologyInstance> gPTInstance;
+
+static const struct
+{
+    PowerModeEnum PowerMode;
+    int64_t Voltage;         // mV
+    int64_t ActiveCurrent;   // mA
+    int64_t ReactiveCurrent; // mA
```
jamesharrow, line 60: I would question if you really need all of these optional measurements for a real device. Voltage, ActiveCurrent, ActivePower and maybe Frequency should be enough for a white goods appliance? But nothing to stop you having them as examples I suppose.

examples/dishwasher-app/silabs/src/ElectricalSensorManager.cpp
```diff
@@ -0,0 +1,251 @@
...
+        if ((updateState < 0) || (updateState > ERROR_STATE_INDEX))
+        {
+            updateState = ERROR_STATE_INDEX;
+        }
+
+        ChipLogDetail(AppServer, "UpdateAllAttributes to Operational State : %d", updateState);
+
+        // Added to support testing using a static array for now
+        gEPMDelegate->SetPowerMode(kAttributes[updateState].PowerMode);
+        gEPMDelegate->SetVoltage(MakeNullable(kAttributes[updateState].Voltage));
```
jamesharrow, line 237: Perhaps some of these might not be needed?

examples/dishwasher-app/silabs/src/DeviceEnergyManager.cpp
```diff
@@ -0,0 +1,167 @@
...
+    VerifyOrReturnError(!gDEMDelegate && !gDEMInstance, CHIP_ERROR_INCORRECT_STATE,
+                        ChipLogError(AppServer, "DEM Delegate or Instance already exist"));
+
+    gDEMDelegate = std::make_unique<DeviceEnergyManagementDelegate>();
+    VerifyOrReturnError(gDEMDelegate, CHIP_ERROR_NO_MEMORY, ChipLogError(AppServer, "Failed to allocate memory for DEM Delegate"));
+
+    // Manufacturer may optionally not support all features, commands & attributes
+    // DeviceEnergyManagement::Feature::kStateForecastReporting is removed from the specification [!PA].a,(STA|PAU|FA|CON),O
+    static uint32_t featureMap = static_cast<uint32_t>(DeviceEnergyManagement::Feature::kPowerForecastReporting) +
+        static_cast<uint32_t>(DeviceEnergyManagement::Feature::kStartTimeAdjustment) +
```
jamesharrow, line 78: I can't see where you are supporting Start Time Adjustment and Power Forecast Adjustment in the code - unless you are reusing the test examples.

It would be nice to put together an example dishwasher power forecast with representative real numbers.

But suspect we need a separate PR for that.

jamesharrow (commented):
I have a PR which is in flight doing some similar things:
https://github.com/project-chip/connectedhomeip/pull/35616

It would be nice to see if we can add a PowerForecast for the dishwasher / washing machine into DEM which shows the estimated duration of each slot etc., but think that is a separate PR to work on.

examples/dishwasher-app/silabs/src/ElectricalEnergyMeasurementInstance.cpp
```diff
@@ -0,0 +1,193 @@
...
+static const MeasurementAccuracyRangeStruct::Type kEnergyAccuracyRanges[] = {
+    { .rangeMin   = 0,
+      .rangeMax   = 1'000'000'000'000'000, // 1 million Mwh
+      .percentMax = MakeOptional(static_cast<chip::Percent100ths>(500)),
+      .percentMin = MakeOptional(static_cast<chip::Percent100ths>(50)) }
+};
+
+static const MeasurementAccuracyStruct::Type kAccuracy = {
+    .measurementType  = MeasurementTypeEnum::kElectricalEnergy,
+    .measured         = true,
```
jamesharrow, line 57: suggest making this False or make a comment to say real appliances should connect this data to a real metering device?

jmartinez-silabs (approved):

bzbarsky-apple (approved):

examples/dishwasher-app/silabs/include/DeviceEnergyManager.h
lpbeliveau-silabs: After review of how the EVSE app is doing its init, I think this file is better of as App Specific. I was even considering merging it into the Dishwasher manager, currently keeping it bu in the long run we might remove it.

examples/dishwasher-app/silabs/include/ElectricalEnergyMeasurementInstance.h
lpbeliveau-silabs: This has app specific data in the cpp and clashes with the current EVSE implementation, I have no strong opinion about it but I would leave this to a further PR.

examples/dishwasher-app/silabs/src/AppTask.cpp
lpbeliveau-silabs: There is quite a lot of rework that needs to be done to refactor this app in a way that removing the energy management is seamless (Electrical Sensor and Device Energy Manager need to be gated by some ifdef, Dishwasher manager also now inherits DEMManufacturerDelegate so it needs some gating, etc.) so I would worry to much about supporting a non EVSE dishwasher app for now. We might update the app in the future if there is a need for it.

examples/dishwasher-app/silabs/src/DeviceEnergyManager.cpp
lpbeliveau-silabs: That will also be put in the back log about improving this app, we are currently re-using the test examples for those just so we can pass matter certs.

PR #35962, opened 2024-10-08
[Darwin] Internal state plumbing for the _XPC classes

jtung-apple:
Currently MTRDevice_XPC and MTRDeviceController_XPC classes have properties / internal state that is queried through XPC every time. This PR aims to plumb initial retrieval and update of this info, to avoid an XPC trip every time.


src/darwin/Framework/CHIP/MTRDeviceController_XPC.mm
```diff
@@ -424,6 +433,62 @@ - (oneway void)device:(NSNumber *)nodeID internalStateUpdated:(NSDictionary *)di
...
+    //     }
+    //     MTRDeviceControllerRegistrationNodeIDsKey: @[
+    //        @{
+    //            MTRDeviceControllerRegistrationNodeIDKey: nodeID,
+    //            MTRDeviceControllerRegistrationDeviceInternalStateKey: deviceInternalStateDictionary
+    //        }
+    //     ]
+    //  }
+
+    NSDictionary * controllerContext = configuration[MTRDeviceControllerRegistrationControllerContextKey];
```
bzbarsky-apple, line 451: Do we need to check that `[controllerContext isKindOfClass:NSDictionary.class]`?  Or more to the point, why are we trusting this is a dictionary, but not trusting that the controllerNodeID is an NSNumber below?

src/darwin/Framework/CHIP/MTRDeviceController_XPC.mm
```diff
@@ -424,6 +433,62 @@ - (oneway void)device:(NSNumber *)nodeID internalStateUpdated:(NSDictionary *)di
...
+    NSNumber * controllerNodeID = controllerContext[MTRDeviceControllerRegistrationControllerNodeIDKey];
+    if (controllerNodeID && [controllerNodeID isKindOfClass:[NSNumber class]]) {
+        _controllerNodeID = controllerContext[MTRDeviceControllerRegistrationControllerNodeIDKey];
+    }
+
+    NSArray * deviceInfoList = configuration[MTRDeviceControllerRegistrationNodeIDsKey];
```
bzbarsky-apple, line 457: Should we be checking that this is `isKindOfClass:NSArray.class`?

src/darwin/Framework/CHIP/MTRDeviceController_XPC.mm
```diff
@@ -424,6 +433,62 @@ - (oneway void)device:(NSNumber *)nodeID internalStateUpdated:(NSDictionary *)di
...
+        NSDictionary * deviceInternalState = deviceInfo[MTRDeviceControllerRegistrationDeviceInternalStateKey];
+        if (deviceInternalState && ![deviceInternalState isKindOfClass:[NSDictionary class]]) {
+            MTR_LOG_ERROR(" - deviceInfo %@ deviceInternalState not NSNumber class %@", deviceInfo, NSStringFromClass([deviceInternalState class]));
+            continue;
+        }
+        if (!nodeID || !deviceInternalState) {
+            MTR_LOG(" - deviceInfo %@ missing elements: nodeID %@ internalState %@", deviceInfo, nodeID, deviceInternalState);
+            continue;
+        }
+        MTRDevice_XPC * device = (MTRDevice_XPC *) [self deviceForNodeID:nodeID];
```
bzbarsky-apple, line 480:
```suggestion
        auto * device = static_cast<MTRDevice_XPC *>([self deviceForNodeID:nodeID]);
```

src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
```diff
@@ -350,6 +350,14 @@ @implementation MTRDevice_Concrete {
     NSDate * _Nullable _mostRecentReportTimeForDescription;
     // Copy of _lastSubscriptionFailureTime that is safe to use in description.
     NSDate * _Nullable _lastSubscriptionFailureTimeForDescription;
+    // Copy of _state that is safe to use in description.
+    MTRDeviceState _deviceStateForDescription;
```
bzbarsky-apple, line 354: In what sense is this ForDescription?  We are not using it in description....

We take the description lock in `_internalProperties` when we are working with the things that are in fact protected by that lock.  It looks to me like we should just take our normal lock there when we need things like state, estimated start time, etc....

Description is special because we might need to get it while holding our normal lock, hence its separate lock.  But would we be calling `_internalProperties` while holding `_lock`?  Looks to me like we are always calling it from an async call on some delegate queue, right?  So just taking `_lock_ and getting the data we want should be fine.

I don't think most of the changes to this file are needed.

src/darwin/Framework/CHIP/MTRDevice_XPC.mm
```diff
@@ -188,16 +190,48 @@ - (oneway void)deviceConfigurationChanged:(NSNumber *)nodeID
...
+    NSNumber * oldStateNumber = self._internalState[kMTRDeviceInternalPropertyDeviceState];
+    NSNumber * newStateNumber = dictionary[kMTRDeviceInternalPropertyDeviceState];
     [self _setInternalState:dictionary];
-    MTR_LOG("%@ internal state updated", self);
+
+    // Call delegate if state changed. State is considered changed if:
+    // 1) old state is nil but new state is not nil
+    // 2) old state is not nil but new state is nil
+    // 3) both old and new state are not nil, and they are not equal
+    if ((!oldStateNumber && newStateNumber) || (oldStateNumber && !newStateNumber) || (oldStateNumber && newStateNumber && ![newStateNumber isEqualToNumber:oldStateNumber])) {
```
bzbarsky-apple, line 204: Could this just use `MTREqualObjects`?

src/darwin/Framework/CHIP/MTRDevice_XPC.mm
```diff
@@ -188,16 +190,48 @@ - (oneway void)deviceConfigurationChanged:(NSNumber *)nodeID
...
+        MTRDeviceState state = static_cast<MTRDeviceState>(newStateNumber ? newStateNumber.unsignedIntegerValue : MTRDeviceStateUnknown);
```
bzbarsky-apple, line 205: Can this just use `self.state`?

src/darwin/Framework/CHIP/MTRDevice_XPC.mm
```diff
@@ -188,16 +190,48 @@ - (oneway void)deviceConfigurationChanged:(NSNumber *)nodeID
...
+    return stateNumber ? static_cast<MTRDeviceState>(stateNumber.unsignedIntegerValue) : MTRDeviceStateUnknown;
+}
+
+- (BOOL)deviceCachePrimed
+{
+    NSNumber * deviceCachePrimedNumber = self._internalState[kMTRDeviceInternalPropertyDeviceCachePrimed];
+    return deviceCachePrimedNumber.boolValue;
+}
+
+- (NSDate *)estimatedStartTime
```
bzbarsky-apple, line 224:
```suggestion
- (nullable NSDate *)estimatedStartTime
```

src/darwin/Framework/CHIP/MTRDevice_XPC.mm
```diff
@@ -188,16 +190,48 @@ - (oneway void)deviceConfigurationChanged:(NSNumber *)nodeID
...
+{
+    return self._internalState[kMTRDeviceInternalPropertyEstimatedStartTime];
+}
 
-MTR_DEVICE_SIMPLE_REMOTE_XPC_GETTER(state, MTRDeviceState, MTRDeviceStateUnknown, getStateWithReply)
-MTR_DEVICE_SIMPLE_REMOTE_XPC_GETTER(deviceCachePrimed, BOOL, NO, getDeviceCachePrimedWithReply)
-MTR_DEVICE_SIMPLE_REMOTE_XPC_GETTER(estimatedStartTime, NSDate * _Nullable, nil, getEstimatedStartTimeWithReply)
-MTR_DEVICE_SIMPLE_REMOTE_XPC_GETTER(estimatedSubscriptionLatency, NSNumber * _Nullable, nil, getEstimatedSubscriptionLatencyWithReply)
+- (NSNumber *)estimatedSubscriptionLatency
```
bzbarsky-apple, line 229:
```suggestion
- (nullable NSNumber *)estimatedSubscriptionLatency
```

src/darwin/Framework/CHIP/MTRDevice_XPC.mm
```diff
@@ -188,16 +190,48 @@ - (oneway void)deviceConfigurationChanged:(NSNumber *)nodeID
...
+        [self _lockAndCallDelegatesWithBlock:^(id<MTRDeviceDelegate> delegate) {
+            [delegate device:self stateChanged:state];
+        }];
+    }
 }
 
-#pragma mark - Remote Commands
+- (MTRDeviceState)state
+{
+    NSNumber * stateNumber = self._internalState[kMTRDeviceInternalPropertyDeviceState];
```
bzbarsky-apple, line 214: I don't see us allocating any _internalState on init, so it might be nil.  What does this square bracket dictionary notation do when the dictionary is nil?  Just return nil?

bzbarsky-apple (dismissed):
Requesting changes for the MTRDevice_Concrete bits.

src/darwin/Framework/CHIP/MTRDevice_XPC.mm
woody-apple: Yes

src/darwin/Framework/CHIP/MTRDevice_XPC.mm
woody-apple: It does

src/darwin/Framework/CHIP/MTRDevice_XPC.mm
woody-apple: Yup

src/darwin/Framework/CHIP/MTRDeviceController_XPC.mm
woody-apple: Done

src/darwin/Framework/CHIP/MTRDevice_XPC.mm
```diff
@@ -211,20 +203,12 @@ - (oneway void)deviceBecameActive:(NSNumber *)nodeID
...
-        MTR_LOG_ERROR("%@ invalid deviceCachePrimed: nodeID: %@", self, nodeID);
-        return;
-    }
-
-    [self _lockAndCallDelegatesWithBlock:^(id<MTRDeviceDelegate> delegate) {
-        if ([delegate respondsToSelector:@selector(deviceCachePrimed:)]) {
-            [delegate deviceCachePrimed:self];
-        }
-    }];
+    // Not needed since this is a state udpate now
```
woody-apple, line 206:
```suggestion
    // Not needed since this is a state update now
```

woody-apple (approved):

src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
woody-apple: @bzbarsky-apple Let's follow up here in an issue.

src/darwin/Framework/CHIP/MTRDeviceController_XPC.mm
```diff
@@ -424,6 +433,60 @@ - (oneway void)device:(NSNumber *)nodeID internalStateUpdated:(NSDictionary *)di
...
+    //        @{
+    //            MTRDeviceControllerRegistrationNodeIDKey: nodeID,
+    //            MTRDeviceControllerRegistrationDeviceInternalStateKey: deviceInternalStateDictionary
+    //        }
+    //     ]
+    //  }
+
+    NSDictionary * controllerContext = MTR_SAFE_CAST(configuration[MTRDeviceControllerRegistrationControllerContextKey], NSDictionary);
+    NSNumber * controllerNodeID = MTR_SAFE_CAST(controllerContext[MTRDeviceControllerRegistrationControllerNodeIDKey], NSNumber);
+    if (controllerNodeID && controllerNodeID) {
```
bzbarsky-apple, line 453:
```suggestion
    if (controllerContext && controllerNodeID) {
```

src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
bzbarsky-apple: No, my point is we should not be making these changes.  I'll go ahead and just fix this part up....

src/darwin/Framework/CHIP/MTRDevice_XPC.mm
```diff
@@ -113,52 +114,44 @@ - (NSString *)description
...
+                     wifi,
+                     thread,
+                     _deviceController.uniqueIdentifier,
+                     (unsigned long) self.state];
 }
 
 - (nullable NSNumber *)vendorID
 {
-    return [[self._internalState objectForKey:kMTRDeviceInternalPropertyKeyVendorID] copy];
+    return [self._internalState objectForKey:kMTRDeviceInternalPropertyKeyVendorID];
```
bzbarsky-apple, line 136:
```suggestion
    return [[self._internalState objectForKey:kMTRDeviceInternalPropertyKeyVendorID] copy];
```
is what our API promises.

src/darwin/Framework/CHIP/MTRDevice_XPC.mm
```diff
@@ -113,52 +114,44 @@ - (NSString *)description
...
 }
 
 - (nullable NSNumber *)productID
 {
-    return [[self._internalState objectForKey:kMTRDeviceInternalPropertyKeyProductID] copy];
+    return [self._internalState objectForKey:kMTRDeviceInternalPropertyKeyProductID];
```
bzbarsky-apple, line 141:
```suggestion
    return [[self._internalState objectForKey:kMTRDeviceInternalPropertyKeyProductID] copy];
```

bzbarsky-apple (approved):

src/darwin/Framework/CHIP/MTRDeviceController.mm
```diff
@@ -714,7 +714,7 @@ - (NSUInteger)_iterateDelegateInfoWithBlock:(void (^_Nullable)(MTRDeviceControll
     }
 }
 
-- (void)_callDelegatesWithBlock:(void (^_Nullable)(id<MTRDeviceControllerDelegate> delegate))block logString:(const char *)logString;
+- (void)_callDelegatesWithBlock:(void (^_Nullable)(id<MTRDeviceControllerDelegate> delegate))block logString:(const char *)logString
```
woody-apple, line 717: need to add delegate

PR #35964, opened 2024-10-08
[Tizen] Enable QEMU RTC driver to sync time on boot

arkq:
### Problem

Tizen QEMU test sometimes fails because during commissioning it's not possible to verify certificate. That happens because Tizen QEMU guest is too slow to synchronize time over NTP.

### Changes

- Synchronize system time with QEMU provided RTC on boot time, instead of using NTP

### Testing

Verified locally that guest has recent date without network enabled.


XToripuru:
The changes function as expected in my local environment.

XToripuru (approved):

andy31415 (approved):

PR #26348, opened 2023-05-03
[ESP32] Use esp_log_writev() for logging rather than ESP_LOGx macros

shubhamdp:
Basically this un-reverts the #26282 with additional changes to fixed the rpc logging problem introduced in #26227.

#### Change Overview
- Earlier a buffer was used to create a log, we won't need that buffer, so getting rid of that buffer.
#26227 tried doing the same thing, but introduced the bug in rpc logging and that PR was reverted with minor changes in #26282.
- Took the changes from #26227 and fixed the RPC logging bug.
- Fix the ESP32 to Chip log level mapping.

#### Tests
- RPC
    - Compiled using ./scripts/build/build_examples.py --target esp32-m5stack-all-clusters-rpc build.
    - Flashed onto ESP32 devkit using ./out/esp32-m5stack-all-clusters-rpc/chip-all-clusters-app.flash.py
    - Attached to pigweed rpc console python3 -m chip_rpc.console --device /dev/cu.usbserial-14340
    - Confirmed that chip and other logs can be seen on rpc console.
    - Tried commissioning using chip-tool
- non-RPC
    - Compiled all-clusters-app/esp32, flashed on ESP32, commissioned the device and verified all `CHIP[..]` and other logs were seen.
- For log level mapping
    - Validated the `build/esp-idf/chip/args.gn`, before change all log levels were set to true in here. After the change, correct values are being set.

PR #35965, opened 2024-10-08
esp32-fix-empty-logs(#35939)

### LINKED ISSUE #35939: CHIP ignore esp-idf log level and print "empty" messages
https://github.com/project-chip/connectedhomeip/blob/bcc1f1b3b8243326ec1818b0d1b84c37464bd0f6/src/platform/ESP32/Logging.cpp#L51

When the chip calls the esp logs, the lines with "printf" ignores the log level, I think this can be fixed as follows:
`esp_log_write(ESP_LOG_DEBUG, tag, LOG_COLOR_D "D (%" PRIu32 ") %s: ", esp_log_timestamp(), tag);`
`esp_log_writev(ESP_LOG_DEBUG, tag, msg, v);`
`esp_log_write(ESP_LOG_DEBUG, tag, LOG_RESET_COLOR "\n");`

mokap13:
Fixes #35939 
Replaced `printf` with `esp_log_write`, that check log level


andy31415 (approved):

bzbarsky-apple (approved):

shubhamdp:
Hi @mokap13,

The earlier changes (#26348) were made with specific requirements in mind. The `printf` statements are not bypassing the log level; rather, they are intentionally designed to print both the initial and later parts of the log.

The log format consists of three parts:
1. Initial Part: This includes the log level, timestamp, and tag.  - `I (117262) chip[ZCL]:`
2. Middle Part: This is the actual log content. - `Endpoint 1 On/off already set to new value`
3. End Part: This contains just the end of the color code character.

`esp_log_write()` internally calls `esp_log_writev()`, now with these changes, it generate 3 logs, where the first and third is pretty useless. (ref: https://github.com/espressif/esp-idf/blob/v5.3/components/log/log.c#L214-L222)

We intentionally added the printf statements to ensure we could apply colors to logs redirected to the console while optimizing the usage of buffer. These changes were introduced in [this PR](https://github.com/project-chip/connectedhomeip/pull/26348).

Given this context, I believe the change in #35965 should be reverted.

PR #35971, opened 2024-10-08
Speed up linux and darwin CI

andy31415:
#### restrict input files in clang-tidy runs to only changed files

This will not catch as many things as headers are not directly checked, so we may only find things as a sideffect, however this should reduce the almost 1-hour long clang-tidy runs to much shorter.

#### Run sanitizers tests only on non-pull-requests

This should speed up CI for pull requests. We still run tests, except without sanitizers and this should be significantly faster (do not run the same test 3x times with some potential for finding issues).


.github/workflows/build.yaml
```diff
@@ -186,6 +189,10 @@ jobs:
             - name: Clean output
               run: rm -rf ./out
             - name: Run Tests with sanitizers
+              # Sanotizer tests are not likely to find extra issues so running the same tests
+              # as above repeateadly on every pull request seems extra time. Instead keep this run
+              # for master onlyu
```
bzbarsky-apple, line 194:
```suggestion
              # Sanitizer tests are not likely to find extra issues so running the same tests
              # as above repeatedly on every pull request seems extra time. Instead keep this run
              # for master only
```

.github/workflows/build.yaml
```diff
@@ -425,14 +452,24 @@ jobs:
...
                   ./scripts/run_in_build_env.sh "./scripts/run_codegen_targets.sh out/default"
+            - name: Find changed files
+              id: changed-files
+              uses: tj-actions/changed-files@v45
             - name: Clang-tidy validation
               # NOTE: clang-tidy crashes on CodegenDataModel_Write due to Nullable/std::optional check.
               #       See https://github.com/llvm/llvm-project/issues/97426
               run: |
+                  touch out/changed_files.txt
+                  for file in ${ALL_CHANGED_FILES}; do
```
bzbarsky-apple, line 463: Two things:

1. What is ALL_CHANGED_FILES here? Does not seem to be set.
2. Does this work correctly for files with spaces in their names?  Was that tested?

andy31415: The constant comes from the env setting (via changed-files).

I tested by seeing how these are reported in logs and locally creating a file with fake cpp files for testing. 
I did not test filenames with spaces .... probably it would consider them separate lines and will not activate, however at the same time at least for posix files we do not have source files with spaces.

andy31415: Ah, for darwin I did not copy the env. Will do so.

andy31415: Updated. Regarding spaces, we do not seem to use C++ files in places containing spaces:

```
src/darwin/CHIPTool/CHIPTool/UI Helpers/CHIPUIViewUtils.m
src/darwin/CHIPTool/CHIPTool/Framework Helpers/FabricKeys.m
src/darwin/CHIPTool/CHIPTool/Framework Helpers/DefaultsUtils.m
src/darwin/CHIPTool/CHIPTool/View Controllers/AppDelegate.m
src/darwin/CHIPTool/CHIPTool/View Controllers/RootViewController.m
src/darwin/CHIPTool/CHIPTool/View Controllers/SceneDelegate.m
src/darwin/CHIPTool/CHIPTool/View Controllers/WiFi/WiFiViewController.m
src/darwin/CHIPTool/CHIPTool/View Controllers/QRCode/QRCodeViewController.m
src/darwin/CHIPTool/CHIPTool/View Controllers/Enumeration/EnumerateViewController.m
src/darwin/CHIPTool/CHIPTool/View Controllers/Fabric/FabricUIViewController.m
src/darwin/CHIPTool/CHIPTool/View Controllers/DeviceSelector.m
src/darwin/CHIPTool/CHIPTool/View Controllers/Temperature Sensor/TemperatureSensorViewController.m
src/darwin/CHIPTool/CHIPTool/View Controllers/Bindings/BindingsViewController.m
src/darwin/CHIPTool/CHIPTool/View Controllers/MultiAdmin/MultiAdminViewController.m
src/darwin/CHIPTool/CHIPTool/View Controllers/UnpairDevices/UnpairDevicesViewController.m
src/darwin/CHIPTool/CHIPTool/View Controllers/OnOffCluster/OnOffViewController.m
```

unsure if we actually clang-tidy those files, we generally tidy cpp source files.

andy31415: we could handle spaces in names if we really wanted (https://github.com/tj-actions/changed-files/issues/216) however for now it does not seem needed.

arkq (approved):

tcarmelveilleux (approved):

PR #35975, opened 2024-10-08
Add script and workflow that is able to "cancel all CI runs for a specific PR/commit"

andy31415:
Intent is to be able to encode a "when a workflow fails, cancel all the rest of the workflows". This seems like something desirable when some of the faster checks fail (e.g. restyle or minimal build or zap, which will require a PR update anyway).


.github/workflows/cancel_workflows_for_pr.yaml
```diff
@@ -0,0 +1,55 @@
...
+                      click       \
+                      coloredlogs \
+                      pygithub    \
+                      && echo "DONE installint python prerequisites"
+            - name: Cancel runs
+              run: |
+                  scripts/tools/cancel_workflows_for_pr.py       \
+                    --pull-request ${{ inputs.pull_request_id }} \
+                    --commit-sha "${{ inputs.commit_sha }}"      \
+                    --token "${{ secrets.GITHUB_TOKEN }}"
```
arkq, line 55: I would consider using a different name, maybe more specific to that purpose, so noone will elevate permission for that token. The case is, that someone might create PR with some typo fix, which for sure will be accepted. Then such person will become collaborator, so CI will run without approval. Then such eve-collaborator might change `cancel_workflows_for_pr.py` to dump the secret to output (after obfuscation of course).... PR will fail and trigger cancellation... and repo might be compromised.

scripts/tools/cancel_workflows_for_pr.py
```diff
@@ -0,0 +1,87 @@
...
+
+@click.command()
+@click.option(
+    "--log-level",
+    default="INFO",
+    type=click.Choice(list(__LOG_LEVELS__.keys()), case_sensitive=False),
+    help="Determines the verbosity of script output.",
+)
+@click.option("--pull-request", type=int, help="Pull request number to consider")
+@click.option("--commit-sha", help="Commit to look at when cancelling pull requests")
```
arkq, line 67: I think that this should be the HEAD commit. Otherwise, we might have a race where someone will push new changes, but this script will kill workflows because of not-killed-yet-by-github run failure, because given SHA is in the list of commits (it's not the last though).

arkq (approved):

scripts/tools/cancel_workflows_for_pr.py
andy31415: Github seems to associate checks with individual commits. My thinking was that we want to cancel workflow runs for the commit that failed but not cancel others. Situation like this:

- bad commit starts CI  (SHA A)
- developer notices bad commit and submits a fix (SHA B)
- new commit starts CI
- CI in SHA A fails and triggers "cancel"

in the above I want SHA-A to only cancel workflow runs for A and not for B.
Does this seem reasonable? what scenarios did you have in mind?

arkq: > Github seems to associate checks with individual commits.

OK I see, that might have sense.

.github/workflows/cancel_workflows_for_pr.yaml
arkq: > I am unsure if/how we can mitigate this as we need at least some "cancel workflows" access.

I've just checked how `secrets.GITHUB_TOKEN` and it seems to be short-lived token which is generated for the workflow only. I was not aware of that.
>The GITHUB_TOKEN expires when a job finishes or after a maximum of 24 hours.

tcarmelveilleux (approved):

cecille (approved):

Issue #35976, opened 2024-10-08
[Feature] Fabric-Admin to dynamically identify the endpoint hosting the aggregator at runtime.


yufengwangca:
### Feature description

...
With the latest update to the FS system diagram in the specification, it is now explicitly required that CCTRL be enabled on the endpoint hosting the aggregator.

CCTRL is a feature of the Aggregator device type when FS is enabled and must be activated on the corresponding endpoint. 

In the example apps, the Aggregator is statically configured on Endpoint 1 using zap, and the Aggregator endpoint is statically set. The next optimization should enable the Fabric-Admin to dynamically identify the endpoint hosting the aggregator at runtime.

### Platform

raspi

PR #35977, opened 2024-10-08
[Fabric-Sync] Use aggregator endpoint instead of root endpoint for CCTRL

yufengwangca:
With the latest update to the FS system diagram in the specification, it is now explicitly required that CCTRL be enabled on the endpoint hosting the aggregator.

CCTRL is a feature of the Aggregator device type when FS is enabled and must be activated on the corresponding endpoint. 

In the example apps, the Aggregator is statically configured on Endpoint 1 using zap, and the Aggregator endpoint is statically set. The next optimization should enable the Fabric-Admin to dynamically identify the endpoint hosting the aggregator at runtime. (https://github.com/project-chip/connectedhomeip/issues/35976)

### DIFF
--- a/src/python_testing/TC_MCORE_FS_1_1.py
+++ b/src/python_testing/TC_MCORE_FS_1_1.py
@@ -51,6 +51,8 @@ from chip.testing.apps import AppServerSubprocess
 from matter_testing_support import MatterBaseTest, TestStep, async_test_body, default_matter_test_main
 from mobly import asserts
 
+_DEVICE_TYPE_AGGREGATOR = 0x000E
+
 
 class TC_MCORE_FS_1_1(MatterBaseTest):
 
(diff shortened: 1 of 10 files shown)


src/python_testing/TC_MCORE_FS_1_1.py
```diff
@@ -125,7 +125,7 @@ def default_timeout(self) -> int:
     @async_test_body
     async def test_TC_MCORE_FS_1_1(self):
         # TODO this value should either be determined or passed in from command line
-        dut_commissioning_control_endpoint = 0
+        dut_commissioning_control_endpoint = 1
```
tehampson, line 128: We will need to get this value some other way we cannot hardcode this. Likely we need this value to come in as a command line argument

src/python_testing/TC_MCORE_FS_1_3.py
```diff
@@ -128,6 +129,7 @@ async def commission_via_commissioner_control(self, controller_node_id: int, dev
 
         await self.send_single_cmd(
             node_id=controller_node_id,
+            endpoint=1,
```
tehampson, line 132: We will need to get this value some other way we cannot hardcode this. Likely we need this value to come in as a command line argument

src/python_testing/TC_MCORE_FS_1_3.py
```diff
@@ -140,6 +142,7 @@ async def commission_via_commissioner_control(self, controller_node_id: int, dev
 
         resp = await self.send_single_cmd(
             node_id=controller_node_id,
+            endpoint=1,
```
tehampson, line 145: We will need to get this value some other way we cannot hardcode this. Likely we need this value to come in as a command line argument

arkq (approved):

tehampson (approved):

PR #35979, opened 2024-10-08
[Silabs] Refactor Sensor code

mkardous-silabs:
#### Description
Goal of the PR is to update the sensor software that was used by the Thermostat to make it more generic.

- Transition the ThermostatSensor code to Si70xxSensor to represent physical leverage and make it less temperature specific. Humidity sensor is used in other sample apps now.
- Update gn arguments and define to fit with the renaming
- make WriteUI() public to avoid needing to set a boolean that had to no impact on the LCD
- Update the thermostat app to always update the value attribute but only trigger a report if the temperature variation is over a given threshold.

####Tests 
Manual tests


examples/platform/silabs/display/lcd.h
```diff
@@ -65,6 +65,7 @@ class SilabsLCD
     int DrawPixel(void * pContext, int32_t x, int32_t y);
     int Update(void);
     void WriteDemoUI(bool state);
+    void WriteDemoUI();
```
lpbeliveau-silabs, line 68: Right above there is `Update(void)`, we might want to define our coding standard regarding these declarations.

third_party/silabs/efr32_sdk.gni
```diff
@@ -91,6 +91,15 @@ declare_args() {
 
   # Enable TestEventTrigger in GeneralDiagnostics cluster
   sl_enable_test_event_trigger = false
+
+  # Temperature Sensor support
+  sl_enable_si70xx_sensor = false
```
lpbeliveau-silabs, line 96: Could we add a comment about which board support this sensor and/or have a check depending on the board used to enable it? Unless it is available on all our matter devboard off course.

third_party/silabs/efr32_sdk.gni
```diff
@@ -91,6 +91,15 @@ declare_args() {
...
+}
+
+declare_args() {
+  # This argument is for retro-compatibility since use_temp_sensor was used by the thermostat app.
+  # It should not be used anymore.
+  use_temp_sensor = sl_enable_si70xx_sensor
```
jmartinez-silabs, line 102: why would we need this still?

lpbeliveau-silabs (approved):

examples/platform/silabs/display/lcd.h
mkardous-silabs: I agree. In my opinion, these class / folder would need a clean and documentation. It is going to in a follow up to avoid muddying the waters of this PR.

examples/thermostat/silabs/src/SensorManager.cpp
```diff
@@ -81,19 +82,20 @@ void SensorManager::SensorTimerEventHandler(void * arg)
     int16_t temperature            = 0;
     static int16_t lastTemperature = 0;
 
-#ifdef USE_TEMP_SENSOR
-    int32_t tempSum   = 0;
-    uint32_t humidity = 0;
+#if defined(SL_MATTER_USE_SI70XX_SENSOR) && SL_MATTER_USE_SI70XX_SENSOR
+    int16_t tempSum   = 0;
```
jmartinez-silabs, line 86: we are taking 100 samples of decidegree Celsius, we could overflow the int16_t. You should keep the int32_t

third_party/silabs/efr32_sdk.gni
mkardous-silabs: I was planning instead of adding that information to the documentation of the sensor app instead of here. 
I can add an assert of the boards that can't support it.

third_party/silabs/efr32_sdk.gni
mkardous-silabs: Really left it in case a customer has their product use that flag. I can remove it we don't really care about that use-case.

examples/thermostat/silabs/src/SensorManager.cpp
```diff
@@ -109,18 +111,20 @@ void SensorManager::SensorTimerEventHandler(void * arg)
...
     if ((temperature >= (lastTemperature + kMinTemperatureDelta)) || temperature <= (lastTemperature - kMinTemperatureDelta))
     {
-        lastTemperature = temperature;
-        PlatformMgr().LockChipStack();
-        // The SensorMagager shouldn't be aware of the Endpoint ID TODO Fix this.
-        // TODO Per Spec we should also apply the Offset stored in the same cluster before saving the temp
-
-        app::Clusters::Thermostat::Attributes::LocalTemperature::Set(kThermostatEndpoint, temperature);
-        PlatformMgr().UnlockChipStack();
+        reportState = MarkAttributeDirty::kYes;
```
jmartinez-silabs, line 121: should it be kIfChanged to compared to the last report

mkardous-silabs: No, since we check that the value differs of a certain threshold. If we get here, we know the value changed for sure.
kYes and kIfChanged would have the same behavior.

third_party/silabs/efr32_sdk.gni
mkardous-silabs: removing the old flags

third_party/silabs/efr32_sdk.gni
lpbeliveau-silabs: +1 for the assert

jmartinez-silabs (approved):

andy31415 (approved):

PR #35981, opened 2024-10-08
Update ZAP XMLs using Alchemy automated generation

ethanzhouyc:
Command used for automated generation:
`alchemy zap --specRoot=../connectedhomeip-spec --sdkRoot=../connectedhomeip --attribute="in-progress" --conformanceXML=true`
After running the command, CI jobs failed. Commits in this PR include generated changes and temporary fixes for the following issues to allow the CI to pass:

(1) Inconsistent XML Attribute Definitions:
Some newly added attributes lack the 'optional', 'writable', or 'reportable' properties, leading to inconsistencies with pre-existing attributes that include these properties.
Examples: `SubjectsPerAccessControlEntry` in `access-control-cluster.xml` lacks 'optional' and 'writable' tags; `CatalogList` in `application-launcher-cluster.xml` has had its 'reportable' tag removed

(2) Duplicate Enums and Structs:
`MeasurementTypeEnum`, `MeasurementAccuracyRangeStruct`, and `MeasurementAccuracyStruct` are duplicated in both `measurement-and-sensing.xml` and global files (`global-enums.xml` or `global-struct.xml`), which caused errors in CI jobs. To pass CI for this PR, one set of the duplicates was removed.

(3) Undefined data types:
`CopyScene` command in `scene.xml` has 2 arguments with type `max 254` , which is correct according to the cluster spec. However, `max 254` is not documented in the core spec or listed in `chip-types.xml`, resulting in failed CI.

(4) Undefined event priority:
The `DoorStateChange`, `LockOperation`, and `LockOperationError` events have `desc` priority as specified in the cluster spec, but `desc` priority is not included in the list of valid priorities in the core spec, nor is it expected by CI jobs, leading to errors.

(5) Parsing issue caused by domain definition change:
Alchemy updated the format of domain definitions within the `<cluster>` tag from `<domain>General</domain>`  to `<domain name="General"/>`. The updated format is not correctly parsed by ZAP. This bug will be fixed by changes in ZAP.

@hasty @andy31415 
Can you review these issues and suggest potential solutions? Thank you!


src/app/zap-templates/zcl/data-model/chip/access-control-cluster.xml
```diff
@@ -14,11 +14,17 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+Source: src/data_model/ACL-Cluster.adoc
+Parameters: in-progress 
+Git: 0.9-fall2024-87-gcdfdb86be-dirty
+-->
+<configurator xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="../../zcl.xsd">
   <domain name="CHIP"/>
-
   <enum name="AccessControlEntryPrivilegeEnum" type="enum8">
     <cluster code="0x001F"/>
+    <cluster code="0x0752"/>
```
bzbarsky-apple, line 27: The spec here is invalid, in referencing a data type from one cluster in another cluster.  The spec needs to be fixed.

More importantly, this is NOT just an "add conformance data" change.

src/app/zap-templates/zcl/data-model/chip/access-control-cluster.xml
```diff
@@ -98,70 +107,101 @@ limitations under the License.
       <feature bit="0" code="EXTS" name="Extension" summary="Device provides ACL Extension attribute">
         <optionalConform/>
       </feature>
-      <feature bit="1" code="MNGD" name="Managed Device" summary="Device is managed">
-        <optionalConform/>
-      </feature>
+      <feature bit="1" code="MNGD" name="ManagedDevice" summary="Device is managed"/>
```
bzbarsky-apple, line 110: Why is this removing the "optionalConform"?  Spec has Conformance as "desc", which is very much not the same thing as "required".

src/app/zap-templates/zcl/data-model/chip/access-control-cluster.xml
```diff
@@ -98,70 +107,101 @@ limitations under the License.
...
+        <feature name="EXTS"/>
+      </mandatoryConform>
+    </attribute>
+    <attribute side="server" code="0x0002" define="SUBJECTS_PER_ACCESS_CONTROL_ENTRY" type="int16u" min="4" default="4">
+      <description>SubjectsPerAccessControlEntry</description>
+      <mandatoryConform/>
+    </attribute>
+    <attribute side="server" code="0x0003" define="TARGETS_PER_ACCESS_CONTROL_ENTRY" type="int16u" min="3" default="3">
+      <description>TargetsPerAccessControlEntry</description>
+      <mandatoryConform/>
```
bzbarsky-apple, line 138: Do we really want to be explicitly listing all mandatory things instead of just defaulting things mandatory unless stated otherwise?

scripts/rules.matterlint
bzbarsky-apple, line 1: If this PR is going to be reviewable, it probably needs to be broken out into multiple PRs, and each of those will need to be carefully compared with the spec for the relevant cluster(s).

bzbarsky-apple (changes requested):

bzbarsky-apple:
> CopyScene command in scene.xml has 2 arguments with type max 254 , which is correct according to the cluster spec. 

This is clearly a bug in the spec that needs to be fixed. That "max 254" should be in the constraint column, and the type should be an actual type.  @lpbeliveau-silabs

hasty:
On the issues listed above:

1. Alchemy leaves out attributes which have their default value. In the example you provided, `SubjectsPerAccessControlEntry` is not writable, and is mandatory, so both attributes are false, which is the default.
2. Those enums and structs predate support for global enums and structs; we should relocate them to the global file.
3. This is an error in the spec; that table has constraints under its type column.
4. The Door Lock ZAP template XML is (to my understanding) one of the first written, and thus is quite idiosyncratic. Most of the other oddities in the ZAP XML can be worked around through the errata system, but Door Lock is particularly unusual.

lpbeliveau-silabs:
> > CopyScene command in scene.xml has 2 arguments with type max 254 , which is correct according to the cluster spec.
> 
> This is clearly a bug in the spec that needs to be fixed. That "max 254" should be in the constraint column, and the type should be an actual type. @lpbeliveau-silabs

PR opened here: https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/10573

src/app/zap-templates/zcl/data-model/chip/access-control-cluster.xml
ethanzhouyc: What does `desc` mean in conformance? I did not see it's explained in the core spec. This irregularity might make Alchemy not parsing it correctly

ethanzhouyc:
> On the issues listed above:
> 
> 1. Alchemy leaves out attributes which have their default value. In the example you provided, `SubjectsPerAccessControlEntry` is not writable, and is mandatory, so both attributes are false, which is the default.
> 2. Those enums and structs predate support for global enums and structs; we should relocate them to the global file.
> 3. This is an error in the spec; that table has constraints under its type column.
> 4. The Door Lock ZAP template XML is (to my understanding) one of the first written, and thus is quite idiosyncratic. Most of the other oddities in the ZAP XML can be worked around through the errata system, but Door Lock is particularly unusual.

Thanks for the explanation! The spec needs to be fixed, and after that, I will work on resolving remaining issues

src/app/zap-templates/zcl/data-model/chip/access-control-cluster.xml
ethanzhouyc: I think leaving mandatoryConform there will easier for paring XML data

ethanzhouyc: If you decide we should remove them, it should be an easy change in Alchemy

scripts/rules.matterlint
ethanzhouyc: Okay thanks for the review, will do that!

src/app/zap-templates/zcl/data-model/chip/access-control-cluster.xml
bzbarsky-apple: The point is: we should not be adding new XMLs as part of this automated PR.  Adding an XML file should be done when we are actually working on implementation for that cluster...

src/app/zap-templates/zcl/data-model/chip/access-control-cluster.xml
bzbarsky-apple: "desc" in conformance means "the actual conformance is described as English prose, sucks to be you if you need machine-readable conformance".

But typically it does NOT mean "M", because if it were "M" it would just say "M".  It typically means some sort of "required not always but under some sort of complicated circumstances".

src/app/zap-templates/zcl/data-model/chip/access-control-cluster.xml
bzbarsky-apple: I'm not sure what you mean by "parsing conformance from XML data".  I think it makes sense to say "conformance is M unless stated otherwise"...

In general, we need to decide whether the XML will always spell out all the things or just list them when they have non-default values. There are reasons to go in either direction, honestly, and tradeoffs in terms of what happens when a new "thing" is added vs how easy it is to extract the relevant information from the XML without having as separate list of defaults somewhere.

Either way, please consider how this is meant to actually get reviewed.  The bigger the deltas, the more time-intensive the review process.

src/app/zap-templates/zcl/data-model/chip/access-control-cluster.xml
bzbarsky-apple: Note that this applies to the event priority too: whenever the spec says "desc" that means "not machine readable, sorry".

ethanzhouyc: Thank you for the explanation!

src/app/zap-templates/zcl/data-model/chip/access-control-cluster.xml
ethanzhouyc: I agree that we should make the deltas smaller to ease the review process. Updating the entire XML introduces too large of a delta, so I'll switch to using Alchemy to only generate conformance data, which should make things more manageable.

ethanzhouyc:
Thanks for your review. Updating the entire ZAP XML introduces too large of a delta to manage and get reviewed effectively, so I will create a new PR focused solely on updating the conformance data.

PR #35982, opened 2024-10-08
Fixing deadlock, and not cleaning up on purpose now

woody-apple:
There's a deadlock here, fixing that

Also removing our cleanup, we should just exit

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice_XPC.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice_XPC.mm
@@ -64,20 +64,14 @@
 
 #import <os/lock.h>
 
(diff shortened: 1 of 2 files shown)


src/darwin/Framework/CHIP/MTRDevice_XPC.mm
```diff
@@ -250,7 +222,7 @@ - (void)_invokeCommandWithEndpointID:(NSNumber *)endpointID
 
     @try {
         [[xpcConnection remoteObjectProxyWithErrorHandler:^(NSError * _Nonnull error) {
-            MTR_LOG_ERROR("Error: %@", error);
```
woody-apple, line 253: add this back in

anush-apple (approved):

PR #35984, opened 2024-10-09
tv-casting-app fix CloseCommissioningWindow upon unexpected CDC Commi…

pgregorr-amazon:
A fix for Github [BUG] Can't connect Android CHPTVCastingApp with Android MatterTvServer #33842: 
https://github.com/project-chip/connectedhomeip/issues/33842

**Change summary**
1.	Updated connectedhomeip/examples/tv-casting-app/tv-casting-common/core/CommissionerDeclarationHandler.cpp to only close the commissioning window upon CDC CommissionerDeclaration message with CommissionerPasscode == true or CancelPasscode == true. The client should handle CDC CommissionerDeclaration ErrorCode appropriately. 

**Testing**
Verified and tested locally with the Linux, Android and iOS tv-casting-app example mobile apps, and the Linux tv-app (CastingPlayer). Able to build and commission with the example apps using both the commissionee and commissioner generated passcode flows.

### DIFF
--- a/examples/tv-casting-app/android/App/app/src/main/java/com/matter/casting/ConnectionExampleFragment.java
+++ b/examples/tv-casting-app/android/App/app/src/main/java/com/matter/casting/ConnectionExampleFragment.java
@@ -29,6 +29,7 @@ import android.widget.EditText;
 import android.widget.TextView;
 import androidx.annotation.Nullable;
 import androidx.fragment.app.Fragment;
+import androidx.fragment.app.FragmentActivity;
 import com.R;
 import com.matter.casting.core.CastingPlayer;
 import com.matter.casting.support.CommissionerDeclaration;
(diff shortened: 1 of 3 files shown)


examples/tv-casting-app/tv-casting-common/core/CommissionerDeclarationHandler.cpp
```diff
@@ -38,39 +38,48 @@ CommissionerDeclarationHandler * CommissionerDeclarationHandler::GetInstance()
...
-    // Close the commissioning window. Since we recived a CommissionerDeclaration message from the Commissioner, we know that
-    // commissioning via AccountLogin cluster failed. We will open a new commissioningWindow prior to sending the next
-    // IdentificationDeclaration Message to the Commissioner.
-    chip::Server::GetInstance().GetCommissioningWindowManager().CloseCommissioningWindow();
-    support::ChipDeviceEventHandler::SetUdcStatus(false);
+    ChipLogProgress(AppServer, "CommissionerDeclarationHandler::OnCommissionerDeclarationMessage()");
+    CastingPlayer * targetCastingPlayer = CastingPlayer::GetTargetCastingPlayer();
+    if (targetCastingPlayer == nullptr)
+    {
+        ChipLogError(AppServer,
```
andy31415, line 48: Why are we not aborting when we find this kind of error? What are codepaths that could cause this?

Otherwise would it be not clearer to return on this error and not follow through with the rest? This would allow us to assume we have a target casting player below and not have extra checks or potential for future code to try to access it.

andy31415: I wonder if this function should be split somehow. We seem to tolerate invalid states and log errors and do several things:

- targetCastingPlayer stop connecting (tolerates casting player being nullptr)
- Commissioning window closing (gated by passcode)
- mCmmissionerDeclarationCallback_ (tolerates it being nullptr)

Overall it seems brittle in terms of tolerating everything not set and I am unclear why we moved the targetCastingPlayer with its nullcheck far away from the usage - this seems to make it more likely to mistake it for being non-null.

andy31415 (dismissed):
Request changes: moving of targetCastingPlayer to the top of the function when we tolerate non-null seems to be more brittle.

examples/tv-casting-app/tv-casting-common/core/CommissionerDeclarationHandler.cpp
pgregorr-amazon: Hey Andrei.

Moving the targetCastingPlayer back to where it it used in `if (cd.GetCancelPasscode())`.

I moved `CastingPlayer * targetCastingPlayer = CastingPlayer::GetTargetCastingPlayer();` and `targetCastingPlayer == nullptr` check up, since I had some additional code (removed from PR)  in `if (cd.GetCommissionerPasscode())` to check if the targetCastingPlayer is in the connecting state and with the CastingPlayer/Commissioner-Generated feature. I removed that additional check but then forgot to move targetCastingPlayer back down into `if (cd.GetCancelPasscode())` where it is used. 

This additional state check is not needed since per the Matter spec section 7.5. UDC with Commissioner-generated Passcode, Step # 9:

9. Commissioner sends CommissionerDeclaration with PasscodeDialogDisplayed and CommissionerPasscode set to true. If a QR code is also displayed, then QRCodeDisplayed is also set to true.

https://github.com/CHIP-Specifications/connectedhomeip-spec/blob/e02a0dfcee3284f6f4b30321ffc7202290ab5bb5/src/rendezvous/UserDirectedCommissioning.adoc#75-udc-with-commissioner-generated-passcode

So we can assume we're in UDC with Commissioner-generated Passcode given the CommissionerPasscode flag.

The spec also calls out PasscodeDialogDisplayed in Step # 9 which doesn't make sense.  PasscodeDialogDisplayed is only used for the "Coordinate PIN Dialogs" feature. I think this step needs to be corrected. I have it in my to do's to submit a correction request for this. 

We also don't abort when targetCastingPlayer is null since we still want to pass the CommissionerDeclaration to the client if the Client provided the mCmmissionerDeclarationCallback_

Per the Matter User Directed Commissioning (UDC) spec: "There is no state or session associated with UDC protocol messages, allowing the recipient to choose to support messages received from different sources at the same time, or choose to ignore messages received while the processing of an earlier message has not completed."

We're opting to pass these messages to the Client for them to decide how to proceed based on the CommissionerDeclaration error code (there are 16 possible error codes).

sharadb-amazon (approved):

chrisdecenzo (approved):

examples/tv-casting-app/android/App/app/src/main/java/com/matter/casting/ConnectionExampleFragment.java
```diff
@@ -208,36 +209,41 @@ public void handle(CommissionerDeclaration cd) {
...
 
-                      getActivity()
-                          .runOnUiThread(
-                              () -> {
-                                connectionFragmentStatusTextView.setText(
-                                    "CommissionerDeclaration message received from Casting Player: \n\n");
-                                if (cd.getCommissionerPasscode()) {
+                      FragmentActivity activity = getActivity();
+                      // Prevent possible NullPointerException. This callback could be called when
+                      // this Fragment is not attached to its host activity or when the fragment's
```
yunhanw-google, line 214: lowercase fragment

yunhanw-google (approved):

PR #35985, opened 2024-10-09
Moving our XPC Checkin to be a general register

woody-apple:
We have too many XPC messages for unregister/register/checkin, let's consolidate them


woody-apple:
Note: I'm not removing the older ones yet, so things don't break, but I've moved things to optional. That will be a followup.

anush-apple (approved):

PR #35997, opened 2024-10-09
[Android] Fix crash when handling jStayActiveMsec

yunhanw-google:
-- jStayActiveMsec is long in java side, and need to be converted via LongToPrimitive
-- locally validated using android chip-tool via commission flow.

### DIFF
--- a/src/controller/java/AndroidDeviceControllerWrapper.cpp
+++ b/src/controller/java/AndroidDeviceControllerWrapper.cpp
@@ -519,9 +519,15 @@ CHIP_ERROR AndroidDeviceControllerWrapper::ApplyICDRegistrationInfo(chip::Contro
                                                         "()Ljava/lang/Long;", &getICDStayActiveDurationMsecMethod);
     ReturnErrorOnFailure(err);
     jobject jStayActiveMsec = env->CallObjectMethod(icdRegistrationInfo, getICDStayActiveDurationMsecMethod);
-    if (jStayActiveMsec != 0)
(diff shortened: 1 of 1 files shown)


src/controller/java/AndroidDeviceControllerWrapper.cpp
```diff
@@ -521,7 +521,13 @@ CHIP_ERROR AndroidDeviceControllerWrapper::ApplyICDRegistrationInfo(chip::Contro
     jobject jStayActiveMsec = env->CallObjectMethod(icdRegistrationInfo, getICDStayActiveDurationMsecMethod);
     if (jStayActiveMsec != 0)
     {
-        params.SetICDStayActiveDurationMsec(chip::JniReferences::GetInstance().IntegerToPrimitive(jStayActiveMsec));
+        jlong stayActiveMsec = chip::JniReferences::GetInstance().LongToPrimitive(jStayActiveMsec);
+        if (!chip::CanCastTo<uint32_t>(stayActiveMsec))
+        {
+            ChipLogError(Controller, "Failed to process stayActiveMsec in %s.", __func__);
```
andy31415, line 527: We could add details stating that this is not a valid 32-bit integer.

andy31415 (approved):

src/controller/java/AndroidDeviceControllerWrapper.cpp
yunhanw-google: updated, thanks

chulspro (approved):

PR #36000, opened 2024-10-09
Add logic to stop full CI if some fast and required checks fail

andy31415:
Added a scheduled job that goes through PRs and cancelles active jobs if any of the "fast" jobs are failing (checking for ZAP, Restyle, lint and spellcheck).

### Changes

- updated workflow and script to go through all active jobs instead of individual job selection

### Also tried but did not work:

Tried to auto-cancel (direct of by starting a workflow) by using a `if: failure()` as a job step, but it turns out that did not work because:
  - github.token does not have write rights on workflows (because from a fork)
  - PRs created from forks have no access to secrets
  
As a result, I switched this to a scheduled workflow, hoping to get more rights. Tested the script locally only for now, it runs in abou 30 seconds (plus the time needed for python install, from what I could see that one is minimal)


scripts/tools/cancel_workflows_for_pr.py
```diff
@@ -32,28 +37,93 @@
...
 class Canceller:
-    def __init__(self, token):
+    def __init__(self, token: str, dry_run: bool):
         self.api = Github(token)
         self.repo = self.api.get_repo(REPOSITORY)
+        self.dry_run = dry_run
+
+    def check_all_pending_prs(self, max_pr_age_days, required_runs):
+        for pr in self.repo.get_pulls(
+            state="open", sort="created_at", direction="desc"
```
tehampson, line 47: I am assuming you already looked into this but can we sort based on last modified.

Essentially if I am pushing an update to a PR I first started working on a week ago it seems like we will no longer cancel the pending things

.github/workflows/cancel_workflows_for_pr.yaml
```diff
@@ -12,22 +12,15 @@
...
-        pull_request_id:
-          description: 'PR number to consider'
-          required: true
-          type: number
-        commit_sha:
-          description: 'Cancel runs for this specific SHA'
-          required: true
-          type: string
+    schedule:
+        - cron: "*/10 * * * *"
```
tehampson, line 19: possibly a dumb question, but this cron job would be running based on what is in master and not a possible uploaded PR for review.

I don't think so, but also want to double check that that someone couldn't maliciously start cancelling jobs by uploading a PR that contains a change to `scripts/tools/cancel_workflows_for_pr.py`

scripts/tools/cancel_workflows_for_pr.py
andy31415: https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests seems to say updated exists. I will change the PR.

.github/workflows/cancel_workflows_for_pr.yaml
andy31415: This will only run on master CI, so changing the python script would only take effect if the PR is approved and merged.

scripts/tools/cancel_workflows_for_pr.py
```diff
@@ -32,28 +37,95 @@
...
 class Canceller:
-    def __init__(self, token):
+    def __init__(self, token: str, dry_run: bool):
         self.api = Github(token)
         self.repo = self.api.get_repo(REPOSITORY)
+        self.dry_run = dry_run
+
+    def check_all_pending_prs(self, max_pr_age_days, required_runs):
+        for pr in self.repo.get_pulls(state="open", sort="updated", direction="desc"):
+            pr_create = (
```
tehampson, line 47: super nit, `pr_create` -> `pr_update_time`

tehampson (approved):

scripts/tools/cancel_workflows_for_pr.py
andy31415: Updated in a followup commit.

arkq (approved):

PR #36004, opened 2024-10-09
Initial implemention of a push av server

fmonniot:
Push AV Server implementation for cameras. This PR is currently a draft to gather early feedback from the camera tiger team. It is not intended to be merged as part of the Matter 1.4 release.

This PR introduce a new tool (`src/tools/push_av_server/server.py`) that offer a web server to test the PUSH AV Transport cluster. The tool also manage TLS certificates for both the server itself as well as the cameras uploading to it. The server can be interacted with via HTTP calls when used as a standalone tool (work on a UI is in progress) or via python functions when used as a module (mostly targeting the test harness).

Note that test harness integration is mostly a work in progress at this point. The added test file in this PR is only intended as a guide (as well as to provide information onto how the tool can be used in that context) to help develop correct APIs and is not intended to be merged as part of this PR.

The web server provide few checks in term of validation for the ingested media. The idea is to have a server permissive enough that record as much information as possible, and then rely on the actual test to validate that the camera is working as it should.

#### Testing

This PR introduce a tool that will be used as part of the camera test harness. Until those tests have been written, the tool can be tested by following the example's instruction in the tool's `README` file.


src/tools/push_av_server/generate_cmaf_content.sh
```diff
@@ -0,0 +1,126 @@
...
+    -content_type "" \
+    -icy 0 \
+    -rw_timeout "$((200 * 1000 * 1000))" \
+    -reconnect 1 \
+    -reconnect_at_eof 1 \
+    -reconnect_on_network_error 1 \
+    -reconnect_on_http_error 4xx,5xx \
+    -reconnect_delay_max 2 \
+    -http_opts $HTTP_OPTS \
+    "$PUBLISHING_ENDPOINT/cmaf/example.str/Switching(video)/video-720p.cmfv" \
```
aronrosenberg, line 62: From the looks of this, this is generating non multiplexed tracks, aka video and audio are in seperate files instead of interleaved m4s / mp4 content.

aronrosenberg: Based on my understanding of the DASH spec we are asking the ecosystem to provide the base URL and the camera then only appends the track name (always fixed since we are a single track) and segment number and file extension onto that to form the complete URL.
Right now the interleaved vs separate tracks is an enum in the spec that only defines interleaved (single track), so we have the ability to change in the future if we want.

src/tools/push_av_server/server.py
```diff
@@ -0,0 +1,582 @@
...
+    # to the server running in the background. This function currently doesn't return.
+    with WorkingDirectory(working_directory) as directory:
+
+        # Sharing state with the various endpoints
+        wd = directory
+
+        dns = "localhost" if dns is None else f"{dns}._http._tcp_.local."
+
+        # Create CA hierarchies (for webserver and devices)
+        device_hierarchy = CAHierarchy(directory.mkdir("certs", "device"), "device")
```
samadDotDev, line 514: Since this function modifies global variables, we'll need a `global` keyword to avoid unintentional local allocation. Note that the reads / accesses are fine without the keyword (python 😒 )
 
```suggestion
        global wd, device_hierarchy

        # Sharing state with the various endpoints
        wd = directory

        dns = "localhost" if dns is None else f"{dns}._http._tcp_.local."

        # Create CA hierarchies (for webserver and devices)
        device_hierarchy = CAHierarchy(directory.mkdir("certs", "device"), "device")
```

src/tools/push_av_server/server.py
```diff
@@ -0,0 +1,582 @@
...
+
+            logging.info(f"CA Hierarchy generated: {self.name}")
+
+    def _save_cert(
+        self,
+        name: str,
+        cert: x509.Certificate,
+        key: Union[CERTIFICATE_PRIVATE_KEY_TYPES, None],
+        bundle_root: bool,
+    ) -> tuple[str, str]:
```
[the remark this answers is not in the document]
fmonniot: I was using python 3.9 as the baseline, but maybe it's not the case? I really have no idea how to find which version is being used. Looking at CI, it's either unspecified or using 3.8, 3.11 or 3.12…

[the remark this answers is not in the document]
fmonniot: I learned that the Test Harness is running on python 3.10.12, so I'm going to assume that version as a baseline for this PR.

fmonniot:
The test plan integration example is currently not up to date, but the CLI example should run just fine nonetheless.

fmonniot:
The test examples should be relatively correct, although they don't run locally for me due to the `chip` module not being available. Probably best to investigate once we have the cluster code generated.

I have added the ability to run multiple servers concurrently (and accessing their CA hierarchies, example in the test file). I have also added an example on how to sign a certificate using the HTTP API.

src/python_testing/push_av_server.py
andy31415, line 1: I am unsure if we should allow symlinks here as we have python_testing scripts generally be copied over.

src/tools/push_av_server/templates/certificates_details.html
andy31415, line 1: Extensions should probably be `jinja2` for clarity.

fmonniot:
The test failure is due to the push av server not being marked as not a test. Once we have resolved the question of where the tool should live, I'll update the exclusion rules as required.

aronrosenberg (approved):

samadDotDev (approved):

chulspro (approved):

PR #36011, opened 2024-10-10
[Electrical Power Measurement Cluster] Returns null for HarmonicPhase…

vivien-apple:
…s and HarmonicCurrents if it has not been measured yet.

#### Problem

The `ElectricalPowerMeasurement::HarmonicPhase`s and `ElectricalPowerMeasurement::HarmonicCurrent` fields are specified as `nullable` in the standard. My understanding is that if no current has been measured, we should return `null` upon reading, rather than an empty array.

That said, I’d appreciate it if someone with deeper knowledge of this part of the spec could review this PR to ensure accuracy.

This PR updates the behavior to return a `null` value when no measurements are available.

### DIFF
--- a/src/app/clusters/electrical-power-measurement-server/electrical-power-measurement-server.cpp
+++ b/src/app/clusters/electrical-power-measurement-server/electrical-power-measurement-server.cpp
@@ -64,6 +64,8 @@ bool Instance::SupportsOptAttr(OptionalAttributes aOptionalAttrs) const
 // AttributeAccessInterface
 CHIP_ERROR Instance::Read(const ConcreteReadAttributePath & aPath, AttributeValueEncoder & aEncoder)
 {
+    CHIP_ERROR err = CHIP_NO_ERROR;
+
     switch (aPath.mAttributeId)
     {
     case FeatureMap::Id:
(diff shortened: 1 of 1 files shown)


woody-apple (approved):

src/app/clusters/electrical-power-measurement-server/electrical-power-measurement-server.cpp
```diff
@@ -188,15 +190,25 @@ CHIP_ERROR Instance::Read(const ConcreteReadAttributePath & aPath, AttributeValu
         VerifyOrReturnError(
             HasFeature(ElectricalPowerMeasurement::Feature::kHarmonics), CHIP_ERROR_UNSUPPORTED_CHIP_FEATURE,
             ChipLogError(Zcl, "Electrical Power Measurement: can not get HarmonicCurrents, feature is not supported"));
-        ReturnErrorOnFailure(
-            aEncoder.EncodeList([this](const auto & encoder) -> CHIP_ERROR { return this->EncodeHarmonicCurrents(encoder); }));
+        err = aEncoder.EncodeList([this](const auto & encoder) -> CHIP_ERROR { return this->EncodeHarmonicCurrents(encoder); });
+        if (CHIP_ERROR_NOT_FOUND == err)
+        {
+            ReturnErrorOnFailure(aEncoder.EncodeNull());
```
bzbarsky-apple, line 196: Null and empty array are supposed to be semantically equivalent, no?  So anyone relying on the difference between the two is broken.

PR #36013, opened 2024-10-10
MCORE-FS-1.4: Wait for dynamic endpoint to appear on TH_FSA

### LINKED ISSUE #35348: TC_MCORE_FS_1_4 Follow-up
### Reproduction steps

Followup items:

- [ ] waiting until commissioned `no-uid-app` shows up on the `PartsList` instead of relying on a log line where the formatting may change
- [x] break out `Subprocess`, `FabricSyncApp` and `AppServer` into a separate file (maybe called fabric_sync_support) to help to compartmentalize the code
- [x] option to search for some "needle" text in the app output in order to start test script instead of a delay option

### Bug prevalence

always

master

### Platform

other

### Platform Version(s)

_No response_

### Anything else?

_No response_

arkq:
### Problem

Fixes #35348

### Changes

- Actively wait for dynamic endpoint instead of relying on the output text

### Testing

CI will verify

### DIFF
--- a/src/python_testing/TC_MCORE_FS_1_4.py
+++ b/src/python_testing/TC_MCORE_FS_1_4.py
@@ -264,10 +264,12 @@ class TC_MCORE_FS_1_4(MatterBaseTest):
             filter=discriminator,
         )
 
+        get_dynamic_endpoint_retries = 60
         th_fsa_bridge_endpoints_new = set(th_fsa_bridge_endpoints)
         # Try to get the dynamic endpoint number for the TH_SERVER_NO_UID on the TH_FSA_BRIDGE.
-        while th_fsa_bridge_endpoints_new == th_fsa_bridge_endpoints:
+        while th_fsa_bridge_endpoints_new == th_fsa_bridge_endpoints and get_dynamic_endpoint_retries > 0:
             await asyncio.sleep(0.5)
+            get_dynamic_endpoint_retries -= 1
             # Get the list of endpoints on the TH_FSA_BRIDGE.
             th_fsa_bridge_endpoints_new.update(await self.read_single_attribute_check_success(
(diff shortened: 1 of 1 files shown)


src/python_testing/TC_MCORE_FS_1_4.py
```diff
@@ -266,16 +264,17 @@ async def test_TC_MCORE_FS_1_4(self):
...
+        # Try to get the dynamic endpoint number for the TH_SERVER_NO_UID on the TH_FSA_BRIDGE.
+        while th_fsa_bridge_endpoints_new == th_fsa_bridge_endpoints:
+            await asyncio.sleep(0.5)
+            # Get the list of endpoints on the TH_FSA_BRIDGE.
+            th_fsa_bridge_endpoints_new.update(await self.read_single_attribute_check_success(
+                cluster=Clusters.Descriptor,
+                attribute=Clusters.Descriptor.Attributes.PartsList,
+                node_id=th_fsa_bridge_th_node_id,
+                endpoint=0,
+            ))
```
tehampson, line 277: nit, do we want to timeout as some point? Previously I think using the `expected_output=` has some default timeout. This does not look like it has a timeout

tehampson (approved):

src/python_testing/TC_MCORE_FS_1_4.py
arkq: I was thinking about that, but in general the entire test has a timeout... if you want I can add something like ~30 seconds (i.e. 100 iterations)

arkq: Added 60 iterations timeout.

saurabhst (approved):

PR #36015, opened 2024-10-10
Matter framework safe conversion for known enum

vivien-apple:
#### Problem

There are conversion issues in `Matter.framework` when building with the `-Wenum-float-conversion` flag enabled. These issues are due to constants that are unlikely to change, so it should be safe to use `static_cast` to address them:

1.  **chip::kMicrosecondsPerSecond**
2. **chip::kMillisecondsPerSecond**
3. **chip::kChipEpochSecondsSinceUnixEpoch**

### DIFF
--- a/src/darwin/Framework/CHIP/MTRBaseDevice.mm
+++ b/src/darwin/Framework/CHIP/MTRBaseDevice.mm
@@ -1894,7 +1894,7 @@ NSTimeInterval MTRTimeIntervalForEventTimestampValue(uint64_t timeValue)
     uint64_t eventTimestampValueSeconds = timeValue / chip::kMillisecondsPerSecond;
     uint64_t eventTimestampValueRemainderMilliseconds = timeValue % chip::kMillisecondsPerSecond;
     NSTimeInterval eventTimestampValueRemainder
-        = NSTimeInterval(eventTimestampValueRemainderMilliseconds) / chip::kMillisecondsPerSecond;
+        = NSTimeInterval(eventTimestampValueRemainderMilliseconds / static_cast<uint64_t>(chip::kMillisecondsPerSecond));
     NSTimeInterval eventTimestampValue = eventTimestampValueSeconds + eventTimestampValueRemainder;
 
     return eventTimestampValue;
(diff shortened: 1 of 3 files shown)


woody-apple (approved):

src/darwin/Framework/CHIP/MTRBaseDevice.mm
```diff
@@ -1894,7 +1894,7 @@ NSTimeInterval MTRTimeIntervalForEventTimestampValue(uint64_t timeValue)
     uint64_t eventTimestampValueSeconds = timeValue / chip::kMillisecondsPerSecond;
     uint64_t eventTimestampValueRemainderMilliseconds = timeValue % chip::kMillisecondsPerSecond;
     NSTimeInterval eventTimestampValueRemainder
-        = NSTimeInterval(eventTimestampValueRemainderMilliseconds) / chip::kMillisecondsPerSecond;
+        = NSTimeInterval(eventTimestampValueRemainderMilliseconds / static_cast<uint64_t>(chip::kMillisecondsPerSecond));
```
bzbarsky-apple, line 1897: This change looks wrong.  `eventTimestampValueRemainderMilliseconds / static_cast<uint64_t>(chip::kMillisecondsPerSecond)` is always 0, no?

What was the actual warning this was trying to fix?

src/darwin/Framework/CHIP/MTRConversion.h
```diff
@@ -38,7 +38,8 @@ AsNumber(chip::Optional<T> optional)
 
 inline NSDate * MatterEpochSecondsAsDate(uint32_t matterEpochSeconds)
 {
-    return [NSDate dateWithTimeIntervalSince1970:(chip::kChipEpochSecondsSinceUnixEpoch + (NSTimeInterval) matterEpochSeconds)];
+    const uint64_t interval = static_cast<uint32_t>(chip::kChipEpochSecondsSinceUnixEpoch) + matterEpochSeconds;
```
bzbarsky-apple, line 41: This change also looks wrong: adding two uint32_t will give you another uint32_t, possibly with overflow.  This should probably cast kChipEpochSecondsSinceUnixEpoch to uint64_t?  Or just cast both to NSTimeInterval.

PR #36018, opened 2024-10-10
python_testing: Add Checks for Command IDs

soares-sergio:
This PR adds checks and unit tests for the command ID ranges defined in Table 79 of the spec.
These checks will be used on the various Python tests, for example, `TC_DeviceBasicComposition.py`.


src/python_testing/global_attribute_ids.py
tcarmelveilleux, line 1: Perhaps this should now be called `global_ids.py` since it's not just attribute IDs

tcarmelveilleux (approved):

soares-sergio:
Moved the changes from here to another PR so they merge together I don't have to sheperd 2 separate ones: https://github.com/project-chip/connectedhomeip/pull/36018

andy31415 (approved):

PR #28375, opened 2023-07-28
[ICD] Add default ICD Behavior to readhandler when establishing a subscription

mkardous-silabs:
#### Description 
PR adds a default behavior for ICDs when establishing a subscription where the ICD tries to set its IdleModeInterval as the MaxInterval of the subscription. 

Default behavior steps :

1. If MinInterval > IdleModeInterval, try to set the MaxInterval to the first interval of IdleModeIntervals above the MinInterval.
2. If the next interval is greater than the MaxIntervalCeiling, use the MaxIntervalCeiling.
3. Otherwise, use IdleModeInterval as MaxInterval

PR #36021, opened 2024-10-10
[ICD] Add subscriber request max interval getter

### LINKED ISSUE #35991: [BUG] Unable to get correct MaxIntervalCeiling in OnSubscriptionRequested for ICD
### Reproduction steps

When using ReadHandler::ApplicationCallback to implement a custom strategy, you will not be able to get the MaxIntervalCeiling in OnSubscriptionRequested for ICD devices. 

We only have GetReportingIntervals API and it get mMaxInterval but is overridden.

https://github.com/project-chip/connectedhomeip/blob/5224cec13d97a8746a8d1a9da70bc16a859d8ee7/src/app/ReadHandler.cpp#L744 https://github.com/project-chip/connectedhomeip/blob/5224cec13d97a8746a8d1a9da70bc16a859d8ee7/src/app/ReadHandler.cpp#L794

### Bug prevalence

100%

5224cec13d97a8746a8d1a9da70bc16a859d8ee7

### Platform

core

### Platform Version(s)

_No response_

### Anything else?

_No response_

mkardous-silabs:
#### Description
 When #28375 was merged to add a default subscription behavior to the subscription establishment, the MaxInterval of the readhandler is changed to the default ICD behavior with the expectation that this behavior can be changed with the OnSubscriptionEstablished callback.
 
The issue is that since we overrule the request MaxInterval, the application API doesn't have a way of knowing the requested max interval from the subscriber to do its decision making. This also impacted the SetMaxInterval function where we were changing the allow maximum which could decrease the spec allowed max interval selection.

PR adds a member that stores the subscriber max interval that can be accessed by the application without impacting the SetMaxInterval function

Fixes #35991

#### Tests
Added a unit test to validate the SetMaxInterval readhandler function

### DIFF
--- a/src/app/ReadHandler.cpp
+++ b/src/app/ReadHandler.cpp
@@ -741,7 +741,9 @@ CHIP_ERROR ReadHandler::ProcessSubscribeRequest(System::PacketBufferHandle && aP
     ReturnErrorOnFailure(err);
 
     ReturnErrorOnFailure(subscribeRequestParser.GetMinIntervalFloorSeconds(&mMinIntervalFloorSeconds));
-    ReturnErrorOnFailure(subscribeRequestParser.GetMaxIntervalCeilingSeconds(&mMaxInterval));
+    ReturnErrorOnFailure(subscribeRequestParser.GetMaxIntervalCeilingSeconds(&mSubscriberRequestedMaxInterval));
+    mMaxInterval = mSubscriberRequestedMaxInterval;
+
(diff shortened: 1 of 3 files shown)


jmartinez-silabs (approved):

yunhanw-google (approved):

fuxiaoming-lumi:
@yunhanw-google @andy31415 I think this commit needs cherry-pick to v1.4-branch.

mkardous-silabs:
@fuxiaoming-lumi Cherry-pick PR is open https://github.com/project-chip/connectedhomeip/pull/36176

PR #36033, opened 2024-10-10
[Fabric-Admin] Refactor to use API methods instead of PushCommand (3/3)

### LINKED ISSUE #34910: [SVE] Switch from string-based cli API to the SDK API using ReadClients.

yufengwangca:

Implement the bridge subscription API.

Fixes https://github.com/project-chip/connectedhomeip/issues/34910

### DIFF
--- a/examples/fabric-admin/device_manager/BridgeSubscription.cpp
+++ b/examples/fabric-admin/device_manager/BridgeSubscription.cpp
@@ -49,6 +49,11 @@ CHIP_ERROR BridgeSubscription::StartSubscription(Controller::DeviceController & 
 {
     assertChipStackLockedByCurrentThread();
 
+    VerifyOrDie(!subscriptionStarted); // Ensure it's not called multiple times.
+
+    // Mark as started
+    subscriptionStarted = true;
+
     mEndpointId = endpointId;
 
     CHIP_ERROR err = controller.GetConnectedDevice(nodeId, &mOnDeviceConnectedCallback, &mOnDeviceConnectionFailureCallback);
@@ -97,20 +102,23 @@ void BridgeSubscription::OnEventData(const app::EventHeader & eventHeader, TLV::
     DeviceMgr().HandleEventData(eventHeader, *data);
 }
 
-void BridgeSubscription::OnReportEnd()
-{
-    // We will call mOnDoneCallback in OnDone.
-}
-
 void BridgeSubscription::OnError(CHIP_ERROR error)
 {
     ChipLogProgress(NotSpecified, "Error on remote fabric sync bridge subscription: %" CHIP_ERROR_FORMAT, error.Format());
+
+    // Reset the subscription state to allow retry
+    subscriptionStarted = false;
 }
 
 void BridgeSubscription::OnDone(ReadClient * apReadClient)
 {
     mClient.reset();
     ChipLogProgress(NotSpecified, "The remote fabric sync bridge subscription is terminated");
+
+    // Reset the subscription state to allow retry
+    subscriptionStarted = false;
+
+    // TODO:(#36092) Fabric-Admin should attempt to re-subscribe when the subscription to the remote bridge is terminated.
 }
 
 void BridgeSubscription::OnDeviceConnected(Messaging::ExchangeManager & exchangeMgr, const SessionHandle & sessionHandle)
--- a/examples/fabric-admin/device_manager/BridgeSubscription.cpp
+++ b/examples/fabric-admin/device_manager/BridgeSubscription.cpp
@@ -140,6 +140,7 @@ void BridgeSubscription::OnDeviceConnected(Messaging::ExchangeManager & exchange
     readParams.mEventPathParamsListSize     = 1;
     readParams.mMinIntervalFloorSeconds     = kSubscribeMinInterval;
     readParams.mMaxIntervalCeilingSeconds   = kSubscribeMaxInterval;
+    readParams.mKeepSubscriptions           = true;
 
     CHIP_ERROR err = mClient->SendRequest(readParams);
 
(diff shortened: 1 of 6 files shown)


examples/fabric-admin/device_manager/BridgeSubscription.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+    ReadPrepareParams readParams(sessionHandle);
+
+    readParams.mpAttributePathParamsList    = readPaths;
+    readParams.mAttributePathParamsListSize = 1;
+    readParams.mpEventPathParamsList        = eventPaths;
+    readParams.mEventPathParamsListSize     = 1;
+    readParams.mMinIntervalFloorSeconds     = kSubscribeMinInterval;
+    readParams.mMaxIntervalCeilingSeconds   = kSubscribeMaxInterval;
+
```
tehampson, line 138: Likely this isn't something that can be done in this PR, but should we filtering the event number only looking for something new. Otherwise you might get old events and start trying to send a CommissionNode (that will fail due to your state machine.

yufengwangca: The event contains the RequestId, which SHALL match the RequestId provided to RequestCommissioningApproval and the ClientNodeId shall match the NodeId of the client which sent the RequestCommissioningApproval command. So we should not process out of date event.

examples/fabric-admin/device_manager/BridgeSubscription.cpp
```diff
@@ -0,0 +1,153 @@
...
+}
+
+} // namespace
+
+BridgeSubscription::BridgeSubscription() :
+    mOnDeviceConnectedCallback(OnDeviceConnectedWrapper, this),
+    mOnDeviceConnectionFailureCallback(OnDeviceConnectionFailureWrapper, this)
+{}
+
+CHIP_ERROR BridgeSubscription::StartSubscription(Controller::DeviceController & controller, NodeId nodeId, EndpointId endpointId)
```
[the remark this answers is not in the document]
yufengwangca: Acked

examples/fabric-admin/device_manager/BridgeSubscription.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+void BridgeSubscription::OnError(CHIP_ERROR error)
+{
+    ChipLogProgress(NotSpecified, "Error on remote fabric sync bridge subscription: %" CHIP_ERROR_FORMAT, error.Format());
+}
+
+void BridgeSubscription::OnDone(ReadClient * apReadClient)
+{
+    mClient.reset();
+    ChipLogProgress(NotSpecified, "The remote fabric sync bridge subscription is terminated");
```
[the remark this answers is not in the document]
yufengwangca: this should be done in the upper layer.  https://github.com/project-chip/connectedhomeip/issues/36092

examples/fabric-admin/device_manager/BridgeSubscription.cpp
```diff
@@ -0,0 +1,153 @@
...
+        ChipLogError(NotSpecified, "Response Failure: No Data");
+        return;
+    }
+
+    DeviceMgr().HandleEventData(eventHeader, *data);
+}
+
+void BridgeSubscription::OnReportEnd()
+{
+    // We will call mOnDoneCallback in OnDone.
```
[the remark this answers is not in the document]
yufengwangca: Copied from other place, removed

examples/fabric-admin/device_manager/BridgeSubscription.cpp
tehampson: But there is a chance an old one could match. It is unlikely, but why open ourselves up if we don't need to.

examples/fabric-admin/device_manager/BridgeSubscription.cpp
```diff
@@ -97,20 +102,23 @@ void BridgeSubscription::OnEventData(const app::EventHeader & eventHeader, TLV::
...
-{
-    // We will call mOnDoneCallback in OnDone.
-}
-
 void BridgeSubscription::OnError(CHIP_ERROR error)
 {
     ChipLogProgress(NotSpecified, "Error on remote fabric sync bridge subscription: %" CHIP_ERROR_FORMAT, error.Format());
+
+    // Reset the subscription state to allow retry
+    subscriptionStarted = false;
```
tehampson, line 110: nit, I think you should wait until OnDone to set this to false

tehampson (approved):

examples/fabric-admin/device_manager/BridgeSubscription.cpp
yufengwangca: We already did, since OnError is always followed by OnDone, so we can remove subscription reset from this method.

examples/fabric-admin/device_manager/BridgeSubscription.cpp
tehampson: Fair what we create in the admin is random (this is assuming nothing is wrong with random generator, which is a relatively safe assumption). I still wonder why you think processing events that we know might be old even if they fail is a good things and not worth considering.

Please note I am okay with not doing this in this PR or even a follow up. I am just wondering if we have considered it and why we are okay with having it process events that we can potentially know are old and not relevant.

The trade-off of developer time is an acceptable reason to not do it, just I want to make sure we at the very least think about what we are trading off. I appreciate the discussion

yufengwangca: I set this event to 'eventPaths[0].mIsUrgentEvent = true;' so that the server won't buffer this event to prevent we retrieve the out of date event.

andy31415:
fast track of 1-week old PR

PR #36035, opened 2024-10-11
Revert "esp32-fix-empty-logs(#35939)"

shubhamdp:
Reverts project-chip/connectedhomeip#35965

Please see https://github.com/project-chip/connectedhomeip/pull/35965#issuecomment-2404123825

### DIFF
--- a/src/platform/ESP32/Logging.cpp
+++ b/src/platform/ESP32/Logging.cpp
@@ -25,6 +25,9 @@ void LogV(const char * module, uint8_t category, const char * msg, va_list v)
     snprintf(tag, sizeof(tag), "chip[%s]", module);
     tag[sizeof(tag) - 1] = 0;
 
+    // We intentionally added the printf statements to ensure we could apply colors to logs redirected to the console.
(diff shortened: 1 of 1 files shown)


src/platform/ESP32/Logging.cpp
```diff
@@ -29,28 +29,28 @@ void LogV(const char * module, uint8_t category, const char * msg, va_list v)
     {
     case kLogCategory_Error: {
         {
-            esp_log_write(ESP_LOG_ERROR, tag, LOG_COLOR_E "E (%" PRIu32 ") %s: ", esp_log_timestamp(), tag);
+            printf(LOG_COLOR_E "E (%" PRIu32 ") %s: ", esp_log_timestamp(), tag);
```
arkq, line 32: Maybe add some comment why printf is uded here, so it is clear that esp_log will break logging.

arkq (approved):

src/platform/ESP32/Logging.cpp
shubhamdp: Thanks @arkq , makes sense, added the comment.

andy31415 (approved):

PR #36041, opened 2024-10-11
Fix dependency requirement for CI cancel

andy31415:
https://github.com/project-chip/connectedhomeip/actions/runs/11293251215/job/31410960679 failed with

```
  File "/home/runner/work/connectedhomeip/connectedhomeip/scripts/tools/cancel_workflows_for_pr.py", line 25, in <module>
    from dateutil.tz import tzlocal
ModuleNotFoundError: No module named 'dateutil'
```

So add this to PIP install


andy31415:
Hotfix for helper script: cannot be validated by CI and minimal dep change.

PR #36042, opened 2024-10-11
Switch `DataModel::Provider` interface to default enabled (except on linux where it stays as check)

andy31415:
This will switch from using ember-compatibility-functions.cpp to using the codegen datamodel provider.

PR should show FLASH and RAM sideffect as well as if any compilation issues may occur.

#### Changes

- enable data model interface usage by default (except linux where the state remains "check" and runs both in parallel)
- Fix build issues and bugs (some split in other PRs). Specifically I have found:
   - k32w0 is very close to full flash, so enabled progress-level logs and above only (no detail)
   - fixed bug in caching being maintained across dynamic endpoint enable/disable (if the same endpoint id is re-used, the same path may reference freed clusters)

### Size notes

- The PR has a non-trivial size increase, generally due to making iteration over attributes & commands configurable rather than fixed-index iteration. 
- It started off at around 1% of extra flash, after which several PRs were created to address sizes and reduce this, including:
  - #36249
  - #36247
  - #36246
  - #36229
  - #36172 
  - #36273
- The unmerged above PRs are included in this change to evaluate the current size overhead.
- Additional size reductions (globally) in PRs:
  - #36253
  - #36251 offers some minute size decreases (however in the data_model implementation those savings seem to zero out)
  - #36286
  - #36288


andy31415:
chip-k32w0x-light-example does not fit in flash anymore ... that is not good at all.

Looking through the rest of the failures and will have separate PRs to fix them.

andy31415:
Converted to draft: we need to figure out all build failures and even then figure out how to manage flash increases.

src/darwin/Framework/CHIP/ServerEndpoint/MTRIMDispatch.mm
```diff
@@ -50,6 +50,13 @@ void emberAfClusterInitCallback(EndpointId endpoint, ClusterId clusterId)
     return Protocols::InteractionModel::Status::UnsupportedAttribute;
 }
 
+Protocols::InteractionModel::Status emberAfWriteAttribute(const ConcreteAttributePath & path, const EmberAfWriteDataInput & input)
+{
+    // All of our attributes are handled via AttributeAccessInterface, so this
```
bzbarsky-apple, line 55: Should have assertChipStackLockedByCurrentThread(), no?

andy31415:
#36104 contains the split out change for `metadata structure generation`

.github/workflows/examples-nxp.yaml
```diff
@@ -70,7 +70,7 @@ jobs:
               run: |
                   scripts/run_in_build_env.sh "\
                       ./scripts/build/build_examples.py \
-                      --target nxp-k32w0-freertos-lighting-factory \
+                      --target nxp-k32w0-freertos-lighting-factory-log-progress \
```
tehampson, line 73: Is this supposed to be in this PR?

tehampson (approved):

.github/workflows/examples-nxp.yaml
andy31415: Yes, k32w0 is low on flash and this makes CI fit by disabling detail log strings. Without this, this PR flash increases do not pass CI.

src/app/dynamic_server/mock_includes/zap-generated/endpoint_config.h
```diff
@@ -21,3 +21,5 @@
 // Some overrides to see that this is not used as a real "endpoint_config" that contains data
 #define GENERATED_DEFAULTS #error "Not a real codegen. This is a temporary include for dynamic-overrides only"
 #define GENERATED_CLUSTERS #error "Not a real codegen. This is a temporary include for dynamic-overrides only"
+
+#define ATTRIBUTE_LARGEST (1003)
```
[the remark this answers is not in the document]
andy31415: src/app/util/ember-io-storage.cpp allocates size for attributes based on this. I will add a comment.

.github/workflows/examples-nxp.yaml
andy31415: Added a comment explaining.

bzbarsky-apple, line 73: What is the current flash increase from turning this on?  What steps are we taking to bring it down?

Ideally, we would bring it down _before_ enabling it on constrained platforms.

src/app/BUILD.gn
```diff
@@ -225,14 +225,17 @@ static_library("interaction-model") {
 
   deps = [ "${chip_root}/src/app:events" ]
 
+  # Temporary dependency: codegen data provider instance should be provided
+  # by the application
```
bzbarsky-apple, line 229: Where is the issue tracking this?  It should be linked here.

src/app/BUILD.gn
```diff
@@ -294,13 +297,23 @@ static_library("interaction-model") {
       "clusters/ota-provider/ota-provider.cpp",
       "dynamic_server/AccessControl.cpp",
       "dynamic_server/AccessControl.h",
+    ]
+
+    # DynamicDispatcher is actually an ember-override that takes over
+    # util/attribute-storage.cpp and util/attribute-table.cpp functions.
+    #
+    # We likely should formalize and change this with a proper DataModel::Provider that
```
bzbarsky-apple, line 305: DynamicDispatcher should be switched to use a DataModel::Provider, indeed.  The problem is right now it has no way to provide one, right?

Need tracking issues for:

1. A way to inject a data model provider during stack bringup instead of defaulting in the codegen one.  For DynamicDispatcher, having this just on the controller factory would be enough.
2. Making DynamicDispatcher actually use a provider.

Both should be linked from here, presumably.

src/app/clusters/microwave-oven-control-server/microwave-oven-control-server.cpp
```diff
@@ -248,11 +248,16 @@ void Instance::HandleSetCookingParameters(HandlerContext & ctx, const Commands::
 
     if (startAfterSetting.HasValue())
     {
+        ConcreteCommandPath commandPath(mEndpointId, OperationalState::Id, OperationalState::Commands::Start::Id);
+
+#if CHIP_CONFIG_USE_EMBER_DATA_MODEL
```
bzbarsky-apple, line 253: This seems like we should just have some sort of "server command exists" API on the IM engine, no?

src/app/dynamic_server/DynamicDispatcher.cpp
```diff
@@ -381,3 +385,62 @@ const EmberAfCluster * emberAfFindServerCluster(EndpointId endpoint, ClusterId c
 
     return nullptr;
 }
+
+unsigned emberAfMetadataStructureGeneration()
+{
+    // DynamicServer at this point hardcodes a single OTA provider cluster
```
bzbarsky-apple, line 391: What is DynamicServer?

src/darwin/Framework/Matter.xcodeproj/project.pbxproj
bzbarsky-apple, line 1: Yeah, all this stuff desperately needs to go away and we need to have a way to inject a provider.  But for now, I guess it will have to do.

bzbarsky-apple (changes requested):
I meant to request changes.  I don't think the flash increases here are acceptable, some of the avenues we discussed for reducing those involve changing the provider APIs, and I don't think we want to enable and start using this while we are still figuring the API out.

src/app/BUILD.gn
andy31415: I cannot do this in one PR: existing code assumes embed is built in and everything evolves around that. My plan would be that every app would have to set a provider before continuing and the default provider (if any) should just be a noop/deny everything.

Before doing that though we have to change every application to set a provider and for controller sides we have to figure out where such provider is set (probably in some sort of platform init call).

.github/workflows/examples-nxp.yaml
andy31415: started with #36172

In the end the extra code is to support interfaces that are not `index in a contiguous array allocated in ram`.

andy31415: #36229 is a followup, saves another 1K of flash.

src/app/BUILD.gn
andy31415: Created #36230 for a generic coverage of things. I will split it up when starting to work on it - for now just keeping the list of checkmarks of todo items.

src/app/BUILD.gn
andy31415: #36230 is the generic coverage for these updates.

.github/workflows/examples-nxp.yaml
andy31415: Resolving as I started a list of flash savings PRs in the summary of this PR.

bzbarsky-apple (approved):

PR #36044, opened 2024-10-11
[darwin-framework-tool] Add an option to use per-controller storage i…

vivien-apple:
…nstead of a global shared storage

#### Problem

Currently, `darwin-framework-tool` starts with shared storage for controllers. It would be useful to have a mechanism to start with `per-controller` storage, as this triggers different code paths.

This PR introduces this capability by adding a new `--commissioner-shared-storage` optional argument to `darwin-framework-too`l. The default is set to **true**, preserving existing behavior.


woody-apple (approved):

examples/darwin-framework-tool/commands/common/CHIPCommandBridge.h
```diff
@@ -41,6 +41,8 @@ class CHIPCommandBridge : public Command {
             "Sets the commissioner node ID of the given "
             "commissioner-name. Interactive mode will only set a single commissioner on the inital command. "
             "The commissioner node ID will be persisted until a different one is specified.");
+        AddArgument("commissioner-shared-storage", 0, 1, &mCommissionerSharedStorage,
+            "Use a shared storage instance instead of individual storage for each commissioner. Defaults is true.");
```
bzbarsky-apple, line 45:
```suggestion
            "Use a shared storage instance instead of individual storage for each commissioner. Default is true.");
```

examples/darwin-framework-tool/commands/common/CHIPCommandBridge.h
```diff
@@ -130,6 +133,8 @@ class CHIPCommandBridge : public Command {
     void StopWaiting();
 
     CHIP_ERROR MaybeSetUpStack();
+    CHIP_ERROR SetUpStackWithSharedStorage(NSArray<NSData *> * productAttestationAuthorityCertificates);
+    CHIP_ERROR SetUpStackWithControllerStorage(NSArray<NSData *> * productAttestationAuthorityCertificates);
```
bzbarsky-apple, line 137:
```suggestion
    CHIP_ERROR SetUpStackWithPerControllerStorage(NSArray<NSData *> * productAttestationAuthorityCertificates);
```

examples/darwin-framework-tool/commands/common/CHIPCommandBridge.h
bzbarsky-apple, line 45: Followup: After this lands, we should consider changing the default to false.

examples/darwin-framework-tool/commands/common/CHIPCommandBridge.mm
```diff
@@ -111,61 +113,111 @@
...
     mOTADelegate = [[OTAProviderDelegate alloc] init];
+    storage = [[CHIPToolPersistentStorageDelegate alloc] init];
 
-    auto factory = [MTRDeviceControllerFactory sharedInstance];
-    if (factory == nil) {
-        ChipLogError(chipTool, "Controller factory is nil");
-        return CHIP_ERROR_INTERNAL;
+    NSError * error;
+    __auto_type * certificateIssuer = [CertificateIssuer sharedInstance];
+    [certificateIssuer startWithStorage:storage error:error];
```
bzbarsky-apple, line 123:
```suggestion
    [certificateIssuer startWithStorage:storage error:&error];
```

examples/darwin-framework-tool/commands/common/CHIPCommandBridge.mm
```diff
@@ -111,61 +113,111 @@
...
+        __auto_type * vendorId = @(mCommissionerVendorId.ValueOr(chip::VendorId::TestVendor1));
+        __auto_type * fabricId = @(i + 1);
+        __auto_type * nodeId = @(chip::kTestControllerNodeId);
+
+        if (commissionerName.compare(identities[i]) == 0 && mCommissionerNodeId.HasValue()) {
+            nodeId = @(mCommissionerNodeId.Value());
+        }
+
+        NSError * error;
+        __auto_type * operational = [certificateIssuer issueOperationalCertificateForNodeID:nodeId fabricID:fabricId error:error];
```
bzbarsky-apple, line 158:
```suggestion
        __auto_type * operational = [certificateIssuer issueOperationalCertificateForNodeID:nodeId fabricID:fabricId error:&error];
```

examples/darwin-framework-tool/commands/common/CHIPCommandBridge.mm
```diff
@@ -111,61 +113,111 @@
...
+                                                                                                        vendorID:vendorId
+                                                                                              operationalKeypair:certificateIssuer.signingKey
+                                                                                          operationalCertificate:operational
+                                                                                         intermediateCertificate:nil
+                                                                                                 rootCertificate:certificateIssuer.rootCertificate];
+        [params setOperationalCertificateIssuer:certificateIssuer queue:controllerStorageQueue];
+        params.productAttestationAuthorityCertificates = productAttestationAuthorityCertificates;
+
+        __auto_type * controller = [[MTRDeviceController alloc] initWithParameters:params error:&error];
+        VerifyOrReturnError(nil != controller, MTRErrorToCHIPErrorCode(error), ChipLogError(chipTool, "Controller startup failure."));
```
bzbarsky-apple, line 175:
```suggestion
        VerifyOrReturnError(nil != controller, MTRErrorToCHIPErrorCode(error), ChipLogError(chipTool, "Controller startup failure: %@", error));
```

examples/darwin-framework-tool/commands/common/CHIPCommandBridge.mm
```diff
@@ -111,61 +113,111 @@
...
+                                                                           nocSigner:certificateIssuer.signingKey];
         if (commissionerName.compare(identities[i]) == 0 && mCommissionerNodeId.HasValue()) {
-            controllerParams.nodeId = @(mCommissionerNodeId.Value());
+            params.nodeId = @(mCommissionerNodeId.Value());
         }
-        // We're not sure whether we're creating a new fabric or using an
-        // existing one, so just try both.
-        auto controller = [factory createControllerOnExistingFabric:controllerParams error:&error];
+
+        params.nodeId = nodeId;
```
bzbarsky-apple, line 211: This looks wrong; it overrides mCommissionerNodeId above, no?

examples/darwin-framework-tool/commands/common/CHIPCommandBridge.mm
```diff
@@ -111,61 +113,111 @@
...
-            controller = [factory createControllerOnNewFabric:controllerParams error:&error];
+            params.vendorID = @(mCommissionerVendorId.ValueOr(chip::VendorId::TestVendor1));
+            controller = [factory createControllerOnNewFabric:params error:&error];
         }
-        if (controller == nil) {
-            ChipLogError(chipTool, "Controller startup failure.");
-            return MTRErrorToCHIPErrorCode(error);
-        }
-
+        VerifyOrReturnError(nil != controller, MTRErrorToCHIPErrorCode(error), ChipLogError(chipTool, "Controller startup failure."));
```
bzbarsky-apple, line 220:
```suggestion
        VerifyOrReturnError(nil != controller, MTRErrorToCHIPErrorCode(error), ChipLogError(chipTool, "Controller startup failure: %@", error));
```

examples/darwin-framework-tool/commands/common/CertificateIssuer.mm
```diff
@@ -0,0 +1,157 @@
...
+    _signingKey = nil;
+    _fabricID = nil;
+    _nextNodeID = nil;
+    _shouldSkipAttestationCertificateValidation = NO;
+
+    return self;
+}
+
+- (void)startWithStorage:(id<MTRStorage>)storage
+                   error:(NSError * _Nullable)error
```
bzbarsky-apple, line 60: That's not a type that will let you return errors to the caller...

examples/darwin-framework-tool/commands/common/CertificateIssuer.mm
```diff
@@ -0,0 +1,157 @@
...
+    _signingKey = signingKey;
+    _ipk = [signingKey getIPK];
+}
+
+- (MTRCertificateDERBytes _Nullable)issueOperationalCertificateForNodeID:(NSNumber *)nodeID
+                                                                fabricID:(NSNumber *)fabricID
+                                                                   error:(NSError * _Nullable)error
+{
+    __auto_type * publicKey = _signingKey.publicKey;
+    return [self issueOperationalCertificateForNodeID:nodeID fabricID:fabricID publicKey:publicKey error:error];
```
bzbarsky-apple, line 86: This is using the private key of our signing cert as the private key of our NOC?  That seems a little odd...

examples/darwin-framework-tool/commands/common/CHIPCommandBridge.mm
```diff
@@ -111,61 +113,111 @@
...
+        __auto_type * operational = [certificateIssuer issueOperationalCertificateForNodeID:nodeId fabricID:fabricId error:error];
+
+        __auto_type * controllerStorage = [[ControllerStorage alloc] initWithControllerID:controllerId];
+        __auto_type * controllerStorageQueue = dispatch_queue_create("com.chip.storage", DISPATCH_QUEUE_SERIAL_WITH_AUTORELEASE_POOL);
+        __auto_type * params = [[MTRDeviceControllerExternalCertificateParameters alloc] initWithStorageDelegate:controllerStorage
+                                                                                            storageDelegateQueue:controllerStorageQueue
+                                                                                                uniqueIdentifier:controllerId
+                                                                                                             ipk:certificateIssuer.ipk
+                                                                                                        vendorID:vendorId
+                                                                                              operationalKeypair:certificateIssuer.signingKey
```
bzbarsky-apple, line 167: This is using the root key as the operational key.  That seems very odd.

examples/darwin-framework-tool/commands/common/ControllerStorage.mm
```diff
@@ -0,0 +1,132 @@
...
+     securityLevel:(MTRStorageSecurityLevel)securityLevel
+       sharingType:(MTRStorageSharingType)sharingType
+{
+#ifdef LOG_DEBUG_CONTROLLER_STORAGE
+    ChipLogError(chipTool, "Controller(%@) Storage - Set value for %@", controller, key);
+#endif // LOG_DEBUG_CONTROLLER_STORAGE
+    NSError * error;
+    NSData * data = [NSKeyedArchiver archivedDataWithRootObject:value requiringSecureCoding:YES error:&error];
+
+    __auto_type * controllerKey = [NSString stringWithFormat:@"%@/%@", _controllerID, key];
```
bzbarsky-apple, line 74: We should factor this key creation operation into a helper, since it _has_ to be the same in all three places, right?

examples/darwin-framework-tool/commands/common/ControllerStorage.mm
```diff
@@ -0,0 +1,132 @@
...
+- (BOOL)controller:(MTRDeviceController *)controller storeValues:(NSDictionary<NSString *, id<NSSecureCoding>> *)values securityLevel:(MTRStorageSecurityLevel)securityLevel sharingType:(MTRStorageSharingType)sharingType
+{
+#ifdef LOG_DEBUG_CONTROLLER_STORAGE
+    ChipLogError(chipTool, "Controller(%@) Storage - store values", controller);
+#endif // LOG_DEBUG_CONTROLLER_STORAGE
+
+    for (NSString * controllerKey in values) {
+        if (![controllerKey hasPrefix:[_controllerID UUIDString]]) {
+            continue;
+        }
```
bzbarsky-apple, line 124: I don't follow this part. These are key values coming from the framework; how would it know to prefix them with your uuid?  This does not look right.

bzbarsky-apple (changes requested):

examples/darwin-framework-tool/commands/common/CHIPCommandBridge.mm
vivien-apple: > This looks wrong; it overrides mCommissionerNodeId above, no?
You’re absolutely right. This does override mCommissionerNodeId from above. It’s a leftover from previous changes.

examples/darwin-framework-tool/commands/common/ControllerStorage.mm
vivien-apple: > I don't follow this part. These are key values coming from the framework; how would it know to prefix them with your uuid? This does not look right.
You’re right; this doesn’t make sense. It was an oversight due to copy-pasting. Thanks for catching that!

examples/darwin-framework-tool/commands/common/CHIPCommandBridge.h
vivien-apple: > Followup: After this lands, we should consider changing the default to false.

Agreed. I initially thought this was the desired default but wasn’t entirely sure, so I kept it set to `true`. Let’s handle changing the default to `false` in a follow-up PR as you suggest, focusing specifically on any potential CI issues that may come up.

examples/darwin-framework-tool/commands/common/CertificateIssuer.mm
```diff
@@ -0,0 +1,162 @@
...
+}
+
+- (void)startWithStorage:(id<MTRStorage>)storage
+                   error:(NSError * __autoreleasing *)error
+{
+    __auto_type * signingKey = [[CHIPToolKeypair alloc] init];
+
+    __auto_type err = [signingKey createOrLoadKeys:storage];
+    if (CHIP_NO_ERROR != err) {
+        *error = [NSError errorWithDomain:@"Error" code:0 userInfo:@{ @"reason" : @"Error creating or loading keys" }];
```
bzbarsky-apple, line 66: Should probably wrap that in `if (error)`.  The API does not say it's nullable, but error outparams implicitly are, iirc.

examples/darwin-framework-tool/commands/common/CertificateIssuer.mm
```diff
@@ -0,0 +1,162 @@
...
+        return;
+    }
+
+    __auto_type * rootCertificate = [MTRCertificates createRootCertificate:signingKey issuerID:nil fabricID:nil error:error];
+    if (nil == rootCertificate) {
+        *error = [NSError errorWithDomain:@"Error" code:0 userInfo:@{ @"reason" : @"Error creating root certificate" }];
```
bzbarsky-apple, line 72: Again, wrap in nil-check.

examples/darwin-framework-tool/commands/common/CertificateIssuer.mm
```diff
@@ -0,0 +1,162 @@
...
+    _ipk = [signingKey getIPK];
+}
+
+- (id<MTRKeypair>)issueOperationalKeypairWithControllerStorage:(ControllerStorage *)storage error:(NSError * __autoreleasing *)error
+{
+    __auto_type * keypair = [[CHIPToolKeypair alloc] init];
+
+    __auto_type err = [keypair createOrLoadKeys:storage];
+    if (CHIP_NO_ERROR != err) {
+        *error = [NSError errorWithDomain:@"Error" code:0 userInfo:@{ @"reason" : @"Error creating or loading keys" }];
```
bzbarsky-apple, line 87: nil-check.

examples/darwin-framework-tool/commands/common/CertificateIssuer.mm
```diff
@@ -0,0 +1,162 @@
...
+                                                               signingCertificate:self.rootCertificate
+                                                             operationalPublicKey:publicKey
+                                                                         fabricID:fabricID
+                                                                           nodeID:nodeID
+                                                            caseAuthenticatedTags:nil
+                                                                            error:error];
+    if (error && *error != nil) {
+        return nil;
+    }
+
```
bzbarsky-apple, line 158:
```suggestion
```

Returning non-nil but setting error would be a very wrong behavior for the callee....

examples/darwin-framework-tool/commands/common/ControllerStorage.mm
```diff
@@ -0,0 +1,159 @@
...
+    return self.storage[controllerKey];
+}
+
+- (void)storeValue:(NSData *)value forKey:key
+{
+    __auto_type * controllerKey = [self _keyToControllerKey:key];
+    self.storage[controllerKey] = value;
+}
+
+- (NSString *)_keyToControllerKey:(NSString *)key
```
bzbarsky-apple, line 144:
```suggestion
- (NSString *)_keyToControllerScopedKey:(NSString *)key
```

examples/darwin-framework-tool/commands/common/ControllerStorage.mm
```diff
@@ -0,0 +1,159 @@
...
+{
+    return [NSString stringWithFormat:@"%@/%@", _controllerID, key];
+}
+
+- (NSString *)_controllerKeyToKey:(NSString *)controllerKey
```
bzbarsky-apple, line 149:
```suggestion
- (NSString *)_controllerScopedKeyToKey:(NSString *)controllerKey
```

examples/darwin-framework-tool/commands/common/ControllerStorage.h
```diff
@@ -0,0 +1,49 @@
...
+       sharingType:(MTRStorageSharingType)sharingType;
+- (BOOL)controller:(MTRDeviceController *)controller
+    removeValueForKey:(NSString *)key
+        securityLevel:(MTRStorageSecurityLevel)securityLevel
+          sharingType:(MTRStorageSharingType)sharingType;
+- (NSDictionary<NSString *, id<NSSecureCoding>> *)valuesForController:(MTRDeviceController *)controller securityLevel:(MTRStorageSecurityLevel)securityLevel sharingType:(MTRStorageSharingType)sharingType;
+- (BOOL)controller:(MTRDeviceController *)controller storeValues:(NSDictionary<NSString *, id<NSSecureCoding>> *)values securityLevel:(MTRStorageSecurityLevel)securityLevel sharingType:(MTRStorageSharingType)sharingType;
+
+- (NSData *)valueForKey:(NSString *)key;
+- (void)storeValue:(NSData *)value forKey:key;
```
bzbarsky-apple, line 46: Where are these two used?

examples/darwin-framework-tool/commands/common/ControllerStorage.mm
```diff
@@ -0,0 +1,159 @@
...
+{
+    __auto_type * controllerPrefix = [NSString stringWithFormat:@"%@/", [_controllerID UUIDString]];
+    return [controllerKey stringByReplacingOccurrencesOfString:controllerPrefix withString:@""];
+}
+
+- (BOOL)_isControllerKey:(NSString *)controllerKey
+{
+    return [controllerKey hasPrefix:[_controllerID UUIDString]];
```
bzbarsky-apple, line 157:
```suggestion
    return [controllerKey hasPrefix:[NSString stringWithFormat:@"%@/", [_controllerID UUIDString]]];
```

examples/darwin-framework-tool/commands/common/ControllerStorage.mm
```diff
@@ -0,0 +1,159 @@
...
+
+- (NSString *)_keyToControllerKey:(NSString *)key
+{
+    return [NSString stringWithFormat:@"%@/%@", _controllerID, key];
+}
+
+- (NSString *)_controllerKeyToKey:(NSString *)controllerKey
+{
+    __auto_type * controllerPrefix = [NSString stringWithFormat:@"%@/", [_controllerID UUIDString]];
+    return [controllerKey stringByReplacingOccurrencesOfString:controllerPrefix withString:@""];
```
bzbarsky-apple, line 152:
```suggestion
    return [controllerKey substringFromIndex:controllerPrefix.length];
```
seems a bit safer.

examples/darwin-framework-tool/commands/common/ControllerStorage.mm
```diff
@@ -0,0 +1,159 @@
...
+}
+
+- (BOOL)_isControllerKey:(NSString *)controllerKey
```
bzbarsky-apple, line 155:
```suggestion
- (BOOL)_isControllerScopedKey:(NSString *)controllerKey
```

examples/darwin-framework-tool/commands/common/ControllerStorage.mm
bzbarsky-apple, line 157: I do wonder whether we should just have a keyScopingPrefix property that has the `[NSString stringWithFormat:@"%@/", [_controllerID UUIDString]]` bit.

bzbarsky-apple (approved):

examples/darwin-framework-tool/commands/common/CertificateIssuer.mm
vivien-apple: > Should probably wrap that in `if (error)`. The API does not say it's nullable, but error outparams implicitly are, iirc.

I don’t see a strong reason for callers to skip passing an NSError. Instead, I’ve updated the method signature with a _Nonnull annotation to ensure error is always provided.

examples/darwin-framework-tool/commands/common/CertificateIssuer.mm
vivien-apple: See https://github.com/project-chip/connectedhomeip/pull/36044#discussion_r1800012966

examples/darwin-framework-tool/commands/common/CertificateIssuer.mm
vivien-apple: See https://github.com/project-chip/connectedhomeip/pull/36044#discussion_r1800012966

examples/darwin-framework-tool/commands/common/ControllerStorage.mm
vivien-apple: > I do wonder whether we should just have a keyScopingPrefix property that has the `[NSString stringWithFormat:@"%@/", [_controllerID UUIDString]]` bit.

Great suggestion! I’ve updated the PR to include a `keyScopingPrefix` property with `[NSString stringWithFormat:@"%@/", [_controllerID UUIDString]]` as you suggested.

examples/darwin-framework-tool/commands/common/ControllerStorage.h
vivien-apple: These are used in `CHIPToolKeypair::createOrLoadKeys`. I’ve updated the method signature to accept an `id` so it can take a `ControllerStorage *` as well. This allows the operational keys to be stored in the dedicated controller storage instead of the global storage where the root key is stored.

PR #36045, opened 2024-10-11
Decouple PumpConfigurationAndControl from EmberCompat

andy31415:
This uses the ember internal function instead of the ember compatibility functions override. It is somewhat unclear why we have an override that replaces one method with a single method that is identical except arguments.

clusters implementations are often coupled to ember, however they should not be coupled to ember-compatibility-functions.cpp 

We should also figure out if there is some other way updating things as relying that we have a "ember default" is somewhat odd here.


tcarmelveilleux (approved):

andy31415:
fast tracking after 5 days.

PR #36048, opened 2024-10-11
Fix constraint check in thermostat cluster to follow the spec.

bzbarsky-apple:
The cluster was excluding valid SystemMode values.

### DIFF
--- a/src/app/clusters/thermostat-server/thermostat-server.cpp
+++ b/src/app/clusters/thermostat-server/thermostat-server.cpp
@@ -778,8 +778,8 @@ MatterThermostatClusterServerPreAttributeChangedCallback(const app::ConcreteAttr
             return Status::InvalidValue;
         }
         auto RequestedSystemMode = static_cast<SystemModeEnum>(*value);
-        if (ControlSequenceOfOperation > ControlSequenceOfOperationEnum::kCoolingAndHeatingWithReheat ||
-            RequestedSystemMode > SystemModeEnum::kFanOnly)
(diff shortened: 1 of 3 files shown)


andy31415 (approved):

andy31415 (approved):

src/app/clusters/thermostat-server/thermostat-server.cpp
```diff
@@ -778,8 +778,8 @@ MatterThermostatClusterServerPreAttributeChangedCallback(const app::ConcreteAttr
             return Status::InvalidValue;
         }
         auto RequestedSystemMode = static_cast<SystemModeEnum>(*value);
-        if (ControlSequenceOfOperation > ControlSequenceOfOperationEnum::kCoolingAndHeatingWithReheat ||
-            RequestedSystemMode > SystemModeEnum::kFanOnly)
+        if (ControlSequenceOfOperation == ControlSequenceOfOperationEnum::kUnknownEnumValue ||
+            RequestedSystemMode == SystemModeEnum::kUnknownEnumValue)
```
bzbarsky-apple, line 782: Actually, this probably needs to call the clamp function.... Need to check why the test is passing like this.

bzbarsky-apple: It's passing because the XML says:
```
    <attribute side="server" code="0x001C" define="SYSTEM_MODE" type="SystemModeEnum" min="0x0" max="0x9" writable="true" default="1">
```
and the ember machinery returns error on values over that max, so for invalid values this code is not in fact reached right now.

woody-apple (approved):

PR #36082, opened 2024-10-15
[Fabric Sync] Update test to run a little faster

### LINKED ISSUE #35229: [Tooling] Optimize CCTRL test script on waiting time

We can speed up the CCTRL test by not blindly waiting 30 seconds. What we can do is we can subscribe to the Fabrics attribute in step 1 and we can wait up to 30 seconds here for a change in subscriptions. We can do that here because we are expecting DUT to be successfully commission TH_SERVER in this case.

### Platform

raspi

### Platform Version(s)

_No response_

### Type

Manually tested with SDK

_No response_

### Anything else?

_No response_

tehampson:
We are now waiting up to 30 seconds and are polling every couple of seconds for some fabric sync checks when we are checking to see if there if device has properly been added.

Alternative considered: Instead of polling we could subscribe to the attribute. This approach was not taken as the amount of code to setup and tear down is more. While from a optics perspective subscription is the more technically "right" option, it offers no additional value and the polling approach taken now results in tests running faster if the device is commissioned right away.

Testplan change that should go in first https://github.com/CHIP-Specifications/chip-test-plans/pull/4752

Fixes: https://github.com/project-chip/connectedhomeip/issues/35229

Tests:
* Ran updated test locally confirmed they pass
* CI passes.

### DIFF
--- a/src/python_testing/TC_CCTRL_2_2.py
+++ b/src/python_testing/TC_CCTRL_2_2.py
@@ -138,8 +138,7 @@ class TC_CCTRL_2_2(MatterBaseTest):
                  TestStep(24, "Reading Event CommissioningRequestResult from DUT, confirm one new event"),
                  TestStep(25, "Send CommissionNode command to DUT with CASE session, with valid parameters"),
                  TestStep(26, "Send OpenCommissioningWindow command on Administrator Commissioning Cluster sent to TH_SERVER"),
-                 TestStep(27, "Wait for DUT to successfully commission TH_SERVER, 30 seconds"),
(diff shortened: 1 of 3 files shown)


src/python_testing/TC_CCTRL_2_2.py
```diff
@@ -138,8 +138,7 @@ def steps_TC_CCTRL_2_2(self) -> list[TestStep]:
                  TestStep(24, "Reading Event CommissioningRequestResult from DUT, confirm one new event"),
```
tehampson, line 138: Blocking comment on https://github.com/CHIP-Specifications/chip-test-plans/pull/4752 merging first

arkq (approved):

yufengwangca (approved):

PR #36085, opened 2024-10-15
[LVL] Add a check for Lvl command's OptionsBitmap arguments

### LINKED ISSUE #34554: [1.3] The Commands in the Level Control Cluster Do Not Obey the Constraint Ranges
### Reproduction steps

OptionsMask and OptionsOverride in OptionsBitmap with Bitmap Parameter Type (Limited to 0-1)

image

However, in the commands: MoveToLevel, Move, Step, Stop, and their variants with OnOff, the commands work normally when setting OptionsMask and OptionsOverride to 3.

```
./chip-tool any command-by-id 0x8 0x1 '{"0":1,"1":1,"2":3,"3":3}' 0x654324 1
```

### Bug prevalence

each time

561d23d0db215a99705ff0696e73853c8edf11b2 

### Platform

other

### Platform Version(s)

1.3

### Type

Spec Compliance Issue

### Anything else?

_No response_

jmartinez-silabs:
Fixes #34554

**Changes**
Add a check in all LVL command handlers that validates if the received `optionsMask` and `optionsOverride` arguments are within the expected values range. Return ConstraintError when they are not.

MISC: change ChipLogProgress to -> ChipLogError for some error message in the lvl cluster server implementation.

Open question; Based on spec, optionsMask and optionsOverride shall be provided.  However, the SDK implementation has allowed omitting them, at the command parsing stage and in the cluster server implementation. This PR doesn't change this behaviour.

**Tests**
Manually tested the out-of-range check.
CI test can validate for regressions.

Should test plan steps be added to test out-of-range argument values?

### DIFF
--- a/src/app/clusters/level-control/level-control.cpp
+++ b/src/app/clusters/level-control/level-control.cpp
@@ -125,6 +125,7 @@ static void writeRemainingTime(EndpointId endpoint, uint16_t remainingTimeMs, bo
 static bool shouldExecuteIfOff(EndpointId endpoint, CommandId commandId, chip::Optional<chip::BitMask<OptionsBitmap>> optionsMask,
                                chip::Optional<chip::BitMask<OptionsBitmap>> optionsOverride);
 
+static Status VerifyOptionsBitmap(Optional<BitMask<OptionsBitmap>> optionsMask, Optional<BitMask<OptionsBitmap>> optionsOverride);
 static Status SetCurrentLevelQuietReport(EndpointId endpoint, EmberAfLevelControlState * state,
                                          DataModel::Nullable<uint8_t> newValue, bool isEndOfTransition);
 
@@ -434,7 +435,7 @@ void emberAfLevelControlClusterServerTickCallback(EndpointId endpoint)
 
     if (status != Status::Success || currentLevel.IsNull())
     {
-        ChipLogProgress(Zcl, "ERR: reading current level %x", to_underlying(status));
+        ChipLogError(Zcl, "ERR: reading current level %x", to_underlying(status));
         state->callbackSchedule.runTime = System::Clock::Milliseconds32(0);
         writeRemainingTime(endpoint, 0);
         return;
@@ -469,7 +470,7 @@ void emberAfLevelControlClusterServerTickCallback(EndpointId endpoint)
     status          = SetCurrentLevelQuietReport(endpoint, state, currentLevel, isTransitionEnd);
     if (status != Status::Success)
     {
-        ChipLogProgress(Zcl, "ERR: writing current level %x", to_underlying(status));
+        ChipLogError(Zcl, "ERR: writing current level %x", to_underlying(status));
         state->callbackSchedule.runTime = System::Clock::Milliseconds32(0);
         writeRemainingTime(endpoint, 0);
         return;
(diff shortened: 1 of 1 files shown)


src/app/clusters/level-control/level-control.cpp
```diff
@@ -886,6 +909,12 @@ static Status moveToLevelHandler(EndpointId endpoint, CommandId commandId, uint8
         return Status::InvalidCommand;
     }
 
+    Status status = VerifyOptionsBitmap(optionsMask, optionsOverride);
+    if (status != Status::Success)
```
rcasallas-silabs, line 913:
```suggestion
    VerifyOrReturnError(Status::Success == status, status);
```

rcasallas-silabs (approved):

src/app/clusters/level-control/level-control.cpp
```diff
@@ -886,6 +909,12 @@ static Status moveToLevelHandler(EndpointId endpoint, CommandId commandId, uint8
         return Status::InvalidCommand;
     }
 
+    Status status = VerifyOptionsBitmap(optionsMask, optionsOverride);
+    VerifyOrReturnError(Status::Success == status, status);
+    {
+        return status;
+    }
```
jmartinez-silabs, line 916:
```suggestion
```

src/app/clusters/level-control/level-control.cpp
```diff
@@ -580,6 +581,28 @@ static void setOnOffValue(EndpointId endpoint, bool onOff)
...
+{
+    constexpr BitMask<OptionsBitmap> maxOptionValue(
+        OptionsBitmap::kExecuteIfOff, OptionsBitmap::kCoupleColorTempToLevel); // add any new bitmap field to this value!
+
+    // Based on spec, we shall expect a value for those arguments. But that's not what has been done in SDK so far.
+    // Should we error out?
+    if (optionsMask.HasValue())
+    {
+        VerifyOrReturnError(optionsMask.Value() <= maxOptionValue, Status::ConstraintError,
+                            ChipLogError(Zcl, "Command error: out of range optionsMask argument"));
```
bzbarsky-apple, line 594: Is this the behavior the spec calls for?

jmartinez-silabs: @bzbarsky-apple I did not find any indication in the cluster spec that specifies what to do in case of unknown bits options bitfield. 
My assumption was to treat them as values outside of the constraint of the expected parameter, like other argument types.

Ticket #34554 complains about unrespected constraints after some fuzz testing found those issues, so I was trying to address that.

However in the main spec, says that Reserved bit/undefined by conveyed from another implementation is ignored

> 1.8.2. Reserved Bit Fields
> An undefined bit or bit field SHALL be considered reserved for future revisions of this standard and SHALL NOT be available for implementation.
> 
> It is RECOMMENDED that a bit stay undefined, rather than defining it as "reserved".
> 
> An implementation of a revision where a bit is reserved SHALL indicate that bit as zero when conveying that bit in an interaction, **and ignore that bit when conveyed from another implementation.**

I should close this PR then and reply, in the issue ticket, that it is an expected behaviour. 

Does that work for you?

bzbarsky-apple: The basic question here is: what should happen when a newer-revision client (which knows about more bits) talks to an older-revision server?  Just ignoring the unknown bits would make sense to me, but really the lighting spec people should figure this out.

jmartinez-silabs:
Closing this PR. On the server side, the undefined OptionsMask and OptionsOverride bits will be ignored on reception and the command will be processed normally taking into account only the known bits.

The Client should however not allow to send undefined bits from its implementation version perspective.
A constraint check could be added on chip-tool to prevent sending undefined bits in those fields. However, it might not me worth the time since chip-tool is a test tool it can be useful to allow show values for testing purposes

PR #36088, opened 2024-10-15
[Fabric Sync] Have ECOINFO mark attribute dirty when new entry added …

### LINKED ISSUE #34553: [Feature] Implements the Ecosystem Information Server in Matter SDK

tehampson:
Fixes: https://github.com/project-chip/connectedhomeip/issues/34553

Tests:
* Unit test pass
* Confirmed that when application calls adds either AddDeviceInfo, or AddLocationInfo that client listening for subscription gets the update


src/app/clusters/ecosystem-information-server/ecosystem-information-server.h
```diff
@@ -144,11 +158,14 @@ class EcosystemLocationStruct
...
 };
 
-class EcosystemInformationServer
+class EcosystemInformationServer : public MatterContext
 {
 public:
     static EcosystemInformationServer & Instance();
 
+    EcosystemInformationServer() : mMatterContext(*this) {};
+    EcosystemInformationServer(TestOnlyParameter _, MatterContext & aMatterContext) : mMatterContext(aMatterContext) {};
```
cecille, line 167: I'm probably missing something, but what is TestOnlyParameter for?

src/app/clusters/ecosystem-information-server/ecosystem-information-server.h
```diff
@@ -212,6 +229,7 @@ class EcosystemInformationServer
     CHIP_ERROR EncodeDeviceDirectoryAttribute(EndpointId aEndpoint, AttributeValueEncoder & aEncoder);
     CHIP_ERROR EncodeLocationStructAttribute(EndpointId aEndpoint, AttributeValueEncoder & aEncoder);
 
+    MatterContext & mMatterContext;
```
cecille, line 232: It seems a bit odd to have the class have a reference to itself. Is there a gain to having the server BE a context as well as having a reference to the context?

src/app/tests/TestEcosystemInformationCluster.cpp
```diff
@@ -49,11 +49,27 @@ const ClusterId kEcosystemInfoClusterId         = EcosystemInformation::Id;
 const AttributeId kDeviceDirectoryAttributeId   = EcosystemInformation::Attributes::DeviceDirectory::Id;
 const AttributeId kLocationDirectoryAttributeId = EcosystemInformation::Attributes::LocationDirectory::Id;
 
+class MockMatterContext : public MatterContext
```
cecille, line 52: totally not a blocker, but given that this mirrors the valve context, do you think it's worthwhile having this be made generic and centralized somewhere? ie, should clusters define their OWN matter context, or should we have one matter context to rule them all with a standard test definition?

src/app/clusters/ecosystem-information-server/ecosystem-information-server.h
tehampson: I think the gain is that I always just call `mMatterContext.MarkDirty(...) without needing to have if conditions

src/app/tests/TestEcosystemInformationCluster.cpp
tehampson: If you are refering to MatterContext I agree that we should have a base class that takes care of the most common cases to avoid boilerplate.

If you are referring to MockMatterContext also wouldn't be a bad idea for a generic on, just there might be some differentiation between tests, but they can extend the mock matter context some more for their cases

cecille (approved):

saurabhst (approved):

PR #36090, opened 2024-10-16
Moving more stuff out of the MTRDevice

woody-apple:
MTRDevice is still using some basic C++ stuff here, we should be not doing this in MTRDevice_XPC, I did a quick and simple hack here. @bzbarsky-apple

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
@@ -1646,11 +1646,6 @@ typedef NS_ENUM(NSUInteger, MTRDeviceWorkItemDuplicateTypeID) {
     }
     [_mostRecentReportTimes addObject:[NSDate now]];
 
-    {
-        std::lock_guard lock(_descriptionLock);
-        _mostRecentReportTimeForDescription = [_mostRecentReportTimes lastObject];
-    }
-
     // Calculate running average and update multiplier - need at least 2 items to calculate intervals
     if (_mostRecentReportTimes.count > 2) {
         NSTimeInterval cumulativeIntervals = 0;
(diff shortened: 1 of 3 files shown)


src/darwin/Framework/CHIP/MTRDevice_XPC.mm
```diff
@@ -309,4 +309,41 @@ - (void)downloadLogOfType:(MTRDiagnosticLogType)type
 // Not Supported via XPC
 // - (oneway void)downloadLogOfType:(MTRDiagnosticLogType)type nodeID:(NSNumber *)nodeID timeout:(NSTimeInterval)timeout completion:(void (^)(NSURL * _Nullable url, NSError * _Nullable error))completion;
 
+#pragma mark - Storage Overrides
+
+- (NSDictionary<MTRClusterPath *, MTRDeviceClusterData *> *)_clusterDataToPersistSnapshot
```
bzbarsky-apple, line 314: If the goal is to not have persistence at all in MTRDevice_XPC, then we should just rip it out of MTRDevice, not do this, I think....

src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
```diff
@@ -1646,11 +1646,6 @@ - (void)_scheduleClusterDataPersistence
     }
     [_mostRecentReportTimes addObject:[NSDate now]];
 
-    {
-        std::lock_guard lock(_descriptionLock);
-        _mostRecentReportTimeForDescription = [_mostRecentReportTimes lastObject];
```
bzbarsky-apple, line 1651: Why is this being removed?

src/darwin/Framework/CHIP/MTRDevice_XPC.mm
woody-apple: Have at it :)

src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
woody-apple: I just copied and pasted the base implementation into concrete, please fix :)

bzbarsky-apple:
I think we should do https://github.com/project-chip/connectedhomeip/pull/36132 instead.

PR #36094, opened 2024-10-16
[Python] revert unnecessary reformat

agners:
We use Python line length of 132 characters. This reverts unnecessary reformat introduced by #32257.

For easier review, this has been implemented with two commits, one which simply reverts #32257, and the second which implements the actual change again.

This has been cross-validated by reformatting with `autopep8` default line length, the effective change ends up to be exactly the same.

### DIFF
--- a/src/controller/python/chip/clusters/Attribute.py
+++ b/src/controller/python/chip/clusters/Attribute.py
@@ -110,11 +110,9 @@ class TypedAttributePath:
         '''Only one of either ClusterType and AttributeType OR Path may be provided.'''
 
         if (self.ClusterType is not None and self.AttributeType is not None) and self.Path is not None:
-            raise ValueError(
-                "Only one of either ClusterType and AttributeType OR Path may be provided.")
+            raise ValueError("Only one of either ClusterType and AttributeType OR Path may be provided.")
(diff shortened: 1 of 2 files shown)


arkq:
I think that the proper solution for that would be to "somehow" add python formatting to restyled and enforce it during PR. Otherwise, everyone will format the code to their likeness - which introduces unnecessary noise in the code and bloats repo size.

I'm not going to mark it as requested changes, but please consider to add codestyle (whatever the codestyle is, IMHO it would be better to have some random and consistent one than noting at all) check to CI.

andy31415 (approved):

agners:
> I think that the proper solution for that would be to "somehow" add python formatting to restyled and enforce it during PR. Otherwise, everyone will format the code to their likeness - which introduces unnecessary noise in the code and bloats repo size.
> 
> I'm not going to mark it as requested changes, but please consider to add codestyle (whatever the codestyle is, IMHO it would be better to have some random and consistent one than noting at all) check to CI.

We do run autopep8 as part of the pipeline, however, it seems that autopep8 only makes sure that lines are shorter than x, and does not reformat if lines got shortened.

Introducing a more forceful approach to formatting probably would introduce quite some change at first.

Issue #35333, opened 2024-08-30
[Fabric Sync Example] fabric-admin DeviceSubscription need better error handling


tehampson:
The fabric-admin example for fabric sync is not able to recover tracking changes to a bridged CADMIN attributes if the DeviceSubscription:
1. Fails to connect
2. After connecting device stops sending out subscriptiong (due to restarting or something like that)

PR #36101, opened 2024-10-16
Add plumbing for BRBINFO reachability change

tehampson:
Progress towards https://github.com/project-chip/connectedhomeip/issues/35333

This will allow for admin side, when it is ready, to call rpc to update reach-ability when there are connection issues with the device.


examples/fabric-bridge-app/fabric-bridge-common/src/BridgedDevice.cpp
```diff
@@ -65,6 +65,29 @@ void BridgedDevice::SetReachable(bool reachable)
     }
 }
 
+void BridgedDevice::ReachableChanged(bool reachable)
+{
+    EndpointId endpointId = mEndpointId;
+    bool reachableChanged = (mReachable != reachable);
+    SetReachable(reachable);
```
yufengwangca, line 72: nit: we only need to SetReachable() if it changed

yufengwangca (approved):

arkq (approved):

PR #36104, opened 2024-10-16
Add `emberAfMetadataStructureGeneration` for codegen data model.

andy31415:
CodegenDataModel is using caching to speed up queries on clusters. However dynamic endpoints can be created/removed at will and in those cases old cluster pointers should not be re-used.

### Changes

- introduce an ember `metadata structure generation` number that increases on every dynamic module change. Did not worry about overflow as even 32-bit integers should last a sufficient amount of time to not introduce caching issues.
- Cluster cache in codegen data model uses the generation number to decide if a cached pointer is likely valid or not.

### DIFF
--- a/src/app/util/attribute-storage.cpp
+++ b/src/app/util/attribute-storage.cpp
@@ -93,6 +93,11 @@ uint8_t singletonAttributeData[ACTUAL_SINGLETONS_SIZE];
 
 uint16_t emberEndpointCount = 0;
 
+/// Determines a incremental unique index for ember
+/// metadata that is increased whenever a structural change is made to the
+/// ember metadata (e.g. changing dynamic endpoints or enabling/disabling endpoints)
+unsigned emberMetadataStructureGeneration = 0;
+
 // If we have attributes that are more than 4 bytes, then
 // we need this data block for the defaults
 #if (defined(GENERATED_DEFAULTS) && GENERATED_DEFAULTS_COUNT)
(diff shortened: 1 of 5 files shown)


arkq (approved):

tehampson (approved):

src/app/util/attribute-storage.cpp
```diff
@@ -93,6 +93,11 @@ uint8_t singletonAttributeData[ACTUAL_SINGLETONS_SIZE];
 
 uint16_t emberEndpointCount = 0;
 
+/// Determines a incremental unique index for ember
+/// metadata that is increased whenever a structural change is made to the
+/// ember metadata (e.g. changing dynamic endpoints or enabling/disabling endpoints)
+unsigned emberMetadataStructureGeneration = 0;
```
bzbarsky-apple, line 99: enable/disable can certainly happen enough to overflow an `unsigned`, no?  Shouldn't this be 64-bit?

andy31415: Sure, it can overflow but will we retain a cache for so long? If we enable/disable every 100ms my calculator says it takes a bit over 13.5 years to overflow and we have a single cached path/cluster pointer. 

For this to be buggy, we would need to both:
   - overflow to the exact same value (not more)
   - try to re-use the same cached path for a cluster (without any cached cluster calls in between)
   
Hence my thinking that 32-bit is sufficient. If you think lets just do 64 bit we can do that as well at a tiny ram/flash cost.

bzbarsky-apple: Ah, good point, it has to overflow to the same value.  Then yeah, this is not an issue in practice.

PR #36111, opened 2024-10-17
[Silabs] Add BRD4343a support

arun-silabs:
This PR has all the changes required for adding brd4343a support.

### DIFF
--- a/.github/workflows/examples-efr32.yaml
+++ b/.github/workflows/examples-efr32.yaml
@@ -106,7 +106,6 @@ jobs:
                 --enable-flashbundle \
                 --target efr32-brd2605a-light-skip-rps-generation \
                 --target efr32-brd4338a-lock-skip-rps-generation \
-                --target efr32-brd4343a-lock-skip-rps-generation \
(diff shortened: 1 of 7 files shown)


.github/workflows/examples-efr32.yaml
```diff
@@ -106,6 +106,7 @@ jobs:
                 --enable-flashbundle \
                 --target efr32-brd2605a-light-skip-rps-generation \
                 --target efr32-brd4338a-lock-skip-rps-generation \
+                --target efr32-brd4343a-lock-skip-rps-generation \
```
jmartinez-silabs, line 109: It isn't worth building the lock app again for a different board when it is the same board family and essentially builds the same files.
Ci is long enough already.

jmartinez-silabs (approved):

.github/workflows/examples-efr32.yaml
arun-silabs: removed for brd4343a

andy31415 (approved):

PR #36086, opened 2024-10-15
[Silabs] Update WiseConnect to v3.3.3

arun-silabs:
**Change**
Updated wifi_sdk submodule from WiseConnect to v3.3.3.

**Testing**
Validated that lighting-app and lock-app compile and run successfully on BRD4338A

PR #36117, opened 2024-10-17
[Silabs] Update WiseConnect-wifi-bt-sdk to 2.10.3 for RS9116

bhmanda-silabs:
**Change**

- Updated WiseConnect-wifi-bt-sdk to 2.10.3 
- Added wpa_wpa2_mixed security type in rsi_if.cpp

**Testing**
Validated commissioning with mg24 + RS9116 combination

### DIFF
--- a/.github/workflows/release_artifacts.yaml
+++ b/.github/workflows/release_artifacts.yaml
@@ -32,7 +32,7 @@ jobs:
         runs-on: ubuntu-latest
 
         container:
-            image: ghcr.io/project-chip/chip-build-esp32:84
+            image: ghcr.io/project-chip/chip-build-esp32:81
 
         steps:
             - name: Checkout
@@ -64,7 +64,7 @@ jobs:
         runs-on: ubuntu-latest
 
         container:
-            image: ghcr.io/project-chip/chip-build-efr32:83
+            image: ghcr.io/project-chip/chip-build-efr32:84
         steps:
             - name: Checkout
               uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


mkardous-silabs (commented):
This requires a docker update in a seperate PR before hand otherwise the CI will never pass.

bhmanda-silabs:
> This requires a docker update in a seperate PR before hand otherwise the CI will never pass.

Done @mkardous-silabs.

.github/workflows/release_artifacts.yaml
```diff
@@ -32,7 +32,7 @@ jobs:
         runs-on: ubuntu-latest
 
         container:
-            image: ghcr.io/project-chip/chip-build-esp32:81
```
mkardous-silabs, line 35: I don't think we should be updating the esp docker image.

mkardous-silabs:
#36086 needs to be merged for this PR to be able to pass the necessary CIs.

integrations/cloudbuild/smoke-test.yaml
mkardous-silabs, line 1: Does this file need to be updated?

mkardous-silabs (approved):

.github/workflows/release_artifacts.yaml
bhmanda-silabs: Updated @mkardous-silabs

jmartinez-silabs (approved):

yufengwangca (approved):

PR #36118, opened 2024-10-17
[NXP][third_party] Border router updates and fixes

marius-preda:
*Fixed selecting correct DnsSd implementation when Border Router is enabled. 

*Fix an issue with the SRP client starting even when the device is running in Border Router mode.

*Removed ChipDnssdShutdown function, there is no use for it. 

*Removed unused OpenThread define.


axelnxp:
Added submodule label because the check is triggered by changes in `third_party` even if no submodule has been changed.

axelnxp (approved):

andy31415 (approved):

PR #36125, opened 2024-10-17
Add missing Water Heater device to matter-devices.xml

soares-sergio:
This PR adds the missing Water Heater device to matter-devices.xml.

The description was generated using the Alchemy tool (https://github.com/project-chip/alchemy) with the following command:

`alchemy zap --attribute="in-progress"  --sdkRoot=./connectedhomeip/ --specRoot=./connectedhomeip-spec/
./connectedhomeip-spec/src/device_types/WaterHeater.adoc`

I manually fixed the device name from `Matter Water Heater` to `Water Heater`.

Test:
* Opened the zap tool and confirmed the new device shows up.


tcarmelveilleux (approved):

src/app/zap-templates/zcl/data-model/chip/matter-devices.xml
```diff
@@ -2698,4 +2698,47 @@ limitations under the License.
...
+                <requireCommand>Identify</requireCommand>
+            </include>
+            <include cluster="Thermostat" client="false" server="true" clientLocked="true" serverLocked="true">
+                <requireAttribute>CONTROL_SEQUENCE_OF_OPERATION</requireAttribute>
+                <requireAttribute>LOCAL_TEMPERATURE</requireAttribute>
+                <requireAttribute>SYSTEM_MODE</requireAttribute>
+                <requireCommand>SetpointRaiseLower</requireCommand>
+            </include>
+            <include cluster="Water Heater Management" client="false" server="true" clientLocked="true" serverLocked="true">
+                <requireAttribute>BOOST_STATE</requireAttribute>
```
cecille, line 2728: Does alchemy put these in? The required attributes here appear to just be the mandatory attributes etc. from the cluster rather than the device type overrides. Seems redundant.

soares-sergio: Yes, it did put it in automatically. Initially I found it strange too but looking at the file, it shows up on other clusters as well, such as: https://github.com/project-chip/connectedhomeip/blob/master/src/app/zap-templates/zcl/data-model/chip/matter-devices.xml#L232

So there's a bit of inconsistency. Some have and some don't.

jamesharrow (approved):
Thanks for putting this in.

I think if this is missing then we must also be missing Heat pump, SolarPV, Battery as well!

The others don’t need any new clusters and just rely upon existing 1.3 clusters (except heat pump also needed water heater as optional).

note that there is no example app for these device types.

soares-sergio:
> > This PR adds the missing Water Heater device to matter-devices.xml.
> > The description was generated using the Alchemy tool (https://github.com/project-chip/alchemy) with the following command:
> > `alchemy zap --attribute="in-progress" --sdkRoot=./connectedhomeip/ --specRoot=./connectedhomeip-spec/ ./connectedhomeip-spec/src/device_types/WaterHeater.adoc`
> > I manually fixed the device name from `Matter Water Heater` to `Water Heater`.
> > Test:
> > 
> > * Opened the zap tool and confirmed the new device shows up.
> 
> Where did you need to fix the device name? When I run alchemy it is crashing with a segfault. Is this a spec issue or something broken somewhere in the SDK?

@jamesharrow 

I fixed it here: https://github.com/project-chip/connectedhomeip/pull/36125/files#diff-a87b1667273c20323f1f0b6774a84ee696ac3bf99365736857b8d70d1bb1890eR2704

Changed 
`<typeName>Matter Water Heater</typeName>`
to
`<typeName>Water Heater</typeName>`

soares-sergio:
@jamesharrow 

Alchemy works for me, just tried again. I don't get a segfault, just some warnings. I'm running on linux. This is my step-by-step:
1. Checkout `alchemy`,  `connectedhomeip` and `connectedhomeip-spec` repos in `~/`
2. `cd alchemy`
3. `go build`
4. cd `~/`
5. `./alchemy/alchemy zap --attribute="in-progress"  --sdkRoot=./connectedhomeip/ --specRoot=./connectedhomeip-spec/ ./connectedhomeip-spec/src/device_types/WaterHeater.adoc`

There's a couple bugs in the output of alchemy:
1. The result is written in `matter-devices.xml` but it deletes everything else. So I just copied the Water Heater generated code, reverted the file back to head and manually added it to the end
2. Changed the name from `Matter Water Heater` to `Water Heater`.

src/app/zap-templates/zcl/data-model/chip/matter-devices.xml
```diff
@@ -2698,4 +2698,47 @@ limitations under the License.
...
+        <name>MA-waterheater</name>
+        <domain>CHIP</domain>
+        <typeName>Water Heater</typeName>
+        <profileId editable="false">0x0103</profileId>
+        <deviceId editable="false">0x050F</deviceId>
+        <class>Simple</class>
+        <scope>Endpoint</scope>
+        <clusters>
+            <include cluster="Descriptor" client="false" server="true" clientLocked="true" serverLocked="true">
+                <requireAttribute>CLIENT_LIST</requireAttribute>
```
bzbarsky-apple, line 2711: Why all the requireAttribute for attributes which are mandatory on that cluster anyway?

soares-sergio: I checked with Hasty and he mentioned he just replicated the pattern in this file. It looks like in the past, we used to include all of these attributes. Over time, when manually adding entries, people stopped including these attributes but the tool still generates those.

soares-sergio: I suggest we leave this as-is for now as this is auto-generated.
If we want to remove those in the future we should probably update Alchemy directly and then regenerate the whole file.
FYI @hasty

andy31415 (approved):

PR #36131, opened 2024-10-17
[Darwin] Some MTRDevice mtr_strongify sites need to guard against nil self

jtung-apple:
In MTRDevice_Concreate, some sites where `mtr_strongify` is used need to guard against a nil self as they have direct ivar references, which can crash if the block is run after MTRDevice object goes away.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
+++ b/src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
@@ -389,8 +389,10 @@ typedef NS_ENUM(NSUInteger, MTRDeviceWorkItemDuplicateTypeID) {
             mtr_weakify(self);
             _systemTimeChangeObserverToken = [[NSNotificationCenter defaultCenter] addObserverForName:NSSystemClockDidChangeNotification object:nil queue:nil usingBlock:^(NSNotification * _Nonnull notification) {
                 mtr_strongify(self);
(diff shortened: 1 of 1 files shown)


kiel-apple (approved):

src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
```diff
@@ -1178,6 +1182,10 @@ - (void)_scheduleSubscriptionPoolWork:(dispatch_block_t)workBlock inNanoseconds:
         MTRAsyncWorkItem * workItem = [[MTRAsyncWorkItem alloc] initWithQueue:self.queue];
         [workItem setReadyHandler:^(id _Nonnull context, NSInteger retryCount, MTRAsyncWorkCompletionBlock _Nonnull completion) {
             mtr_strongify(self);
+            if (!self) {
```
ksperling-apple, line 1185: `VerifyOrReturn(self);` ? Would be nice to have a common pattern for this that's used everywhere.

ksperling-apple (approved):

src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
jtung-apple: Good idea - I've made this change.

PR #36139, opened 2024-10-18
Fix DevicePairingDelegate::OnPairingComplete is not called if pairing code is wrong

evg-trushkin:
Only the `DevicePairingDelegate::OnStatusUpdate(Status::SecurePairingFailed)` callback will be called in case of pairing device with wrong manual code. The signature of OnStatusUpdate() does not include CHIP_ERROR.

Seems that the correct behavior is to call DevicePairingDelegate::OnPairingComplete(error) as well. 
Error will likely hold `CHIP Error 0x00000032: Timeout`.

This way looking at `OnPairingComplete`, `OnCommissioningSuccess` and `OnCommissioningFailure` is enough to keep track on commissioning process. And all of them provide CHIP_ERROR for diagnostics.

### DIFF
--- a/src/controller/CHIPDeviceController.cpp
+++ b/src/controller/CHIPDeviceController.cpp
@@ -1141,11 +1141,10 @@ void DeviceCommissioner::RendezvousCleanup(CHIP_ERROR status)
         // for IP commissioning, we have taken a reference to the
         // operational node to send the completion command.
         ReleaseCommissioneeDevice(mDeviceInPASEEstablishment);
-
-        if (mPairingDelegate != nullptr)
-        {
-            mPairingDelegate->OnPairingComplete(status);
-        }
+    }
+    if (CHIP_NO_ERROR != status && mPairingDelegate != nullptr)
+    {
+        mPairingDelegate->OnPairingComplete(status);
     }
 }
 


src/controller/CHIPDeviceController.cpp
```diff
@@ -1141,11 +1141,10 @@ void DeviceCommissioner::RendezvousCleanup(CHIP_ERROR status)
         // for IP commissioning, we have taken a reference to the
         // operational node to send the completion command.
         ReleaseCommissioneeDevice(mDeviceInPASEEstablishment);
-
-        if (mPairingDelegate != nullptr)
-        {
-            mPairingDelegate->OnPairingComplete(status);
-        }
+    }
+    if (CHIP_NO_ERROR != status && mPairingDelegate != nullptr)
```
bzbarsky-apple, line 1145: The actual behavior here is very complicated (e.g. if your pairing code is _right_ but the device is not advertising that's the same situation as wrong pairing code!), and worse yet existing clients depend on it, so probably cannot be changed without breaking lots of shipping software.

The right answer here is that we should have a sane error reporting API, not what we have right now, but we should be _extremely_ careful about any changes to the existing behavior.

PR #36151, opened 2024-10-18
Add an MTRDevice_Concrete method for checking whether a data-value satisfies some expectation.

bzbarsky-apple:


src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
```diff
@@ -3319,6 +3319,133 @@ - (BOOL)_attributeDataValue:(NSDictionary *)one isEqualToDataValue:(NSDictionary
     return [one[MTRTypeKey] isEqual:theOther[MTRTypeKey]] && ((one[MTRValueKey] == theOther[MTRValueKey]) || [one[MTRValueKey] isEqual:theOther[MTRValueKey]]);
 }
 
+// _attributeDataValue:isEquivalentToDataValue: differs from
```
ksperling-apple, line 3322: Maybe this should be a free function instead of an MTRDevice method?

src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
```diff
@@ -3319,6 +3319,133 @@ - (BOOL)_attributeDataValue:(NSDictionary *)one isEqualToDataValue:(NSDictionary
...
+// _attributeDataValue:isEqualToDataValue in that it does not insist on the
+// order of fields in structs matching between the two values.  While in theory
```
ksperling-apple, line 3324: Should this potentially be an asymmetric test where the actual value can have additional struct fields that are not included in the expected value?

src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
bzbarsky-apple: I considered that... I basically did it by analogy with the _attributeDataValue:isEqualToDataValue: thing, but doing it as a free function is certainly doable.

src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
bzbarsky-apple: Hmm.  That is a good question.  As in, allow specifying only the parts of the struct you care about in terms of doing the comparison?

I'll think about that a bit as a followup, but for now I think testing for having the same fields is good.

src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
bzbarsky-apple: Making this a free function makes it hard to log the device, so not going to do that.

src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
bzbarsky-apple: It's also worth thinking a bit about whether list order should be enforced, by the way.  At some point we might need some configuration flags for this or something.  But for now, let's stick to this.

kiel-apple (approved):

bzbarsky-apple:

PR #36160, opened 2024-10-19
Improve Exit Error Handling for failures in PASE pairing requests

Alami-Amine:
- In the event of failures in Pairing Requests, in the method `PASESession::Pair`, the exit is not as clean as it should be.
- this was discovered while fuzzing; Sanitizer Crashes resulted before the below fix was integrated.

---
If a failure happens in the call to `PASESession::Init` , (triggered by passing a bigger setup passcode when fuzzing), the Exchange Context is not properly destructed becuase the `ReferenceCount` is not equal to 0.
- This happens before the PASESession takes ownership of the passed exchange context.
- Fix: Close the exchange context that is passed as a pointer using `exchangeCtxt->Close();`

Error Log before Fix:
```cpp
[1729359422.539] [2090338:2090338] [IN] SecureSession[0x512000000940]: Released - Type:1 LSID:45924
[1729359422.539] [2090338:2090338] [FP] Shutting down FabricTable
[1729359422.539] [2090338:2090338] [TS] Pending Last Known Good Time: 2023-10-14T01:16:48
[1729359422.539] [2090338:2090338] [TS] Previous Last Known Good Time: 2023-10-14T01:16:48
[1729359422.539] [2090338:2090338] [TS] Reverted Last Known Good Time to previous value
//The following two lines are error codes
[1729359422.539] [2090338:2090338] [EM] ExchangeContext: 7302i delegate=?
[1729359422.539] [2090338:2090338] [SPT] VerifyOrDie failure at src/messaging/ExchangeContext.cpp:337: mExchangeMgr != nullptr && GetReferenceCount() == 0

```

### DIFF
--- a/src/protocols/secure_channel/PASESession.cpp
+++ b/src/protocols/secure_channel/PASESession.cpp
@@ -232,7 +232,18 @@ CHIP_ERROR PASESession::Pair(SessionManager & sessionManager, uint32_t peerSetUp
 exit:
     if (err != CHIP_NO_ERROR)
     {
+        // If a failure happens in PASESession::Init, we need to ensure that the passed exchange context is closed to prevent
+        // resource leaks.
+        exchangeCtxt->Close();
+
+        // If a failure happens in SendPBKDFParamRequest, we need to Discard the exchange (mExchangeCtxt) so that Clear() doesn't
+        // try closing it.  The exchange will handle that.
+        DiscardExchange();
         Clear();
+        ChipLogError(SecureChannel, "Failed during PASE session pairing request: %" CHIP_ERROR_FORMAT, err.Format());
+        MATTER_TRACE_COUNTER("PASEFail");
+        // Do this last in case the delegate frees us.
+        NotifySessionEstablishmentError(err);
     }
     return err;
 }
--- a/src/protocols/secure_channel/PASESession.cpp
+++ b/src/protocols/secure_channel/PASESession.cpp
@@ -232,18 +232,15 @@ CHIP_ERROR PASESession::Pair(SessionManager & sessionManager, uint32_t peerSetUp
 exit:
     if (err != CHIP_NO_ERROR)
     {
-        // If a failure happens in PASESession::Init, we need to ensure that the passed exchange context is closed to prevent
-        // resource leaks.
-        exchangeCtxt->Close();
-
-        // If a failure happens in SendPBKDFParamRequest, we need to Discard the exchange (mExchangeCtxt) so that Clear() doesn't
-        // try closing it.  The exchange will handle that.
-        DiscardExchange();
+        // If a failure happens before we have placed the incoming exchange into `mExchangeCtxt`, we need to make
+        // sure to close the exchange to fulfill our API contract.
+        if (!mExchangeCtxt.HasValue())
+        {
+            exchangeCtxt->Close();
+        }
         Clear();
         ChipLogError(SecureChannel, "Failed during PASE session pairing request: %" CHIP_ERROR_FORMAT, err.Format());
         MATTER_TRACE_COUNTER("PASEFail");
-        // Do this last in case the delegate frees us.
-        NotifySessionEstablishmentError(err);
     }
     return err;
 }
--- a/src/protocols/secure_channel/PASESession.h
+++ b/src/protocols/secure_channel/PASESession.h
@@ -99,6 +99,10 @@ public:
      *                            ownership of the exchangeCtxt to PASESession object. PASESession
      *                            will close the exchange on (successful/failed) handshake completion.
      * @param delegate            Callback object
+     *                            Note: The delegate will not be notified if a failure occurs at this stage.
+     *                            The delegate will only be notified of an error during session establishment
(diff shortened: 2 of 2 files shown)


src/protocols/secure_channel/PASESession.cpp
```diff
@@ -232,7 +232,18 @@ CHIP_ERROR PASESession::Pair(SessionManager & sessionManager, uint32_t peerSetUp
 exit:
     if (err != CHIP_NO_ERROR)
     {
+        // If a failure happens in PASESession::Init, we need to ensure that the passed exchange context is closed to prevent
+        // resource leaks.
+        exchangeCtxt->Close();
+
+        // If a failure happens in SendPBKDFParamRequest, we need to Discard the exchange (mExchangeCtxt) so that Clear() doesn't
+        // try closing it.  The exchange will handle that.
+        DiscardExchange();
```
bzbarsky-apple, line 241: This seems pretty confused.

Fundamentally, there are two cases here:

1. We failed out after setting `mExchangeCtxt`.  In that case, we don't need to mess with the exchange; that will get handled automatically by `Clear()`.
2. We failed out before setting `mExchangeCtxt`.  In that case, we should close the exchange.  So the right logic here seems to be:

```suggestion
        // If a failure happens before we have placed the incoming exchange into `mExchangeCtxt`, we need to make
        // sure to close the exchange to fulfill our API contract.
        if (!mExchangeCtxt.HasValue()) {
            exchangeCtxt->Close();
        }
```

Or we could just emplace `mExchangeCtxt` before we call `Init`, and not need any special cleanup here at all?

src/protocols/secure_channel/PASESession.cpp
```diff
@@ -232,7 +232,18 @@ CHIP_ERROR PASESession::Pair(SessionManager & sessionManager, uint32_t peerSetUp
...
         Clear();
+        ChipLogError(SecureChannel, "Failed during PASE session pairing request: %" CHIP_ERROR_FORMAT, err.Format());
+        MATTER_TRACE_COUNTER("PASEFail");
+        // Do this last in case the delegate frees us.
+        NotifySessionEstablishmentError(err);
```
bzbarsky-apple, line 246: What's the API contract around this stuff?  It needs to be clearly documented.  For the exchange this is documented already, and we were just not fulfilling the API contract.  But for the delegate: do callers expect to get a call to the delegate if the `Pair()` call fails out?  I would not expect so, and certainly the one in CHIPDeviceController does not seem to expect it.  Concretely, in CHIPDeviceController.cpp the logic basically looks like this:

```
    err = device->GetPairing().Pair(*mSystemState->SessionMgr(), params.GetSetupPINCode(), GetLocalMRPConfig(), exchangeCtxt, this);
    if (err != CHIP_NO_ERROR)
    {
        ReleaseCommissioneeDevice(device);
    }
```

but now if we get `OnSessionEstablishmentError` before the `Pair()` call returns, that will call `RendezvousCleanup` which will do `ReleaseCommissioneeDevice()` on that same exact device pointer, which will do a double-free on the device.   Which might be a really bad idea.  So I think we should not be calling `NotifySessionEstablishmentError` here, and we should clearly document the API to explain when it _is_ called: Only after certain initial calls returned success.

bzbarsky-apple (changes requested):

src/protocols/secure_channel/PASESession.cpp
Alami-Amine: - I might have gotten confused, I am trying to reproduce my "2nd" Error Log, but for the moment I wasnt able to...

- Regarding your suggestion of emplacing mExchangeCtxt before Init, it wouldnt be possible since we call `Clear()` inside `PASESession::Init` which will clear mExchangeCtxt

src/protocols/secure_channel/PASESession.cpp
Alami-Amine: I agree, fixed.
where exactly do we document API contracts anyways, above the declaration of the method?

src/protocols/secure_channel/PASESession.cpp
Alami-Amine: integrated

src/protocols/secure_channel/PASESession.h
```diff
@@ -99,6 +99,10 @@ class DLL_EXPORT PASESession : public Messaging::UnsolicitedMessageHandler,
      *                            ownership of the exchangeCtxt to PASESession object. PASESession
      *                            will close the exchange on (successful/failed) handshake completion.
      * @param delegate            Callback object
+     *                            Note: The delegate will not be notified if a failure occurs at this stage.
+     *                            The delegate will only be notified of an error during session establishment
+     *                            if it happens after the Pair() call succeeds and the PBKDFParamRequest
+     *                            is sent; i.e., after the session establishment process has started.
```
bzbarsky-apple, line 105:
```suggestion
     *                            Note: The delegate will be notified if and only if Pair() returns success.
```

bzbarsky-apple (approved):

andy31415 (approved):

PR #36163, opened 2024-10-20
create a cirque virtualenv

### LINKED ISSUE #33950: [Cirque] use virtualenv
After upgrading to 24.04 it becomes very obvious that cirque should use a virtualenv instead of dirtying the system python packages.

We should instead create a virtualenv for every pip package install and python usage. We also seem to have significant copy&paste throughout the tests for pip install, those should be consolidated.

FergusHuang:
Fixes #33950 [Cirque] use virtualenv
create a cirque virtualenv for every pip package installation without the break-system-packages option.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,6 +54,7 @@ docs/html
 
 # Python venv
 .venv
+cirque_venv
 
 # Documentation
 docs/_build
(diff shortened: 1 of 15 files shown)


yunhanw-google (approved):

.gitignore
```diff
@@ -54,6 +54,7 @@ docs/html
 
 # Python venv
 .venv
+cirque_venv
```
andy31415, line 57: adding more top_level directories is somewhat inconvenient. Could we use `out` for example? we do so for default python builds.

andy31415 (commented):
@FergusHuang please add a `Testing` section to the PR summary that describes how this PR is tested. If the testing is manual, please describe how it was tested in detail and why automated testing is impossible.

PR #35522, opened 2024-09-10
[Silabs] Adding Queue for the ble events and removing the semaphore

chirag-silabs:
**Problem:**
The Ble on the Rs911x devices was using bitmaps for the events and having semaphores between source files via extern keywords

**Fix**
Moving to queue based implementation of the ble events which will not poll everytime and will be done in the FIFO fashion. Sending the details via the structure in queue

**Testing**
Tested locally with 917 SoC

PR #36167, opened 2024-10-21
[Silabs] Fixing hardfault issue with RS9116 ble manager

bhmanda-silabs:
**Problem:**
The Ble manager for Rs911x devices code refactored by below PR and in this no where the memory is not allocated for _bleEvent.eventData_ member.
https://github.com/project-chip/connectedhomeip/pull/35522
Due to which hardfault is coming when this struct member is getting defined.

**Fix**
Fixed the issue by allocating

**Testing**
Tested commissioning with 917 SoC, RS9116+EFR mg24.

### DIFF
--- a/src/platform/silabs/rs911x/wfx_sl_ble_init.cpp
+++ b/src/platform/silabs/rs911x/wfx_sl_ble_init.cpp
@@ -294,6 +294,7 @@ void SilabsBleWrapper::rsi_ble_add_char_val_att(void * serv_handler, uint16_t ha
 
 uint32_t SilabsBleWrapper::rsi_ble_add_matter_service(void)
 {
+    bleEvent.eventData                                      = new SilabsBleWrapper::sl_wfx_msg_t();
     uuid_t custom_service                                   = { RSI_BLE_MATTER_CUSTOM_SERVICE_UUID };
     custom_service.size                                     = RSI_BLE_MATTER_CUSTOM_SERVICE_SIZE;
     custom_service.val.val16                                = RSI_BLE_MATTER_CUSTOM_SERVICE_VALUE_16;


rosahay-silabs (approved):

src/platform/silabs/rs911x/wfx_sl_ble_init.cpp
```diff
@@ -294,6 +294,7 @@ void SilabsBleWrapper::rsi_ble_add_char_val_att(void * serv_handler, uint16_t ha
 
 uint32_t SilabsBleWrapper::rsi_ble_add_matter_service(void)
 {
+    bleEvent.eventData                                      = new SilabsBleWrapper::sl_wfx_msg_t();
```
mkardous-silabs, line 297: Instead of using dynamic allocation, we could just change the structure type to avoid using a pointer for the eventData. From i can tell, there doesn't seem to be an explicitly requirement and the variable is global.

mkardous-silabs (changes requested):
As a general comment, do we need to use a global variable for the `bleEvent`?
`osMessageQueuePut` copies the the buffer to the queue to not require the user of the API to maintain the initial event data.

We can just create a local variable each time we want to pass an event to the message queue.

chirag-silabs:
> local

@mkardous-silabs The reason it can't be saved as a local variable and passed through the queue is that we need the data which is saved in the same struct during the init/previous event data during the some of events. So instead of having multiple global variables it is better to have a single struct.

mkardous-silabs:
> > As a general comment, do we need to use a global variable for the bleEvent?
> 
> @mkardous-silabs The reason it can't be saved as a local variable and passed through the queue is that we need the data which is saved in the same struct during the init/previous event data during the some of events. So instead of having multiple global variables it is better to have a single struct.

@chirag-silabs From what i can see from the code, the only spot i see where the ble event is used differently is `rsi_ble_add_matter_service`. In that function, we set a two values that don't change other than at init making them the equivalent of "constants. 

I'm not sure why this would require a global variable. Am i missing something?

bhmanda-silabs:
This change is not required

PR #36178, opened 2024-10-21
[Matter.framework] Move various #define to src/platform/Darwin/CHIPPl…

vivien-apple:
…atformConfig.h by defautl instead of relying on CHIPProjectAppConfig.h

#### Problem

Some build flags in `config/ios/CHIPProjectConfig.h` are now obsolete. Additionally, a few others could potentially be moved to `src/platform/Darwin/CHIPPlatformConfig.h` for better organization.


src/darwin/Framework/CHIP/include/CHIPProjectConfig.h
```diff
@@ -16,16 +17,9 @@
 
 /**
  *    @file
- *      Project configuration for Darwin Framework Tool.
+ *      CHIP project configuration for standalone builds.
```
bzbarsky-apple, line 20: Is this really about "standalone" builds?

src/darwin/Framework/CHIP/include/SystemProjectConfig.h
```diff
@@ -0,0 +1,25 @@
...
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+/**
+ *    @file
+ *      chip::System project configuration for standalone builds.
```
bzbarsky-apple, line 20: Again, how is this "standalone"?

src/darwin/Framework/Matter.xcodeproj/project.pbxproj
```diff
@@ -2431,6 +2447,8 @@
 					"MTR_NO_AVAILABILITY=1",
 					"MTR_ENABLE_PROVISIONAL=1",
 					"MTR_ENABLE_UNSTABLE_API=1",
+					"CONFIG_BUILD_FOR_HOST_UNIT_TEST=1",
+					"CHIP_DEVICE_CONFIG_ENABLE_TEST_SETUP_PARAMS=1",
```
bzbarsky-apple, line 2451: So this is adding those for the framework build in general, right?  Why do we want that?  I don't think we want CHIP_DEVICE_CONFIG_ENABLE_TEST_SETUP_PARAMS for the framework at all.  And ideally not CONFIG_BUILD_FOR_HOST_UNIT_TEST either.

src/platform/Darwin/CHIPPlatformConfig.h
```diff
@@ -53,9 +55,16 @@ extern "C" int __cxa_atexit(void (*f)(void *), void * p, void * d);
 #define CHIP_CONFIG_MAX_UNSOLICITED_MESSAGE_HANDLERS 8
 #endif // CHIP_CONFIG_MAX_UNSOLICITED_MESSAGE_HANDLERS
 
-#ifndef CHIP_CONFIG_MAX_EXCHANGE_CONTEXTS
-#define CHIP_CONFIG_MAX_EXCHANGE_CONTEXTS 8
-#endif // CHIP_CONFIG_MAX_EXCHANGE_CONTEXTS
+//
+// Default of 8 ECs is not sufficient for some of the unit tests
+// that try to validate multiple simultaneous interactions.
```
bzbarsky-apple, line 60: This should only matter when building those unit tests, right?

That said, in general we might in fact want more than 8 on darwin.....

src/platform/Darwin/CHIPPlatformConfig.h
```diff
@@ -66,5 +75,18 @@ extern "C" int __cxa_atexit(void (*f)(void *), void * p, void * d);
 #endif // CHIP_CONFIG_BDX_MAX_NUM_TRANSFERS
 
 #ifndef CHIP_CONFIG_KVS_PATH
+#if TARGET_OS_IPHONE
+#define CHIP_CONFIG_KVS_PATH "chip.store"
```
bzbarsky-apple, line 79: Shouldn't this be for all the cases that are not TARGET_OS_OSX or something?

config/ios/CHIPProjectConfig.h
```diff
@@ -25,28 +25,4 @@
  */
 #ifndef CHIPPROJECTCONFIG_H
 #define CHIPPROJECTCONFIG_H
-
```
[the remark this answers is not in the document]
ksperling-apple: Or can we just remove the file?

examples/tv-casting-app/darwin/args.gni
```diff
@@ -20,7 +20,7 @@ chip_device_project_config_include = "<CHIPProjectAppConfig.h>"
 chip_project_config_include = "<CHIPProjectAppConfig.h>"
 chip_system_project_config_include = "<SystemProjectConfig.h>"
 
-chip_project_config_include_dirs += [
+chip_project_config_include_dirs = [
```
ksperling-apple, line 23: Should this file not be importing `${chip_root}/config/standalone/args.gni` instead of `config/ios/args.gni`, and this line can remain `+=` ?

src/darwin/Framework/CHIP/include/CHIPProjectConfig.h
ksperling-apple: What's going to be in this file that's not in `Darwin/SystemPlatformConfig.h` ?

src/darwin/Framework/chip_xcode_build_connector.sh
```diff
@@ -122,6 +122,9 @@ declare -a args=(
     "target_cflags=$(format_gn_list "${target_cflags[@]}")"
     "target_cflags_c=$(format_gn_list "${target_cflags_c[@]}")"
     "target_cflags_cc=$(format_gn_list "${target_cflags_cc[@]}")"
+    'chip_project_config_include="<CHIPProjectConfig.h>"'
```
ksperling-apple, line 125: Rather than hard-coding more and more settings in the shell script, I think we should add an `args.gni` file for the framework that contains those settings that aren't derived by the script itself from xcodebuild environment variables.

src/platform/Darwin/CHIPPlatformConfig.h
```diff
@@ -66,5 +75,18 @@ extern "C" int __cxa_atexit(void (*f)(void *), void * p, void * d);
...
 #define CHIP_CONFIG_KVS_PATH "/tmp/chip_kvs"
+#endif // TARGET_OS_IPHONE
 #endif // CHIP_CONFIG_KVS_PATH
+
+#define CHIP_SYSTEM_CONFIG_PACKETBUFFER_POOL_SIZE 0
+
+// The session pool size limits how many subscriptions we can have live at
+// once.  Home supports up to 1000 accessories, and we subscribe to all of them,
+// so we need to make sure the pool is big enough for that.
+#define CHIP_CONFIG_SECURE_SESSION_POOL_SIZE 1000
```
ksperling-apple, line 90: Do we even need to hard-limit the number of secure sessions in this way? `SecureSessionTable::EvictAndAllocate` does this rather complicated eviction logic... can we just remove the limit and compile out the eviction code entirely? (Probably something for a follow-up change)

src/platform/Darwin/CHIPPlatformConfig.h
vivien-apple: It looks like this macro is being used to size an array as well.

from `src/lib/core/CHIPConfig.h`
```
#define CHIP_CONFIG_MCSP_RECEIVE_TABLE_SIZE (CHIP_CONFIG_MAX_EXCHANGE_CONTEXTS - 2)
```

from `src/protocols/secure_channel/MessageCounterManager.h`:
```
ReceiveTableEntry mReceiveTable[CHIP_CONFIG_MCSP_RECEIVE_TABLE_SIZE];
```

config/ios/CHIPProjectConfig.h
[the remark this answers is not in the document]
vivien-apple: I’ve removed that file.

src/darwin/Framework/CHIP/include/CHIPProjectConfig.h
vivien-apple: I’ve removed that file.

src/darwin/Framework/CHIP/include/SystemProjectConfig.h
vivien-apple: I’ve removed that file.

src/darwin/Framework/Matter.xcodeproj/project.pbxproj
vivien-apple: I’ve removed those definitions.

examples/tv-casting-app/darwin/args.gni
vivien-apple: I have updated this file to use `config/darwin/args.gni`.
And the line still need to be changed from `+=` to `=` since there is no definition of `chip_project_config_include_dirs` there.

src/platform/Darwin/CHIPPlatformConfig.h
vivien-apple: > When does this actually take effect? The default KVS isn't used from the Framework right? Also, what's the effect of having a relative paths here?

I assumed it might be used via `chip-tool`.

src/darwin/Framework/chip_xcode_build_connector.sh
vivien-apple: I have moved most those additional lines into `config/darwin/args.gni`
About moving the others, that looks like a different issue to me.

src/platform/Darwin/CHIPPlatformConfig.h
bzbarsky-apple: Followup at best; the secure session situation is complicated.

woody-apple (approved):

PR #36186, opened 2024-10-21
Generate Conformance Data with Alchemy

ethanzhouyc:
**Description:**
This PR adds conformance data to attributes, commands, and events in ZAP XMLs using automated generation from Alchemy.

**Generation Method:**
Since Alchemy does not provide a command to directly generate conformance, the changes in this PR were produced by generating the diff between running Alchemy commands with and without the conformance flag. As a result, some unrelated changes are also included: Alchemy removes attributes with default values inside `<attribute>` tags. They are redundant and will not affect XML parsing. For example: `optional="false"` and `min="0"` (for unsigned int types) are omitted.

**Validation and Discoveries:**
To validate the generation result, Paul and I spent an entire day on reviewing and comparing all conformance-related updates against the spec element by element to ensure accuracy. Here are key findings:

1. Conformance data for elements within the cluster extension of `color-control-cluster.xml` was not generated by Alchemy and has been manually added based on the spec.
2. An empty conformance field in the spec is translated into `mandatoryConform` in XML.
3. If the conformance field is `desc` in the spec, no conformance will be added to the XML. When combined with other conformance types, for example, `P, desc`, it translates to a `provisionalConform` inside `otherwiseConform`, and `desc` translates to nothing. It is wrong and should just be `provisionalConform`

### DIFF
--- a/src/app/zap-templates/zcl/data-model/chip/access-control-cluster.xml
+++ b/src/app/zap-templates/zcl/data-model/chip/access-control-cluster.xml
@@ -113,29 +113,58 @@ limitations under the License.
       <description>ACL</description>
       <access op="read" privilege="administer"/>
       <access op="write" privilege="administer"/>
+      <mandatoryConform/>
     </attribute>
 
     <attribute side="server" code="0x0001" define="EXTENSION" type="array" entryType="AccessControlExtensionStruct" writable="true" optional="true">
       <description>Extension</description>
       <access op="read" privilege="administer"/>
       <access op="write" privilege="administer"/>
+      <mandatoryConform>
+        <feature name="EXTS"/>
+      </mandatoryConform>
     </attribute>
 
-    <attribute side="server" code="0x0002" define="SUBJECTS_PER_ACCESS_CONTROL_ENTRY" type="int16u" min="4" default="4">SubjectsPerAccessControlEntry</attribute>
-    <attribute side="server" code="0x0003" define="TARGETS_PER_ACCESS_CONTROL_ENTRY" type="int16u" min="3" default="3">TargetsPerAccessControlEntry</attribute>
-    <attribute side="server" code="0x0004" define="ACCESS_CONTROL_ENTRIES_PER_FABRIC" type="int16u" min="4" default="4">AccessControlEntriesPerFabric</attribute>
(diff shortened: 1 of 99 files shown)


andy31415 (approved):

jmartinez-silabs (approved):

src/app/zap-templates/zcl/data-model/chip/access-control-cluster.xml
```diff
@@ -113,29 +113,58 @@ limitations under the License.
...
+      </mandatoryConform>
     </attribute>
 
-    <attribute side="server" code="0x0002" define="SUBJECTS_PER_ACCESS_CONTROL_ENTRY" type="int16u" min="4" default="4">SubjectsPerAccessControlEntry</attribute>
-    <attribute side="server" code="0x0003" define="TARGETS_PER_ACCESS_CONTROL_ENTRY" type="int16u" min="3" default="3">TargetsPerAccessControlEntry</attribute>
-    <attribute side="server" code="0x0004" define="ACCESS_CONTROL_ENTRIES_PER_FABRIC" type="int16u" min="4" default="4">AccessControlEntriesPerFabric</attribute>
-    <attribute code="0x0005" side="server" define="COMMISSIONING_ARL" type="array" entryType="CommissioningAccessRestrictionEntryStruct" optional="true">CommissioningARL</attribute>
-    <attribute code="0x0006" side="server" define="ARL" type="array" entryType="AccessRestrictionEntryStruct" optional="true">ARL</attribute>
-
+    <attribute side="server" code="0x0002" define="SUBJECTS_PER_ACCESS_CONTROL_ENTRY" type="int16u" min="4" default="4">
```
bzbarsky-apple, line 128: I raised this on Slack, but this merged before I had a chance to comment on the PR @jmartinez-silabs @andy31415 @ethanzhouyc :

This attribute syntax is because it is really hard to tell where the attribute name should go.  But now we forced it on all attributes, but adding lots of `<mandatoryConform/>` everywhere instead of just assuming M conformance if none is listed.  Ideally we would just not have the noise, but if we really have to, we need to make the syntax make some sort of sense: the attribute name is not a "description".

andy31415: Rolled back in #36220

PR #30965, opened 2023-12-12
Not call onError in BufferedReadCallback

yunhanw-google:
If onError has been called from BufferedReadCallback when processing multiple attrbitues, potentially onError or onAttribute or onEvent could be called again from ReadClient, as a result, the behavior in application could be unexpected, crash is possible. onError should be called only with  close path from readClient.

PR #36187, opened 2024-10-21
[IM] Handle internal fatal error in BufferReadCallback

yunhanw-google:
Revive https://github.com/project-chip/connectedhomeip/pull/30965

Issue: If onError has been called from BufferedReadCallback when processing multiple attrbitues, potentially onError or onAttribute or onEvent could be called again from ReadClient, as a result, the behavior in application could be unexpected, crash is possible. note: the errors in BufferredReadCallback are treated as fatal ones.

Solution: In order to resolve the above issue, we add mDataBufferingError to store internal error in BufferredReadCallback, we call a new introduced GetLastError callback in readClient after each onAttribute and onEvent, if any fatal error happens, we propagate the error to Close in ReadClient, for any new callback class that inherits ReadClient::Callback, if there is a fatal error, ReadClient can get aware of this error from GetLastError.

### DIFF
--- a/src/app/BufferedReadCallback.cpp
+++ b/src/app/BufferedReadCallback.cpp
@@ -37,12 +37,8 @@ void BufferedReadCallback::OnReportBegin()
 void BufferedReadCallback::OnReportEnd()
 {
     CHIP_ERROR err = DispatchBufferedData(mBufferedPath, StatusIB(), true);
-    if (err != CHIP_NO_ERROR)
-    {
-        mCallback.OnError(err);
-        return;
-    }
-
+    SetFatalReportError(err);
+    ReturnOnFailure(err);
     mCallback.OnReportEnd();
 }
 
@@ -230,13 +226,12 @@ CHIP_ERROR BufferedReadCallback::DispatchBufferedData(const ConcreteAttributePat
 void BufferedReadCallback::OnAttributeData(const ConcreteDataAttributePath & aPath, TLV::TLVReader * apData,
                                            const StatusIB & aStatus)
 {
-    CHIP_ERROR err;
-
     //
     // First, let's dispatch to our registered callback any buffered up list data from previous calls.
     //
-    err = DispatchBufferedData(aPath, aStatus);
-    SuccessOrExit(err);
+    CHIP_ERROR err = DispatchBufferedData(aPath, aStatus);
+    SetFatalReportError(err);
+    ReturnOnFailure(err);
 
     //
     // We buffer up list data (only if the status was successful)
@@ -244,7 +239,8 @@ void BufferedReadCallback::OnAttributeData(const ConcreteDataAttributePath & aPa
     if (aPath.IsListOperation() && aStatus.mStatus == Protocols::InteractionModel::Status::Success)
     {
         err = BufferData(aPath, apData);
-        SuccessOrExit(err);
+        SetFatalReportError(err);
+        ReturnOnFailure(err);
     }
     else
     {
@@ -255,12 +251,6 @@ void BufferedReadCallback::OnAttributeData(const ConcreteDataAttributePath & aPa
     // Update our latched buffered path.
     //
     mBufferedPath = aPath;
-
-exit:
-    if (err != CHIP_NO_ERROR)
-    {
-        mCallback.OnError(err);
-    }
 }
 
 } // namespace app
--- a/src/app/BufferedReadCallback.h
+++ b/src/app/BufferedReadCallback.h
@@ -44,7 +44,7 @@ public:
 
 private:
     /*
-     * Generates the reconsistuted TLV array from the stored individual list elements
+     * Generates the reconstituted TLV array from the stored individual list elements
      */
     CHIP_ERROR GenerateListTLV(TLV::ScopedBufferTLVReader & reader);
 
--- a/src/app/ReadClient.h
+++ b/src/app/ReadClient.h
@@ -202,6 +202,8 @@ public:
          * This object MUST continue to exist after this call is completed. The application shall wait until it
          * receives an OnDone call to destroy the object.
          *
+         * This callback is expected to be called inside the ReadClient.
+         *
          * @param[in] aError       A system error code that conveys the overall error code.
          */
         virtual void OnError(CHIP_ERROR aError) {}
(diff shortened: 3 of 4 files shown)


src/app/ReadClient.h
```diff
@@ -284,6 +284,11 @@ class ReadClient : public Messaging::ExchangeDelegate
          * things like min/max intervals based on the session parameters).
          */
         virtual void OnCASESessionEstablished(const SessionHandle & aSession, ReadPrepareParams & aSubscriptionParams) {}
+
+        /*
+         * return the internal fatal error in callback
+         */
+        virtual CHIP_ERROR CheckInternalError() { return CHIP_NO_ERROR; }
```
andy31415, line 291: The name is `Check` and description says `returns the internal....`.

Could we rename as this does not check anything? Something like `GetProcessingError` or `GetLastError` and update the description explaining when this is set and what it represents. the `internal fatal error` is not as clear.

Also this seems to be `const`

src/app/ReadClient.h
```diff
@@ -284,6 +284,11 @@ class ReadClient : public Messaging::ExchangeDelegate
          * things like min/max intervals based on the session parameters).
          */
         virtual void OnCASESessionEstablished(const SessionHandle & aSession, ReadPrepareParams & aSubscriptionParams) {}
+
+        /*
+         * Get the last internal fatal error in callback
+         */
+        virtual CHIP_ERROR GetLastError() const { return CHIP_NO_ERROR; }
```
bzbarsky-apple, line 291: Why is this new API needed?  This is making all ReadClient consumers pay extra flash costs for an internal implementation details of BufferedReadCallback (which is meant for "capable clients" only, and is better able to shoulder such costs).

If this is trying to implement the suggestion from https://github.com/project-chip/connectedhomeip/pull/30965#discussion_r1424596155, then that's not what the suggestion was.  The suggestion was that BufferedReadCallback store some internal state that will allow it to maintain the relevant invariants and just start ignoring calls from ReadClient after it reaches that error state, no?  As far as the ReadClient is concerned, there would be no changes in behavior of any sort here.

bzbarsky-apple (changes requested):

src/app/BufferedReadCallback.cpp
```diff
@@ -36,13 +36,8 @@ void BufferedReadCallback::OnReportBegin()
...
 {
-    CHIP_ERROR err = DispatchBufferedData(mBufferedPath, StatusIB(), true);
-    if (err != CHIP_NO_ERROR)
-    {
-        mCallback.OnError(err);
-        return;
-    }
-
+    mLastError = DispatchBufferedData(mBufferedPath, StatusIB(), true);
+    ReturnOnFailure(mLastError);
```
bzbarsky-apple, line 40: OK, so this skips calling OnReportEnd and removes the OnError call.  But API guarantees either OnReportEnd or OnError, no?  I don't understand how this is supposed to work.

That said, in the _subscription_ case OnError means something quite specific which does not match what this code ends up doing if we did call OnError.

So what is the actual behavior we are trying for here?

src/app/BufferedReadCallback.h
```diff
@@ -131,6 +132,7 @@ class BufferedReadCallback : public ReadClient::Callback
     ConcreteDataAttributePath mBufferedPath;
     std::vector<System::PacketBufferHandle> mBufferedList;
     Callback & mCallback;
+    CHIP_ERROR mLastError = CHIP_NO_ERROR;
```
bzbarsky-apple, line 135: This should probably be called mDataBufferingError?

src/app/BufferedReadCallback.h
```diff
@@ -74,6 +74,7 @@ class BufferedReadCallback : public ReadClient::Callback
     void OnError(CHIP_ERROR aError) override
     {
         mBufferedList.clear();
+        mLastError = CHIP_NO_ERROR;
```
bzbarsky-apple, line 77: Why?

src/app/BufferedReadCallback.h
```diff
@@ -74,6 +74,7 @@ class BufferedReadCallback : public ReadClient::Callback
     void OnError(CHIP_ERROR aError) override
```
bzbarsky-apple, line 74: Somewhere in here OnEventData should also do nothing if we have errored, no?

src/app/BufferedReadCallback.h
yunhanw-google: done, thanks

src/app/BufferedReadCallback.h
yunhanw-google: sorry, I forgot to push latest change, could you re-review?

src/app/BufferedReadCallback.h
yunhanw-google: sorry, I forgot to push latest change, could you re-review? this has been removed

src/app/BufferedReadCallback.cpp
yunhanw-google: sorry, I forgot to push latest change, could you re-review?

src/app/ReadClient.h
yunhanw-google: just updated following the suggestion here.

src/app/BufferedReadCallback.h
bzbarsky-apple: I see no changes to OnEventData in what's here right now, 5 hours later...

src/app/BufferedReadCallback.cpp
bzbarsky-apple: Again: "what is the actual behavior we are trying for here?"  For reads, and for subscriptions (which may want different behaviors, because OnError has very different meanings for them).

src/app/BufferedReadCallback.h
yunhanw-google: yes, onEventData needs to be skipped as well when there is the internal error, thanks

src/app/ReadClient.h
yunhanw-google: if we introduce GetLastError as new callback in ReadClient, we check GetLastError in the end of onAttributeData or onEventData in ReadClient, no matter what kind of chained callback we have, we could lead to Close call in readClient for read or subscribe, for resubscription case,  it would depend on which error OnResubscriptionNeeded returns, OnResubscriptionNeeded is not only conveying this error to contrloller applicaiton, but also decides how to translate the error, for example, application can skip all errors, or application can treat the errors except timeout as no error, in all situations, onError would only be called inside ReadClient, only inside Close, we may also update comments for onError, and clearly mention onError is not expected called outside ReadClient

PR #36195, opened 2024-10-22
Add a BigEndian::Reader to be the equivalent for LittleEndian::Reader

andy31415:
I will need the ability to handle both little and big endian logic since ember data store is in native-endianness. Adding bigendian readers as a result (we already have writers).

Only a subset of methods is added/supported for now, can grow as needed.

Also minor re-arrangement of logic, which seems to result in minute flash savings.

### DIFF
--- a/src/lib/support/BufferReader.h
+++ b/src/lib/support/BufferReader.h
@@ -396,8 +396,8 @@ public:
      * Create a buffer reader from a given byte span.
      *
      * @param buffer The octet buffer byte span from which to read.  The caller must ensure
-     *               that the buffer outlives the reader.  The buffer's ByteSpan .data() pointer
-     *               is is nullptr, length is automatically overridden to zero, to avoid accesses.
+     *               that the buffer outlives the reader.  If the buffer's ByteSpan .data() pointer
+     *               is nullptr, length is automatically overridden to zero, to avoid accesses.
      */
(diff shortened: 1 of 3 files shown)


tcarmelveilleux (approved):

yunhanw-google (approved):

src/lib/support/BufferReader.h
```diff
@@ -290,46 +367,80 @@ class Reader
...
+    Reader(const uint8_t * buffer, size_t buf_len) : BufferReader(buffer, buf_len) {}
 
-private:
     /**
-     * Our buffer start.
+     * Create a buffer reader from a given byte span.
+     *
+     * @param buffer The octet buffer byte span from which to read.  The caller must ensure
+     *               that the buffer outlives the reader.  The buffer's ByteSpan .data() pointer
+     *               is is nullptr, length is automatically overridden to zero, to avoid accesses.
```
bzbarsky-apple, line 400:
```suggestion
     *               that the buffer outlives the reader.  If the buffer's ByteSpan .data() pointer
     *               is nullptr, length is automatically overridden to zero, to avoid accesses.
```

bzbarsky-apple (approved):

jmartinez-silabs (approved):

tehampson (approved):

PR #36204, opened 2024-10-22
Camera AV Stream Management Server Impl

pidarped:
* Delegate definition
  - Support for Command handlers
  - Helper interfaces for loading of pre-allocated streams from storage

* Attributes support
  - Read/Write support
  - Setters/Getters for all attributes
  - Storage handling for most attributes, except streams.
  
* Add attributes to json files

* Autogen code changes to zzz_generated/


src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1801 @@
...
+CHIP_ERROR CameraAVStreamMgmtServer::StoreViewport(const ViewportStruct & viewport)
+{
+    uint8_t buffer[kViewportStructMaxSerializedSize];
+    TLV::TLVWriter writer;
+
+    writer.Init(buffer);
+    ReturnErrorOnFailure(viewport.Encode(writer, TLV::AnonymousTag()));
+
+    return mPersistentStorage->SyncSetKeyValue(DefaultStorageKeyAllocator::CameraAVStreamMgmtViewport().KeyName(), buffer,
+                                               static_cast<uint16_t>(writer.GetLengthWritten()));
```
bzbarsky-apple, line 1347: I do wonder: is there a reason to not use the key format that the other attributes (via SafeAttributePersistenceProvider) use?

And the reason I ask that is this: as soon as you have two instances of this cluster on different endpoints, this setup breaks completely, with the global key name you have here.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1801 @@
...
+}
+
+CHIP_ERROR CameraAVStreamMgmtServer::ClearViewport()
+{
+    // TODO
```
bzbarsky-apple, line 1352: Needs to be at least tracked in an issue that is referenced here.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1801 @@
...
+}
+
+CHIP_ERROR CameraAVStreamMgmtServer::StoreRankedVideoStreamPriorities()
+{
+    uint8_t buffer[kRankedVideoStreamPrioritiesTlvSize];
+    TLV::TLVWriter writer;
+    writer.Init(buffer);
+
+    TLV::TLVType arrayType;
+    ReturnErrorOnFailure(writer.StartContainer(TLV::AnonymousTag(), TLV::kTLVType_Array, arrayType));
```
bzbarsky-apple, line 1382: This, and the way viewport is stored, assumes there will never be format/versioning changes, right?  Is that a good assumption?

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1801 @@
...
+CHIP_ERROR CameraAVStreamMgmtServer::LoadRankedVideoStreamPriorities()
+{
+    uint8_t buffer[kRankedVideoStreamPrioritiesTlvSize];
+
+    uint16_t len = kRankedVideoStreamPrioritiesTlvSize;
+    ReturnErrorOnFailure(mPersistentStorage->SyncGetKeyValue(
+        DefaultStorageKeyAllocator::CameraAVStreamMgmtRankedVideoStreamPriorities().KeyName(), buffer, len));
+
+    TLV::TLVReader reader;
+    reader.Init(buffer);
```
bzbarsky-apple, line 1404: Shouldn't this somehow take `len` into account?  The viewport codepath did this right.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1801 @@
...
+    {
+        if (reader.Next(TLV::kTLVType_UnsignedInteger, TLV::AnonymousTag()) != CHIP_NO_ERROR)
+        {
+            break;
+        }
+
+        ReturnErrorOnFailure(reader.Get(mRankedVideoStreamPriorities[i]));
+    }
+
+    ReturnErrorOnFailure(reader.ExitContainer(arrayType));
```
bzbarsky-apple, line 1420: Whoa, wait.  If our Next() failed because the TLV was malformed, we need to return at that point.  Next() is only allowed to return "end of TLV" or success for us to keep going.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1801 @@
...
+    bool isWaterMarkEnabled = false;
+    bool isOSDEnabled       = false;
+
+    if (HasFeature(Feature::kWatermark))
+    {
+        isWaterMarkEnabled = commandData.watermarkEnabled.ValueOr(false);
+    }
+    else
+    {
+        VerifyOrExit(!commandData.watermarkEnabled.HasValue(), status = Status::InvalidCommand;
```
bzbarsky-apple, line 1584: I strongly suggest not using "go to exit".  Have this function return a status and have the caller call AddStatus on failure?

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1801 @@
...
+    auto & minBitRate       = commandData.minBitRate;
+    auto & maxBitRate       = commandData.maxBitRate;
+    auto & minFragmentLen   = commandData.minFragmentLen;
+    auto & maxFragmentLen   = commandData.maxFragmentLen;
+    bool isWaterMarkEnabled = false;
+    bool isOSDEnabled       = false;
+
+    if (HasFeature(Feature::kWatermark))
+    {
+        isWaterMarkEnabled = commandData.watermarkEnabled.ValueOr(false);
```
bzbarsky-apple, line 1580: This allows invalid commands where we have the WMARK feature but the command is missing the field.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1801 @@
...
+    }
+    else
+    {
+        VerifyOrExit(!commandData.watermarkEnabled.HasValue(), status = Status::InvalidCommand;
+                     ChipLogError(Zcl, "CameraAVStreamMgmt: Failed to allocate Video stream. Unsupported watermark feature"));
+    }
+
+    if (HasFeature(Feature::kOnScreenDisplay))
+    {
+        isOSDEnabled = commandData.OSDEnabled.ValueOr(false);
```
bzbarsky-apple, line 1590: Again, this allows invalid commands.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1801 @@
...
+    }
+    else
+    {
+        VerifyOrExit(!commandData.OSDEnabled.HasValue(), status = Status::InvalidCommand;
+                     ChipLogError(Zcl, "CameraAVStreamMgmt: Failed to allocate Video stream. Unsupported OnScreenDisplay feature"));
+    }
+
+    // Call the delegate
```
bzbarsky-apple, line 1598: What happened to verifying that the passed-in enums have valid valies?  That the min/max constraints of all sorts are satisfied?  All of that needs to happen here, because relying on the delegate to do that checking means it won't happen.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1801 @@
...
+    status =
+        mDelegate.VideoStreamAllocate(streamUsage, videoCodec, minFrameRate, maxFrameRate, minResolution, maxResolution, minBitRate,
+                                      maxBitRate, minFragmentLen, maxFragmentLen, isWaterMarkEnabled, isOSDEnabled);
+
+    if (status == Status::Success)
+    {
+        ctx.mCommandHandler.AddResponse(ctx.mRequestPath, response);
```
bzbarsky-apple, line 1605: I don't get it.  Where did `response.videoStreamID` get set?

Was this tested at all?

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1801 @@
...
+                                                       const Commands::VideoStreamModify::DecodableType & commandData)
+{
+    Status status           = Status::Success;
+    bool isWaterMarkEnabled = false;
+    bool isOSDEnabled       = false;
+    auto & videoStreamID    = commandData.videoStreamID;
+
+    if (HasFeature(Feature::kWatermark))
+    {
+        isWaterMarkEnabled = commandData.watermarkEnabled.ValueOr(false);
```
bzbarsky-apple, line 1623: This has all the same bugs as VideoStreamAllocate in terms of allowing invalid commands, not validating enums, etc.

I'm stopping here in the .cpp.  Please fix the issues pointed out so far, including in the code I have not read yet.

bzbarsky-apple (changes requested):

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1801 @@
...
+
+CHIP_ERROR CameraAVStreamMgmtServer::RemoveFromFabricsUsingCamera(chip::FabricIndex aFabricIndex)
+{
+    mFabricsUsingCamera.erase(aFabricIndex);
+    return CHIP_NO_ERROR;
+}
+
+CHIP_ERROR CameraAVStreamMgmtServer::SetRankedVideoStreamPriorities(const StreamUsageEnum newPriorities[kNumOfStreamUsageTypes])
+{
+    std::copy(newPriorities, newPriorities + kNumOfStreamUsageTypes, mRankedVideoStreamPriorities);
```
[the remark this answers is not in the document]
pidarped: Replaced.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.h
```diff
@@ -0,0 +1,627 @@
...
+     *
+     *   @return Success if the allocation is successful and a VideoStreamID was
+     *   produced; otherwise, the command SHALL be rejected with an appropriate
+     *   error.
+     */
+    virtual Protocols::InteractionModel::Status
+    VideoStreamAllocate(StreamUsageEnum streamUsage, CameraAvStreamManagement::VideoCodecEnum videoCodec,
+                        const uint16_t minFrameRate, const uint16_t maxFrameRate, VideoResolutionStruct minResolution,
+                        VideoResolutionStruct maxResolution, const uint32_t minBitRate, const uint32_t maxBitRate,
+                        const uint16_t minFragmentLen, const uint16_t maxFragmentLen, bool waterMarkEnabled, bool osdEnabled) = 0;
```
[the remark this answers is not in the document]
pidarped: I presume for this call you are talking about the WMARK and OSD features. The checks for the feature are done in the handler before calling this delegate function. If feature is not supported these would default to false.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.h
```diff
@@ -0,0 +1,627 @@
...
+    /**
+     *   @brief Handle Command Delegate for SetStreamPriorities.
+     *
+     *   @param streamPriorities    Indicates the list of stream priorities to set, in decreasing order of priorities.
+     *
+     *   @return Success if the setting of stream priorities is successful; otherwise, the command SHALL be rejected with an
+     * appropriate error.
+     */
+    virtual Protocols::InteractionModel::Status
+    SetStreamPriorities(const DataModel::DecodableList<StreamUsageEnum> streamPriorities) = 0;
```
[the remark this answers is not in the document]
ksperling-apple: `Span<StreamUsageEnum>` ?

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.h
```diff
@@ -0,0 +1,656 @@
...
+     *  The delegate application is responsible for creating and persisting
+     *  these streams(based on the Allocation commands). These Load APIs would be
+     *  used to load the pre-allocated stream context information into the cluster server list,
+     *  at initialization.
+     *  Once loaded, the cluster server would be serving Reads on these
+     *  attributes. The list is updatable via the Add/Remove functions for the
+     *  respective streams.
+     */
+    virtual Protocols::InteractionModel::Status
+    LoadAllocatedVideoStreams(std::vector<VideoStreamStruct> & allocatedVideoStreams) = 0;
```
ksperling-apple, line 254: Has there been a conscious decision to use `std::vector` in this code? Most core code avoids using it.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.h
```diff
@@ -0,0 +1,627 @@
...
+constexpr uint8_t kMaxMicrophoneLevel       = 254;
+constexpr uint16_t kMaxImageRotationDegrees = 359;
+
+constexpr size_t kViewportStructMaxSerializedSize = TLV::EstimateStructOverhead(sizeof(uint16_t) * 4);
+
+// The number of different use case types for the streams(Recording, Liveview, etc).
+constexpr size_t kNumOfStreamUsageTypes = 4;
+
+// StreamUsageEnum + Anonymous tag(1 byte)
+constexpr size_t kStreamUsageTlvSize = sizeof(StreamUsageEnum) + 1;
```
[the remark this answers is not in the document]
pidarped: To avoid discrepancy, I am thinking of using a sizeof(uint8_t) for the StreamUsageEnum.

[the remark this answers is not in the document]
bzbarsky-apple: No, that's not the point.  The point is that technically nothing requires a number that fits in uint8 to be encoded as one byte in the TLV.  Our TLV encoder does happen to do that, and if we are depending on that, we should probably fail out during encoding if that's violated.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.h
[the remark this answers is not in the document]
bzbarsky-apple: My question is why we would pass false to the delegate instead of correctly representing the "not present" state?

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.h
```diff
@@ -0,0 +1,627 @@
...
+     * to be registered and called by the interaction model at the appropriate times.
+     *
+     * @param aDelegate                         A pointer to the delegate to be used by this server.
+     *                                          Note: the caller must ensure that the delegate lives throughout the instance's
+     *                                          lifetime.
+     *
+     * @param aEndpointId                       The endpoint on which this cluster exists. This must match the zap configuration.
+     * @param aClusterId                        The ID of the Camera AV Stream Management cluster to be instantiated.
+     * @param aFeature                          The bitmask value that identifies which features are supported by this instance.
+     * @param aOptionalAttrs                    The bitmask value that identifies the optional attributes supported by this
```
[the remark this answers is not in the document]
bzbarsky-apple: Because in the spec a "mapN" doesn't mean each bit is independent; some mapN in the spec use two bits to represent stuff.  So you can't use BitFlags for spec mapN when generating code, because you don't know the details of those semantics....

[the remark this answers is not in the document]
bzbarsky-apple: Why wouldn't we just fail out of Init on invalid input, especially since the API docs say we do?

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.h
```diff
@@ -0,0 +1,627 @@
...
+
+    CHIP_ERROR AddSnapshotStream(const SnapshotStreamStruct & snapshotStream);
+
+    CHIP_ERROR RemoveSnapshotStream(uint16_t snapshotStreamId);
+
+private:
+    CameraAVStreamMgmtDelegate & mDelegate;
+    EndpointId mEndpointId;
+    ClusterId mClusterId;
+    BitMask<Feature> mFeature;
```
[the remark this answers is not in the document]
pidarped: Have made the BitFlags to be const.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
pidarped: Removed TODO and added the call to delete key.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.h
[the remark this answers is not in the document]
pidarped: Converted to optional args and passed as-is to the delegate.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.h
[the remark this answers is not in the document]
pidarped: I realized that I had added the stream priorities as a member array. So, no need to pass to the delegate. Instead, there is a notification callback that is invoked on the delegate when the stream priorities is set via the Command.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1801 @@
...
+    AttributeAccessInterfaceRegistry::Instance().Unregister(this);
+}
+
+CHIP_ERROR CameraAVStreamMgmtServer::Init()
+{
+    LoadPersistentAttributes();
+
+    VerifyOrReturnError(AttributeAccessInterfaceRegistry::Instance().Register(this), CHIP_ERROR_INTERNAL);
+    ReturnErrorOnFailure(CommandHandlerInterfaceRegistry::Instance().RegisterCommandHandler(this));
+    return CHIP_NO_ERROR;
```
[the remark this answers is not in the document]
pidarped: Added the feature and optional attribute checks.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
pidarped: Added sanity checks for args for all the commands.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.h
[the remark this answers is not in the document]
pidarped: Added optional bitflag checks along with the feature checks in the Init function.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.h
pidarped: I have used vectors here mainly for ease of use for managing dynamic lists, and especially, since cameras are expected to be relatively resourceful. What is the preferred approach to represent dynamic lists that are owned?

yufengwangca: If the objects are very large/non-copyable, then we might use std::vector<std::unique_ptr<VideoStreamStruct>>?
 
std::unique_ptr clearly expresses unique ownership: the vector “owns” the objects. This helps avoid memory leaks and clarifies that nothing else should be responsible for those objects’ lifetimes.

yunhanw-google (commented):
```
2025-02-12 19:21:41 fcd452e3b655 PythonMatterControllerTEST[4749] ERROR Unexpected error occurred when running tests: Cannot decode value for path 2/<class \'chip.clusters.Objects.CameraAvStreamManagement.Attributes.SupportedSnapshotParams\'>, got error: \'InteractionModelError: Failure (0x1)\', raw TLV data: \'None\'
2025-02-12 19:21:41 fcd452e3b655 PythonMatterControllerTEST[4749] ERROR Cannot decode value for path 2/<class \'chip.clusters.Objects.CameraAvStreamManagement.Attributes.SupportedSnapshotParams\'>, got error: \'InteractionModelError: Failure (0x1)\', raw TLV data: \'None\'
Traceback...
```

From cirque mobile test, we see the above error, it looks there is decoding error for CameraAvStreamManagement.Attributes.SupportedSnapshotParam

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1765 @@
...
+
+    VerifyOrReturn(minFrameRate <= maxFrameRate, ctx.mCommandHandler.AddStatus(ctx.mRequestPath, Status::ConstraintError));
+
+    VerifyOrReturn(minBitRate <= maxBitRate, ctx.mCommandHandler.AddStatus(ctx.mRequestPath, Status::ConstraintError));
+
+    VerifyOrReturn(minFragmentLen <= maxFragmentLen, ctx.mCommandHandler.AddStatus(ctx.mRequestPath, Status::ConstraintError));
+
+    // Call the delegate
+    status =
+        mDelegate.VideoStreamAllocate(streamUsage, videoCodec, minFrameRate, maxFrameRate, minResolution, maxResolution, minBitRate,
```
bzbarsky-apple, line 1504: Where is the constraint validation for fields of minResolution/maxResolution?

bzbarsky-apple (changes requested):

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1765 @@
...
+CHIP_ERROR CameraAVStreamMgmtServer::StoreViewport(const ViewportStruct & viewport)
+{
+    uint8_t buffer[kViewportStructMaxSerializedSize];
+    TLV::TLVWriter writer;
+
+    writer.Init(buffer);
+    ReturnErrorOnFailure(viewport.Encode(writer, TLV::AnonymousTag()));
+
+    StorageKeyName key =
+        DefaultStorageKeyAllocator::AttributeValue(mEndpointId, CameraAvStreamManagement::Id, Attributes::Viewport::Id);
```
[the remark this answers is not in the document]
pidarped: Didn't see the SafeWriteValue/SafeReadValue() APIs earlier for writing/reading TLV encoded bytespans. Have changed to that. Thanks.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1765 @@
...
+    return mPersistentStorage->SyncSetKeyValue(key.KeyName(), buffer, static_cast<uint16_t>(writer.GetLengthWritten()));
+}
+
+CHIP_ERROR CameraAVStreamMgmtServer::ClearViewport()
```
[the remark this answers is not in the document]
pidarped: I initially thought I would clear before storing but it seems it is not required. Removed it.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1765 @@
...
+    TLV::TLVReader reader;
+    reader.Init(bufferSpan);
+
+    ReturnErrorOnFailure(reader.Next(TLV::kTLVType_Array, TLV::AnonymousTag()));
+    TLV::TLVType arrayType;
+    ReturnErrorOnFailure(reader.EnterContainer(arrayType));
+
+    for (uint8_t i = 0; i < kNumOfStreamUsageTypes; i++)
+    {
+        ReturnErrorOnFailure(reader.Next(TLV::kTLVType_UnsignedInteger, TLV::AnonymousTag()));
```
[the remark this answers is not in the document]
pidarped: Have changed the attribute to be a vector. There are some changes that are coming in the Spec where the device would have a SupportedStreamUsages attribute and we will check against that when processing this command. So, changing it to a vector now, and will be updating with those checks when the spec changes land.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1765 @@
...
+
+    auto & streamPriorities = commandData.streamPriorities;
+    // auto & streamPriorities = commandData.streamPriorities;
+
+    auto iter = streamPriorities.begin();
+    StreamUsageEnum rankedStreamPriorities[kNumOfStreamUsageTypes];
+    int i = 0;
+    while (iter.Next())
+    {
+        auto & streamUsage          = iter.GetValue();
```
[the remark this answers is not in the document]
pidarped: Added check.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1765 @@
...
+        rankedStreamPriorities[i++] = streamUsage;
```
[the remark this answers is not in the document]
pidarped: Changed to a vector and not a fixed sized array.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1765 @@
...
+    }
+
+    VerifyOrReturn(i == kNumOfStreamUsageTypes, {
```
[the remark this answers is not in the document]
pidarped: Changed to a vector for storing the StreamUsageEnums, which can have less number of items than kNumOfStreamUsageTypes. When the new Spec attributed of SupportedStreamUsages lands, we would need to add checks against that when this command is processed.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1765 @@
...
+
+    ctx.mCommandHandler.AddStatus(ctx.mRequestPath, Protocols::InteractionModel::Status::Success);
+}
+
+void CameraAVStreamMgmtServer::HandleSetStreamPriorities(HandlerContext & ctx,
+                                                         const Commands::SetStreamPriorities::DecodableType & commandData)
+{
+
+    auto & streamPriorities = commandData.streamPriorities;
+    // auto & streamPriorities = commandData.streamPriorities;
```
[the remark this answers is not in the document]
pidarped: Removed spurious commented line.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1850 @@
...
+        break;
+    case Viewport::Id:
+        VerifyOrReturnError(
+            HasFeature(Feature::kVideo), CHIP_IM_GLOBAL_STATUS(UnsupportedAttribute),
+            ChipLogError(Zcl, "CameraAVStreamMgmt[ep=%d]: can not get Viewport, feature is not supported", mEndpointId));
+        ReturnErrorOnFailure(aEncoder.Encode(mViewport));
+        break;
+    case SpeakerMuted::Id:
+        VerifyOrReturnError(
+            HasFeature(Feature::kAudio) && HasFeature(Feature::kSpeaker), CHIP_IM_GLOBAL_STATUS(UnsupportedAttribute),
```
bzbarsky-apple, line 528: Just check Speaker.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1850 @@
...
+            ChipLogError(Zcl, "CameraAVStreamMgmt[ep=%d]: can not get SpeakerMuted, feature is not supported", mEndpointId));
+        ReturnErrorOnFailure(aEncoder.Encode(mSpeakerMuted));
+        break;
+    case SpeakerVolumeLevel::Id:
+        VerifyOrReturnError(
+            HasFeature(Feature::kAudio) && HasFeature(Feature::kSpeaker), CHIP_IM_GLOBAL_STATUS(UnsupportedAttribute),
```
bzbarsky-apple, line 534: Just check Speaker.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1850 @@
...
+            ChipLogError(Zcl, "CameraAVStreamMgmt[ep=%d]: can not get SpeakerVolumeLevel, feature is not supported", mEndpointId));
+        ReturnErrorOnFailure(aEncoder.Encode(mSpeakerVolumeLevel));
+        break;
+    case SpeakerMaxLevel::Id:
+        VerifyOrReturnError(
+            HasFeature(Feature::kAudio) && HasFeature(Feature::kSpeaker), CHIP_IM_GLOBAL_STATUS(UnsupportedAttribute),
```
bzbarsky-apple, line 540: Just check Speaker.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1850 @@
...
+            ChipLogError(Zcl, "CameraAVStreamMgmt[ep=%d]: can not get SpeakerMaxLevel, feature is not supported", mEndpointId));
+        ReturnErrorOnFailure(aEncoder.Encode(mSpeakerMaxLevel));
+        break;
+    case SpeakerMinLevel::Id:
+        VerifyOrReturnError(
+            HasFeature(Feature::kAudio) && HasFeature(Feature::kSpeaker), CHIP_IM_GLOBAL_STATUS(UnsupportedAttribute),
```
bzbarsky-apple, line 546: Just check Speaker.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1850 @@
...
+}
+
+CHIP_ERROR CameraAVStreamMgmtServer::Write(const ConcreteDataAttributePath & aPath, AttributeValueDecoder & aDecoder)
+{
+    VerifyOrDie(aPath.mClusterId == CameraAvStreamManagement::Id);
+
+    switch (aPath.mAttributeId)
+    {
+    case HDRModeEnabled::Id: {
+        // Optional Attribute if Video is supported
```
bzbarsky-apple, line 639: Not in the spec PR this is claiming to implement.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1850 @@
...
+        VerifyOrReturnError(
+            HasFeature(Feature::kVideo), CHIP_IM_GLOBAL_STATUS(UnsupportedAttribute),
+            ChipLogError(Zcl, "CameraAVStreamMgmt[ep=%d]: can not set Viewport, feature is not supported", mEndpointId));
+        ViewportStruct viewPort;
+        ReturnErrorOnFailure(aDecoder.Decode(viewPort));
+        return SetViewport(viewPort);
+    }
+    case SpeakerMuted::Id: {
+        VerifyOrReturnError(
+            HasFeature(Feature::kAudio) && HasFeature(Feature::kSpeaker), CHIP_IM_GLOBAL_STATUS(UnsupportedAttribute),
```
bzbarsky-apple, line 697: Just Speaker.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1850 @@
...
+            ChipLogError(Zcl, "CameraAVStreamMgmt[ep=%d]: can not set SpeakerMuted, feature is not supported", mEndpointId));
+        bool speakerMuted;
+        ReturnErrorOnFailure(aDecoder.Decode(speakerMuted));
+        return SetSpeakerMuted(speakerMuted);
+    }
+    case SpeakerVolumeLevel::Id: {
+        VerifyOrReturnError(
+            HasFeature(Feature::kAudio) && HasFeature(Feature::kSpeaker), CHIP_IM_GLOBAL_STATUS(UnsupportedAttribute),
```
bzbarsky-apple, line 705: Just Speaker.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.h
```diff
@@ -0,0 +1,662 @@
...
+constexpr size_t kMaxSnapshotImageSize = kMaxLargeSecureSduLengthBytes - 100;
+
+constexpr size_t kViewportStructMaxSerializedSize =
+    TLV::EstimateStructOverhead(sizeof(uint16_t), sizeof(uint16_t), sizeof(uint16_t), sizeof(uint16_t));
+
+// The number of possible values of StreamUsageEnum.
+constexpr size_t kNumOfStreamUsageTypes = 4;
+
+// StreamUsageEnum + Anonymous tag ( 1 byte ).
+// Assumes min-size encoding ( 1 byte ) for the integer.
```
bzbarsky-apple, line 67:
```suggestion
// StreamUsageEnum + Anonymous tag (1 byte).
// Assumes min-size encoding (1 byte) for the integer.
```

"parentheticals" and "parentheses" are not the same thing.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.h
```diff
@@ -0,0 +1,662 @@
...
+{
+public:
+    CameraAVStreamMgmtDelegate() = default;
+
+    virtual ~CameraAVStreamMgmtDelegate() = default;
+
+    struct VideoStreamAllocateArgs
+    {
+        StreamUsageEnum
+            streamUsage; // Indicates the type of usage of stream ( Recording, Liveview, etc ) that this allocation is for.
```
bzbarsky-apple, line 99: Look, things with parentheses should be like this: "abc (def ghi) something".  Space before open and after close (which you used to not have) and no spaces after open and before close (which you added for some reason all over here).

Please make this stuff look sane.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1850 @@
...
+CHIP_ERROR CameraAVStreamMgmtServer::StoreViewport(const ViewportStruct & viewport)
+{
+    uint8_t buffer[kViewportStructMaxSerializedSize];
+    TLV::TLVWriter writer;
+
+    writer.Init(buffer);
+    ReturnErrorOnFailure(viewport.Encode(writer, TLV::AnonymousTag()));
+
+    auto path = ConcreteAttributePath(mEndpointId, CameraAvStreamManagement::Id, Attributes::Viewport::Id);
+    return GetSafeAttributePersistenceProvider()->SafeWriteValue(path, ByteSpan(buffer));
```
bzbarsky-apple, line 1323: This is wrong.  It should be creating a span with the right length.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1850 @@
...
+
+CHIP_ERROR CameraAVStreamMgmtServer::LoadViewport(ViewportStruct & viewport)
+{
+    uint8_t buffer[kViewportStructMaxSerializedSize];
+    MutableByteSpan bufferSpan(buffer);
+    uint16_t size = static_cast<uint16_t>(bufferSpan.size());
+
+    auto path = ConcreteAttributePath(mEndpointId, CameraAvStreamManagement::Id, Attributes::Viewport::Id);
+    ReturnErrorOnFailure(GetSafeAttributePersistenceProvider()->SafeReadValue(path, bufferSpan));
+    bufferSpan.reduce_size(size);
```
bzbarsky-apple, line 1334: This is _very very_ wrong.  Please fix.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1850 @@
...
+
+CHIP_ERROR CameraAVStreamMgmtServer::LoadRankedVideoStreamPriorities()
+{
+    uint8_t buffer[kRankedVideoStreamPrioritiesTlvSize];
+    MutableByteSpan bufferSpan(buffer);
+    uint16_t size = static_cast<uint16_t>(bufferSpan.size());
+
+    auto path = ConcreteAttributePath(mEndpointId, CameraAvStreamManagement::Id, Attributes::RankedVideoStreamPrioritiesList::Id);
+    ReturnErrorOnFailure(GetSafeAttributePersistenceProvider()->SafeReadValue(path, bufferSpan));
+    bufferSpan.reduce_size(size);
```
bzbarsky-apple, line 1375: This is wrong.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1850 @@
...
+
+    uint8_t i = 0;
+    CHIP_ERROR err;
+    while ((err = reader.Next(TLV::kTLVType_UnsignedInteger, TLV::AnonymousTag())) == CHIP_NO_ERROR && i < kNumOfStreamUsageTypes)
+    {
+        ReturnErrorOnFailure(reader.Get(mRankedVideoStreamPriorities[i]));
+        i++;
+    }
+
+    VerifyOrReturnError(err == CHIP_END_OF_TLV, err);
```
bzbarsky-apple, line 1392: OK, so here if the stored TLV had more than kNumOfStreamUsageTypes entries, we will return CHIP_NO_ERROR at this point, right, without doing the ExitContainer/VerifyEndOfContainer below.  Is that the expected behavior?

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1850 @@
...
+                                                       const Commands::VideoStreamModify::DecodableType & commandData)
+{
+    Status status             = Status::Success;
+    auto & isWaterMarkEnabled = commandData.watermarkEnabled;
+    auto & isOSDEnabled       = commandData.OSDEnabled;
+    auto & videoStreamID      = commandData.videoStreamID;
+
+    // VideoStreamModify should have either the WMARK or OSD feature supported
+    VerifyOrReturn(HasFeature(Feature::kWatermark) || HasFeature(Feature::kOnScreenDisplay),
+                   ctx.mCommandHandler.AddStatus(ctx.mRequestPath, Status::InvalidCommand));
```
bzbarsky-apple, line 1616: Why is this checked here and not in the dispatch where Video feature is checked?

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1850 @@
...
+        ChipLogError(Zcl, "CameraAVStreamMgmt[ep=%d]: Invalid image codec", mEndpointId);
+        ctx.mCommandHandler.AddStatus(ctx.mRequestPath, Status::InvalidCommand);
+    });
+
+    VerifyOrReturn(commandData.maxFrameRate > 0, {
+        ChipLogError(Zcl, "CameraAVStreamMgmt[ep=%d]: Invalid maxFrameRate", mEndpointId);
+        ctx.mCommandHandler.AddStatus(ctx.mRequestPath, Status::ConstraintError);
+    });
+
+    VerifyOrReturn(commandData.bitRate > 0, {
```
bzbarsky-apple, line 1734: Where are the constraint checks on minResolution/maxResolution?

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1850 @@
...
+    {
+        auto & streamUsage = iter.GetValue();
+        if (streamUsage == StreamUsageEnum::kUnknownEnumValue)
+        {
+            ctx.mCommandHandler.AddStatus(ctx.mRequestPath, Status::InvalidCommand);
+            return;
+        }
+        rankedStreamPriorities.push_back(streamUsage);
+    }
+
```
bzbarsky-apple, line 1795: Again: Where's the success check for the list iteration?

As in, https://github.com/project-chip/connectedhomeip/pull/36204#discussion_r1958725972 was not addressed.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1850 @@
...
+void CameraAVStreamMgmtServer::HandleCaptureSnapshot(HandlerContext & ctx,
+                                                     const Commands::CaptureSnapshot::DecodableType & commandData)
+{
+
+    Commands::CaptureSnapshotResponse::Type response;
+    auto & snapshotStreamID    = commandData.snapshotStreamID;
+    auto & requestedResolution = commandData.requestedResolution;
+    ImageSnapshot image;
+
```
bzbarsky-apple, line 1817: Where is the constraint validation for requestedResolution?

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1850 @@
...
+    // Call the delegate
+    Status status = mDelegate.CaptureSnapshot(snapshotStreamID, requestedResolution, image);
+
+    if (status != Status::Success)
+    {
+        ctx.mCommandHandler.AddStatus(ctx.mRequestPath, status);
+        return;
+    }
+
+    if (image.data.size() > kMaxSnapshotImageSize)
```
bzbarsky-apple, line 1827: Where is kMaxSnapshotImageSize coming from?  I don't see it in the spec.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1850 @@
...
+    {
+        ctx.mCommandHandler.AddStatus(ctx.mRequestPath, Status::ResourceExhausted);
+        return;
+    }
+
+    // copy the image data into the response
+    memcpy(&response.data, image.data.data(), image.data.size());
```
bzbarsky-apple, line 1834: You just smashed the stack.  This is a critical security bug right here.  Please let me know if you can't figure out why and how to do this part right.

pidarped:
Trying to incorporate changes in https://github.com/CHIP-Specifications/connectedhomeip-spec/pull/11309 which was in flux. Not all changes from that spec PR have been included yet. Will indicate here once it is ready for review again.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.h
```diff
@@ -0,0 +1,662 @@
...
+ *  Specifically, it defines interfaces for the command handling and loading of the allocated streams.
+ */
+class CameraAVStreamMgmtDelegate
+{
+public:
+    CameraAVStreamMgmtDelegate() = default;
+
+    virtual ~CameraAVStreamMgmtDelegate() = default;
+
+    struct VideoStreamAllocateArgs
```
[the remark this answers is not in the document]
pidarped: The VideoStreamStruct is similar with a ref count field. Replaced with that.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.h
```diff
@@ -0,0 +1,662 @@
...
+        VideoResolutionStruct maxResolution;     // Indicates the maximum resolution of the video stream.
+        uint32_t minBitRate;                     // Indicates the minimum bit rate ( bits/second ) of the video stream.
+        uint32_t maxBitRate;                     // Indicates the maximum bit rate ( bits/second ) of the video stream.
+        uint16_t minFragmentLen;                 // Indicates the minimum length ( msecs ) of a clip fragment for the video stream.
+        uint16_t maxFragmentLen;                 // Indicates the maximum length ( msecs ) of a clip fragment for the video stream.
+        chip::Optional<bool> isWaterMarkEnabled; // Indicates  whether a watermark can be applied on the video stream.
+        chip::Optional<bool> isOSDEnabled;       // Indicates  whether the on-screen display can be applied on the video stream.
+    };
+
+    struct AudioStreamAllocateArgs
```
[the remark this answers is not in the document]
pidarped: Replaced with AudioStreamStruct::Type.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.h
```diff
@@ -0,0 +1,662 @@
...
+        StreamUsageEnum
+            streamUsage; // Indicates the type of usage of stream ( Recording, Liveview, etc ) that this allocation is for.
+        AudioCodecEnum audioCodec; // Indicates the type of audio codec the stream should support.
+        uint8_t channelCount;      // Indicates the the number of channels used by the stream, e.g., Mono ( 1 ), Stereo ( 2 ), etc.
+        uint32_t sampleRate;       // Indicates the sampling rate of the audio stream in Hz.
+        uint32_t bitRate;          // Indicates the bitrate ( bits/sec ) of the specified audio codec.
+        uint8_t bitDepth;          // Indicates the number of information bits ( 8, 16, 24 or 32 ) used to represent each sample.
+    };
+
+    struct SnapshotStreamAllocateArgs
```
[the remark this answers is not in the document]
pidarped: Replaced with SnapshotStreamStruct::type

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
pidarped: Presumed the response.data to be a byte array. Changed to a ByteSpan.

examples/all-clusters-app/all-clusters-common/src/camera-av-stream-delegate-impl.cpp
```diff
@@ -0,0 +1,214 @@
...
+    for (SnapshotStream & stream : snapshotStreams)
+    {
+        if (!stream.isAllocated && stream.codec == allocateArgs.imageCodec)
+        {
+            stream.isAllocated = true;
+            outStreamID        = stream.id;
+        }
+    }
+
+    return Status::Success;
```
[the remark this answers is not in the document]
pidarped: Updated comment.

examples/all-clusters-app/all-clusters-common/src/camera-av-stream-delegate-impl.cpp
```diff
@@ -0,0 +1,214 @@
...
+{
+    for (AudioStream & stream : audioStreams)
+    {
+        if (stream.id == streamID && stream.isAllocated)
+        {
+            stream.isAllocated = false;
+        }
+    }
+
+    return Status::Success;
```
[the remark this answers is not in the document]
pidarped: Updated comment.

examples/all-clusters-app/all-clusters-common/src/camera-av-stream-delegate-impl.cpp
```diff
@@ -0,0 +1,214 @@
...
+    for (AudioStream & stream : audioStreams)
+    {
+        if (!stream.isAllocated && stream.codec == allocateArgs.audioCodec)
+        {
+            stream.isAllocated = true;
+            outStreamID        = stream.id;
+        }
+    }
+
+    return Status::Success;
```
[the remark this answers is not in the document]
pidarped: Updated comment.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.h
pidarped: Allocation should be infrequent and the number of streams should be relatively small. Keeping it as a vector of structs seems simpler to manage.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
pidarped: kNumOfStreamUsageTypes is expected to hold the max number of supported stream usages. So, the stored list is not expected to have more than kNumOfStreamUsageTypes entries. However, we have added a new attribute `SupportedStreamUsages` and the `RankedStreamPriorities` list should conform to that. I have filed an issue(https://github.com/project-chip/connectedhomeip/issues/37933) for adding that logic.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.h
```diff
@@ -0,0 +1,636 @@
...
+{
+public:
+    CameraAVStreamMgmtDelegate() = default;
+
+    virtual ~CameraAVStreamMgmtDelegate() = default;
+
+    /**
+     *   @brief Handle Command Delegate for Video stream allocation with the provided parameter list.
+     *
+     *   @param[in]  allocateArgs   Structure with parameters for video stream allocation.
```
bzbarsky-apple, line 98: Should document that callee must ignore the videoStreamID and referenceCount values?  Similar for the other Allocate methods.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.cpp
```diff
@@ -0,0 +1,1908 @@
...
+    mRankedVideoStreamPriorities.clear();
+    uint8_t i = 0;
+    CHIP_ERROR err;
+    while ((err = reader.Next(TLV::kTLVType_UnsignedInteger, TLV::AnonymousTag())) == CHIP_NO_ERROR)
+    {
+        ReturnErrorOnFailure(reader.Get(mRankedVideoStreamPriorities[i]));
+        i++;
+    }
+
+    ReturnErrorOnFailure(reader.ExitContainer(arrayType));
```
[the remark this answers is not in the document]
pidarped: Yes, added the `VerifyOrReturnError(err == CHIP_ERROR_END_OF_TLV)`

examples/all-clusters-app/all-clusters-common/src/camera-av-stream-delegate-impl.cpp
```diff
@@ -0,0 +1,373 @@
...
+                                                                               uint16_t & outStreamID)
+{
+    outStreamID               = kInvalidStreamID;
+    bool foundAvailableStream = false;
+
+    for (VideoStream & stream : videoStreams)
+    {
+        if (!stream.isAllocated)
+        {
+            foundAvailableStream = true;
```
bzbarsky-apple, line 55: This works, but why do we need a separate loop?  Why can't we have one loop which does:
```
        if (!stream.isAllocated)
        {
            foundAvailableStream = true;
            if (stream.codec == allocateArgs.videoCodec)
            {
                stream.isAllocated = true;
                outStreamID        = stream.id;
                return Status::Success;
            }
        }
```
kind of thing?

examples/all-clusters-app/all-clusters-common/src/camera-av-stream-delegate-impl.cpp
```diff
@@ -0,0 +1,373 @@
...
+        // Fake allocation with just matching codec
+        if (!stream.isAllocated && stream.codec == allocateArgs.videoCodec)
+        {
+            stream.isAllocated = true;
+            outStreamID        = stream.id;
+            return Status::Success;
+        }
+    }
+
+    return Status::Failure;
```
bzbarsky-apple, line 76: Still needs a spec issue and comment pointing to it here that the behavior on codec mismatch is not defined.  FAILURE is not a good error code to be returning in general.

examples/all-clusters-app/all-clusters-common/src/camera-av-stream-delegate-impl.cpp
```diff
@@ -0,0 +1,373 @@
...
+}
+
+Protocols::InteractionModel::Status CameraAVStreamManager::AudioStreamAllocate(const AudioStreamStruct & allocateArgs,
+                                                                               uint16_t & outStreamID)
+{
+    outStreamID = kInvalidStreamID;
+
+    bool foundAvailableStream = false;
+
+    for (AudioStream & stream : audioStreams)
```
bzbarsky-apple, line 117: See above.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.h
```diff
@@ -0,0 +1,644 @@
...
+{
+public:
+    CameraAVStreamMgmtDelegate() = default;
+
+    virtual ~CameraAVStreamMgmtDelegate() = default;
+
+    /**
+     *   @brief Handle Command Delegate for Video stream allocation with the provided parameter list.
+     *
+     *   @note The videoStreamID and referenceCount fields in the struct are
```
bzbarsky-apple, line 98:
```suggestion
     *   @note The videoStreamID and referenceCount fields in the struct must be
```

bzbarsky-apple, line 98: Shouldn't this be part of the param definition?  Or after it and say "fields in allocateArgs"?  Because the obvious question here is "what struct"?

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.h
```diff
@@ -0,0 +1,644 @@
...
+     *
+     *   @return Success       Stream shall be de-allocated, if found.
+     *
+     */
+    virtual Protocols::InteractionModel::Status VideoStreamDeallocate(const uint16_t streamID) = 0;
+
+    /**
+     *   @brief Handle Command Delegate for Audio stream allocation.
+     *
+     *   @note The audioStreamID and referenceCount fields in the struct are
```
bzbarsky-apple, line 143: See above.

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.h
```diff
@@ -0,0 +1,644 @@
...
+     *   @param streamID       Indicates the streamID to deallocate.
+     *
+     *   @return Success       Stream shall be de-allocated, if found.
+     */
+    virtual Protocols::InteractionModel::Status AudioStreamDeallocate(const uint16_t streamID) = 0;
+
+    /**
+     *   @brief Handle Command Delegate for Snapshot stream allocation.
+     *
+     *   @note The snapshotStreamID and referenceCount fields in the struct are
```
bzbarsky-apple, line 169: See above.

bzbarsky-apple (approved):

examples/all-clusters-app/all-clusters-common/src/camera-av-stream-delegate-impl.cpp
pidarped: https://github.com/CHIP-Specifications/connectedhomeip-spec/issues/11384

src/app/clusters/camera-av-stream-management-server/camera-av-stream-management-server.h
pidarped: Moved it under allocateArgs.

yufengwangca (approved):

PR #36205, opened 2024-10-22
Add an API on MTRDevice to wait for attributes to reach certain values.

bzbarsky-apple:


src/darwin/Framework/CHIP/MTRDevice.h
```diff
@@ -323,6 +323,25 @@ MTR_AVAILABLE(ios(16.1), macos(13.0), watchos(9.1), tvos(16.1))
...
+ * Wait until a set of attributes reaches certain values or the provided timeout
+ * (if any) is reached.  The completion is called at that point, with an error
+ * if a timeout occurred.
+ *
+ * The completion is also called with an error on input errors.
+ *
+ * The values to wait for are represented as an array of response-values which
+ * have an MTRAttributePathKey and MTRDataKey.
+ *
+ * The timeout argument can be nil to indicate no timeout. Otherwise it must
```
ksperling-apple, line 337: Is it a good idea to allow no timeout here, or even arbitrarily large timeouts? What if the attributes never reach the expected state?

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -79,6 +79,75 @@ - (BOOL)callDelegateSynchronouslyWithBlock:(void (^)(id<MTRDeviceDelegate>))bloc
...
+
+@end
+
+@implementation MTRAttributeValueWaiter
+- (instancetype)initWithValues:(NSArray<NSDictionary<NSString *, id> *> *)values queue:(dispatch_queue_t)queue completion:(MTRStatusCompletion)completion
+{
+    if (self = [super init]) {
+        auto * valueExpectations = [NSMutableArray arrayWithCapacity:values.count];
+        for (NSDictionary<NSString *, id> * value in values) {
+            auto * valueExpectation = [NSMutableDictionary dictionaryWithCapacity:2];
```
ksperling-apple, line 111: Can we not use a class for this instead of a dictionary?

src/darwin/Framework/CHIP/MTRDevice.h
bzbarsky-apple: Then you never get notified.

This seems useful to me for the "I want to know if this ever happens" use case.  You might not be expecting it to happen, but would like to know if it does.

src/darwin/Framework/CHIP/MTRDevice.mm
bzbarsky-apple: We could, sure.  Wasn't sure it was worth creating another class, but I could add one.

src/darwin/Framework/CHIP/MTRDevice.h
ksperling-apple: Well, I'm wondering about forgotten Waiter objects hanging around in memory forever. A fire-and-forget API like this seems fine if the work is reasonably time-bounded (like I'm sending some command and then expect some state to be reached in e.g. 30 seconds). If these are going to hang around for a long time, I think we'd need to either have a way to cancel them, or expose the actual Waiter object and make the client hold the reference to it as long as it's wanted?

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -711,6 +793,109 @@ - (BOOL)_attributeDataValue:(MTRDeviceDataValueDictionary)observed satisfiesValu
...
+        return;
+    }
+
+    // Otherwise, wait for the values to arrive or our timeout.
+    [self.attributeValueWaiters addObject:attributeWaiter];
+
+    if (timeout != nil) {
+        mtr_weakify(attributeWaiter);
+        mtr_weakify(self);
+        dispatch_after(dispatch_time(DISPATCH_TIME_NOW, static_cast<uint64_t>(timeout.doubleValue * NSEC_PER_SEC)), self.queue, ^{
```
ksperling-apple, line 843: If timeouts can be long, it would be better to use a dispatch_source that can be cancelled, otherwise the dispatched block will remain in memory until the timeout comes around even if we successfully arrived at the expected state.

src/darwin/Framework/CHIP/MTRDevice.h
bzbarsky-apple: Talked to some plausible API consumers and added a 5-minute cap.

src/darwin/Framework/CHIP/MTRDevice.mm
bzbarsky-apple: Done.

src/darwin/Framework/CHIP/MTRDevice.mm
bzbarsky-apple: Added canceling of timers.

src/darwin/Framework/CHIP/MTRDevice.h
```diff
@@ -337,6 +337,24 @@ MTR_AVAILABLE(ios(16.1), macos(13.0), watchos(9.1), tvos(16.1))
...
     MTR_AVAILABLE(ios(17.6), macos(14.6), watchos(10.6), tvos(17.6));
+
+/**
+ * Wait until a set of attributes reaches certain values or the provided timeout
+ * (if any) is reached.  The completion is called at that point, with an error
+ * if a timeout occurred.
+ *
+ * The completion is also called with an error on input errors.
+ *
+ * The values to wait for are represented as an array of response-values which
```
ksperling-apple, line 348: Is that really a convenient representation for clients here? The whole thing could be a dictionary from attribute paths to values for example.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -78,6 +78,98 @@ - (BOOL)callDelegateSynchronouslyWithBlock:(void (^)(id<MTRDeviceDelegate>))bloc
...
+@property (nonatomic, retain, readonly) MTRDeviceResponseValueDictionary value;
+
+- (instancetype)initWithValue:(MTRDeviceResponseValueDictionary)value;
+@end
+
+@implementation MTRAwaitedAttributeState
+- (instancetype)initWithValue:(MTRDeviceResponseValueDictionary)value
+{
+    if (self = [super init]) {
+        _valueSatisfied = NO;
```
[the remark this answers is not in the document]
bzbarsky-apple: True, but I prefer being more explicit about this sort of thing.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -78,6 +78,98 @@ - (BOOL)callDelegateSynchronouslyWithBlock:(void (^)(id<MTRDeviceDelegate>))bloc
...
+        _value = value;
+    }
+
+    return self;
+}
+@end
+
+// Represents the data passed to waitForAttributeValues:
+@interface MTRAttributeValueWaiter : NSObject
```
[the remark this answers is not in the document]
bzbarsky-apple: Done.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -78,6 +78,98 @@ - (BOOL)callDelegateSynchronouslyWithBlock:(void (^)(id<MTRDeviceDelegate>))bloc
 #endif
 @end
 
+#pragma mark - MTRAttributeValueWaiter
+
+// Represents the state of a single attribute being waited for: has the path and
+// value being waited for in the response-value and whether the value has been reached.
+@interface MTRAwaitedAttributeState : NSObject
```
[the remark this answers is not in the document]
bzbarsky-apple: Done.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -631,7 +728,15 @@ - (BOOL)_attributeDataValue:(MTRDeviceDataValueDictionary)observed satisfiesValu
                 return NO;
             }
 
-            if (![self _attributeDataValue:observedEntry[MTRDataKey] satisfiesValueExpectation:expectedEntry[MTRDataKey]]) {
+            MTRDeviceDataValueDictionary observedDataValue = observedEntry[MTRDataKey];
+            MTRDeviceDataValueDictionary expectedDataValue = expectedEntry[MTRDataKey];
+
+            if (![observedDataValue isKindOfClass:NSDictionary.class] || ![expectedDataValue isKindOfClass:NSDictionary.class]) {
```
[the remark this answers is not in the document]
bzbarsky-apple: Done.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -106,6 +202,7 @@ - (instancetype)initForSubclassesWithNodeID:(NSNumber *)nodeID controller:(MTRDe
         _deviceController = controller;
         _nodeID = nodeID;
         _state = MTRDeviceStateUnknown;
+        _attributeValueWaiters = [NSMutableSet set];
```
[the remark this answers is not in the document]
bzbarsky-apple: Done.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -710,6 +815,128 @@ - (BOOL)_attributeDataValue:(MTRDeviceDataValueDictionary)observed satisfiesValu
...
+    dispatch_resume(timerSource);
+}
+
+- (void)_attributeValue:(MTRDeviceDataValueDictionary)value reportedForPath:(MTRAttributePath *)path
+{
+    os_unfair_lock_assert_owner(&_lock);
+
+    // Check whether anyone was waiting for this attribute.  Iterate over a copy
+    // of the set, so we can remove waiters that are done waiting.
+    for (MTRAttributeValueWaiter * attributeValueWaiter in [self.attributeValueWaiters copy]) {
```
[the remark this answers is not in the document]
bzbarsky-apple: Yeah, good point.  Done.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -710,6 +815,128 @@ - (BOOL)_attributeDataValue:(MTRDeviceDataValueDictionary)observed satisfiesValu
...
 
+#pragma mark - Handling of waits for attribute values
+
+- (void)waitForAttributeValues:(NSArray<MTRDeviceResponseValueDictionary> *)values timeout:(NSTimeInterval)timeout queue:(dispatch_queue_t)queue completion:(void (^)(NSError * _Nullable error))completion
+{
+    // Check that all the values are in fact attribute response-values.
+    for (MTRDeviceResponseValueDictionary value in values) {
+        if (!value[MTRAttributePathKey] || !value[MTRDataKey]) {
+            MTR_LOG_ERROR("%@ waitForAttributeValues passed a non-attribute value %@", self, value);
+            dispatch_async(queue, ^{
```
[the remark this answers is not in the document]
bzbarsky-apple: OK.  Done.

src/darwin/Framework/CHIP/MTRDevice.h
bzbarsky-apple: Hmm, that's a good point.  Will change that.

bzbarsky-apple: It's a bit of a tough call, in that the format here is what we actually report attributes as to consumers... But I agree that a typical use of this would probably do better with the paths-to-values dictionary.

ksperling-apple (dismissed):

src/darwin/Framework/CHIP/MTRDevice.h
```diff
@@ -337,6 +338,27 @@ MTR_AVAILABLE(ios(16.1), macos(13.0), watchos(9.1), tvos(16.1))
...
+ * completion is called, the completion will be called with an
+ * MTRErrorCodeCancelled error.
+ *
+ * The attributes and values to wait for are represented as a dictionary which
+ * has the attribute paths as keys and the expected data-values as values.
+ *
+ * Nil will be returned if for some reason the wait could not be started
+ * (e.g. invalid input).
+ */
+- (MTRAttributeValueWaiter * _Nullable)waitForAttributeValues:(NSDictionary<MTRAttributePath *, NSDictionary<NSString *, id> *> *)values
```
woody-apple, line 357: If this is nullable, then it needs to return an NSError ** out param. 

Per our discussion as well, I feel this is mixing two worlds together... I prefer delegate, but not a hill i'll die on :)

bzbarsky-apple: So I talked about the delegate idea to some of the planned API consumer, and it has a drawback, which is that it makes it hard to take different actions depending on which set of attributes you were waiting for; you sort of have to define a new protocol implementation for each new thing you decide you might need, and if you need to change the behavior at runtime at all you end up having to do blocks for the behavior anyway, but now held by the delegate.  So just in terms of how people want to use this API it does not feel like a great fit.

Good point about NSError**; if this stays nullable (as opposed to switching to "just assert on bad input"), I will do that.

woody-apple: Personally, completions (pun intended) are a trap here, but if that's what the people want, sounds fine. :)

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -710,6 +733,130 @@ - (BOOL)_attributeDataValue:(MTRDeviceDataValueDictionary)observed satisfiesValu
...
+
+    // Check whether anyone was waiting for this attribute.
+    NSHashTable * satisfiedWaiters = [NSHashTable weakObjectsHashTable];
+    for (MTRAttributeValueWaiter * attributeValueWaiter in self.attributeValueWaiters) {
+        if ([attributeValueWaiter _attributeValue:value reportedForPath:path byDevice:self] && attributeValueWaiter.allValuesSatisfied) {
+            [satisfiedWaiters addObject:attributeValueWaiter];
+        }
+    }
+
+    [self.attributeValueWaiters minusHashTable:satisfiedWaiters];
```
ksperling-apple, line 816:
```suggestion
    NSMutableArray * satisfiedWaiters;
    for (MTRAttributeValueWaiter * attributeValueWaiter in self.attributeValueWaiters) {
        if ([attributeValueWaiter _attributeValue:value reportedForPath:path byDevice:self] && attributeValueWaiter.allValuesSatisfied) {
            if (!satisfiedWaiters) {
                satisfiedWaiter = [[NSMutableArray alloc] init];
            }
            [satisfiedWaiters addObject:attributeValueWaiter];
        }
    }
```

I think we can avoid allocating the satisfiedWaiters container until needed (the `for in` loop below is a no-op if satisfiedWaiters is nil). Also making it weak seems harder to reason about for no benefit (unless I'm missing something), and `_notify...` already handles the removal from `self.attributeValueWaiters`.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -124,6 +131,9 @@ - (void)dealloc
 {
     // TODO: retain cycle and clean up https://github.com/project-chip/connectedhomeip/issues/34267
     MTR_LOG("MTRDevice dealloc: %p", self);
+
+    std::lock_guard lock(_lock);
```
ksperling-apple, line 135: There shouldn't be a need to lock in dealloc

ksperling-apple (approved):

src/darwin/Framework/CHIP/MTRDevice.h
```diff
@@ -337,6 +338,27 @@ MTR_AVAILABLE(ios(16.1), macos(13.0), watchos(9.1), tvos(16.1))
                     queue:(dispatch_queue_t)queue
                completion:(void (^)(NSURL * _Nullable url, NSError * _Nullable error))completion
     MTR_AVAILABLE(ios(17.6), macos(14.6), watchos(10.6), tvos(17.6));
+
+/**
+ * Wait until a set of attributes reaches certain values or the provided timeout
+ * (if any) is reached.  The completion is called at that point, with an
+ * MTREErrorCodeTimeout error if a timeout occurred.
```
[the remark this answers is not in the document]
bzbarsky-apple: The "if any" dates back to when it was nullable, indeed.  Will fix this comment up.  It should probably mention cancellation too, in the set of things that cause the completion to be called.

src/darwin/Framework/CHIP/MTRAttributeValueWaiter.mm
```diff
@@ -0,0 +1,137 @@
...
+
+@interface MTRAttributeValueWaiter ()
+@property (nonatomic, retain) NSDictionary<MTRAttributePath *, MTRAwaitedAttributeState *> * valueExpectations;
+// Protected by the MTRDevice's lock.
+@property (nonatomic, readwrite, retain) dispatch_queue_t queue;
+@property (nonatomic, readwrite, nullable) MTRStatusCompletion completion;
+@end
+
+@implementation MTRAttributeValueWaiter {
+    MTRDevice * _device;
```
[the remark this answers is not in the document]
bzbarsky-apple: No, this one could be a property too.  Some hysteresis happened here.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -710,6 +733,130 @@ - (BOOL)_attributeDataValue:(MTRDeviceDataValueDictionary)observed satisfiesValu
...
+    mtr_weakify(timerSource);
+    dispatch_source_set_event_handler(timerSource, ^{
+        mtr_strongify(self);
+        mtr_strongify(attributeWaiter);
+        mtr_strongify(timerSource);
+        if (self != nil && attributeWaiter != nil) {
+            [self _attributeWaitTimedOut:attributeWaiter];
+        } else if (timerSource != nil) {
+            // Make sure to cancel the timer ourselves, if we can't call _attributeWaitTimedOut.
+            dispatch_source_cancel(timerSource);
```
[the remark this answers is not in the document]
bzbarsky-apple: > Wouldn't it be simpler for the block to hold a strong reference to the timer source and cancel it unconditionally before doing anything else?

Is it safe to dispatch_source_cancel a timer source multiple times?  If so, that would be doable, yes, since then the cancels that happen in callees of the block will just be no-ops.

src/darwin/Framework/CHIP/MTRDevice_XPC.mm
```diff
@@ -165,6 +165,33 @@ - (oneway void)device:(NSNumber *)nodeID receivedAttributeReport:(NSArray<MTRDev
...
+
+    std::lock_guard lock(_lock);
+    for (NSDictionary<NSString *, id> * report in attributeReport) {
+        if (!MTR_SAFE_CAST(report, NSDictionary)) {
+            MTR_LOG_ERROR("%@ handed a response-value that is not a dictionary: %@", self, report);
+            continue;
+        }
+
+        MTRAttributePath * path = report[MTRAttributePathKey];
+        if (!MTR_SAFE_CAST(path, MTRAttributePath)) {
```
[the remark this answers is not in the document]
bzbarsky-apple: Yep, will do.

src/darwin/Framework/CHIP/MTRDevice.mm
bzbarsky-apple: It was a weak thing so I could do the minusHashTable.  I'll double-check that the removals happen in some other way, and if so adjust accordingly.

src/darwin/Framework/CHIP/MTRAttributeValueWaiter.h
```diff
@@ -0,0 +1,38 @@
...
++ (instancetype)new NS_UNAVAILABLE;
+
+/**
+ * Cancel the wait for the set of attribute path/value pairs represented by this
+ * MTRAttributeValueWaiter.  If the completion has not been called yet, it will
+ * becalled with MTRErrorCodeCancelled.
+ */
+- (void)cancel;
+
+@property (readonly, nonatomic) NSUUID * waiterID;
```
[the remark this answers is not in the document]
bzbarsky-apple: Done.

src/darwin/Framework/CHIP/MTRDevice.h
bzbarsky-apple: Done.

src/darwin/Framework/CHIP/MTRDevice.h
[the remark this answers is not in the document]
bzbarsky-apple: Done.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -96,6 +98,11 @@ - (BOOL)unitTestSuppressTimeBasedReachabilityChanges:(MTRDevice *)device;
 @end
 #endif
 
+@interface MTRDevice ()
```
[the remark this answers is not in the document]
bzbarsky-apple: Done.

src/darwin/Framework/CHIP/MTRAttributeValueWaiter.mm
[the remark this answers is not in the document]
bzbarsky-apple: Made this a property.

src/darwin/Framework/CHIP/MTRDevice.mm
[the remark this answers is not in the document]
bzbarsky-apple: Looks like dispatch_source_cancel is in fact idempotent, so made this change.

src/darwin/Framework/CHIP/MTRDevice.mm
bzbarsky-apple: Added a comment.

bzbarsky-apple:
Fast-tracking platform-specific fix with platform owner review

PR #36221, opened 2024-10-23
[Silabs] Configure Light-Switch as a LIT ICD app

mkardous-silabs:
#### Description
PR configures the Light-Swith as a LIT ICD with Dynamic Sit Lit Support
- Update ZAP file with the necessary configs
- Add Common ICD shell that is enabled if the device is an ICD and shell is enabled
- Clean defines checks
- Moves shell files into their respective directory
- Adds a LIT ICD Light-Switch Zap file to avoid causing build issues or changing the behavior of other platforms sample apps.

PR also pulls latest docker image which reducdes the size of the efr32 docker image.

#### Tests
Manual tests to validate everything works as expected


jmartinez-silabs (approved):

src/lib/shell/BUILD.gn
```diff
@@ -57,9 +58,14 @@ static_library("shell") {
   } else if (chip_device_platform == "efr32" ||
              chip_device_platform == "SiWx917") {
     sources += [
+      # Temporary workaround to prevent a circular dependecy while restructuring the silabs shell files
+      "${chip_root}/examples/platform/silabs/shell/MatterShell.h",
       "MainLoopSilabs.cpp",
       "streamer_silabs.cpp",
     ]
+
+    # Temporary workaround to prevent a circular dependecy while restructuring the silabs shell files
```
jmartinez-silabs, line 67: Can we track this in a internal ticket please

examples/light-switch-app/light-switch-common/BUILD.gn
```diff
@@ -14,8 +14,13 @@
 
 import("//build_overrides/chip.gni")
 import("${chip_root}/src/app/chip_data_model.gni")
+import("${chip_root}/src/app/icd/icd.gni")
 
 chip_data_model("light-switch-common") {
-  zap_file = "light-switch-app.zap"
+  if (chip_enable_icd_lit) {
+    zap_file = "lit-icd-light-switch-app.zap"
```
andy31415, line 21: should the name order be consistent? one is lit-icd and the config is icd_lit

examples/platform/silabs/BaseApplication.cpp
```diff
@@ -37,9 +37,13 @@
 #endif // QR_CODE_ENABLED
 #endif // DISPLAY_ENABLED
 
-#if CHIP_CONFIG_ENABLE_ICD_SERVER == 1
+#if defined(CHIP_CONFIG_ENABLE_ICD_SERVER) && CHIP_CONFIG_ENABLE_ICD_SERVER
```
andy31415, line 40: We should not use defined/ifdef.

Please include relevant headers and only use `#if` and `#if !`.

andy31415 (changes requested):
Changes requested: do not use `ifdef` or `defined(....)`

src/lib/shell/BUILD.gn
```diff
@@ -57,9 +58,15 @@ static_library("shell") {
...
     sources += [
+      # Temporary workaround to prevent a circular dependecy while restructuring the silabs shell files
+      # Due to some odd dependency issue, we need the source file and the include_dir for the build to pass.
+      "${chip_root}/examples/platform/silabs/shell/MatterShell.h",
       "MainLoopSilabs.cpp",
       "streamer_silabs.cpp",
     ]
+
+    # Temporary workaround to prevent a circular dependecy while restructuring the silabs shell files
+    include_dirs = [ "${chip_root}/examples/platform/silabs/shell/" ]
```
andy31415, line 69: Please describe the circular dependency. Generally this is not ok: we cannot link from src into examples. You may need to fix this as a prerequisite before this PR can go in.

src/lib/shell/BUILD.gn
```diff
@@ -57,9 +58,15 @@ static_library("shell") {
   } else if (chip_device_platform == "efr32" ||
              chip_device_platform == "SiWx917") {
     sources += [
+      # Temporary workaround to prevent a circular dependecy while restructuring the silabs shell files
+      # Due to some odd dependency issue, we need the source file and the include_dir for the build to pass.
+      "${chip_root}/examples/platform/silabs/shell/MatterShell.h",
```
andy31415, line 63: This does not look ok. Saying in a comment "I know it does not look ok" does not make it better.

What kind of a commitment can we get for this to get fixed? I would approve if I am allowed to roll the PR back if not fixed within X time where X is some reasonably picked but short timeframe, like 2-4 weeks.

andy31415:
> Changes requested: do not use `ifdef` or `defined(....)`

Also making lib depend on examples is something that needs fixing.

examples/light-switch-app/light-switch-common/BUILD.gn
mkardous-silabs: done

examples/platform/silabs/BaseApplication.cpp
mkardous-silabs: done

src/lib/shell/BUILD.gn
mkardous-silabs: reverted these changes until i have time to do the full cleanup.

src/lib/shell/BUILD.gn
mkardous-silabs: reverted these changes until i have time to do the full cleanup.

yunhanw-google (approved):

andy31415 (approved):

PR #36223, opened 2024-10-23
[ICD] Set MaximumCheckInBackOff to external and fix define guards for unit test members

mkardous-silabs:
#### Description
PR addresses two issues in the ICD server code :
- MaximumCheckInBackOff attribute should have been set to external in the zcl json files. This was not done when the attributes were initially added to the xml.
- In the ICDManager, a define guard was set to `idef` when the define was always set to 0 or 1. Updated the guard to correctly check the define value

Only the first comment as functionnal changes. The second is just the generation.

####
CI and generation to validate the changes

### DIFF
--- a/src/app/zap-templates/zcl/zcl-with-test-extensions.json
+++ b/src/app/zap-templates/zcl/zcl-with-test-extensions.json
@@ -290,7 +290,8 @@
             "ActiveModeThreshold",
             "RegisteredClients",
             "ICDCounter",
-            "ClientsSupportedPerFabric"
+            "ClientsSupportedPerFabric",
+            "MaximumCheckInBackOff"
         ],
         "Occupancy Sensing": ["HoldTimeLimits", "HoldTime", "FeatureMap"],
         "Operational Credentials": [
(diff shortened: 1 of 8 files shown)


jmartinez-silabs (approved):

yunhanw-google (approved):

src/app/zap-templates/zcl/zcl-with-test-extensions.json
```diff
@@ -290,7 +290,8 @@
             "ActiveModeThreshold",
             "RegisteredClients",
             "ICDCounter",
-            "ClientsSupportedPerFabric"
+            "ClientsSupportedPerFabric",
+            "MaximumCheckInBackOff"
```
gmarcosb, line 294: `MaximumCheckInBackOff` is tagged as `apiMaturity="provisional"` - should it be here?

mkardous-silabs: Where is this set? I'm not familiar with the `apiMaturity` tag

gmarcosb: @ https://github.com/project-chip/connectedhomeip/blob/73ff58ff14c91c590c679a40c9570e434dbdd88f/src/app/zap-templates/zcl/data-model/chip/icd-management-cluster.xml#L4

bzbarsky-apple: > MaximumCheckInBackOff is tagged as apiMaturity="provisional" - should it be here?

@gmarcosb apiMaturity has absolutely nothing to do with whether things are listed in the JSON metadata.

PR #36229, opened 2024-10-24
Use custom ember-buffer decode on codegen data model Read as well.

andy31415:
This saves about 1KB of flash compared to the alternative (more template-heavy) approach.

image

### DIFF
--- a/src/app/codegen-data-model-provider/EmberAttributeDataBuffer.cpp
+++ b/src/app/codegen-data-model-provider/EmberAttributeDataBuffer.cpp
@@ -462,15 +462,15 @@ CHIP_ERROR EmberAttributeDataBuffer::EncodeInteger(chip::TLV::TLVWriter & writer
         return writer.Put(tag, static_cast<uint32_t>(value.uint_value));
     case ZCL_INT40U_ATTRIBUTE_TYPE: // Unsigned 40-bit integer
     case ZCL_INT48U_ATTRIBUTE_TYPE: // Unsigned 48-bit integer
-    case ZCL_INT56U_ATTRIBUTE_TYPE: // Unsigned 56-bit integer
-    case ZCL_INT64U_ATTRIBUTE_TYPE: // Unsigned 64-bit integer
+    case ZCL_INT56U_ATTRIBUTE_TYPE: // Signed 56-bit integer
+    case ZCL_INT64U_ATTRIBUTE_TYPE: // Signed 64-bit integer
         return writer.Put(tag, static_cast<uint64_t>(value.uint_value));
-    case ZCL_INT8S_ATTRIBUTE_TYPE: // Unsigned 8-bit integer
+    case ZCL_INT8S_ATTRIBUTE_TYPE: // Signed 8-bit integer
         return writer.Put(tag, static_cast<int8_t>(value.int_value));
-    case ZCL_INT16S_ATTRIBUTE_TYPE: // Unsigned 16-bit integer
+    case ZCL_INT16S_ATTRIBUTE_TYPE: // Signed 16-bit integer
         return writer.Put(tag, static_cast<int16_t>(value.int_value));
-    case ZCL_INT24S_ATTRIBUTE_TYPE: // Unsigned 24-bit integer
-    case ZCL_INT32S_ATTRIBUTE_TYPE: // Unsigned 32-bit integer
+    case ZCL_INT24S_ATTRIBUTE_TYPE: // Signed 24-bit integer
+    case ZCL_INT32S_ATTRIBUTE_TYPE: // Signed 32-bit integer
         return writer.Put(tag, static_cast<int32_t>(value.int_value));
     default:
         return writer.Put(tag, static_cast<int64_t>(value.int_value));
--- a/src/app/codegen-data-model-provider/EmberAttributeDataBuffer.cpp
+++ b/src/app/codegen-data-model-provider/EmberAttributeDataBuffer.cpp
@@ -16,14 +16,14 @@
  */
 #include <app/codegen-data-model-provider/EmberAttributeDataBuffer.h>
 
-#include <app/util/attribute-metadata.h>
-#include <lib/core/TLVTags.h>
-#include <lib/core/TLVWriter.h>
 #include <app-common/zap-generated/attribute-type.h>
+#include <app/util/attribute-metadata.h>
 #include <app/util/attribute-storage-null-handling.h>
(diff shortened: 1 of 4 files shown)


tcarmelveilleux (approved):

woody-apple (approved):

src/app/codegen-data-model-provider/EmberAttributeDataBuffer.cpp
```diff
@@ -321,6 +375,200 @@ CHIP_ERROR EmberAttributeDataBuffer::Decode(chip::TLV::TLVReader & reader)
...
+    uint8_t raw_bytes[8];
+
+    bool isSigned = (mAttributeType == ZCL_INT8S_ATTRIBUTE_TYPE) //
+        || (mAttributeType == ZCL_INT16S_ATTRIBUTE_TYPE)         //
+        || (mAttributeType == ZCL_INT24S_ATTRIBUTE_TYPE)         //
+        || (mAttributeType == ZCL_INT32S_ATTRIBUTE_TYPE)         //
+        || (mAttributeType == ZCL_INT40S_ATTRIBUTE_TYPE)         //
+        || (mAttributeType == ZCL_INT48S_ATTRIBUTE_TYPE)         //
+        || (mAttributeType == ZCL_INT56S_ATTRIBUTE_TYPE)         //
+        || (mAttributeType == ZCL_INT64S_ATTRIBUTE_TYPE);
```
bzbarsky-apple, line 394: Is there a reason we are reinventing `chip::app::IsSignedAttributeType`?

src/app/codegen-data-model-provider/EmberAttributeDataBuffer.cpp
```diff
@@ -321,6 +375,200 @@ CHIP_ERROR EmberAttributeDataBuffer::Decode(chip::TLV::TLVReader & reader)
...
+    for (int i = static_cast<int>(byteCount) - 1; i >= 0; i--)
+    {
+#endif
+        value.uint_value <<= 8;
+        value.uint_value = (value.uint_value & ~0xFFULL) | raw_bytes[i];
+    }
+
+    if (mIsNullable && (value.uint_value == nullValue))
+    {
+        // MaxValue is used for NULL setting
```
bzbarsky-apple, line 450: This comment does not make sense for signed types.

src/app/codegen-data-model-provider/EmberAttributeDataBuffer.cpp
```diff
@@ -321,6 +375,200 @@ CHIP_ERROR EmberAttributeDataBuffer::Decode(chip::TLV::TLVReader & reader)
...
+        return writer.PutNull(tag);
+    }
+
+    switch (mAttributeType)
+    {
+    case ZCL_INT8U_ATTRIBUTE_TYPE: // Unsigned 8-bit integer
+        return writer.Put(tag, static_cast<uint8_t>(value.uint_value));
```
bzbarsky-apple, line 457: Why are we doing this?  Assuming the value is in range at all, we can just `Put()` as a 64-bit signed/unsigned respectively depending on isSigned and don't need this whole big switch and the casting.

src/app/codegen-data-model-provider/EmberAttributeDataBuffer.cpp
```diff
@@ -321,6 +375,200 @@ CHIP_ERROR EmberAttributeDataBuffer::Decode(chip::TLV::TLVReader & reader)
...
+        {
+            return endianReader.StatusCode();
+        }
+        switch (value)
+        {
+        case 0:
+        case 1:
+            return writer.PutBoolean(tag, value != 0);
+        case 0xFF:
+            return writer.PutNull(tag);
```
bzbarsky-apple, line 500: This will incorrectly encode null even if `!mIsNullable`, instead of correctly erroring out.

src/app/codegen-data-model-provider/EmberAttributeDataBuffer.cpp
```diff
@@ -321,6 +375,200 @@ CHIP_ERROR EmberAttributeDataBuffer::Decode(chip::TLV::TLVReader & reader)
...
+            float value;
+        } value;
+
+        if (!endianReader.ReadBytes(value.raw, sizeof(value)).IsSuccess())
+        {
+            return endianReader.StatusCode();
+        }
+        if (NumericAttributeTraits<float>::IsNullValue(value.value))
+        {
+            return writer.PutNull(tag);
```
bzbarsky-apple, line 536: Again, this needs to check mIsNullable!

src/app/codegen-data-model-provider/EmberAttributeDataBuffer.cpp
```diff
@@ -321,6 +375,200 @@ CHIP_ERROR EmberAttributeDataBuffer::Decode(chip::TLV::TLVReader & reader)
...
+            double value;
+        } value;
+
+        if (!endianReader.ReadBytes(value.raw, sizeof(value)).IsSuccess())
+        {
+            return endianReader.StatusCode();
+        }
+        if (NumericAttributeTraits<double>::IsNullValue(value.value))
+        {
+            return writer.PutNull(tag);
```
bzbarsky-apple, line 553: Needs to check mIsNullable.

src/app/codegen-data-model-provider/EmberAttributeDataBuffer.h
```diff
@@ -73,6 +85,10 @@ class EmberAttributeDataBuffer
     /// Takes into account internal mIsNullable.
     CHIP_ERROR DecodeSignedInteger(chip::TLV::TLVReader & reader, EndianWriter & writer);
 
+    /// Encodes the UNSIGNED integer into `writer`.
```
bzbarsky-apple, line 88: The code is not just for unsigned. What is this comment talking about?

src/app/codegen-data-model-provider/EmberAttributeDataBuffer.cpp
andy31415: probably me not knowing all the functions we have ... will update it.

src/app/codegen-data-model-provider/EmberAttributeDataBuffer.cpp
andy31415: I will re-write it. It does rely that we have a union and we store the null value as a uint64_t ... so we are looking at a static u64 cast on both sides here ...

PR #36232, opened 2024-10-24
TC-DA-1.4 - Updates YAML per VS steps

raul-marquez-csa:
Addresses
[[TC-DA-1.4] - Manual script has to be updated as per VS changes #357](https://github.com/project-chip/matter-test-scripts/issues/357)

#### Testing


andy31415 (approved):

Rajashreekalmane (approved):

src/app/tests/suites/certification/Test_TC_DA_1_4.yaml
```diff
@@ -33,13 +33,9 @@ tests:
 
     - label: "Pre-Conditions"
       verification: |
```
cecille, line 35: Some questions in general about these test cases:
- this is an entirely manual test case because it is for commissioners - where is this script used? Does this appear in the TH? Is this script just for QA use?
- The commissioners under test will almost certainly NOT show logs like in the examples. It's also permissible to just warn. So is having those logs there helpful if this is actually for cert? This seems like a lot of work to generate those logs when the test step could just say the exact same thing as the test plan

raul-marquez-csa: The test shows up in the TH.
@Rajashreekalmane, can you address Cecille's other comments?

Rajashreekalmane: Since this is a manual test case listed under the UI-Manual section, the example logs are provided just to guide. If the commissioners don't show the exact logs, it's fine to issue a warning instead of trying to match them exactly.

cecille: ok so then...what's the purpose of forcing these files to be in sync with the vs docs? This seems like work for the sake of work.

Rajashreekalmane: Hi @cecille, understood. While the test is manual, we still need to keep the YAML script in sync with the latest VS to ensure consistency across the YAML, Test-plan and verification document. We can reduce the number of log entries included in the Excel sheet to keep things simpler, but the documented steps must still align with the test plan.

cecille: why? Why are you keeping reference logs in two places if people don't even need to exactly match them? It's two sources of not-quite-truth. If there's a manual yaml, why does the vscode doc not just say "see yaml" and logs that need to be updated can be updated once? Many of the updates here are basically just updates of the log timestamps.

Rajashreekalmane: That’s a fair point — maintaining logs in both the YAML and VS docs does create duplication, especially when only timestamps or minor log text changes are involved.

The main reason we’ve been keeping them in sync is to ensure that anyone reviewing only the VS docs still sees the same context as in the YAML. But I agree that referencing the YAML directly from the VS doc could reduce redundancy and avoid multiple sources of truth.

I can propose simplifying this by:

1. Keeping the detailed example logs only in the YAML.
2. Updating the VS docs to reference the YAML for log details, while still preserving the step alignment with the Test Plan.


Would that approach work better?

cecille: I think that would be a great solution

cecille (approved):

cecille:
old, manual test with just log updates. Labeling as sdk maintainer for that reason.

PR #36236, opened 2024-10-24
Fixed import module for chip.testing.tasks

rquidute:
The goal of this PR is to fix the import `chip.testing.tasks` at `TC_MCORE_FS_1_1` and `TC_MCORE_FS_1_4` test cases.

### Related Issue
https://github.com/project-chip/certification-tool/issues/473

### DIFF
--- a/src/python_testing/TC_MCORE_FS_1_1.py
+++ b/src/python_testing/TC_MCORE_FS_1_1.py
@@ -47,7 +47,7 @@ import time
 
 import chip.clusters as Clusters
 from chip import ChipDeviceCtrl
-from chip.testing.tasks import Subprocess
+from matter_testing_infrastructure.chip.testing.tasks import Subprocess
 from matter_testing_support import MatterBaseTest, TestStep, async_test_body, default_matter_test_main
 from mobly import asserts
 
--- a/src/python_testing/TC_MCORE_FS_1_4.py
+++ b/src/python_testing/TC_MCORE_FS_1_4.py
@@ -49,7 +49,7 @@ import tempfile
 import chip.clusters as Clusters
 from chip import ChipDeviceCtrl
 from chip.interaction_model import Status
-from chip.testing.tasks import Subprocess
+from matter_testing_infrastructure.chip.testing.tasks import Subprocess
 from matter_testing_support import MatterBaseTest, TestStep, async_test_body, default_matter_test_main, type_matches
 from mobly import asserts
 from TC_MCORE_FS_1_1 import AppServer


cecille (approved):

tehampson (approved):

andy31415:
Is this not run in CI? It seems we have arguments for it....should this not have failed CI?

src/python_testing/TC_MCORE_FS_1_1.py
```diff
@@ -47,7 +47,7 @@
 
 import chip.clusters as Clusters
 from chip import ChipDeviceCtrl
-from chip.testing.tasks import Subprocess
```
andy31415, line 50: in 1.4 branch, the path seems to be chip.testing.tasks for the WHL:

https://github.com/rquidute/connectedhomeip/blob/v1.4-branch/src/python_testing/matter_testing_infrastructure/BUILD.gn#L34

andy31415 (changes requested):
`matter_testing_infrastructure.chip.testing.tasks` is not a valid path out of a WHL installation.

it looks like `chip.testing.tasks` is a valid path. https://github.com/project-chip/connectedhomeip/blob/v1.4-branch/scripts/tests/run_python_test.py#L31 in particular as an example uses `chip.testing` without the `matter_testing_infrastructure` prefix.

saurabhst (approved):

hiltonlima:
> `matter_testing_infrastructure.chip.testing.tasks` is not a valid path out of a WHL installation.
> 
> it looks like `chip.testing.tasks` is a valid path. https://github.com/project-chip/connectedhomeip/blob/v1.4-branch/scripts/tests/run_python_test.py#L31 in particular as an example uses `chip.testing` without the `matter_testing_infrastructure` prefix.

Hi @andy31415 ,

From what I understand, the example you mentioned uses only imports:
```
from chip.testing.metadata import Metadata, MetadataReader
from chip.testing.tasks import Subprocess

```
and in this test file, both forms are used:
```
from chip import ChipDeviceCtrl
from chip.testing.tasks import Subprocess
```
I believe this causes the base module (chip) to point to a different place and not find the testing module.

We have the chip in two paths:
```
src/controller/python/chip
src/python_testing/matter_testing_infrastructure/chip
```

src/python_testing/TC_MCORE_FS_1_1.py
```diff
@@ -47,7 +47,7 @@
...
+from matter_testing_infrastructure.chip.testing.tasks import Subprocess
```
rquidute, line 50:
```suggestion
try:
    from chip.testing.tasks import Subprocess
except:
    from matter_testing_infrastructure.chip.testing.tasks import Subprocess
```

src/python_testing/TC_MCORE_FS_1_4.py
```diff
@@ -49,7 +49,7 @@
 import chip.clusters as Clusters
 from chip import ChipDeviceCtrl
 from chip.interaction_model import Status
-from chip.testing.tasks import Subprocess
+from matter_testing_infrastructure.chip.testing.tasks import Subprocess
```
rquidute, line 52:
```suggestion
try:
    from chip.testing.tasks import Subprocess
except:
    from matter_testing_infrastructure.chip.testing.tasks import Subprocess
```

rquidute (commented):
@andy31415 How about having both imports, since when we are running using TH (not using vent) we are getting a `chip.testing` module not found.

arkq:
Can we properly install the chip-testing package in the TH docker image, instead of hacking python import?

The hack for this (the one which will work properly), would be to add the path for the `chip.testing` module `/root/python_testing/scripts/sdk/matter_testing_infrastructure` to the `sys.path` before calling `import chip.testing`. Something like this:

```python
import sys
sys.path.append("/root/python_testing/scripts/sdk/matter_testing_infrastructure")
from chip.testing.tasks import Subprocess
```

I'm not able to test it, though, because I do not have the `chip-cert-bins` Docker image yet (I'm building it now).

src/python_testing/TC_MCORE_FS_1_1.py
andy31415: These are workarounds of not having the chip test wheel installed. Test harness should install the wheel.

arkq (approved):
Approving in case when the try/except will be merged. Also (after a Slack chat), I will prepare a proper fix (pip installing chip-testing package) for master.

rquidute:
Closing because we got a better solution as discussed offline.

PR #36240, opened 2024-10-24
[Fabric-Sync] Add startup script

yufengwangca:
The goal of this script is similar to fabric-admin/scripts/run_fabric_sync.sh. It is designed to facilitate a seamless transition from fabric-admin/fabric-bridge-app to fabric-sync. This script streamlines the process by handling the cleanup (rm -rf /tmp/chip_*) and ensuring the appropriate instance of fabric-sync runs.

### DIFF
--- a/examples/fabric-sync/scripts/run_fabric_sync.sh
+++ b/examples/fabric-sync/scripts/run_fabric_sync.sh
@@ -1,0 +1,59 @@
+#!/bin/bash
+
+# Default paths
+DEFAULT_SYNC_CHOICES=(
+    "./fabric-sync"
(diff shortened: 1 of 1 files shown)


andy31415 (changes requested):
Please update the summary of this PR explaining why we add such a script.

To me this looks like a `rm -rf /tmp/chip_* && ./fabric-sync` (with some path finding rules on finding the fabric sync binary). This is a lot of lines of shell for this...

yufengwangca:
after we merge fabric-admin/fabric-bridge, this scripts is not necessary, just keep to old test setup working without change

PR #36241, opened 2024-10-25
[Fabric-Sync] Port 'add-bridge' and 'remove-bridge' commands

yufengwangca:
Port 'add-bridge' and 'remove-bridge' commands from fabric-admin to fabric-sync app.

### DIFF
--- a/examples/fabric-sync/admin/DeviceManager.cpp
+++ b/examples/fabric-sync/admin/DeviceManager.cpp
@@ -80,6 +80,12 @@ CHIP_ERROR DeviceManager::PairRemoteFabricBridge(NodeId nodeId, uint32_t setupPI
 
 CHIP_ERROR DeviceManager::UnpairRemoteFabricBridge()
 {
+    if (mRemoteBridgeNodeId == kUndefinedNodeId)
+    {
+        ChipLogError(NotSpecified, "Remote bridge node ID is undefined; cannot unpair device.");
+        return CHIP_ERROR_INCORRECT_STATE;
+    }
+
     CHIP_ERROR err = PairingManager::Instance().UnpairDevice(mRemoteBridgeNodeId);
     if (err != CHIP_NO_ERROR)
     {
--- a/examples/fabric-sync/admin/DeviceManager.cpp
+++ b/examples/fabric-sync/admin/DeviceManager.cpp
@@ -1,0 +1,105 @@
+/*
+ *
+ *    Copyright (c) 2024 Project CHIP Authors
+ *    All rights reserved.
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+#include "DeviceManager.h"
+
+#include <crypto/RandUtils.h>
+#include <lib/support/StringBuilder.h>
+
+#include <cstdio>
+#include <string>
+
+using namespace chip;
+
+// Define the static member
+DeviceManager DeviceManager::sInstance;
+
+void DeviceManager::Init()
+{
+    // TODO: (#34113) Init mLastUsedNodeId from chip config file
+    mLastUsedNodeId = 1;
+    mInitialized    = true;
+
+    ChipLogProgress(NotSpecified, "DeviceManager initialized: last used nodeId " ChipLogFormatX64,
+                    ChipLogValueX64(mLastUsedNodeId));
+}
+
+NodeId DeviceManager::GetNextAvailableNodeId()
+{
+    mLastUsedNodeId++;
+    VerifyOrDieWithMsg(mLastUsedNodeId < std::numeric_limits<NodeId>::max(), NotSpecified, "No more available NodeIds.");
+
+    return mLastUsedNodeId;
+}
+
+void DeviceManager::UpdateLastUsedNodeId(NodeId nodeId)
+{
+    if (nodeId > mLastUsedNodeId)
+    {
+        mLastUsedNodeId = nodeId;
+        ChipLogProgress(NotSpecified, "Updating last used NodeId to " ChipLogFormatX64, ChipLogValueX64(mLastUsedNodeId));
+    }
+}
+
+void DeviceManager::SetRemoteBridgeNodeId(chip::NodeId nodeId)
+{
+    mRemoteBridgeNodeId = nodeId;
+}
+
+CHIP_ERROR DeviceManager::PairRemoteFabricBridge(NodeId nodeId, uint32_t setupPINCode, const char * deviceRemoteIp,
+                                                 uint16_t deviceRemotePort)
+{
+    CHIP_ERROR err = PairingManager::Instance().PairDevice(nodeId, setupPINCode, deviceRemoteIp, deviceRemotePort);
+
+    if (err != CHIP_NO_ERROR)
+    {
+        ChipLogError(NotSpecified,
+                     "Failed to pair remote fabric bridge: Node ID " ChipLogFormatX64 " with error: %" CHIP_ERROR_FORMAT,
+                     ChipLogValueX64(nodeId), err.Format());
+        return err;
+    }
+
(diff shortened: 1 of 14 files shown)


examples/fabric-sync/main.cpp
```diff
@@ -35,6 +40,21 @@ int main(int argc, char * argv[])
 
     VerifyOrDie(ChipLinuxAppInit(argc, argv) == 0);
 
+#if defined(ENABLE_CHIP_SHELL)
+#if CHIP_DEVICE_CONFIG_APP_PLATFORM_ENABLED
```
tehampson, line 44: Not sure I follow why there is a check for `CHIP_DEVICE_CONFIG_APP_PLATFORM_ENABLED`

examples/fabric-sync/shell/ShellCommands.cpp
```diff
@@ -0,0 +1,131 @@
...
+#include <lib/support/CHIPArgParser.hpp>
+#include <lib/support/CHIPMem.h>
+#include <lib/support/CodeUtils.h>
+#include <platform/CHIPDeviceLayer.h>
+
+namespace chip {
+namespace Shell {
+
+// Global pointer to hold the AddBridgeCommand instance
+static std::unique_ptr<AddBridgeCommand> gAddBridgeCommand;
```
tehampson, line 37: What is this for?

examples/fabric-sync/admin/DeviceManager.h
```diff
@@ -0,0 +1,99 @@
...
+ */
+
+#pragma once
+
+#include "PairingManager.h"
+
+#include <app-common/zap-generated/cluster-objects.h>
+#include <platform/CHIPDeviceLayer.h>
+
+class DeviceManager : public PairingDelegate
```
tehampson, line 26: I always thought that this class could benefit with documentation to let reader know this manages devices on the remote bridge and not on the local bridge.

Do you think it is worth adding now?

examples/fabric-sync/admin/DeviceManager.cpp
```diff
@@ -0,0 +1,105 @@
...
+        return err;
+    }
+
+    ChipLogProgress(NotSpecified, "Successfully paired remote fabric bridge: Node ID " ChipLogFormatX64, ChipLogValueX64(nodeId));
+    return CHIP_NO_ERROR;
+}
+
+CHIP_ERROR DeviceManager::UnpairRemoteFabricBridge()
+{
+    CHIP_ERROR err = PairingManager::Instance().UnpairDevice(mRemoteBridgeNodeId);
```
tehampson, line 83: Should we check that mRemoteBridgeNodeId != kUndefinedNodeId?

examples/fabric-sync/admin/DeviceManager.cpp
```diff
@@ -0,0 +1,105 @@
...
+    if (err != CHIP_NO_ERROR)
+    {
+        ChipLogError(NotSpecified, "Failed to unpair remote bridge device " ChipLogFormatX64, ChipLogValueX64(mRemoteBridgeNodeId));
+        return err;
+    }
+
+    ChipLogProgress(NotSpecified, "Successfully unpaired remote fabric bridge: Node ID " ChipLogFormatX64,
```
tehampson, line 90: If successful, should we set the value of mRemoteBridgeNodeId to kUndefinedNodeId. Or is that expected to be done by calling UnpairDevice?

examples/fabric-sync/shell/AddBridgeCommand.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+CHIP_ERROR AddBridgeCommand::RunCommand()
+{
+    if (DeviceMgr().IsFabricSyncReady())
+    {
+        // print to console
+        Shell::streamer_t * sout = Shell::streamer_get();
+        Shell::streamer_printf(sout, "Remote Fabric Bridge has already been configured.\n");
+
+        return CHIP_NO_ERROR;
```
tehampson, line 76: It is not clear to me why this isn't an error like BUSY or something. Someone might be calling add bridge with different credentials and this success might give the wrong impression that it was successful

examples/fabric-sync/shell/AddBridgeCommand.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+#include "AddBridgeCommand.h"
+
+#include <admin/DeviceManager.h>
+#include <lib/shell/streamer.h>
+
+using namespace ::chip;
+
+// Define the global unique pointer
+std::unique_ptr<AddBridgeCommand> gAddBridgeCmd = nullptr;
```
tehampson, line 27: Is there any way we can avoid a global here? If not can we added a comment on why things need to be this way?

examples/fabric-sync/shell/RemoveBridgeCommand.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+#include "RemoveBridgeCommand.h"
+
+#include <admin/DeviceManager.h>
+#include <lib/shell/streamer.h>
+
+using namespace ::chip;
+
+// Define the global unique pointer
+std::unique_ptr<RemoveBridgeCommand> gRemoveBridgeCmd = nullptr;
```
tehampson, line 27: Same, Is there any way we can avoid a global here? If not, can we added a comment on why things need to be this way?

examples/fabric-sync/shell/ShellCommands.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+static CHIP_ERROR HandleRemoveBridgeCommand(int argc, char ** argv)
+{
+    if (argc != 1)
+    {
+        fprintf(stderr, "Invalid arguments. Usage: app remove-bridge\n");
+        return CHIP_ERROR_INVALID_ARGUMENT;
+    }
+
+    gRemoveBridgeCmd.reset();
```
tehampson, line 82: Is this safe? What happens if you reset here, but the RunCommand below fails and doesn't end up calling SetPairingDelegate. Would you have a use after free?

Essentially should there be a dtor for RemoveBridgeCommand

examples/fabric-sync/shell/ShellCommands.cpp
```diff
@@ -0,0 +1,131 @@
...
+    {
+        return HandleAddBridgeCommand(argc, argv);
+    }
+    else if (strcmp(argv[0], "remove-bridge") == 0)
+    {
+        return HandleRemoveBridgeCommand(argc, argv);
+    }
+    else if (strcmp(argv[0], "sync-device") == 0)
+    {
+        // TODO
```
tehampson, line 110: empty TODO

examples/fabric-sync/main.cpp
yufengwangca: The Shell commission commands depend on examples/platform/linux/CommissionerMain.cpp to implement their functionalities, which require this flag. Without it, the following compilation errors will occur:

```
../../../examples/fabric-sync/third_party/connectedhomeip/examples/platform/linux/CommissionerMain.cpp: In constructor ‘PairingCommand::PairingCommand()’:
../../../examples/fabric-sync/third_party/connectedhomeip/examples/platform/linux/CommissionerMain.cpp:243:9: error: class ‘PairingCommand’ does not have any field named ‘mOnDeviceConnectedCallback’
  243 |         mOnDeviceConnectedCallback(OnDeviceConnectedFn, this),
```

tehampson: Would some sort of static assert here be better? Is the app going to have some functionality without being able to call `Shell::RegisterCommands();`?

Essentially if I enable `ENABLE_CHIP_SHELL`, but don't enable `CHIP_DEVICE_CONFIG_APP_PLATFORM_ENABLED`, why do we let things still compile that prevents `ENABLE_CHIP_SHELL` from working?

examples/fabric-sync/admin/DeviceManager.cpp
yufengwangca: This is done in void RemoveBridgeCommand::OnDeviceRemoved(NodeId deviceId, CHIP_ERROR err)

examples/fabric-sync/admin/DeviceManager.cpp
yufengwangca: acked

examples/fabric-sync/shell/ShellCommands.cpp
yufengwangca: removed

examples/fabric-sync/admin/DeviceManager.h
yufengwangca: In fabric-admin, DeviceManager manages both local bridge and remote bridge, in fabric-sync, it currently only manages the remote bridge for 1-1 fabric-sync, but we might also need to manage the local bridge later, lets keep it as it is now.

tehampson (approved):

joonhaengHeo (approved):

PR #36242, opened 2024-10-25
FanControl: modify percentSetting/percentSpeedSetting return value in case NULL is written (match spec)

arun-silabs:
As per the spec, in fan-control cluster, when ```NULL``` is written to ```percentSetting``` attribute or ```speedSetting``` attribute by client, it should return INVALID_IN_STATE. But currently WRITE_IGNORED is being returned.

image

The return value in these 2 cases is corrected in this PR.


src/app/clusters/fan-control-server/fan-control-server.cpp
```diff
@@ -198,7 +198,7 @@ MatterFanControlClusterServerPreAttributeChangedCallback(const ConcreteAttribute
                 }
                 else
                 {
-                    res = Status::WriteIgnored;
+                    res = Status::InvalidInState;
```
bzbarsky-apple, line 201: We apparently have no test coverage, since nothing started failing.

Please add a basic YAML test that exercises this.

arun-silabs: I've modified the "TestFanControl.yaml" file to include coverage for this particular case.

bzbarsky-apple (approved):

src/app/tests/suites/TestFanControl.yaml
```diff
@@ -69,6 +69,8 @@ tests:
       attribute: "PercentSetting"
       arguments:
           value: null
+      response:
+          error: INVALID_IN_STATE
```
tcarmelveilleux, line 73: We need to add a cert test for this in TC-FAN-XXXX. Please open an issue to update a test script plan and create SDK issue to update test.

tcarmelveilleux (approved):

src/app/tests/suites/TestFanControl.yaml
arun-silabs: This is already covered in the fan control cert test python scripts, @yinyihu-silabs confirmed it.

yinyihu-silabs: [TC_FAN_3_1.py](https://github.com/project-chip/connectedhomeip/blob/master/src/python_testing/TC_FAN_3_1.py)
I understand there is plan to update fan cluster test plan.

PR #36244, opened 2024-10-25
[darwin-framework-tool] [darwin-framework-tool] Add an optional parameter to the storage view command to only show the entries for a given commissioner

vivien-apple:
#### Problem

This PR adds an option to filter by controller (`identity [alpha, beta, gamma]`) when inspecting the contents of per-controller storage, making it easier to navigate. It also improves consistency in the use of `fabricId` alongside `controllerId`.

### DIFF
--- a/examples/darwin-framework-tool/commands/common/ControllerStorage.mm
+++ b/examples/darwin-framework-tool/commands/common/ControllerStorage.mm
@@ -143,6 +143,20 @@ NSString * const kDarwinFrameworkToolControllerDomain = @"com.apple.darwin-frame
     self.storage[controllerKey] = value;
 }
 
+- (void)print
+{
+    NSLog(@"%@ (%@)", kDarwinFrameworkToolControllerDomain, _keyScopingPrefix);
+    for (NSString * controllerKey in self.storage) {
+        if (![self _isControllerScopedKey:controllerKey]) {
+            continue;
+        }
+
+        __auto_type * key = [self _controllerScopedKeyToKey:controllerKey];
+        __auto_type * data = self.storage[controllerKey];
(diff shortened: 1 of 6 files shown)


kiel-apple (approved):

examples/darwin-framework-tool/commands/common/ControllerStorage.mm
```diff
@@ -143,6 +143,20 @@ - (void)storeValue:(NSData *)value forKey:key
...
+- (void)print
+{
+    NSLog(@"%@ (%@)", kDarwinFrameworkToolControllerDomain, _keyScopingPrefix);
+    for (NSString * controllerKey in self.storage) {
+        if (![self _isControllerScopedKey:controllerKey]) {
+            continue;
+        }
+
+        __auto_type * key = [self _controllerScopedKeyToKey:controllerKey];
+        __auto_type * data = [self.storage objectForKeyedSubscript:controllerKey];
```
bzbarsky-apple, line 155: Shouldn't that just use `objectForKey`?

bzbarsky-apple (approved):

examples/darwin-framework-tool/commands/common/ControllerStorage.mm
vivien-apple: > Shouldn't that just use `objectForKey`?

No. But that is true that I can simplify it: `__auto_type * data = self.storage[controllerKey]`

vivien-apple: I have pushed an updated version.

bzbarsky-apple:

PR #36247, opened 2024-10-25
Update First/Next accepted command iteration in CodegenDataModelProvider to save flash

andy31415:
This saves about 160 bytes of flash on an nrf light build.

#### Changes

- drop `FindCommandEntry` and always use `FindCommandId` - this saves about 136 bytes
- re-use some code for first/next/exact finding in a new `FindCommand` method with quite a few arguments. This saves another 24 bytes (really minute given the delta, however I figured since I made the changes I may as well keep them)

### DIFF
--- a/src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
+++ b/src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
@@ -14,6 +14,7 @@
  *    See the License for the specific language governing permissions and
  *    limitations under the License.
  */
+#include "lib/support/CodeUtils.h"
 #include <app/codegen-data-model-provider/CodegenDataModelProvider.h>
 
 #include <access/AccessControl.h>
@@ -77,9 +78,6 @@ public:
     ///    - kInvalidCommandId if the find failed (but command handler interface does provide a list)
     ///    - valid id if command handler interface usage succeeds
     std::optional<CommandId> FindCommandId(Operation operation, const ConcreteCommandPath & path);
-
-    /// Uses FindCommandId to find the given command and loads the command entry data
-    std::optional<DataModel::CommandEntry> FindCommandEntry(Operation operation, const ConcreteCommandPath & path);
 
 private:
     HandlerCallbackFunction mCallback;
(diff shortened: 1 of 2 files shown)


src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
```diff
@@ -682,106 +662,58 @@ std::optional<DataModel::AttributeInfo> CodegenDataModelProvider::GetAttributeIn
...
+    EnumeratorCommandFinder handlerFinder(&CommandHandlerInterface::EnumerateGeneratedCommands);
 
-    std::optional<CommandId> commandId = mGeneratedCommandsIterator.Next(cluster->generatedCommandList, before.mCommandId);
-    VerifyOrReturnValue(commandId.has_value(), kInvalidCommandPath);
+    CommandId commandId = FindCommand(before, handlerFinder, detail::EnumeratorCommandFinder::Operation::kFindNext,
+                                      mGeneratedCommandsIterator, GeneratedCommands);
 
-    return ConcreteCommandPath(before.mEndpointId, before.mClusterId, *commandId);
+    VerifyOrReturnValue(commandId != kInvalidCommandId, kInvalidCommandPath);
+    return ConcreteCommandPath(before.mEndpointId, before.mClusterId, commandId);
```
bzbarsky-apple, line 716: Followup: Doing the return-by-value of the whole path, not just the command id which is the one thing caller does not have, seems a bit space-wasteful too....

bzbarsky-apple (approved):

arkq (approved):

src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
andy31415: I can take a look. When I had both command and path versions it was obviously smaller, so I picked the commandid return because this is what "generated commands" needs. That was obviously larger than a single method, hence this updated code.

bzbarsky-apple: I guess my point is: why do we need to return the whole path at all?

andy31415: It made the API easier to iterate through. All other outputs have a path embedded even though they realistically do not need it and only the ID is needed.

None of the APIs recurse, so they would not really need paths, however ergonomically having full "path + data" seems clearer than "id without prefix + data".

PR #36249, opened 2024-10-25
Make datamodel-provider inteface logging optional/configurable

andy31415:
CodegenDatamodelProvider (and DataModel::Provider) has more checks and error logs than the methods it replaces. Since CodegenDataModel switch in #36246 has concerns on codesize, make this extra logging configurable.

Logging is enabled on known large platforms (linux, darwin and android) and disabled otherwise. Depending on flash amount, we may want to start enabling this on other platforms.

This saves about 490 bytes of flash when the extra logging is disabled.

### DIFF
--- a/src/app/common_flags.gni
+++ b/src/app/common_flags.gni
@@ -45,5 +45,6 @@ declare_args() {
   # This is an optimization for small-flash size devices as extra logging requires
   # additional flash for messages & code for formatting.
   chip_data_model_extra_logging =
-      current_os == "linux" || current_os == "ios" || current_os == "mac" || current_os == "android"
+      current_os == "linux" || current_os == "ios" || current_os == "mac" ||
+      current_os == "android"
 }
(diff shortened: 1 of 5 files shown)


src/app/common_flags.gni
```diff
@@ -40,4 +40,10 @@ declare_args() {
   # If/once the chip_use_data_model_interface flag is removed or does not support
   # a `check` option, this should alwo be removed
   chip_data_model_check_die_on_failure = false
+
+  # This controls if more logging is supposed to be enabled into the data models.
+  # This is an optimization for small-flash size devices as extra logging requires
+  # additional flash for messages & code for formatting.
+  chip_data_model_extra_logging =
+      current_os == "linux" || current_os == "darwin" || current_os == "android"
```
bzbarsky-apple, line 48:
```suggestion
      current_os == "linux" || current_os == "ios" || current_os == "mac" || current_os == "android"
```
`darwin` is not a value of `current_os` at all; it's a value of `chip_device_platform`.  I wish ninja had better typing for this stuff.

bzbarsky-apple (approved):

yufengwangca (approved):

PR #36251, opened 2024-10-25
Event validation simplification - pull logic into InteractionModelEngine.cpp only

andy31415:
Since code generation now does not have eventlist attribute, ember code cannot be used to provide any extra validation compared to existing iteration though endpoints and clusters.

Moved the code for event path validation back to interactionmodelengine and provided a DataModel::Provider implementation for it.


src/app/InteractionModelEngine.cpp
```diff
@@ -53,12 +57,148 @@
...
+namespace {
+
+/**
+ * Helper to handle wildcard events in the event path.
+ *
+ * Validates that ACL access is permitted to:
+ *    - Cluster::View in case the path is a wildcard for the event id
+ *    - Event read if the path is a concrete event path
+ */
+bool HasValidEventPathForEndpointAndCluster(const ConcreteClusterPath & path, const EventPathParams & aEventPath,
```
bzbarsky-apple, line 71: I know you based this naming on the naming of the variables in the IM engine, but I would call this function something like "MayHaveAccessibleEventPathForEndpointAndCluster".

It's not checking "valid", it's checking "something you could access".  Path could be valid but ACL denying access....

It's also not an exact check, so HasAccessibleEventPathForEndpointAndCluster would not be quite right.  If this function returns true, you might have such a path, or you might not.  But if it returns false, you certainly don't.

I would suggest similar naming adjustments to the other functions here.

The original "valid" naming is in a very narrow context of "valid to subscribe to", but these functions are sort of divorced from that meaning as standalone functions....

andy31415: Unresolving to rename all ...

tehampson (approved):

arkq (approved):

PR #36253, opened 2024-10-25
Move decodable lists bits as a non-template class to save flash

andy31415:
This looks like significant savings, since decode takes a lot of space if templated and it is not actually template-dependent. On a NRF light this looks like 6K of flash savings.

image

### DIFF
--- a/src/app/data-model/DecodableList.h
+++ b/src/app/data-model/DecodableList.h
@@ -27,6 +27,68 @@ namespace chip {
 namespace app {
 namespace DataModel {
 
+namespace detail {
+
+/*
+ * Base class of DecodableList to minimize template usage
+ */
+class DecodableListBase
+{
+public:
+    DecodableListBase() { ClearReader(); }
+
+    /*
+     * @brief
+     *
+     * This call stores a TLV reader positioned on the list this class is to manage.
+     *
+     * Specifically, the passed-in reader should be pointing into the list just after
+     * having called `OpenContainer` on the list element.
+     */
+    void SetReader(const TLV::TLVReader & reader) { mReader = reader; }
+
+    /*
+     * @brief
+     *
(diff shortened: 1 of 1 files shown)


src/app/data-model/DecodableList.h
```diff
@@ -47,30 +109,13 @@ namespace DataModel {
...
-    /*
-     * @brief
-     *
-     * This call clears the TLV reader managed by this class, so it can be reused.
-     */
-    void ClearReader() { mReader.Init(nullptr, 0); }
-
     template <typename T0 = T, std::enable_if_t<DataModel::IsFabricScoped<T0>::value, bool> = true>
     void SetFabricIndex(FabricIndex fabricIndex)
     {
```
bzbarsky-apple, line 121: Followup: we might get some more wins by moving bits of the iterator into a non-templated super-class too.

bzbarsky-apple (approved):
Thank you for fixing this stuff up!

arkq (approved):

src/app/data-model/DecodableList.h
```diff
@@ -47,30 +109,13 @@ namespace DataModel {
...
-
-    /*
-     * @brief
-     *
-     * This call clears the TLV reader managed by this class, so it can be reused.
-     */
-    void ClearReader() { mReader.Init(nullptr, 0); }
-
     template <typename T0 = T, std::enable_if_t<DataModel::IsFabricScoped<T0>::value, bool> = true>
     void SetFabricIndex(FabricIndex fabricIndex)
```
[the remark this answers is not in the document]
andy31415: created #36264 for followups

src/app/data-model/DecodableList.h
andy31415: created #36264 for followups

PR #36256, opened 2024-10-25
chef: Fix FanControl mode/speed interdependence behavior

soares-sergio:
This PR implements the FanControl logic to ensure the interdependent behavior between Mode/Speed/Percent is coherent.

This is based on what was done in the AirPurifier examples.

Changes:
* Removed Custom Write/Read
* Updated Zap file `SpeedMax` default to 10 instead of 100
* Forward `MatterPostAttributeChangeCallback` FanControl-related changes to `ChefFanControlManager`
* Implements Mode/Speed/Percent interdependency logic based on the AirPurifier example

Test:
* Verified that Mode, Speed and SpeedPercent change accordingly when any of those attributes change and cross a boundary.


bzbarsky-apple (changes requested):

examples/chef/common/chef-fan-control-manager.cpp
```diff
@@ -99,98 +112,223 @@ Status ChefFanControlManager::HandleStep(StepDirectionEnum aDirection, bool aWra
...
+        }
+        break;
+    }
+    case FanControl::FanModeEnum::kSmart:
+    case FanControl::FanModeEnum::kAuto: {
+        ChipLogProgress(NotSpecified, "ChefFanControlManager::FanModeWriteCallback: Auto");
+        break;
+    }
+    case FanControl::FanModeEnum::kUnknownEnumValue: {
+        ChipLogProgress(NotSpecified, "ChefFanControlManager::FanModeWriteCallback: Unknown");
```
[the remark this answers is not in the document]
soares-sergio: I'm new to this code but it looks like the cluster code only handles `kSmart` and `kOn`  on pre-callback. Everything else it just passes down: https://github.com/project-chip/connectedhomeip/blob/master/src/app/clusters/fan-control-server/fan-control-server.cpp#L183

Are you suggesting I go ahead and update the cluster code to check if it's `kUnknownValue` and return `Status::ConstraintError` instead of passing down?
I could do that. But searching the codebase, I don't see other clusters handling this on pre-callback, so I'm not sure what's the rationale here. Was the intention to just let the app/products handle this, maybe?

examples/chef/common/chef-fan-control-manager.cpp
```diff
@@ -99,98 +112,223 @@ Status ChefFanControlManager::HandleStep(StepDirectionEnum aDirection, bool aWra
...
 
-        // Ensure new speed in percent is valid.
-        if (!newPercentSetting.IsNull() && newPercentSetting.Value() > 100)
+    default: {
+        break;
+    }
+    }
+}
+
+void ChefFanControlManager::PercentSettingWriteCallback(uint8_t aNewPercentSetting)
```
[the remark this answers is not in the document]
soares-sergio: > Looking at the spec, none of this code should be reached if null is written (cluster should ignore the write and return with an error per spec), no?

It seems this is only true when the `client` is trying to write to null.
The cluster code can still set this to null and we use the post attribute change callback to detect the change in PercentSetting so we can update PercentCurrent.

See: https://github.com/project-chip/connectedhomeip/blob/master/src/app/clusters/fan-control-server/fan-control-server.cpp#L232

examples/chef/common/chef-fan-control-manager.cpp
```diff
@@ -99,98 +112,223 @@ Status ChefFanControlManager::HandleStep(StepDirectionEnum aDirection, bool aWra
...
+    switch (attributeId)
     {
-    case SpeedSetting::Id: {
-        Nullable<uint8_t> newSpeedSetting;
-        ReturnErrorOnFailure(aDecoder.Decode(newSpeedSetting));
-
-        // Ensure new speed is in bounds
+    case FanControl::Attributes::PercentSetting::Id: {
+        ChipLogProgress(NotSpecified, "ChefFanControlManager::HandleFanControlAttributeChange  PercentSetting");
+        DataModel::Nullable<Percent> percentSetting = static_cast<DataModel::Nullable<uint8_t>>(*value);
```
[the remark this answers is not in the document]
soares-sergio: I'm sorry, I'm new to the codebase and not yet familiar with all of the types yet. I followed what's currently done in the `air-purifier-app`, which is likely wrong as well: https://github.com/project-chip/connectedhomeip/blob/master/examples/air-purifier-app/air-purifier-common/src/air-purifier-manager.cpp#L157

> Was this tested at all?
Yes, it was tested and it seems to have coherent values.

I would appreciate if you could point me to a good example and I can update both of the incorrect apps. Thanks!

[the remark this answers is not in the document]
bzbarsky-apple: Yeah, that air purifier code is definitely wrong too....

When this was tested, was it tested with an actual null value?

What I would is that the actual type at the callsite of `value` here is `uint8_t`, with null represented as `0xFF`.

examples/chef/common/chef-fan-control-manager.cpp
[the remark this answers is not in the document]
bzbarsky-apple: > I'm new to this code but it looks like the cluster code only handles kSmart and kOn on pre-callback. 

That sounds like a bug in the cluster code; it should deal with invalid input in the way the spec specifies, so that applications do not have to.

>

examples/chef/common/chef-fan-control-manager.cpp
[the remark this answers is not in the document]
soares-sergio: Yeah, the spec doesn't specify what to do with `PercentCurrent` when `PercentSetting` goes to null (a.k.a when the FanMode goes to kAuto).
We're deciding to set this to 0 in the app (as `PercentCurrent` can't be null as well).

examples/chef/common/chef-fan-control-manager.cpp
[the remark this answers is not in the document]
soares-sergio: I updated the PR, let me know if it looks better now. I needed a little lesson from Andrei on how ember encodes null. Thanks for catching this, I'll fix the other app in a separate PR.

examples/chef/common/chef-fan-control-manager.cpp
```diff
@@ -99,98 +112,223 @@ Status ChefFanControlManager::HandleStep(StepDirectionEnum aDirection, bool aWra
...
+    }
 
-            // Mark both the setting AND the current dirty, since the current always
-            // tracks the target for our product.
-            MatterReportingAttributeChangeCallback(mEndpoint, FanControl::Id, Attributes::SpeedSetting::Id);
-            MatterReportingAttributeChangeCallback(mEndpoint, FanControl::Id, Attributes::SpeedCurrent::Id);
+    case FanControl::Attributes::SpeedSetting::Id: {
+        ChipLogProgress(NotSpecified, "ChefFanControlManager::HandleFanControlAttributeChange  SpeedSetting");
+
+        DataModel::Nullable<uint8_t> speedSetting = static_cast<DataModel::Nullable<uint8_t>>(*value);
```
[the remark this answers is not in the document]
soares-sergio: Should be fixed now.

examples/chef/common/chef-fan-control-manager.cpp
```diff
@@ -99,98 +113,205 @@ Status ChefFanControlManager::HandleStep(StepDirectionEnum aDirection, bool aWra
...
+        // value for the Fan percentage. The spec doesn't specify what to do in this case.
+        // If not Null, PercentCurrent tracks PercentSetting's value.
+        SetPercentCurrent(percentSetting.IsNull() ? 0 : percentSetting.Value());
+        break;
+    }
+
+    case FanControl::Attributes::SpeedSetting::Id: {
+        ChipLogProgress(NotSpecified, "ChefFanControlManager::HandleFanControlAttributeChange  SpeedSetting");
+
+        // DataModel::Nullable<uint8_t> speedSetting = static_cast<DataModel::Nullable<uint8_t>>(*value);
```
andy31415, line 137:
```suggestion
```

examples/chef/common/chef-fan-control-manager.cpp
```diff
@@ -99,98 +113,205 @@ Status ChefFanControlManager::HandleStep(StepDirectionEnum aDirection, bool aWra
...
+}
 
-    switch (aPath.mAttributeId)
+FanControl::FanModeEnum ChefFanControlManager::SpeedToFanMode(uint8_t speed)
+{
+    if (speed == 0)
+    {
+        return FanControl::FanModeEnum::kOff;
+    }
+    else if (speed <= FAN_MODE_LOW_UPPER_BOUND)
```
andy31415, line 165: clang-tidy does not run on this, however if it would, it would complain and ask for `no else after return` (https://clang.llvm.org/extra/clang-tidy/checks/readability/else-after-return.html)

examples/chef/common/chef-fan-control-manager.cpp
```diff
@@ -99,98 +113,205 @@ Status ChefFanControlManager::HandleStep(StepDirectionEnum aDirection, bool aWra
...
-        ReturnErrorOnFailure(aDecoder.Decode(newSpeedSetting));
+        return FanControl::FanModeEnum::kHigh;
+    }
+    return FanControl::FanModeEnum::kUnknownEnumValue;
+}
 
-        // Ensure new speed is in bounds
+void ChefFanControlManager::SetPercentCurrent(uint8_t aNewPercentCurrent)
+{
+    if (aNewPercentCurrent != mPercentCurrent)
```
andy31415, line 182: nit: you could decrease indend by doing:

```cpp
if (aNewPercentCurrent != mPercentCurrent)
{
  return;
}
// .. code that is now not indented
```

examples/chef/common/chef-fan-control-manager.cpp
```diff
@@ -99,98 +113,205 @@ Status ChefFanControlManager::HandleStep(StepDirectionEnum aDirection, bool aWra
...
-            Protocols::InteractionModel::Status status = SpeedMax::Get(mEndpoint, &maxSpeedSetting);
-            VerifyOrReturnError(status == Protocols::InteractionModel::Status::Success, CHIP_IM_GLOBAL_STATUS(Failure));
-
-            if (!newSpeedSetting.IsNull() && newSpeedSetting.Value() > maxSpeedSetting)
-            {
-                return CHIP_IM_GLOBAL_STATUS(ConstraintError);
-            }
+            ChipLogError(NotSpecified, "ChefFanControlManager::SetPercentCurrent: failed to set PercentCurrent attribute: %d",
+                         to_underlying(status));
+            return;
```
andy31415, line 191: this has a return and SetSpeedCurrent does not. I would make it consistent (maybe decrease indent in both and no return, since we use this status for log only)

examples/chef/common/chef-fan-control-manager.cpp
soares-sergio: Good idea. Fixed, thanks!

examples/chef/common/chef-fan-control-manager.cpp
soares-sergio: Fixed thanks!

examples/chef/common/chef-fan-control-manager.cpp
soares-sergio: Right, made them consistent. Thanks!

examples/chef/common/chef-fan-control-manager.cpp
```diff
@@ -99,98 +113,205 @@ Status ChefFanControlManager::HandleStep(StepDirectionEnum aDirection, bool aWra
...
-            MatterReportingAttributeChangeCallback(mEndpoint, FanControl::Id, Attributes::SpeedCurrent::Id);
+            ChipLogError(NotSpecified, "ChefFanControlManager::SetSpeedCurrent: failed to set SpeedCurrent attribute: %d",
+                         to_underlying(status));
         }
+    }
+}
 
+void ChefFanControlManager::FanModeWriteCallback(FanControl::FanModeEnum aNewFanMode)
+{
+    ChipLogDetail(NotSpecified, "ChefFanControlManager::FanModeWriteCallback: %d", (uint8_t) aNewFanMode);
```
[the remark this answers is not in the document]
soares-sergio: Fixed, thanks!

examples/chef/common/chef-fan-control-manager.cpp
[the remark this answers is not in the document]
bzbarsky-apple: @soares-sergio The right behavior is that in auto mode PercentCurrent should reflect the actual speed of the fan, as a percentage of max, no?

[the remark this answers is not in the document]
soares-sergio: Yes!

In a real application, I would expect the app to keep updating `SpeedCurrent` and `PercentCurrent` to the real Fan speed/percentage while in auto mode.
For this example app, we simplified to writing to a fixed value (0, could be any other valid value).

A more sophisticated approach for this Example could be to build an "Auto" mode simulator that's enabled when FanMode is set to `Auto` and then automatically update `SpeedCurrent` and `PercentCurrent` (maybe keep looping through a few speeds?).

Is that what you're suggesting?

examples/chef/common/chef-fan-control-manager.cpp
```diff
@@ -38,20 +38,34 @@ using namespace chip::app::Clusters::FanControl::Attributes;
...
+    uint8_t mPercentCurrent;
+    uint8_t mSpeedCurrent;
+
+    // Fan Mode Limits
+    static constexpr int FAN_MODE_LOW_LOWER_BOUND    = 1;
+    static constexpr int FAN_MODE_LOW_UPPER_BOUND    = 3;
+    static constexpr int FAN_MODE_MEDIUM_LOWER_BOUND = 4;
+    static constexpr int FAN_MODE_MEDIUM_UPPER_BOUND = 7;
+    static constexpr int FAN_MODE_HIGH_LOWER_BOUND   = 8;
+    static constexpr int FAN_MODE_HIGH_UPPER_BOUND   = 10;
```
andy31415, line 62: nit: we generally use `kCamelCase` for C++ constant naming

andy31415: I also wonder if we should generally have suffix for things .... there is no real unit of measurement here, but maybe `kFanModeLowSpeedLowerBound` (i.e. speed suffix) would make code obvious in `if (speed <= kFanModeLowSpeedLowerBound)`

andy31415: Maybe outside of this PR, but for maintainability, it seems that these are a `range` so I would want to have code like:

```cpp
constexpr Range kLowFanSpeed = {1, 3};
constexpr Range kMediumFanSpeed = {4, 7};
constexpr Range kHighFanSpeed = {8, 10};

static_assert(kLowFanSpeed.High() + 1 == kMediumFanSpeed.Low());
static_assert(kMediumFanSpeed.High() + 1 == kHighFanSpeed.Low());
// ....

if (kLowFanSpeed.Contains(speed))
{
    return FanControl::FanModeEnum::kLow;
}

// ...

if (!kLowFanSpeed.Contains(mSpeedCurrent))
{
    DataModel::Nullable<uint8_t> speedSetting(kLowFanSpeed.Low());
    // ...
```

examples/chef/common/chef-fan-control-manager.cpp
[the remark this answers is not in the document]
bzbarsky-apple: > For this example app, we simplified to writing to a fixed value (0, could be any other valid value).

Ah, I see.  I think that's fine but should be documented, because it looked like "it's null so we don't know what to do" vs "it's null and this is the thing we are doing for this reason".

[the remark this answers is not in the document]
soares-sergio: Cool. Added a comment in the code, where we set those to 0.
```
        // The cluster code in fan-control-server.cpp is the only one allowed to set speedSetting to null.
        // This happens as a consequence of setting the FanMode to kAuto. In auto mode, speedCurrent should continue to report the
        // real fan speed. In this example, we set SpeedCurrent to 0 here as we don't have a real value for the Fan speed or a
        // FanAutoMode simulator.
        // When not Null, SpeedCurrent tracks SpeedSetting's value.
```

I'll add an AutoMode simulator in a future PR. I think it's a good idea anyway to provide an example for that. Thanks!

examples/chef/common/chef-fan-control-manager.cpp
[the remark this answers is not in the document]
soares-sergio: Makes sense. I'll address that in a separate PR as I don't want to mix changes to cluster code and example code together.

examples/chef/common/chef-fan-control-manager.cpp
soares-sergio: Good ideas, done all that! Thanks Andrei.

examples/chef/common/chef-fan-control-manager.cpp
```diff
@@ -38,20 +38,45 @@ using namespace chip::app::Clusters::FanControl::Attributes;
...
-    Nullable<uint8_t> mSpeedSetting{};
+    uint8_t mPercentCurrent;
+    uint8_t mSpeedCurrent;
+
+    // Fan Mode Limits
+    struct Range
+    {
+        const bool Contains(int value) const { return value >= low && value <= high; }
+        const int Low() const { return low; }
+        const int High() const { return High(); }
```
andy31415, line 61:
```suggestion
        const int High() const { return high; }
```

examples/chef/common/chef-fan-control-manager.cpp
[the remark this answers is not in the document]
bzbarsky-apple: Yes, separate PR is a good idea.

examples/chef/common/chef-fan-control-manager.cpp
```diff
@@ -38,20 +38,45 @@ using namespace chip::app::Clusters::FanControl::Attributes;
...
+    void HandleFanControlAttributeChange(AttributeId attributeId, uint8_t type, uint16_t size, uint8_t * value);
     Status HandleStep(StepDirectionEnum aDirection, bool aWrap, bool aLowestOff) override;
+    DataModel::Nullable<uint8_t> GetSpeedSetting();
+    DataModel::Nullable<Percent> GetPercentSetting();
 
 private:
-    Nullable<uint8_t> mPercentSetting{};
-    Nullable<uint8_t> mSpeedSetting{};
+    uint8_t mPercentCurrent;
+    uint8_t mSpeedCurrent;
```
andy31415, line 54:
```suggestion
    uint8_t mPercentCurrent = 0;
    uint8_t mSpeedCurrent = 0;
```

examples/chef/common/chef-fan-control-manager.cpp
```diff
@@ -99,98 +124,213 @@ Status ChefFanControlManager::HandleStep(StepDirectionEnum aDirection, bool aWra
...
+        ChipLogProgress(NotSpecified, "ChefFanControlManager::HandleFanControlAttributeChange  PercentSetting");
+        DataModel::Nullable<Percent> percentSetting =
+            NumericAttributeTraits<DataModel::Nullable<Percent>>::StorageToWorking(*value);
+
+        // The cluster code in fan-control-server.cpp is the only one allowed to set PercentSetting to null.
+        // This happens as a consequence of setting the FanMode to kAuto. In auto mode, percentCurrent should continue to report the
+        // real fan speed percentage. In this example, we set PercentCurrent to 0 here as we don't have a real value for the Fan
+        // speed or a FanAutoMode simulator.
+        // When not Null, SpeedCurrent tracks SpeedSetting's value.
+        SetPercentCurrent(percentSetting.IsNull() ? 0 : percentSetting.Value());
```
andy31415, line 142:
```suggestion
        SetPercentCurrent(percentSetting.ValueOr(0));
```

examples/chef/common/chef-fan-control-manager.cpp
```diff
@@ -99,98 +124,213 @@ Status ChefFanControlManager::HandleStep(StepDirectionEnum aDirection, bool aWra
...
+    case FanControl::Attributes::SpeedSetting::Id: {
+        ChipLogProgress(NotSpecified, "ChefFanControlManager::HandleFanControlAttributeChange  SpeedSetting");
+
+        DataModel::Nullable<uint8_t> speedSetting = NumericAttributeTraits<DataModel::Nullable<uint8_t>>::StorageToWorking(*value);
+        // The cluster code in fan-control-server.cpp is the only one allowed to set speedSetting to null.
+        // This happens as a consequence of setting the FanMode to kAuto. In auto mode, speedCurrent should continue to report the
+        // real fan speed. In this example, we set SpeedCurrent to 0 here as we don't have a real value for the Fan speed or a
+        // FanAutoMode simulator.
+        // When not Null, SpeedCurrent tracks SpeedSetting's value.
+        SetSpeedCurrent(speedSetting.IsNull() ? 0 : speedSetting.Value());
```
andy31415, line 155:
```suggestion
        SetSpeedCurrent(speedSetting.ValueOr(0));
```

examples/chef/common/chef-fan-control-manager.cpp
```diff
@@ -99,98 +124,213 @@ Status ChefFanControlManager::HandleStep(StepDirectionEnum aDirection, bool aWra
...
-    VerifyOrDie(aPath.mClusterId == FanControl::Id);
-    VerifyOrDie(aPath.mEndpointId == mEndpoint);
-
-    switch (aPath.mAttributeId)
+    if (aNewSpeedSetting.IsNull())
     {
-    case PercentCurrent::Id: {
-        // Current percents always tracks setting immediately in our implementation.
-        return aEncoder.Encode(mPercentSetting.ValueOr(0));
+        ChipLogError(NotSpecified, "ChefFanControlManager::SetSpeedSetting: invalid value");
```
andy31415, line 284:
```suggestion
        ChipLogError(NotSpecified, "ChefFanControlManager::SetSpeedSetting: null value is invalid");
```

I think we know explicitly what value is invalid, so the message could be clearer

examples/chef/common/chef-fan-control-manager.cpp
```diff
@@ -99,98 +124,213 @@ Status ChefFanControlManager::HandleStep(StepDirectionEnum aDirection, bool aWra
...
-        return aEncoder.Encode(mSpeedSetting);
+}
+
+void ChefFanControlManager::Init()
+{
+    DataModel::Nullable<Percent> percentSetting = GetPercentSetting();
+    SetPercentCurrent(percentSetting.IsNull() ? 0 : percentSetting.Value());
+
+    DataModel::Nullable<uint8_t> speedSetting = GetSpeedSetting();
+    SetSpeedCurrent(speedSetting.IsNull() ? 0 : speedSetting.Value());
```
andy31415, line 305:
```suggestion
    SetPercentCurrent(GetPercentSetting().ValueOr(0));
    SetSpeedCurrent(GetSpeedSetting().ValueOr(0));
```

andy31415 (approved):

examples/chef/common/chef-fan-control-manager.cpp
soares-sergio: Thanks, Andrei. Learned lots of new tricks today!

examples/chef/common/chef-fan-control-manager.cpp
```diff
@@ -99,98 +124,210 @@ Status ChefFanControlManager::HandleStep(StepDirectionEnum aDirection, bool aWra
...
 {
-    VerifyOrDie(aPath.mClusterId == FanControl::Id);
-    VerifyOrDie(aPath.mEndpointId == mEndpoint);
+    ChipLogProgress(NotSpecified, "ChefFanControlManager::HandleFanControlAttributeChange");
+    switch (attributeId)
+    {
+    case FanControl::Attributes::PercentSetting::Id: {
+        ChipLogProgress(NotSpecified, "ChefFanControlManager::HandleFanControlAttributeChange  PercentSetting");
+        DataModel::Nullable<Percent> percentSetting =
+            NumericAttributeTraits<DataModel::Nullable<Percent>>::StorageToWorking(*value);
```
bzbarsky-apple, line 135: The fact that this compiles at all is sort of broken..... It does not do the right thing, for sure:
```
NumericAttributeTraits<DataModel::Nullable<Percent>>::StorageToWorking(255).IsNull()
```
returns false, right?

I think you want something more like this:
```suggestion
        DataModel::Nullable<Percent> percentSetting;
        if (!NumericAttributeTraits<Percent>::IsNullValue(*value)) {
            percentSetting.SetNonNull(NumericAttributeTraits<Percent>::StorageToWorking(*value));
        }
```

examples/chef/common/chef-fan-control-manager.cpp
```diff
@@ -99,98 +124,210 @@ Status ChefFanControlManager::HandleStep(StepDirectionEnum aDirection, bool aWra
...
+        // speed or a FanAutoMode simulator.
+        // When not Null, SpeedCurrent tracks SpeedSetting's value.
+        SetPercentCurrent(percentSetting.ValueOr(0));
+        break;
+    }
+
+    case FanControl::Attributes::SpeedSetting::Id: {
+        ChipLogProgress(NotSpecified, "ChefFanControlManager::HandleFanControlAttributeChange  SpeedSetting");
+
+        DataModel::Nullable<uint8_t> speedSetting = NumericAttributeTraits<DataModel::Nullable<uint8_t>>::StorageToWorking(*value);
```
bzbarsky-apple, line 149: Again, this very much does not do the right thing.

examples/chef/common/chef-fan-control-manager.cpp
```diff
@@ -99,98 +124,210 @@ Status ChefFanControlManager::HandleStep(StepDirectionEnum aDirection, bool aWra
...
+        SetSpeedCurrent(speedSetting.ValueOr(0));
+        // Determine if the speed change should also change the fan mode
+        FanControl::Attributes::FanMode::Set(mEndpoint, SpeedToFanMode(mSpeedCurrent));
+        break;
+    }
+
+    case FanControl::Attributes::FanMode::Id: {
+        ChipLogProgress(NotSpecified, "ChefFanControlManager::HandleFanControlAttributeChange  FanMode");
+
+        FanControl::FanModeEnum fanMode = static_cast<FanControl::FanModeEnum>(*value);
```
bzbarsky-apple, line 164: Should probably static_assert that sizeof(FanModeEnum) is 1 here.

examples/chef/common/chef-fan-control-manager.cpp
```diff
@@ -99,98 +124,210 @@ Status ChefFanControlManager::HandleStep(StepDirectionEnum aDirection, bool aWra
...
-    case SpeedSetting::Id: {
-        Nullable<uint8_t> newSpeedSetting;
-        ReturnErrorOnFailure(aDecoder.Decode(newSpeedSetting));
+        return FanControl::FanModeEnum::kMedium;
+    }
+    if (kFanModeHighSpeedRange.Contains(speed))
+    {
+        return FanControl::FanModeEnum::kHigh;
+    }
+    return FanControl::FanModeEnum::kUnknownEnumValue;
```
bzbarsky-apple, line 193: Might be OK depending on how this is used, but I would be _really_ careful about ever returning kUnknownEnumValue.  It must never be encoded on the wire, for example, or written into the data store, and this is very much returning a value that will be written into the data store.

I'd suggest a followup restructuring this to make sure that we never write kUnknownEnumValue into the data store.

bzbarsky-apple (approved):

examples/chef/common/chef-fan-control-manager.cpp
bzbarsky-apple: https://github.com/project-chip/connectedhomeip/pull/36459 to make this not compile.

soares-sergio: Good catch, Thanks! Yes, I assumed that would convert properly. Thanks for the follow up PR to make this mistake no longer possible.

examples/chef/common/chef-fan-control-manager.cpp
soares-sergio: Good idea!

examples/chef/common/chef-fan-control-manager.cpp
soares-sergio: Fixed like the other one, thank you so much!

PR #36267, opened 2024-10-28
[darwin-framework-tool][interactive] The multiline prompt makes it di…

vivien-apple:
…fficult to modify command with backspace

#### Problem

The interactive mode in `darwin-framework-tool` is cumbersome to use, especially when editing commands, due to a multiline prompt. Since this prompt only provides instructions, it doesn’t need to be part of the command input line.

This PR updates the instructions to print separately from the prompt, making them clearer and easier to follow. The instructions have also been styled for readability.

Extra bonus: Added `quit` as an alias for `quit()`.

### DIFF
--- a/examples/darwin-framework-tool/commands/interactive/InteractiveCommands.mm
+++ b/examples/darwin-framework-tool/commands/interactive/InteractiveCommands.mm
@@ -27,12 +27,17 @@
 #include <editline.h>
 #include <stdlib.h>
 
-constexpr char kInteractiveModePrompt[] = "Stop and restart stack: [Ctrl+_] & [Ctrl+^]\n"
-                                          "Trigger exit(0): [Ctrl+@]\n"
-                                          "Quit Interactive: 'quit()'\n"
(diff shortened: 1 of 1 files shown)


examples/darwin-framework-tool/commands/interactive/InteractiveCommands.mm
```diff
@@ -391,7 +401,7 @@ el_status_t ExitFunction()
 
 bool InteractiveCommand::ParseCommand(char * command, int * status)
 {
-    if (strcmp(command, kInteractiveModeStopCommand) == 0) {
+    if (strcmp(command, kInteractiveModeStopCommand) == 0 || strcmp(command, kInteractiveModeStopAlternateCommand) == 0) {
```
bzbarsky-apple, line 404: Might be worth adding that to chip-tool too, in a separate PR.

bzbarsky-apple (approved):

bzbarsky-apple:

PR #36275, opened 2024-10-28
Update Coding style documentation

cecille:
- removed most of the motivation sections as it was somewhat
      repetitive and made the doc quite verbose
    - removed the drawing because it doesn't match the SDK implementation
      in reality
    - removed C discussion - I don't think we have any
    - added python versions. Didn't add the other languages because I
      dont' know but other folks can easily add a follow up
    - added sections on
      + formatters
      + anonymous namespaces for internal stuff in cpp files
      + singletons
      + std containers to heap allocation sections
      + CopySpanToMutableSpan
      + std::optional
      + python section
    - removed the code samples from heap allocation because it implies
      that people show re-implement these things when the preference is
      to use the provided support classes
    - removed the version table - we have git history.

Also changed from adoc to md, since that's why it was not showing up on the documentation site.


docs/style/CODING_STYLE_GUIDE.md
```diff
@@ -0,0 +1,162 @@
...
+original code.
+
+If you find the conventions so foreign or otherwise confusing, it may be best to
+let whoever owns the file make the necessary changes or seek the counsel of
+others in the group to find out what the right thing to do is. Never just start
+changing code wholesale for personal reasons without consulting others first.
+
+#### Commenting Out or Disabling Code
+
+Unused code shall not be disabled by commenting it out with C- or C++-style
```
arkq, line 57: Maybe add an information what to do instead of commenting the code out, like: "simply remove it". Otherwise, one might get the impression that the unused code should not be disabled/removed at all.

docs/style/CODING_STYLE_GUIDE.md
```diff
@@ -0,0 +1,162 @@
...
+#### Auto-formatters
+
+We use the following auto-formatters on code:
+
+| Language   | Formatter          | Style File                                                                                                  |
+| ---------- | ------------------ | ----------------------------------------------------------------------------------------------------------- |
+| C++        | clang-format       | [.clang-format](https://github.com/project-chip/connectedhomeip/blob/master/.clang-format)                  |
+| ObjectiveC | clang-format       | [.clang-format](https://github.com/project-chip/connectedhomeip/blob/master/.clang-format)                  |
+| java       | google-java-format | N/A                                                                                                         |
+| Python     | pep8, isort        | [.restyled.yaml](https://github.com/project-chip/connectedhomeip/blob/master/.restyled.yaml) (command line) |
```
arkq, line 69: .restyled.yaml is not a style file for python, it's a configuration for restyled. However I can see that it has command line options for autopep8.... which is not perfect because autopep8 run locally will not see these options... How about moving them to one of these: https://github.com/hhatto/autopep8?tab=readme-ov-file#id8 Also, we have `ruff` rules: https://github.com/project-chip/connectedhomeip/blob/master/ruff.toml

config for isort: https://github.com/project-chip/connectedhomeip/blob/master/.isort.cfg

docs/style/CODING_STYLE_GUIDE.md
```diff
@@ -0,0 +1,162 @@
...
+
+If the class truly should be a singleton (ex. if it is controlling access to a
+hardware resource)
+
+-   The standard function name for accessing an SDK singleton is GetInstance().
+-   Singleton classes should delete copy and move constructors
+
+#### Avoid Heap-based Resource Allocation and auto-resizing std containers
+
+Heap-based resource allocation should be avoided. This includes any container
```
arkq, line 118: I think that there is an exception from that rule for platform code which runs on platforms with sufficient amount of RAM. It would be nice to emphasize that this mostly applies to common SDK codebase and RAM-constrained platforms.

arkq (approved):

docs/style/CODING_STYLE_GUIDE.md
cecille: Good call. Updated this section to indicate this applies only to core SDK code, and controller and platforms may allow heap allocation.

docs/style/CODING_STYLE_GUIDE.md
cecille: Added config for isort. Going to defer the move of the autopep stuff to avoid mixing documentation and CI changes in the same PR.

So we RUN ruff in the CI? I'm not sure I've ever failed it and I'd DEFINITELY done some weird stuff that should have failed it.

arkq: > So we RUN ruff in the CI?

https://github.com/project-chip/connectedhomeip/blob/master/.github/workflows/lint.yml#L307

> I've ever failed it and I'd DEFINITELY done some weird stuff that should have failed it.

It has lots of warning disabled. Most of these warning are style related, so the CI will not fail on "style" issues, but only on potential code bugs. The style, IMHO, should be handled by restyled, which provides a patch for a fixup. I was never configured though, because as far as I know there are some issues with line length automatic reflow. But maybe it has changed now...

andy31415 (approved):

docs/style/CODING_STYLE_GUIDE.md
cecille: oh, the linter. Cool. OK, I'll add that - thanks

docs/style/CODING_STYLE_GUIDE.md
```diff
@@ -0,0 +1,167 @@
...
+| ObjectiveC | clang-format       | [.clang-format](https://github.com/project-chip/connectedhomeip/blob/master/.clang-format)                                                                                                                                                                                |
+| java       | google-java-format | N/A                                                                                                                                                                                                                                                                       |
+| Python     | pep8, isort, ruff  | [.restyled.yaml](https://github.com/project-chip/connectedhomeip/blob/master/.restyled.yaml) (command line),[isort](https://github.com/project-chip/connectedhomeip/blob/master/.isort.cfg) [ruff](https://github.com/project-chip/connectedhomeip/blob/master/ruff.toml) |
```
arkq, line 70: nit: Maybe for readability's sake you can use link "links", i.e.:

```md
Some nice web page [here][link]
 
[link]: https://github.com/project-chip/connectedhomeip/pull/36275/files
```

Also there is no space after the first comma and there is no comma on the second space :D

cecille: sounds good - updated.

bzbarsky-apple (approved):

docs/style/CODING_STYLE_GUIDE.md
```diff
@@ -0,0 +1,174 @@
...
+from many different companies and the SDK code therefore uses several different
+coding styles throughout the code base. Stylistically, code should attempt to
+conform to the dominant style of the code being modified, while also adhering to
+the guidelines below.
+
+## Language standard
+
+Code in the SDK conforms to the following standards. Changes to the C++ standard
+happen infrequently and there has been only one change from C++14 to C++17 over
+the course of the SDK lifetime. Changes to the Python version happen more
```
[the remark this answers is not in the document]
cecille: I'm not sure either TBH. I will update this to make it more vague. The exact history doesn't really matter too much.

docs/style/CODING_STYLE_GUIDE.md
```diff
@@ -0,0 +1,174 @@
...
+
+#### Use C++ _cstdint_ for Plain Old Data Types
+
+Standard, scalar data types defined in _cstdint_ should be used for basic signed
+and unsigned integer types, especially when size and serialization to
+non-volatile storage or across a network is concerned.
+
+Examples of these are: `uint8_t`, `int8_t`, etc.
+
+#### Avoid `using namespace` Statements in Headers
```
cecille, line 95:
```suggestion
#### Avoid top-level `using namespace` Statements in Headers
```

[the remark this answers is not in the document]
cecille: I'm just going to update this to say "top-level `using namespace` statements.

docs/style/CODING_STYLE_GUIDE.md
```diff
@@ -0,0 +1,174 @@
...
+
+By doing this, you are effectively forcing every other module that includes the
+header to also be using the namespace. This causes namespace pollution and
+generally defeats the purposes of namespaces. Fully-qualified symbols should be
+used instead.
```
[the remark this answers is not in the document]
cecille: good point - will update. Thanks.

docs/style/CODING_STYLE_GUIDE.md
```diff
@@ -0,0 +1,172 @@
...
+-   Pool-based allocators
+-   Platform-defined and -assigned allocators
+
+[CHIPMem.h](https://github.com/project-chip/connectedhomeip/blob/master/src/lib/support/CHIPMem.h)
+provides support for platform defined allocators.
+
+[Pool.h](https://github.com/project-chip/connectedhomeip/blob/master/src/lib/support/Pool.h)
+is the Matter SDK pool allocator implementation.
+
+#### Prefer CopySpanToMutableSpan over memcpy when using spans
```
Alami-Amine, line 147: Should we also add a section for: Use Span/ByteSpan instead of C-style pointers for buffers ?

docs/style/CODING_STYLE_GUIDE.md
```diff
@@ -0,0 +1,172 @@
...
+    https://github.com/project-chip/connectedhomeip/blob/master/.restyled.yaml
+[isort_link]:
+    https://github.com/project-chip/connectedhomeip/blob/master/.isort.cfg
+[ruff_link]:
+    https://github.com/project-chip/connectedhomeip/blob/master/ruff.toml
+
+All pull requests run formatting checks using these tools before merge is
+allowed. Generated code is not run through restyle.
+
+### C++
```
Alami-Amine, line 83: should we add a couple of sections on casting? something like:
- Use C++-style casts (e.g. static_cast) for type conversions instead of C-style casts.
- Whenever possible, use SafePointerCast.h instead of reinterpret_cast for pointer casting, since it is safer and adds alignment and type safety checks. 
- Avoid casting function pointers to `void *`, as this is considered undefined behavior in C++.

jmartinez-silabs (approved):

docs/style/CODING_STYLE_GUIDE.md
```diff
@@ -0,0 +1,172 @@
+# Coding Style Guide
```
jmartinez-silabs, line 1: Shall we present some of those conventions

- Variable, class, file naming?
- PascalCasing
- constexpr preferred to defines
- class enum
- Maybe some guidance for .gn file structure

Alami-Amine (approved):

docs/style/CODING_STYLE_GUIDE.md
cecille: Tracking suggestions for new content in a separate issue: https://github.com/project-chip/connectedhomeip/issues/36544

docs/style/CODING_STYLE_GUIDE.md
cecille: Tracking new content suggestions in an issue for later: https://github.com/project-chip/connectedhomeip/issues/36544

docs/style/CODING_STYLE_GUIDE.md
cecille: Tracking new content suggestions in an issue: https://github.com/project-chip/connectedhomeip/issues/36544

PR #36286, opened 2024-10-29
Make TLVReader/Writer consume slightly less flash

andy31415:
Some of the methods seemed large in reports. I could not find very large savings, however this still saves a bit over 100 bytes of flash for a sample NRF light.

Readability may also be ever so slightly improved (at least less lines of code)

### DIFF
--- a/src/lib/core/TLVReader.cpp
+++ b/src/lib/core/TLVReader.cpp
@@ -755,27 +755,16 @@ CHIP_ERROR TLVReader::SkipToEndOfContainer()
 
 CHIP_ERROR TLVReader::ReadElement()
 {
-    CHIP_ERROR err;
-    uint8_t stagingBuf[17]; // 17 = 1 control byte + 8 tag bytes + 8 length/value bytes
-    const uint8_t * p;
-    TLVElementType elemType;
-
     // Make sure we have input data. Return CHIP_END_OF_TLV if no more data is available.
-    err = EnsureData(CHIP_END_OF_TLV);
-    if (err != CHIP_NO_ERROR)
-        return err;
+    ReturnErrorOnFailure(EnsureData(CHIP_END_OF_TLV));
+    VerifyOrReturnError(mReadPoint != nullptr, CHIP_ERROR_INVALID_TLV_ELEMENT);
 
-    if (mReadPoint == nullptr)
-    {
-        return CHIP_ERROR_INVALID_TLV_ELEMENT;
-    }
     // Get the element's control byte.
     mControlByte = *mReadPoint;
 
     // Extract the element type from the control byte. Fail if it's invalid.
-    elemType = ElementType();
-    if (!IsValidTLVType(elemType))
-        return CHIP_ERROR_INVALID_TLV_ELEMENT;
+    TLVElementType elemType = ElementType();
+    VerifyOrReturnError(IsValidTLVType(elemType), CHIP_ERROR_INVALID_TLV_ELEMENT);
 
     // Extract the tag control from the control byte.
     TLVTagControl tagControl = static_cast<TLVTagControl>(mControlByte & kTLVTagControlMask);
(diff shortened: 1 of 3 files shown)


jmartinez-silabs (approved):

src/lib/core/TLVReader.cpp
```diff
@@ -787,54 +761,34 @@ CHIP_ERROR TLVReader::ReadElement()
...
-        p = mReadPoint;
-        mReadPoint += elemHeadBytes;
-        mLenRead += elemHeadBytes;
-    }
+    // 17 = 1 control byte + 8 tag bytes + 8 length/value bytes
+    uint8_t stagingBuf[17];
+
+    // Odd workaround: clang-tidy claims garbage value otherwise as it does not
+    // understand that ReadData initializes stagingBuf
+    stagingBuf[1] = 0;
```
andy31415, line 775: This is the only questionable bit. clang-tidy complains on 

```
`../../third_party/nlio/repo/include/nlio-base.h:84:5: error: Undefined or garbage value returned to caller [clang-analyzer-core.uninitialized.UndefReturn,-warnings-as-errors]   
   84 |     return *(const uint8_t *)(p); 
```

and I could not find a reasonable way to suppress it. setting staging[1] to 0 is smaller code than `= {0}` or `memset(stagingBuf, 0, sizeof(stagingBug))`;

https://github.com/project-chip/connectedhomeip/actions/runs/11577116971/job/32227734524 for the full error

src/lib/core/TLVReader.cpp
```diff
@@ -787,54 +761,34 @@ CHIP_ERROR TLVReader::ReadElement()
...
+
+    // Read to handle also if the head of the element overlaps the end of the input buffer.
+    // This reads the bytes into the staging buffer and arrange to parse them from there.
```
bzbarsky-apple, line 778:
```suggestion
    // If the head of the element goes past the end of the current input buffer, 
    // we need to read it into the staging buffer to parse it.  Just do that unconditionally,
    // even if the head does not go past end of current buffer, to save codesize.
```

src/lib/core/TLVReader.cpp
```diff
@@ -787,54 +761,34 @@ CHIP_ERROR TLVReader::ReadElement()
...
+    ReturnErrorOnFailure(ReadData(stagingBuf, elemHeadBytes));
 
-    // Skip over the control byte.
-    p++;
+    const uint8_t * p = stagingBuf + 1;
```
bzbarsky-apple, line 781: Why did this lose the "skip over the control byte" comment?  That seems still needed...

src/lib/core/TLVReader.cpp
```diff
@@ -787,54 +761,34 @@ CHIP_ERROR TLVReader::ReadElement()
...
-    case kTLVFieldSize_4Byte:
-        mElemLenOrVal = LittleEndian::Read32(p);
-        break;
-    case kTLVFieldSize_8Byte:
-        mElemLenOrVal = LittleEndian::Read64(p);
-        VerifyOrReturnError(!TLVTypeHasLength(elemType) || (mElemLenOrVal <= UINT32_MAX), CHIP_ERROR_NOT_IMPLEMENTED);
-        break;
-    }
+    memcpy(&mElemLenOrVal, p, valOrLenBytes);
+    LittleEndian::HostSwap(mElemLenOrVal);
```
bzbarsky-apple, line 789: This is worth a comment that it works because for a little-endian integer memcopying it into the start of a larger-sized little-endian integer preserves the value.

src/lib/core/TLVTypes.h
```diff
@@ -186,26 +186,32 @@ inline bool TLVTypeIsUTF8String(TLVElementType type)
...
-inline uint8_t TLVFieldSizeToBytes(TLVFieldSize fieldSize)
+constexpr uint8_t TLVFieldSizeToBytes(TLVFieldSize fieldSize)
 {
     // We would like to assert fieldSize < 7, but that gives us fatal
     // -Wtautological-constant-out-of-range-compare warnings...
     return static_cast<uint8_t>((fieldSize != kTLVFieldSize_0Byte) ? (1 << fieldSize) : 0);
 }
 
+static_assert(TLVFieldSizeToBytes(kTLVFieldSize_0Byte) == 0);
+static_assert(TLVFieldSizeToBytes(kTLVFieldSize_1Byte) == 1);
```
bzbarsky-apple, line 211: Thank you for adding these.

bzbarsky-apple (approved):

src/lib/core/TLVReader.cpp
andy31415: Added it back.

src/lib/core/TLVReader.cpp
andy31415: Added comment.

PR #36287, opened 2024-10-29
[darwin-framework-tool] Add a command to initialize a MTRDevice such …

vivien-apple:
…the the initial subscription is setted up without having to issue an additional command (e.g read or write)

#### Problem

This PR adds a command to `darwin-framework-tool` that enables the initial subscription from `MTRDevice` to be set up without requiring an additional command (like `read-by-id` or `write-by-id` for now).

Additionally, it introduces extra configuration options for the device delegate and supports having a separate device delegate per device instance (rather than a single shared delegate).

### DIFF
--- a/examples/darwin-framework-tool/commands/configuration/SetUpDeviceCommand.h
+++ b/examples/darwin-framework-tool/commands/configuration/SetUpDeviceCommand.h
@@ -1,0 +1,82 @@
+/*
+ *   Copyright (c) 2024 Project CHIP Authors
+ *   All rights reserved.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
+ *   You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *   Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


woody-apple (approved):

examples/darwin-framework-tool/commands/configuration/SetUpDeviceCommand.h
```diff
@@ -0,0 +1,82 @@
...
+#include <commands/common/Command.h>
+
+class SetUpDeviceCommand : public CHIPCommandBridge {
+public:
+    SetUpDeviceCommand()
+        : CHIPCommandBridge("device", "Creates and configures an instance of a device.")
+    {
+        AddArgument("node-id", 0, UINT64_MAX, &mNodeId, "The Node ID of the device instance to create.");
+        AddArgument("pretend-thread-enabled", 0, 1, &mPretendThreadEnabled,
+            "When the device is configured using an MTRDevice (via -use-mtr-device), instructs the MTRDevice to treat the "
```
bzbarsky-apple, line 32: I don't get this comment. This command _always_ sets up an MTRDevice...  What device is "configured" with "-use-mtr-device" for purposes of this command?  (And shouldn't it be --use-mtr-device anyway?)

PR #36288, opened 2024-10-29
Scenes-cluster: move non-template things into non-template functions to save flash

andy31415:
Scenes seems to use templates quite a bit so there is a code size multipler for template code. Switched code to use non-template functions for some parts that do not need to be templates.

This saves some flash


tehampson (approved):

lpbeliveau-silabs (approved):

lpbeliveau-silabs:
FYI, Opened issue for the AddResponseOnError here: https://github.com/project-chip/connectedhomeip/issues/36309

PR #36290, opened 2024-10-29
[Telink] Add support for groups multicast control

interfer:
Add support for up to 8 groups for lighting-app.
Tested on Telink B9x, W91 platform. 
Increased Thread multicast configs for ble-thread plarforms

### DIFF
--- a/examples/lighting-app/telink/include/CHIPProjectConfig.h
+++ b/examples/lighting-app/telink/include/CHIPProjectConfig.h
@@ -35,3 +35,9 @@
 //  Until this is improved in OpenThread we need to increase the retransmission
 //  interval to survive the stall.
 #define CHIP_CONFIG_MRP_LOCAL_ACTIVE_RETRY_INTERVAL (1000_ms32)
+
+// ========== Multicast groups Configuration Overrides =========
+#undef CHIP_CONFIG_MAX_GROUP_ENDPOINTS_PER_FABRIC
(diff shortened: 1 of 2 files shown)


examples/lighting-app/telink/include/CHIPProjectConfig.h
damien0x0023, line 1: Hey, @interfer.

We need to increase the maximum group number per fabric to 8 and the definition `CHIP_CONFIG_MAX_GROUPS_PER_FABRIC` is for this purpose. 
See src/lib/core/CHIPConfig.h (https://github.com/s07641069/connectedhomeip/blob/72e0a68127e9513a97855d3d199d58167b0e7ec1/src/lib/core/CHIPConfig.h#L1072) for details.

However, `CHIP_CONFIG_MAX_GROUP_ENDPOINTS_PER_FABRIC` is the maximum group endpoint number per fabric.
If we set it to 8, the maximum group number per fabric will be 32.
Please check  L1058~1082 to see the differences.

damien0x0023 (changes requested):

s07641069 (approved):

examples/lighting-app/telink/include/CHIPProjectConfig.h
interfer: @damien0x0023 
Thanks for the hint, agree with your point. Changes applied, tested again and force-pushed

andy31415 (approved):

damien0x0023 (approved):

Issue #36289, opened 2024-10-29
Missing Clusters on Endpoint ID 1 in Linux TV Casting commissioner-generated-passcode flow


shaoltan-amazon:
**Description:**
For the commissioner-generated passcode commissioning flow of the Linux TV casting experience, we run demo interactions with Endpoint ID 1. However, when running in an Ubuntu environment, no Clusters are available on Endpoint ID 1. This issue prevents us from subscribing to the Media Playback cluster and sending/receiving the LaunchURL between the tv-app and the tv-casting-app.

Below is a snippet from the Linux tv-casting-app logs where it shows that there are 0 Clusters for Endpoint ID 1:

```
[1730222035.829705][12461:12461] CHIP:SVR: simple-app-helper.cpp::LogEndpointsDetails() Number of Endpoints: 3
[1730222035.829723][12461:12461] CHIP:SVR: Endpoint::LogDetail() Endpoint ID: 1, Vendor ID: 0, Product ID: 0, Clusters: 0
[1730222035.829739][12461:12461] CHIP:SVR: Endpoint::LogDetail() Endpoint ID: 2, Vendor ID: 0, Product ID: 0, Clusters: 2
[1730222035.829755][12461:12461] CHIP:SVR: Endpoint::LogDetail() Endpoint ID: 4, Vendor ID: 65521, Product ID: 32769, Clusters: 6
[1730222035.829772][12461:12461] CHIP:SVR: simple-app-helper.cpp::ConnectionHandler(): Triggering demo interactions with CastingPlayer (ID: D83ADD0E2E015650). Endpoint ID: 1
```

Note: We do not receive any more messages between the tv-app and the tv-casting-app after this point, likely due to subscribing to an Endpoint with no Clusters.

On the other hand, when running the same Linux commissioning flow on MacOS, we do see 7 Clusters on Endpoint ID 1, which allows us to subscribe to the Media Playback cluster and successfully send and receive the LaunchURL between the tv-app and the tv-casting-app. This is the expected behavior when we run in an Ubuntu environment as well.

Below is a snippet of the expected output in the Linux tv-casting-app logs:

```
[1730156548.558] [9408:32653093] [SVR] simple-app-helper.cpp::LogEndpointsDetails() Number of Endpoints: 3
[1730156548.558] [9408:32653093] [SVR] Endpoint::LogDetail() Endpoint ID: 1, Vendor ID: 0, Product ID: 0, Clusters: 7
[...]

PR #36291, opened 2024-10-29
Fix and re-enable Linux TV commissioner-generated-passcode CI test

### LINKED ISSUE #34598: run_tv_casting_test.py is flaky for commissioner generated passcode
In CI, it seems like `python3 ./scripts/tests/run_tv_casting_test.py --commissioner-generated-passcode=True` is flaky.

Example:

https://github.com/project-chip/connectedhomeip/actions/runs/10148134972/job/28060156339

Error seems timeouts on

```
TestStepException: commissioner_generated_passcode_test - Did not find the expected output string(s) in the tv-app subprocess within the timeout: ['Commissioner Declaration Start', 'commissioner passcode: true', 'Commissioner Declaration End']
```

from the application logs, sometimes all the strings are missing and sometimes only a subset of strings are missing.

shaoltan-amazon:
`Fixes #34598`

The Linux TV commissioner-generated-passcode test flow was flaky due to buffering issues when attempting to read from the output (and error) stream(s). 

**Changes made:**
1. Added line buffering improvements to the code and also "flushed" after every "write".
2. Modified `_io_thread` to read all the output available in the output stream and store them directly to the output queue.

Note:
Commented out the last few steps in the commissioner-generated-passcode test flow as there seems to be a new issue that came up recently where we are subscribing to an endpoint with 0 clusters (which should not be the case). See https://github.com/project-chip/connectedhomeip/issues/36289 for more details. These test sequence steps should be enabled once #36289 is fixed.

**Testing**
- Ran the test script 20 times in Ubuntu environment with 100% passing rate for both the commissioner-generated-passcode test flow and the commissionee-generated-passcode test flow.
- Ran the CI test 20 times in local branch with 100% passing rate: https://github.com/shaoltan-amazon/connectedhomeip/actions/runs/11583471460

### DIFF
--- a/scripts/tests/linux/log_line_processing.py
+++ b/scripts/tests/linux/log_line_processing.py
@@ -96,12 +96,12 @@ class ProcessOutputCapture:
             text=True,
             bufsize=1,  # Enable line buffering for immediate output from subprocess
         )
+        self.output_file = open(self.output_path, "wt", buffering=1)  # Enable line buffering for immediate output
+        self._write_to_file(f"### PROCESS START: {time.ctime()} ###\n")
         self.stdout_thread = threading.Thread(target=self._stdout_thread)
         self.stderr_thread = threading.Thread(target=self._stderr_thread)
         self.stdout_thread.start()
         self.stderr_thread.start()
-        self.output_file = open(self.output_path, "wt", buffering=1)  # Enable line buffering
-        self._write_to_file(f"### PROCESS START: {time.ctime()} ###\n")
         return self
 
     def __exit__(self, exception_type, exception_value, traceback):
--- a/scripts/tests/linux/log_line_processing.py
+++ b/scripts/tests/linux/log_line_processing.py
@@ -63,26 +63,30 @@ class ProcessOutputCapture:
         err_wait = select.poll()
         err_wait.register(self.process.stderr, select.POLLIN | select.POLLHUP)
 
-        with open(self.output_path, "wt") as f:
+        with open(self.output_path, "wt", buffering=1) as f:
             f.write("PROCESS START: %s\n" % time.ctime())
+            f.flush()
             while not self.done:
-                changes = out_wait.poll(0.1)
-                if changes:
-                    out_line = self.process.stdout.readline()
-                    if not out_line:
-                        # stdout closed (otherwise readline should have at least \n)
-                        continue
-                    f.write(out_line)
-                    self.output_lines.put(out_line)
+                out_changes = out_wait.poll(0.1)
+                for fd, _ in out_changes:
+                    if fd == self.process.stdout.fileno():
+                        while True:
+                            out_line = self.process.stdout.readline()
+                            if not out_line:
+                                break
+                            f.write(out_line)
+                            f.flush()
+                            self.output_lines.put(out_line)
 
-                changes = err_wait.poll(0)
-                if changes:
-                    err_line = self.process.stderr.readline()
-                    if not err_line:
-                        # stderr closed (otherwise readline should have at least \n)
-                        continue
-                    f.write(f"!!STDERR!! : {err_line}")
+                err_changes = err_wait.poll(0)
+                for fd, _ in err_changes:
+                    if fd == self.process.stderr.fileno():
+                        err_line = self.process.stderr.readline()
+                        if err_line:
+                            f.write(f"!!STDERR!! : {err_line}")
+                            f.flush()
             f.write("PROCESS END: %s\n" % time.ctime())
+            f.flush()
 
     def __enter__(self):
         self.done = False
@@ -92,6 +96,7 @@ class ProcessOutputCapture:
             stdout=subprocess.PIPE,
             stderr=subprocess.PIPE,
             text=True,
+            bufsize=1,
         )
         self.io_thread = threading.Thread(target=self._io_thread)
         self.io_thread.start()
--- a/scripts/tests/linux/log_line_processing.py
+++ b/scripts/tests/linux/log_line_processing.py
@@ -16,6 +16,7 @@ import logging
 import queue
 import subprocess
 import threading
+import time
 from typing import List
 
 
@@ -43,6 +44,7 @@ class ProcessOutputCapture:
     def __init__(self, command: List[str], output_path: str):
         # in/out/err are pipes
         self.command = command
+        self.output_file = None  # Output file handle
         self.output_path = output_path
         self.output_lines = queue.Queue()
         self.process = None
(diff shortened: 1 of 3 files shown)


scripts/tests/linux/log_line_processing.py
```diff
@@ -63,26 +63,30 @@ def _io_thread(self):
         err_wait = select.poll()
         err_wait.register(self.process.stderr, select.POLLIN | select.POLLHUP)
 
-        with open(self.output_path, "wt") as f:
+        with open(self.output_path, "wt", buffering=1) as f:
```
andy31415, line 66: Please add a comment on what buffering=1 means (it seems to be line buffered).

scripts/tests/linux/log_line_processing.py
```diff
@@ -92,6 +96,7 @@ def __enter__(self):
             stdout=subprocess.PIPE,
             stderr=subprocess.PIPE,
             text=True,
+            bufsize=1,
```
andy31415, line 99: add a comment this meaning line buffering and explaining why this is correct.

scripts/tests/linux/log_line_processing.py
```diff
@@ -63,26 +63,30 @@ def _io_thread(self):
...
-                if changes:
-                    err_line = self.process.stderr.readline()
-                    if not err_line:
-                        # stderr closed (otherwise readline should have at least \n)
-                        continue
-                    f.write(f"!!STDERR!! : {err_line}")
+                out_changes = out_wait.poll(0.1)
+                for fd, _ in out_changes:
+                    if fd == self.process.stdout.fileno():
+                        while True:
```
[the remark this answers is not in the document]
shaoltan-amazon: The `while True` loop allows us to read all available data from the output stream and store it in the output queue. This change seems to alleviate the race condition we had with the test script. If I remember correctly from debugging, previously we were adding one line to the output queue at a time when data was available. However, if the output queue was empty while trying to retrieve and process data from it, we seemed to repeatedly enter the `_io_thread`, poll for output, and then timing out without reaching `self.process.stdout.readline()`. 

When running the tv-app and the tv-casting-app manually in separate terminals, I've observed that we receive a lot of messages at once with intermittent pauses (of maybe a few milliseconds) before the next big stream of messages appear. This suggests, as you pointed out, that we likely don't block for long when doing `self.process.stdout.readline()`.

I just did a quick check and it looks like once we enter the `while True`, we do exit the loop and enter it again.

I've also re-ran the [CI test 20 times in my local branch](https://github.com/shaoltan-amazon/connectedhomeip/actions/runs/11583471460) and this change passes for all 20 attempts.

scripts/tests/linux/log_line_processing.py
shaoltan-amazon: Yes, I've added it in the latest commit.

scripts/tests/linux/log_line_processing.py
shaoltan-amazon: I've added a comment in the latest commit.

scripts/tests/linux/log_line_processing.py
```diff
@@ -63,26 +63,30 @@ def _io_thread(self):
...
-                changes = err_wait.poll(0)
-                if changes:
-                    err_line = self.process.stderr.readline()
-                    if not err_line:
-                        # stderr closed (otherwise readline should have at least \n)
-                        continue
-                    f.write(f"!!STDERR!! : {err_line}")
+                out_changes = out_wait.poll(0.1)
+                for fd, _ in out_changes:
+                    if fd == self.process.stdout.fileno():
```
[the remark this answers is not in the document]
shaoltan-amazon: Great suggestion! Updated in the latest commit.

scripts/tests/linux/log_line_processing.py
```diff
@@ -63,26 +63,28 @@ def _io_thread(self):
...
-                    out_line = self.process.stdout.readline()
-                    if not out_line:
-                        # stdout closed (otherwise readline should have at least \n)
-                        continue
-                    f.write(out_line)
-                    self.output_lines.put(out_line)
-
-                changes = err_wait.poll(0)
-                if changes:
+                out_changes = out_wait.poll(0.1)
```
andy31415, line 70: since we do the `filno in ...` thing, do we need an out and err wait that are separate anymore? it seems we could have a single poll object and maybe even poll for longer (the only reason we timeout the poll seems to stop ourselves)

scripts/tests/linux/log_line_processing.py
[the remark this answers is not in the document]
andy31415: Unresolving: I have tested your PR and we only re-enter the loop after the file closes (i.e. we seem to always read all the items, then we repeateadly read 0 a couple of times). The logs I have look like:

- `casting app` reads 1507 lines
- `casting app` reads 0 lines
- `casting app` reads 0 lines
- ...
- `casting app` reads 0 lines
- `tv app` reads 5585 lines

This effectively makes this entire read single-threaded. This would be ok if we would not split out stdout and stderr. Maybe if we really need this we should split int two I/O threads that are doing blocking readson on each of stdout and stderr.

scripts/tests/linux/log_line_processing.py
shaoltan-amazon: Discussed implementation changes offline. Marking this as resolved since we will move forward with removing the polling logic. Instead we will now create two threads (stdout and stderr) and read all lines.

scripts/tests/linux/log_line_processing.py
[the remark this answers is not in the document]
shaoltan-amazon: Discussed offline. Made the changes to now create two threads (stdout and stderr) and read all lines.

sharadb-amazon (approved):

scripts/tests/linux/log_line_processing.py
```diff
@@ -48,41 +46,48 @@ def __init__(self, command: List[str], output_path: str):
...
-    def _io_thread(self):
-        """Reads process lines and writes them to an output file.
+    def _write_to_file(self, line: str, is_error_line=False):
+        """Writes a line to an output file in a thread-safe manner.
+
+        Opens file in append-text mode ('at') with line buffering 
+        for immediate output to ensure all writes are appended.
+        """
+        with self.lock:
+            with open(self.output_path, "at", buffering=1) as f:
```
andy31415, line 61: you are re-opening this file at every line, this seems not ideal. Can't we keep the file open and just lock around f.write() ?

andy31415 (approved):

scripts/tests/linux/log_line_processing.py
shaoltan-amazon: Updated in the latest commit.

scripts/tests/linux/log_line_processing.py
```diff
@@ -92,9 +94,14 @@ def __enter__(self):
...
             text=True,
+            bufsize=1,  # Enable line buffering for immediate output from subprocess
         )
-        self.io_thread = threading.Thread(target=self._io_thread)
-        self.io_thread.start()
+        self.stdout_thread = threading.Thread(target=self._stdout_thread)
+        self.stderr_thread = threading.Thread(target=self._stderr_thread)
+        self.stdout_thread.start()
+        self.stderr_thread.start()
+        self.output_file = open(self.output_path, "wt", buffering=1)  # Enable line buffering
```
andy31415, line 103: open should be moved before the thread start, otherwise threads may _write to the file that is not yet open.

shaoltan-amazon: Should I move it to before line 99?

shaoltan-amazon: Moved it in the latest commit.

PR #36293, opened 2024-10-30
Fix Chef unable to set audio-output through RPC

erwinpan1:
The current `AudioOutputManager` use a member variable `mCurrentOutput` to store the current output. Though we can set the output through Matter protocol, however, writing to this value through RPC (which write to the real attribute storage) will not reflect the change to the `AudioOutputManager`'s `mCurrentOutput`. 

Thus, I remove the variable `mCurrentOutput` and just read/write it through Attribute storage


erwinpan1:
Unable to fix restyled issue. Close this and recreate another

PR #36297, opened 2024-10-30
Create CHIP-specific backends for pigweed assert and logs

andy31415:
Without this, we funnel into pigweed basic logging (which uses sys_io (generally stdout) and does not hook into general chip logging) and asserts are not using chip assert macros.

This would allow us to eventually use more of pigweed generic modules (e.g. bytes, base64, functions) without losing matter module integration.

#### Changes

- Standardized the assert and logging backends for pigweed to use chip-sdk specific functions
- Added `keep-going` to locals.py to support debugging failing tests
- made locals.py show colorful fail/pass on tests with timing and show timing by default
- stopped fabric admin from forcing BLE and used no-ble-no-wifi builds for CI for fabric admin and fabric bridge
- updated `host.py` to fully disable BLE on noble (since it seems otherwise we could disable the BLE layer while enabling BLE, which makes no sense)
- MCORE.FS 1.5 was not "waiting up to 10 seconts for changes" but instead checking the first change, so it seemed racy and it hit my changes that altered timing. Updated MCORE.FS to start checking only when the first value matches
- pw_rpc was asserting that Listen/Accept passes and that is not the case when we shut down INET. Made it stop the thread instead.


src/backends/assert/handler.cpp
bzbarsky-apple, line 1: Nothing about the naming of this file anywhere up the chain suggests it's pigweed-specific.  `src/backends` is _very_ generic.  If these are pigweed backends, it should be named accordingly, no?

src/platform/Linux/Logging.cpp
```diff
@@ -43,7 +43,6 @@ void LogV(const char * module, uint8_t category, const char * msg, va_list v)
     // indicate the error occurred during getting time.
     gettimeofday(&tv, nullptr);
 
-#if !CHIP_USE_PW_LOGGING
```
bzbarsky-apple, line 46: Do we still need the `pw_log/log.h` include above?

src/platform/NuttX/Logging.cpp
```diff
@@ -43,7 +43,6 @@ void LogV(const char * module, uint8_t category, const char * msg, va_list v)
     // indicate the error occurred during getting time.
     gettimeofday(&tv, nullptr);
 
-#if !CHIP_USE_PW_LOGGING
```
bzbarsky-apple, line 46: Likewise: is the pigweed log.h include still needed?

src/platform/Linux/Logging.cpp
bzbarsky-apple, line 46: Do we still need CHIP_USE_PW_LOGGING and chip_use_pw_logging?

andy31415: Removed all references.

src/backends/assert/handler.cpp
andy31415: renamed to pw_backends.

bzbarsky-apple (approved):

andy31415:
FabricSync tests fail and I believe they are the ones that use RPC. Will have to debug them, likely as a prerequisite I will fix up `local.py` to actually see them and share the python excludes into one place.

andy31415:
Local script fixes in #36359

src/python_testing/TC_MCORE_FS_1_5.py
```diff
@@ -304,7 +304,11 @@ async def test_TC_MCORE_FS_1_5(self):
                 endpoint, attribute, value = item['endpoint'], item['attribute'], item['value']
 
                 # Record arrival of an expected subscription change when seen
-                if endpoint == newly_added_endpoint and attribute == Clusters.AdministratorCommissioning.Attributes.WindowStatus:
+                if endpoint == newly_added_endpoint and attribute == cadmin_attr.WindowStatus:
+                    if value != th_server_direct_cadmin[cadmin_attr.WindowStatus]:
+                        logging.info("Window status is %r, waiting for %r", value,
+                                     th_server_direct_cadmin[cadmin_attr.WindowStatus])
+                        continue
```
tehampson, line 311: You might still need `time_remaining = report_waiting_timeout_delay_sec - elapsed` before continue

tehampson (approved):

src/python_testing/TC_MCORE_FS_1_5.py
andy31415: Moved the elapsed/remaining bits into a `finally` block.

PR #36298, opened 2024-10-30
[Matter.framework] Invalidate the CASE session if something calls Tri…

vivien-apple:
…ggerResubscriptionWithReason and it has not been established yet

#### Problem

When setting up the initial subscription and it appears to be stuck, let’s cancel the session establishment. This will cause the session to fail, and the error will be caught by the following GitHub repository: https://github.com/project-chip/connectedhomeip/blob/7df767603817b737bd2d173d30da9751492bca39/src/darwin/Framework/CHIP/MTRDevice_Concrete.mm#L2456

### DIFF
--- a/src/app/CASESessionManager.cpp
+++ b/src/app/CASESessionManager.cpp
@@ -188,6 +188,12 @@ Optional<SessionHandle> CASESessionManager::FindExistingSession(const ScopedNode
         peerId, MakeOptional(Transport::SecureSession::Type::kCASE), transportPayloadCapability);
 }
 
+void CASESessionManager::ReleaseSession(const ScopedNodeId & peerId)
+{
+    auto * session = mConfig.sessionSetupPool->FindSessionSetup(peerId, false);
+    ReleaseSession(session);
+}
+
 void CASESessionManager::ReleaseSession(OperationalSessionSetup * session)
 {
     if (session != nullptr)
--- a/src/app/CASESessionManager.h
+++ b/src/app/CASESessionManager.h
@@ -142,6 +142,7 @@ public:
 #endif // CHIP_DEVICE_CONFIG_ENABLE_AUTOMATIC_CASE_RETRIES
                                 TransportPayloadCapability transportPayloadCapability = TransportPayloadCapability::kMRPPayload);
 
+    void ReleaseSession(const ScopedNodeId & peerId);
     void ReleaseSessionsForFabric(FabricIndex fabricIndex);
 
     void ReleaseAllSessions();
--- a/src/darwin/Framework/CHIP/MTRDeviceController_Concrete.mm
+++ b/src/darwin/Framework/CHIP/MTRDeviceController_Concrete.mm
@@ -1619,6 +1619,17 @@ static inline void emitMetricForSetupPayload(MTRSetupPayload * payload)
     [self syncRunOnWorkQueue:block error:nil];
 }
 
+- (void)invalidateCASESessionEstablishmentForNode:(NSNumber *)nodeID;
+{
+    auto block = ^{
+        auto caseSessionMgr = self->_cppCommissioner->CASESessionMgr();
+        VerifyOrDie(caseSessionMgr != nullptr);
+        caseSessionMgr->ReleaseSession(self->_cppCommissioner->GetPeerScopedId(nodeID.unsignedLongLongValue));
+    };
+
+    [self syncRunOnWorkQueue:block error:nil];
+}
+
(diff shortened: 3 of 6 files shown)


woody-apple (approved):

src/app/CASESessionManager.cpp
```diff
@@ -188,6 +188,12 @@ Optional<SessionHandle> CASESessionManager::FindExistingSession(const ScopedNode
         peerId, MakeOptional(Transport::SecureSession::Type::kCASE), transportPayloadCapability);
 }
 
+void CASESessionManager::ReleaseSession(const ScopedNodeId & peerId)
+{
+    auto * session = mConfig.sessionSetupPool->FindSessionSetup(peerId, false);
+    ReleaseSession(session);
```
bzbarsky-apple, line 194: Doesn't this need to check whether session is null?

src/app/CASESessionManager.h
```diff
@@ -142,6 +142,7 @@ class CASESessionManager : public OperationalSessionReleaseDelegate, public Sess
 #endif // CHIP_DEVICE_CONFIG_ENABLE_AUTOMATIC_CASE_RETRIES
                                 TransportPayloadCapability transportPayloadCapability = TransportPayloadCapability::kMRPPayload);
 
+    void ReleaseSession(const ScopedNodeId & peerId);
```
bzbarsky-apple, line 145: Please document.

src/darwin/Framework/CHIP/MTRDeviceController_Concrete.mm
```diff
@@ -1619,6 +1619,17 @@ - (void)invalidateCASESessionForNode:(NSNumber *)nodeID;
...
 
+- (void)invalidateCASESessionEstablishmentForNode:(NSNumber *)nodeID;
+{
+    auto block = ^{
+        auto caseSessionMgr = self->_cppCommissioner->CASESessionMgr();
+        VerifyOrDie(caseSessionMgr != nullptr);
+        caseSessionMgr->ReleaseSession(self->_cppCommissioner->GetPeerScopedId(nodeID.unsignedLongLongValue));
+    };
+
+    [self syncRunOnWorkQueue:block error:nil];
```
bzbarsky-apple, line 1630: Why do we need a sync dispatch here?

src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
```diff
@@ -898,6 +898,12 @@ - (void)_triggerResubscribeWithReason:(NSString *)reason nodeLikelyReachable:(BO
             subscriptionCallback->ResetResubscriptionBackoff();
         }
         readClientToResubscribe->TriggerResubscribeIfScheduled(reason.UTF8String);
+    } else if (_internalDeviceState == MTRInternalDeviceStateSubscribing && nodeLikelyReachable) {
+        // If we have reason to suspect that the node is now reachable and we haven’t established a
+        // CASE session yet, let’s consider it to be stalled and invalidate the pairing session.
+        dispatch_async(self.queue, ^{
+            [[self _concreteController] invalidateCASESessionEstablishmentForNode:self->_nodeID];
```
bzbarsky-apple, line 905: We shouldn't add new controller API here, I would think.  Just `asyncGetCommissionerOnMatterQueue` and go from there.

That said, do we have any indication that we are not just going to get a slew of _triggerResubscribeWithReason from DNS-SD stuff while nothing has really changed?  I'm really worried about us getting into a state where we just keep canceling things and never manage to subscribe.  It does help that we only do this when we have no CASE session yet, at least....

src/app/CASESessionManager.cpp
vivien-apple: ReleaseSession is doing a null check already. It is 2 lines under this method and this is why I have not added one here

bzbarsky-apple: Ah, OK.

src/darwin/Framework/CHIP/MTRDevice_Concrete.mm
vivien-apple: https://github.com/project-chip/connectedhomeip/pull/36315

src/darwin/Framework/CHIP/MTRDeviceController_Concrete.mm
vivien-apple: https://github.com/project-chip/connectedhomeip/pull/36315

PR #36302, opened 2024-10-30
[darwin-framework-tool][interactive] Add Ctrl+Z as a shortcut key to …

vivien-apple:
…suspend or resume the running controllers

#### Problem

This PR introduces a new shortcut, ‘Ctrl+Z’, to suspend or resume the controllers in interactive mode.

### DIFF
--- a/examples/darwin-framework-tool/commands/common/CHIPCommandBridge.mm
+++ b/examples/darwin-framework-tool/commands/common/CHIPCommandBridge.mm
@@ -358,6 +358,16 @@ void CHIPCommandBridge::ShutdownCommissioner()
     [[MTRDeviceControllerFactory sharedInstance] stopControllerFactory];
 }
 
+void CHIPCommandBridge::SuspendOrResumeCommissioners()
+{
+    for (auto & pair : mControllers) {
+        __auto_type * commissioner = pair.second;
+        if (commissioner.running) {
+            commissioner.suspended ? [commissioner resume] : [commissioner suspend];
+        }
+    }
+}
+
 CHIP_ERROR CHIPCommandBridge::StartWaiting(chip::System::Clock::Timeout duration)
 {
     auto waitingUntil = std::chrono::system_clock::now() + std::chrono::duration_cast<std::chrono::seconds>(duration);
--- a/examples/darwin-framework-tool/commands/interactive/InteractiveCommands.mm
+++ b/examples/darwin-framework-tool/commands/interactive/InteractiveCommands.mm
@@ -30,9 +30,10 @@
 constexpr char kInteractiveModeInstruction[] = "╔══════════════════════════════════════════════════════════════════╗\n"
                                                "║                        Interactive Mode                          ║\n"
                                                "╠══════════════════════════════════════════════════════════════════╣\n"
-                                               "║ Stop and restart stack: [Ctrl+_] & [Ctrl+^ ]                     ║\n"
-                                               "║ Trigger exit(0)       : [Ctrl+@]                                 ║\n"
-                                               "║ Quit Interactive      : 'quit()' or `quit`                       ║\n"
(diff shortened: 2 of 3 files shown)


examples/darwin-framework-tool/commands/interactive/InteractiveCommands.mm
```diff
@@ -30,9 +30,10 @@
...
                                                "║                        Interactive Mode                          ║\n"
                                                "╠══════════════════════════════════════════════════════════════════╣\n"
-                                               "║ Stop and restart stack: [Ctrl+_] & [Ctrl+^ ]                     ║\n"
-                                               "║ Trigger exit(0)       : [Ctrl+@]                                 ║\n"
-                                               "║ Quit Interactive      : 'quit()' or `quit`                       ║\n"
+                                               "║ Stop and restart stack    : [Ctrl+_] & [Ctrl+^ ]                 ║\n"
+                                               "║ Suspend/Resume controllers: [Ctrl+Z]                             ║\n"
+                                               "║ Trigger exit(0)           : [Ctrl+@]                             ║\n"
+                                               "║ Quit Interactive          : 'quit()' or `quit`                   ║\n"
```
ksperling-apple, line 36: While we're doing hot keys, it would be nice if Ctrl-D did a quit like how it does in shell.

examples/darwin-framework-tool/commands/interactive/InteractiveCommands.mm
```diff
@@ -333,6 +357,7 @@ el_status_t ExitFunction()
     el_bind_key(CTL('^'), RestartFunction);
     el_bind_key(CTL('_'), StopFunction);
     el_bind_key(CTL('@'), ExitFunction);
+    el_bind_key(CTL('Z'), SuspendOrResumeFunction);
```
bzbarsky-apple, line 360: This will make it impossible to use Ctrl-Z to suspect the tool itself and get back to the shell...  Can we use a different char here, possibly?

vivien-apple: I’m curious to know if `Ctrl-Z` works for you in interactive mode. By that, I mean does it suspend the interactive mode and return to the shell? It doesn’t work for me, and I assumed this was because `Ctrl-Z` is already handled by `editline`. I understand that it’s currently bound to `el_ring_bell`, which is why replacing it doesn’t cause any issues.
But maybe it works for others ?

examples/darwin-framework-tool/commands/common/CHIPCommandBridge.mm
```diff
@@ -358,6 +358,16 @@
     [[MTRDeviceControllerFactory sharedInstance] stopControllerFactory];
 }
 
+void CHIPCommandBridge::SuspendOrResumeCommissioners()
+{
+    for (auto & pair : mControllers) {
+        __auto_type * commissioner = pair.second;
+        if (commissioner.running) {
+            commissioner.suspended ? [commissioner resume] : [commissioner suspend];
```
bzbarsky-apple, line 366: Would be nice to log whether we are resuming or suspending (and what the commissioner involved is).

bzbarsky-apple (approved):

examples/darwin-framework-tool/commands/interactive/InteractiveCommands.mm
vivien-apple: I can try that in a follow up as it seems a new shortcut and I like to keep PR as small as possible. And additionally I took a quick try at it and I have not yet understood why it fails on first try 😗

examples/darwin-framework-tool/commands/common/CHIPCommandBridge.mm
vivien-apple: The framework is already quite verbose. ;)

bzbarsky-apple:

PR #36310, opened 2024-10-30
Add a largeMessage property support to Commands based on the quality setting in xml.

pidarped:
* Define a constant bool property in Commands via the corresponding zap template.
* Set the largeMessage=true for the 'CaptureSnapshot' command in AVStreamMgmt cluster xml.

### DIFF
--- a/src/app/data-model-provider/MetadataTypes.h
+++ b/src/app/data-model-provider/MetadataTypes.h
@@ -83,10 +83,12 @@ struct AttributeEntry
     static const AttributeEntry kInvalid;
 };
 
+// Bitmask values for different Command qualities.
 enum class CommandQualityFlags : uint32_t
 {
     kFabricScoped = 0x0001,
     kTimed        = 0x0002, // `T` quality on commands
+    kLargeMessage = 0x0004, // `L` quality on commands
 };
 
 struct CommandInfo
--- a/src/app/zap-templates/templates/app/cluster-objects-src.zapt
+++ b/src/app/zap-templates/templates/app/cluster-objects-src.zapt
@@ -261,5 +261,21 @@ bool CommandIsFabricScoped(ClusterId aCluster, CommandId aCommand)
     return false;
 }
 
+bool CommandHasLargePayload(ClusterId aCluster, CommandId aCommand)
+{
+    {{#zcl_clusters}}
+      {{#zcl_commands}}
+          {{#if isLargeMessage}}
+          if ((aCluster == Clusters::{{asUpperCamelCase parent.name}}::Id) &&
+		       (aCommand == Clusters::{{asUpperCamelCase parent.name}}::Commands::{{asUpperCamelCase name}}::Id))
+		  {
+              return true;
+		  }
+          {{/if}}
+      {{/zcl_commands}}
+    {{/zcl_clusters}}
+	return false;
+}
+
 } // namespace app
 } // namespace chip
--- a/zzz_generated/app-common/app-common/zap-generated/cluster-objects.cpp
+++ b/zzz_generated/app-common/app-common/zap-generated/cluster-objects.cpp
@@ -35101,5 +35101,20 @@ bool CommandIsFabricScoped(ClusterId aCluster, CommandId aCommand)
     return false;
 }
 
+bool CommandHasLargePayload(ClusterId aCluster, CommandId aCommand)
+{
+    if ((aCluster == Clusters::CameraAvStreamManagement::Id) &&
+        (aCommand == Clusters::CameraAvStreamManagement::Commands::CaptureSnapshot::Id))
+    {
+        return true;
+    }
+    if ((aCluster == Clusters::CameraAvStreamManagement::Id) &&
+        (aCommand == Clusters::CameraAvStreamManagement::Commands::CaptureSnapshotResponse::Id))
+    {
+        return true;
+    }
+    return false;
+}
+
 } // namespace app
 } // namespace chip
--- a/src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
+++ b/src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp
@@ -241,6 +241,8 @@ DataModel::CommandEntry CommandEntryFrom(const ConcreteClusterPath & clusterPath
     entry.info.flags.Set(DataModel::CommandQualityFlags::kFabricScoped,
                          CommandIsFabricScoped(clusterPath.mClusterId, clusterCommandId));
 
+    entry.info.flags.Set(DataModel::CommandQualityFlags::kLargeMessage,
+                         CommandHasLargePayload(clusterPath.mClusterId, clusterCommandId));
     return entry;
 }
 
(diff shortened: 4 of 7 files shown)


src/app/zap-templates/templates/app/ids/Commands.zapt
```diff
@@ -20,6 +20,7 @@ namespace Commands {
 {{/first}}
 namespace {{asUpperCamelCase label}} {
 static constexpr CommandId Id = {{asMEI manufacturerCode code}};
+static constexpr bool kRequiresLargePayloadTransport = {{isLargeMessage}};
```
bzbarsky-apple, line 23: This does not seem like the right place for this.  This sort of metadata generally lives in cluster-objects, not the ids header.  The one really is just for the ids.

bzbarsky-apple (changes requested):

src/app/zap-templates/templates/app/ids/Commands.zapt
andy31415: This looks like something that should eventually be plumbed through https://github.com/project-chip/connectedhomeip/blob/master/src/app/data-model-provider/MetadataTypes.h#L86 or similar (not sure if this is a quality or not, but it sure looks like a boolean flag on commands)

andy31415: This is how the codegen data model currently loads those: https://github.com/project-chip/connectedhomeip/blob/master/src/app/codegen-data-model-provider/CodegenDataModelProvider.cpp#L233

It seems like you may want to plumb this similarely to `CommandNeedsTimedInvoke` and `CommandIsFabricScoped`.

pidarped: Yes, looking at the auto-gen'ed code, it felt this was probably not the right place. Was looking to find the appropriate location. Thanks for the pointers.

pidarped: Modified to a `CommandIsLarge` method in the cluster-objects zapt

bzbarsky-apple: So the key here is that we will need to go from a dynamic command id to whether it's large, not from a statically known command name to whether it's large, right?  @pidarped

src/app/data-model-provider/MetadataTypes.h
```diff
@@ -87,6 +87,7 @@ enum class CommandQualityFlags : uint32_t
 {
     kFabricScoped = 0x0001,
     kTimed        = 0x0002, // `T` quality on commands
+    kIsLarge      = 0x0003, // `L` quality on commands
```
bzbarsky-apple, line 90:
```suggestion
    kLargeMessage      = 0x0004, // `L` quality on commands
```
These are bitmask values, not bit offsets.  And the name of the quality in the spec is "Large Message", though maybe the flag should be called "Large Payload"...

The good news is that CI caught the functional bug, at least....

src/app/zap-templates/templates/app/cluster-objects-src.zapt
```diff
@@ -261,5 +261,32 @@ bool CommandIsFabricScoped(ClusterId aCluster, CommandId aCommand)
     return false;
 }
 
+bool CommandIsLarge(ClusterId aCluster, CommandId aCommand)
```
bzbarsky-apple, line 264: CommandHasLargePayload?

src/app/zap-templates/templates/app/cluster-objects-src.zapt
```diff
@@ -261,5 +261,32 @@ bool CommandIsFabricScoped(ClusterId aCluster, CommandId aCommand)
...
+    switch (aCluster)
+    {
+    {{#zcl_clusters}}
+      {{#zcl_commands}}
+      {{#first}}
+      case Clusters::{{asUpperCamelCase parent.name}}::Id:
+      {
+          switch (aCommand) {
+      {{/first}}
+          {{#if isLargeMessage}}
```
bzbarsky-apple, line 275: See, _here_ it's named the right thing...

pidarped: Changed to CommandHasLargePayload.
Earlier, I tried to be consistent with what was in zap, i.e., isLargeMessage. But CommandIsLargeMessage does not make sense, so kept it as CommandIsLarge. Don't care either way. Both seem ok to me.

src/app/zap-templates/templates/app/ids/Commands.zapt
pidarped: Yes, I think so. But to also note, that ultimately the `L` quality is gleaned from what is specified in the xml for the Command.

zzz_generated/app-common/app-common/zap-generated/cluster-objects.cpp
```diff
@@ -35101,5 +35101,584 @@ bool CommandIsFabricScoped(ClusterId aCluster, CommandId aCommand)
...
 
+bool CommandHasLargePayload(ClusterId aCluster, CommandId aCommand)
+{
+    switch (aCluster)
+    {
+    case Clusters::Identify::Id: {
+        switch (aCommand)
+        {
+        default:
+            return false;
```
andy31415, line 35112: This is huge. How about we just switch for the "true" set and that is it?

bzbarsky-apple (approved):

zzz_generated/app-common/app-common/zap-generated/cluster-objects.cpp
bzbarsky-apple: The problem is that ZAP does not give us a clean way to enumerate "the true set", right?  @brdandu

pidarped: Yes. That is what I gathered as well. We have to go through all the clusters to check their list of commands and then check the specific quality.

yufengwangca (approved):

aronrosenberg (approved):

zzz_generated/app-common/app-common/zap-generated/cluster-objects.cpp
andy31415: You can loop over everything and have a `if true` I imagine. It just needs compressing the switch into a compare.

andy31415: I am generally wondering if this could read:

```cpp
bool CommandHasLargePayload(ClusterId aCluster, CommandId aCommand)
{
  if ((aClusterId == Clusters::CameraAvStreamManagement::Id) 
      && (aCommandId == Clusters::CameraAvStreamManagement::Commands::CaptureSnapshot::Id)
  {
    return true;
  }
  
   if ((aClusterId == Clusters::CameraAvStreamManagement::Id) 
       && (aCommandId == Clusters::CameraAvStreamManagement::Commands::CaptureSnapshotResponse::Id)
  {
    return true;
  }
  
  return false;
}
```

We keep accepting sub-part code because it is code generated. I think we should not do that.

bzbarsky-apple: Yes, you could certainly generate that code.

pidarped: Yes. I was not too conversant with the zap templates and their code generation model. Was looking at only the switch based templates. But the template for generating the above code seems pretty straight-forward. Thanks.

PR #36318, opened 2024-10-31
[Silabs] Rename Wifi interface sources files

mkardous-silabs:
#### Description
PR structures the Wi-Fi headers and sources to start implementing a Silabs Wifi Abstraction layer.
PR has no functionnal changes.

- rename files to match the interface they are implementing
- Seperate the wifi support source_set for it to be used by other modules
- Delete files that weren't being used

#### Tests
Manual tests with the BRD4338A
Ci for build validation


jmartinez-silabs (approved):
Should `WifiInterfaceAbstraction` live in `src/platform/silabs/platformAbstraction` along with the <wifiplatform>WifiInterface.cpp

it could be done later and I'm not sure if there would be some circular dependencies issue. but it seems like it would be the correct final destination.

andy31415 (approved):

PR #36319, opened 2024-10-31
Enforce `DataModel::Provider` everywhere

andy31415:
Remove the non-provider code to only maintain a single code path for data model handling.

This should enable code cleanups in the future:
  - ACL can be validated outside of read/write implementations (directly in IM)
  - IM could handle global attributes directly as it has metadata access

#### Changes

- delete `ember-compatibility-functions`
- Maintain only the `enabled` flags and remove all ifdef conditionals and code splits
   - In particular moved reporting/Read-DataModel and AttributePathExpandIterator-DataModel to be the main intefaces as a `h` and `cpp` file.
- cleanup several overrides within tests and dynamic files (a few that were reused remained and were marked as static)


src/app/AttributePathExpandIterator.h
```diff
@@ -17,30 +17,113 @@
...
+ *
+ * AttributePathExpandIterator will expand attribute paths with wildcards, and only emit existing paths for
+ * AttributePathParams with wildcards. For AttributePathParams with a concrete path (i.e. does not contain wildcards),
+ * AttributePathExpandIterator will emit them as-is.
+ *
+ * The typical use of AttributePathExpandIterator may look like:
+ * ConcreteAttributePath path;
+ * for (AttributePathExpandIterator iterator(AttributePathParams); iterator.Get(path); iterator.Next()) {...}
+ *
+ * The iterator does not copy the given AttributePathParams, The given AttributePathParams must be valid when using the iterator.
```
bzbarsky-apple, line 40:
```suggestion
 * The iterator does not copy the given AttributePathParams. The given AttributePathParams must remain valid when using the iterator.
```

src/app/AttributePathExpandIterator.h
```diff
@@ -17,30 +17,113 @@
...
+ * If the set of endpoints, clusters, or attributes that are supported changes, AttributePathExpandIterator must be
+ * reinitialized.
+ *
+ * A initialized iterator will return the first valid path, no need to call Next() before calling Get() for the first time.
+ *
+ * Note: The Next() and Get() are two separate operations by design since a possible call of this iterator might be:
```
bzbarsky-apple, line 46:
```suggestion
 * Note: Next() and Get() are two separate operations by design since a possible call of this iterator might be:
```

src/app/CommandHandlerImpl.cpp
```diff
@@ -475,25 +475,6 @@ Status CommandHandlerImpl::ProcessGroupCommandDataIB(CommandDataIB::Parser & aCo
...
-    // to ACL and command presence checks does not matter, because the behavior
-    // is the same for all of them: ignore the path.
-#if !CHIP_CONFIG_USE_DATA_MODEL_INTERFACE
-
-    // Without data model interface, we can query individual commands.
-    // Data model interface queries commands by a full path so we need endpointID as well.
-    //
-    // Since this is a performance update and group commands are never timed,
-    // missing this should not be that noticeable.
-    if (CommandNeedsTimedInvoke(clusterId, commandId))
```
bzbarsky-apple, line 490: So in the CHIP_CONFIG_USE_DATA_MODEL_INTERFACE case (the only case now): where is this check done?

src/app/WriteHandler.h
```diff
@@ -200,10 +200,8 @@ class WriteHandler : public Messaging::ExchangeDelegate
     Optional<ConcreteAttributePath> mProcessingAttributePath;
     Optional<AttributeAccessToken> mACLCheckCache = NullOptional;
 
-#if CHIP_CONFIG_USE_DATA_MODEL_INTERFACE
     DataModel::Provider * mDataModelProvider = nullptr;
     std::optional<ConcreteAttributePath> mLastSuccessfullyWrittenPath;
```
bzbarsky-apple, line 204: Followup: can we have mProcessingAttributePath and mLastSuccessfullyWrittenPath set at the same time?  If not there is some possible savings here...

src/app/reporting/Read.cpp
bzbarsky-apple, line 1: So this is, now that we have things in their final shape, really oddly names.  This file (header and cpp) has precisely two functions:

1) A data version equality checker (which is not even read-specific; writes might need to do that check!)
2) A function to encode the value of an attribute to TLV, whether it's for read or subscribe.

I guess this should be a followup, to not block this, but we should probably split this up into separate files with names that makes sense.

Alternately, we could move RetrieveClusterData back into Engine.cpp, which is where it really makes the most sense, and still do something sane with IsClusterDataVersionEqualTo.

src/app/reporting/Read.cpp
```diff
@@ -26,21 +26,13 @@
...
 DataModel::ActionReturnStatus RetrieveClusterData(DataModel::Provider * dataModel,
                                                   const Access::SubjectDescriptor & subjectDescriptor, bool isFabricFiltered,
                                                   AttributeReportIBs::Builder & reportBuilder,
                                                   const ConcreteReadAttributePath & path, AttributeEncodeState * encoderState)
 {
-    // Odd ifdef is to only do this if the `Read-Check` does not do it already.
-#if !CHIP_CONFIG_USE_EMBER_DATA_MODEL
-    ChipLogDetail(DataManagement, "<RE:Run> Cluster %" PRIx32 ", Attribute %" PRIx32 " is dirty", path.mClusterId,
-                  path.mAttributeId);
-    DataModelCallbacks::GetInstance()->AttributeOperation(DataModelCallbacks::OperationType::Read,
```
bzbarsky-apple, line 40: Why is this being removed?

src/app/server/Server.cpp
```diff
@@ -95,30 +93,12 @@ class DeviceTypeResolver : public chip::Access::AccessControl::DeviceTypeResolve
         {
             if (type->deviceTypeId == deviceType)
```
bzbarsky-apple, line 94: This looks like the second copy of this function we have sitting around somewhere.  Should this just be a non-virtual member method on DataModel::Provider?  Followup for that, obviously.

src/darwin/Framework/CHIP/ServerEndpoint/MTRServerAccessControl.mm
```diff
@@ -47,7 +46,14 @@
 public:
     bool IsDeviceTypeOnEndpoint(DeviceTypeId deviceType, EndpointId endpoint) override
     {
-        return app::IsDeviceTypeOnEndpoint(deviceType, endpoint);
+        chip::app::DataModel::Provider * model = chip::app::InteractionModelEngine::GetInstance()->GetDataModelProvider();
```
bzbarsky-apple, line 49:
```suggestion
        app::DataModel::Provider * model = app::InteractionModelEngine::GetInstance()->GetDataModelProvider();
```
But note this is now the _third_ copy of this code.

src/app/CommandHandlerImpl.cpp
andy31415: This was an optimization that was possible with direct ember calls because the command flags were global. In the data model interface we need the full path, so we cannot early-optimize (unless we do more work).

The individual check (that matches spec text more closely) was added in `mpCallback->ValidateCommandCanBeDispatched` (in #35897). Specifically timed invoke validation is done in `CheckCommandFlags`: https://github.com/project-chip/connectedhomeip/blob/master/src/app/InteractionModelEngine.cpp#L1859C61-L1859C78

src/app/reporting/Read.cpp
andy31415: Bad grep instruction ... thank you for catching.

src/app/server/Server.cpp
andy31415: Started #36327 to track these things.

Overall I would be weary of adding too much into DataModel::Provider if at all possible, however all for code reuse however we figure out it should look like.

src/app/reporting/Read.cpp
andy31415: added this work into #36327

src/app/WriteHandler.h
andy31415: Added to #36327

src/app/CommandHandlerImpl.cpp
bzbarsky-apple: Great, "ValidateCommandCanBeDispatched call below" is perfect.

bzbarsky-apple (approved):

src/app/tests/BUILD.gn
```diff
@@ -171,7 +171,7 @@ source_set("app-test-stubs") {
     "test-ember-api.cpp",
     "test-ember-api.h",
 
-    # The overrides in these files are overrides from ember-compatibility-functions
+    # The overrides in these files are overrides from ember code generated code
```
jmartinez-silabs, line 174:
```suggestion
    # The overrides in these files are overrides from ember generated code
```

jmartinez-silabs (approved):

PR #36320, opened 2024-10-31
Remove "run on work queue" bits from non-concrete controllers.

bzbarsky-apple:
There is no "work queue" in the sense of the Matter queue involved unless we have a concrete controller.

* syncRunOnWorkQueueWithReturnValue was unused on MTRDeviceController, so can just be removed.

* syncRunOnWorkQueueWithBoolReturnValue was unused on MTRDeviceController, so can just be removed.

* After that syncRunOnWorkQueue becomes unused and can be removed.

* At this point chipWorkQueue is not used on the base class and can be moved into the subclasses as needed.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRDeviceController_Internal.h
+++ b/src/darwin/Framework/CHIP/MTRDeviceController_Internal.h
@@ -68,8 +68,6 @@ NS_ASSUME_NONNULL_BEGIN
 @property (readonly, assign) os_unfair_lock_t deviceMapLock;
 
 @property (readwrite, nonatomic) NSUUID * uniqueIdentifier;
-
-// queue used to serialize all work performed by the MTRDeviceController
 // (moved here so subclasses can initialize differently)
 
 - (instancetype)initForSubclasses:(BOOL)startSuspended;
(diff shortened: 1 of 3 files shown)


src/darwin/Framework/CHIP/MTRDeviceController_Internal.h
```diff
@@ -71,7 +71,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 // queue used to serialize all work performed by the MTRDeviceController
 // (moved here so subclasses can initialize differently)
```
vivien-apple, line 73: Do we need to keep the comment ?

src/darwin/Framework/CHIP/MTRDeviceController_Internal.h
```diff
@@ -71,7 +71,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 // queue used to serialize all work performed by the MTRDeviceController
```
jtung-apple, line 72: The comments for the removed property could be removed as well.

vivien-apple (approved):

jtung-apple (approved):

src/darwin/Framework/CHIP/MTRDeviceController_Internal.h
bzbarsky-apple: Good catch, removing.

bzbarsky-apple, line 72:
```suggestion
```

bzbarsky-apple:

PR #36328, opened 2024-10-31
Move concurrentSubscriptionPool property to MTRDeviceController_Concrete.

bzbarsky-apple:
The non-concrete case does not directly create subscriptions, so has no need of a pool for them.


jtung-apple (approved):

bzbarsky-apple:

PR #36332, opened 2024-11-01
[ESP32]: fix compile error for c6 bt lib path

DejinChen:
#### Problem
The path of esp32c6-bt-lib has changed in idf/release/v5.3 branch, Modify CMakeLists to be compatible with the old and new paths.

#### Test
Compiled esp32c6 example on idf/release/v5.3 (new path) and idf/v5.3.1 tag (old path).


chshu (approved):

config/esp32/components/chip/CMakeLists.txt
```diff
@@ -427,7 +427,11 @@ if(CONFIG_BT_ENABLED)
     if((target_name STREQUAL "esp32h2") OR (target_name STREQUAL "esp32c2") OR (target_name STREQUAL "esp32c6"))
         idf_component_get_property(bt_dir bt COMPONENT_DIR)
         list(APPEND chip_libraries $<TARGET_FILE:${bt_lib}>)
-        list(APPEND chip_libraries "${bt_dir}/controller/lib_${target_name}/${target_name}-bt-lib/libble_app.a")
+        if (EXISTS ${bt_dir}/controller/lib_${target_name}/${target_name}-bt-lib/libble_app.a)
+            list(APPEND chip_libraries "${bt_dir}/controller/lib_${target_name}/${target_name}-bt-lib/libble_app.a")
+        elseif(EXISTS ${bt_dir}/controller/lib_${target_name}/${target_name}-bt-lib/${target_name}/libble_app.a)
```
andy31415, line 432: is it normal for non of the libble_app.a to exists? The previous code seems to always want to add a libble_app.a, this version may silently fail to add such a archive if it does not exist. Should we have a else branch with at least a warning?

DejinChen: The existing situation is already covered with If condition and elseif condition. But we can indeed add warnings in case of future changes. Have added warning message.

andy31415: My point was that this maybe should be `if/else` not `if/elseif`. because if you have the latter, both if conditions could fail ... so then you are not adding any ble library even though you seem to need one. So I would expect `if/elseif/else { ERROR_MESSAGE }`

andy31415 (approved):

PR #36348, opened 2024-11-03
[ICD] Shutdown icd client storage when destroying android controller

### LINKED ISSUE #36346: [BUG] Reinitializing ChipController on Android faults in ICDClientStorage
### Reproduction steps

@yunhanw-google 
My Android app supports multiple homes, when I switch homes I need to make another ChipController object because I have to initialize using a different operational key. Now when I make the second controller I am faulting in the ICD code. I have no ICD specific code in my Android app. I don't explicitly delete the first controller instance, I was relying on garbage collection to recover it so there will be two controller instances around until that happens. This is code which had been working for several years.  Do I have to do something special now for ICD support? 

Note that I am a little bit behind head (Aug 27), if this has been fixed I can update. But I looked in the git logs and didn't see anything that looked like a fix.
~~~
2024-11-01 23:33:00.500 30717-30717 AndroidRuntime          com.lowpan.m2                        E  FATAL EXCEPTION: main (Ask Gemini)
                                                                                                    Process: com.lowpan.m2, PID: 30717
                                                                                                    chip.devicecontroller.ChipDeviceControllerException: src/app/icd/client/DefaultICDClientStorage.cpp:159: CHIP Error 0x00000003: Incorrect state
                                                                                                    	at chip.devicecontroller.ChipDeviceController.newDeviceController(Native Method)
                                                                                                    	at chip.devicecontroller.ChipDeviceController.<init>(ChipDeviceController.java:69)
                                                                                                    	at com.lowpan.m2.chip.ChipClient$1$1.emit(ChipClient.kt:196)
                                                                                                    	at com.lowpan.m2.chip.ChipClient$1$1.emit(ChipClient.kt:96)
                                                                                                    	at kotlinx.coroutines.flow.StateFlowImpl.collect(StateFlow.kt:396)
                                                                                                    	at kotlinx.coroutines.flow.StateFlowImpl$collect$1.invokeSuspend(Unknown Source:15)
                                                                                                    	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
                                                                                                    	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:108)
                                                                                                    	at android.os.Handler.handleCallback(Handler.java:959)
                                                                                                    	at android.os.Handler.dispatchMessage(Handler.java:100)
                                                                                                    	at android.os.Looper.loopOnce(Looper.java:232)
                                                                                                    	at android.os.Looper.loop(Looper.java:317)
                                                                                                    	at android.app.ActivityThread.main(ActivityThread.java:8705)
                                                                                                    	at java.lang.reflect.Method.invoke(Native Method)
                                                                                                    	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:580)
                                                                                                    	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:886)
~~~

~~~
                    chipDeviceController = ChipDeviceController(
                        ControllerParams.newBuilder(operationalKeyConfig()).setUdpListenPort(0)
                            .setControllerVendorId(Lowpan.COMPANY).build()
                    ).also { chipDeviceController ->
~~~

### Bug prevalence

always

ea679d2dc674f576f4d391d1d71af1489010e580

### Platform

android

### Platform Version(s)

_No response_

### Anything else?

_No response_

yunhanw-google:
Summary: In android controller , DefaultICDClientStorage has been handled as static instance, which is reused across multiple controllers, without this PR, it would crash when second controller reinitialize the ICDClientStorage

In order to resolve this issue, we need shutdown icd client storage when destroying android controller

fixes: https://github.com/project-chip/connectedhomeip/issues/36346

### DIFF
--- a/src/app/icd/client/DefaultICDClientStorage.cpp
+++ b/src/app/icd/client/DefaultICDClientStorage.cpp
@@ -542,5 +542,14 @@ CHIP_ERROR DefaultICDClientStorage::ProcessCheckInPayload(const ByteSpan & paylo
     iterator->Release();
     return CHIP_ERROR_NOT_FOUND;
 }
+
+void DefaultICDClientStorage::Shutdown()
+{
+    mICDClientInfoIterators.ReleaseAll();
+    mpClientInfoStore = nullptr;
+    mpKeyStore        = nullptr;
+    mFabricList.clear();
+}
+
 } // namespace app
 } // namespace chip
--- a/src/app/icd/client/DefaultICDClientStorage.h
+++ b/src/app/icd/client/DefaultICDClientStorage.h
@@ -119,10 +119,10 @@ public:
 
     CHIP_ERROR ProcessCheckInPayload(const ByteSpan & payload, ICDClientInfo & clientInfo,
                                      Protocols::SecureChannel::CounterType & counter) override;
-    
+
     /**
      * Shutdown DefaultICDClientStorage
-     * 
+     *
      */
     void Shutdown();
 
--- a/src/app/icd/client/DefaultICDClientStorage.h
+++ b/src/app/icd/client/DefaultICDClientStorage.h
@@ -119,6 +119,12 @@ public:
 
     CHIP_ERROR ProcessCheckInPayload(const ByteSpan & payload, ICDClientInfo & clientInfo,
                                      Protocols::SecureChannel::CounterType & counter) override;
+    
+    /**
+     * Shutdown DefaultICDClientStorage
+     * 
+     */
+    void Shutdown();
 
 #if CONFIG_BUILD_FOR_HOST_UNIT_TEST
     size_t GetFabricListSize() { return mFabricList.size(); }
--- a/src/app/icd/client/DefaultICDClientStorage.h
+++ b/src/app/icd/client/DefaultICDClientStorage.h
@@ -121,7 +121,7 @@ public:
                                      Protocols::SecureChannel::CounterType & counter) override;
 
     /**
-     * Shutdown DefaultICDClientStorage
+     * Shut down DefaultICDClientStorage
      *
      */
     void Shutdown();
--- a/src/controller/java/AndroidDeviceControllerWrapper.cpp
+++ b/src/controller/java/AndroidDeviceControllerWrapper.cpp
@@ -50,6 +50,7 @@ using namespace TLV;
 
 AndroidDeviceControllerWrapper::~AndroidDeviceControllerWrapper()
 {
+    getICDClientStorage()->Shutdown();
     mController->Shutdown();
 
     if (mKeypairBridge != nullptr)


drempelg (approved):

src/app/icd/client/DefaultICDClientStorage.cpp
```diff
@@ -161,7 +161,13 @@ void DefaultICDClientStorage::ICDClientInfoIteratorImpl::Release()
 CHIP_ERROR DefaultICDClientStorage::Init(PersistentStorageDelegate * clientInfoStore, Crypto::SymmetricKeystore * keyStore)
 {
     VerifyOrReturnError(clientInfoStore != nullptr && keyStore != nullptr, CHIP_ERROR_INVALID_ARGUMENT);
+    if (mpClientInfoStore != nullptr && mpKeyStore != nullptr)
```
arkq, line 164: nit: Maybe use `VerifyOrReturnValue` to follow the convention of params checking.

arkq (approved):

src/app/icd/client/DefaultICDClientStorage.cpp
yunhanw-google: prefer to keep the current style for multiple check here for readability consideration.

bzbarsky-apple (approved):

src/app/icd/client/DefaultICDClientStorage.cpp
```diff
@@ -161,7 +161,13 @@ void DefaultICDClientStorage::ICDClientInfoIteratorImpl::Release()
 CHIP_ERROR DefaultICDClientStorage::Init(PersistentStorageDelegate * clientInfoStore, Crypto::SymmetricKeystore * keyStore)
 {
     VerifyOrReturnError(clientInfoStore != nullptr && keyStore != nullptr, CHIP_ERROR_INVALID_ARGUMENT);
+    if (mpClientInfoStore != nullptr && mpKeyStore != nullptr && mpClientInfoStore == clientInfoStore && mpKeyStore == keyStore)
+    {
+        ChipLogProgress(ICD, "DefaultICDClientStorage has already been initialized");
```
bzbarsky-apple, line 166: Though realistically: the right response to multiple calls on `Init()` on an object is to error out of all but the first, because clearly the API consumer is just completely confused.  So actually, I don't understand why we would want to make this change.

bzbarsky-apple (changes requested):

src/app/icd/client/DefaultICDClientStorage.cpp
yunhanw-google: yes, I agree I think we may introduce a function call IsInitialized() and a state(Uninitialized/Initialized)for ICDClientStorage, it seems to be more natural to me
just update the code to reflect this proposal

src/controller/java/AndroidDeviceControllerWrapper.cpp
```diff
@@ -184,13 +184,16 @@ AndroidDeviceControllerWrapper * AndroidDeviceControllerWrapper::AllocateNew(
         chip::Credentials::SetDeviceAttestationVerifier(GetDefaultDACVerifier(testingRootStore));
     }
 
-    *errInfoOnFailure = getICDClientStorage()->Init(wrapperStorage, &wrapper->mSessionKeystore);
-    if (*errInfoOnFailure != CHIP_NO_ERROR)
+    auto icdStorage = getICDClientStorage();
+    if (icdStorage != nullptr && !(icdStorage->IsInitialized()))
```
bzbarsky-apple, line 188: OK, but what if the wrapper's storage and keystore don't match the existing ones on the icdStorage?  How is this supposed to work?

yunhanw-google: we don't support changing wrapper's storage and keystore, right?

bzbarsky-apple: > we don't support changing wrapper's storage and keystore, right?

I don't know; the wrapper is part of the Android API and I am not familiar with it.  

But if there are two calls to `AndroidDeviceControllerWrapper::AllocateNew`, won't those have different `wrapperStorage` and `&wrapper->mSessionKeystore` values?  Is calling `AllocateNew` twice just not supported?  But this whole issue that this PR is claiming to fix seems to be about "across multiple controoler
 [sic]", which means multiple calls to `AndroidDeviceControllerWrapper::AllocateNew`, I would think.

Let me put it this way: at this point, short of an actual test exercising this code in the scenario of https://github.com/project-chip/connectedhomeip/issues/36346, I don't believe this fixes https://github.com/project-chip/connectedhomeip/issues/36346 at all.

src/app/icd/client/DefaultICDClientStorage.cpp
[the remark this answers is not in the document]
yunhanw-google: Just update code to shutdown icd client storage when destroying android controller

src/app/icd/client/DefaultICDClientStorage.cpp
yunhanw-google: Just update code to shutdown icd client storage when destroying android controller, thanks

src/controller/java/AndroidDeviceControllerWrapper.cpp
yunhanw-google: Just update code to shutdown icd client storage when destroying android controller, thanks

src/app/icd/client/DefaultICDClientStorage.h
```diff
@@ -120,6 +120,12 @@ class DefaultICDClientStorage : public ICDClientStorage
     CHIP_ERROR ProcessCheckInPayload(const ByteSpan & payload, ICDClientInfo & clientInfo,
                                      Protocols::SecureChannel::CounterType & counter) override;
 
+    /**
+     * Shutdown DefaultICDClientStorage
```
bzbarsky-apple, line 124:
```suggestion
     * Shut down DefaultICDClientStorage
```

bzbarsky-apple (approved):

PR #36357, opened 2024-11-04
Add skip_bytes functionality to OTA Provider and BDX downloader class

lucicop:
Implementation to extend BDX class with the capability to skip bytes during OTA download - functionality is provisioned for, but not yet implemented. This PR fixes that.

### DIFF
--- a/examples/ota-provider-app/ota-provider-common/BdxOtaSender.cpp
+++ b/examples/ota-provider-app/ota-provider-common/BdxOtaSender.cpp
@@ -124,16 +124,24 @@ void BdxOtaSender::HandleTransferSessionOutput(TransferSession::OutputEvent & ev
 
         break;
     }
-    case TransferSession::OutputEventType::kQueryReceived: {
+    case TransferSession::OutputEventType::kQueryReceived:
+    case TransferSession::OutputEventType::kQueryWithSkipReceived: {
         TransferSession::BlockData blockData;
         uint16_t blockSize   = mTransfer.GetTransferBlockSize();
         uint16_t bytesToRead = blockSize;
+        uint64_t bytesToSkip = 0;
+
+        if (event.EventType == TransferSession::OutputEventType::kQueryWithSkipReceived)
+        {
+            bytesToSkip = event.bytesToSkip.BytesToSkip;
+        }
+        uint64_t seekOffset = mNumBytesSent + bytesToSkip;
 
         // TODO: This should be a utility function in TransferSession
-        if (mTransfer.GetTransferLength() > 0 && mNumBytesSent + blockSize > mTransfer.GetTransferLength())
+        if ((mTransfer.GetTransferLength() > 0) && ((seekOffset + blockSize) > mTransfer.GetTransferLength()))
         {
             // cast should be safe because of condition above
-            bytesToRead = static_cast<uint16_t>(mTransfer.GetTransferLength() - mNumBytesSent);
+            bytesToRead = static_cast<uint16_t>(mTransfer.GetTransferLength() - seekOffset);
         }
 
         chip::System::PacketBufferHandle blockBuf = chip::System::PacketBufferHandle::New(bytesToRead);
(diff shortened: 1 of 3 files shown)


bzbarsky-apple (approved):

examples/ota-provider-app/ota-provider-common/BdxOtaSender.cpp
```diff
@@ -124,16 +124,24 @@ void BdxOtaSender::HandleTransferSessionOutput(TransferSession::OutputEvent & ev
...
+    case TransferSession::OutputEventType::kQueryReceived:
+    case TransferSession::OutputEventType::kQueryWithSkipReceived: {
         TransferSession::BlockData blockData;
         uint16_t blockSize   = mTransfer.GetTransferBlockSize();
         uint16_t bytesToRead = blockSize;
+        uint32_t bytesToSkip = 0;
+
+        if (event.EventType == TransferSession::OutputEventType::kQueryWithSkipReceived)
+        {
+            bytesToSkip = static_cast<uint32_t>(event.bytesToSkip.BytesToSkip);
```
tehampson, line 136: BytesToSkip is uint64_t, right? If so should we at least assert that it would fix into uint32_t?

examples/ota-provider-app/ota-provider-common/BdxOtaSender.cpp
```diff
@@ -124,16 +124,24 @@ void BdxOtaSender::HandleTransferSessionOutput(TransferSession::OutputEvent & ev
...
+        }
+        uint32_t seekOffset = mNumBytesSent + bytesToSkip;
```
tehampson, line 138: Should we make sure that this value doesn't overflow in some way? If not, should there be comment on why we are not expect this to overflow?

tehampson (approved):

tehampson:
@tehampson so I get emails about this PR

examples/ota-provider-app/ota-provider-common/BdxOtaSender.cpp
```diff
@@ -124,16 +124,24 @@ void BdxOtaSender::HandleTransferSessionOutput(TransferSession::OutputEvent & ev
 
         break;
     }
-    case TransferSession::OutputEventType::kQueryReceived: {
+    case TransferSession::OutputEventType::kQueryReceived:
+    case TransferSession::OutputEventType::kQueryWithSkipReceived: {
```
tcarmelveilleux, line 128: How is a client supposed to know that this is now supported?

This was mandatory to support in 1.0. How come our SDK examples did not support it?

examples/ota-provider-app/ota-provider-common/BdxOtaSender.cpp
lucicop: Indeed, BytesToSkip is uint64_t, but mNumBytesSent is uint32_t. I think it makes sense to change it to uint64_t as well - to make it as future proof as possible :) What do you think ?

examples/ota-provider-app/ota-provider-common/BdxOtaSender.cpp
lucicop: I have no idea... I guess future implementations could take advantage of it, if needed.

lucicop: Would it be a possible to signal support for SkipBytes in the FeatureMap attribute of OTA Provider cluster ?

examples/ota-provider-app/ota-provider-common/BdxOtaSender.cpp
lucicop: On a second thought, if we take into account that 4 bytes for the image size mean 4.194.303 blocks of maximum size 1024 bytes, and that it takes at least (most optimistically) 100msec to transfer one block - realistically one sec or more, we end up with transfer time of minimum 419430 seconds for an image with size fitting 4 bytes. This is 116 hours - which I don't think it a delay somebody is willing to wait for a new software version update :) So just an assert that it fits 4 bytes and doesn't overflow should be enough.

tehampson: SGTM. Sorry for the delayed response, my email filters seems to have stopped working :(

lucicop: I have set it to uint64_t, to simplify the logic.

examples/ota-provider-app/ota-provider-common/BdxOtaSender.cpp
lucicop: In this context, an overflow is highly unlikely because:
- Both mNumBytesSent and bytesToSkip are of type uint64_t, which can represent extremely large values (up to ~18 exabytes).
- The data being sent is constrained to realistic file sizes far smaller than the maximum representable value of a uint64_t.

PR #36358, opened 2024-11-04
Fix: do not try to resubscribe on invalid action error

j0tunn:
For now if device doesn't support some clusters client endlessly tries to resubscribe getting `IM Error 0x00000580: General error: 0x80 (INVALID_ACTION)` in response. And while this happens no other commands could be sent to device.

This PR fixes resubscription policy to just fail subscription on `INVALID_ACTION`

### DIFF
--- a/src/app/ReadClient.cpp
+++ b/src/app/ReadClient.cpp
@@ -43,6 +43,7 @@ namespace chip {
 namespace app {
 
 using Status = Protocols::InteractionModel::Status;
+using ClusterStatusCode = Protocols::InteractionModel::ClusterStatusCode;
 
 ReadClient::ReadClient(InteractionModelEngine * apImEngine, Messaging::ExchangeManager * apExchangeMgr, Callback & apCallback,
                        InteractionType aInteractionType) :
@@ -1269,6 +1270,15 @@ CHIP_ERROR ReadClient::DefaultResubscribePolicy(CHIP_ERROR aTerminationCause)
     {
         ChipLogProgress(DataManagement, "ICD device is inactive, skipping scheduling resubscribe within DefaultResubscribePolicy");
         return CHIP_ERROR_LIT_SUBSCRIBE_INACTIVE_TIMEOUT;
+    }
+
+    if (aTerminationCause.IsIMStatus() && ClusterStatusCode(aTerminationCause).GetStatus() == Status::InvalidAction)
+    {
+        // InvalidAction in most cases means some Interaction Model mismatch with the target device
+        // That's not something that could be fixed by retry
+        ChipLogError(DataManagement, "Stop resubscription due to unrecoverable error %" CHIP_ERROR_FORMAT,
+                     aTerminationCause.Format());
+        return aTerminationCause;
     }
 
     VerifyOrReturnError(IsIdle(), CHIP_ERROR_INCORRECT_STATE);


src/app/ReadClient.cpp
```diff
@@ -1271,6 +1271,13 @@ CHIP_ERROR ReadClient::DefaultResubscribePolicy(CHIP_ERROR aTerminationCause)
         return CHIP_ERROR_LIT_SUBSCRIBE_INACTIVE_TIMEOUT;
     }
 
+    if (aTerminationCause.IsIMStatus() &&
+        aTerminationCause.GetSdkCode() == static_cast<uint8_t>(Protocols::InteractionModel::Status::InvalidAction))
```
andy31415, line 1275: How about using ClusterStatusCode from a CHIP_ERROR: https://github.com/project-chip/connectedhomeip/blob/master/src/protocols/interaction_model/StatusCode.h#L71

That way the logic of what status is contained in an error is shared.

src/app/ReadClient.cpp
```diff
@@ -1271,6 +1271,13 @@ CHIP_ERROR ReadClient::DefaultResubscribePolicy(CHIP_ERROR aTerminationCause)
...
+    {
+        ChipLogError(DataManagement, "Stop resubscription due to unrecoverable error %" CHIP_ERROR_FORMAT, aTerminationCause.Format());
```
andy31415, line 1277: needs a comment on why `InvalidAction` is never recoverable.

src/app/ReadClient.cpp
```diff
@@ -1271,6 +1272,15 @@ CHIP_ERROR ReadClient::DefaultResubscribePolicy(CHIP_ERROR aTerminationCause)
         return CHIP_ERROR_LIT_SUBSCRIBE_INACTIVE_TIMEOUT;
     }
 
+    if (aTerminationCause.IsIMStatus() && ClusterStatusCode(aTerminationCause).GetStatus() == Status::InvalidAction)
```
andy31415, line 1275:
```suggestion
    if (ClusterStatusCode(aTerminationCause).GetStatus() == Status::InvalidAction)
```

src/app/ReadClient.cpp
```diff
@@ -1271,6 +1272,15 @@ CHIP_ERROR ReadClient::DefaultResubscribePolicy(CHIP_ERROR aTerminationCause)
...
+    {
+        // InvalidAction in most cases means some Interaction Model mismatch with the target device
```
andy31415, line 1277: the `in most cases` seems vague, so does `some Interaction Model mismatch`. This feels like a "we do not know what the problem is, but this kind of works" ... we need stronger root cause investigation I believe.

Looking through the code on ReadHandler we seem to return InvalidAction on:
   - OnMessageReceived when not having received a StatusResponse
   - OnInitialRequest when processing the read or subscribe fails without setting a different IM status.

Trying to see what the specification states, it seems to say in `8.2.5.3 Incoming Action` that `If this layer receives a message for an action that is not expected semantically, has invalid action information, or has an error not described in this specification, a Status Response action with an INVALID_ACTION Status Code SHALL be generated as defined in Status Response Action, and the associated transaction and interaction SHALL terminate.`

And `Invalid Action` says `The action is malformed, has missing fields, or fields with invalid values. Action not carried out.`

Maybe the the invalid action definition would help here, however I would like a sanity review from someone more familiar with the spec and how this works. @bzbarsky-apple @tennessee-google @yunhanw-google please help with opinions here. I sugges:

```suggestion
        // InvalidAction means "The action is malformed, has missing fields, or fields with invalid 
        // values. Action not carried out." according to the spec and retrying will not fix this as we would
        // send out the same parameters (that are malformed, missing fields or invalid) so there is
        // no point to retry.
```

src/app/ReadClient.cpp
```diff
@@ -1271,6 +1272,15 @@ CHIP_ERROR ReadClient::DefaultResubscribePolicy(CHIP_ERROR aTerminationCause)
...
+        // That's not something that could be fixed by retry
```
bzbarsky-apple, line 1278: InvalidAction for a subscription in practice means either "you tried to subscribe to things that don't exist" or "you don't have the right ACLs".

First of all, the documentation should say that.

Second, for that second case, it may be the case that retrying until something fixes the ACLs is in fact the right behavior.

The PR description says "while this happens no other commands could be sent to device", but that sounds _very_ odd.  Why can't you send anything else while resubscribes are happening, especially if they go into backoff?

And given the backoff behavior, it's very much not clear to me that we want to make this change: the client will just back off, so in practice there should be minimal impact here.

src/app/ReadClient.cpp
```diff
@@ -1271,6 +1272,15 @@ CHIP_ERROR ReadClient::DefaultResubscribePolicy(CHIP_ERROR aTerminationCause)
...
+        ChipLogError(DataManagement, "Stop resubscription due to unrecoverable error %" CHIP_ERROR_FORMAT,
```
bzbarsky-apple, line 1279: I disagree that this is an unrecoverable error.  We just don't know, based on the information we have whether it's recoverable....

src/app/ReadClient.cpp
j0tunn: > The PR description says "while this happens no other commands could be sent to device", but that sounds very odd. Why can't you send anything else while resubscribes are happening, especially if they go into backoff?

I'll take a look more thoroughly on our code. Thanks. Closing for now

Issue #36353, opened 2024-11-04
[BUG] Attribute change cannot be reported for the data model provider which doesn't use ember


wqx6:
### Reproduction steps

For the attributes that are managed by the AttributeAccessInterface, when their changes are not caused by the WriteAttribute function in the data model provider, these changes cannot be reported if the Data Model provider doesn't use ember.

For instance, `LastNetworkingStatus` in `NetworkCommissioning` cluster will be changed after network connection changes but the `WriteAttribute` will not be called. `MatterReportingAttributeChangeCallback()` will be called instead. 

Should we provide an interface `ReportAttributeChanged` in data model provider and call that function in `MatterReportingAttributeChangeCallback()`?

### Bug prevalence

Always

### GitHub hash of the SDK that was being used

744f45cc886210ee95404ea659fd0327f818e891

### Platform

core

bzbarsky-apple:
@andy31415

andy31415:
The path build for this in `DataModel::Provider` is to use a https://github.com/project-chip/connectedhomeip/blob/master/src/app/data-model-provider/ProviderChangeListener.h#L42 (which is mapped by the InteractionModelEngine to report attribute changes). 

The way this gets passed is that a https://github.com/project-chip/connectedhomeip/blob/master/src/app/data-model-provider/Context.h#L32 contains a ProviderChangeListener and this context is passed to the Provider init and is accessible in GetCurrentContext https://github.com/project-chip/connectedhomeip/blob/master/src/app/data-model-provider/Provider.h#L60

The bindings however mean that the Provider is currently not piped through into the cluster implementations. Once could use `InteractionModelEngine::GetInstance().GetDataModelProvider()` however that is a bit coupled.


So for the long term, what we need is AAI to be aware of the context (passed in) or maybe define a "cluster implementation" that actually has all data and combines AAI and CHI because this is what people really want (see examples/tv-casting-app/tv-casting-common/core/BaseCluster.h having done this or https://github.com/project-chip/connectedhomeip/pull/33480 that wanted to propose a base class)

This is what I have on my medium TODO list:
  - want to have something that combines AAI + CHI + metadata and context access
  - have clusters implement/use that instead of what we have today
  
No short-term fix though, except maybe `InteractionModelEngine::GetInstance()->GetDataModelProvider()->CurrentContext().dataModelChangeListener->MarkDirty(attributePath)` (eek at how long that is!)

wqx6:
[...]

PR #36364, opened 2024-11-05
data-model-provider: Add an interface to report attribute changes which are not caused by WriteAttribute.

wqx6:
Short term fix for #36353


src/app/data-model-provider/Provider.h
```diff
@@ -105,6 +105,16 @@ class Provider : public ProviderMetadataTree
...
 
+    /// Report attribute changed
+    ///
+    /// When this is invoked, the caller is expected to increase the cluster data version and the attribute path
+    /// should be marked as dirty by the data model provider listener so that the reporter can report the attribute
+    /// changes to the subscriber.
+    ///
+    /// It should be invoked when attribute managed by access override is changed but the change is not caused by
+    /// WriteAttribute(Such as LastNetworkingStatus in NetworkCommissioning cluster).
+    virtual void ReportAttributeChanged(const AttributePathParams & path) = 0;
```
andy31415, line 116: This feels off: the data model provider is supposed to be the "owner of the data" including all implementations of WriteAttribute. The fact that it uses AttributeAccessInterface for example should be transparent. 

I do not think this is the right placement for a method as it seems to invert dependencies at a first glance.

andy31415 (changes requested):
I am generally weary of adding things to DataModelProvider.

In particular its design is such that it is the interface to query for data and it owns all the underlying data.

Please provide a use case and design description in the summary at least, so I understand it. Right now the data model intentionally did not provide version updates as it is supposed to externally notify changes by using its context, however it is not designed to get pushes to version. 

The underlying concept is that the `DataModel::Provider` owns its data and versions. So you cannot tell it "I changed your data" because you do not own it.

src/app/reporting/reporting.cpp
```diff
@@ -39,8 +40,8 @@ void MatterReportingAttributeChangeCallback(const ConcreteAttributePath & aPath)
     // applications notifying about changes from their end.
     assertChipStackLockedByCurrentThread();
 
-    emberAfAttributeChanged(aPath.mEndpointId, aPath.mClusterId, aPath.mAttributeId,
```
andy31415, line 42: These should indeed be updated out: libCHIP should not bind into ember. I think these changes independently are actually correct, however we have to pipe then through something else than `MatterRerportingAttributeChangeCallback` ... the callback assumes too many globals and data ownership differences.

andy31415: We could maybe mark this as temporary and have `TermpoaryReportAttributeChanged` instead of `ReportAttributeChanged` and a specific TODO to not use this anymore (please add a bug for this ... I can add some interfaces to AAI/CHI to allow context usage to report attribute changes).

src/app/data-model-provider/Provider.h
andy31415:
```suggestion
   [[deprecated("Workaround until all code can use a context object instead")]]
    virtual void Temporary_ReportAttributeChanged(const AttributePathParams & path) = 0;
```

With some documentation explaining this will be replaced.

wqx6: Added the documentation

src/app/reporting/reporting.cpp
wqx6: Added the bug https://github.com/project-chip/connectedhomeip/issues/36381

andy31415 (approved):

src/app/codegen-data-model-provider/CodegenDataModelProvider_Write.cpp
```diff
@@ -261,5 +261,18 @@ DataModel::ActionReturnStatus CodegenDataModelProvider::WriteAttribute(const Dat
...
+void CodegenDataModelProvider::Temporary_ReportAttributeChanged(const AttributePathParams & path)
+{
+    ContextAttributesChangeListener change_listener(CurrentContext());
+    if (path.mClusterId != kInvalidClusterId)
+    {
+        emberAfAttributeChanged(path.mEndpointId, path.mClusterId, path.mAttributeId, &change_listener);
+    }
+    else
+    {
+        change_listener.MarkDirty(path);
```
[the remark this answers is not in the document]
wqx6: Added the document

wqx6:
The deprecated mark will cause CI failures, as the lots of CI buildings are using -Werror=deprecated-declarations compile option. Removed the deprecated mark.

src/app/codegen-data-model-provider/CodegenDataModelProvider_Write.cpp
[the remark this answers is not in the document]
bzbarsky-apple: This comment doesn't actually answer the question.  Why can't we just `MarkDirty(path)` in the case when the cluster _is_ valid?  Presumably the answer has to do with data version maintenance or something, but that needs to be made clear, so someone doesn't "simplify" this to just `change_listener.MarkDirty(path)`.

Alternately, why is the `MarkDirty` case not calling `emberAfEndpointChanged`?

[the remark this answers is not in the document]
wqx6: Done, I am using the `emberAfEndpointChanged` now.

src/app/data-model-provider/Provider.h
```diff
@@ -105,6 +105,24 @@ class Provider : public ProviderMetadataTree
...
+    /// For example, if the LastNetworkingStatus attribute changes because the NetworkCommissioning driver detects a
+    /// network connection status change and calls SetLastNetworkingStatusValue(). The data model provider can recognize
+    /// this change by invoking this function at the point of change.
+    ///
+    /// This is a workaround function as we cannot notify the attribute change to the data model provider. The provider
+    /// should own its data and versions.
+    ///
+    /// TODO: We should remove this function when the AttributeAccessInterface/CommandHandlerInterface is able to report
+    /// the attribute changes.
+    virtual void Temporary_ReportAttributeChanged(const AttributePathParams & path) = 0;
```
andy31415, line 124: Should we maybe move this to `ProviderMetadataTree` instead?

My reasoning is that attribute changed seems more metadata-related (i.e. it increases cluster version). This would keep the provider being the action status on top of metadata.

wqx6: agreed, moved it to the ProviderMetadataTree.

andy31415:
Fast tracking: comments were addressed and PR is open for 1 month.

PR #36373, opened 2024-11-05
[Silabs]Update our refrigerator app zap configs.

jmartinez-silabs:
Zap config updates

- Enable Identify cluster on ep 1
- Set default Ref alarm mask attribute to 1. It should start at 1 and user action on the refrigerator can disable it
- Since we already have 2 distinct zap files for Thread and Wifi platforms for this app, make the network commissioning
cluster spec-compliant : Enable SupportedWifiBands or ThreadVersion and SupportedThreadFeatures respectively of the network type.
- Remove Diagnostic log. We do not support it

### DIFF
--- a/examples/refrigerator-app/silabs/data_model/refrigerator-thread-app.matter
+++ b/examples/refrigerator-app/silabs/data_model/refrigerator-thread-app.matter
@@ -241,6 +241,56 @@ struct AtomicAttributeStatusStruct {
   status statusCode = 1;
 }
 
+/** Attributes and commands for putting a device into Identification mode (e.g. flashing a light). */
+cluster Identify = 3 {
+  revision 4;
+
+  enum EffectIdentifierEnum : enum8 {
+    kBlink = 0;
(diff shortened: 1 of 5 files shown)


mkardous-silabs (approved):

andy31415 (dismissed):

examples/refrigerator-app/silabs/data_model/refrigerator-thread-app.matter
```diff
@@ -1641,6 +1644,19 @@ endpoint 0 {
   device type ma_rootdevice = 22, version 1;
 
 
+  server cluster Identify {
```
andy31415, line 1647: Linter caught this: Identify is not allowed on endpoint 0

jmartinez-silabs: Oh, I wasn't aware of that. I'll move it to ep1. Thanks

andy31415 (approved):

PR #36386, opened 2024-11-06
Migrate from deprecated OpenSSL SHA256_* APIs to EVP APIs

Alami-Amine:
Migrating from deprecated OpenSSL SHA256_* APIs to the higher-level EVP APIs:

- The deprecation impacts the APIs: `SHA256_Init` , `SHA256_Update` and `SHA256_Final`, details [here](https://docs.openssl.org/master/man3/SHA256_Init/#synopsis)
- The Recommended replacement APIs seem to be present since at least OpenSSL 0.9.7, --> backward compatibility isn't an issue.
- These APIs work for both OpenSSL and BoringSSL.

 **NOTE**: The Digest Context `EVP_MD_CTX`  with these new APIs is Opaque, and thus `HashSHA256OpaqueContext` was used to store the pointer to SHA256 Context (`EVP_MD_CTX`) instead of the structure itself (as was originally intended for `HashSHA256OpaqueContext`).

## Testing 
- Unit Tests and Fuzz Tests (with OpenSSL and BoringSSL)
- CI Testing

### DIFF
--- a/src/crypto/CHIPCryptoPALOpenSSL.cpp
+++ b/src/crypto/CHIPCryptoPALOpenSSL.cpp
@@ -433,36 +433,58 @@ CHIP_ERROR Hash_SHA1(const uint8_t * data, const size_t data_length, uint8_t * o
     return CHIP_NO_ERROR;
 }
 
-Hash_SHA256_stream::Hash_SHA256_stream() {}
+// For OpenSSL, we store a pointer to the digest context (EVP_MD_CTX) since EVP_MD_CTX is Opaque.
+static_assert(kMAX_Hash_SHA256_Context_Size >= sizeof(void *),
+              "kMAX_Hash_SHA256_Context_Size needs to at least be able to store a pointer");
+
+// Storing a pointer to EVP_MD_CTX in HashSHA256OpaqueContext instead of the actual EVP_MD_CTX structure, as EVP_MD_CTX was made
+// opaque by OpenSSL and is dynamically allocated.
+static inline void set_inner_hash_evp_md_ctx(HashSHA256OpaqueContext * context, EVP_MD_CTX * evp_ctx)
+{
+    *SafePointerCast<EVP_MD_CTX **>(context) = evp_ctx;
+}
+
+static inline EVP_MD_CTX * to_inner_hash_evp_md_ctx(HashSHA256OpaqueContext * context)
+{
+    return *SafePointerCast<EVP_MD_CTX **>(context);
+}
+
+Hash_SHA256_stream::Hash_SHA256_stream()
+{
+    set_inner_hash_evp_md_ctx(&mContext, nullptr);
+    mInitialized = false;
+}
 
 Hash_SHA256_stream::~Hash_SHA256_stream()
 {
     Clear();
 }
 
-static_assert(kMAX_Hash_SHA256_Context_Size >= sizeof(SHA256_CTX),
-              "kMAX_Hash_SHA256_Context_Size is too small for the size of underlying SHA256_CTX");
-
-static inline SHA256_CTX * to_inner_hash_sha256_context(HashSHA256OpaqueContext * context)
-{
-    return SafePointerCast<SHA256_CTX *>(context);
-}
-
 CHIP_ERROR Hash_SHA256_stream::Begin()
 {
-    SHA256_CTX * const context = to_inner_hash_sha256_context(&mContext);
 
-    const int result = SHA256_Init(context);
+    EVP_MD_CTX * mdctx = EVP_MD_CTX_new();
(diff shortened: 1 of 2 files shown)


src/crypto/CHIPCryptoPALOpenSSL.cpp
```diff
@@ -433,71 +433,116 @@ CHIP_ERROR Hash_SHA1(const uint8_t * data, const size_t data_length, uint8_t * o
...
+    return *SafePointerCast<EVP_MD_CTX **>(context);
+}
 
-static inline SHA256_CTX * to_inner_hash_sha256_context(HashSHA256OpaqueContext * context)
+Hash_SHA256_stream::Hash_SHA256_stream()
 {
-    return SafePointerCast<SHA256_CTX *>(context);
+    set_inner_hash_evp_md_ctx(&mContext, nullptr);
+    mInitialized = false;
+}
```
andy31415, line 456:
```suggestion
}
```

src/crypto/CHIPCryptoPALOpenSSL.cpp
```diff
@@ -433,71 +433,116 @@ CHIP_ERROR Hash_SHA1(const uint8_t * data, const size_t data_length, uint8_t * o
...
 
-static_assert(kMAX_Hash_SHA256_Context_Size >= sizeof(SHA256_CTX),
-              "kMAX_Hash_SHA256_Context_Size is too small for the size of underlying SHA256_CTX");
+static inline EVP_MD_CTX * to_inner_hash_evp_md_ctx(HashSHA256OpaqueContext * context)
+{
+    return *SafePointerCast<EVP_MD_CTX **>(context);
+}
 
-static inline SHA256_CTX * to_inner_hash_sha256_context(HashSHA256OpaqueContext * context)
+Hash_SHA256_stream::Hash_SHA256_stream()
```
andy31415, line 452:
```suggestion
Hash_SHA256_stream::Hash_SHA256_stream() : mInitialized(false)
```

src/crypto/CHIPCryptoPALOpenSSL.cpp
```diff
@@ -433,71 +433,116 @@ CHIP_ERROR Hash_SHA1(const uint8_t * data, const size_t data_length, uint8_t * o
...
 }
 
 void Hash_SHA256_stream::Clear()
 {
+    EVP_MD_CTX * mdctx = to_inner_hash_evp_md_ctx(&mContext);
+    if (mdctx != nullptr)
+    {
+        EVP_MD_CTX_free(mdctx);
+        set_inner_hash_evp_md_ctx(&mContext, nullptr);
+        mInitialized = false;
```
andy31415, line 544: Should this be not unconditional regardless of if we have a context or not as clear resets things?

src/crypto/CHIPCryptoPALOpenSSL.cpp
```diff
@@ -433,71 +433,116 @@ CHIP_ERROR Hash_SHA1(const uint8_t * data, const size_t data_length, uint8_t * o
...
+static_assert(kMAX_Hash_SHA256_Context_Size >= sizeof(void *),
+              "kMAX_Hash_SHA256_Context_Size needs to at least be able to store a pointer");
 
-Hash_SHA256_stream::~Hash_SHA256_stream()
+// Storing a pointer to EVP_MD_CTX in HashSHA256OpaqueContext instead of the actual EVP_MD_CTX structure, as EVP_MD_CTX was made
+// opaque by OpenSSL and is dynamically allocated.
+static inline void set_inner_hash_evp_md_ctx(HashSHA256OpaqueContext * context, EVP_MD_CTX * evp_ctx)
 {
-    Clear();
+    *SafePointerCast<EVP_MD_CTX **>(context) = evp_ctx;
```
andy31415, line 444: Is context guaranteed to be properly aligned for this operation?

andy31415: Consider adding a comment that HashSHA256OpaqueContext is aligned at HashSHA256OpaqueContext which is size_t or `sizeof(psa_hash_operation_t)` ... although I am not sure what the latter is.

Should we `static_assert(CHIP_CONFIG_SHA256_CONTEXT_ALIGN >= sizeof(EVP_MD_CTX *))` just in case (not sure if that is fully sane, but it seems reasonable)

andy31415: Ah, SafePointerCast already checks alignment. Never mind then.

andy31415 (approved):

andy31415 (approved):

jmartinez-silabs (approved):

PR #36387, opened 2024-11-06
Add LWIP IPv6 zone index in chip::Inet::IPAddress

db-shelly:
If using LWIP without any patches and with LWIP_IPV6_SCOPES=1, outbound connections cannot be routed to the correct interface, because zone index is missing in chip::Inet::IPAddress.

This is a simple fix to keep the ip6_addr_t zone in IPAddress if needed.

### DIFF
--- a/src/inet/IPAddress.h
+++ b/src/inet/IPAddress.h
@@ -201,6 +201,9 @@ public:
      *  address in network byte order.
      */
     uint32_t Addr[4];
+#if CHIP_SYSTEM_CONFIG_USE_LWIP && LWIP_IPV6_SCOPES
+    uint8_t Zone;
+#endif
 
     /**
      * @brief   Test whether address is IPv6 compatible.
--- a/src/inet/IPAddress.h
+++ b/src/inet/IPAddress.h
@@ -202,7 +202,7 @@ public:
      */
     uint32_t Addr[4];
 #if CHIP_SYSTEM_CONFIG_USE_LWIP && LWIP_IPV6_SCOPES
-    uint8_t Zone;
+    uint8_t Zone = IP6_NO_ZONE;
 #endif
 
     /**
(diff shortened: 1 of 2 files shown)


src/inet/IPAddress.h
```diff
@@ -201,6 +201,9 @@ class DLL_EXPORT IPAddress
      *  address in network byte order.
      */
     uint32_t Addr[4];
+#if CHIP_SYSTEM_CONFIG_USE_LWIP && LWIP_IPV6_SCOPES
+    uint8_t Zone;
```
andy31415, line 205: should this default to `= IP6_NO_ZONE` ? Or are we always setting it to something? 

Generally how about other constructors:
  - The default constructor
  - IPV4 constructor
  - the otIp6Address (is that even a thing? would it be supported with LWIP and scopes enabled?)

db-shelly: I think `Zone` should be set to something whenever `Addr` is set to something. Defaulting to IP6_NO_ZONE won't hurt though.

IPv4 doesn't need zones and I think OT has nothing to do with LWIP routing. Correct me if I am wrong.

andy31415:
@db-shelly please sign the CLA

andy31415 (approved):

andy31415:
It seems that we cannot set an assigment, we should modify all constructors instead....

```
INFO    ../../examples/light-switch-app/asr/third_party/connectedhomeip/src/inet/IPAddress.h:714:43: error: static assertion failed: IPAddress is not trivial
INFO      714 | static_assert(std::is_trivial<IPAddress>::value, "IPAddress is not trivial");
```

src/inet/IPAddress.h
```diff
@@ -201,6 +201,9 @@ class DLL_EXPORT IPAddress
      *  address in network byte order.
      */
     uint32_t Addr[4];
+#if CHIP_SYSTEM_CONFIG_USE_LWIP && LWIP_IPV6_SCOPES
+    uint8_t Zone = IP6_NO_ZONE;
```
bzbarsky-apple, line 205: Does LwIP use zones for non-link-local bits?

Because for link-local, PeerAddress is supposed to have an interface ID already, which is what the zone bits are, no?  Is that not being handled correctly in our LwIP integration?  Because if that's what's going on, the right fix is to fix it there, not to add things to IPAddress that are also duplicated-but-unused-it's-a-trap in PeerAddress.

PR #36391, opened 2024-11-06
Add a helper function to allow ErrorStr not using static char array

yyzhong-g:
Provide another ErrorStr function that enable user to specify the buffer they writes to. 

Using a static char array for writing the error str causes the data race when there is more than one thread.


src/lib/core/ErrorStr.h
```diff
@@ -49,6 +49,7 @@ struct ErrorFormatter
 };
 
 extern const char * ErrorStr(CHIP_ERROR err, bool withSourceLocation = true);
+extern void ErrorStr(CHIP_ERROR err, bool withSourceLocation, char * buf, uint16_t bufSize);
```
arkq, line 52: How about using `MutableByteSpan` instead of pointer and size? Also for convenience, maybe this function should return `const char *` just like the other one?

src/lib/core/ErrorStr.cpp
```diff
@@ -50,8 +50,25 @@ static ErrorFormatter * sErrorFormatterList = nullptr;
...
+ * which describes the provided error.
+ *
+ * @param[in] err                      The error for format and describe.
+ * @param[in] withSourceLocation       Whether or not to include the source
+ * @param[in] buf                      Provided buf to write into
+ * @param[in] bufSize                  Size of the buf
+ * location in the output string. Only used if CHIP_CONFIG_ERROR_SOURCE &&
+ * !CHIP_CONFIG_SHORT_ERROR_STR. Defaults to true.
+ */
+DLL_EXPORT void ErrorStr(CHIP_ERROR err, bool withSourceLocation, char * buf, uint16_t bufSize)
```
andy31415, line 68: Use a span for ptr + size

src/lib/core/ErrorStr.cpp
```diff
@@ -50,8 +50,25 @@ static ErrorFormatter * sErrorFormatterList = nullptr;
...
+{
+    char * formattedError   = buf;
+    uint16_t formattedSpace = sizeof(buf);
```
andy31415, line 71: this seems to be sizeof(char *) ... is this right?

src/lib/core/ErrorStr.cpp
andy31415: Maybe even consider something like a dedicated storage type like https://github.com/project-chip/connectedhomeip/blob/master/src/app/data-model-provider/ActionReturnStatus.h#L100

andy31415 (dismissed):
Please add unit tests as well

src/lib/core/ErrorStr.cpp
yyzhong-g: Using a span in ErrorStr.cpp seems inevitably introducing circular dependancy.
Span uses CodeUtils.h and CodeUtils.h uses Errors which I think is pretty hard to break up.

src/lib/core/ErrorStr.cpp
yyzhong-g: Oops, you are right.

src/lib/core/ErrorStr.h
yyzhong-g: Will change it to return const char*, but using a span here creates circualr dependancy.

src/lib/core/ErrorStr.cpp
yyzhong-g: Ah, didn't see your comment about the storage type, will try to see if that helps.

src/lib/core/ErrorStr.h
```diff
@@ -48,7 +48,15 @@ struct ErrorFormatter
     ErrorFormatter * Next;
 };
 
+struct ErrorStrStorage
+{
+    uint16_t Size() { return sizeof(buff); }
```
andy31415, line 53:
```suggestion
    constexpr static uint16_t Size() { return sizeof(buff); }
```

or do we really want this as a method? Why is it not a `constexpr uint16_t kBufferSize = sizeof(buff)` ?

andy31415 (approved):

arkq (approved):

PR #36406, opened 2024-11-06
[Fan Control]Circular callback fix

lpbeliveau-silabs:
Proposed fix to circumvent the circular callback when a percentage change happens with a max speed value below 100.

Basically, currently, with a max speed of 10, what happens is:
Attribute write PercentSetting 25
PercentSetting Change to 25 Callback
SpeedSetting Change 3 Callback
PercentSetting Change 30 Callback.

Added AttributeAccessInterface and used it in All-cluster app.

### DIFF
--- a/src/app/clusters/fan-control-server/fan-control-server.cpp
+++ b/src/app/clusters/fan-control-server/fan-control-server.cpp
@@ -24,6 +24,7 @@
 #include <app-common/zap-generated/cluster-objects.h>
 #include <app-common/zap-generated/ids/Attributes.h>
 #include <app-common/zap-generated/ids/Clusters.h>
+#include <app/AttributeAccessInterfaceRegistry.h>
 #include <app/CommandHandler.h>
 #include <app/ConcreteCommandPath.h>
 #include <app/clusters/fan-control-server/fan-control-server.h>
@@ -389,7 +390,7 @@ void MatterFanControlClusterServerAttributeChangedCallback(const app::ConcreteAt
             // Plus 99 then integer divide by 100 instead of multiplying 0.01 to avoid floating point precision error
             uint8_t speedSetting = static_cast<uint8_t>((speedMax * percent + 99) / 100);
 
-            if (currentSpeedSetting.IsNull() || speedSetting != currentSpeedSetting.Value())
+            if (currentSpeedSetting != speedSetting)
             {
                 status = SpeedSetting::Set(attributePath.mEndpointId, speedSetting);
                 VerifyOrReturn(Status::Success == status,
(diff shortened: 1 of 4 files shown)


jmartinez-silabs (approved):

src/app/clusters/fan-control-server/fan-control-server.cpp
```diff
@@ -413,33 +414,75 @@ void MatterFanControlClusterServerAttributeChangedCallback(const app::ConcreteAt
...
+                ReturnLogErrorOnFailure(StatusIB(status).ToChipError());
+
+                DataModel::Nullable<Percent> currentPercentSetting;
+                status = PercentSetting::Get(aPath.mEndpointId, currentPercentSetting);
+                ReturnLogErrorOnFailure(StatusIB(status).ToChipError());
+
+                float speed            = speedSetting.Value();
+                Percent percentSetting = static_cast<Percent>(speed / speedMax * 100);
+
+                if (currentPercentSetting.IsNull() || percentSetting != currentPercentSetting.Value())
```
andy31415, line 473: Optional and nullable have an operator== ... could that be used?

https://github.com/project-chip/connectedhomeip/blob/master/src/app/data-model/Nullable.h#L111

andy31415 (approved):

rcasallas-silabs (approved):

src/app/clusters/fan-control-server/fan-control-server.cpp
```diff
@@ -413,33 +414,75 @@ void MatterFanControlClusterServerAttributeChangedCallback(const app::ConcreteAt
...
             }
 
-            // Adjust PercentSetting from a speed value change for SpeedSetting
-            // percent = floor( speed/SpeedMax * 100 )
-            uint8_t speedMax;
-            status = SpeedMax::Get(attributePath.mEndpointId, &speedMax);
-            VerifyOrReturn(Status::Success == status,
-                           ChipLogError(Zcl, "Failed to get SpeedMax with error: 0x%02x", to_underlying(status)));
+            // Adjust PercentSetting from a speed value change for SpeedSetting only when the SpeedSetting change was received
+            // on a write command, not when it was changed by the server or the app logic. This avoids circular logic such as
```
bzbarsky-apple, line 418: There is no "write command" in Matter.  That's Zigbee terminology.  Please fix this comment to make sense for Matter.

src/app/clusters/fan-control-server/fan-control-server.cpp
```diff
@@ -413,33 +414,75 @@ void MatterFanControlClusterServerAttributeChangedCallback(const app::ConcreteAt
...
+            // Skip the last step if we are writing NULL
+            VerifyOrReturnValue(!speedSetting.IsNull(), CHIP_NO_ERROR);
 
-            if (currentPercentSetting.IsNull() || percentSetting != currentPercentSetting.Value())
+            if (SupportsMultiSpeed(aPath.mEndpointId))
             {
-                status = PercentSetting::Set(attributePath.mEndpointId, percentSetting);
-                VerifyOrReturn(Status::Success == status,
-                               ChipLogError(Zcl, "Failed to set PercentSetting with error: 0x%02x", to_underlying(status)));
+                // If SpeedSetting is set to 0, the server SHALL set the FanMode attribute value to Off.
```
bzbarsky-apple, line 453: So why are we doing that both in here and in MatterFanControlClusterServerAttributeChangedCallback?  That seems pretty odd to me.

src/app/clusters/fan-control-server/fan-control-server.cpp
```diff
@@ -413,33 +414,75 @@ void MatterFanControlClusterServerAttributeChangedCallback(const app::ConcreteAt
...
+    }
+}
 
-            DataModel::Nullable<Percent> currentPercentSetting;
-            status = PercentSetting::Get(attributePath.mEndpointId, currentPercentSetting);
-            VerifyOrReturn(Status::Success == status,
-                           ChipLogError(Zcl, "Failed to get PercentSetting with error: 0x%02x", to_underlying(status)));
+CHIP_ERROR FanControlAttributeAccessInterface::Write(const ConcreteDataAttributePath & aPath, AttributeValueDecoder & aDecoder)
+{
+    Status status = Status::Success;
```
bzbarsky-apple, line 433: Why is this declared here, not where it's actually used?

src/app/clusters/fan-control-server/fan-control-server.h
```diff
@@ -19,13 +19,23 @@
...
 #include <app-common/zap-generated/cluster-objects.h>
+#include <app/AttributeAccessInterface.h>
 #include <app/util/af-types.h>
 
 namespace chip {
 namespace app {
 namespace Clusters {
 namespace FanControl {
 
+class FanControlAttributeAccessInterface : public AttributeAccessInterface
```
bzbarsky-apple, line 30: This needs documentation.  The fact that all users of this cluster now need to inherit from this thing, and will get non-spec-compliant behavior unless they do (which is an API break, by the way), is _extremely_ non-obvious.

What should have happened here is that if we need an AttributeAccessInterface registered for proper cluster functionality then the _cluster_ needs to register it.  And if we have things the app needs to handle, the cluster should be calling out into the app for those things, via the relevant delegate interface, instead of the app implementing AttributeAccessInterface directly.

src/app/clusters/fan-control-server/fan-control-server.cpp
```diff
@@ -24,6 +24,7 @@
 #include <app-common/zap-generated/cluster-objects.h>
 #include <app-common/zap-generated/ids/Attributes.h>
 #include <app-common/zap-generated/ids/Clusters.h>
+#include <app/AttributeAccessInterfaceRegistry.h>
```
bzbarsky-apple, line 27: Is this used anywhere in this file?

PR #36407, opened 2024-11-06
Attempt to fix cirque by replacing OTBR agent startup string that we look for.

andy31415:
This is after https://github.com/openthread/ot-br-posix/pull/2574 removed the log line we were looking for.


yunhanw-google (approved):

andy31415:
Hotfixing this in as I don't think it makes this worse... github CI is backlogged and github support has not yet responded to my open ticket.

PR #36413, opened 2024-11-07
[Silabs][WiFi] Wifi scan fixes and additional advertising fix

chirag-silabs:
WiFi scan was broken due to the refactor and failing with the ecosystem commissioning and normal scan commands. 
Additional advertising build was also failing due to incorrect signature of the API. 

Fixing the scan to copy the correct ssid and fixing the additional advertising API. Enabling the additional advertising for one of the NCP so the build failure can be captured in the CI

Tested locally

### DIFF
--- a/examples/platform/silabs/SiWx917/SiWxWifiInterface.cpp
+++ b/examples/platform/silabs/SiWx917/SiWxWifiInterface.cpp
@@ -328,9 +328,8 @@ sl_status_t InitiateScan()
     sl_wifi_scan_configuration_t wifi_scan_configuration = default_wifi_scan_configuration;
 
     ssid.length = wfx_rsi.sec.ssid_length;
-
-    // TODO: workaround because the string management with the null termination is flawed
-    chip::Platform::CopyString((char *) &ssid.value[0], ssid.length + 1, wfx_rsi.sec.ssid);
+    ssid.length = std::min<size_t>(wfx_rsi.sec.ssid_length, sizeof(ssid_arg.value) - 1);
+    chip::Platform::CopyString((char *) &ssid.value[0], ssid.length + 1, wfx_rsi.sec.ssid); // +1 for null termination
     sl_wifi_set_scan_callback(ScanCallback, NULL);
 
     osSemaphoreAcquire(sScanInProgressSemaphore, osWaitForever);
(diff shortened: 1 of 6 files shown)


examples/platform/silabs/SiWx917/SiWxWifiInterface.cpp
```diff
@@ -328,9 +328,8 @@ sl_status_t InitiateScan()
     sl_wifi_scan_configuration_t wifi_scan_configuration = default_wifi_scan_configuration;
 
     ssid.length = wfx_rsi.sec.ssid_length;
-
-    // TODO: workaround because the string management with the null termination is flawed
-    chip::Platform::CopyString((char *) &ssid.value[0], ssid.length + 1, wfx_rsi.sec.ssid);
+    ssid.length = std::min<size_t>(wfx_rsi.sec.ssid_length, sizeof(ssid_arg.value) - 1);
```
jmartinez-silabs, line 331:
```suggestion
    ssid.length = std::min<size_t>(wfx_rsi.sec.ssid_length, sizeof(ssid.value) - 1);
```

jmartinez-silabs (approved):

examples/platform/silabs/SiWx917/SiWxWifiInterface.cpp
```diff
@@ -328,9 +328,8 @@ sl_status_t InitiateScan()
     sl_wifi_scan_configuration_t wifi_scan_configuration = default_wifi_scan_configuration;
 
     ssid.length = wfx_rsi.sec.ssid_length;
-
-    // TODO: workaround because the string management with the null termination is flawed
-    chip::Platform::CopyString((char *) &ssid.value[0], ssid.length + 1, wfx_rsi.sec.ssid);
+    ssid.length = std::min<size_t>(wfx_rsi.sec.ssid_length, sizeof(ssid.value) - 1);
```
bzbarsky-apple, line 331: Why are we setting `ssid.length` two lines in a row?  If the second line had used the value from the first, that would make sense, but it does not.

chirag-silabs: Ohh didn't noticed that the assign was added not modified. Removed the first assignment.

examples/platform/silabs/SiWx917/SiWxWifiInterface.cpp
```diff
@@ -609,11 +607,13 @@ sl_status_t show_scan_results(sl_wifi_scan_result_t * scan_result)
 
         cur_scan_result.ssid_length = strnlen((char *) scan_result->scan_info[idx].ssid,
                                               std::min<size_t>(sizeof(scan_result->scan_info[idx].ssid), WFX_MAX_SSID_LENGTH));
-        chip::Platform::CopyString(cur_scan_result.ssid, cur_scan_result.ssid_length, (char *) scan_result->scan_info[idx].ssid);
+        // cur_scan_result.ssid is of size WFX_MAX_SSID_LENGTH+1, we are safe with the cur_scan_result.ssid_length calculated above
+        chip::Platform::CopyString(cur_scan_result.ssid, cur_scan_result.ssid_length + 1,
```
andy31415, line 611: why is the cast to char * required here? Should this not be const char *? Also same question for line 331 above: why are we casting (and using a C-style cast instead of C++)?

Also wondering about copy-string usage: this copies until the first null terminator, however the syntax on scan_results seem to be length-delimited instead. It seems we would rather want some helpers that use ByteSpan or similar.

chirag-silabs:
PR not needed any more hence closing this.

PR #36415, opened 2024-11-07
Ensure SYSTEM_STATS_INCREMENT only on successful memory allocation in SystemPacketBuffer.cpp

BoB13-Matter:
### Description

- Fixes an issue in `SystemPacketBuffer.cpp` where `SYSTEM_STATS_INCREMENT` is called even when memory allocation fails, leading to incorrect packet buffer statistics.

- Moved `SYSTEM_STATS_INCREMENT` to occur only after verifying that `MemoryAlloc` did not return `nullptr`, ensuring `kSystemLayer_NumPacketBufs` is incremented only when allocation is successful.

### Changes

- Adjusted the position of `SYSTEM_STATS_INCREMENT(chip::System::Stats::kSystemLayer_NumPacketBufs)` in `SystemPacketBuffer.cpp` to be called after `nullptr` validation.

### DIFF
--- a/src/system/SystemPacketBuffer.cpp
+++ b/src/system/SystemPacketBuffer.cpp
@@ -620,7 +620,6 @@ PacketBufferHandle PacketBufferHandle::New(size_t aAvailableSize, uint16_t aRese
     // checked to fit in a size_t.
     const size_t lBlockSize = static_cast<size_t>(sumOfSizes);
     lPacket                 = reinterpret_cast<PacketBuffer *>(chip::Platform::MemoryAlloc(lBlockSize));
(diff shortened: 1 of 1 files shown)


arkq (approved):

andy31415 (approved):

kkasperczyk-no:
@BoB13-Matter I've spotted that this change introduced regression regarding `stat peak` shell command usage (and it's using HighWatermarks that are probably incorrect). Before this commit it was actually presenting the peak usage of packet buffers, however after your change it's like a sum of packet buffers used from the start of an application.

For example, I have 11 packet buffers pool and after commissioning the `stats peak` shows `Packet Buffers: 71` (before it was 4). After several additional reads packet buffers are `91`, which is way bigger than the max pool size, so the value is invalid.

PR #36416, opened 2024-11-07
Move more of the MTRAttributeValueWaiter logic into the waiter.

bzbarsky-apple:
This logic should not live in MTRDevice.

### DIFF
--- a/src/darwin/Framework/CHIP/MTRAttributeValueWaiter.mm
+++ b/src/darwin/Framework/CHIP/MTRAttributeValueWaiter.mm
@@ -79,10 +79,10 @@ MTR_DIRECT_MEMBERS
 - (void)cancel
 {
     [self.device _forgetAttributeWaiter:self];
-    [self _cancelWithoutRemovingFromDevice];
+    [self _notifyCancellation];
 }
 
-- (void)_cancelWithoutRemovingFromDevice
+- (void)_notifyCancellation
 {
     [self _notifyWithError:[MTRError errorForCHIPErrorCode:CHIP_ERROR_CANCELLED]];
 }
--- a/src/darwin/Framework/CHIP/MTRAttributeValueWaiter.mm
+++ b/src/darwin/Framework/CHIP/MTRAttributeValueWaiter.mm
@@ -15,12 +15,15 @@
  */
 
 #import <Foundation/Foundation.h>
+#import <os/lock.h>
 
 #import <Matter/MTRError.h>
 
 #import "MTRAttributeValueWaiter_Internal.h"
 #import "MTRDevice_Internal.h"
+#import "MTRError_Internal.h"
 #import "MTRLogging_Internal.h"
+#import "MTRUnfairLock.h"
 
 @implementation MTRAwaitedAttributeState
 - (instancetype)initWithValue:(MTRDeviceDataValueDictionary)value
@@ -40,10 +43,14 @@ MTR_DIRECT_MEMBERS
 // Protected by the MTRDevice's lock.
 @property (nonatomic, readwrite, retain) dispatch_queue_t queue;
 @property (nonatomic, readwrite, copy, nullable) MTRStatusCompletion completion;
+@property (nonatomic, retain, readwrite, nullable) dispatch_source_t expirationTimer;
 @property (nonatomic, readonly, retain) MTRDevice * device;
 @end
 
-@implementation MTRAttributeValueWaiter
+@implementation MTRAttributeValueWaiter {
+    // Protects queue/completion and expirationTimer.
+    os_unfair_lock _lock;
+}
 
 - (instancetype)initWithDevice:(MTRDevice *)device values:(NSDictionary<MTRAttributePath *, MTRDeviceDataValueDictionary> *)values queue:(dispatch_queue_t)queue completion:(MTRStatusCompletion)completion
 {
@@ -58,6 +65,7 @@ MTR_DIRECT_MEMBERS
         _completion = completion;
         _device = device;
         _UUID = [NSUUID UUID];
+        _lock = OS_UNFAIR_LOCK_INIT;
     }
 
     return self;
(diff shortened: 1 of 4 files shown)


src/darwin/Framework/CHIP/MTRAttributeValueWaiter.mm
```diff
@@ -120,15 +137,45 @@ - (void)_notifyWithError:(NSError * _Nullable)error
...
-    auto completion = self.completion;
-    auto queue = self.queue;
-    self.completion = nil;
-    self.queue = nil;
+    auto * device = self.device;
     dispatch_async(queue, ^{
+        // Call into the device from our async dispatch, to avoid problems when
+        // _notifyWithError is happening while the device lock is held.
+        //
+        // But this seems to trigger TSAN errors about null-derefs.  Probably
```
ksperling-apple, line 145: Yeah you can't dispatch away in code that's called from `dealloc`. Sadly the runtime doesn't catch it when dealloc has attempted to "revive" the object and you just end up with UAFs.

ksperling-apple: I think the best thing to do here is to not call `_forget...` from `_notifyWithError`. Instead leave it up to the caller.
- When MTRDevice calls `_notifyWithError`, remove the waiter explicitly first
- When MTRDevice cancels all waiters, make it call a new `_notifyOfCancellation` method on the waiter that just calls `_notifyWithError` but does not call back into MTRDevice
- Only `cancel` (or dealloc) called directly on the waiter needs to call `_forgetWaiter` on the device along with calling `_notifyOfCancellation`.

There's also the option to rely the weak NSHashTable to do some of its own cleanup. For example from `dealloc` we could simply call `_notifyOfCancellation` because we know there's nobody else holding the waiter alive, so the slot in the hash table is reclaimable the next time it's needed without doing explicit work.

For explicit `cancel` and for successful notifications from the MTRDevice it probably makes sense to explicitly forget the waiter; that way the table won't end up growing wildly even if the client is for some reason (bug?) hanging on to waiters that have already notified.

src/darwin/Framework/CHIP/MTRAttributeValueWaiter.mm
```diff
@@ -120,15 +137,45 @@ - (void)_notifyWithError:(NSError * _Nullable)error
...
+        // when this code is called from dealloc, in which case self cannot
+        // really be captured, since it's dying.
+        [device _forgetAttributeWaiter:self];
         completion(error);
     });
 }
 
+- (void)_startTimer:(NSTimeInterval)timeout withQueue:(dispatch_queue_t)queue
+{
+    dispatch_source_t timerSource = dispatch_source_create(DISPATCH_SOURCE_TYPE_TIMER, 0, 0, queue);
```
ksperling-apple, line 155: Using `self.device.queue` may be more obvious than passing in the queue? Could also consider doing this directly in `init`?

bzbarsky-apple: Doing it in init would in some cases start and immediately stop the timer because our values are in fact satisfied.  So I'd rather not start in init.

`self.device.queue` I can do.

src/darwin/Framework/CHIP/MTRAttributeValueWaiter.mm
```diff
@@ -70,7 +78,13 @@ - (void)dealloc
 
 - (void)cancel
 {
-    [self.device _attributeWaitCanceled:self];
+    [self.device _forgetAttributeWaiter:self];
+    [self _cancelWithoutRemovingFromDevice];
+}
+
+- (void)_cancelWithoutRemovingFromDevice
```
ksperling-apple, line 85: Nit: Could call this _notifyCancellation since that's really all it does.

src/darwin/Framework/CHIP/MTRAttributeValueWaiter.mm
```diff
@@ -120,15 +143,41 @@ - (void)_notifyWithError:(NSError * _Nullable)error
...
+    dispatch_source_set_event_handler(timerSource, ^{
+        dispatch_source_cancel(timerSource);
+        mtr_strongify(self);
+        if (self != nil) {
+            [self.device _forgetAttributeWaiter:self];
+            [self _notifyWithError:[MTRError errorForCHIPErrorCode:CHIP_ERROR_TIMEOUT]];
+        }
+    });
+
+    {
```
ksperling-apple, line 173: Locking here isn't even strictly necessary right? The timer can't kick in until it's resumed, and the waiter hasn't been returned to the client yet.

src/darwin/Framework/CHIP/MTRDevice.mm
```diff
@@ -764,37 +763,25 @@ - (MTRAttributeValueWaiter *)waitForAttributeValues:(NSDictionary<MTRAttributePa
...
+        // no need to remove it before notifying.
         [attributeWaiter _notifyWithError:nil];
         return attributeWaiter;
     }
 
-    // Otherwise, wait for the values to arrive or our timeout.
-    if (!self.attributeValueWaiters) {
-        self.attributeValueWaiters = [NSHashTable weakObjectsHashTable];
+    // Otherwise, wait for one of our termination conditions.  It's important to
+    // not take the lock earlier, because the _notifyWithError call above would
```
ksperling-apple, line 773: This sounds outdated now since `_notifyWithError` doesn't go via the device lock anymore?

ksperling-apple (approved):

src/darwin/Framework/CHIP/MTRAttributeValueWaiter.mm
bzbarsky-apple: Good idea, done.

src/darwin/Framework/CHIP/MTRAttributeValueWaiter.mm
bzbarsky-apple: True, but that global state is non-obvious from here.  So it's safer to just do our invariants, which is locking around the relevant field.

src/darwin/Framework/CHIP/MTRDevice.mm
bzbarsky-apple: Good catch, fixed.

kiel-apple (approved):

bzbarsky-apple:

PR #36419, opened 2024-11-07
Allow to easy install CHIP wheels in current virtual env

arkq:
### Problem

It's not easy to build and install all CHIP packages in the current python virtual environment. Current use case:
```sh
./scripts/build_python.sh
# find all wheels generated in the output directory (where is the output, though)
# uninstall them, and install again...
```

In order to simplify that, the `build_python.sh` now provides option to automatically install all generated packages without the need of creating new virtual env (which is time costly since new env is re-created by default), i.e.:
```sh
source ./scripts/activate.sh
./scripts/build_python.sh -i
# do some modifications in the code
./scripts/build_python.sh -i
# ... repeat ...
```

### Changes

- added `-i`/`--install-wheels` option to install generated packages
- renamed `-i`/`--create_virtual_env` to  `-e`/`--create-virtual-env` (the `-e` better align with "environment", though, this change breaks backward compatibility)
- renamed some long options to use "-" instead of "_"

### Testing

CI will verify

### DIFF
--- a/scripts/build_python.sh
+++ b/scripts/build_python.sh
@@ -35,7 +35,7 @@ echo_bold_white() {
 }
 
 check_one_of() {
-    local value=$1
+    local value="$1"
     shift
     for v in "$@"; do
         if [ "$value" = "$v" ]; then
--- a/scripts/build_python.sh
+++ b/scripts/build_python.sh
@@ -34,16 +34,27 @@ echo_bold_white() {
     echo -e "\033[1;37m$*\033[0m"
 }
 
+check_one_of() {
+    local value=$1
+    shift
+    for v in "$@"; do
+        if [ "$value" = "$v" ]; then
+            return 0
+        fi
+    done
+    return 1
+}
+
 CHIP_ROOT=$(_normpath "$(dirname "$0")/..")
 OUTPUT_ROOT="$CHIP_ROOT/out/python_lib"
 
 declare enable_ble=true
 declare chip_detail_logging=false
 declare chip_mdns
-declare case_retry_delta
 declare install_virtual_env
 declare clean_virtual_env=yes
-declare install_pytest_requirements=yes
+declare install_wheels=no
+declare install_pytest_requirements=unknown
 declare install_jupyterlab=no
 
 help() {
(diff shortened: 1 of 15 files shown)


andy31415 (changes requested):
@arkq I believe installing chip wheels in an activated package is a path for errors as the modules that we install may conflict with the source code itself and we had frequent bug reports from people saying "update master" did not work.

I would like to explicitly disallow:

```
source ./scripts/activate.sh
./scripts/build_python.sh -i
```

which this script seems to intend to allow.

andy31415:
> @arkq I believe installing chip wheels in an activated package is a path for errors as the modules that we install may conflict with the source code itself and we had frequent bug reports from people saying "update master" did not work.
> 
> I would like to explicitly disallow:
> 
> ```
> source ./scripts/activate.sh
> ./scripts/build_python.sh -i
> ```
> 
> which this script seems to intend to allow.

Maybe enfoce some extra  command ... not sure how but like `--i-know-this-may-break-my-builds` or something.

andy31415:
> 
> Maybe enfoce some extra command ... not sure how but like `--i-know-this-may-break-my-builds` or something.

Thinking of it more, I think we should allow "just reinstall wheels without clearing a virtualenv" however ensuring we do have a virtualenv path. Never allow to install in `.environment/$PW_ENVIRONMENT_ROOT`

scripts/build_python.sh
```diff
@@ -60,16 +71,16 @@ Input Options:
...
-                                                            Defaults to yes.
-  --include_pytest_deps  <yes|no>                           Install requirements.txt for running scripts/tests and
-                                                            src/python_testing scripts.
-                                                            Defaults to yes.
-  -j, --jupyter-lab                                         Install jupyterlab requirements.
-  --extra_packages PACKAGES                                 Install extra Python packages from PyPI
-  -z --pregen_dir DIRECTORY                                 Directory where generated zap files have been pre-generated.
+  -e, --create-virtual-env  <path>                          Create Python virtual environment in the specified path
+  -c, --clean-virtual-env   <yes|no>                        Clean the virtual environment if it exists; defaults to yes
+  -i, --install-wheels                                      Install generated wheels; this option is enabled by default
```
andy31415, line 76: This will create a conflict with people copy & pasting in history. It is inconvenient for people to remember what version they checked out for commands to work. How about keeping `-i/--install-virtual-env` and maybe `-w/--install-wheels` for the new option?

scripts/build_python.sh
```diff
@@ -206,20 +228,30 @@ if [ -n "$install_virtual_env" ]; then
...
+fi
 
-    if [ "$install_pytest_requirements" = "yes" ]; then
-        echo_blue "Installing python test dependencies ..."
-        "$ENVIRONMENT_ROOT"/bin/pip install -r "$CHIP_ROOT/scripts/tests/requirements.txt"
-        "$ENVIRONMENT_ROOT"/bin/pip install -r "$CHIP_ROOT/src/python_testing/requirements.txt"
-    fi
+if [ "$install_wheels" = "yes" ]; then
+    # Uninstall the existing wheels and install the new ones to make sure we
+    # have the latest versions even if the wheel version was not bumped.
```
andy31415, line 236: Please add a check that the virtualenv should not be the build virtual environment. Maybe we can also support some path somehow.

arkq:
> Please add a check that the virtualenv should not be the build virtual environment

The whole reason for this PR was to allow easy install of all CHIP python packages in the build environment :D I've been always using it that way without any issues. However, I understand that for non-developers such configuration might sometimes causes difficulties, especially for ones not familiar with python ecosystem. I will keep using my local scripts for that, so let's drop this change.

PR #36424, opened 2024-11-07
Fix android CHIPTest linkage after #36297 .

andy31415:
Tested locally that `android-arm64-chip-test` builds.


andy31415:
Hotfixing it in as this is only tested in master CI currently and I checked compilation locally.

Issue #36381, opened 2024-11-06
Remove the temporary ReportAttributeChanged function when the AAI/CHI is able to report the attribute changes.


wqx6:
See [comment](https://github.com/project-chip/connectedhomeip/pull/36364#pullrequestreview-2415975918)

We cannot notify the attribute change to the data model provider as the provider should own its data and versions. The workaround function should be replaced when the AAI/CHI is able to report the attribute changes.

PR #36425, opened 2024-11-07
Add the current provider to AttributeAccessInterface

andy31415:
This is currently done on Startup of the DataModel::Provider.

We do not yet have a set ordering for Startup, so we may have to iterate among `register attribute access interface` and usage of DataModel::Provider startup. However for now the basics should be there for addressing #36381

### DIFF
--- a/src/app/AttributeAccessInterface.h
+++ b/src/app/AttributeAccessInterface.h
@@ -18,11 +18,11 @@
 
 #pragma once
 
-#include <app/data-model-provider/ProviderChangeListener.h>
 #include <app/AttributeReportBuilder.h>
 #include <app/AttributeValueDecoder.h>
 #include <app/AttributeValueEncoder.h>
 #include <app/data-model-provider/Provider.h>
+#include <app/data-model-provider/ProviderChangeListener.h>
 #include <lib/core/CHIPError.h>
 
 /**
--- a/src/app/AttributeAccessInterface.h
+++ b/src/app/AttributeAccessInterface.h
@@ -18,7 +18,7 @@
 
 #pragma once
 
-#include "data-model-provider/ProviderChangeListener.h"
+#include <app/data-model-provider/ProviderChangeListener.h>
 #include <app/AttributeReportBuilder.h>
 #include <app/AttributeValueDecoder.h>
 #include <app/AttributeValueEncoder.h>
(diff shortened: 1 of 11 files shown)


src/app/AttributeAccessInterface.h
```diff
@@ -48,6 +50,36 @@ class AttributeAccessInterface
...
+     *
+     * @param [in] provider - use it for read/write/invoke and for GetCurrentContext
+     * @param [in] attributeChangeListener - use this to `MarkDirty` a specific parth that
+     *        this AttributeAccessInterfaces manages. The paths SHOULD match mEndpointID/mClusterId
+     *        for this AttributeAccessInterface
+     *
+     * Lifetime of provider and attributeChangeListener is valid until DetachProvider is called
+     * with the same provider value.
+     */
+    virtual void AttachToProvider(DataModel::Provider * provider, DataModel::ProviderChangeListener * attributeChangeListener) {}
```
bzbarsky-apple, line 73: I don't think this really scales once we have multiple providers around, including some that delegate some endpoints to one provider and some to another provider, possibly with the same cluster on both sets of endpoints using a single AttributeAccessInterface.

Put another way, this breaks as soon as you have a bridge that wants to use ZAP for its static features and a different provider for the dynamic bridge functionality.

I suspect the right way to do this is to pass the relevant context to the AttributeAccessInterface via whatever APIs are going to need to have that information eventually, not this binding setup.

bzbarsky-apple (changes requested):

src/app/AttributeAccessInterface.h
andy31415: Regarding multiple providers, I believe long term the providers should handle their attribute access interfaces or equivalent. Having a single `AttributeAccessInterfaceRegistry` and similar for commands is essentially a global singleton and that starts forcing singletons within our providers and that is too much coupling.

Since the InteractionModelEngine logic now does not directly call AAI/CHI anymore, I believe the providers should handle their AAI. If an application has multiple providers, those providers should manage their AAI.

andy31415:
Closing PR: after an offline discussion, another concern popping up was that we have dynamic AAI being added and then the order of `attach/detach` is not so obvious anymore. The dynamic part could also `attach` however that makes for awkward/more brittle code.

The approach that will be taken instead will be to have construction-time arguments for AAI that need provider or similar variables. Our ember-specific code can assume CodegenDataModelProvider and other applications can call different versions as needed in their main.

